PR #6261, opened 2021-12-03
Rewrite MauiComboBox tests

### LINKED ISSUE #3617: MAUI tests fail frequently and aren't easily diagnosable
These are an attempt to shim in the old internal tests (note that MAUI here is an old internal test framework). The old tests worked quite differently and don't fit in well with the current test infrastructure. We're getting [frequent intermittent failures](https://dev.azure.com/dnceng/public/_build/results?buildId=737706&view=ms.vss-test-web.build-test-results-tab&runId=22844842&resultId=100000&paneView=history) like so:

```
System.Windows.Forms.Maui.IntegrationTests.WinformsMauiButtonTests.MauiButtonTest(scenarioName: \"Hotkey_Fires_OnClick\")

Assert.Equal() Failure\r\n          ↓ (pos 0)\r\nExpected: Pass\r\nActual:   Fail\r\n          ↑ (pos 0)

Stack trace
   at System.Windows.Forms.Maui.IntegrationTests.MauiTestHelper.ValidateScenarioPassed(String projectName, String scenarioName) in /_/src/System.Windows.Forms/tests/IntegrationTests/System.Windows.Forms.Maui.IntegrationTests/MauiTestHelper.cs:line 90
   at System.Windows.Forms.Maui.IntegrationTests.WinformsMauiButtonTests.MauiButtonTest(String scenarioName) in /_/src/System.Windows.Forms/tests/IntegrationTests/System.Windows.Forms.Maui.IntegrationTests/WinformsMauiButtonTests.cs:line 35
``` 

This doesn't really help you much and it isn't obvious as to what you can do to move forward. In addition, the MAUI infrastructure sources aren't available so it is difficult to understand how the whole system works.

My recommendation (and current POR):

1. Disable the MAUI tests for now
2. Make them more xUnit-like and have any shim code in the repo
3. If we can make the tests plug in as a "normal" xUnit test then re-enable them for the CI, otherwise leave them for manual runs

cc: @AdamYoblick, @RussKie 

Failed with #3600

Danil-Andrianov:
Partially implements #3617

## Proposed changes

- Replaced ComboBoxMAUI tests into UI IntegrationTests project

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6261)


src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ComboBoxTests.cs
```diff
@@ -0,0 +1,411 @@
...
+                var szItems = CreateItems(NumItems);
+
+                Assert.True(AddItems(comboBox, szItems, NumItems));
+
+                //scroll items using the down arrow key
+                Assert.True(SelectItemWithKeyboard(comboBox, "{DOWN}", numKeyPresses));
+
+                Assert.Equal(_numEvents, numKeyPresses);
+
+                //scroll items using the up arrow key
```
vladimir-krestov, line 176: Remove blank lines between Aserts
```suggestion
                Assert.True(AddItems(comboBox, szItems, NumItems));
                //scroll items using the down arrow key
                Assert.True(SelectItemWithKeyboard(comboBox, "{DOWN}", numKeyPresses));
                Assert.Equal(_numEvents, numKeyPresses);
                //scroll items using the up arrow key
```

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/System.Windows.Forms.UI.IntegrationTests.csproj
```diff
@@ -22,6 +22,12 @@
     <ProjectReference Include="..\System.Windows.Forms.IntegrationTests.Common\System.Windows.Forms.IntegrationTests.Common.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Reference Include="WinformsTestLib">
+      <HintPath>..\MauiTests\lib\WinformsTestLib.dll</HintPath>
+    </Reference>
+  </ItemGroup>
```
vladimir-krestov, line 29: Why do we need it? Where did you get it?

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ComboBoxTests.cs
```diff
@@ -0,0 +1,411 @@
...
+{
+    public class ComboBoxTests : ControlTestBase
+    {
+        private const int NumItems = 200;
+        private RandomUtil _randomUtil;
+        private int _numEvents;
+        public ComboBoxTests(ITestOutputHelper testOutputHelper)
+            : base(testOutputHelper)
+        {
+            _numEvents = 0;
```
RussKie, line 20: This is redundant, value types are automatically initialised to 0.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ComboBoxTests.cs
```diff
@@ -0,0 +1,411 @@
...
+            {
+                InitializeItems(comboBox);
+
+                Assert.True(FindItem(comboBox, true));
+                return Task.CompletedTask;
+            });
+        }
+
+        [WinFormsFact]
+        public async Task Select_Item_By_KeyboardAsync()
```
RussKie, line 161: All other tests except this look like unit tests. Unless there is a good reason they must be moved (or removed if we already have tests for the matching functionality) to the unit tests project.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ComboBoxTests.cs
```diff
@@ -0,0 +1,411 @@
...
+        {
+            await RunComboBoxTestAsync((form, comboBox) =>
+            {
+                comboBox.SelectedIndexChanged += (x, y) => _numEvents++;
+                var numKeyPresses = _randomUtil.GetRange(1, comboBox.Items.Count);
+                var szItems = CreateItems(NumItems);
+
+                Assert.True(AddItems(comboBox, szItems, NumItems));
+
+                //scroll items using the down arrow key
```
RussKie, line 171: All comments must be in sentence case
```suggestion
                // Scroll items using the down arrow key
```

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ComboBoxTests.cs
```diff
@@ -0,0 +1,411 @@
...
+                {
+                    expectedMatchingIndex = i;
+                    break;
+                }
+            }
+
+            return (matchingIndex == expectedMatchingIndex);
+        }
+
+        private bool SelectItemWithKeyboard(ComboBox comboBox, string key, int numKeyPresses)
```
RussKie, line 319: https://github.com/dotnet/winforms/pull/6262#discussion_r762652712

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/System.Windows.Forms.UI.IntegrationTests.csproj
RussKie: No external dependencies, unless there's a clear justification.

RussKie (changes requested):

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
```diff
@@ -2074,6 +2074,332 @@ public void ComboBox_DroppedDownProperty_Set_False_Invoke_RaiseAutomationPropert
...
+        public void ComboBox_Remove_Items_By_Object()
+        {
+            using ComboBox comboBox = new()
+            {
+                Location = new Point(0, 0),
+            };
+            InitializeItems(comboBox, 10);
+
+            // Remove 5 items by object
+            bool result = RemoveItems(comboBox, false, 5);
```
SergeySmirnov-Akvelon, line 2124: It seems to me that this is a bad approach, when instead of `Assert` for a specific condition, we use a method that checks some conditions inside itself and returns only `true`/`false`.
Firstly, we cannot immediately understand which condition has not been met in the logs of CI build.  
Secondly, even locally, only debugging can help you understand which condition was not met.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
```diff
@@ -2074,6 +2074,332 @@ public void ComboBox_DroppedDownProperty_Set_False_Invoke_RaiseAutomationPropert
...
+
+            for (int i = 0; i < numItems; i++)
+            {
+                bool result = items[i].Equals(comboBox.Items[i]);
+                Assert.True(result);
+            }
+        }
+
+        [WinFormsFact]
+        public void ComboBox_SelectedIndex_Changes_SelectedItem_And_Fires_OnSelectedIndexChanged()
```
SergeySmirnov-Akvelon, line 2158: I think this test is too difficult to understand. In fact, we get a random `index` and set it as `SelectedIndex` property. If the `index` is already equal to the `SelectedIndex` property, then we consider that the event should not be triggered, otherwise it should.

Firstly, it is difficult to understand and not obvious from the name of the test.
Secondly, the condition when the `index` is already equal to the `SelectedIndex` property may not be met, since we are using random values.

I believe there should be two separate tests that will test both cases. We shouldn't have one test in which checking both cases depends on luck.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
```diff
@@ -2074,6 +2074,332 @@ public void ComboBox_DroppedDownProperty_Set_False_Invoke_RaiseAutomationPropert
...
+            var items = new string[numItems];
+            for (var i = 0; i < numItems; i++)
+            {
+                items[i] = "item" + i;
+            }
+
+            return items;
+        }
+
+        private bool AddItems(ComboBox c, string[] items, int numToAdd)
```
SergeySmirnov-Akvelon, line 2263: For me, this is also a very bad approach. When in the name of the method we see just adding items, but in fact we do a lot of manipulations and checks for different conditions. I would create a separate test that checks what is happening here if really necessary. In other tests, I would just add items without additional checks.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
```diff
@@ -2074,6 +2074,332 @@ public void ComboBox_DroppedDownProperty_Set_False_Invoke_RaiseAutomationPropert
...
+                if (!items[j].Equals(c.Items[j]))
+                {
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
+        private bool RemoveItems(ComboBox comboBox, bool isByIndex, int numToRemove)
```
SergeySmirnov-Akvelon, line 2293: I don't like this approach. As I said above, here we rely on random actions (random `index`) and make checks that are not available for checking in the logs. It is also possible that a random combination will indeed give some kind of error, but we will never be able to reproduce this, since this is a random actions. I would write a simple few tests to cover what is going on here.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
```diff
@@ -2074,6 +2074,332 @@ public void ComboBox_DroppedDownProperty_Set_False_Invoke_RaiseAutomationPropert
...
+                    {
+                        return false;
+                    }
+                }
+            }
+
+            return true;
+        }
+
+        private bool FindItem(ComboBox comboBox, bool isExactMatch)
```
SergeySmirnov-Akvelon, line 2327: I would also add some tests that check specific cases, rather than operate on random indices.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
```diff
@@ -2074,6 +2074,282 @@ public void ComboBox_DroppedDownProperty_Set_False_Invoke_RaiseAutomationPropert
...
+        [InlineData("item1984", 9, -1)]
+        public void ComboBox_FindString(string value, int index, int expected)
+        {
+            using ComboBox comboBox = new()
+            {
+                Location = new(0, 0),
+            };
+            InitializeItems(comboBox, 10);
+
+            int actual = comboBox.FindString(value, index);
```
dreddy-work, line 2237: Given that `FindString `returns insensitive substring matching,  I would add `inlinedata` for substring match aswell.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
```diff
@@ -2074,6 +2074,282 @@ public void ComboBox_DroppedDownProperty_Set_False_Invoke_RaiseAutomationPropert
...
+        {
+            for (int i = 0; i < numItems; i++)
+            {
+                comboBox.Items.Add("item" + i);
+            }
+
+            Assert.Equal(numItems, comboBox.Items.Count);
+        }
+
+        private int GetRange(int nMin, int nMax)
```
dreddy-work, line 2306: This is a private method and called once. Necessary?

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ComboBoxTests.cs
```diff
@@ -0,0 +1,97 @@
...
+
+using System.Drawing;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace System.Windows.Forms.UITests
+{
+    public class ComboBoxTests : ControlTestBase
+    {
+        private int _numEvents;
```
SergeySmirnov-Akvelon, line 13: Please add blank line

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ComboBoxTests.cs
```diff
@@ -0,0 +1,97 @@
...
+
+                // Scroll items using the up arrow key
+                for (int i = 0; i < numKeyPresses; i++)
+                {
+                    await InputSimulator.SendAsync(
+                        form,
+                        inputSimulator => inputSimulator.Keyboard.KeyPress(WindowsInput.Native.VirtualKeyCode.UP));
+                }
+
+                Assert.Equal(numKeyPresses, _numEvents);
```
SergeySmirnov-Akvelon, line 39:
```suggestion
                comboBox.SelectedIndexChanged += (x, y) => _numEvents++;
                AddItems(comboBox, 6);
                int numKeyPresses = 5;

                comboBox.SelectedIndex = comboBox.Items.Count - 1;
                _numEvents = 0;

                // Scroll items using the up arrow key
                for (int i = 0; i < numKeyPresses; i++)
                {
                    await InputSimulator.SendAsync(
                        form,
                        inputSimulator => inputSimulator.Keyboard.KeyPress(WindowsInput.Native.VirtualKeyCode.UP));
                }

                Assert.Equal(numKeyPresses, _numEvents);
```

Despite the fact that the test passes, it has two incorrect places that compensate for each other.

First, we execute `comboBox.SelectedIndex = comboBox.Items.Count - 1;` code that changes the value of the `_numEvents` property even before we get into the check loop.

In the cycle itself, we press the `Up` button 5 times, but we are already on the 5th element. In 4 clicks we reach the first element and the last (fifth) pressing of the `Up` button does not affect anything.

Thus, we have a check loop that works as expected only 4 out of 5 times and a change in the value outside the check loop, which leads to the correct, but not expected result.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
```diff
@@ -2074,6 +2074,282 @@ public void ComboBox_DroppedDownProperty_Set_False_Invoke_RaiseAutomationPropert
...
+            {
+                Location = new Point(0, 0),
+            };
+            InitializeItems(comboBox, 10);
+
+            // Add 5 items by index
+            for (var i = 0; i < 5; i++)
+            {
+                var count = comboBox.Items.Count;
+                var index = GetRange(0, count - 1);
```
SergeySmirnov-Akvelon, line 2090:
```suggestion
                var index = GetRange(0, comboBox.Items.Count - 1);
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
```diff
@@ -2074,6 +2074,282 @@ public void ComboBox_DroppedDownProperty_Set_False_Invoke_RaiseAutomationPropert
...
+        [WinFormsTheory]
+        [InlineData(10, 0, 5)]
+        [InlineData(10, 2, 4)]
+        public void ComboBox_Remove_Items_By_Index(int numberOfItems, int index, int number)
+        {
+            using ComboBox comboBox = new()
+            {
+                Location = new(0, 0),
+            };
+            InitializeItems(comboBox, numberOfItems);
```
SergeySmirnov-Akvelon, line 2108: Please add blank line

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
```diff
@@ -2074,6 +2074,282 @@ public void ComboBox_DroppedDownProperty_Set_False_Invoke_RaiseAutomationPropert
...
+        [WinFormsTheory]
+        [InlineData(10, 0, 5)]
+        [InlineData(10, 2, 4)]
+        public void ComboBox_Remove_Items_By_Object(int numberOfItems, int index, int number)
+        {
+            using ComboBox comboBox = new()
+            {
+                Location = new(0, 0),
+            };
+            InitializeItems(comboBox, numberOfItems);
```
SergeySmirnov-Akvelon, line 2128: Please add blank line

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
```diff
@@ -2074,6 +2074,282 @@ public void ComboBox_DroppedDownProperty_Set_False_Invoke_RaiseAutomationPropert
...
+        }
+
+        [WinFormsFact]
+        public void ComboBox_Sorted_Sorts_By_Ascending()
+        {
+            using ComboBox comboBox = new()
+            {
+                Location = new Point(0, 0),
+            };
+            int numItems = 200;
```
SergeySmirnov-Akvelon, line 2146: Please add blank line

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
```diff
@@ -2074,6 +2074,282 @@ public void ComboBox_DroppedDownProperty_Set_False_Invoke_RaiseAutomationPropert
...
+                comboBox.Items.Add(item);
+            }
+
+            comboBox.Sorted = true;
+            Array.Sort(items);
+
+            for (int i = 0; i < numItems; i++)
+            {
+                bool result = items[i].Equals(comboBox.Items[i]);
+                Assert.True(result);
```
SergeySmirnov-Akvelon, line 2163:
```suggestion
                Assert.True(items[i].Equals(comboBox.Items[i]));
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
```diff
@@ -2074,6 +2074,282 @@ public void ComboBox_DroppedDownProperty_Set_False_Invoke_RaiseAutomationPropert
...
+        [WinFormsTheory]
+        [InlineData(1)]
+        [InlineData(5)]
+        public void ComboBox_SelectedIndexDoesNotChangeSelectedItemAndDoesNotFireOnSelectedIndexChanged(int index)
+        {
+            using ComboBox comboBox = new()
+            {
+                Location = new(0, 0),
+            };
+            InitializeItems(comboBox, 10);
```
SergeySmirnov-Akvelon, line 2176: Please add blank line

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
```diff
@@ -2074,6 +2074,282 @@ public void ComboBox_DroppedDownProperty_Set_False_Invoke_RaiseAutomationPropert
...
+
+            int numEvents = 0;
+            comboBox.SelectedIndex = index;
+            comboBox.SelectedIndexChanged += (x, y) => numEvents++;
+            comboBox.Sorted = false;
+            int numEventsExpected = 0;
+            numEvents = 0;
+            int selectedIndex = comboBox.SelectedIndex;
+            comboBox.SelectedIndex = index;
```
SergeySmirnov-Akvelon, line 2185:
```suggestion
            comboBox.SelectedIndex = index;
            int numEvents = 0;
            comboBox.SelectedIndexChanged += (x, y) => numEvents++;
            comboBox.Sorted = false;
            int selectedIndex = comboBox.SelectedIndex;

            comboBox.SelectedIndex = index;
```

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ComboBoxTests.cs
Danil-Andrianov: That's tricky moment. Thanks for explanation.

Tanya-Solyanik: >First, we execute comboBox.SelectedIndex = comboBox.Items.Count - 1; code that changes the value of the _numEvents property even before we get into the check loop.

Please add a comment about it to the code.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
sharwell: Any reason to not use `Assert.Equal` instead?

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ComboBoxTests.cs
```diff
@@ -0,0 +1,98 @@
...
+
+using System.Drawing;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace System.Windows.Forms.UITests
+{
+    public class ComboBoxTests : ControlTestBase
+    {
+        private int _numEvents;
```
RussKie, line 13: Please remove the field, and use local variables where appropriate/required.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
```diff
@@ -2098,6 +2098,247 @@ public void ComboBox_DroppedDownProperty_Set_False_Invoke_RaiseAutomationPropert
...
+            InitializeItems(comboBox, 10);
+
+            // Add 5 items by index
+            for (var i = 0; i < 5; i++)
+            {
+                var count = comboBox.Items.Count;
+                var index = rand.Next(0, count - 1);
+                var item = "new item" + i;
+                comboBox.Items.Insert(index, item);
+                bool result = item.Equals(comboBox.Items[index]) && comboBox.Items.Count == ++count;
```
RussKie, line 2118: Please don't do this, this is very difficult to read.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
```diff
@@ -2098,6 +2098,247 @@ public void ComboBox_DroppedDownProperty_Set_False_Invoke_RaiseAutomationPropert
...
+
+                Assert.True(result);
+            }
+        }
+
+        [WinFormsTheory]
+        [InlineData(10, 0, 5)]
+        [InlineData(10, 2, 4)]
+        public void ComboBox_Remove_Items_By_Index(int numberOfItems, int index, int number)
```
RussKie, line 2127:
```suggestion
        public void ComboBox_Remove_Items_By_Index(int numberOfItems, int index, int numberOfItemsToRemove)
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
```diff
@@ -2098,6 +2098,247 @@ public void ComboBox_DroppedDownProperty_Set_False_Invoke_RaiseAutomationPropert
...
+                Location = new Point(0, 0),
+            };
+            int numItems = 200;
+
+            var items = new string[numItems];
+
+            // All the items are between a - z
+            for (var j = 0; j < numItems; j++)
+            {
+                var item = (GetInt(true) % 26 + 97).ToString();
```
RussKie, line 2178: `GetInt` method is not needed. it can be simplified and inlined.
Also please avoid magic numbers (not everyone remembers ASCII table by heart 😃), i.e. `97 -> (char)'a'`

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
```diff
@@ -2098,6 +2098,247 @@ public void ComboBox_DroppedDownProperty_Set_False_Invoke_RaiseAutomationPropert
...
+            for (int i = 0; i < 3; i++)
+            {
+                MeasureItemEventArgs e = derivedComboBox.MeasureItemEventArgs[i];
+                Assert.NotNull(e.Graphics);
+
+                Assert.Equal(e.Index, i);
+
+                Assert.Equal(18, e.ItemHeight);
+
+                Assert.Equal(0, e.ItemWidth);
```
RussKie, line 2300: Keep similar actions together
```suggestion
                Assert.NotNull(e.Graphics);
                Assert.Equal(e.Index, i);
                Assert.Equal(18, e.ItemHeight);
                Assert.Equal(0, e.ItemWidth);
```

SergeySmirnov-Akvelon:
azp /run

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ComboBoxTests.cs
```diff
@@ -0,0 +1,108 @@
...
+            {
+                int numEvents = 0;
+                comboBox.SelectedIndexChanged += (x, y) => numEvents++;
+                AddItems(comboBox, 10);
+                int numKeyPresses = 9;
+
+                // SelectedIndex should be initialized before numEvents because of SelectedIndexChanged is raised
+                // and affects on the numEvents.
+                comboBox.SelectedIndex = 0;
+                numEvents = 0;
```
RussKie, line 31: We already initialised this earlier.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ComboBoxTests.cs
```diff
@@ -0,0 +1,108 @@
...
+                form.Show();
+                form.Focus();
```
RussKie, line 33: The dialog is already shown and is focused.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ComboBoxTests.cs
```diff
@@ -0,0 +1,108 @@
...
+                Application.DoEvents();
```
RussKie, line 34: Why is this needed?

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ComboBoxTests.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        [WinFormsFact]
+        public async Task ComboBox_Select_Item_By_DownArrowKeyAsync()
+        {
+            await RunComboBoxTestAsync(async (form, comboBox) =>
+            {
+                int numEvents = 0;
+                comboBox.SelectedIndexChanged += (x, y) => numEvents++;
+                AddItems(comboBox, 10);
+                int numKeyPresses = 9;
```
RussKie, line 26: Move it down where it is used.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ComboBoxTests.cs
```diff
@@ -0,0 +1,108 @@
...
+        {
+        }
+
+        [WinFormsFact]
+        public async Task ComboBox_Select_Item_By_DownArrowKeyAsync()
+        {
+            await RunComboBoxTestAsync(async (form, comboBox) =>
+            {
+                int numEvents = 0;
+                comboBox.SelectedIndexChanged += (x, y) => numEvents++;
```
RussKie, line 24: This can be wired after the combobox is fully initialised, i.e. after `AddItems` is called.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
```diff
@@ -2098,6 +2098,233 @@ public void ComboBox_DroppedDownProperty_Set_False_Invoke_RaiseAutomationPropert
             Assert.Equal(2, ((CustomComboBoxAccessibleObject)comboBox.AccessibilityObject).RaiseAutomationCallCount);
         }
 
+        [WinFormsFact]
+        public void ComboBox_Add_Items_By_Index()
+        {
+            using ComboBox comboBox = new()
+            {
+                Location = new Point(0, 0),
+            };
```
RussKie, line 2107: Why do we need `Location` in the unit tests?

RussKie (changes requested):

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ComboBoxTests.cs
Danil-Andrianov: Yes, but after line 30 `numEvents` is equal 1 and we should nullify it.

RussKie: What is the reason to init it before line:30?

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ComboBoxTests.cs
```diff
@@ -0,0 +1,98 @@
...
+            await RunComboBoxTestAsync(async (form, comboBox) =>
+            {
+                AddItems(comboBox, 10);
+                int numEvents = 0;
+                comboBox.SelectedIndexChanged += (x, y) => numEvents++;
+
+                // SelectedIndex should be initialized before numEvents because of SelectedIndexChanged is raised
+                // and affects on the numEvents.
+                comboBox.SelectedIndex = comboBox.Items.Count - 1;
+                numEvents = 0;
```
RussKie, line 29:
```suggestion
                AddItems(comboBox, 10);
                comboBox.SelectedIndex = comboBox.Items.Count - 1;

                int numEvents = 0;
                comboBox.SelectedIndexChanged += (x, y) => numEvents++;
```

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ComboBoxTests.cs
```diff
@@ -0,0 +1,92 @@
...
+        private async Task RunComboBoxTestAsync(Func<Form, ComboBox, Task> runTest)
+        {
+            await RunSingleControlTestAsync(
+                testDriverAsync: runTest,
+                createControl: () =>
+                {
+                    ComboBox control = new();
+
+                    return control;
+                },
```
RussKie, line 82: I'm pretty certain this would work:

```suggestion
        private async Task RunComboBoxTestAsync(Func<Form, ComboBox, Task> runTest, int numberOfItemsToAdd, int selectedIndex)
        {
            await RunSingleControlTestAsync(
                testDriverAsync: runTest,
                createControl: () =>
                {
                    ComboBox control = new();
                    AddItems(comboBox, numberOfItemsToAdd);
                    comboBox.SelectedIndex = selectedIndex;

                    return control;
                },
```

```cs
        [WinFormsFact]
        public async Task ComboBox_Select_Item_By_DownArrowKeyAsync()
        {
            await RunComboBoxTestAsync(async (form, comboBox) =>
            {
                int numEvents = 0;
                comboBox.SelectedIndexChanged += (x, y) => numEvents++;

                // ...
            },
            numberOfItemsToAdd: 10,
            selectedIndex: 0);
        }
```

RussKie (commented):
I see several feedback comments thumbed up but not corresponding changes. Have those changes been pushed?

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ComboBoxTests.cs
Danil-Andrianov: Sorry, I forgot about one amendment.

RussKie:
Some tests are failing fairly consistently. Setting as draft until the failures are investigated.

PR #6275, opened 2021-12-04
Enable nullability in FileDialog

gpetrou:
## Proposed changes

- Enable nullability in FileDialog

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6275)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.Vista.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.Vista.cs
@@ -224,7 +224,7 @@ namespace System.Windows.Forms
             }
         }
 
-        private static COMDLG_FILTERSPEC[] GetFilterItems(string filter)
+        private static COMDLG_FILTERSPEC[] GetFilterItems(string? filter)
         {
             // Expected input types
             // "Text files (*.txt)|*.txt|All files (*.*)|*.*"
--- a/src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.Vista.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.Vista.cs
@@ -89,7 +89,7 @@ namespace System.Windows.Forms
                 }
             }
 
-            dialog.SetTitle(Title!);
+            dialog.SetTitle(Title);
             dialog.SetOptions(GetOptions());
             SetFileTypes(dialog);
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.Vista.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.Vista.cs
@@ -89,7 +89,7 @@ namespace System.Windows.Forms
                 }
             }
 
-            dialog.SetTitle(Title);
+            dialog.SetTitle(Title!);
             dialog.SetOptions(GetOptions());
             SetFileTypes(dialog);
 
@@ -144,7 +144,7 @@ namespace System.Windows.Forms
         {
             int saveOptions = _options;
             int saveFilterIndex = FilterIndex;
-            string[] saveFileNames = _fileNames;
+            string[]? saveFileNames = _fileNames;
             bool ok = false;
 
             try
--- a/src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.cs
@@ -2,10 +2,9 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.ComponentModel;
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using System.Drawing.Design;
 using System.Runtime.InteropServices;
 using static Interop;
@@ -27,14 +26,14 @@ namespace System.Windows.Forms
 
         private protected int _options;
 
-        private string _title;
-        private string _initialDirectory;
-        private string _defaultExt;
-        private string[] _fileNames;
-        private string _filter;
+        private string? _title;
+        private string? _initialDirectory;
+        private string? _defaultExt;
+        private string[]? _fileNames;
+        private string? _filter;
         private bool _ignoreSecondFileOkNotification;
         private int _okNotificationCount;
-        private UnicodeCharBuffer _charBuffer;
+        private UnicodeCharBuffer? _charBuffer;
         private IntPtr _dialogHWnd;
 
         /// <summary>
@@ -122,6 +121,7 @@ namespace System.Windows.Forms
         [SRCategory(nameof(SR.CatBehavior))]
         [DefaultValue("")]
         [SRDescription(nameof(SR.FDdefaultExtDescr))]
+        [AllowNull]
         public string DefaultExt
         {
             get => _defaultExt ?? string.Empty;
@@ -135,7 +135,7 @@ namespace System.Windows.Forms
                     }
                     else if (value.Length == 0)
                     {
-                        value = null;
+                        value = null!;
                     }
                 }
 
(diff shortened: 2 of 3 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.cs
```diff
@@ -135,7 +135,7 @@ public string DefaultExt
                     }
                     else if (value.Length == 0)
                     {
-                        value = null;
+                        value = null!;
```
dreddy-work, line 138: we could avoid this by refactoring this code.

src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.cs
```diff
@@ -534,11 +539,11 @@ protected unsafe override IntPtr HookProc(IntPtr hWnd, int msg, IntPtr wparam, I
                             MoveToScreenCenter(_dialogHWnd);
                             break;
                         case -602: /* CDN_SELCHANGE */
-                            NativeMethods.OPENFILENAME_I ofn = Marshal.PtrToStructure<NativeMethods.OPENFILENAME_I>(notify->lpOFN);
+                            NativeMethods.OPENFILENAME_I? ofn = Marshal.PtrToStructure<NativeMethods.OPENFILENAME_I>(notify->lpOFN);
 
                             // Get the buffer size required to store the selected file names.
                             int sizeNeeded = (int)User32.SendMessageW(_dialogHWnd, (User32.WM)1124 /*CDM_GETSPEC*/);
-                            if (sizeNeeded > ofn.nMaxFile)
+                            if (ofn is not null && sizeNeeded > ofn.nMaxFile)
```
dreddy-work, line 546: How are you deciding on choosing between null forgiving `!` vs `null check` ?

src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.Vista.cs
```diff
@@ -89,7 +89,7 @@ private void OnBeforeVistaDialog(IFileDialog dialog)
                 }
             }
 
-            dialog.SetTitle(Title);
+            dialog.SetTitle(Title!);
```
RussKie, line 92: This feels wrong.
How can it be null?

src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.cs
RussKie: If we the compiler doesn't accept `value` as nullable, I think we should restructure the assignment to avoid the bang altogether.

src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.cs
```diff
@@ -404,17 +409,17 @@ public bool ValidateNames
         /// </summary>
         private bool DoFileOk(IntPtr lpOFN)
         {
-            NativeMethods.OPENFILENAME_I ofn = Marshal.PtrToStructure<NativeMethods.OPENFILENAME_I>(lpOFN);
+            NativeMethods.OPENFILENAME_I? ofn = Marshal.PtrToStructure<NativeMethods.OPENFILENAME_I>(lpOFN);
```
RussKie, line 412: This can't be null.
Also we should replace `Marshal` with the direct unsafe cast, like we do in other places. Something like
```suggestion
            NativeMethods.OPENFILENAME_I* ofn = (NativeMethods.OPENFILENAME_I*)lpOFN;
```

I'm on a phone so can't give a better suggestion.

src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.cs
```diff
@@ -662,7 +667,7 @@ private bool ProcessFileNames()
             if ((_options & (int)Comdlg32.OFN.NOVALIDATE) == 0)
             {
                 string[] extensions = FilterExtensions;
-                for (int i = 0; i < _fileNames.Length; i++)
+                for (int i = 0; i < _fileNames!.Length; i++)
```
RussKie, line 670: Use the property instead of the field here

src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.cs
gpetrou: Sorry, not sure what you mean here. Can you elaborate?

RussKie:
```cs
            set
            {
                string? defaultExt = value;
                if (defaultExt is not null)
                {
                    if (defaultExt.StartsWith("."))
                    {
                        defaultExt = defaultExt.Substring(1);
                    }
                    else if (defaultExt.Length == 0)
                    {
                        defaultExt = null;
                    }
                }

                _defaultExt = defaultExt;
            }
```

src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.cs
RussKie: See https://github.com/dotnet/winforms/pull/6275/files#r776884216

dreddy-work:
@gpetrou , please rebase after addressing last feedback here.

src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.cs
gpetrou: The property is using `Clone`. I assume I shouldn't do that and keep what we have?

src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.cs
gpetrou: I am observing a few errors when I do this change.

src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.cs
```diff
@@ -122,24 +121,26 @@ public bool CheckPathExists
...
         [DefaultValue("")]
         [SRDescription(nameof(SR.FDdefaultExtDescr))]
+        [AllowNull]
         public string DefaultExt
         {
             get => _defaultExt ?? string.Empty;
             set
             {
-                if (value is not null)
+                string? defaultExt = value;
```
dreddy-work, line 130: Why you needed this local variable here?

gpetrou: This is based on https://github.com/dotnet/winforms/pull/6275#discussion_r795321173

dreddy-work:
@RussKie , can take a final look here and help merge?

RussKie:
I'll have a look next week.

src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.cs
RussKie: Yes, the [`OPENFILENAMEW` struct](https://docs.microsoft.com/windows/win32/api/commdlg/ns-commdlg-openfilenamew) is currently declared as a class, and it needs to be converted to a blittable equivalent.
This definitely falls beyond the scope of this PR.

src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.cs
RussKie: Thank you for checking.
In this case we need to the same null check as everywhere else when dereferencing this field.

RussKie:
Other than https://github.com/dotnet/winforms/pull/6275#discussion_r776884308, looks good.

src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.cs
gpetrou: When `ProcessFileNames` is called, we seem to always have a non-null _fileNames. So, perhaps the null check is not needed in this case?

RussKie: If it's not obvious, it's better to be safe than sorry.

dreddy-work (approved):
LGTM.

PR #6277, opened 2021-12-04
Enable nullability in ControlCollection

gpetrou:
## Proposed changes

- Enable nullability in ControlCollection.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6277)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlCollection.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.Collections;
 using System.ComponentModel;
 using System.Diagnostics;
@@ -34,7 +32,7 @@ namespace System.Windows.Forms
             /// <summary>
             ///  Returns true if the collection contains an item with the specified key, false otherwise.
             /// </summary>
-            public virtual bool ContainsKey(string key)
+            public virtual bool ContainsKey(string? key)
             {
                 return IsValidIndex(IndexOfKey(key));
             }
@@ -44,7 +42,7 @@ namespace System.Windows.Forms
             ///  the child control list. If the control is already a child of another control it
             ///  is first removed from that control.
             /// </summary>
-            public virtual void Add(Control value)
+            public virtual void Add(Control? value)
             {
                 if (value is null)
                 {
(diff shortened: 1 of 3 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlCollection.cs
```diff
@@ -204,7 +202,10 @@ public bool Contains(Control control)
             /// </summary>
             public Control[] Find(string key, bool searchAllChildren)
             {
-                key.ThrowIfNullOrEmptyWithMessage(SR.FindKeyMayNotBeEmptyOrNull);
+                if (string.IsNullOrEmpty(key))
+                {
+                    throw new ArgumentNullException(nameof(key), SR.FindKeyMayNotBeEmptyOrNull);
+                }
```
RussKie, line 208:
```suggestion
                key.ThrowIfNullOrEmptyWithMessage(SR.FindKeyMayNotBeEmptyOrNull);
```

RussKie (commented):
👍

src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlCollection.cs
```diff
@@ -385,7 +383,7 @@ public virtual void RemoveByKey(string key)
                             string.Format(SR.IndexOutOfRange, index.ToString(CultureInfo.CurrentCulture)));
```
dreddy-work, line 383: what about method return type here?

gpetrou: Sorry, what do you mean? Are you suggesting changing to `public new virtual Control? this[int index]` or are you talking about something else?

dreddy-work: that was it. Did you leave it out intentionally?

gpetrou: I noticed that there is a `Debug.Assert` call, so I thought that this should not be null.

dreddy-work: But the control being returning is already nullable at line 386. In general use, we shouldn't be expecting a null and given that we are asserting, maybe we should go with null forgiving operator on line 386.

dreddy-work (approved):

PR #6281, opened 2021-12-04
Enable nullability in Com2Enum

gpetrou:
## Proposed changes

- Enable nullability in Com2Enum

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6281)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ComponentModel/COM2Interop/COM2Enum.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ComponentModel/COM2Interop/COM2Enum.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.Globalization;
 
 namespace System.Windows.Forms.ComponentModel.Com2Interop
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ComponentModel/COM2Interop/COM2Enum.cs
```diff
@@ -113,7 +111,7 @@ public virtual string ToString(object value)
 
             // We have to do this to compensate for small discrepancies in a lot of objects in COM2
             // (DWORD -> VT_IU4, value we get is VT_I4, which convert to Int32, UInt32 respectively)
-            string stringValue = value.ToString();
+            string? stringValue = value?.ToString();
```
RussKie, line 114: I don't think this correctly expresses the intent - i.e. an enum is not a nullable structure. Whilst we're operating with objects here, we never expect a `null` as a result. In an unlikely event when we receive a `null`, we should just return
```suggestion
            if (value?.ToString() is not string stringValue)
            {
                return string.Empty;
            }
```

src/System.Windows.Forms/src/System/Windows/Forms/ComponentModel/COM2Interop/COM2Enum.cs
```diff
@@ -92,7 +90,7 @@ protected void PopulateArrays(string[] names, object[] values)
         /// <summary>
         ///  Retrieves the string name of a given value.
         /// </summary>
-        public virtual string ToString(object value)
+        public virtual string? ToString(object? value)
```
RussKie, line 93:
```suggestion
        public virtual string ToString(object? value)
```

RussKie (changes requested):

RussKie (approved):

PR #6284, opened 2021-12-04
Enable nullability in DataFormats

gpetrou:
## Proposed changes

- Enable nullability in DataFormats.


src/System.Windows.Forms/src/System/Windows/Forms/DataFormats.cs
```diff
@@ -139,15 +137,15 @@ public static partial class DataFormats
...
+        public static readonly string StringFormat = typeof(string).FullName!;
 
         /// <summary>
         ///  Specifies a format that encapsulates any type of WinForms object.
         ///  This <see langword='static'/> field is read-only.
         /// </summary>
         public static readonly string Serializable = Application.WindowsFormsVersion + "PersistentObject";
 
-        private static Format[] s_formatList;
+        private static Format[]? s_formatList;
```
RussKie, line 148: Perhaps we could do a little better than sprinkle `s_formatList!` all over the file. `EnsureFormatSpace()` and `EnsurePredefined()` must be annotated with `MemberNotNull(nameof(s_formatList)` - this should address all of warnings. If not, let's see what else we can do.

RussKie (changes requested):

src/System.Windows.Forms/src/System/Windows/Forms/DataFormats.cs
gpetrou: There is one new warning in such case. Also, one `s_formatList!` is still needed.

src/System.Windows.Forms/src/System/Windows/Forms/DataFormats.cs
```diff
@@ -247,7 +247,7 @@ private static void EnsureFormatSpace(int size)
                 Format[] newList = new Format[newSize];
                 for (int n = 0; n < s_formatCount; n++)
                 {
-                    newList[n] = s_formatList[n];
+                    newList[n] = s_formatList![n];
```
RussKie, line 250: This looks like a bug - `s_formatList` can be `null` here:
https://github.com/dotnet/winforms/blob/d46ad2e8dc76248739d9ae22b28b399a6a6b299e/src/System.Windows.Forms/src/System/Windows/Forms/DataFormats.cs#L243-L253

Looks like there's a missing null check before the `for` loop.

dreddy-work:
@gpetrou ,  please check unit tests that are failing here as well.

src/System.Windows.Forms/src/System/Windows/Forms/DataFormats.cs
Tanya-Solyanik: If `s_formatList` is null, then `s_formatCount` is zero, so `!` looks reasonable in this spot.

dreddy-work (dismissed):

src/System.Windows.Forms/src/System/Windows/Forms/DataFormats.cs
dreddy-work: @RussKie , i am little confused on the use of `MemberNotNull` and then checking for null in this PR. Please take a final look here and help merge this.

dreddy-work:
ping @RussKie to unblock this PR.

src/System.Windows.Forms/src/System/Windows/Forms/DataFormats.cs
RussKie: @dreddy-work when the method returns `s_formatList` is guaranteed to be not null. Either because it was already initialised, or because it gets initialised from `newList`. Does this make sense?

RussKie (approved):

PR #6288, opened 2021-12-05
Enable nullability in ControlPaint

gpetrou:
## Proposed changes

- Enable nullability in ControlPaint

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6288)

PR #6289, opened 2021-12-05
Enable nullability in ButtonRenderer

gpetrou:
## Proposed changes

- Enable nullability in ButtonRenderer

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6289)


src/System.Windows.Forms/src/System/Windows/Forms/ButtonRenderer.cs
```diff
@@ -83,8 +82,8 @@ internal static void DrawButton(IDeviceContext deviceContext, Rectangle bounds,
             }
             else
             {
-                Graphics graphics = deviceContext.TryGetGraphics(create: true);
-                ControlPaint.DrawButton(graphics, bounds, ConvertToButtonState(state));
+                Graphics? graphics = deviceContext.TryGetGraphics(create: true);
+                ControlPaint.DrawButton(graphics!, bounds, ConvertToButtonState(state));
```
RussKie, line 86: This doesn't look correct: theoretically `graphics` can be a `null`, and we can't assume it isn't on the following line. Please add a condition like you did in https://github.com/dotnet/winforms/pull/6288.

RussKie (changes requested):

RussKie (approved):

PR #639, opened 2019-03-27
Porting ToolStrip PerMonitorV2 HighDpi changes from classic framework

KlausLoeffelmann:
fixes #215

This PR is a port of the basic ToolStrip PerMonitorV2 functionality from the classic framework, but without taking the HighDpi Quirks from the classic runtime into account. It addresses those ToolStrip HighDpi issues which were needed at the time to make Visual Studio render ToolStrip based items in PMV2 correctly.

Basically, the problem has to be solved that if a form is moved from one monitor to another with a different pixel density, the size of the form and also the size of all controls placed on it have to be adapted to the new resolution. 

In order to get the right point in time within the normally excluded control classes when the DPI change (the transition from one monitor to another) occurs, we enable ToolStrip itself in a way other controls/components can register at the ToolStrip to receive a kind of Rescale event. This is done with the new field variable rescaleConstsCallbackDelegate. Because ToolStrip itself is placed on a form or container, its RescaleConstantsForDPI is called accordingly, so it performs the Invoke based on this delegate, triggering all listening components. However, this event must be initially wired for all of the ToolStripItem-based components it is hosting, and this is done in ToolStripItem's SetOwner method.

This means that ToolStrip, all controls derived from ToolStrip and also all components derived from ToolStripItem are triggered when the DPI changes and the required size adjustments are due. A problem remains with the ToolStripDropDown containers, which host only ToolStripDropDownItem-based components, but which again are derived from ToolStripItem. The latter enables us to intervene in the rescaling event, determine if the affected ToolStripItem "has" a DropDown , and then make any necessary rescaling here (and in any subordinate) as well. This happens in the ProcessDropDownItems method of the ToolStripDropDownItem class, which in turn is called by ToolStrip_RescaleConstants (and which in turn is triggered by the new delegate in ToolStrip and the wiring of ToolStripItem).

Code changes in all remaining classes then serve as far as possible to make the actual size adjustments for the new DPI value.

PR #6302, opened 2021-12-06
Enable nullability in ToolStripRenderer

gpetrou:
## Proposed changes

- Enable nullability in ToolStripRenderer

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6302)


RussKie:
MC

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripRenderer.cs
```diff
@@ -36,10 +34,10 @@ public abstract class ToolStripRenderer
 
         private static readonly object EventRenderStatusStripSizingGrip = new object();
 
-        private static ColorMatrix disabledImageColorMatrix;
+        private static ColorMatrix? _disabledImageColorMatrix;
```
RussKie, line 37:
```suggestion
        private static ColorMatrix? s_disabledImageColorMatrix;
```

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripRenderer.cs
```diff
@@ -628,12 +626,12 @@ protected virtual void OnRenderArrow(ToolStripArrowRenderEventArgs e)
...
                 //middle.X += (dropDownRect.Width % 2);
 
-                Point[] arrow = null;
+                Point[]? arrow = null;
 
                 // We need the Elvis operator here, since at design time at this point Item can be null.
                 if (e.Item?.DeviceDpi != previousDeviceDpi && DpiHelper.IsPerMonitorV2Awareness)
                 {
-                    previousDeviceDpi = e.Item.DeviceDpi;
+                    previousDeviceDpi = e.Item!.DeviceDpi;
```
RussKie, line 634: This looks like a bug introduced in https://github.com/dotnet/winforms/pull/639 - if `e.Item == null` we'll end up with a NRE here.
I think line:632 must look as
```cs
                if (e.Item is not null && e.Item.DeviceDpi != previousDeviceDpi && DpiHelper.IsPerMonitorV2Awareness)
```

/cc: @KlausLoeffelmann

KlausLoeffelmann: Sadly, yes,

RussKie (approved):

PR #6309, opened 2021-12-07
Improving accessible object for label edit in ListView

### LINKED ISSUE #4810: [Accessibility]Inspect: There is an additional “”edit node in Inspect when ListView item in edit Status
* .NET Core Version: 
.Net 6.0.100-preview.4.21218.5 from main branch

Yes

**More info:**
1.	In .NET Framework 4.7.1, it has the same result.
2.	This issue can repro no matter enable virtual mode or disable virtual mode.

**Minimal repro:**
1.	Extract the “noVirtual_Listview_detail.zip” and open it in VS.
[noVirtual_Listview_detail.zip](https://github.com/dotnet/winforms/files/6334696/noVirtual_Listview_detail.zip)
2.	Build and run this application.
3.	Start Inspect and focus on one ListView item(e.g. item1), observe the structure in Inspect.
4.	Make ListView item in edit status and focus it, observe the structure in Inspect.

**Actual behavior:**
After step3, it focus on “item1” edit node in Inspect.
step3
After step4, there is an additional “”edit node in Inspect when ListView item in edit Status.
step4

**Expected behavior:**
After step4, it should focus on “item1” edit node, and no additional “”edit node in Inspect.

WPMGPRoSToTeMa:
Fixes #4810. Implements an accessible object for label edit control in ListView. This object now is the last child of the ListView (the same behavior as in the Windows Explorer). The object now also has a non-empty name according to its value (Windows Explorer behavior). The original object wasn't producing automation events related to the text, it was done by the UIAutomationCore proxy -- this was also implemented along with the text pattern support for the new accessible object, so the Narrator announces its text correctly.

## Proposed changes

- Implemented native window wrapper and accessible object for label edit control in ListView
- Implemented text pattern provider for label edit control (and added triggering automation events about label edit text)
- Made label edit accessible object be the last child of ListView (Windows Explorer behavior)
- Made label edit name value set according to its text value (Windows Explorer behavior)

## Customer Impact

- Narrator now announces the name of the label edit control

## Regression? 

- No

## Risk

- Minimal

### Before

image

### After

image

## Test methodology 

- Manual (using inspect and Narrator tools to verify hierarchy and text pattern)
- Unit-tests (not currently presented)
- CTI (planned)

 

## Test environment(s) 

- Microsoft Windows [Version 10.0.19044.1348]
- .NET 7.0.0-alpha.1.21602.1

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6309)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/Interop/User32/Interop.SetWinEventHook.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/User32/Interop.SetWinEventHook.cs
@@ -1,0 +1,19 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class User32
+    {
+        [DllImport(Libraries.User32, ExactSpelling = true, SetLastError = true)]
+        public static extern nint SetWinEventHook(
+            uint eventMin, uint eventMax, nint hmodWinEventProc, WINEVENTPROC pfnWinEventProc, uint idProcess, uint idThread, WINEVENT dwFlags);
+
+        public static nint SetWinEventHook(uint eventId, WINEVENTPROC pfnWinEventProc)
+            => SetWinEventHook(
+                eventId, eventId, Kernel32.GetModuleHandleW(null), pfnWinEventProc, (uint)Environment.ProcessId, Kernel32.GetCurrentThreadId(), WINEVENT.INCONTEXT);
+    }
+}
--- a/src/System.Windows.Forms.Primitives/src/Interop/User32/Interop.SetWinEventHook.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/User32/Interop.SetWinEventHook.cs
@@ -9,11 +9,11 @@ internal static partial class Interop
     internal static partial class User32
     {
         [DllImport(Libraries.User32, ExactSpelling = true, SetLastError = true)]
-        public static extern nint SetWinEventHook(
-            uint eventMin, uint eventMax, nint hmodWinEventProc, WINEVENTPROC pfnWinEventProc, uint idProcess, uint idThread, WINEVENT dwFlags);
+        private static extern nint SetWinEventHook(uint eventMin, uint eventMax, nint hmodWinEventProc,
+            WINEVENTPROC pfnWinEventProc, uint idProcess, uint idThread, WINEVENT dwFlags);
 
         public static nint SetWinEventHook(uint eventId, WINEVENTPROC pfnWinEventProc)
-            => SetWinEventHook(
-                eventId, eventId, Kernel32.GetModuleHandleW(null), pfnWinEventProc, (uint)Environment.ProcessId, Kernel32.GetCurrentThreadId(), WINEVENT.INCONTEXT);
+            => SetWinEventHook(eventId, eventId, Kernel32.GetModuleHandleW(null), pfnWinEventProc,
+                (uint)Environment.ProcessId, Kernel32.GetCurrentThreadId(), WINEVENT.INCONTEXT);
     }
 }
--- a/src/System.Windows.Forms.Primitives/src/Interop/User32/Interop.WINEVENT.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/User32/Interop.WINEVENT.cs
@@ -6,7 +6,7 @@ internal static partial class Interop
 {
     internal static partial class User32
     {
-        public enum WINEVENT : int
+        public enum WINEVENT : uint
         {
             OUTOFCONTEXT   = 0,
             SKIPOWNTHREAD  = 1,
--- a/src/System.Windows.Forms.Primitives/src/Interop/User32/Interop.WINEVENT.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/User32/Interop.WINEVENT.cs
@@ -1,0 +1,17 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+internal static partial class Interop
+{
+    internal static partial class User32
+    {
+        public enum WINEVENT : int
+        {
+            OUTOFCONTEXT   = 0,
+            SKIPOWNTHREAD  = 1,
+            SKIPOWNPROCESS = 2,
+            INCONTEXT      = 4,
+        }
+    }
+}
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
@@ -179,7 +179,7 @@ namespace System.Windows.Forms
 
         private bool _blockLabelEdit;
 
-        private ListViewLabelEditNativeWindow _labelEdit;
+        private ListViewLabelEditNativeWindow? _labelEdit;
 
         // Background image stuff
         // Because we have to create a temporary file and the OS does not clean up the temporary files from the machine
@@ -6468,9 +6468,12 @@ namespace System.Windows.Forms
 
                 case (int)LVN.BEGINLABELEDITW:
                     {
+                        Debug.Assert(_labelEdit is null,
+                            "A new label editing shouldn't start before the previous one ended");
                         if (_labelEdit is not null)
                         {
-                            throw new InvalidOperationException("A new label editing can't be started before the previous one was ended.");
+                            _labelEdit.ReleaseHandle();
+                            _labelEdit = null;
                         }
 
                         bool cancelEdit;
@@ -6524,9 +6527,10 @@ namespace System.Windows.Forms
 
                 case (int)LVN.ENDLABELEDITW:
                     {
+                        Debug.Assert(_labelEdit is not null, "There is no active label edit to end");
                         if (_labelEdit is null)
                         {
-                            throw new InvalidOperationException("There is no active label edit to end");
+                            break;
                         }
 
                         _labelEdit.ReleaseHandle();
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
@@ -179,6 +179,8 @@ namespace System.Windows.Forms
 
         private bool _blockLabelEdit;
 
+        private ListViewLabelEditNativeWindow _labelEdit;
+
         // Background image stuff
         // Because we have to create a temporary file and the OS does not clean up the temporary files from the machine
         // we have to do that ourselves
@@ -6466,6 +6468,11 @@ namespace System.Windows.Forms
 
                 case (int)LVN.BEGINLABELEDITW:
                     {
+                        if (_labelEdit is not null)
+                        {
+                            throw new InvalidOperationException("A new label editing can't be started before the previous one was ended.");
+                        }
+
                         bool cancelEdit;
                         if (_blockLabelEdit)
                         {
@@ -6481,6 +6488,13 @@ namespace System.Windows.Forms
 
                         m.ResultInternal = cancelEdit ? 1 : 0;
                         _listViewState[LISTVIEWSTATE_inLabelEdit] = !cancelEdit;
+
+                        if (!cancelEdit)
+                        {
+                            _labelEdit = new ListViewLabelEditNativeWindow(this);
+                            _labelEdit.AssignHandle(User32.SendMessageW(this, (User32.WM)LVM.GETEDITCONTROL));
+                        }
+
                         break;
                     }
 
@@ -6510,6 +6524,14 @@ namespace System.Windows.Forms
 
                 case (int)LVN.ENDLABELEDITW:
                     {
+                        if (_labelEdit is null)
+                        {
+                            throw new InvalidOperationException("There is no active label edit to end");
+                        }
+
+                        _labelEdit.ReleaseHandle();
+                        _labelEdit = null;
+
                         _listViewState[LISTVIEWSTATE_inLabelEdit] = false;
                         NMLVDISPINFO* dispInfo = (NMLVDISPINFO*)m.LParamInternal;
                         string? text = dispInfo->item.pszText is null ? null : new string(dispInfo->item.pszText);
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditAccessibleObject.cs
@@ -2,7 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System.Diagnostics;
+using System.Drawing;
 using System.Runtime.InteropServices;
 using static Interop;
 
@@ -12,57 +12,57 @@ namespace System.Windows.Forms
     {
         private const string LIST_VIEW_LABEL_EDIT_AUTOMATION_ID = "1";
 
-        private ListView _owner;
-        private IntPtr _handle;
+        private readonly ListView _owningListView;
+        private readonly IntPtr _handle;
+        private readonly ListViewLabelEditUiaTextProvider _textProvider;
         private int[]? _runtimeId;
 
-        public ListViewLabelEditAccessibleObject(ListView owner, IHandle handle)
+        public ListViewLabelEditAccessibleObject(ListView owningListView, IHandle handle)
         {
-            _owner = owner;
+            _owningListView = owningListView.OrThrowIfNull();
             _handle = handle.Handle;
             UseStdAccessibleObjects(_handle);
-
-            ListViewLabelEditUiaTextProvider textProvider = new(owner, handle, this);
-            UseTextProviders(textProvider, textProvider);
+            _textProvider = new ListViewLabelEditUiaTextProvider(owningListView, handle, this);
         }
 
-        internal override object? GetPropertyValue(UiaCore.UIA propertyID)
-            => propertyID switch
-            {
-                UiaCore.UIA.RuntimeIdPropertyId => RuntimeId,
-                UiaCore.UIA.BoundingRectanglePropertyId => Bounds,
-                UiaCore.UIA.ProcessIdPropertyId => Environment.ProcessId,
-                UiaCore.UIA.ControlTypePropertyId => UiaCore.UIA.EditControlTypeId,
-                UiaCore.UIA.NamePropertyId => Name,
-                UiaCore.UIA.AccessKeyPropertyId => string.Empty,
-                UiaCore.UIA.HasKeyboardFocusPropertyId => true,
-                UiaCore.UIA.IsKeyboardFocusablePropertyId => (State & AccessibleStates.Focusable) == AccessibleStates.Focusable,
-                UiaCore.UIA.IsEnabledPropertyId => Enabled,
-                UiaCore.UIA.AutomationIdPropertyId => LIST_VIEW_LABEL_EDIT_AUTOMATION_ID,
-                UiaCore.UIA.HelpTextPropertyId => Help ?? string.Empty,
-                UiaCore.UIA.IsContentElementPropertyId => true,
-                UiaCore.UIA.IsPasswordPropertyId => false,
-                UiaCore.UIA.NativeWindowHandlePropertyId => _handle,
-                UiaCore.UIA.IsOffscreenPropertyId => false,
-                UiaCore.UIA.IsTextPatternAvailablePropertyId => IsPatternSupported(UiaCore.UIA.TextPatternId),
-                UiaCore.UIA.IsTextPattern2AvailablePropertyId => IsPatternSupported(UiaCore.UIA.TextPattern2Id),
-                UiaCore.UIA.IsValuePatternAvailablePropertyId => IsPatternSupported(UiaCore.UIA.ValuePatternId),
-                UiaCore.UIA.IsLegacyIAccessiblePatternAvailablePropertyId => IsPatternSupported(UiaCore.UIA.LegacyIAccessiblePatternId),
-                _ => base.GetPropertyValue(propertyID),
-            };
+        private protected override string AutomationId => LIST_VIEW_LABEL_EDIT_AUTOMATION_ID;
 
         internal override UiaCore.IRawElementProviderFragment? FragmentNavigate(UiaCore.NavigateDirection direction)
         {
-            AccessibleObject parent = _owner.AccessibilityObject;
+            AccessibleObject parent = _owningListView.AccessibilityObject;
 
             return direction switch
             {
                 UiaCore.NavigateDirection.Parent => parent,
                 UiaCore.NavigateDirection.NextSibling => parent.GetChildIndex(this) is int childId and >= 0 ? parent.GetChild(childId + 1) : null,
                 UiaCore.NavigateDirection.PreviousSibling => parent.GetChildIndex(this) is int childId and >= 0 ? parent.GetChild(childId - 1) : null,
-                UiaCore.NavigateDirection.FirstChild or UiaCore.NavigateDirection.LastChild => null,
                 _ => base.FragmentNavigate(direction),
             };
+        }
+
+        internal override UiaCore.IRawElementProviderFragmentRoot FragmentRoot => _owningListView.AccessibilityObject;
+
+        internal override object? GetPropertyValue(UiaCore.UIA propertyID)
+            => propertyID switch
+            {
+                UiaCore.UIA.ProcessIdPropertyId => Environment.ProcessId,
+                UiaCore.UIA.ControlTypePropertyId => UiaCore.UIA.EditControlTypeId,
+                UiaCore.UIA.AccessKeyPropertyId => string.Empty,
+                UiaCore.UIA.HasKeyboardFocusPropertyId => true,
+                UiaCore.UIA.IsKeyboardFocusablePropertyId => (State & AccessibleStates.Focusable) == AccessibleStates.Focusable,
+                UiaCore.UIA.IsEnabledPropertyId => _owningListView.Enabled,
+                UiaCore.UIA.IsContentElementPropertyId => true,
+                UiaCore.UIA.NativeWindowHandlePropertyId => _handle,
+                _ => base.GetPropertyValue(propertyID),
+            };
+
+        internal override UiaCore.IRawElementProviderSimple? HostRawElementProvider
+        {
+            get
+            {
+                UiaCore.UiaHostProviderFromHwnd(new HandleRef(this, _handle), out UiaCore.IRawElementProviderSimple provider);
+                return provider;
+            }
         }
 
         internal override bool IsPatternSupported(UiaCore.UIA patternId) => patternId switch
@@ -74,28 +74,32 @@ namespace System.Windows.Forms
             _ => base.IsPatternSupported(patternId),
         };
 
-        internal override int[] RuntimeId => _runtimeId ??= new int[] { RuntimeIDFirstItem, PARAM.ToInt(_handle) };
-
-        internal override UiaCore.IRawElementProviderFragmentRoot? FragmentRoot => _owner.AccessibilityObject;
-
-        internal override UiaCore.IRawElementProviderSimple? HostRawElementProvider
-        {
-            get
-            {
-                UiaCore.UiaHostProviderFromHwnd(new HandleRef(this, _handle), out UiaCore.IRawElementProviderSimple provider);
-                return provider;
-            }
-        }
-
         public override string Name => User32.GetWindowText(_handle);
 
-        private bool Enabled
-        {
-            get
-            {
-                Debug.Assert(_owner.Enabled);
-                return _owner.Enabled;
-            }
-        }
+        internal override int[] RuntimeId => _runtimeId ??= new int[] { RuntimeIDFirstItem, PARAM.ToInt(_handle) };
+
+        internal override UiaCore.ITextRangeProvider DocumentRangeInternal
+            => _textProvider.DocumentRange;
+
+        internal override UiaCore.ITextRangeProvider[]? GetTextSelection()
+            => _textProvider.GetSelection();
+
+        internal override UiaCore.ITextRangeProvider[]? GetTextVisibleRanges()
+            => _textProvider.GetVisibleRanges();
+
+        internal override UiaCore.ITextRangeProvider? GetTextRangeFromChild(UiaCore.IRawElementProviderSimple childElement)
+            => _textProvider.RangeFromChild(childElement);
+
+        internal override UiaCore.ITextRangeProvider? GetTextRangeFromPoint(Point screenLocation)
+            => _textProvider.RangeFromPoint(screenLocation);
+
+        internal override UiaCore.SupportedTextSelection SupportedTextSelectionInternal
+            => _textProvider.SupportedTextSelection;
+
+        internal override UiaCore.ITextRangeProvider? GetTextCaretRange(out BOOL isActive)
+            => _textProvider.GetCaretRange(out isActive);
+
+        internal override UiaCore.ITextRangeProvider GetRangeFromAnnotation(UiaCore.IRawElementProviderSimple annotationElement)
+            => _textProvider.RangeFromAnnotation(annotationElement);
     }
 }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditAccessibleObject.cs
@@ -1,0 +1,101 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+using static Interop;
+
+namespace System.Windows.Forms
+{
+    internal class ListViewLabelEditAccessibleObject : AccessibleObject
+    {
+        private const string LIST_VIEW_LABEL_EDIT_AUTOMATION_ID = "1";
+
+        private ListView _owner;
+        private IntPtr _handle;
+        private int[]? _runtimeId;
+
+        public ListViewLabelEditAccessibleObject(ListView owner, IHandle handle)
+        {
+            _owner = owner;
+            _handle = handle.Handle;
+            UseStdAccessibleObjects(_handle);
+
+            ListViewLabelEditUiaTextProvider textProvider = new(owner, handle, this);
+            UseTextProviders(textProvider, textProvider);
+        }
+
+        internal override object? GetPropertyValue(UiaCore.UIA propertyID)
+            => propertyID switch
+            {
+                UiaCore.UIA.RuntimeIdPropertyId => RuntimeId,
+                UiaCore.UIA.BoundingRectanglePropertyId => Bounds,
+                UiaCore.UIA.ProcessIdPropertyId => Environment.ProcessId,
+                UiaCore.UIA.ControlTypePropertyId => UiaCore.UIA.EditControlTypeId,
+                UiaCore.UIA.NamePropertyId => Name,
+                UiaCore.UIA.AccessKeyPropertyId => string.Empty,
+                UiaCore.UIA.HasKeyboardFocusPropertyId => true,
+                UiaCore.UIA.IsKeyboardFocusablePropertyId => (State & AccessibleStates.Focusable) == AccessibleStates.Focusable,
+                UiaCore.UIA.IsEnabledPropertyId => Enabled,
+                UiaCore.UIA.AutomationIdPropertyId => LIST_VIEW_LABEL_EDIT_AUTOMATION_ID,
+                UiaCore.UIA.HelpTextPropertyId => Help ?? string.Empty,
+                UiaCore.UIA.IsContentElementPropertyId => true,
+                UiaCore.UIA.IsPasswordPropertyId => false,
+                UiaCore.UIA.NativeWindowHandlePropertyId => _handle,
+                UiaCore.UIA.IsOffscreenPropertyId => false,
+                UiaCore.UIA.IsTextPatternAvailablePropertyId => IsPatternSupported(UiaCore.UIA.TextPatternId),
+                UiaCore.UIA.IsTextPattern2AvailablePropertyId => IsPatternSupported(UiaCore.UIA.TextPattern2Id),
+                UiaCore.UIA.IsValuePatternAvailablePropertyId => IsPatternSupported(UiaCore.UIA.ValuePatternId),
+                UiaCore.UIA.IsLegacyIAccessiblePatternAvailablePropertyId => IsPatternSupported(UiaCore.UIA.LegacyIAccessiblePatternId),
+                _ => base.GetPropertyValue(propertyID),
+            };
+
+        internal override UiaCore.IRawElementProviderFragment? FragmentNavigate(UiaCore.NavigateDirection direction)
+        {
+            AccessibleObject parent = _owner.AccessibilityObject;
+
+            return direction switch
+            {
+                UiaCore.NavigateDirection.Parent => parent,
+                UiaCore.NavigateDirection.NextSibling => parent.GetChildIndex(this) is int childId and >= 0 ? parent.GetChild(childId + 1) : null,
+                UiaCore.NavigateDirection.PreviousSibling => parent.GetChildIndex(this) is int childId and >= 0 ? parent.GetChild(childId - 1) : null,
+                UiaCore.NavigateDirection.FirstChild or UiaCore.NavigateDirection.LastChild => null,
+                _ => base.FragmentNavigate(direction),
+            };
+        }
+
+        internal override bool IsPatternSupported(UiaCore.UIA patternId) => patternId switch
+        {
+            UiaCore.UIA.TextPatternId => true,
+            UiaCore.UIA.TextPattern2Id => true,
+            UiaCore.UIA.ValuePatternId => true,
+            UiaCore.UIA.LegacyIAccessiblePatternId => true,
+            _ => base.IsPatternSupported(patternId),
+        };
+
+        internal override int[] RuntimeId => _runtimeId ??= new int[] { RuntimeIDFirstItem, PARAM.ToInt(_handle) };
+
+        internal override UiaCore.IRawElementProviderFragmentRoot? FragmentRoot => _owner.AccessibilityObject;
+
+        internal override UiaCore.IRawElementProviderSimple? HostRawElementProvider
+        {
+            get
+            {
+                UiaCore.UiaHostProviderFromHwnd(new HandleRef(this, _handle), out UiaCore.IRawElementProviderSimple provider);
+                return provider;
+            }
+        }
+
+        public override string Name => User32.GetWindowText(_handle);
+
+        private bool Enabled
+        {
+            get
+            {
+                Debug.Assert(_owner.Enabled);
+                return _owner.Enabled;
+            }
+        }
+    }
+}
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditNativeWindow.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditNativeWindow.cs
@@ -1,0 +1,183 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.InteropServices;
+using static Interop;
+
+namespace System.Windows.Forms
+{
+    internal class ListViewLabelEditNativeWindow : NativeWindow
+    {
+        private const uint TextSelectionChanged = 0x8014;
+
+        private ListView _owner;
+
+        private AccessibleObject? _accessibilityObject;
+
+        private User32.WINEVENTPROC? _winEventProc;
+
+        private nint _valueChangeHook;
+        private nint _textSelectionChangedHook;
+
+        private bool _winEventHooksInstalled;
+
+        public ListViewLabelEditNativeWindow(ListView owner)
+        {
+            _owner = owner.OrThrowIfNull();
+        }
+
+        public AccessibleObject AccessibilityObject => _accessibilityObject ??= new ListViewLabelEditAccessibleObject(_owner, this);
+
+        public override void ReleaseHandle()
+        {
+            if (_winEventHooksInstalled)
+            {
+                User32.UnhookWinEvent(_valueChangeHook);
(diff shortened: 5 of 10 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ListViewChildLabelEditNativeWindow.cs
```diff
@@ -0,0 +1,161 @@
...
+                m.ResultInternal = UiaCore.UiaReturnRawElementProvider(
+                    this,
+                    m.WParamInternal,
+                    m.LParamInternal,
+                    AccessibilityObject);
+
+                return;
+            }
+
+            // See "How to Handle WM_GETOBJECT" in MSDN
```
Tanya-Solyanik, line 128: replace this comment with an actual link to docs?

src/System.Windows.Forms/src/System/Windows/Forms/ListViewChildLabelEditNativeWindow.cs
```diff
@@ -0,0 +1,161 @@
...
+
+        public AccessibleObject AccessibilityObject
+        {
+            get
+            {
+                if (_accessibilityObject is null)
+                {
+                    _accessibilityObject = new ListViewChildLabelEditAccessibleObject(_owner, this);
+
+                    if (!_winEventHooksInstalled)
```
Tanya-Solyanik, line 38: hook should not be installed if we don't have the acc obj

src/System.Windows.Forms/src/System/Windows/Forms/ListViewChildLabelEditNativeWindow.cs
```diff
@@ -0,0 +1,161 @@
...
+        private User32.WINEVENTPROC? _winEventProc;
+
+        private nint _valueChangeHook;
+        private nint _textSelectionChangedHook;
+
+        private bool _winEventHooksInstalled;
+
+        public ListViewChildLabelEditNativeWindow(ListView owner)
+        {
+            _owner = owner;
```
Tanya-Solyanik, line 27: THrow is argument is null.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewChildLabelEditNativeWindow.cs
```diff
@@ -0,0 +1,161 @@
...
+                    User32.UnhookWinEvent(_valueChangeHook);
+                    User32.UnhookWinEvent(_textSelectionChangedHook);
+
+                    _winEventHooksInstalled = false;
+                }
+
+                return;
+            }
+
+            if (UiaCore.UiaClientsAreListening().IsFalse())
```
Tanya-Solyanik, line 65: Please move this test into `InstallWinHooks()`

Tanya-Solyanik (commented):
Looks good. I added cleanup-y comments only. 
I did not review text patterns and pinvoke definitions yet.

src/System.Windows.Forms.Primitives/src/Interop/User32/Interop.SetWinEventHook.cs
```diff
@@ -0,0 +1,19 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class User32
+    {
+        [DllImport(Libraries.User32, ExactSpelling = true, SetLastError = true)]
```
Tanya-Solyanik, line 11: We are not checking the last error, we don't need to set this attribute, it makes interop a little less preformant

src/System.Windows.Forms.Primitives/src/Interop/User32/Interop.SetWinEventHook.cs
```diff
@@ -0,0 +1,19 @@
...
+        public static extern nint SetWinEventHook(uint eventMin, uint eventMax, nint hmodWinEventProc, WINEVENTPROC pfnWinEventProc, int idProcess, uint idThread, WINEVENT dwFlags);
+
+        public static nint SetWinEventHook(uint eventId, WINEVENTPROC pfnWinEventProc)
+        {
+            return SetWinEventHook(eventId, eventId, Kernel32.GetModuleHandleW(null), pfnWinEventProc, Environment.ProcessId, Kernel32.GetCurrentThreadId(), WINEVENT.INCONTEXT);
```
Tanya-Solyanik, line 16: install https://marketplace.visualstudio.com/items?itemName=PaulHarrington.EditorGuidelinesPreview, it will suggest where to put line breaks in the VS

src/System.Windows.Forms/src/System/Windows/Forms/ListViewChildLabelEditNativeWindow.cs
```diff
@@ -0,0 +1,161 @@
...
+                return;
+            }
+
+            // See "How to Handle WM_GETOBJECT" in MSDN
+            if ((int)m.LParamInternal == User32.OBJID.CLIENT)
+            {
+                // Get the IAccessible GUID
+                Guid IID_IAccessible = new Guid(NativeMethods.uuid_IAccessible);
+
+                // Get an Lresult for the accessibility Object for this control
```
Tanya-Solyanik, line 134: nit:  comments should be proper sentences ending with a dot

src/System.Windows.Forms/src/System/Windows/Forms/ListViewChildLabelEditAccessibleObject.cs
```diff
@@ -0,0 +1,114 @@
...
+            _handle = handle.Handle;
+            UseStdAccessibleObjects(_handle);
+
+            ListViewChildLabelEditUiaTextProvider textProvider = new(owner, handle, this);
+            UseTextProviders(textProvider, textProvider);
+        }
+
+        internal override object? GetPropertyValue(UiaCore.UIA propertyID)
+        {
+            switch (propertyID)
```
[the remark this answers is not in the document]
WPMGPRoSToTeMa: There was a problem that `Enabled` case occupied two lines of code (because of `Debug.Assert`). I moved it to a separate property and applied pattern syntax to this switch.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewChildLabelEditAccessibleObject.cs
```diff
@@ -0,0 +1,114 @@
...
+        internal override bool IsPatternSupported(UiaCore.UIA patternId) => patternId switch
+        {
+            UiaCore.UIA.TextPatternId => true,
+            UiaCore.UIA.TextPattern2Id => true,
+            UiaCore.UIA.ValuePatternId => true,
+            UiaCore.UIA.LegacyIAccessiblePatternId => true,
+            _ => base.IsPatternSupported(patternId),
+        };
+
+        internal override int[] RuntimeId => new int[] { 0x2A, PARAM.ToInt(_handle) };
```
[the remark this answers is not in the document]
WPMGPRoSToTeMa: Added caching with a backing field. We may want to use a read-only auto property instead of that and initialize it directly in the constructor (runtime id is almost always retrieved by the accessibility tools anyway) -- this will make code a bit cleaner I think.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewChildLabelEditAccessibleObject.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        internal override UiaCore.IRawElementProviderFragmentRoot? FragmentRoot => _owner.AccessibilityObject;
+
+        internal override UiaCore.IRawElementProviderSimple? HostRawElementProvider
+        {
+            get
+            {
+                UiaCore.UiaHostProviderFromHwnd(new HandleRef(this, _handle), out UiaCore.IRawElementProviderSimple provider);
```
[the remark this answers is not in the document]
WPMGPRoSToTeMa: Do we really need to do that? I checked the other uses of `UiaHostProviderFromHwnd` through the code and it is not tested anywhere.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewChildLabelEditAccessibleObject.cs
```diff
@@ -0,0 +1,114 @@
...
+                return provider;
+            }
+        }
+
+        public override string Name => User32.GetWindowText(_handle);
```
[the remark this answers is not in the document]
WPMGPRoSToTeMa: `GetWindowText` implementation is pretty tricky, but the return value can't be null.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewChildLabelEditNativeWindow.cs
```diff
@@ -0,0 +1,161 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.InteropServices;
+using static Interop;
+
+namespace System.Windows.Forms
+{
+    internal class ListViewChildLabelEditNativeWindow : NativeWindow
+    {
+        private const uint TextSelectionChanged = 0x8014;
```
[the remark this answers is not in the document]
WPMGPRoSToTeMa: Yep, it should be defined in `AccessibleEvents` enum, but modifying `AccessibleEvents` involves changing of the public API. Should I do that?

src/System.Windows.Forms/src/System/Windows/Forms/ListViewChildLabelEditNativeWindow.cs
```diff
@@ -0,0 +1,161 @@
...
+            }
+
+            InstallWinEventHooks();
+        }
+
+        protected override void WndProc(ref Message m)
+        {
+            switch (m.MsgInternal)
+            {
+                case User32.WM.GETOBJECT:
```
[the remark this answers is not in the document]
WPMGPRoSToTeMa: > Should we unhook winevents hooks on wm_destroy?

As I can see `LVN_ENDLABELEDIT` notification is always sent when a listview control is destroyed with an active label edit. Therefore it is enough to do the cleanup on `ReleaseHandle`.

> And also unregister the UIAProvider - call UiaReturnRawElementProvider with all nulls

Thanks, we also need to call `UiaDisconnectProvider` to decrement the ref counter on accessible object (I've checked it using the VS memory profiler and it really works).

src/System.Windows.Forms/src/System/Windows/Forms/ListViewChildLabelEditNativeWindow.cs
WPMGPRoSToTeMa: We actually have an accessible object at this moment, but I changed the logic a bit anyway.

src/System.Windows.Forms.Primitives/src/Interop/User32/Interop.SetWinEventHook.cs
WPMGPRoSToTeMa: I basically copied it from the `SetWindowsHookExW` definition, the interesting thing is that its calling code is also not checking for the last error. The performance is not crucial in this case as it is called very rarely. But anyway I think I will remove it as you suggested.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewChildLabelEditUiaTextProvider.cs
```diff
@@ -0,0 +1,368 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Diagnostics;
+using System.Drawing;
+using System.Windows.Forms.Automation;
+using static Interop;
+using static Interop.User32;
```
Tanya-Solyanik, line 9: In this code base it's more common to keep binary name with the Interop function for context. I.e. we say everywhere else User32.SendMessage rather than SendMessage. The thing is that we have another SendMessage in ComCtl namespace, so it might be confusing.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewChildLabelEditUiaTextProvider.cs
```diff
@@ -0,0 +1,368 @@
...
+                return;
+            }
+
+            SendMessageW(_owningChildEdit, (WM)EM.SETSEL, start, end);
+        }
+
+        private int GetCharIndexFromPosition(Point pt)
+        {
+            int index = (int)SendMessageW(_owningChildEdit, (WM)EM.CHARFROMPOS, 0, PARAM.FromPoint(pt));
+            index = PARAM.LOWORD(index);
```
Tanya-Solyanik, line 312: We have a `LOWORD` helper that takes a `nint` as an argument, extra cast is not needed.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewChildLabelEditUiaTextProvider.cs
```diff
@@ -0,0 +1,368 @@
...
+
+            if (index < 0)
```
Tanya-Solyanik, line 314: This is the case when the point is outside the control, I don't see why we are returning the first char.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewChildLabelEditUiaTextProvider.cs
```diff
@@ -0,0 +1,368 @@
...
+            {
+                index = 0;
+            }
+            else
+            {
+                string t = Text;
+                // EM_CHARFROMPOS will return an invalid number if the last character in the RichEdit
+                // is a newline.
+                //
```
Tanya-Solyanik, line 323: Please remove the empty line

Tanya-Solyanik (commented):
I did not finish the review

src/System.Windows.Forms.Primitives/src/Interop/User32/Interop.SetWinEventHook.cs
RussKie: We definitely want `SetLastError = true`, and _ideally_ check the error.

src/System.Windows.Forms.Primitives/src/Interop/User32/Interop.SetWinEventHook.cs
```diff
@@ -0,0 +1,19 @@
...
+
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class User32
+    {
+        [DllImport(Libraries.User32, ExactSpelling = true, SetLastError = true)]
+        public static extern nint SetWinEventHook(
+            uint eventMin, uint eventMax, nint hmodWinEventProc, WINEVENTPROC pfnWinEventProc, int idProcess, uint idThread, WINEVENT dwFlags);
```
RussKie, line 13:
```suggestion
            uint eventMin, uint eventMax, nint hmodWinEventProc, WINEVENTPROC pfnWinEventProc, uint idProcess, uint idThread, WINEVENT dwFlags);
```

src/System.Windows.Forms.Primitives/src/Interop/User32/Interop.SetWinEventHook.cs
Tanya-Solyanik: Why do we want SetLastError?

RussKie: Apologies, I must have looked at the wrong docs. Unlike [`SetWindowsHookExW` docs](https://docs.microsoft.com/windows/win32/api/winuser/nf-winuser-setwindowshookexw#return-value), [`SetWinEventHook` docs](https://docs.microsoft.com/windows/win32/api/winuser/nf-winuser-setwineventhook) don't say anything about setting the error.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditAccessibleObject.cs
```diff
@@ -0,0 +1,101 @@
...
+    {
+        private const string LIST_VIEW_LABEL_EDIT_AUTOMATION_ID = "1";
+
+        private ListView _owner;
+        private IntPtr _handle;
+        private int[]? _runtimeId;
+
+        public ListViewLabelEditAccessibleObject(ListView owner, IHandle handle)
+        {
+            _owner = owner;
```
RussKie, line 21: I vaguely remember a discussion with @vladimir-krestov  and @SergeySmirnov-Akvelon about prefixing the first parameter on non public types with "owning", e.g. `owningListView`.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditNativeWindow.cs
```diff
@@ -0,0 +1,183 @@
...
+        private ListView _owner;
+
+        private AccessibleObject? _accessibilityObject;
+
+        private User32.WINEVENTPROC? _winEventProc;
+
+        private nint _valueChangeHook;
+        private nint _textSelectionChangedHook;
+
+        private bool _winEventHooksInstalled;
```
RussKie, line 23: Please keep these together

src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditNativeWindow.cs
```diff
@@ -0,0 +1,183 @@
...
+            base.OnHandleChange();
+
+            if (Handle == IntPtr.Zero)
+            {
+                return;
+            }
+
+            // Install winevent hooks only in the case when the parent listview already has an accessible object created.
+            // If we won't install hooks at the label edit startup then Narrator won't announce the text pattern for it.
+            // If we will just check for UiaClientsAreListening then we will have hooks installed even if Narrator isn't currently run.
```
RussKie, line 70: Did I get correctly the intent?
```suggestion
            // Install winevent hooks *only* when the parent listview has an existing accessible object.
            // If we don't install hooks at the label edit startup then assistive tech (e.g. Narrator) won't announce the text pattern for it.
            // By invoking UiaClientsAreListening we will have hooks installed even if the assistive tech isn't currently run.
```

Also "_...then assistive tech (e.g. Narrator) won't announce the text pattern for **it**._" What is "_it_" in this context?

src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditNativeWindow.cs
```diff
@@ -0,0 +1,183 @@
...
+            _winEventHooksInstalled = true;
+        }
+
+        private AccessibleObject RequestAccessibilityObject()
+        {
+            AccessibleObject result = AccessibilityObject;
+
+            // Accessibility object was likely requested by some accessibility tool (because of WM_GETOBJECT message).
+            // The tool that requested the object may be Narrator in which case we may need to install winevent hooks to
+            // produce the automation events related to the text pattern.
```
RussKie, line 112:
```suggestion
            // Accessibility object was likely requested by an assistive tech (which sent WM_GETOBJECT message).
            // We may need to install winevent hooks to produce the automation events related to the text pattern.
```

src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditNativeWindow.cs
```diff
@@ -0,0 +1,183 @@
...
+            _winEventProc = new User32.WINEVENTPROC(WinEventProc);
+            _valueChangeHook = User32.SetWinEventHook((uint)AccessibleEvents.ValueChange, _winEventProc);
+            _textSelectionChangedHook = User32.SetWinEventHook(TextSelectionChanged, _winEventProc);
+
+            _winEventHooksInstalled = true;
+        }
+
+        private AccessibleObject RequestAccessibilityObject()
+        {
+            AccessibleObject result = AccessibilityObject;
```
RussKie, line 108: Why is it called `result`?

src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditNativeWindow.cs
```diff
@@ -0,0 +1,183 @@
...
+            }
+
+            _winEventProc = new User32.WINEVENTPROC(WinEventProc);
+            _valueChangeHook = User32.SetWinEventHook((uint)AccessibleEvents.ValueChange, _winEventProc);
+            _textSelectionChangedHook = User32.SetWinEventHook(TextSelectionChanged, _winEventProc);
+
+            _winEventHooksInstalled = true;
+        }
+
+        private AccessibleObject RequestAccessibilityObject()
```
RussKie, line 106: I am finding the method name at odds with its implementation. The name suggests it _sends_ or _initiate_ a request somewhere to get an accessibility object. In reality it just deferences `AccessibilityObject` property and install winhooks. The last bit is a side effect, which has an [element of surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditNativeWindow.cs
```diff
@@ -0,0 +1,183 @@
...
+{
+    internal class ListViewLabelEditNativeWindow : NativeWindow
+    {
+        private const uint TextSelectionChanged = 0x8014;
+
+        private ListView _owner;
+
+        private AccessibleObject? _accessibilityObject;
+
+        private User32.WINEVENTPROC? _winEventProc;
```
RussKie, line 18:
```suggestion
        private User32.WINEVENTPROC? _winEventProcCallback;
```

src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditNativeWindow.cs
```diff
@@ -0,0 +1,183 @@
...
+        private void InstallWinEventHooks()
+        {
+            if (UiaCore.UiaClientsAreListening().IsFalse())
+            {
+                return;
+            }
+
+            _winEventProc = new User32.WINEVENTPROC(WinEventProc);
+            _valueChangeHook = User32.SetWinEventHook((uint)AccessibleEvents.ValueChange, _winEventProc);
+            _textSelectionChangedHook = User32.SetWinEventHook(TextSelectionChanged, _winEventProc);
```
RussKie, line 101: Why can't we use the local variable? It doesn't appear to be used anywhere else...

src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditNativeWindow.cs
```diff
@@ -0,0 +1,183 @@
...
+            // produce the automation events related to the text pattern.
+            if (!_winEventHooksInstalled)
+            {
+                InstallWinEventHooks();
+            }
+
+            return result;
+        }
+
+        private void WinEventProc(nint hWinEventHook, uint eventId, nint hwnd, int idObject, int idChild, uint idEventThread, uint dwmsEventTime)
```
RussKie, line 121:
```suggestion
        private void WinEventProcCallback(nint hWinEventHook, uint eventId, nint hwnd, int idObject, int idChild, uint idEventThread, uint dwmsEventTime)
```

src/System.Windows.Forms/src/System/Windows/Forms/ListViewChildLabelEditNativeWindow.cs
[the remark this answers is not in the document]
Tanya-Solyanik: Yes, this is an acceptable change for a new release.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditAccessibleObject.cs
```diff
@@ -0,0 +1,101 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+using static Interop;
+
+namespace System.Windows.Forms
+{
+    internal class ListViewLabelEditAccessibleObject : AccessibleObject
```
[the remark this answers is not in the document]
Tanya-Solyanik: I thought that we are sorting members logically, not alphabetically.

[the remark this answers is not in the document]
RussKie: We are maintained a legacy codebase with established rules and guidelines. We should continue to maintain and adhere to those as much as practical. We had discussed and agreed on this with @vladimir-krestov and @SergeySmirnov-Akvelon.

src/System.Windows.Forms.Primitives/src/Interop/User32/Interop.SetWinEventHook.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    internal static partial class User32
+    {
+        [DllImport(Libraries.User32, ExactSpelling = true, SetLastError = true)]
+        public static extern nint SetWinEventHook(
+            uint eventMin, uint eventMax, nint hmodWinEventProc, WINEVENTPROC pfnWinEventProc, uint idProcess, uint idThread, WINEVENT dwFlags);
+
+        public static nint SetWinEventHook(uint eventId, WINEVENTPROC pfnWinEventProc)
+            => SetWinEventHook(
+                eventId, eventId, Kernel32.GetModuleHandleW(null), pfnWinEventProc, (uint)Environment.ProcessId, Kernel32.GetCurrentThreadId(), WINEVENT.INCONTEXT);
```
vladimir-krestov, line 17: Consider:
```suggestion
            => SetWinEventHook(eventId, eventId, Kernel32.GetModuleHandleW(null), pfnWinEventProc,
            (uint)Environment.ProcessId, Kernel32.GetCurrentThreadId(), WINEVENT.INCONTEXT);
```

src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditAccessibleObject.cs
```diff
@@ -0,0 +1,101 @@
...
+    {
+        private const string LIST_VIEW_LABEL_EDIT_AUTOMATION_ID = "1";
+
+        private ListView _owner;
+        private IntPtr _handle;
+        private int[]? _runtimeId;
+
+        public ListViewLabelEditAccessibleObject(ListView owner, IHandle handle)
```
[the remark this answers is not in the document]
vladimir-krestov: Yes, we decided to make every owning field in one style - `_owning[Control]`, eg. _owningListView

src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditAccessibleObject.cs
```diff
@@ -0,0 +1,101 @@
...
+using static Interop;
+
+namespace System.Windows.Forms
+{
+    internal class ListViewLabelEditAccessibleObject : AccessibleObject
+    {
+        private const string LIST_VIEW_LABEL_EDIT_AUTOMATION_ID = "1";
+
+        private ListView _owner;
+        private IntPtr _handle;
```
vladimir-krestov, line 16:
```suggestion
        private readonly ListView _owner;
        private readonly IntPtr _handle;
```

src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditAccessibleObject.cs
```diff
@@ -0,0 +1,101 @@
...
+        {
+            _owner = owner;
+            _handle = handle.Handle;
+            UseStdAccessibleObjects(_handle);
+
+            ListViewLabelEditUiaTextProvider textProvider = new(owner, handle, this);
+            UseTextProviders(textProvider, textProvider);
+        }
+
+        internal override object? GetPropertyValue(UiaCore.UIA propertyID)
```
vladimir-krestov, line 29: Please check the base class and remove unnecessary duplicated cases

src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditAccessibleObject.cs
```diff
@@ -0,0 +1,101 @@
...
+            => propertyID switch
+            {
+                UiaCore.UIA.RuntimeIdPropertyId => RuntimeId,
+                UiaCore.UIA.BoundingRectanglePropertyId => Bounds,
+                UiaCore.UIA.ProcessIdPropertyId => Environment.ProcessId,
+                UiaCore.UIA.ControlTypePropertyId => UiaCore.UIA.EditControlTypeId,
+                UiaCore.UIA.NamePropertyId => Name,
+                UiaCore.UIA.AccessKeyPropertyId => string.Empty,
+                UiaCore.UIA.HasKeyboardFocusPropertyId => true,
```
vladimir-krestov, line 38: It should be true if the owner is in focus

src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditAccessibleObject.cs
```diff
@@ -0,0 +1,101 @@
...
+                UiaCore.UIA.IsKeyboardFocusablePropertyId => (State & AccessibleStates.Focusable) == AccessibleStates.Focusable,
+                UiaCore.UIA.IsEnabledPropertyId => Enabled,
+                UiaCore.UIA.AutomationIdPropertyId => LIST_VIEW_LABEL_EDIT_AUTOMATION_ID,
+                UiaCore.UIA.HelpTextPropertyId => Help ?? string.Empty,
+                UiaCore.UIA.IsContentElementPropertyId => true,
+                UiaCore.UIA.IsPasswordPropertyId => false,
+                UiaCore.UIA.NativeWindowHandlePropertyId => _handle,
+                UiaCore.UIA.IsOffscreenPropertyId => false,
```
vladimir-krestov, line 46: Please remove this case, it should be like in the base class

src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditAccessibleObject.cs
```diff
@@ -0,0 +1,101 @@
...
+        internal override UiaCore.IRawElementProviderFragment? FragmentNavigate(UiaCore.NavigateDirection direction)
+        {
+            AccessibleObject parent = _owner.AccessibilityObject;
+
+            return direction switch
+            {
+                UiaCore.NavigateDirection.Parent => parent,
+                UiaCore.NavigateDirection.NextSibling => parent.GetChildIndex(this) is int childId and >= 0 ? parent.GetChild(childId + 1) : null,
+                UiaCore.NavigateDirection.PreviousSibling => parent.GetChildIndex(this) is int childId and >= 0 ? parent.GetChild(childId - 1) : null,
+                UiaCore.NavigateDirection.FirstChild or UiaCore.NavigateDirection.LastChild => null,
```
vladimir-krestov, line 63: Please remove this case for consistency with the rest AOs, the base class works for it.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditAccessibleObject.cs
```diff
@@ -0,0 +1,101 @@
...
+            UiaCore.UIA.TextPatternId => true,
+            UiaCore.UIA.TextPattern2Id => true,
+            UiaCore.UIA.ValuePatternId => true,
+            UiaCore.UIA.LegacyIAccessiblePatternId => true,
+            _ => base.IsPatternSupported(patternId),
+        };
+
+        internal override int[] RuntimeId => _runtimeId ??= new int[] { RuntimeIDFirstItem, PARAM.ToInt(_handle) };
+
+        internal override UiaCore.IRawElementProviderFragmentRoot? FragmentRoot => _owner.AccessibilityObject;
```
vladimir-krestov, line 79: Can AccessibilityObject value be null?
```suggestion
        internal override UiaCore.IRawElementProviderFragmentRoot FragmentRoot => _owner.AccessibilityObject;
```

src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditAccessibleObject.cs
```diff
@@ -0,0 +1,101 @@
...
+            }
+        }
+
+        public override string Name => User32.GetWindowText(_handle);
+
+        private bool Enabled
+        {
+            get
+            {
+                Debug.Assert(_owner.Enabled);
```
vladimir-krestov, line 96: You didn't add this for the rest members, so I would remove this and add a unit test.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditNativeWindow.cs
```diff
@@ -0,0 +1,183 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.InteropServices;
+using static Interop;
+
+namespace System.Windows.Forms
+{
+    internal class ListViewLabelEditNativeWindow : NativeWindow
+    {
+        private const uint TextSelectionChanged = 0x8014;
```
vladimir-krestov, line 12: I would put all these fields together

src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditNativeWindow.cs
```diff
@@ -0,0 +1,183 @@
...
+        {
+            if (_winEventHooksInstalled)
+            {
+                User32.UnhookWinEvent(_valueChangeHook);
+                User32.UnhookWinEvent(_textSelectionChangedHook);
+
+                _winEventHooksInstalled = false;
+            }
+
+            if (_accessibilityObject is not null)
```
vladimir-krestov, line 42: I would create IsAccessibilityObjectCreated created for readability, like we did for ControlAO

src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditNativeWindow.cs
```diff
@@ -0,0 +1,183 @@
...
+            {
+                InstallWinEventHooks();
+            }
+
+            return result;
+        }
+
+        private void WinEventProc(nint hWinEventHook, uint eventId, nint hwnd, int idObject, int idChild, uint idEventThread, uint dwmsEventTime)
+        {
+            if (hwnd != Handle || idObject != User32.OBJID.CLIENT)
```
vladimir-krestov, line 123: Do we need to check if AO is created before?

src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditUiaTextProvider.cs
```diff
@@ -0,0 +1,339 @@
...
+        ///  Since the label edit inside the ListView is always single-line, for optimization
+        ///  we always return 0 as the index of lines
+        /// </summary>
+        private const int OwnerChildEditLineIndex = 0;
+
+        /// <summary>
+        ///  Since the label edit inside the ListView is always single-line, for optimization
+        ///  we always return 1 as the number of lines
+        /// </summary>
+        private const int OwnerChildEditLinesCount = 1;
```
vladimir-krestov, line 24: nit: Put them together

src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditUiaTextProvider.cs
```diff
@@ -0,0 +1,339 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Diagnostics;
+using System.Drawing;
+using System.Windows.Forms.Automation;
+using static Interop;
+
+namespace System.Windows.Forms
+{
+    internal class ListViewLabelEditUiaTextProvider : UiaTextProvider2
```
vladimir-krestov, line 12: Is this possible to inherit from TextBoxBase provider?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewLabelEditAccessibleObjectTests.cs
```diff
@@ -0,0 +1,152 @@
...
+        Assert.Empty((string)accessibilityObject.GetPropertyValue(UiaCore.UIA.HelpTextPropertyId));
+        Assert.True((bool)accessibilityObject.GetPropertyValue(UiaCore.UIA.IsContentElementPropertyId));
+        Assert.False((bool)accessibilityObject.GetPropertyValue(UiaCore.UIA.IsPasswordPropertyId));
+        Assert.Equal(labelEdit.Handle, accessibilityObject.GetPropertyValue(UiaCore.UIA.NativeWindowHandlePropertyId));
+        Assert.False((bool)accessibilityObject.GetPropertyValue(UiaCore.UIA.IsOffscreenPropertyId));
+        Assert.True((bool)accessibilityObject.GetPropertyValue(UiaCore.UIA.IsTextPatternAvailablePropertyId));
+        Assert.True((bool)accessibilityObject.GetPropertyValue(UiaCore.UIA.IsTextPattern2AvailablePropertyId));
+        Assert.True((bool)accessibilityObject.GetPropertyValue(UiaCore.UIA.IsValuePatternAvailablePropertyId));
+        Assert.True((bool)accessibilityObject.GetPropertyValue(UiaCore.UIA.IsLegacyIAccessiblePatternAvailablePropertyId));
+    }
```
vladimir-krestov, line 41: Please add IsHandleCreated check for the listView and labelEdit

vladimir-krestov (commented):
Good work!

src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditNativeWindow.cs
dmitrii-drobotov: Thanks for cleaning up the texts. I assume "it" means the label edit control.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditAccessibleObject.cs
dmitrii-drobotov: It wouldn't work because owner (ListView) would not be in focus when edit is opened.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditAccessibleObject.cs
dmitrii-drobotov: Done, and added assert to `ListViewLabelEditAccessibleObject_GetPropertyValue_ReturnsExpected`:
```
Assert.Equal(listView.Enabled, (bool)accessibilityObject.GetPropertyValue(UiaCore.UIA.IsEnabledPropertyId));
```

src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditUiaTextProvider.cs
dmitrii-drobotov: No, to use `TextBoxBaseUiaTextProvider` ListViewLabelEdit would need to inherit from `TextBoxBase` but it inherits only from `NativeWindow`

src/System.Windows.Forms.Primitives/src/Interop/User32/Interop.SetWinEventHook.cs
```diff
@@ -0,0 +1,19 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class User32
+    {
+        [DllImport(Libraries.User32, ExactSpelling = true, SetLastError = true)]
+        public static extern nint SetWinEventHook(
```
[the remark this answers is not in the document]
dmitrii-drobotov: It indeed doesn't so I made it private.

RussKie:
> or example on this screenshot it could be where the arrow points.
> Should I make this change or keep it as is?

Great observation! To me your proposal sounds logical, though I'm not an expect in this area. What are we doing in similar cases for other controls?

WPMGPRoSToTeMa:
> Should I make this change or keep it as is?

Could you please check this behavior in Windows Explorer?

dmitrii-drobotov:
For Winforms controls it's not very consistent, e.g. for DataGridView the edit control is put inside an item:
image
And for for nodes of TreeView, the edit is on top (same as it was for ListView before this PR):
image

But in File Explorer for grouped list views it's in the bottom:
image

Since in this issue we take Explorer as a reference, I think we should go with its way, i.e. keep it as is.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewLabelEditAccessibleObjectTests.cs
Tanya-Solyanik: `Assert.True(IsHandleCreated)` is reasonable for documentation purposes

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
```diff
@@ -6418,6 +6420,11 @@ private unsafe void WmReflectNotify(ref Message m)
 
                 case (int)LVN.BEGINLABELEDITW:
                     {
+                        if (_labelEdit is not null)
```
Tanya-Solyanik, line 6423: Can our code recover from this condition? I think we could release that edit box and re-allocate it. Then the exception could be replaced with an Assert.

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
```diff
@@ -6462,6 +6476,14 @@ private unsafe void WmReflectNotify(ref Message m)
 
                 case (int)LVN.ENDLABELEDITW:
                     {
+                        if (_labelEdit is null)
+                        {
+                            throw new InvalidOperationException("There is no active label edit to end");
```
Tanya-Solyanik, line 6481: If there is no edit control, we could ignore this message and do an Assert instead of an exception.

Tanya-Solyanik:
@dmitrii-drobotov - the placement of edit box seems incorrect to me in the file explorer (and this PR). DGV placement looks more logical. However priority-wise this is not as high as other issues, since File Explorer had not fixed it in win11. Please open a new issue with this info - https://github.com/dotnet/winforms/pull/6309#issuecomment-1147464953 and we'll triage it.

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
dmitrii-drobotov: Yeah, it seems like we can do it. Do you know how I can test it to make sure it will be handled correctly? I can't quite trigger this case.

Tanya-Solyanik: I would set this field to null in the debugger 😄

Tanya-Solyanik: Let me know if you have time to get this in right now. If not, we can finish up this particular comment in NET8

src/System.Windows.Forms.Primitives/src/Interop/User32/Interop.WINEVENT.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+internal static partial class Interop
+{
+    internal static partial class User32
+    {
+        public enum WINEVENT : int
```
Tanya-Solyanik, line 9: This should be a `uint` because it will be used as the last parameter into https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setwineventhook which is a DWORD

Tanya-Solyanik (commented):
Please address comment about the uint in the enum type.

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
dmitrii-drobotov: Yes, I can take a look now.

dmitrii-drobotov:
> Please address comment about the uint in the enum type. When had this been tested by the CTIs, should it be re-tested or was the delta since then minor?

This PR actually hasn't been tested by the CTI yet. I'll make a request once I address these code review points.

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
dmitrii-drobotov: I managed to test it (it's a different case than setting to null - I needed to keep it *not null* instead) and it does works well. Handle can be released and then recreated without issues.

src/System.Windows.Forms.Primitives/src/Interop/User32/Interop.WINEVENT.cs
dmitrii-drobotov: Nice catch! Fixed.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewLabelEditNativeWindow.cs
```diff
@@ -0,0 +1,183 @@
...
+            _winEventProcCallback = new User32.WINEVENTPROC(WinEventProcCallback);
+            _valueChangeHook = User32.SetWinEventHook((uint)AccessibleEvents.ValueChange, _winEventProcCallback);
+            _textSelectionChangedHook = User32.SetWinEventHook(TextSelectionChanged, _winEventProcCallback);
+
+            _winEventHooksInstalled = true;
+        }
+
+        private bool IsAccessibilityObjectCreated => _accessibilityObject is not null;
+
+        public bool IsHandleCreated => Handle != IntPtr.Zero;
```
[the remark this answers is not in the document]
dmitrii-drobotov: Thanks for the insight! But yeah, I will keep it as is then.

dmitrii-drobotov:
There were 2 issues in the CTI test report, but both of them are related to Inspect not updating its UIA tree automatically and would be resolved after making a refresh. It's also the same behavior as in other Winforms controls with edit like TreeView or DataGridView and in Windows file explorer. Therefore I can say that both issues are expected and don't need to be fixed.

There are no other issues found, so this PR passes the testing.

Tanya-Solyanik (approved):

RussKie:
@artem-golubikhin this is causing the main to fail with
```
D:\a\_work\1\s\src\System.Windows.Forms\src\System\Windows\Forms\ListViewLabelEditUiaTextProvider.cs(12,20): error CS0534: 'ListViewLabelEditUiaTextProvider' does not implement inherited abstract member 'UiaTextProvider.RectangleToScreen(Rectangle)' [D:\a\_work\1\s\src\System.Windows.Forms\src\System.Windows.Forms.csproj]
##[error]src\System.Windows.Forms\src\System\Windows\Forms\ListViewLabelEditUiaTextProvider.cs(12,20): error CS0534: (NETCORE_ENGINEERING_TELEMETRY=Build) 'ListViewLabelEditUiaTextProvider' does not implement inherited abstract member 'UiaTextProvider.RectangleToScreen(Rectangle)'

Build FAILED.
```

PR #6328, opened 2021-12-11
Enable nullability in ArraySubsetEnumerator

gpetrou:
## Proposed changes

- Enable nullability in ArraySubsetEnumerator.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6328)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ArraySubsetEnumerator.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ArraySubsetEnumerator.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.Collections;
 using System.Diagnostics;
 
@@ -19,7 +17,7 @@ namespace System.Windows.Forms
         {
             Debug.Assert(count == 0 || array is not null, "if array is null, count should be 0");
             Debug.Assert(array is null || count <= array.Length, "Trying to enumerate more than the array contains");
-            _array = array;
+            _array = array!;
             _total = count;
             _current = -1;
         }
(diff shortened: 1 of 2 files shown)


dreddy-work (approved):

src/System.Windows.Forms/src/System/Windows/Forms/ArraySubsetEnumerator.cs
```diff
@@ -2,20 +2,18 @@
...
 {
     internal class ArraySubsetEnumerator : IEnumerator
     {
-        private readonly object[] _array;
+        private readonly object[]? _array;
         private readonly int _total;
         private int _current;
 
-        public ArraySubsetEnumerator(object[] array, int count)
+        public ArraySubsetEnumerator(object[]? array, int count)
```
RussKie, line 16: I would say that this is a non-nullable.
Also looking at how it is used - `ListViewSubItemCollection.GetEnumerator()` has a null-check before invoking this API, whilst `TreeNodeCollection.GetEnumerator()` doesn't - and this is a deficiency would need to address in the future.

RussKie (changes requested):

src/System.Windows.Forms/src/System/Windows/Forms/ArraySubsetEnumerator.cs
dreddy-work: We don't stop or check for null. `Movenext ` handle passing null value.

RussKie: This is an `internal` API so we control the input parameters. We also have an assertion that the array isn't null. We don't check however anywhere that the array's elements aren't null, though looking at how we use the API we never expect neither a null array or null elements.
The use of `!` is generally an indication of a code smell.

We could add a null check to the .ctor - `array.ThrowIfNull();`, but I don't really see a value in this.

dreddy-work (approved):

PR #6331, opened 2021-12-11
Enable nullability in WebBrowserHelper

gpetrou:
## Proposed changes

- Enable nullability in WebBrowserHelper

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6331)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/WebBrowserHelper.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/WebBrowserHelper.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.Collections.Specialized;
 using System.ComponentModel;
 using System.ComponentModel.Design;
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/WebBrowserHelper.cs
```diff
@@ -94,20 +91,12 @@ internal static int LogPixelsY
...
-                if (o is ISelectionService)
-                {
-                    return (ISelectionService)o;
-                }
-            }
+            ISelectionService? selectionService = ctl.Site?.GetService<ISelectionService>();
+            Debug.Assert(selectionService is null, "service must implement ISelectionService");
 
-            return null;
+            return selectionService;
```
RussKie, line 99: The change doesn't look equivalent. Currently we assert only if `ctl.Site is not null`. With this change we'll be asserting in other cases too.

dreddy-work: you are right. @gpetrou , would you mind  taking `ctl.Site` out and still keep simplification?

dreddy-work (approved):

PR #6334, opened 2021-12-13
Fixed unstable Integration tests

### LINKED ISSUE #6333: Some rewritten MAUI tests fails
* .NET Core Version: 
7.0.0-alpha.1.21576.4

No

**Problem description:**
Some Integration tests occasionally fails.
image

Danil-Andrianov:
Fixes #6333 

## Proposed changes

- Fixed two Integration Button tests
- Fixed one Integration MonthCalendar test
- Fixed one Integration TabControl test

## Regression? 

- Yes

## Risk

- Minimal

## Test methodology 

- Manual testing

## Test environment(s) 

- Microsoft Windows [Version 10.0.22000.318]
- .NET Core SDK: 7.0.0-alpha.1.21562.1


src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/MonthCalendarTests.cs
```diff
@@ -108,30 +108,6 @@ public async Task MonthCalendar_Click_Date_InvokeEventsAsync(int delta)
             });
         }
 
-        [WinFormsTheory]
-        [InlineData(-1)]
-        [InlineData(1)]
-        [InlineData(2)]
-        public async Task MonthCalendar_DoubleClick_Date_InvokeEventsAsync(int delta)
```
RussKie, line 115: Why is this test being removed?

Danil-Andrianov: This test didn't cover any potential bug, but required a lot of work to maintain all edge cases, so we decided just remove it.

RussKie: Please elaborate what edge cases you're referring to, and what maintenance is required.

Danil-Andrianov: As example, when we trying to switch to the date from previous month, it requires additional actions to switch the month respectively. It makes tests harder for understanding and requires a lot of investigation for editing both tests and helper methods.

vladimir-krestov: I would keep this test, but "skip" it. And fix all edge cases later as a special PR.

vladimir-krestov (commented):
Looks good.

ArtemTatarinov (approved):
LGTM ✔

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/MonthCalendarTests.cs
RussKie: I'm still not clear what problems you're facing. The test is passing on CI.

dreddy-work: @Danil-Andrianov , what are the expectations when selecting date from previous month? As i understand this is unstable when run during the months end/start days. We may still want a test to verify them. If you need to investigate further here, please open tests issue for it and link here. Also, add the linked issue in `skip `comment for future reference until it resolved.

dreddy-work:
Rest of it look good to me. Please rebase.

SergeySmirnov-Akvelon (commented):
LGTM.

RussKie:
Questions from https://github.com/dotnet/winforms/pull/6334#discussion_r770191552 are still outstanding.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/MonthCalendarTests.cs
Danil-Andrianov: Actually, situation is more complicated than it looks. MonthCalendar is native control and by default it's created with focus on the first month. Imagine that we want to select a date which is 100 days ago from today. It will require a lot of actions, which will make test difficult for understanding: the problem is that we can't display previous months by adjusting `CalendarDimensions` property, so me should switch months by yourself. Therefore, for sure, we firstly should switch to the month containing selected date from the past. It could be achieved by clicking on the arrow in the left corner (please, see the first picture; actually, for now I have now idea how to do that. It's another drawback of MonthCalendar native nature. ) or by pressing the PageDown button several times. In both cases we also will need the not elementary formula, which would compute number of months that we should omit.
And also we have the case when selected date is in previous month but still on the screen (please, see the second picture). 
Are those cases acceptable for just one test? And it's not clear for me why the testing of such cases is so important at all?
image
image

Danil-Andrianov: In a call we came to terms that this test should test concrete case without binding to the current date.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ButtonTests.cs
```diff
@@ -340,5 +381,34 @@ public async Task Button_Hotkey_DoesNotFire_OnClickAsync()
                 Assert.False(wasClicked);
             });
         }
+
+        public static Point GetCenter(Rectangle rect)
```
RussKie, line 385:
```suggestion
        private static Point GetCenter(Rectangle rect)
```

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/MonthCalendarTests.cs
```diff
@@ -15,6 +15,10 @@ namespace System.Windows.Forms.UITests
 {
     public class MonthCalendarTests : ControlTestBase
     {
+        private readonly DateTime _initialDate = new(2021, 12, 1);
```
RussKie, line 18:
```suggestion
        private static readonly DateTime CurrentDate = new(2021, 12, 1);
        private static readonly DateTime MaxDate = CurrentDate.AddDays(DaysAdd);
        private static readonly DateTime MinDate = CurrentDate.AddDays(-DaysSubtract);
```

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/MonthCalendarTests.cs
```diff
@@ -228,6 +214,30 @@ private async Task ClickOnDateTwiceAsync(Form form, MonthCalendar calendar, Date
...
+        {
+            await RunSingleControlTestAsync(
+                testDriverAsync: runTest,
+                createControl: () =>
+                {
+                    MonthCalendar control = new()
+                    {
+                        Location = new Point(0, 0),
+                        MinDate = _initialDate.AddDays(-DaysSubtract),
+                        MaxDate = _initialDate.AddDays(DaysAdd)
```
RussKie, line 227:
```suggestion
                        MinDate = MinDate,
                        MaxDate = MaxDate
```

RussKie (commented):
👍

dreddy-work:
Tests failed in this PR are also causing major issue in the build and CI process. @Danil-Andrianov /Team, can you guys take a look and prioritize it?. We may have to take it to 7.0 preview 1 as well if needed.

RussKie (approved):

PR #6262, opened 2021-12-03
Rewrite printPreviewDialog, label and DataGridView MAUI tests

Danil-Andrianov:
Fixes #3617

## Proposed changes

- Replaced DataGridViewMAUI tests into UI IntegrationTests project
- Replaced LabelMAUI tests into UI IntegrationTests project
- Replaced PrintPreviewDialogMAUI tests into UI IntegrationTests project

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6262)

PR #6335, opened 2021-12-13
Fix Form snap layout when in maximized state

### LINKED ISSUE #6153: Windows 11 snap layout does not work correctly when initiated from a maximized window
When the Windows 11 snap layout functionality is invoked, the window is not resized correctly when it is started from a maximized state.

* .NET Core Version: 

.net core 3.1

**Problem description:**

**Expected behavior:**

The window should be sized correctly.

**Minimal repro:**

- Create a new winforms project
- Maximize the window
- Hover over the maximize button and select one of the snap modes

jbhensley:
Fixes #6153. 

## Proposed changes

`System.Windows.Forms.Form` caches size and location upon becoming maximized and restores it when exiting maximized state. This creates a problem when using Windows snap layout since the form will be brought out of a maximized state and "docked" to a specific location by Windows only have its size and location subsequently changed to the internally cached values when the form processes `WM_WINDOWPOSCHANGED`.

~~If it were possible to not use internally cached values when coming out of a maximized state due to a snap layout command, this problem should be resolved. It does not appear that Windows provides any way to detect that. Instead, this PR assumes that any position change not initiated by the user or by a programmatic change of `Form.WindowState` is instead initiated by a window manager and cached positional values are not used.~~

~~Note: for determining if the change was initiated by the user, it was observed that a window message of `WM_SYSCOMMAND` containing a `wParam` of `SC_RESTORE` was sent regardless of whether the max/min button was clicked, the title bar was double-clicked, or if the restore command was used from the window's context menu. I was not able to find any way for the user to exit maximized state without triggering the `WM_SYSCOMMAND`. If that were possible, this fix would not work.~~

This PR fixes the issue by removing the unconditional caching of size/location that occurred when a form is maximized/minimized. By only attempting to set these values when necessary (e.g. a form is minimized and `Size` or `Location` is changed programmatically) it allows Windows to otherwise handle positioning.

The original attempt to fix left unconditional caching alone and only addressed the scenario of a form coming out of maximized state due to Windows snap layout. However, there were additional scenarios identified in which Windows attempted to set the geometry only to have been overridden by unconditionally cached values.

## Customer Impact

Customers should observe that `Form` objects respond properly when using snap layout while in a maximized state.

## Regression? 

- No

## Risk

~~Risk is that manner of determining whether Form position change is due to Windows snap layout versus user/programmatic change is deficient and does not cover all scenarios.~~

- It is unclear why form geometry was unconditionally cached and restored. There is risk that the original author had an presently unidentified edge case in mind. This change should be tested thoroughly.

## Test methodology

- Unit tests
- UI integration tests
- Manually manipulated a `Form` object with the mouse and keyboard to verify behavior

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6335)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
@@ -2382,7 +2382,6 @@ namespace System.Windows.Forms
 
                 if (IsHandleCreated && Visible)
                 {
-                    IntPtr hWnd = Handle;
                     switch (value)
                     {
                         case FormWindowState.Normal:
@@ -5931,12 +5930,15 @@ namespace System.Windows.Forms
                         SuspendLayoutForMinimize();
                     }
 
-                    _restoredWindowBounds.Size = ClientSize;
-                    _formStateEx[FormStateExWindowBoundsWidthIsClientSize] = 1;
-                    _formStateEx[FormStateExWindowBoundsHeightIsClientSize] = 1;
-                    _restoredWindowBoundsSpecified = BoundsSpecified.Size;
-                    _restoredWindowBounds.Location = Location;
-                    _restoredWindowBoundsSpecified |= BoundsSpecified.Location;
+                    if (!OsVersion.IsWindows11_OrGreater)
+                    {
+                        _restoredWindowBounds.Size = ClientSize;
+                        _formStateEx[FormStateExWindowBoundsWidthIsClientSize] = 1;
+                        _formStateEx[FormStateExWindowBoundsHeightIsClientSize] = 1;
+                        _restoredWindowBoundsSpecified = BoundsSpecified.Size;
+                        _restoredWindowBounds.Location = Location;
+                        _restoredWindowBoundsSpecified |= BoundsSpecified.Location;
+                    }
 
                     // stash off restoreBounds As well...
                     _restoreBounds.Size = Size;
--- a/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FormTests.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FormTests.cs
@@ -1,0 +1,157 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Drawing;
+using WindowsInput.Native;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace System.Windows.Forms.UITests
+{
+    public class FormTests : ControlTestBase
+    {
+        // When using the keyboard for snap layout menu, there are various times when
+        // a delay is needed. This value may need to be adjusted if tests fail
+        // in CI/different environment
+        private const int SnapLayoutDelayMS = 500;
+
+        public FormTests(ITestOutputHelper testOutputHelper)
+            : base(testOutputHelper)
+        {
+        }
+
+        [WinFormsTheory]
+        [InlineData(FormWindowState.Normal)]
+        [InlineData(FormWindowState.Maximized)]
+        public async Task Form_SnapsLeftAsync(FormWindowState windowState)
+        {
+            if (!OsVersion.IsWindows11_OrGreater)
+            {
+                return;
+            }
+
+            await RunEmptyFormTestAsync(async form =>
+            {
+                form.Location = new Point(20, 21);
+                form.Size = new Size(300, 310);
+
+                form.WindowState = windowState;
+
+                await InputSimulator.SendAsync(
+                    form,
+                    inputSimulator => inputSimulator.Keyboard.ModifiedKeyStroke(VirtualKeyCode.LWIN, VirtualKeyCode.VK_Z));
+
+                // inputSimulator.Sleep appears wildly inconsistent with snap panel timing. Task.Delay does not
+                await Task.Delay(SnapLayoutDelayMS);
+
+                // Snap left
+                await InputSimulator.SendAsync(
+                    form,
+                    inputSimulator => inputSimulator.Keyboard.KeyPress(VirtualKeyCode.RIGHT)
+                                                             .KeyPress(VirtualKeyCode.RETURN));
+
+                await Task.Delay(SnapLayoutDelayMS);
+
+                // At this point, Windows displays a panel containing all running applications so the
+                // user can select one to dock next to our form. It also takes the keyboard focus away.
+                // If left in this state, subsequently run tests will fail since keyboard focus is not
+                // given to any newly launched window until this panel is dismissed.
+                await InputSimulator.SendAsync(
+                    form,
+                    inputSimulator => inputSimulator.Keyboard.KeyPress(VirtualKeyCode.ESCAPE));
+
+                await Task.Delay(SnapLayoutDelayMS);
+
+                var screenWorkingArea = Screen.FromControl(form).WorkingArea;
+                int borderSize = (form.Width - form.ClientRectangle.Width) / 2;
+
+                Assert.True(form.Left <= screenWorkingArea.X);
+                Assert.True(form.Left >= screenWorkingArea.X - borderSize);
+
+                Assert.True(form.Top <= screenWorkingArea.Y);
+                Assert.True(form.Top >= screenWorkingArea.Y - borderSize);
+
+                Assert.True(form.Height >= screenWorkingArea.Height);
+                Assert.True(form.Height <= screenWorkingArea.Height + (borderSize * 2));
+
+                Assert.True(form.Width >= screenWorkingArea.Width / 2);
+                Assert.True(form.Width <= (screenWorkingArea.Width / 2) + (borderSize * 2));
+            });
+        }
+
+        [WinFormsTheory]
+        [InlineData(FormWindowState.Normal)]
+        [InlineData(FormWindowState.Maximized)]
+        public async Task Form_SnapsRightAsync(FormWindowState windowState)
+        {
+            if (!OsVersion.IsWindows11_OrGreater)
+            {
+                return;
+            }
+
+            await RunEmptyFormTestAsync(async form =>
+            {
+                form.Location = new Point(20, 21);
+                form.Size = new Size(300, 310);
+
+                form.WindowState = windowState;
+
+                await InputSimulator.SendAsync(
+                    form,
+                    inputSimulator => inputSimulator.Keyboard.ModifiedKeyStroke(VirtualKeyCode.LWIN, VirtualKeyCode.VK_Z));
+
+                // inputSimulator.Sleep appears wildly inconsistent with snap panel timing. Task.Delay does not
+                await Task.Delay(SnapLayoutDelayMS);
+
+                // Snap right
+                await InputSimulator.SendAsync(
+                    form,
+                    inputSimulator => inputSimulator.Keyboard.KeyPress(VirtualKeyCode.RIGHT)
+                                                             .KeyPress(VirtualKeyCode.RIGHT)
+                                                             .KeyPress(VirtualKeyCode.RETURN));
+
+                await Task.Delay(SnapLayoutDelayMS);
+
+                // At this point, Windows displays a panel containing all running applications so the
+                // user can select one to dock next to our form. It also takes the keyboard focus away.
+                // If left in this state, subsequently run tests will fail since keyboard focus is not
+                // given to any newly launched window until this panel is dismissed.
+                await InputSimulator.SendAsync(
+                    form,
+                    inputSimulator => inputSimulator.Keyboard.KeyPress(VirtualKeyCode.ESCAPE));
+
+                await Task.Delay(SnapLayoutDelayMS);
+
+                var screenWorkingArea = Screen.FromControl(form).WorkingArea;
+                int screenMiddleX = screenWorkingArea.X + (screenWorkingArea.Width / 2);
+                int borderSize = (form.Width - form.ClientRectangle.Width) / 2;
+
+                Assert.True(form.Left <= screenMiddleX);
+                Assert.True(form.Left >= screenMiddleX - borderSize);
+
+                Assert.True(form.Top <= screenWorkingArea.Y);
+                Assert.True(form.Top >= screenWorkingArea.Y - borderSize);
+
+                Assert.True(form.Height >= screenWorkingArea.Height);
+                Assert.True(form.Height <= screenWorkingArea.Height + (borderSize * 2));
+
+                Assert.True(form.Width >= screenWorkingArea.Width / 2);
+                Assert.True(form.Width <= (screenWorkingArea.Width / 2) + (borderSize * 2));
+            });
+        }
+
+        private async Task RunEmptyFormTestAsync(Func<Form, Task> testDriverAsync)
+        {
+            await RunFormWithoutControlAsync(
+                () =>
+                {
+                    var form = new Form();
+                    form.TopMost = true;
+
+                    return form;
+                },
+                testDriverAsync);
+        }
+    }
+}
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
@@ -975,6 +975,64 @@ namespace System.Windows.Forms.Tests
             Assert.False(control.IsHandleCreated);
         }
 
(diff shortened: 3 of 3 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
```diff
@@ -2384,10 +2385,10 @@ public FormWindowState WindowState
 
                 if (IsHandleCreated && Visible)
                 {
-                    IntPtr hWnd = Handle;
                     switch (value)
                     {
                         case FormWindowState.Normal:
+                            formStateEx[FormStateExUserOrProgrammaticRestore] = 1;
```
Tanya-Solyanik, line 2391: Consider adding dedicates Get/Set state methods to improve readability.

Tanya-Solyanik (commented):
Looks reasonable. On what versions of windows had it been tested? WIn11/10/7?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
```diff
@@ -975,6 +975,78 @@ public void Form_Parent_SetWithNonNullOldParent_GetReturnsExpected(Control value
...
+
+            Assert.Equal(FormWindowState.Maximized, form.WindowState);
+            Assert.NotEqual(new Point(20, 21), form.Location);
+            Assert.NotEqual(new Size(300, 310), form.Size);
+
+            form.WindowState = FormWindowState.Normal;
+
+            Assert.Equal(new Point(20, 21), form.Location);
+            Assert.Equal(new Size(300, 310), form.Size);
+        }
```
RussKie, line 1048: Nice!

Now I wonder if we could write integration tests (UIIntegrationTests) that would interact with the new snap layout options.

.gitignore
```diff
@@ -375,3 +375,4 @@ Temporary Items
 InteropTests/NativeTests/out
 /src/System.Windows.Forms/src/comctl32.dll
 *.received.*
+/src/System.Windows.Forms/tests/InteropTests/NativeTests/out/build/x64-Debug
```
RussKie, line 378: Is line:375 not working?

jbhensley:
> Looks reasonable. On what versions of windows had it been tested? WIn11/10/7?

Tested:

- Win10 21H1 (19043.1415)
- Win11 21H2 (22000.376)

.gitignore
jbhensley: Sorry, not sure how that got in there. Reverted

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
jbhensley: Interesting. Seems like we'd need to know where to position the mouse. I tried calculating center of the maximize button using `SystemInformation.CaptionButtonSize` in conjunction with `Width` and then with `ClientRectangle.Width`, but the numbers don't quite come out right either way. The crude method I'm using is below.

``` C#
protected override void OnPreviewKeyDown(PreviewKeyDownEventArgs e)
{
	base.OnPreviewKeyDown(e);

	if (e.KeyCode != Keys.A)
		return;

	var builder = new StringBuilder();
	builder.AppendLine(PointToClient(MousePosition).ToString());
	builder.AppendLine($"\tWidth: {Width}");
	builder.AppendLine($"\tClientRectangle: {ClientRectangle.Width}");
	builder.AppendLine($"\tButton width: {SystemInformation.CaptionButtonSize.Width}");
	builder.AppendLine($"\tCenter from width: {Width - 8 - (SystemInformation.CaptionButtonSize.Width * 1.5)}");
	builder.AppendLine($"\tCenter from client width: {ClientRectangle.Width - 8 - (SystemInformation.CaptionButtonSize.Width * 1.5)}");

	Debug.WriteLine(builder.ToString());
}
```
With my cursor reasonably centered:

image

I get:

{X=730,Y=-17}
	Width: 816
	ClientRectangle: 800
	Button width: 36
	Center from width: 754
	Center from client width: 738

I'm afraid things would get worse from there when trying to move to the layouts. Of course, I could be going about this all wrong.

jbhensley: Quick follow up. The snap layout options panel has built in delays. I think that would be the biggest hurtle. 

As far as mouse, it is possible to get needed mouse position to hover over the max button by sending `WM_GETTITLEBARINFOEX` message to the window. Then you have to account for the hover delay. Positioning from there to the layout you want might be tricky.

For keyboard, you can send Win+Z to open the menu and navigate with arrow keys. I started down this route, but also encountered timing issues.

``` C#
[WinFormsFact]
public async Task Form_RespondsToSnapLayoutAsync()
{
	await RunEmptyFormTestAsync(async form =>
	{
		form.Location = new Point(20, 21);
		form.Size = new Size(300, 310);

		// Windows Key + Z opens the snap layout menu
		await InputSimulator.SendAsync(
			form,
			inputSimulator => inputSimulator.Keyboard.ModifiedKeyStroke(VirtualKeyCode.LWIN, VirtualKeyCode.VK_Z));

		// 65 or less and the snap layout panel is not yet ready for keyboard input
		await Task.Delay(66);

		// Right arrow selects the first snap layout (dock to left)
		await InputSimulator.SendAsync(
			form,
			inputSimulator => inputSimulator.Keyboard.KeyPress(VirtualKeyCode.RIGHT));

		// Press enter to snap
		await InputSimulator.SendAsync(
			form,
			inputSimulator => inputSimulator.Keyboard.KeyPress(VirtualKeyCode.RETURN));

		// Pause here so we can visually observe the form
		await Task.Delay(5000);
	});
}
```

I'm concerned that the needed delay may vary.

`RunEmptyFormTestAsync` is a method I added to `ControlTestBase` just for this.

RussKie:
Nice! Thank you for adding such tests.
I don't think we need to exercise all of them, one or two will be sufficient to ensure it is working in general.

RE: timing - I think it's totally acceptable to account for a delay. IIRC the InputSimulator has an extension to add a wait fluently. I think I saw it in some other test.

I'm currently OOF, and won't be able to formally review your change for a few weeks.

jbhensley:
Added a couple of snap layout UI integration tests. They do feel a bit fragile, but are passing on my box.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
dreddy-work: May be we should rely on windows notification here. Have you noticed any windows messages/notifications when snap layout pops up?

jbhensley: Doesn't look like it.

I modified `RunEmptyFormTestAsync` to be generic so I could use a custom form that dumps `WndProc` to the debugger. Changing the test code a bit to:

``` C#
form.LogMessagesToDebug = true;
Diagnostics.Debug.WriteLine($"{DateTime.Now.Millisecond}: logging on");

await InputSimulator.SendAsync(
	form,
	inputSimulator => inputSimulator.Keyboard.ModifiedKeyStroke(VirtualKeyCode.LWIN, VirtualKeyCode.VK_Z));

Diagnostics.Debug.WriteLine($"{DateTime.Now.Millisecond}: input sent");

// inputSimulator.Sleep appears wildly inconsistent with snap panel timing. Task.Delay does not
await Task.Delay(SnapLayoutDelayMS);

form.LogMessagesToDebug = false;
Diagnostics.Debug.WriteLine($"{DateTime.Now.Millisecond}: logging off");
```

I get:

MS | msgid | msg/descr
-- | -- | --
376 |   | logging on
383 | 70 | WM_WINDOWPOSCHANGING
392 | 70 | WM_WINDOWPOSCHANGING
396 | 133 | WM_NCPAINT
399 | 20 | WM_ERASEBKGND
404 | 14 | WM_GETTEXTLENGTH
407 | 13 | WM_GETTEXT
411 | 71 | WM_WINDOWPOSCHANGED
415 | 15 | WM_PAINT
421 | 14 | WM_GETTEXTLENGTH
423 | 13 | WM_GETTEXT
441 | 135 | WM_GETDLGCODE
450 | 135 | WM_GETDLGCODE
455 | 256 | WM_KEYFIRST
464 | 257 | WM_KEYUP
468 | 257 | WM_KEYUP
477 |   | input sent
487 | 134 | WM_NCACTIVATE
491 | 6 | WM_ACTIVATE
493 | 28 | WM_WININICHANGE
495 | 8 | WM_KILLFOCUS
498 | 14 | WM_GETTEXTLENGTH
500 | 13 | WM_GETTEXT
502 | 14 | WM_GETTEXTLENGTH
505 | 13 | WM_GETTEXT
516 | 641 | WM_IME_SETCONTEXT
523 | 642 | WM_IME_NOTIFY
995 |   | logging off

So there's only 46 milliseconds between when the code resumes execution from sending the key sequence and when the last `WndProc` msg is received. The smallest workable delay I've observed was 66 ms.

src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
```diff
@@ -1204,6 +1205,22 @@ public bool IsMdiContainer
...
+        {
+            get
+            {
+                return formStateEx[FormStateExUserOrProgrammaticRestore] == 1;
+            }
+            set
+            {
+                formStateEx[FormStateExUserOrProgrammaticRestore] = value ? 1 : 0;
+            }
+        }
```
RussKie, line 1222:
```suggestion
        private bool IsUserOrProgrammaticRestore
        {
            get => formStateEx[FormStateExUserOrProgrammaticRestore] == 1;
            set => formStateEx[FormStateExUserOrProgrammaticRestore] = value ? 1 : 0;
        }
```

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FormTests.cs
```diff
@@ -0,0 +1,141 @@
...
+                    form,
+                    inputSimulator => inputSimulator.Keyboard.ModifiedKeyStroke(VirtualKeyCode.LWIN, VirtualKeyCode.VK_Z));
+
+                // inputSimulator.Sleep appears wildly inconsistent with snap panel timing. Task.Delay does not
+                await Task.Delay(SnapLayoutDelayMS);
+
+                // Snap right
+                await InputSimulator.SendAsync(
+                    form,
+                    inputSimulator => inputSimulator.Keyboard.KeyPress(VirtualKeyCode.RIGHT).KeyPress(VirtualKeyCode.RIGHT).KeyPress(VirtualKeyCode.RETURN));
```
RussKie, line 109: Makes it easier to read
```suggestion
                    inputSimulator => inputSimulator.Keyboard.KeyPress(VirtualKeyCode.RIGHT)
                                                             .KeyPress(VirtualKeyCode.RIGHT)
                                                             .KeyPress(VirtualKeyCode.RETURN));
```

ditto in other places

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/ControlTestBase.cs
```diff
@@ -146,6 +146,19 @@ protected async Task MoveMouseAsync(Form window, Point point, bool assertCorrect
...
+            await RunFormAsync(
+                () =>
+                {
+                    var form = new Form();
+                    form.TopMost = true;
+
+                    return form;
+                },
+                testDriverAsync);
+        }
```
RussKie, line 160: Move it in `FormTests`

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/ControlTestBase.cs
```diff
@@ -215,6 +228,37 @@ protected async Task RunControlPairTestAsync<T>(Func<Form, (T control1, T contro
                 testDriverAsync);
         }
 
+        protected async Task RunFormAsync(Func<Form> createDialog, Func<Form, Task> testDriverAsync)
```
RussKie, line 231: We have another PR in flight that uses the same method https://github.com/dotnet/winforms/pull/6262/files#diff-b701242799ecc2572782b1786e9da6de99a8b2894ee42a6e96bf472abede001a

RussKie (commented):
🚀

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/ControlTestBase.cs
jbhensley: Awesome. I can wait for that PR to merge and then use `RunFormWithoutControlAsync` introduced there. Looks like it might be about ready.

Let me know if you want me to keep this PR rebased to a single commit or if it just gets squashed on merge to `main`.

Lastly, no pressure to review if your still OOF

jbhensley: Pulled changes from #6262 and refactored

RussKie:
Overall it is working much better, though I seem to have a scenario where the original form geometry isn't getting restored

1. Maximise a .NET form
2. Select one of the snap layouts and fill the other positions.
3. If we drag the form from its placement the form retains the original (i.e. start geometry).

If instead of 3\. we maximise the form again and then drag it - the form will be restored to the snap layout's geometry, not to the original.

I'll add a recording later.

jbhensley:
Ah, yes, that makes sense. The form caches geometry when it becomes maximized, so the snap position was the last known geometry at the time.

Maybe this should be set about a different way. The purpose for caching is apparently to allow size/location change while not in normal state, but yet size/location is cached unconditionally upon maximize (and minimize):

https://github.com/dotnet/winforms/blob/6fb22da6bb1468c46252ce612424dc2940eec632/src/System.Windows.Forms/src/System/Windows/Forms/Form.cs#L4748-L4754

https://github.com/dotnet/winforms/blob/6fb22da6bb1468c46252ce612424dc2940eec632/src/System.Windows.Forms/src/System/Windows/Forms/Form.cs#L5918-L5920

Perhaps this logic should be smarter. Windows does indeed handle restoring position of native windows, including in the scenario you have outlined. This can be tested with Notepad. My original attempt was to be minimally invasive and provide the least opportunity to break something. At this point, though, it feels like we're fighting the window manager. I can explore the alternative approach.

jbhensley:
I have updated the approach and changed the PR description to match. 

This feels more like the "right" way in that we let Windows do what it does and only take over when necessary. That said, it also feels more risky since the original author obviously had something in mind when they added the code that has been removed. Even though unconditionally caching geometry seems wrong and is definitely the source of our snap layout problems, there may have been a reason for it. I don't suppose there is any way we can access history for Framework 4.8 and earlier to get an idea why this was there?

jbhensley:
I'm a lowly peon. I don't suppose `sudo /azp run` would work 😆

RussKie:
> I'm a lowly peon. I don't suppose `sudo /azp run` would work 😆

No, sorry. It requires WRITE access to re-run the build this way. Mere mortals need to close/re-open PRs to restart the build. :)

RussKie:
Apologies for the delay, we've not forgotten about this one.

RussKie (commented):
I think we should retain the original behaviour for all OSes, and only disable it for Win11.

jbhensley:
That's probably safer, with any potential side effect of the change only affecting a subset of users. Code updated to remove unconditional caching of geometry for Win11 or greater.

RussKie:
@jbhensley apologies, it fell of my radar. Do you mind rebasing on top of the latest main, so we can do a test pass over this change before we commit it. Thank you.

dreddy-work:
@Olina-Zhang , can you help validate this change before we commit?

Olina-Zhang:
> @Olina-Zhang , can you help validate this change before we commit?

Will test it today.

John-Qiao:
@dreddy-work Tested the built private binaries for this pull based on latest .NET 7.0 SDK, now snap layout works correctly when initiated from the maximized window, no new issue found during manual testing. But the migrated automation cases blocked by the known issue: [6952](https://github.com/dotnet/winforms/issues/6952), we will re-run the migrated automation cases once that known issue be fixed and update result here.

RussKie:
Thank you @John-Qiao, #6952 is likely caused by an unrelated change.

RussKie (approved):

RussKie:
Thank you @jbhensley for you work and your patience.

Issue #4876, opened 2021-05-07
Inspect can't find PropertyGrid's highlighted dropdown list in the accessibility tree


SergeySmirnov-Akvelon:
**.NET Core Version:**
- 6.0.100-preview.3.21202.5

**Have you experienced this same bug with .NET Framework?:**
- Yes

**Minimal repro:**
1. Create an app with a PropertyGrid and button.
2. Set button as "SelectedObject" property of PropertyGrid
3. Run the app
4. Run  the Inspect
5. Expand the "AccessibleRole" dropdown list and highlight the drop-down list using the Inspect

**Actual behavior:**
The highlighted drop-down list is not displayed in the accessibility tree.
![image](https://user-images.githubusercontent.com/23376742/117441815-9d7d4d80-af3e-11eb-8e62-e10c914b4a2a.png)

**Expected behavior:**
The highlighted drop-down list should be displayed in the accessibility tree.

vladimir-krestov:
Looks like there is an incorrect accessibility tree. "Available values" list and DropDownHolder are different components.
DropDownHolder includes the values list. 
- So the first way is to add a new "DropDownHolder" layer (red in the pic), like this:
![image](https://user-images.githubusercontent.com/49272759/141472510-82eea00f-c203-400e-9bd4-a16939932f19.png)
- The second way is to change HitTest logic (get element from mouse point) and the dropped list box Bounds, when mouse hovers DropDownHolder we may return GridViewListBoxAO instead of DropDownHolderAO, like that:
![AR0X7imhBk](https://user-images.githubusercontent.com/49272759/141612743-77653d33-0e24-498c-82d9-74a7bbfc5677.gif)

@dreddy-work, which way looks more acceptable?

Tanya-Solyanik:
Could you please elaborate on these options? 
>So the first way is to add a new "DropDownHolder" layer (red in the pic), like this:

Where are you suggesting to add DroppDownHolder to? Are you suggesting to return it's accessibility object from the navigation APIs?

The second tree looks cleaner, and might be preferable to user experience. But I would think that navigation interfces would need to change as well to ensure parent-child link between the propertygrid entry and the list of available items.

What would you prefer?

vladimir-krestov:
[...]

PR #6341, opened 2021-12-14
Fixing PGV accessibility tree for property descriptors and dropdown holders

### LINKED ISSUE #2848: [Inspect]There is no "Font" edit item below "Font" tree item node when Font property is Expanded
-  .NET Core Version: .Net Core 3.1.200-preview-014978 from 3.1 Release branch + VS2019 with latest core designer
- Have you experienced this same bug with .NET Framework?: Yes

**More Info:**
1.	This issue can reproduce on Font, Location, Margin, Size… properties.
2.	This issue can also reproduce on Framework project with latest .Net 4.8.
3.	Using Inspect to focus Font property, there is "Font" edit item in the first time for collapse status, no "Font" edit item in the second time. Please see the result of Step5.

**Steps to reproduce:**
1.	Open the attached “Font_Inspect_issue.zip” in VS. 
[Font_Inspect_issue.zip](https://github.com/dotnet/winforms/files/4191735/Font_Inspect_issue.zip)
2.	Build and run this application, select the PropertyGrid control. 
3.	Open Inspect and focus on the Font property, observe the inspect behavior. 
4.	Expand the Font property and observe the inspect behavior. 
5.	Collapse the Font property and observe the inspect behavior.

**Actual behavior:**
After step 3, there is "Font" edit item below "Font" tree item node.
Step3
After step 4, there is no "Font" edit item below "Font" tree item node.
Step4
After step 5, there is no "Font" edit item below "Font" tree item node. Please see the Gif:
Font_expand_collapse_issue

**Expected behavior:**
After step 4, there should be "Font" edit item below "Font" tree item node.
After step 5, there should be "Font" edit item below "Font" tree item node, the result should be same as step3.

WPMGPRoSToTeMa:
Resolves dotnet#2848 by fixing sibiling relationship between child property descriptors and selected entry controls. Also resolves dotnet#4876 by fixing parent-child relationship between drop-down holder and listbox.

The fixes are implemented by adding helper methods to retrieve a sibling node for a specified child node. This implementation removes the need for copy-paste throughout different child nodes of property descriptor entry, it greatly simplifies the implementation and helps to avoid mistakes in future because the implementation is now located in a single place.

## Proposed changes

- Fix sibling relationship between child property descriptors and selected entry controls
- Fix parent-child relationship between dropdown holder and listbox

## Customer Impact

- Accessibility tools may incorrectly navigate through the descriptor entry children (controls, other entries) before this PR

## Regression? 

- No

## Risk

- Minimal

## Test methodology 

- CTI
- Manual testing
- Unit-tests

 

## Test environment(s) 

- Microsoft Windows [Version 10.0.19044.1348]
- .NET 7.0.0-alpha.1.21602.1

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6341)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/DropDownButton.DropDownButtonAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/DropDownButton.DropDownButtonAccessibleObject.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using static System.Windows.Forms.PropertyGridInternal.PropertyDescriptorGridEntry;
 using static Interop;
 
 namespace System.Windows.Forms.PropertyGridInternal
@@ -47,18 +48,19 @@ namespace System.Windows.Forms.PropertyGridInternal
             /// <returns>Returns the element in the specified direction.</returns>
             internal override UiaCore.IRawElementProviderFragment? FragmentNavigate(UiaCore.NavigateDirection direction)
             {
-                if (direction == UiaCore.NavigateDirection.Parent &&
-                    _owningPropertyGrid?.SelectedGridEntry is not null &&
-                    _owningDropDownButton.Visible)
+                if (!_owningDropDownButton.Visible
+                    || _owningPropertyGrid?.SelectedGridEntry?.AccessibilityObject is not PropertyDescriptorGridEntryAccessibleObject parent)
                 {
-                    return _owningPropertyGrid.SelectedGridEntry?.AccessibilityObject;
-                }
-                else if (direction == UiaCore.NavigateDirection.PreviousSibling)
-                {
-                    return _owningPropertyGrid?.EditAccessibleObject;
+                    return null;
                 }
 
-                return base.FragmentNavigate(direction);
+                return direction switch
+                {
+                    UiaCore.NavigateDirection.Parent => parent,
+                    UiaCore.NavigateDirection.NextSibling => parent.GetNextChild(this),
+                    UiaCore.NavigateDirection.PreviousSibling => parent.GetPreviousChild(this),
+                    _ => base.FragmentNavigate(direction),
+                };
             }
 
             /// <summary>
--- a/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyDescriptorGridEntry.PropertyDescriptorGridEntryAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyDescriptorGridEntry.PropertyDescriptorGridEntryAccessibleObject.cs
@@ -2,13 +2,14 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Diagnostics;
 using static Interop;
 
 namespace System.Windows.Forms.PropertyGridInternal
 {
     internal partial class PropertyDescriptorGridEntry
     {
-        protected class PropertyDescriptorGridEntryAccessibleObject : GridEntryAccessibleObject
+        internal class PropertyDescriptorGridEntryAccessibleObject : GridEntryAccessibleObject
         {
             private readonly PropertyDescriptorGridEntry _owningPropertyDescriptorGridEntry;
 
@@ -38,6 +39,109 @@ namespace System.Windows.Forms.PropertyGridInternal
                 }
             }
 
+            public override AccessibleObject? GetChild(int index)
+            {
+                Debug.Assert(index >= 0);
+
+                // Child controls exist in tree only when the entry is selected.
+                if (GetPropertyGridView() is { } propertyGridView && propertyGridView.SelectedGridEntry == _owningPropertyDescriptorGridEntry)
+                {
+                    // DropDownControlHolder exists in the tree if the drop-down holder is visible.
+                    if (propertyGridView.DropDownVisible)
+                    {
+                        if (index == 0)
+                        {
+                            return propertyGridView.DropDownControlHolder.AccessibilityObject;
+                        }
+
+                        index--;
+                    }
+
+                    // TextBox exists in the tree if it's created.
+                    if (propertyGridView.IsEditTextBoxCreated)
+                    {
+                        if (index == 0)
+                        {
+                            return propertyGridView.EditAccessibleObject;
+                        }
+
+                        index--;
+                    }
+
+                    // DropDownButton exists in the tree if the drop-down button is visible.
+                    if (propertyGridView.DropDownButton is { Visible: true } dropDownButton)
+                    {
+                        if (index == 0)
+                        {
+                            return dropDownButton.AccessibilityObject;
+                        }
+
+                        index--;
+                    } // DialogButton exists in the tree if the ellipsis button is visible.
+                    else if (propertyGridView.DialogButton is { Visible: true } dialogButton)
+                    {
+                        if (index == 0)
+                        {
+                            return dialogButton.AccessibilityObject;
+                        }
+
+                        index--;
+                    }
+                }
+
+                // Child entries exist in the tree if the entry has child entries and is expanded.
+                if (_owningPropertyDescriptorGridEntry is { ChildCount: > 0, Expanded: true })
+                {
+                    if (index < _owningPropertyDescriptorGridEntry.ChildCount)
+                    {
+                        return _owningPropertyDescriptorGridEntry.Children[index].AccessibilityObject;
+                    }
+
+                    // Uncomment the following line in case there will be another children going after the child entries:
+                    // index -= _owningPropertyDescriptorGridEntry.ChildCount;
+                }
+
+                return null;
+            }
+
+            public override int GetChildCount()
+            {
+                int count = 0;
+                // Child controls exist in tree only when the entry is selected.
+                if (GetPropertyGridView() is { } propertyGridView && propertyGridView.SelectedGridEntry == _owningPropertyDescriptorGridEntry)
+                {
+                    // DropDownControlHolder exists in the tree if the drop-down holder is visible.
+                    if (propertyGridView.DropDownVisible)
+                    {
+                        count++;
+                    }
+
+                    // TextBox exists in the tree if it's created.
+                    if (propertyGridView.IsEditTextBoxCreated)
+                    {
+                        count++;
+                    }
+
+                    // DropDownButton exists in the tree if the drop-down button is visible.
+                    if (propertyGridView.DropDownButton.Visible)
+                    {
+                        count++;
+                    } // DialogButton exists in the tree if the ellipsis button is visible.
+                    else if (propertyGridView.DialogButton.Visible)
+                    {
+                        count++;
+                    }
+                }
+
+                // Child entries exist in the tree if the entry has child entries and is expanded.
+                if (_owningPropertyDescriptorGridEntry.Expanded)
+                {
+                    count += _owningPropertyDescriptorGridEntry.ChildCount;
+                }
+
+                return count;
+            }
+
             internal override void Collapse()
             {
                 if (ExpandCollapseState == UiaCore.ExpandCollapseState.Expanded)
@@ -60,24 +164,114 @@ namespace System.Windows.Forms.PropertyGridInternal
             /// <param name="direction">Indicates the direction in which to navigate.</param>
             /// <returns>Returns the element in the specified direction.</returns>
             internal override UiaCore.IRawElementProviderFragment? FragmentNavigate(UiaCore.NavigateDirection direction)
-            {
-                switch (direction)
+                => direction switch
                 {
-                    case UiaCore.NavigateDirection.NextSibling:
-                        var propertyGridViewAccessibleObject = (PropertyGridView.PropertyGridViewAccessibleObject?)Parent;
-                        var propertyGridView = propertyGridViewAccessibleObject?.Owner as PropertyGridView;
-                        return propertyGridViewAccessibleObject?.GetNextGridEntry(_owningPropertyDescriptorGridEntry, propertyGridView?.TopLevelGridEntries, out _);
-                    case UiaCore.NavigateDirection.PreviousSibling:
-                        propertyGridViewAccessibleObject = (PropertyGridView.PropertyGridViewAccessibleObject?)Parent;
-                        propertyGridView = propertyGridViewAccessibleObject?.Owner as PropertyGridView;
-                        return propertyGridViewAccessibleObject?.GetPreviousGridEntry(_owningPropertyDescriptorGridEntry, propertyGridView?.TopLevelGridEntries, out _);
-                    case UiaCore.NavigateDirection.FirstChild:
-                        return GetFirstChild();
-                    case UiaCore.NavigateDirection.LastChild:
-                        return GetLastChild();
+                    UiaCore.NavigateDirection.NextSibling => GetNextSibling(),
+                    UiaCore.NavigateDirection.PreviousSibling => GetPreviousSibling(),
+                    UiaCore.NavigateDirection.FirstChild => GetFirstChild(),
+                    UiaCore.NavigateDirection.LastChild => GetLastChild(),
+                    _ => base.FragmentNavigate(direction),
+                };
+
+            internal override int GetChildIndex(AccessibleObject? child)
+            {
+                Debug.Assert(child is not null);
+
+                int index = 0;
+                // Child controls exist in tree only when the entry is selected.
+                if (GetPropertyGridView() is { } propertyGridView && propertyGridView.SelectedGridEntry == _owningPropertyDescriptorGridEntry)
+                {
+                    // DropDownControlHolder exists in the tree if the drop-down holder is visible.
+                    if (propertyGridView.DropDownVisible)
+                    {
+                        if (child == propertyGridView.DropDownControlHolder.AccessibilityObject)
+                        {
+                            return index;
+                        }
+
+                        index++;
+                    }
+
+                    // TextBox exists in the tree if it's created.
+                    if (propertyGridView.IsEditTextBoxCreated)
+                    {
+                        if (child == propertyGridView.EditAccessibleObject)
+                        {
+                            return index;
+                        }
+
+                        index++;
+                    }
+
+                    // DropDownButton exists in the tree if the drop-down button is visible.
+                    if (propertyGridView.DropDownButton is { Visible: true } dropDownButton)
+                    {
+                        if (child == dropDownButton.AccessibilityObject)
+                        {
+                            return index;
+                        }
+
+                        index++;
+                    } // DialogButton exists in the tree if the ellipsis button is visible.
+                    else if (propertyGridView.DialogButton is { Visible: true } dialogButton)
+                    {
+                        if (child == dialogButton.AccessibilityObject)
+                        {
+                            return index;
+                        }
+
+                        index++;
+                    }
                 }
 
-                return base.FragmentNavigate(direction);
+                // Child entries exist in the tree if the entry has child entries and is expanded.
+                if (_owningPropertyDescriptorGridEntry is { ChildCount: > 0, Expanded: true })
+                {
+                    foreach (GridEntry childEntry in _owningPropertyDescriptorGridEntry.Children)
+                    {
+                        if (child == childEntry.AccessibilityObject)
+                        {
+                            return index;
+                        }
+
+                        index++;
+                    }
+                }
+
+                return -1;
+            }
+
+            internal AccessibleObject? GetNextChild(AccessibleObject child)
+            {
+                Debug.Assert(child is not null);
+
+                int index = GetChildIndex(child);
+                int lastChildIndex = GetChildCount() - 1;
+
+                Debug.Assert(index <= lastChildIndex);
+
+                // Ensure that it is a valid child and not the last child.
+                if (index == -1 || index == lastChildIndex)
+                {
+                    return null;
+                }
+
+                return GetChild(index + 1);
+            }
+
+            internal AccessibleObject? GetPreviousChild(AccessibleObject child)
+            {
+                Debug.Assert(child is not null);
+
+                int index = GetChildIndex(child);
+
+                // Ensure that it is a valid child and not the first child.
+                if (index == -1 || index == 0)
+                {
+                    return null;
+                }
+
+                return GetChild(index - 1);
             }
 
             internal override object? GetPropertyValue(UiaCore.UIA propertyID)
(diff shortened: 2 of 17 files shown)


dreddy-work:
@WPMGPRoSToTeMa , if this is draft, we should avoid "Waiting for review" label.

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/DropDownButton.DropDownButtonAccessibleObject.cs
```diff
@@ -49,18 +49,18 @@ public override void DoDefaultAction()
...
-                {
-                    return _owningPropertyGrid.EditAccessibleObject;
+                    return null;
                 }
 
-                return base.FragmentNavigate(direction);
+                return direction switch
+                {
+                    UiaCore.NavigateDirection.Parent => parent,
+                    UiaCore.NavigateDirection.NextSibling => parent.GetChildIndex(this) is int childId and >= 0 ? parent.GetChild(childId + 1) : null,
```
Tanya-Solyanik, line 60: The condition `ChildId >= 0` seems to be incorrect when going to the previous sibling. It should be comparing against the total count.

Consider renaming `ChildId` to `index`

WPMGPRoSToTeMa: `GetChild` returns `null` for out-of-range indexes, so this is actually correct though not obvious from the code. `childId >= 0` is more like a safety check against `-1`, it actually should be `childId != -1`. For the previous case we are also passing an out-of-range index `-1` in case `GetChildIndex` returns `0`. The current condition only prevents us from passing `-1 + 1` for the next sibling case.

I think we can make the code more explicit by throwing an exception (or at least use `Debug.Assert`) in case `GetChildIndex` returns `-1` (specifically on the call site) as this is a bug in the program's logic.

But I'm not sure if we should explicitly check if `+ 1` or `- 1` will give us an out-of-range index.

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyDescriptorGridEntry.PropertyDescriptorGridEntryAccessibleObject.cs
```diff
@@ -62,24 +163,103 @@ internal override void Expand()
...
-                    case UiaCore.NavigateDirection.FirstChild:
-                        return GetFirstChild();
-                    case UiaCore.NavigateDirection.LastChild:
-                        return GetLastChild();
-                }
-
-                return base.FragmentNavigate(direction);
+                if (_owningPropertyDescriptorGridEntry.ParentGridEntry is PropertyDescriptorGridEntry { AccessibilityObject: { } parent })
+                {
+                    return parent.GetChildIndex(this) is int childId and >= 0 ? parent.GetChild(childId - 1) : null;
```
Tanya-Solyanik, line 179: Is condition `ChildId >= 0` correct? It might give you `GetChild(-1)`, GetChild(-1) should return a `null1, but there is no reason to call it.

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/DropDownButton.DropDownButtonAccessibleObject.cs
Tanya-Solyanik: Thank you for the explanation.

>I think we can make the code more explicit by throwing an exception (or at least use Debug.Assert) in case GetChildIndex returns -1

This sounds reasonable, Debug.Assert gives the complete callstack, we don't have to put it on the call-site.

Some other PR (yet unmerged) introduces an `InvalidIndex` constant, please use it here.

>At least I can add a comment that it is expected that it can be out-of-range and it is handled by the GetChild method.

Sounds good!

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/TestPropertyGrid.cs
```diff
@@ -0,0 +1,84 @@
...
+internal class TestPropertyGrid : PropertyGrid
+{
+    private const User32.WM WM_DELAYEDEXECUTION = User32.WM.APP;
+
+    //public TestPropertyGrid(object selectedObject) : base()
+    //{
+    //    SelectedObject = selectedObject;
+    //}
+
+    public void PopupEditorAndClose(Action? callback = null)
```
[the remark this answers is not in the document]
RussKie: Few more ideas: `onClosingHandler`, `onClosingAction`.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/TestPropertyGrid.cs
```diff
@@ -0,0 +1,84 @@
...
+
+using System.Runtime.ExceptionServices;
+using System.Runtime.InteropServices;
+using System.Windows.Forms.PropertyGridInternal;
+using Xunit;
+using static Interop;
+
+namespace System.Windows.Forms.UITests;
+
+internal class TestPropertyGrid : PropertyGrid
```
RussKie, line 13: As per the established UnitTests conventions the subclassed types are
1. prefixed with "Sub", e.g., [`SubComboBox`](https://github.com/dotnet/winforms/blob/2e1174ee26dd6a9b1a6f0bf9b1e022b8635a559a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs#L2141), [`SubLabel`](https://github.com/dotnet/winforms/blob/2e1174ee26dd6a9b1a6f0bf9b1e022b8635a559a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/LabelTests.cs#L286), etc.
2. placed in the same file as the test - this ensures we don't introduce cross-test coupling to mocks.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/UserControlWithObjectCollectionEditor.cs
```diff
@@ -0,0 +1,37 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.ComponentModel;
+using System.Drawing.Design;
+using System.Windows.Forms.Design;
+
+namespace System.Windows.Forms.UITests;
+
+internal class UserControlWithObjectCollectionEditor : UserControl
```
RussKie, line 11: Any test specific mocks/stubs/etc. should reside in the same file as a test using those to avoid risks of introducing cross-test coupling.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyGridInternal/DropDownButton.DropDownButtonAccessibleObjectTests.cs
```diff
@@ -44,25 +44,29 @@ public void DropDownButtonAccessibleObject_Role_IsPushButton_ByDefault()
...
         }
 
-        [WinFormsFact]
-        public void DropDownButtonAccessibleObject_FragmentNavigate_SiblingsAreExpected()
-        {
-            using PropertyGrid control = new();
-            using Button button = new();
-            control.SelectedObject = button;
-            control.SelectedGridItem = control.GetCurrentEntries()[1].GridItems[5]; // FlatStyle property
+        //[WinFormsFact]
```
RussKie, line 47: Why is this commented out, and what's the plan for it?

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/DropDownButtonAccessibleObjectTests.cs
```diff
@@ -0,0 +1,72 @@
...
+
+using System.Windows.Forms.PropertyGridInternal;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace System.Windows.Forms.UITests;
+
+public class DropDownButtonAccessibleObjectTests : ControlTestBase
+{
+    public DropDownButtonAccessibleObjectTests(ITestOutputHelper testOutputHelper) : base(testOutputHelper)
```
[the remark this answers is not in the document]
RussKie: Yes, please see https://xunit.net/docs/capturing-output.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyGridInternal/PropertyDescriptorGridEntry.PropertyDescriptorGridEntryAccessibleObjectTests.cs
```diff
@@ -194,5 +498,61 @@ public DropDownEditablePropertyDescriptorGridEntry(PropertyGrid ownerGrid, GridE
 
             public override bool NeedsDropDownButton => true;
         }
+
+        private class SubPropertyGrid : PropertyGrid
```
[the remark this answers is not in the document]
RussKie: This is the right way of doing this.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/AccessibleObjectHelpers.cs
```diff
@@ -0,0 +1,56 @@
...
+using Xunit;
+using static Interop;
+
+namespace System.Windows.Forms.UITests;
+
+internal static class AccessibleObjectHelpers
+{
+    public static void VerifyHierarchy(this AccessibleObject accessibilityObject, AccessibleObject? expectedParent,
+                                 AccessibleObject? expectedPreviousSibling, AccessibleObject? expectedNextSibling,
+                                 IEnumerable<AccessibleObject> expectedChildren)
```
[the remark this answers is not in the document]
RussKie: I think "expected" prefix is quite fitting in this context.

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyDescriptorGridEntry.PropertyDescriptorGridEntryAccessibleObject.cs
```diff
@@ -10,7 +10,7 @@ namespace System.Windows.Forms.PropertyGridInternal
 {
```
RussKie, line 10: I'd like to see this class NRT annotated _before_ any more code is added.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/DropDownButtonAccessibleObjectTests.cs
```diff
@@ -0,0 +1,72 @@
...
+    [WinFormsFact]
+    public async Task DropDownAccessibleObject_Hierarchy_ParentWithoutChildEntries_IsCorrectAsync()
+    {
+        using Button button = new();
+        await RunSingleControlTestAsync<PropertyGrid>((form, propertyGrid) =>
+        {
+            propertyGrid.SelectedObject = button;
+            PropertyGridView propertyGridView = propertyGrid.TestAccessor().Dynamic._gridView;
+
+            propertyGridView.Focus();
```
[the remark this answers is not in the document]
WPMGPRoSToTeMa: Sometimes PG doesn't give the focus to the PGV (more precisely it's not always the active control of the PG).

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyDescriptorGridEntry.PropertyDescriptorGridEntryAccessibleObject.cs
WPMGPRoSToTeMa: Should this be done in a separate PR?

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/DropDownHolderAccessibleObjectTests.cs
```diff
@@ -0,0 +1,74 @@
...
+    public DropDownHolderAccessibleObjectTests(ITestOutputHelper testOutputHelper) : base(testOutputHelper)
+    {
+    }
+
+    [WinFormsFact]
+    public async Task DropDownHolderAccessibleObject_Hierarchy_WithListBox_IsCorrectAsync()
+    {
+        await RunSingleControlTestAsync<TestPropertyGrid>((form, propertyGrid) =>
+        {
+            propertyGrid.SelectedObject = new Button();
```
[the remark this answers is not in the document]
WPMGPRoSToTeMa: `SelectedObject` is automatically disposed in `TestPropertyGrid` class.

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyDescriptorGridEntry.PropertyDescriptorGridEntryAccessibleObject.cs
RussKie: Either way works. Maybe a separate PR would be better, as we probably review and merge it quicker.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/TestPropertyGrid.cs
```diff
@@ -0,0 +1,84 @@
...
+    {
+        object selectedObject = SelectedObject;
+
+        base.Dispose(disposing);
+
+        if (disposing && selectedObject is IDisposable disposable)
+        {
+            disposable.Dispose();
+        }
+    }
```
RussKie, line 72: Whilst this is not wrong, it adds an element of surprise - i.e., it's an unexpected behaviour, and callers wouldn't know about it. It also transfers the ownership over the life-time of a control to the property grid, which is not necessarily a good thing.
I suggest removing this, and updating the code as per https://github.com/dotnet/winforms/pull/6341/files#r795269508.

Tanya-Solyanik:
@WPMGPRoSToTeMa  - please copy over cleanup changes from netfx 4.8.1 into this PR

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyDescriptorGridEntry.PropertyDescriptorGridEntryAccessibleObject.cs
```diff
@@ -62,24 +163,103 @@ internal override void Expand()
...
+                }
+
+                return GetChild(index + 1);
+            }
+
+            internal AccessibleObject GetPreviousChild(AccessibleObject child)
+            {
+                int index = GetChildIndex(child);
+
+                if (index == -1 || index == 0)
```
SergeySmirnov-Akvelon, line 257:
```suggestion
                if (index <= 0)
```
Perhaps this option is easier?

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyGridView.DropDownHolder.DropDownHolderAccessibleObject.cs
```diff
@@ -20,17 +21,30 @@ public DropDownHolderAccessibleObject(DropDownHolder dropDownHolder) : base(drop
...
-                    => direction switch
+                {
+                    if (!ExistsInAccessibleTree)
+                    {
+                        return null;
+                    }
+
+                    PropertyGridView? gridView = _owningDropDownHolder._gridView;
+                    GridEntry? selectedEntry = gridView?.SelectedGridEntry;
+                    if (selectedEntry?.AccessibilityObject is not PropertyDescriptorGridEntryAccessibleObject parent)
```
SergeySmirnov-Akvelon, line 32: In the old implementation, if `_owningDropDownHolder._gridView.SelectedGridEntry is null`, then this did not affect the code for getting `NextSibling`, but in the new implementation we will immediately return `null`. Is this behavior expected?

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyGridView.GridViewListBoxAccessibleObject.cs
```diff
@@ -36,17 +36,16 @@ public GridViewListBoxAccessibleObject(GridViewListBox owningGridViewListBox) :
...
             internal override UiaCore.IRawElementProviderFragment? FragmentNavigate(UiaCore.NavigateDirection direction)
             {
-                if (direction == UiaCore.NavigateDirection.Parent && _owningPropertyGridView.SelectedGridEntry is not null)
+                if (_owningPropertyGridView is not { DropDownVisible: true, SelectedGridEntry: not null })
                 {
-                    return _owningPropertyGridView.SelectedGridEntry.AccessibilityObject;
+                    return null;
                 }
 
-                if (direction == UiaCore.NavigateDirection.NextSibling)
```
SergeySmirnov-Akvelon, line 44: Is it expected that we removed the implementation for `NextSibling`?

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/DropDownButtonAccessibleObjectTests.cs
```diff
@@ -0,0 +1,72 @@
...
+            Assert.True(propertyGridView.Focused);
```
SergeySmirnov-Akvelon, line 27: Please add a blank line before the `Assert` here and below.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyGridInternal/PropertyGridView.DropDownHolder.DropDownHolderAccessibleObjectTests.cs
```diff
@@ -36,40 +36,45 @@ public void DropDownHolder_AccessibilityObject_Constructor_throws_error_if_passe
             });
         }
 
-        [WinFormsFact]
-        public void DropDownHolder_AccessibilityObject_ReturnsExpected()
-        {
-            using PropertyGrid propertyGrid = new PropertyGrid();
-            PropertyGridView propertyGridView = propertyGrid.TestAccessor().GridView;
+        //[WinFormsFact]
```
SergeySmirnov-Akvelon, line 39: Why are these tests ignored?

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyDescriptorGridEntry.PropertyDescriptorGridEntryAccessibleObject.cs
Tanya-Solyanik: Alternatively - add a Debug.Assert

Tanya-Solyanik (approved):
Looks good, but please reply to @SergeySmirnov-Akvelon comments.
@vladimir-krestov - Could you please review this change?

I'm approving pending @SergeySmirnov-Akvelon  and @vladimir-krestov review.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/PropertyGridInternal/PropertyDescriptorGridEntry.PropertyDescriptorGridEntryAccessibleObjectTests.cs
```diff
@@ -0,0 +1,277 @@
...
+            return Task.CompletedTask;
+        });
+    }
+
+    [WinFormsFact]
+    public async Task PropertyDescriptorGridEntryAccessibleObject_FragmentNavigate_LastChild_IsNullAsync()
+    {
+        await RunSingleControlTestAsync<SubPropertyGrid<Button>>((form, grid) =>
+        {
+            Assert.Null(grid[nameof(Button.Font)].AccessibilityObject.FragmentNavigate(NavigateDirection.LastChild));
```
Tanya-Solyanik, line 199: nit: `nameof(Button.Font)` -> `nameof(Control.Font)`  becasue this is not a button...  resulting string is the same but `Button` looks confusing.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/AccessibleObjectHelpers.cs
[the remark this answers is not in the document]
Tanya-Solyanik: It is implied and its presence makes code less readable.

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyGridView.DropDownHolder.DropDownHolderAccessibleObject.cs
Tanya-Solyanik: Yes, this is an error condition

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyGridView.GridViewListBoxAccessibleObject.cs
Tanya-Solyanik: Listbox is the only child

Tanya-Solyanik (approved):

Tanya-Solyanik:
@Olina-Zhang  - could you please test this change?

Amy-Li03:
Completed testing with no new issue found.

vladimir-krestov:
Is reviewing it...

vladimir-krestov (approved):
Looks good.
Great work!

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/DropDownButton.DropDownButtonAccessibleObject.cs
WPMGPRoSToTeMa: It was converted to the helper methods inside the property descriptor entry accessible object class.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/AccessibleObjectHelpers.cs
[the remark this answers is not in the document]
WPMGPRoSToTeMa: This helper method was removed. Instead I've made more fine-grained tests for the specific navigate constants.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/DropDownButtonAccessibleObjectTests.cs
```diff
@@ -0,0 +1,72 @@
...
+
+            GridEntry selectedEntry = propertyGrid.GetCurrentEntries().Single(e => e.PropertyName == "Appearance")
+                .GridItems.Cast<GridEntry>().Single(e => e.PropertyName == "FlatStyle");
```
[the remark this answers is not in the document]
WPMGPRoSToTeMa: Added an indexer to `SubPropertyGrid` to greatly simplify this.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/TestPropertyGrid.cs
WPMGPRoSToTeMa: This will add boilerplate code to each test method. `SelectedObject` is currently created and disposed inside the `SubPropertyGrid` class (and it's generalized on the type of `SelectedObject`), so I think it will be less surprising than before.

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyGridView.GridViewListBoxAccessibleObject.cs
WPMGPRoSToTeMa: That's because we're also fixing #4876 in this PR. Previously we had:
`Entry > ListBox`, so there were some siblings.
Now we have:
`Entry > DropDownHolder > ListBox`, so `ListBox` is the only child of `DropDownHolder`.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/PropertyGridInternal/PropertyDescriptorGridEntry.PropertyDescriptorGridEntryAccessibleObjectTests.cs
WPMGPRoSToTeMa: I use `Button` because the type of `SelectedObject` is `Button`.

WPMGPRoSToTeMa:
It would be great if someone will add before/after screenshots for the both bugs. I'm currently not able to do that, sorry that haven't done this before.

PR #4910, opened 2021-05-14
Separating ListViewItemAccessibleObject

SergeySmirnov-Akvelon:
Fixes #4748 

## Proposed changes
During the separating of `ListViewAccessibilityObject`, two ways of separating were considered: `View` and `ListViewGroup` presence. 
Unfortunately, it makes no sense to separate `ListViewAccessibilityObject` by `View`, since the class contains almost no code specific to any `View`.
A `ListViewAccessibilityObject`  has code specific to the presence or absence of `ListViewGroups`, but the display of a `ListViewGroup` depends on many factors:
`ShowGroups` property;
`VirtualMode` property;
`View` property;
`Groups` property;
`Visual styles`.
Keeping track of all of these factors will lead to the need to add changes that will be much more difficult to support than the current code. Therefore, it was decided to leave the current code for the `ListViewAccessibilityObject`.

When separating the `ListViewItemAccessibleObject`, we took into account how the `ListViewItemAccessibleObject` works with `ListViewGroups` and `ListViewItems` depending on the view.

**ListViewItemBaseAccessibleObject**
The `ListViewItemBaseAccessibleObject` class has been added. It contains the implementation for most of the properties and methods of the `ListViewItemAccessibleObject` class. In addition, its implementation fully corresponds to the behavior of elements when the tree is in `LargeIcon` or `SmallIcon` view.

**ListViewItemListAccessibleObject**
For the `List` view, an `ListViewItemListAccessibleObject` class has been added. It contains an overridden implementation for the `Bounds` and `RuntimeId` properties and the `FragmentNavigate` method. Since this view does not support `ListViewGroup`, it allows us to remove some unnecessary checks.

**ListViewItemTileAccessibleObject**
For the `Tile` mode, an `ListViewItemTileAccessibleObject` class has been added. It contains implementations for methods such as `FragmentNavigate`, `GetChild`, `GetChildInternal`, `GetChildCount`, `GetChildIndex` and `GetSubItemBounds` methods. This is necessary because this mode supports `ListViewSubItems` and has complex logic to work with them. In addition, we have a special method `GetChildInternal`, which is necessary because inside the classes we operate with real indexes of `ListViewSubItems` , and not visible to the user.

**ListViewItemDetailsAccessibleObject**
[...]

PR #6249, opened 2021-11-30
Fixing NullReferenceException when user removes ListViewItem

SergeySmirnov-Akvelon:
Fixes #6230 

## Proposed changes
- The issue is reproduced because we believed that if an `ListViewItem` does not have a `ListView`, then it is inaccessible to the user and has no interaction with the Accessibility tool (Inspect, Narrator).
- Unfortunately, after deleting an `ListViewItem`, the Accessibility tool still tries to get information about it, although the `ListViewItem` no longer has a `ListView`. Therefore, the logic has been updated to work with an `ListViewItem` that does not have a `ListView`.
- Also, after the focused `ListViewItem` is removed, it remains cached in the `FocusedItem` property. It also forces the Accessibility tool to try to retrieve the data of the `ListViewItem` that is not displayed. As a fix, a logic was added to reset the `Focused` flag from the `ListViewItem`.

## Customer Impact
**Before fix:**
Issue-6230-before

**After fix:**
Issue-6230

## Regression? 
- Yes.

## Risk
- Minimal.

## Test methodology 
-  CTI team;
- Unit tests.

## Accessibility testing  
- Accessibility Insights;
- Inspect;
- Narrator.

## Test environment(s) 
- Microsoft Windows [Version 10.0.19041.1348]
- .NET Core SDK: 7.0.0-alpha.1.21576.4

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6249)

PR #6352, opened 2021-12-16
Fixing issue with getting ListViewItem.AccessibilityObject when ListView does not exist

### LINKED ISSUE #6230: ListView.Items.RemoveAt causes NullReferenceException
* .NET Core Version: 
6.0.0

Classic .NET - No
.NET 5 - No
.NET 6 - Yes

**Problem description:**
Deleting a selected item from ListView causes NullReferenceException on calling ListView.Items.RemoveAt(index).
The bug reproduces stable in my app in the following conditions:
1. The item being deleted must be selected. If it's not selected, no error.
2. To workaround the error, we can call ListView.Focus() before delete.
3. ListView is a Details view.

The code that fails:
`

    // listView1.Focus(); // helps to avoid error

    listView1.Items.RemoveAt(0); // error is here if item 0 is selected

    MessageBox.Show("a"); // shows after error
`

Despite the error, the item is deleted.

Error stack:
   at System.Windows.Forms.ListViewItem.get_AccessibilityObject()
   at System.Windows.Forms.ListView.ListViewAccessibleObject.GetFocus()
   at System.Windows.Forms.Control.WmSetFocus(Message& m)
   at System.Windows.Forms.Control.WndProc(Message& m)
   at System.Windows.Forms.ListView.WndProc(Message& m)
   at System.Windows.Forms.Control.ControlNativeWindow.WndProc(Message& m)

**Expected behavior:**
No error in the above case. Because it worked well before .NET 6.

**Minimal repro:**
I couldn't create a small app to reproduce the error, because there is no error in a simple app. However, I hope that a developer who works daily on WinForms, can find the possible mistake in ListViewItem.get_AccessibilityObject()
Edit: the test app is attached below.

SergeySmirnov-Akvelon:
Fixes #6230 

## Proposed changes
- The issue is reproduced because we have cached the `ListViewItem.AccessibilityObject`. After the changes in #4910, we added logic that, with each call to the `AccessibilityObject` property, checked that the `ListView` exists and, if absent, threw an exception. Now we return null as a `ListViewItem.AccessibilityObject` instead of returning an exception.

- Also, after the focused `ListViewItem` is removed, it remains cached in the `FocusedItem` property. It also forces the Accessibility tool to try to retrieve the data of the `ListViewItem` that is not displayed. As a fix, a logic was added to reset the Focused flag from the `ListViewItem`.

## Customer Impact
**Before fix:**
Issue-6230

**After fix:**
Issue-6230-SecondFix

## Regression? 
- Yes (from #4910)

## Risk
- Minimal

## Test methodology
- Manual testing
- CTI team
- Unit tests

## Accessibility testing  
- Inspect 
- Narrator
- Accessibility Insights

## Test environment(s)
- .NET Core SDK: 7.0.0-alpha.1.21576.4

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6352)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListView.ListViewNativeItemCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListView.ListViewNativeItemCollection.cs
@@ -427,6 +427,8 @@ namespace System.Windows.Forms
 
                 owner.ApplyUpdateCachedItems();
                 int itemID = DisplayIndexToID(index);
+
+                this[index].Focused = false;
                 this[index].UnHost(true);
 
                 if (owner.IsHandleCreated)
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.cs
@@ -227,7 +227,13 @@ namespace System.Windows.Forms
         {
             get
             {
-                ListView owningListView = listView ?? Group.ListView ?? throw new InvalidOperationException(nameof(listView));
+                ListView owningListView = listView ?? Group?.ListView;
+                if (owningListView is null)
+                {
+                    _accessibilityObject = null;
+                    return _accessibilityObject;
+                }
+
                 if (_accessibilityObject is null || owningListView.View != _accessibilityObjectView)
                 {
                     _accessibilityObjectView = owningListView.View;
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItemTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItemTests.cs
@@ -852,15 +852,16 @@ namespace System.Windows.Forms.Tests
             yield return new object[] { Color.Red, Color.Red };
         }
 
-        [Theory]
+        [WinFormsTheory]
         [MemberData(nameof(BackColor_Set_TestData))]
         public void ListViewItem_BackColor_GetWithOwner_ReturnsExpected(Color value, Color expected)
         {
-            var listView = new ListView
+            using ListView listView = new()
             {
                 BackColor = value
             };
-            var item = new ListViewItem();
+
+            ListViewItem item = new();
             listView.Items.Add(item);
             Assert.Equal(expected, item.BackColor);
 
@@ -884,12 +885,12 @@ namespace System.Windows.Forms.Tests
             Assert.Equal(expected, item.BackColor);
         }
 
-        [Theory]
+        [WinFormsTheory]
         [MemberData(nameof(BackColor_Set_TestData))]
         public void ListViewItem_BackColor_SetWithOwner_GetReturnsExpected(Color value, Color expected)
         {
-            var listView = new ListView();
-            var item = new ListViewItem();
+            using ListView listView = new();
+            ListViewItem item = new();
             listView.Items.Add(item);
 
             item.BackColor = value;
@@ -908,15 +909,16 @@ namespace System.Windows.Forms.Tests
             yield return new object[] { Color.Red, Color.Red };
         }
 
-        [Theory]
+        [WinFormsTheory]
         [MemberData(nameof(ForeColor_Set_TestData))]
         public void ListViewItem_ForeColor_GetWithOwner_ReturnsExpected(Color value, Color expected)
         {
-            var listView = new ListView
+            using ListView listView = new()
             {
                 ForeColor = value
             };
-            var item = new ListViewItem();
+
+            ListViewItem item = new();
             listView.Items.Add(item);
             Assert.Equal(expected, item.ForeColor);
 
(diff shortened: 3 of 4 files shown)


SergeySmirnov-Akvelon:
Hi @Tanya-Solyanik. Could you take a look at this fix? Here we have fixed the error with getting the `ListViewItem.AccessibilityObject`.

dreddy-work (dismissed):
LGTM. waiting @Tanya-Solyanik's sign off.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItemTests.cs
```diff
@@ -1063,5 +1063,23 @@ private static void AssertEqualListViewSubItem(ListViewItem.ListViewSubItem[] ex
...
+
+        [Fact]
+        public void ListViewItem_AccessibilityObject_ReturnsNull_IfListViewNotExists()
+        {
+            ListViewItem item = new();
+
+            Assert.Null(item.AccessibilityObject);
+        }
+
+        [Fact]
```
vladimir-krestov, line 1075: @RussKie, why do we use [Fact] here for all tests?

src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.cs
```diff
@@ -229,7 +229,12 @@ internal virtual AccessibleObject AccessibilityObject
         {
             get
             {
-                ListView owningListView = listView ?? Group.ListView ?? throw new InvalidOperationException(nameof(listView));
+                ListView owningListView = listView ?? Group?.ListView;
+                if (owningListView is null)
+                {
+                    return _accessibilityObject = null;
```
vladimir-krestov, line 235: May we catch NRE when using an item's accessible object anywhere?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItemTests.cs
WPMGPRoSToTeMa: I think we don't need WinForms context because we don't create controls here.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.cs
Tanya-Solyanik: nit: i find this syntax hard to read: `return _accessibilityObject = null;`. Consider one statement on one line instead.

src/System.Windows.Forms/src/System/Windows/Forms/ListView.ListViewNativeItemCollection.cs
```diff
@@ -427,6 +427,8 @@ public void RemoveAt(int index)
 
                 owner.ApplyUpdateCachedItems();
                 int itemID = DisplayIndexToID(index);
+
+                this[index].Focused = false;
```
Tanya-Solyanik, line 431: I think I had asked it before, but to double-check - had you reviewed if we have the same issue in other container controls?

src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.cs
Tanya-Solyanik: >May we catch NRE when using an item's accessible object anywhere?

We can review all callers in ListView code and ensure that we handle `null` correctly.

SergeySmirnov-Akvelon: In case we want to be sure that we will not get NRE, I would choose a #6249. In this PR, we are fixing the customer issue, but errors will continue to be thrown for us and the accessibility tools.

Tanya-Solyanik: I would prefer this fix(6352) for servicing branches. And 6249 for 7.0 once we understand why narrator caches providers for removed items.

Tanya-Solyanik (approved):
@dreddy-work  - I don't mind fixing this particular regression in the 6.0 branches, but in 7.0, we should keep working on this issue, and understand why narrator calls into released providers, see PR #6352

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItemTests.cs
RussKie: Yes, we need to mark this as `WinFormsFact` because we create `ListView` control.
I had a cursory glance and it looks like we have multiple issues in this test file that need to be fixed:
1. Use `WinFormsFact` instead of `Fact`
2. Dispose (i.e. use the `using`) controls in tests

src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.cs
RussKie, line 235: ❗ Do no combine multiple operation like this. It severely inhibits the readability and comprehension.

dreddy-work:
If band aid fix releasing to servicing but not available in 7.0 previews, may still create some noise. I think we should take this fix into both servicing and main branch while still working on full fix in 7.0 with #6249.

src/System.Windows.Forms/src/System/Windows/Forms/ListView.ListViewNativeItemCollection.cs
SergeySmirnov-Akvelon: I found a similar bug today (#6612 ). Tomorrow I plan to run some more tests to make sure everything works fine.

dreddy-work:
We are taking this band aid into servicing 6.0.3 as well. can you create one for servicing?

Issue #5960, opened 2021-10-13
ListView not firing GroupCollapsedStateChanged when user expands/collapses a ListViewGroup using the keyboard


SergeySmirnov-Akvelon:
.NET Core Version: 
- .NET Core SDK: 7.0.0-alpha.1.21512.1

Have you experienced this same bug with .NET Framework?: 
- No. `ListViewGroup.CollapsedState` was implemented in .NET Core 5.0 (#3155)

**Repro steps:**
1. Create a `Form` with a `ListView`.
2. Add `ListViewGroups` to the `ListView` (`ListViewGroups` should have `CollapsedState = ListViewGroupCollapsedState.Expanded` or `CollapsedState = ListViewGroupCollapsedState.Collapsed`).
3. Add `EventHandler` for `GroupCollapsedStateChanged`.
4. Run the application.
5. Expand / collapse a `ListViewGroup` using the keyboard.

**Actual behavior:**
`ListView` not firing `GroupCollapsedStateChanged` when user expands/collapses a `ListViewGroup` using the keyboard
![5960](https://user-images.githubusercontent.com/23376742/137115469-013fdfbc-4036-4d40-8558-53b763edba67.gif)

**Expected behavior:**
`ListView` should firing `GroupCollapsedStateChanged` when user expands/collapses a `ListViewGroup` using the keyboard

Jenna-Zheng:
Verified this issue with .NET SDK 7.0.100-alpha.1.22060.5, it is fixed.
1. Collapsed and expanded ListViewGroup from the keyboard with writing to log at the event handler:
![ListView_Keyboard](https://user-images.githubusercontent.com/64949616/148894977-06ebbace-0001-4089-8e09-7a0a3680b174.gif)
2. Examined Narrator's text while collapsing and expanding ListViewGroup from the keyboard:
![ListView_Narrator](https://user-images.githubusercontent.com/64949616/148895045-eb3ce92f-47b5-4764-99f5-fdd35fe631f1.gif)
3. Watched the ExpandCollapse.ExpandCollapseState property at the Inspect tool after expanding/collapsing group from the keyboard:
![image](https://user-images.githubusercontent.com/64949616/148895072-71cf019a-b5de-4b96-aa96-89ecf5c2a8a1.png)
![image](https://user-images.githubusercontent.com/64949616/148895251-73016361-02ac-4256-aaf3-3947e8c3f88f.png)

Lydia-Shi:
[...]

PR #6354, opened 2021-12-16
Fixing NRE when the user press the left / right arrow in ListView

### LINKED ISSUE #6353: An exception appears when the user press the left / right arrow in the ListView without ListViewGroup
.NET Core Version: 
 - NET Core SDK: 7.0.0-alpha.1.21606.7

- Yes (from #5960)

**Problem description:**

1. Create a `Form` with a `ListView`.
2. Add an `ListViewItem` to the `ListView`.
3. Run the app.
4. Press the left / right arrow.

**Actual behavior:**
User sees an exception.
Issue-6353

**Expected behavior:**
User should not see exception.

SergeySmirnov-Akvelon:
Fixes #6353

## Proposed changes
- The issue is reproduced because we are trying to access a `ListViewGroup` that the `ListViewItem` may not contain it.
- Added a condition so that the code for working with `ListViewGroups` is called only when the `ListViewGroups` are enabled (`GroupsEnabled` property).
- Fixed condition for cases when the `Group` property has `null` value. 

## Customer Impact
**Before fix:**
Issue-6353

**After fix:**
Issue-6353_AfterFix

## Regression? 
- Yes (from #5960)

## Risk
- Minimal

## Test methodology 
- Unit tests
- CTI team

## Test environment(s) 
- .NET Core SDK: 7.0.0-alpha.1.21576.4

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6354)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
@@ -6931,14 +6931,13 @@ namespace System.Windows.Forms
                 case User32.WM.KEYUP:
                     int key = (int)m.WParamInternal;
 
-                    if ((key == User32.VK.LEFT || key == User32.VK.RIGHT) && SelectedItems.Count > 0)
+                    if (GroupsDisplayed && (key is User32.VK.LEFT or User32.VK.RIGHT) && SelectedItems.Count > 0)
                     {
                         ListViewGroup group = SelectedItems[0].Group;
-                        ListViewGroupCollapsedState groupCollapsedState = group.CollapsedState;
 
-                        if (group is null || groupCollapsedState == ListViewGroupCollapsedState.Default
-                            || (key == User32.VK.LEFT && groupCollapsedState == ListViewGroupCollapsedState.Collapsed)
-                            || (key == User32.VK.RIGHT && groupCollapsedState == ListViewGroupCollapsedState.Expanded))
+                        if (group is null || group.CollapsedState is ListViewGroupCollapsedState.Default
+                            || (key == User32.VK.LEFT && group.CollapsedState is ListViewGroupCollapsedState.Collapsed)
+                            || (key == User32.VK.RIGHT && group.CollapsedState is ListViewGroupCollapsedState.Expanded))
                         {
                             break;
                         }
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewTests.cs
@@ -5369,6 +5369,28 @@ namespace System.Windows.Forms.Tests
             }
         }
 
+        [WinFormsTheory]
+        [InlineData(Keys.Right)]
+        [InlineData(Keys.Left)]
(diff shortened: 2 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
```diff
@@ -6931,14 +6931,13 @@ protected override void WndProc(ref Message m)
                 case User32.WM.KEYUP:
                     int key = (int)m.WParamInternal;
 
-                    if ((key == User32.VK.LEFT || key == User32.VK.RIGHT) && SelectedItems.Count > 0)
+                    if (GroupsDisplayed && (key == User32.VK.LEFT || key == User32.VK.RIGHT) && SelectedItems.Count > 0)
```
WPMGPRoSToTeMa, line 6934:
```suggestion
                    if (GroupsDisplayed && (key is User32.VK.LEFT or User32.VK.RIGHT) && SelectedItems.Count > 0)
```

WPMGPRoSToTeMa:
```suggestion
                    if (GroupsDisplayed && (key == User32.VK.LEFT || key == User32.VK.RIGHT) && SelectedItems.Any())
```
Though I'm not sure about performance, https://stackoverflow.com/a/63569609 (see .NET Core).

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewTests.cs
```diff
@@ -5369,6 +5369,47 @@ public void ListView_OnMouseClick_EditLabel_AsExpected(int subItemIndex, bool la
...
+            // The MSDN page tells us what bits of lParam to use for each of the parameters.
+            // All we need to do is some bit shifting to assemble lParam
+            // lParam = repeatCount | (scanCode << 16)
+            nint keyCode = (nint)Keys.Right;
+            nint lParam = 0x00000001 | keyCode << 16;
+            User32.SendMessageW(listView, User32.WM.KEYUP, keyCode, lParam);
+
+            Assert.True(listView.IsHandleCreated);
+        }
+
```
WPMGPRoSToTeMa, line 5412: I think these tests could be merged into a single theory based on a `keyCode` value.

WPMGPRoSToTeMa (approved):
LGTM

ArtemTatarinov:
LGTM

ArtemTatarinov (dismissed):
LGTM

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
SergeySmirnov-Akvelon: Hi @Tanya-Solyanik, @dreddy-work, @RussKie. Could you please share your opinion on which is better to use `SelectedItems.Count > 0` or `SelectedItems.Any()`.

dreddy-work: Not so sure but Count would seem fine here.

dreddy-work (approved):

PR #6356, opened 2021-12-16
Do not attempt to cache CheckedItems if in virtual mode

### LINKED ISSUE #4042: ListView is in virtual mode throws IOE when user disables checkboxes
- .NET Core Version:
.NET SDK 5.0.100-rc.1.20420.14

- Have you experienced this same bug with .NET Framework?:
Yes

**Minimal repro:**

1. Create application with ListView in virtual mode
2. Enable checkboxes
3. Disable checkboxes
4. Or use attached [WindowsFormsApp3.zip](https://github.com/dotnet/winforms/files/5310459/WindowsFormsApp3.zip)

ListView - checkboxes
**StackTrace:**
```
   at System.Windows.Forms.ListView.CheckedListViewItemCollection.get_Count()
   at System.Windows.Forms.ListView.set_CheckBoxes(Boolean value)
   at WindowsFormsApp3.Form1.Button1_Click(Object sender, EventArgs e)
   at System.Windows.Forms.Control.OnClick(EventArgs e)
   at System.Windows.Forms.Button.OnClick(EventArgs e)
```

**Expected behavior:**
Checkboxes should be disabled without exceptions

**Actual behavior:**
'Cannot access the checked items collection when the ListView is in virtual mode.' exception is displayed

jbhensley:
Fixes #4042 

## Proposed changes

`ListView` attempts to cache `CheckedItems` when setting `CheckBoxes` to `false`. This results in an `InvalidOperationException` in `CheckedItems.Count` when the `ListView` is in virtual mode. Since `CheckedItems` cannot be used in virtual mode anyway, the fix is to skip this caching logic when `VirtualMode == true`.

Note: there was a unit test that ensured this behavior. It's not clear why since it seems undesirable. The error is trivial to repro. All you have to do is drop a new `ListView` into the forms designer, set `VirtualMode` to `true`,  and then try toggling `CheckBoxes` to `true` and then back to `false`, all within the properties tool window. The existing unit test was modified to ensure the error does not happen.

## Customer Impact

Customer should be able to toggle `CheckBoxes` from `true` to `false` when in virtual mode.

## Regression? 

- No

## Risk

Can't identify any

## Test methodology 

- Unit test ensures functionality 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6356)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
@@ -468,7 +468,7 @@ namespace System.Windows.Forms
                             throw new NotSupportedException(SR.ListViewCheckBoxesNotSupportedInTileView);
                         }
 
-                        if (CheckBoxes)
+                        if (CheckBoxes && !VirtualMode)
                         {
                             // Save away the checked items just in case we re-activate checkboxes
                             //
@@ -515,7 +515,7 @@ namespace System.Windows.Forms
                             throw new NotSupportedException(SR.ListViewCheckBoxesNotSupportedInTileView);
                         }
 
-                        if (CheckBoxes)
+                        if (CheckBoxes && !VirtualMode)
                         {
                             // Save away the checked items just in case we re-activate checkboxes
                             //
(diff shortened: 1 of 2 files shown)


dreddy-work:
@SergeySmirnov-Akvelon, this is waiting on your review.

ArtemTatarinov:
Just in case I've examined it too: looks good to me.

vladimir-krestov (approved):
Looks good to me ✔️
@dreddy-work, I would ask CTI to check this fix. I will send a request.

Olina-Zhang:
Tested the PR with no new issue found.

dreddy-work (approved):

PR #6358, opened 2021-12-16
Enable nullability in ToolStripItemAccessibleObject

gpetrou:
## Proposed changes

- Enable nullability in ToolStripItemAccessibleObject.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6358)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.ToolStripItemAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.ToolStripItemAccessibleObject.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.Diagnostics;
 using System.Drawing;
 using System.Globalization;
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.ToolStripItemAccessibleObject.cs
```diff
@@ -132,7 +130,7 @@ internal override object GetPropertyValue(UiaCore.UIA propertyID)
                 return base.GetPropertyValue(propertyID);
             }
 
-            public override string Name
+            public override string? Name
```
dreddy-work, line 133: I don't see this can be null. It can be `string.empty` but not null.

dreddy-work (changes requested):

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.ToolStripItemAccessibleObject.cs
gpetrou: I other places such as ControlAccessibleObject this is nullable, I guess because TextWithoutMnemonics can return null.

dreddy-work (approved):

PR #6368, opened 2021-12-20
Fixing multiple adding of `ControlBox` buttons to MDI child form

### LINKED ISSUE #6339: Setting `MainMenuStrip` at the MDI child form several times leads to multiplying `ControlBox` buttons
* .NET Core Version: 
7.0

Yes, at 4.8

**Problem description:**
If the `MainMenuStrip` property of the child MDI `Form` class is set to some `MenuStrip` instance, then set to null, and then is set again to the `MenuStrip` instance, then the `ControlBox` buttons (`System`, `Minimize`, `Restore`, `Close`) are cloned:

mdiMultipleButtons

The handler for this checkbox does the following:

image

**Expected behavior:**
The set of the `ControlBox` buttons should always be single per window.

**Minimal repro:**

1. Run the internal `WinformsControlsTest` test app
2. Click the `MDI Parent` button
3. Check and uncheck the `Set parent menustrip` checkbox several times.

ArtemTatarinov:
Fixes #6339

## Proposed changes

- The fix is based on PR #4465. Looks like something was changed in WinAPI: now not the latest build of 5.0 nor this specifically checked out commit doesn't fix the issue.
- Now we need to recreate the dummy menu every time we're setting the `MainMenuStrip` property of the `Form`.

## Customer Impact
**Before the fix** 
cloneBeforeTheFix

**After the fix** 
cloneAfterTheFix

## Regression? 
- Yes

## Risk
- Minimal

## Test methodology 
- Manual testing (see the screenshots above)
- Unit tests
- CTI team

## Test environment(s) 
Windows 10 10.0.19043.1415
.NET SDK 7.0.0-alpha.1.21606.7

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6368)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
@@ -1354,7 +1354,7 @@ namespace System.Windows.Forms
                 Properties.SetObject(PropMainMenuStrip, value);
                 if (IsHandleCreated)
                 {
-                    UpdateMenuHandles();
+                    UpdateMenuHandles(recreateMenu: true);
                 }
             }
         }
@@ -5540,7 +5540,7 @@ namespace System.Windows.Forms
             }
         }
 
-        private void UpdateMenuHandles()
+        private void UpdateMenuHandles(bool recreateMenu = false)
         {
             if (!IsHandleCreated)
             {
@@ -5567,7 +5567,7 @@ namespace System.Windows.Forms
                     // an ole menu is being removed.
                     // Make MDI forget the mdi item position.
                     IntPtr? dummyMenu = Properties.GetObject(PropDummyMdiMenu) as IntPtr?;
-                    if (!dummyMenu.HasValue || dummyMenu.Value == IntPtr.Zero)
+                    if (!dummyMenu.HasValue || dummyMenu.Value == IntPtr.Zero || recreateMenu)
                     {
                         dummyMenu = User32.CreateMenu();
                         Properties.SetObject(PropDummyMdiMenu, dummyMenu);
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
```diff
@@ -5540,7 +5540,7 @@ private void UpdateLayered()
             }
         }
 
-        private void UpdateMenuHandles()
+        private void UpdateMenuHandles(bool forceDummyMenuRecreating = false)
```
WPMGPRoSToTeMa, line 5543: Consider:
```suggestion
        private void UpdateMenuHandles(bool forceDummyMenuRecreation = false)
```

WPMGPRoSToTeMa (dismissed):
LGTM

src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
```diff
@@ -1354,7 +1354,7 @@ public MenuStrip MainMenuStrip
                 Properties.SetObject(PropMainMenuStrip, value);
                 if (IsHandleCreated)
                 {
-                    UpdateMenuHandles();
+                    UpdateMenuHandles(forceDummyMenuRecreating: true);
```
RussKie, line 1357:
```suggestion
                    UpdateMenuHandles(recreateMenu: true);
```

RussKie (approved):
👍

dreddy-work (approved):

Olina-Zhang:
Tested this PR with no new issue found, linked issue is fixed.

vladimir-krestov (approved):
Approved ✔️

dreddy-work:
Given that this has been regressed by windows but not ours and its been happening in .NET framework 4.8, lets wait on servicing and take it only to 7.0 for now.

PR #6376, opened 2021-12-21
Enable nullability in DataObject

gpetrou:
## Proposed changes

- Enable nullability in DataObject.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6376)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataObject.DataStore.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataObject.DataStore.cs
@@ -2,7 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System.Collections;
 using System.Collections.Specialized;
 using System.ComponentModel;
 using System.Diagnostics;
@@ -18,17 +17,17 @@ namespace System.Windows.Forms
         {
             private class DataStoreEntry
             {
-                public object Data { get; }
+                public object? Data { get; }
                 public bool AutoConvert { get; }
 
-                public DataStoreEntry(object data, bool autoConvert)
+                public DataStoreEntry(object? data, bool autoConvert)
                 {
                     Data = data;
                     AutoConvert = autoConvert;
                 }
             }
 
-            private readonly Hashtable _data = new Hashtable(BackCompatibleStringComparer.Default);
+            private readonly Dictionary<string, DataStoreEntry> _data = new Dictionary<string, DataStoreEntry>(BackCompatibleStringComparer.Default);
 
             public DataStore()
             {
@@ -43,9 +42,8 @@ namespace System.Windows.Forms
                     return null;
                 }
 
-                DataStoreEntry? dse = (DataStoreEntry?)_data[format];
                 object? baseVar = null;
-                if (dse is not null)
+                if (_data.TryGetValue(format, out DataStoreEntry? dse))
                 {
                     baseVar = dse.Data;
                 }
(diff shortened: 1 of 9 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/DataObject.DataStore.cs
```diff
@@ -20,55 +18,55 @@ private class DataStore : IDataObject
         {
             private class DataStoreEntry
             {
-                public object data;
-                public bool autoConvert;
+                public object Data { get; }
```
Tanya-Solyanik, line 21: Doesn't this class participate in binary serialization? If this is the case, we can't change field names, or change fields to properties. It will break round trips between different framerworks

src/System.Windows.Forms/src/System/Windows/Forms/DataObject.FormatEnumerator.cs
```diff
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
```
RussKie, line 6: This file hasn't been annotated correctly. E.g., line:35 and line:41 contradict each other.

src/System.Windows.Forms/src/System/Windows/Forms/DataObject.DataStore.cs
RussKie: It's not obvious. How do we verify?

JeremyKuhne: If this class was `[Serializable]` then it would be a concern. You can't serialize with `BinaryFormatter` without that attribute across the object graph. If `[Serializable]` types implement `ISerializable` then they custom serialize and the names may not matter either- you'd have to look at the implementation of `ISerializable` on said class to see if they're doing any sort of type name reflection.

src/System.Windows.Forms/src/System/Windows/Forms/DataObject.DataStore.cs
```diff
@@ -20,55 +18,55 @@ private class DataStore : IDataObject
...
                 {
-                    this.data = data;
-                    this.autoConvert = autoConvert;
+                    Data = data;
+                    AutoConvert = autoConvert;
                 }
             }
 
-            private readonly Hashtable data = new Hashtable(BackCompatibleStringComparer.Default);
+            private readonly Hashtable _data = new Hashtable(BackCompatibleStringComparer.Default);
```
JeremyKuhne, line 31: If we're going to annotate this class we should try to change this to a typed Dictionary so we can accurately track the types involved.

JeremyKuhne: That will avoid a lot of the `!` usage.

dreddy-work:
@Tanya-Solyanik / @JeremyKuhne / @RussKie , can you take a final look here?

Tanya-Solyanik (approved):
Thank you, looks good!

src/System.Windows.Forms/src/System/Windows/Forms/DataObject.DataStore.cs
```diff
@@ -20,58 +17,56 @@ private class DataStore : IDataObject
...
                 {
-                    this.data = data;
-                    this.autoConvert = autoConvert;
+                    Data = data;
+                    AutoConvert = autoConvert;
                 }
             }
 
-            private readonly Hashtable data = new Hashtable(BackCompatibleStringComparer.Default);
+            private readonly Dictionary<string, DataStoreEntry> _data = new Dictionary<string, DataStoreEntry>(BackCompatibleStringComparer.Default);
```
JeremyKuhne, line 30: Because we expose a list of data types through the public methods this _may_ technically be a breaking change. I'm not sure how the ordering will be impacted, but we've not guaranteed a particular order. I don't think it will be a problem, but if we find it to be so we could write a typed wrapper for `Hashtable` that would keep the existing ordering behavior and satisfy our nullability requirements.

JeremyKuhne (approved):
Thanks for you work here!

PR #6386, opened 2021-12-22
Enable nullability in some Application classes

gpetrou:
## Proposed changes

- Enable nullability in some Application classes.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6386)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Application.ModalApplicationContext.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Application.ModalApplicationContext.cs
@@ -2,15 +2,13 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 namespace System.Windows.Forms
 {
     public sealed partial class Application
     {
         private class ModalApplicationContext : ApplicationContext
         {
-            private ThreadContext _parentWindowContext;
+            private ThreadContext? _parentWindowContext;
 
             private delegate void ThreadWindowCallback(ThreadContext context, bool onlyWinForms);
 
@@ -20,7 +18,7 @@ namespace System.Windows.Forms
 
             public void DisableThreadWindows(bool disable, bool onlyWinForms)
             {
-                Control parentControl = null;
+                Control? parentControl = null;
 
                 // Get ahold of the parent HWND -- if it's a different thread we need to do the disable
                 // over there too.  Note we only do this if we're parented by a Windows Forms parent.
@@ -33,16 +31,17 @@ namespace System.Windows.Forms
                     parentControl = Control.FromHandle(parentHandle);
 
                     _parentWindowContext = parentControl is not null && parentControl.InvokeRequired
-                        ? GetContextForHandle(parentControl) : null;
+                        ? GetContextForHandle(parentControl)
+                        : null;
                 }
 
                 // If we got a thread context, that means our parent is in a different thread, make the call on that thread.
                 if (_parentWindowContext is not null)
                 {
                     // In case we've already torn down, ask the context for this.
-                    parentControl ??= _parentWindowContext.ApplicationContext.MainForm;
+                    parentControl ??= _parentWindowContext.ApplicationContext!.MainForm;
 
-                    parentControl.Invoke(
+                    parentControl!.Invoke(
                         disable ? new ThreadWindowCallback(DisableThreadWindowsCallback) : new ThreadWindowCallback(EnableThreadWindowsCallback),
                         new object[] { _parentWindowContext, onlyWinForms });
                 }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.ComponentModel;
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
@@ -41,18 +39,18 @@ namespace System.Windows.Forms
             private static msoloop s_baseLoopReason;
 
             [ThreadStatic]
-            private static ThreadContext t_currentThreadContext;
+            private static ThreadContext? t_currentThreadContext;
 
-            internal ThreadExceptionEventHandler _threadExceptionHandler;
-            internal EventHandler _idleHandler;
-            internal EventHandler _enterModalHandler;
-            internal EventHandler _leaveModalHandler;
+            internal ThreadExceptionEventHandler? _threadExceptionHandler;
+            internal EventHandler? _idleHandler;
+            internal EventHandler? _enterModalHandler;
+            internal EventHandler? _leaveModalHandler;
 
             // Parking window list
             private readonly List<ParkingWindow> _parkingWindows = new();
-            private Control _marshalingControl;
-            private List<IMessageFilter> _messageFilters;
-            private List<IMessageFilter> _messageFilterSnapshot;
+            private Control? _marshalingControl;
+            private List<IMessageFilter>? _messageFilters;
+            private List<IMessageFilter>? _messageFilterSnapshot;
             private int _inProcessFilters;
             private HANDLE _handle;
             private readonly uint _id;
@@ -61,17 +59,17 @@ namespace System.Windows.Forms
             private int _modalCount;
 
             // Used for correct restoration of focus after modality
-            private WeakReference _activatingControlRef;
+            private WeakReference? _activatingControlRef;
 
             // IMsoComponentManager stuff
-            private IMsoComponentManager _componentManager;
+            private IMsoComponentManager? _componentManager;
             private bool _externalComponentManager;
             private bool _fetchingComponentManager;
 
             // IMsoComponent stuff
             private UIntPtr _componentID = s_invalidId;
-            private Form _currentForm;
-            private ThreadWindows _threadWindows;
+            private Form? _currentForm;
+            private ThreadWindows? _threadWindows;
             private int _disposeCount;   // To make sure that we don't allow
                                          // reentrancy in Dispose()
 
@@ -84,7 +82,7 @@ namespace System.Windows.Forms
             private bool _ourModalLoop;
 
             // A private field on Application that stores the callback delegate
-            private MessageLoopCallback _messageLoopCallback;
+            private MessageLoopCallback? _messageLoopCallback;
 
             /// <summary>
             ///  Creates a new thread context object.
@@ -114,13 +112,13 @@ namespace System.Windows.Forms
                 }
             }
 
-            public ApplicationContext ApplicationContext { get; private set; }
+            public ApplicationContext? ApplicationContext { get; private set; }
 
             /// <summary>
             ///  Retrieves the component manager for this process.  If there is no component manager
             ///  currently installed, we install our own.
             /// </summary>
-            internal unsafe IMsoComponentManager ComponentManager
+            internal unsafe IMsoComponentManager? ComponentManager
             {
                 get
                 {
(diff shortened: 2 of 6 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Application.ModalApplicationContext.cs
```diff
@@ -45,10 +43,10 @@ public void DisableThreadWindows(bool disable, bool onlyWinForms)
                     // In case we've already torn down, ask the context for this.
                     if (parentControl is null)
                     {
-                        parentControl = _parentWindowContext.ApplicationContext.MainForm;
+                        parentControl = _parentWindowContext.ApplicationContext?.MainForm;
                     }
 
-                    parentControl.Invoke(
+                    parentControl?.Invoke(
```
dreddy-work, line 49: we were expecting this as non-null earlier and throw exception if null.  I see MainForm could be null but ignoring null is a behavior change here. I would go with `!` operator here to preserve the existing behavior.

dreddy-work:
@Olina-Zhang, Can you please run some Unaware/SystemAware/PermonV2 scenarios against this change?

John-Qiao:
@dreddy-work we have completed the testing for Unaware/SystemAware/PermonV2 scenarios against this change, no new issue found.

src/System.Windows.Forms/src/System/Windows/Forms/Application.ModalApplicationContext.cs
dreddy-work: ping @gpetrou for feedback/correction here.

dreddy-work (approved):

src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs
```diff
@@ -568,15 +569,19 @@ private void DisposeParkingWindow()
                     // It is important that we just call DestroyHandle here
                     // and do not call Dispose.  Otherwise we would destroy
                     // controls that are living on the parking window.
+                    uint hwndThread = 0;
+                    if (_parkingWindows[0] is not null)
```
Tanya-Solyanik, line 573: Can parking window that is added to the list be null?

src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs
```diff
@@ -42,18 +40,18 @@ internal sealed class ThreadContext : MarshalByRefObject, IMsoComponent
...
+            internal EventHandler? _idleHandler;
+            internal EventHandler? _enterModalHandler;
+            internal EventHandler? _leaveModalHandler;
 
             // Parking window list
-            private readonly List<ParkingWindow> _parkingWindows = new List<ParkingWindow>();
-            private Control _marshalingControl;
-            private List<IMessageFilter> _messageFilters;
-            private List<IMessageFilter> _messageFilterSnapshot;
+            private readonly List<ParkingWindow?> _parkingWindows = new List<ParkingWindow?>();
```
dreddy-work, line 51: We may need to define this non nullable list and handle dispose properly.

src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs
```diff
@@ -315,19 +313,22 @@ internal ParkingWindow GetParkingWindowForContext(IntPtr context)
                 }
 
                 // Supported OS scenario.
-                foreach (ParkingWindow p in _parkingWindows)
+                foreach (ParkingWindow? p in _parkingWindows)
                 {
-                    if (User32.AreDpiAwarenessContextsEqual(p.DpiAwarenessContext, context))
+                    if (p is not null)
```
Tanya-Solyanik, line 318: `p == null` is an error condition and should not be ignored.

src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs
```diff
@@ -1067,7 +1072,11 @@ private unsafe void RunMessageLoopInner(msoloop reason, ApplicationContext conte
                     // If the owner window of the dialog is still enabled, disable it now.
                     // This can happen if the owner window is from a different thread or
                     // process.
-                    hwndOwner = User32.GetWindowLong(_currentForm, User32.GWL.HWNDPARENT);
+                    if (_currentForm is not null)
```
Tanya-Solyanik, line 1075: I would remove this test and let windows report error

src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs
```diff
@@ -1114,7 +1123,7 @@ private unsafe void RunMessageLoopInner(msoloop reason, ApplicationContext conte
 
                     if ((!fullModal && !localModal) || ComponentManager is ComponentManager)
                     {
-                        result = ComponentManager.FPushMessageLoop(_componentID, reason, null).IsTrue();
+                        result = ComponentManager!.FPushMessageLoop(_componentID, reason, null).IsTrue();
```
Tanya-Solyanik, line 1126: `is` operator should return `false`  for `null`... it's strange that the null-forgiving operator is required here...

src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs
```diff
@@ -1256,10 +1265,13 @@ internal bool ProcessFilters(ref User32.MSG msg, out bool modified)
                 // should not update the snapshot until the next message.
                 if (_messageFilters is not null && !GetState(STATE_FILTERSNAPSHOTVALID) && _inProcessFilters == 0)
                 {
-                    _messageFilterSnapshot.Clear();
-                    if (_messageFilters.Count > 0)
+                    if (_messageFilterSnapshot is not null)
```
Tanya-Solyanik, line 1268: This null-check is hiding an error that used to be reported.

src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs
RussKie: I think the condition should be written as 
```cs
                    if ((!fullModal && !localModal) || ComponentManager is ComponentManager componentManager )
                    {
                        result = componentManager.FPushMessageLoop(_componentID, reason, null).IsTrue();
```

src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs
RussKie: I don't think this would be the right course of action though. Above and below we're checking `_currentForm` for null, and this operation only makes sense if `_currentForm != null`.

src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs
```diff
@@ -687,7 +692,7 @@ private static void ExitCommon(bool disposing)
                         {
                             if (ctxs[i].ApplicationContext is not null)
                             {
-                                ctxs[i].ApplicationContext.ExitThread();
+                                ctxs[i].ApplicationContext!.ExitThread();
                             }
```
RussKie, line 696:
```suggestion
                            if (ctxs[i].ApplicationContext is ApplicationContext context)
                            {
                                context.ExitThread();
                            }
```

src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs
RussKie: The reason is `||` condition - if `(!fullModal && !localModal) == true` `ComponentManager` isn't checked, and hence can be null.

It is totally not obvious if `ComponentManager` is not null when `fullModal || localModal`....

RussKie (commented):
Here's my version of annotated ~[Application.ThreadContext.cs](https://github.com/dotnet/winforms/files/7879288/Application.ThreadContext.cs.txt)~ [Application.ThreadContext.cs](https://github.com/dotnet/winforms/files/7879610/Application.ThreadContext.cs.txt)

src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs
RussKie: `_messageFilters` and `_messageFilterSnapshot` are initialised simultaneously:
https://github.com/dotnet/winforms/blob/75dee0b5ba4ce9f1f433598c2aca7b1701f16893/src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs#L386-L396

There are no other places where either of these is initialised or nulled out.

src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs
Tanya-Solyanik: The difference is any side effects that `GetWindowLong` could have, which is probably none :) 
> If the function fails, the return value is zero. 

https://docs.microsoft.com/windows/win32/api/winuser/nf-winuser-getwindowlonga

RussKie: We will fail with NRE as `GetWindowLong` will try to get `Handle` property from a null reference.

src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs
kant2002: I believe removal of this lines
https://github.com/dotnet/winforms/blob/77b1fbc40e0bcc931bec28bfe0af2bc579793984/src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs#L581-L584

is fine, since all references to `_parkingWindows[i] ` would be removed after subsequent call to `Clear` method.

RussKie: This suggestion is still outstanding.

RussKie (approved):

kant2002:
@gpetrou does all feedback addressed? This PR in same location where I would like to try make additional improvements. @dreddy-work are you still have any improvements in this these changes?

dreddy-work:
> @gpetrou does all feedback addressed? This PR in same location where I would like to try make additional improvements. @dreddy-work are you still have any improvements in this these changes?

Yes. This needs to be looked with respect to ParkingWindows. I will have to go over those scenarios and make sure everything is alright. We can rebase if your improvements are outside of parkingWindows but this is not on priority list at the moment.

gpetrou:
I am not planning to do any more changes. I think I have addressed all feedback.

RussKie (approved):

gpetrou:
@dreddy-work perhaps we could merge this now?

gpetrou:
@dreddy-work, @Tanya-Solyanik @JeremyKuhne do you need me to do any more changes in this PR or could we merge it?

src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs
dreddy-work: We are adding null to list. I need relook and see if we could rearrange code in a better way. But for this nullability changes, it makes sense.

dreddy-work (approved):

src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs
JeremyKuhne: @dreddy-work is there a tracking issue for that?

dreddy-work: I didn't have but just created one: https://github.com/dotnet/winforms/issues/8252.

PR #6388, opened 2021-12-22
Enable nullability in WinCategoryAttribute

gpetrou:
## Proposed changes

- Enable nullability in WinCategoryAttribute.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6388)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/WinCategoryAttribute.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/WinCategoryAttribute.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.ComponentModel;
 using System.Diagnostics;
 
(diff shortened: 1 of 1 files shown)


dreddy-work (approved):

Tanya-Solyanik (approved):
Thank you!

src/System.Windows.Forms/src/System/Windows/Forms/WinCategoryAttribute.cs
```diff
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
```
RussKie, line 2: This type has only 2 uses, and it doesn't appear we have any "WinFormsCategory*" resources... Perhaps this class was/was meant to be used wider but wasn't.
Perhaps we should retire it completely

dreddy-work: I see bunch in `AxHost`.

dreddy-work (approved):

PR #6390, opened 2021-12-22
Enable nullability in GroupBoxRenderer

gpetrou:
## Proposed changes

- Enable nullability in GroupBoxRenderer.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6390)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/GroupBoxRenderer.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/GroupBoxRenderer.cs
@@ -2,9 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using System.Drawing;
 using System.Windows.Forms.VisualStyles;
 using static Interop;
@@ -18,7 +17,7 @@ namespace System.Windows.Forms
     {
         // Make this per-thread, so that different threads can safely use these methods.
         [ThreadStatic]
-        private static VisualStyleRenderer t_visualStyleRenderer = null;
+        private static VisualStyleRenderer? t_visualStyleRenderer = null;
         private static readonly VisualStyleElement s_groupBoxElement = VisualStyleElement.Button.GroupBox.Normal;
         private const int TextOffset = 8;
         private const int BoxHeaderWidth = 7;    // The groupbox frame shows 7 pixels before the caption.
@@ -79,7 +78,7 @@ namespace System.Windows.Forms
         /// <summary>
         ///  Renders a GroupBox control. Uses the text color specified by the theme.
         /// </summary>
-        public static void DrawGroupBox(Graphics g, Rectangle bounds, string groupBoxText, Font font, GroupBoxState state)
+        public static void DrawGroupBox(Graphics g, Rectangle bounds, string? groupBoxText, Font? font, GroupBoxState state)
             => DrawGroupBox(g, bounds, groupBoxText, font, TextFormatFlags.Top | TextFormatFlags.Left, state);
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/GroupBoxRenderer.cs
```diff
@@ -208,8 +207,11 @@ private static void DrawThemedGroupBoxNoText(Graphics g, Rectangle bounds, Group
 
             // Calculate area for background box
             Rectangle boxBounds = bounds;
-            boxBounds.Y += font.Height / 2;
-            boxBounds.Height -= font.Height / 2;
+            if (font is not null)
```
dreddy-work, line 210: Q: what happen when font is null?  when do we expect it to be null here?

dreddy-work: @gpetrou , I see you matched this to native method `MeasureText` that was imported here. However, subsequent `font` uses expect it to be non-null  and throw exception if happen to be null. this is altering that behavior and I do not recall a scenario where we could get font null and still draw `groupbox`. Did you experiment with Font null?

dreddy-work:
ping @gpetrou to act on last feedback.

src/System.Windows.Forms/src/System/Windows/Forms/GroupBoxRenderer.cs
```diff
@@ -18,7 +17,7 @@ public static class GroupBoxRenderer
     {
         // Make this per-thread, so that different threads can safely use these methods.
         [ThreadStatic]
-        private static VisualStyleRenderer t_visualStyleRenderer = null;
+        private static VisualStyleRenderer? t_visualStyleRenderer = null;
```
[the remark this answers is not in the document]
gpetrou: We use `[MemberNotNull(nameof(t_visualStyleRenderer))]` in `InitializeRenderer` which is called in a few places in this class. We have used a similar approach in other renderer classes.

src/System.Windows.Forms/src/System/Windows/Forms/GroupBoxRenderer.cs
gpetrou: I was trying to be consistent with the changes we did in other renderer classes such as in https://github.com/dotnet/winforms/pull/6321 and https://github.com/dotnet/winforms/pull/6392. Perhaps what I should do is change the Font in the other classes and here to non-nullable? @dreddy-work and @RussKie what do you think?

RussKie: @dreddy-work this is a public API that never asserted that font must be not null, so it _may_ be null. Also the code downstream doesn't require font to be not null either.

RussKie (approved):

src/System.Windows.Forms/src/System/Windows/Forms/GroupBoxRenderer.cs
dreddy-work: @RussKie , if font was null, we were throwing null reference earlier( line 212) when we access font.Height.  Now we don't. Why did you say downstream code doesn't require font to be not null?

RussKie: Both `TextRenderer.MeasureText(...)` and `TextRenderer.DrawText(...)` calls directly above accept `font` as null.
Also this method is called from a public API and we never made any assumptions or assertions that the supplied font isn't a null:
https://github.com/dotnet/winforms/blob/1f4032c1ba6e1bb97aee586605616a4f01cbaf45/src/System.Windows.Forms/src/System/Windows/Forms/GroupBoxRenderer.cs#L130-L157

It's incorrect to annotate this method in its current form as requiring `font` to be not-null, and to throw an ANE would be a breaking change:
https://github.com/dotnet/winforms/blob/1f4032c1ba6e1bb97aee586605616a4f01cbaf45/src/System.Windows.Forms/src/System/Windows/Forms/GroupBoxRenderer.cs#L130-L138

RussKie (approved):

RussKie:
Taking as is, we can revisit any concerns later.

RussKie (approved):

PR #4931, opened 2021-05-18
Enable nullability in some ToolStrip EventArgs

gpetrou:
## Proposed changes

- Enable nullability in some ToolStrip EventArgs.

PR #6391, opened 2021-12-22
Enable nullability in ToolStripSystemRenderer

gpetrou:
## Proposed changes

- Enable nullability in ToolStripSystemRenderer.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6391)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripSystemRenderer.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripSystemRenderer.cs
@@ -187,11 +187,11 @@ namespace System.Windows.Forms
         /// </summary>
         protected override void OnRenderToolStripBackground(ToolStripRenderEventArgs e)
         {
-            ToolStrip? toolStrip = e.ToolStrip;
+            ToolStrip toolStrip = e.ToolStrip;
             Graphics g = e.Graphics;
             Rectangle bounds = e.AffectedBounds;
 
-            if (toolStrip is not null && !ShouldPaintBackground(toolStrip))
+            if (!ShouldPaintBackground(toolStrip))
             {
                 return;
             }
@@ -210,7 +210,7 @@ namespace System.Windows.Forms
                 {
                     FillBackground(g, bounds, (toolStrip is ToolStripDropDown) ? SystemColors.ControlLight : e.BackColor);
                 }
-                else if (toolStrip is not null && toolStrip.IsDropDown)
+                else if (toolStrip.IsDropDown)
                 {
                     FillBackground(g, bounds, (!ToolStripManager.VisualStylesEnabled) ?
                                          e.BackColor : SystemColors.Menu);
@@ -222,12 +222,9 @@ namespace System.Windows.Forms
                 }
                 else if (ToolStripManager.VisualStylesEnabled && VisualStyleRenderer.IsElementDefined(VisualStyleElement.Rebar.Band.Normal))
                 {
-                    VisualStyleRenderer? vsRenderer = VisualStyleRenderer;
-                    if (vsRenderer is not null)
-                    {
-                        vsRenderer.SetParameters(VisualStyleElement.ToolBar.Bar.Normal);
-                        vsRenderer.DrawBackground(g, bounds);
-                    }
+                    VisualStyleRenderer vsRenderer = VisualStyleRenderer!;
+                    vsRenderer.SetParameters(VisualStyleElement.ToolBar.Bar.Normal);
+                    vsRenderer.DrawBackground(g, bounds);
                 }
                 else
                 {
@@ -242,12 +239,8 @@ namespace System.Windows.Forms
         /// </summary>
         protected override void OnRenderToolStripBorder(ToolStripRenderEventArgs e)
         {
-            ToolStrip? toolStrip = e.ToolStrip;
-            Rectangle bounds = Rectangle.Empty;
-            if (toolStrip is not null)
-            {
-                bounds = toolStrip.ClientRectangle;
-            }
+            ToolStrip toolStrip = e.ToolStrip;
+            Rectangle bounds = toolStrip.ClientRectangle;
 
             if (toolStrip is StatusStrip)
             {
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ToolStripSystemRenderer.cs
```diff
@@ -224,10 +222,12 @@ protected override void OnRenderToolStripBackground(ToolStripRenderEventArgs e)
                 }
                 else if (ToolStripManager.VisualStylesEnabled && VisualStyleRenderer.IsElementDefined(VisualStyleElement.Rebar.Band.Normal))
                 {
-                    VisualStyleRenderer vsRenderer = VisualStyleRenderer;
-
-                    vsRenderer.SetParameters(VisualStyleElement.ToolBar.Bar.Normal);
-                    vsRenderer.DrawBackground(g, bounds);
+                    VisualStyleRenderer? vsRenderer = VisualStyleRenderer;
+                    if (vsRenderer is not null)
```
dreddy-work, line 226: This is behavioral change. We may want to go with `! `here. Do you have a scenario where this could be null and expected behavior is correct?

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripSystemRenderer.cs
```diff
@@ -292,27 +295,30 @@ protected override void OnRenderGrip(ToolStripGripRenderEventArgs e)
 
             if (ToolStripManager.VisualStylesEnabled && VisualStyleRenderer.IsElementDefined(VisualStyleElement.Rebar.Gripper.Normal))
             {
-                VisualStyleRenderer vsRenderer = VisualStyleRenderer;
+                VisualStyleRenderer? vsRenderer = VisualStyleRenderer;
 
                 if (verticalGrip)
                 {
-                    vsRenderer.SetParameters(VisualStyleElement.Rebar.Gripper.Normal);
+                    vsRenderer?.SetParameters(VisualStyleElement.Rebar.Gripper.Normal);
```
dreddy-work, line 302: same here.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripSystemRenderer.cs
```diff
@@ -189,11 +187,11 @@ private static ToolBarState GetToolBarState(ToolStripItem item)
         /// </summary>
         protected override void OnRenderToolStripBackground(ToolStripRenderEventArgs e)
         {
-            ToolStrip toolStrip = e.ToolStrip;
+            ToolStrip? toolStrip = e.ToolStrip;
```
Tanya-Solyanik, line 190: Would it make sense to declare ToolStripRenderEventArgs.ToolStrip property as not nullable?

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripSystemRenderer.cs
Tanya-Solyanik: `!` sounds reasonable, because we are within the "visual styles are enabled" case.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripSystemRenderer.cs
RussKie: It was annotated as nullable in https://github.com/dotnet/winforms/pull/4931, but in hindsight it was probably a mistake. We should check and throw if it's null.

gpetrou: I have created https://github.com/dotnet/winforms/pull/6484 about this.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripSystemRenderer.cs
```diff
@@ -784,22 +814,25 @@ private static void RenderStatusStripBackground(ToolStripRenderEventArgs e)
         private static void RenderLabelInternal(ToolStripItemRenderEventArgs e)
         {
             // don't call RenderItemInternal, as we NEVER want to paint hot.
-            ToolStripItem item = e.Item;
+            ToolStripItem? item = e.Item;
```
gpetrou, line 817: Looks like I should also make this non-nullable.

Tanya-Solyanik: Makes sense

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripSystemRenderer.cs
```diff
@@ -150,7 +148,7 @@ private static ToolBarState GetSplitButtonToolBarState(ToolStripSplitButton butt
         /// <summary>
         ///  translates the ToolStrip item state into a toolbar state, which is something the renderer understands
         /// </summary>
-        private static ToolBarState GetToolBarState(ToolStripItem item)
+        private static ToolBarState GetToolBarState(ToolStripItem? item)
```
RussKie, line 151: I've looked through the call stacks and I don't think this can be null.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripSystemRenderer.cs
gpetrou: I created https://github.com/dotnet/winforms/pull/6507 about this.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripSystemRenderer.cs
```diff
@@ -506,8 +496,7 @@ protected override void OnRenderSeparator(ToolStripSeparatorRenderEventArgs e)
         protected override void OnRenderToolStripStatusLabelBackground(ToolStripItemRenderEventArgs e)
         {
             RenderLabelInternal(e);
-            ToolStripStatusLabel item = e.Item as ToolStripStatusLabel;
-
+            ToolStripStatusLabel item = (e.Item as ToolStripStatusLabel)!;
```
RussKie, line 499: We need to null-check here (like we do in `OnRenderMenuItemBackground`). We can't assume `e.Item` is `ToolStripStatusLabel`.
Ditto below.

gpetrou: Oh, I got confused by the previous comments.

RussKie (approved):

RussKie (approved):

PR #6395, opened 2021-12-22
Enable nullability in MouseHoverTimer

gpetrou:
## Proposed changes

- Enable nullability in MouseHoverTimer.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6395)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/MouseHoverTimer.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/MouseHoverTimer.cs
@@ -6,7 +6,7 @@ namespace System.Windows.Forms
 {
     internal class MouseHoverTimer : IDisposable
     {
-        private Timer? _mouseHoverTimer = new Timer();
+        private Timer _mouseHoverTimer = new Timer();
 
         // Consider - weak reference?
         private ToolStripItem? _currentItem;
@@ -25,7 +25,7 @@ namespace System.Windows.Forms
             }
 
             _currentItem = item;
-            if (_currentItem is not null && _mouseHoverTimer is not null)
+            if (_currentItem is not null)
             {
                 _mouseHoverTimer.Enabled = true;
             }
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/MouseHoverTimer.cs
```diff
@@ -2,16 +2,14 @@
...
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 namespace System.Windows.Forms
 {
     internal class MouseHoverTimer : IDisposable
     {
-        private Timer _mouseHoverTimer = new Timer();
+        private Timer? _mouseHoverTimer = new Timer();
```
dreddy-work, line 9: why?

src/System.Windows.Forms/src/System/Windows/Forms/MouseHoverTimer.cs
```diff
@@ -27,22 +25,26 @@ public void Start(ToolStripItem item)
             }
 
             _currentItem = item;
```
dreddy-work, line 27: `item `should be nullable here.

Tanya-Solyanik (approved):
Thank you!

src/System.Windows.Forms/src/System/Windows/Forms/MouseHoverTimer.cs
```diff
@@ -2,47 +2,49 @@
...
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 namespace System.Windows.Forms
 {
     internal class MouseHoverTimer : IDisposable
     {
-        private Timer _mouseHoverTimer = new Timer();
+        private Timer? _mouseHoverTimer = new Timer();
```
dreddy-work, line 9: @gpetrou , the way this is initialized, we do not expect `_mouseHoverTimer` to be null in any practical scenarios here. I understand that `null` assignment statement in the `dispose ` method is forcing this.  We think the right thing to do here is remove the `null `assignment statement in the dispose method and keep this non-nullable.

dreddy-work (changes requested):

dreddy-work (approved):

Tanya-Solyanik (approved):
Thank you!

PR #6401, opened 2021-12-22
Use generic list in Lines property of TextBoxBase

gpetrou:
## Proposed changes

- Use generic list in Lines property of TextBoxBase.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6401)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.cs
@@ -4,7 +4,6 @@
 
 #nullable disable
 
-using System.Collections;
 using System.Collections.Specialized;
 using System.ComponentModel;
 using System.Diagnostics;
@@ -613,7 +612,7 @@ namespace System.Windows.Forms
             get
             {
                 string text = Text;
-                ArrayList list = new ArrayList();
+                List<string> list = new();
 
                 int lineStart = 0;
                 while (lineStart < text.Length)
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.cs
```diff
@@ -613,7 +612,7 @@ public string[] Lines
             get
             {
                 string text = Text;
-                ArrayList list = new ArrayList();
+                List<string> list = new List<string>();
```
dreddy-work, line 615:
```suggestion
                List<string> list = new ();
```

dreddy-work:
LGTM

gpetrou:
@dreddy-work can we merge this?

src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.cs
```diff
@@ -648,10 +647,10 @@ public string[] Lines
                 // Corner case -- last character in Text is a new line; need to add blank line to list
                 if (text.Length > 0 && (text[text.Length - 1] == '\r' || text[text.Length - 1] == '\n'))
                 {
-                    list.Add("");
+                    list.Add(string.Empty);
                 }
 
-                return (string[])list.ToArray(typeof(string));
+                return list.ToArray();
```
dreddy-work, line 653: @gpetrou , i just put it on hold because `list `is not explicitly disposed here after creating array and returning it. I do not have a suggestion if there is a better way to do that. Will merge if i do not have any other suggestion here.

RussKie: List<> is not disposable, so it will be collected once it goes out of scope.

RussKie (approved):

src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.cs
dreddy-work: I know. I was thinking if revisiting logic here to avoid recreation of array. Taking as-is for now. we have implemented pooled lists in designer to improve performance.

PR #6404, opened 2021-12-22
Enable nullability in DataStreamFromComStream

gpetrou:
## Proposed changes

- Enable nullability in DataStreamFromComStream.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6404)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataStreamFromComStream.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataStreamFromComStream.cs
@@ -2,19 +2,17 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using static Interop;
 
 namespace System.Windows.Forms
 {
     internal class DataStreamFromComStream : Stream
     {
-        private Ole32.IStream comStream;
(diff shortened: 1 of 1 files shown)


dreddy-work (commented):
Behavioral change. holding on team.

Tanya-Solyanik (approved):

src/System.Windows.Forms/src/System/Windows/Forms/DataStreamFromComStream.cs
```diff
@@ -2,19 +2,17 @@
...
-#nullable disable
-
 using static Interop;
 
 namespace System.Windows.Forms
 {
     internal class DataStreamFromComStream : Stream
     {
-        private Ole32.IStream comStream;
+        private Ole32.IStream? _comStream;
```
RussKie, line 11: How is this nullable if in the .ctor it's declared as not nullable?

gpetrou: I had set this as nullable because we set it as null in Dispose. I used `null!` instead. Perhaps that is better?

RussKie (approved):

PR #6407, opened 2021-12-23
Enable nullability in EnumUnknown

gpetrou:
## Proposed changes

- Enable nullability in EnumUnknown.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6407)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/AxHost.EnumUnknown.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/AxHost.EnumUnknown.cs
@@ -46,16 +46,13 @@ namespace System.Windows.Forms
                 }
                 else
                 {
-                    if (_array is not null)
+                    for (; _location < _size && fetched < celt; ++_location)
                     {
-                        for (; _location < _size && fetched < celt; ++_location)
+                        if (_array![_location] is not null)
                         {
-                            if (_array[_location] is not null)
-                            {
-                                Marshal.WriteIntPtr(rgelt, Marshal.GetIUnknownForObject(_array[_location]));
-                                rgelt = (IntPtr)((long)rgelt + (long)sizeof(IntPtr));
-                                ++fetched;
-                            }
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/AxHost.EnumUnknown.cs
```diff
@@ -13,21 +11,21 @@ public abstract partial class AxHost
...
+            private int _loc;
+            private readonly int _size;
 
             internal EnumUnknown(object[] arr)
             {
                 //if (AxHTraceSwitch.TraceVerbose) Debug.WriteObject(arr);
-                this.arr = arr;
-                loc = 0;
-                size = (arr is null) ? 0 : arr.Length;
+                _arr = arr;
```
dreddy-work, line 21: seems `arr `can be null

dreddy-work (changes requested):

dreddy-work (approved):

gpetrou:
@dreddy-work maybe we could merge this now that the build completed successfully?

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.EnumUnknown.cs
```diff
@@ -48,13 +46,16 @@ unsafe HRESULT Ole32.IEnumUnknown.Next(uint celt, IntPtr rgelt, uint* pceltFetch
                 }
                 else
                 {
-                    for (; _location < _size && fetched < celt; ++_location)
+                    if (_array is not null)
```
dreddy-work, line 49: This is unnecessary check. If location < _size,   why would _array be null. Old code was not checking for it either.

gpetrou: I used `!` instead.

dreddy-work (approved):

PR #6412, opened 2021-12-23
Enable nullability in ActiveXImpl

gpetrou:
## Proposed changes

- Enable nullability in ActiveXImpl.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6412)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/AxHost.PropertyBagStream.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/AxHost.PropertyBagStream.cs
@@ -33,12 +33,13 @@ namespace System.Windows.Forms
                 }
             }
 
-            HRESULT Oleaut32.IPropertyBag.Read(string pszPropName, ref object pVar, Oleaut32.IErrorLog pErrorLog)
+            HRESULT Oleaut32.IPropertyBag.Read(string pszPropName, out object pVar, Oleaut32.IErrorLog pErrorLog)
             {
                 Debug.WriteLineIf(s_axHTraceSwitch.TraceVerbose, $"Reading property {pszPropName} from OCXState propertybag.");
 
                 if (!_bag.Contains(pszPropName))
                 {
+                    pVar = null;
                     return HRESULT.E_INVALIDARG;
                 }
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
@@ -1677,7 +1677,7 @@ namespace System.Windows.Forms
                     /// </summary>
                     public ComConnectionPoint FindConnectionPoint(Type eventInterface)
                     {
-                        FindConnectionPointD findConnectionPoint = (FindConnectionPointD)Marshal.GetDelegateForFunctionPointer(_vtbl!.FindConnectionPointPtr, typeof(FindConnectionPointD));
+                        FindConnectionPointD findConnectionPoint = (FindConnectionPointD)Marshal.GetDelegateForFunctionPointer(_vtbl.FindConnectionPointPtr, typeof(FindConnectionPointD));
 
                         Guid iid = eventInterface.GUID;
                         int hresult = findConnectionPoint.Invoke(handle, ref iid, out IntPtr result);
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
@@ -1302,11 +1302,13 @@ namespace System.Windows.Forms
                         }
                     }
 
+                    object? obj;
+
                     // Special properties that we care about
                     switch (dispID)
                     {
                         case DispatchID.AMBIENT_UIDEAD:
-                            if (GetAmbientProperty(DispatchID.AMBIENT_UIDEAD, out object? obj))
+                            if (GetAmbientProperty(DispatchID.AMBIENT_UIDEAD, out obj))
                             {
                                 _activeXState[s_uiDead] = (bool)obj!;
                             }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
@@ -1304,7 +1304,7 @@ namespace System.Windows.Forms
                     }
 
                     // Special properties that we care about
-                    object? obj = new object();
+                    object? obj = null;
 
                     switch (dispID)
                     {
@@ -1930,7 +1930,7 @@ namespace System.Windows.Forms
                 }
 
                 // Get the ambient properties that effect us.
-                object? obj = new object();
+                object? obj = null;
                 if (GetAmbientProperty(DispatchID.AMBIENT_UIDEAD, ref obj))
                 {
                     _activeXState[s_uiDead] = (bool)obj!;
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
@@ -1662,15 +1662,14 @@ namespace System.Windows.Forms
 
                     private readonly VTABLE _vtbl;
 
+                    [StructLayout(LayoutKind.Sequential)]
                     private struct VTABLE
                     {
-#pragma warning disable CS0649
                         public IntPtr QueryInterfacePtr;
                         public IntPtr AddRefPtr;
                         public IntPtr ReleasePtr;
                         public IntPtr EnumConnectionPointsPtr;
                         public IntPtr FindConnectionPointPtr;
-#pragma warning restore CS0649
                     }
 
                     /// <summary>
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
@@ -148,7 +148,7 @@ namespace System.Windows.Forms
                             try
                             {
                                 Debug.Assert(obj is not null, "GetAmbientProperty failed");
-                                Debug.WriteLineIf(CompModSwitches.ActiveX.TraceInfo, "Object font type=" + obj!.GetType().FullName);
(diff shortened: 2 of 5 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
```diff
@@ -136,20 +134,20 @@ internal Color AmbientBackColor
...
                     if (prop.Empty)
                     {
-                        object obj = null;
+                        object? obj = null;
                         if (GetAmbientProperty(DispatchID.AMBIENT_FONT, ref obj))
                         {
                             try
                             {
-                                Debug.WriteLineIf(CompModSwitches.ActiveX.TraceInfo, "Object font type=" + obj.GetType().FullName);
+                                Debug.WriteLineIf(CompModSwitches.ActiveX.TraceInfo, "Object font type=" + obj!.GetType().FullName);
```
Tanya-Solyanik, line 150: Are we guaranteed that obj is not null? Then next function in this file has a null-test after a successful call to `GetAmbientProperty`

Tanya-Solyanik (commented):
Please see my question, it is applicable in multiple places in this change.

src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
gpetrou: There is a Debug.Assert here, this is why I did not change it.

RussKie: I would swap the statements around, so that we assert first, and only then writeline.

src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
```diff
@@ -136,21 +134,21 @@ internal Color AmbientBackColor
...
                     {
-                        object obj = null;
+                        object? obj = null;
                         if (GetAmbientProperty(DispatchID.AMBIENT_FONT, ref obj))
                         {
                             try
                             {
-                                Debug.WriteLineIf(CompModSwitches.ActiveX.TraceInfo, "Object font type=" + obj.GetType().FullName);
                                 Debug.Assert(obj is not null, "GetAmbientProperty failed");
+                                Debug.WriteLineIf(CompModSwitches.ActiveX.TraceInfo, "Object font type=" + obj!.GetType().FullName);
```
RussKie, line 151:
```suggestion
                                Debug.WriteLineIf(CompModSwitches.ActiveX.TraceInfo, "Object font type=" + obj.GetType().FullName);
```

src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
```diff
@@ -1663,7 +1660,7 @@ public ComConnectionPointContainer(object obj, bool addRefIntPtr)
                         _vtbl = LoadVtable<VTABLE>();
                     }
 
-                    private readonly VTABLE _vtbl;
+                    private readonly VTABLE? _vtbl;
```
RussKie, line 1663: `VTABLE` should be turned into a `struct`, and then there are no issues with NRTA.
Also `protected V LoadVtable<V>() where V : struct`.

@JeremyKuhne @AaronRobinsonMSFT please correct, if it's wrong.

AaronRobinsonMSFT: Not wrong, you are correct.

The `Marshal.PtrToStructure<>()` should only return a `null` value in cases where the input is `IntPtr.Zero`. In general, I find the nullable logic to be of dubious value in interop code. It breeds a false sense of safety since almost any use of the `Marshal` class is inherently unsafe.

src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
```diff
@@ -1307,22 +1304,22 @@ internal void OnAmbientPropertyChange(DispatchID dispID)
                     }
 
                     // Special properties that we care about
-                    object obj = new object();
+                    object? obj = new object();
```
AaronRobinsonMSFT, line 1307:
```suggestion
                    object? obj = null;
```

No reason to allocate something that is never used.

src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
```diff
@@ -583,7 +581,7 @@ internal unsafe HRESULT GetAdvise(DVASPECT* pAspects, ADVF* pAdvf, IAdviseSink[]
             ///  Helper function to retrieve an ambient property.  Returns false if the
             ///  property wasn't found.
             /// </summary>
-            private bool GetAmbientProperty(DispatchID dispid, ref object obj)
+            private bool GetAmbientProperty(DispatchID dispid, ref object? obj)
```
AaronRobinsonMSFT, line 584: It seems that the intent of this is to really be an out so...

```suggestion
            private bool GetAmbientProperty(DispatchID dispid, out object? obj)
```

src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
```diff
@@ -1933,15 +1930,15 @@ internal void SetClientSite(IOleClientSite value)
                 }
 
                 // Get the ambient properties that effect us.
-                object obj = new object();
+                object? obj = new object();
```
AaronRobinsonMSFT, line 1933:
```suggestion
                object? obj = null;
```

src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
```diff
@@ -2518,7 +2515,7 @@ internal void Read(IStream istream)
                     }
                 }
 
-                HRESULT Oleaut32.IPropertyBag.Read(string pszPropName, ref object pVar, Oleaut32.IErrorLog pErrorLog)
+                HRESULT Oleaut32.IPropertyBag.Read(string pszPropName, ref object? pVar, Oleaut32.IErrorLog? pErrorLog)
```
AaronRobinsonMSFT, line 2518: Seems like the intent of this interface should be `out object? pVar`

src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
```diff
@@ -1307,22 +1303,20 @@ internal void OnAmbientPropertyChange(DispatchID dispID)
                     }
 
                     // Special properties that we care about
-                    object obj = new object();
-
                     switch (dispID)
                     {
                         case DispatchID.AMBIENT_UIDEAD:
-                            if (GetAmbientProperty(DispatchID.AMBIENT_UIDEAD, ref obj))
+                            if (GetAmbientProperty(DispatchID.AMBIENT_UIDEAD, out object? obj))
```
AaronRobinsonMSFT, line 1309: Wow. This is a bug in my opinion. Declaring a local in one `case` shouldn't be usable in another case. Can we move the single declaration of `object? obj` to be outside of the `switch` instead of buried in the first `case`?

gpetrou: Sure.

src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
```diff
@@ -1680,7 +1677,7 @@ private class VTABLE
                     /// </summary>
                     public ComConnectionPoint FindConnectionPoint(Type eventInterface)
                     {
-                        FindConnectionPointD findConnectionPoint = (FindConnectionPointD)Marshal.GetDelegateForFunctionPointer(_vtbl.FindConnectionPointPtr, typeof(FindConnectionPointD));
+                        FindConnectionPointD findConnectionPoint = (FindConnectionPointD)Marshal.GetDelegateForFunctionPointer(_vtbl!.FindConnectionPointPtr, typeof(FindConnectionPointD));
```
RussKie, line 1680: Since `VTABLE` is a struct, it can't be null
```suggestion
                        FindConnectionPointD findConnectionPoint = (FindConnectionPointD)Marshal.GetDelegateForFunctionPointer(_vtbl.FindConnectionPointPtr, typeof(FindConnectionPointD));
```

src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
```diff
@@ -43,17 +41,17 @@ private unsafe class ActiveXImpl : MarshalByRefObject, IWindowTarget
...
             private readonly IWindowTarget _controlWindowTarget;
             private Rectangle? _lastClipRect;
 
-            private IOleClientSite _clientSite;
-            private IOleInPlaceUIWindow _inPlaceUiWindow;
-            private IOleInPlaceFrame _inPlaceFrame;
+            private IOleClientSite? _clientSite;
+            private IOleInPlaceUIWindow? _inPlaceUiWindow;
+            private IOleInPlaceFrame? _inPlaceFrame;
             private readonly ArrayList _adviseList;
```
RussKie, line 53: We can probably change this to
```suggestion
            private readonly List<IAdviseSink> _adviseList;
```
...as this will remove associated bang operators . We can do it later though.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.PropertyBagStream.cs
```diff
@@ -33,12 +33,13 @@ internal void Read(Stream stream)
                 }
             }
 
-            HRESULT Oleaut32.IPropertyBag.Read(string pszPropName, ref object pVar, Oleaut32.IErrorLog pErrorLog)
+            HRESULT Oleaut32.IPropertyBag.Read(string pszPropName, out object pVar, Oleaut32.IErrorLog pErrorLog)
```
AaronRobinsonMSFT, line 36: Doesn't this signature need to be updated with `?`?

gpetrou: We have not enabled nullability on this file yet. I guess it will be done in a future PR.

RussKie (approved):

PR #6422, opened 2021-12-26
Enable nullability in ToolStripPanelRenderEventArgs

gpetrou:
## Proposed changes

- Enable nullability in ToolStripPanelRenderEventArgs.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6422)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripPanelRenderEventArgs.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripPanelRenderEventArgs.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.Drawing;
 
 namespace System.Windows.Forms
@@ -13,9 +11,9 @@ namespace System.Windows.Forms
         /// <summary>
         ///  This class represents all the information to render the toolStrip
         /// </summary>
-        public ToolStripPanelRenderEventArgs(Graphics g, ToolStripPanel toolStripPanel)
+        public ToolStripPanelRenderEventArgs(Graphics g, ToolStripPanel? toolStripPanel)
         {
-            Graphics = g;
+            Graphics = g.OrThrowIfNull();
             ToolStripPanel = toolStripPanel;
         }
 
@@ -27,7 +25,7 @@ namespace System.Windows.Forms
         /// <summary>
         ///  Represents which toolStrip was affected by the click
         /// </summary>
-        public ToolStripPanel ToolStripPanel { get; }
+        public ToolStripPanel? ToolStripPanel { get; }
 
         public bool Handled { get; set; }
     }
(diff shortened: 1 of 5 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ToolStripPanelRenderEventArgs.cs
```diff
@@ -13,9 +11,9 @@ public class ToolStripPanelRenderEventArgs : EventArgs
         /// <summary>
         ///  This class represents all the information to render the toolStrip
         /// </summary>
-        public ToolStripPanelRenderEventArgs(Graphics g, ToolStripPanel toolStripPanel)
+        public ToolStripPanelRenderEventArgs(Graphics g, ToolStripPanel? toolStripPanel)
         {
-            Graphics = g;
+            Graphics = g.OrThrowIfNull();
```
dreddy-work, line 16: how come tests calling with `null `are passing here? @gpetrou  can you check if there is inconsistency in tests where they are calling with null for `Graphics `?

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripPanelRenderEventArgs.cs
```diff
@@ -27,7 +25,7 @@ public ToolStripPanelRenderEventArgs(Graphics g, ToolStripPanel toolStripPanel)
         /// <summary>
         ///  Represents which toolStrip was affected by the click
         /// </summary>
-        public ToolStripPanel ToolStripPanel { get; }
+        public ToolStripPanel? ToolStripPanel { get; }
```
dreddy-work, line 28: same here. Would this be null apart from tests?

RussKie: Similar to https://github.com/dotnet/winforms/pull/6391#discussion_r778588402

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripPanelRenderEventArgs.cs
gpetrou: In which tests are you referring to?

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripPanelRenderEventArgs.cs
gpetrou: I will make this non-nullable.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripPanelRenderEventArgs.cs
dreddy-work: @gpetrou , `ToolStripPanelRenderEventArgs_TestData` is passing graphics `null` and is consumed by `ToolStripRenderer_DrawToolStripPanelBackground_Invoke_CallsRenderToolStripPanelBackground`.

gpetrou: I don't see any null Graphics there. Only ToolStripPanelRenderEventArgs is null which is similar to what happens in many other tests.

RussKie (approved):

RussKie:
Docs: https://github.com/dotnet/docs/issues/26720

PR #6428, opened 2021-12-26
Enable nullability in SendKeys

gpetrou:
## Proposed changes

- Enable nullability in SendKeys.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6428)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/SendKeys.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/SendKeys.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.ComponentModel;
 using System.Globalization;
 using System.Security;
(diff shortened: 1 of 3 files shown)


Tanya-Solyanik (approved):
Thank you!

src/System.Windows.Forms/src/System/Windows/Forms/SendKeys.cs
```diff
@@ -674,12 +672,12 @@ private unsafe static void SendInput(ReadOnlySpan<byte> oldKeyboardState, SKEven
 
                 try
                 {
-                    eventsTotal = s_events.Count;
+                    eventsTotal = s_events!.Count;
```
RussKie, line 675: I find this rather unsettling. The code is rather convoluted and I couldn't confirm that we would have called `AddEvent()` before getting here. So instead of using a `!` it probably would be more correct to check `s_events` for null and exit, if it is so.

gpetrou: Since this was already merged I created https://github.com/dotnet/winforms/pull/6470.

src/System.Windows.Forms/src/System/Windows/Forms/SendKeys.cs
```diff
@@ -81,12 +79,12 @@ public partial class SendKeys
...
         private static bool s_stopHook;
         private static IntPtr s_hhook;
-        private static User32.HOOKPROC s_hook;
+        private static User32.HOOKPROC? s_hook;
 
         /// <summary>
         ///  Vector of events that we have yet to post to the journaling hook.
         /// </summary>
-        private static Queue<SKEvent> s_events;
+        private static Queue<SKEvent>? s_events;
```
dreddy-work, line 87: Ia right fix seems to be making this as a `readonly` assignment and make it non nullable.

PR #6435, opened 2021-12-28
Extracting test control items from resource files to code

### LINKED ISSUE #6430: Extract test controls from resources to code
Some items at some forms (`ListViewTest`, `TreeViewTest`) of our internal test application (`WinformsControlsTest`) are created from the resources:
image
image

These items are not visible at the Managed Resources Editor:
image

We should extract the creation of such items from the `resx`-files to the direct code initialization to make their properties readable, simplify their modification if needed, etc.

ArtemTatarinov:
Fixes #6430

## Proposed changes
- Extracted control items for test forms at our internal test application from the resource files to the direct code initialization
- Renamed some items so it would be easier to differ one item from another

**Before the fix**
image
image
image

**After the fix**
image
image
image

## Customer Impact
- No

## Regression? 
- No

## Risk
- Minimal

## Test methodology 
- Manual (see the screenshots above)

## Test environment(s) 
.NET SDK 7.0.0-alpha.1.21606.7
Microsoft Windows [Version 10.0.19043.1415]

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6435)


WPMGPRoSToTeMa (approved):
👍

RussKie:
One of the reasons to have resources is to ensure the associated code is exercised, i.e. we can serialise and deserialise the resources both at design- and runtime.

ArtemTatarinov:
We've discussed this refactoring with Igor and Vladimir via e-mail and decided to perform it.

dreddy-work (approved):
LGTM.

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/ListViewTest.Designer.cs
```diff
@@ -82,6 +86,33 @@ private void InitializeComponent()
             this.listView1.SmallImageList = this.imageList1;
             this.listView1.TabIndex = 0;
             this.listView1.UseCompatibleStateImageBehavior = false;
+            //
+            // listView1_Group1
+            //
+            this.listView1_Group1.Header = "ListViewGroup";
```
Tanya-Solyanik, line 92: Is it not possible to put these in the resources?

ArtemTatarinov: We can put the string with a group header name into the resources, but since it's only our test application, it's more convenient to have it as a plain string as other control item names, I guess. Also, it's handier to have all controls creation and initialization directly at the code, so we can easily view and/or modify them. But maybe I don't quite get the review point?

Tanya-Solyanik: I don't understand the reason for this change.

>Extracted control items for test forms at our internal test application from the resource files to the direct code initialization

 After this change we no longer test resource serialization and no new serialization tests are added to compensate. XML resource file is still quite readable in the text editor, and names will be visible there, I don't think that the win in readability is worth the loss in test coverage.

>Renamed some items so it would be easier to differ one item from another

This is a reasonable change, but it could be done in the resources as well.

ArtemTatarinov: Maybe I should write special unit tests for checking serialization and deserialization (using the `ResXResourceReader` and `ResXResourceWriter` classes and/or without them)? This check will be more reliable than deserialization by pressing `TreeView, ImageList`/`ListView` buttons at our test app (I guess, these buttons may not be pressed for days/weeks), and serialization will be covered too, and our test forms will have unified control initialization gathered in one place without the need to check/modify it at two places (code and XML).

Tanya-Solyanik: Integration tests are useful to manually try out scenarios, not to prevent regressions as unit tests are. It's reasonable to have integration test that resembles customer use, i.e. initialization split into 2 files.

dreddy-work: Intention of this test was not serialization. hence, either is fine to me. @ArtemTatarinov  can you please followup on this and get this to closure. We will be branching soon and clear as many PRs as possible here.

Tanya-Solyanik: I strongly disagree about intention of this test. Integration tests verify how different features come together, thus removing one of these features is making test less useful.

dreddy-work: @ArtemTatarinov , do you mind updating tests here and help unblock this?

ArtemTatarinov: We discussed this topic with Vladimir Krestov and decided to leave it as it is and close this PR now. In the future maybe we could add new unit tests for regular automated checking of the resource (de)serialization. @RussKie, would you agree?

RussKie: > In the future maybe we could add new unit tests for regular automated checking of the resource (de)serialization.

👍 

> We discussed this topic with Vladimir Krestov and decided to leave it as it is and close this PR now.

👍

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/ListViewTest.cs
```diff
@@ -142,7 +142,7 @@ private void AddCollapsibleGroupToListView()
             listView1.Groups.Add(lvgroup1);
             listView1.Items.Add(new ListViewItem
             {
-                Text = "Item",
+                Text = "Item4",
```
Tanya-Solyanik, line 145: @ArtemTatarinov  - could you please create a new PR with the clean up changes for item names when you have a chance?

ArtemTatarinov: Created a new PR: https://github.com/dotnet/winforms/pull/6848

Issue #5487, opened 2021-08-20
A DateTimePicker control should support state announcing on Calendar opening


DmitryGorokhov:
* More info:
Not regression. This is an addition functionality for improving experience.
This opportunity appears after adding support Expand-Collapse pattern for DateTimePicker control.

* Linked items:
       - #4099 Issue
       - #5245 PR 

* .NET Core Version: 
6.0.100-rc.2.21419.27

* Have you experienced this same bug with .NET Framework?: 
Yes 

**Problem description:**

An DateTimePicker control should announce state on Calendar opening.

**Expected behavior:**

An DateTimePicker control announce state on Calendar opening action. You can take a look at ComboBox control as an example.

**Minimal repro:**

1. Create a winforms project with a DateTimePicker control.
2. Build and run the project.
3. Launch Narrator (and Narrator Buddy optional).
4. Expand calendar using arrow-button. Collapse calendar using arrow-button or select any value in calendar.

You can detect that date-time-picker don't provide state on open calendar and provide on close. So you task to fix it.

![image](https://user-images.githubusercontent.com/58004471/130207452-463aa09c-0c20-4d53-a040-4a98c3222b53.png)

vladimir-krestov:
@Olina-Zhang, could you please test this issue? It seems, that #6620 fixed this issue and we can close it.
Please use F4 button or mouse to expand the calendar.
As I understand, it's expected behavior:

![image](https://user-images.githubusercontent.com/49272759/159918145-b1f6209d-45ac-432e-9234-efe4e3bcb9ef.png)

Please compare with previous versions.

Olina-Zhang:
Verified this issue in the latest .Net 7.0 build, it is fixed: when expand datetimepicker calendar by F4, Alt+Down or mouse, Narrator can announce the expand state now.
![image](https://user-images.githubusercontent.com/26474449/160337284-c101a43e-fc31-40b2-a1f8-f718e1e1b707.png)

PR #6436, opened 2021-12-29
Fixing empty value of DefaultAction of DateTimePicker control

### LINKED ISSUE #6317: Accessibility: Inspect: The LegacyIAccessible.DefaultAction property should not be empty for Collapsed/Expanded DateTimePicker
* .NET Core Version: 
.Net 7.0 SDK build: 7.0.100-alpha.1.21568.2

Yes

**Minimal repro:**
1. Create a Winforms .Net Core application
2. Add a DateTimePicker control to form designer
3. Build and run application, then use Inspect to detect it when DateTimePicker is collapsed and expanded, pay attention to the LegacyIAccessible.DefaultAction property

**Actual behavior:**
LegacyIAccessible.DefaultAction property is empty when DateTimePicker is collapsed or expanded
Collapse:
image
Expand:
image
**Expected behavior:**
LegacyIAccessible.DefaultAction property should be same as ComboBox control when it is collapsed or expanded
ComboBox is collapsed:
image
ComboBox is expanded:
image

DmitryGorokhov:
Fixes #6317 

## Proposed changes

- Overridden `DefaultAction` property of DateTimePicker AO to return `Expand` value if the control is collapsed. Otherwise, return `Collapse` if the control is expanded.
- Overridden `DoDefaultAction` method to perform `Collapse()` or `Expand()` methods.
- Added test case for it.

## Customer Impact

- Receive the correct `DefaultAction` value of a dtp control and improve accessibility.

## Regression? 

- No

## Risk

- Minimal

### Before

Property returns empty value.

#### Collapsed

DTP_col_b

#### Expanded

DTP_exp_b

### After

Returns `Expand` or `Collapse` value.

#### Collapsed

DTP_col

#### Expanded

DTP_exp

#### Accessibility Insights

Also has no additional errors in AI.

AI

## Test methodology 

- Manually
- Unit tests
- CTI

## Accessibility testing  

- Inspect
- AI
- Narrator
 
## Test environment(s) 

- .Net version:   6.0.100
- OS version: 10.0.19044

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6436)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObject.cs
@@ -116,10 +116,34 @@ namespace System.Windows.Forms
                 };
 
             internal override bool IsPatternSupported(UiaCore.UIA patternId)
-                => (patternId == UiaCore.UIA.TogglePatternId && ((DateTimePicker)Owner).ShowCheckBox) ||
-                    patternId == UiaCore.UIA.ExpandCollapsePatternId ||
-                    patternId == UiaCore.UIA.ValuePatternId ||
-                    base.IsPatternSupported(patternId);
+                => patternId switch
+                {
+                    UiaCore.UIA.TogglePatternId when ((DateTimePicker)Owner).ShowCheckBox => true,
+                    UiaCore.UIA.ExpandCollapsePatternId => true,
+                    UiaCore.UIA.ValuePatternId => true,
+                    _ => base.IsPatternSupported(patternId)
+                };
+
+            public override string DefaultAction
+                => ExpandCollapseState switch
+                {
+                    UiaCore.ExpandCollapseState.Collapsed => SR.AccessibleActionExpand,
+                    UiaCore.ExpandCollapseState.Expanded => SR.AccessibleActionCollapse,
+                    _ => string.Empty
+                };
+
+            public override void DoDefaultAction()
+            {
+                switch (ExpandCollapseState)
+                {
+                    case UiaCore.ExpandCollapseState.Collapsed:
+                        Expand();
+                        break;
+                    case UiaCore.ExpandCollapseState.Expanded:
+                        Collapse();
+                        break;
+                }
+            }
 
             #region Toggle Pattern
 
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObjectTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObjectTests.cs
@@ -290,5 +290,55 @@ namespace System.Windows.Forms.Tests
             Assert.Equal(testAccessibleName, dateTimePicker.AccessibilityObject.Name);
             Assert.False(dateTimePicker.IsHandleCreated);
         }
+
+        public static IEnumerable<object[]> DateTimePickerAccessibleObject_DefaultAction_ReturnsExpected_TestData()
+        {
+            // Expanded dtp control has "Collapse" as a default action, else "Expand".
+            yield return new object[] { true, SR.AccessibleActionCollapse };
+            yield return new object[] { false, SR.AccessibleActionExpand };
+        }
+
+        [WinFormsTheory]
+        [MemberData(nameof(DateTimePickerAccessibleObject_DefaultAction_ReturnsExpected_TestData))]
+        public void DateTimePickerAccessibleObject_DefaultAction_ReturnsExpected(bool isExpanded, string expected)
+        {
+            using DateTimePicker dateTimePicker = new();
+            dateTimePicker.CreateControl();
+
+            AccessibleObject accessibleObject = dateTimePicker.AccessibilityObject;
+
+            if (isExpanded)
+            {
+                accessibleObject.Expand();
+            }
+
+            Assert.Equal(expected, accessibleObject.DefaultAction);
+            Assert.True(dateTimePicker.IsHandleCreated);
+        }
+
+        [WinFormsTheory]
+        [InlineData(true, (int)UiaCore.ExpandCollapseState.Collapsed)]
+        [InlineData(false, (int)UiaCore.ExpandCollapseState.Expanded)]
+        public void DateTimePickerAccessibleObject_DoDefaultAction_IfHandleIsCreated_ReturnsExpected(bool isExpanded, int expected)
+        {
+            using DateTimePicker dateTimePicker = new();
+            dateTimePicker.CreateControl();
+
+            AccessibleObject accessibleObject = dateTimePicker.AccessibilityObject;
+
+            Assert.Equal(UiaCore.ExpandCollapseState.Collapsed, accessibleObject.ExpandCollapseState);
+
+            if (isExpanded)
+            {
+                accessibleObject.Expand();
+
+                Assert.Equal(UiaCore.ExpandCollapseState.Expanded, accessibleObject.ExpandCollapseState);
+            }
+
+            accessibleObject.DoDefaultAction();
+
+            Assert.Equal((UiaCore.ExpandCollapseState)expected, accessibleObject.ExpandCollapseState);
+            Assert.True(dateTimePicker.IsHandleCreated);
+        }
     }
 }
--- a/src/System.Windows.Forms/src/PublicAPI.Unshipped.txt
+++ b/src/System.Windows.Forms/src/PublicAPI.Unshipped.txt
@@ -2,3 +2,5 @@
 ~override System.Windows.Forms.PrintPreviewControl.CreateAccessibilityInstance() -> System.Windows.Forms.AccessibleObject
 ~override System.Windows.Forms.DateTimePicker.DateTimePickerAccessibleObject.Name.get -> string
 ~override System.Windows.Forms.DateTimePicker.OnGotFocus(System.EventArgs e) -> void
+override System.Windows.Forms.DateTimePicker.DateTimePickerAccessibleObject.DoDefaultAction() -> void
+~override System.Windows.Forms.DateTimePicker.DateTimePickerAccessibleObject.DefaultAction.get -> string


vladimir-krestov (commented):
Looks good, but I think we should implement `DoDefaultAction` as well.

DmitryGorokhov:
I also added `Collapse()` or `Expand()` execution on calling `DefaultAction` method. Thanks for reminder

vladimir-krestov (dismissed):
Approved ✔️

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObjectTests.cs
```diff
@@ -240,5 +241,61 @@ public void DateTimePickerAccessibleObject_Collapse_IfHandleIsCreated_ReturnsExp
...
+        public void DateTimePickerAccessibleObject_DefaultAction_ReturnsExpected(bool isExpanded)
+        {
+            using DateTimePicker dateTimePicker = new();
+
+            dateTimePicker.CreateControl();
+            var accessibleObject = (DateTimePickerAccessibleObject)dateTimePicker.AccessibilityObject;
+
+            if (isExpanded)
+            {
+                accessibleObject.Expand();
```
dreddy-work, line 256: Q: isn't this trying to expand?  if it was already expanded, do we expect it to collapse inside `expand()` ?

DmitryGorokhov: No, if it was already expanded, we expect `Expand()` does nothing. We check this behavior in other test cases. 
In this case I am trying to test different variants of `DefaultAction` that depends on `ExpandCollapseState`. As a default the control has collapsed state. The most expected way to change state is by factual expand the control, so `Expand()` do this.

RussKie: I suggest to restructure the test to pass two input arguments, e.g. pseudocode: `InlineData(expanded: false, expected: SR.AccessibleActionExpand)`. This way the test has clear purpose and cyclomatic complexity of 1.

dreddy-work: at the least, you should describe,  as comment, on how default action description changes with these actions here.

src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObject.cs
```diff
@@ -116,6 +116,23 @@ internal override bool IsPatternSupported(UiaCore.UIA patternId)
...
                     base.IsPatternSupported(patternId);
 
+            public override string DefaultAction
+                => ExpandCollapseState == UiaCore.ExpandCollapseState.Collapsed
+                    ? SR.AccessibleActionExpand
+                    : SR.AccessibleActionCollapse;
+
+            public override void DoDefaultAction()
+            {
+                if (ExpandCollapseState == UiaCore.ExpandCollapseState.Collapsed)
```
Tanya-Solyanik, line 126: nit: could you please follow the same pattern as on line 120, i.e. ternary conditional operator?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObjectTests.cs
DmitryGorokhov: Ok, I got it. Thanks

src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObject.cs
DmitryGorokhov: I can't do this. I see the same message if I use curly braces instead of an arrow.
image

Tanya-Solyanik: I see, it returns a void. You are right!

Tanya-Solyanik (approved):

DmitryGorokhov:
I've sent it for testing.

DmitryGorokhov:
CTI has found an issue. For a collapsed dtp, when expanded it, then collapsed it, the `DefaultAction` is not correct. It has unexpected value `Press` instead of `Expand`. So pr is a draft while I'm resolving the issue.

dreddy-work:
> CTI has found an issue. For a collapsed dtp, when expanded it, then collapsed it, the `DefaultAction` is not correct. It has unexpected value `Press` instead of `Expand`. So pr is a draft while I'm resolving the issue.

What is the state of this PR now @DmitryGorokhov ?

src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObject.cs
```diff
@@ -116,6 +116,23 @@ internal override bool IsPatternSupported(UiaCore.UIA patternId)
                     patternId == UiaCore.UIA.ExpandCollapsePatternId ||
                     base.IsPatternSupported(patternId);
 
+            public override string DefaultAction
+                => ExpandCollapseState == UiaCore.ExpandCollapseState.Collapsed
```
Tanya-Solyanik, line 120: I see, there are more states than expanded and collapsed - https://docs.microsoft.com/en-us/dotnet/api/system.windows.automation.expandcollapsestate?view=windowsdesktop-6.0

DmitryGorokhov: Yes, it is. At least dtp can't be a `LeafNode` in all way, but I don't know what about `PartiallyExpanded`. But as I can see in the control class `ExpandCollapseState` can return only 2 state of this.

DmitryGorokhov:
> > CTI has found an issue. For a collapsed dtp, when expanded it, then collapsed it, the `DefaultAction` is not correct. It has unexpected value `Press` instead of `Expand`. So pr is a draft while I'm resolving the issue.
> 
> What is the state of this PR now @DmitryGorokhov ?

WIP still. A `Draft` label has been removed by Tanya's comment. I am continue watching found issue.

src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObject.cs
Tanya-Solyanik: I would suggest looking up what the corresponding windows control is doing.

src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObject.cs
```diff
@@ -114,8 +114,26 @@ internal override object GetPropertyValue(UiaCore.UIA propertyID)
             internal override bool IsPatternSupported(UiaCore.UIA patternId)
                 => (patternId == UiaCore.UIA.TogglePatternId && ((DateTimePicker)Owner).ShowCheckBox) ||
                     patternId == UiaCore.UIA.ExpandCollapsePatternId ||
+                    patternId == UiaCore.UIA.LegacyIAccessiblePatternId ||
                     base.IsPatternSupported(patternId);
```
WPMGPRoSToTeMa, line 118: Could you please refactor this using switch expression? @Tanya-Solyanik what do you think about readability of switch expression in this case?
```suggestion
                => patternId switch
                {
                    UiaCore.UIA.TogglePatternId when ((DateTimePicker)Owner).ShowCheckBox: true,
                    UiaCore.UIA.ExpandCollapsePatternId: true,
                    UiaCore.UIA.LegacyIAccessiblePatternId: true,
                    _ => base.IsPatternSupported(patternId)
                };
```

src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObject.cs
WPMGPRoSToTeMa: I think we can use pattern matching here and return an empty string (or throw an exception) in case state is neither collapsed nor expanded. But in this case it would be better to also change `DoDefaultAction` to use switch statement probably.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObjectTests.cs
```diff
@@ -240,5 +240,66 @@ public void DateTimePickerAccessibleObject_Collapse_IfHandleIsCreated_ReturnsExp
             Assert.Equal(UiaCore.ExpandCollapseState.Collapsed, accessibleObject.ExpandCollapseState);
             Assert.True(dateTimePicker.IsHandleCreated);
         }
+
+        public static IEnumerable<object[]> DateTimePickerAccessibleObject_DefaultAction_ReturnsExpected_TestData()
+        {
+            // Expanded dtp control has "Collapse" as a default action, else "Expand".
+            yield return new object[] { true, SR.AccessibleActionCollapse };
+            yield return new object[] { false, SR.AccessibleActionExpand };
```
WPMGPRoSToTeMa, line 248: Inline data is not applicable here because localized strings aren't constants, right?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObjectTests.cs
```diff
@@ -240,5 +240,66 @@ public void DateTimePickerAccessibleObject_Collapse_IfHandleIsCreated_ReturnsExp
...
+        }
+
+        [WinFormsTheory]
+        [MemberData(nameof(DateTimePickerAccessibleObject_DefaultAction_ReturnsExpected_TestData))]
+        public void DateTimePickerAccessibleObject_DefaultAction_ReturnsExpected(bool isExpanded, string expected)
```
WPMGPRoSToTeMa, line 253: nit: consider:
```suggestion
        public void DateTimePickerAccessibleObject_DefaultAction_ReturnsExpected(bool expand, string expected)
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObjectTests.cs
```diff
@@ -240,5 +240,66 @@ public void DateTimePickerAccessibleObject_Collapse_IfHandleIsCreated_ReturnsExp
...
+        {
+            using DateTimePicker dateTimePicker = new();
+
+            dateTimePicker.CreateControl();
+            var accessibleObject = (DateTimePickerAccessibleObject)dateTimePicker.AccessibilityObject;
```
WPMGPRoSToTeMa, line 258: nit:
```suggestion
            using DateTimePicker dateTimePicker = new();
            dateTimePicker.CreateControl();
            
            var accessibleObject = (DateTimePickerAccessibleObject)dateTimePicker.AccessibilityObject;
```

Also, do we need `accessibleObject` variable and do we really need this cast (e.g. what are the advantages of having it)?

DmitryGorokhov:
I've sent it to retesting.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObjectTests.cs
```diff
@@ -240,5 +240,66 @@ public void DateTimePickerAccessibleObject_Collapse_IfHandleIsCreated_ReturnsExp
...
+
+            accessibleObject.Expand();
+
+            Assert.Equal(UiaCore.ExpandCollapseState.Expanded, accessibleObject.ExpandCollapseState);
+
+            accessibleObject.DoDefaultAction();
+
+            Assert.Equal(UiaCore.ExpandCollapseState.Collapsed, accessibleObject.ExpandCollapseState);
+            Assert.True(dateTimePicker.IsHandleCreated);
+        }
```
WPMGPRoSToTeMa, line 303: I think you can merge these tests into a single theory.

WPMGPRoSToTeMa (approved):
LGTM, some comments with nits and refactorings.

src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObject.cs
Tanya-Solyanik: I prefer the switch expression in terms of readability, the general guidance is to modernize the code base as we go along. 
In this particular case, `IsPatternSupported` is implemented as a `switch` statement/expression in other classes, so additionally it makes the code more consistent.

Olina-Zhang:
Tested this PR with no new issue found, hitting known GH issue: [5487](https://github.com/dotnet/winforms/issues/5487).

src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObject.cs
DmitryGorokhov: It looks better, thanks. Reworked.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObjectTests.cs
DmitryGorokhov: Right. I have to use `MemberData` here.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObjectTests.cs
DmitryGorokhov: We can remove the cast, because the base class has all methods to call. So this unnecessary cast has been deleted.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObjectTests.cs
DmitryGorokhov: Yes, looks reasonable. Reworked. Also the unnecessary cast has been deleted.

src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObject.cs
DmitryGorokhov: I also reworked `DefaultAction` property using the match pattern. But seems I can't do this for `DoDefaultAction` method due to default case where I can't call some function or return something because the method has void return-type. So situation the same as when use a ternary conditional.

src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObject.cs
DmitryGorokhov: Good idea! Thank you. Reworked in the comment below.

src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObject.cs
DmitryGorokhov: Don't mind, reworked `DoDefaultAction` using switch.

src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObject.cs
```diff
@@ -112,9 +112,31 @@ internal override object GetPropertyValue(UiaCore.UIA propertyID)
...
+                    _ => string.Empty
+                };
+
+            public override void DoDefaultAction()
+            {
+                switch (ExpandCollapseState)
+                {
+                    case UiaCore.ExpandCollapseState.Collapsed: Expand(); break;
+                    case UiaCore.ExpandCollapseState.Expanded: Collapse(); break;
+                    default: break;
```
WPMGPRoSToTeMa, line 137: Default case can be removed.
```suggestion
```

src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObject.cs
```diff
@@ -112,9 +112,31 @@ internal override object GetPropertyValue(UiaCore.UIA propertyID)
...
+                    UiaCore.ExpandCollapseState.Expanded => SR.AccessibleActionCollapse,
+                    _ => string.Empty
+                };
+
+            public override void DoDefaultAction()
+            {
+                switch (ExpandCollapseState)
+                {
+                    case UiaCore.ExpandCollapseState.Collapsed: Expand(); break;
+                    case UiaCore.ExpandCollapseState.Expanded: Collapse(); break;
```
WPMGPRoSToTeMa, line 136: Thanks!

src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObject.cs
DmitryGorokhov: Fixed, thanks

DmitryGorokhov:
@dreddy-work, seems work on it is finished, because all comments resolved and testers approved this changes.

vladimir-krestov (dismissed):
LGTM ✔️

Tanya-Solyanik:
@DmitryGorokhov - please rebase this change, otherwise it's ready to go!

DmitryGorokhov:
I should check it one more time. Seems, `IsPatternSupported` contains duplicate `LegacyIAccessiblePatternId` I due to dtp has already support uia provider and this pattern.

DmitryGorokhov:
I've removed the duplicate pattern definition. So, this is ready for merge.
@Tanya-Solyanik PTAL

Tanya-Solyanik (approved):
Thank you, looks good!

RussKie (approved):

PR #6445, opened 2021-12-31
Enable nullability in ListBindingConverter

gpetrou:
## Proposed changes

- Enable nullability in ListBindingConverter.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6445)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListBindingConverter.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListBindingConverter.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.Collections;
 using System.ComponentModel;
 using System.ComponentModel.Design.Serialization;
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ListBindingConverter.cs
```diff
@@ -134,29 +132,25 @@ private InstanceDescriptor GetInstanceDescriptorFromValues(Binding b)
...
-                //
                 if (ConstructorParameterProperties[lastItem] is null)
                 {
                     break;
                 }
 
                 // get the property and see if it needs to be serialized.
-                //
-                PropertyDescriptor prop = TypeDescriptor.GetProperties(b)[ConstructorParameterProperties[lastItem]];
+                PropertyDescriptor? prop = TypeDescriptor.GetProperties(b)[ConstructorParameterProperties[lastItem]!];
```
dreddy-work, line 141: This can not be null. It may return empty collection.

RussKie, line 141: Instead of a `!` please refactor to something like:
```cs
                var constructorParameterProperty = ConstructorParameterProperties[lastItem];
                if (constructorParameterProperty is null)
                {
                    break;
                }

                PropertyDescriptor? prop = TypeDescriptor.GetProperties(b)[constructorParameterProperty];
```

src/System.Windows.Forms/src/System/Windows/Forms/ListBindingConverter.cs
```diff
@@ -188,7 +181,7 @@ private InstanceDescriptor GetInstanceDescriptorFromValues(Binding b)
                         val = b.BindingMemberInfo.BindingMember;
                         break;
                     default:
-                        val = TypeDescriptor.GetProperties(b)[ConstructorParameterProperties[i]].GetValue(b);
+                        val = TypeDescriptor.GetProperties(b)[ConstructorParameterProperties[i]!]?.GetValue(b);
```
RussKie, line 184: How can we be sure `ConstructorParameterProperties[i]!` is _not_ null?

RussKie (approved):

PR #6448, opened 2021-12-31
Enable nullability in KeysConverter

gpetrou:
## Proposed changes

- Enable nullability in KeysConverter.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6448)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
@@ -17,7 +17,7 @@ namespace System.Windows.Forms
     /// </summary>
     public class KeysConverter : TypeConverter, IComparer
     {
-        private IDictionary? _keyNames;
+        private IDictionary<string, Keys>? _keyNames;
         private List<string>? _displayOrder;
         private StandardValuesCollection? _values;
 
@@ -31,7 +31,7 @@ namespace System.Windows.Forms
         [MemberNotNull(nameof(_displayOrder))]
         private void Initialize()
         {
-            _keyNames = new Hashtable(34);
+            _keyNames = new Dictionary<string, Keys>(34);
             _displayOrder = new List<string>(34);
 
             AddKey(SR.toStringEnter, Keys.Return);
@@ -77,7 +77,7 @@ namespace System.Windows.Forms
         ///  Access to a lookup table of name/value pairs for keys.  These are localized
         ///  names.
         /// </summary>
-        private IDictionary KeyNames
+        private IDictionary<string, Keys> KeyNames
         {
             get
             {
@@ -252,12 +252,11 @@ namespace System.Windows.Forms
 
                     // First, iterate through and do the modifiers. These are
                     // additive, so we support things like Ctrl + Alt
-                    //
                     for (int i = 0; i < DisplayOrder.Count; i++)
                     {
                         string keyString = DisplayOrder[i];
-                        Keys keyValue = (Keys)_keyNames![keyString]!;
-                        if (((int)(keyValue) & (int)modifiers) != 0)
+                        Keys keyValue = _keyNames![keyString];
+                        if (((int)keyValue & (int)modifiers) != 0)
                         {
                             if (asString)
                             {
@@ -290,7 +289,7 @@ namespace System.Windows.Forms
                     for (int i = 0; i < DisplayOrder.Count; i++)
                     {
                         string keyString = DisplayOrder[i];
-                        Keys keyValue = (Keys)_keyNames![keyString]!;
+                        Keys keyValue = _keyNames![keyString];
                         if (keyValue.Equals(keyOnly))
                         {
                             if (asString)
@@ -355,7 +354,7 @@ namespace System.Windows.Forms
             {
                 ArrayList list = new ArrayList();
 
-                ICollection keys = KeyNames.Values;
+                ICollection<Keys> keys = KeyNames.Values;
 
                 foreach (object o in keys)
                 {
--- a/src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
@@ -2,11 +2,10 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.Collections;
 using System.ComponentModel;
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.Text;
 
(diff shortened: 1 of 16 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
```diff
@@ -156,27 +157,24 @@ public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo c
...
                     }
 
                     // Now lookup each key token in our key hashtable.
-                    //
                     Keys key = (Keys)0;
                     bool foundKeyCode = false;
 
                     for (int i = 0; i < tokens.Length; i++)
                     {
-                        object obj = KeyNames[tokens[i]];
```
dreddy-work, line 173: this is nullable.

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
```diff
@@ -156,27 +157,24 @@ public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo c
...
+                        object obj = KeyNames[tokens[i]]!;
```
RussKie, line 172: I don't think it is - `private IDictionary KeyNames`. Instead of using `!` I think we should change the signature of `KeyNames` to `private IDictionary<string, Keys> KeyNames`.

dreddy-work: I do not have practical example here but following code was written with intention that `obj` could be null.

dreddy-work:
@gpetrou , can you rebase here?

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
RussKie: I followed the code and I can't see how we can have null in the dictionary. Maybe I'm missing some flow...

RussKie (commented):
We should avoid bang operators, and in this case I do not see them as appropriate.
I believe it should be annotated like this:
```diff
diff --git a/src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs b/src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
index 024ab7ee1..dd124b10c 100644
--- a/src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
@@ -21,12 +21,6 @@ namespace System.Windows.Forms
         private List<string>? _displayOrder;
         private StandardValuesCollection? _values;
 
-        private void AddKey(string key, Keys value)
-        {
-            _keyNames![key] = value;
-            _displayOrder!.Add(key);
-        }
-
         [MemberNotNull(nameof(_keyNames))]
         [MemberNotNull(nameof(_displayOrder))]
         private void Initialize()
@@ -71,12 +65,20 @@ namespace System.Windows.Forms
             AddKey("7", Keys.D7);
             AddKey("8", Keys.D8);
             AddKey("9", Keys.D9);
+
+            void AddKey(string key, Keys value)
+            {
+                _keyNames[key] = value;
+                _displayOrder!.Add(key);
+            }
         }
 
         /// <summary>
         ///  Access to a lookup table of name/value pairs for keys.  These are localized
         ///  names.
         /// </summary>
+        [MemberNotNull(nameof(_keyNames))]
+        [MemberNotNull(nameof(_displayOrder))]
         private IDictionary<string, Keys> KeyNames
         {
             get
@@ -87,10 +89,14 @@ namespace System.Windows.Forms
                     Initialize();
                 }
 
+#pragma warning disable CS8774 // Member must have a non-null value when exiting.
                 return _keyNames;
+#pragma warning restore CS8774 // Member must have a non-null value when exiting.
             }
         }
 
+        [MemberNotNull(nameof(_keyNames))]
+        [MemberNotNull(nameof(_displayOrder))]
         private List<string> DisplayOrder
         {
             get
@@ -101,7 +107,9 @@ namespace System.Windows.Forms
                     Initialize();
                 }
 
+#pragma warning disable CS8774 // Member must have a non-null value when exiting.
                 return _displayOrder;
+#pragma warning restore CS8774 // Member must have a non-null value when exiting.
             }
         }
 
@@ -255,7 +263,7 @@ namespace System.Windows.Forms
                     for (int i = 0; i < DisplayOrder.Count; i++)
                     {
                         string keyString = DisplayOrder[i];
-                        Keys keyValue = _keyNames![keyString];
+                        Keys keyValue = _keyNames[keyString];
                         if (((int)keyValue & (int)modifiers) != 0)
                         {
                             if (asString)
@@ -289,7 +297,7 @@ namespace System.Windows.Forms
                     for (int i = 0; i < DisplayOrder.Count; i++)
                     {
                         string keyString = DisplayOrder[i];
-                        Keys keyValue = _keyNames![keyString];
+                        Keys keyValue = _keyNames[keyString];
                         if (keyValue.Equals(keyOnly))
                         {
                             if (asString)
```

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
```diff
@@ -70,45 +65,59 @@ private void Initialize()
             AddKey("7", Keys.D7);
             AddKey("8", Keys.D8);
             AddKey("9", Keys.D9);
+
+            void AddKey(string key, Keys value)
+            {
+                _keyNames[key] = value;
+                _displayOrder!.Add(key);
```
RussKie, line 72:
```suggestion
                _displayOrder.Add(key);
```

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
```diff
@@ -70,45 +65,59 @@ private void Initialize()
...
-                if (keyNames is null)
+                if (_keyNames is null)
                 {
-                    Debug.Assert(displayOrder is null);
+                    Debug.Assert(_displayOrder is null);
                     Initialize();
                 }
 
-                return keyNames;
+#pragma warning disable CS8774 // Member must have a non-null value when exiting.
```
RussKie, line 92: Apologies, it appears I posted an out-dated suggestion in the last review
```suggestion
#pragma warning disable CS8774 // Member must have a non-null value when exiting: Initialize() inits both _keyNames and _displayOrder.
```

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
```diff
@@ -70,45 +65,59 @@ private void Initialize()
...
-                if (displayOrder is null)
+                if (_displayOrder is null)
                 {
-                    Debug.Assert(keyNames is null);
+                    Debug.Assert(_keyNames is null);
                     Initialize();
                 }
 
-                return displayOrder;
+#pragma warning disable CS8774 // Member must have a non-null value when exiting.
```
RussKie, line 110: ditto here

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
```diff
@@ -156,68 +165,47 @@ public override object ConvertFrom(ITypeDescriptorContext context, CultureInfo c
...
                     }
 
                     // Now lookup each key token in our key hashtable.
-                    //
                     Keys key = (Keys)0;
                     bool foundKeyCode = false;
 
                     for (int i = 0; i < tokens.Length; i++)
                     {
                         object obj = KeyNames[tokens[i]];
```
RussKie, line 180: There's no need to box the value
```suggestion
                        Keys currentKey = KeyNames[tokens[i]];
```

RussKie (commented):
👍

RussKie (approved):

dreddy-work (approved):

PR #6472, opened 2022-01-06
Enable nullability in Button and IButtonControl

gpetrou:
## Proposed changes

- Enable nullability in Button and IButtonControl.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6472)

### DIFF
--- a/src/System.Windows.Forms/src/PublicAPI.Shipped.txt
+++ b/src/System.Windows.Forms/src/PublicAPI.Shipped.txt
@@ -6090,8 +6090,8 @@ System.Windows.Forms.Button
 System.Windows.Forms.Button.AutoSizeMode.get -> System.Windows.Forms.AutoSizeMode
 System.Windows.Forms.Button.AutoSizeMode.set -> void
 System.Windows.Forms.Button.Button() -> void
-System.Windows.Forms.Button.DoubleClick -> System.EventHandler!
-System.Windows.Forms.Button.MouseDoubleClick -> System.Windows.Forms.MouseEventHandler!
+System.Windows.Forms.Button.DoubleClick -> System.EventHandler?
+System.Windows.Forms.Button.MouseDoubleClick -> System.Windows.Forms.MouseEventHandler?
 System.Windows.Forms.Button.PerformClick() -> void
 System.Windows.Forms.ButtonBase
 System.Windows.Forms.ButtonBase.AutoEllipsis.get -> bool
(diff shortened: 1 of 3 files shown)


src/System.Windows.Forms/src/PublicAPI.Shipped.txt
```diff
@@ -6307,8 +6307,8 @@ System.Windows.Forms.Button
 System.Windows.Forms.Button.AutoSizeMode.get -> System.Windows.Forms.AutoSizeMode
 System.Windows.Forms.Button.AutoSizeMode.set -> void
 System.Windows.Forms.Button.Button() -> void
-System.Windows.Forms.Button.DoubleClick -> System.EventHandler
-System.Windows.Forms.Button.MouseDoubleClick -> System.Windows.Forms.MouseEventHandler
+System.Windows.Forms.Button.DoubleClick -> System.EventHandler!
+System.Windows.Forms.Button.MouseDoubleClick -> System.Windows.Forms.MouseEventHandler!
```
gpetrou, line 6311: Perhaps these should be nullable? I see in ScrollBar we have nullable EventHandler, so not sure which one is correct.

Tanya-Solyanik: I would change `ScrollBar` event handler nullability. I don't see when it would be null...

Tanya-Solyanik (dismissed):

src/System.Windows.Forms/src/PublicAPI.Shipped.txt
RussKie: @gpetrou could you please elaborate?

gpetrou: In ScrollBar that already has nullability enabled we have:
```
[Browsable(false)]
public new event MouseEventHandler? MouseDoubleClick
{
    add => base.MouseDoubleClick += value;
    remove => base.MouseDoubleClick -= value;
}
```
Here I used:
```
[Browsable(false)]
public new event MouseEventHandler MouseDoubleClick
{
    add => base.MouseDoubleClick += value;
    remove => base.MouseDoubleClick -= value;
}
```
Looking at the references in the codebase we only seem to have a couple for each in tests such as:
```
 control.MouseDoubleClick += handler;
```
So, I am not sure whether the MouseEventHandler and other similar handlers should be nullable or not. Any thoughts?

RussKie: I think event handlers should be annotated as nullable - unless wired to these are null.

RussKie (approved):

PR #6473, opened 2022-01-06
Enable nullability in ContainerInfo

gpetrou:
## Proposed changes

- Enable nullability in ContainerInfo.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6473)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Layout/TableLayout.ContainerInfo.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Layout/TableLayout.ContainerInfo.cs
@@ -4,6 +4,7 @@
 
 using System.Collections.Specialized;
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using System.Drawing;
 
 namespace System.Windows.Forms.Layout
@@ -309,7 +310,7 @@ namespace System.Windows.Forms.Layout
                     if (HasChildWithAbsolutePositioning)
                     {
                         int index = 0;
-                        for (int i = 0; i < _childInfo!.Length; i++)
+                        for (int i = 0; i < _childInfo.Length; i++)
                         {
                             if (_childInfo[i].IsAbsolutelyPositioned)
                             {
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Layout/TableLayout.ContainerInfo.cs
```diff
@@ -308,7 +306,7 @@ public LayoutInfo[] FixedChildrenInfo
                     Debug.Assert(ChildInfoValid, "Fetched invalid information");
                     // we only get this in a cached scenario - so we don't have to worry about caching it.
                     LayoutInfo[] fixedChildren = new LayoutInfo[_countFixedChildren];
-                    if (HasChildWithAbsolutePositioning)
+                    if (HasChildWithAbsolutePositioning && _childInfo is not null)
```
Tanya-Solyanik, line 309: I'm not sure that it's valid to have `_childInfo == null` when we know that we have ` _countFixedChildren > 0`.  I would prefer an exception in this case.

dreddy-work (dismissed):

src/System.Windows.Forms/src/System/Windows/Forms/Layout/TableLayout.ContainerInfo.cs
```diff
@@ -311,7 +309,7 @@ public LayoutInfo[] FixedChildrenInfo
                     if (HasChildWithAbsolutePositioning)
                     {
                         int index = 0;
-                        for (int i = 0; i < _childInfo.Length; i++)
+                        for (int i = 0; i < _childInfo!.Length; i++)
```
RussKie, line 312: Please don't do this.

I believe this should address this:
```cs
            [MemberNotNullWhen(true, nameof(_childInfo))]
            public bool HasChildWithAbsolutePositioning
            {
                get { return _countFixedChildren > 0; }
            }
```

RussKie (changes requested):

RussKie (dismissed):

gpetrou:
@RussKie perhaps you need to manually merge this since the PrintPreviewDialogTests cause too many failures?

PR #6476, opened 2022-01-06
Enable nullability in classes that inherit from ToolStripDropDownItemAccessibleObject

gpetrou:
## Proposed changes

- Enable nullability in classes that inherit from ToolStripDropDownItemAccessibleObject.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6476)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripOverflowButton.ToolStripOverflowButtonAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripOverflowButton.ToolStripOverflowButtonAccessibleObject.cs
@@ -2,6 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Diagnostics.CodeAnalysis;
+
 namespace System.Windows.Forms
 {
     public partial class ToolStripOverflowButton
@@ -12,7 +14,8 @@ namespace System.Windows.Forms
             {
             }
 
-            public override string? Name
+            [AllowNull]
+            public override string Name
             {
                 get => Owner.AccessibleName ?? SR.ToolStripOptions;
                 set => base.Name = value;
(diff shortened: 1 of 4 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ToolStripOverflowButton.ToolStripOverflowButtonAccessibleObject.cs
```diff
@@ -14,7 +12,7 @@ public ToolStripOverflowButtonAccessibleObject(ToolStripOverflowButton owner) :
             {
             }
 
-            public override string Name
+            public override string? Name
```
dreddy-work, line 15: why would this be null?

dreddy-work:
@gpetrou , thank you for making time to make repo better.  can you please also look at your PRs that were marked with `waiting-author-feedback` label and make them complete?

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripOverflowButton.ToolStripOverflowButtonAccessibleObject.cs
Tanya-Solyanik: `null` is an allowed value. This is the way to clear the accessible name. https://docs.microsoft.com/dotnet/api/system.windows.forms.accessibleobject.name?view=netframework-4.8

I don't see a reason to make it non-nullable on the base class (on `ToolStripItemAccessibleObject`). However in this class we have several alternatives because we know that the getter does not return null and the setter is not doing anything new. 
```
            public override string Name
            {
                get => Owner.AccessibleName ?? SR.ToolStripOptions;
            }
```
 or apply a pre-condition that allows setting this property to null to match the 
 base class type (https://docs.microsoft.com/dotnet/csharp/language-reference/attributes/nullable-analysis#preconditions-allownull-and-disallownull):

```
            [Diagnostics.CodeAnalysis.AllowNull]
            public override string Name
            {
                get => Owner.AccessibleName ?? SR.ToolStripOptions;
                set => base.Name = value;
            }
```

Issue #6421, opened 2021-12-26
Setting ShowInTaskbar to false causes the form to close automatically (Abnormal exit).


roland5572:
Setting ShowInTaskbar to false causes the form to close automatically (**Abnormal exit**).

https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.form

Test Code:
.NET 6.0 (Version: 6.0.1)(This problem also appears in .net framework 4.8)
Win10 system (21H2).

```
using System;
using System.Windows.Forms;

namespace FormShowInTaskbarTest
{
    internal static class Program
    {
        private static Form form1;
        private static Form form2;

        [STAThread]
        static void Main()
        {
            ApplicationConfiguration.Initialize();

            form1 = new Form()
            {
                Width = 800,
                Height = 600,
                StartPosition = FormStartPosition.CenterScreen,
            };

            var button1 = new Button()
            {
                Text = "Click here to open new Form.",

                Width = 200,
                Height = 100,
            };

            button1.Click += Button1_Click;
            form1.Controls.Add(button1);

            Application.Run(form1);
        }

        private static void Button1_Click(object sender, EventArgs e)
        {
            form2 = new Form()
            {
                Width = 600,
                Height = 400,

                StartPosition = FormStartPosition.CenterScreen,
            };

            var button2 = new Button()
            {
                Text = "Click here to test ShowInTaskbar.",

                Width = 200,
                Height = 100,
            };

            button2.Click += Button2_Click;
            form2.Controls.Add(button2);

            form2.ShowDialog(form1);
        }

        private static void Button2_Click(object sender, EventArgs e)
        {
            form2.ShowInTaskbar = false;
        }
    }
}
```

Steps to reproduce:
1. Run the appliction.
2. Click the first button, the application will open a new Form.
3. Click the first button in new form.

AraHaan:
This is odd because in my program it works just fine (even when using 7.0.0-dev runtime) 🤔.

roland5572:
> This is odd because in my program it works just fine (even when using 7.0.0-dev runtime) 🤔 (And I change it in a Timer when the application's own settings change from displaying the icon in taskbar to tray or to both).

Did you use Win7 or Win11? I am running on Win10 system (21H2).

AraHaan:
I use Windows 11 on my old computer (that had 11 since it was in preview), and my newer one and it seems to work fine.

roland5572:
[...]

PR #2262, opened 2019-11-01
Check for the return value of native SetParent calls in Control class

chenxinyanc:
Fixes #2163

## Proposed changes
* Check for the return value and GetLastError in `RecreateHandleCore` and `SetParentHandle` methods.
* Did not add the UT for this yet, because when `Control.CreateHandle` calls `Application.ParkHandle`, it does not specify the DPI awareness (`DPI_AWARENESS_CONTEXT_UNSPECIFIED`), so `ParkHandle` will choose a parking window for `UNSPECIFIED` awareness context. When constructing the parking window for `UNSPECIFIED` context, the DPI awareness of current thread will be used (e.g. PerMonitorV2). This will causes the subsequently **parked** window uses the very same DPI awareness setting, regardless of their owning thread's current DPI awareness at the point of control instance / handle construction.
    * I'm not sure whether it's by design or not, but perhaps we need to specify the DPI awareness explicitly when calling `ParkHandle` in `CreateHandle` method.
    * Nevertheless, I've tested the behavior with https://github.com/chenxinyanc/WinFormSetParentTest and it throws `Win32Exception` accordingly.

## Customer Impact

- 
- 

## Regression? 

- Yes / No

## Risk

-

## Test methodology 

- 
- 
- 

## Test environment(s) 

- 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/winforms/pull/2262)

PR #6477, opened 2022-01-07
Fixing incorrect state set for State.Created flag on the Control

### LINKED ISSUE #6464: Catastrophic regression introduced in .NET 6.0.1 WinForms, completely broke Paint.NET v4.3.6
* .NET Core Version: 6.0.1

**Problem description:**
Yesterday I rolled out a new update for Paint.NET, version 4.3.5. After quickly pushing out a hotfix (v4.3.6) for something that _I_ broke in the code, I then received numerous reports of another catastrophic breaking issue that rendered the app completely broken. 

After many hours it was finally determined that the culprit was having a certain app setting disabled (the default is enabled): "Translucent windows".
image

I'm pretty sure this was caused by a bug (or bug "fix" ?) introduced in .NET 6.0.1, as it doesn't happen in the previous version of my app (v4.3.4) on .NET 6.0, I had very few changes in the new version, and the workaround I found suggests a change in behavior surrounding `Handle`, `IsHandleCreated`, `OnHandleCreated()`, `OnHandleDestroyed()`. I've since published a new update with that workaround fix applied and everything works fine.

There's a big discussion on my forum that details the issue, the pleas for help and rollbacks, the troubleshooting, the discovery of a workaround in the app, and subsequently a test build that saved the day: https://forums.getpaint.net/topic/119185-many-problems-with-436/

Here are some highlights:

- Dialogs always opening at the top-left of the screen, instead of centered or literally anywhere they're supposed to be

- Truncated window size. This dialog should be 3-4x taller 
image

- Progress dialogs that never complete and can't be closed, resulting in a soft hang that requires Task Manager -> End Process
image

- Completely scrambled dialog layout for things like File->New, Image->Resize, etc.
image 

**Here is the change in 6.0.1 that is _highly_ suspicious:** https://github.com/dotnet/winforms/pull/6114 cc @RussKie 

In my code, I'm setting the window's opacity myself via `SetWindowLongPtrW()` and `SetLayeredWindowAttributes()`. This is because, way back with .NET 1.1 in 2004, there was some bug with `Form.Opacity` that I don't remember the details of. (I may be able to remove this code now, but like ... it ain't broke so don't fix it? maybe?) When "Translucent windows" is _disabled_, this ends up causing the window handle to be created in my `Form`s constructor (in a common base class that all other forms derive from), after `SuspendLayout()` and `InitializeComponent()`, and right before `ResumeLayout(false)`. I do consider this to be a bug in my code, but it should only be a performance bug, and it's been working fine for literally 18 years (Paint.NET 1.0 was originally released in May 2004 on .NET 1.1).

My fix, which is in the frantically released 4.3.7 update, is to first check `this.IsHandleCreated` before grabbing `this.Handle` to manually set the form's opacity (via `SetWindowLongPtrW()` and `SetLayeredWindowAttributes()`).

The behavior of the bug(s) seems to indicate some kind of disconnect between what WinForms thinks the handle is, and what handle Win32 is using, or maybe a disconnect between various parts of WinForms itself. Dialogs aren't closing properly, layout is all broken, every new dialog is at the top-left corner of the screen instead of centered, etc. The app will soft hang because modal progress dialogs don't close when they're supposed to. It's just a total disaster.

**Expected behavior:**
Everything works great. In reality, not even close :(

**Minimal repro:**
Here's a ZIP of the "cursed" Paint.NET v4.3.6 build, in portable (self-contained deployment) form: https://www.dotpdn.com/files/zip/test/paint.net.4.3.6.portable.x64.zip

Launch the app by double-clicking `paintdotnet.exe`. You should be able to launch things like File->New, Image->Resize, Layers->Rotate/Zoom, and run effects and adjustments and everything works fine.

Next, go into Settings (gear icon at top right of main window), then uncheck the box "Translucent windows". This is supposed to only affect the floating tool forms, since they're the only forms that use `Opacity`. Anyway, close the window and then try all the things listed above (File->New, etc.). _Everything will be completely, catastrophically broken._

I'm able to provide access to my solution/code for MSFT employees thanks to an NDA/license that @richlander and I set up a few years back. I can then easily show you which lines of code to change to make things broken/fixed.

dreddy-work:
This PR is the result of regressions caused by #6114 and #2262.

`State.Created` must be set only when `Handle` and the `Control `associated with it is created. Otherwise, recreating `Control `scenarios would be out of sync.  Changes in this PR is basically bringing the source code back to 6.0 GA state with respect to setting flag on `State.Created`.

Setting `State.Created` when the `Handle `is created but the associated `Control `is not, have repercussions mentioned in #6464. We cache this state and check before creating the control. Incorrect state would skip this. ex: https://github.com/dotnet/winforms/blob/0d4484331f57113d3303275c6aa5717fcd4b6c25/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs#L5044

fixes #6464 
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6477)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
@@ -9900,9 +9900,6 @@ namespace System.Windows.Forms
                     // the call shouldn't fail.
                     // However, it could fail if this.CreateParams.Parent is changed outside our control.
                     CreateHandle();
-
-                    // DestroyHandle resets the state, but CreateHandle doesn't set the state back.
-                    SetState(States.Created, true);
                 }
                 catch (Exception)
                 {
@@ -9932,7 +9929,7 @@ namespace System.Windows.Forms
                     // - or -
                     // CreateHandle is successful.
                     //      We will move the child controls to the new parent.
-                    if (controlSnapshot is not null && GetState(States.Created))
+                    if (controlSnapshot is not null && IsHandleCreated)
                     {
                         for (int i = 0; i < controlSnapshot.Length; i++)
                         {


Tanya-Solyanik (approved):

RussKie (commented):
Thank you.

We need to add a regression test(s). I'm not sure if a unit test can be devised, but we can certainly add UI integration tests.

rickbrew:
Alright I'm trying to validate this with a build of Paint.NET but for the life of me I can't seem to find the binaries to download so I can craft up a build ... can someone point me to them?

dreddy-work:
> Alright I'm trying to validate this with a build of Paint.NET but for the life of me I can't seem to find the binaries to download so I can craft up a build ... can someone point me to them?

This PR is targeting 7.0 so I created a pull request [here](https://github.com/rickbrew/WinForms6464_Data/pull/1)  with binaries that you need to replace for .net6.0.1. You need to build framework dependent pant.Net app and then it shoudl pick up these replaced binaries.

rickbrew:
Thanks, I'll be able to test this around lunch time PST.

rickbrew:
Okay I just tested this out and it does seem to work for fixing the issue described in #6464 ...

However, there is also a huge performance regression. When I go to launch the Image -> Resize dialog, it takes noticeably longer to launch. Not just the first time, which could easily because of R2R vs. JIT, but always.

In addition, resizing the main window is very slow now, whereas normally it's very fluid. I'll try and get a video attached here in a minute.

rickbrew:
Okay here are the videos. First one is Paint.NET v4.3.4, framework-dependent build, running on .NET 6.0.1 stock. Second video is with the bits from this PR


https://user-images.githubusercontent.com/10873410/148609621-4967c05f-8a05-4b88-ac08-fcb9e00eb9d4.mp4


https://user-images.githubusercontent.com/10873410/148609636-f0bf9f75-60dc-473c-b9b9-3a0d6cefd728.mp4

rickbrew:
@dreddy-work is this a debug build of the SWF.dll ? That could explain it ...

dreddy-work:
> @dreddy-work is this a debug build of the SWF.dll ? That could explain it ...

yes. they are debug builds.  I wouldn't worry much on performance for now. And, thanks for confirming the fix.

AraHaan:
@dreddy-work did this also fix #6421?

PR #6488, opened 2022-01-11
Fixing incorrect mouse tooltip for TabPage of TabControl

### LINKED ISSUE #4750: [Accessibility] The mouse tooltip of second TabPage is not correct when setting TabControl, tabpages' tooltip, and another control's tooltip
* .NET Core Version: 
.Net 6.0 SDK build: 6.0.100-Preview.4.21180.17
No, framework is not supported TabPage's keyborad tooltip
* More Info
1. This is a regression issue, found after the fix of PR: https://github.com/dotnet/winforms/pull/4421
2. This issue cannot repro if don't do following step4

**Repro steps:**
1.	Create a Winforms .Net Core app
2.	Add a TabControl, a Tooltip, a Button to form designer 
3.	Focus on TabControl, go to Properties window, set TabControl's ShowToolTips as True, then set TabControl's Tooltip on toolTip1 property as "TabControl", set Tabpages' TooltipText with "TabPage1" and "TabPage2"
4.	Set anther control: button's Tooltip on toolTip1 property as "button1" --- **it is necessary step**
5.	Build and run app

Or open attached application: 
[WinFormsApp12.zip](https://github.com/dotnet/winforms/files/6235818/WinFormsApp12.zip)

**Actual behavior:**
Keyboard to focus on Button, then Tab to Tabpage1, then mouse-over Tagpage2 to show mouse tooltip, Tabpage2 display is not correct: Tabpage2 tooltip shows TabControl's
image
TabPageTooltipIssue

**Expected behavior:**
Tabpage2 shows "TabPage2" tooltip

WPMGPRoSToTeMa:
Fixes #4750. The DELTOOL message of tooltip causes moving of the last tool to the slot of the deleted tool in the tool array (e.g. deleting `Button` from the `[Button, TabControl, TabPage1, TabPage2]` results in `[TabPage1, TabControl, TabPage2]`). This breaks the hit-test logic of the tooltip that checks the tools in the reverse order. Thus the TabControl may be checked earlier than any of TabPages and tooltip will be shown for the TabControl and not for the TabPage.

A possible fix is to reorder the tools in the tools array, but there is no way to access it directly. It is possible to change the array through GETTOOLINFO and SETTOOLINFO messages but this may have undesirable side-effects of breaking caches and other things.

The other way is to remove deletion from the keyboard tooltips logic. But it won't actually help because deletion may be called directly.

The only appropriate solution is to not add a tool for the TabControl itself. It doesn't break anything because the TabControl relays mouse messages to the tooltip by itself (no subclassing is needed) and there is no area where the tool for the TabControl will be shown (hit-testing detects only TabPages, the other things are transparent).

## Proposed changes

- Removed adding a tool for the `TabControl` in the tooltip logic

## Customer Impact

- The user sees the correct tooltip in the correct rectangle for a tab page of tab control

## Regression? 

- No

## Risk

- Minimal

## Test methodology 

- Manual testing
- Unit-tests
- CTI (planned)

 

## Test environment(s) 

- Microsoft Windows [Version 10.0.19044.1415]
- .NET 7.0.0-alpha.1.22058.4

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6488)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolTip.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolTip.cs
@@ -842,6 +842,17 @@ namespace System.Windows.Forms
 
         private void SetToolInfo(Control control, string caption)
         {
+            // Don't add a tool for the TabControl itself. It's not needed because:
+            // 1. TabControl already relays all mouse events to its tooltip:
+            // https://docs.microsoft.com/windows/win32/controls/tab-controls#default-tab-control-message-processing
+            // 2. Hit-testing against TabControl detects only TabPages which are separate tools added by TabControl.
+            // This prevents a bug when a TabPage tool is placed before the TabControl tool after reordering caused by a tool deletion.
+            // Also prevents double handling of mouse messages which is caused by subclassing altogether with the message relaying from the TabControl internals.
+            if (control is TabControl)
+            {
+                return;
+            }
+
             IntPtr result = GetTOOLINFO(control, caption).SendMessage(this, (User32.WM)TTM.ADDTOOLW);
 
             if ((control is TreeView tv && tv.ShowNodeToolTips)
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolTipTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolTipTests.cs
@@ -866,6 +866,29 @@ namespace System.Windows.Forms.Tests
                 Times.Once);
         }
 
+        [WinFormsFact]
+        public void ToolTip_SetToolTip_TabControl_DoesNotAddToolForTabControlItself()
+        {
+            // We need a Form because tooltips don't work on controls without a valid parent.
+            using Form form = new();
+            using ToolTip toolTip = new();
+            using TabControl tabControl = new() { ShowToolTips = true };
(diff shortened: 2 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ToolTip.cs
```diff
@@ -842,6 +842,16 @@ private void DestroyAllRegions()
 
         private void SetToolInfo(Control control, string caption)
         {
+            // Don't add a tool for the TabControl itself. It's not needed because:
+            // 1. TabControl already relays all mouse events to its tooltip.
```
Tanya-Solyanik, line 846: Did you find code that is doing git? Or doc reference?

Tanya-Solyanik (dismissed):
Looks good, please send to testing!

src/System.Windows.Forms/src/System/Windows/Forms/ToolTip.cs
WPMGPRoSToTeMa: Added link to the Windows docs.

Tanya-Solyanik (approved):

Amy-Li03:
Testing has completed with no new issue found.

Tanya-Solyanik (approved):

PR #2074, opened 2019-10-11
[Accessibility][ScreenReader]: Fixing tooltip text announcing by NVDA/Narrator

vladimir-krestov:
Fixes #2073
Original bug: [987952](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/987952)

## Proposed changes

- Add AnnounceText method to raise UIA notification (ScreenReader says notification text) when a tooltip is shown

<del>- Add IToolTip interface and implement it to unite elements (Control, ToolStripItem, DataGridViewElement) which can have a tooltip (this is necessary to the correct, simple and short implementation of raising UIA notification.)</del>

> ScreenReader reads the tooltip text of elements of in turn.

## Customer Impact

- A user can hear a text of a shown tooltip

## Regression? 

- No

## Risk

- No

### Before
image

### After
- **A screen reader announces a text with a title of all tooltips (mouse and keyboard). &#x1F4D7;**

image

## Test methodology 

- Manual UI testing

## Accessibility testing  

- Using NVDA and Narrator

 

## Test environment(s) 

- .NET Core Version: 3.1.0-preview1.19458.7
- Microsoft Windows [Version 10.0.18362.356]

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/winforms/pull/2074)

PR #6489, opened 2022-01-11
Fixing announcing the tooltip for the TabPage

### LINKED ISSUE #5835: Narrator is announcing the wrong tooltip for the TabPage
.NET Core Version:
- .Net 6.0 SDK: 6.0.100-rc.1.21430.12

- No (regression from #2538)

**Repro steps:**

1. Create a Winforms .Net Core app.
2. Add a TabControl, a Tooltip, a Button to form designer.
3. Add 3 TabPages.
4. Set tooltips for pages and controls.
5. Set ShowToolTips = true.
6. Build and run app.
7. Set focus to button.
8. Hover over the TabPage.

**Actual behavior:**
The narrator announces a tooltip for TabControl
Issue-5835

**Expected behavior:**
The narrator should announces a tooltip for TabPage

WPMGPRoSToTeMa:
Fixes #5835. There are some controls that are adding their own tools (e.g. TabControls adds a tool for each TabPage) and for these tools we don't have an associated caption. Moreover these tools are typically retrieving the caption using TTN_NEEDTEXT notification.

The simplest way to fix this issue is to retrieve the caption using the `GetWindowText` function. At `WmShow` we know that the caption was already retrieved and set by the tooltip, so it can be retrieved using this API function. This way works for both static and dynamic captions.

The more complex way is to change the logic of the current tool retrieval and also the most important to change the logic of the caption retrieval, to support the dynamic ones.

The first approach was chosen as it is very simple and works without any downsides. The second approach also means that we are actually retrieving the caption twice: once by the tooltip internal logic and then once by our announcing logic.

## Proposed changes

- Changed `GetCaptionForTool` to `GetWindowText` in the caption retrieval logic inside `WmShow` method

## Customer Impact

- The user of Narrator hears the correct caption for a tab page's tooltip

## Regression? 

- The announcing feature was introduced in #2074 and it already had issue for tab pages

## Risk

- Minimal

## Test methodology 

- Manual testing
- Unit-tests
- CTI (planned)

## Test environment(s) 

- Microsoft Windows [Version 10.0.19044.1415]
- .NET 7.0.0-alpha.1.22058.4

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6489)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolTip.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolTip.cs
@@ -642,7 +642,7 @@ namespace System.Windows.Forms
                 return;
             }
 
-            tool.AccessibilityObject.RaiseAutomationNotification(
+            tool.AccessibilityObject.InternalRaiseAutomationNotification(
                 Automation.AutomationNotificationKind.ActionCompleted,
                 Automation.AutomationNotificationProcessing.All,
                 $"{ToolTipTitle} {text}");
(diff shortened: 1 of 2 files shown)


Tanya-Solyanik (approved):
Looks good!

dreddy-work (approved):

Olina-Zhang:
Tested this change on Win11, Win10 and Win8.1, and confirmed the only issue on Win8.1 is expected behavior.

PR #6493, opened 2022-01-11
Update ListViewGroup.CollapsedState from native when changed by keyboard

### LINKED ISSUE #6468: ListView will fire the GroupCollapsedStateChanged when focus is on the ListViewItem and user expands/collapses a ListViewGroup using the keyboard
.NET Core Version: 

-  .NET Core SDK: 7.0.0-alpha.1.21568.0

- No. `ListViewGroup.CollapsedState` was implemented in .NET Core 5.0 (#3155) and from (#5960)

**Problem description:**
When the user uses the right or left key to expand/collapse the `ListViewGroup`, the focus will be on the `ListViewItem` in the group. At this time, continue to use the left/right key to collapse/expand the `ListViewGroup`, and the `GroupCollapsedStateChanged` event will be triggered, but the state doesn't change correspondingly.
ExpandCollapsed

**Expected behavior:**
The expected result should be one of the following:
1. Continue to use the left/right key to collapse/expand the `ListViewGroup`, and the `GroupCollapsedStateChanged` event will be triggered and the state changed correspondingly.
2. Continue to use the left/right key to collapse/expand the `ListViewGroup`, and the `GroupCollapsedStateChanged` event will not be triggered and the state doesn't change correspondingly.

**Minimal repro:**
1. Create a `Form` with a `ListView`.
2. Add` ListViewGroups` to the `ListView` (`ListViewGroups` should have `CollapsedState = ListViewGroupCollapsedState.Expanded` or `CollapsedState = ListViewGroupCollapsedState.Collapsed`).
3. Add `EventHandler` for `GroupCollapsedStateChanged`.
4. Run the application.
5. Expand/collapse a `ListViewGroup` using the keyboard.
[ListViewGroupState.zip](https://github.com/dotnet/winforms/files/7819708/ListViewGroupState.zip)

**More Info:**
This issue can't reproduce when using mouse to expand/collapse `ListViewGroup`.

jbhensley:
Fixes #6468

## Proposed changes

Previously, `ListView` would toggle `CollapsedState` of current `ListViewGroup` between `Expanded` and `Collapsed` if a `ListViewItem` was selected and:

- The left arrow key was pressed and the group state was `Expanded`
- The right arrow key was pressed and the group state was `Collapsed`

It did this regardless of whether the key press triggered an actual change in native state resulting in repeated `GroupCollapsedStateChanged` events and the group's `CollapsedState` property becoming out of sync with native state.

This change reads native state on arrow key press and updates `CollapsedState`/raises `GroupCollapsedStateChanged` only when the state is different.

## Customer Impact

`ListViewGroup.CollapsedState` should no longer be out of sync with native state and phantom `GroupCollapsedStateChanged` events should no longer fire.

## Regression? 

- No. Appears that issue existed in initial commit of work.

## Test methodology 

- Unit tests
- UI integration test
- Manual testing

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6493)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
@@ -6941,7 +6941,7 @@ namespace System.Windows.Forms
                             break;
                         }
 
-                        var nativeState = group.GetNativeCollapsedState();
+                        ListViewGroupCollapsedState nativeState = group.GetNativeCollapsedState();
                         if (nativeState != group.CollapsedState)
                         {
                             group.SetCollapsedStateInternal(nativeState);
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
@@ -6931,22 +6931,22 @@ namespace System.Windows.Forms
                 case User32.WM.KEYUP:
                     int key = (int)m.WParamInternal;
 
+                    // User can collapse/expand a group using the keyboard by focusing the group header and using left/right
                     if (GroupsDisplayed && (key is User32.VK.LEFT or User32.VK.RIGHT) && SelectedItems.Count > 0)
                     {
                         ListViewGroup group = SelectedItems[0].Group;
 
-                        if (group is null || group.CollapsedState is ListViewGroupCollapsedState.Default
-                            || (key == User32.VK.LEFT && group.CollapsedState is ListViewGroupCollapsedState.Collapsed)
-                            || (key == User32.VK.RIGHT && group.CollapsedState is ListViewGroupCollapsedState.Expanded))
+                        if (group is null || group.CollapsedState is ListViewGroupCollapsedState.Default)
                         {
                             break;
                         }
 
-                        group.SetCollapsedStateInternal(group.CollapsedState == ListViewGroupCollapsedState.Expanded
-                                                ? ListViewGroupCollapsedState.Collapsed
-                                                : ListViewGroupCollapsedState.Expanded);
-
-                        OnGroupCollapsedStateChanged(new ListViewGroupEventArgs(Groups.IndexOf(group)));
+                        var nativeState = group.GetNativeCollapsedState();
+                        if (nativeState != group.CollapsedState)
+                        {
+                            group.SetCollapsedStateInternal(nativeState);
+                            OnGroupCollapsedStateChanged(new ListViewGroupEventArgs(Groups.IndexOf(group)));
+                        }
                     }
 
                     break;
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroup.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroup.cs
@@ -413,6 +413,11 @@ namespace System.Windows.Forms
                 throw new InvalidOperationException(nameof(ListView));
             }
 
+            if (!ListView.GroupsEnabled)
+            {
+                return ListViewGroupCollapsedState.Default;
+            }
+
             LVGS state = (LVGS)User32.SendMessageW(ListView, (User32.WM)LVM.GETGROUPSTATE, ID, (nint)(LVGS.COLLAPSIBLE | LVGS.COLLAPSED));
             if (!state.HasFlag(LVGS.COLLAPSIBLE))
             {
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroup.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroup.cs
@@ -408,13 +408,13 @@ namespace System.Windows.Forms
 
         internal ListViewGroupCollapsedState GetNativeCollapsedState()
         {
-            if(ListView is null)
+            if (ListView is null)
             {
-                throw new InvalidOperationException();
+                throw new InvalidOperationException(nameof(ListView));
             }
 
             LVGS state = (LVGS)User32.SendMessageW(ListView, (User32.WM)LVM.GETGROUPSTATE, ID, (nint)(LVGS.COLLAPSIBLE | LVGS.COLLAPSED));
-            if(!state.HasFlag(LVGS.COLLAPSIBLE))
+            if (!state.HasFlag(LVGS.COLLAPSIBLE))
             {
                 return ListViewGroupCollapsedState.Default;
             }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroup.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroup.cs
@@ -6,6 +6,8 @@ using System.ComponentModel;
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.Runtime.Serialization;
+using static Interop;
+using static Interop.ComCtl32;
 
 namespace System.Windows.Forms
 {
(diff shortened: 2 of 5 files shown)


jbhensley:
CI has turned up failing accessibility tests. Initial quick review reveals these may be out of sync with the native state of the control also. Will look into further.

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
```diff
@@ -6931,22 +6931,22 @@ protected override void WndProc(ref Message m)
...
                         {
                             break;
                         }
 
-                        group.SetCollapsedStateInternal(group.CollapsedState == ListViewGroupCollapsedState.Expanded
-                                                ? ListViewGroupCollapsedState.Collapsed
-                                                : ListViewGroupCollapsedState.Expanded);
-
-                        OnGroupCollapsedStateChanged(new ListViewGroupEventArgs(Groups.IndexOf(group)));
+                        var nativeState = group.GetNativeCollapsedState();
```
RussKie, line 6944: We only use `var` when the type is obvious (e.g. `new Type()`). In this case it isn't, please use the actual type.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroup.cs
```diff
@@ -404,6 +406,22 @@ private void Deserialize(SerializationInfo info, StreamingContext context)
             }
         }
 
+        internal ListViewGroupCollapsedState GetNativeCollapsedState()
+        {
+            if(ListView is null)
```
RussKie, line 411: [nit] missing space. Ditto below. 
Looks like some rule is missing from the .editorconfig.... 🤔

src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroup.cs
```diff
@@ -404,6 +406,22 @@ private void Deserialize(SerializationInfo info, StreamingContext context)
...
+            {
+                throw new InvalidOperationException();
```
RussKie, line 413:
```suggestion
                throw new InvalidOperationException(nameof(ListView));
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewGroup.ListViewGroupAccessibleObjectTests.cs
```diff
@@ -702,30 +703,43 @@ public void ListViewGroupAccessibleObject_ExpandCollapseStateFromKeyboard_Return
             // The MSDN page tells us what bits of lParam to use for each of the parameters.
             // All we need to do is some bit shifting to assemble lParam
             // lParam = repeatCount | (scanCode << 16)
-            nint keyCode = (nint)Keys.Left;
+            nint keyCode = (nint)Keys.Up;
```
RussKie, line 706: Why is this changed?

RussKie (commented):
👍

RussKie (dismissed):

src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroup.cs
```diff
@@ -404,6 +406,22 @@ private void Deserialize(SerializationInfo info, StreamingContext context)
...
 
+        internal ListViewGroupCollapsedState GetNativeCollapsedState()
+        {
+            if (ListView is null)
+            {
+                throw new InvalidOperationException(nameof(ListView));
+            }
+
+            LVGS state = (LVGS)User32.SendMessageW(ListView, (User32.WM)LVM.GETGROUPSTATE, ID, (nint)(LVGS.COLLAPSIBLE | LVGS.COLLAPSED));
+            if (!state.HasFlag(LVGS.COLLAPSIBLE))
```
Tanya-Solyanik, line 417: `COLLAPSIBLE` is defined only in common controls v6, perhaps we need additional logic to support cases when visual styles are disabled(v5)?
><html>
<body>


LVGS_COLLAPSIBLE | Version 6.00 and later. The group can be collapsed.
-- | --



</body>
</html>

https://docs.microsoft.com/windows/win32/api/commctrl/ns-commctrl-lvgroup

RussKie: ListView doesn't support groups without visual styles
image

RussKie: https://github.com/dotnet/winforms/blob/12d87648900ba1668c0bad7e63c4cba8f58c5abb/src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs#L1020-L1021

dreddy-work:
@Olina-Zhang can you get this change tested?

VinoWang01:
@jbhensley, tested the private binaries that we built for GH issue [6468](https://github.com/dotnet/winforms/issues/6468)(GH Pull: [#6493](https://github.com/dotnet/winforms/pull/6493)), the issue is fixed and except the issue you mentioned in [here](https://github.com/dotnet/winforms/pull/6493#issuecomment-1010314986) we find a not regression issue [6699](https://github.com/dotnet/winforms/issues/6699), it also can reproduce without private, so we open a new issue to track it.

In addition, all migrated RT core case passed.

After Fix:
FixIssue

vladimir-krestov (approved):
LGTM ✔️ Please rework Tanya's and Igor's points.

RussKie:
I'd like to get a clarification for https://github.com/dotnet/winforms/pull/6493#discussion_r785666208, other than this - 👍

jbhensley:
> I'd like to get a clarification for [#6493 (comment)](https://github.com/dotnet/winforms/pull/6493#discussion_r785666208), other than this - 👍

There were a series of errors all predicated upon an initial logical flaw. That was that selecting an item and pressing left will collapse a group. In reality, selecting an item and pressing left will navigate leftward among items until an initial one is reached, at which point it will do nothing. The correct way to collapse a group with the keyboard is to select the group header and then press left. Selecting the group header can be done by pressing the up arrow. 

This flaw was coded into `ListView` by setting `ListViewGroup.CollapsedState` to `Collapsed` if left arrow were pressed and `SelectedItems.Count > 0`. In that case, managed `CollapsedState` and native state would not match.

The test carried this flaw forward by selecting an item, sending a left arrow key press, and then verifying that `CollapsedState` == `Collapsed`. The intention of the test was to collapse the group via keyboard. However, even though `CollapsedState` reported this to be true, the native state of the control was still expanded.

The flaw was corrected in `ListView` by checking if the native collapsed state has changed upon arrow key press and updating `CollapsedState` accordingly. This means that `CollapseState` no longer became `Collapsed` by simply selecting an item and pressing left. At this point, the test failed. 

The test was updated in two ways:

- To first send `Keys.Up` so that the header would be selected before sending `Keys.Left`. This is the necessary combination to collapse the group
- Asserting the native state of the control matches what we expect. E.g. if the intention of the key combination was to collapse the group then assert that the group really is collapsed natively. This gives us confidence when we then assert that the managed `CollapsedState` property is correct

jbhensley:
Yikes! I just spotted another flaw in the original code. Not the same as what's being fixed with this PR, but it's very related. This:

https://github.com/dotnet/winforms/blob/150aabc4cf47c303f980fbc6313c329e6da8583c/src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs#L6934-L6937

I tested and it's completely possible to do the following:

- Select an item in one group (group 1)
- Hold down Ctrl and click the group header for a different group (group 2)
- Use the left arrow key, which will collapse the second group since its header is selected

The code will pull the group from `SelectedItems[0]` which will be group 1, not group 2.

I'm going to open a separate issue for this.

RussKie (approved):

RussKie:
Thank you.

PR #6059, opened 2021-10-25
Fixing of the contrast of the borders "Popup" and "Standard" buttons

SergeySmirnov-Akvelon:
Fixes #6021

## Proposed changes
- The problem is reproduced because the default colors are too low contrast.
- As a fix for the `Popup` button, added logic to darken the current color.
- As a fix for the `Standard` button, added logic for drawing the border with the necessary contrast.

**Notes:**
`Flat` buttons do not need to be fixed as they already have the necessary contrast ratio:

image

The fix does not affect the `System` button, since the Windows team is responsible for it rendering

## Customer Impact
<details>
<summary>"Popup" button</summary>

**Before fix:**
image

**After fix:**
image
</details>

<details>
<summary>"Popup" button in the "Hovered" state</summary>

**Before fix:**
image

**After fix:**
image
</details>

<details>
<summary>"Standard" button</summary>

**Before fix:**
image

**After fix:**
image
</details>

## Regression? 
- No

## Risk
- Minimal

## Test methodology 
- CTI team

## Accessibility testing  
- Accessibility Insights

## Test environment(s) 
- Microsoft Windows [Version 10.0.19041.804]
- .NET Core SDK: 7.0.0-alpha.1.21522.9

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6059)

PR #6497, opened 2022-01-13
Fixing incorrect rendering of button rounded borders for Windows 11

### LINKED ISSUE #6187: We are not rounding corners in button borders on Win11
* .NET Version: 7.0 Preview1

**Problem description:**

https://github.com/dotnet/winforms/pull/6059 updated the colour of the `Button`'s borders, however it regressed the rendering on W11

Here's .NET 7.0 vs .NET 6.0
image
...magnified:
image

**Expected behavior:**

The borders are rendered correctly

SergeySmirnov-Akvelon:
Fixes #6187

## Proposed changes
- The issue is reproduced because we have always drawn rectangular borders, although Windows 11 uses rounded borders.
- Added logic for drawing a rounded border.
- Added logic to draw a line at the bottom to simulate a gradient effect

## Customer Impact
**Before fix:**
image

**After fix:**
image
image
image

## Regression impact
**Before fix:**
Rendering 500 standard buttons on local machine takes an average of **2208** seconds.
Rendering 500 standard buttons on a virtual machine takes an average of **1.999** seconds.

**After fix:**
Rendering 500 standard buttons on local machine takes an average of **2229** seconds.
Rendering 500 standard buttons on a virtual machine takes an average of **1.879** seconds.

## Regression? 
- Yes (from #6059)

## Risk
- Minimal

## Test methodology 
- CTI team

## Test environment(s) 
- Microsoft Windows [Version 10.0.22000.434]
- .NET Core SDK: 7.0.0-alpha.1.22061.11

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6497)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ButtonInternal/ButtonBaseAdapter.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ButtonInternal/ButtonBaseAdapter.cs
@@ -21,6 +21,9 @@ namespace System.Windows.Forms.ButtonInternal
         // Coefficient for darkening the border color of the "Popup" and "Standard" buttons
         protected const float ButtonBorderDarkerOffset = 0.8f;
 
+        // This coefficient was added to simulate the original gradient as in Windows 11.
+        protected const float ButtonBottomBorderDarkerOffset = 0.7f;
+
         internal ButtonBaseAdapter(ButtonBase control) =>
             Control = control.OrThrowIfNull();
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ButtonInternal/ButtonStandardAdapter.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ButtonInternal/ButtonStandardAdapter.cs
@@ -4,6 +4,7 @@
 
 using System.Diagnostics;
 using System.Drawing;
+using System.Drawing.Drawing2D;
 using System.Windows.Forms.Layout;
 using System.Windows.Forms.VisualStyles;
 using static Interop;
@@ -72,12 +73,38 @@ namespace System.Windows.Forms.ButtonInternal
                 pbState,
                 DpiHelper.IsScalingRequirementMet ? Control.HandleInternal : IntPtr.Zero);
 
-            if (!SystemInformation.HighContrast && pbState == PushButtonState.Normal)
+            if (!SystemInformation.HighContrast
+                && ButtonRenderer.RenderWithVisualStyles
+                && !Control.Focused
+                && pbState == PushButtonState.Normal
+                && Control.BackColor == SystemColors.Control)
             {
-                ControlPaint.DrawBorderSimple(
-                    e.GraphicsInternal,
-                    Rectangle.Inflate(bounds, -1, -1),
-                    ControlPaint.Darker(SystemColors.ButtonShadow, ButtonBorderDarkerOffset));
+                if (OsVersion.IsWindows11_OrGreater)
+                {
+                    int radius = 3;
+                    Rectangle rectangle = Rectangle.Inflate(bounds, -1, -1);
+                    rectangle.Height -= 1;
+                    rectangle.Width -= 1;
+
+                    using (Pen pen = ControlPaint.Darker(SystemColors.ButtonShadow, ButtonBorderDarkerOffset).CreateStaticPen(DashStyle.Solid))
+                    {
+                        ControlPaint.DrawRoundedBorder(e.Graphics, pen, rectangle, radius);
+                    }
+
+                    using (Pen pen = ControlPaint.Darker(SystemColors.ButtonShadow, ButtonBottomBorderDarkerOffset).CreateStaticPen(DashStyle.Solid))
+                    {
+                        Point startPoint = new Point(rectangle.X + radius, rectangle.Y + rectangle.Height);
+                        Point endPoint = new Point(rectangle.X + rectangle.Width - radius, rectangle.Y + rectangle.Height);
+                        e.GraphicsInternal.DrawLine(pen, startPoint, endPoint);
+                    }
+                }
+                else
+                {
+                    ControlPaint.DrawBorderSimple(
+                        e.GraphicsInternal,
+                        Rectangle.Inflate(bounds, -1, -1),
+                        ControlPaint.Darker(SystemColors.ButtonShadow, ButtonBorderDarkerOffset));
+                }
             }
 
             // Now overlay the background image or backcolor (the former overrides the latter), leaving a margin.
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ControlPaint.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ControlPaint.cs
@@ -150,6 +150,40 @@ namespace System.Windows.Forms
         public static Color ContrastControlDark
             => SystemInformation.HighContrast ? SystemColors.WindowFrame : SystemColors.ControlDark;
 
+        internal static void DrawRoundedBorder(this Graphics graphics, Pen pen, Rectangle bounds, int cornerRadius)
+        {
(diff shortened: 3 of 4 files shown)


merriemcgaw:
How are you isolating this to Win11 machines? That would be required for both Core and NetFX.

merriemcgaw:
@JeremyKuhne can you take a look at this one too?

src/System.Windows.Forms/src/System/Windows/Forms/ControlPaint.cs
```diff
@@ -150,6 +150,46 @@ internal static Rectangle CalculateBackgroundImageRectangle(Rectangle bounds, Im
...
+
+            // bottom left arc
+            rectangle.X = bounds.Left;
+            path.AddArc(rectangle, 90, 90);
+
+            path.CloseFigure();
+            return path;
+        }
+
+        internal static void DrawRoundedRectangle(this Graphics graphics, Pen pen, Rectangle bounds, int cornerRadius)
```
Tanya-Solyanik, line 185: Would be nice to have "windows11" in the function name or in a comment

src/System.Windows.Forms/src/System/Windows/Forms/ButtonInternal/ButtonStandardAdapter.cs
```diff
@@ -72,12 +73,38 @@ private void PaintThemedButtonBackground(PaintEventArgs e, Rectangle bounds, boo
                 pbState,
                 DpiHelper.IsScalingRequirementMet ? Control.HandleInternal : IntPtr.Zero);
 
-            if (!SystemInformation.HighContrast && pbState == PushButtonState.Normal)
+            if (!SystemInformation.HighContrast
+                && ButtonRenderer.RenderWithVisualStyles
+                && !Control.Focused
+                && pbState == PushButtonState.Normal
+                && Control.BackColor == SystemColors.Control)
```
Tanya-Solyanik, line 80: How do buttons with non-default background look?

Tanya-Solyanik (commented):
I added some nits, and I'm not clear why buttons with custom back color are excluded, otherwise looks good. 

Do we have additional issues about selected/focused/etc borders?

Tanya-Solyanik:
@merriemcgaw  -
>How are you isolating this to Win11 machines? That would be required for both Core and NetFX.

We look up the version using windows API that does not depend on the manifest version

src/System.Windows.Forms/src/System/Windows/Forms/ButtonInternal/ButtonStandardAdapter.cs
```diff
@@ -72,12 +73,38 @@ private void PaintThemedButtonBackground(PaintEventArgs e, Rectangle bounds, boo
                 pbState,
                 DpiHelper.IsScalingRequirementMet ? Control.HandleInternal : IntPtr.Zero);
 
-            if (!SystemInformation.HighContrast && pbState == PushButtonState.Normal)
+            if (!SystemInformation.HighContrast
+                && ButtonRenderer.RenderWithVisualStyles
```
JeremyKuhne, line 77: Why are you adding all these extra cases to `DrawBorderSimple`? Shouldn't they be included with `if (OsVersion.IsWindows11_OrGreater)`?

JeremyKuhne (changes requested):

src/System.Windows.Forms/src/System/Windows/Forms/ControlPaint.cs
```diff
@@ -150,6 +150,46 @@ internal static Rectangle CalculateBackgroundImageRectangle(Rectangle bounds, Im
         public static Color ContrastControlDark
             => SystemInformation.HighContrast ? SystemColors.WindowFrame : SystemColors.ControlDark;
 
+        private static GraphicsPath RoundedRect(Rectangle bounds, int radius)
```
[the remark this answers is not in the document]
JeremyKuhne: If the only caller is DrawRoundedRectangle it shouldn't be a separate method. Move it inline for clarity and to avoid the extra call frame.

JeremyKuhne:
This needs a lot more testing and documentation. There should be screenshots of Windows 11 buttons at different DPIs, with different high contrast settings, and anything else that can be modified with comparisons against our implementation. I'd want to see actual color values as well.

JeremyKuhne:
Was our initial rendering actually different than what Windows draws directly? If not, I don't like the idea of trying to "improve" Windows rendering- we're just asking for major headaches trying to keep in line with Windows changes. If they don't render at a contrast that meets accessibility guidelines that shouldn't be something we should be "patching".

I'll try and get some shots of actual direct Win32 renders and add them.

src/System.Windows.Forms/src/System/Windows/Forms/ButtonInternal/ButtonBaseAdapter.cs
```diff
@@ -20,6 +20,7 @@ internal abstract partial class ButtonBaseAdapter
 
         // Coefficient for darkening the border color of the "Popup" and "Standard" buttons
         protected const float ButtonBorderDarkerOffset = 0.8f;
+        protected const float ButtonBottomBorderDarkerOffset = 0.6f;
```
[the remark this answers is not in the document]
SergeySmirnov-Akvelon: This coefficient was added to simulate the original gradient. The contrast ratio in the original is **1.252**, we have a contrast ratio of **1.253**.

SergeySmirnov-Akvelon:
> How are you isolating this to Win11 machines? That would be required for both Core and NetFX.

Hi @merriemcgaw. We use the `OsVersion.IsWindows11_OrGreater` flag to check the operating system. Depending on the value of the flag, we draw either rectangular or rounded borders. It is available in both frameworks.

SergeySmirnov-Akvelon:
Thanks @Tanya-Solyanik  for your questions.

> I added some nits, and I'm not clear why buttons with custom back color are excluded, otherwise looks good.

By default, we have the following colors for the system button border:
Windows 10 - RGB(173, 173, 173)
Windows 11 - RGB(201, 201, 201)

These colors are not system colors. Should we keep the border colors stored in properties? Or should we get the border colors each time from the button to ensure it's correct?

If we want to support custom colors, then what color should we rely on when calculating a new color for the button border? `Parent.BackColor` or `Button.BackColor`?

> Do we have additional issues about selected/focused/etc borders?

By default, the **selected** and **hovered** button borders have acceptable contrast on Windows 10 and Windows 11.

For **Windows 10**:
image
**Selected** state (yellow rectangle): `Button.BackColor` - **3.44**; `Form.BackColor` - **3.947**
**Hovered** state (rose rectangle): `Button.BackColor` - **3.921**; `Form.BackColor` - **3.947**

For **Windows 11**:
image
**Selected** state (yellow rectangle): `Button.BackColor` - **4.338**; `Form.BackColor` - **3.973**
**Hovered** state (rose rectangle): `Button.BackColor` - **3.559**; `Form.BackColor` - **3.973**

By default, the checked and pressed button borders have low contrast on Windows 11.
For **Windows 10**:
image
**Unchecked** state: `Button.BackColor` - **3.02**; `Form.BackColor` - **3.465**
**Checked** state: `Button.BackColor` - **5.867**; `Form.BackColor` - **3.754**
**Pressed** state: `Button.BackColor` - **3.921**; `Form.BackColor` - **3.947**

For **Windows 11**:
image
**Unchecked** state: `Button.BackColor` - **3.783**; `Form.BackColor` - **3.465**
**Checked** state: `Button.BackColor` - **1.167**; `Form.BackColor` - **1.136**
**Pressed** state: `Button.BackColor` - **1.167**; `Form.BackColor` - **1.136**

I plan on Monday to fix pressed and checked states for Windows 11

Tanya-Solyanik:
@SergeySmirnov-Akvelon  - I thought we decided to close this PR and pursue a windows fix?

PR #6516, opened 2022-01-18
Improve build on Linux

kant2002:
I know, I know, but sometimes I work from Linux and want to react to PR, this is not whole Linux build, but at least I can smoke test compilation errors. Given that you guys do not like to put pressure on CI, I would like to have this capability

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6516)


eng/packageContent.targets
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <_DotnetApiDocsPackageName>Microsoft.Private.Intellisense</_DotnetApiDocsPackageName>
```
RussKie, line 4: Why not just change this?

RussKie (dismissed):

eng/packageContent.targets
kant2002: I guarantee that everybody forget about my case. It’s quite unusual and as such easy to revert that during some cleanup. Name of package is written right, it’s pa to creation which is “wrong”

kant2002:
Can somebody re-run this. I assume this is just test flackiness. If these tests are flacky, should some issues to be created to fix them?

dreddy-work:
> Can somebody re-run this. I assume this is just test flackiness. If these tests are flacky, should some issues to be created to fix them?

Those tests are being worked. Can you please elaborate on what sort of failure you are seeing on Linux on how this is helping?

kant2002:
> Can you please elaborate on what sort of failure you are seeing on Linux

Build complains that "blah, blah, blah cannot find intellisense docs for `Microsoft.Private.Intellisense`. That was because path was constructed like `~/.nuget/packages/Microsoft.Private.Intellisense/.....`, but on Linux path is `~/.nuget/packages/microsoft.private.intellisense/.....`. So I just lowecase package name.

> how this is helping?

It helps me when I'm react to code review, where most implementation on PR is done, and I apply stylistic suggestions, or something which slips. Usually if's something small, I just need to run compilation (with reasonable hope that everything is ok and I do nothing stupid). Right now from what I see, after this change all C# code compiled flawlesly. Obviously I cannot run tests, but still even compilation alone helps me do not do stupid typos, missing commas, etc.

Obviously I do not expect that this is fall into "Supported" scenario, only if something fails on Linux, I may politely ask for inclusion of such trivial changes which I consider neutral for you guys, but really helps me.

eng/packageContent.targets
Tanya-Solyanik: @kant2002  - please add a comment to the targets file that `ToLower` enables compiling on Linux, even though Linux is not supported

kant2002: Done

dreddy-work (approved):

eng/packageContent.targets
```diff
@@ -2,7 +2,8 @@
...
-    <_DotnetApiDocsFilesRoot>$([MSBuild]::NormalizeDirectory('$(NuGetPackageRoot)', '$(_DotnetApiDocsPackageName)', '$(MicrosoftPrivateIntellisenseVersion)', 'IntellisenseFiles', 'windowsdesktop'))</_DotnetApiDocsFilesRoot>
+    <!-- Using ToLower in line below unlock Linux build, event if it's unsupported. -->
```
Tanya-Solyanik, line 5:
```suggestion
    <!-- Using ToLower below to unlock Linux build, even if it's unsupported. -->
```

RussKie (approved):

PR #6523, opened 2022-01-19
Enable nullability in PageSetupDialog

gpetrou:
## Proposed changes

- Enable nullability in PageSetupDialog.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6523)


src/System.Windows.Forms/src/System/Windows/Forms/Printing/PageSetupDialog.cs
```diff
@@ -201,7 +199,7 @@ private Comdlg32.PSD GetFlags()
                 flags |= Comdlg32.PSD.MINMARGINS;
             }
 
-            if (_pageSettings.Margins is not null)
+            if (_pageSettings!.Margins is not null)
```
RussKie, line 202:
```suggestion
            if (_pageSettings?.Margins is not null)
```

src/System.Windows.Forms/src/System/Windows/Forms/Printing/PageSetupDialog.cs
```diff
@@ -238,14 +236,16 @@ private void ResetMinMargins()
         /// </summary>
         private bool ShouldSerializeMinMargins()
         {
-            return _minMargins.Left != 0
+            return _minMargins!.Left != 0
```
RussKie, line 239: We can't do this since `_minMargins` **can** be null. We must check and return false, if it's null.

RussKie (changes requested):

RussKie (approved):

PR #6524, opened 2022-01-19
Do not use flat style in high contrast mode

### LINKED ISSUE #6155: Comboboxes with "Flat" and "Popup" flat styles are not displayed in High Contrast mode
.NET Core Version: 
- 7.0.0-alpha.1.21556.5

- Yes

**Problem description:**
1. Create a `Form` with a `Combobox`
2. Set `Popup` or `Flat` flat style
3. Turn on High Contrast mode
4. Run the application

**Actual behavior:**
Comboboxes should be displayed are not displayed in High Contrast mode:
image

**Expected behavior:**
Comboboxes should be displayed in High Contrast mode.

jbhensley:
Fixes #6155.

## Proposed changes

`ComboBox.FlatStyle` has two modes that do not show up correctly when in high contrast mode:

- `FlatStyle.Flat` renders the `ComboBox` with no border
- `FlatStyle.Popup` renders the same as `FlatStyle.Flat` except when focused/moused over, in which case it attempts to simulate the natively drawn border

Since high contrast mode uses the same color for `ComboBox.BackColor` as it does for `Form.BackColor` and since the control has no border, it is not visually perceptible except for its chevron. This PR checks for high contrast mode and renders the same as `FlatStyle.Standard`.

## Customer Impact

- `ComboBox` will be visible in high contrast mode

## Regression? 

- No

## Risk

Should be minimal with the exception of fringe cases such as:
 
- Possibility of a developer deriving from `ComboBox` and performing their own custom drawing logic that is somehow dependent on the current rendering behavior of the control while in high contrast mode
- Possibility that a visually imperceptible `ComboBox` is the desired outcome when in high contrast mode

### Before

- Left: normal
- Right: high contrast
- Styles (top to bottom): `Standard`, `Flat`, `Popup`

image

### After

Same order as before

image

## Test methodology 

- Manual testing

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6524)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.cs
@@ -3903,7 +3903,7 @@ namespace System.Windows.Forms
                     break;
 
                 case WM.PAINT:
-                    if (GetStyle(ControlStyles.UserPaint) == false && (FlatStyle == FlatStyle.Flat || FlatStyle == FlatStyle.Popup) && !DisplayInformation.HighContrast)
+                    if (GetStyle(ControlStyles.UserPaint) == false && (FlatStyle == FlatStyle.Flat || FlatStyle == FlatStyle.Popup) && !SystemInformation.HighContrast)
                     {
                         using var dropDownRegion = new Gdi32.RegionScope(FlatComboBoxAdapter._dropDownRect);
                         using var windowRegion = new Gdi32.RegionScope(Bounds);
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.cs
@@ -3903,7 +3903,7 @@ namespace System.Windows.Forms
                     break;
 
                 case WM.PAINT:
-                    if (GetStyle(ControlStyles.UserPaint) == false && (FlatStyle == FlatStyle.Flat || FlatStyle == FlatStyle.Popup))
+                    if (GetStyle(ControlStyles.UserPaint) == false && (FlatStyle == FlatStyle.Flat || FlatStyle == FlatStyle.Popup) && !DisplayInformation.HighContrast)
                     {
                         using var dropDownRegion = new Gdi32.RegionScope(FlatComboBoxAdapter._dropDownRect);
                         using var windowRegion = new Gdi32.RegionScope(Bounds);
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.cs
@@ -3903,7 +3903,7 @@ namespace System.Windows.Forms
                     break;
 
                 case WM.PAINT:
-                    if (GetStyle(ControlStyles.UserPaint) == false && (FlatStyle == FlatStyle.Flat || FlatStyle == FlatStyle.Popup) && !SystemInformation.HighContrast)
+                    if (GetStyle(ControlStyles.UserPaint) == false && (FlatStyle == FlatStyle.Flat || FlatStyle == FlatStyle.Popup) && !(SystemInformation.HighContrast && BackColor == SystemColors.Window))
                     {
                         using var dropDownRegion = new Gdi32.RegionScope(FlatComboBoxAdapter._dropDownRect);
                         using var windowRegion = new Gdi32.RegionScope(Bounds);


dreddy-work:
Tagging @Olina-Zhang for testing.

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.cs
```diff
@@ -3903,7 +3903,7 @@ protected override void WndProc(ref Message m)
                     break;
 
                 case WM.PAINT:
-                    if (GetStyle(ControlStyles.UserPaint) == false && (FlatStyle == FlatStyle.Flat || FlatStyle == FlatStyle.Popup))
+                    if (GetStyle(ControlStyles.UserPaint) == false && (FlatStyle == FlatStyle.Flat || FlatStyle == FlatStyle.Popup) && !DisplayInformation.HighContrast)
```
vladimir-krestov, line 3906: We use `SystemInformation` for it:
```suggestion
                    if (GetStyle(ControlStyles.UserPaint) == false && (FlatStyle == FlatStyle.Flat || FlatStyle == FlatStyle.Popup) && !SystemInformation.HighContrast)
```

vladimir-krestov (commented):
Looks good

Olina-Zhang:
Tested this PR change for comboBox control with flatStyle: Pop and Flat on Windows 10 and Windows 11 high contrast modes. Issue: [6155](https://github.com/dotnet/winforms/issues/6155) is fixed and no new issue found. See testing result as following:

- Before on Windows 10
image

- After on Windows 10
image

- Before on Windows 11
image

- After on Windows 11
image

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.cs
```diff
@@ -3903,7 +3903,7 @@ protected override void WndProc(ref Message m)
                     break;
 
                 case WM.PAINT:
-                    if (GetStyle(ControlStyles.UserPaint) == false && (FlatStyle == FlatStyle.Flat || FlatStyle == FlatStyle.Popup))
+                    if (GetStyle(ControlStyles.UserPaint) == false && (FlatStyle == FlatStyle.Flat || FlatStyle == FlatStyle.Popup) && !SystemInformation.HighContrast)
```
Tanya-Solyanik, line 3906: Why did you change `DisplayInformation` to `SystemInformation`? Is the former not updated properly?

jbhensley: Suggested by @vladimir-krestov here:

https://github.com/dotnet/winforms/pull/6524#discussion_r807610248

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.cs
Tanya-Solyanik: @vladimir-krestov - did you find any problems with `DisplayInformation`? If there are any, we should fix them or get rid of this class.

vladimir-krestov: No, I didn't. It's just for consistency. Will take a look at these classes

Tanya-Solyanik (dismissed):

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.cs
RussKie: I think `DisplayInformation.HighContrast` is the right approach:
https://github.com/dotnet/winforms/blob/10b647bb2f19bea7b29189669fe29952c1e6f6ec/src/System.Windows.Forms/src/System/Windows/Forms/DisplayInformation.cs#L67-L80

RussKie: I agree with @Tanya-Solyanik on this on, `DisplayInformation.HighContrast` is the right approach:
https://github.com/dotnet/winforms/blob/10b647bb2f19bea7b29189669fe29952c1e6f6ec/src/System.Windows.Forms/src/System/Windows/Forms/DisplayInformation.cs#L67-L80

jbhensley: Reverted last commit, returning to use of `DisplayInformation`

Tanya-Solyanik (approved):

Tanya-Solyanik:
@SergeySmirnov-Akvelon  - please take a look.

SergeySmirnov-Akvelon:
This fix leads to changes in the behavior of the `ComboBoxes`, which are possibly a regression.

In the past, in the .NET Framework (4.7.2 and 4.8) and in all versions of the .NET Core (3.1, 5.0, 6.00, we colored the `ComboBoxe using the color from the `BackColor` property.
image

Now we color only `ComboBoxes` that have `DropDownStyle = ComboBoxStyle.Simple` settings. 
image

@Tanya-Solyanik, @merriemcgaw, is this behavior expected?

Tanya-Solyanik:
@SergeySmirnov-Akvelon  - excellent investigation! Thank you. I agree that we should not ignore the back ground color, this is a problem. I think we are honoring the developer-set color for buttons in most cases. 
@jbhensley  - Are you interested in fixing this case?

jbhensley:
Sorry, I haven't had a lot of time recently to look at this.

After some testing, I determined that the native `ComboBox` does not honor `SetBkMode` (the native way of setting the background color) if it has the `CBS_DROPDOWNLIST` style. This is likely because the text portion of the control is disabled in that mode (i.e. user cannot type into the text part) and so it is rendered with a disabled color.

Below is a screenshot of all combinations of `ComboBox.FlatStyle` and `ComboBox.DropDownStyle` with a background color set. The styles `FlatStyle.Flat` and `FlatStyle.Popup` are not native styles and rendering is handled by the managed framework, which uses the control's `BackColor`. On the other hand, `FlatStyle.Standard` lets the native control do the rendering, which ignores `BackColor` for `DropDownList`.

The change in this PR is to treat `FlatStyle.Flat` and `FlatStyle.Popup` the same as `FlatStyle.Standard` when in high contrast mode, allowing the native control to render. That is the source of the change in behavior.

The solution here is not immediately obvious. Trying to fix back color for `DropDownList` in `Standard` mode deviates from native behavior and is a larger change from existing. Leaving it be doesn't allow us to get around the high contrast issue by just letting native render the control and probably means a more complex implementation of `FlatStyle.Flat`/`FlatStyle.Popup` that takes into account high contrast mode. I can try exploring this second option, but it may take a couple of weeks before I can do so.

image

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.cs
Tanya-Solyanik, line 3906: Consider changing the added condition to "&& (Not High contrast OR (High Contrast AND no custom background color))"

Then the case that @SergeySmirnov-Akvelon found does not have the border but keeps the custom background...

Tanya-Solyanik:
@jbhensley  - I'm not sure if the issue related to custom background is worth making the existing implementation more complex. I suggest to keep the .NET Framework behavior in the case of custom back color and high contrast.

SergeySmirnov-Akvelon (approved):

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.cs
jbhensley: The conditional is getting a bit ugly, but this is an elegant solution.

jbhensley:
High contrast flat mode `ComboBox` with default `BackColor` (top) and custom `BackColor` (bottom) after latest change:

image

RussKie (approved):

RussKie:

Tanya-Solyanik (approved):
Looks good, thank you!

PR #6538, opened 2022-01-24
Move FabricBot rules to Configuration-As-Code

RussKie:
Migrates FabricBot automation for the dotnet-api-docs to [Config-as-Code](https://microsoft.sharepoint.com/teams/FabricBot/SitePages/Config-as-code.aspx). The file is a dump of the automation rules that are currently active and doesn't contain any additional changes.

Similar to https://github.com/dotnet/runtime/pull/62297 and https://github.com/dotnet/dotnet-api-docs/pull/7442.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6538)

### DIFF
--- a/.github/fabricbot.json
+++ b/.github/fabricbot.json
@@ -1,0 +1,1261 @@
+[
+  {
+    "taskType": "trigger",
+    "capabilityId": "CodeFlowLink",
+    "subCapability": "CodeFlowLink",
+    "version": "1.0",
+    "config": {
+      "taskName": "Add a CodeFlow link to new pull requests"
+    }
+  },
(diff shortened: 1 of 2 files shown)


dreddy-work:
If it can be edited only in fabric tool, isn't it going to be double the work?

RussKie:
> If it can be edited only in fabric tool, isn't it going to be double the work?

The config can be edited manually, that's the whole point.

dreddy-work (approved):

PR #6544, opened 2022-01-25
Clean up code in AxHost/DataObject/ContainerControl

JeremyKuhne:
Specifically targeting a few things:

- Use interpolated strings
- Clean up field naming style
- General reduction of blue info messages

I tweaked some interop definitions to allow using readonly Guids when requesting interfaces.

Note that AxHost.State custom serializes- I went through it carefully to make sure updating the fields wouldn't impact serialization.

Most of the strings are in Debug tracing. Moving to interpolated strings will avoid a lot of needless string formatting when debugging. I'm skeptical of the value of maintaining these at all actually, but for now I'm not going to push leaving them in. (They make reading the code significantly more difficult.)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6544)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/AxHost.State.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/AxHost.State.cs
@@ -6,7 +6,6 @@
 
 using System.ComponentModel;
 using System.Diagnostics;
-using System.Globalization;
 using System.Runtime.InteropServices;
 using System.Runtime.Serialization;
 using static Interop;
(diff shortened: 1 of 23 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/AxHost.State.cs
```diff
@@ -369,19 +357,19 @@ void ISerializable.GetObjectData(SerializationInfo si, StreamingContext context)
...
                 {
                     try
                     {
                         using var propertyBagBinaryStream = new MemoryStream();
-                        PropertyBagBinary.Write(propertyBagBinaryStream);
-                        si.AddValue(nameof(PropertyBagBinary), propertyBagBinaryStream.ToArray());
+                        _propertyBag.Write(propertyBagBinaryStream);
+                        si.AddValue(PropertyBagSerializationName, propertyBagBinaryStream.ToArray());
                     }
                     catch (Exception e)
```
RussKie, line 370:
```suggestion
                    catch (Exception ex)
```

RussKie (approved):

RussKie:
Since you spent some time looking through these classes, can you please review https://github.com/dotnet/winforms/pull/6376?

PR #6545, opened 2022-01-25
Enable nullability in ButtonBase

gpetrou:
## Proposed changes

- Enable nullability in ButtonBase.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6545)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.cs
@@ -471,7 +471,7 @@ namespace System.Windows.Forms
                 }
 
                 // If they were previously using indexes - this should clear out the image index field.
-                _imageIndex.Key = value!;
+                _imageIndex.Key = value;
                 Invalidate();
             }
         }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.ComponentModel;
 using System.Diagnostics;
 using System.Drawing;
@@ -25,9 +23,9 @@ namespace System.Windows.Forms
         private ContentAlignment _textAlign = ContentAlignment.MiddleCenter;
         private TextImageRelation _textImageRelation = TextImageRelation.Overlay;
         private readonly ImageList.Indexer _imageIndex = new ImageList.Indexer();
-        private FlatButtonAppearance _flatAppearance;
-        private ImageList _imageList;
-        private Image _image;
+        private FlatButtonAppearance? _flatAppearance;
+        private ImageList? _imageList;
+        private Image? _image;
 
         private const int FlagMouseOver = 0x0001;
         private const int FlagMouseDown = 0x0002;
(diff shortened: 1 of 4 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.cs
```diff
@@ -1164,7 +1158,7 @@ protected override void OnPaint(PaintEventArgs pevent)
             if (GetStyle(ControlStyles.UserPaint))
             {
                 Animate();
-                ImageAnimator.UpdateFrames(Image);
+                ImageAnimator.UpdateFrames(Image!);
```
RussKie, line 1161: This at least deserves a comment that [`ImageAnimator`](https://github.com/dotnet/runtime/blob/bc6d349ecd72eec162f0532e6b6218fdbaa07ddb/src/libraries/System.Drawing.Common/src/System/Drawing/ImageAnimator.cs#L103-L108) isn't NRTA'ed, and accepts nulls.

So maybe we should just add a null check and avoid the bang operator at all.

src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.cs
```diff
@@ -470,7 +471,7 @@ public string ImageKey
                 }
 
                 // If they were previously using indexes - this should clear out the image index field.
-                _imageIndex.Key = value;
+                _imageIndex.Key = value!;
```
RussKie, line 474: `Indexer` type isn't annotated correctly, as `Key` accepts nulls:
https://github.com/dotnet/winforms/blob/6365af0d23685c467a74bbfeb9363d3ddec3d50c/src/System.Windows.Forms/src/System/Windows/Forms/ImageList.Indexer.cs#L29-L35

RussKie (changes requested):

src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.cs
gpetrou: I see in https://github.com/dotnet/runtime/blob/7d30da68a91d1cced11a0302cc7a8aa9a0ea9af1/src/libraries/System.Drawing.Common/src/System/Drawing/ImageAnimator.cs#L103 there is already a null check and that nullability is enabled in https://github.com/dotnet/runtime/blob/7d30da68a91d1cced11a0302cc7a8aa9a0ea9af1/src/libraries/System.Drawing.Common/src/System/Drawing/ImageAnimator.cs#L103.

RussKie: The actual type doesn't look like it was annotated (correctly). @jeffhandley would you know the story behind it?

gpetrou: Perhaps I should just add the null check on our side to unblock this PR?

jeffhandley: Sorry; I'm overlooking the issue. It looks to me like ImageAnimator.UpdateFrames expects a non-null Image, but it gracefully no-ops if null is passed in; that seems OK to me--is there an issue with that approach?

My recommendation would be to add the null check on this call site.

dreddy-work (approved):
LGTM.

src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.cs
RussKie: @jeffhandley my understanding of NRTA is that if a method accepts nulls (like in this situation) it should be annotated as nullable.

jeffhandley: > @jeffhandley my understanding of NRTA is that if a method accepts nulls (like in this situation) it should be annotated as nullable.

Sorry for the delayed response, @RussKie. OK, I'm convinced. :wink:. I'll submit a PR to change it to `public static void ImageAnimator.UpdateFrames(Image? image)`.

jeffhandley: Created https://github.com/dotnet/runtime/pull/65949 to change the annotation.

RussKie: Thank you!

PR #6563, opened 2022-01-27
Fix ListViewGroupAccessibleObject.GetNativeGroupId returning incorrect value

### LINKED ISSUE #6306: The accessibility rectangle is drawn incorrectly for a dynamically added ListViewGroup.
.NET Core Version: 
- 7.0.0-alpha.1.21576.4

- No (from .NET Core 6.0)

**Repro steps:**
1. Create a `Form` with a `ListView` and a `Button`.
2. Add an event for a `Button` to add multiple `ListViewGroups` with `ListViewItems`.
3. Run the application
4. Add `ListViewGroups` using the `Button`
5. Get accessibility rectangle for `ListViewGroup` (using `Inspect`)

**Actual behavior:**
The rectangle for the `ListViewGroup` is displayed incorrectly.
image

**Expected behavior:**
The rectangle for the `ListViewGroup` should be displayed correctly.
image

jbhensley:
Fixes #6306

## Proposed changes

`ListViewGroupAccessibleObject.GetNativeGroupId` attempted to retrieve information about a group from the native control based on its positional index. However, the position in `ListView.Groups` does not always correspond to the position in the native control (see more information here: https://github.com/dotnet/winforms/issues/6306#issuecomment-1020301155). Specifically, there is a difference when:

- The user adds a `ListView` to a `Form` and groups to the `ListView` using the visual designer, in which case the `ListView` handle will be created after groups have been added
- The user adds a `ListView` to a `Form` and then creates groups programmatically after the form has been displayed. In this case, the `ListView` handle will be created before groups are added

In the former case, the native control will not contain the default group, while in the latter case it will. This changes the positional index of groups in the native control. 

We can bypass all of the concern regarding position by simply using `ListViewGroup.ID`, which is an incrementing unique identifier passed to Win32 as `LVGROUP.iGroupId` upon creation.

## Customer Impact

- The accessibility rectangle for `ListViewGroups` should show correctly in cases where groups are added after handle creation

## Risk

- Should be minimal

### Before

Here we can see pre-fix that highlighting group 1 with Accessibility Insights returns information for group 2

image

### After

Here we can see post-fix that highlighting group 1 with Accessibility Insights returns the correct information

image

## Test methodology 

- Unit tests
- Accessibility Insights for Windows tool

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6563)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroup.ListViewGroupAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroup.ListViewGroupAccessibleObject.cs
@@ -166,17 +166,14 @@ namespace System.Windows.Forms
                 return LVGS.FOCUSED == (LVGS)User32.SendMessageW(_owningListView, (User32.WM)LVM.GETGROUPSTATE, nativeGroupId, (nint)LVGS.FOCUSED);
             }
 
-            private unsafe int GetNativeGroupId()
+            private int GetNativeGroupId()
             {
-                LVGROUPW lvgroup = new LVGROUPW
+                if (User32.SendMessageW(_owningListView, (User32.WM)LVM.HASGROUP, _owningGroup.ID) == 0)
                 {
-                    cbSize = (uint)sizeof(LVGROUPW),
-                    mask = LVGF.GROUPID,
-                };
+                    return -1;
+                }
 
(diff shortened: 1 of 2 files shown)


SergeySmirnov-Akvelon (approved):

Tanya-Solyanik (approved):
Looks good, thank you!

dreddy-work:
Thank you @jbhensley.

PR #6432, opened 2021-12-27
Implementing support UIA Provider for TreeView control

DmitryGorokhov:
Implemented accessible objects for TreeView and TreeNode. Added needed automation events and unit tests.

Implements part of #3421

## Proposed changes

- Switched on `SupportsUiaProviders` property of `TreeView` class.
- Implemented accessible objects for `TreeView` and `TreeNode` classes.
- Added needed automation events and unit tests.

## Customer Impact

- Improving development experience for `TreeView` control accessibility.

## Regression? 

- No

## Risk

- Minimal

#### `TreeView` before and after

treeInspect

#### Default `TreeNode` before and after

nodeInspect

#### Expanded `TreeNode` before and after

nodeExpandedInspect

#### Leaf `TreeNode` before and after

nodeLeafInspect

#### CheckBox `TreeNode` before and after

NodeIsCheckBox

#### Invisible `TreeNode` before and after

invisibleNode

#### Editable `TreeNode` before and after

nodeEditable

### Accessibility Insights

There is no specific errors. You can see errors on next image, but they are not about accessibility changes were created in this PR.

AI

#### Expand and collapse default node before and after

NBExpandCollapse

#### Check and uncheck node before and after

NBCheckedBox

#### Edit default node before and after

NBEdit

#### `ScrollIntoView` method before

ScrollIntoViewBefore

#### `ScrollIntoView` method after

ScrollIntoViewAfter

This realization sets calling node as `TopNode`  of `TreeView` if the node is invisible. Looks like it works the same. 

### Focus on `TreeView` issue

By default, when we focus on a `TreeView` control, Inspect and Narrator highlight the `SelectedNode` of
 this control. As I know it takes raise an events for a selected node accessible object to notify it was focused. But calling AccessibilityObject forces creation the object. It takes accessible object of the tree, so forces creating the object if it doesn't exist yet.
 To avoid force creation if this is not needed should use check `IsAccessibilityObjectCreated`. But I found next problem when I try to add this check. Highlight after focus doesn't work for the first time if the control has first `TabIndex` on form.  You can see this issue behavior below.
 
 #### Focus on the control with check `IsAccessibilityObjectCreated`
 
It works incorrect.
 

BrokenFocusOnTreeView

 #### Focus on the control without check `IsAccessibilityObjectCreated`

It works great, but not safety due to forcing the object creating.

FocusOnTreeView

[...]

PR #6571, opened 2022-01-28
Remove using empty base AO ctor

DmitryGorokhov:
Resolve comment from pr #6432 

## Proposed changes

- Remove call empty ctor of Accessible Object class in inherited classes.

## Customer Impact

- No

## Regression? 

- No

## Risk

- No

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6571)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.TopRowAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.TopRowAccessibleObject.cs
@@ -16,7 +16,7 @@ namespace System.Windows.Forms
             private int[]? runtimeId;
             private DataGridView? _ownerDataGridView;
 
-            public DataGridViewTopRowAccessibleObject() : base()
+            public DataGridViewTopRowAccessibleObject()
             {
             }
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainItemAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainItemAccessibleObject.cs
@@ -11,7 +11,7 @@ namespace System.Windows.Forms
             private string? _name;
             private readonly DomainItemListAccessibleObject _parent;
 
-            public DomainItemAccessibleObject(string? name, AccessibleObject parent) : base()
+            public DomainItemAccessibleObject(string? name, AccessibleObject parent)
             {
                 _name = name;
                 _parent = (DomainItemListAccessibleObject)parent.OrThrowIfNull();
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainItemListAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainItemListAccessibleObject.cs
@@ -11,7 +11,7 @@ namespace System.Windows.Forms
             const string DefaultName = "Items";
             private readonly DomainUpDownAccessibleObject _parent;
 
-            public DomainItemListAccessibleObject(DomainUpDownAccessibleObject parent) : base()
+            public DomainItemListAccessibleObject(DomainUpDownAccessibleObject parent)
             {
                 _parent = parent;
             }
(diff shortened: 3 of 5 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.TopRowAccessibleObject.cs
```diff
@@ -16,11 +16,11 @@ protected class DataGridViewTopRowAccessibleObject : AccessibleObject
             private int[]? runtimeId;
             private DataGridView? _ownerDataGridView;
 
-            public DataGridViewTopRowAccessibleObject() : base()
+            public DataGridViewTopRowAccessibleObject()
             {
             }
```
WPMGPRoSToTeMa, line 21: Can we also remove the parameterless constructor?

WPMGPRoSToTeMa (approved):

WPMGPRoSToTeMa:
It's safe to remove `: base()` even when it calls non-empty parameterless constructor because it's called implicitly:
> In a derived class, if a base-class constructor is not called explicitly by using the base keyword, the parameterless constructor, if there is one, is called implicitly.
https://docs.microsoft.com/dotnet/csharp/programming-guide/classes-and-structs/using-constructors

The only question whether we want be explicit about calling base parameterless constructors or not.

cc: @RussKie

vladimir-krestov (dismissed):
Approved ✔️

RussKie:
The default .ctors are called implicitly if those are present and no other .ctor is explicitly called.

RussKie (commented):
:shipit: once the comment is addressed.

WPMGPRoSToTeMa:
> The default .ctors are called implicitly if those are present and no other .ctor is explicitly called.

I mean is there a code style rule that will warn when the base parameterless .ctor is called explicitly? (to avoid this in future)

vladimir-krestov (dismissed):

Tanya-Solyanik (approved):

RussKie:
> > The default .ctors are called implicitly if those are present and no other .ctor is explicitly called.
> 
> I mean is there a code style rule that will warn when the base parameterless .ctor is called explicitly? (to avoid this in future)

The compiler will tell you :)

If there is no .ctor defined - one is implicitly added and called. If there is a non-default .ctor (i.e. with parameters) - it will always be called, no default .ctor will be generated.

src/System.Windows.Forms/src/PublicAPI.Shipped.txt
```diff
@@ -6816,7 +6816,6 @@ System.Windows.Forms.DataGridView.DataGridViewAccessibleObject
 System.Windows.Forms.DataGridView.DataGridViewAccessibleObject.DataGridViewAccessibleObject(System.Windows.Forms.DataGridView! owner) -> void
 System.Windows.Forms.DataGridView.DataGridViewControlCollection
 System.Windows.Forms.DataGridView.DataGridViewTopRowAccessibleObject
-System.Windows.Forms.DataGridView.DataGridViewTopRowAccessibleObject.DataGridViewTopRowAccessibleObject() -> void
```
RussKie, line 6819: We must not alter the public API surface. See https://docs.microsoft.com/dotnet/api/system.windows.forms.datagridview.datagridviewtoprowaccessibleobject.-ctor

RussKie (changes requested):

WPMGPRoSToTeMa:
> 
> The compiler will tell you :)
> 
> If there is no .ctor defined - one is implicitly added and called. If there is a non-default .ctor (i.e. with parameters) - it will always be called, no default .ctor will be generated.

This is what I actually meant:
```
public DataGridViewTopRowAccessibleObject() : base()
                                              ^~~~~~
                                              |
                                              Warning: base parameterless constructor is called explicitly, please don't do that
```
And as I can see there is no rule for that :(

src/System.Windows.Forms/src/PublicAPI.Shipped.txt
DmitryGorokhov: Ok, restored paramless ctor and tis file. Thanks

RussKie (approved):

PR #6574, opened 2022-01-29
Operate open file dialog using ComWrappers

kant2002:
That's piggyback on the previous work. I have to provide CCW for `IFileDialogEvents`. The multiselect did not use `ComWrappers`, since it requires me to provide RCW for `IShellItemArray`, but selection works using build-in COM.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6574)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/Interop/Ole32/Interop.CoCreateInstance.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/Ole32/Interop.CoCreateInstance.cs
@@ -18,10 +18,10 @@ internal partial class Interop
 
         [DllImport(Libraries.Ole32, ExactSpelling = true)]
         public static extern HRESULT CoCreateInstance(
-            ref Guid rclsid,
+            in Guid rclsid,
             IntPtr punkOuter,
             CLSCTX dwClsContext,
-            ref Guid riid,
+            in Guid riid,
             out IntPtr ppv);
     }
 }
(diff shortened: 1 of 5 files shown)


RussKie (commented):
LGTM. Can you please describe how you tested it?

kant2002:
I test mostly in my testbed for NativeAOT.
 "Open File dialog", "Open file dialog" with Multiple=True, Save File Dialog and FolderBrowser dialog since they are related areas. Did not test with test app in WinForms repo (named WinFormsTest probably)

RussKie (dismissed):

kant2002:
I made small change. During that change there something whch puzzles me. I switch to using `UniqueInstance` to create new RCW during each call. I geniunely thought that I should Release aquired objects using dispose, but instead of that it produce crashes. Probably due to double dispose.

Observations. QI on IShellItem in callbacks are returning same pointer. So I speculate that actual instance of IShellItem passed so I thought that maybe QI on IShellItem does not call AddRef?

kant2002:
I undo change `Unwrap`=> `UniqueInstance` since that's identical in that case and `Unwrap` is what built-in COM do. I made change because I have issues in NativeAOT which was due to my errors in testing wrong executable. 

Observation about `IShellItem` and reference counting still valid. Hopefully I did not miss something simple.

RussKie (approved):

RussKie:
@JeremyKuhne @AaronRobinsonMSFT it would be great if you could review as well. Thanks

kant2002:
Friendly ping.

JeremyKuhne (approved):

PR #5319, opened 2021-07-22
Add support for IFileOpenDialog

kant2002:
CCW Wrapper in the PR was autogenerated using source generator
and tailored based on my previous experience.
As such I have strange local variables sometimes.
Also I do not implemented anything yet for following interfaces
- [ ] `Shell32.IFileDialogEvents` (CCW needed)
- [x] `Shell32.IShellItem` (RCW from built-in COM)
- [ ] `Shell32.IShellItemArray` (RCW from built-in COM)
- [ ] `Shell32.IFileDialogCustomize` (FileOpenDialogWrapper RCW need [support for that](https://github.com/dotnet/winforms/blob/81774abbcbb885fe906f1c0f813e8175eef8705c/src/System.Windows.Forms/src/System/Windows/Forms/FolderBrowserDialog.cs#L277) interface)

I plan to implement them, when collect all feedback on quality of this PR, so all improvements would be passed to next work.

What to do with code which maybe never called. How to detect that code?  or should I bother with that and provide implementation for all methods?
For example I do not see errors in tests from `Shell32.IFileDialogEvents`, even if this interface used by `VistaDialogEvents` and passed to `IFileOpenDialog.Advise` method. So removing code and look where it fails in tests would be not practical.

Would like to ask question about `IShellItem` and `IShellItemArray` relationships. I will lookup on MSDN, but maybe collective conscionesness maybe help me better.
So question: Does these completely unrelated interfaces, or there some `IShellItem` implementation which implement `IShellItemArray` as well. Answer affects how I implement RCW for these interfaces.

Contributes to #5163

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/5319)

PR #6575, opened 2022-01-29
Prevent accidental changing of IID.IAccessible

kant2002:
I cannot make `IID.IAccessible` as auto-property, since I have to pass that value by reference. I prefer to mark function as `in` instead of `ref`

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6575)


kant2002:
That's follow-up to https://github.com/dotnet/winforms/pull/5319#discussion_r787353071

RussKie (approved):

Issue #6597, opened 2022-02-02
Several DataGridView tests are flaky in arm64 CI build


vladimir-krestov:
* .NET Core Version: 7.0

* Have you experienced this same bug with .NET Framework?: No

**Problem description:**
These unit tests fail sometime in arm64 CI build for successful PRs, that blocking deployment of fixes into sdk.
**Stack trace** ([System.Windows.Forms.Tests_net7.0_arm64.log](https://dev.azure.com/dnceng/public/_build/results?buildId=1584807&view=artifacts&pathAsName=false&type=publishedArtifacts)):

- DataGridView_OnColumnHeadersHeightChanged_InvokeWithHandle_CallsColumnHeadersHeightChanged:
```
System.Windows.Forms.Tests.DataGridViewTests.DataGridView_OnColumnHeadersHeightChanged_InvokeWithHandle_CallsColumnHeadersHeightChanged(columnHeadersWidthSizeMode: EnableResizing, columnHeadersVisible: True, eventArgs: null) [FAIL]
      Assert.Equal() Failure
      Expected: 0
      Actual:   2
      Stack Trace:
        /_/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewTests.cs(1984,0): at System.Windows.Forms.Tests.DataGridViewTests.DataGridView_OnColumnHeadersHeightChanged_InvokeWithHandle_CallsColumnHeadersHeightChanged(DataGridViewColumnHeadersHeightSizeMode columnHeadersWidthSizeMode, Boolean columnHeadersVisible, EventArgs eventArgs)
    System.Windows.Forms.Tests.DataGridViewTests.DataGridView_OnColumnHeadersHeightChanged_InvokeWithHandle_CallsColumnHeadersHeightChanged(columnHeadersWidthSizeMode: EnableResizing, columnHeadersVisible: True, eventArgs: EventArgs { }) [FAIL]
      Assert.Equal() Failure
      Expected: 0
      Actual:   2
      Stack Trace:
        /_/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewTests.cs(1984,0): at System.Windows.Forms.Tests.DataGridViewTests.DataGridView_OnColumnHeadersHeightChanged_InvokeWithHandle_CallsColumnHeadersHeightChanged(DataGridViewColumnHeadersHeightSizeMode columnHeadersWidthSizeMode, Boolean columnHeadersVisible, EventArgs eventArgs)
[...]

PR #6599, opened 2022-02-02
Added "Skip" attribute for fail unit tests in arm64 and x86

vladimir-krestov:
Related Issues:
- #6597
- #6609 
- #6610
- #6635 

## Proposed changes
- Skipped unit tests that fail sometimes in arm64 and x86 environments, that blocks fixes deployment.

## Customer Impact
- No

## Regression? 
- No

## Risk
- Minimal

## Test methodology 
- CI build

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6599)

### DIFF
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewTests.cs
@@ -5,6 +5,7 @@
 using System.ComponentModel;
 using System.Data;
 using System.Drawing;
+using System.Runtime.InteropServices;
 using System.Windows.Forms.TestUtilities;
 using Xunit;
 
@@ -170,7 +171,8 @@ namespace System.Windows.Forms.Tests
         }
 
         [ActiveIssue("https://github.com/dotnet/winforms/issues/6597")]
-        [WinFormsTheory(Skip = "Flaky tests, see: https://github.com/dotnet/winforms/issues/6597")]
+        [ConditionalWinFormsTheory(UnsupportedArchitecture = Architecture.Arm64,
+            Skip = "Flaky tests, see: https://github.com/dotnet/winforms/issues/6597")]
         [MemberData(nameof(ColumnHeadersHeight_SetWithHandle_TestData))]
         public void DataGridView_ColumnHeadersHeight_SetWithHandle_GetReturnsExpected(DataGridViewColumnHeadersHeightSizeMode columnHeadersWidthSizeMode, bool columnHeadersVisible, bool autoSize, int value, int expectedValue, int expectedInvalidatedCallCount)
         {
@@ -238,7 +240,8 @@ namespace System.Windows.Forms.Tests
         }
 
         [ActiveIssue("https://github.com/dotnet/winforms/issues/6597")]
-        [WinFormsTheory(Skip = "Flaky tests, see: https://github.com/dotnet/winforms/issues/6597")]
+        [ConditionalWinFormsTheory(UnsupportedArchitecture = Architecture.Arm64,
+            Skip = "Flaky tests, see: https://github.com/dotnet/winforms/issues/6597")]
         [MemberData(nameof(ColumnHeadersHeight_SetWithParentWithHandle_TestData))]
         public void DataGridView_ColumnHeadersHeight_SetWithParentWithHandle_GetReturnsExpected(DataGridViewColumnHeadersHeightSizeMode columnHeadersWidthSizeMode, bool columnHeadersVisible, bool autoSize, int value, int expectedValue, int expectedInvalidatedCallCount, int expectedLayoutCallCount, int expectedParentLayoutCallCount)
         {
@@ -527,7 +530,8 @@ namespace System.Windows.Forms.Tests
         }
 
         [ActiveIssue("https://github.com/dotnet/winforms/issues/6597")]
-        [WinFormsTheory(Skip = "Flaky tests, see: https://github.com/dotnet/winforms/issues/6597")]
+        [ConditionalWinFormsTheory(UnsupportedArchitecture = Architecture.Arm64,
+            Skip = "Flaky tests, see: https://github.com/dotnet/winforms/issues/6597")]
         [MemberData(nameof(ColumnHeadersHeightSizeMode_SetWithHandle_TestData))]
         public void DataGridView_ColumnHeadersHeightSizeMode_SetWithHandle_GetReturnsExpected(bool columnHeadersVisible, DataGridViewColumnHeadersHeightSizeMode value, int expectedColumnHeadersHeight, int expectedInvalidatedCallCount)
         {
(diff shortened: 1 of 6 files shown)


src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewTests.cs
```diff
@@ -169,7 +169,7 @@ public static IEnumerable<object[]> ColumnHeadersHeight_SetWithHandle_TestData()
             }
         }
 
-        [WinFormsTheory]
+        [WinFormsTheory(Skip = "https://github.com/dotnet/winforms/issues/6597")]
```
RussKie, line 172: We historically skip like this:
https://github.com/dotnet/winforms/blob/81774abbcbb885fe906f1c0f813e8175eef8705c/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TextRendererTests.cs#L460-L463

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripTests.cs
```diff
@@ -7118,7 +7118,7 @@ public void ToolStrip_WndProc_InvokeMouseActivate_Success()
             Assert.True(control.IsHandleCreated);
         }
 
-        [WinFormsFact]
+        [WinFormsFact(Skip = "https://github.com/dotnet/winforms/issues/6597")]
```
RussKie, line 7121: I suppose it's ok to raise a single issue for flaky tests related to the same control, but please separate unrelated controls' tests into own issues. 

Also when raising issues please establish the context of failures, i.e. include stack traces, etc.

RussKie (changes requested):

RussKie:
> Maybe we should skip them only for ARM64 arc.

Right now we only have x86/x64 detection. dotnet/runtime has decorations, but I don't know if they work out of the box or require elbow grease to get those working for us. https://github.com/dotnet/runtime/blob/main/docs/workflow/testing/libraries/filtering-tests.md.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripTests.cs
vladimir-krestov: I created additional issues, please take a look:
DGV: https://github.com/dotnet/winforms/issues/6597
ToolStrip: https://github.com/dotnet/winforms/issues/6610
RichTextBox: https://github.com/dotnet/winforms/issues/6609

vladimir-krestov:
> > Maybe we should skip them only for ARM64 arc.
> 
> Right now we only have x86/x64 detection. dotnet/runtime has decorations, but I don't know if they work out of the box or require elbow grease to get those working for us. https://github.com/dotnet/runtime/blob/main/docs/workflow/testing/libraries/filtering-tests.md.

I tried to find some solution, but didn't find exist one. So created my own. Please check the second commit as example.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewTests.cs
```diff
@@ -169,7 +169,8 @@ public static IEnumerable<object[]> ColumnHeadersHeight_SetWithHandle_TestData()
             }
         }
 
-        [WinFormsTheory]
+        [ActiveIssue("https://github.com/dotnet/winforms/issues/6597")]
+        [WinFormsTheory(Skip = "Flaky tests, see: https://github.com/dotnet/winforms/issues/6597")]
```
dreddy-work, line 173: SkipCondition default to `false`. Does this skip the test or doesn't? unclear from reading the code.

vladimir-krestov: To use `SkipCondition` you need to use `WinFormsTheoryEx`. I added it only to one file for example. Maybe Igor has a better way

dreddy-work: Got it.

dreddy-work: I am fine if we are forced to skip for all archs here in case if we do not agree on this implementation as long as we have tracking issues that we prioritize in this release

RussKie (commented):
Great idea! 

Perhaps we could extend the API a little furhter, so that we could write something like this:
```cs
        [ConditionalWinFormsFact(UnsupportedArchitecture = Architecture.Arm64)]
        public void ToolStrip_WndProc_InvokeMouseActivateWithHandle_Success()
```
...or
```cs
        [ConditionalWinFormsFact(SupportedArchitecture = Architecture.x86)]
        public void ToolStrip_WndProc_InvokeMouseActivateWithHandle_Success()
```
The later will allow us to apply the same attributes to `PARAMTests` and Co tests.

dreddy-work:
@vladimir-krestov , lets get this merged ASAP. CI is making lot of noise.

src/System.Windows.Forms/tests/TestUtilities/WinFormsFactEx.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            get => base.Skip;
+            set
+            {
+                _skipReason = value;
+
+                if (_skipCondition)
+                {
+                    base.Skip = value;
+                }
```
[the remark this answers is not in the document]
vladimir-krestov: We can't do this just like that, because at the moment when we set `Skip`, `UnsupportedArchitecture` is not set yet. So we will always have the exception.

src/System.Windows.Forms/tests/TestUtilities/WinFormsFactEx.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            get => _skipArchitecture;
+            set
+            {
+                _skipArchitecture = value;
+                _skipCondition = RuntimeInformation.ProcessArchitecture == value;
+                if (_skipCondition && _skipReason is not null)
+                {
+                    Skip = _skipReason;
+                }
```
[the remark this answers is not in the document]
vladimir-krestov: null `_skipReason` means, that we didn't set `Skip` yet. In my environment, `Skip` is set first, and then `UnsupportedArchitecture` is set. I'm not sure, that `Skip` is always first. So we may have the exception, even if `Skip` is set. I plan to find some more universal solution, eg. using the attribute constructor. But let's merge it as is to not block Devendar with CI. And continue discussing in the next refactoring PR.

dreddy-work:
Seems there is one more test in debug mode here.  Would you mind adding that to skip list as well.

vladimir-krestov:
Added `Skip` for `ImageList_FinalizerReleasesNativeHandle_ReturnsExpected` as well. But didn't add the conditional attribute, because this test is in the project, that doesn't have references to unit test utilities. So we need to decide how to add conditional attributes for UI tests (add dependecy or copy-paste attributes). But I suggest to do this in the next PR.

dreddy-work:
> Added `Skip` for `ImageList_FinalizerReleasesNativeHandle_ReturnsExpected` as well. But didn't add the conditional attribute, because this test is in the project, that doesn't have references to unit test utilities. So we need to decide how to add conditional attributes for UI tests (add dependecy or copy-paste attributes). But I suggest to do this in the next PR.

Please create a tracking issue so we don't loose it.

dreddy-work (approved):

dreddy-work:
FOllowing is one more to add in your next PR: 
image

dreddy-work:
Another one.

image

image

vladimir-krestov:
> > Added `Skip` for `ImageList_FinalizerReleasesNativeHandle_ReturnsExpected` as well. But didn't add the conditional attribute, because this test is in the project, that doesn't have references to unit test utilities. So we need to decide how to add conditional attributes for UI tests (add dependecy or copy-paste attributes). But I suggest to do this in the next PR.
> 
> Please create a tracking issue so we don't loose it.

Created https://github.com/dotnet/winforms/issues/6653

PR #6608, opened 2022-02-03
Finish ComWrappers for file dialogs

kant2002:
I extract from FileOpenDialogWrapper base class FileDialogWrapper to handle just IFileDialog and derive FileSaveDialogWrapper form it.
- [x] Add `IFileSaveDialog`
- [x] Add `IShellItemArray`
- [x] Add `IFileDialogCustomize`

This put closure on the file/directory browsing dialogs in the current form.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6608)


kant2002:
This PR requires https://github.com/dotnet/winforms/pull/6574 to be merged because of changes in `CoCreateInstance`.

kant2002:
This is ready for review

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.FileDialogWrapper.cs
```diff
@@ -0,0 +1,317 @@
...
+                Marshal.Release(_wrappedInstance);
+                _wrappedInstance = IntPtr.Zero;
+            }
+
+            HRESULT Shell32.IFileDialog.Show(IntPtr parent)
+            {
+                HRESULT retVal;
+                retVal = ((delegate* unmanaged<IntPtr, IntPtr, HRESULT>)(*(*(void***)_wrappedInstance + 3)))
+                    (_wrappedInstance, parent);
+                return retVal;
```
dreddy-work, line 34:
```suggestion
                return ((delegate* unmanaged<IntPtr, IntPtr, HRESULT>)(*(*(void***)_wrappedInstance + 3)))
                    (_wrappedInstance, parent);

```

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.FileDialogWrapper.cs
```diff
@@ -0,0 +1,317 @@
...
+            }
+
+            HRESULT Shell32.IFileDialog.SetFileTypes(uint cFileTypes, Shell32.COMDLG_FILTERSPEC[] rgFilterSpec)
+            {
+                HRESULT retVal;
```
dreddy-work, line 39: i see some places defined as `result `and in some places `retVal`. Good to unify them.

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.FileDialogWrapper.cs
```diff
@@ -0,0 +1,317 @@
...
+                    retVal = ((delegate* unmanaged<IntPtr, uint, Shell32.COMDLG_FILTERSPEC*, HRESULT>)(*(*(void***)_wrappedInstance + 4)))
+                        (_wrappedInstance, cFileTypes, rgFilterSpec_local);
+                }
+
+                return retVal;
+            }
+
+            HRESULT Shell32.IFileDialog.SetFileTypeIndex(uint iFileType)
+            {
+                HRESULT retVal;
```
dreddy-work, line 54: same here.  I do not see a need for local variable here.

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.FileDialogWrapper.cs
```diff
@@ -0,0 +1,317 @@
...
+                    (_wrappedInstance, psi_local);
+                if (result.Failed())
+                {
+                    Marshal.ThrowExceptionForHR((int)result);
+                }
+            }
+
+            void Shell32.IFileDialog.GetFolder(out Shell32.IShellItem? ppsi)
+            {
+                HRESULT result;
```
dreddy-work, line 149: and can be moved to inline?

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.FileDialogWrapper.cs
JeremyKuhne: We generally strive to use standard English terms for variable names unless there is a specific alignment to native definitions. So "result" is the better choice here (and elsewhere).

JeremyKuhne: And this one also does not even need a variable- you can just return from within the fixed block.

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.FileDialogWrapper.cs
```diff
@@ -0,0 +1,317 @@
...
+            void Shell32.IFileDialog.SetTitle(string pszTitle)
+            {
+                HRESULT result;
+                fixed (char* pszTitle_local = pszTitle)
+                {
+                    result = ((delegate* unmanaged<IntPtr, char*, HRESULT>)(*(*(void***)_wrappedInstance + 17)))
+                        (_wrappedInstance, pszTitle_local);
+                }
+
+                if (result.Failed())
```
JeremyKuhne, line 208: As we're doing this over and over again it would be good to add an extension to `HResultExtensions` and start using it:

``` C#
public static void ThrowIfFailed(this HRESULT hr)
{
    if hr.Failed()
    {
        Marshal.ExceptionForHR((int)hr);
    }
}
```

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.FileDialogWrapper.cs
```diff
@@ -0,0 +1,317 @@
...
+                Marshal.FreeCoTaskMem(pszName_local);
+                if (result.Failed())
+                {
+                    Marshal.ThrowExceptionForHR((int)result);
+                }
+            }
+
+            void Shell32.IFileDialog.SetTitle(string pszTitle)
+            {
+                HRESULT result;
```
JeremyKuhne, line 201: Should check `pszTitle` for null and throw here.

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.FileDialogWrapper.cs
```diff
@@ -0,0 +1,317 @@
...
+                {
+                    Marshal.ThrowExceptionForHR((int)result);
+                }
+            }
+
+            void Shell32.IFileDialog.GetFolder(out Shell32.IShellItem? ppsi)
+            {
+                HRESULT result;
+                IntPtr ppsi_local;
+                result = ((delegate* unmanaged<IntPtr, IntPtr*, HRESULT>)(*(*(void***)_wrappedInstance + 13)))
```
JeremyKuhne, line 151: Just throw right away if the called method failed as we're going to throw anyway. There shouldn't be a case where we return a valid object when we throw. Same for other instances.

JeremyKuhne (changes requested):
Some general clean-up feedback. I'll review again once you've made another pass.

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.FileDialogWrapper.cs
```diff
@@ -0,0 +1,230 @@
...
+                ((delegate* unmanaged<IntPtr, IntPtr*, HRESULT>)(*(*(void***)_wrappedInstance + 13)))
+                    (_wrappedInstance, &ppsi_local).ThrowIfFailed();
+                ppsi = ppsi_local == IntPtr.Zero ? null : (Shell32.IShellItem)WinFormsComWrappers.Instance.GetOrCreateObjectForComInstance(ppsi_local, CreateObjectFlags.Unwrap);
+            }
+
+            HRESULT Shell32.IFileDialog.GetCurrentSelection(out Shell32.IShellItem? ppsi)
+            {
+                IntPtr ppsi_local;
+                HRESULT result = ((delegate* unmanaged<IntPtr, IntPtr*, HRESULT>)(*(*(void***)_wrappedInstance + 14)))
+                    (_wrappedInstance, &ppsi_local);
```
dreddy-work, line 118: shouldn't this check for result value?

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.FileDialogWrapper.cs
```diff
@@ -0,0 +1,230 @@
...
+
+            void Shell32.IFileDialog.GetFileName(out string? pszName)
+            {
+                HRESULT result;
+                IntPtr pszName_local;
+                result = ((delegate* unmanaged<IntPtr, IntPtr*, HRESULT>)(*(*(void***)_wrappedInstance + 16)))
+                    (_wrappedInstance, &pszName_local);
+                pszName = Marshal.PtrToStringUni(pszName_local);
+                Marshal.FreeCoTaskMem(pszName_local);
+                if (result.Failed())
```
dreddy-work, line 141: this method can be cleaned further to use `extension `method on `HRESULT `and inline `result `declaration.

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.FileOpenDialogWrapper.cs
```diff
@@ -140,17 +140,40 @@ void Shell32.IFileOpenDialog.GetResults(out Shell32.IShellItemArray? ppenum)
...
                 return result;
             }
+
+            public void AddText(uint dwIDCtl, string pszText)
+            {
+                var targetInterface = IID.IFileDialogCustomize;
+                var result = Marshal.QueryInterface(_wrappedInstance, ref targetInterface, out var thisPtr);
+                if (result != 0)
+                {
+                    throw new System.InvalidCastException();
```
RussKie, line 161:
```suggestion
                    throw new InvalidCastException();
```

src/System.Windows.Forms.Primitives/src/Interop/Interop.IID.cs
```diff
@@ -12,9 +12,15 @@ internal static class IID
         // D57C7288-D4AD-4768-BE02-9D969532D960
```
RussKie, line 12: Please keep entries in this file in alphabetical order

src/System.Windows.Forms/src/System/Windows/Forms/OpenFileDialog.cs
```diff
@@ -143,15 +143,22 @@ private protected override string[] ProcessVistaFiles(IFileDialog dialog)
...
+                    {
+                        results.GetItemAt(i, out IShellItem item);
+                        files[unchecked((int)i)] = GetFilePathFromShellItem(item);
+                    }
+
+                    return files;
+                }
+                finally
+                {
+                    ((IDisposable)results).Dispose();
```
RussKie, line 160:
```suggestion
                    (results as IDisposable)?.Dispose();
```

RussKie (commented):
👍

src/System.Windows.Forms/src/System/Windows/Forms/OpenFileDialog.cs
kant2002: Null check is not needed. We control right now exact type of the COM wrapper instance. So we know that that type implement IDisposable always.

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.FileDialogWrapper.cs
kant2002: Obviously!!

kant2002: Actually not. We return result. That's intended.

dreddy-work:
👍

src/System.Windows.Forms/src/System/Windows/Forms/OpenFileDialog.cs
kant2002: If you take a look at this https://github.com/dotnet/winforms/pull/6685/commits/5e9697957fb7cdfa5fd070ee09641b4d427fd5d4#diff-6fb59eb16f6df895c60c72cf64adbca50feb1556190727a7c648817a891cd0cfR21

you may notice that by knowing explicit type of RCW you can get rid of COM interfaces. I did not go that route in this PR because even this way seems beneficial, it definitely require discussion, or at least not obvious choice.

RussKie: In this linked commit it is obvious. Here we operate on interfaces, so in theory it is possible that a given instance may not implement `IDisposable`. Hence it is better to be safe than sorry :)

kant2002: I can swear in blood that this will never happens. My plan is to remove interfaces where possible and have only concrete classes. If I introduce that in this PR it will produce a lot of noise and distrat with more important problem. Validating that wrapper do not have leaks and correctly convert data. 

In that location we can change `IFileDialog` to `FileDialogWrapper` and all your concerns would disappear since they would be inapplicable. That change will also affect Open File dialog.

If you insist on being pedantic, I will gladly made change, because the faster this is lands, the better I can move to other places.

RussKie: Alright, sign in blood here 📜. 👹

On a serious note, if you're planning to update this code with wrappers in the near future - we can leave it as is. Otherwise, let's add a check. How does this sound?

kant2002: That would be my next PR.

RussKie: Awesome, keeping it as is then.

RussKie:
@msftbot merge if @JeremyKuhne approves.

JeremyKuhne (approved):

PR #6619, opened 2022-02-04
Fixing issue with getting ListViewItem.AccessibilityObject when ListView does not exist (port to 6.0)

### LINKED ISSUE #6230: ListView.Items.RemoveAt causes NullReferenceException

SergeySmirnov-Akvelon:
Fixes #6230 

## Proposed changes
- The issue is reproduced because we have cached the `ListViewItem.AccessibilityObject`. After the changes in #4910, we added logic that, with each call to the `AccessibilityObject` property, checked that the `ListView` exists and, if absent, threw an exception. Now we return null as a `ListViewItem.AccessibilityObject` instead of returning an exception.

- Also, after the focused `ListViewItem` is removed, it remains cached in the `FocusedItem` property. It also forces the Accessibility tool to try to retrieve the data of the `ListViewItem` that is not displayed. As a fix, a logic was added to reset the Focused flag from the `ListViewItem`.

## Customer Impact
**Before fix:**
Issue-6230

**After fix:**
Issue-6230-SecondFix

## Regression? 
- Yes (from #4910)

## Risk
- Minimal

## Test methodology
- Manual testing
- CTI team
- Unit tests

## Accessibility testing  
- Inspect 
- Narrator
- Accessibility Insights

## Test environment(s)
- .NET Core SDK: 6.0.1

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6619)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListView.ListViewNativeItemCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListView.ListViewNativeItemCollection.cs
@@ -432,6 +432,8 @@ namespace System.Windows.Forms
 
                 owner.ApplyUpdateCachedItems();
                 int itemID = DisplayIndexToID(index);
+
+                this[index].Focused = false;
                 this[index].UnHost(true);
 
                 if (owner.IsHandleCreated)
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroup.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroup.cs
@@ -78,17 +78,7 @@ namespace System.Windows.Forms
         }
 
         internal AccessibleObject AccessibilityObject
-        {
-            get
-            {
-                if (_accessibilityObject is null)
-                {
-                    _accessibilityObject = new ListViewGroupAccessibleObject(this, ListView?.Groups.Contains(this) == false);
-                }
-
-                return _accessibilityObject;
-            }
-        }
+            => _accessibilityObject ??= new ListViewGroupAccessibleObject(this, ListView?.Groups.Contains(this) == false);
 
         /// <summary>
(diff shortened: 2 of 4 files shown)


Tanya-Solyanik (approved):

Tanya-Solyanik:
@SteveMCarroll had approved over e-mail

Issue #3421, opened 2020-06-10
Epic - [Accessibility] Implement UIA provider for all controls


vladimir-krestov:
* Base issue: #2146
* Base PR: #3037 

* .NET Core Version: 6.0

**Problem description:**
- Many .Net Core WinForms controls support `MSAA` proxy, which is unmanaged for us. These controls don't support `UIA` provider. This doesn't allow us to change some Accessibility properties or methods to fix some Accessibility bugs. We can't manage our AccessibleObjects fully.

**Expected behavior:**
- All .Net controls support UIA provider

**Minimal repro:**
- Create a .Net Winforms app
- Add any control to the MainForm
- Check `ProviderDescription` Accessibility property of this control using Inspect

## Controls
<details>
  <summary>Done</summary>

- [x]  Button (#3215)
- [x]  CheckBox (#3228)
- [x] ComboBox (#4540)
- [x] DataGridView
- [x] DataGridViewEditingPanel
- [x] DataGridViewTextBoxEditingControl
- [x] ErrorProvider (#2381)
- [x] DomainUpDown (#2382)
- [x] Form (#2096)
- [x] GroupBox 
- [x] Label (#3206)
- [x] ListBox (#935)
- [x] ListView (#3224)
- [x] MonthCalendar (#2090)
- [x] NumericUpDown (#2382)
- [x] PictureBox (#3233)
- [x] ProgressBar
- [x] PropertyGrid
- [x] DocComment
- [x] DropDownButton
- [x] HotCommands
- [x] PropertyGridView
- [x] DropDownHolder
- [x] GridViewEdit
- [x] PictureBox (#3233)
- [x] PropertyGridView
- [x] RadioButton (#3244)
- [x] Splitter (#3246)
- [x] StatusStrip
- [x] ToolStrip
- [x] ToolStripComboBoxControl
- [x] ToolStripNumericUpDownControl
- [x] ToolStripProgressBarControl
- [x] ToolStripTextBoxControl
- [x] UpDownButtons
- [x] ScrollBar/HScrollBar/VScrollBar (#5044)
- [x] TrackBar (#4939)
- [x] LinkLabel (#5157)
- [x] TabControl (#5380)
- [x] CheckedListBox (#5276)
- [x] Panel (#5485)
- [x] TabPage (#5380)
- [x] MenuStrip
- [x] ContextMenuStrip
- [x] DateTimePicker (#6620)
- [x] ToolStripContainer (#6568) (Danil Andrianov)
[...]

PR #6620, opened 2022-02-04
Implementing uia provider support for DTP control

DmitryGorokhov:
Implements part of #3421

## Proposed changes

- Switched on `SupportsUiaProviders` property of `DateTimePicker` class.
- Added needed automation events to fix behavior in Narrator.
- Reworked `DateTimePickerAccessibleObject`, added unit tests.
- Added workaround to check `AccessKey` for dtp in `ControlAccessibleObjectTests`.  

## Customer Impact

- Improving development experience for `DateTimePicker` control accessibility.

## Regression? 

- No

## Risk

- Minimal

## Screenshots 

Changes in accessibility tools.

#### Before

AI

#### After

a

### Narrator & Buddy

The behavior in Narrator was restored after the fix. Seems was resolved problem described in issue #5487.

NB New

### Calendar issue

The Narrator doesn't work with dtp calendar. Seems, it is native control problem, because the Narrator doesn't work with dtp in MFC app at all. You can see it on following gif below.
I attach the test MFC app with dtp: [MFCApplication.zip](https://github.com/dotnet/winforms/files/8098994/MFCApplication.zip)

DateTimePickerMFC

## Test methodology 

- Manually
- Unit testing
- CTI

## Accessibility testing  

- Inspect
- Accessibility Insights
- Narrator
 

## Test environment(s) 

- .Net version: 7.0.100-preview.1.22078.7
- OS version: 10.0.19044

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6620)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.cs
@@ -820,6 +820,8 @@ namespace System.Windows.Forms
             }
         }
 
+        internal override bool SupportsUiaProviders => true;
+
         /// <summary>
         ///  Overrides Text to allow for setting of the value via a string.  Also, returns
         ///  a formatted Value when getting the text.  The DateTime class will throw
@@ -1071,9 +1073,13 @@ namespace System.Windows.Forms
             _onCloseUp?.Invoke(this, eventargs);
             _expandCollapseState = UiaCore.ExpandCollapseState.Collapsed;
 
+            // Raise automation event to annouce new state.
             if (IsAccessibilityObjectCreated)
             {
-                AccessibilityObject.RaiseAutomationEvent(UiaCore.UIA.AutomationFocusChangedEventId);
+                AccessibilityObject.RaiseAutomationPropertyChangedEvent(
+                    UiaCore.UIA.ExpandCollapseExpandCollapseStatePropertyId,
+                    oldValue: UiaCore.ExpandCollapseState.Expanded,
+                    newValue: UiaCore.ExpandCollapseState.Collapsed);
             }
         }
 
@@ -1084,6 +1090,15 @@ namespace System.Windows.Forms
         {
             _onDropDown?.Invoke(this, eventargs);
             _expandCollapseState = UiaCore.ExpandCollapseState.Expanded;
+
+            // Raise automation event to annouce new state.
+            if (IsAccessibilityObjectCreated)
+            {
+                AccessibilityObject.RaiseAutomationPropertyChangedEvent(
+                    UiaCore.UIA.ExpandCollapseExpandCollapseStatePropertyId,
+                    oldValue: UiaCore.ExpandCollapseState.Collapsed,
+                    newValue: UiaCore.ExpandCollapseState.Expanded);
+            }
         }
 
         protected virtual void OnFormatChanged(EventArgs e)
@@ -1091,6 +1106,18 @@ namespace System.Windows.Forms
             if (Events[s_formatChangedEvent] is EventHandler eh)
             {
                 eh(this, e);
+            }
+        }
+
+        protected override void OnGotFocus(EventArgs e)
+        {
+            base.OnGotFocus(e);
+
+            // Raise automation event to annouce the control.
+            if (IsAccessibilityObjectCreated)
+            {
+                _expandCollapseState = UiaCore.ExpandCollapseState.Collapsed;
+                AccessibilityObject.RaiseAutomationEvent(UiaCore.UIA.AutomationFocusChangedEventId);
             }
         }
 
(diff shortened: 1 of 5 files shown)


RussKie (approved):

src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.cs
```diff
@@ -1132,6 +1159,16 @@ protected override void OnEnabledChanged(EventArgs e)
...
         {
             _onValueChanged?.Invoke(this, eventargs);
+
+            // Raise automation event to annouce changed value.
+            if (IsAccessibilityObjectCreated)
+            {
+                string name = AccessibilityObject.Name;
+                AccessibilityObject.RaiseAutomationPropertyChangedEvent(
+                    UiaCore.UIA.NamePropertyId,
+                    oldValue: name,
```
dreddy-work, line 1169: Q: what is changing here? I see both old and new values are same here.

DmitryGorokhov: A dtp should announce its name again due to date has been changed by user actions. If date is changed so name of dtp is changed too. But in this section I can't receive the previous name value, so I have to use current value twice. Anyway it doesn't matter because the Narrator pronounces actual AO name.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Control.ControlAccessibleObjectTests.cs
```diff
@@ -38,6 +38,14 @@ public class Control_ControlAccessibleObjectTests : IClassFixture<ThreadExceptio
                 typeof(WebBrowser),
         };
 
+        // These controls have special conditions for setting the Text property.
+        // Please check if the control type isn't contained here in cases where text change is needed
+        // otherwise an error can be thrown.
+        private static Type[] s_controlsIgnoringTextChangesForTests = new Type[]
+        {
+                typeof(DateTimePicker),
```
RussKie, line 46: [nit] Looks like a bad indent.

src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.cs
RussKie: It may worth adding this explanation as a comment.

RussKie (commented):
👍

src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.cs
DmitryGorokhov: Looks good. Added. Thanks

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Control.ControlAccessibleObjectTests.cs
DmitryGorokhov: Fixed.

Tanya-Solyanik:
@Olina-Zhang - could you please test this change?

DmitryGorokhov:
Fixed 2 found issues here. Sent it to retest.

Tanya-Solyanik (dismissed):

Amy-Li03:
Completed testing with no new issue found now.

src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.cs
```diff
@@ -1132,6 +1159,20 @@ protected override void OnEnabledChanged(EventArgs e)
...
+
+            // Raise automation event to annouce changed value.
+            if (IsAccessibilityObjectCreated)
+            {
+                // If date is changed so dtp name is changed too.
+                // But I can't receive the previous name value here,
+                // so I have to use current value twice.
+                // Anyway it doesn't matter because the Narrator pronounces actual AO name.
+                string name = AccessibilityObject.Name;
+                AccessibilityObject.RaiseAutomationPropertyChangedEvent(
```
vladimir-krestov, line 1171: If we take, that `AccessibilityObject.Name` does not reflect `Value`, do we need it at all?

vladimir-krestov (commented):
Looks good ✔️

src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.cs
DmitryGorokhov: Narrator works with Value property the same, so let's use Value instead of Name here. Thanks, reworked.

dreddy-work (dismissed):

dreddy-work:
@DmitryGorokhov , this needs rebase.

Tanya-Solyanik (approved):

Philip-Wang01:
Verified with .NET SDK 7.0.100-preview.3.22153.11 build, this issue is fixed.
image
image
image

PR #6626, opened 2022-02-05
Enable nullability in classes that inherit from NativeWindow

gpetrou:
## Proposed changes

- Enable nullability in classes that inherit from NativeWindow.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6626)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlNativeWindow.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlNativeWindow.cs
@@ -16,7 +16,7 @@ namespace System.Windows.Forms
 
             internal ControlNativeWindow(Control control)
             {
-                _control = control;
+                _control = control.OrThrowIfNull();
                 WindowTarget = this;
             }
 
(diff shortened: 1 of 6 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlNativeWindow.cs
```diff
@@ -74,12 +72,7 @@ public void OnMessage(ref Message m)
             // non-released controls will show what control wasn't released.
             public override string ToString()
             {
-                if (_control is not null)
-                {
-                    return _control.GetType().FullName;
-                }
-
-                return base.ToString();
+                return _control.GetType().FullName!;
```
dreddy-work, line 75: `_control `is not expected `null`, we should throw null exception in the constructor assignment if happened to be null

RussKie (approved):

dreddy-work (approved):

PR #6632, opened 2022-02-06
Enable nullability in SelectedIndexCollection and SelectedObjectCollection

gpetrou:
## Proposed changes

- Enable nullability in SelectedIndexCollection and SelectedObjectCollection.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6632)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListBox.SelectedObjectCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListBox.SelectedObjectCollection.cs
@@ -279,7 +279,7 @@ namespace System.Windows.Forms
                     ObjectCollection items = _owner.Items;
                     if (items is not null && value is not null)
                     {
-                        int index = items!.IndexOf(value!);
+                        int index = items.IndexOf(value);
                         if (index != -1 && GetSelected(index))
                         {
                             _owner.SetSelected(index, false);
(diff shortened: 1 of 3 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ListBox.SelectedObjectCollection.cs
```diff
@@ -283,7 +279,7 @@ public void Remove(object value)
                     ObjectCollection items = _owner.Items;
                     if (items is not null & value is not null)
```
RussKie, line 280: This looks like a typo, which is flagged by the analyzer with the NRTA
```suggestion
                    if (items is not null && value is not null)
```

RussKie (changes requested):

src/System.Windows.Forms/src/System/Windows/Forms/ListBox.SelectedObjectCollection.cs
```diff
@@ -281,9 +277,9 @@ public void Remove(object value)
                 if (_owner is not null)
                 {
                     ObjectCollection items = _owner.Items;
-                    if (items is not null & value is not null)
+                    if (items is not null && value is not null)
                     {
-                        int index = items.IndexOf(value);
+                        int index = items!.IndexOf(value!);
```
RussKie, line 282: The bangs aren't necessary now

PR #6633, opened 2022-02-06
Enable nullability in SnappableControl and CommandsPane

gpetrou:
## Proposed changes

- Enable nullability in SnappableControl and CommandsPane.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6633)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/CommandsPane.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/CommandsPane.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.ComponentModel;
 using System.ComponentModel.Design;
 using System.Drawing;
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/CommandsPane.cs
```diff
@@ -113,11 +111,11 @@ public override int GetOptimalHeight(int width)
         /// <inheritdoc />
         internal override bool SupportsUiaProviders => true;
 
-        private void LinkClicked(object sender, LinkLabelLinkClickedEventArgs e)
+        private void LinkClicked(object? sender, LinkLabelLinkClickedEventArgs e)
         {
             try
             {
-                if (!e.Link.Enabled)
+                if (!e.Link!.Enabled)
```
RussKie, line 118: This may result in a NRE, we can't do this.
```suggestion
                if (e.Link is null || !e.Link.Enabled)
```

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/CommandsPane.cs
```diff
@@ -207,7 +205,7 @@ private void InitializeLabelLinks()
         {
             Label.Links.Clear();
             StringBuilder sb = new();
-            Point[] links = new Point[_verbs.Length];
+            Point[] links = new Point[_verbs!.Length];
```
RussKie, line 208: `_verbs` is not guaranteed to be initialised here.

RussKie (changes requested):

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/CommandsPane.cs
gpetrou: If I understand correctly, in `SetVerbs` method we use a non-null value before calling this.

dreddy-work (approved):

PR #6641, opened 2022-02-08
Refactoring AutomationId at the accessible object classes

### LINKED ISSUE #6582: Move more properties from the `GetPropertyValue` method to the base class
Refactor properties `IsPasswordPropertyId`, `FrameworkIdPropertyId`, `HelpTextPropertyId` and `AccessKeyPropertyId` the same way as in PRs: #5745, #6429.

Original issue:
- #5494

Related issues:
- #6103
- #6236
- #6324

### LINKED ISSUE #6603: Refactoring `AutomationId` at the `GetPropertyValue` method
We should consider refactoring the `AutomationId` property at `AccessibleObject` class and its descendants. Now 7 descendants of the `ControlAccessibleObject` class return the `Owner.Name` value for the `AutomationId` property at the `GetPropertyValue` method, and 4 descendants of the `AccessibleObject` class return `AutomationId` (which is not virtual now as each class implements it on its own).
 
We could refactor this property like that: 
- move the AutomationId property to the AccessibleObject class (it will be returned at the GetPropertyValue method here)
- make it virtual with the default implementation returning a `null` value
- keep the current implementations in the 4 classes derived from AO,  but mark them as overrides 
- move returning of the `Owner.Name` to the `ControlAccessibleObjectObject` class itself to make it universal for all its descendants.

This could be risky since some customers could have automation tests that rely on `AutomationId` and have custom controls derived from our control class that don't implement custom `AutomationIdProperty`, but maybe code simplification outweighs the risk and at the worst-case scenario we could just revert the future PR with the single `AutomationId` refactoring.

Original post: https://github.com/dotnet/winforms/pull/6429#issuecomment-1009023639

ArtemTatarinov:
Fixes #6603

## Proposed changes
- The `AutomationId` property is moved to the `AccessibleObject` base class
- This property returns `null` at `AccessibleObject`
- It is returned at the `GetPropertyValue` method of `AccessibleObject` for the case `AutomationIdPropertyId`
- The current implementations in the 4 classes derived from AO are kept the same way, but they are marked as overrides
- Returning of the `Owner.Name` is moved to the `ControlAccessibleObjectObject` class to make it universal for all its descendants
- All duplicate code with the same logic is cleaned out in the derived classes
- No unit tests are added yet, but they will be after these changes will be discussed
- Initial discussion on this topic is here: https://github.com/dotnet/winforms/pull/6429#issuecomment-1009286397

## Customer Impact
- Tests based on the old implementation of AutomationId could become obsolete

## Regression? 
- No

## Risk
- Minimal

## Test methodology
- Unit tests
- CTI testing

## Test environment(s)
.NET SDK 7.0.0-preview-2.22103.2
Microsoft Windows [Version 10.0.19043.1466]

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6641)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlAccessibleObject.cs
@@ -34,6 +34,8 @@ namespace System.Windows.Forms
                 InitHandle(ownerControl);
             }
 
+            private protected override string AutomationId => Owner.Name;
+
             // If the control is used as an item of a ToolStrip via ToolStripControlHost,
             // its accessible object should provide info about the owning ToolStrip and items-siblings
             // to build a correct ToolStrip accessibility tree.
--- a/src/System.Windows.Forms/src/System/Windows/Forms/AccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/AccessibleObject.cs
@@ -391,6 +391,8 @@ namespace System.Windows.Forms
 
         internal virtual UiaCore.IRawElementProviderSimple? HostRawElementProvider => null;
 
+        private protected virtual string? AutomationId => null;
+
         /// <summary>
         ///  Returns the value of the specified <paramref name="propertyID"/> from the element.
         /// </summary>
@@ -428,6 +430,7 @@ namespace System.Windows.Forms
                 UiaCore.UIA.IsPasswordPropertyId => false,
                 UiaCore.UIA.FrameworkIdPropertyId => NativeMethods.WinFormFrameworkId,
                 UiaCore.UIA.AccessKeyPropertyId => KeyboardShortcut ?? string.Empty,
+                UiaCore.UIA.AutomationIdPropertyId => AutomationId,
                 _ => null
             };
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Button.ButtonAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Button.ButtonAccessibleObject.cs
@@ -17,8 +17,6 @@ namespace System.Windows.Forms
             internal override object? GetPropertyValue(UIA propertyID)
                 => propertyID switch
                 {
-                    UIA.AutomationIdPropertyId
-                        => Owner.Name,
                     UIA.ControlTypePropertyId
                         // If we don't set a default role for Button and ButtonBase accessible objects
                         // it will be retrieved from Windows.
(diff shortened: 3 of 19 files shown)


Tanya-Solyanik:
nit: @ArtemTatarinov - wasn't the intention to remove
```
                    case UiaCore.UIA.AutomationIdPropertyId:
                        return Owner.Name;
```

from ControlAccessibleObject and replace it with `AutomationId` property override?

ArtemTatarinov:
> nit: @ArtemTatarinov - wasn't the intention to remove
> 
> ```
>                     case UiaCore.UIA.AutomationIdPropertyId:
>                         return Owner.Name;
> ```
> 
> from ControlAccessibleObject and replace it with `AutomationId` property override?

@Tanya-Solyanik You're absolutely right, fixed it.

src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlAccessibleObject.cs
```diff
@@ -34,6 +34,8 @@ internal ControlAccessibleObject(Control ownerControl, int accObjId)
                 InitHandle(ownerControl);
             }
 
+            internal override string AutomationId => Owner.Name;
```
vladimir-krestov, line 37: AOs of all controls now will have this behavior, we have to make sure, that it's correct for the rest controls, that weren't implemented.

vladimir-krestov (commented):
Looks great! But we need to check if adding `AutomationId` for all controls is correct.

src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlAccessibleObject.cs
ArtemTatarinov: I've manually tested with the Inspect tool boths control with AutomationId implemented and not implemented and also checked their behavior before the fix. Looks like behavior isn't changing: if we were returning specific AutomationId, then it will still be returned after the fix. And if there were no AutomationId (returning `_ => null` at the `GetPropertyValue` method at the `AccessibleObject` class), then it won't be after the fix too, because our default implementation of the virtual `AutomationId` property returns null. But maybe it need some further investigation too.

Tanya-Solyanik:
@dkazennov  - what does inspect display when AccessibleObject returns `null` for `AutomationId`?

dkazennov:
> @dkazennov - what does inspect display when AccessibleObject returns `null` for `AutomationId`?
For the sake of testing I commented the override field AutomationId in the ListViewItemBaseAccessibleObject class and tested ListView control. If AutomationId is null then Inspect will display no AutomationId property at all. 
It shows the property and value if the field is overridden and is not null (a normal case).

As for this PR I've created all necessary unit tests, but I was being unable to push my changes to ArtemTatarinov/winforms fork. I've asked Vladimir for necessary rights.

RussKie:
You may need to push a the changes into your own fork and open a new PR.

PR #6643, opened 2022-02-08
Use RegexGenerator in WinForms

stephentoub:
The regex source generator is new in .NET 7.  It does all the parsing and compilation of the regex that would have been done at run-time instead at compile time, improving startup and enabling additional optimizations that wouldn't otherwise be performed.  This updates the regexes used in winforms to use the generator.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6643)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
@@ -4738,12 +4738,7 @@ namespace System.Windows.Forms
         private void SnapFocus(IntPtr otherHwnd)
         {
 #if DEBUG
-            if (s_snapFocusDebug.TraceVerbose)
-            {
-                string stackTrace = new StackTrace().ToString();
-                Regex regex = new Regex("FocusInternal");
-                Debug.WriteLine(!regex.IsMatch(stackTrace), "who is setting focus to us?");
-            }
+            Debug.WriteLineIf(s_snapFocusDebug.TraceVerbose, !Regex.IsMatch(Environment.StackTrace, "FocusInternal"), "who is setting focus to us?");
 #endif
             // we need to know who sent us focus so we know who to send it back to later.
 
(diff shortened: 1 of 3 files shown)


Tanya-Solyanik (approved):

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
```diff
@@ -4738,12 +4738,7 @@ internal void SnapMouseLocation()
         private void SnapFocus(IntPtr otherHwnd)
         {
 #if DEBUG
-            if (s_snapFocusDebug.TraceVerbose)
-            {
-                string stackTrace = new StackTrace().ToString();
-                Regex regex = new Regex("FocusInternal");
-                Debug.WriteLine(!regex.IsMatch(stackTrace), "who is setting focus to us?");
-            }
+            Debug.WriteLineIf(s_snapFocusDebug.TraceVerbose, !Regex.IsMatch(Environment.StackTrace, "FocusInternal"), "who is setting focus to us?");
```
joperezr, line 4741: NIT: kind of separate question but should this instead just be using an `IndexOf("FocusInternal")` not equals to -1?

stephentoub: Yeah, or Contains.  Didn't want to mess with it, though.  And it's debug-only.

joperezr (approved):
LGTM other than the small NIT I added.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
RussKie: I don't think the `if` should be collapsed. Previous we would only execute the regex if the trace switch was enabled, now we always execute it:
image

stephentoub: If it matters, it could be expanded back into an if or alternatively the argument put into an interpolated string, which would then bind to the overload employing conditional evaluation.

RussKie: I think going back to `if` would be the right move since we avoid running a regex and walking the stack constructing the stacktrace.

PR #6648, opened 2022-02-09
Merging internal commits for release/6.0

vseanreesermsft:


dreddy-work:
@vseanreesermsft , what is happenign here? why second PR with same commits? Is this auto generated?

dreddy-work:
Or because it was squash merged?

RussKie:
Aye, internal->public merges must not be squash-merged.

dreddy-work:
Given that earlier one squashed, how should we go about this? @vseanreesermsft / @RussKie ?

vseanreesermsft:
I don't see anything but automated dependency updates in either PR, I would close this one unless you see something I don't.
It was generated because the commits were squashed and then only because the automation was run a second time for 6.0.2xx.

dreddy-work:
> I don't see anything but automated dependency updates in either PR, I would close this one unless you see something I don't. It was generated because the commits were squashed and then only because the automation was run a second time for 6.0.2xx.

if automation was expecting the commit history to be available, do we expect similar issue when you run automation for next release cycle on the same branch then? @vseanreesermsft . What is the draw back to take this change?

vseanreesermsft:
@dreddy-work I think you're right. These commits would show up again the next time a merge was run. I've also learned that the dependency updates need to be preserved. We should revert the original commit and then merge this one.

dreddy-work (approved):

vseanreesermsft:
@dreddy-work Please make sure the commit history for this merge ends up in the repository. This PR got squashed as well.

dreddy-work:
> @dreddy-work Please make sure the commit history for this merge ends up in the repository. This PR got squashed as well.

Sorry about that. It was my mistake to set auto merge. Can you please create new one and I will make sure this doesn’t happen again.

PR #6658, opened 2022-02-09
Reworking conditional `Skip` attributes

### LINKED ISSUE #6653: Needs to add tests attributes with conditional skipping for UI tests
* .NET Core Version: 7.0

**Problem description:**
- Needs to add the ability to skip some tests in a specific architecture for UI tests and MAUI tests like we did it for unit tests (see [#6599](https://github.com/dotnet/winforms/pull/6599)). We can't use the attributes from unit tests utilities, because UI tests doesn't have a reference to them.
Based on this [comment](https://github.com/dotnet/winforms/pull/6599#issuecomment-1031969657)

**Solutions:**
- Copy-paste `ConditionalWinFormsFactAttribute` and `ConditionalWinFormsTheoryAttribute`
- Or add the dependency to UI tests project

**Will this feature affect UI controls?** No

vladimir-krestov:
Fixes #6653

## Proposed changes

- Reworked the conditional attributes to have an ability to skip tests in several architectures

## Customer Impact

- No

## Regression? 

- No

## Risk

- Minimal

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6658)


JeremyKuhne:
I would follow along with how the runtime repo does things.  https://github.com/dotnet/runtime/blob/main/docs/workflow/testing/libraries/filtering-tests.md https://github.com/dotnet/arcade/tree/main/src/Microsoft.DotNet.XUnitExtensions

So basically `[SkipOnArchitecture(...)]`. If we follow the same patterns we can better leverage what already exists and have the option of moving things like this to the shared assembly.

src/System.Windows.Forms/tests/TestUtilities/ConditionalWinFormsTheoryAttribute.cs
```diff
@@ -3,45 +3,42 @@
...
-                // Don't skip a test case if the skip condition is false.
-            }
+            set => _skipReason = value;
         }
 
         /// <summary>
         ///  Defines the architecture to skip the test on. This is complementary to <see cref="Skip"/> parameter.
         /// </summary>
-        public Architecture UnsupportedArchitecture
+        public Architecture[] UnsupportedArchitectures
```
kant2002, line 25: Can `Architecture UnsupportedArchitecture` property still be available, but just piggyback on the new `Architecture[] UnsupportedArchitectures`? That way this PR have less noise, and it would be much clearner to work when only one arch should be skipped?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewTests.cs
```diff
@@ -2446,8 +2439,7 @@ public void DataGridView_OnRowHeadersWidthChanged_Invoke_CallsRowHeadersWidthCha
         }
 
         [ActiveIssue("https://github.com/dotnet/winforms/issues/6597")]
-        [ConditionalWinFormsTheory(UnsupportedArchitecture = Architecture.Arm64,
-            Skip = "Flaky tests, see: https://github.com/dotnet/winforms/issues/6597")]
+        [WinFormsTheory(Skip = "Flaky tests, see: https://github.com/dotnet/winforms/issues/6597")]
```
Tanya-Solyanik, line 2442: This and some other tests were conditional on ARM64 previously, I assume that you removed condition on ARM since we can't distinguish ARM64 from AMD64, however I wonder if this test is flaky for x86 as well?

Tanya-Solyanik (commented):
Since we can't distinguish ARM64 from AMD64, perhaps it makes sense to add only x86 and x64 to this attribute?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewTests.cs
dmitrii-drobotov: I checked in Runfo and found just one test that is flaky in x86 too (https://runfo.azurewebsites.net/view/build/?number=1719382). For other tests in this file I'll set to skip on X64 because it indeed makes sense to run at least in x86.

dmitrii-drobotov:
@Tanya-Solyanik 
> Since we can't distinguish ARM64 from AMD64, perhaps it makes sense to add only x86 and x64 to this attribute?
Thanks for suggestion, I'll keep only x86 and x64.

Tanya-Solyanik (approved):
Looks good!

src/System.Windows.Forms/tests/TestUtilities/ConditionalWinFormsTheoryAttribute.cs
Tanya-Solyanik: Not applicable anymore.

src/System.Windows.Forms/tests/TestUtilities/SkipOnArchitectureAttribute.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace System
+{
+    /// <summary>
+    ///  Apply this attribute to your test method or class to skip it on a certain architecture.
+    /// </summary>
+    /// <remarks>
+    ///  This attribute isn't able to distinguish Arm64 from x64.
+    ///  To skip tests on Arm64 use <see cref="TestArchitectures.X64"/>.
+    ///  See https://github.com/dotnet/winforms/issues/7013.
+    /// </remarks>
```
RussKie, line 16: With this we're skipping tests on both x64 and arm64, which isn't very good. Our build pipelines are run on each platform independently, and it feels that we're overlooking this fact. That is, the CI sets the current arch (arm64), which should be discoverable by `SkipOnArchitectureDiscoverer`, which then should apply the filtering logic.

dmitrii-drobotov: Yes, this is indeed not an ideal solution. But the property `RuntimeInformation.ProcessArchitecture`, which is used in `SkipOnArchitectureDiscoverer`, returns `X64` when the tests are run for arm64 in CI. I couldn't find another way to determine the arm64 arch, so I decided to leave it this way. I created #7013 so that if it gets fixed, e.g. by running arm64 tests on an actual arm64 machine, we could fix this attribute as well.

PR #6667, opened 2022-02-10
Pin assembly versions to prevent revving those in servicing releases

### LINKED ISSUE #6663: 6.0.2 SDK update has broken WinForms apps
ℹ️  **UPDATE:** [Workaround here for those who need it.](https://github.com/dotnet/core/issues/7176)

. . .

There appears to be a critical problem with the new 6.0.2 update which makes apps compiled with it fail to start on machines running 6.0.1.

To reproduce, do the following:
1. Visit Windows Update to update the runtime on your **dev machine**, and also update Visual Studio 2022 using Visual Studio Installer.
2. On a **client/test machine**, install the 6.0.1 desktop runtime. [Direct Link](https://dotnetcli.azureedge.net/dotnet/WindowsDesktop/6.0.1/windowsdesktop-runtime-6.0.1-win-x64.exe)
3. On your dev machine, create a new WinForms project, build it and copy the executable files to the client/test machine.
4. Run it and it will not start. Event logs show the following:
```
Application: WinFormsApp1.exe
CoreCLR Version: 6.0.121.56705
.NET Version: 6.0.1
Description: The process was terminated due to an unhandled exception.
Exception Info: System.IO.FileLoadException: Could not load file or assembly 'System.Windows.Forms, Version=6.0.2.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'. The located assembly's manifest definition does not match the assembly reference. (0x80131040)
File name: 'System.Windows.Forms, Version=6.0.2.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'
   at WinFormsApp1.Program.Main()
```

This problem did not happen on .NET 5. I was under the impression patch versions are generally compatible with each other, and I use `<RollForward>LatestPatch</RollForward>` in my real app, but even with that setting it still crashes.

This is a blocking issue for me because if I distribute an update for my app built on 6.0.2 and the client is still running 6.0.1, that will be bad if the app suddenly stops working.

RussKie:
Resolves #6663

## Proposed changes

* Lock runtime and targeting pack assemblies at 6.0.2.0 so that any applications, libraries or NuGets built against 6.0.2 or later can work. 
This assembly versions pinning is similar to https://github.com/dotnet/runtime/blob/bbc766a7c65716305919bc691d696b1e46fa6f62/eng/Versions.props#L14 and https://github.com/dotnet/aspnetcore/blob/c0575788ecadd6e5cfeb4eab635c13b5fd433d37/Directory.Build.targets#L88.
* Add a validation step when we run packaging to verify that the assemblies we package into the transport package are correctly versioned, i.e., analyzers/generators are revved for every release for sxs, and those part of the ref pack are pinned to X.Y.2.0.

## Customer Impact

Currently Windows Forms applications built against 6.0.2 won't run on 6.0 GA or 6.0.1
With the fix applications, libraries or NuGets built against 6.0.2 or later can work. Any applications built on 6.0.2 or later that need to target 6.0.0 or 6.0.1 will need to use the workaround provided in https://github.com/dotnet/core/issues/7176. It is not ideal, but is believed to have a lesser impact than going back to 6.0.0.0 versions, and have 6.0.2 broken.

## Regression? 

- Yes 

## Risk

- Medium. With this fix we'll stop supporting the folks that compiled against 6.0.2 or later and target 6.0.0 or 6.0.1 without the workaround.
In general, since 6.0.1 and 6.0.2 are security releases, customers are encouraged to update to the latest version.

### Before

6.0.1
image
image

6.0.2
image
image

### After

A private built for this fix:
image
image
image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6667)

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -16,6 +16,34 @@
     <LangVersion>preview</LangVersion>
     <ImplicitUsings>enable</ImplicitUsings>
     <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
+
+    <IsAnalyzerProject>false</IsAnalyzerProject>
+    <IsAnalyzerProject Condition="!$(IsTestProject) and
+                                  ($(MSBuildProjectName.EndsWith('.Analyzers')) or $(MSBuildProjectName.EndsWith('.Analyzers.CSharp')))"
+                                  >true</IsAnalyzerProject>
+  </PropertyGroup>
+
+  <!--
+    Set assembly version to align with major and minor version, as for the patches and revisions should be manually
+    updated per assembly if it is serviced.
+
+    Note, any components that aren't exposed as references in the targeting pack (like analyzers/generators) those should rev
+    so that they can exist SxS, as the compiler relies on different version to change assembly version for caching purposes.
+
+    Because it is possible to build on a lower version and run on a higher version, we're locking the targeting pack and the runtime
+    assembly vesions to 6.0.2.
+    There's still a risk that some customers building on 6.0.2+ and trying to run on 6.0.0 and 6.0.1 will still have issue, but
+    those issues can be mitigated with a workaround described in https://github.com/dotnet/core/issues/7176.
+    -->
+  <PropertyGroup Condition="'$(IsAnalyzerProject)' != 'true'">
+    <AssemblyVersion>6.0.2.0</AssemblyVersion>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(IsAnalyzerProject)' == 'true'">
+    <IsRoslynComponent>true</IsRoslynComponent>
+
+    <!-- Mark it so eng/packageContent.targets knows how to correctly package it -->
+    <DefineConstants>WINFORMS_ANALYZERS</DefineConstants>
   </PropertyGroup>
 
   <!-- ApplicationConfiguration specific settings -->
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -1,11 +1,12 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <VersionPrefix>6.0.3</VersionPrefix>
+    <MajorVersion>6</MajorVersion>
+    <MinorVersion>0</MinorVersion>
+    <PatchVersion>3</PatchVersion>
     <!-- version in our package name #.#.#-below.#####.## -->
     <PreReleaseVersionLabel>servicing</PreReleaseVersionLabel>
-    <PreReleaseVersionIteration>
-    </PreReleaseVersionIteration>
+    <VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
     <!-- Use the compiler in the CLI instead of in the sdk, since the sdk one doesn't work with netcoreapp5.0 yet -->
     <UsingToolMicrosoftNetCompilers>false</UsingToolMicrosoftNetCompilers>
   </PropertyGroup>
--- a/pkg/Microsoft.Private.Winforms/Directory.Build.targets
+++ b/pkg/Microsoft.Private.Winforms/Directory.Build.targets
@@ -3,10 +3,12 @@
   <Import Project="..\..\Directory.Build.targets" />
 
   <PropertyGroup>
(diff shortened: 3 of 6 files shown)


eng/Versions.props
```diff
@@ -1,7 +1,13 @@
...
 <Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <VersionPrefix>6.0.3</VersionPrefix>
+    <MajorVersion>6</MajorVersion>
+    <MinorVersion>0</MinorVersion>
+    <PatchVersion>3</PatchVersion>
+    <VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
+    <!-- Set assembly version to align with major and minor version,
+         as for the patches and revisions should be manually updated per assembly if it is serviced. -->
+    <AssemblyVersion>$(MajorVersion).$(MinorVersion).0.0</AssemblyVersion>
```
ericstj, line 10: Note that this will be a downgrade from 6.0.2 and will make it so applications/libraries compiled against the 6.0.2 ref pack will fail to run on this version.

ericstj: Also, if you build any components that aren't exposed as references in the targeting pack those should rev so that they can exist sxs.  In particular analyzers/generators -- the compiler relies on different version to change assembly version for caching purposes. https://github.com/dotnet/runtime/pull/58836.   MSBuild tasks might have a similar requirement.

RussKie: Thank you.
I update the logic and added a validation step when we run packaging to verify that the assemblies we package into the transport package are correctly versioned.

RussKie:
Here's a test package https://dev.azure.com/dnceng/internal/_artifacts/feed/general-testing-internal/NuGet/Microsoft.Private.Winforms/6.0.3-servicing.22110.14/

RussKie:
I've spoken with @dsplaisted, and determined that we should lock at 6.0.2.0 so that any applications, libraries or NuGets built against 6.0.2 or later can work. Any applications built on 6.0.2 or later that need to target 6.0.0 or 6.0.1 will need to use the workaround provided in https://github.com/dotnet/core/issues/7176. It is not ideal, but is believed to have a lesser impact than going back to 6.0.0.0 versions, and have 6.0.2 broken.
In general, since 6.0.1 and 6.0.2 are security releases, customers are encouraged to update to the latest version.

Directory.Build.props
```diff
@@ -16,6 +16,34 @@
...
+    Note, any components that aren't exposed as references in the targeting pack (like analyzers/generators) those should rev
+    so that they can exist sxs. The compiler relies on different version to change assembly version for caching purposes.
+
+    Because it is possible to build on a lower version and run on a higher version, we're locking the targeting pack and the runtime
+    assembly vesions to 6.0.2.
+    There's still a risk that some customers building on 6.0.2+ and trying to run on 6.0.0 and 6.0.1 will still have issue, but
+    those issues can be mitigated with a workaround described in https://github.com/dotnet/core/issues/7176.
+    -->
+  <PropertyGroup Condition="'$(IsAnalyzerProject)' != 'true'">
+    <AssemblyVersion>$(MajorVersion).$(MinorVersion).2.0</AssemblyVersion>
```
dsplaisted, line 39: We only want to use the `.2` patch for 6.0... for 7.0 and onwards the ref assemblies should use the `.0` patch.

```suggestion
    <AssemblyVersion>$(MajorVersion).$(MinorVersion).0.0</AssemblyVersion>
    <AssemblyVersion Condition="'$(MajorVersion).$(MinorVersion)' == '6.0'">6.0.2.0</AssemblyVersion>
```

RussKie: This is a 6.0 branch. For the 7.0 branch it is locked at Major.Minor.0.0, see https://github.com/dotnet/winforms/pull/6668.
Is there a situation where you expect 6.0 spill into another branch?

dsplaisted: In our repos we merge changes forward.  If you don't do that it should be fine, but if that's the case it might as well just be hardcoded to `6.0.2.0` instead of `$(MajorVersion).$(MinorVersion).2.0`.

RussKie: We never merge forward, and always do dual check ins for the release branches.
I'll burn 6.0.2.0 in.

RussKie:
Approved over the email

PR #6669, opened 2022-02-10
Fixing text pattern of a multiline text box

### LINKED ISSUE #6661: [TextPattern] Getting incorrect rectangles for multiline TextBox
* .NET Core Version: 7.0

**Problem description:**
- TextPattern doesn't show empty lines (RTL and LTR modes):
image
It's important, when the caret is in the empty line, Narrator announces it incorrectly:
qzQ3n0qO2M

- In RTL mode some text ranges are not shown:
image

**Expected behavior:**
- UIA shows all text ranges correctly:
image

**Minimal repro:**
- Create a .NET 7.0 app
- Add a multiline TextBox to the app
- Run the app and use Narrator and Inspect to check this

vladimir-krestov:
Fixes #6661

## Proposed changes

- Rework `GetMultilineBoundingRectangles` method to:
     + calculate correct rectangles
     + make it simpler

## Customer Impact

- Now a user can see more correct text ranges in a muliline TextBox

## Regression? 

- No

## Risk

- Minimal

### Before

- TextPattern doesn't show empty lines;
- In RTL mode some text ranges are not shown:
image

### After

- UIA shows all text ranges correctly:
image

## Test methodology 

- Manual testing
- CTI
- Unit tests

## Accessibility testing  

- Using Narrator and Isnpect

## Test environment(s) 

- Windows 10
- .NET 7.0.0-preview.2.22108.17

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6669)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextProvider.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextProvider.cs
@@ -69,6 +69,8 @@ namespace System.Windows.Forms.Automation
 
         public abstract Point PointToScreen(Point pt);
 
+        public abstract Rectangle RectangleToScreen(Rectangle rect);
+
         public abstract void SetSelection(int start, int end);
 
         public static ES GetEditStyle(IHandle hWnd) => (ES)GetWindowLong(hWnd, GWL.STYLE);
--- a/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextRange.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextRange.cs
@@ -276,11 +276,9 @@ namespace System.Windows.Forms.Automation
 
         double[] ITextRangeProvider.GetBoundingRectangles()
         {
-            Rectangle ownerBounds = Drawing.Rectangle.Empty;
-
-            if (_enclosingElement.GetPropertyValue(UIA.BoundingRectanglePropertyId) is Rectangle boundsPropertyValue)
+            if (_enclosingElement.GetPropertyValue(UIA.BoundingRectanglePropertyId) is not Rectangle ownerBounds)
             {
-                ownerBounds = boundsPropertyValue;
+                return Array.Empty<double>();
             }
 
             // We accumulate rectangles onto a list.
@@ -597,9 +595,6 @@ namespace System.Windows.Forms.Automation
         /// </summary>
         private List<Rectangle> GetMultilineBoundingRectangles(string text, Point mapClientToScreen, Rectangle clippingRectangle)
         {
-            // Remember the line height.
-            int height = Math.Abs(_provider.Logfont.lfHeight);
-
             // Get the starting and ending lines for the range.
             int start = Start;
             int end = End;
@@ -623,31 +618,68 @@ namespace System.Windows.Forms.Automation
                 end = _provider.GetLineIndex(endLine) - 1;
             }
 
+            // Remember the line height.
+            int lineHeight = Math.Abs(_provider.Logfont.lfHeight);
+
             // Adding a rectangle for each line.
             List<Rectangle> rects = new List<Rectangle>();
-            int nextLineIndex = _provider.GetLineIndex(startLine);
 
-            for (int i = startLine; i <= endLine; i++)
+            for (int lineIndex = startLine; lineIndex <= endLine; lineIndex++)
             {
-                // Determine the starting coordinate on this line.
-                Point startPoint = _provider.GetPositionFromChar(i == startLine ? start : nextLineIndex);
+                // Get the left text position in the line.
+                int lineStartIndex = lineIndex == startLine ? start : _provider.GetLineIndex(lineIndex);
+                Point lineStartPoint = _provider.GetPositionFromChar(lineStartIndex);
 
-                // Determine the ending coordinate on this line.
-                Point endPoint;
+                // Get the right text position in the line.
+                int lineEndIndex = lineIndex == endLine
+                    // Just take the end of the range for the last line.
+                    // `end` is a caret position after the last character in the range,
+                    // so subtract 1 to get the last character index.
+                    ? end - 1
+                    // Or get the first index of the next line and take the previous character.
+                    // This is a workaround to get the last index of the line,
+                    // because Windows doesn't provide API for it.
+                    : _provider.GetLineIndex(lineIndex + 1) - 1;
+                Point lineEndPoint = _provider.GetPositionFromCharForUpperRightCorner(lineEndIndex, text);
 
-                if (i == endLine)
+                int lineTextLength = lineEndIndex - lineStartIndex + 1;
+
+                // If the line is long and takes up all the space,
+                // there is no space for the end of line rectangle. Adjust it.
+                if (lineEndPoint.X > clippingRectangle.Right
+                    // Or if the line is empty (just contains transition to the next line),
+                    // we need to offset the start point on end of line width (equals 2 px)
+                    // to show its rectangle in RTL mode. For LTR mode
+                    // `GetPositionFromCharForUpperRightCorner` do it for us.
+                    // Also, we have to check the line text length, because Windows may provide
+                    // incorrect endpoints for RTL TextBox, in this case we will catch an exception.
+                    || (_provider.IsReadingRTL && lineTextLength > 0
+                        && text.Substring(lineStartIndex, lineTextLength) == Environment.NewLine))
                 {
-                    endPoint = _provider.GetPositionFromCharForUpperRightCorner(end - 1, text);
-                }
-                else
-                {
-                    nextLineIndex = _provider.GetLineIndex(i + 1);
-                    endPoint = _provider.GetPositionFromChar(nextLineIndex - 1);
+                    lineStartPoint.X -= UiaTextProvider.EndOfLineWidth;
                 }
 
-                // Add a bounding rectangle for this line if it is nonempty.
-                // Add 2 to Y and 1 to Height to get a correct size of a rectangle around a range
-                Rectangle rect = new Rectangle(startPoint.X, startPoint.Y + 2, endPoint.X - startPoint.X, height + 1);
+                if (_provider.IsReadingRTL && lineEndPoint.X - lineStartPoint.X < 0)
+                {
+                    // TextBox in RTL mode may have incorrect character display order,
+                    // in this case the caret "jumps" between characters in the line
+                    // instead of direct moving through them. In this case Windows provides
+                    // incorrect characters indexes and their positions in the line.
+                    // This is a bug or the native control, because the caret "jums"
+                    // when selecting and the selected text is torn. Moreover, the caret position
+                    // is incorrect for some whitespaces, thereby Windows provides incorrect text range endpoints.
+                    // There are 2 ways to get rectangles:
+                    //    1) Go through all characters in the line and get min and max positions.
+                    //       (Low performance and will be useless when the native RTL TextBox is fixed)
+                    //    2) Take the owning control rectangle width for the line text range.
+                    //       (Fast and simple, won't be used when the native RTL TextBox is fixed)
+                    lineStartPoint.X = clippingRectangle.Left;
+                    lineEndPoint.X = clippingRectangle.Right;
+                }
+
+                // Add a bounding rectangle for a line, if it's nonempty.
+                // Increase Y by 2 to Y to get a correct size of the rectangle around a text range.
+                Rectangle rect = new(lineStartPoint.X, lineStartPoint.Y + 2, lineEndPoint.X - lineStartPoint.X, lineHeight);
                 rect.Intersect(clippingRectangle);
                 if (rect.Width > 0 && rect.Height > 0)
                 {
--- a/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextRange.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextRange.cs
@@ -659,26 +659,55 @@ namespace System.Windows.Forms.Automation
                     lineStartPoint.X -= UiaTextProvider.EndOfLineWidth;
                 }
 
-                if (_provider.IsReadingRTL && lineEndPoint.X - lineStartPoint.X < 0)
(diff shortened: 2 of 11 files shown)


src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextRange.cs
```diff
@@ -623,31 +618,55 @@ private List<Rectangle> GetMultilineBoundingRectangles(string text, Point mapCli
...
+                    // `end` is a caret position after the last character in the range,
+                    // so subtract 1 to get the last character index.
+                    lineEndIndex = end - 1;
                 }
                 else
                 {
-                    nextLineIndex = _provider.GetLineIndex(i + 1);
-                    endPoint = _provider.GetPositionFromChar(nextLineIndex - 1);
+                    // Get a supporting point in the right side to get an index of the last character in a line.
+                    // Add 1 to Y for correct GetCharIndexFromPosition method work.
```
RussKie, line 653:
```suggestion
                    // Get a supporting point on the right side to get an index of the last character in the line.
                    // Increase Y by one so that GetCharIndexFromPosition yields correct results.
```

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextRange.cs
```diff
@@ -623,31 +618,55 @@ private List<Rectangle> GetMultilineBoundingRectangles(string text, Point mapCli
...
+                    Point lineRightPoint = new(clippingRectangle.Right, lineStartPoint.Y + 1);
+                    lineEndIndex = _provider.GetCharIndexFromPosition(lineRightPoint);
+
+                    if (_provider.GetLineFromCharIndex(lineEndIndex) != lineIndex)
+                    {
+                        // If a text in a line is too long and its part is moved to the next line,
+                        // Windows returns the next line index, that is incorrect. Adjust it.
```
RussKie, line 660:
```suggestion
                        // If the text in a line is too long and wraps to the next line, Windows returns the next line index,
                        // which is incorrect, so we adjust it.
```

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextRange.cs
```diff
@@ -623,31 +618,55 @@ private List<Rectangle> GetMultilineBoundingRectangles(string text, Point mapCli
...
+                    }
                 }
 
-                // Add a bounding rectangle for this line if it is nonempty.
-                // Add 2 to Y and 1 to Height to get a correct size of a rectangle around a range
-                Rectangle rect = new Rectangle(startPoint.X, startPoint.Y + 2, endPoint.X - startPoint.X, height + 1);
+                Point lineEndPoint = _provider.GetPositionFromCharForUpperRightCorner(lineEndIndex, text);
+
+                // Add a bounding rectangle for a line, if it's nonempty.
+                // Add 2 to Y to get a correct size of a rectangle around a text range.
```
RussKie, line 668:
```suggestion
                // Increase Y by 2 to get a correct size of the rectangle around a text range.
```

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextRange.cs
```diff
@@ -623,31 +618,55 @@ private List<Rectangle> GetMultilineBoundingRectangles(string text, Point mapCli
...
+                Point lineStartPoint = _provider.GetPositionFromChar(lineStartIndex);
 
-                // Determine the ending coordinate on this line.
-                Point endPoint;
+                char ch = text[lineStartIndex];
+                if (_provider.IsReadingRTL && char.IsControl(ch))
+                {
+                    if (ch == '\r' || ch == '\n')
+                    {
+                        lineStartPoint.X -= UiaTextProvider.EndOfLineWidth; // Add 2 px to left to show the end of a line in RTL environment.
```
RussKie, line 637: The comment refers to "2px", which is not obvious from the code.

RussKie (commented):
👍

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextRange.cs
```diff
@@ -623,31 +618,55 @@ private List<Rectangle> GetMultilineBoundingRectangles(string text, Point mapCli
...
-                Point startPoint = _provider.GetPositionFromChar(i == startLine ? start : nextLineIndex);
+                int lineStartIndex = lineIndex == startLine ? start : _provider.GetLineIndex(lineIndex);
+                Point lineStartPoint = _provider.GetPositionFromChar(lineStartIndex);
 
-                // Determine the ending coordinate on this line.
-                Point endPoint;
+                char ch = text[lineStartIndex];
+                if (_provider.IsReadingRTL && char.IsControl(ch))
+                {
+                    if (ch == '\r' || ch == '\n')
```
Tanya-Solyanik, line 635: What happens if `\r` is not followed by `\n` on windows? Why are you not testing 2 characters? Why are you not using `Environment.NewLine`?

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextRange.cs
```diff
@@ -623,31 +618,55 @@ private List<Rectangle> GetMultilineBoundingRectangles(string text, Point mapCli
...
                 }
 
-                // Add a bounding rectangle for this line if it is nonempty.
-                // Add 2 to Y and 1 to Height to get a correct size of a rectangle around a range
-                Rectangle rect = new Rectangle(startPoint.X, startPoint.Y + 2, endPoint.X - startPoint.X, height + 1);
+                Point lineEndPoint = _provider.GetPositionFromCharForUpperRightCorner(lineEndIndex, text);
+
+                // Add a bounding rectangle for a line, if it's nonempty.
+                // Add 2 to Y to get a correct size of a rectangle around a text range.
+                Rectangle rect = new(lineStartPoint.X, lineStartPoint.Y + 2, lineEndPoint.X - lineStartPoint.X, lineHeight);
```
Tanya-Solyanik, line 669: Constant values seem to be pixels in this case, so they should be scaled for the DPI.

Tanya-Solyanik (commented):
Looks good. Please get this change tested.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextRange.cs
vladimir-krestov: Reworked on using `Environment.NewLine`. Thank you

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextRange.cs
vladimir-krestov: I checked several DPI modes, this solution works the same, because it moves a rectangle down on 2 px around a TextBox border. This border is 1 px in all DPI modes, so rectangle position is the same for them.
- For example (175% DPI): 
image

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextProvider.cs
```diff
@@ -126,6 +126,8 @@ public unsafe int SendKeyboardInputVK(short vk, bool press)
             return SendInput(1, ref keyboardInput, sizeof(INPUT));
         }
 
+        internal virtual int GetCharIndexFromPosition(Point pt) => default;
```
vladimir-krestov, line 129: To me: this change may be removed, because it's unnecessary now.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextRange.cs
Tanya-Solyanik: Please add a comment to this code that these pixels work well in HDPI

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextProvider.cs
Tanya-Solyanik: Please clarify, why did you add this line in the first place?

Tanya-Solyanik: I see, you never call it from the base provider class, Makes sense to define individual method on the derived Combo provider. Note that The text box base provider only redirects to the TextBoxBase, and is not actually invoked in this code.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextRange.cs
```diff
@@ -623,31 +618,97 @@ private List<Rectangle> GetMultilineBoundingRectangles(string text, Point mapCli
...
+                    {
+                        // TextBox in RTL mode may have incorrect character display order,
+                        // in this case the caret "jumps" between characters in the line
+                        // instead of direct moving through them. In this case Windows provides
+                        // incorrect characters indexes and their positions in the line.
+                        // This is a bug or the native control, because the caret "jums"
+                        // when selecting and the selected text is torn. Moreover, the caret position
+                        // is incorrect for some whitespaces, thereby Windows provides incorrect text range endpoints.
+                        // There are 2 ways to get rectangles:
+                        //    1) Go through all characters in the line and get min and max positions.
```
vladimir-krestov, line 680: The previous commit used the second way. This one uses the first - through all characters in an incorrect RTL line.
@Tanya-Solyanik, which way is better for incorrect Windows lines? (check the code above):
1) Slower, but more attractive:  
image

2) Simpler, faster, but looks like a bug:
image


PS. That's all because Windows returns incorrect values for RTL text box in LTR environment (CTI is checking RTL environment)
For example: look at the 3rd line `"and some   "`, a(111) X = 107, n(112) X = 45: 
image
Also, you can see something like that in RTL TextBox (not related to TextPattern):
- mouse selecting:
wN3o8O8Rmq

- keyboard navigation - ckick <- left arrow only (the caret is jumping):
Navigation

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextProvider.cs
vladimir-krestov: I would keep the previous implementation as is to concetrate on TextPattern

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextRange.cs
Tanya-Solyanik: >because Windows returns incorrect values for RTL text box in LTR environment

If this is confirmed, then we can ignore it as not a common case and go with the simpler solution

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TextBoxBase.TextBoxBaseUiaTextProviderTests.cs
```diff
@@ -658,8 +644,8 @@ public static IEnumerable<object[]> TextBoxBaseUiaTextProvider_GetVisibleRangePo
                 yield return new object[] { new Size(0, 0), 0, 0 };
                 yield return new object[] { new Size(0, 20), 0, 0 };
                 yield return new object[] { new Size(30, 30), 0, 5 };
-                yield return new object[] { new Size(50, 20), 0, 11 };
-                yield return new object[] { new Size(120, 20), 0, 26 };
+                yield return new object[] { new Size(50, 20), 0, 5 };
+                yield return new object[] { new Size(120, 20), 0, 19 };
```
dmitrii-drobotov, line 648: FYI: the old values were actually incorrect. This PR fixes `GetVisibleRangePoints` implementation and new values represent actual range. For 50x20 TextBox visible range is 5 symbols (`"Some "`):
Screenshot 2022-06-17 140842
And for 120x20 it should be 19 symbols (`"Some test text for "`):
Screenshot 2022-06-17 141042

src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.TextBoxBaseUiaTextProvider.cs
```diff
@@ -306,8 +305,22 @@ public override void GetVisibleRangePoints(out int visibleStart, out int visible
...
                 Point ptEnd = new Point(rectangle.Right - 1, rectangle.Bottom - 1);
 
-                visibleStart = _owningTextBoxBase.GetCharIndexFromPosition(ptStart);
-                visibleEnd = _owningTextBoxBase.GetCharIndexFromPosition(ptEnd) + 1; // Add 1 to get a caret position after received character
+                if (IsMultiline)
+                {
+                    visibleStart = GetLineIndex(FirstVisibleLine);
+
+                    int lastVisibleLine = FirstVisibleLine + LinesPerPage - 1;
+                    visibleEnd = GetLineIndex(lastVisibleLine + 1); // Index of the next line is the end caret position of the previous line.
```
Tanya-Solyanik, line 313: >// Index of the next line is the end caret position of the previous line. [](http://example.com/codeflow?start=68&length=73)

nit: In this repo we prefer comments to be placed on a dedicated line, above the line they apply to.

Tanya-Solyanik (dismissed):
@vladimir-krestov - had this fix changed significantly enough to justify CTI retesting it? If not, please merge!

dmitrii-drobotov:
> had this fix changed significantly enough to justify CTI retesting it? If not, please merge!

I think it's worth retesting because since the last test there was a fix to an issue reported by CTI — it would be good to confirm it. I'll send a request.

src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.TextBoxBaseUiaTextProvider.cs
dmitrii-drobotov: Fixed

Tanya-Solyanik (approved):
Thank you!

dmitrii-drobotov:
CTI approved the fix, and the previous issue is also fixed now.

vladimir-krestov (commented):
Approved ✅

PR #6675, opened 2022-02-10
Enable nullability in StatusStripAccessibleObject

gpetrou:
## Proposed changes

- Enable nullability in StatusStripAccessibleObject.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6675)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/StatusStrip.StatusStripAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/StatusStrip.StatusStripAccessibleObject.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using static Interop;
 
 namespace System.Windows.Forms
@@ -30,9 +28,9 @@ namespace System.Windows.Forms
                 }
             }
 
-            internal override UiaCore.IRawElementProviderFragment FragmentNavigate(UiaCore.NavigateDirection direction)
+            internal override UiaCore.IRawElementProviderFragment? FragmentNavigate(UiaCore.NavigateDirection direction)
             {
-                if (!(Owner is StatusStrip statusStrip) || statusStrip.Items.Count == 0)
+                if (Owner is not StatusStrip statusStrip || statusStrip.Items.Count == 0)
                 {
                     return base.FragmentNavigate(direction);
                 }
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/StatusStrip.StatusStripAccessibleObject.cs
```diff
@@ -30,21 +28,21 @@ public override AccessibleRole Role
...
                 switch (direction)
                 {
                     case UiaCore.NavigateDirection.FirstChild:
-                        AccessibleObject firstChild = null;
+                        AccessibleObject? firstChild;
                         for (int i = 0; i < GetChildCount(); i++)
                         {
                             firstChild = GetChild(i);
-                            if (firstChild is not null && !(firstChild is ControlAccessibleObject))
+                            if (firstChild is not null && firstChild is not ControlAccessibleObject)
```
dreddy-work, line 45: This condition is confusing me. How come child is not a `ControlAccessibleObject`? @vladimir-krestov ?

vladimir-krestov: `GetChild` always returns `ToolStripItem`, that is a component. Only one case, when `GetChild` can return `ControlAccessibleObject` is when this child is `ToolStripGrip`, that is a control.
So this logic means - exclude `ToolStripGrip` from the accessibility tree for `StatusStrip`. It looks right, because I didn't find any way, how to show a Grip for a `StatusStrip`.
So here should be checked the following:
```cs
if (firstChild is not null && firstChild is not ToolStripGrip.ToolStripGripAccessibleObject)
{
   return firstChild;
}
```
... and added a comment about it. But in this case we should make CTI testing to make sure, that I'm right.
The second way is to refactor `GetChild` method. But anyway, these suggestions are not for this PR, because it's about nullabiliti.
@dreddy-work, should we rework it as another PR?

vladimir-krestov (approved):
LGTM ✔️

src/System.Windows.Forms/src/System/Windows/Forms/StatusStrip.StatusStripAccessibleObject.cs
dreddy-work: Okay. Now i got it. Thank you for explanation. Workign on refactoring is low pri.

dreddy-work (approved):

PR #6684, opened 2022-02-12
Option to not use WebRequest for PictureBox

### LINKED ISSUE #1756: `PictureBox` to use `HttpClient` instead of using `WebRequest`
* .NET Core Version: 3.0 

**Problem description:**

`PictureBox` uses an instance of `WebRequest` to load images. 
It needs to be replaces with a cached instance of `HttpClient` as suggest in https://github.com/dotnet/winforms/pull/1696#discussion_r320565246

**Expected behavior:**

**Minimal repro:**

https://github.com/dotnet/winforms/blob/release/3.0/src/System.Windows.Forms/src/System/Windows/Forms/PictureBox.cs#L558-L564

kant2002:
Indirectly related to #1756 because if that would work, I would be able to fix #1756 using the same approach.
WebRequest bring ton of code to WinForms which is old, and not really interesting. NativeAOT suffer from this.

Inspired by https://github.com/dotnet/runtime/pull/38397

By itself if applying `<RuntimeHostConfigurationOption Include="System.Windows.Forms.PictureBox.UseWebRequest" Value="false" Trim="true" />` it will shave 1Mb from 20Mb  of my sample application which compiled under NativeAOT. I include all controls in that application.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6684)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/PictureBox.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/PictureBox.cs
@@ -597,8 +597,7 @@ namespace System.Windows.Forms
                             int progress = (int)(100 * (((float)_totalBytesRead) / ((float)_contentLength)));
                             if (_currentAsyncLoadOperation is not null)
                             {
-                                _currentAsyncLoadOperation.Post(_loadProgressDelegate,
-                                        new ProgressChangedEventArgs(progress, null));
+                                _currentAsyncLoadOperation.Post(_loadProgressDelegate, new ProgressChangedEventArgs(progress, null));
                             }
                         }
                     }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/PictureBox.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/PictureBox.cs
@@ -486,7 +486,7 @@ namespace System.Windows.Forms
                     }
                     else
                     {
-                        throw new NotSupportedException("Loading from remove location disabled.");
+                        throw new NotSupportedException(SR.PictureBoxRemoteLocationNotSupported);
                     }
                 }
             }
@@ -583,7 +583,7 @@ namespace System.Windows.Forms
                 }
                 else
                 {
-                    throw new NotSupportedException("Loading from remove location disabled.");
+                    throw new NotSupportedException(SR.PictureBoxRemoteLocationNotSupported);
                 }
             }
         }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/PictureBox.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/PictureBox.cs
@@ -606,8 +606,7 @@ namespace System.Windows.Forms
                         _tempDownloadStream.Seek(0, SeekOrigin.Begin);
                         if (_currentAsyncLoadOperation is not null)
                         {
-                            _currentAsyncLoadOperation.Post(_loadProgressDelegate,
-                                        new ProgressChangedEventArgs(100, null));
+                            _currentAsyncLoadOperation.Post(_loadProgressDelegate, new ProgressChangedEventArgs(100, null));
                         }
 
                         PostCompleted(null, false);
--- a/src/System.Windows.Forms/src/System/Windows/Forms/PictureBox.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/PictureBox.cs
@@ -579,7 +579,7 @@ namespace System.Windows.Forms
                 var uri = CalculateUri(_imageLocation);
                 if (uri.IsFile)
                 {
-                    Task.Run(() => LoadFromFileAsync());
+                    LoadFromFileAsync();
                 }
                 else
                 {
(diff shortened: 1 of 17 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/PictureBox.cs
```diff
@@ -549,7 +558,75 @@ public void LoadAsync()
...
+                    LoadFromFileAsync();
+                }
+                else
+                {
+                    throw new NotSupportedException("Loading from remove location disabled.");
+                }
+            }
+        }
+
+        private async void LoadFromFileAsync()
```
RussKie, line 579:
```suggestion
        private async Task LoadFromFileAsync()
```

src/System.Windows.Forms/src/System/Windows/Forms/PictureBox.cs
```diff
@@ -549,7 +558,75 @@ public void LoadAsync()
...
+                        _tempDownloadStream.Write(_readBuffer, 0, bytesRead);
 
+                        // Report progress thus far, but only if we know total length.
+                        if (_contentLength != -1)
+                        {
+                            int progress = (int)(100 * (((float)_totalBytesRead) / ((float)_contentLength)));
+                            if (_currentAsyncLoadOperation is not null)
+                            {
+                                _currentAsyncLoadOperation.Post(_loadProgressDelegate,
+                                        new ProgressChangedEventArgs(progress, null));
```
RussKie, line 601:
```suggestion
                                _currentAsyncLoadOperation.Post(_loadProgressDelegate, new ProgressChangedEventArgs(progress, null));
```

src/System.Windows.Forms/src/System/Windows/Forms/PictureBox.cs
```diff
@@ -549,7 +558,75 @@ public void LoadAsync()
...
+                            }
+                        }
+                    }
+                    else
+                    {
+                        _tempDownloadStream.Seek(0, SeekOrigin.Begin);
+                        if (_currentAsyncLoadOperation is not null)
+                        {
+                            _currentAsyncLoadOperation.Post(_loadProgressDelegate,
+                                        new ProgressChangedEventArgs(100, null));
```
RussKie, line 611:
```suggestion
                            _currentAsyncLoadOperation.Post(_loadProgressDelegate, new ProgressChangedEventArgs(100, null));
```

src/System.Windows.Forms/src/System/Windows/Forms/PictureBox.cs
kant2002: If I apply that change, then analyzer complain about this function at callsite. Whole body in `try/catch`. Do you think having `Task` is important enough?

kant2002:
If subsequently apply https://github.com/kant2002/winforms/commit/7ce82f2a4b307588b2c232c449617ba20c3ded5c then application trimmed to 15Mb and that with reflection

src/System.Windows.Forms/src/System/Windows/Forms/PictureBox.cs
kant2002: Should switch to `Task` and annotate code
or should I keep `async void`?

src/System.Windows.Forms/src/System/Windows/Forms/PictureBox.cs
```diff
@@ -466,26 +468,25 @@ public void Load()
...
+                    {
+                        _localImageStreamReader = new StreamReader(uri.LocalPath);
+                        Image img = Image.FromStream(_localImageStreamReader.BaseStream);
+                        InstallNewImage(img, ImageInstallationType.FromUrl);
+                    }
+                    else
                     {
-                        _uriImageStream = wc.OpenRead(uri.ToString());
-                        img = Image.FromStream(_uriImageStream);
+                        throw new NotSupportedException("Loading from remove location disabled.");
```
dreddy-work, line 489: should this string be localized?

src/System.Windows.Forms/src/System/Windows/Forms/PictureBox.cs
```diff
@@ -549,7 +570,73 @@ public void LoadAsync()
...
+                    }
+                    else
+                    {
+                        _tempDownloadStream.Seek(0, SeekOrigin.Begin);
+                        if (_currentAsyncLoadOperation is not null)
+                        {
+                            _currentAsyncLoadOperation.Post(_loadProgressDelegate, new ProgressChangedEventArgs(100, null));
+                        }
+
+                        PostCompleted(null, false);
```
dreddy-work, line 624: you do not need to post  `PostCompleted `here

kant2002:
@RussKie do you think, this PR require some additional work?

RussKie:
Apologies for the delay, I’m working on multiple streams and stretched pretty thin atm. I’ll try to get to this asapp.

src/System.Windows.Forms/src/System/Windows/Forms/PictureBox.cs
JeremyKuhne: @kant2002 I'd put the option inside of the existing `Task.Run` and not mix in `async` here. Better to stick with the existing patterns (which predate `async`).

JeremyKuhne (changes requested):
I'd stick with putting this in the existing `Task.Run` and avoid introducing `async`.

kant2002:
@JeremyKuhne did I infer your proposal correctly?

src/System.Windows.Forms/src/System/Windows/Forms/PictureBox.cs
```diff
@@ -549,7 +570,71 @@ public void LoadAsync()
...
+            if (UseWebRequest())
+            {
+                StartLoadViaWebRequest();
+            }
+            else
+            {
+                var uri = CalculateUri(_imageLocation);
+                if (uri.IsFile)
+                {
+                    Task.Run(() => LoadFromFileAsync());
```
JeremyKuhne, line 582: When you do this you're spinning two workers. What I was suggesting was that you remove `async` from the helper and not use `ReadAsync()`. Something like `Task.Run(() => LoadFromFileWithProgress())` or have `void LoadFromFileAsync()` with a `Task.Run` in it.

kant2002: > or have void LoadFromFileAsync() with a Task.Run in it.

@JeremyKuhne Hopefully I get it right this time.

src/System.Windows.Forms/src/System/Windows/Forms/PictureBox.cs
```diff
@@ -549,7 +570,74 @@ public void LoadAsync()
...
+                        {
+                            _tempDownloadStream.Seek(0, SeekOrigin.Begin);
+                            if (_currentAsyncLoadOperation is not null)
+                            {
+                                _currentAsyncLoadOperation.Post(_loadProgressDelegate, new ProgressChangedEventArgs(100, null));
+                            }
+                        }
+                    }
+                    while (bytesRead > 0);
+                    PostCompleted(null, false);
```
JeremyKuhne, line 628: nit: when passing `true` or `false` it is better to explicitly name the parameter for clarity. You should also consider naming the parameter when passing `null`.

src/System.Windows.Forms/src/System/Windows/Forms/PictureBox.cs
```diff
@@ -549,7 +570,74 @@ public void LoadAsync()
...
+                    LoadFromFileAsync();
+                }
+                else
+                {
+                    throw new NotSupportedException(SR.PictureBoxRemoteLocationNotSupported);
+                }
+            }
+        }
 
+        private void LoadFromFileAsync()
```
JeremyKuhne, line 591: Looking at this again I wonder why we don't just use the existing `ReadCallback()`? It would be clearer to reutilize the existing code instead of duplicating it. Then you'd also get cancellation support, which you haven't currently implemented.

``` C#
private void LoadFromFileAsync()
{
    // Shouldn't need to check CancellationPendingState here if all we're doing is Task.Run(() => LoadFromFileAsync())
    FileStream stream = null;

    try
    {
        stream = File.OpenRead(_imageLocation);
        _contentLength = stream.Length;
        _totalBytesRead = 0;

        stream.BeginRead(
            _readBuffer,
            0,
            ReadBlockSize,
            new AsyncCallback(ReadCallBack),
            responseStream);
    }
    catch (Exception error)
    {
        PostCompleted(error, cancelled: false);
    }
    finally
    {
         stream?.Dispose();
    }
}
```

JeremyKuhne (changes requested):
Spending some more time looking at this I think you'd be better off using the existing `ReadCallBack()`. Let me know if that doesn't work for you or if you have any other questions.

JeremyKuhne:
@kant2002 appreciate your effort here.

kant2002:
@JeremyKuhne reason why I did use `async/await` in "new" implementation is because I think they are easier to reason about. And  seems here appetite eventually move to HttpClient so `async/await` is would be natural choice. Since this is new code path, "some regressions" maybe acceptable for target audience.

Anyway, I update as you suggest the code. I do not have questions and this approach is fine.

JeremyKuhne:
> I think they are easier to reason about.

Normally, I'd agree. As this code predates `async` you introduce a fair amount of cognitive overhead if you mix and match approaches. While the code could be completely rewritten to leverage new patterns I'd hesitate to do so without a strong justification.

> Since this is new code path, "some regressions" maybe acceptable for target audience.

I think the default position should be "no differences in functionality". We can, of course, alter that based on data and discussion.

Again, thanks for your effort here.

JeremyKuhne (approved):
Looks good!

PR #6685, opened 2022-02-12
Autocomplete using ComWrappers

kant2002:
Provide RCW for
- [x] IAutoComplete2

Provides CCW
- [x] IEnumString

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6685)


src/System.Windows.Forms/src/System/Windows/Forms/StringSource.cs
```diff
@@ -20,41 +18,29 @@ internal class StringSource : IEnumString
...
-        /// </summary>
-        private static Guid autoCompleteClsid = new Guid("{00BB2763-6A77-11D0-A535-00C04FD7D062}");
+        private WinFormsComWrappers.AutoCompleteWrapper? _autoCompleteObject2;
 
         /// <summary>
         ///  Constructor.
         /// </summary>
         public StringSource(string[] strings)
         {
-            Array.Clear(strings, 0, size);
```
kant2002, line 35: Based on the usage this will never ever happens.

src/System.Windows.Forms/src/System/Windows/Forms/StringSource.cs
```diff
@@ -82,20 +68,14 @@ public void ReleaseAutoComplete()
...
             {
-                Marshal.ReleaseComObject(_autoCompleteObject2);
+                _autoCompleteObject2.Dispose();
                 _autoCompleteObject2 = null;
             }
         }
 
         public void RefreshList(string[] newSource)
         {
-            Array.Clear(strings, 0, size);
```
kant2002, line 92: Also based on the usage this will never happens.

kant2002: Or better say, this is no-op.

src/System.Windows.Forms/src/System/Windows/Forms/StringSource.cs
kant2002: Or better say, this is no-op.

src/System.Windows.Forms/src/System/Windows/Forms/StringSource.cs
```diff
@@ -82,20 +68,14 @@ public void ReleaseAutoComplete()
...
-
-            if (strings is not null)
-            {
-                strings = newSource;
-            }
-
+            strings = newSource;
             current = 0;
             size = (strings is null) ? 0 : strings.Length;
```
dreddy-work, line 80: Is this still needed then?

kant2002: Line 78 - yes. But Line 80 definitely can be simplified, since `strings` never null.

RussKie (approved):

kant2002:
Does this require additional review?

dreddy-work:
> Does this require additional review?

Yes. @JeremyKuhne may want to look and merge. give it a day or two more.

src/System.Windows.Forms/src/System/Windows/Forms/StringSource.cs
JeremyKuhne: While it is true that there are no callers that pass in a null array, not clearing the passed in array is a change in behavior. I wouldn't remove the clear in this change.

kant2002: Done.

dreddy-work (approved):

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.AutoCompleteWrapper.cs
```diff
@@ -0,0 +1,44 @@
...
+
+internal partial class Interop
+{
+    internal unsafe partial class WinFormsComWrappers
+    {
+        internal class AutoCompleteWrapper
+        {
+            private IntPtr _wrappedInstance;
+
+            public AutoCompleteWrapper(IntPtr wrappedInstance)
```
willibrandon, line 17: IAutoComplete2 has methods _Init_, _Enable_, _SetOptions_, and _GetOptions_. Here for AutoCompleteWrapper I only see _Init_ and _SetOptions_. Is that intentional?

kant2002: Yes. This is intentional. There no need for declaring methods in RCW if they would not be used in the application. WinForms use just 2 methods, so I leave only these 2 method. That mentality was shown for me in the dotnet/runtime repo, and based on my experience working with WinForms team, same applied here as well. Additional cost (albeit very small) is paid for all WinForms consumers.

willibrandon: That is good information to know for ComWrappers, I will keep this in mind. Thank you for the insight.

Issue #5163, opened 2021-06-28
ComWrappers support


kant2002:
I want to raise questions about plans for adding ComWrappers support in WinForms.
This make WinForms application IL trimming friendly. For me that would be big win for NativeAOT, but I understand that this is not priority.
Work in .NET repository started https://github.com/dotnet/runtime/pull/54636 and I think similar approach to some degree can happens here.

I initially try to find simplest interfaces where this can be tested, but everything so intermixed with ActiveX support, so I do not know where to start.
So far easiest targets to get started is following
- IPicture, IStream, IPersistStream.
- IClassFactory2
- IMallocSpy

Each these interfaces has their own challenges
# IPicture, IStream, IPersistStream
Initially can be implemented for Cursor class only. IPicture used a lot in ActiveX code and that part is extremely problematic and maybe require usage of IDynamicCastable or similar complicated approach. Also probably it will require `OleCreatePictureIndirect` split of implementation on two parts, one for ComWrappers, second is for ComWrappers, and that would be for some time, since ActiveX seems impenetrable target for now.

# IClassFactory2
This is very limited in usage and can be used for probing ground how to start make ActiveX support friendly for ComWrappers. Interface in itself not a problem, but `CreateInstanceLic` return object instance which as far as I understand, should be universal RCW. I personally may go with temporary using existing COM infrastructure, using `Marshal.GetObjectForIUnknown` until better solution arise. 

# IMallocSpy
This interface used only for tests, so not a big win in terms of implementation. Maybe good candidate to prepare tests for world without no built-in COM.

# Testing
As mentioned in PR for System.Drawing.Common, they have to support .NET Core 3.1 and this requires a bit of changes in testing. I'm not sure how applicable this concerns for this repo, and how do you manage older .NET Core. Require discussion too.

Also a lot of tests rely on the built-in COM infrastructure so this is also should be factored somehow.

dreddy-work:
@JeremyKuhne would be right person on winforms team to take this discussion further.

JeremyKuhne:
We do want/intend to move to function pointers for COM, it is just a matter of resource allocation. If folks are interested in contributing to make this happen we'd welcome it. Tests will be important and depending on risk and timing some changes might have to go into .NET 7.

kant2002:
@JeremyKuhne Yes. I potentially willing to contribute in that area. 
[...]

PR #6692, opened 2022-02-15
Explicitly use ComWrappers for file dialogs

kant2002:
Do not use COM interfaces for file dialogs at all except IFileDialogEvents
Right now each CoClass created using CoCreateInstance represented as unique RCW
That RCW still uses interfaces, but all important methods are implicitly declares
No explicit interface method of IFileDialog, IOpenFileDialog and ISaveFileDialog
are used.
IShellItem and IShellItemArray still used, but that usage comes from IXXXFileDialog
interfaces. If I remove them, I can be explicit about wrappers.

After this point, there two options
- Leave interfaces as is, wasteful RCW, good for documentation
- Remove interfaces, simplify RCW even more. Keeping track of method table would be problematic.

Related to #5163

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6692)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.FileDialogWrapper.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.FileDialogWrapper.cs
@@ -26,7 +26,7 @@ internal partial class Interop
                 _wrappedInstance = IntPtr.Zero;
             }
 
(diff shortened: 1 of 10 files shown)


RussKie (approved):

kant2002:
Does this also require additional review?

kant2002:
Can somebody merge this?

JeremyKuhne (approved):

PR #6723, opened 2022-02-20
Explain localization process

kant2002:
Steps for localization process not clearly explained in the documentation

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -62,6 +62,15 @@ Note that this does **not** build using your machine-wide installed version of t
 
 To create the Microsoft.Private.Winforms package, run `.\build -pack`
 
+## Localization
+
+If you need create localization string do the following steps
+
+- Modify `Resource\SR.resx` file by adding nescessary strings
+- Update localization files by running command `.dotnet/dotnet build /t:UpdateXlf src/System.Windows.Forms/src/`
+
+In case you want to redo localization, and nothing seems to be not updated you may want to cleanup temporary files. Please delete `artifacts/obj/System.Windows.Forms` folder.
+
 [comment]: <> (URI Links)
 
 [corefx-windows-instructions]: https://github.com/dotnet/corefx/blob/master/Documentation/building/windows-instructions.md


docs/building.md
```diff
@@ -62,6 +62,15 @@ Note that this does **not** build using your machine-wide installed version of t
...
 To create the Microsoft.Private.Winforms package, run `.\build -pack`
 
+## Localization
+
+If you need create localization string do the following steps
+
+- Modify `Resource\SR.resx` file by adding nescessary strings
+- Update localization files by running command `.dotnet/dotnet build /t:UpdateXlf src/System.Windows.Forms/src/`
+
+In case you want to redo localization, and nothing seems to be not updated you may want to cleanup temporary files. Please delete `artifacts/obj/System.Windows.Forms` folder.
```
dreddy-work, line 72:
```suggestion
In case you want to redo localization, and nothing seems to be not updated you may want to remove artifacts and rebuild with `/t:UpdateXlf`. Use `git -clean` to remove artifacts.
```

docs/building.md
```diff
@@ -62,6 +62,15 @@ Note that this does **not** build using your machine-wide installed version of t
 
 To create the Microsoft.Private.Winforms package, run `.\build -pack`
 
+## Localization
+
+If you need create localization string do the following steps
```
dreddy-work, line 67:
```suggestion
If you need to create localization string do the following in the corresponding namespace/binary.  
```

docs/building.md
```diff
@@ -62,6 +62,15 @@ Note that this does **not** build using your machine-wide installed version of t
...
+
+- Modify `Resource\SR.resx` file by adding nescessary strings
+- Update localization files by running command `.dotnet/dotnet build /t:UpdateXlf src/System.Windows.Forms/src/`
+
+In case you want to redo localization, and nothing seems to be not updated you may want to cleanup temporary files. Please delete `artifacts/obj/System.Windows.Forms` folder.
+
```
RussKie, line 73:
```suggestion
## Localization

If you need to create a new or updated an existing localization string, follow these steps:

- Modify `Resource\SR.resx` file adding or updating necessary strings in the project that contains the said strings. (It is often faster/easier to open SR.resx in an XML editor).
- Regenerate the localization files by rebuilding the solution/project from Visual Studio, or by executing `.\build.cmd` command. You can also build just the modified project by running  `dotnet build` from the project's root.

```

docs/building.md
RussKie: To be honest I find this suggestion to a degree harmful, because it will wipe .dotnet folder along with other local changes you may have. If artifacts folder needs to be cleaned, then `git clean -xdf` should be run under `.\artifacts` folder.

I haven't seen resources not updating correctly on build, and would be interested in learning how you get there.

One common mistake I've seen people do, is that they use the same repo for building different versions (e.g., .NET 6 and .NET 7). This leads to various unexplainable errors, random failures, incorrect package resolutions and other red herrings. This is not Windows Forms repo specific issue, the spans even beyond dotnet/* repos. To avoid this, one either needs to have a separate clone for each version, or use git worktrees.

kant2002: 1. I prepare PR without localization
2. I edit `SR.resx` on the github.dev since I assume that touching `SR.locale.resx` is not my responsibility.
3. Then I start looking at the error message and try to find information on the process.

I do know that my suggestion is hackish and maybe not willing to be supported, but they seems to be simple and reasonable so I decide to give it a try. I'm not big fan of `git clean -xdf` since this is last measure for me.

RussKie: > * I edit `SR.resx` on the github.dev...

Can you run a build on github.dev though? If not, then this scenario is not supported.


> I'm not big fan of `git clean -xdf` since this is last measure for me.

Same here, if I run it - I run it in specific folders (like .\\src or .\\pkg).

kant2002: > If not, then this scenario is not supported.

That's fine. But github.dev is so convinient and "working" on dotnet/runtime, so I assume it will work here.

docs/building.md
```diff
@@ -62,6 +62,13 @@ Note that this does **not** build using your machine-wide installed version of t
 
 To create the Microsoft.Private.Winforms package, run `.\build -pack`
 
+## Localization
+
+If you need to create a new or updated an existing localization string, follow these steps:
```
RussKie, line 67: My grammar was clearly asleep at the wheel...
```suggestion
If you need to add a new localization string or update an existing one, follow these steps:
```

RussKie (approved):

docs/building.md
RussKie: "Windows" in the name sets the limits :)

RussKie:
@msftbot merge if @dreddy-work  approves

kant2002:
@dreddy-work May I ping you on this PR?

dreddy-work (approved):

Issue #4649, opened 2021-03-05
Determine linker compatibility for .NET 6


agocke:
Right now winforms produces many linker warnings (some of them in the core framework, which will be handled separately). To provide linker compatibility, these warnings should be reviewed.

Unfortunately, there are too many warnings to list directly, so here's an attached file:

[warnings.txt](https://github.com/dotnet/winforms/files/6093559/warnings.txt)

RussKie:
Useful docs:
* https://github.com/mono/linker/blob/main/docs/design/reflection-flow.md#linker-friendly-reflection
* https://devblogs.microsoft.com/dotnet/app-trimming-in-net-5/

weltkante:
I find no mention of COM interop, the warnings about COM interop not being supported are worrying (`IL2050: xxx declares a parameter with COM marshalling. Correctness of COM interop cannot be guaranteed after trimming. Interfaces and interface members might be removed.`) since this is fundamental functionality to communicate with Windows (WinForms without clipboard, drag'n'drop, RichEdit, accessibility etc. would not be the same)

Furthermore the text mentions trimming of COM interfaces yet only reports P/Invoke signatures. That has me worried that ComImport interfaces also get broken, trimming them is certainly not allowed (changes method order which is fatal for IUnknown based interfaces)

merriemcgaw:
Adding this to our dotnet/planning/issues/27 theme

kant2002:
I did try trim recent version of WinForms and what are the log which I receive 640 rows.
[with_design.txt](https://github.com/dotnet/winforms/files/8407199/with_design.txt)

Most issues which I see is from Design namespace which for most application does not make sense. Also I do not think that in practice that would make sense to target that namespace without additional work on Designer which happens in private.

So I remove from log these warnings and receive smaller file of 260 rows.
[without_design.txt](https://github.com/dotnet/winforms/files/8407219/without_design.txt)

I look again, and next target for questionable warnings would be Converters which is mostly used during design-time. Also `DefaultValueAttribute` somewhat irrelevant for "trimming-app".

So I would like to ask questions
- How that Design namespace issues can be hidden from trim warnings?
- Can whole issue be splitted in something more manageable. For example I see "design" "typeconverters" "binding", "resources" areas which can be worked separately. Please correct me if I wrong here. Since this issue looks so big, nobody want to bother, but if split issue on something smaller, it would be easier to handle.
[...]

PR #4685, opened 2021-03-14
Fix inversion of conditional

weltkante:
Contributes to #4370

## Proposed changes

This regression to Desktop Framework was found while investigating #4370. It is unclear what the effects of this regression are, but [comparing](https://referencesource.microsoft.com/#System.Windows.Forms/winforms/Managed/System/WinForms/Control.cs,15700) with Desktop Framework it seems its a mistake from refactoring

When running a debug build of WinForms during investigation of #4370 I had turned on `CompModSwitches.ActiveX` and the broken code resulted in warnings of the ActiveX host passing wrong flags (which of course is wrong, using MFC or Office as host you can assume they pass the right flags).

## Customer Impact

- unknown, probably changing behavior of hosted controls due to wrong flags being passed to host

## Regression? 

- yes

## Risk

- low, restoring code to how it was before. There was another recent regression fix for #4370 so this should be fine to accompany the other fix

### Before

- running the sample from #4370 (or generally, hosting a WinForms control in a MFC app or other ActiveX host) outputs warnings when using a WinForms debug build with ActiveX tracing turned on

### After

- no warnings in debug build

## Test methodology

- source inspection
- running the sample from #4370 with a debug build of WinForms, making sure the previously observed warning is gone

## Test environment(s) 

- MFC application (stock app from wizard) initializing the WinForms control from the sample code
- Outlook 16.0.13801.20240 (32 bit) hosting the WinForms control from the sample code

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/4685)

PR #6743, opened 2022-02-23
Use ComWrappers for IErrorInfo

kant2002:
Progress towards #5163 and #4649 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6743)


RussKie:
The change looks sensible. Can you please elaborate how you tested it? Also can you please at least manual test to WinformsControlsTest project.

kant2002:
I add small window which display COM Object which supports `ISuportsErrorInfo` but unfortunately I cannot find CLSID for an object which has properties which trigger `IErrorInfo` code path. I manage to test only `GetErrorInfo() == S_FALSE` path.

kant2002:
@RussKie this is hard time, and if you have spare cycles, can you take a look?

RussKie:
> I add small window which display COM Object which supports `ISuportsErrorInfo`

Are you able to add your test harness to the solution? Either as an extension to one of the existing integration projects (e.g., under IntegrationTests or InteropTests) or add a new stand-alone project.

src/System.Windows.Forms.Primitives/src/Interop/OleAut32/Interop.GetErrorInfo.cs
```diff
@@ -9,6 +9,16 @@ internal partial class Interop
     internal static partial class Oleaut32
     {
         [DllImport(Libraries.Oleaut32, ExactSpelling = true)]
-        public static extern HRESULT GetErrorInfo(uint dwReserved, out IErrorInfo pperrinfo);
+        public static extern HRESULT GetErrorInfo(uint dwReserved, out IntPtr pperrinfo);
```
RussKie, line 12: Is this used outside this file? If not, please make it private.

RussKie:
>  I manage to test only GetErrorInfo() == S_FALSE path.

Can we make a custom COM object that returns S_OK? For example, in https://github.com/dotnet/winforms/pull/4685 @weltkante added a native project that loads the managed code. Maybe we can do something like this too.

RussKie:
We (the team) have a extremely limited test capabilities and limited understanding of real scenarios in which COM is used in conjunction with Windows Forms apps. This makes it very difficult to accept changes in this area without good and verifiable test coverage.

kant2002:
Tests are in place. I have to only fix infra issues. Creation of IDispatch COM objects rely on TLB file which produced by MIDL and this is failing right now. It works for me locally in the Developer command prompt, but not in regular PowerShell for example. If somebody have an idea, I would like to know what should I do.

weltkante:
> It works for me locally in the Developer command prompt, but not in regular PowerShell for example. If somebody have an idea, I would like to know what should I do.

Developer command prompt is setting additional search paths & environment variables, but the powershell scripts are using the arcade build infrastructure, so needs someone who knows how to configure C++ projects for that I guess.

kant2002:
I the dotnet/runtime this is done like that https://github.com/dotnet/runtime/blob/57bfe474518ab5b7cfe6bf7424a79ce3af9d6657/eng/native/init-vs-env.cmd so maybe I'll start with this approach.

kant2002:
Progress:
1. building locally from Powershell
2. Release build issues on CI, no sure what it is.
3. Debug issues with running tests. I have this one locally, but think that I solve them by making test fixture to run sequentially. Seems to be I done that wrong. 

Would gladly accept suggestions how to run specific test fixture sequentially. Or how I can run Reg-free COM in XUint that way tests would be more realistic.

weltkante:
> Or how I can run Reg-free COM in XUint that way tests would be more realistic.

In my experience you usually have the manifest for regfree COM alongside the exe, which with unit tests is a bit problematic as it will be a test runner not under your control.

You could start a separate process for the test, where you put the manifest along that processes exe

Or you could try to use [activation contexts](https://docs.microsoft.com/en-us/windows/win32/sbscs/activation-contexts) to load a manifest for a library loaded at runtime. I've never used activation contexts so I can't guarantee that it works for .NET libraries this way, would need some research/experimentation to see if that route is viable.

kant2002:
Activation contexts scares me. Even if see snippets for .Net I do not fully understand them. If it would be acceptable just run `regsvr32` I probably would be happy

weltkante:
> Activation contexts scares me. Even if see snippets for .Net I do not fully understand them.

I did a quick test and it works seamlessly by [doing a few interop calls](https://gist.github.com/weltkante/dd0ed188ab3bb0ba3c0f96a854959818) to load the manifest manually

RussKie:
> I the dotnet/runtime this is done like that [dotnet/runtime@`57bfe47`/eng/native/init-vs-env.cmd](https://github.com/dotnet/runtime/blob/57bfe474518ab5b7cfe6bf7424a79ce3af9d6657/eng/native/init-vs-env.cmd) so maybe I'll start with this approach.

I'm sure the runtime team won't mind if we borrowed the script. We can invoke it from start-vs.cmd.

RussKie:
I'm phoning a friend - @AaronRobinsonMSFT :) May be he can help...

RussKie:
> If it would be acceptable just run `regsvr32` I probably would be happy

I'm not sure other contributors would be happy if we'd reg'ed some random COM object in their system, nor if we'd be able to do this on CI boxes either. @MattGal any thoughts?

AaronRobinsonMSFT:
@RussKie Thanks for adding me to the conversation. Per usual, @weltkante is providing solid guidance here.

>If it would be acceptable just run regsvr32 I probably would be happy

Yep, I get that, but that is not something any test should be doing. The interop team took great pains to ensure all COM related testing avoided global registration. The RegFree/ActivationCxt approach is recommended for these cases.

> Activation contexts scares me. Even if see snippets for .Net I do not fully understand them.

Agreed. It is a convoluted system and has corner cases that have kept me, and the rest of the Interop team stuck debugging weird issues at times. However, we have tamed the beast for .NET Core testing and have a well-worn path in dotnet/runtime. See [here](https://github.com/dotnet/runtime/blob/466b0f66aafe63a31696c16dca43a727cd4bbce5/src/coreclr/hosts/corerun/corerun.hpp#L261-L296) for `corerun`. @weltkante's example above is the same logic, but in C# - thank you!

A manifest example can be found in dotnet/runtime for one of the COM tests - [`App.manifest`](https://github.com/dotnet/runtime/blob/main/src/tests/Interop/COM/NETClients/IDispatch/App.manifest).

kant2002:
@RussKie can you, or somebody else from the team look at what I done with csproj and sln for ARM64? I do not have hardware, so cannot test this configuration and something is off

kant2002:
> Oops, I missed this tag and came here to check why I was getting so many updates

Thankfully for me, @MattGal all these updates right now from the regular dev activity. I implement reg-free COM with help and guidance of others. Breakages in CI right now from my lack of access to ARM64

> registering random COM objects on their machine at best leaves stuff behind polluting their registry and at worst can cause unexpected behavior / stability issues.

Yeah, that was desperate request to somehow get unblocked. Thankfully my advisors all insist that I would not be lazy and implement proper solution. Deeply thankful for that.

> If you go one of the other routes and it comes around to affect other users, we will chat in the future :)

I appreciate that people on Dotnet organization so talkative, and like to make new friends. I probably reserve chat for with you, for more interesting topics. :)

kant2002:
@RussKie I think I make it works, but right now I see some issues with build itself. Once more then one project with `<NativeProjectReference Include="xxx\CMakeLists.txt" CMakeProject="xxx\xxx.proj" BuildNative="true" />` then current build start have race conditions. I thought I have just one issue with XML files which I fix by do not work with XML files in native projects https://github.com/dotnet/winforms/pull/6743/files#diff-50e91c82311ea26f2a73202525dfdf2b0a89c178ee666b2bf3ad4c84ac4c06dcR5, but now I see some issues with building CMake projects. I cannot explain issues which I see in any way as concurrency issues. I thought that this is because `BuildNative="true"` is set. And in order to have native projects, we probably have to move building of native projects to separate step.

Commit [9b54574](https://github.com/dotnet/winforms/pull/6743/commits/9b545744fd7d398013666ef627e742797617b535) has issues only with ARM64, and no funny CMake related issues. Since I change just solution configuration, I cannot explain what's going on currently.

I would like that somebody check my thought process. and look at the issues in last couple builds.

RussKie:
I'll have a look early next week.

kant2002:
@RussKie can somebody take a look again? I think my build issue was coincidental with GitHub issue at this day.

RussKie:
> can somebody take a look again? I think my build issue was coincidental with GitHub issue at this day.

It's building:
image
...or is there something else?

kant2002:
Not sure, but some of my builds there was some race conditions during Xml Intellisense files. There was no such issues locally, and after I rebase all issues are gone. But at least once agents timed out. 

Also I implement tests, so all my changes are covered now. I would like finish this PR

src/System.Windows.Forms/tests/UnitTests/System.Windows.Forms.Tests.csproj
```diff
@@ -24,6 +27,10 @@
     <ProjectReference Include="..\TestUtilities\System.Windows.Forms.TestUtilities.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <NativeProjectReference Include="WindowsFormsTests\CMakeLists.txt" CMakeProject="WindowsFormsTests\WindowsFormsTests.proj" BuildNative="true" />
```
RussKie, line 31: Can we rename to align with https://github.com/dotnet/winforms/tree/main/src/System.Windows.Forms/tests/InteropTests/NativeTests?
```suggestion
    <NativeProjectReference Include="WindowsFormsTests\CMakeLists.txt" CMakeProject="NativeTests\NativeTests.proj" BuildNative="true" />
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyGrid.IErrorInfoSupportTests.cs
```diff
@@ -0,0 +1,162 @@
...
+using System.ComponentModel;
+using System.Runtime.InteropServices;
+using Xunit;
+
+namespace System.Windows.Forms.Tests
+{
+    [Collection("Sequential")]
+    public class PropertyGrid_IErrorInfoSupportTests
+    {
+        public const int DISP_E_MEMBERNOTFOUND = unchecked((int)0x80020003);
```
RussKie, line 14: We have this defined in:
https://github.com/dotnet/winforms/blob/1f4032c1ba6e1bb97aee586605616a4f01cbaf45/src/System.Windows.Forms.Primitives/src/Interop/Interop.HRESULT.cs#L50

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyGrid.IErrorInfoSupportTests.cs
```diff
@@ -0,0 +1,162 @@
...
+        public const string System_Windows_Forms_NativeTests = "System_Windows_Forms_NativeTests";
```
RussKie, line 15: Where is this used?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyGrid.IErrorInfoSupportTests.cs
```diff
@@ -0,0 +1,162 @@
...
+            public string lpAssemblyDirectory; // LPCWSTR
+            public string lpResourceName; // LPCWSTR
+            public string lpApplicationName; // LPCWSTR
+            public IntPtr hModule; // HMODULE
+        }
+
+        private object CreateComObjectWithRawIErrorInfoUsage()
+        {
+            var clsid = new Guid("0ED8EE0D-22E3-49EA-850C-E69B20D1F296");
+            var IID_IUnknown = new Guid("{00000000-0000-0000-C000-000000000046}");
```
RussKie, line 121: We have this one defined in:
https://github.com/dotnet/winforms/blob/1f4032c1ba6e1bb97aee586605616a4f01cbaf45/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/NativeMethods.cs#L169

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyGrid.IErrorInfoSupportTests.cs
```diff
@@ -0,0 +1,162 @@
...
+            public int cbSize; // ULONG
+            public int dwFlags; // DWORD
+            public string lpSource; // LPCWSTR
+            public short wProcessorArchitecture; // USHORT
+            public short wLangId; // LANGID
+            public string lpAssemblyDirectory; // LPCWSTR
+            public string lpResourceName; // LPCWSTR
+            public string lpApplicationName; // LPCWSTR
+            public IntPtr hModule; // HMODULE
+        }
```
RussKie, line 116: We have this defined in:
https://github.com/dotnet/winforms/blob/1f4032c1ba6e1bb97aee586605616a4f01cbaf45/src/System.Windows.Forms.Primitives/src/Interop/Kernel32/Interop.ACTCTXW.cs#L11-L23

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyGrid.IErrorInfoSupportTests.cs
```diff
@@ -0,0 +1,162 @@
...
+            return result;
+        }
+
+        [DllImport("ole32.dll", ExactSpelling = true, PreserveSig = false)]
+        private static extern void CoCreateInstance(
+            ref Guid rclsid,
+            IntPtr punkOuter,
+            int dwClsContext,
+            ref Guid riid,
+            [MarshalAs(UnmanagedType.Interface)] out object ppv);
```
RussKie, line 148: This one is defined in:
https://github.com/dotnet/winforms/blob/1f4032c1ba6e1bb97aee586605616a4f01cbaf45/src/System.Windows.Forms.Primitives/src/Interop/Ole32/Interop.CoCreateInstance.cs#L11-L25

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyGrid.IErrorInfoSupportTests.cs
```diff
@@ -0,0 +1,162 @@
...
+        private static extern IntPtr CreateActCtxW(in ACTCTXW pActCtx);
+
+        [DllImport("kernel32", SetLastError = true)]
+        private static extern void ReleaseActCtx(IntPtr hActCtx);
+
+        [DllImport("kernel32", SetLastError = true)]
+        private static extern bool ActivateActCtx(IntPtr hActCtx, out IntPtr lpCookie);
+
+        [DllImport("kernel32")]
+        private static extern bool DeactivateActCtx(int dwFlags, IntPtr ulCookie);
```
RussKie, line 160: These are too defined in the Interop namespace (except for `ReleaseActCtx` I think)

src/System.Windows.Forms/tests/UnitTests/WindowsFormsTests/System_Windows_Forms_NativeTests.X.manifest
```diff
@@ -0,0 +1,20 @@
...
+
+<file name = "System_Windows_Forms_NativeTests.dll">
+  <!-- RawErrorInfoUsageTest -->
+  <comClass
+    clsid="{0ED8EE0D-22E3-49EA-850C-E69B20D1F296}"
+    threadingModel="Both" />
+    <!-- StandardErrorInfoUsageTest Test -->
+    <comClass
+      clsid="{EA1FCB3A-277C-4C79-AB85-E2ED3E858201}"
+      threadingModel="Both" />
```
RussKie, line 17: [nit] Incorrect indent

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyGrid.IErrorInfoSupportTests.cs
```diff
@@ -0,0 +1,162 @@
...
+            public short wLangId; // LANGID
+            public string lpAssemblyDirectory; // LPCWSTR
+            public string lpResourceName; // LPCWSTR
+            public string lpApplicationName; // LPCWSTR
+            public IntPtr hModule; // HMODULE
+        }
+
+        private object CreateComObjectWithRawIErrorInfoUsage()
+        {
+            var clsid = new Guid("0ED8EE0D-22E3-49EA-850C-E69B20D1F296");
```
RussKie, line 120:
```suggestion
            Guid clsidRawErrorInfoUsageTest = new("0ED8EE0D-22E3-49EA-850C-E69B20D1F296");
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyGrid.IErrorInfoSupportTests.cs
```diff
@@ -0,0 +1,162 @@
...
+                IntPtr.Zero,
+                1,
+                ref IID_IUnknown,
+                out object result);
+            return result;
+        }
+
+        private object CreateComObjectWithStandardIErrorInfoUsage()
+        {
+            var clsid = new Guid("EA1FCB3A-277C-4C79-AB85-E2ED3E858201");
```
RussKie, line 132:
```suggestion
            Guid clsidStandardErrorInfoUsageTest = new("EA1FCB3A-277C-4C79-AB85-E2ED3E858201");
```

src/System.Windows.Forms/tests/UnitTests/WindowsFormsTests/WindowsFormsTests.proj
```diff
@@ -0,0 +1,6 @@
+<Project Sdk="Microsoft.DotNet.CMake.Sdk">
```
RussKie, line 1: Can you please add a readme file for those who aren't very good with C++ and COM (e.g., me 😊) explaining what this project does. 
If I understand it correctly, this project contains two COM objects RawErrorInfoUsageTest (CLSID: 0ED8EE0D-22E3-49EA-850C-E69B20D1F296) and StandardErrorInfoUsageTest (CLSID: EA1FCB3A-277C-4C79-AB85-E2ED3E858201), that are used in PropertyGrid_IErrorInfoSupportTests tests.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyGrid.IErrorInfoSupportTests.cs
```diff
@@ -0,0 +1,162 @@
...
+                    var encodingEntry = entries[0].Children.First(_ => _.PropertyName == "Int_Property");
+                    try
+                    {
+                        encodingEntry.SetPropertyTextValue("333");
+                        Assert.False(true);
+                    }
+                    catch (ExternalException ex)
+                    {
+                        Assert.Equal(DISP_E_MEMBERNOTFOUND, ex.HResult);
+                    }
```
RussKie, line 37: It would be great if you could add some comments explaining the test mechanics, and why we're asserting in the `catch` clause. Ditto for `ISupportErrorInfo_Supported_WithIErrorInfoGiven`.

RussKie (commented):
This is a 🚀 science 😄.

eng/init-vs-env.cmd
```diff
@@ -0,0 +1,72 @@
+@if not defined _echo @echo off
```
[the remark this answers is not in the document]
kant2002: This script right now run as part of build. For the developer no need to do anything with this script.

src/System.Windows.Forms/tests/UnitTests/System.Windows.Forms.Tests.csproj
kant2002: It did not work for me. Have  issues with CMake SDK where both projects use `NativeTests` as output folder. Maybe `NativeTests\WindowsFormsTests.proj` is good compromise. Not sure that this will works.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyGrid.IErrorInfoSupportTests.cs
kant2002: Letftover. Will remove.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyGrid.IErrorInfoSupportTests.cs
kant2002: I would like to keep tests unrelated to Winforms code if possible. Yes, this will lead to a bit of duplication, but that way we would not touch code.

Specifically this `[MarshalAs(UnmanagedType.Interface)] out object ppv);` require have built-in COM interop. I would like to push code which does not rely on it. End-goal is this declaration would not be present at all in the Primitives project, and it anyway have to be reintroduced in tests.

Maybe `IDispatch` support never let me do that with current ComWrappers approach, but `IDynamicInterfaceCastable` should do the trick eventually.

Winforms.sln
```diff
@@ -193,22 +193,22 @@ Global
...
+		{AB38E262-F206-4820-8F29-23C5F72A4A16}.Debug|x86.ActiveCfg = Debug|x86
+		{AB38E262-F206-4820-8F29-23C5F72A4A16}.Debug|x86.Build.0 = Debug|x86
+		{AB38E262-F206-4820-8F29-23C5F72A4A16}.Release|Any CPU.ActiveCfg = Release|x86
+		{AB38E262-F206-4820-8F29-23C5F72A4A16}.Release|Any CPU.Build.0 = Release|x86
+		{AB38E262-F206-4820-8F29-23C5F72A4A16}.Release|arm64.ActiveCfg = Release|x64
+		{AB38E262-F206-4820-8F29-23C5F72A4A16}.Release|arm64.Build.0 = Release|x64
+		{AB38E262-F206-4820-8F29-23C5F72A4A16}.Release|x64.ActiveCfg = Release|x64
+		{AB38E262-F206-4820-8F29-23C5F72A4A16}.Release|x64.Build.0 = Release|x64
+		{AB38E262-F206-4820-8F29-23C5F72A4A16}.Release|x86.ActiveCfg = Release|x86
+		{AB38E262-F206-4820-8F29-23C5F72A4A16}.Release|x86.Build.0 = Release|x86
```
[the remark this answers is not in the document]
kant2002: CMake did not like when I have AnyCPU for the `csproj` with CMake reference.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyGrid.IErrorInfoSupportTests.cs
kant2002: I prefer tests to be self-contained. I would not insist on that, but what if have separate set of constants in Primitives and somewhere in the tests infra?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyGrid.IErrorInfoSupportTests.cs
kant2002: I would try, but GridEntry and their behavior is a bit black box. I just capture existing behaviour. Will try to comeup with some explanation for catch, etc.

src/System.Windows.Forms/tests/UnitTests/WindowsFormsTests/WindowsFormsTests.proj
kant2002: That's for now. I expect to have more COM objects in this project eventually, since you would not let me off the hook with my changes. Will add README where I try to give rationale for project and Reg-free com configuration.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyGrid.IErrorInfoSupportTests.cs
kant2002: I can use them from Primitives project, but would like to keep declarations in the test projects if possible. All such cases except CoCreateInstance is not important IMO for have non-changeable tests.

kant2002:
@RussKie I left some comments about code duplication in tests. Let me know about your opinion, since immutable tests is pattern which I mimic from runtime repo. After that I will start cleaning up and adding comments.

Winforms.sln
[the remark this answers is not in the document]
kant2002: Also seems that ARM64 tests executed under x64 simulation (?) because it is important that ARM64 configuraiton build using x64, otherwise build did not build/pickup native parts.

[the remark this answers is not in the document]
weltkante: When I was looking at native tests last there were no ARM tests, the project files were building X64 just so they would have something to build, but no tests were executed. I don't know if the situation changed, quite some time has gone by.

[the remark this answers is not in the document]
kant2002: > When I was looking at native tests last there were no ARM tests

There ARM64 unit tests running. https://dev.azure.com/dnceng/public/_build/results?buildId=1677777&view=logs&j=a8f71490-549e-54d9-6b98-8fc4837d22f9&t=9ed0a6c0-fd3e-5751-41bf-59e617722956

But because I do not have hardware, I just make tests pass in CI. But I have to say that ARM64 solution configuration should build x64 DLL. You can see that in SLN file. When I was "smart" and try to say "arm64" it blow-up on CI.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyGrid.IErrorInfoSupportTests.cs
RussKie: a duplication is a maintenance overhead. If there's more than one definition, a contributor has a dilemma - which definition is to use? And if the definitions differ in some way - which definition is correct? And so on...

> End-goal is this declaration would not be present at all in the Primitives project, and it anyway have to be reintroduced in tests.

We are free to move anything that's not a part of a public API, and if we have definitions that are "test only" - we can move those across.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyGrid.IErrorInfoSupportTests.cs
```diff
@@ -0,0 +1,147 @@
...
+                    try
+                    {
+                        encodingEntry.SetPropertyTextValue("333");
+                        Assert.False(true, "Invalid set values should produce ExternalException which will be presenttted to the user.");
+                    }
+                    catch (ExternalException ex)
+                    {
+                        // Most default C++ implementation when Invoke return error code
+                        // consult IErrorInfo object and populate EXCEPINFO structure
+                        // So grid entry knows only about error code.
```
RussKie, line 36: 👍 
```suggestion
                        // Most default C++ implementations when Invoke return error code.
                        // Consult IErrorInfo object and populate EXCEPINFO structure so grid entry knows only about the error code.
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyGrid.IErrorInfoSupportTests.cs
```diff
@@ -0,0 +1,147 @@
...
+                    try
+                    {
+                        encodingEntry.SetPropertyTextValue("123");
+                        Assert.False(true, "Invalid set values should produce ExternalException which will be presenttted to the user.");
+                    }
+                    catch (ExternalException ex)
+                    {
+                        // If C++ implementation of Invoke did not populate EXCEPINFO structure
+                        // from IErrorInfo then we read that information about error call.
+                        // and display that error message to the user.
```
RussKie, line 69: 👍 

```suggestion
                        // If C++ implementation of Invoke did not populate EXCEPINFO structure
                        // from IErrorInfo, then we read that information about error call and display that error message to the user.
```

src/System.Windows.Forms/tests/UnitTests/WindowsFormsTests/README.md
```diff
@@ -0,0 +1,18 @@
+Supplementary native parts for WinForms unit tests
+==================================================
+
+This project is native DLL which provide unmanaged parts required to testing of WinForms components. Most interesting part of the tests would be COM objects which given to controls.
+This project builds as native DLL, which can be used as reg-free COM host. That allow run tests and do not worry about polluting the developer/build machines.
```
RussKie, line 5:
```suggestion
This project is native DLL which provide unmanaged parts required for testing WinForms components. Most interesting part of the tests would be COM objects which given to controls.
This project builds as a native DLL, which can be used as a reg-free COM host. That allow running tests without worrying about "polluting" developer/build machines.
```

RussKie (commented):
Thank you for the comments and the readme 👍

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyGrid.IErrorInfoSupportTests.cs
kant2002: > a duplication is a maintenance overhead.

I think not. I think so, because I prefer having tests in write-only mode. This does not works well with new projects, but works perfect with mature projects like this one. If you never touch test then you do not care how things are declared.

If we have same declaration in code and tests we may slightly change test scenario when we refactor. Since as you told, team not always have full understanding of consequences of changes, Also that allow refactoring spree, and do not care that you may break tests by that.

In a sense writing tests is already "maintainance overhead" for regular developer, since writing tests slightly different mindset.

Would like to summon @weltkante on this topic, and other persistent contributors to say on the matter.

I will comply in this PR, since for me it is more important to move forward, but expect me to be vocal about this testing strategy on each opportunity. I hope I can change team's mind on this topic.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyGrid.IErrorInfoSupportTests.cs
kant2002: This change meaning. `Invoke` method look a the IErrorInfo and populate `EXCEPINFO` where you assume that it is developer/caller job to look at IErrorInfo. It does not mean that my English is fine 😄

kant2002: I rewrote explanation.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyGrid.IErrorInfoSupportTests.cs
weltkante: I don't have a strong opinion here. I also haven't been following closely what the "state of the art" for the new COM interop is - but `CoCreateInstance` was redundant under the old model, you'd just annotate the interface/class correctly and could `new` it up directly. Under the new model it seems wrong that its annotated with `MarshalAs`.

kant2002: Thanks for feedback. I really do not think that in tests we should go with “new” ways only. But for this particular function, “old” way is better, since it allow write test easier. Anyway. I do apply @RussKie suggestion so I can move forward

kant2002:
@RussKie can you take a look, to keep me busy?

RussKie:
The team  is currently busy with urgent work, and it may be few days before I may get to this. Please bear with me.

kant2002:
Thank you for telling me. For sure I will wait how long it is necessary.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyGrid.IErrorInfoSupportTests.cs
AaronRobinsonMSFT: @kant2002 I agree with @weltkante here. Having a single definition is preferrable. I appreciate your statement regarding "new" vs "old" ways in which COM interop works though. It should be possible to simply do the object marshal manually using [`Marshal.GetObjectForIUnknown`](https://docs.microsoft.com/dotnet/api/system.runtime.interopservices.marshal.getobjectforiunknown). The benefit here is the same API can be called for either and the tests explicit indicate what they want to test as opposed to relying on the magic of the generated IL stub.

kant2002:
@AaronRobinsonMSFT I would like to ask your opinion on my manual approach to reducing size of RCW for IErrorInfo for example. I provide just necessary methods which is require to run WinForms. Let’s say in the future somebody want to use additional methods from the COM interface, so it has to manually check for method number in Vtbl. That’s error prone. At the same time if RCW would be generated automatically all methods would be added. Isn’t this “expensive” and what do you think about this in general?

src/System.Windows.Forms/tests/UnitTests/WindowsFormsTests/StandardErrorInfoUsageTest.h
```diff
@@ -0,0 +1,53 @@
...
+
+class StandardErrorInfoUsageTest : public DispatchImpl, public IBasicTest, public ISupportErrorInfo
+{
+public:
+    StandardErrorInfoUsageTest()
+        : DispatchImpl(IID_IBasicTest, static_cast<IBasicTest*>(this))
+    {
+    }
+
+    ~StandardErrorInfoUsageTest()
```
AaronRobinsonMSFT, line 19: The destructor isn't needed here. If it is desired, then I'd use the default.

```c++
~StandardErrorInfoUsageTest() = default;
```

src/System.Windows.Forms/tests/UnitTests/WindowsFormsTests/RawErrorInfoUsageTest.h
```diff
@@ -0,0 +1,53 @@
...
+
+class RawErrorInfoUsageTest : public DispatchImpl, public IBasicTest, public ISupportErrorInfo
+{
+public:
+    RawErrorInfoUsageTest()
+        : DispatchImpl(IID_IBasicTest, static_cast<IBasicTest*>(this))
+    {
+    }
+
+    ~RawErrorInfoUsageTest()
```
AaronRobinsonMSFT, line 19: Remove or use default.

```cpp
~RawErrorInfoUsageTest() = default;
```

src/System.Windows.Forms/tests/UnitTests/WindowsFormsTests/StandardErrorInfoUsageTest.cpp
```diff
@@ -0,0 +1,43 @@
...
+    /* [retval][out] */ int *ret)
+{
+    *ret = _int;
+    return S_OK;
+}
+
+HRESULT STDMETHODCALLTYPE StandardErrorInfoUsageTest::put_Int_Property(
+    /* [in] */ int val)
+{
+    ICreateErrorInfo* cei;
```
AaronRobinsonMSFT, line 17: `ComSmartPtr<>`

src/System.Windows.Forms/tests/UnitTests/WindowsFormsTests/StandardErrorInfoUsageTest.cpp
```diff
@@ -0,0 +1,43 @@
...
+    if (SUCCEEDED(::CreateErrorInfo(&cei)))
+    {
+        if (SUCCEEDED(cei->SetGUID(IID_IBasicTest)))
+        {
+            if (SUCCEEDED(cei->SetDescription(L"Error From StandardErrorInfoUsageTest")))
+            {
+                IErrorInfo* errorInfo;
```
AaronRobinsonMSFT, line 24: `ComSmartPtr<>`

src/System.Windows.Forms/tests/UnitTests/WindowsFormsTests/DispatchImpl.cpp
```diff
@@ -0,0 +1,58 @@
...
+    HRESULT hr = ::LoadTypeLibEx(tlbToLoad, REGKIND::REGKIND_NONE, &_typeLib);
+    if (FAILED(hr))
+        throw hr;
+
+    hr = _typeLib->GetTypeInfoOfGuid(guid, &_typeInfo);
+    if (FAILED(hr))
+        throw hr;
+}
+
+DispatchImpl::~DispatchImpl()
```
AaronRobinsonMSFT, line 26: This destructor can be turned into `virtual ~DispatchImpl() = default;` in the header if `ComSmartPtr<>` is used for `IUnknown` types.

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.ErrorInfoWrapper.cs
```diff
@@ -0,0 +1,41 @@
...
+                Marshal.Release(_wrappedInstance);
+                _wrappedInstance = IntPtr.Zero;
+            }
+
+            public bool GetDescription([NotNullWhen(true)] out string? pBstrDescription)
+            {
+                IntPtr descriptionPtr;
+                var result = ((delegate* unmanaged<IntPtr, IntPtr*, HRESULT>)(*(*(void***)_wrappedInstance + 5 /* IErrorInfo.GetDescription */)))
+                    (_wrappedInstance, &descriptionPtr);
+                pBstrDescription = Marshal.PtrToStringUni(descriptionPtr);
```
AaronRobinsonMSFT, line 35: These should only be checked if `result` is non-failure.

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.ErrorInfoWrapper.cs
```diff
@@ -0,0 +1,41 @@
...
+
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.InteropServices;
+using System.Windows.Forms;
+
+internal partial class Interop
+{
+    internal unsafe partial class WinFormsComWrappers
+    {
+        internal class ErrorInfoWrapper
```
AaronRobinsonMSFT, line 13: I would add the methods for the rest of the `IErrorInfo` interface. They can all just throw `NotImplementedException`.

src/System.Windows.Forms.Primitives/src/Interop/OleAut32/Interop.IErrorInfo.cs
```diff
@@ -1,37 +0,0 @@
...
-using System.Runtime.InteropServices;
-
-internal partial class Interop
-{
-    internal static partial class Oleaut32
-    {
-        [ComImport]
-        [Guid("1CF2B120-547D-101B-8E65-08002B2BD119")]
-        [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
-        public unsafe interface IErrorInfo
```
AaronRobinsonMSFT, line 14: I don't understand why we are getting rid of the `IErrorInfo` type here. That is what we are working with not an `ErrorInfoWrapper`.

src/System.Windows.Forms/tests/UnitTests/WindowsFormsTests/DispatchImpl.h
```diff
@@ -0,0 +1,43 @@
...
+    DispatchImpl& operator=(DispatchImpl&&) = default;
+
+protected:
+    HRESULT DoGetTypeInfoCount(UINT* pctinfo);
+    HRESULT DoGetTypeInfo(UINT iTInfo, ITypeInfo** ppTInfo);
+    HRESULT DoGetIDsOfNames(LPOLESTR* rgszNames, UINT cNames, DISPID* rgDispId);
+    HRESULT DoInvoke(DISPID dispIdMember, WORD wFlags, DISPPARAMS* pDispParams, VARIANT* pVarResult, EXCEPINFO* pExcepInfo, UINT* puArgErr);
+
+private:
+    ITypeLib *_typeLib;
```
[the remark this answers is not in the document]
kant2002: That's borrowed from https://github.com/dotnet/runtime/blob/71ad9f5382b1936db7bb6ee03a75efdb8c7cb675/src/tests/Interop/COM/Dynamic/Server/DispatchImpl.h#L29-L31 
Should go back and change there too?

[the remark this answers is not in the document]
AaronRobinsonMSFT: Boo. I didn't realize we missed that in dotnet/runtime. Yes, we should update that as well.

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.ErrorInfoWrapper.cs
kant2002: Ok. I was thinking that general sentiment is to remove methods https://github.com/dotnet/runtime/blob/890fde7c5a57ca3f367131f173773c410e4fc26f/src/libraries/System.Drawing.Common/src/System/Drawing/DrawingCom.COMWrappers.cs#L284-L293

Also Jeremy has different idea, to comment out method/#ifdef so I see slight twists on the matter in each person. If you see I remove interface altogether, since I do not think that having artificial interface is anybetter then just directly work with class. I notice that just so you have visibility of the usage for that pattern which you probably interested in.

src/System.Windows.Forms.Primitives/src/Interop/OleAut32/Interop.IErrorInfo.cs
kant2002: I do not fully understand what do you mean. Technically in this piece of code there no `IErrorInfo` usage anymore and I get rid of it.  I think you saying something which I do not get.

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.ErrorInfoWrapper.cs
AaronRobinsonMSFT: That is a fair perspective to have. My personal preference is to have the well-known COM interface around instead of having to search for it at some point in the future. Simply operating on the class itself isn't a bad idea, just slightly confusing from my perspective. I say that because if I come into this space the API to use is `GetErrorInfo()` and that is documented to return an `IErrorInfo` but in this code base I don't see that used and so am confused as to why. This is just my perspective, likely because I am so familiar with COM.

Jeremy's perspective to have the methods but ifdef them out also seems reasonable. I don't have a strong preference other than someone coming into this space is likely to at least be passingly familiar with COM and `GetErrorInfo()`. However, how it is currently designed avoids anything familiar with respect to COM and forces the new person to start from the beginning and ask: how does all this fit together?

src/System.Windows.Forms.Primitives/src/Interop/OleAut32/Interop.IErrorInfo.cs
AaronRobinsonMSFT: My point is that using the `GetErrorInfo()` and `IErrorInfo` is more intuitive with the correct types that match the official documentation. We can define these types nicely in C# so I would suggest doing that. However, I will defer to the WinForms team if their preference is what you have written here.

RussKie:
> Failures seem to be a parallel build issue. Does the `NativeTests.proj` need to have access to the `xml` doc?

Not at all. Raised https://github.com/dotnet/winforms/issues/6928 to track the reentrancy.

RussKie:
> Is there something I can help review here?

@AaronRobinsonMSFT yes, your help is very much appreciated 👍

kant2002:
@AaronRobinsonMSFT I do think I address all main questions. @RussKie can you also take a look again?

RussKie (commented):
I'm trying to catch up, and I'd appreciate it if you could help me understand the chosen organisation of the tests.

To me PropertyGrid.IErrorInfoSupportTests look like integration tests (arguably most of our unit tests are not "true" unit tests as our implementations involve Win32, etc., but some tests are "more integration tests" than the other). And logically these tests, I'd think, belong to System.Windows.Forms.Interop.Tests project. The C++ implementations could reside in the System.Windows.Forms.Interop.Tests/NativeTests project.
Was there any reason why you decided not to place the new tests there?

Thank you

kant2002:
My rationale is following. Tests for me mostly are specifications of behaviours for the component. Displaying COM object inside Property Grid is just one of these requirements. I suspect if wrote properly we can have up to 10 COM objects of different shapes which test all possible cases of OLE/IDispatch magic. Integration testing for me is more of when multiple things come into interplay. For example: I setup `PropertyGrid`, set `SelectObject`, let somebody indirectly change that object and indirectly trigger `PropertyGrid` update. Then this would be qualified for me as integration testing.

The more requirements for same component living together, the better. I do not think that COM interop is something special for `PropertyGrid`. Consider that this eventually would be not one-off job, but more `ComWrappers` work, the more tests should I have to cover existing codepaths for ugly cases.

Anyway, is rule of thumb that all "unit-tests" which involve COM should be in the `System.Windows.Forms.Interop.Tests` I definitely can live with that.

I will take a look at the new location for tests, moving things should not be that difficult.

RussKie:
> Anyway, is rule of thumb that all "unit-tests" which involve COM should be in the `System.Windows.Forms.Interop.Tests` I definitely can live with that.
> 
> I will take a look at the new location for tests, moving things should not be that difficult.

🚀

RussKie (approved):
👍

RussKie:
@JeremyKuhne can you please review as well

JeremyKuhne (changes requested):
Just some formatting nits- otherwise I'm good with this. Thanks for your effort!

kant2002:
@RussKie @JeremyKuhne I think I'm done here.

JeremyKuhne (approved):
:shipit:

PR #6752, opened 2022-02-23
Fixing layout issues of Anchored/Docked controls in the MDI child forms

### LINKED ISSUE #6259: ChildMdi Container control rendering issues when not Maximized
* .NET Core Version: 
dotnet-sdk-7.0.100-alpha.1.21568.2-win-x64

No

**Issue** When moving an MdiParent to a higher DPI monitor, the ChildMdi (not maximized) TabControl on the bottom becomes slightly wider and slightly shorter after multiple moves.  

Run the project and click MdiParent - Simple Layout on the 100% monitor
Click on the Maximize/Minimize to un-maximize the ChildMDI form, then drag the MdiParent to the 150% monitor.  The Width = 613 and the Height = 261 as displayed in the text box inside the tab control.  After moving it to the 100% and back again, the size becomes 619x233.  After doing this again, it becomes 816x204 and the tab control now runs past the right edge.
ChildMDI Not Maximized
[PerMonitorHDPICore - NET 7.0.zip](https://github.com/dotnet/winforms/files/7646074/PerMonitorHDPICore.-.NET.7.0.zip)

dreddy-work:
Fixes #6259 

## Proposed changes

When moving the Form to a new monitor with different DPI settings, Windows changing the Dpi
on the handle and sending DPI changed message to the Form are not in sync. Due to this,
method User32.GetClientRect are returning the incorrect value.  This method is not
accurate with respect to adornments metrics and their Dpi. Hence this work around.

This method is expected to be called whenever there is WINDOWSPOSITION_CHANGED message is triggered.
We will be calling UpdateBounds(x, y, width, height) explicitly, instead, for MDI child windows
if we had to change the size. for every other control, bounds are changed here.

## Regression? 

- No

## Risk

- Low. Impacts only MDI child forms

## Test methodology 

- Manual. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6752)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
@@ -10854,7 +10854,7 @@ namespace System.Windows.Forms
                             // NOTE: SetWindowPos causes a WM_WINDOWPOSCHANGED which is processed
                             // synchonously so we effectively end up in UpdateBounds immediately following
                             // SetWindowPos.
-                            // In case of MDI child windows, UpdateBOund() is made a no-op with respect
+                            // In case of MDI child windows, UpdateBounds() is made no-op with respect
                             // to bounds due to incorrect size returned by GetClientRect() method when
                             // UpdateBounds() is triggered as a result of DPI_CHNAGED event. As a workaround,
                             // we would be explicitly calling UpdateBounds(x, y, width, height).
@@ -11539,13 +11539,13 @@ namespace System.Windows.Forms
                 User32.MapWindowPoints(IntPtr.Zero, User32.GetParent(new HandleRef(_window, InternalHandle)), ref rect);
 
                 // When moving the Form to a new monitor with different DPI settings, Windows changing the Dpi
-                // on the handle and sendign DPI changed message to the Form are not in sync. Due to this,
-                // methods User32.GetClientRect are returnign the incorrect value.  This method is not
-                // accurate with respect to adornments metrics and their Dpi. Hence this work around.
+                // on the handle and sending DPI changed message to the Form are not in sync. Due to this,
+                // methods User32.GetClientRect are returning the incorrect value. User32.GetClientRect method is not
+                // accurate with respect to adornments metrics, their Dpi and current layout transaction. Hence this work around.
 
-                // This method is expected to be called whenever there is WINDOWSPOSITION_CHANGED message is triggered.
-                // We will be calling UpdateBounds(x, y, width, height) explicitly, instead, for MDI child windows
-                // if we had to chnage the size. for every other control, bounds are changed here.
+                // UpdateBounds() is expected to be called whenever there is WINDOWSPOSITION_CHANGED message is triggered.
+                // We will be calling UpdateBounds(x, y, width, height) explicitly, instead, for MDI child windows if we
+                // had to change the size. For every other control, bounds are changed here.
                 if (this is Form form && form.IsMdiChild)
                 {
                     clientWidth = form.ClientRectangle.Width;
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
@@ -10854,8 +10854,15 @@ namespace System.Windows.Forms
                             // NOTE: SetWindowPos causes a WM_WINDOWPOSCHANGED which is processed
                             // synchonously so we effectively end up in UpdateBounds immediately following
                             // SetWindowPos.
-                            //
-                            //UpdateBounds(x, y, width, height);
+                            // In case of MDI child windows, UpdateBOund() is made a no-op with respect
+                            // to bounds due to incorrect size returned by GetClientRect() method when
+                            // UpdateBounds() is triggered as a result of DPI_CHNAGED event. As a workaround,
+                            // we would be explicitly calling UpdateBounds(x, y, width, height).
+
+                            if (this is Form form && form.IsMdiChild)
+                            {
+                                UpdateBounds(x, y, width, height);
+                            }
                         }
                     }
                 }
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -11523,16 +11530,36 @@ protected internal void UpdateBounds()
...
+            int width = rect.right - rect.left;
+            int height = rect.bottom - rect.top;
+
             if (!GetTopLevel())
             {
                 User32.MapWindowPoints(IntPtr.Zero, User32.GetParent(new HandleRef(_window, InternalHandle)), ref rect);
+
+                // When moving the Form to a new monitor with different DPI settings, Windows changing the Dpi
+                // on the handle and sendign DPI changed message to the Form are not in sync. Due to this,
+                // methods User32.GetClientRect are returnign the incorrect value.  This method is not
```
RussKie, line 11543: typo: returnign

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -11523,16 +11530,36 @@ protected internal void UpdateBounds()
...
+                // accurate with respect to adornments metrics and their Dpi. Hence this work around.
```
RussKie, line 11544: I'm finding "_This method is not accurate_" confusing, I can't infer from the context which method it refers to
```suggestion
                // methods User32.GetClientRect are returnign the incorrect value.  This method is not
                // accurate with respect to adornments metrics and their Dpi. Hence this work around.
```

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -11523,16 +11530,36 @@ protected internal void UpdateBounds()
...
+                // This method is expected to be called whenever there is WINDOWSPOSITION_CHANGED message is triggered.
+                // We will be calling UpdateBounds(x, y, width, height) explicitly, instead, for MDI child windows
+                // if we had to chnage the size. for every other control, bounds are changed here.
+                if (this is Form form && form.IsMdiChild)
+                {
+                    clientWidth = form.ClientRectangle.Width;
+                    clientHeight = form.ClientRectangle.Height;
+                    width = form.Bounds.Width;
+                    height = form.Bounds.Height;
+                }
```
RussKie, line 11555: Same here, the comment states "_This method_", yet there are no method calls here.

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -11523,16 +11530,36 @@ protected internal void UpdateBounds()
...
                 User32.MapWindowPoints(IntPtr.Zero, User32.GetParent(new HandleRef(_window, InternalHandle)), ref rect);
+
+                // When moving the Form to a new monitor with different DPI settings, Windows changing the Dpi
+                // on the handle and sendign DPI changed message to the Form are not in sync. Due to this,
+                // methods User32.GetClientRect are returnign the incorrect value.  This method is not
+                // accurate with respect to adornments metrics and their Dpi. Hence this work around.
+
+                // This method is expected to be called whenever there is WINDOWSPOSITION_CHANGED message is triggered.
+                // We will be calling UpdateBounds(x, y, width, height) explicitly, instead, for MDI child windows
+                // if we had to chnage the size. for every other control, bounds are changed here.
```
RussKie, line 11548: typo "chnage"
typo ". for"

RussKie (commented):
I like the comment, we just need to fix typos and reword the comments to make them a little clearer.

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -10854,8 +10854,15 @@ protected virtual void SetBoundsCore(int x, int y, int width, int height, Bounds
                             // NOTE: SetWindowPos causes a WM_WINDOWPOSCHANGED which is processed
                             // synchonously so we effectively end up in UpdateBounds immediately following
                             // SetWindowPos.
-                            //
-                            //UpdateBounds(x, y, width, height);
+                            // In case of MDI child windows, UpdateBOund() is made a no-op with respect
```
Tanya-Solyanik, line 10857: typo: `UpdateBOunds`
and `is made a no-op` -> 'is 'no-op'

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -10854,8 +10854,15 @@ protected virtual void SetBoundsCore(int x, int y, int width, int height, Bounds
...
+                            // to bounds due to incorrect size returned by GetClientRect() method when
+                            // UpdateBounds() is triggered as a result of DPI_CHNAGED event. As a workaround,
+                            // we would be explicitly calling UpdateBounds(x, y, width, height).
+
+                            if (this is Form form && form.IsMdiChild)
```
Tanya-Solyanik, line 10862: We need the `UpdateBounds` on line 10864 when processing position changed after a DPI changed. Is it possible to add a condition to test that?

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -11523,16 +11530,36 @@ protected internal void UpdateBounds()
...
+
+            int width = rect.right - rect.left;
+            int height = rect.bottom - rect.top;
+
             if (!GetTopLevel())
             {
                 User32.MapWindowPoints(IntPtr.Zero, User32.GetParent(new HandleRef(_window, InternalHandle)), ref rect);
+
+                // When moving the Form to a new monitor with different DPI settings, Windows changing the Dpi
+                // on the handle and sendign DPI changed message to the Form are not in sync. Due to this,
```
Tanya-Solyanik, line 11542: typo: `sendign `

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -11523,16 +11530,36 @@ protected internal void UpdateBounds()
             int clientWidth = rect.right;
             int clientHeight = rect.bottom;
             User32.GetWindowRect(new HandleRef(_window, InternalHandle), ref rect);
```
Tanya-Solyanik, line 11532: Would it make sense to exclude this interop call for MDI windows?

Tanya-Solyanik (commented):
Added some minor comments. Otherwise looks good.

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
dreddy-work: For MDI, position is relative to parent and that shouldn't change. Any other scenario here?

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -11523,16 +11530,36 @@ protected internal void UpdateBounds()
...
+            int height = rect.bottom - rect.top;
+
             if (!GetTopLevel())
             {
                 User32.MapWindowPoints(IntPtr.Zero, User32.GetParent(new HandleRef(_window, InternalHandle)), ref rect);
+
+                // When moving the Form to a new monitor with different DPI settings, Windows changing the Dpi
+                // on the handle and sending DPI changed message to the Form are not in sync. Due to this,
+                // methods User32.GetClientRect are returning the incorrect value. User32.GetClientRect method is not
+                // accurate with respect to adornments metrics, their Dpi and current layout transaction. Hence this work around.
```
RussKie, line 11544: [nit] inconsistent use of "DPI" and "Dpi".

RussKie (approved):
👍

PR #6640, opened 2022-02-08
Moving 3 properties at `GetPropertyValue` method to the base `AccessibleObject` class

ArtemTatarinov:
Fixes #6582

## Proposed changes
- `IsPasswordPropertyId`, `FrameworkIdPropertyId` and `AccessKeyPropertyId` are refactored by moving to the base `AccessibleObject` class
- `AccessKeyPropertyId` is returning an empty string now if `KeyboardShortcut` is null, I guess that's a suitable change of behavior because it allows us to clean a lot of places in the code where an empty string is returned, also the `GetPropertyValue` method is an `internal` method anyway, and UIA tools like Inspect and Accessibility Insights treat `null` and `string.Empty` the same way, in the codebase `KeyboardShortcut` is checked for both null/string.Empty.

## Customer Impact
- None

## Regression? 
- No

## Risk
- Minimal

## Test methodology
- Unit tests
- CTI testing

## Test environment(s)
.NET SDK 7.0.0-preview-2.22103.2
Microsoft Windows [Version 10.0.19043.1466]

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6640)

PR #6761, opened 2022-02-25
Removing empty AccessKey properties which appeared after refactoring

### LINKED ISSUE #6760: Empty AccessKey values appeared after refactoring
* .NET Core Version: 7.0

After PR #6640 was merged some accessible objects have received the empty `AccessKey` values at the accessibility tools (like `Inspect`). We discussed it with the team and considered it as a regression. Their behavior should be restored as it were before this PR.

ArtemTatarinov:
Fixes #6760

## Proposed changes
- `AccessibleObject` returns just `KeyboartShortcut` without additional conditions, so if it's `null`, the `null` value will be returned, not the `string.Empty` value like in the previous PR
- Behavior for accessible objects which return `string.Empty` is restored, so there will be no changes for them

## Customer Impact
- Accessibility tools will have the same behavior as before refactoring

## Regression? 
- Yes, from PR #6640

## Risk
- Minimal

## Test methodology 
- Manual (viewed properties with the Inspect tool)
- Unit test
- CTI testing

## Test environment(s) 
- Microsoft Windows [Version 10.0.19043.1526]
- .NET 7.0.0-preview-3.22123.2

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6761)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/AccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/AccessibleObject.cs
@@ -427,7 +427,7 @@ namespace System.Windows.Forms
                 UiaCore.UIA.ValueValuePropertyId => !string.IsNullOrEmpty(Value) ? Value : null,
                 UiaCore.UIA.IsPasswordPropertyId => false,
                 UiaCore.UIA.FrameworkIdPropertyId => NativeMethods.WinFormFrameworkId,
-                UiaCore.UIA.AccessKeyPropertyId => KeyboardShortcut ?? string.Empty,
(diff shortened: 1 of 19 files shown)


Tanya-Solyanik (approved):
Looks good, thank you! 
Waiting for testing to merge.

dreddy-work:
@Olina-Zhang / @Tanya-Solyanik , work around is posted as below untill fix is in:

copy :
.\.dotnet\shared\Microsoft.NETCore.App to C:\Program Files\dotnet\shared\Microsoft.NETCore.App.

Tanya-Solyanik:
This issue is actually not a regression from the .NET Framework, closing the PR.

PR #6778, opened 2022-03-02
Fixing Anchor Info calculations.

### LINKED ISSUE #5774: Parent `PerformLayout()` sets negative child `Location` at scale factor >1
* .NET Core Version: **.NET 5.0.9 and .NET 6.0 RC2**

* Have you experienced this same bug with .NET Framework?: **No**

**Problem description:**

The problem is originally reported at
* https://github.com/gitextensions/gitextensions/issues/9402, 
* https://github.com/gitextensions/gitextensions/issues/9491.

### The following report is based on https://github.com/gitextensions/gitextensions/issues/9402

The app on .NET Framework 4.x behaves as expected (150% sf):
ge-net4x

The ported version to .NET 5.0/6.0 exhibit the **issue at scale factor > 100%** (at 100% everything works as expected):
ge-net60

During a high level debugging it looks like the panel in question gets assigned negative v-coords when the parent tabpage performs the layout:
ge-net60-debugger

**Minimal repro:**

1. set the 
1. clone https://github.com/RussKie/Test-InvisiblePanel
2. open the solution
3. in `Program.cs` set the start form to `FormPush`: `Application.Run(new FormPush());`
3. set the breakpoint in FormPush.Designer.cs line:577
4. run

dreddy-work:
Fixes #5774
Updating `AnchoriInfo `is only needed when control is ready for layout. `InitLayoutCore()` does call `UpdateAnchorInfo()`.
At the least, we are checking if control is parented before updating `AnchorInfo`. This helps avoid calculating
`AnchorInfo `with default initial values of the Control. They are always overridden when layout happens.

We have one other issue in the `AnchorInfo `calculations that is being investigated but this change will indirectly fix the #5774.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6778)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
@@ -3430,7 +3430,7 @@ namespace System.Windows.Forms
             get => new Size(_width, _height);
             set
             {
-                if (Anchor != CommonProperties.DefaultAnchor)
+                if (CommonProperties.GetNeedsAnchorLayout(this))
                 {
                     // Reset AnchorInfo that may have calculated based on default Size of the control or
                     // with the previous size of the control.  Especially in the designer scenario.                  .
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
@@ -3432,9 +3432,9 @@ namespace System.Windows.Forms
             {
                 if (CommonProperties.GetNeedsAnchorLayout(this))
                 {
-                    // Reset AnchorInfo that may have calculated based on default Size of the control or
+                    // Reset AnchorInfo that may have been calculated based on default Size of the control or
                     // with the previous size of the control.  Especially in the designer scenario.                  .
-                    DefaultLayout.SetAnchorInfo(this, null);
+                    DefaultLayout.SetAnchorInfo(element: this, value: null);
                 }
 
                 SetBounds(_x, _y, value.Width, value.Height, BoundsSpecified.Size);
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
@@ -3428,7 +3428,17 @@ namespace System.Windows.Forms
         public Size Size
         {
             get => new Size(_width, _height);
-            set => SetBounds(_x, _y, value.Width, value.Height, BoundsSpecified.Size);
+            set
+            {
+                if (Anchor != CommonProperties.DefaultAnchor)
+                {
+                    // Reset AnchorInfo that may have calculated based on default Size of the control or
+                    // with the previous size of the control.  Especially in the designer scenario.                  .
+                    DefaultLayout.SetAnchorInfo(this, null);
+                }
+
+                SetBounds(_x, _y, value.Width, value.Height, BoundsSpecified.Size);
+            }
         }
 
         [SRCategory(nameof(SR.CatPropertyChanged))]
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Layout/DefaultLayout.AnchorInfo.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Layout/DefaultLayout.AnchorInfo.cs
@@ -1,0 +1,20 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#nullable disable
+
+
+namespace System.Windows.Forms.Layout
+{
+    internal partial class DefaultLayout
+    {
+        internal sealed class AnchorInfo
+        {
+            public int Left;
+            public int Top;
+            public int Right;
+            public int Bottom;
+        }
+    }
+}
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Layout/DefaultLayout.AnchorInfo.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Layout/DefaultLayout.AnchorInfo.cs
@@ -4,7 +4,6 @@
 
 #nullable disable
 
-
 namespace System.Windows.Forms.Layout
 {
     internal partial class DefaultLayout
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Layout/DefaultLayout.AnchorInfo.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Layout/DefaultLayout.AnchorInfo.cs
@@ -2,6 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+#nullable disable
+
 namespace System.Windows.Forms.Layout
 {
     internal partial class DefaultLayout
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Layout/DefaultLayout.AnchorInfo.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Layout/DefaultLayout.AnchorInfo.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 namespace System.Windows.Forms.Layout
 {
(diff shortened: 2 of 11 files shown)


Tanya-Solyanik (approved):
Looks good. Could you please add a new page to the integration test?

RussKie:
I've installed the latest nightly SDK https://aka.ms/dotnet/7.0.1xx/daily/dotnet-sdk-win-x64.exe (7.0.100-preview.3.22128.6, ab332adae54d8e089fa0a9e81f30afddf63c9eb2), and retested the sample from https://github.com/dotnet/winforms/issues/5774 and I get the following results, which are somewhat surprising.

- A single monitor at 100% everything is good, no changes.
- The primary monitor is at 100% and the secondary is at 150% - the issue is not reproducible, even when the app is launched on the secondary monitor. Tested .NET 5/6/7. Looks like the app is launched with the primary monitor DPI settings.... Worth noting that I haven't signed off/on when changed the scaling factor. I wouldn't worry too much about this, but posted for completeness. 
- The primary monitor is at 150% and the secondary is at 150% - the issue is reproducible on .NET 5 and 6, but doesn't repro on .NET 7 (with and without the private build!). I guess on of the previous fixes addressed the issue, which would be great to find and service back to .NET 6.

---

The nightly SDK has an outdated version of Microsoft.NETCore.App: 7.0.0-preview.2.xxx. This will result in the sample app failures with the following error:
```
Exception Info: System.TypeLoadException: Could not load type 'System.Runtime.CompilerServices.Unsafe' from assembly 'System.Runtime, Version=7.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'.
```

To get around the issue you need to copy the latest version of Microsoft.NETCore.App workload from `.\.dotnet\shared\Microsoft.NETCore.App` to `C:\Program Files\dotnet\shared\Microsoft.NETCore.App`.

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -3436,7 +3436,17 @@ public override ISite Site
...
         {
             get => new Size(_width, _height);
-            set => SetBounds(_x, _y, value.Width, value.Height, BoundsSpecified.Size);
+            set
+            {
+                if (CommonProperties.GetNeedsAnchorLayout(this))
+                {
+                    // Reset AnchorInfo that may have calculated based on default Size of the control or
+                    // with the previous size of the control.  Especially in the designer scenario.                  .
+                    DefaultLayout.SetAnchorInfo(this, null);
```
RussKie, line 3445: Please add arg name to both parameters.

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -3436,7 +3436,17 @@ public override ISite Site
         public Size Size
         {
             get => new Size(_width, _height);
-            set => SetBounds(_x, _y, value.Width, value.Height, BoundsSpecified.Size);
+            set
+            {
+                if (CommonProperties.GetNeedsAnchorLayout(this))
+                {
+                    // Reset AnchorInfo that may have calculated based on default Size of the control or
+                    // with the previous size of the control.  Especially in the designer scenario.                  .
```
RussKie, line 3444:
```suggestion
                    // Reset AnchorInfo that may have been calculated based on default Size of the control or
                    // with the previous size of the control. Especially in the designer scenario.
```

src/System.Windows.Forms/src/System/Windows/Forms/Layout/DockAndAnchorLayout.cs
```diff
@@ -748,7 +748,10 @@ public static void SetAnchor(IArrangedElement container, IArrangedElement elemen
 
                 CommonProperties.xSetAnchor(element, value);
 
-                if (CommonProperties.GetNeedsAnchorLayout(element))
+                // Updating AnchoriInfo is only needed when control is ready for layout. InitLayoutCore() does call UpdateAnchorInfo().
+                // At the least, we are checking if control is parented before updating AnchorInfo. This helps avoid calculating
+                // AnchorInfo with default initial values of the Control. They are always overriden when layout happen.
```
RussKie, line 753: Few comments on the comment:
* I'm struggling to draw the connection between the first and the second sentences: "_Updating AnchoriInfo is only needed when control is ready for layout. InitLayoutCore() does call UpdateAnchorInfo()._". Do you think you could help me understand the connection?
* "_At the least,_" doesn't look like it fits in the context too, as it kind of implies "one of the above options", but in the code we're doing "and" operation. Perhaps "_In addition_" would work better here?
* "_They are always overriden when layout happen._" - to me "override" sounds like we're overriding a method. Were you meant to say something that has "updated" or "recalculated" meaning?

```suggestion
                // AnchorInfo only needs to be updated when the control is ready for layout. InitLayoutCore() call UpdateAnchorInfo().
                // In addition we need to check whether the control is parented before updating AnchorInfo. This helps to avoid calculating
                // AnchorInfo with the control's default initial values, as those are always overriden when layout is performed.
```

src/System.Windows.Forms/src/System/Windows/Forms/Layout/DockAndAnchorLayout.cs
```diff
@@ -1014,7 +1024,7 @@ public static bool IsAnchored(AnchorStyles anchor, AnchorStyles desiredAnchor)
             return (anchor & desiredAnchor) == desiredAnchor;
         }
 
-        private sealed class AnchorInfo
+        internal sealed class AnchorInfo
```
RussKie, line 1027: This would be good to move out to its own file. And consider if we can turn this into a struct, as it would help with perf aspects.
Not expecting these changes in the scope of this PR.

RussKie:
image

src/System.Windows.Forms/src/System/Windows/Forms/Layout/DefaultLayout.AnchorInfo.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#nullable disable
```
gpetrou, line 5: Is this needed?

dreddy-work: Thank you. it is not. Auto move into new file generated this.

src/System.Windows.Forms/src/System/Windows/Forms/Layout/DockAndAnchorLayout.cs
```diff
@@ -748,7 +748,10 @@ public static void SetAnchor(IArrangedElement container, IArrangedElement elemen
 
                 CommonProperties.xSetAnchor(element, value);
 
-                if (CommonProperties.GetNeedsAnchorLayout(element))
+                // Updating AnchoriInfo is only needed when control is ready for layout. Oneway to check this precondition is to
```
hyperquantum, line 751: Looks like a typo: `AnchoriInfo`

src/System.Windows.Forms/src/System/Windows/Forms/Layout/DockAndAnchorLayout.cs
```diff
@@ -797,8 +800,15 @@ public static void SetDock(IArrangedElement element, DockStyle value)
                         {
                             // We are transitioning from docked to not docked, restore the original bounds.
                             element.SetBounds(CommonProperties.GetSpecifiedBounds(element), BoundsSpecified.None);
-                            // Restore Anchor information as its now relevant again.
-                            UpdateAnchorInfo(element);
+
+                            // Updating AnchoriInfo is only needed when control is ready for layout. InitLayoutCore() does call UpdateAnchorInfo().
```
hyperquantum, line 804: Looks like a typo: `AnchoriInfo`

dreddy-work:
> Just to let you know: we'd be very happy to see this fixed as it would let GitExtension release its new version, which a lot of users are eagerly awaiting (myself included !). Thanks !

Thank you. Understand that. While validating this PR, we also found a related issue. I am looking if i can root cause it as well to make it complete. I am trying to get some help from native experts on why i am seeing few discrepancies with respect to bounds when moved between monitors with different DPIs. Will get this merged as soon as i can.

Alovchin91:
Hi, will this be serviced in .NET 6 as well?

PR #6782, opened 2022-03-02
Removed old constraint related to CAS.

Tanya-Solyanik:
This should had been removed when we removed CAS-related code. ControlAccessibleObject was demanding unmanaged code permissions and that is why we did not allow controls to override it. In .NET CAS is deprecated, and this type check is not relevant any more.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6782)


dreddy-work (approved):

Tanya-Solyanik:
Thank you!

PR #6793, opened 2022-03-04
Enable nullability in ToolStripControlHost

gpetrou:
## Proposed changes

- Enable nullability in ToolStripControlHost.

We seem to have many tests where we call `Dispose` and therefore `_control = null!;`. Since this is the only place where we set _control to null, what should we do? Update/remove the tests or make `_control` nullable? For example, in https://github.com/dotnet/winforms/pull/6404 we decided to use `_comStream = null!;` in Dispose.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6793)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.cs
@@ -2,9 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.ComponentModel;
+using System.Diagnostics.CodeAnalysis;
 using System.Drawing;
 using System.Runtime.CompilerServices;
 using System.Windows.Forms.Layout;
@@ -50,7 +49,8 @@ namespace System.Windows.Forms
             OnSubscribeControlEvents(c);
         }
 
-        public ToolStripControlHost(Control c, string name) : this(c)
+        public ToolStripControlHost(Control c, string name)
+            : this(c)
         {
             Name = name;
         }
@@ -68,7 +68,7 @@ namespace System.Windows.Forms
         [SRCategory(nameof(SR.CatAppearance))]
         [SRDescription(nameof(SR.ToolStripItemImageDescr))]
         [DefaultValue(null)]
-        public override Image BackgroundImage
+        public override Image? BackgroundImage
         {
             get => ControlInternal.BackgroundImage;
             set => ControlInternal.BackgroundImage = value;
@@ -135,7 +135,7 @@ namespace System.Windows.Forms
             }
         }
 
-        internal AccessibleObject ControlAccessibilityObject => _control?.AccessibilityObject;
+        internal AccessibleObject? ControlAccessibilityObject => _control?.AccessibilityObject;
 
         /// <summary>
         ///  Deriving classes can override this to configure a default size for their control.
@@ -167,7 +167,7 @@ namespace System.Windows.Forms
 
         [Browsable(false)]
         [EditorBrowsable(EditorBrowsableState.Never)]
-        public new event EventHandler DisplayStyleChanged
+        public new event EventHandler? DisplayStyleChanged
         {
             add => Events.AddHandler(s_displayStyleChangedEvent, value);
             remove => Events.RemoveHandler(s_displayStyleChangedEvent, value);
@@ -187,6 +187,7 @@ namespace System.Windows.Forms
             set => base.DoubleClickEnabled = value;
         }
 
+        [AllowNull]
         public override Font Font
         {
             get => ControlInternal.Font;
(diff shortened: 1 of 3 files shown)


RussKie:
@gpetrou it doesn't appear this change builds.

dreddy-work:
ping @gpetrou

gpetrou:
@RussKie and @dreddy-work I have updated the PR, but please check the question I had in the description. I wasn't sure about what to do here. Essentially the question is: What do we expect to happen when we first call Dispose and then a property/method of the disposed object?

Tanya-Solyanik:
>For example, in https://github.com/dotnet/winforms/pull/6404 we decided to use _comStream = null!; in Dispose.

Makes sense to follow this example as we preserve nullability checks through out the code, and access after dispose should throw.

Tanya-Solyanik:
@gpetrou  - regarding null-assignment after disposal, @dreddy-work had reminded me [here ](https://github.com/dotnet/winforms/pull/7046#discussion_r853337071) that we had discussed this in the past and decided to unify code towards the following pattern

private readonly SomeType _someField = new();

 protected override void Dispose(bool disposing)
 {
<skip>
            if ( disposing)
            {
                _someField .Dispose();
            }
}

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.StubSite.cs
```diff
@@ -40,27 +38,33 @@ public StubSite(IComponent control, IComponent host)
...
             ///  When implemented by a class, gets the container associated with the <see cref="ISite"/>.
             /// </summary>
-            IContainer ISite.Container => _owner.Site.Container;
+            IContainer? ISite.Container => _owner.Site?.Container;
 
             /// <summary>
             ///  When implemented by a class, determines whether the component is in design mode.
             /// </summary>
-            bool ISite.DesignMode => _owner.Site.DesignMode;
+            bool ISite.DesignMode => _owner.Site is null ? false : _owner.Site.DesignMode;
```
vladimir-krestov, line 46: Consider this way, like we used to do.
```suggestion
            bool ISite.DesignMode => _owner.Site?.DesignMode ?? false;
```

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.cs
```diff
@@ -132,7 +130,7 @@ public ContentAlignment ControlAlign
         [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
         public Control Control => _control;
 
-        internal AccessibleObject ControlAccessibilityObject => Control?.AccessibilityObject;
+        internal AccessibleObject? ControlAccessibilityObject => Control?.AccessibilityObject;
```
vladimir-krestov, line 133: Why if Control can't be null here? Look at line 36

dreddy-work:
@gpetrou , are you still waiting for any input from us here?

gpetrou:
> @gpetrou , are you still waiting for any input from us here?

I guess I am still puzzled about using `null!` in `Dispose`. In https://github.com/dotnet/winforms/pull/7046#discussion_r853337071 we ended up keeping it. Should I keep it here as well?

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.cs
gpetrou: If I remember correctly, this is related to using `null!` in `Dispose` or not.

RussKie: I think @vladimir-krestov has a point here - line:133 `public Control Control => _control;`, which means `Control` is never a null.
So depending on whether `_control.AccessibilityObject` can be null or not this should looks either:
```suggestion
        internal AccessibleObject? ControlAccessibilityObject => Control.AccessibilityObject;
```
...or
```suggestion
        internal AccessibleObject ControlAccessibilityObject => Control.AccessibilityObject;
```

RussKie (commented):
Overall 👍

RussKie (dismissed):

RussKie:
Some tests fail, e.g. `ToolStripControlHost_CreateAccessibilityInstance_InvokeDisposed_ReturnsExpected`.

gpetrou:
@dreddy-work any feedback on this one?
BTW, it looks like the bot does not remove the `draft` label even though I set it to ready for review.

dreddy-work:
> BTW, it looks like the bot does not remove the `draft` label even though I set it to ready for review.

I will have to check bot if we added such action.

dreddy-work:
Final change LGTM.

dreddy-work (dismissed):

gpetrou:
@dreddy-work could we merge this now?

dreddy-work (approved):

PR #6802, opened 2022-03-07
Enable additional C/C++ compiler warnings

### LINKED ISSUE #6786: SDL signoff requirements - please enable additional C/C++ compiler warnings
> __Please prioritize this work.__ Depending on the size of the code base, there could be hundreds or thousands of warnings which require addressing.
> 
> __This work must be completed before RC1 snap.__

Recent SDL changes require that we enable additional compiler warnings across our C/C++ projects. The requirements are:

* Compile with warning level 3 or higher. These warnings do not have to be fixed, but they must at least be triaged.
* Enable the following additional warnings, which __must__ be fixed if they occur:
  * __C4018__ - 'expression' : signed/unsigned mismatch
  * __C4055__ - 'conversion' : from data pointer 'type1' to function pointer 'type2'
  * __C4146__ - unary minus operator applied to unsigned type, result still unsigned
  * __C4242__ - 'identifier' : conversion from 'type1' to 'type2', possible loss of data
  * __C4244__ - 'conversion' conversion from 'type1' to 'type2', possible loss of data
  * __C4267__ - 'var' : conversion from 'size_t' to 'type', possible loss of data
  * __C4302__ - 'conversion' : truncation from 'type 1' to 'type 2'
  * __C4308__ - negative integral constant converted to unsigned type
  * __C4509__ - nonstandard extension used: 'function' uses SEH and 'object' has destructor
  * __C4510__ - 'class' : default constructor could not be generated
  * __C4532__ - 'continue' : jump out of __finally/finally block has undefined behavior during termination handling
  * __C4533__ - initialization of 'variable' is skipped by 'instruction'
  * __C4610__ - object 'class' can never be instantiated - user-defined constructor required
  * __C4611__ - interaction between 'function' and C++ object destruction is non-portable
  * __C4700__ - uninitialized local variable 'name' used
  * __C4701__ - Potentially uninitialized local variable 'name' used
  * __C4703__ - Potentially uninitialized local pointer variable 'name' used
  * __C4789__ - destination of memory copy is too small
  * __C4995__ - 'function': name was marked as #pragma deprecated
  * __C4996__ - 'function': was declared deprecated also 'std::': Function call with parameters that are potentially unsafe - this call relies on the caller to check that the passed values are correct. To disable this warning, use -D_SCL_SECURE_NO_WARNINGS. See documentation on how to use Visual C++ 'Checked Iterators'

See [the SDL guidelines](https://liquid.microsoft.com/Web/Object/Read/ms.security/Requirements/Microsoft.Security.SystemsADM.10086) (MSFT internal only) for further information.

To make auditing code easier prior to signoff and release, **please include the string _Microsoft.Security.SystemsADM.10086_** somewhere in the vcxproj / cmake / whatever file which is responsible for C/C++ compilation where all of these flags are defined. That gives the code auditors a target string to search against and validate that the proper warnings are enabled. See below an example of how to do this in a cmake file.

```cmake
# [[! Microsoft.Security.SystemsADM.10086 !]] - SQL required warnings
add_compile_options($<$<COMPILE_LANGUAGE:C,CXX>:/W3>) # warning level 3
add_compile_options($<$<COMPILE_LANGUAGE:C,CXX>:/WX>) # treat warnings as errors
add_compile_options($<$<COMPILE_LANGUAGE:C,CXX>:/we4018>) # 'expression' : signed/unsigned mismatch
add_compile_options($<$<COMPILE_LANGUAGE:C,CXX>:/we4055>) # 'conversion' : from data pointer 'type1' to function pointer 'type2'
# ... and so on ...
```

Once the work is finished, or if there is no work to perform, please feel free to close this issue.

Thanks for your assistance!

## Quick FAQ

__What code is bound to this requirement?__

This affects only _production_ code. Production code is generally defined as code which ships as part of the product and which runs on customer machines or which manages infrastructure, such as our build labs. Unit and functional test projects are not considered production code.

__Does this need to be backported?__

No backporting plans at this time. If actual bugs are found during this process, individual product teams have discretion to selectively backport into the next downlevel servicing vehicle.

__What about forked third-party code?__

This requirement applies to all code that MSFT builds from source, regardless of its provenance. Ideally any changes that we make to local forked copies can be submitted upstream as a PR so that the wider ecosystem can enjoy their benefits.

If this is impractical, exceptions to this requirement can be sought on an as-needed basis. However, exceptions are: (a) not guaranteed to be granted; and (b) time-constrained. The exception process is not intended to provide a permanent deferral of this work. Please contact the _fxsecurity_ alias if an exception is needed.

__What about C# and other languages?__

This requirement only affects C/C++ code. Requirements for other languages will be filed as separate issues.

RussKie:
Resolves #6786

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6802)

### DIFF
--- a/src/System.Windows.Forms/tests/IntegrationTests/NativeHost/NativeHost.vcxproj
+++ b/src/System.Windows.Forms/tests/IntegrationTests/NativeHost/NativeHost.vcxproj
@@ -110,12 +110,14 @@
   </ItemDefinitionGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
     <ClCompile>
-      <WarningLevel>Level3</WarningLevel>
+      <WarningLevel>Level4</WarningLevel>
       <SDLCheck>true</SDLCheck>
       <PreprocessorDefinitions>_DEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <ConformanceMode>true</ConformanceMode>
       <PrecompiledHeader>Use</PrecompiledHeader>
       <PrecompiledHeaderFile>pch.h</PrecompiledHeaderFile>
+      <TreatWarningAsError>true</TreatWarningAsError>
+      <DisableSpecificWarnings>4100</DisableSpecificWarnings>
     </ClCompile>
     <Link>
       <SubSystem>Windows</SubSystem>
--- a/src/System.Windows.Forms/tests/InteropTests/NativeTests/CMakeLists.txt
+++ b/src/System.Windows.Forms/tests/InteropTests/NativeTests/CMakeLists.txt
@@ -4,6 +4,14 @@ set(CMAKE_MACOSX_RPATH 1)
 set(CMAKE_CXX_STANDARD 11)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
+
+# [[! Microsoft.Security.SystemsADM.10086 !]] - SQL required warnings
+add_compile_options($<$<COMPILE_LANGUAGE:C,CXX>:/W3>) # warning level 3
+add_compile_options($<$<COMPILE_LANGUAGE:C,CXX>:/WX>) # treat warnings as errors
+add_compile_options($<$<COMPILE_LANGUAGE:C,CXX>:/we4018>) # 'expression' : signed/unsigned mismatch
+add_compile_options($<$<COMPILE_LANGUAGE:C,CXX>:/we4055>) # 'conversion' : from data pointer 'type1' to function pointer 'type2'
+
+
 add_library(NativeTests SHARED
     AccessibleObjectTests.cpp
     WebBrowserSiteBaseInteropTests.cpp


src/System.Windows.Forms/tests/IntegrationTests/NativeHost/NativeHost.vcxproj
```diff
@@ -110,12 +110,14 @@
...
     <ClCompile>
-      <WarningLevel>Level3</WarningLevel>
+      <WarningLevel>Level4</WarningLevel>
       <SDLCheck>true</SDLCheck>
       <PreprocessorDefinitions>_DEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <ConformanceMode>true</ConformanceMode>
       <PrecompiledHeader>Use</PrecompiledHeader>
       <PrecompiledHeaderFile>pch.h</PrecompiledHeaderFile>
+      <TreatWarningAsError>true</TreatWarningAsError>
+      <DisableSpecificWarnings>4100</DisableSpecificWarnings>
```
dreddy-work, line 120: what is this for?

dreddy-work (approved):

src/System.Windows.Forms/tests/IntegrationTests/NativeHost/NativeHost.vcxproj
RussKie: [C4100](https://docs.microsoft.com/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4100): The formal parameter is not referenced in the body of the function. 

Suppressing to stop it complaining about unused arguments in these methods:
https://github.com/dotnet/winforms/blob/190a0b615c3b3a3e9f5ae7e5402eb4acc9c0fabe/src/System.Windows.Forms/tests/IntegrationTests/NativeHost/NativeHost.cpp#L41
https://github.com/dotnet/winforms/blob/190a0b615c3b3a3e9f5ae7e5402eb4acc9c0fabe/src/System.Windows.Forms/tests/IntegrationTests/NativeHost/NativeHost.cpp#L47

PR #6835, opened 2022-03-12
Enable nullability in ListBox

gpetrou:
## Proposed changes

- Enable nullability in ListBox.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6835)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs
@@ -57,58 +57,59 @@ namespace System.Windows.Forms
         private static readonly object EVENT_DRAWITEM = new object();
         private static readonly object EVENT_MEASUREITEM = new object();
 
-        SelectedObjectCollection? selectedItems;
-        SelectedIndexCollection? selectedIndices;
-        ObjectCollection? itemsCollection;
+        private SelectedObjectCollection? _selectedItems;
+        private SelectedIndexCollection? _selectedIndices;
+        private ObjectCollection? _itemsCollection;
 
-        int itemHeight = DefaultItemHeight;
-        int columnWidth;
-        int requestedHeight;
-        int topIndex;
-        int horizontalExtent;
-        int maxWidth = -1;
-        int updateCount;
+        private int _itemHeight = DefaultItemHeight;
+        private int _columnWidth;
+        private int _requestedHeight;
+        private int _topIndex;
+        private int _horizontalExtent;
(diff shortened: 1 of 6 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs
```diff
@@ -59,58 +57,59 @@ public partial class ListBox : ListControl
...
+        private bool _integralHeightAdjust;
+        private bool _multiColumn;
+        private bool _horizontalScrollbar;
+        private bool _useTabStops = true;
+        private bool _useCustomTabOffsets;
+        private bool _fontIsChanged;
+        private bool _doubleClickFired;
+        private bool _selectedValueChangedFired;
+
+        private DrawMode drawMode = DrawMode.Normal;
```
dreddy-work, line 84: why were these not renamed given those are also private?

src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs
```diff
@@ -1339,7 +1334,7 @@ protected virtual ObjectCollection CreateItemCollection()
         internal virtual int ComputeMaxItemWidth(int oldMax)
         {
             // pass LayoutUtils the collection of strings
-            string[] strings = new string[Items.Count];
+            string?[] strings = new string[Items.Count];
```
dreddy-work, line 1337: I see other instances of `GetItemText` could return null. why do you make this change?

dreddy-work, line 1337: I see other instances of `GetItemText` could return null. why do you make this change?

gpetrou: Because `GetItemText` returns a nullable string. So, the array should contain nullable strings.

gpetrou: Same reply as in my other comment.

RussKie (approved):

src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs
dreddy-work: i meant to say there were other instances that may need nullable . I looked again but not seen any. I was mistaken.

dreddy-work (approved):

PR #6853, opened 2022-03-16
Fixing MonthCalendar accessible object leaks

### LINKED ISSUE #6852: MonthCalendar accessible objects are leaking in memory
* .NET Core Version: 7.0

**Problem description:**
MonthCalendar accessible objects are left in memory after the MonthCalendar control instance was inspected with accessibility tool (like Accessibility Insights or Inspect) and the form with MonthCalendar is closed, garbage is collected, but accessibility tool is still opened.

Report from WinDbg:
image
 
**Expected behavior:**

**Minimal repro:**
1. Create a parent form with two buttons: one will be opening a child form with the MonthCalendar control and the other will collect the garbage.
2. Press the second button to open the child form with MonthCalendar.
4. Run automated checks for the MonthCalendar control using AI.
5. Close the child form.
7. DON'T close the AI tool.
8. Look for MonthCalendar accessible objects left in memory (using WinDbg, or dotMemory, or another suitable tool).

ArtemTatarinov:
Fixes #6852

## Proposed changes
- Use the `UiaDisconnectProvider` method cascade through the hierarchy of the classes to free objects held by accessibility tool
- Only static comparers are left at memory and their amount aren't increasing after several recurrences of the repro steps described in issue

## Customer Impact
- MonthCalendar memory will stop leaking

## Regression? 
- No

## Risk
- Minimal

### Before
image

### After
image

## Test methodology
- Manual testing (_using WinDbg, see the screenshots above_)
- CTI team

## Test environment(s)
- Microsoft Windows [Version 10.0.19043.1586]
- .NET 7.0.0-preview.3.22159.12

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6853)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/MonthCalendar.CalendarAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/MonthCalendar.CalendarAccessibleObject.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Diagnostics;
 using System.Drawing;
 using static Interop;
 using static Interop.ComCtl32;
@@ -36,6 +37,23 @@ namespace System.Windows.Forms
                 // otherwise the calendar accessibility tree will be rebuilt.
                 // So save this value one time to avoid sending messages to Windows every time.
                 _initName = initName;
+            }
+
+            internal void DisconnectChildren()
+            {
+                Debug.Assert(OsVersion.IsWindows8OrGreater);
+                if (_calendarHeaderAccessibleObject is not null)
+                {
+                    HRESULT result = UiaCore.UiaDisconnectProvider(_calendarHeaderAccessibleObject);
+                    Debug.Assert(result == 0);
+                }
+
+                if (_calendarBodyAccessibleObject is not null)
+                {
+                    _calendarBodyAccessibleObject.DisconnectChildren();
+                    HRESULT result = UiaCore.UiaDisconnectProvider(_calendarBodyAccessibleObject);
+                    Debug.Assert(result == 0);
+                }
             }
 
             public override Rectangle Bounds
--- a/src/System.Windows.Forms/src/System/Windows/Forms/MonthCalendar.CalendarRowAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/MonthCalendar.CalendarRowAccessibleObject.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Diagnostics;
 using System.Drawing;
 using System.Globalization;
 using static Interop;
@@ -96,6 +97,27 @@ namespace System.Windows.Forms
             }
 
             internal void ClearChildCollection() => _cellsAccessibleObjects = null;
+
+            internal void DisconnectChildren()
+            {
+                Debug.Assert(OsVersion.IsWindows8OrGreater);
+                if (_weekNumberCellAccessibleObject is not null)
+                {
+                    HRESULT result = UiaCore.UiaDisconnectProvider(_weekNumberCellAccessibleObject);
+                    Debug.Assert(result == 0);
+                }
+
+                if (_cellsAccessibleObjects is null)
+                {
+                    return;
+                }
+
+                foreach (CalendarCellAccessibleObject cell in _cellsAccessibleObjects)
+                {
+                    HRESULT result = UiaCore.UiaDisconnectProvider(cell);
+                    Debug.Assert(result == 0);
+                }
+            }
 
             public override string? Description
             {
--- a/src/System.Windows.Forms/src/System/Windows/Forms/MonthCalendar.MonthCalendarAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/MonthCalendar.MonthCalendarAccessibleObject.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Diagnostics;
 using System.Drawing;
 using static Interop;
 using static Interop.ComCtl32;
@@ -76,6 +77,47 @@ namespace System.Windows.Forms
                     }
 
                     return _calendarsAccessibleObjects;
+                }
+            }
+
+            // This function should be called from a single place in the root of MonthCalendar object that already tests for availability of this API
+            internal void DisconnectChildren()
+            {
+                Debug.Assert(OsVersion.IsWindows8OrGreater);
+                if (_previousButtonAccessibleObject is not null)
+                {
+                    HRESULT result = UiaCore.UiaDisconnectProvider(_previousButtonAccessibleObject);
+                    Debug.Assert(result == 0);
+                }
+
+                if (_nextButtonAccessibleObject is not null)
+                {
+                    HRESULT result = UiaCore.UiaDisconnectProvider(_nextButtonAccessibleObject);
+                    Debug.Assert(result == 0);
+                }
+
+                if (_todayLinkAccessibleObject is not null)
+                {
+                    HRESULT result = UiaCore.UiaDisconnectProvider(_todayLinkAccessibleObject);
+                    Debug.Assert(result == 0);
+                }
+
+                if (_focusedCellAccessibleObject is not null)
+                {
+                    HRESULT result = UiaCore.UiaDisconnectProvider(_focusedCellAccessibleObject);
+                    Debug.Assert(result == 0);
+                }
+
+                if (_calendarsAccessibleObjects is null)
+                {
+                    return;
+                }
+
+                foreach (CalendarAccessibleObject calendarAccessibleObject in _calendarsAccessibleObjects)
+                {
+                    calendarAccessibleObject.DisconnectChildren();
+                    HRESULT result = UiaCore.UiaDisconnectProvider(calendarAccessibleObject);
+                    Debug.Assert(result == 0);
                 }
             }
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/MonthCalendar.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/MonthCalendar.cs
@@ -2325,6 +2325,19 @@ namespace System.Windows.Forms
                     break;
                 case User32.WM.DESTROY:
                     base.WndProc(ref m);
+                    if (IsHandleCreated && IsAccessibilityObjectCreated)
+                    {
+                        UiaCore.UiaReturnRawElementProvider(Handle, wParam: IntPtr.Zero, lParam: IntPtr.Zero, el: null);
+
+                        if (OsVersion.IsWindows8OrGreater)
+                        {
+                            ((MonthCalendarAccessibleObject)AccessibilityObject).DisconnectChildren();
+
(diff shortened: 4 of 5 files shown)


src/System.Windows.Forms/src/misc/ApiHelper.cs
```diff
@@ -0,0 +1,54 @@
...
+
+using System.Collections.Concurrent;
+using static Interop;
+
+namespace System.Windows.Forms
+{
+    /// <summary>
+    /// Internal APiHelper class. Helps identify if API is avaialble at runtime.
+    /// </summary>
+    internal static class ApiHelper
```
Tanya-Solyanik, line 13: .NET Core removed this class. Instead, it goes with the documented versions of windows where the API should be available. Search for previous call sites of UIADisconnectProvider for details.

src/System.Windows.Forms/src/System/Windows/Forms/MonthCalendar.MonthCalendarAccessibleObject.cs
```diff
@@ -79,6 +80,47 @@ public MonthCalendarAccessibleObject(MonthCalendar owner) : base(owner)
...
+                    Debug.Assert(result == 0);
+                }
+
+                if (_focusedCellAccessibleObject is not null)
+                {
+                    HRESULT result = UiaCore.UiaDisconnectProvider(_focusedCellAccessibleObject);
+                    Debug.Assert(result == 0);
+                }
+
+                if (_calendarsAccessibleObjects == null)
```
Tanya-Solyanik, line 111: nit: `is null`

src/System.Windows.Forms/src/System/Windows/Forms/MonthCalendar.MonthCalendarAccessibleObject.cs
```diff
@@ -79,6 +80,47 @@ public MonthCalendarAccessibleObject(MonthCalendar owner) : base(owner)
                 }
             }
 
+            // This function should be called from a single place in the root of MonthCalendar object that already tests for availability of this API
+            internal void DisconnectChildren()
+            {
+                Debug.Assert(ApiHelper.IsApiAvailable(Libraries.UiaCore, nameof(UiaCore.UiaDisconnectProvider)));
+                if (_previousButtonAccessibleObject is not null)
```
Tanya-Solyanik, line 87: Is it possible to test if provider had been created for this object? Had it been cached?

Tanya-Solyanik (commented):
Thank you, I added some comments.

src/System.Windows.Forms/src/System/Windows/Forms/MonthCalendar.MonthCalendarAccessibleObject.cs
ArtemTatarinov: I've tried different approaches, but, unfortunately, couldn't find the way to determine if provider is created for children accessible objects yet. I've looked through the UIA methods here (https://docs.microsoft.com/en-us/windows/win32/api/uiautomationcoreapi/) to find some kind of symmetrical method for `UiaDisconnectProvider`, something which connects or creates a provider, but no luck. The closest one, I guess, is `UiaReturnRawElementProvider` which is called at the `WmGetObject` method at the `Control` class, but it runs under condition `m.LParamInternal == NativeMethods.UiaRootObjectId`, so it looks like quite similar to the `IsAccessibilityObjectCreated` flag itself - for parent objects only.

vladimir-krestov: @Tanya-Solyanik, we don't create providers for subitems of MonthCalendar. These objects are just parts of MonthCalendarAccessibleObject. We don't have controls for these parts, we just get some info from native MonthCanedar and give it to UIA. So only MonthCalendar AO has a provider and we need to check if these objects are `null` only, like Artem did. So I think, we should keep it as it is made now.

src/System.Windows.Forms/src/System/Windows/Forms/MonthCalendar.CalendarRowAccessibleObject.cs
```diff
@@ -97,6 +98,27 @@ public override Rectangle Bounds
...
+                    HRESULT result = UiaCore.UiaDisconnectProvider(_weekNumberCellAccessibleObject);
+                    Debug.Assert(result == 0);
+                }
+
+                if (_cellsAccessibleObjects == null)
+                {
+                    return;
+                }
+
+                foreach (var cell in _cellsAccessibleObjects)
```
vladimir-krestov, line 115: Please explicit `var`

src/System.Windows.Forms/src/System/Windows/Forms/MonthCalendar.CalendarAccessibleObject.cs
```diff
@@ -38,6 +39,26 @@ public CalendarAccessibleObject(MonthCalendarAccessibleObject calendarAccessible
...
+                if (_calendarHeaderAccessibleObject is not null)
+                {
+                    HRESULT result = UiaCore.UiaDisconnectProvider(_calendarHeaderAccessibleObject);
+                    Debug.Assert(result == 0);
+                }
+
+                if (_calendarBodyAccessibleObject is not null)
+                {
+                    _calendarBodyAccessibleObject.DisconnectChildren();
+                    if (_calendarBodyAccessibleObject is not null)
```
vladimir-krestov, line 54: Why do we need to check `null` one more time?

vladimir-krestov (commented):
Looks good to me ✔️

src/System.Windows.Forms/src/System/Windows/Forms/MonthCalendar.MonthCalendarAccessibleObject.cs
Tanya-Solyanik: Makes sense, null check is sufficient

Tanya-Solyanik:
Please make sure this is tested for regressions. Please use the accessibility test application.

src/System.Windows.Forms/src/System/Windows/Forms/MonthCalendar.CalendarRowAccessibleObject.cs
```diff
@@ -97,6 +98,27 @@ public override Rectangle Bounds
...
+            internal void DisconnectChildren()
+            {
+                Debug.Assert(OsVersion.IsWindows8OrGreater);
+                if (_weekNumberCellAccessibleObject is not null)
+                {
+                    HRESULT result = UiaCore.UiaDisconnectProvider(_weekNumberCellAccessibleObject);
+                    Debug.Assert(result == 0);
+                }
+
+                if (_cellsAccessibleObjects == null)
```
Tanya-Solyanik, line 110: `is null` ?

src/System.Windows.Forms/src/System/Windows/Forms/MonthCalendar.cs
```diff
@@ -2325,6 +2325,22 @@ protected override void WndProc(ref Message m)
                     break;
                 case User32.WM.DESTROY:
                     base.WndProc(ref m);
+                    if (IsAccessibilityObjectCreated)
+                    {
+                        if (IsHandleCreated)
+                        {
+                            UiaCore.UiaReturnRawElementProvider(Handle, IntPtr.Zero, IntPtr.Zero, null);
```
Tanya-Solyanik, line 2332: Please add parameter names in front of the literals.

src/System.Windows.Forms/src/System/Windows/Forms/MonthCalendar.cs
```diff
@@ -2325,6 +2325,22 @@ protected override void WndProc(ref Message m)
                     break;
                 case User32.WM.DESTROY:
                     base.WndProc(ref m);
+                    if (IsAccessibilityObjectCreated)
+                    {
+                        if (IsHandleCreated)
```
Tanya-Solyanik, line 2330: Shouldn't we exit is handle is not created?

ArtemTatarinov:
> Please make sure this is tested for regressions. Please use the accessibility test application.

Tested different cases of using of MonthCalendar with Accessibility Insights turned on, including autochecks, closing the form with MonthCalendar or its children controls focused in AI, etc, everything seems fine. Also, I hope the CTI team will check for regressions too.

vladimir-krestov (approved):
Approved ✔️

src/System.Windows.Forms/src/System/Windows/Forms/MonthCalendar.CalendarAccessibleObject.cs
```diff
@@ -38,6 +39,23 @@ public CalendarAccessibleObject(MonthCalendarAccessibleObject calendarAccessible
                 _initName = initName;
             }
 
+            internal void DisconnectChildren()
+            {
+                Debug.Assert(OsVersion.IsWindows8OrGreater);
+                if (_calendarHeaderAccessibleObject is not null)
+                {
+                    HRESULT result = UiaCore.UiaDisconnectProvider(_calendarHeaderAccessibleObject);
+                    Debug.Assert(result == 0);
```
RussKie, line 48: I don't think it's correct to compare a HRESULT with 0 because success is indicated by a non-negative value. We have extensions to check success or failure:
https://github.com/dotnet/winforms/blob/1c6b1978d1c40314d1f4cf1da661f8b8ccf68d18/src/System.Windows.Forms.Primitives/src/Interop/Interop.HRESULT.cs#L83-L92

RussKie (commented):
👍 with a comment.

Please holler when the testing is done.

src/System.Windows.Forms/src/System/Windows/Forms/MonthCalendar.CalendarAccessibleObject.cs
Tanya-Solyanik: We want to Assert on S_FALSE, only S_OK is expected.

Tanya-Solyanik:
@Olina-Zhang  - Could you please test this change?

Tanya-Solyanik (approved):

John-Qiao:
@Tanya-Solyanik we have Tested the built private binaries for GH pull: [6853](https://github.com/dotnet/winforms/pull/6853) based on latest .NET 7.0 SDK, MonthCalendar control works well and no new issue found. All the migrated automation cases are passed too.

src/System.Windows.Forms/src/System/Windows/Forms/MonthCalendar.CalendarAccessibleObject.cs
```diff
@@ -38,6 +39,23 @@ public CalendarAccessibleObject(MonthCalendarAccessibleObject calendarAccessible
                 _initName = initName;
             }
 
+            internal void DisconnectChildren()
+            {
+                Debug.Assert(OsVersion.IsWindows8OrGreater);
```
RussKie, line 44: What benefits do we get from this assertion? We still support W7SP1.

Tanya-Solyanik: We should not invoke DisconnectChildren on win7.

RussKie: Asserts don't prevent us from doing this.

Tanya-Solyanik: of cause not, this is only a documentation

PR #4554, opened 2021-02-11
Fix IPerPropertyBrowsing on 64 bit

JeremyKuhne:
IPerPropertyBrowsing.GetPredefinedStrings would fail on 64bit as it would try to marshal an array of integers as IntPtrs.

The first commit makes a more targeted change. The code was terribly inefficient, so the 3rd commit simplifies the marshalling significantly. 2nd commit is just some style clean in prep for the third.

Please do not squash the PR so we can reference the fix for the original marshalers if we need to port to .NET Framework.

Also:

- Ordered HRESULTs
- Add test helpers to validate COM memory handling

We don't have a great way to test this end-to-end as there aren't COM Controls readily available on a standard Win10 box that implement this API (afaik).

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/4554)

PR #6866, opened 2022-03-18
Remove not used interfaces

kant2002:
- IOmNavigator and IHTMLScreen do not used
- IMallocSpy belongs to test project and as such moved there

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6866)


src/System.Windows.Forms.Primitives/tests/TestUtilities/IMallocSpy.cs
```diff
@@ -0,0 +1,71 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
dreddy-work, line 1: if this is used only in tests, why/what are the tests for?

kant2002: For tracking memory allocations. Follow the link for locations where actual usage happens https://github.com/dotnet/winforms/search?q=MallocSpyScope

dreddy-work:
👍

src/System.Windows.Forms.Primitives/tests/TestUtilities/IMallocSpy.cs
RussKie: It was added in https://github.com/dotnet/winforms/pull/4554.

RussKie (approved):

PR #6870, opened 2022-03-18
Enable nullability in FontDialog

gpetrou:
## Proposed changes

- Enable nullability in FontDialog.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6870)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/FontDialog.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/FontDialog.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.ComponentModel;
 using System.Diagnostics;
 using System.Drawing;
@@ -29,7 +27,7 @@ namespace System.Windows.Forms
         private const int defaultMaxSize = 0;
 
         private Comdlg32.CF options;
-        private Font font;
+        private Font? font;
         private Color color;
         private int minSize = defaultMinSize;
         private int maxSize = defaultMaxSize;
@@ -155,7 +153,7 @@ namespace System.Windows.Forms
         {
             get
             {
-                Font result = font;
+                Font? result = font;
                 if (result is null)
                 {
                     result = Control.DefaultFont;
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/FontDialog.cs
```diff
@@ -155,7 +153,7 @@ public Font Font
         {
             get
             {
-                Font result = font;
+                Font? result = font;
```
dreddy-work, line 156:
```suggestion
                Font? result = font ?? Control.DefaultFont;
```

src/System.Windows.Forms/src/System/Windows/Forms/FontDialog.cs
```diff
@@ -155,11 +153,7 @@ public Font Font
         {
             get
             {
-                Font result = font;
-                if (result is null)
-                {
-                    result = Control.DefaultFont;
-                }
+                Font? result = font ?? Control.DefaultFont;
```
RussKie, line 156: `Control.DefaultFont` can't be null
https://github.com/dotnet/winforms/blob/09120ef4871dccfa3b4c09e33ad96648cf3f4c64/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs#L1811-L1823
```suggestion
                Font result = font ?? Control.DefaultFont;
```

gpetrou: `SystemFonts.MessageBoxFont` is nullable though. I guess when we enable nullability in Control we could use the null-forgiving operator there.

RussKie: Ah, right. All good then

src/System.Windows.Forms/src/System/Windows/Forms/FontDialog.cs
```diff
@@ -155,11 +153,7 @@ public Font Font
         {
             get
             {
-                Font result = font;
-                if (result is null)
-                {
-                    result = Control.DefaultFont;
-                }
+                Font result = font ?? Control.DefaultFont;
```
RussKie, line 156:
```suggestion
                Font? result = font ?? Control.DefaultFont;
```

RussKie (approved):

PR #6901, opened 2022-03-25
Partially enable nullability in Control

gpetrou:
## Proposed changes

- Partially enable nullability in Control.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6901)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/Interop/UiaCore/Interop.UiaReturnRawElementProvider.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/UiaCore/Interop.UiaReturnRawElementProvider.cs
@@ -9,9 +9,9 @@ internal partial class Interop
     internal static partial class UiaCore
     {
         [DllImport(Libraries.UiaCore, CharSet = CharSet.Unicode)]
-        public static extern nint UiaReturnRawElementProvider(IntPtr hwnd, nint wParam, nint lParam, IRawElementProviderSimple el);
+        public static extern nint UiaReturnRawElementProvider(IntPtr hwnd, nint wParam, nint lParam, IRawElementProviderSimple? el);
 
-        public static nint UiaReturnRawElementProvider(IHandle hwnd, nint wParam, nint lParam, IRawElementProviderSimple el)
+        public static nint UiaReturnRawElementProvider(IHandle hwnd, nint wParam, nint lParam, IRawElementProviderSimple? el)
         {
             nint result = UiaReturnRawElementProvider(hwnd.Handle, wParam, lParam, el);
             GC.KeepAlive(hwnd);
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
@@ -8,6 +8,7 @@ using System.ComponentModel;
 using System.ComponentModel.Design;
 using System.ComponentModel.Design.Serialization;
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using System.Drawing;
 using System.Globalization;
 using System.Runtime.CompilerServices;
@@ -9722,7 +9723,7 @@ namespace System.Windows.Forms
         {
             ((PaintEventHandler)Events[s_paintEvent])?.Invoke(this, e);
         }
-
+#nullable enable
         private void RemovePendingMessages(User32.WM msgMin, User32.WM msgMax)
         {
             if (!IsDisposed)
@@ -9849,7 +9850,7 @@ namespace System.Windows.Forms
 
                 HandleRef parentHandle = new HandleRef(this, User32.GetParent(this));
 
-                Control[] controlSnapshot = null;
+                Control?[]? controlSnapshot = null;
                 SetState(States.Recreate, true);
 
                 try
@@ -9861,7 +9862,7 @@ namespace System.Windows.Forms
                     // This behavior can be overridden in OnParentHandleRecreat* and is in ListView.
 
                     //fish out control collection w/o demand creating one.
-                    ControlCollection controlsCollection = (ControlCollection)Properties.GetObject(s_controlsCollectionProperty);
+                    ControlCollection? controlsCollection = (ControlCollection?)Properties.GetObject(s_controlsCollectionProperty);
                     if (controlsCollection is not null && controlsCollection.Count > 0)
                     {
                         controlSnapshot = new Control[controlsCollection.Count];
@@ -9926,7 +9927,7 @@ namespace System.Windows.Forms
                     {
                         for (int i = 0; i < controlSnapshot.Length; i++)
                         {
-                            Control childControl = controlSnapshot[i];
+                            Control? childControl = controlSnapshot[i];
                             if (childControl is not null && childControl.IsHandleCreated)
                             {
                                 // Re-parent the control.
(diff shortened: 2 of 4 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -8156,15 +8157,15 @@ protected virtual void OnGotFocus(EventArgs e)
 
             ((EventHandler)Events[s_gotFocusEvent])?.Invoke(this, e);
         }
-
+#nullable enable
```
gpetrou, line 8160: There are many changes when annotating the entire file, so I decided to split them in order to make the review easier.

src/System.Windows.Forms.Primitives/src/Interop/UiaCore/Interop.UiaReturnRawElementProvider.cs
```diff
@@ -9,9 +9,9 @@ internal partial class Interop
     internal static partial class UiaCore
     {
         [DllImport(Libraries.UiaCore, CharSet = CharSet.Unicode)]
-        public static extern nint UiaReturnRawElementProvider(IntPtr hwnd, nint wParam, nint lParam, IRawElementProviderSimple el);
+        public static extern nint UiaReturnRawElementProvider(IntPtr hwnd, nint wParam, nint lParam, IRawElementProviderSimple? el);
```
gpetrou, line 12: Shouldn't we have `nuint  wParam` instead? Perhaps we could define LRESULT, LPARAM, WPARAM, etc... structs that contain the uint/nuint values, to make it easier to read the DllImports. I have not checked the performance impact though.

RussKie: This is a good observation. According to the [interop best practices](https://docs.microsoft.com/en-us/dotnet/standard/native-interop/best-practices#common-windows-data-types), `WPARAM` maps to `UIntPtr`. Though in our codebase we've been annotating it as `IntPtr`.
I think we should go and fix this holistically as a separate piece of work.

gpetrou:
> @gpetrou , this need more attention and may take little longer to review/merge. Just heads-up.

@dreddy-work perhaps I should revert some of the changes to make this easier to review?

gpetrou:
I reduced the amount of changes. Hopefully, this is easier to review now.

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -10166,7 +10167,7 @@ internal void SetAcceptDrops(bool accept)
                         Debug.WriteLineIf(CompModSwitches.DragDrop.TraceInfo, "   ret:" + n.ToString(CultureInfo.CurrentCulture));
                         if (n != HRESULT.S_OK && n != HRESULT.DRAGDROP_E_ALREADYREGISTERED)
                         {
-                            throw Marshal.GetExceptionForHR((int)n);
+                            throw Marshal.GetExceptionForHR((int)n)!;
```
dreddy-work, line 10170: We don't know this. It could be `null` ex: S_FALSE.  We cannot use   `ThrowExceptionForHR` either.  We may need to customize in such cases here. @Tanya-Solyanik  any suggestions?

dreddy-work:
> I reduced the amount of changes. Hopefully, this is easier to review now.

Thanks @gpetrou . This is much better.

dreddy-work:
@Olina-Zhang can you run a sanity check for our runtime tests with these privates once `ready-for-merge`?

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
Tanya-Solyanik: Special-case S_FALSE.

dreddy-work: @gpetrou  can you see if this is the only constant that could thow null here? or any positive value end up throwing null?

gpetrou: I assume you are referring to https://github.com/dotnet/runtime/blob/215b39abf947da7a40b0cb137eab4bceb24ad3e3/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.cs#L643?
In previous PRs, we have used the null forgiving operator with `GetExceptionForHR`. Do you want to throw a different exception if null is returned?

gpetrou: `RegisterDragDrop` does not seem to return S_FALSE though as indicated in https://docs.microsoft.com/en-us/windows/win32/api/ole2/nf-ole2-registerdragdrop#return-value

dreddy-work: > I assume you are referring to https://github.com/dotnet/runtime/blob/215b39abf947da7a40b0cb137eab4bceb24ad3e3/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.cs#L643? In previous PRs, we have used the null forgiving operator with `GetExceptionForHR`. Do you want to throw a different exception if null is returned?

yes i was referring to it.  In such case, we may want to just return error code instead of mapping. Did you check any of those error codes mentioned in the document for `RegisterDragDrop` has positive values?

RussKie: This is an existing code, so I'm not sure we should start rewriting this....
We also have this extension:
https://github.com/dotnet/winforms/blob/97777ed2310f23a0f6662540e9e393dba922de57/src/System.Windows.Forms.Primitives/src/Interop/Interop.HRESULT.cs#L94-L100

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -9926,7 +9927,7 @@ internal virtual void RecreateHandleCore()
                     {
                         for (int i = 0; i < controlSnapshot.Length; i++)
                         {
-                            Control childControl = controlSnapshot[i];
+                            Control? childControl = controlSnapshot[i];
```
RussKie, line 9930: How is this can be `null`?

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
gpetrou: We have the following values, so I assume that we can just keep the current changes.
| Name | Value |
| ---: | :--- |
| DRAGDROP_E_INVALIDHWND | 0x80040102 |
| DRAGDROP_E_ALREADYREGISTERED | 0x80040101 |
| E_OUTOFMEMORY | 0x8007000E |

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
gpetrou: We have `controlSnapshot[i] = null;` in line 9884,

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
dreddy-work: I wish we had a better solution but do not want to rewrite this part given this has no impact at runtime behavior, i am fine leaving it as-is.

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
RussKie: Thank you, I missed it.

RussKie (approved):

RussKie:
Standing by for a confirmation from the test team.

Olina-Zhang:
> @Olina-Zhang can you run a sanity check for our runtime tests with these privates once `ready-for-merge`?

Tested this PR with doing some runtime testing and running our internal automation cases, with no new issue found.

gpetrou:
@dreddy-work perhaps we could merge this?

PR #6904, opened 2022-03-26
Enable nullability in Splitter

gpetrou:
## Proposed changes

- Enable nullability in Splitter.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6904)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Splitter.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Splitter.cs
@@ -2,9 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.ComponentModel;
+using System.Diagnostics.CodeAnalysis;
 using System.Drawing;
 using System.Globalization;
 using System.Runtime.InteropServices;
@@ -34,7 +33,7 @@ namespace System.Windows.Forms
         private int minSize = 25;
         private int minExtra = 25;
         private Point anchor = Point.Empty;
-        private Control splitTarget;
+        private Control? splitTarget;
         private int splitSize = -1;
         private int splitterThickness = 3;
         private int initTargetSize;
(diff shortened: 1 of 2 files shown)


dreddy-work (dismissed):

src/System.Windows.Forms/src/System/Windows/Forms/Splitter.cs
```diff
@@ -567,7 +567,7 @@ private void DrawSplitBar(int mode)
         private Rectangle CalcSplitLine(int splitSize, int minWeight)
         {
             Rectangle r = Bounds;
-            Rectangle bounds = splitTarget.Bounds;
+            Rectangle bounds = splitTarget!.Bounds;
```
RussKie: Instead I propose we pass `splitTarget` in, e.g.:
```cs
private Rectangle CalcSplitLine(Control splitTarget, int splitSize, int minWeight)
```

src/System.Windows.Forms/src/System/Windows/Forms/Splitter.cs
```diff
@@ -793,16 +793,16 @@ private int GetSplitSize(int x, int y)
             switch (Dock)
```
RussKie, line 793: ditto for `GetSplitSize`

RussKie (changes requested):

src/System.Windows.Forms/src/System/Windows/Forms/Splitter.cs
```diff
@@ -844,15 +844,7 @@ protected override void OnMouseMove(MouseEventArgs e)
...
-            if (splitTarget is not null)
-            {
-                int x = e.X + Left;
-                int y = e.Y + Top;
-                Rectangle r = CalcSplitLine(GetSplitSize(e.X, e.Y), 0);
-                int xSplit = r.X;
-                int ySplit = r.Y;
-                SplitEnd(true);
-            }
+            SplitEnd(true);
```
RussKie, line 847: Can you please help me understand this change?

gpetrou: If I had to guess, I would say that this code was a copy/paste from `OnMouseMove`. However, here we call `SplitEnd` instead of `OnSplitterMoving`. So, `xSplit` and  `ySplit` are not used. Same applies for `r`, `x` and `y`. Inside `SplitEnd` we use `_splitTarget = null;`, so I removed the `if (splitTarget is not null)` check.

RussKie: That makes sense. Thank you.

RussKie (approved):

PR #6909, opened 2022-03-27
Enable nullability in ListViewGroupCollection and ListViewGroupItemCollection

gpetrou:
## Proposed changes

- Enable nullability in ListViewGroupCollection and ListViewGroupItemCollection.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6909)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroupItemCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroupItemCollection.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.Collections;
 
 namespace System.Windows.Forms
@@ -11,7 +9,7 @@ namespace System.Windows.Forms
     internal class ListViewGroupItemCollection : ListView.ListViewItemCollection.IInnerList
     {
         private readonly ListViewGroup _group;
-        private ArrayList _items;
+        private ArrayList? _items;
 
         public ListViewGroupItemCollection(ListViewGroup group)
         {
@@ -20,7 +18,7 @@ namespace System.Windows.Forms
 
         public int Count => Items.Count;
 
-        private ArrayList Items => _items ?? (_items = new ArrayList());
+        private ArrayList Items => _items ??= new ArrayList();
 
         public bool OwnerIsVirtualListView => _group.ListView is not null && _group.ListView.VirtualMode;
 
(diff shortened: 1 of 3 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroupItemCollection.cs
```diff
@@ -20,22 +18,22 @@ public ListViewGroupItemCollection(ListViewGroup group)
...
+        private ArrayList Items => _items ??= new ArrayList();
 
         public bool OwnerIsVirtualListView => _group.ListView is not null && _group.ListView.VirtualMode;
 
         public bool OwnerIsDesignMode => _group.ListView?.Site is not null && _group.ListView.Site.DesignMode;
 
         public ListViewItem this[int index]
         {
-            get => (ListViewItem)Items[index];
+            get => (ListViewItem)Items[index]!;
```
RussKie, line 29: 💭  A possible solution to `!` operator is to convert `ArrayList` to `List<ListViewItem>`. 

The only issue with this is the following method:
```diff
- public void CopyTo(Array dest, int index) => Items.CopyTo(dest, index);
+ public void CopyTo(Array dest, int index) => Array.Copy(Items.ToArray(), dest, index);
```
...which makes an extra allocation... 🤔 

@stephentoub, I'm curious - are there any ways to convert a `List<T>` to an array without extra allocations or resorting to a reflection? Thanks

stephentoub: > are there any ways to convert a List<T> to an array without extra allocations or resorting to a reflection?

To an array? No.  But you can access `List<T>`'s backing array as a span via https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.collectionsmarshal.asspan?view=net-6.0#system-runtime-interopservices-collectionsmarshal-asspan-1(system-collections-generic-list((-0))).

stephentoub: > The only issue with this is the following method

`List<T>` implements the non-generic `ICollection`, so such a method would presumably just be:
```C#
public void CopyTo(Array dest, int index) => ((ICollection)Items).CopyTo(dest, index);
```

RussKie (approved):

src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroupItemCollection.cs
RussKie: Thank you!

PR #6923, opened 2022-03-29
Enable nullability in KeyboardToolTipStateMachine

gpetrou:
## Proposed changes

- Enable nullability in KeyboardToolTipStateMachine.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6923)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/KeyboardToolTipStateMachine.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/KeyboardToolTipStateMachine.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using static Interop;
 
 namespace System.Windows.Forms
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/KeyboardToolTipStateMachine.cs
```diff
@@ -39,16 +37,16 @@ public static KeyboardToolTipStateMachine Instance
...
 
         private SmState _currentState = SmState.Hidden;
-        private IKeyboardToolTip _currentTool;
+        private IKeyboardToolTip? _currentTool;
         private readonly InternalStateMachineTimer _timer = new();
-        private SendOrPostCallback _refocusDelayExpirationCallback;
+        private SendOrPostCallback? _refocusDelayExpirationCallback;
 
-        private readonly WeakReference<IKeyboardToolTip> _lastFocusedTool = new(null);
+        private readonly WeakReference<IKeyboardToolTip> _lastFocusedTool = new(null!);
```
dreddy-work, line 49: this doesn't seem right.

gpetrou: Indeed. It looks like that I should use nullable `IKeyboardToolTip` and `ToolTip` instead.

dreddy-work: had one more suggestion but rest look good to me.

dreddy-work (approved):

RussKie (approved):

PR #6925, opened 2022-03-29
Correctly retrieve the embedded resource

### LINKED ISSUE #6915: NET 6, Windows.Forms ColorEditor, missing resources
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/t/NET-6-WindowsForms-ColorEditor-missin/1700906)._

---
Hi, the screenshot should explain everything. 
A missing resource in Windows.Forms.Design.dll. 
The problem is easy to reproduce, for example: 
In a PropertyGrid Control, Color Property, DropDown, right click on a Custom Color Field what should show the Colorpicker.
Regards, 
Christian Ohle

---
#### Feedback Bot on 3/24/2022, 02:38 PM: 

<p>We have directed your feedback to the appropriate engineering team for further evaluation. The team will review the feedback and notify you about the next steps.</p>

#### Olina Zhang [MSFT] on 3/24/2022, 06:08 PM: 

<p>Thanks for your feedback!</p>
<p>Unfortunately we cannot reproduce this issue on our side, in order to investigate this further, could you please provide detail reproduce steps or a minimum sample app how to repro it?<br>
Here is we are trying:</p>
<ol>
<li>Create a Winforms .Net 6.0 application</li>
<li>Add a PropertyGrid control to form designer, bind SelectedObject property to Form1</li>
<li>Build and run application</li>
<li>Find backColor property in PropertyGrid, click dropdown, focus on Custom color tab, pick up any color</li>
</ol>
<p>Result: All work well, please see following gif:<br>
</p>

#### Christian Ohle on 3/24/2022, 11:01 PM: 

<p>Hi, it happens on right click on one of the the custom colors fields in the “user defined colors” tab.<br>
I hope it helps</p>

#### Christian Ohle on 3/24/2022, 11:15 PM: 

<p>VS screen record from the page seems not to work for me. Therefore again as screenshots:</p>
<p></p>
<p></p>
<p></p>
<p></p>

#### Olina Zhang [MSFT] on 3/28/2022, 11:33 AM: 

<p>Thanks for your update, I can repro your issue now.</p>

---
### Original Solutions
(no solutions)

RussKie:
Resolves #6915

The regression was caused by the migration of types and resources from System.Drawing.Design to System.Windows.Forms.Design, and the resource is now getting embedded as `System.Windows.Forms.Design.colordlg.data` instead of `System.Drawing.Design.colordlg.data`.
* .NET 6 version embeds the resource as follows:
image
* The .NET Framework version embeds it as:
image

There are few ways of addressing this regression - update the `GetManifestResourceStream()` call and pass an updated resource name, or embed the resource under a different name. The latter felt a little more opaque from the maintenance perspective.

I presume we missed this because:
* we had no test coverage for this dialog as it can only be invoked via the `PropertyGrid` when editing a color, and it's not possible to invoke the dialog directly,
* color editing is not believed to be a major use case in runtime scenarios.

## Customer Impact

- Customers will be able to use custom color editor again

    **Actual Result:** an error dialog pops up:
    image

    **Fixed result:** Define color dialog pops up, following is the behavior in framework app
    image

## Regression? 

- Yes

## Risk

- Minimal

## Test methodology 

- manual tests
- automated tests

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6925)

### DIFF
--- a/src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/Dialogs.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/Dialogs.cs
@@ -6,6 +6,8 @@
 
 using System.ComponentModel;
 using System.ComponentModel.DataAnnotations;
+using System.Drawing;
+using System.Drawing.Design;
 using System.Windows.Forms;
 
 namespace WinformsControlsTest
@@ -25,7 +27,7 @@ namespace WinformsControlsTest
 
             _btnOpen = new("&Open dialog")
             {
-                Image = (System.Drawing.Bitmap?)(resources.GetObject("OpenDialog")),
+                Image = (Bitmap?)resources.GetObject("OpenDialog"),
                 Enabled = false,
             };
 
@@ -49,6 +51,14 @@ namespace WinformsControlsTest
             toolbar.Items.Add(_btnOpen);
         }
 
+        private void DisposeIfNeeded()
+        {
+            if (propertyGrid1.SelectedObject is IDisposable disposable)
+            {
+                disposable.Dispose();
+            }
+        }
+
         private ToolStrip GetToolbar()
         {
             foreach (Control control in propertyGrid1.Controls)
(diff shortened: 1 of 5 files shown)


src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/Dialogs.cs
```diff
@@ -49,6 +51,14 @@ public Dialogs()
             toolbar.Items.Add(_btnOpen);
         }
 
+        private void DisposeIfRequired()
```
KlausLoeffelmann, line 54: The naming sounds a bit weird to me. Like, we should dispose it when it's required (the object), which would be bad. Maybe it's just my foreign language feeling, but `DisposeOnDemand` or `DisposeIfNecessary` would sound better to me.

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/Dialogs.cs
```diff
@@ -63,28 +73,48 @@ private ToolStrip GetToolbar()
             throw new MissingMemberException("Unable to find the toolstrip in the PropertyGrid.");
         }
 
+        private void btnColorDialog_Click(object sender, EventArgs e)
```
KlausLoeffelmann, line 76: I have started to name EventHandlers starting with capital letters, to observe the analyzer rules.
Should we do that in general?

KlausLoeffelmann (dismissed):
Just a few naming nits.

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/Dialogs.cs
RussKie: The designer never capitalised the first char. I oppose this new development in general :)

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/Dialogs.cs
RussKie: IMO, in this case, "necessary" and "required" carry the same meaning - i.e., _invoke dispose, if it is required/necessary_. The the designer codebase we have the same method which is called `DisposeIfNeeded`, I'll rename this to match.

RussKie:
Thank you

PR #6935, opened 2022-03-30
Add some tests for classes that inherit from Control.ControlCollection

gpetrou:
## Proposed changes

- Add some tests for classes that inherit from Control.ControlCollection.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6935)

### DIFF
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridView.DataGridViewControlCollectionTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridView.DataGridViewControlCollectionTests.cs
@@ -1,0 +1,24 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace System.Windows.Forms.Tests;
+
+public class DataGridView_DataGridViewControlCollectionTests
+{
+    [WinFormsFact]
+    public void DataGridViewControlCollection_Ctor_Control()
+    {
+        using DataGridView owner = new();
+        DataGridView.DataGridViewControlCollection collection = new(owner);
+
+        Assert.Empty(collection);
+        Assert.False(collection.IsReadOnly);
+        Assert.Same(owner, collection.Owner);
+    }
+
+    [WinFormsFact]
+    public void DataGridViewControlCollection_Ctor_NullOwner_ThrowsArgumentNullException()
+    {
+        Assert.Throws<ArgumentNullException>("owner", () => new DataGridView.DataGridViewControlCollection(null));
+    }
+}
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ReadOnlyControlCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ReadOnlyControlCollection.cs
@@ -12,7 +12,8 @@ internal class ReadOnlyControlCollection : Control.ControlCollection
 {
     private readonly bool _isReadOnly;
 
-    public ReadOnlyControlCollection(Control owner, bool isReadOnly) : base(owner)
+    public ReadOnlyControlCollection(Control owner, bool isReadOnly)
+        : base(owner)
     {
         _isReadOnly = isReadOnly;
     }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TypedControlCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TypedControlCollection.cs
@@ -14,13 +14,15 @@ internal class TypedControlCollection : ReadOnlyControlCollection
     private readonly Type _typeOfControl;
     private readonly Control _ownerControl;
 
-    public TypedControlCollection(Control owner, Type typeOfControl, bool isReadOnly) : base(owner, isReadOnly)
+    public TypedControlCollection(Control owner, Type typeOfControl, bool isReadOnly)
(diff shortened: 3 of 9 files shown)


src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TabControl.ControlCollectionTests.cs
```diff
@@ -881,11 +881,11 @@ public void TabControlControlCollection_Add_InvalidValue_ThrowsArgumentException
         }
 
         [WinFormsFact]
-        public void TabControlControlCollection_Add_NullValue_ThrowsNullReferenceException()
+        public void TabControlControlCollection_Add_NullValue_ThrowsArgumentException()
         {
             using var owner = new TabControl();
             var collection = new TabControl.ControlCollection(owner);
-            Assert.Throws<NullReferenceException>(() => collection.Add(null));
```
dreddy-work, line 888: you did not change any exception types in this PR. how come this test was passing earlier?

gpetrou: Before, we had:
`throw new ArgumentException(string.Format(SR.TabControlInvalidTabPageType, value.GetType().Name));`
Now, we have:
`throw new ArgumentException(string.Format(SR.TabControlInvalidTabPageType, value?.GetType().Name));`

dreddy-work: Isn't test expecting `NullReferenceException` exception before?

gpetrou: Not sure what you are asking, but before `value.GetType()` would throw a `NullReferenceException` in the specific test. Now with `value?.GetType()` it will throw `ArgumentException`. Unfortunately, the implementations of the `Add` methods are inconsistent between the classes. Some have nop, some have `ArgumentNullException`, some have ArgumentException.

gpetrou: Since it doesn't make much sense to Add/Remove null controls, perhaps what we should do is throw ArgumentNullException in all cases. I have added a separate commit for that.

src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlCollection.cs
```diff
@@ -42,13 +42,8 @@ public virtual bool ContainsKey(string? key)
...
             ///  is first removed from that control.
             /// </summary>
-            public virtual void Add(Control? value)
+            public virtual void Add(Control value!!)
             {
-                if (value is null)
-                {
-                    return;
-                }
-
```
RussKie, line 51: ❗  This is a breaking change. We can't do this.

Applies universally.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.ControlCollection.cs
```diff
@@ -14,20 +12,19 @@ public new class ControlCollection : Control.ControlCollection
...
 
-            public override void Add(Control value)
+            public override void Add(Control? value)
             {
-                if (!(value is TabPage))
+                if (value is not TabPage tabPage)
                 {
-                    throw new ArgumentException(string.Format(SR.TabControlInvalidTabPageType, value.GetType().Name));
+                    throw new ArgumentException(string.Format(SR.TabControlInvalidTabPageType, value?.GetType().Name));
                 }
```
RussKie, line 26: This looks like an inconsistent oversight. In other collections we ignore `null` value, and I think we should do the same here. Whilst technically it is a change of behaviour I don't believe it will result in a breaking change.
```cs
            public override void Add(Control? value)
            {
                if (value is null)
                {
                    return;
                }

                if (value is not TabPage tabPage)
                {
                    throw new ArgumentException(string.Format(SR.TabControlInvalidTabPageType, value.GetType().Name));
                }
```

RussKie (changes requested):

dreddy-work:
@gpetrou, can take a look at current state of this PR and see if it needs to be out of draft?

lonitra:
@gpetrou Apologies for this longstanding PR. It looks like this PR now just adds tests. Could you update PR title and fix test errors?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridView.DataGridViewControlCollectionTests.cs
```diff
@@ -0,0 +1,23 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace System.Windows.Forms.Tests;
+
+public class DataGridView_DataGridViewControlCollectionTests
+{
+    [WinFormsFact]
+    public void DataGridViewControlCollection_Ctor_Control()
+    {
+        using var owner = new DataGridView();
```
lonitra, line 11: Could you change the objects we are instaniating in tests to have type on left hand side? e.g.
```suggestion
        using DataGridView owner = new();
```
Otherwise LGTM, thanks for adding tests! 👍

lonitra (approved):

PR #6940, opened 2022-03-31
Fixing issue with announcing expand/collapse action for TreeNode

### LINKED ISSUE #6732: [Accessibility] Narrator cannot announce the expand/collapse state for treeView with checkbox enabled when expand/collapse any treeNode
* .NET Core Version: 
7.0.100-preview.2.22116.19

Yes

**Minimal repro:**
1. Create a Winforms .Net Core application with treeView control added
2. In treeView, add several root treeNodes and child treeNodes

Or use following attached application:
[WinFormsApp4.zip](https://github.com/dotnet/winforms/files/8107369/WinFormsApp4.zip)

3. Build and run this application
4. Launch Narrator to test expand/collapse root treeNode

**Actual result:**
Narrator cannot announce the expand/collapse state for treeview nodes when expanding or collapsing them. 
image
TreeView_CheckBox

**Expected result:**
Narrator can announce the expand/collapse state for treeview nodes when expanding or collapsing them

SergeySmirnov-Akvelon:
Fixes #6732

## Proposed changes
- The issue is reproduced due to the incorrect operation of the `RaiseAutomationPropertyChangedEvent` for the `TreeNode`. Under different conditions, the message may not be announced or announced twice.
- Using `InternalRaiseAutomationNotification` helps to get around this issue, as it ensures that the message is called from .NET Core side, and not from the side of the native control.
- Added `AnnouncementStopped` flag to avoid announcing `TreeNode` status message on bulk action (e.g. `ExpandAll` method)
- Added unit tests

## Customer Impact
**Before fix:**
154913766-afd3b6f7-cc49-41f1-9e1e-511b52930d85

**After fix:**
PR-6940

## Regression? 
- No

## Risk
- Minimal

## Test methodology 
- CTI team
- Unit tests

## Accessibility testing  
- Narrator
- JAWS
- NVDA

 ## Test environment(s) 
- Microsoft Windows [Version 10.0.19044.1526]
- .NET Core SDK: 7.0.0-preview.4.22179.1

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6940)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TreeNode.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TreeNode.cs
@@ -227,6 +227,12 @@ namespace System.Windows.Forms
         }
 
         /// <summary>
+        ///  This flag is necessary to avoid announcing the node status message
+        ///  during a bulk action (e.g. <see cref="TreeNode.ExpandAll"/> method)
+        /// </summary>
+        internal bool IsAnnouncementStopped { get; private set; }
+
+        /// <summary>
         ///  The bounding rectangle for the node (text area only). The coordinates
         ///  are relative to the upper left corner of the TreeView control.
         /// </summary>
@@ -1767,11 +1773,14 @@ namespace System.Windows.Forms
         /// </summary>
         public void ExpandAll()
         {
+            IsAnnouncementStopped = true;
             Expand();
             for (int i = 0; i < childCount; i++)
             {
                 children[i].ExpandAll();
             }
+
+            IsAnnouncementStopped = false;
         }
 
         /// <summary>
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TreeNode.TreeNodeAccessibleObjectTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TreeNode.TreeNodeAccessibleObjectTests.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Windows.Forms.Automation;
 using System.Windows.Forms.TestUtilities;
 using Xunit;
 using static System.Windows.Forms.TreeNode;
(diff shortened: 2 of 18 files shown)


dreddy-work (dismissed):
LGTM. @Tanya-Solyanik , can you take a look when you get chance?

src/System.Windows.Forms/src/System/Windows/Forms/TreeNode.cs
```diff
@@ -226,6 +226,12 @@ public Color BackColor
             }
         }
 
+        /// <summary>
+        ///  This flag is necessary to avoid announcing the node status message
+        ///  during a bulk action (e.g. <see cref="TreeNode.ExpandAll"/> method)
+        /// </summary>
+        internal bool AnnouncementStopped { get; set; }
```
RussKie, line 233:
```suggestion
        internal bool AnnouncementStopped { get; private set; }
```

I'd call it `IsAnnouncingNodeStatus`

RussKie (commented):
👍 

@vladimir-krestov you may need to take over it.

vladimir-krestov (dismissed):

src/System.Windows.Forms/src/System/Windows/Forms/TreeNode.cs
```diff
@@ -1767,11 +1773,14 @@ public void Expand()
         /// </summary>
         public void ExpandAll()
         {
+            IsAnnouncementStopped = true;
             Expand();
```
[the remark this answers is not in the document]
Tanya-Solyanik: This sets the flag on the parent node, but it looks like you are reading it on the child node, no?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TreeNode.TreeNodeAccessibleObjectTests.cs
```diff
@@ -485,5 +486,110 @@ public void TreeNodeAccessibleObject_Value_EqualsText()
             Assert.Equal(testText, node.AccessibilityObject.Value);
             Assert.False(control.IsHandleCreated);
         }
+
+        [WinFormsTheory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public void TreeNodeAccessibleObject_Expand_Collapse_Calls_RaiseAutomationNotification_TreeNode(bool checkBoxesEnabled)
```
Tanya-Solyanik, line 493: Please reference the bug you are testing - https://github.com/dotnet/winforms/issues/6732

Tanya-Solyanik (commented):
Please describe the scenario that you are fixing - when is the announcement skipped?

vladimir-krestov:
I have prepared a new implementation of the fix.
This PR doesn't look safe and good to me. Narrator doesn't announce expand-collapse state of a node, like it does without checkboxes. 
So here is the new way:
image

Proposed change: use TreeItem control type for TreeNodes despite CheckBoxes property. 
Risks: **users used to hear "checkbox" for these nodes. We will change .Net Framework behavior as well.**

I think, nodes should be "treeitem" always, they are tree items first, even if they have checkboxes.

This PR behavior:
image


My approach behavior:
image

How it works without CheckBoxes in **main** branch:
image


@merriemcgaw, @Tanya-Solyanik, I need your approval to change the control type.

Tanya-Solyanik:
Once you change the control type, does Acc. Insights detect any problems? Narrator output looks reasonable, but do we lose any patterns, can the default action be invoked correctly? I think both checkbox and tree item are reasonable types as long as toggle pattern is available.

Tanya-Solyanik:
Feel free to start a new PR.

vladimir-krestov:
> Once you change the control type, does Acc. Insights detect any problems? Narrator output looks reasonable, but do we lose any patterns, can the default action be invoked correctly? I think both checkbox and tree item are reasonable types as long as toggle pattern is available.

1) There are no any AI warnings for the both cases (CheckBox vs TreeItem)
2) All Inspect properties the same for the both cases (except control type)
3) All Inspect actions the same for the both cases, DefailtAction, Toggle, Expand/Collapse, work correctly as well.

PS. ExpandCollapse pattern available for any tree node (even in leaf mode), that is not correct and should be fixed.

Will create a PR for the new fix.

vladimir-krestov:
Closed this PR, because the new implementation (#7044) of the fix was merged.

PR #6946, opened 2022-03-31
Fixing PropertyGrid memory leaks

### LINKED ISSUE #6943: PropertyGrid accessible objects are leaking in memory
### .NET version

7.0

No

No

### Issue description

`PropertyGrid` accessible objects are left in memory after the `PropertyGrid` control instance was inspected with accessibility tool (like `Accessibility Insights` or `Inspect`) and the form with `PropertyGrid` is closed, garbage is collected, but accessibility tool is still opened.

Report from WinDbg:
image

**Expected behavior:**

### Steps to reproduce

1. Create a parent form with two buttons: one will be opening a child form with the `PropertyGrid` (`SelectedObject` set to `this`) control and the other will collect the garbage.
2. Press the second button to open the child form with `PropertyGrid`.
3. Start the `Accessibility Insights` tools.
4. Run automated checks for the `PropertyGrid` control using AI.
5. Close the child form.
7. DON'T close the AI tool.
8. Look for `PropertyGrid` accessible objects left in memory (using `WinDbg`, or `dotMemory`, or another suitable tool).

ArtemTatarinov:
Fixes #6943

## Proposed changes
- Amount of leaked objects decreased from 595 to 39
- Use the `UiaDisconnectProvider` method through the hierarchy of the classes to free objects held by accessibility tool
- Only static comparers should be left at memory as amount of memory allocated for them isn't increasing after several recurrences of the repro steps described in issue

**This PR requires further development:**
- 39 objects are still left in memory, some of them are accessible objects. They should be tried to clean, except for the Comparers/Sorters which are most possibly are static system objects.
- Quite possible that nullifying `_gridView` and other objects won't be necessary when all provider will be disconnected property
- I've met a crash at the code line `Debug.Assert(result == 0)` for the result of disconnecting the third button of the `_viewSortButtons` collection if debugger is detached, and E_FAIL result if attached - it requires further investigation
- There was a crash scenario at NetFx:  start AI, run internal test app, click "Dialogs" button, click "Print dialog" button, then "Folder browser dialog" - disconnecting `gridEntry.AccessibilityObject` at the `DisconnectChildren` method at the `PropertyGridView` class caused NRE. I couldn't reproduce this issue at the Core branch, but it's better to double check it.

## Customer Impact
- PropertyGrid memory will stop leaking

## Regression? 
- No

## Risk
- Minimal

### Before
image

### After
image

## Test methodology
- Manual testing (_using WinDbg, see the screenshots above_)
- CTI team

## Test environment(s)
- Microsoft Windows [Version 10.0.19043.1586]
- .NET 7.0.0-preview.3.22159.12

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6946)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGrid.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGrid.cs
@@ -26,11 +26,11 @@ namespace System.Windows.Forms
     [SRDescription(nameof(SR.DescriptionPropertyGrid))]
     public partial class PropertyGrid : ContainerControl, IComPropertyBrowser, Ole32.IPropertyNotifySink
     {
-        private readonly HelpPane _helpPane;
+        private HelpPane _helpPane;
         private int _helpPaneSizeRatio = -1;
         private int _commandsPaneSizeRatio = -1;
-        private readonly CommandsPane _commandsPane;
-        private readonly ToolStrip _toolStrip;
+        private CommandsPane _commandsPane;
+        private ToolStrip _toolStrip;
 
         private bool _helpVisible = true;
         private bool _toolbarVisible = true;
@@ -43,7 +43,7 @@ namespace System.Windows.Forms
         private Bitmap _categoryBitmap;
         private Bitmap _propertyPageBitmap;
 
-        private readonly List<TabInfo> _tabs = new();
+        private List<TabInfo> _tabs = new();
         private TabInfo _selectedTab;
         private bool _tabsDirty = true;
 
@@ -61,7 +61,7 @@ namespace System.Windows.Forms
         private readonly ToolStripSeparator _separator2;
 
         // Our main view
-        private readonly PropertyGridView _gridView;
+        private PropertyGridView _gridView;
 
         private IDesignerHost _designerHost;
         private IDesignerEventService _designerEventService;
@@ -1826,12 +1826,15 @@ namespace System.Windows.Forms
 
                 ActiveDesigner = null;
 
-                foreach (var tabInfo in _tabs)
+                if (_tabs != null)
                 {
-                    tabInfo.Tab.Dispose();
-                }
+                    foreach (var tabInfo in _tabs)
+                    {
+                        tabInfo.Tab.Dispose();
+                    }
 
-                _tabs.Clear();
+                    _tabs.Clear();
+                }
 
                 _normalButtonImages?.Dispose();
                 _normalButtonImages = null;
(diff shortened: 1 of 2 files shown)


Tanya-Solyanik:
@vladimir-krestov  - please prioritize this work because we introduced these leak when adding UIA providers

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGrid.cs
```diff
@@ -4255,6 +4258,54 @@ protected unsafe override void WndProc(ref Message m)
 
                     m.ResultInternal = 1;
                     return;
+                case User32.WM.DESTROY:
+                    base.WndProc(ref m);
+                    if (OsVersion.IsWindows8OrGreater && IsAccessibilityObjectCreated)
+                    {
+                        _gridView?.DisconnectChildren();
+                        HRESULT result = UiaCore.UiaDisconnectProvider(_gridView?.AccessibilityObject);
```
Tanya-Solyanik, line 4266: Do not call `UiaDisconnectProvide` if _gridView is null or _gridView does not have accessibility object created. And the same applies to all other subcontrols.

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGrid.cs
```diff
@@ -4255,6 +4258,54 @@ protected unsafe override void WndProc(ref Message m)
...
+                        Debug.Assert(result == 0);
+                        result = UiaCore.UiaDisconnectProvider(_helpPane?.AccessibilityObject);
+                        Debug.Assert(result == 0);
+                        result = UiaCore.UiaDisconnectProvider(_commandsPane?.AccessibilityObject);
```
Tanya-Solyanik, line 4270: Are we really creating providers for all inner controls?

Tanya-Solyanik: Please use `HRESULT.S_OK` if possible

RussKie: We should use these extensions
https://github.com/dotnet/winforms/blob/5d454f4bcee2bf6eced54624720f2ddf0980e9c8/src/System.Windows.Forms.Primitives/src/Interop/Interop.HRESULT.cs#L83-L92

Tanya-Solyanik: We want to Debug.Assert on S_FALSE

dmitrii-drobotov:
We'll create a new PR once we get around to it and use this one as a starting point.

PR #6956, opened 2022-04-03
System.Windows.Forms.DataGridView.RefreshColumns() improvements

kirsan31:
## Proposed changes
- Not generate bound columns if `AutoGenerateColumns == false`.
- Removed double memory allocation in `GetCollectionOfBoundDataGridViewColumns`.

## Customer Impact

- Performance improvements in `AutoGenerateColumns == false`  - no need to generate bound columns in this mode.
- Performance improvements in `AutoGenerateColumns == true`  - no more double memory allocation and coping.

## Regression? 

- No

## Risk

- Minimal.

## Test methodology

Existing tests and manual testing.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6956)


src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.DataConnection.cs
```diff
@@ -991,14 +991,14 @@ private static DataGridViewColumn GetDataGridViewColumnFromType(Type type)
...
-            public DataGridViewColumn[] GetCollectionOfBoundDataGridViewColumns()
+            public List<DataGridViewColumn> GetCollectionOfBoundDataGridViewColumns()
             {
                 if (_props is null)
                 {
                     return null;
                 }
 
-                ArrayList cols = new ArrayList();
+                List<DataGridViewColumn> cols = new List<DataGridViewColumn>();
```
RussKie, line 1001: nit
```suggestion
                List<DataGridViewColumn> columns = new();
```

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.Methods.cs
```diff
@@ -1624,7 +1624,7 @@ private void AutoGenerateDataBoundColumns(DataGridViewColumn[] boundColumns)
             // 4. Add new columns for the Fields which were not data bound previously ( ie, for fields which do not have a clone ).
             if (boundColumns is not null)
             {
-                for (int j = 0; j < boundColumns.Length; j++)
+                for (int j = 0; j < boundColumns.Count; j++)
                 {
                     if (boundColumns[j] is not null && boundColumns[j].IsBrowsableInternal)
```
RussKie, line 1629: Slight OT.
I think we can further simplify this, though NRT will make it obvious.
We know that `AutoGenerateDataBoundColumns` has a single callsite - `RefreshColumns`. The input parameter can either be `null`, or comes from `DataConnection.GetCollectionOfBoundDataGridViewColumns()` call - which can either be `null` or a list of non-nulls. This means `boundColumns[j] is not null` is always `true`.

RussKie (approved):

RussKie (approved):

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.DataConnection.cs
```diff
@@ -1032,9 +1032,7 @@ public DataGridViewColumn[] GetCollectionOfBoundDataGridViewColumns()
                     cols.Add(dataGridViewColumn);
                 }
 
-                DataGridViewColumn[] ret = new DataGridViewColumn[cols.Count];
-                cols.CopyTo(ret);
-                return ret;
```
dreddy-work, line 1037: by any chance, this copy was modifiable earlier and now not without impacting the original columns?

kirsan31: We have not here _original_ columns. We create here new List and fill it with new created columns. 
But in any case, it is then not modified.

coderb:
Please use the list constructor that takes a capacity when it is known.  This can save memory and also avoid garbage.
`List<DataGridViewColumn> cols = new();`  => `List<DataGridViewColumn> cols = new(_props.Count);`

kirsan31:
> Please use the list constructor that takes a capacity when it is known. This can save memory and also avoid garbage. `List<DataGridViewColumn> cols = new();` => `List<DataGridViewColumn> cols = new(_props.Count);`

I thought about it too, but capacity will not be `_props.Count` in all cases (may be less) and I decided to leave original realization...

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.Methods.cs
```diff
@@ -1624,39 +1624,41 @@ private void AutoGenerateDataBoundColumns(DataGridViewColumn[] boundColumns)
...
-                        if (addNewColumn)
-                        {
-                            dataGridViewCols.Add(boundColumns[j]);
-                        }
-                        else
+                    bool addNewColumn = true;
+                    // Go thru the list of cloned columns and see if there is another column w/ the same data property name.
+                    int clonedColIndex = 0;
+                    for (; clonedColIndex < clonedColumnsCount; clonedColIndex++)
+                    {
```
dreddy-work, line 1638: can you  simplify this block by moving  `finalClonedColumns[clonedColIndex]` in to local ? also, you won't need `clonedIndex` outside of the for loop then.

dreddy-work: i did not see local for `finalClonedColumns[clonedColIndex]`.  You are accessing indexed location multiple times. why not store it local and reuse?

kirsan31: So I misunderstood you the first time...

RussKie (approved):

Issue #6930, opened 2022-03-29
DataGridView unshare rows in virtual model despite all conditions to prevent it are met.


isadkvc:
### .NET version

6.0.201

### Did it work in .NET Framework?

No

### Did it work in any of the earlier releases of .NET Core or .NET 5+?

Tested with .NET Framework 4.6.1, same issue.

### Issue description

When creating `DataGridView` and setting `VirtualMode` to `true` it is not possible to prevent rows from becoming unshared.

As far as i see, i did everything that is described in this [article](https://docs.microsoft.com/en-us/dotnet/desktop/winforms/controls/how-to-implement-virtual-mode-in-the-windows-forms-datagridview-control?view=netframeworkdesktop-4.8) to prevent rows becoming unshared, but no luck. This article is not updated for .NET Core or later.

Please take a look at the video, rows becoming unshared on row (cell) select (in 4.6.1 same behavior) but in 6.0.2 it also for some reason incrementally unshare all rows from 0 to the end in inconsistent manner (sometimes it won't start it unless row changed with arrow keys or some other woodoo)
https://user-images.githubusercontent.com/102589197/160618007-a6077a18-8869-4388-9cbd-52ea604f4fec.mp4

### Steps to reproduce

Code below is not exactly what i'm trying to achieve, it is a result of experiments to see if i can solve this issue.
My goal is to display large set of data using virtual mode without unsharing rows.

```cs
namespace WinFormsApp1;

internal static class Program
{
    [STAThread]
    static void Main()
    {
        ApplicationConfiguration.Initialize();
        Application.Run(new Form1());
    }
}

public partial class Form1 : Form
{
    public Form1()
    {
        InitializeComponent();

        grid.ReadOnly = true;
        grid.AllowUserToResizeColumns = false;
        grid.AllowUserToResizeRows = false;
        grid.ShowCellToolTips = false;
        grid.Columns.Add(new DataGridViewColumn(new DataGridViewTextBoxCell()) { HeaderText = "One", Name = "c_one1" });
        grid.Columns.Add(new DataGridViewColumn(new DataGridViewTextBoxCell()) { HeaderText = "Two", Name = "c_one2" });
        grid.Columns.Add(new DataGridViewColumn(new DataGridViewTextBoxCell()) { HeaderText = "Three", Name = "c_one3" });
        grid.Columns.Add(new DataGridViewColumn(new DataGridViewTextBoxCell()) { HeaderText = "Four", Name = "c_one4" });
[...]

Issue #6859, opened 2022-03-17
DataGridView not dispose old elements


kirsan31:
* .NET Core Version: all up to 7.0 p2

* Have you experienced this same bug with .NET Framework?: Yes

**Problem description:**
As I mentioned in https://github.com/dotnet/winforms/issues/6858 `DataGridView` not dispose old data on `DataSource` change. This is lead to 2 problems:
1. All this stuff are going to finalization query because we have unnecessary finalizer here #6858.
2. If we have a `ContextMenuStrip` bound to Cell/Column/Row we will have a memory leak. `DataGridView` with 3 columns and `ContextMenuStrip` bound to 1000 rows, result of DS refresh:
![image](https://user-images.githubusercontent.com/17767561/158846287-aebb7710-09c2-45f5-8428-6f4ebe9812cd.png)

**Expected behavior:**
Old data must be property disposed. Probably in `DataGridView.RefreshColumnsAndRows()` (in `DataGridView.Methods.cs`)

**Minimal repro:**
[DataGridViewLeak.zip](https://github.com/dotnet/winforms/files/8288127/DataGridViewLeak.zip)

dreddy-work:
Thanks @kirsan31. ownership and lifespan of the DataSource need to be investigated here. Given that this has been the case since .NET framework, this will be on low priority for the team but should be able to help you if you get chance to investigate this further.

kirsan31:
> ownership and lifespan of the DataSource need to be investigated here.

No it's not about DataSource, it's about that all elements (`DataGridViewCell`, `DataGridViewRow`, `DataGridViewColumn`) are replaced on `dataGridVie.DataSource = dataSource;` (and may some other scenarios too).

> Given that this has been the case since .NET framework, this will be on low priority for the team

https://social.msdn.microsoft.com/Forums/silverlight/en-US/fdba08b0-5c62-47a4-be73-48d11372fa02/colossal-memory-leak-contextmenu-on-datagridrow
https://stackoverflow.com/questions/20270367/gdi-objects-memory-leak-custom-toolstripcontrolhost-within-contextmenustrip-is
https://stackoverflow.com/questions/44224884/c-sharp-contextmenustrip-of-datagridviewrow-what-is-the-equivalent-of-sourcecon
And so on...
The general conclusion people come to is not to use `ContextMenuStrip` with `DataGridView` elements.

The problem is in _good_ known pattern - ContextMenuStrip lifetime tracking:
[...]

PR #6957, opened 2022-04-03
Fixed unsharing of last shared row in DataGridViewRowCollection.

kirsan31:
## Proposed changes

With current implementation if currently unsharing row is the last one instance (of this shared row) in DataGridViewRowCollection we will clone it and therefore current instance will be lost. The only exception is if it is the very last row in the collection. It's more correct in this situation to always just update the index and return the current row.

## Customer Impact

- In some situation old behavior can mislead ppl and theoretically lead to bugs. For example lets see at [DataGridViewCell_Displayed_GetWithSharedDataGridViewWithHandle_ReturnsExpected](https://github.com/dotnet/winforms/blob/be52b79bfe1553d69118e1a630767f65a8c653c1/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewCellTests.cs#L530) test: 
```cs
public void DataGridViewCell_Displayed_GetWithSharedDataGridViewWithHandle_ReturnsExpected(bool gridVisible, bool rowHeadersVisible, bool columnHeadersVisible, bool columnVisible)
{
    using var control = new DataGridView
    {
        Visible = gridVisible,
        RowHeadersVisible = rowHeadersVisible,
        ColumnHeadersVisible = columnHeadersVisible
    };
    using var cellTemplate = new SubDataGridViewCell();
    using var column = new DataGridViewColumn
    {
        CellTemplate = cellTemplate,
        Visible = columnVisible
    };
    control.Columns.Add(column);
    // We create default row with State = DataGridViewElementStates.Visible;
    // After that we have only one such shared row in the collection.
    // And there 2 rows in the collection overall (+ 1 new_row because of AllowUserToAddRows == true).
    control.Rows.Add();
    // Store it. Note that if we put this line after Assert.NotEqual(IntPtr.Zero, control.Handle),
    // with previews implementation (prior #6957) test will fail.
    DataGridViewRow row = control.Rows.SharedRow(0);
    // While calling control.Handle we will create a new handle.
    // During this process DataGridView call MakeFirstDisplayedCellCurrentCell().
    // Where if we have cell to display (columnVisible) we will set it as CurrentCell.
    // Which in turn will lead to unsharing of the raw in any case.
    // MakeFirstDisplayedCellCurrentCell() -> SetAndSelectCurrentCellAddress() -> SetCurrentCellAddressCore()
    //   -> OnCurrentCellChanged() -> CurrentCell.get() -> Rows[index]
    // See https://github.com/dotnet/winforms/issues/6930#issuecomment-1090213559.
    // Note that the cell is Displayed (Displayed == true) only if the owning row IS NOT shared and Displayed and owning column is Displayed.
    // So we have these options:
    // 1. columnVisible == false - our row remain shared and therefore cell will not be Displayed;
    // 2. gridVisible == false - our row and column are not Displayed, so and the cell too.
    // 3. (gridVisible && columnVisible) == true - our row became unshared and Displayed.
    //    But because this is the only one such row in the collection and Count of all rows = 2,
    //    with previous implementation (prior #6957) we will clone it and LOST.
    Assert.NotEqual(IntPtr.Zero, control.Handle);
    int invalidatedCallCount = 0;
    control.Invalidated += (sender, e) => invalidatedCallCount++;
    int styleChangedCallCount = 0;
    control.StyleChanged += (sender, e) => styleChangedCallCount++;
    int createdCallCount = 0;
    control.HandleCreated += (sender, e) => createdCallCount++;

    DataGridViewCell cell = row.Cells[0];
    // Here 3 our ways with previous implementation (prior #6957):
    // 1. All correct - we have shared and therefore not Displayed row.
    // 2. All correct - we have not Displayed row.
    // 3. control.Rows.SharedRow(0) unshared and Displayed BUT our stored row (LOST) still shared and therefore not Displayed.
    //    So test will pass of course.
    // Test must check: Assert.Equal(gridVisible && columnVisible, cell.Displayed);
    // And with new implementation it will work no mater where you put row = control.Rows.SharedRow(0);
    Assert.False(cell.Displayed);
    Assert.Equal(0, invalidatedCallCount);
    Assert.Equal(0, styleChangedCallCount);
}
```

- ~We will not lost any rows and they will not end in Freachable queue #6859~ (Not relevant after https://github.com/dotnet/winforms/pull/6978). Also we will not create a new row in this case - performance improvement. But we need to do additional run over rows collection - performance decrease.
  Performance impact overall: if iteration over rows is faster then row cloning - we will get performance increase, otherwise - decrease. See benchmark results: https://github.com/dotnet/winforms/pull/6957#issuecomment-1200465985
- Memory consumption decrease in some cases. See benchmark results: https://github.com/dotnet/winforms/pull/6957#issuecomment-1200465985

## Regression? 

- No

## Risk

- Low in current implementation (see comments in code).

## Test methodology 

Existing tests (with corrections) and manual testing.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6957)


src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRowCollection.cs
```diff
@@ -211,12 +211,12 @@ protected DataGridView DataGridView
                 DataGridViewRow dataGridViewRow = SharedRow(index);
                 if (dataGridViewRow.Index == -1)
                 {
-                    if (index == 0 && items.Count == 1)
+                    if (SharedList.LastIndexOf(dataGridViewRow) == index && SharedList.IndexOf(dataGridViewRow) == index)
```
kirsan31, line 214: This is error free implementation, but will introduce some slowdown (o(n) when the number of such rows approaches to1). We can have a counter for each shared row - how many references of this row are in our collection. If this counter is = 1, then this is the last row. This approach won't negatively impact performance, but it will increase the chance of errors when reference counting...

RussKie: > but will introduce some slowdown (o(n) when the number of such rows approaches to1).

Not quite sure if I fully understand your comment. Typo? 
O(N) is not too bad, especially in the context of DGV.

RussKie:
Is this a fix to an open issue? If so - could you please link it. If not - could you please provide a sample that demonstrates the issue.

kirsan31:
> Is this a fix to an open issue? If so - could you please link it. If not - could you please provide a sample that demonstrates the issue.

I was looking for leaks while working on the https://github.com/dotnet/winforms/issues/6859 and found this problem. After fixing it I ran into the fact that two tests do not work. Actually, the fact that they worked before and indicates a problem :) I will try to update 1 post...

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRowCollection.cs
kirsan31: O(n) in worst case is not too bad yea, but before it was O(1) in any case...

kirsan31:
I've updated the 1 post with more detailed explanations.

RussKie:
> * I can put these comments to source if needed.

Thank you for the comments! They do help.
And if you could add them to the code - it'd be great! I'll help with the wording once they're in.

RussKie:
> ```cs
> // So we have 2 ways:
>     // 1. gridVisible && columnVisible == false; our row remain shared and therefore invisible.
>     // 2. gridVisible && columnVisible == true; our row became unshared and therefore visible. But because this is the only one such row in the collection and Count of all rows = 2,
>     // with previous implementation we will clone it and LOST.
> ```

This sounds like good candidates for individual tests. That is, add a test to verify that a row gets shared and another to verify the opposite.

> ```cs
> // Here 2 our ways:
>     // 1. All correct - we have shared and therefore invisible row.
>     // 2. control.Rows.SharedRow(0) unshared and therefore visible BUT our stored row (LOST) still shared and therefore invisible :( So test will pass of course.
>     // Test must check: Assert.Equal(gridVisible && columnVisible, cell.Displayed); And with new implementation it will work no mater where you put row = control.Rows.SharedRow(0);
> ```

Same here.

kirsan31:
I will definitely look into it (and everything else), the only thing I don't know how soon, it's a complete mess with the time for now :(

RussKie:
Totally understand. No rush, take your time.
Thank you.

KlausLoeffelmann:
@kirsan31: I have just started to take a look at this.
What do you mean by "Shared" in general in this context?

KlausLoeffelmann:
Forget the comment. I misread.

KlausLoeffelmann:
@kirsan31: Is this something - from your assessment - people may have encountered before and implemented their own workaround? If that is the case, how likely is it that after this fix their workaround will no longer work?
(Sorry for those basic questions - only ramping up with this).

kirsan31:
@KlausLoeffelmann 
> @kirsan31: Is this something - from your assessment - people may have encountered before and implemented their own workaround? If that is the case, how likely is it that after this fix their workaround will no longer work?
> (Sorry for those basic questions - only ramping up with this).

The only workaround for this case that come to my mind is - refresh cached (and may be lost) shared raw with call to `SharedRow(index)`.
So after the fix everything seems to be working. But I might be missing something, of course...

kirsan31:
Sorry for long delay.
I once again studied the code, regarding the implementation through the reference counter. And I'm afraid that it will be very problematic (at least for me) and error prone :(
Then I started writing a benchmark (BenchmarkDotNet) to test the performance drop when using `IndexOf`. But ran into a problem - [can't use BenchmarkDotNet with newly build WinForms dll](https://github.com/dotnet/BenchmarkDotNet/issues/1197#issuecomment-1173126572) 🤷‍♂️
Any suggestions?

RussKie:
I guess if @adamsitnik doesn't have an answer, the only way to run the benchmark would be with [this step](https://github.com/dotnet/winforms/blob/main/docs/debugging.md#1-copy-your-changes-into-the-sdk).

kirsan31:
Ok, I've tested this change (thanks to @adamsitnik advice):
Modified 6.0.7 branch: https://github.com/kirsan31/winforms/tree/DGVRowUnsharingBench
Benchmark (need to be run from `NewRel` configuration): https://github.com/kirsan31/DataGridViewRowUnsharingBench
**Job**
- _Original_ - without modifications.
- _New_ - testing change.

**Method**:
- _EnumAll_1Uniq_ - All rows are shared with one shared row.
- _EnumAll_2Uniq_ - First halve of rows are shared with one shared row and second halve with another shared row.
- _EnumAll_HalfUniq_ - Every 2 rows are shared with one individual shared row.
- _EnumAll_AllUniq_ - Every row is individual shared row.
Results:
Snipaste_2022-07-31_19-51-20

In my opinion, the drop in performance in some cases is acceptable. Also, as noted in 1 post, in some cases we get an increase in performance and a decrease in memory used.

RussKie (dismissed):

kirsan31:
@RussKie what about [this](https://github.com/dotnet/winforms/pull/6957#issuecomment-1119297512) and [this](https://github.com/dotnet/winforms/pull/6957#issuecomment-1119309471) your comments?
I was going to implement them after a performance check...

KlausLoeffelmann:
Does this change affect virtual mode in any way?

KlausLoeffelmann:
What I mean is: Have you thought of any _additional_ cases to take into account when the DGV is virtual mode.
If my question doesn't make sense, then please disregard. Just wanted to make sure we covered all important scenarios.

kirsan31:
> What I mean is: Have you thought of any _additional_ cases to take into account when the DGV is virtual mode.
> If my question doesn't make sense, then please disregard. Just wanted to make sure we covered all important scenarios.

Theoretically unsharring rows through `DataGridViewRow this[int index]` should work the same for all modes, but I will doublecheck.

Also I almost finished with comments and new checks in sources, will update later...

KlausLoeffelmann:
> but I will doublecheck.

Thanks a lot for that!

kirsan31:
> Theoretically unsharring rows through `DataGridViewRow this[int index]` should work the same for all modes, but I will doublecheck.

I've looked at the code again, in my understanding, there is (and should not be) any difference in processing of `DataGridViewRowCollection.DataGridViewRow this[int index]` for any DGV mode.

Comments and one new check added (1 post was edited too).

P.s. if we will fix #6930 (explanation [here](https://github.com/dotnet/winforms/issues/6930#issuecomment-1091111422)). Then these tests (and may be some others) will also need to be redo again.

RussKie:
Totally understand. No rush, take your time.
Thank you.

kirsan31:
@RussKie 
I want to clarify, all the work has [already been done](https://github.com/dotnet/winforms/pull/6957#issuecomment-1207169530), we are waiting for a review from @KlausLoeffelmann here...

KlausLoeffelmann:
Just one last ask for the moment:
If the rows are generated through Data Binding, will that have any breaking changes?
Do you have tests for that?

kirsan31:
@KlausLoeffelmann 
> Just one last ask for the moment:
> If the rows are generated through Data Binding, will that have any breaking changes?
> Do you have tests for that?

Oh, it was so long time ago... As I remember answer must be the [same](https://github.com/dotnet/winforms/pull/6957#issuecomment-1207169530):

> I've looked at the code again, in my understanding, there is not (and should not be) any difference in processing of DataGridViewRow DataGridViewRowCollection[int index] for any DGV mode.

But of course to be sure I have to raise (remember) my knowledge on this topic and doublecheck...

KlausLoeffelmann:
Thanks for the quick answer!
I know, it's been a while, and I was really busy with... 😉 

Let's just make sure to think of everything. I really want this, but I also don't want to have any breaking changes we have not thought of!

KlausLoeffelmann:
> But of course to be sure I have to raise (remember) my knowledge on this topic and doublecheck...

Did you get the chance to check for this?

@Olina-Zhang: Do we have any DataGridView-binding-related bigger manual test scenarios we could apply for this with a set of private binaries resulting from this branch?

Also, @kirsan31, if you get the chance, could you resolve the merge-conflicts?

kirsan31:
@KlausLoeffelmann 

> Did you get the chance to check for this?

Sorry not yet. And I don't know when I can, as I said, it will take extra time to get to the bottom of the problem again :(

> Also, @kirsan31, if you get the chance, could you resolve the merge-conflicts?

Do you think it is advisable to do this without answering the previous question?

KlausLoeffelmann:
Maybe for testing-now purposes, but please don't feel rushed!
But no, this is not-at-all pressuring, please take your time!
And as always: Really appreciate your thoroughness, here!

Will add the waiting-author-feedback label, though.

Olina-Zhang:
> @Olina-Zhang: Do we have any DataGridView-binding-related bigger manual test scenarios we could apply for this with a set of private binaries resulting from this branch?

We have some .Net framework scenarios about DataGridView-binding, and can follow them in .Net Core for testing.

kirsan31:
Ok, I looked at it again. My thoughts...
- We do not initially have any code that depends on the way data is provided. Therefore, this method should not depend on it in any way.
- All the cloning logic here is aimed at the fact that the consumer from the outside, requesting an instance of the row, received an individual instance, and not a common one. Our change does not violate this.
- We will not add any new logic here, we will expand the existing one:
  image

KlausLoeffelmann (approved):
Look good to me.
Let's take it!

PR #6399, opened 2021-12-22
Move ENPROTECTED to Interop Richedit

gpetrou:
## Proposed changes

- Move ENPROTECTED to Interop Richedit.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6399)

PR #2851, opened 2020-02-13
fix: `PrintDialog` not shown

RussKie:
Resolves #2814

## Proposed changes

The bug was caused by the conversion of a struct fields to auto-properties in  #2594. This broke the interop binary layout.

Rework the code to
- correctly define x86 and x64 structs
- make the structs blittable
- add tests verifying the structs layouts

## Customer Impact

- `PrintDialog` is now shown as expected
- 

## Regression? 

- Yes

## Test methodology 

manual

* x86
```powershell
dotnet restore --runtime win-x86
dotnet run --runtime win-x86
```
image

* x64
```powershell
dotnet restore --runtime win-x64
dotnet run --runtime win-x64
```
image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/winforms/pull/2851)

Issue #2814, opened 2020-02-07
 Invoke PrintDialog.ShowDialog() does not show Print dialog successfully


Thea-Zhao:
* .NET Core Version: .NET Core SDK 5.0.100-preview.1.20106.1 + the VSIX Inserted into Dev16 16.5.0 [29803.141.master]

* Have you experienced this same bug with .NET Framework?: No

**Problem description:**
 Invoke PrintDialog.ShowDialog() does not show Print dialog successfully.

**More Info:**
1. This is a regression issue, it's not reproducing when targeting .NET Core 3.1.
2. This issue also cannot reproduce with Framework project.

**Actual behavior:**
No expected Print dialog shown.
.NET Core 5.0 screenshot:
![PrintDialogIssue5 0](https://user-images.githubusercontent.com/53589182/74011630-291fbd00-493d-11ea-9879-f54897d6b34a.gif)

**Expected behavior:**
The Print dialog should be shown successfully.
.NET Core 3.1 screenshot:
![PrintDialogIssue3 1](https://user-images.githubusercontent.com/53589182/74011467-c3cbcc00-493c-11ea-9b47-81ecb1128738.gif)

**Minimal repro:**

1. Create a Winforms Core application.
2. Drag and drop a button to the form.
3. Double click the button1 to generate the button1_click event, then add following code to it:
PrintDialog pdlg = new PrintDialog();
pdlg.ShowDialog();
4. Build and run.
5. Click the button1 to see whether the Print dialog is popping up.
[PrintDialogIssue.zip](https://github.com/dotnet/winforms/files/4169604/PrintDialogIssue.zip)

RussKie:
I haven't bisected, but looking at the history we probably disturbed it in https://github.com/dotnet/winforms/pull/2594.
Guessing these classes, having one of the fields modified, caused some misalignment, and caused the call to `UnsafeNativeMethods.PrintDlg(data)` fail.

/cc: @gpetrou

weltkante:
@RussKie the PR you linked moved the `int m_Flags;` field which broke the binary layout. I'm adding comments to the PR to highlight the mistakes.

RussKie:
Thank you @weltkante 👍  somehow I failed to see it myself

Thea-Zhao:
Verified with latest Dev16 16.6.0 Preview 2.0 [29904.160.master] + latest .NET Core 5.0 5.0.100-preview.1.20154.12 build, this issue is fixed now.
[...]

PR #6962, opened 2022-04-04
Remove Message.GetLParam since it is linker unfriendly

kant2002:
That remove one warning from linker.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6962)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/Interop/Richedit/Interop.CHARRANGE.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/Richedit/Interop.CHARRANGE.cs
@@ -8,6 +8,7 @@ internal partial class Interop
 {
     internal static partial class Richedit
     {
+        // All structures from richedit.h is packed at 4
         [StructLayout(LayoutKind.Sequential, Pack = 4)]
         public struct CHARRANGE
         {
--- a/src/System.Windows.Forms.Primitives/src/Interop/Richedit/Interop.CHARRANGE.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/Richedit/Interop.CHARRANGE.cs
@@ -8,7 +8,7 @@ internal partial class Interop
 {
     internal static partial class Richedit
     {
-        [StructLayout(LayoutKind.Sequential)]
+        [StructLayout(LayoutKind.Sequential, Pack = 4)]
         public struct CHARRANGE
         {
             public int cpMin;
--- a/src/System.Windows.Forms.Primitives/src/Interop/Richedit/Interop.CHARRANGE.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/Richedit/Interop.CHARRANGE.cs
@@ -8,8 +8,7 @@ internal partial class Interop
 {
     internal static partial class Richedit
     {
-        // All structures from richedit.h is packed at 4
-        [StructLayout(LayoutKind.Sequential, Pack = 4)]
+        [StructLayout(LayoutKind.Sequential, Pack = RichEditPack)]
         public struct CHARRANGE
         {
             public int cpMin;
--- a/src/System.Windows.Forms.Primitives/src/Interop/Richedit/Interop.GTL.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/Richedit/Interop.GTL.cs
@@ -14,7 +14,7 @@ internal partial class Interop
             PRECISE = 2,
             CLOSE = 4,
             NUMCHARS = 8,
-            NUMBYTES = 6,
+            NUMBYTES = 16,
         }
     }
 }
--- a/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/NativeMethods.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/NativeMethods.cs
@@ -122,26 +122,6 @@ namespace System.Windows.Forms
             public int FlagsEx;
         }
 
-        [StructLayout(LayoutKind.Sequential, Pack = 4)]
-        public struct ENLINK
-        {
-            public User32.NMHDR nmhdr;
-            public int msg;
-            public IntPtr wParam;
-            public IntPtr lParam;
-            public Richedit.CHARRANGE charrange;
-        }
-
-        [StructLayout(LayoutKind.Sequential, CharSet = CharSet.Auto, Pack = 4)]
-        public struct ENPROTECTED
-        {
-            public User32.NMHDR nmhdr;
-            public int msg;
-            public IntPtr wParam;
-            public IntPtr lParam;
-            public Richedit.CHARRANGE chrg;
-        }
-
         public class ActiveX
         {
             public const int ALIGN_MIN = 0x0;
--- a/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/NativeMethods.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/NativeMethods.cs
@@ -122,38 +122,24 @@ namespace System.Windows.Forms
             public int FlagsEx;
         }
 
-        [StructLayout(LayoutKind.Sequential)]
-        public class ENLINK
+        [StructLayout(LayoutKind.Sequential, Pack = 4)]
+        public struct ENLINK
         {
             public User32.NMHDR nmhdr;
             public int msg;
-            public IntPtr wParam = IntPtr.Zero;
-            public IntPtr lParam = IntPtr.Zero;
+            public IntPtr wParam;
+            public IntPtr lParam;
             public Richedit.CHARRANGE charrange;
         }
 
-        [StructLayout(LayoutKind.Sequential)]
-        public class ENLINK64
-        {
-            [MarshalAs(UnmanagedType.ByValArray, SizeConst = 56)]
-            public byte[] contents = new byte[56];
-        }
-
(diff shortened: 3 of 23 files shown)


src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Message.cs
```diff
@@ -85,6 +85,11 @@ public IntPtr Result
         /// </summary>
         public object? GetLParam(Type cls) => Marshal.PtrToStructure(LParamInternal, cls);
 
+        /// <summary>
+        ///  Gets the <see cref="LParam"/> value, and converts the value to an object of given type.
+        /// </summary>
+        public T? GetLParam<T>() => Marshal.PtrToStructure<T>(LParamInternal);
```
RussKie, line 91: This is a new public API, and _if_ it's really needed it needs to go through the API review process. However, I don't believe this API is necessary. Please read on...

src/System.Windows.Forms/src/System/Windows/Forms/ToolTip.cs
```diff
@@ -2052,7 +2052,7 @@ private void WmMouseActivate(ref Message message)
         /// </summary>
         private void WmWindowFromPoint(ref Message message)
         {
-            var point = (Point)message.GetLParam(typeof(Point));
+            var point = message.GetLParam<Point>();
```
RussKie, line 2055: We can do an unsafe cast
image

kant2002: You are right. Too much automatic replacements hurting my reasoning capabilities.

RussKie:
I'm happy to take ToolTip changes, but RTB-related changes require tests and more thinking. 

I guess there was a reason why `ENLINK` et al were duplicated for x64, this comment provides some insight but sadly lacks details, which we need to find:
https://github.com/dotnet/winforms/blob/1616cf133035beadf3c97573c8abecbdb6750967/src/System.Windows.Forms/src/System/Windows/Forms/RichTextBox.cs#L3278-L3279

For example, packing may differ between platforms, and we've already stepped on this rake before in https://github.com/dotnet/winforms/issues/2814 (fixed in https://github.com/dotnet/winforms/pull/2851). 

To have better a comfort level with the RTB-related changes we'd need to check native definitions and write tests that verify the struct layouts similar to https://github.com/dotnet/winforms/pull/2851.

RussKie:
Related work https://github.com/dotnet/winforms/pull/6399

kant2002:
@RussKie think I’m done. Take a look at the approach

kant2002:
Maybe @Tanya-Solyanik or @dreddy-work has time for this?

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/NativeMethods.cs
```diff
@@ -122,25 +122,18 @@ public class OPENFILENAME_I
             public int FlagsEx;
         }
 
-        [StructLayout(LayoutKind.Sequential)]
-        public class ENLINK
+        [StructLayout(LayoutKind.Sequential, Pack = 4)]
```
JeremyKuhne, line 125: We should call out that everything from `richedit.h` is packed at 4.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/NativeMethods.cs
```diff
@@ -122,25 +122,18 @@ public class OPENFILENAME_I
...
+        public struct ENLINK
```
JeremyKuhne, line 126: Can you please move these out of `NativeMethods` to `Interop.Richedit`? `NativeMethods` is a TODO chunk of code.

JeremyKuhne (changes requested):
Thanks! If you could move these types to the new `Interop` location that would be fantastic.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/NativeMethods.cs
```diff
@@ -122,25 +122,18 @@ public class OPENFILENAME_I
...
         {
             public User32.NMHDR nmhdr;
             public int msg;
-            public IntPtr wParam = IntPtr.Zero;
-            public IntPtr lParam = IntPtr.Zero;
+            public IntPtr wParam;
```
JeremyKuhne, line 130: Can you change these to nint/nuint too please? We need to get off of IntPtr for things that are actually nint/nuint to avoid checked.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/NativeMethods.cs
kant2002: Do you want apply that to all structs in `Interop.Richedit` and write tests for them in similar manner?

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/NativeMethods.cs
kant2002: Hope I'm done what you ask me to do.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/NativeMethods.cs
RussKie: That'd be great 👍

RussKie (approved):

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/NativeMethods.cs
kant2002: Can I do that in follow up PR?

kant2002:
Hey, I even found one funny bug with these new tests.

src/System.Windows.Forms/tests/InteropTests/NativeTests/RichTextBoxTests.cpp
```diff
@@ -0,0 +1,119 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#include "testhelpers.h"
+#include <cassert>
+#include <richedit.h>
+
+extern "C" __declspec(dllexport) void WINAPI RichTextBox_EnLink(ENLINK* value)
+{
+    value->nmhdr.code = 132;
```
Tanya-Solyanik, line 11: Would be nice to share the magic numbers between this code and the .cs code. Though I don't have any idea how to do it other than a text file.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/NativeMethods.cs
Tanya-Solyanik: >We should call out that everything from richedit.h is packed at 4.

@kant2002  - I did not see a comment to this effect added, could you please add one?

src/System.Windows.Forms/tests/InteropTests/NativeTests/RichTextBoxTests.cpp
RussKie: A comment in the managed test code stating these numbers come from the unmanaged side (with a link to the file) would suffice imo.

src/System.Windows.Forms.Primitives/src/Interop/Richedit/Interop.CHARRANGE.cs
```diff
@@ -8,6 +8,7 @@ internal partial class Interop
 {
     internal static partial class Richedit
     {
+        // All structures from richedit.h is packed at 4
```
RussKie, line 11: I suggest going one step further and create a const that we can reuse across the lot. This way we don't have to dup the comment and the value across multiple fies.
A good question where to define the said const... Maybe put it in src/Interop/Richedit/Interop.RichEditPack.cs file?

src/System.Windows.Forms.Primitives/src/Interop/Richedit/Interop.GTL.cs
```diff
@@ -14,7 +14,7 @@ public enum GTL : uint
             PRECISE = 2,
             CLOSE = 4,
             NUMCHARS = 8,
-            NUMBYTES = 6,
+            NUMBYTES = 16,
```
RussKie, line 17: 🎉

src/System.Windows.Forms/tests/InteropTests/RichTextBoxTests.cs
```diff
@@ -0,0 +1,201 @@
...
+
+using System.Runtime.InteropServices;
+using Xunit;
+using static Interop;
+using static Interop.Richedit;
+
+namespace System.Windows.Forms.Interop.Tests;
+
+// Magic numbrs in this file are coming from unmanaged part of the test
+// at https://github.com/dotnet/winforms/blob/main/src/System.Windows.Forms/tests/InteropTests/NativeTests/RichTextBoxTests.cpp
```
RussKie, line 13:
```suggestion
// at src/System.Windows.Forms/tests/InteropTests/NativeTests/RichTextBoxTests.cpp
```

RussKie (commented):
👍 with few nits

RussKie (approved):

RussKie:
Thank you

PR #6971, opened 2022-04-06
Fix nested/User control scaling.

### LINKED ISSUE #6381: Anchors of user control don't work when scaling on high DPI
* .NET Core Version: 
6.0

No

**Problem description:**

When components are inside a user control, their anchors work incorrect when scaling. It affects a project a lot, because using docking instead of anchors for components inside a user control is difficult in some cases.
Scale 100%
scale100
Scale 150%
scale150

**Expected behavior:**
Anchors works well like it was on classic .NET.

**Minimal repro:**
Sample app is attached. Run it when screen is scaled 150% to reproduce the bug.
[ControlScalingBug.zip](https://github.com/dotnet/winforms/files/7756538/ControlScalingBug.zip)

dreddy-work:
Fixes #6381.

Nested controls are disabled for scaling once their parent layout is finished.
However, Parent might be reparented and could go through a new layout based on
 its parent's scale dimensions.

We are maintaining multiple flags around scaling here and can be cleaned but require full testing. Making a scoped fix here to limit the risk.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6971)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
@@ -637,18 +637,6 @@ namespace System.Windows.Forms
         }
 
         /// <summary>
-        ///  Recursively enables required scaling from the given control
-        /// </summary>
-        private void EnableRequiredScaling(Control start, bool enable)
-        {
-            start.RequiredScalingEnabled = enable;
-            foreach (Control c in start.Controls)
-            {
-                EnableRequiredScaling(c, enable);
-            }
-        }
-
-        /// <summary>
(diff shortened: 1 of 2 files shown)


JeremyKuhne (approved):
Looks ok. Presumably you'll give CTI a heads up on this one.

RussKie (commented):
👍

Tanya-Solyanik (approved):

dreddy-work:
Testing found that nested controls inside `TabControl` still has issues. Investigatin.

dreddy-work:
Anchoring inside user control needs bigger change and can not make it to this release.  Taking this PR as-is.

PR #6972, opened 2022-04-06
Deprecate Permonitor mode intended only for Windows 8.1

dreddy-work:
WinForms never fully supported this mode. Windows 10 and above versions use 'PermonitorV2' mode.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6972)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/HighDpiMode.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/HighDpiMode.cs
@@ -37,8 +37,9 @@ namespace System.Windows.Forms
         SystemAware,
 
         /// <summary>
-        ///  The window checks for DPI when it's created and adjusts scale factor when the DPI changes.
+        /// The window checks for DPI when it's created and adjusts scale factor when the DPI changes.
         /// </summary>
+        [Obsolete("Permonitor mode is applicable only on Windows 8.1. Winforms does not support this mode", true)]
         PerMonitor,
 
         /// <summary>


src/System.Windows.Forms.Primitives/src/System/Windows/Forms/HighDpiMode.cs
```diff
@@ -37,8 +37,9 @@ public enum HighDpiMode
         SystemAware,
 
         /// <summary>
-        ///  The window checks for DPI when it's created and adjusts scale factor when the DPI changes.
+        /// The window checks for DPI when it's created and adjusts scale factor when the DPI changes.
         /// </summary>
+        [Obsolete("Permonitor mode is applicable only on Windows 8.1. Winforms does not support this mode", true)]
```
RussKie, line 42:
```suggestion
        [Obsolete("PerMonitor mode is applicable only on Windows 8.1", true)]
```

1. What would be the experience if an app is build/run on Windows 8.1?
2. The obsoletion must contain diagnostics as described in https://github.com/dotnet/designs/blob/main/accepted/2020/better-obsoletion/better-obsoletion.md. https://github.com/dotnet/runtime/blob/main/docs/project/list-of-diagnostics.md may be useful as well.

RussKie (changes requested):
Please raise a formal API proposal discussing the reasons and possible outcomes. The proposal must go through the Team triage first, and possibly vetted by the API Review Board.

dreddy-work:
Will be looking before next year.

PR #2096, opened 2019-10-16
Implementing UIA supporting for Form control

vladimir-krestov:
## Issue description
- Form doesn't support UIA so it blocks #2074 because Narrator can't announce tooltip text if no controls on a form don't support UIA provider.

## Proposed changes
- Set SupportsUiaProviders property as true
- Use FormAccessibleObject as AccessibleObject instead default

**This allows customizing Form AccessibleObject how we want. 
For example, fix UIA navigation in the accessibility tree. It changes the focus order of Narrator. 
Form AccessibleObject works as before.**

## Customer Impact

- <del>Improved Bounds</del> No

## Regression? 

- No

## Risk

- No

## Test methodology 

- Manual UI testing

## Accessibility testing  
- Using Inspect tool

 

## Test environment(s) 
- .NET Core Version: 3.1.0-preview1.19458.7
- Microsoft Windows [Version 10.0.18362.356]

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/winforms/pull/2096)

PR #6974, opened 2022-04-06
Fixed Narrator announcement of an empty form or its inner controls when opening

### LINKED ISSUE #6896: Screenreaders don't announce window title when application opens.
### .NET version

6.0

Yes

I haven't checked .net core, but it did the same in .net 5.0.

### Issue description

When an application loads, the form is in focus, but the active screen reader doesn't announce it's title when it should.
I have tried several workarounds, non of them make sense and actually should be done because this should work without any workarounds, but it didn't fix the issue. I have used nvda and windows10 narrator screen readers.

### Steps to reproduce

1. Create a standard windows forms app using the template without changing it.
2. Get a screen reader ([nvda](https://nvaccess.org) or windows10/11 narrator (ctrl+windows+enter) should do), launch it.
3. Build and launch.
4. The window opens and is in focus, but the screen reader doesn't announce it's title (the Form.Text property) when it pops up, but it should. To read the window title, you should alt+tab out then back again, as window title checking commands (Insert/capslock+t for nvda and capslock+t for narrator report wrong information).
I've made sure that the issue is reproducible not only on my pc, but on others as well.

vladimir-krestov:
Fixes #6896
This is a regression from .NET 3.1 after #2096. (.NET Core 3.0 and .NET Framework work well)

## Proposed changes

- Raise the focus event for a form or its active control, when the form is opening. Thereby a screen reader will focus on the form and announce its title.
**We can't raise the UIA focus event in `OnGotFocus`, `OnShown`, `OnLoad` event handlers, because when they work AccessibleObject is not created yet and we can't be sure that a UIA tool (eg. Narrator) is working. It's a Form control issue only, when it's showing and if the form is empty. So we have to use `OnAccessibleObjectCreated` and check a focus state there.
If the form has enabled and visible controls, we need to do nothing.**

## Customer Impact

- Now a user can know from Narrator that a form has been opened

## Regression? 

- Yes (from #2096)

## Risk

- Minimal

## Test methodology 

- Manual testing
- CTI

## Accessibility testing 
- Using Narrator, AI, and Inspect.
   Tested the following cases:
     - Open an empty form - Narrator should focus on the form and announce its title
     - Open a form with one control - Narrator should focus on this control and announce the form title and then the control info. Also checked, if a control, that overrides `OnGotFocus`, doesn't conflict with this fix raising focus event:
        - ListBox
        - DateTimePicker
        - ListView
        - MonthCalendar
        - TabControl
     - Open a form with several controls - Narrator should focus on an active control (usually, that has a minimal TabIndex) and announce the form title and then the active control info.
     - Open a form with one disabled or invisible control (so we can't focus on it) - Narrator should focus on the form and announce its title
    

## Test environment(s) 

- Windows 11
- .NET 7.0.0-preview.4.22179.1

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6974)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
@@ -3087,7 +3087,23 @@ namespace System.Windows.Forms
 
         protected override AccessibleObject CreateAccessibilityInstance()
         {
-            return new FormAccessibleObject(this);
+            AccessibleObject accessibleObject = new FormAccessibleObject(this);
+
+            // Try to raise UIA focus event for the form, if it's focused.
+            // Try it after the accessible object creation, because a screen reader
+            // gets the accessible object after "OnGotFocus", "OnLoad", "OnShown" handlers,
+            // and the object is not created yet, when these methods work, so we can't raise
+            // the event while the object is not created. It's the Form control's feature only,
+            // for the rest controls the accessibility tree will be built, when they get focus.
+            // This case works for an empty form or a form with disabled or invisible controls
+            // to have consistent behavior with .NET Framework.
+            // If the form has any control (ActiveControl is true), a screen reader will focus on it instead.
+            if (Focused && ActiveControl is null)
+            {
+                accessibleObject.RaiseAutomationEvent(UiaCore.UIA.AutomationFocusChangedEventId);
+            }
+
+            return accessibleObject;
         }
 
         [EditorBrowsable(EditorBrowsableState.Advanced)]
@@ -3946,6 +3962,19 @@ namespace System.Windows.Forms
             }
 
             base.OnFontChanged(e);
+        }
+
+        protected override void OnGotFocus(EventArgs e)
+        {
+            base.OnGotFocus(e);
+
+            // Raise the UIA focus event for an empty form (form with no ActiveControl),
+            // when it gets focus to a screen reader can focus on it and announce its title text.
+            // If the form has any control, a screen reader will focus on it instead.
+            if (Focused && IsAccessibilityObjectCreated && ActiveControl is null)
+            {
+                AccessibilityObject.RaiseAutomationEvent(UiaCore.UIA.AutomationFocusChangedEventId);
+            }
         }
 
         /// <summary>
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Form.FormAccessibleObjectTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Form.FormAccessibleObjectTests.cs
@@ -74,5 +74,84 @@ namespace System.Windows.Forms.Tests
             Assert.Equal(AccessibleRole.Client, actual);
             Assert.False(form.IsHandleCreated);
         }
+
+        [WinFormsTheory]
+        [InlineData("null")] // The form is empty
+        [InlineData("invisible")]  // The form has invisible control (looks empty)
+        [InlineData("disabled")] // The form has disabled control
+        public void FormAccessibleObject_RaiseFocusEvent_WhenFormGetsFocus_WithoutActiveControl(string controlCase)
+        {
+            using Form form = new FocusEventsCounterForm();
+            using Control control = controlCase switch
+            {
+                "null" => null,
+                "invisible" => new Button() { Visible = false },
+                "disabled" => new Button() { Enabled = false },
+                _ => null
+            };
+            form.Controls.Add(control);
+            form.CreateControl(true);
+            var accessibleObject = (FocusEventsCounterFormAccessibleObject)form.AccessibilityObject;
+
+            Assert.NotNull(accessibleObject);
+            Assert.Equal(0, accessibleObject.RaiseAutomationFocusEventCallsCount);
+            Assert.True(form.IsHandleCreated);
+
+            form.Visible = true;
+            form.Focus();
+
+            Assert.Null(form.ActiveControl);
+            Assert.Equal(1, accessibleObject.RaiseAutomationFocusEventCallsCount);
+        }
+
+        [WinFormsFact]
+        public void FormAccessibleObject_RaiseFocusEvent_WhenFormGetsFocus_WithActiveControl()
+        {
+            using Form form = new FocusEventsCounterForm();
+            using Button control = new();
+            form.Controls.Add(control);
+            form.CreateControl(true);
+            var accessibleObject = (FocusEventsCounterFormAccessibleObject)form.AccessibilityObject;
+
+            Assert.NotNull(accessibleObject);
+            Assert.Equal(0, accessibleObject.RaiseAutomationFocusEventCallsCount);
+            Assert.True(form.IsHandleCreated);
+
+            form.Visible = true;
+            control.Visible = true;
+            form.Focus();
+
+            Assert.NotNull(form.ActiveControl);
+
+            // The child control gets the focus changed event instead of the form.
+            // Native control does it itself, so a screen reader should focus on the inner control.
+            Assert.Equal(0, accessibleObject.RaiseAutomationFocusEventCallsCount);
+        }
+
+        private class FocusEventsCounterForm : Form
+        {
+            protected override AccessibleObject CreateAccessibilityInstance()
+                => new FocusEventsCounterFormAccessibleObject(this);
+        }
+
+        private class FocusEventsCounterFormAccessibleObject : FormAccessibleObject
+        {
+            public FocusEventsCounterFormAccessibleObject(Form owner) : base(owner)
+            {
+                RaiseAutomationFocusEventCallsCount = 0;
+            }
+
+            public int RaiseAutomationFocusEventCallsCount { get; private set; }
+
+            internal override bool RaiseAutomationEvent(UiaCore.UIA eventId)
+            {
+                if (eventId == UiaCore.UIA.AutomationFocusChangedEventId)
+                {
+                    RaiseAutomationFocusEventCallsCount++;
+                }
+
+                return base.RaiseAutomationEvent(eventId);
+            }
+        }
     }
 }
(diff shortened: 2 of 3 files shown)


Tanya-Solyanik (approved):
Looks good

RussKie:
Has it been tested by the CTI?

vladimir-krestov:
> Has it been tested by the CTI?

Not yet, I need to make sure, that we need to raise the focus event for an active control. Because Narrator doesn't focus on an active control in different cases.

vladimir-krestov:
@Tanya-Solyanik, @RussKie, I cardinally changed the implementation. Careful testing showed, that raising the UIA focus event inside `OnShown` works very unstable and often a screen reader doesn't focus on a form. It's because the form's accessible object is created after `OnGotFocus`, `OnShown`, `OnLoad` methods work and `IsAccessibleObjectCreated == false` when they work. This is an edge case, when an empty form (or contains invisble or disabled controls inside) is just opened. So I have to use `OnAccessibleObjectCreated` handler. But it may look incorrect. Please check it.
If the form has any controls, Narrator focuses on them correctly (except some cases).

Tanya-Solyanik:
@vladimir-krestov  - why did you add "no-merge" label?

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -499,6 +499,11 @@ public AccessibleObject AccessibilityObject
                     accessibleObject = CreateAccessibilityInstance();
 
                     Properties.SetObject(s_accessibilityProperty, accessibleObject);
+
+                    if (accessibleObject is not null)
+                    {
+                        OnAccessibilityObjectCreated(EventArgs.Empty);
```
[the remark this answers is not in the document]
vladimir-krestov: Yes, move it into `Form.CreateAccessibilityInstance`

vladimir-krestov:
CTI approved the fix ✔️

src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
```diff
@@ -3948,6 +3964,19 @@ protected override void OnFontChanged(EventArgs e)
             base.OnFontChanged(e);
         }
 
+        protected override void OnGotFocus(EventArgs e)
+        {
+            base.OnGotFocus(e);
+
+            // Raise the UIA focus event for an empty form (there is no any ActiveControl),
```
Tanya-Solyanik, line 3971:
```suggestion
            // Raise the UIA focus event for an empty form (form with no ActiveControl),
```

Tanya-Solyanik (commented):
Could you please add a unit test?

vladimir-krestov:
> Could you please add a unit test?

It seems, we can't create unit tests for `CreateAccessibilityInstance`, because we can't override the object, that is not created yet.
But added tests for `OnGotFocus`

Tanya-Solyanik:
FormAccessibleObject is internal, are you saying you can't inherit from it?

vladimir-krestov:
> FormAccessibleObject is internal, are you saying you can't inherit from it?

I mean, there is no way to override `CreateAccessibilityInstance` in unit tests to register `RaiseAutomationEvent` calls count for `FormAccessibleObject`. This class doesn't have this registration mechanism. Even if we override `CreateAccessibilityInstance`, call `base.CreateAccessibilityInstance` and return `SubFormAccessibleObject` there, anyway call count will be 0.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Form.FormAccessibleObjectTests.cs
```diff
@@ -74,5 +74,84 @@ public void FormAccessibleObject_Role_IsClient_ByDefault()
...
+        {
+            protected override AccessibleObject CreateAccessibilityInstance()
+                => new SubFormAccessibleObject(this);
+        }
+
+        private class SubFormAccessibleObject : FormAccessibleObject
+        {
+            public SubFormAccessibleObject(Form owner) : base(owner)
+            {
+                RaiseAutomationEventCallsCount = 0;
```
Tanya-Solyanik, line 145: You are counting all events, not the specific one you had added in this fix.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Form.FormAccessibleObjectTests.cs
```diff
@@ -74,5 +74,84 @@ public void FormAccessibleObject_Role_IsClient_ByDefault()
...
+            finally
+            {
+                control?.Dispose();
+            }
+        }
+
+        [WinFormsFact]
+        public void FormAccessibleObject_RaiseFocusEvent_WhenFormGetsFocus_WithActiveControl()
+        {
+            using Form form = new SubForm();
```
Tanya-Solyanik, line 115: @RussKie  - do you mind getting rid of the rule about decorating test controls with prefix `Sub`. Not only is it an abbreviation, but it does not improve code readability. We need specific names that describe what the control is testing.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Form.FormAccessibleObjectTests.cs
```diff
@@ -74,5 +74,84 @@ public void FormAccessibleObject_Role_IsClient_ByDefault()
...
+            Assert.Equal(0, subAccessibleObject.RaiseAutomationEventCallsCount);
+
+            Assert.True(form.IsHandleCreated);
+
+            form.Visible = true;
+            control.Visible = true;
+            form.Focus();
+
+            Assert.NotNull(form.ActiveControl);
+            Assert.Equal(0, subAccessibleObject.RaiseAutomationEventCallsCount);
```
Tanya-Solyanik, line 131: nit: add a comment that the child control got focus changed event instead of the form. Or better yet, add a fucus changed count to the child control.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Form.FormAccessibleObjectTests.cs
```diff
@@ -74,5 +74,84 @@ public void FormAccessibleObject_Role_IsClient_ByDefault()
...
+        }
+
+        [WinFormsFact]
+        public void FormAccessibleObject_RaiseFocusEvent_WhenFormGetsFocus_WithActiveControl()
+        {
+            using Form form = new SubForm();
+            using Button control = new();
+            form.Controls.Add(control);
+            form.CreateControl(true);
+            SubFormAccessibleObject subAccessibleObject = (SubFormAccessibleObject)form.AccessibilityObject;
```
Tanya-Solyanik, line 119:
```suggestion
            var subAccessibleObject = (SubFormAccessibleObject)form.AccessibilityObject;
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Form.FormAccessibleObjectTests.cs
```diff
@@ -74,5 +74,84 @@ public void FormAccessibleObject_Role_IsClient_ByDefault()
...
+
+            Assert.True(form.IsHandleCreated);
+
+            form.Visible = true;
+            control.Visible = true;
+            form.Focus();
+
+            Assert.NotNull(form.ActiveControl);
+            Assert.Equal(0, subAccessibleObject.RaiseAutomationEventCallsCount);
+            Assert.True(form.IsHandleCreated);
```
Tanya-Solyanik, line 132: Not sure what you are testing here, since Handle creation was tested on line 124, and there is no reason for it to get destroyed yet.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Form.FormAccessibleObjectTests.cs
RussKie: We don't have any specific rules about "Sub" prefix, I think this naming convention may have organically formed as we've been building the test suite.
`SubForm` is a subclass of a `Form` required for the test. 
I'm not sure what the issue is tbh.

Tanya-Solyanik: It's not clear what this control is doing.

vladimir-krestov: Named it as `FocusEventsCounterForm`

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Form.FormAccessibleObjectTests.cs
vladimir-krestov: Added the comment. But we can't count events for an inner control, because we don't call the method for it. Native control do it itself.

vladimir-krestov:
@RussKie, @dreddy-work, @Tanya-Solyanik, what do you think about `try-finally` in `FormAccessibleObject_RaiseFocusEvent_WhenFormGetsFocus_WithoutActiveControl` test? Will it work for theory objects?

Tanya-Solyanik:
>, what do you think about try-finally in FormAccessibleObject_RaiseFocusEvent_WhenFormGetsFocus_WithoutActiveControl test? Will it work for theory objects?

You are not re-using control that you are freeing, seems OK  to me. Is it possible to make this theory private?

Tanya-Solyanik (dismissed):
@Olina-Zhang - could you please test this change?

RussKie:
> what do you think about try-finally in FormAccessibleObject_RaiseFocusEvent_WhenFormGetsFocus_WithoutActiveControl test? Will it work for theory objects?


Yes, it would work in principle. Though we shouldn't dispose a control that was given to us. And in general we should avoid passing controls as theory arguments.Message ID: ***@***.***>

vladimir-krestov:
> what do you think about try-finally in FormAccessibleObject_RaiseFocusEvent_WhenFormGetsFocus_WithoutActiveControl test? Will it work for theory objects?
> Yes, it would work in principle. Though we shouldn't dispose a control that was given to us. And in general we should avoid passing controls as theory arguments.

I can rework it to something like that:
```cs
[WinFormsTheory]
[InlineData("null")]
[InlineData("invisible")]
[InlineData("disabled")]
public void FormAccessibleObject_RaiseFocusEvent_WhenFormGetsFocus_WithoutActiveControl(string case)
{
     using Form form = new FocusEventsCounterForm();
     using Control control = case switch {
        "null" => null,
        "invisible" => new Button() { Visible = false },
        "disabled" => new Button() { Enabled = false },
        _ => null
     };
     form.Controls.Add(control);
     ...
```

Should I rework it or keep it as is?

RussKie:
It's not great, but in this situation i think it'd be better to go with this idea.
👍

Tanya-Solyanik (approved):

PR #6978, opened 2022-04-06
Workaround for empty finalizers in DataGridView.

### LINKED ISSUE #6858: Empty finalizers
* .NET Core Version: _7.0 now_

* Have you experienced this same bug with .NET Framework?: _probably yes_

**Problem description:**

List of empty finalizers through `Dispose(false)` that doing nothing:
~[ResXResourceReader](https://github.com/dotnet/winforms/blob/55e1755b579395e7e22fecf340905c79ccfa9b1a/src/System.Windows.Forms/src/System/Resources/ResXResourceReader.cs#L194)~ and ~[ResXResourceWriter](https://github.com/dotnet/winforms/blob/55e1755b579395e7e22fecf340905c79ccfa9b1a/src/System.Windows.Forms/src/System/Resources/ResXResourceWriter.cs#L549)~ [We leave it as is](https://github.com/dotnet/winforms/pull/7431#issuecomment-1188739332).
- [x] [ApplicationContext](https://github.com/dotnet/winforms/blob/55e1755b579395e7e22fecf340905c79ccfa9b1a/src/System.Windows.Forms/src/System/Windows/Forms/ApplicationContext.cs#L91) https://github.com/dotnet/winforms/pull/7431
- [x] [DataGridViewBand](https://github.com/dotnet/winforms/blob/55e1755b579395e7e22fecf340905c79ccfa9b1a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewBand.cs#L817) https://github.com/dotnet/winforms/pull/6978
- [x] [DataGridViewCell](https://github.com/dotnet/winforms/blob/55e1755b579395e7e22fecf340905c79ccfa9b1a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCell.cs#L1313) https://github.com/dotnet/winforms/pull/6978
- [x] [HtmlShim](https://github.com/dotnet/winforms/blob/55e1755b579395e7e22fecf340905c79ccfa9b1a/src/System.Windows.Forms/src/System/Windows/Forms/HtmlShim.cs#L107) https://github.com/dotnet/winforms/pull/6878
- [x] [HtmlShimManager](https://github.com/dotnet/winforms/blob/55e1755b579395e7e22fecf340905c79ccfa9b1a/src/System.Windows.Forms/src/System/Windows/Forms/HtmlShimManager.cs#L229) https://github.com/dotnet/winforms/pull/6878
- [x] [FeedbackRectangle](https://github.com/dotnet/winforms/blob/55e1755b579395e7e22fecf340905c79ccfa9b1a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripPanel.FeedbackRectangle.cs#L55) https://github.com/dotnet/winforms/pull/6878

**Expected behavior:**
Due that we cannot remove finalizers from an unsealed public type we can remove only `FeedbackRectangle`, `HtmlShimManager`, `HtmlShim`. And for others we can try to do [this](https://github.com/dotnet/winforms/pull/6878#issuecomment-1087197261) workaround...

Stumbled upon this by setting `DataGridView.DataSource` - old rows/columns/cells are not properly disposed (this is another issue  and I have a strong feeling that with `ContextMenuStrip` we will have a memory leak) and ending up in Freachable queue...

kirsan31:
Contribute to #6858.
This is continuation of https://github.com/dotnet/winforms/pull/6878 based on discussion.
Documentation for calling `GC.SuppressFinalize` on users subclasses also need to be updated some how...

## Test methodology

- Existing UT.
- New UT for all subclasses coverage.
- Manual.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6978)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewBand.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewBand.cs
@@ -34,10 +34,8 @@ namespace System.Windows.Forms
         {
         }
 
-        /// <summary>
-        /// Currently this finalizer is unneeded (empty). See https://github.com/dotnet/winforms/issues/6858.<br/>
-        /// All classes that are not need to be finalized contains in <see cref="DataGridViewElement.s_typesWithEmptyFinalizer"/> collection. Consider to modify it if needed.
-        /// </summary>
+        // NOTE: currently this finalizer is unneeded (empty). See https://github.com/dotnet/winforms/issues/6858.
+        // All classes that are not need to be finalized contains in DataGridViewElement.s_typesWithEmptyFinalizer collection. Consider to modify it if needed.
         ~DataGridViewBand() => Dispose(false);
 
         internal int CachedThickness { get; set; }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewBand.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewBand.cs
@@ -34,6 +34,10 @@ namespace System.Windows.Forms
         {
         }
 
+        /// <summary>
+        /// Currently this finalizer is unneeded (empty). See https://github.com/dotnet/winforms/issues/6858.<br/>
+        /// All classes that are not need to be finalized contains in <see cref="DataGridViewElement.s_typesWithEmptyFinalizer"/> collection. Consider to modify it if needed.
+        /// </summary>
         ~DataGridViewBand() => Dispose(false);
 
         internal int CachedThickness { get; set; }
@@ -824,6 +828,9 @@ namespace System.Windows.Forms
                     contextMenuStrip.Disposed -= new EventHandler(DetachContextMenuStrip);
                 }
             }
+
+            // If you are adding releasing unmanaged resources code here (disposing == false), you need to remove this class type (and all of its subclasses) from DataGridViewElement.s_typesWithEmptyFinalizer!
+            // Also consider to modify ~DataGridViewBand() description.
         }
 
         internal void GetHeightInfo(int rowIndex, out int height, out int minimumHeight)
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewElement.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewElement.cs
@@ -18,6 +18,9 @@ namespace System.Windows.Forms
         /// </summary>
         public DataGridViewElement()
         {
+            if (NeedSuppressFinalize(GetType()))
+                GC.SuppressFinalize(this);
+
             State = DataGridViewElementStates.Visible;
         }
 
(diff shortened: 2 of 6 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewElement.cs
```diff
@@ -89,5 +92,20 @@ protected void RaiseMouseWheel(MouseEventArgs e)
...
+
+        /// <summary>
+        /// Determines whether <see cref="GC.SuppressFinalize(object)" /> should be called in the constructor for the given <paramref name="type"/>.<br/>
+        /// <see cref="GC.SuppressFinalize(object)" /> must be called on known internal subclasses of <see cref="DataGridViewBand" /> / <see cref="DataGridViewCell" /> which don't need a finalizer.
+        /// </summary>
+        /// <param name="type">The type.</param>
+        /// <returns><see langword="true"/> if need to call <see cref="GC.SuppressFinalize(object)" />, <see langword="false"/> otherwise.</returns>
+        private static bool NeedSuppressFinalize(Type type)
+        {
+            return type == typeof(DataGridViewBand) || type == typeof(DataGridViewColumn) || type == typeof(DataGridViewButtonColumn) || type == typeof(DataGridViewCheckBoxColumn) ||
```
Tanya-Solyanik, line 104: This code will be hard to maintain. In the unlikely scenario where we would change one of the finalizers, we will be very likely to miss this condition. @kirsan31 - do you see a better option? At least we could add comments to each empty finalizer.

Also this code might be more readable if all these derived types are added to a static container and the constructor checks if the type is contained in there.

if (s_typesWithEmptyFinalizer.Contains(GetType()))

kirsan31: > This code will be hard to maintain. In the unlikely scenario where we would change one of the finalizers, we will be very likely to miss this condition. @kirsan31 - do you see a better option?

 I am afraid we have only 2 options here:
- One, that I've implemented - In the base class we check that type is belong to known types.
- Put the check in every constructor of known types, Like:
   ```cs
  internal DataGridViewBand()
  {
      if (GetType() == typeof(DataGridViewBand))
      {
        GC.SuppressFinalize(this);
      }
  }
   ```
   This approach will not solve the problem you mentioned, but add more troubles with UT new subclasses :(

> At least we could add comments to each empty finalizer.

Will do.

> Also this code might be more readable if all these derived types are added to a static container and the constructor checks if the type is contained in there.
> 
> if (s_typesWithEmptyFinalizer.Contains(GetType()))

Will try...

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewBand.cs
```diff
@@ -824,6 +828,9 @@ protected virtual void Dispose(bool disposing)
                     contextMenuStrip.Disposed -= new EventHandler(DetachContextMenuStrip);
                 }
             }
+
+            // If you are adding releasing unmanaged resources code here (disposing == false), you need to remove this class type (and all of its subclasses) from DataGridViewElement.s_typesWithEmptyFinalizer!
+            // Also consider to modify ~DataGridViewBand() description.
```
RussKie, line 833: Would this work?
```suggestion
#if DEBUG
            else
            {
                Debug.Fail($"Remove this class type (and all of its subclasses) from {nameof(DataGridViewElement.s_typesWithEmptyFinalizer)} and modify {nameof(~DataGridViewBand)} description");
            }
#endif
```

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewElement.cs
```diff
@@ -11,13 +11,44 @@ namespace System.Windows.Forms
     /// </summary>
     public class DataGridViewElement
     {
+        /// <summary>
+        /// These are subclasses of the <see cref="DataGridViewElement"/> for which you don't need to call the finalizer, because it's empty.
```
RussKie, line 15:
```suggestion
        /// These are subclasses of the <see cref="DataGridViewElement"/> for which we don't need to call the finalizer, because it's empty.
```

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewBand.cs
kirsan31: I think not... Because the reason of the comment -  send a developer to the right place who purposefully decides to add clearing of unmanaged resources logic to the Dispose (false). But this assert can be a bit confusing. Especially in case if we will add a new subclass (without unmanaged resources) and run test app - we will see this assert and it;s message will be completely wrong, because in this case we need **to add** new class to `DataGridViewElement.s_typesWithEmptyFinalizer`.

RussKie: The only developer who may see this assert is the one who's building the SDK and makes changes in this area - we only ship Release binaries.

> we will see this assert and it's message will be completely wrong, because in this case we need **to add** new class to `DataGridViewElement.s_typesWithEmptyFinalizer`.

Ok, so we just need to improve the wording?

kirsan31: > The only developer who may see this assert is the one who's building the SDK and makes changes in this area - we only ship Release binaries.

Yes of course.

> Ok, so we just need to improve the wording? ...or is there something else I'm missing?

In principle, yes, but I find it difficult to write this correctly and briefly 😟 It will also 100% duplicate the functionality of the new test added.

Tanya-Solyanik: > It will also 100% duplicate the functionality of the new test added.

Test sounds like a sufficient safeguard to me.

Tanya-Solyanik (dismissed):
Looks good!

RussKie:
Please fix the comments and we can take this.

kirsan31:
> Please fix the comments and we can take this.

Done, sorry for delay.

RussKie (approved):

PR #3676, opened 2020-07-30
Several fixes to Control.AccessibleName property

SergeySmirnov-Akvelon:
Fixes #3673

## Proposed changes
- Added supporting of custom AccessibleName for Form control
- Removed unneeded using "LegacyIAccessible" pattern from controls.
- Removed default names for several controls, where default names were synonyms of the control type

## Customer Impact
Customers have ability to set custom accessible name for Form control.
Now, It's app developer responsibility to set default name for several controls, and AI tooling correctly reports errors if developer had not provided AccessibleName.
- DomainUpDown;
- NumericUpdown;
- ProgressBar;
- DataGridView;
- PropertyGrid;

## Regression? 

- Yes, from .Net Framework 4.8

## Risk

- Minimal

## Test methodology 

- Unit tests
- Manual testing

## Test environment(s) 

- Microsoft Windows [Version 10.0.18363.900]
- .NET Core 5.0: 5.0.100-rc.1.20379.24

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/winforms/pull/3676)

PR #6983, opened 2022-04-07
Do not dispose imagelist bound to a toolstrip

### LINKED ISSUE #6881: Closing an old WinForm affects ToolStrip ImageLists in a new Winform
### .NET version

.Net Core 6.0.200

Yes

_No response_

### Issue description

For our WinForms Application, we have a number of global classes which are used between various forms. One of which is for ToolStrips where we store a global ImageList class, which is then loaded to each ToolStrip that is created on a form. This is to ensure only one ImageList needs to be created and icons are kept common between the ToolStrips we use throughout the application.

However, we have found that when changing forms (by calling .Show() on the new form and .Close() straight after on the old form), it appeared to set the ImageList properties to empty in the ToolStrips on the new form, meaning none of the images would show on the ToolStrip. After debugging and stepping through every part of the code that may have cause the property to set to null, I found that the .Close() call on the old form was setting the ImageList on the ToolStrips to empty.

The forms and the code behind them were not changed at all when converting the project from .Net Framework to .Net Core. This only seems to affect the .Net Core project.

I've created an example project with 2 identical forms, both having a tool strip with a single item and a button to open another form. The button image shows on Form1, and on Form2 when you open that, however the next time you click the button to switch forms, the image is no longer there.

[ImageListExample.zip](https://github.com/dotnet/winforms/files/8317886/ImageListExample.zip)

Not sure why this works on Framework and not Core. I have been able to get around the issue by loading the global ImageList each time rather than setting it at program start, but it's an extra processing step which could potentially lead to (albeit minor) slowdowns.

### Steps to reproduce

- Run the attached project
- Click button1 on Form1, notice the toolstrip at the top of Form1 and Form2 have an image (just a black square)
- Click button1 on Form2, notice the toolstrip no longer has an image displaying

RussKie:
Revert https://github.com/dotnet/winforms/pull/3676/commits/05c132f1231a5824d8e0b1db83cf51b0d095f9fa
Fixes #6881

## Proposed changes

- Do not dispose imagelist bound to a toolstrip, because the imagelist may be shared and outlive the controls it's bound to.

## Regression? 

- Yes, introduced in https://github.com/dotnet/winforms/pull/3676

## Risk

- Minimal

## Test methodology 

- Manual tests
- Automated tests

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6983)


dreddy-work (approved):

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ToolStripTests.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            _sharedImageList = new ImageList();
+            _sharedImageList.Images.Add("square", Image.FromFile("./Resources/image.png"));
+        }
+
+        // Regression test for https://github.com/dotnet/winforms/issues/6881
+        [WinFormsFact]
+        public void ToolStrip_shared_imagelist_should_not_get_disposed_when_toolstrip_does()
+        {
+            // establish a scope
```
Tanya-Solyanik, line 26: Would a helper method that creates and disposes a form work instead of relying on a comment to explain presence braces?

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ToolStripTests.cs
```diff
@@ -0,0 +1,91 @@
...
+            {
+                using Form f1 = new();
```
Tanya-Solyanik, line 28: We spell out names in the source, it makes sense to follow the same convention in tests.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ToolStripTests.cs
RussKie: Yes, DRY and all :)

Issue #2734, opened 2020-01-17
CurrentCulture in unit tests?


weltkante:
While updating the PaddingConverter tests in PR #2679 I noticed that it contains tests which only succeed on english (or similar) environments which use comma as separator. Some languages use semicolon due to comma being used as decimal separator. This leads to false-positive failing tests on developer machines.

Failing tests are [PaddingConverter_ConvertFrom_String_ReturnsExpected](https://github.com/dotnet/winforms/blob/fde3805ea206227ad01266c7d25ce0ed2df4996f/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PaddingConverterTests.cs#L41) on e.g. [this input](https://github.com/dotnet/winforms/blob/fde3805ea206227ad01266c7d25ce0ed2df4996f/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PaddingConverterTests.cs#L30) as well as [PaddingConverter_ConvertTo_String_ReturnsExpected](https://github.com/dotnet/winforms/blob/fde3805ea206227ad01266c7d25ce0ed2df4996f/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PaddingConverterTests.cs#L80). (I'm only selectively running tests locally so I'd assume there could be more instances of this problem in other tests.)

Do you think tests should be running under fixed CurrentCulture (maybe some sort of fixture)? Maybe check how corefx is solving this problem when they are testing their converters/parsers.

weltkante:
Just for the record, there are currently 105 failing tests, at least 94 of them are due to tests assuming english culture.

This also means you probably have no test coverage against culture dependent regressions in CI.

If you want an overview over which tests are failing I have a mostly working branch [here](https://github.com/dotnet/winforms/compare/master...weltkante:culture), trying to get myself unstuck because currently running tests is a pain (I can't see if I'm adding a new test failure when running tests locally) - but I wouldn't want to PR that branch, there are probably better solutions than manually identifying which tests need english culture.

Personally I'd probably prefer if all tests are run under the same culture and then have two passes over all tests in two representative cultures, say one for english and one for a culture which is sufficiently different (de-DE and es-ES are candidates that differ in at least 3 separator tokens commonly used in WinForms). That should catch most generic culture regressions, of course regressions in a specific culture are only caught by exhaustive testing of every language which has its own localized strings, probably not worth for CI that runs on every PR.

[...]

PR #6985, opened 2022-04-08
Run selected tests in en-US culture

RussKie:
Mu OS default locale is set to "en-AU", and these tests kept failing for me. Add two new XUnit attributes to specify the desired locale.

/cc: @weltkante @vladimir-krestov I know that in the past you had troubles with some locale-dependent tests on your ends. If you could run these changes and confirm that (at least) these tests pass for you - it would be awesome. 
Also, if you see any other tests that may benefit from the new decoration - please let me know (or sent a PR in :)).
Thank you.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6985)

### DIFF
--- a/src/Common/tests/TestUtilities/XUnit/UseCultureAttribute.cs
+++ b/src/Common/tests/TestUtilities/XUnit/UseCultureAttribute.cs
@@ -1,0 +1,136 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+// The original code was borrowed from https://github.com/xunit/samples.xunit/blob/93f87d5/UseCulture/UseCultureAttribute.cs
+// Licensed under http://www.apache.org/licenses/LICENSE-2.0.
+
+using System.Globalization;
+using System.Reflection;
+using System.Runtime.InteropServices;
+using Xunit.Sdk;
+
+namespace Xunit
+{
+    /// <summary>
+    ///  Apply this attribute to your test method to replace the <see cref="Thread.CurrentThread" /> <see cref="CultureInfo.CurrentCulture" /> and
+    ///  <see cref="CultureInfo.CurrentUICulture" /> with another culture.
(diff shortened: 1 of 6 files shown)


weltkante:
Thanks, I'll cross reference with my local branch I pull in to adjust broken tests (an older version is also [here](https://github.com/dotnet/winforms/compare/main...weltkante:culture) but I haven't pushed it for a long time, that was just for commenting on an issue back then). Do you want this to be the resolution for #2734? If so I can do a PR to adress the affected tests.

src/Common/tests/TestUtilities/XUnit/UseCultureAttribute.cs
```diff
@@ -0,0 +1,136 @@
...
+        /// and replaces them with the new cultures defined in the constructor.
+        /// </summary>
+        /// <param name="methodUnderTest">The method under test</param>
+        public override unsafe void Before(MethodInfo methodUnderTest)
+        {
+            _originalCulture = Thread.CurrentThread.CurrentCulture;
+            _originalUICulture = Thread.CurrentThread.CurrentUICulture;
+
+            CultureInfo.DefaultThreadCurrentCulture =
+                Thread.CurrentThread.CurrentCulture = Culture;
```
Tanya-Solyanik, line 74:
```suggestion
            CultureInfo.DefaultThreadCurrentCulture = Culture;
            Thread.CurrentThread.CurrentCulture = Culture;
```

Tanya-Solyanik (approved):
Looks good.

src/Common/tests/TestUtilities/XUnit/UseCultureAttribute.cs
```diff
@@ -0,0 +1,136 @@
...
+            Thread.CurrentThread.CurrentUICulture = UICulture;
+
+            if (SetUnmanagedUiThreadCulture)
```
dreddy-work, line 77: you may want to check current culture before setting it, if it needs to be set.

PR #6986, opened 2022-04-08
Cut working incorrectly in ToolStripTextBox

kant2002:
I assume that this is a bug which we all living with .NET 2.0
Highly unlikely this is ever would be fixed, but would like to discuss possible options.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6986)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripTextBox.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripTextBox.cs
@@ -506,7 +506,7 @@ namespace System.Windows.Forms
         public void Clear() { TextBox.Clear(); }
         public void ClearUndo() { TextBox.ClearUndo(); }
         public void Copy() { TextBox.Copy(); }
-        public void Cut() { TextBox.Copy(); }
+        public void Cut() { TextBox.Cut(); }
         public void DeselectAll() { TextBox.DeselectAll(); }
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ToolStripTextBox.cs
```diff
@@ -506,7 +506,7 @@ public bool WordWrap
         public void Clear() { TextBox.Clear(); }
         public void ClearUndo() { TextBox.ClearUndo(); }
         public void Copy() { TextBox.Copy(); }
-        public void Cut() { TextBox.Copy(); }
+        public void Cut() { TextBox.Cut(); }
```
RussKie, line 509: Looks like it's a very old bug...
https://referencesource.microsoft.com/#System.Windows.Forms/winforms/Managed/System/WinForms/ToolStripTextBox.cs,595

kant2002: I think 20-years old qualifies to be young adult. So it is by definition young.

dreddy-work: May be there wasn't many who used this operation.

dreddy-work (approved):

vladimir-krestov (approved):

PR #6989, opened 2022-04-08
Fixing ShowInTaskbar=false doesn't work in some cases (#6421)

roland5572:
Fixes  https://github.com/dotnet/winforms/issues/6421.

If you set ShowInTaskbar=false, the winforms will recreate the Handle. It will call RecreateHandleCore() from Control.cs.

In method WmNCDestroy(ref Message m) from class Form.cs, The program needs to confirm whether the handle is being recreated before setting the value of the DialogResult.

If the program is in the process of recreating the handle, setting DialogResult = DialogResult.Cancel will cause the Form to be automatically closed.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6989)

### DIFF
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
@@ -1820,6 +1820,32 @@ namespace System.Windows.Forms.Tests
             Assert.Equal(0, createdCallCount);
         }
 
+        [ActiveIssue("https://github.com/dotnet/winforms/issues/6421")]
+        [WinFormsFact]
+        public void Form_ShowInTaskBar_SetFalse_GetReturnsExpected()
+        {
+            using var control = new Form
+            {
+                ShowInTaskbar = true,
+            };
+
+            var prefValue = DialogResult.OK;
+
+            control.Load += (object sender, EventArgs e) =>
+            {
+                control.ShowInTaskbar = false;
+            };
+
+            control.Shown += (object sender, EventArgs e) =>
+            {
+                control.DialogResult = prefValue;
+            };
+
+            control.ShowDialog();
+
+            Assert.Equal(prefValue, control.DialogResult);
+        }
+
         public static IEnumerable<object[]> Visible_Set_TestData()
         {
             foreach (DialogResult dialogResult in Enum.GetValues(typeof(DialogResult)))
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
@@ -1820,30 +1820,35 @@ namespace System.Windows.Forms.Tests
             Assert.Equal(0, createdCallCount);
         }
 
-        [ActiveIssue("https://github.com/dotnet/winforms/issues/6421")]
         [WinFormsFact]
-        public void Form_ShowInTaskBar_SetFalse_GetReturnsExpected()
+        public void Form_ShowInTaskbar_SetFalse_GetReturnsExpected()
         {
-            using var control = new Form
+            // Regression test for https://github.com/dotnet/winforms/issues/6421
+
+            using var form = new Form
             {
                 ShowInTaskbar = true,
             };
 
-            var prefValue = DialogResult.OK;
+            bool expectedIsHandleCreated = true;
+            DialogResult expectedDialogResult = DialogResult.OK;
 
-            control.Load += (object sender, EventArgs e) =>
+            form.Load += (object sender, EventArgs e) =>
             {
-                control.ShowInTaskbar = false;
+                IntPtr formHandle = form.Handle;
+                form.ShowInTaskbar = false;
+
+                Assert.Equal(expectedIsHandleCreated, form.IsHandleCreated);
+                Assert.NotEqual(formHandle, form.Handle);
             };
 
-            control.Shown += (object sender, EventArgs e) =>
+            form.Shown += (object sender, EventArgs e) =>
             {
-                control.DialogResult = prefValue;
+                form.DialogResult = expectedDialogResult;
             };
 
-            control.ShowDialog();
-
-            Assert.Equal(prefValue, control.DialogResult);
+            Assert.Equal(expectedDialogResult, form.ShowDialog());
+            Assert.Equal(expectedDialogResult, form.DialogResult);
         }
 
         public static IEnumerable<object[]> Visible_Set_TestData()
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
@@ -1830,7 +1830,6 @@ namespace System.Windows.Forms.Tests
                 ShowInTaskbar = true,
             };
 
-            bool expectedIsHandleCreated = true;
             DialogResult expectedDialogResult = DialogResult.OK;
 
             form.Load += (object sender, EventArgs e) =>
@@ -1838,7 +1837,7 @@ namespace System.Windows.Forms.Tests
                 IntPtr formHandle = form.Handle;
                 form.ShowInTaskbar = false;
 
-                Assert.Equal(expectedIsHandleCreated, form.IsHandleCreated);
+                Assert.True(form.IsHandleCreated);
                 Assert.NotEqual(formHandle, form.Handle);
             };
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
@@ -6361,10 +6361,7 @@ namespace System.Windows.Forms
 
             if (Modal && _dialogResult == DialogResult.None)
             {
-                if (GetState(States.Recreate) == false)
-                {
-                    DialogResult = DialogResult.Cancel;
-                }
+                DialogResult = DialogResult.Cancel;
             }
         }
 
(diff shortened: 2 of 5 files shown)


Tanya-Solyanik:
Looks good, could you please add unit test for this case

roland5572:
@Tanya-Solyanik Thank you. I have tested it on my machine. I am new to this, do you have a spec document so I can learn how to make unit tests that conform to the .net spec? Because this change actually only adjusts one line of the statement. I'm not sure about the test system's rules for dealing with such changes.

Tanya-Solyanik:
@roland5572  - you could use this https://github.com/dotnet/winforms/blob/main/docs/testing.md as a starting point.

roland5572:
@Tanya-Solyanik Thank you. Do you prefer a unit test for current issue, or a unit test specifically tests a function that has been fixed?

Tanya-Solyanik:
>Do you prefer a unit test for current issue, or a unit test specifically tests a function that has been fixed?

Ha-ha, may I have both please?

I meant the current issue. We are aiming to accompany fixes with "regression" tests that cover specific issues being fixed. But if you have time to cover this functionality, such a PR will always be welcome.

RussKie:
I'm currently OOF, and unable to review. Just want to note that if tests need to show forms, those tests should be thought of as integration tests. And the best place for those is UIIntegrationTests project. Before adding an automated test please start with a manual test, it should be added to WinFormsControlTests project (writing project names from memory).

One of the best ways to approach a complex issue is to write a test (and have it failed) in the first commit, then add a fix in a second commit (at this point the test must pass). This approach works great if you're new to testing as well.

dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/winforms?pullRequest=6989) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/winforms?pullRequest=6989) before we can accept your contribution.<br/><br/>:x: roland5572 [sign now](https://cla.dotnetfoundation.org/dotnet/winforms?pullRequest=6989)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/winforms?pullRequest=6989) it.</sub>

roland5572:
Thank you, I have added unit test.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
```diff
@@ -1820,6 +1820,32 @@ public void Form_TransparencyKey_SetNotTopLevelWithHandle_GetReturnsExpected(boo
             Assert.Equal(0, createdCallCount);
         }
 
+        [ActiveIssue("https://github.com/dotnet/winforms/issues/6421")]
+        [WinFormsFact]
+        public void Form_ShowInTaskBar_SetFalse_GetReturnsExpected()
```
Tanya-Solyanik, line 1825: nit: `Taskbar`  - B is not capitalized

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
```diff
@@ -1820,6 +1820,32 @@ public void Form_TransparencyKey_SetNotTopLevelWithHandle_GetReturnsExpected(boo
...
+            using var control = new Form
+            {
+                ShowInTaskbar = true,
+            };
+
+            var prefValue = DialogResult.OK;
+
+            control.Load += (object sender, EventArgs e) =>
+            {
+                control.ShowInTaskbar = false;
```
Tanya-Solyanik, line 1836: Is it possible to confirm that handle had been re-created? Perhaps compare handle values? Or read the state?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
```diff
@@ -1820,6 +1820,32 @@ public void Form_TransparencyKey_SetNotTopLevelWithHandle_GetReturnsExpected(boo
...
+            };
+
+            control.Shown += (object sender, EventArgs e) =>
+            {
+                control.DialogResult = prefValue;
+            };
+
+            control.ShowDialog();
+
+            Assert.Equal(prefValue, control.DialogResult);
```
Tanya-Solyanik, line 1846: Please confirm that the handle had been created.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
```diff
@@ -1820,6 +1820,32 @@ public void Form_TransparencyKey_SetNotTopLevelWithHandle_GetReturnsExpected(boo
...
+            {
+                control.ShowInTaskbar = false;
+            };
+
+            control.Shown += (object sender, EventArgs e) =>
+            {
+                control.DialogResult = prefValue;
+            };
+
+            control.ShowDialog();
```
Tanya-Solyanik, line 1844: ShowDialog should return the same value, you could verify that as well.

Tanya-Solyanik (commented):
Thank you for adding a test! I added some nits

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
```diff
@@ -1820,6 +1820,32 @@ public void Form_TransparencyKey_SetNotTopLevelWithHandle_GetReturnsExpected(boo
...
+        [ActiveIssue("https://github.com/dotnet/winforms/issues/6421")]
+        [WinFormsFact]
+        public void Form_ShowInTaskBar_SetFalse_GetReturnsExpected()
+        {
+            using var control = new Form
+            {
+                ShowInTaskbar = true,
+            };
+
+            var prefValue = DialogResult.OK;
```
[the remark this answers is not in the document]
roland5572: I think it is preference value. If you prefer "expected", I'll use expected all in the future.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
```diff
@@ -1820,6 +1820,32 @@ public void Form_TransparencyKey_SetNotTopLevelWithHandle_GetReturnsExpected(boo
             Assert.Equal(0, createdCallCount);
         }
 
+        [ActiveIssue("https://github.com/dotnet/winforms/issues/6421")]
+        [WinFormsFact]
+        public void Form_ShowInTaskBar_SetFalse_GetReturnsExpected()
+        {
+            using var control = new Form
```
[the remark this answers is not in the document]
roland5572: I can see many "using var control = new Form" from FormTests.cs, So I'm not sure which is better. I'll use form in the future.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
roland5572: Assert.Equal(prefValue, control.DialogResult) has implemented the check. My thinking is to check the property at the end is also correct. Do you tend to check both?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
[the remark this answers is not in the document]
Tanya-Solyanik: Code style guide recommends against using abbreviations. The rest is up to you.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
roland5572: In this Issue, Handle is bound to be created successfully. This Issue is not caused by the failure to create the Handle, but in the process of creating the Handle, a statement calls the statement that automatically closes the window.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
Tanya-Solyanik: We expect them to match, so makes sense to check both

roland5572:
@Tanya-Solyanik Thank you. I have modified all the code according to your suggestion, and added a manual test.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
```diff
@@ -1820,6 +1820,37 @@ public void Form_TransparencyKey_SetNotTopLevelWithHandle_GetReturnsExpected(boo
...
+
+            bool expectedIsHandleCreated = true;
+            DialogResult expectedDialogResult = DialogResult.OK;
+
+            form.Load += (object sender, EventArgs e) =>
+            {
+                IntPtr formHandle = form.Handle;
+                form.ShowInTaskbar = false;
+
+                Assert.Equal(expectedIsHandleCreated, form.IsHandleCreated);
```
Tanya-Solyanik, line 1841: nit: `Assert.True(form.IsHandleCreated)`

Tanya-Solyanik (dismissed):
Thank you, Looks good

Tanya-Solyanik (approved):
Thank you!

AraHaan:
Will this go into a servicing release for .NET 6?

dreddy-work:
@AraHaan , given this has been the behavior since .NET framework, it wouldn't meet the servicing bar unless there is a business justification on customer impact.

AraHaan:
So basically winforms programs targeting .NET 6 (and might be production quality) will still be broken by this as it won't be serviced?

RussKie:
> So basically winforms programs targeting .NET 6 (and might be production quality) will still be broken by this as it won't be serviced as they cannot target .NET 7 yet until RC at least?

Since this was an existing issue apps targeting .NET 6.0 aren't affected by this issue any more than apps targeting .NET Framework.

PR #6997, opened 2022-04-10
Enable nullability in OpenFileDialog

gpetrou:
## Proposed changes

- Enable nullability in OpenFileDialog.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6997)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.Vista.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.Vista.cs
@@ -153,7 +153,7 @@ namespace System.Windows.Forms
                 dialog.GetFileTypeIndex(out uint filterIndexTemp);
                 FilterIndex = unchecked((int)filterIndexTemp);
                 _fileNames = ProcessVistaFiles(dialog);
-                if (ProcessFileNames())
+                if (ProcessFileNames(_fileNames))
                 {
                     CancelEventArgs ceevent = new CancelEventArgs();
                     if (NativeWindow.WndProcShouldBeDebuggable)
--- a/src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.Vista.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.Vista.cs
@@ -144,7 +144,7 @@ namespace System.Windows.Forms
         {
             int saveOptions = _options;
             int saveFilterIndex = FilterIndex;
-            string?[]? saveFileNames = _fileNames;
+            string[]? saveFileNames = _fileNames;
             bool ok = false;
 
             try
--- a/src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.Vista.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.Vista.cs
@@ -138,13 +138,13 @@ namespace System.Windows.Forms
             return ret;
         }
 
-        private protected abstract string[] ProcessVistaFiles(Interop.WinFormsComWrappers.FileDialogWrapper dialog);
+        private protected abstract string?[] ProcessVistaFiles(WinFormsComWrappers.FileDialogWrapper dialog);
 
-        private bool HandleVistaFileOk(Interop.WinFormsComWrappers.FileDialogWrapper dialog)
+        private bool HandleVistaFileOk(WinFormsComWrappers.FileDialogWrapper dialog)
         {
             int saveOptions = _options;
             int saveFilterIndex = FilterIndex;
-            string[]? saveFileNames = _fileNames;
+            string?[]? saveFileNames = _fileNames;
             bool ok = false;
 
             try
--- a/src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.Vista.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.Vista.cs
@@ -138,7 +138,7 @@ namespace System.Windows.Forms
             return ret;
         }
 
-        private protected abstract string?[] ProcessVistaFiles(WinFormsComWrappers.FileDialogWrapper dialog);
+        private protected abstract string[] ProcessVistaFiles(WinFormsComWrappers.FileDialogWrapper dialog);
 
         private bool HandleVistaFileOk(WinFormsComWrappers.FileDialogWrapper dialog)
         {
@@ -250,11 +250,11 @@ namespace System.Windows.Forms
             return extensions.ToArray();
         }
 
-        private protected static string? GetFilePathFromShellItem(IShellItem item)
+        private protected static string GetFilePathFromShellItem(IShellItem item)
         {
             HRESULT hr = item.GetDisplayName(SIGDN.DESKTOPABSOLUTEPARSING, out string? filename);
             hr.ThrowIfFailed();
-            return filename;
+            return filename!;
         }
 
         private readonly FileDialogCustomPlacesCollection _customPlaces = new();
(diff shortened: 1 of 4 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/OpenFileDialog.cs
```diff
@@ -139,18 +137,23 @@ private protected override bool RunFileDialog(NativeMethods.OPENFILENAME_I ofn)
...
+
                 try
                 {
                     results.GetCount(out uint count);
                     string[] files = new string[count];
                     for (uint i = 0; i < count; ++i)
                     {
                         results.GetItemAt(i, out IShellItem item);
-                        files[unchecked((int)i)] = GetFilePathFromShellItem(item);
+                        files[unchecked((int)i)] = GetFilePathFromShellItem(item) ?? string.Empty;
```
kant2002, line 156: Isn’t this a potential breaking change for somebody?

gpetrou: I added some more changes to remove `string.Empty` usages.

src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.cs
```diff
@@ -670,21 +670,21 @@ private bool ProcessFileNames()
                 string[] extensions = FilterExtensions;
                 for (int i = 0; i < _fileNames!.Length; i++)
                 {
-                    string fileName = _fileNames[i];
+                    string? fileName = _fileNames[i];
```
dreddy-work, line 673: you could make this non-nullable and avoid `!` in rest of the places.

dreddy-work, line 673: you could make this non-nullable and avoid `!` in rest of the places.

src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.cs
```diff
@@ -804,9 +804,9 @@ private bool RunDialogOld(IntPtr hWndOwner)
             try
             {
                 _charBuffer = new UnicodeCharBuffer(FileBufferSize);
-                if (_fileNames is not null)
+                if (_fileNames is not null && _fileNames[0] is not null)
                 {
-                    _charBuffer.PutString(_fileNames[0]);
+                    _charBuffer.PutString(_fileNames[0]!);
```
dreddy-work, line 809: this change seems breaking. Did you check when we would have filename in the array as `null` ?

dreddy-work, line 809: this change seems breaking. Did you check when we would have filename in the array as `null` ?

dreddy-work: Never mind i see native method from `ProcessVistaFiles` that could return null.  `!` is necessary even after checking for not null?

dreddy-work: Never mind i see native method from `ProcessVistaFiles` that could return null.  `!` is necessary even after checking for not null?

src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.cs
RussKie: I'd do a check for null and exit early in this method.

src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.cs
RussKie, line 809: Use `FileName` instead.

src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.Vista.cs
```diff
@@ -138,13 +138,13 @@ private FOS GetOptions()
             return ret;
         }
 
-        private protected abstract string[] ProcessVistaFiles(Interop.WinFormsComWrappers.FileDialogWrapper dialog);
+        private protected abstract string?[] ProcessVistaFiles(WinFormsComWrappers.FileDialogWrapper dialog);
```
RussKie, line 141: I don't think this is correct. Which path do you see that could result in a null? I think `GetFilePathFromShellItem` _may_ be annotated incorrectly, as it returns the string is the call succeeded (see https://docs.microsoft.com/windows/win32/api/shobjidl_core/nf-shobjidl_core-ishellitem-getdisplayname).

RussKie (changes requested):

src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.Vista.cs
gpetrou: You are right. Since we throw on failure, we can use the null forgiving operator.

src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.cs
gpetrou: Yes, there is a warning if we remove this. We had similar cases in previous PRs.

gpetrou: `Filename` does the same check internally, so we would duplicate the check. Perhaps we should leave this as it is?

RussKie: The whole block can be simplified to `_charBuffer.PutString(FileName);

src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.cs
```diff
@@ -29,7 +29,7 @@ public abstract partial class FileDialog : CommonDialog
         private string? _title;
         private string? _initialDirectory;
         private string? _defaultExt;
-        private string[]? _fileNames;
+        private string?[]? _fileNames;
```
RussKie, line 32: I don't believe this is correct. Where do you see we set a `null` entry?

src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.cs
RussKie: To avoid `!` operators we need to do the following:
```diff
         /// </summary>
-        private bool ProcessFileNames()
+        private bool ProcessFileNames(string[] fileNames)
         {
             if ((_options & (int)Comdlg32.OFN.NOVALIDATE) == 0)
             {
                 string[] extensions = FilterExtensions;
-                for (int i = 0; i < _fileNames!.Length; i++)
+                for (int i = 0; i < fileNames.Length; i++)
                 {
-                    string? fileName = _fileNames[i];
+                    string fileName = fileNames[i];
                     if ((_options & AddExtensionOption) != 0 && !Path.HasExtension(fileName))
                     {
                         bool fileMustExist = (_options & (int)Comdlg32.OFN.FILEMUSTEXIST) != 0;
@@ -699,7 +699,7 @@ namespace System.Windows.Forms
                             }
                         }
 
-                        _fileNames[i] = fileName;
+                        fileNames[i] = fileName;
                     }
```

src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.cs
gpetrou: In current main, we have `if (_fileNames is null || string.IsNullOrEmpty(_fileNames[0]))` in `FileName` property. So, I thought that we could have null string elements. Should I remove that and the check?

RussKie: I've checked how `_fileNames` is populated (a double check is welcome), and I couldn't see how we can end up with a null element.

RussKie (approved):

PR #7002, opened 2022-04-11
Enable nullability in ColumnHeaderCollection

gpetrou:
## Proposed changes

- Enable nullability in ColumnHeaderCollection.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7002)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListView.ColumnHeaderCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListView.ColumnHeaderCollection.cs
@@ -15,11 +15,11 @@ namespace System.Windows.Forms
         [ListBindable(false)]
         public class ColumnHeaderCollection : IList
         {
-            private readonly ListView owner;
+            private readonly ListView _owner;
 
             public ColumnHeaderCollection(ListView owner)
             {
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ListView.ColumnHeaderCollection.cs
```diff
@@ -32,16 +30,16 @@ public ColumnHeaderCollection(ListView owner)
...
                         throw new ArgumentOutOfRangeException(nameof(index), index, string.Format(SR.InvalidArgument, nameof(index), index));
                     }
 
-                    return owner._columnHeaders[index];
+                    return _owner._columnHeaders[index];
                 }
             }
 
-            object IList.this[int index]
+            object? IList.this[int index]
```
dreddy-work, line 42: when does this return null?

gpetrou: Here, we implement the interface indexer which is declared as nullable. So, we have to make this nullable as well.

dreddy-work (approved):

PR #7018, opened 2022-04-14
Enable nullability in CheckedIndexCollection

gpetrou:
## Proposed changes

- Enable nullability in CheckedIndexCollection.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7018)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListView.CheckedIndexCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListView.CheckedIndexCollection.cs
@@ -12,12 +12,12 @@ namespace System.Windows.Forms
         [ListBindable(false)]
         public class CheckedIndexCollection : IList
         {
-            private readonly ListView owner;
+            private readonly ListView _owner;
 
             /* C#r: protected */
             public CheckedIndexCollection(ListView owner)
             {
-                this.owner = owner;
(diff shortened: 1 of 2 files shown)


Tanya-Solyanik (approved):

src/System.Windows.Forms/src/System/Windows/Forms/ListView.CheckedIndexCollection.cs
```diff
@@ -14,12 +12,12 @@ public partial class ListView
...
         public class CheckedIndexCollection : IList
         {
-            private readonly ListView owner;
+            private readonly ListView _owner;
 
             /* C#r: protected */
             public CheckedIndexCollection(ListView owner)
             {
-                this.owner = owner;
+                _owner = owner;
```
RussKie, line 20: I don't believe this correct. The argument was never asserted to be non null, so it can be a null.
Since a null would lead to a multitude of NREs, we can assert and throw here - it'd be a breaking change but in line with work we've been doing.

gpetrou: I created https://github.com/dotnet/winforms/pull/7091 about this.

PR #7029, opened 2022-04-16
Replace hashtable with generic dictionary in ListView

gpetrou:
## Proposed changes

- Replace hashtable with generic dictionary in ListView.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7029)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListView.ListViewNativeItemCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListView.ListViewNativeItemCollection.cs
@@ -80,7 +80,7 @@ namespace System.Windows.Forms
 
                         if (owner.IsHandleCreated && !owner.ListViewHandleDestroyed)
                         {
-                            return (ListViewItem)owner._listItemsTable[DisplayIndexToID(displayIndex)];
+                            return owner._listItemsTable[DisplayIndexToID(displayIndex)];
                         }
                         else
                         {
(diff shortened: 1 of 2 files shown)


dreddy-work (approved):

src/System.Windows.Forms/src/System/Windows/Forms/ListView.ListViewNativeItemCollection.cs
```diff
@@ -80,7 +80,7 @@ public bool OwnerIsDesignMode
 
                         if (owner.IsHandleCreated && !owner.ListViewHandleDestroyed)
                         {
-                            return (ListViewItem)owner._listItemsTable[DisplayIndexToID(displayIndex)];
+                            return owner._listItemsById[DisplayIndexToID(displayIndex)];
```
RussKie, line 83: @gpetrou the breaking change here is very subtle and not obvious by looking at the code. I ran at this few times recently myself.
`Hashtable[non_existent_key]` returns a null ([docs](https://docs.microsoft.com/dotnet/api/system.collections.hashtable.item#property-value)), however, `Dictionary<K, V>[non_existent_key]` throws KNFE ([docs](https://docs.microsoft.com/dotnet/api/system.collections.generic.dictionary-2.item#property-value)).

We can iterate on this in the future with some added tests. I'm going to revert this change for now.

gpetrou: As I suggested in the issue, we can use `TryGetValue` instead. We have used that in other similar PRs. Would you like me to open a PR?

RussKie: I'd love that :) 
I'd want some tests though to make sure we don't accidentally break something else though.

PR #7034, opened 2022-04-17
Improve roadmap

kant2002:
I find this document via announcement to .NET 7 Preview 3
and I think it give bad vibes, and not adequately show efforts which is made in this repository.

So I take liberty and slice and dice this document with explanation how I think it should be improved and why.

- Removed *Past* section, since this is roadmap. WinForms deliver past achievements and even if it was glorious past, it may outshine Future goals. Also since it is on first place, it outshines even present
- Move *Present* to backyard. Technically that section is fine, but `Windows Forms runtime` is so bland and gives impression that there no goals for WinForms. Everybody who is familiar with corporate speak may think that "we consider everything, but do not commit to anything". I do not know how to rephrase, but want to point to the wording, which give not enough appreciation of work which happens in this repo. 
- My impression that Future and the Present is really the same, because you actively fixing bugs in Accessibility and work in High DPI support all the time. Because this project is stable, and really everything sounds like ongoing work, but maybe there some way how to present information which will have more *commitment*, if that's possible within your operational limitations. Maybe something like https://github.com/microsoft/microsoft-ui-xaml/blob/main/docs/roadmap.md#features-by-release where would be concrete goals tied to controls/features like Dark Theme support for example.

Please treat this is open critique. If you have better ideas how to improve roadmap then mine please just discard this changes.

### DIFF
--- a/docs/roadmap.md
+++ b/docs/roadmap.md
@@ -2,51 +2,19 @@
 
 This roadmap communicates priorities for evolving and extending the scope of Windows Forms. For general information regarding .NET plans, see [.NET roadmap](https://github.com/dotnet/core/blob/master/roadmap.md).
 
-* [Past](#past)
-* [Present](#present)
 * [Future](#future)
+* [Present](#present)
 
 This repository is a community effort and we welcome community feedback on our plans. The best way to give your feedback is to open an issue in this repo.
 We also invite contributions. The [up-for-grabs issues](https://github.com/dotnet/winforms/issues?q=is%3Aopen+is%3Aissue+label%3Aup-for-grabs) on GitHub are a good place to start.
 
-## Past 
-
-### Windows Forms runtime
-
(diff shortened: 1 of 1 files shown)


merriemcgaw:
Thank you so much for taking the initiative here! I'll be taking a look at this tonight as will @OliaG. You're right on track with how our roadmap is shaping up.

merriemcgaw (approved):
This looks good to me.
@RussKie , @dreddy-work ?

dreddy-work:
> This looks good to me. @RussKie , @dreddy-work ?
> 
> as @kant2002 mentioned, Should we move HDPI and accessibility from future to Present or wait until .NET 8?

We can move them to present and in progress.

dreddy-work (approved):

merriemcgaw (approved):

PR #3541, opened 2020-07-07
Update FxCopAnalyzers

gpetrou:
## Proposed changes

- Update FxCopAnalyzers.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/winforms/pull/3541)

PR #7035, opened 2022-04-17
Use Microsoft.CodeAnalysis.NetAnalyzers instead of Microsoft.CodeAnalysis.FxCopAnalyzers

gpetrou:
## Proposed changes

- Use Microsoft.CodeAnalysis.NetAnalyzers instead of Microsoft.CodeAnalysis.FxCopAnalyzers which is deprecated.
- Similarly to https://github.com/dotnet/winforms/pull/3541, I took the settings from https://github.com/dotnet/runtime/blob/release/6.0/eng/CodeAnalysis.src.globalconfig of the release/6.0 branch of the runtime repository. I disabled some rules that produced new warnings with `TODO:`.
- I also took `_SkipUpgradeNetAnalyzersNuGetWarning` from https://github.com/dotnet/roslyn-analyzers/blob/3393ecfccd4331ced6423e9b875187ee377f1ce2/Directory.Build.props#L28 to avoid build warnings as described in https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/overview.
- I removed some `TODO`s in subsequent commits.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7035)

### DIFF
--- a/eng/CodeAnalysis.test.globalconfig
+++ b/eng/CodeAnalysis.test.globalconfig
@@ -1,7 +1,7 @@
 is_global = true
 
-# Disable until https://github.com/dotnet/roslyn-analyzers/issues/4750 is fixed
-dotnet_diagnostic.AD0001.severity = none
+# Analyzer threw an exception
+dotnet_diagnostic.AD0001.severity = suggestion
 
 # Do not declare static members on generic types
 dotnet_diagnostic.CA1000.severity = none
@@ -24,7 +24,7 @@ dotnet_diagnostic.CA1008.severity = none
 # Generic interface should also be implemented
 dotnet_diagnostic.CA1010.severity = none
 
-# Abstract types should not have constructors
+# Abstract types should not have public constructors
 dotnet_diagnostic.CA1012.severity = none
 
 # Mark assemblies with CLSCompliant
@@ -100,15 +100,15 @@ dotnet_diagnostic.CA1050.severity = warning
 dotnet_diagnostic.CA1051.severity = none
 
 # Static holder types should be Static or NotInheritable
-dotnet_diagnostic.CA1052.severity = none
+dotnet_diagnostic.CA1052.severity = none # TODO: warning
 
-# Uri parameters should not be strings
+# URI-like parameters should not be strings
 dotnet_diagnostic.CA1054.severity = none
 
-# Uri return values should not be strings
+# URI-like return values should not be strings
 dotnet_diagnostic.CA1055.severity = none
 
-# Uri properties should not be strings
+# URI-like properties should not be strings
 dotnet_diagnostic.CA1056.severity = none
 
 # Types should not extend certain base types
@@ -133,10 +133,10 @@ dotnet_diagnostic.CA1064.severity = none
 dotnet_diagnostic.CA1065.severity = none
 
 # Implement IEquatable when overriding Object.Equals
-dotnet_diagnostic.CA1066.severity = none
+dotnet_diagnostic.CA1066.severity = none # TODO: warning
 
 # Override Object.Equals(object) when implementing IEquatable<T>
-dotnet_diagnostic.CA1067.severity = none
+dotnet_diagnostic.CA1067.severity = warning
 
 # CancellationToken parameters must come last
 dotnet_diagnostic.CA1068.severity = none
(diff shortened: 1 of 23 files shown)


eng/CodeAnalysis.test.globalconfig
```diff
@@ -219,18 +234,12 @@ dotnet_diagnostic.CA1712.severity = none
 # Events should not have 'Before' or 'After' prefix
 dotnet_diagnostic.CA1713.severity = none
 
-# Flags enums should have plural names
-dotnet_diagnostic.CA1714.severity = none
```
RussKie, line 223: Are we losing this check? ...or is it covered by some other new one?

RussKie (commented):
👍

eng/CodeAnalysis.test.globalconfig
gpetrou: Based on https://github.com/dotnet/roslyn-analyzers/blob/7a0e601fda115bbeb9e4ca446891b6f302a99af5/src/Text.Analyzers/Core/AnalyzerReleases.Unshipped.md seems like some rules have been removed. They don't seem to be present in the NuGet package.

eng/CodeAnalysis.test.globalconfig
```diff
@@ -256,7 +265,7 @@ dotnet_diagnostic.CA1805.severity = warning
 dotnet_diagnostic.CA1806.severity = none
 
 # Initialize reference type static fields inline
-dotnet_diagnostic.CA1810.severity = none
+dotnet_diagnostic.CA1810.severity = none # TODO: warning
```
Tanya-Solyanik, line 268: FYI: This was set to none previously in the test version of this file.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/DesignerSerializationManager.cs
```diff
@@ -848,7 +848,7 @@ Type IDesignerSerializationManager.GetType(string typeName)
                         break;
                     }
 
-                    typeName = typeName.Substring(0, dotIndex) + "+" + typeName.Substring(dotIndex + 1, typeName.Length - dotIndex - 1);
+                    typeName = string.Concat(typeName.AsSpan(0, dotIndex), "+", typeName.AsSpan(dotIndex + 1, typeName.Length - dotIndex - 1));
```
Tanya-Solyanik, line 851: This could be string interpolation instead of concat. It might be more readable.

src/System.Windows.Forms/src/System/Resources/ResXFileRef.cs
```diff
@@ -95,7 +95,7 @@ private static string PathDifference(string path1, string path2, bool compareCas
                 }
             }
 
-            return relPath.ToString() + path2.Substring(si + 1);
+            return string.Concat(relPath.ToString(), path2.AsSpan(si + 1));
```
Tanya-Solyanik, line 98: Consider string interpolation.

Tanya-Solyanik (approved):
Thank you, looks good.

eng/CodeAnalysis.test.globalconfig
gpetrou: This only produces a couple of warnings as far as I remember, but sure I can remote the TODO,

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/DesignerSerializationManager.cs
gpetrou: I used the auto-fixer for these warnings. Do you still want me to use string interpolation?

eng/CodeAnalysis.test.globalconfig
Tanya-Solyanik: Just prioritize it below other TODOs

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/DesignerSerializationManager.cs
Tanya-Solyanik: Yes, I think it is more readable. Whenever you have a chance.

PR #7046, opened 2022-04-18
Enable nullability in NotifyIcon

gpetrou:
## Proposed changes

- Enable nullability in NotifyIcon.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7046)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/NotifyIcon.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/NotifyIcon.cs
@@ -36,34 +36,34 @@ namespace System.Windows.Forms
         private const int WM_TRAYMOUSEMESSAGE = (int)User32.WM.USER + 1024;
         private static readonly User32.WM WM_TASKBARCREATED = User32.RegisterWindowMessageW("TaskbarCreated");
 
-        private readonly object syncObj = new object();
+        private readonly object _syncObj = new object();
 
-        private Icon? icon;
-        private string text = string.Empty;
-        private readonly uint id;
-        private bool added;
-        private NotifyIconNativeWindow window;
-        private ContextMenuStrip? contextMenuStrip;
-        private ToolTipIcon balloonTipIcon;
-        private string balloonTipText = string.Empty;
-        private string balloonTipTitle = string.Empty;
+        private Icon? _icon;
+        private string _text = string.Empty;
+        private readonly uint _id;
+        private bool _added;
+        private NotifyIconNativeWindow _window;
+        private ContextMenuStrip? _contextMenuStrip;
+        private ToolTipIcon _balloonTipIcon;
+        private string _balloonTipText = string.Empty;
+        private string _balloonTipTitle = string.Empty;
         private static uint s_nextId;
-        private object? userData;
-        private bool doubleClick; // checks if doubleclick is fired
+        private object? _userData;
+        private bool _doubleClick; // checks if doubleclick is fired
 
         // Visible defaults to false, but the NotifyIconDesigner makes it seem like the default is
         // true.  We do this because while visible is the more common case, if it was a true default,
         // there would be no way to create a hidden NotifyIcon without being visible for a moment.
-        private bool visible;
+        private bool _visible;
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/NotifyIcon.cs
```diff
@@ -245,15 +245,15 @@ public string Text
                     value = string.Empty;
                 }
 
-                if (value is not null && !value.Equals(text))
+                if (value is not null && !value.Equals(_text))
```
dreddy-work, line 248: unnecessary check for null.

src/System.Windows.Forms/src/System/Windows/Forms/NotifyIcon.cs
```diff
@@ -37,34 +36,34 @@ public sealed partial class NotifyIcon : Component
...
-        private ToolTipIcon balloonTipIcon;
-        private string balloonTipText = string.Empty;
-        private string balloonTipTitle = string.Empty;
+        private readonly object _syncObj = new object();
+
+        private Icon? _icon;
+        private string _text = string.Empty;
+        private readonly uint _id;
+        private bool _added;
+        private NotifyIconNativeWindow _window;
```
dreddy-work, line 45: can be readonly?

src/System.Windows.Forms/src/System/Windows/Forms/NotifyIcon.cs
```diff
@@ -395,25 +395,24 @@ protected override void Dispose(bool disposing)
...
                 {
-                    icon = null;
+                    _icon = null;
                     Text = string.Empty;
                     UpdateIcon(false);
-                    window.DestroyHandle();
-                    window = null;
-                    contextMenuStrip = null;
+                    _window.DestroyHandle();
+                    _window = null!;
```
dreddy-work, line 404: this doesn't seem right.

gpetrou: We had done something similar in https://github.com/dotnet/winforms/pull/6404. If we remove the null assignment and there is a subsequent call to a NotifyIcon method that uses _window it might behave differently.

dreddy-work: After dispose, why would we expect it to be called. We were looking to avoid these kinds of objects being declared as non nullable and end up assigning to null. @JeremyKuhne , can you weigh in on this?

Tanya-Solyanik: When accessing after dispose we'll get an `ObjectDisposed` exception, which is good enough. We don't have to assign a null after a dispose.

RussKie: It'd be good to add a test a verify this scenario though

vladimir-krestov (commented):
Looks good, except `_window` filed.

src/System.Windows.Forms/src/System/Windows/Forms/NotifyIcon.cs
RussKie: I've looked through the code, and I think the correct way to resolve this would be something like this:
1. `_window` is only assigned at the .ctor, so it can never be a null. 
2. All check `_window is null` should be replaced with `_window.Handle == IntPtr.Zero` as it essentially what's determining whether these is a native window or not.

gpetrou: What I meant was if we use something like:
```
[WinFormsFact]
public void AfterDisposeTest()
{
    var notifyIcon = new NotifyIcon();
    notifyIcon.Dispose();

    notifyIcon.Visible = true;
}
```
if we keep `_window = null!;` it will result in a NullReferenceException. If we remove `_window = null!;` it will not throw any error. If this is what we want, then I will remove it.

dreddy-work (approved):

RussKie (approved):

PR #7047, opened 2022-04-18
Enable nullability in AutoCompleteStringCollection

gpetrou:
## Proposed changes

- Enable nullability in AutoCompleteStringCollection.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7047)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/AutoCompleteStringCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/AutoCompleteStringCollection.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.Collections;
 using System.ComponentModel;
 
@@ -14,8 +12,8 @@ namespace System.Windows.Forms;
 /// </summary>
 public class AutoCompleteStringCollection : IList
 {
-    private CollectionChangeEventHandler onCollectionChanged;
-    private readonly List<string> data = new();
+    private CollectionChangeEventHandler? _onCollectionChanged;
+    private readonly List<string> _data = new();
 
     public AutoCompleteStringCollection()
     {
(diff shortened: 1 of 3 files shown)


Tanya-Solyanik (approved):
Looks good!

src/System.Windows.Forms/src/System/Windows/Forms/AutoCompleteStringCollection.cs
```diff
@@ -194,51 +193,75 @@ public void RemoveAt(int index)
...
+        object? IList.this[int index]
         {
             get
             {
                 return this[index];
             }
             set
             {
-                this[index] = (string)value;
+                if (value is string valueAsString)
```
Tanya-Solyanik, line 204: Before the change if value were not a string, this was an exception, now we handle in gracefully. @dreddy-work - thoughts?

dreddy-work: while some of the other interface overrides have return value to check on, this one ( couple below) are silently skipping. Given these are public, and documentation isn't very clear on what is expected incase of non string insertions, Maybe we should keep the existing behavior here.

Tanya-Solyanik: I agree, silent skipping opens opportunities for misuse, I would prefer an in-your-face  CastException.

Tanya-Solyanik (dismissed):
Please retain cast exceptions where applicable.

src/System.Windows.Forms/src/System/Windows/Forms/AutoCompleteStringCollection.cs
kant2002: Should this behavior be captured using test?

src/System.Windows.Forms/src/System/Windows/Forms/AutoCompleteStringCollection.cs
```diff
@@ -14,8 +12,8 @@ namespace System.Windows.Forms
     /// </summary>
     public class AutoCompleteStringCollection : IList
     {
-        CollectionChangeEventHandler onCollectionChanged;
-        private readonly ArrayList data = new ArrayList();
+        CollectionChangeEventHandler? _onCollectionChanged;
```
Tanya-Solyanik, line 15: nit: please add explicit access modifier

dreddy-work:
@gpetrou , can you address feedback here and i will help merge it?

dreddy-work:
@gpetrou, can you take a relook at this and rebase?

dreddy-work (approved):

PR #7049, opened 2022-04-19
Enable and fix CA1822 warnings in src files

gpetrou:
## Proposed changes

- Enable and fix CA1822 warnings in src files.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7049)


gpetrou:
@RussKie I see some warnings with regards to the `System.Windows.Forms.Common.TestUtilities` and `WinformsControlsTest` projects. Perhaps we need to fix something related to `IsShipping` property for these projects? We use it in https://github.com/dotnet/winforms/blob/main/eng/CodeStyle.targets#L8 to determine which globalconfig file to use. Looking at the codebase, I don't see where/how this is set.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ComponentTray.cs
```diff
@@ -1868,7 +1868,9 @@ private void PositionControl(TrayControl c)
             }
         }
 
+#pragma warning disable CA1822 // Mark members as static
         internal void RearrangeInAutoSlots(Control c, Point pos)
+#pragma warning restore CA1822 // Mark members as static
```
Tanya-Solyanik, line 1873: Please remove `#if Debug` instead.

Tanya-Solyanik (commented):
Looks good. But waiting for team to weigh in on `IsShipping` vs `IsTestProject` conditions in targets file

RussKie:
> Perhaps we need to fix something related to IsShipping property for these projects? We use it in https://github.com/dotnet/winforms/blob/main/eng/CodeStyle.targets#L8<https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fdotnet%2Fwinforms%2Fblob%2Fmain%2Feng%2FCodeStyle.targets%23L8&data=05%7C01%7Cigor.velikorossov%40microsoft.com%7Cb5063b89671642b7de4708da21cbb6af%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C637859456079977845%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=GIz55aKbBrDtdExTXHi6wuuhOPXs8IvvqaeZogpUL4c%3D&reserved=0> to determine which globalconfig file to use. Looking at the codebase, I don't see where/how this is set.

Arcade sets a number of variables. You can check the binlog under ./artifacts/logs and see when/how this is set. And you can also search for the variable in the Arcade's props/targets that are included from the NuGet package under %user%/.nuget.

gpetrou:
Looking at https://github.com/dotnet/arcade/blob/e7ede87875f41a9b3df898ae08da5ebc96e24f56/src/Microsoft.DotNet.Arcade.Sdk/tools/Version.BeforeCommonTargets.targets and https://github.com/dotnet/arcade/blob/e7ede87875f41a9b3df898ae08da5ebc96e24f56/src/Microsoft.DotNet.Arcade.Sdk/tools/Tests.props it seems expected to have the results we get. Both `TestUtilities` and `WinformsControlsTest` don't seem to satisfy the naming conditions. So, the change I did in `eng/CodeStyle.targets` might be an acceptable workaround.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ComponentTray.cs
gpetrou: I moved `Debug.Assert` instead to avoid using `IndexOf` in Release.

Tanya-Solyanik: You can inline `Indexof` and get rid of #definfes

I was suggesting removing all debugging code, line 1873 - 1876 though

eng/CodeStyle.targets
```diff
@@ -5,7 +5,7 @@
     <EditorConfigFiles Include="$(MSBuildThisFileDirectory)CodeAnalysis.src.globalconfig" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(IsShipping)' == 'false'">
+  <ItemGroup Condition="'$(IsTestProject)' == 'true' Or $(MSBuildProjectName.Contains('Test'))">
```
Tanya-Solyanik, line 8: Please add a comment that explains that TestUtilities and WinfromsControlsTest project names don't satisfy the arcade convention defined in  https://github.com/dotnet/arcade/blob/e7ede87875f41a9b3df898ae08da5ebc96e24f56/src/Microsoft.DotNet.Arcade.Sdk/tools/Tests.props. And that we test presence of word `Test` as a workaround.

gpetrou: Good idea.

Tanya-Solyanik (approved):
Thank you!

eng/CodeStyle.targets
RussKie: As noted in https://github.com/dotnet/winforms/pull/7065#discussion_r858185470 we should revert this change.

PR #6102, opened 2021-11-02
Fixing no focus border on the empty DGV in the Narrator

DmitryGorokhov:
Fixes #2178  

## Proposed changes

- Reworked the property `HasKeyboardFocusPropertyId` of DGV AO. If no inner item then focus on the entire DGV else focus on the inner cell instead of entire DGV. This works also for DGV in Virtual mode. This doesn't break focusing border on the cell if DGV is not empty. It works also when form doesn't have any row (e.g. row for edit) see this case in after screens below.
- Turned off `TablePattern` support for DGV without rows. It needs to avoid incorrect announcement of the control:
if focus on empty DGV then it has no message "enter table", although if leave then it has message "exit table". So we turn off "enter table" and "exit table" messages for the empty DGV. This approach doesn't make a new accessibility problems in Accessibility Insights.
- Added unit tests for it.

## Customer Impact

- The Narrator has correct focus on empty DGV. 

## Regression? 

- No

## Risk

- Minimal

### Before

No a blue border focus on the empty DGV (only dashed border).

before

### After

The focusing blue border is set on the empty DGV.

after fix

If DGV is non editable e.g. DGV has `ReadOnly = true` and `AllowUserToAddRows = false` properties and DGV has no rows then the Narrator set the blue border on the entire DGV.

## Test methodology

- Manually
- Unit
- CTI

## Test environment(s) 

- Dotnet version 6.0.100
- OS Version 10.0.19043

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6102)

Issue #2178, opened 2019-10-24
The empty DataGridView control cannot get a blue Narrator rectangle, just get the focus dot line


Vino-Wang:
* .NET Core Version: (.NET Core SDK 3.1 Preview 2 3.1.100.14533)
* Have you experienced this same bug with .NET Framework?: Yes

**Problem description:**
The empty DataGridView control cannot get a blue Narrator rectangle, just get the focus dot line, so no any announcement about the empty dataGridView.
But when adding some columns to the DataGridView, the blue rectangle will on the currently specific cells with corresponding announcement by Narrator reading. Maybe this issue is by design, just report here and find the final confirmation from you guys.

This issue also can reproduce with latest .NET Framework 4.8 project.

**Actual behavior:** 
The empty DataGridView control cannot get a blue Narrator rectangle, just get the focus dot line.

**Expected behavior:**
The empty DataGridView control should get a blue Narrator rectangle.

**Minimal repro:**
1.	Run the attached app,
2.	Launch the Narrator and tab to navigate to the controls on the form.
3.	Observe the narrator blue rectangle between controls and DataGridView.
[WindowsFormsApp3.zip](https://github.com/dotnet/winforms/files/3766270/WindowsFormsApp3.zip)

merriemcgaw:
@Vino-Wang is this still happening? Is Narrator reading the right stuff?

Zheng-Li01:
@merriemcgaw verified the issue with latest .NET Core SDK 3.1[3.1.200.14989] build, the issue still can preproduce.

merriemcgaw:
@Zheng-Li01 does Narrator read it properly and it just doesn't have the blue indicator, or are you saying you can't get Narrator to recognize it has focus at all?

Zheng-Li01:
@merriemcgaw, the empty DGV can't get Narrator to recognize it has focus at all. Please see below reproduce gif.
![core](https://user-images.githubusercontent.com/38325459/75226530-762ddb00-57e7-11ea-917c-02adf6ca3285.gif)

merriemcgaw:
Thanks @Zheng-Li01 I think we will leave this marked as future. There isn't any real user impact as this is the empty datagrid view, and that's not a common scenario for customers.

Lydia-Shi:
Verified this issue with .NET 7.0.100-alpha.1.21566.13 from latest main branch, this issue is fixed that empty DGV can get blue Narrator rectangle.
[...]

PR #7052, opened 2022-04-19
Fix empty DataGridView HasKeyboardFocus property being always true

### LINKED ISSUE #7026: Focus in an empty DGV
@vladimir-krestov  - should this be `return owner.Focused && RowCount == 0;`

_Originally posted by @Tanya-Solyanik in https://github.com/dotnet/winforms/pull/6102#discussion_r851568124_

dmitrii-drobotov:
Property value was ignoring whether DGV was focused or not.
Now for empty DGV  `HasKeyboardFocusProperty` will have the same value as owner's Focused property. For non-empty DGV it will remain always false as focus will be acquired by inner cells.
Fixes #7026

## Proposed changes

- Change empty `DataGridView`'s `HasKeyboardFocus` property calculation to reflect DGV focused state.

## Customer Impact

- `HasKeyboardFocus` will have correct value for empty `DataGridView`.

## Regression? 

- No

## Risk

- Minimal

### Before

Without focus - `HasKeyboardFocus` is true:
before_unfocused_true

With focus (note the dotted line on DGV) - `HasKeyboardFocus` is true:
before_focused_true

### After

Without focus - `HasKeyboardFocus` is false:
after_notfocused_false

With focus (note the dotted line on DGV) - `HasKeyboardFocus` is true:
after_focused_true

## Test methodology 

- Manual (with Inspect tool)
- Unit
- CTI

## Accessibility testing  

Tested following cases with Inspect tool:
- Have an empty DataGridView without focus on it - `HasKeyboardFocus` should be false
- Set focus on empty DataGridView - `HasKeyboardFocus` should be true

## Test environment(s) 

- .NET 7.0
- Windows 10

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7052)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.AccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.AccessibleObject.cs
@@ -244,7 +244,7 @@ namespace System.Windows.Forms
                     case UiaCore.UIA.HasKeyboardFocusPropertyId:
                         // If no inner cell entire DGV should be announced as focused by Narrator.
                         // Else only inner cell should be announced as focused by Narrator but not entire DGV.
-                        return RowCount == 0;
+                        return _ownerDataGridView.Focused;
                     case UiaCore.UIA.IsKeyboardFocusablePropertyId:
                         return _ownerDataGridView.CanFocus;
                     case UiaCore.UIA.IsControlElementPropertyId:
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewAccessibleObjectTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewAccessibleObjectTests.cs
@@ -509,49 +509,87 @@ namespace System.Windows.Forms.Tests
             Assert.False(dataGridView.IsHandleCreated);
         }
 
-        [WinFormsFact]
-        public void DataGridViewAccessibleObject_GetPropertyValue_HasKeyboardFocus_IsExpected_ForEmptyDGV()
+        [WinFormsTheory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public void DataGridViewAccessibleObject_GetPropertyValue_HasKeyboardFocus_IsExpected_ForEmptyDGV(bool focused)
         {
-            using DataGridView dataGridView = new();
+            using DataGridView dataGridView = new FakeFocusDataGridView(focused);
 
-            Assert.True((bool)dataGridView.AccessibilityObject.GetPropertyValue(UiaCore.UIA.HasKeyboardFocusPropertyId));
+            bool actual = (bool)dataGridView.AccessibilityObject
+                .GetPropertyValue(UiaCore.UIA.HasKeyboardFocusPropertyId);
+
+            Assert.Equal(focused, dataGridView.Focused);
+            Assert.Equal(0, dataGridView.RowCount); // DGV is empty
+            Assert.Equal(focused, actual);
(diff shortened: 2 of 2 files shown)


src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewAccessibleObjectTests.cs
```diff
@@ -509,51 +509,89 @@ public void DataGridViewAccessibleObject_GetPropertyValue_ControlType_IsExpected
...
-
+            using DataGridView dataGridView = new FakeFocusDataGridView(focused);
             dataGridView.Columns.Add(new DataGridViewButtonColumn());
 
-            Assert.False((bool)dataGridView.AccessibilityObject.GetPropertyValue(UiaCore.UIA.HasKeyboardFocusPropertyId));
+            bool actual = (bool)dataGridView.AccessibilityObject
+                .GetPropertyValue(UiaCore.UIA.HasKeyboardFocusPropertyId);
+
+            Assert.Equal(focused, dataGridView.Focused);
+            Assert.Equal(1, dataGridView.RowCount); // One new row for editing
```
Tanya-Solyanik, line 540: I would add a comment about that single row stealing focus from the DGV.

Tanya-Solyanik (approved):
Look good!

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.AccessibleObject.cs
```diff
@@ -244,7 +244,7 @@ internal override int[] RuntimeId
                     case UiaCore.UIA.HasKeyboardFocusPropertyId:
                         // If no inner cell entire DGV should be announced as focused by Narrator.
                         // Else only inner cell should be announced as focused by Narrator but not entire DGV.
-                        return RowCount == 0;
+                        return _ownerDataGridView.Focused && RowCount == 0;
```
vladimir-krestov, line 247: It seems, it should be like this:
```suggestion
                        return _ownerDataGridView.Focused;
```
DGV should have `HasKeyboardFocus` when its cell is in focus.
So https://github.com/dotnet/winforms/pull/6102 way was incorrect.
But it should be carefully tested because we had cases when Narrator did an extra focus jump on the whole control.

dmitrii-drobotov: Done, thanks

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewAccessibleObjectTests.cs
dmitrii-drobotov: Done, thanks

dmitrii-drobotov:
Changed property value to return `_ownerDataGridView.Focused;`. For empty DGV behavior is the same as it was with `return _ownerDataGridView.Focused && RowCount == 0;`. For not empty it's following:
* When a cell inside a DGV is focused, DGV is also focused. Therefore `HasKeyboardFocus` is also true.
Screenshot 2022-04-21 135649
* When it's not focused,  `HasKeyboardFocus` is false as expected.
* There are no focus change issues in Narrator like in https://github.com/dotnet/winforms/issues/2178

Tanya-Solyanik (approved):
Looks good!

Tanya-Solyanik:
@Olina-Zhang - this change had been tested by your team for the 481 servicing release, so we'll skip the pre-checkin testing. Please test it during the regular .NET 7.0 testing.

PR #7059, opened 2022-04-20
Use ILockBytes using ComWrappers

kant2002:
That's first step to make RichTextBox to be operable in trimming mode.
ILockBytes used together with IStorage which used by IRichTextboxOleCallback.
Since IStorage deeply involved with ActiveX moving slowly and carefully.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7059)


src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.LockBytesComWrapper.cs
```diff
@@ -0,0 +1,39 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.InteropServices;
+using System.Windows.Forms;
+
+internal partial class Interop
+{
+    internal unsafe partial class WinFormsComWrappers
```
gpetrou, line 10: Perhaps such files should be in their own `ComWrappers` folder? Also, the file name could be Interop.ComWrappers.LockBytesComWrapper? I guess we could remove `WinForms` from the name.

gpetrou, line 10: Perhaps such files should be in their own `ComWrappers` folder? Also, the file name could be Interop.ComWrappers.LockBytesComWrapper? I guess we could remove `WinForms` from the name.

kant2002: 1. Separate folder is idea which I support. 
2. File name I honestly indifferent. But I notice that I should name it `LockBytesWrapper` to be the same with other classes.
3. Remove WinForms from the name probably not, since I suspect that some parts (like ActiveX) require different strategy for wrappers creation so we will have different class. Also we cannot name it `ComWrappers` to not confuse with [System.Runtime.InteropServices.ComWrappers](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.comwrappers)

src/System.Windows.Forms.Primitives/src/Interop/Ole32/Interop.CreateILockBytesOnHGlobal.cs
```diff
@@ -8,7 +8,18 @@ internal partial class Interop
...
-        public static extern ILockBytes CreateILockBytesOnHGlobal(IntPtr hGlobal, BOOL fDeleteOnRelease);
+        [DllImport(Libraries.Ole32, PreserveSig = false, EntryPoint = "CreateILockBytesOnHGlobal")]
+        private static extern IntPtr CreateILockBytesOnHGlobalRaw(IntPtr hGlobal, BOOL fDeleteOnRelease);
+
+        public static WinFormsComWrappers.LockBytesWrapper? CreateILockBytesOnHGlobal(IntPtr hGlobal, BOOL fDeleteOnRelease)
+        {
+            var ptr = CreateILockBytesOnHGlobalRaw(hGlobal, fDeleteOnRelease);
+            if (ptr == IntPtr.Zero)
+            {
+                return null;
```
RussKie, line 19: The only concern I have is that [`CreateILockBytesOnHGlobal`](https://docs.microsoft.com/windows/win32/api/coml2api/nf-coml2api-createilockbytesonhglobal) doesn't appear to return null, instead it's expected to set a non success HRESULT. This won't affect the end-users as the call is wrapped in try-catch, and we never surface the error, but makes the debug experience slightly more convoluted - we'd failed with a NRE in `CreateStorage()` implementation instead of failing here with a more targeted exception.

RussKie (commented):
Looks good.
How did you test this change?

src/System.Windows.Forms.Primitives/src/Interop/Ole32/Interop.CreateILockBytesOnHGlobal.cs
kant2002: I assume that `CreateILockBytesOnHGlobal` can create NULL only on low-memory conditions. We do not test for this scenario in any way? I can throw exception here, but what kind of exception is there.

I rely solely on test suite for this change. I expect either RichEdit Clipboard testing either present, or I add missing bits when I would add support for `IRichTextboxOleCallback`.

RussKie: If we rewrite the signature to use HRESULT we can use ThrowForHResult method, can we?

kant2002: Yes. we can. I rewrote, but we already have that behavior with built-in marshalling

kant2002: Was too quick to react without second thought. But I like explicitness of the approach. And no need to wait until `LibraryImportGenerator` became public.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/MultilineStringEditor.OleCallback.cs
```diff
@@ -27,13 +27,12 @@ public HRESULT GetNewStorage(out Ole32.IStorage storage)
             {
                 Debug.WriteLineIf(RichTextDebug.TraceVerbose, "IRichTextBoxOleCallback::GetNewStorage");
 
-                Ole32.ILockBytes pLockBytes = Ole32.CreateILockBytesOnHGlobal(IntPtr.Zero, BOOL.TRUE);
+                WinFormsComWrappers.LockBytesWrapper pLockBytes = Ole32.CreateILockBytesOnHGlobal(IntPtr.Zero, BOOL.TRUE);
                 Debug.Assert(pLockBytes is not null, "pLockBytes is NULL!");
```
RussKie, line 31: We can now delete this

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.State.cs
```diff
@@ -165,21 +163,20 @@ private void CreateStorage()
                 try
                 {
                     _iLockBytes = Ole32.CreateILockBytesOnHGlobal(hglobal, BOOL.TRUE);
+                    Debug.Assert(_iLockBytes is not null, "_iLockBytes is NULL!");
```
RussKie, line 166: Ditto

RussKie (commented):
👍

RussKie:
> I rely solely on test suite for this change. I expect either RichEdit Clipboard testing either present, or I add missing bits when I would add support for `IRichTextboxOleCallback`.

I don't believe we have any tests that invoke `GetNewStorage` method. I tried tinkering with the "RichTextBox" sample in the WinformsControlsTest project, but I found no way to hit it either.
With `richTextBox1.EnableAutoDragDrop = true;` I can d'n'd images and text from Windows Explorer and text editors (e.g., Word), as well as use the standard shortcuts. But none of these operations invoke the require method....

RussKie:
Actually, I stand corrected - I found a way to manually test it. If you copy-paste an OLE object (such as a graph from Excel) - I can hit the method:
image

I'm not sure we can automate this test.... ODBC?

kant2002:
Custom OLE object like I did for `IErrorInfo` ?

RussKie:
Maybe? :)

Once we have a confirmation that it's working (continues to work) - it's good to merge.


Thank you

kant2002:
Hmm. What kind of confirmation do you need?. Once you paste into clipboard, and let it run, it does not crash. Not kind of validation which I would like to have, but at least it's working.

I prefer create test separately if possible

RussKie:
> What kind of confirmation do you need?.

An ack that you've manually tested before and after, and confirmed that there were no obvious regressions (like we had with drag and drop).

> I prefer create test separately if possible

It's preferrable to add tests with the change. a) it makes the change complete, b) it makes it easier to review because you can see all that changed, and c) it helps avoiding rework, if there was a regression.

kant2002:
Okay, I create small test which test code path which I change, I cannot create integration test for now.

kant2002:
For more complex testing where I put actual OLE DataObject I plan to use approach from https://github.com/acaly/SharpOle . I found this library here: https://stackoverflow.com/questions/40901712/how-to-implement-ole-server-in-c-sharp

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RichTextBoxTests.cs
```diff
@@ -7933,6 +7933,20 @@ protected unsafe override void WndProc(ref Message m)
             }
         }
 
+        [WinFormsFact]
+        public void RichTextBox_OleObject_HandleJunkOleObject()
```
RussKie, line 7937: The original name was good - it provides information on what path is tested
```suggestion
        public void RichTextBox_RichEditOleCallback_CreateStorage()
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RichTextBoxTests.cs
```diff
@@ -7933,6 +7933,20 @@ protected unsafe override void WndProc(ref Message m)
...
+            using var control = new RichTextBox();
+            Assert.NotEqual(IntPtr.Zero, control.Handle);
+
+            using var memoryStream = new MemoryStream();
+            using var bitmap = new Bitmap(100, 100);
+            bitmap.Save(memoryStream, Drawing.Imaging.ImageFormat.Png);
+            Clipboard.SetData("Embed Source", memoryStream);
+
+            control.Paste();
+        }
```
RussKie, line 7948: Can we assert that the image is pasted? Something to indicate that the test has passed?

RussKie (commented):
👍

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RichTextBoxTests.cs
kant2002: I can revert back.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RichTextBoxTests.cs
kant2002: Not in this case otherwise I have to add a lot of other machinery which I not yet fully understand. That’s why I add links, because information is lost

kant2002: To be clear. Nothing pasted. Because I produce bad clipboard object

RussKie: Right, then we can add a comment explaining this and assert that the RTB is still empty, right? :)

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RichTextBoxTests.cs
RussKie: In general, we want to avoid words like "junk" in the code base.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RichTextBoxTests.cs
kant2002: Done.

RussKie (approved):

RussKie:
Thank you

PR #7065, opened 2022-04-21
Enable and fix CA1052 warnings in src files

gpetrou:
## Proposed changes

- Enable and fix CA1052 warnings in src files.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7065)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Layout/CommonProperties.DockAnchorMode.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Layout/CommonProperties.DockAnchorMode.cs
@@ -4,7 +4,7 @@
 
 namespace System.Windows.Forms.Layout
 {
-    internal partial class CommonProperties
+    internal static partial class CommonProperties
     {
         private enum DockAnchorMode
         {
--- a/eng/CodeAnalysis.src.globalconfig
+++ b/eng/CodeAnalysis.src.globalconfig
@@ -100,7 +100,7 @@ dotnet_diagnostic.CA1050.severity = warning
 dotnet_diagnostic.CA1051.severity = none
 
 # Static holder types should be Static or NotInheritable
-dotnet_diagnostic.CA1052.severity = none # TODO: warning
+dotnet_diagnostic.CA1052.severity = warning
 dotnet_code_quality.CA1052.api_surface = private, internal
 
 # URI-like parameters should not be strings
--- a/src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/Helpers/FileSystemUtils.vb
+++ b/src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/Helpers/FileSystemUtils.vb
@@ -13,7 +13,7 @@ Namespace Microsoft.VisualBasic.CompilerServices
     ''' <summary>
     '''  Internal utilities from Microsoft.VisualBasic.FileIO.FileSystem.
     ''' </summary>
-    Friend Class FileSystemUtils
+    Friend NotInheritable Class FileSystemUtils
 
         ''' <summary>
         ''' Normalize the path, but throw exception if the path ends with separator.
--- a/src/System.Windows.Forms.Primitives/src/Interop/Imm32/Interop.CPS.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/Imm32/Interop.CPS.cs
@@ -4,7 +4,7 @@
 
 internal partial class Interop
 {
-    internal partial class Imm32
+    internal static partial class Imm32
     {
         public enum CPS : uint
         {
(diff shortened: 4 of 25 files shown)


Tanya-Solyanik (approved):
Looks good

src/System.Windows.Forms/src/System/Windows/Forms/Layout/CommonProperties.DockAnchorMode.cs
```diff
@@ -4,7 +4,7 @@
 
 namespace System.Windows.Forms.Layout
 {
-    internal partial class CommonProperties
+    internal static partial class CommonProperties
```
dreddy-work, line 7: what about changing the main `CommonProperties` class to `static`?

gpetrou: I used the auto-fixer to address these warnings. I assume it applies the keyword to only one file which should be enough. Would you like me to add the keyword to the other files for consistency?

dreddy-work: Yes. that would make it complete.

dreddy-work: I didn't get that. If i see same class declared as non-static and then static partial in other. It is confusing.

eng/CodeStyle.targets
```diff
@@ -5,7 +5,7 @@
     <EditorConfigFiles Include="$(MSBuildThisFileDirectory)CodeAnalysis.src.globalconfig" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(IsShipping)' == 'false'">
+  <ItemGroup Condition="'$(IsTestProject)' == 'true' Or $(MSBuildProjectName.Contains('Test'))">
```
RussKie, line 8: I want to look into this when I'm back from a vacation next week.

src/System.Windows.Forms/src/System/Windows/Forms/Layout/CommonProperties.DockAnchorMode.cs
RussKie: The main file (i.e., the file with the shortest name, e.g.: MyCode.cs vs MyCode.Inner.cs) must have all declarations and modifiers - visibility, accessibility, etc. The compiler doesn't require to re-declare these on all partial files. And you can think of these re-declarations as a "maintenance overhead", this is one of the reasons we don't do it in our templates:
https://github.com/dotnet/winforms/blob/a9dbf1b4904a6b2c2fa8d106339f99cc629fd742/pkg/Microsoft.Dotnet.WinForms.ProjectTemplates/content/WinFormsApplication-CSharp/net6.0/Form1.Designer.cs#L4

eng/CodeStyle.targets
RussKie: I don't see any reasons to remove `IsShipping` check, it is set to `true` for all `IsTestProject`s:
https://github.com/dotnet/arcade/blob/5145e86df0c491e082b589aa31d69eea300adc02/src/Microsoft.DotNet.Arcade.Sdk/tools/Tests.props#L35-L37

I suggest to revert the change in this file, and do the following instead:
```diff
diff --git a/Directory.Build.targets b/Directory.Build.targets
index cdb7a8fbb..89440a8cc 100644
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -21,6 +21,13 @@
     <EnableSingleFileAnalyzer>true</EnableSingleFileAnalyzer>
   </PropertyGroup>
 
+  <!-- Mark all projects located under **\tests\* folders as non-shipping -->
+  <PropertyGroup>
+    <_ProjectDirectoryRelativePath>$([MSBuild]::MakeRelative($(RepoRoot), $(MSBuildProjectDirectory)))</_ProjectDirectoryRelativePath>
+    <_IsUnderTestsFolder>$(_ProjectDirectoryRelativePath.Contains("\tests\"))</_IsUnderTestsFolder>
+    <IsShipping Condition="$(_IsUnderTestsFolder)">false</IsShipping>
+  </PropertyGroup>
+
   <PropertyGroup Condition="'$(IsTestProject)' == 'true' ">
     <TargetArchitecture Condition="'$(TargetArchitecture)' == ''">x64</TargetArchitecture>
     <Platform Condition="'$(Platform)' == 'AnyCPU' or '$(Platform)' == 'Any CPU'">$(TargetArchitecture)</Platform>
```

RussKie (approved):

RussKie:
Thank you

PR #7087, opened 2022-04-26
Use ComWrappers for Clipboard

kant2002:
Create missing RCW and CCW
Testing happens using existing test suite which is quite good.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7087)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/Interop/Ole32/Interop.OleSetClipboard.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/Ole32/Interop.OleSetClipboard.cs
@@ -10,6 +10,16 @@ internal static partial class Interop
     internal static partial class Ole32
     {
         [DllImport(Libraries.Ole32, ExactSpelling = true)]
-        public static extern HRESULT OleSetClipboard(IDataObject? pDataObj);
+        private static extern HRESULT OleSetClipboard(IntPtr pDataObj);
+
+        public static HRESULT OleSetClipboard(IDataObject? pDataObj)
+        {
+            if (pDataObj == null)
+            {
+                return OleSetClipboard(IntPtr.Zero);
+            }
+
+            return OleSetClipboard(WinFormsComWrappers.Instance.GetComPointer(pDataObj, IID.IDataObject));
+        }
     }
 }
--- a/src/System.Windows.Forms.Primitives/src/Interop/Ole32/Interop.OleSetClipboard.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/Ole32/Interop.OleSetClipboard.cs
@@ -14,7 +14,7 @@ internal static partial class Interop
 
         public static HRESULT OleSetClipboard(IDataObject? pDataObj)
         {
-            if (pDataObj == null)
+            if (pDataObj is null)
             {
                 return OleSetClipboard(IntPtr.Zero);
             }
--- a/src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.AgileDataObjectWrapper.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.AgileDataObjectWrapper.cs
@@ -1,0 +1,82 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.InteropServices;
+using System.Runtime.InteropServices.ComTypes;
+
+internal partial class Interop
+{
+    internal unsafe partial class WinFormsComWrappers
+    {
+        internal class AgileDataObjectWrapper : IDataObject
+        {
+            private AgileReferenceWrapper _wrappedInstance;
+
+            public AgileDataObjectWrapper(AgileReferenceWrapper wrappedInstance)
+            {
+                _wrappedInstance = wrappedInstance;
+            }
+
+            private DataObjectWrapper Unwrap()
+            {
+                var instance = _wrappedInstance.Resolve(IID.IDataObject);
+                return new DataObjectWrapper(instance);
+            }
+
+            public void GetData(ref FORMATETC format, out STGMEDIUM medium)
+            {
+                using var dataObject = Unwrap();
+                dataObject.GetData(ref format, out medium);
+            }
+
+            public void GetDataHere(ref FORMATETC format, ref STGMEDIUM medium)
+            {
+                using var dataObject = Unwrap();
+                dataObject.GetDataHere(ref format, ref medium);
+            }
+
+            public int QueryGetData(ref FORMATETC format)
+            {
+                using var dataObject = Unwrap();
+                return dataObject.QueryGetData(ref format);
+            }
+
+            public int GetCanonicalFormatEtc(ref FORMATETC formatIn, out FORMATETC formatOut)
+            {
+                using var dataObject = Unwrap();
+                return dataObject.GetCanonicalFormatEtc(ref formatIn, out formatOut);
+            }
+
+            public void SetData(ref FORMATETC formatIn, ref STGMEDIUM medium, bool release)
+            {
+                using var dataObject = Unwrap();
+                dataObject.SetData(ref formatIn, ref medium, release);
+            }
+
+            public IEnumFORMATETC EnumFormatEtc(DATADIR direction)
+            {
+                using var dataObject = Unwrap();
+                return dataObject.EnumFormatEtc(direction);
+            }
+
+            public int DAdvise(ref FORMATETC pFormatetc, ADVF advf, IAdviseSink adviseSink, out int connection)
+            {
+                using var dataObject = Unwrap();
+                return dataObject.DAdvise(ref pFormatetc, advf, adviseSink, out connection);
+            }
+
+            public void DUnadvise(int connection)
+            {
+                using var dataObject = Unwrap();
+                dataObject.DUnadvise(connection);
+            }
+
+            public int EnumDAdvise(out IEnumSTATDATA? enumAdvise)
+            {
+                using var dataObject = Unwrap();
+                return dataObject.EnumDAdvise(out enumAdvise);
+            }
+        }
+    }
+}
--- a/src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.AgileDataObjectWrapper.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.AgileDataObjectWrapper.cs
@@ -2,7 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System.Runtime.InteropServices;
 using System.Runtime.InteropServices.ComTypes;
 
 internal partial class Interop
--- a/src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.DataObjectWrapper.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.DataObjectWrapper.cs
@@ -64,7 +64,7 @@ internal partial class Interop
                 {
                     STGMEDIUM_Raw mediumRaw = new()
                     {
-                        pUnkForRelease = medium.pUnkForRelease == null ? IntPtr.Zero : Marshal.GetIUnknownForObject(medium.pUnkForRelease),
+                        pUnkForRelease = medium.pUnkForRelease is null ? IntPtr.Zero : Marshal.GetIUnknownForObject(medium.pUnkForRelease),
                         tymed = medium.tymed,
                         unionmember = medium.unionmember,
                     };
@@ -108,7 +108,7 @@ internal partial class Interop
                 {
                     STGMEDIUM_Raw mediumRaw = new()
                     {
-                        pUnkForRelease = medium.pUnkForRelease == null ? IntPtr.Zero : Marshal.GetIUnknownForObject(medium.pUnkForRelease),
+                        pUnkForRelease = medium.pUnkForRelease is null ? IntPtr.Zero : Marshal.GetIUnknownForObject(medium.pUnkForRelease),
                         tymed = medium.tymed,
                         unionmember = medium.unionmember,
                     };
--- a/src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.DataObjectWrapper.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.DataObjectWrapper.cs
@@ -10,7 +10,7 @@ internal partial class Interop
 {
     internal unsafe partial class WinFormsComWrappers
     {
-        internal class DataObjectWrapper : IDataObject
+        internal sealed class DataObjectWrapper : IDataObject, IDisposable
         {
             private IntPtr _wrappedInstance;
 
--- a/src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.DataObjectWrapper.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.DataObjectWrapper.cs
@@ -70,6 +70,8 @@ internal partial class Interop
                     };
                     ((delegate* unmanaged<IntPtr, FORMATETC*, STGMEDIUM_Raw*, HRESULT>)(*(*(void***)_wrappedInstance + 4)))
                         (_wrappedInstance, formatPtr, &mediumRaw).ThrowIfFailed();
+                    // Because we do not know if COM interface implementation change value of mediumRaw.pUnkForRelease during the call,
+                    // unmarshal it again.
                     medium = new()
                     {
                         pUnkForRelease = mediumRaw.pUnkForRelease == IntPtr.Zero ? null : Marshal.GetObjectForIUnknown(mediumRaw.pUnkForRelease),
--- a/src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.DataObjectWrapper.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.DataObjectWrapper.cs
@@ -1,0 +1,142 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.InteropServices;
+using System.Runtime.InteropServices.ComTypes;
+using System.Windows.Forms;
+
+internal partial class Interop
+{
+    internal unsafe partial class WinFormsComWrappers
+    {
+        internal class DataObjectWrapper : IDataObject
+        {
+            private IntPtr _wrappedInstance;
+
+            public DataObjectWrapper(IntPtr wrappedInstance)
+            {
+                _wrappedInstance = wrappedInstance.OrThrowIfZero();
+            }
+
+            internal IntPtr Instance => _wrappedInstance;
+
+            public void Dispose()
+            {
+                Marshal.Release(_wrappedInstance);
+                _wrappedInstance = IntPtr.Zero;
+            }
+
+            public void GetData(ref FORMATETC format, out STGMEDIUM medium)
+            {
+                fixed (FORMATETC* formatPtr = &format)
+                {
+                    STGMEDIUM_Raw mediumRaw;
+                    ((delegate* unmanaged<IntPtr, FORMATETC*, STGMEDIUM_Raw*, HRESULT>)(*(*(void***)_wrappedInstance + 3)))
+                        (_wrappedInstance, formatPtr, &mediumRaw).ThrowIfFailed();
+                    medium = new()
+                    {
+                        pUnkForRelease = mediumRaw.pUnkForRelease == IntPtr.Zero ? null : Marshal.GetObjectForIUnknown(mediumRaw.pUnkForRelease),
+                        tymed = mediumRaw.tymed,
+                        unionmember = mediumRaw.unionmember,
+                    };
+                }
+            }
+
+            public void GetDataHere(ref FORMATETC format, ref STGMEDIUM medium)
+            {
+                fixed (FORMATETC* formatPtr = &format)
+                {
+                    STGMEDIUM_Raw mediumRaw = new()
+                    {
+                        pUnkForRelease = medium.pUnkForRelease == null ? IntPtr.Zero : Marshal.GetIUnknownForObject(medium.pUnkForRelease),
+                        tymed = medium.tymed,
+                        unionmember = medium.unionmember,
+                    };
+                    ((delegate* unmanaged<IntPtr, FORMATETC*, STGMEDIUM_Raw*, HRESULT>)(*(*(void***)_wrappedInstance + 4)))
+                        (_wrappedInstance, formatPtr, &mediumRaw).ThrowIfFailed();
+                    medium = new()
+                    {
+                        pUnkForRelease = mediumRaw.pUnkForRelease == IntPtr.Zero ? null : Marshal.GetObjectForIUnknown(mediumRaw.pUnkForRelease),
+                        tymed = mediumRaw.tymed,
+                        unionmember = mediumRaw.unionmember,
+                    };
+                }
+            }
+
+            public int QueryGetData(ref FORMATETC format)
+            {
+                fixed (FORMATETC* formatPtr = &format)
+                {
+                    return (int)((delegate* unmanaged<IntPtr, FORMATETC*, HRESULT>)(*(*(void***)_wrappedInstance + 5)))
+                        (_wrappedInstance, formatPtr);
+                }
+            }
+
+            public int GetCanonicalFormatEtc(ref FORMATETC formatIn, out FORMATETC formatOut)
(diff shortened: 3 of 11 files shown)


kant2002:
@RussKie can I ask you start looking into this masterpiece.

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.DataObjectWrapper.cs
```diff
@@ -0,0 +1,142 @@
...
+            public void GetData(ref FORMATETC format, out STGMEDIUM medium)
+            {
+                fixed (FORMATETC* formatPtr = &format)
+                {
+                    STGMEDIUM_Raw mediumRaw;
+                    ((delegate* unmanaged<IntPtr, FORMATETC*, STGMEDIUM_Raw*, HRESULT>)(*(*(void***)_wrappedInstance + 3)))
+                        (_wrappedInstance, formatPtr, &mediumRaw).ThrowIfFailed();
+                    medium = new()
+                    {
+                        pUnkForRelease = mediumRaw.pUnkForRelease == IntPtr.Zero ? null : Marshal.GetObjectForIUnknown(mediumRaw.pUnkForRelease),
```
RussKie, line 39: Do we need to release `mediumRaw.pUnkForRelease`?

RussKie:
Looks good. I've run some local manual tests too.

@AaronRobinsonMSFT @JeremyKuhne I'd appreciate it if you could review as well.

src/System.Windows.Forms/src/System/Windows/Forms/Clipboard.cs
```diff
@@ -155,15 +155,10 @@ public static void SetDataObject(object data, bool copy, int retryTimes, int ret
...
-            while (hr != 0);
+            while (!success);
 
             if (dataObject is not null)
             {
-                if (dataObject is IDataObject ido && !Marshal.IsComObject(dataObject))
-                {
-                    return ido;
-                }
-
```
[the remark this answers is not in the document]
kant2002: This is effectively dead code after `ComVisible` removal. My current plan is provide for `DataObject` CCW's which supports 2 interfaces `IComDataObject` and `IWinFormsDataObject` which resurrect code in some form.

`dataObject` is always `DataObjectWrapper` here which implements only `IComDataObject` for now. And previously we have `_ComObject` which react just to `IComDataObject`, that why @filipnavara create issue in first place.

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.EnumFORMATETCWrapper.cs
```diff
@@ -0,0 +1,71 @@
...
+            private IntPtr _wrappedInstance;
+
+            public EnumFORMATETCWrapper(IntPtr wrappedInstance)
+            {
+                _wrappedInstance = wrappedInstance.OrThrowIfZero();
+            }
+
+            internal IntPtr Instance => _wrappedInstance;
+
+            public void Dispose()
```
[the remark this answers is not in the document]
kant2002: I add finalizer after this comment. Also I leave this method for consistency with other wrappers.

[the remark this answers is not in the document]
AaronRobinsonMSFT: I don't think the `Dispose` pattern shouldn't be used in this case. Consistency isn't the priority here, but rather correctness. See the note under https://docs.microsoft.com/dotnet/standard/native-interop/tutorial-comwrappers#step-4--handle-native-object-wrapper-lifetime-details.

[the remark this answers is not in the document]
kant2002: I can attribute my ignorance to the fact that there now documentation when I start 😉 , but thank for keeping me responsible.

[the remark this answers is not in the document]
kant2002: And crap. I add finalizer in different place. In the DataObject....

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.DataObjectWrapper.cs
```diff
@@ -0,0 +1,159 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.InteropServices;
+using System.Runtime.InteropServices.ComTypes;
+using System.Windows.Forms;
+
+internal partial class Interop
+{
+    internal unsafe partial class WinFormsComWrappers
```
AaronRobinsonMSFT, line 11: I don't see anything here around apartment marshalling. If I recall, the entire clipboard stack is required to be on the STA. This doesn't enforce that or even bother to check, which means we can get into data corruption or random crashes. This type only ever going to be used on the STA and if so, how is that is that enforced?

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.IEnumFORMATETCVtbl.cs
```diff
@@ -0,0 +1,90 @@
...
+                }
+
+                return result;
+            }
+
+            [UnmanagedCallersOnly]
+            private static int Skip(IntPtr thisPtr, int celt)
+            {
+                IEnumFORMATETC instance = ComInterfaceDispatch.GetInstance<IEnumFORMATETC>((ComInterfaceDispatch*)thisPtr);
+                return instance.Skip(celt);
```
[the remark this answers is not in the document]
kant2002: That's done.

src/System.Windows.Forms.Primitives/src/Interop/Ole32/Interop.OleSetClipboard.cs
```diff
@@ -10,6 +10,16 @@ internal static partial class Interop
     internal static partial class Ole32
     {
         [DllImport(Libraries.Ole32, ExactSpelling = true)]
-        public static extern HRESULT OleSetClipboard(IDataObject? pDataObj);
+        private static extern HRESULT OleSetClipboard(IntPtr pDataObj);
```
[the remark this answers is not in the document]
RussKie: As in like this?
```suggestion
        public static HRESULT OleSetClipboard(IDataObject? pDataObj)
        {
            if (pDataObj is null)
            {
                return OleSetClipboard(IntPtr.Zero);
            }

            return OleSetClipboard(WinFormsComWrappers.Instance.GetComPointer(pDataObj, IID.IDataObject));

            [DllImport(Libraries.Ole32, ExactSpelling = true)]
            static extern HRESULT OleSetClipboard(IntPtr pDataObj);
        }
```

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.DataObjectWrapper.cs
```diff
@@ -0,0 +1,159 @@
...
+                    {
+                        pUnkForRelease = medium.pUnkForRelease is null ? IntPtr.Zero : Marshal.GetIUnknownForObject(medium.pUnkForRelease),
+                        tymed = medium.tymed,
+                        unionmember = medium.unionmember,
+                    };
+                    ((delegate* unmanaged<IntPtr, FORMATETC*, STGMEDIUM_Raw*, HRESULT>)(*(*(void***)_wrappedInstance + 4)))
+                        (_wrappedInstance, formatPtr, &mediumRaw).ThrowIfFailed();
+                    medium = new()
+                    {
+                        pUnkForRelease = mediumRaw.pUnkForRelease == IntPtr.Zero ? null : Marshal.GetObjectForIUnknown(mediumRaw.pUnkForRelease),
```
RussKie, line 69: Could you please help me understand why we need to marshal again, and can't just copy `mediumRaw.pUnkForRelease` here?

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.DataObjectWrapper.cs
RussKie: In the clipboard we check:
https://github.com/dotnet/winforms/blob/93edc2d840887225d23eed00d348c5b24b4781c0/src/System.Windows.Forms/src/System/Windows/Forms/Clipboard.cs#L40-L43

Perhaps we should be doing the same here as well.

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.DataObjectWrapper.cs
```diff
@@ -0,0 +1,159 @@
...
+                ((delegate* unmanaged<IntPtr, int, HRESULT>)(*(*(void***)_wrappedInstance + 10)))
+                    (_wrappedInstance, connection).ThrowIfFailed();
+            }
+
+            public int EnumDAdvise(out IEnumSTATDATA? enumAdvise)
+            {
+                IntPtr enumAdvisePtr;
+                var result = ((delegate* unmanaged<IntPtr, IntPtr*, HRESULT>)(*(*(void***)_wrappedInstance + 11)))
+                    (_wrappedInstance, &enumAdvisePtr);
+                enumAdvise = result.Succeeded() ? null : (IEnumSTATDATA)Marshal.GetObjectForIUnknown(enumAdvisePtr);
```
RussKie, line 154: @AaronRobinsonMSFT @kant2002  could you please help me understand the best practices for `Marshal.GetObjectForIUnknown`? According to the [docs](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshal.getobjectforiunknown?view=net-6.0#remarks) it increments the ref count. Does it mean we need to release `enumAdvisePtr` before we exit?

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.EnumFORMATETCWrapper.cs
```diff
@@ -0,0 +1,76 @@
...
+            ~EnumFORMATETCWrapper()
+            {
+                this.DisposeInternal();
+            }
+
+            private void DisposeInternal()
+            {
+                Marshal.Release(_wrappedInstance);
+                _wrappedInstance = IntPtr.Zero;
+            }
```
RussKie, line 33: Any reason we can't inline these together?

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.IEnumFORMATETCVtbl.cs
```diff
@@ -0,0 +1,97 @@
...
+                {
+                    IEnumFORMATETC instance = ComInterfaceDispatch.GetInstance<IEnumFORMATETC>((ComInterfaceDispatch*)thisPtr);
+                    instance.Reset();
+                }
+                catch (Exception ex)
+                {
+                    return ex.HResult;
+                }
+
+                return S_OK;
```
RussKie, line 76: This makes it easier to follow the flow:
```suggestion
                try
                {
                    IEnumFORMATETC instance = ComInterfaceDispatch.GetInstance<IEnumFORMATETC>((ComInterfaceDispatch*)thisPtr);
                    instance.Reset();
                    return S_OK;
                }
                catch (Exception ex)
                {
                    return ex.HResult;
                }
```

ditto other places

RussKie (commented):
Few questions.

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.DataObjectWrapper.cs
kant2002: I'm all for multilayered checks, but I wonder, how .NET already handling that? Also My understnading was that I should use `IAgileReference` like it is here https://github.com/RussKie/ClipboardRedux

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.DataObjectWrapper.cs
```diff
@@ -0,0 +1,159 @@
...
+                }
+            }
+
+            public void GetDataHere(ref FORMATETC format, ref STGMEDIUM medium)
+            {
+                fixed (FORMATETC* formatPtr = &format)
+                {
+                    STGMEDIUM_Raw mediumRaw = new()
+                    {
+                        pUnkForRelease = medium.pUnkForRelease is null ? IntPtr.Zero : Marshal.GetIUnknownForObject(medium.pUnkForRelease),
```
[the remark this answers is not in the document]
kant2002: Then what we pass to unmanged side.

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.DataObjectWrapper.cs
kant2002: what if somebody replace value, stored in the `mediumRaw.pUnkForRelease`. We do not know if somebody change value of `pUnkForRelease`. That's highly unlikely, but who knows.

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.DataObjectWrapper.cs
kant2002: Release is responsibility of RCW finalizer.

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.EnumFORMATETCWrapper.cs
kant2002: Technically we can. That was needed because I thought that having Dispose method is valuable, but right now there no reason.

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.DataObjectWrapper.cs
AaronRobinsonMSFT: All .NET objects are free-threaded so COM servers implemented in .NET can run on any apartment. For built-in RCWs, this is incredibly complex and handled in the RCW implementation - see [`UnmarshalIUnknownForCurrContext`](https://github.com/dotnet/runtime/blob/41419131095d36fb5b811600ad0dab3b0d804269/src/coreclr/vm/comcache.cpp#L609-L611) for an example of the marshalling across apartments.

AaronRobinsonMSFT: Yes, `IAgileReference` is the best way to do this on Windows 8+. However, it doesn't work on Windows 7, which means we need to have a fallback if Winforms is still required to support that OS. On pre-Windows 7 using the [Global Interface Table](https://docs.microsoft.com/windows/win32/com/when-to-use-the-global-interface-table) (GIT) is required.

I believe this is mentioned in the ClipboardRedux example above.

src/System.Windows.Forms.Primitives/src/Interop/Ole32/Interop.OleSetClipboard.cs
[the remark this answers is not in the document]
kant2002: I think that's what Aaron means. Do you want me apply this change?

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.DataObjectWrapper.cs
kant2002: Win 7 SP1 ESU from what I see will be supported till [Jan 20 2023](https://docs.microsoft.com/en-us/troubleshoot/windows-client/windows-7-eos-faq/windows-7-extended-security-updates-faq). Release of .NET 7 I assume happens somewhere around Nov 2022, which leave 3 month of support for Win7. 

I understand that according to .NET support policy Win 7 should be supported. But as lazy person, I would like to ask WinForms team, or .NET team, is this necessary?

AaronRobinsonMSFT: > I understand that according to .NET support policy Win 7 should be supported. But as lazy person, I would like to ask WinForms team, or .NET team, is this necessary?

I would say, yes. Microsoft support for a product doesn't mean other products don't support running on that platform. This is annoying for sure, but how enterprise software works unfortunately. I do not know the current plan, but as long as the platform is supported at release, it is likely it will be a supported target.

@richlander Would know the support perspective for sure.

src/System.Windows.Forms.Primitives/src/Interop/Ole32/Interop.OleSetClipboard.cs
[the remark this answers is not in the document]
kant2002: How will it play if we switch to `LibraryImport` later on? Will source gen produce code for this?

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.DataObjectWrapper.cs
kant2002: Would it be reasonable for me target only `IAgileReference` and create issue for tracking support for Win7 ESU? Based on answer I will either add support for Win7 separately, or would not. No answer I will consider as "Please implement Win7 ESU". 

In my defense, I would tell - why I even consider that you may degrade Win7 ESU support for .NET 7 is because .NET 7 is non LTS version and combination of extra conservative platform (Win7 ESU) + fast cadence platform (.NET 7) seems wierd for me.

AaronRobinsonMSFT: > Would it be reasonable for me target only IAgileReference and create issue for tracking support for Win7 ESU?

Absolutely. Regardless whether Win 7 is supported or not, the "fast path" using `IAgileReference` should be separate and not cluttered by Win 7 support, which could be slower if we determine it is required.

AaronRobinsonMSFT: I am still working to clarify if Win 7 support is needed. I should have an official answer closer to the end of the month.

richlander: Independent of the answer on Windows 7 support, can we just implement new Windows Forms features the new best way and say that we'll document which features are not supported on Windows 7 going forward if that proves necessary?

Put another way, if we decide to support Windows 7 again -- which I _really_ hope not -- it's not really intended as a first-class citizen.

kant2002: @AaronRobinsonMSFT and @RussKie can we consider current implementation as "limited support" for Win7? I'm not sure that `Clipboard` can be considered `new feature`. Or maybe for Win7 I can revert back to built-in Com Interop, but how all this playout with LibraryImport work. Just throwing ideas.

RussKie: My take on this is that we add OS checks to use the new functionality on Win8+ and use the original implementations for Win7.

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.AgileDataObjectWrapper.cs
```diff
@@ -0,0 +1,81 @@
...
+
+            public void GetDataHere(ref FORMATETC format, ref STGMEDIUM medium)
+            {
+                using var dataObject = Unwrap();
+                dataObject.GetDataHere(ref format, ref medium);
+            }
+
+            public int QueryGetData(ref FORMATETC format)
+            {
+                using var dataObject = Unwrap();
```
AaronRobinsonMSFT, line 40: If this is how apartment support for `IDataObject` is being provided, then I'm not convinced we should permit it to be `IDisposable`. That pattern has non-trivial performance impact and is likely to be noticeable in large applications. We should instead defer to the built-in lifetime management provided by the runtime.

src/System.Windows.Forms.Primitives/src/Interop/Ole32/Interop.OleSetClipboard.cs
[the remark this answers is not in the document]
RussKie: I guess we'll cross that bridge when we get to it :)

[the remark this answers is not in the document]
RussKie: I guess we'll cross that bridge when we get to it :)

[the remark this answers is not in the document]
kant2002: Isn’t #7172 practically done. And it will land definitely faster then this one. Am I miss something?

src/System.Windows.Forms/src/System/Windows/Forms/Clipboard.cs
```diff
@@ -138,13 +138,13 @@ public static void SetDataObject(object data, bool copy, int retryTimes, int ret
         /// </remarks>
         private static IDataObject? GetDataObject(int retryTimes, int retryDelay)
         {
-            IComDataObject? dataObject = null;
```
RussKie, line 141: Since W7 support is still in scope, we need to retain the original implementation, and conditionally invoke the new implementation. E.g., using `OsVersion.IsWindows8OrGreater` or doing something like:
```cs
        private static IDataObject? GetDataObject(int retryTimes, int retryDelay)
        {
            // new code
        }

        private static IDataObject? GetDataObjectWin7(int retryTimes, int retryDelay)
        {
            // original code
        }
```

src/System.Windows.Forms.Primitives/src/Interop/Ole32/Interop.OleSetClipboard.cs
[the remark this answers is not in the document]
RussKie: Judging by the @JeremyKuhne's [comments](https://github.com/dotnet/winforms/pull/7172#discussion_r880829193) there are perf implications that may need to be addressed first.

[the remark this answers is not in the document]
kant2002: My understanding on this, that this comment applied only if custom marshaller would be used with `HandleRef` or `IHandle`, but with `IntPtr` these concerns are not applicable. So that PR looks like it can be merged as is right now.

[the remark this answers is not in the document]
JeremyKuhne: > How will it play if we switch to LibraryImport later on? Will source gen produce code for this?

If we're just switching to `LibraryImport` it won't necessitate changing anything here.

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.DataObjectWrapper.cs
JeremyKuhne: Since there was a question here consider a clarifying comment

src/System.Windows.Forms/src/System/Windows/Forms/Clipboard.cs
kant2002: Thanks! does agility an issue for `OleSetClipboard` in same way as for `OleGetClipboard`?

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.DataObjectWrapper.cs
RussKie: FYI [.NET support for Windows 7 and 8.1 will end in January 2023](https://github.com/dotnet/announcements/issues/226)

kant2002: Seems to be Did not matter. I have to somehow drag to that point and I need old implementation hanging around. 

Still looking for approximate list of test cases for clipboard.

JeremyKuhne:
@kant2002 unfortunately resource constraints have pushed this out of .NET 8. We hope to tackle the rest of the COM interop for .NET 9.

I appreciate all of the effort here and if you're interested, I'll keep you looped in. As the COM code has changed significantly, I think it's better to close this one and use it as a reference in a new PR.

Note: One particular difficulty here is that we need to make sure we've fully understood and documented changes in castability. Without built-in COM interop you don't get the "magic" casting of pointers back to their original objects or out to arbitrary implemented `[ComImport]` interfaces on the types. I haven't had the time to deep-dive into mapping this out, but I suspect we may have to have `AppContext` switches in some places. I'll capture this in more detail as I get our tracking issues into shape.

kant2002:
@JeremyKuhne I agree with you on this one. And since I did not push this in time we should rethink how it can be done in .NET 9. I still interested in this area, but cannot allocate cycles for pushing this forward.

PR #7108, opened 2022-04-30
Enable nullability in BindingContext and HashKey

gpetrou:
## Proposed changes

- Enable nullability in BindingContext and HashKey.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7108)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/BindingContext.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/BindingContext.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.Collections;
 using System.ComponentModel;
 using System.Diagnostics;
@@ -83,7 +81,7 @@ namespace System.Windows.Forms
         ///  Gets the System.Windows.Forms.BindingManagerBase associated with the specified
         ///  data source and data member.
         /// </summary>
-        public BindingManagerBase this[object dataSource, string dataMember]
+        public BindingManagerBase this[object dataSource, string? dataMember]
         {
             get => EnsureListManager(dataSource, dataMember);
         }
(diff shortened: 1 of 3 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/BindingContext.cs
```diff
@@ -47,7 +44,7 @@ int ICollection.Count
         void ICollection.CopyTo(Array ar, int index)
         {
             ScrubWeakRefs();
-            _listManagers.CopyTo(ar, index);
+            new Hashtable(_listManagers).CopyTo(ar, index);
```
gpetrou, line 47: I am not sure about this change. If we use `((ICollection)_listManagers).CopyTo(` instead, some tests such as `BindingContext_CopyTo_Invoke_Success` will fail because the type has changed from DictionaryEntry to KeyValuePair. If modifying the tests is acceptable, then the next issue is that `HashKey` is private, so we cannot just use that in the assertions. Assuming that we don't just want to make `HashKey` internal, there might be an alternative approach to achieve what the current assertions do?

RussKie: In this case updating the test is totally justified

src/System.Windows.Forms/src/System/Windows/Forms/BindingContext.cs
```diff
@@ -332,15 +332,15 @@ private static void CheckPropertyBindingCycles(BindingContext newBindingContext,
 
         private void ScrubWeakRefs()
         {
-            ArrayList cleanupList = null;
-            foreach (DictionaryEntry de in _listManagers)
+            List<HashKey> cleanupList = null;
+            foreach (KeyValuePair<HashKey, WeakReference> de in _listManagers)
```
RussKie, line 333: Can we do deconstruction here? E.g., something like:
```suggestion
            foreach ((HashKey key, WeakReference wRef) in _listManagers)
```

elachlan:
Related #8143

Great work with this! I'd love to see it merged.

gpetrou:
@dreddy-work could we also merge this one? https://github.com/dotnet/winforms/pull/8235 should take care of the ArrayList/Hashtable replacements.

dreddy-work (approved):
LGTM.

PR #7134, opened 2022-05-04
rename file DockAndAnchorLayout.cs to DefaultLayout.cs

tbolon:
This is a simple change to rename DockAndAnchorLayout.cs to DefaultLayout.cs

I just spend 15 minutes looking for the definition of DefaultLayout class, and had to clone the repo to find it.
I think the class has been renamed long ago, but the file was not renamed.

The only reference left is an internal method named [GetNeedsDockAndAnchorLayout](https://github.com/dotnet/winforms/blob/main/src/System.Windows.Forms/src/System/Windows/Forms/Layout/CommonProperties.cs#L416) in `CommonProperties` which could also be renamed.

I did not create an issue since no template seems to really match this PR.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7134)


RussKie (approved):

RussKie:
Thank you

PR #7143, opened 2022-05-05
Introduce Command Binding, Bindable ToolStripItems and DataContext to WinForms

### LINKED ISSUE #4895: [API Proposal]: Introduce Commands and DataContext to WinForms to adapt modern Binding scenarios
## Rationale:

We want our customers to modernize their existing WinForms Apps and allow them to adapt new Azure-based technologies. It is important for our customers, that the business logic of their WinForms LOB apps to this end becomes more flexible and can be rearchitected away from the WinForms-typical code-behind and towards UI Controller or ViewModels, which can be easy reused and unit tested. This is also an incentive to pivot away from the 15 years old Typed DataSets in WinForms, which have very limited support in .NET WinForms Apps over .NET Framework, since the necessary Data Source Provider Service is not available for .NET Apps altogether. Feedback for suggesting this as our [new Binding strategy](https://devblogs.microsoft.com/dotnet/databinding-with-the-oop-windows-forms-designer/#new-approaches-when-using-object-data-source-binding) was highly engaged, very positive, and shows that we're on the right way with that.

WinForms has a pretty powerful binding engine, but to achieve those things, a few essential features are missing:

* ToolStripItems (the items that make pull-down menus, status bar labels and other status bar controls) are _components_ rather than controls, and components deriving from ToolStripItem are currently not bindable.
* Buttons and ToolstripItems don't provide properties to store or bind to Commands based on `System.Windows.Input.ICommand`.
* There is no central hooking point on a Form or a UserControl to distribute a data source for data binding purposes from the parent down to every child it's hosting, like the `DataContext` property does that in WPF (or `BindingContext` in Maui Apps).

Having these features in WinForms, enables a series of important scenarios:

* **Using ViewModels/Controllers in WinForms:** Customer’s LOB logic should be leverage to other UI technologies (Maui, UWP, WPF) and Azure Services, which is not possible or at least only quite difficult to achieve with the typical existing WinForms code-behind methodology.

MauiWinForms

* **Easier adaptation of modern ORMs:** Since for the .NET Designer, the Data Source Provider Service is not available, with this new feature, it becomes easier to adapt and migrate away from ancient WinForms data layer architectures: While typed `DataSet` and `DataTable` were clearly designed with the WinForms code-behind model and direct binding to UI Elements of WinForms in mind, with the new data binding features present, modernized WinForms Apps could more easily consume typical ViewModels and Models as their Controller and Data Layer abstractions, and, what's important: Models in this scenarios being POCO classes which are generated and maintained by Entity Framework or Entity Framework Core.

* **Making it possible to unit test WinForms Apps business logic:** Especially with the introduction of commands and command parameters, it would be way easier to use ViewModels/UI-Controller in binding scenarios for WinForms, that can be reused in Maui scenarios. While WinForms will not be able to achieve the same flexibility as compared to XAML based data binding, binding commands is a considerable step in that direction, Customers have repeatedly asked for that, and making ToolStripItems bindable at the same time, would be equally helpful in easier adapting ViewModels/UI Controller Models.

* It's almost impossible to unit test Code-behind apps, which mixes business logic and UI-Control code. But that's how WinForms Apps are typically "architected". The new API makes it way simple, to separated the Business Logic from the actual Code-Behind code, so it can be properly unit and integration tested.

### Github Repo for Sample-App

https://github.com/KlausLoeffelmann/WinFormsCommandBinding

## API-Changes, Overview:

Note: For .NET 7, we want to put this feature out in Preview, attributing respective new classes and properties with the [`RequiresPreviewFeature` attribute](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.versioning.requirespreviewfeaturesattribute?view=net-6.0), collect feedback and then enable it for .NET 8.

To provide a more ViewModel/View Controller compatible way for a WinForms LOB architecture, I propose the following API-Changes:

* Introduce the abstract classes `BindableComponent`, `CommandComponent` and `CommandControl` which describes the necessary methods and properties for extending a WinForms Component or a WinForms Control with the command logic. Since a Component is not bindable by default, we introduce a new abstract class which is bringing all the necessary infrastructure to make a Component bindable. CommandComponent (literally) on top of that introduces a `Command` property, allows to executes an `Action` assigned to `Command`, when the context allows for the Command to be executed. The action assigned to `Command` is being passed the parameter which is stored in the Component's or Control's new property `CommandParameter`. This is the same way, commands work in WPF, UWP, Xamarin and Maui.

* Inherit `ButtonBase` from `CommandControl` instead of `Control`, so it will get all the necessary infrastructure for binding to commands, and for executing commands - if the context allows - `OnClick`.
* Inherit `ToolStripItem` for `CommandComponent`, to make a) all `ToolStripItem` derived components bindable, and b) introduce a bindable `Command` property on all `ToolStripItem` derived components which can be clicked. Note: `IBindableComponent` is not a new API. It was especially developed to provide Data Binding for components (and it's of course implemented in `Control` to make _it_ bindable). Implementing this Interface makes the new APIs `BindingContext` and `OnBindingContextChanged` necessary. The functionality of `BindingContext` is handed over to the existing static method `BindingContext.UpdateBindings`, which was especially designed to handle bindable Components. The idea is to seperate `BindableComponent` and `CommandComponent`, so other components can derive from `BindableComponent` to provide principle binding functionality, without necessarily to provide command binding.
* Introduce a property `DataContext` on `Control` as an ambient property, and - to meet the WinForms DataBinding conventions - implement a virtual protected method `OnDataContextChanged` and the `DataContextChanged` event along with it. This way a child (User)Control can marshal a new/changing data source to whatever `BindingSourceComponent` needs to be assigned to.

## FAQ

A few of most frequently asked questions up to now:

### Why do we need `On[Propertyname]Changed` methods and `[Propertyname]Changed Events` for each property

That's the convention in WinForms to support data binding back to the source: A property is bindable in the direction to the data source only, if the property can raise a respective `xxxChanged` event. A further convention in WinForms is that for each event there is a `Onxxx` method, that can be overwritten in a derived class, which raises the event, since events cannot be raised directly in a derived class.

### What's the reason for the naming

Naming pf new properties is based on precedence in existing UI stacks (`Command`, `CommandParameter`, `DataContext` in WPF). Naming of additional new events and methods is based on either existing interfaces or naming conventions in WinForms.

### Why are we only introducing Commands on `ButtonBase` and `ToolStripItem`

Because only those make sense in the context of executing commands, and that's the way is done also in WPF. We have been discussing LinkLabel, but came to the conclusion that it would probably too special a use case. Since we are providing an easy way to extend controls with commands by letting exsting Controls or Commands inherit from `CommandComponent` or `CommandControl`, users can easily implement additional scenarios in controls how they see fit. The same is true for control vendors.

### Do we need to extend the Designer?

No. As can be seen in the screenshots here, the Designer can deal with every of these changes out of the box. There is no necessity for additional work in the WinForms Designer.

If we have a typical Controller or ViewModel based on a [MVVM](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93viewmodel)-Framework like the [MVVM Toolkit](https://docs.microsoft.com/en-us/windows/communitytoolkit/mvvm/introduction) (or just implementing `INotifyPropertyChanged` directly like in the following sample), then the Designer can already work with that (see following screenshot):

```CS
using System;
using System.ComponentModel;

namespace WinFormsCommandBinding.Models.Controller
{
    public class SimpleEditorController : INotifyPropertyChanged
    {
        public event PropertyChangedEventHandler? PropertyChanged;

        private string? _textDocument;
        
        // Simple class which implements the details of ICommand.
        private RelayCommand? _newCommand;

        public SimpleEditorController()
        {
            _newCommand = new(ExecuteRelayCommand, CanExecuteRelayCommand);
        }

        private void ExecuteRelayCommand(object? commandParameter)
            => TextDocument = String.Empty;

        private bool CanExecuteRelayCommand(object? commandParameter)
            => TextDocument != String.Empty;

        public RelayCommand? NewCommand
        {
            get => _newCommand;

            set
            {
                if (!Equals(_newCommand, value))
                {
                    _newCommand = value;
                    PropertyChanged?.Invoke(this, new(nameof(NewCommand)));
                }
            }
        }

        public string? TextDocument
        {
            get => _textDocument;

            set
            {
                var wasEmpty = String.IsNullOrEmpty(_textDocument);

                if (!Equals(_textDocument, value))
                {
                    _textDocument = value;
                    PropertyChanged?.Invoke(this, new(nameof(TextDocument)));

                    if (wasEmpty ^ string.IsNullOrEmpty(_textDocument))
                    {
                        // Execution-Context changed because the TextDocument was empty
                        // and now it's not anymore or viceversa.
                        NewCommand?.RaiseCanExecuteChanged();
                    }
                }
            }
        }
    }
}
```

image

### Introducing the new abstract component `BindableComponent`

**Note:** Even though on first glance this might sense to put in the `System.Component` namespace and in the runtime, this is for the WinForms binding infrastructure only: `IBindableComponent` is in `System.Windows.Forms` and so is `BindingContext`.
Providing the infrastructure for Updating/Managing binding for Components has been implemented in WinForms all along, see here: [BindingContext.cs (dot.net)](https://source.dot.net/#System.Windows.Forms/System/Windows/Forms/BindingContext.cs,394)

```CS
{
    public abstract class BindableComponent : Component, IBindableComponent
    {
        public event EventHandler? BindingContextChanged;
        public ControlBindingsCollection? DataBindings { get; }
        public BindingContext? BindingContext { get; set; }
        protected virtual void OnBindingContextChanged(EventArgs e) { }
    }
}
```

### Introducing the new abstract component `CommandComponent`

```CS

    internal abstract class CommandComponent : BindableComponent
    {
        public event EventHandler? CommandCanExecuteChanged;
        public event EventHandler? CommandChanged;
        public event EventHandler? CommandParameterChanged;

        public abstract bool Enabled { get; set; }
        public System.Windows.Input.ICommand? Command { get; set; }
        public object? CommandParameter { get; set; }
        protected virtual void OnCommandChanged(EventArgs e) { }
        protected virtual void OnCommandCanExecuteChanged(EventArgs e) { }
        protected virtual void OnCommandParameterChanged(EventArgs e) { }
        protected virtual void OnRequestCommandExecute(EventArgs e) { }
    }
}
```

### Introducing the new abstract control `CommandControl`

```CS
{
    internal abstract class CommandControl : Control
    {
        public event EventHandler? CommandCanExecuteChanged;
        public event EventHandler? CommandChanged;
        public event EventHandler? CommandParameterChanged;

        public System.Windows.Input.ICommand? Command { get; set; }
        public object? CommandParameter { get; set; }
        protected virtual void OnCommandChanged(EventArgs e) { }
        protected virtual void OnCommandCanExecuteChanged(EventArgs e) { }
        protected virtual void OnCommandParameterChanged(EventArgs e) { }
        protected virtual void OnRequestCommandExecute(EventArgs e) { }
    }
}
```

### Inheriting `ButtonBase` from `CommandControl`

```diff
namespace System.Windows.Forms
{
    public abstract partial class ButtonBase
-        : Control
+        : CommandControl
    {
          .
          .
          .
    }
}
```

### Inheriting `ToolStripItem` from `CommandComponent`

```diff
namespace System.Windows.Forms
{
    public abstract partial class ToolStripItem 
-                            : Component,
+                            : CommandComponent,
                              IDropTarget,
                              ISupportOleDropSource,
                              IArrangedElement,
                              IKeyboardToolTip
    {
    .
    .
    .
    }
}
```

### Introducing `DataContext` on the existing class `class Control`

We need an extension point for assigning the data sources not only to a `Form` or a `UserControl`, but at the same time make that data source available to each of the child (User)Controls which this root control is hosting. To this end, we introduce the `DataContext` Property of type `object` on `Control`, and we implement it as an ambient property. This way, it doesn't use any additional memory resources, as long as it is not used. When it is assigned to a Form, its `OnDataContextChanged` for the Form (and for each respective child control) is called. `OnDataContextChanged` then raises the `DataContextChanged` event. The root control or/and the children can handle that event to decide, what to do with the data source: Usually, the data source then gets assigned to the `BindingSource` component(s) the Control uses, which then initiates the actual data binding. 

```CS
{
    public partial class Control :
        Component,
        Ole32.IOleControl,
        Ole32.IOleObject,
        Ole32.IOleInPlaceObject,
        Ole32.IOleInPlaceActiveObject,
        Ole32.IOleWindow,
        Ole32.IViewObject,
        Ole32.IViewObject2,
        Ole32.IPersist,
        Ole32.IPersistStreamInit,
        Oleaut32.IPersistPropertyBag,
        Ole32.IPersistStorage,
        Ole32.IQuickActivate,
        ISupportOleDropSource,
        IDropTarget,
        ISynchronizeInvoke,
        IWin32Window,
        IArrangedElement,
        IBindableComponent,
        IKeyboardToolTip,
        IHandle
    {
        public event EventHandler DataContextChanged;

        public virtual Object? DataContext { get; set; }

        protected virtual void OnDataContextChanged(EventArgs e) { }
        protected virtual void OnParentDataContextChanged(EventArgs e) { }

    }
}
```

KlausLoeffelmann:
This fixes #4895.

- [x] Implement `ICommandBindingTargetProvider` with Default Interface Methods holding 90% of the `Command` logic, so it can be easily reimplemented in other Components/Controls where is makes sense. This is no longer public, but internal, just to not repeat implementation code.
- [x] Implement `BindableComponent` to have a Component which can be bound at Design time the same way it works with `Control` already (both are implementing `IBindableComponent`).
- [x] Implement `ICommandBindingTargetProvider` into `ToolStripItem` to let it have a `Command` property of type `ICommand` which can be bound to from Controllers/ViewModels.
- [x] Implement `ICommandBindingTargetProvider` into `ButtonBase` to let it have a `Command` property of type `ICommand` which can be bound to from Controllers/ViewModels.
- [x] Implement `DataContext` Property as an ambient property along with `DataContextChanged` event, `OnDataContextChanged` and `OnParentDataContextChanged`.
- [x] Update demo to test implementations and to show the general usage of Controller/Models for WinForms Apps. See https://github.com/KlausLoeffelmann/WinFormsCommandBinding/tree/main/src. For this to be working, compile PR, patch .NET 7 binaries (assemblies, refs) and use ModifiedRuntime sample.
- [x] Review the general approach in the internal WinForms Technical Sync.
- [x] Apply changes from discussions (In addition: removed `CommandEventArgs`, added `CommandParameter` property).
- [x] Add a Maui sample to the demo, which uses the same ViewModel/UIController.
- [x] Write/Modify the demo for this, and test the hell out of it.
- [x] Go through API review.
- [x] Apply changes from API review.
- [x] Write/Update XML comments.
- [x] Add unit tests.

Update: After having an internal team discussion, we decided to rename `BindableCommand` back to be just `Command`.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7143)


src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
```diff
@@ -407,14 +490,6 @@ public virtual Image BackgroundImage
             }
         }
 
-        // Every ToolStripItem needs to cache its last/current Parent's DeviceDpi
-        // for PerMonitorV2 scaling purposes.
-        internal virtual int DeviceDpi
```
RussKie, line 412: Alphabetising members is a great idea, however please do it in a dedicated commit, separate from all other changes.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
```diff
@@ -329,6 +342,30 @@ public virtual bool AllowDrop
...
+        [Browsable(false)]
+        [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
+        [DefaultValue(CommonProperties.DefaultAnchor)]
+        public AnchorStyles Anchor
+        {
+            get
+            {
+                // since we don't support DefaultLayout go directly against the CommonProperties
+                return CommonProperties.xGetAnchor(this);
+            }
```
RussKie, line 354: [nit]
```suggestion
            get => CommonProperties.xGetAnchor(this);
```

src/System.Windows.Forms/src/System/Windows/Forms/IBindableCommandProvider.cs
```diff
@@ -0,0 +1,131 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
RussKie, line 1: Filenames must match types contained within.

src/System.Windows.Forms/src/System/Windows/Forms/IBindableCommandProvider.cs
```diff
@@ -0,0 +1,131 @@
...
+            HandleCommandCanExecuteChanged(sender, commandEventArgs);
+
+            if (!commandEventArgs.Cancel)
+            {
+                Enabled = Command?.CanExecute(commandEventArgs.Parameter) ?? false;
+            }
+        }
+    }
+
+    internal class CommandControl : Control, ICommandProvider
```
RussKie, line 91: One type per file

src/System.Windows.Forms/src/System/Windows/Forms/ExecuteCommandEventArgs.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+#nullable disable
```
RussKie, line 5: All new API must be NRT annotated

src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.cs
```diff
@@ -1040,6 +1114,33 @@ internal virtual TextFormatFlags CreateTextFormatFlags()
...
         }
 
+        /// <summary>
+        ///  Raises the <see cref="ToolStripItem.CommandChanged"/> event.
+        ///  Inheriting classes should override this method to handle this event.
+        ///  Call base.CommandChanged to send this event to any registered event listeners.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Advanced)]
+        protected virtual void OnCommandChanged(EventArgs e)
+            => ((EventHandler)Events[s_commandChangedEvent]!)?.Invoke(this, e);
```
RussKie, line 1124:
```suggestion
            => ((EventHandler?)Events[s_commandChangedEvent])?.Invoke(this, e);
```

src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.cs
```diff
@@ -821,6 +883,18 @@ protected override void OnGotFocus(EventArgs e)
             Invalidate();
         }
 
+        // Called by the ICommandProvider's internal DIM-based logic.
+        void ICommandProvider.HandleCommandChanged(EventArgs e)
+            => OnCommandChanged(e);
+
+        // Called by the ICommandProvider's internal DIM-based logic.
+        void ICommandProvider.HandleCommandCanExecuteChanged(object sender, CommandEventArgs e)
```
RussKie, line 891: To align with .NET runtime annotations:
```suggestion
        void ICommandProvider.HandleCommandCanExecuteChanged(object? sender, CommandEventArgs e)
```

ditto for other instances.

RussKie (commented):
As discussed, please bring the API proposal up-to-date. Few nits and suggestions.

src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.cs
```diff
@@ -1040,6 +1145,42 @@ internal virtual TextFormatFlags CreateTextFormatFlags()
...
+            => ((EventHandler)Events[s_commandCanExecuteChangedEvent]!)?.Invoke(sender, e);
+
+        /// <summary>
+        ///  Raises the <see cref="ButtonBase.CommandParameterChanged"/> event.
+        ///  Inheriting classes should override this method to handle this event.
+        ///  Call base.CommandParameterChanged to send this event to any registered event listeners.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Advanced)]
+        protected virtual void OnCommandParameterChanged(EventArgs e)
+            => ((EventHandler)Events[s_commandParameterChangedEvent]!)?.Invoke(this, e);
```
RussKie, line 1182: All of these are incorrect. Please refer to https://github.com/dotnet/winforms/pull/7143/files#r869843036

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -890,6 +893,47 @@ internal Gdi32.HBRUSH BackColorBrush
...
         }
 
+        /// <summary>
+        ///  An ambient property to hold a data source for binding purposes.
+        ///  Reassignment to BindingSources or other binding mechanism is
+        ///  the responsibility of the control once notified by
+        ///  the <see cref="Control.DataContextChanged"/> event.
+        /// </summary>
+        [SRCategory(nameof(SR.CatData))]
+        public virtual Object? DataContext
```
RussKie, line 903:
```suggestion
        public virtual object? DataContext
```

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -7309,6 +7365,32 @@ protected virtual void OnCursorChanged(EventArgs e)
...
+            {
+                eventHandler(this, e);
+            }
+
+            ControlCollection controlsCollection = (ControlCollection)Properties.GetObject(s_controlsCollectionProperty);
+            if (controlsCollection is not null)
+            {
+                // PERFNOTE: This is more efficient than using Foreach.  Foreach
+                // forces the creation of an array subset enum each time we
+                // enumerate
```
RussKie, line 7386: Whilst this is a good note, it's not required - we don't use linq in the codebase.

KlausLoeffelmann: I can delete it, yes, but I think it didn't refer to LINQ. It's a copy-paste note from the original source. And I am not sure if it related to LINQ, but to For Each.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
KlausLoeffelmann: This was marked as new through resorting, but I never touched it.

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
KlausLoeffelmann: Good catch, thanks!

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -7501,6 +7580,19 @@ protected virtual void OnParentCursorChanged(EventArgs e)
             }
         }
 
+        [EditorBrowsable(EditorBrowsableState.Advanced)]
+        protected virtual void OnParentDataContextChanged(EventArgs e)
+        {
+            // We don't test on null here, because a root's (Form's) DataContext 
+            // could be {}, this control's parent then became set to null, and now 
+            // its children in that case _need_ to receive the DataContextChanged event as well.
+            // The parent DataContext value, though, in that case would be null.
```
RussKie, line 7589: Please reformat to 120 chars long.
Applies to the added xml-docs as well.

RussKie, line 7589: It'd be great if you could help me understand this comment. 

Here's what I understand:
- The form may have `DataContext` set.
- The container (i.e., parent of this control) may have had its `DataContext` set to null.

But from this point I get lost. The children of which control need to be notified?
And more generally, why do we care if `DataContext` is null to begin with?

src/System.Windows.Forms/src/System/Windows/Forms/ICommandProvider.cs
```diff
@@ -0,0 +1,155 @@
...
+using System.ComponentModel;
+using System.Diagnostics.CodeAnalysis;
+using System.Windows.Input;
+
+namespace System.Windows.Forms
+{
+    public interface ICommandProvider
+    {
+        /// <summary>
+        /// Occurs, when the Command property has changed.
```
RussKie, line 14:
```suggestion
        /// Occurs when the Command property has changed.
```

ditto other places

src/System.Windows.Forms/src/System/Windows/Forms/ICommandProvider.cs
```diff
@@ -0,0 +1,155 @@
...
+        /// </summary>
+        /// <param name="sender">The sender of the event, passed from the data source.</param>
+        /// <param name="e">An empty <see cref="EventArgs"/> instance.</param>
+        protected void RaiseCommandCanExecuteChanged([AllowNull] object sender, EventArgs e);
+
+        /// <summary>
+        /// An implementation should raise the <see cref="CommandExecuting"/> event 
+        /// by calling component's or control's OnCommandExecuting method.
+        /// </summary>
+        /// <param name="e"></param>
```
RussKie, line 70: Missing. Ditto other places.

src/System.Windows.Forms/src/System/Windows/Forms/ICommandProvider.cs
```diff
@@ -0,0 +1,155 @@
...
+
+            if (!e.Cancel && (commandComponent.Command?.CanExecute(commandComponent.CommandParameter) ?? false))
+            {
+                commandComponent.Command?.Execute(commandComponent.CommandParameter);
+            }
+        }
+
+        private void CommandSetter(
+            ICommand? newCommand,
+            ref ICommand? commandBackingField)
```
RussKie, line 111: Keep this on the same line

src/System.Windows.Forms/src/System/Windows/Forms/ICommandProvider.cs
```diff
@@ -0,0 +1,155 @@
...
+                    return;
+                }
+
+                commandBackingField.CanExecuteChanged += CommandCanExecuteChangedProc;
+                PreviousEnabledStatus ??= Enabled;
+                Enabled = commandBackingField.CanExecute(CommandParameter);
+            }
+        }
+
+        private void CommandCanExecuteChangedProc([AllowNull] object sender, EventArgs e)
```
RussKie, line 144:
```suggestion
        private void Command_CanExecuteChanged(object? sender, EventArgs e)
```
ditto other places

src/System.Windows.Forms/src/System/Windows/Forms/ICommandProvider.cs
```diff
@@ -0,0 +1,155 @@
...
+        /// Method which should be called by the class implementing this
+        /// interface, when the assigned <see cref="Command"/> should be executed.
+        /// </summary>
+        /// <remarks>
+        /// As an example, a <see cref="Button"/> should call this method inside the method which
+        /// also raises the <see cref="Control.Click"/> Event of that Button, which
+        /// would be the <see cref="Button.OnClick(EventArgs)"/> OnClick method.
+        /// </remarks>
+        /// <param name="commandComponent"></param>
+        protected static void RequestCommandExecute(ICommandProvider commandComponent)
```
RussKie, line 98: We need to validate the args on the public API, and add xml-docs for the ANE.
```suggestion
        protected static void RequestCommandExecute(ICommandProvider!! commandComponent)
```
See https://github.com/dotnet/runtime/pull/64720.

Ditto other places that allow non-nullable input args.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
```diff
@@ -481,19 +624,36 @@ public virtual Color BackColor
...
-        {
-            add => Events.AddHandler(s_backColorChangedEvent, value);
-            remove => Events.RemoveHandler(s_backColorChangedEvent, value);
-        }
+        /// <summary>
+        /// The bounds of the item
+        /// </summary>
+        [Browsable(false)]
+        public virtual Rectangle Bounds
+            => _bounds;
```
RussKie, line 632: We generally don't wrap short lines.
```suggestion
        public virtual Rectangle Bounds => _bounds;
```

Ditto other places

RussKie:
@gewarren when you have a chance, may I get you to review from the docs perspective?

src/System.Windows.Forms/src/System/Windows/Forms/ICommandProvider.cs
```diff
@@ -0,0 +1,155 @@
...
+        /// <param name="e">An empty <see cref="EventArgs"/> instance.</param>
+        protected void RaiseCommandChanged(EventArgs e);
+
+        /// <summary>
+        /// An implementation should raise the <see cref="CommandCanExecuteChanged"/> event 
+        /// by calling component's or control's OnCommandCanExecuteChanged method.
+        /// </summary>
+        /// <param name="sender">The sender of the event, passed from the data source.</param>
+        /// <param name="e">An empty <see cref="EventArgs"/> instance.</param>
+        protected void RaiseCommandCanExecuteChanged([AllowNull] object sender, EventArgs e);
```
RussKie, line 64: This makes me uncomfortable - we don't have any public API like that (i.e., `RaiseXyz`) that accept `sender`. 
Why is `sender` necessary?

src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.cs
```diff
@@ -1042,6 +1147,42 @@ internal virtual TextFormatFlags CreateTextFormatFlags()
             return Adapter.CreateTextFormatFlags();
         }
 
+        /// <summary>
+        ///  Raises the <see cref="ButtonBase.CommandChanged"/> event.
+        ///  Inheriting classes should override this method to handle this event.
+        ///  Call base.CommandChanged to send this event to any registered event listeners.
```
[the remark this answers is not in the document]
gewarren: Add a `<block>` element like this:

```
<block subset="none" type="overrides">
  Override this method to handle the event. In the overriding method, call ...
</block>
```

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -890,6 +893,47 @@ internal Gdi32.HBRUSH BackColorBrush
             }
         }
 
+        /// <summary>
+        ///  An ambient property to hold a data source for binding purposes.
```
gewarren, line 897:
```suggestion
        ///  Gets or sets the data source for binding purposes.
```

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -890,6 +893,47 @@ internal Gdi32.HBRUSH BackColorBrush
...
+        ///  Reassignment to BindingSources or other binding mechanism is
```
gewarren, line 898: This might belong in remarks instead of summary?

src/System.Windows.Forms/src/System/Windows/Forms/ICommandProvider.cs
```diff
@@ -0,0 +1,155 @@
...
+        /// <param name="e"></param>
+        protected void RaiseCommandExecuting(CancelEventArgs e);
+
+        /// <summary>
+        /// Method which should be called by a class implementing this 
+        /// interface in the setter of the <see cref="Command"/> property.
+        /// </summary>
+        /// <param name="commandComponent">Instance of the class implementing this interface.</param>
+        /// <param name="newCommand">The new value of the <see cref="Command"/>
+        /// which should be assigned to the property.</param>
```
gewarren, line 79:
```suggestion
        /// that should be assigned to the property.</param>
```

src/System.Windows.Forms/src/System/Windows/Forms/ICommandProvider.cs
```diff
@@ -0,0 +1,155 @@
...
+            ICommand? newCommand,
+            ref ICommand? commandBackingField)
+            => commandComponent.CommandSetter(newCommand, ref commandBackingField);
+
+        /// <summary>
+        /// Method which should be called by the class implementing this
+        /// interface, when the assigned <see cref="Command"/> should be executed.
+        /// </summary>
+        /// <remarks>
+        /// As an example, a <see cref="Button"/> should call this method inside the method which
```
gewarren, line 93:
```suggestion
        /// As an example, a <see cref="Button"/> should call this method inside the method that
```

src/System.Windows.Forms/src/System/Windows/Forms/ICommandProvider.cs
```diff
@@ -0,0 +1,155 @@
...
+        /// also raises the <see cref="Control.Click"/> Event of that Button, which
```
gewarren, line 94:
```suggestion
        /// also raises the <see cref="Control.Click"/> event of that Button, which
```

src/System.Windows.Forms/src/System/Windows/Forms/ICommandProvider.cs
```diff
@@ -0,0 +1,155 @@
...
+        /// would be the <see cref="Button.OnClick(EventArgs)"/> OnClick method.
```
gewarren, line 95:
```suggestion
        /// is the <see cref="Button.OnClick(EventArgs)"/> method.
```

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
```diff
@@ -329,6 +344,30 @@ public virtual bool AllowDrop
...
+                    if (ParentInternal is not null)
+                    {
+                        LayoutTransaction.DoLayout(this, ParentInternal, PropertyNames.Anchor);
+                    }
+                }
+            }
+        }
+
         /// <summary>
         ///  Determines whether we set the ToolStripItem to its preferred size
```
gewarren, line 372:
```suggestion
        ///  Gets or sets a value that indicates whether to set the ToolStripItem to its preferred size.
```

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
```diff
@@ -481,19 +624,36 @@ public virtual Color BackColor
...
 
-        [SRCategory(nameof(SR.CatPropertyChanged))]
-        [SRDescription(nameof(SR.ToolStripItemOnBackColorChangedDescr))]
-        public event EventHandler BackColorChanged
-        {
-            add => Events.AddHandler(s_backColorChangedEvent, value);
-            remove => Events.RemoveHandler(s_backColorChangedEvent, value);
-        }
+        /// <summary>
+        /// The bounds of the item
```
gewarren, line 628:
```suggestion
        /// Gets the bounds of the item.
```

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
```diff
@@ -481,19 +624,36 @@ public virtual Color BackColor
...
+        /// </summary>
+        [Browsable(false)]
+        public virtual Rectangle Bounds
+            => _bounds;
 
         /// <summary>
-        ///  The bounds of the item
+        ///  Determines whether or not the item can be selected.
```
gewarren, line 635:
```suggestion
        ///  Gets a value that indicates whether or not the item can be selected.
```

gewarren (commented):
In case it's helpful, the guidelines for documenting event-related APIs are here: https://github.com/dotnet/dotnet-api-docs/wiki/Summary%3A-Event-related

src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.cs
```diff
@@ -166,6 +177,78 @@ public override Color BackColor
...
+        [SRCategory(nameof(SR.CatData))]
+        public System.Windows.Input.ICommand? Command
+        {
+            get => _command;
+            set => ICommandProvider.CommandSetter(this, value, ref _command);
+        }
+
+        /// <summary>
+        /// Occurs when the Command has changed
+        /// </summary>
```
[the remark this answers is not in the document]
RussKie: An extra offset increases readability, e.g.:
image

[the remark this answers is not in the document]
KlausLoeffelmann: We had problems in the past (Sandcastle) that those extra spaces where then in the docs. That's no longer a problem then anymore, I assume?

Also, we are not doing it in the runtime. So, why are we doing it differently?
Shouldn't we change it to be in parity with the runtime?

I personally don't like that additional character.

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
KlausLoeffelmann: `DataContext` is an ambient property. Like `Font`. So, if it's defined once in a parent control, children take that setting simply by referring to it. So, consider Form's DC is something, then a user control in that Form would also have that something as DC - because it's ambient. _Now_ you set the UserControl's DC explicitly to `null` in addition. In that moment you want to make sure, that those children again receive the DC changed, _even_ though technically theirs _is_ already `null` and their direct parent's is null.

src/System.Windows.Forms/src/System/Windows/Forms/ICommandProvider.cs
KlausLoeffelmann: I was struggling with that as well, to be honest, but it seemed to be the most logical choice.
In this case, the ViewModel/DataSource had technically indicated the Command's execution context by raising that event. So, it should also be the ViewModel to be indicated as the sender, because _it is_ technically the sender.

Now, if it's more about obtaining the standard than the logic, that would also be fine with me.
I guess in that case, we would need to introduce additional `EventArgs`, and introduce something like `OriginalSender`. Would you feel more comfortable with that?

I'd also be interested in additional ideas around that. @Tanya-Solyanik, @JeremyKuhne?

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
KlausLoeffelmann: Oh, absolutely!

src/System.Windows.Forms/src/System/Windows/Forms/ICommandProvider.cs
```diff
@@ -0,0 +1,155 @@
...
+        private void CommandCanExecuteChangedProc([AllowNull] object sender, EventArgs e)
+        {
+            CancelEventArgs cancelEventArgs = new();
+            RaiseCommandCanExecuteChanged(sender, cancelEventArgs);
+
+            if (!cancelEventArgs.Cancel)
+            {
+                Enabled = Command?.CanExecute(CommandParameter) ?? false;
+            }
+        }
```
RussKie, line 153: This look strange to say the least....

src/System.Windows.Forms/src/System/Windows/Forms/ICommandProvider.cs
RussKie: As dicussed you may wnat to verify this is all we need:
```suggestion
        private void Command_CanExecuteChanged(object? sender, EventArgs e)
        {
           Command?.RaiseCommandCanExecuteChanged();
           Enabled = Command?.CanExecute(CommandParameter) ?? false;
        }
```

src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.cs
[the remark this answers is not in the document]
RussKie: We don't use Sandcastle. The docs are xml that gets turned into html, and html is indifferent to spaces.

> Also, we are not doing it in the runtime. So, why are we doing it differently?
> Shouldn't we change it to be in parity with the runtime?

Which runtime are you referring to? :) There are many differences between Windows Forms runtime and .NET runtime in various points.

src/System.Windows.Forms/src/System/Windows/Forms/CommandComponent.cs
```diff
@@ -0,0 +1,132 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.ComponentModel;
+
+namespace System.Windows.Forms
+{
+    public abstract class CommandComponent : BindableComponent, ICommandPropertyProvider
```
RussKie, line 9: All new public API must be xml-doc annotated.

src/System.Windows.Forms/src/System/Windows/Forms/CommandComponent.cs
```diff
@@ -0,0 +1,132 @@
...
+    {
+        // Holds the logic for controlling a command's execution context.
+        private ICommandPropertyProvider.CommandProviderManager _commandProviderManager;
+
+        // Backing fields for the infrastructure to make ToolStripItem bindable and introduce (bindable) ICommand.
+        private System.Windows.Input.ICommand? _command;
+        private object? _commandParameter;
+
+        internal static readonly object s_commandChangedEvent = new object();
```
RussKie, line 18:
```suggestion
        internal static readonly object s_commandChangedEvent = new();
```

ditto other places

src/System.Windows.Forms/src/System/Windows/Forms/CommandComponent.cs
```diff
@@ -0,0 +1,132 @@
...
+        internal static readonly object s_commandParameterChangedEvent = new object();
+        internal static readonly object s_commandCanExecuteChangedEvent = new object();
+
+        public CommandComponent() : base()
+        {
+            _commandProviderManager = new(this);
+        }
+
+        public abstract bool Enabled { get; set; }
```
RussKie, line 27: This should be placed below `CommandParameterChanged`

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -890,6 +893,47 @@ internal Gdi32.HBRUSH BackColorBrush
...
+        ///  the responsibility of the control once notified by
+        ///  the <see cref="Control.DataContextChanged"/> event.
+        /// </summary>
+        [SRCategory(nameof(SR.CatData))]
+        public virtual object? DataContext
```
RussKie, line 903: Should data context be a part of `CommandControl`?

src/System.Windows.Forms/src/System/Windows/Forms/BindableComponent.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        private ControlBindingsCollection? _dataBindings;
+        private BindingContext? _bindingContext;
+
+        [Browsable(false)]
+        [EditorBrowsable(EditorBrowsableState.Advanced)]
+        [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
+        [SRCategory(nameof(SR.CatData))]
+        [SRDescription(nameof(SR.ToolStripItemBindingContextDescr))]
+        public BindingContext BindingContext
```
RussKie, line 21: What's the difference between `BindingContext` (used here) and `DataContext` (used in `Control`)?

src/System.Windows.Forms/src/System/Windows/Forms/ICommandPropertyProvider.cs
```diff
@@ -0,0 +1,114 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Windows.Input;
+
+namespace System.Windows.Forms
+{
+    internal interface ICommandPropertyProvider
```
RussKie, line 9: https://github.com/dotnet/winforms/issues/4895#issuecomment-1184777610 gives me an impression that the interface should go... Am I reading it correctly?

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
KlausLoeffelmann: Since we refactored the whole thing after the API review, the original code was restored (except one line), so this is no longer an issue.

src/System.Windows.Forms/src/System/Windows/Forms/ICommandProvider.cs
KlausLoeffelmann: No longer applies.

src/System.Windows.Forms/src/System/Windows/Forms/ICommandProvider.cs
KlausLoeffelmann: It's gone.

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
KlausLoeffelmann: No. `DataContext` is a carrier for data sources which need to be passed down the line. For example: A form consists of a series of UserControls. `DataContext` - as an ambient property - provides a central place to assign the data source. Now, since it's ambient - the data source is handed down to each respective child control in every container's `ControlsControllection`. The receiving [User]Controls get notified and can assign their BindingSources as they see fit. 

This API has been cleared by the API Review already.

src/System.Windows.Forms/src/System/Windows/Forms/ICommandPropertyProvider.cs
KlausLoeffelmann: Well, yes and no. Certainly we don't need it publicly, because best practice is now to just inherit from `BindableComponent`, `CommandComponent` and `CommandControl`. This interface just helps to internally to not have redundant code. It's not exposed, so we meet their request.

src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.cs
```diff
@@ -629,6 +711,17 @@ internal bool MouseIsOver
...
 
+        /// <summary>
+        /// Provides the previous Enabled value for the ICommandProvider implementation.
+        /// </summary>
+        /// <remarks>
+        /// The interface uses DFI to bring a consistent Command binding implementation along.
+        /// It therefore needs a backing field to save a previous Enabled setting, since
+        /// a components/controls Enabled property is automatically controlled by the Command
+        /// binding through the CanExecute functionality of the Command binding.
+        /// </remarks>
```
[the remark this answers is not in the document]
KlausLoeffelmann: Does no longer apply.

src/System.Windows.Forms/src/System/Windows/Forms/BindableComponent.cs
```diff
@@ -0,0 +1,74 @@
...
+        [EditorBrowsable(EditorBrowsableState.Advanced)]
+        protected virtual void OnBindingContextChanged(EventArgs e)
+        {
+            if (_bindingContext is not null)
+            {
+                for (int i = 0; i < DataBindings.Count; i++)
+                {
+                    BindingContext.UpdateBinding(BindingContext, DataBindings[i]);
+                }
+            }
```
RussKie, line 60: Mixing backing fields (i.e., `_bindingContext`) and properties (i.e., `BindingContext`) often makes it harder to reason about the state. Can we use one or the other?
Also using `DataBindings` may force creation of the backing field, so should this be as follows?
```suggestion
            if (_bindingContext is not null && _dataBindings is not null)
            {
                for (int i = 0; i < _dataBindings.Count; i++)
                {
                    _bindingContext.UpdateBinding(_bindingContext, _dataBindings[i]);
                }
            }
```

Btw, this looks very strange: `BindingContext.UpdateBinding(BindingContext, ...)`

src/System.Windows.Forms/src/System/Windows/Forms/CommandComponent.cs
```diff
@@ -0,0 +1,132 @@
...
+        // Holds the logic for controlling a command's execution context.
+        private readonly ICommandPropertyProvider.CommandProviderManager _commandProviderManager;
+
+        // Backing fields for the infrastructure to make ToolStripItem bindable and introduce (bindable) ICommand.
+        private System.Windows.Input.ICommand? _command;
+        private object? _commandParameter;
+
+        internal static readonly object s_commandChangedEvent = new();
+        internal static readonly object s_commandParameterChangedEvent = new();
+        internal static readonly object s_commandCanExecuteChangedEvent = new();
```
RussKie, line 20: Why are these leaking outside this type?
Ditto below

src/System.Windows.Forms/src/System/Windows/Forms/ICommandPropertyProvider.cs
```diff
@@ -0,0 +1,113 @@
...
+    {
+        public class CommandProviderManager
```
RussKie, line 11: If we don't expect this to be inherited let's seal it, as it'd allow the compiler to do a better job.
```suggestion
        public sealed class CommandProviderManager
```

src/System.Windows.Forms/src/System/Windows/Forms/ICommandPropertyProvider.cs
```diff
@@ -0,0 +1,113 @@
...
+            public CommandProviderManager(ICommandPropertyProvider commandPropertyProvider)
+            {
+                _commandPropertyProvider = commandPropertyProvider;
+            }
+
+            public void CommandSetter(
+                ICommand? newCommand,
+                ref ICommand? commandBackingField)
+            {
+                if (!Equals(commandBackingField, newCommand))
```
RussKie, line 25: Invert to reduce nesting

RussKie (commented):
Looks good. 

As mentioned before all public API (i.e., marked as `public`, `protected` or explicit interface implementations) must be correctly xml-doc annotated.

src/System.Windows.Forms/src/System/Windows/Forms/BindableComponent.cs
KlausLoeffelmann: In this case `BindingContext` is the type name - we're calling the _static_ method 'UpdateBinding'. It may look a bit weird, but it should be fine like this.

src/System.Windows.Forms/src/System/Windows/Forms/ICommandPropertyProvider.cs
KlausLoeffelmann: Outdated.

src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.cs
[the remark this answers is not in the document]
KlausLoeffelmann: These comments don't make sense anymore, since we got rid of the respective base classes.

src/System.Windows.Forms/src/System/Windows/Forms/ICommandBindingTargetProvider.cs
```diff
@@ -0,0 +1,130 @@
...
+
+using System.Diagnostics.CodeAnalysis;
+using System.Windows.Input;
+
+namespace System.Windows.Forms
+{
+    internal interface ICommandBindingTargetProvider
+    {
+        /// <summary>
+        /// Occurs, when the Command property has changed.
```
RussKie, line 13:
```suggestion
        /// Occurs when the <see cref="Command"/> property has changed.
```

ditto below

src/System.Windows.Forms/src/System/Windows/Forms/ICommandBindingTargetProvider.cs
```diff
@@ -0,0 +1,130 @@
...
+        ICommand? Command { get; set; }
+
+        /// <summary>
+        /// Gets or sets the CommandParameter for the Component or Control.
+        /// </summary>
+        object? CommandParameter { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether the control 
+        /// can respond to user interaction.
```
RussKie, line 35: This line is well under 120 chars, please keep it on the same line. Ditto other similar instances.

Please install the [editor guidelines extension](https://github.com/dotnet/winforms/blob/main/docs/developer-guide.md#machine-setup), which will help you in such situations.

src/System.Windows.Forms/src/System/Windows/Forms/ICommandBindingTargetProvider.cs
```diff
@@ -0,0 +1,130 @@
...
+        /// the <see cref="Command"/> property.</param>
+        protected static void CommandSetter(
+            ICommandBindingTargetProvider commandComponent,
+            ICommand? newCommand,
+            ref ICommand? commandBackingField)
+            => commandComponent.CommandSetter(newCommand, ref commandBackingField);
+
+        /// <summary>
+        /// Method which should be called by the class implementing this
+        /// interface, when the assigned <see cref="Command"/> should be executed.
```
RussKie, line 76: The summary should explain what the method does not who should call it. 
Who and/or when should invoke the method could be explained in the remarks section.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripComboBox.cs
```diff
@@ -398,7 +398,8 @@ private void HandleTextUpdate(object sender, EventArgs e)
         {
             OnTextUpdate(e);
         }
-
+        
+#pragma warning disable CA2252
```
RussKie, line 402: As discussed please have comments added for all suppression/restorations similar to how we do the same in other places
image

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
```diff
@@ -407,6 +418,89 @@ public virtual Image BackgroundImage
             }
         }
 
+        /// <summary>
+        ///  Gets or sets the <see cref="System.Windows.Input.ICommand"/> object whose <see cref="System.Windows.Input.ICommand.Execute(object?)"/>
+        ///  method will be called when the ToolStripItem's <see cref="Click"/> event gets invoked and
+        ///  <see cref="System.Windows.Input.ICommand.CanExecute(object?)"/> indicates that the command can be executed.
+        /// </summary>
```
RussKie, line 425: I don't think the last bit is nessecary. Ditto for the ButtonBase class
```suggestion
        /// <summary>
        ///  Gets or sets the <see cref="System.Windows.Input.ICommand"/> object whose <see cref="System.Windows.Input.ICommand.Execute(object?)"/>
        ///  method will be called when <see cref="Click"/> event is invoked.
        /// </summary>
```

RussKie (changes requested):
Looking good!

src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.cs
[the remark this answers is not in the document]
KlausLoeffelmann: Yes, you're right.

src/System.Windows.Forms/src/System/Windows/Forms/BindableComponent.cs
```diff
@@ -0,0 +1,93 @@
...
+        public event EventHandler BindingContextChanged
+        {
+            add => Events.AddHandler(s_bindingContextChangedEvent, value);
+            remove => Events.RemoveHandler(s_bindingContextChangedEvent, value);
+        }
+
+        /// <summary>
+        ///  Raises the <see cref="BindableComponent.BindingContextChanged"/> event.
+        ///  Inheriting classes should override this method to handle this event.
+        ///  Call base.OnBindingContextChanged to send this event to any registered event listeners.
```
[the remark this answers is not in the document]
KlausLoeffelmann: @gewarren: I would like to know if all that isn't some standard-ceremony kind of event lyrics, which (can) get inserted automatically?

src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.cs
```diff
@@ -167,6 +176,89 @@ public override Color BackColor
...
+        [Browsable(false)]
+        [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
+        [SRCategory(nameof(SR.CatData))]
+        [SRDescription(nameof(SR.CommandComponentCommandParameterDescr))]
+        public object? CommandParameter
+        {
+            [RequiresPreviewFeatures]
+            get => _commandParameter;
+
+            // We need to opt into previre features here, because we calling a preview feature from the setter.
```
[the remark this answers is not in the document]
KlausLoeffelmann: It'll not hurt either. If there is a need to document the conditional compilation directive to ignore the `RequiresPreviewFeatures`, then we should also include a hint, why this doesn't sit on top of the property, but rather on the getter and setter individually. I am changing the comment though, to reflect this clearer.

src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.cs
```diff
@@ -1042,6 +1146,55 @@ internal virtual TextFormatFlags CreateTextFormatFlags()
...
+            => RaiseEvent(s_commandChangedEvent, e);
+
+        /// <summary>
+        ///  Raises the <see cref="ButtonBase.CommandCanExecuteChanged"/> event.
+        /// </summary>
+        /// <param name="e">An empty <see cref="EventArgs"/> instance.</param>
+        [RequiresPreviewFeatures]
+        [EditorBrowsable(EditorBrowsableState.Advanced)]
+        protected virtual void OnCommandCanExecuteChanged(EventArgs e)
+            // TODO: Is passing 'this' correct here?
```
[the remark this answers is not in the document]
KlausLoeffelmann: No. I need to talk with @JeremyKuhne about this.

src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.cs
```diff
@@ -1042,6 +1146,55 @@ internal virtual TextFormatFlags CreateTextFormatFlags()
...
+        [RequiresPreviewFeatures]
+        protected virtual void OnRequestCommandExecute(EventArgs e)
+            => ICommandBindingTargetProvider.RequestCommandExecute(this);
+
+        // Called by the CommandProviderManager's internal DIM-based logic.
+        [RequiresPreviewFeatures]
+        void ICommandBindingTargetProvider.RaiseCommandChanged(EventArgs e)
+            => OnCommandChanged(e);
+
+        // Called by the CommandProviderManager's internal DIM-based logic.
```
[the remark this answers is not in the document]
KlausLoeffelmann: Rephrased.

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -890,6 +893,58 @@ internal Gdi32.HBRUSH BackColorBrush
...
+        ///  Gets or sets the data context for the purpose of data binding.
+        ///  This is an ambient property.
+        /// </summary>
+        /// <remarks>
+        ///  Data context is a concept that allows elements to inherit information from their parent elements
+        ///  about the data source that is used for binding. It's the duty of deriving controls which inherit from
+        ///  this class to handle the provided data source accordingly. For example, UserControls, which using
+        ///  <see cref="BindingSource"/> components for data binding scenarios could either handle the
+        ///  <see cref="DataContextChanged"/> event or overwriting <see cref="OnDataContextChanged(EventArgs)"/> to provide
+        ///  the relevant data from the data context to a BindingSource component's <see cref="BindingSource.DataSource"/>.
```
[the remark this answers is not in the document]
KlausLoeffelmann: Yes. But `BindingSource` was tagged a few words before, so I decided to not tag it again.
I saw that a few times in the docs, that APIs were tagged only on first occurrence, so I assumed that was best practice?

src/System.Windows.Forms/src/System/Windows/Forms/BindableComponent.cs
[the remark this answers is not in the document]
gewarren: I don't see anything about the standard text in the remarks either [here](https://github.com/dotnet/dotnet-api-docs/wiki/Summary%3A-Event-related) or in the Remarks section of the wiki. So I don't think you need to add it unless you want to.

src/System.Windows.Forms/src/System/Windows/Forms/ICommandBindingTargetProvider.cs
KlausLoeffelmann: Note, that this is all none-public.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
KlausLoeffelmann: Let's put that in the remarks then.

src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.cs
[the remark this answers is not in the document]
JeremyKuhne: @KlausLoeffelmann What else would you pass? The sender is this class. It would be odd to pass something different. I'd recommend reading the updated guidance on events in the latest Framework Design Guidelines to make sure we're in alignment.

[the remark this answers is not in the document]
KlausLoeffelmann: Of course, I know the guidance on events, but I was wondering if it is that simple here. 
In WinForms, we don't have the concept of [RoutedEvents](https://docs.microsoft.com/en-us/dotnet/desktop/wpf/events/routed-events-overview?view=netdesktop-6.0) which are bubbling up, like in WPF. And even if this doesn't apply here in the same way, technically, this Control is not the originator in this case, but a bound Data Source is, and the event is kind of passed on. It would be confusing, not have the Control as the sender all right, but at least I wanted to raise awareness of this.

If this becomes an issue, we can add the dedicated EventArgs with the original sender later. I don't think so it will though, since the originator will always be the assigned Data Source.

RussKie:
@KlausLoeffelmann I just come across this blog https://devblogs.microsoft.com/dotnet/announcing-the-dotnet-community-toolkit-800/. 
Do you think this work would make us compatible with [MVVM Toolkit](https://devblogs.microsoft.com/dotnet/announcing-the-dotnet-community-toolkit-800/#mvvm-toolkit)?

KlausLoeffelmann:
I don't see why not!
I will certainly take it into account, when we'll preparing the WinForms/MAUI example.

RussKie (dismissed):

src/System.Windows.Forms/src/System/Windows/Forms/BindableComponent.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// Gets or sets the <see cref="BindingContext"/> for this bindable <see cref="Component"/>.
+        /// </summary>
+        [Browsable(false)]
+        [EditorBrowsable(EditorBrowsableState.Advanced)]
+        [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
+        [SRCategory(nameof(SR.CatData))]
+        [SRDescription(nameof(SR.BindingComponentBindingContextDescr))]
+        public BindingContext? BindingContext
+        {
+            [RequiresPreviewFeatures]
```
dreddy-work, line 32: Q:  I see not all public properties has `RequiresPreviewFeatures` attribute. How are you deciding on this?

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -890,6 +893,57 @@ internal Gdi32.HBRUSH BackColorBrush
...
+                return ParentInternal?.DataContext;
+            }
+            set
+            {
+                if (Equals(value, DataContext))
+                {
+                    return;
+                }
+                
+                if (Properties.ContainsObject(s_dataContextProperty) && Equals(ParentInternal?.DataContext, value))
```
dreddy-work, line 929: Can you add a comment for this condition to make it clear? It seems you are also checking parent Internals context should be different from current context before setting the new value.

dreddy-work (dismissed):
RLGTM.

src/System.Windows.Forms/src/System/Windows/Forms/BindableComponent.cs
KlausLoeffelmann: Actually, the individual properties had been attributed before, and then we changed this to be on class level (line 14, which is including all the public members). So yes, these attributes are now redundant, good catch!

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
KlausLoeffelmann: Yes. If `DataContext` was different than its parent `DataContext`, and it's now about to become the same, we're removing the object altogether, since it's an ambient property and should inherit the (same) value from its parent. Added a comment.

JeremyKuhne (dismissed):

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -926,6 +926,8 @@ internal Gdi32.HBRUSH BackColorBrush
                     return;
                 }
                 
+                // When DataContext was different than its parent before, but now is is about to become the same,
```
dreddy-work:
```suggestion
                // When DataContext was different than its parent before, but now it is about to become the same,
```

dreddy-work (approved):

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
KlausLoeffelmann: Yeah, I saw that already... 😦 
Corrected!

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
```diff
@@ -2635,7 +2730,13 @@ protected virtual void OnBoundsChanged()
             InternalLayout.PerformLayout();
         }
 
-        protected virtual void OnClick(EventArgs e) => RaiseEvent(s_clickEvent, e);
+        protected virtual void OnClick(EventArgs e)
+        {
+            RaiseEvent(s_clickEvent, e);
+
+            // We won't let the preview feature warnings bubble further up beyond this point.
```
Tanya-Solyanik, line 2737: I don't really follow this comment. Do you mean that application that overrides this method will not  show the warning?

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
```diff
@@ -2677,6 +2778,50 @@ private void OnAnimationFrameChanged(object o, EventArgs e)
...
+        /// <summary>
+        ///  Raises the <see cref="ToolStripItem.CommandParameterChanged"/> event.
+        /// </summary>
+        /// <param name="e">An empty <see cref="EventArgs"/> instance.</param>
+        [RequiresPreviewFeatures]
+        [EditorBrowsable(EditorBrowsableState.Advanced)]
+        protected virtual void OnCommandParameterChanged(EventArgs e) => RaiseEvent(s_commandParameterChangedEvent, e);
+
+        /// <summary>
+        ///  Called in the context of <see cref="OnClick(EventArgs)"/> to invoke <see cref="System.Windows.Input.ICommand.Execute(object?)"/> if the context allows.
```
Tanya-Solyanik, line 2808: nit; this line is too long

src/System.Windows.Forms/src/Resources/SR.resx
```diff
@@ -6913,7 +6913,40 @@ Stack trace where the illegal operation occurred was:
   <data name="ExpandedStateName" xml:space="preserve">
     <value>Expanded</value>
   </data>
+  <data name="ToolStripItemOnCommandChangedDescr" xml:space="preserve">
+    <value>Occurs whenever the bindable command changes.</value>
```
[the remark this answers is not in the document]
RussKie: This wording is common for event handlers, and it's used in other parts of the codebase.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
KlausLoeffelmann: Corrected.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
KlausLoeffelmann: No, but at one point we needed to protect the WinForms runtime against the Preview warnings themselves. It's gone now.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripSeparator.cs
```diff
@@ -304,7 +304,9 @@ protected override void OnPaint(PaintEventArgs e)
         protected override void OnFontChanged(EventArgs e)
         {
             // Perf: don't call base, we don't care if the font changes
+#pragma warning disable CA2252 // Suppress 'Opt in to preview features' (https://aka.ms/dotnet-warnings/preview-features)
```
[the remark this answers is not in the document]
KlausLoeffelmann: Cut off the dependencies, when the consumer does not use the Preview feature directly. If we didn't do this, also using `OnFontChange` would have been considered a Preview feature, and so on, and so on.

PR #7144, opened 2022-05-06

gpetrou:
## Proposed changes

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7144)

### DIFF
--- a/src/System.Windows.Forms/src/Resources/SR.resx
+++ b/src/System.Windows.Forms/src/Resources/SR.resx
@@ -5194,8 +5194,8 @@ Stack trace where the illegal operation occurred was:
   <data name="ScrollableControlHorizontalScrollDescr" xml:space="preserve">
     <value>Gets the horizontal scroll bar for this ScrollableControl.</value>
   </data>
-  <data name="ScrollableControlParentInternalNull" xml:space="preserve">
-    <value>ParentInternal cannot be null.</value>
+  <data name="ScrollableControlActiveControlParentNull" xml:space="preserve">
+    <value>Parent property of the active control cannot be null.</value>
   </data>
   <data name="ScrollableControlRaiseMouseEnterLeaveEventsForScrollBarsDescr" xml:space="preserve">
     <value>Property being set to false will not raise MouseEnter and MouseLeave events until the mouse leaves the client area and the scrollbars.</value>
--- a/src/System.Windows.Forms/src/Resources/SR.resx
+++ b/src/System.Windows.Forms/src/Resources/SR.resx
@@ -5194,6 +5194,9 @@ Stack trace where the illegal operation occurred was:
   <data name="ScrollableControlHorizontalScrollDescr" xml:space="preserve">
     <value>Gets the horizontal scroll bar for this ScrollableControl.</value>
   </data>
+  <data name="ScrollableControlParentInternalNull" xml:space="preserve">
+    <value>ParentInternal cannot be null.</value>
+  </data>
   <data name="ScrollableControlRaiseMouseEnterLeaveEventsForScrollBarsDescr" xml:space="preserve">
     <value>Property being set to false will not raise MouseEnter and MouseLeave events until the mouse leaves the client area and the scrollbars.</value>
   </data>
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
@@ -3081,7 +3081,7 @@ namespace System.Windows.Forms
         {
         }
 
-        internal virtual void RemoveToolTip(ToolTip? toolTip)
+        internal virtual void RemoveToolTip(ToolTip toolTip)
         {
             // Control doesn't have a specific logic after a toolTip is removed
         }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
@@ -907,7 +907,7 @@ namespace System.Windows.Forms
                     return c;
                 }
 
-                Control p = ParentInternal;
+                Control? p = ParentInternal;
                 if (p is not null && p.CanAccessProperties)
                 {
                     c = p.BackColor;
@@ -1099,7 +1099,7 @@ namespace System.Windows.Forms
                 }
 
                 // Otherwise, see if the parent has one for us.
-                Control p = ParentInternal;
+                Control? p = ParentInternal;
                 if (p is not null && p.CanAccessProperties)
                 {
                     return p.BindingContext;
@@ -1636,7 +1636,7 @@ namespace System.Windows.Forms
                 }
                 else
                 {
-                    Control parent = ParentInternal;
+                    Control? parent = ParentInternal;
                     if (parent is not null)
                     {
                         return parent.ValidationCancelled;
@@ -1713,7 +1713,7 @@ namespace System.Windows.Forms
                     return localDefault;
                 }
 
-                Control p = ParentInternal;
+                Control? p = ParentInternal;
                 if (p is not null)
                 {
                     return p.Cursor;
@@ -2380,7 +2380,7 @@ namespace System.Windows.Forms
                 return null;
             }
         }
-#nullable disable
+
         [EditorBrowsable(EditorBrowsableState.Advanced)]
         public virtual Size GetPreferredSize(Size proposedSize)
         {
@@ -2482,7 +2482,7 @@ namespace System.Windows.Forms
         {
             get
             {
-                ControlCollection controls = (ControlCollection)Properties.GetObject(s_controlsCollectionProperty);
+                ControlCollection? controls = (ControlCollection?)Properties.GetObject(s_controlsCollectionProperty);
                 return controls is not null && controls.Count > 0;
             }
         }
@@ -2578,7 +2578,7 @@ namespace System.Windows.Forms
 
                 RECT temp = new RECT();
                 Region working;
-                Control parent = ParentInternal;
+                Control? parent = ParentInternal;
                 if (parent is not null)
                 {
                     while (parent.ParentInternal is not null)
@@ -2715,7 +2715,7 @@ namespace System.Windows.Forms
         public bool IsAncestorSiteInDesignMode =>
             GetSitedParentSite(this) is ISite parentSite ? parentSite.DesignMode : false;
 
-        private ISite GetSitedParentSite(Control control)
+        private ISite? GetSitedParentSite(Control control)
         {
             ArgumentNullException.ThrowIfNull(control);
             return (control.Site is not null && control.Site.DesignMode) || control.Parent is null ?
@@ -2804,7 +2804,7 @@ namespace System.Windows.Forms
 
         [SRCategory(nameof(SR.CatPropertyChanged))]
         [SRDescription(nameof(SR.ControlOnLocationChangedDescr))]
-        public event EventHandler LocationChanged
+        public event EventHandler? LocationChanged
         {
             add => Events.AddHandler(s_locationEvent, value);
             remove => Events.RemoveHandler(s_locationEvent, value);
@@ -2835,7 +2835,7 @@ namespace System.Windows.Forms
 
         [SRCategory(nameof(SR.CatLayout))]
         [SRDescription(nameof(SR.ControlOnMarginChangedDescr))]
-        public event EventHandler MarginChanged
+        public event EventHandler? MarginChanged
         {
             add => Events.AddHandler(s_marginChangedEvent, value);
             remove => Events.RemoveHandler(s_marginChangedEvent, value);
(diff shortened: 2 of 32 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/SplitContainer.cs
```diff
@@ -1555,13 +1555,13 @@ private int GetSplitterDistance(int x, int y)
         /// </summary>
         private bool ProcessArrowKey(bool forward)
         {
-            Control group = this;
+            Control? group = this;
```
RussKie, line 1558: How can this be nullable?

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.cs
```diff
@@ -621,7 +622,7 @@ protected override void OnPaintBackground(PaintEventArgs e)
             }
         }
 
-        internal override void RemoveToolTip(ToolTip toolTip)
+        internal override void RemoveToolTip(ToolTip? toolTip)
```
RussKie, line 625: There's only one call site (ignoring the tests), and that invocation passes `this`, so it can't be nullable. We can remove the check below and update the tests.

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -3027,7 +3028,7 @@ internal virtual Control ParentInternal
                     }
                     else
                     {
-                        _parent.Controls.Remove(this);
+                        _parent!.Controls.Remove(this);
```
RussKie, line 3031: Since `_parent` can be null, I don't think this is safe... 🤔 

I think we should update as follows:
```diff
diff --git a/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs b/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
index c9119e0ca..076db89ea 100644
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
@@ -3025,7 +3025,7 @@ namespace System.Windows.Forms
                     {
                         value.Controls.Add(this);
                     }
-                    else
+                    else if (_parent is not null)
                     {
                         _parent.Controls.Remove(this);
                     }
```

May worth reducing the nesting too, while we're at it:
```diff
diff --git a/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs b/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
index c9119e0ca..5a37727dc 100644
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
@@ -3019,19 +3019,21 @@ namespace System.Windows.Forms
             get => _parent;
             set
             {
-                if (_parent != value)
+                if (_parent == value)
                 {
+                    return;
+                }
+
                 if (value is not null)
                 {
                     value.Controls.Add(this);
                 }
-                    else
+                else if (_parent is not null)
                 {
                     _parent.Controls.Remove(this);
                 }
             }
         }
-        }
 ```

File: src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AxHostTests.cs [hunk 4]
```diff
@@ -2233,8 +2243,8 @@ public void AxHost_InvokeEditMode_InvokeWithSiteDesignModeWithComponentSelectedV
             Assert.Equal(0, invalidatedCallCount);
             Assert.Equal(0, styleChangedCallCount);
             Assert.Equal(0, createdCallCount);
-            mockSite.Verify(s => s.GetService(typeof(ISelectionService)), Times.Exactly(2));
-            mockSelectionService.Verify(s => s.GetComponentSelected(control), Times.Once());
+            mockSite.Verify(s => s.GetService(typeof(ISelectionService)), Times.Exactly(3));
+            mockSelectionService.Verify(s => s.GetComponentSelected(control), Times.Exactly(2));
```
RussKie, line 2247: Can you please help me understand what change led to this? Thank you.

RussKie (commented):
👍 with few comments

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
gpetrou: We only enter the statement if `_parent` and `value` are different. We enter the `else` statement if `value` is null, which means that the `_parent` is not null. Do you still want me to add the extra check?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AxHostTests.cs
gpetrou: If you try to debug these tests in current main, you will see that there are some NullReference exceptions that eventually get ignored. The problematic code seems to be `lpFrameInfo->hwndFrame = _host.ParentInternal.Handle;`. If we modify it to be like the code we have in WebBrowserSiteBase, we won't get these exceptions anymore, but we have to modify these values, so that the tests complete successfully.

gpetrou:
@RussKie any more feedback on this?

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
RussKie: I'd add an extra check just for clarity.

RussKie (dismissed):

src/System.Windows.Forms/src/System/Windows/Forms/ScrollableControl.cs
```diff
@@ -885,7 +885,7 @@ protected virtual Point ScrollToControl(Control activeControl)
             if (activeControl.ParentInternal != this)
             {
                 Debug.WriteLineIf(s_autoScrolling!.TraceVerbose, "not direct child, original bounds: " + bounds);
-                bounds = RectangleToClient(activeControl.ParentInternal.RectangleToScreen(bounds));
+                bounds = RectangleToClient(activeControl.ParentInternal?.RectangleToScreen(bounds) ?? Rectangle.Empty);
```
dreddy-work, line 888: this change doesn't seem right. If you assign empty rectangle, rest of the code doesn't make sense. Can you explain why this change?

src/System.Windows.Forms/src/System/Windows/Forms/SplitContainer.cs
```diff
@@ -1555,13 +1555,13 @@ private int GetSplitterDistance(int x, int y)
...
             if (ActiveControl is not null)
             {
                 group = ActiveControl.ParentInternal;
             }
 
-            return group.SelectNextControl(ActiveControl, forward, false, false, true);
+            return group?.SelectNextControl(ActiveControl, forward, false, false, true) ?? false;
```
dreddy-work, line 1564: it seems it was never null here before otherwise throws exception. Did you get chance to run the scenario here?

RussKie: `group` can be null, and we need to return a boolean.

src/System.Windows.Forms/src/System/Windows/Forms/ScrollableControl.cs
gpetrou: `ParentInternal` is nullable. You want me to return `Point.Empty` instead if it is null?

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
gpetrou: I added the extra check.

src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.UpDownButtons.UpDownButtonsAccessibleObject.DirectionButtonAccessibleObject.cs
```diff
@@ -44,7 +44,7 @@ public override Rectangle Bounds
                             }
 
                             // Convert to screen coords
-                            return (((UpDownButtons)_parent.Owner).ParentInternal).RectangleToScreen(bounds);
+                            return ((UpDownButtons?)_parent.Owner)?.ParentInternal?.RectangleToScreen(bounds) ?? Rectangle.Empty;
```
RussKie, line 47: `_parent.Owner` can't be null here, we'd have failed on line:32 with a NRE.
```suggestion
                            return ((UpDownButtons)_parent.Owner).ParentInternal?.RectangleToScreen(bounds) ?? Rectangle.Empty;
```

src/System.Windows.Forms/src/System/Windows/Forms/ScrollableControl.cs
RussKie: Should we check `activeControl.ParentInternal is null` at the beginning and bail out, if so?

gpetrou: I added a null check.

src/System.Windows.Forms/src/System/Windows/Forms/SplitContainer.cs
dreddy-work: I see that. Earlier it would have been throwing null reference if it was null.  But i don't think it ever did. So wanted to find scenario experience here If checking for null and throwing exception still make sense?

src/System.Windows.Forms/src/System/Windows/Forms/ScrollableControl.cs
```diff
@@ -885,6 +885,12 @@ protected virtual Point ScrollToControl(Control activeControl)
             if (activeControl.ParentInternal != this)
             {
                 Debug.WriteLineIf(s_autoScrolling!.TraceVerbose, "not direct child, original bounds: " + bounds);
+
+                if (activeControl.ParentInternal is null)
+                {
+                    return Point.Empty;
```
dreddy-work, line 891: Same comment here. I am assuming, returning empty will place scroll bar back at (0,0) position. By understanding scenario here, ParentInternal may be null if tab goes out of scrollable control container. And in such case, scrollbars shouldn't reset to (0,0) position. @Tanya-Solyanik any thoughts here?

Tanya-Solyanik: Winforms invokes this method from a single place - https://github.com/dotnet/winforms/blob/0b2ee90e0c1c08458fdebaf72c81f0833e90b8f1/src/System.Windows.Forms/src/System/Windows/Forms/ScrollableControl.cs#L841 where we are guaranteed that we don't enter this method when control in unparented because we test that control is a descendant. If user code called `ScrollToControl` internally, they were getting exception in the past, so we could throw a more descriptive exception now. So I would rather do `InvalidOperationException`.

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.cs
```diff
@@ -497,7 +497,8 @@ internal override void AssignParent(Control value)
         /// </summary>
         public static TabPage? GetTabPageOfComponent(object? comp)
         {
-            if (comp is not Control c)
+            Control? c = comp as Control;
+            if (c is null)
```
RussKie, line 501: Why change this?

src/System.Windows.Forms/src/Resources/SR.resx
```diff
@@ -5194,6 +5194,9 @@ Stack trace where the illegal operation occurred was:
   <data name="ScrollableControlHorizontalScrollDescr" xml:space="preserve">
     <value>Gets the horizontal scroll bar for this ScrollableControl.</value>
   </data>
+  <data name="ScrollableControlParentInternalNull" xml:space="preserve">
+    <value>ParentInternal cannot be null.</value>
```
RussKie, line 5198: `ParentInternal` is an implementation detail, and it's not part of the public API surface.
```suggestion
    <value>Parent property of the active control cannot be null.</value>
```

similar to `DataGridViewCell_ValueTypeNull` and `DataGridViewCell_FormattedValueTypeNull`.

src/System.Windows.Forms/src/Resources/SR.resx
```diff
@@ -5194,6 +5194,9 @@ Stack trace where the illegal operation occurred was:
...
+  </data>
```
RussKie, line 5199: `ParentInternal` is an implementation detail, and it's not part of the public API surface.

```suggestion
  <data name="ScrollableControlActiveControlParentNull" xml:space="preserve">
    <value>Parent property of the active control cannot be null.</value>
  </data>
```
similar to `DataGridViewCell_ValueTypeNull` and `DataGridViewCell_FormattedValueTypeNull`.

RussKie (commented):
👍

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.cs
gpetrou: We need `c` to be nullable since we use `c.ParentInternal` below.

RussKie (approved):

RussKie:
Thank you

PR #7146, opened 2022-05-06
Fix disabling a focused TextBox draws TextBox wrong

### LINKED ISSUE #4856: Disabling a focused TextBox draws TextBox wrong
**.NET Core Version:**

SDK: 5.0.202 (db7cc87d51)
Runtime: 5.0.5 (2f740adc14)

**Problem description:**

If a focused textbox is disabled while having the focus, it does not display correctly:

Here are 2 textboxes where the top one has focus:
image

The top one is now disabled, notice the weird looking border, it should look like the bottom one:
image

**Repro:**

I have attached a sample project showing the problem. There are 2 textboxes and a button. The button disables both textboxes. The textbox with the focus shows the problem:
[TestDisableTextBox.zip](https://github.com/dotnet/winforms/files/6397707/TestDisableTextBox.zip)

Balkoth:
Fixes #4856

## Proposed changes

- override OnEnabledChanged() and force RecreateHandle

## Customer Impact

- RecreateHandle might be unexpected

## Regression? 

- No

## Risk

- Low

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7146)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TextBox.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TextBox.cs
@@ -648,10 +648,7 @@ namespace System.Windows.Forms
             // There is a bug in the underlying Win32 control which prevents it from correctly redrawing its borders when switching between enabled and disabled state.
             // The border color is never updated when switching state, so the control will always paint using the color set when the handle was created (either enabled or disabled).
             // The only way to force the control to update its border color is to force a handle creation.
-            if (IsHandleCreated && Application.RenderWithVisualStyles)
-            {
-                RecreateHandle();
-            }
+            RecreateHandle();
         }
 
         protected override void OnKeyUp(KeyEventArgs e)
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TextBox.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TextBox.cs
@@ -641,6 +641,19 @@ namespace System.Windows.Forms
             base.OnHandleDestroyed(e);
         }
 
+        protected override void OnEnabledChanged(EventArgs e)
+        {
+            base.OnEnabledChanged(e);
+
+            // There is a bug in the underlying Win32 control which prevents it from correctly redrawing its borders when switching between enabled and disabled state.
+            // The border color is never updated when switching state, so the control will always paint using the color set when the handle was created (either enabled or disabled).
+            // The only way to force the control to update its border color is to force a handle creation.
+            if (IsHandleCreated && Application.RenderWithVisualStyles)
+            {
+                RecreateHandle();
+            }
+        }
+
         protected override void OnKeyUp(KeyEventArgs e)
         {
             base.OnKeyUp(e);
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/TextBox.cs
```diff
@@ -641,6 +641,19 @@ protected override void OnHandleDestroyed(EventArgs e)
...
+        protected override void OnEnabledChanged(EventArgs e)
+        {
+            base.OnEnabledChanged(e);
+
+            // There is a bug in the underlying Win32 control which prevents it from correctly redrawing its borders when switching between enabled and disabled state.
+            // The border color is never updated when switching state, so the control will always paint using the color set when the handle was created (either enabled or disabled).
+            // The only way to force the control to update its border color is to force a handle creation.
+            if (IsHandleCreated && Application.RenderWithVisualStyles)
+            {
+                RecreateHandle();
```
dreddy-work, line 653: Recreating handle does lot more than just creating a new handle and bunch of code around those events(In PermonV2 apps, even more). Did you explore alternatives like redrawing the boarder to match expected behavior?

Balkoth: I did it this way, because this was suggested in the linked issue. I am all open for better approaches to fix this problem in winforms.

As i understand the TextBox control is a wrapper around the common control EDIT. How dou you get that to redraw it's border?

dreddy-work:
:/cc @Tanya-Solyanik / @JeremyKuhne .

src/System.Windows.Forms/src/System/Windows/Forms/TextBox.cs
RussKie: @dreddy-work we call this API in many places across the codebase.

@Balkoth `IsHandleCreated` isn't strictly needed, `RecreateHandle` performs this check within. We don't appear to check for `Application.RenderWithVisualStyles` anywhere else in the code when recreating the handle, so for consistency sake I would remove this too.

Balkoth:
@RussKie Updated with your suggestions.
As you mention in the linked issue, the same problem exists with MaskedTextBox. Should i submit a commit here with the same fix or create a new PR?

src/System.Windows.Forms/src/System/Windows/Forms/TextBox.cs
dreddy-work: > @dreddy-work we call this API in many places across the codebase.

Sure. But that doesn't invalidate the amount of code we would be calling. Especially for an issue that has been there since Framework and coul repro in Win32 aswell. If windows fix this (for any reason), we wouldn't be tracking that and revisiting and removing this fix.

dreddy-work:
@Balkoth, @Tanya-Solyanik suggested following work around at the application level. Can you give it a try?
    The root cause looks like a race between the native control updating in response to the focus loss and managed code setting disabled state([here](https://github.com/dotnet/winforms/blob/9ea06044776cd10bd1968ee141449bed1a8e6945/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs#L10692). Workaround on the app level is simple:

       private void button1_Click(object sender, EventArgs e)
        {​​​​​​
            BeginInvoke(new Action(() =>
            {​​​​​​
                textBox1.Enabled = !textBox1.Enabled;
            }​​​​​​));
        }​​​​​​

Given this is not a regression and work around is simple and Windows `may` fix it in later versions from which we can benefit, I am hesitant to take this change.

Balkoth:
Imho and as @RussKie pointed out, this will never get fixed by the Windows team. This has been a problem since Vista. Enabling and disabling controls is not something that is happening so often that `RecreateHandle` should be a performance issue. On the other hand fixing this in the framework benefits anyone using it.

dreddy-work:
`RecreateHandle` executes bunch of code and trigger events that were not there before. In the past, we have seen regression issues due to this. Given the work around is simple at the application level, scenario is very specific (disabling control without moving focus) and is not a regression, team decided not to pursue this.

Thank you for understanding @Balkoth.

PR #6976, opened 2022-04-06
Provide CCW for IDataObject

kant2002:
Progress towards #5163

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6976)

PR #6576, opened 2022-01-29
Add support for drag images and drop descriptions

willibrandon:
Allow drop targets to display the drag image while the cursor is over the target and add support for setting drop descriptions.

Allow drag sources to specify the drag image bitmap.

A possible implementation for #5884.

## Proposed changes

- Maintain the current default `DropTarget` behavior. By default, do not display the drag image and drop description.

- Add a new public enumeration; `DropImageType`.

```csharp
namespace System.Windows.Forms
{
    /// <summary>
    /// Specifies the drop description image type.
    /// </summary>
    public enum DropImageType
    {
        /// <summary>
        /// No drop image preference; use the default image.
        /// </summary>
        Invalid = -1,

        /// <summary>
        /// A red bisected circle such as that found on a "no smoking" sign.
        /// </summary>
        None = 0,

        /// <summary>
        /// A plus sign (+) that indicates a copy operation.
        /// </summary>
        Copy = DragDropEffects.Copy,

        /// <summary>
        /// An arrow that indicates a move operation.
        /// </summary>
        Move = DragDropEffects.Move,

        /// <summary>
        /// An arrow that indicates a link.
        /// </summary>
        Link = DragDropEffects.Link,

        /// <summary>
        /// A tag icon that indicates that the metadata will be changed.
        /// </summary>
        Label = 6,

        /// <summary>
        /// A yellow exclamation mark that indicates that a problem has been encountered in the operation.
        /// </summary>
        Warning = 7,

        /// <summary>
        /// Windows 7 and later. Use no drop image.
        /// </summary>
        NoImage = 8
    }
}
```

- Add 3 new public `DragEventArgs` properties; `DropImageType`, `Message`, and `MessageReplacementToken`.

```csharp
namespace System.Windows.Forms
{
  public class DragEventArgs : EventArgs
  {
      /// <summary>
      /// Gets or sets the drop description image type.
      /// </summary>
      public DropImageType DropImageType { get; set; }
  
        /// <summary>
        /// Gets or sets the drop description text such as "Move to %1" when "Documents" is specified in <see cref="MessageReplacementToken"/>.
        /// </summary>
        /// <remarks>
        /// <para>
        /// UI coloring is applied to the text in <see cref="MessageReplacementToken"/> if used by specifying %1 in <see cref="Message"/>.
        /// </para>
        /// </remarks>
        public string? Message { get; set; }
  
        /// <summary>
        /// Gets or sets the drop description text such as "Documents" when %1 is specified in <see cref="Message"/>.
[...]

PR #7162, opened 2022-05-10
Add IEnumFormatEtc ComWrapper

kant2002:
This is extract of https://github.com/dotnet/winforms/pull/7087 which should help fix https://github.com/dotnet/winforms/pull/6976#discussion_r867278650 and unblock https://github.com/dotnet/winforms/pull/6576

I expect #7087 would take a bit of time if Win 7 support would be required, so I think if fast-track only this part it would bring main in more healthy state.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7162)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.EnumFORMATETCWrapper.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.EnumFORMATETCWrapper.cs
@@ -27,7 +27,7 @@ internal partial class Interop
                 _wrappedInstance = IntPtr.Zero;
             }
 
-            public int Next(int celt, FORMATETC[] rgelt, int[] pceltFetched)
+            public int Next(int celt, FORMATETC[] rgelt, int[]? pceltFetched)
             {
                 fixed (FORMATETC* rgeltPtr = &MemoryMarshal.GetArrayDataReference(rgelt))
                 {
--- a/src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.EnumFORMATETCWrapper.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.EnumFORMATETCWrapper.cs
@@ -1,0 +1,76 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.InteropServices;
+using System.Runtime.InteropServices.ComTypes;
+using System.Windows.Forms;
+
+internal partial class Interop
+{
+    internal unsafe partial class WinFormsComWrappers
+    {
+        internal class EnumFORMATETCWrapper : IEnumFORMATETC
+        {
+            private IntPtr _wrappedInstance;
+
+            public EnumFORMATETCWrapper(IntPtr wrappedInstance)
+            {
+                _wrappedInstance = wrappedInstance.OrThrowIfZero();
+            }
+
+            internal IntPtr Instance => _wrappedInstance;
+
+            ~EnumFORMATETCWrapper()
+            {
+                this.DisposeInternal();
+            }
+
+            private void DisposeInternal()
+            {
+                Marshal.Release(_wrappedInstance);
+                _wrappedInstance = IntPtr.Zero;
+            }
+
+            public int Next(int celt, FORMATETC[] rgelt, int[] pceltFetched)
+            {
+                fixed (FORMATETC* rgeltPtr = &MemoryMarshal.GetArrayDataReference(rgelt))
+                {
+                    if (pceltFetched == null)
+                    {
+                        return ((delegate* unmanaged<IntPtr, int, FORMATETC*, int*, int>)(*(*(void***)_wrappedInstance + 3)))
+                            (_wrappedInstance, celt, rgeltPtr, null);
+                    }
+                    else
+                    {
+                        fixed (int* pceltFetchedPtr = &MemoryMarshal.GetArrayDataReference(pceltFetched))
+                        {
+                            return ((delegate* unmanaged<IntPtr, int, FORMATETC*, int*, int>)(*(*(void***)_wrappedInstance + 3)))
+                                (_wrappedInstance, celt, rgeltPtr, pceltFetchedPtr);
+                        }
+                    }
+                }
+            }
+
+            public int Skip(int celt)
+            {
+                return ((delegate* unmanaged<IntPtr, int, int>)(*(*(void***)_wrappedInstance + 4)))
+                    (_wrappedInstance, celt);
+            }
+
+            public int Reset()
+            {
+                return ((delegate* unmanaged<IntPtr, int>)(*(*(void***)_wrappedInstance + 5)))
+                    (_wrappedInstance);
+            }
+
+            public void Clone(out IEnumFORMATETC newEnum)
+            {
+                IntPtr resultPtr;
+                ((delegate* unmanaged<IntPtr, IntPtr*, HRESULT>)(*(*(void***)_wrappedInstance + 6)))
+                    (_wrappedInstance, &resultPtr).ThrowIfFailed();
+                newEnum = (IEnumFORMATETC)WinFormsComWrappers.Instance.GetOrCreateObjectForComInstance(resultPtr, CreateObjectFlags.Unwrap);
+            }
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.EnumFORMATETCWrapper.cs
```diff
@@ -0,0 +1,76 @@
...
+            ~EnumFORMATETCWrapper()
+            {
+                this.DisposeInternal();
+            }
+
+            private void DisposeInternal()
+            {
+                Marshal.Release(_wrappedInstance);
+                _wrappedInstance = IntPtr.Zero;
+            }
```
RussKie, line 33:
```suggestion
            ~EnumFORMATETCWrapper()
            {
                Marshal.Release(_wrappedInstance);
                _wrappedInstance = IntPtr.Zero;
            }
```

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.EnumFORMATETCWrapper.cs
```diff
@@ -0,0 +1,76 @@
...
+
+            public int Next(int celt, FORMATETC[] rgelt, int[] pceltFetched)
+            {
+                fixed (FORMATETC* rgeltPtr = &MemoryMarshal.GetArrayDataReference(rgelt))
+                {
+                    if (pceltFetched == null)
```
RussKie, line 39:
```suggestion
                    if (pceltFetched is null)
```

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.IDataObjectVtbl.cs
```diff
@@ -177,9 +177,9 @@ private static unsafe HRESULT EnumDAdvise(IntPtr thisPtr, IntPtr* pEnumAdvise)
 
             internal struct STGMEDIUM_Raw
             {
-                public IntPtr pUnkForRelease;
                 public TYMED tymed;
                 public IntPtr unionmember;
+                public IntPtr pUnkForRelease;
```
RussKie, line 182: 👍

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.EnumFORMATETCWrapper.cs
```diff
@@ -0,0 +1,76 @@
...
+                this.DisposeInternal();
+            }
+
+            private void DisposeInternal()
+            {
+                Marshal.Release(_wrappedInstance);
+                _wrappedInstance = IntPtr.Zero;
+            }
+
+            public int Next(int celt, FORMATETC[] rgelt, int[] pceltFetched)
```
RussKie, line 35:
```suggestion
            public int Next(int celt, FORMATETC[] rgelt, int[]? pceltFetched)
```

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.IEnumFORMATETCVtbl.cs
```diff
@@ -0,0 +1,102 @@
...
+
+            [UnmanagedCallersOnly]
+            private static int Next(IntPtr thisPtr, int celt, FORMATETC* rgelt, int* pceltFetched)
+            {
+                try
+                {
+                    IEnumFORMATETC instance = ComInterfaceDispatch.GetInstance<IEnumFORMATETC>((ComInterfaceDispatch*)thisPtr);
+                    FORMATETC[] elt = new FORMATETC[celt];
+                    int[] celtFetched = new int[1];
+                    var result = instance.Next(celt, elt, pceltFetched == null ? null! : celtFetched);
```
RussKie, line 37: Would this work?
```suggestion
                    var result = instance.Next(celt, elt, pceltFetched ?? celtFetched);
```

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.IEnumFORMATETCVtbl.cs
```diff
@@ -0,0 +1,102 @@
...
+                    for (var i = 0; i < celt; i++)
+                    {
+                        rgelt[i] = elt[i];
+                    }
+
+                    if (pceltFetched != null)
```
RussKie, line 43:
```suggestion
                    if (pceltFetched is not null)
```

RussKie (commented):
👍

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.EnumFORMATETCWrapper.cs
kant2002: I cannot teach myself on that style ☹️

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.IEnumFORMATETCVtbl.cs
kant2002: This one I believe you got wrong.

kant2002: Obviously it would work, but it seems we change behavior of how `IEnumFORMATETC.Next` should work.

RussKie: Why?

RussKie: Actually, I think I did get it wrong. :)
It should be `var result = instance.Next(celt, elt, pceltFetched is null ? null : celtFetched);`

kant2002: That way I always pass non-null `pceltFetched` value which feels wrong because it is slightly different then passing NULL, but I would not fight over that specific location, since it's really works as expected for outside callers.

kant2002: come on. Why again `pceltFetched is null` 😄

kant2002:
@RussKie can you take a look?

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.IEnumFORMATETCVtbl.cs
```diff
@@ -0,0 +1,102 @@
...
+
+            [UnmanagedCallersOnly]
+            private static int Next(IntPtr thisPtr, int celt, FORMATETC* rgelt, int* pceltFetched)
+            {
+                try
+                {
+                    IEnumFORMATETC instance = ComInterfaceDispatch.GetInstance<IEnumFORMATETC>((ComInterfaceDispatch*)thisPtr);
+                    FORMATETC[] elt = new FORMATETC[celt];
+                    int[] celtFetched = new int[1];
+                    var result = instance.Next(celt, elt, pceltFetched is null ? null! : celtFetched);
```
RussKie, line 37: The `!` operator is not needed here. If it's still needed then `Next` isn't annotated correctly.

RussKie (commented):
Overall LGTM

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.IEnumFORMATETCVtbl.cs
kant2002: https://github.com/dotnet/runtime/pull/68537

Aaron ask me to correct all annotations or do not touch them.

kant2002: And yes. You are right, IEnumFORMATETC annotated incorrectly.

RussKie: Ah! Thank you.
Can you please add a TODO comment to the code referencing the runtime issue?

kant2002: Done.

RussKie (approved):

RussKie:
Thank you

kant2002:
@willibrandon since this is lands main, you may want merge with main it again and check that you work no longer affected.

RussKie:
I recommend to always rebase instead of doing tracking merges, as it provides a significantly cleaner view of actual changes (especiallyfor reviews outside GitHub), and allows for a better devex (e.g., fixup, selective squash, etc.).

PR #6370, opened 2021-12-20
Targeting TFM to 7.0

dreddy-work:
contributes to #6323

Upgrading TFM for WinForms main branch to 7.0.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/6370)

PR #1612, opened 2019-08-13
Fix tooltip crash and harden

JeremyKuhne:
The tooltip class will cause Windows to AV when attempting to get the tooltip bubble size when there isn't an actively showing tooltip. (Really Windows should be checking before derefing their internal pointer, but this has been there forever.) Fixing this by making sure to check that one is active when sending the query message.

The existing TooInfo usage was not safe and could crash with bad timing with a GC. I've hardened usage of the TTTOOLINFO struct and made it blittable, which should make things both safe and more performant as we're not instantiating a class anymore.

Fixes #1602.

## Proposed changes

- Refactor TTOOLINFO to guard against AVs
- Fix the explicit AV in #1602

## Customer Impact

- Won't easily AV
- Better performance

## Regression? 

- Yes - Introduced in 4.8

## Risk

- Reasonably low- it is much easier to reason about the code with this change

## Test methodology

- Manual testing with Tooltip
- Debugging into Windows to validate behavior

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/winforms/pull/1612)

PR #7172, opened 2022-05-13
Use LibraryImport in some Interop User32

gpetrou:
Now that https://github.com/dotnet/runtime/issues/60595 is closed, perhaps we could start using LibraryImport instead of DllImport.

## Proposed changes

- Use LibraryImport in some Interop User32.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7172)

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "tools": {
-    "dotnet": "7.0.100-preview.3.22154.3",
+    "dotnet": "7.0.100-preview.5.22228.6",
     "runtimes": {
       "dotnet/x64": [
         "$(VSRedistCommonNetCoreSharedFrameworkx6470PackageVersion)"
@@ -11,7 +11,7 @@
     }
   },
   "sdk": {
-    "version": "7.0.100-preview.3.22154.3"
+    "version": "7.0.100-preview.5.22228.6"
   },
   "msbuild-sdks": {
     "Microsoft.DotNet.Arcade.Sdk": "7.0.0-beta.22301.2",
--- a/src/System.Windows.Forms.Primitives/src/Interop/User32/Interop.CopyImage.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/User32/Interop.CopyImage.cs
@@ -8,8 +8,8 @@ internal static partial class Interop
 {
     internal static partial class User32
     {
-        [DllImport(Libraries.User32, ExactSpelling = true)]
-        public static extern IntPtr CopyImage(IntPtr hImage, IMAGE type, int cx, int cy, LR flags);
+        [LibraryImport(Libraries.User32)]
+        public static partial IntPtr CopyImage(IntPtr hImage, IMAGE type, int cx, int cy, LR flags);
 
         public static IntPtr CopyImage(IHandle hImage, IMAGE type, int cx, int cy, LR flags)
         {
--- a/src/System.Windows.Forms.Primitives/src/Interop/User32/Interop.ActivateKeyboardLayout.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/User32/Interop.ActivateKeyboardLayout.cs
@@ -8,7 +8,7 @@ internal static partial class Interop
 {
     internal static partial class User32
     {
-        [DllImport(Libraries.User32, ExactSpelling = true)]
-        public static extern IntPtr ActivateKeyboardLayout(IntPtr hkl, uint Flags);
+        [LibraryImport(Libraries.User32)]
+        public static partial IntPtr ActivateKeyboardLayout(IntPtr hkl, uint Flags);
     }
 }
--- a/src/System.Windows.Forms.Primitives/src/Interop/User32/Interop.AreDpiAwarenessContextsEqual.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/User32/Interop.AreDpiAwarenessContextsEqual.cs
@@ -9,8 +9,8 @@ internal static partial class Interop
 {
     internal static partial class User32
     {
-        [DllImport(Libraries.User32, ExactSpelling = true, EntryPoint = "AreDpiAwarenessContextsEqual", SetLastError = true)]
-        private static extern BOOL AreDpiAwarenessContextsEqualInternal(IntPtr dpiContextA, IntPtr dpiContextB);
+        [LibraryImport(Libraries.User32, EntryPoint = "AreDpiAwarenessContextsEqual", SetLastError = true)]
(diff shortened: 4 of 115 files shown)


global.json
```diff
@@ -1,6 +1,6 @@
 {
   "tools": {
-    "dotnet": "7.0.100-preview.3.22154.3",
```
dreddy-work, line 3: these upgrades normally come throw a dependency flow. If your changes need this updated runtime, it may better to wait couple more days /week here.

gpetrou: We need a newer preview version in order to use the `LibraryImport`. As far as I understand, these lines don't change that often. I see that they were last changed in https://github.com/dotnet/winforms/pull/6767 by the bot and before that by you in https://github.com/dotnet/winforms/pull/6370. Only two times this year. So, perhaps it is OK to change this?

gpetrou: We need a newer preview version in order to use the `LibraryImport`. As far as I understand, these lines don't change that often. I see that they were last changed in https://github.com/dotnet/winforms/pull/6767 by the bot and before that by you in https://github.com/dotnet/winforms/pull/6370. Only two times this year. So, perhaps it is OK to change this?

RussKie: @gpetrou you're absolutely correct, these don't get updated automatically.

src/System.Windows.Forms.Primitives/src/Interop/User32/Interop.CopyImage.cs
```diff
@@ -8,8 +8,8 @@ internal static partial class Interop
 {
     internal static partial class User32
     {
-        [DllImport(Libraries.User32, ExactSpelling = true)]
-        public static extern IntPtr CopyImage(IntPtr hImage, IMAGE type, int cx, int cy, LR flags);
+        [LibraryImport(Libraries.User32)]
+        public static partial IntPtr CopyImage(IntPtr hImage, IMAGE type, int cx, int cy, LR flags);
 
         public static IntPtr CopyImage(IHandle hImage, IMAGE type, int cx, int cy, LR flags)
```
AaronRobinsonMSFT, line 14: Many of these small helper functions could be removed and the source generator will generate them on your behalf. /cc @jkoritzinsky @elinor-fung @JeremyKuhne

gpetrou: If I understand correctly, you mean creating a CustomTypeMarshaller? Something like:
```csharp
[CustomTypeMarshaller(typeof(IHandle), Direction = CustomTypeMarshallerDirection.In, Features = CustomTypeMarshallerFeatures.UnmanagedResources | CustomTypeMarshallerFeatures.TwoStageMarshalling)]
internal struct HandleMarshaller
{
    private IHandle _handle;

    public HandleMarshaller(IHandle handle)
    {
        _handle = handle;
    }

    public IntPtr ToNativeValue() => _handle.Handle;

    public void FreeNative() => GC.KeepAlive(_handle.Handle);
}
```
And then use it like:
```csharp
[LibraryImport(Libraries.User32)]
public static partial IntPtr CopyImage([MarshalUsing(typeof(HandleMarshaller))] IHandle hImage, IMAGE type, int cx, int cy, LR flags);
```
BTW, it would be nice to make https://github.com/dotnet/runtime/blob/bcded447f1558e7f891f3411f566db49f2ced6ec/src/libraries/Common/src/System/Runtime/InteropServices/HandleRefMarshaller.cs public to avoid copying/pasting the same code here for HandleRef.

AaronRobinsonMSFT: > BTW, it would be nice to make https://github.com/dotnet/runtime/blob/bcded447f1558e7f891f3411f566db49f2ced6ec/src/libraries/Common/src/System/Runtime/InteropServices/HandleRefMarshaller.cs public to avoid copying/pasting the same code here for HandleRef.

We are still waiting to see the impact of not including that marshaller. Our guess is it is very limited to certain codebases and not worth the cost. The reason here is uses of `HandleRef` really should be converted to a `SafeHandle` derivative, which would get marshalled automatically. The goal is to get rid of `HandleRef` from code bases so we intentionally made it less easy.

AaronRobinsonMSFT: Your usage should replace `IHandle` with `HandleRef` I believe, but otherwise the signature is what I would expect.

RussKie: IIRC we introduced `IHandle` addressing an AV in https://github.com/dotnet/winforms/pull/1612, and to remove the general indirection of `HandleRef`. @JeremyKuhne can provide more information on that.
_If_ we need to go back to `HandleRef` I suppose we can consider adding an extension method that would convert `IHandle` to `HandleRef`.

RussKie (commented):
👍 LGTM

src/System.Windows.Forms.Primitives/src/Interop/User32/Interop.CopyImage.cs
jkoritzinsky: Instead of using `MarshalUsing` at every usage site, you can use the `[NativeTypeMarshalling]` attribute on `IHandle` to set a default handler.

gpetrou: > Instead of using `MarshalUsing` at every usage site, you can use the `[NativeTypeMarshalling]` attribute on `IHandle` to set a default handler.

Thanks. I wasn't aware of that.

gpetrou:
@JeremyKuhne any thoughts on this PR? Could we merge it as it is and continue work on subsequent PRs?

src/System.Windows.Forms.Primitives/src/Interop/User32/Interop.CopyImage.cs
gpetrou: What do you want me to do with this PR, so that we can merge it? Perhaps the changes for IHandle and HandleRef can be done in a separate PR?

JeremyKuhne: @gpetrou just keep the wrapper methods. We can iterate when https://github.com/dotnet/runtime/issues/69403 is resolved.

JeremyKuhne (approved):
Took another look. @gpetrou, thanks for doing this, and thanks for your patience.

RussKie:
Thank you all

PR #7183, opened 2022-05-16
Fix TreeView overflow exception

### LINKED ISSUE #7098: TreeView OverflowException
### .NET version

net6 x86

Yes

unknown

### Issue description

I'm hitting an OverflowException in TreeView.cs due to the cast to IntPtr on x86 (32-bit) with multiple monitors.

It looks like this is likely fixed under net7 with https://github.com/dotnet/winforms/pull/5791 .

`User32.SendMessageW(this, User32.WM.CONTEXTMENU, Handle, (IntPtr)User32.GetMessagePos());`

The problem is that `GetMessagePos()` could return 0xFFFEFDB5 which overflows on 32-bit
```
            //var x = GetMessagePos(); // -> 0xFFFEFDB5
            uint n = 0xFFFEFDB5;
            var nn = (IntPtr)n; // throws OverflowException
```

Is it possible to get a fix for this in net6 since net7 is still 1/2 year+ away?

### Steps to reproduce

see above

dreddy-work:
Fixing an overflow exception in `TreeView` control in 32 bit process due to incorrect cast from `uint -> IntPtr`.

Fixes #7098 

## Proposed changes

- Cast return value (`nint`) of `GetMessagePos()` in the `TreeView `control to `int` before casting it to `IntPtr`

## Customer Impact

- Customer should be able to use `TreeView `control in a 32-bit process.

## Regression? 

- Regressed from .NET 3.1

## Risk

- Low. Similar change is in .NET 7.0 since Preview 1

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7183)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TreeView.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TreeView.cs
@@ -3095,7 +3095,7 @@ namespace System.Windows.Forms
                             else
                             {
                                 treeViewState[TREEVIEWSTATE_showTreeViewContextMenu] = true;
-                                User32.SendMessageW(this, User32.WM.CONTEXTMENU, Handle, (IntPtr)User32.GetMessagePos());
+                                User32.SendMessageW(this, User32.WM.CONTEXTMENU, Handle, (IntPtr)(int)User32.GetMessagePos());
                             }
 
                             m.Result = (IntPtr)1;


JeremyKuhne:
This is a regression from .NET Framework.

JeremyKuhne:
Additionally, the change has not technically been in since 7.0. A different (but effectively similar) change was made in 7.0.

JeremyKuhne (approved):

JeremyKuhne:
This was also regressed in Core. The problem did not exist in 3.1.

https://github.com/dotnet/winforms/blob/f4f009f89078f24712714f75edadd1fb3ccd6ef6/src/Common/src/SafeNativeMethods.cs#L99

dreddy-work:
approved via email.

RussKie:
What about a regression test?

PR #7187, opened 2022-05-17
Enable nullability in SelectedListViewItemCollection

gpetrou:
## Proposed changes

- Enable nullability in SelectedListViewItemCollection.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7187)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListView.CheckedListViewItemCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListView.CheckedListViewItemCollection.cs
@@ -22,7 +22,7 @@ namespace System.Windows.Forms
             /* C#r: protected */
             public CheckedListViewItemCollection(ListView owner)
             {
-                _owner = owner;
+                _owner = owner.OrThrowIfNull();
             }
 
             /// <summary>
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListView.SelectedListViewItemCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListView.SelectedListViewItemCollection.cs
@@ -25,7 +25,7 @@ namespace System.Windows.Forms
             /* C#r: protected */
             public SelectedListViewItemCollection(ListView owner)
             {
-                _owner = owner;
+                _owner = owner.OrThrowIfNull();
             }
 
             private ListViewItem[] SelectedItemArray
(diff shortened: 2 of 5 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ListView.SelectedListViewItemCollection.cs
```diff
@@ -17,37 +15,37 @@ public partial class ListView
...
             ///  We use an index here rather than control so that we don't have lifetime
             ///  issues by holding on to extra references.
-            private int lastAccessedIndex = -1;
+            private int _lastAccessedIndex = -1;
 
             /* C#r: protected */
             public SelectedListViewItemCollection(ListView owner)
             {
-                this.owner = owner;
+                _owner = owner;
```
Tanya-Solyanik, line 28: `OrThrowIfNull` is an allowed breaking change in this case because we will throw a less specific exception on line 35 or elsewhere. Not sure why anyone would new-up this class, but it is public. https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/breaking-change-rules.md#exceptions.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListView.CheckedListViewItemCollectionTests.cs
```diff
@@ -0,0 +1,17 @@
...
+using Xunit;
+
+namespace System.Windows.Forms.Tests
+{
+    public class ListView_CheckedListViewItemCollectionTests : IClassFixture<ThreadExceptionFixture>
+    {
+        [WinFormsFact]
+        public void CheckedListViewItemCollection_Ctor_OwnerIsNull_ThrowsArgumentNullException()
+        {
+            Assert.Throws<ArgumentNullException>("owner", () => { new ListView.CheckedListViewItemCollection(null); });
```
Tanya-Solyanik, line 14: Nice!

Tanya-Solyanik (approved):
Thank you!

src/System.Windows.Forms/src/System/Windows/Forms/ListView.CheckedListViewItemCollection.cs
```diff
@@ -22,7 +22,7 @@ public class CheckedListViewItemCollection : IList
             /* C#r: protected */
             public CheckedListViewItemCollection(ListView owner)
             {
-                _owner = owner;
+                _owner = owner.OrThrowIfNull();
```
Tanya-Solyanik, line 25: Nice catch!

PR #7145, opened 2022-05-06
Implement StandardAccessble RCW and AccessibleObject CCW

kant2002:
I realize that EnumVariantWrapper RCW is essentially always implement 3 interfaces which is needed
IEnumVariant, IAccessible and IOleWindow
Some internal things of Win32 prevent me using just single call to `CreateStdAccessibleObject`
Also add CCW for AccessibleObject

With this change you can run WinForms application with disabled built-in COM except following controls:
- RichTextBox
- WebBrowser
- AxHost

That's important step forward #5163 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7145)

PR #7191, opened 2022-05-18
Test that numeric up-down successfully focused

kant2002:
This is successfully catches bug in the https://github.com/dotnet/winforms/pull/7145 implementation
I need just call to `GetFocused`, but if some additional check can be made, I can add them.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7191)

### DIFF
--- a/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/NumericUpDownTests.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/NumericUpDownTests.cs
@@ -17,16 +17,15 @@ namespace System.Windows.Forms.UITests
         [WinFormsFact]
         public async Task NumericUpDownAccessibleObject_Focused_ReturnsCorrectValueAsync()
         {
-            await RunSingleControlTestAsync<NumericUpDown>((form, control) =>
+            await RunSingleControlTestAsync<NumericUpDown>(async (form, control) =>
             {
                 var accessibleObject = control.AccessibilityObject;
+                await MoveMouseToControlAsync(control);
                 form.Activate();
                 control.Focus();
 
                 var focused = accessibleObject.GetFocused();
                 Assert.NotNull(focused);
-
-                return Task.CompletedTask;
             });
         }
     }
--- a/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/NumericUpDownTests.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/NumericUpDownTests.cs
@@ -1,0 +1,34 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Drawing;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace System.Windows.Forms.UITests
+{
+    public class NumericUpDownTests : ControlTestBase
+    {
+        public NumericUpDownTests(ITestOutputHelper testOutputHelper)
+            : base(testOutputHelper)
+        {
+        }
+
+        [WinFormsFact]
+        public async Task NumericUpDownAccessibleObject_Focused_ReturnsCorrectValueAsync()
+        {
+            await RunSingleControlTestAsync<NumericUpDown>((form, control) =>
+            {
+                control.Focus();
+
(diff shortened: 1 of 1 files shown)


kant2002 (commented):
Remove unused using

Tanya-Solyanik (dismissed):
Thank you!

kant2002:
This test is flacky. I do not get what happens and what trick to apply.
Between call to SetFocus and checking focus state somehow focus gets lost. Would like to have advice on what I can do here to make test more reliable.

RussKie:
```
  Starting:    System.Windows.Forms.UI.IntegrationTests
    System.Windows.Forms.UITests.NumericUpDownTests.NumericUpDownAccessibleObject_Focused_ReturnsCorrectValueAsync [FAIL]
      NumericUpDown's text box accessbile object state should be focused
      Expected: True
      Actual:   False
      Stack Trace:
        /_/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/NumericUpDownTests.cs(28,0): at System.Windows.Forms.UITests.NumericUpDownTests.<NumericUpDownAccessibleObject_Focused_ReturnsCorrectValueAsync>b__1_0(Form form, NumericUpDown control)
        /_/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/ControlTestBase.cs(242,0): at System.Windows.Forms.UITests.ControlTestBase.<>c__DisplayClass28_0`1.<<RunFormAsync>b__0>d.MoveNext()
        --- End of stack trace from previous location ---
           at Microsoft.VisualStudio.Threading.JoinableTask.JoinAsync(CancellationToken cancellationToken)
        /_/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/ControlTestBase.cs(261,0): at System.Windows.Forms.UITests.ControlTestBase.RunFormAsync[T](Func`1 createDialog, Func`3 testDriverAsync)
        /_/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/ControlTestBase.cs(162,0): at System.Windows.Forms.UITests.ControlTestBase.RunSingleControlTestAsync[T](Func`3 testDriverAsync)
        /_/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/NumericUpDownTests.cs(20,0): at System.Windows.Forms.UITests.NumericUpDownTests.NumericUpDownAccessibleObject_Focused_ReturnsCorrectValueAsync()
        --- End of stack trace from previous location ---
      Output:
        Textbox state: Focusable
```

kant2002:
Does you have this error locally?

kant2002:
So situation is following. 
If run just single test in VS or via `dotnet test --filter NumericUpDownAccessibleObject_Focused_ReturnsCorrectValueAsync` then it fails 0 times out of 100.

If I run this test as part of whole project, then it fails 25 times out of 100.

If I add delay for 1_000 ms between `Control,Focus()` and call to `GetFocused()` then nothing changes, and it fails 28 times our of 100 runs. 

My understanding is that some app steal focus from main app and this make `GetFocused()`  return not-focused accessible object. I was able reliable break test if inside VS once form shown, I click somewhere else to switch focus from test form to VS instance.

kant2002:
After adding `Form.Activate()` I was managed to reach 3 failures out of 100 runs. I think that's acceptable for now. If somebody have any ideas how to make test fully reliable I would like to know.

RussKie:
Thinking out loud. May the failures have anything to do with a position of a mouse? May be move the mouse explicitly to (0, 0)?

kant2002:
Subjectively adding `await MoveMouseToControlAsync(control);` make test slightly more reliable.

| Run | Failures | Attempts |
|-|-|-|
| 1 | 3 | 100 |
| 2 | 1 | 100 |
| 3 | 0 | 100 |

kant2002:
I run 1000 runs of `dotnet test System.Windows.Forms.UI.IntegrationTests.csproj` and this specific test was not fail even once. My previous results probably have too many failures because I by mistake count all failures in tests runs, and not just failures from this particular test.
I definitely remember that second run do not have NumericUpDown failures, but first one I may see, or may not. My PC crash and I lost that information. 

when I receive notifications from other apps, like Discord test suite almost immidiately start to failing here and there. Just observation.

Tanya-Solyanik (approved):

PR #7194, opened 2022-05-19
VB fix title `MsgBox` and `InputBox` APIs

### LINKED ISSUE #3161: The default title of MsgBox is not correct
* .NET Core Version: 
5.0.100-preview.3.20216.6

No

**Problem description:**

The default title is "Microsoft.VisualBasic.Core" instead of the name of the calling assembly.
image

**Expected behavior:**
The default title should be the name of the calling assembly.
image

**Minimal repro:**
Run the following code in a VB winforms project:
```vbnet
MsgBox("Content")
```

hamarb123:
Fixes #3161

## Proposed changes

- Fix the regression of `MsgBox` and `InputBox` APIs specifying no title and getting `Microsoft.VisualBasic.Core` (or other assembly name based on inlining) as their title instead of the calling assembly

## Customer Impact

- Fixes a regression

## Regression? 

- No

## Risk

- N/A

### Before

image
image

### After

image
image

## Test methodology 

- I modified one of the projects in the solution to test it, I just removed the title parameter and gave it appropriate command-line arguments. Project name is in images above
- I also tested it with the product property set in the project file, and it worked

## Accessibility testing  

Not sure if I need to run this, please advise
 

## Test environment(s) 

 

200% scaling
English

```
.NET SDK (reflecting any global.json):
 Version:   7.0.100-preview.3.22154.3
 Commit:    5fe093bc35

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.19044
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   <redacted>\winforms\.dotnet\sdk\7.0.100-preview.3.22154.3\

global.json file:
  Not found

Host:
  Version:      7.0.0-preview.5.22267.11
  Architecture: x64
  Commit:       5b1fd50332

.NET SDKs installed:
  7.0.100-preview.3.22154.3 [<redacted>\winforms\.dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 7.0.0-preview.3.22152.6 [<redacted>\winforms\.dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 7.0.0-preview.3.22151.6 [<redacted>\winforms\.dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 7.0.0-preview.5.22267.11 [<redacted>\winforms\.dotnet\shared\Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 7.0.0-preview.3.22128.8 [<redacted>\winforms\.dotnet\shared\Microsoft.WindowsDesktop.App]

Download .NET:
  https://aka.ms/dotnet-download

Learn about .NET Runtimes and SDKs:
  https://aka.ms/dotnet/runtimes-sdk-info
```

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7194)

### DIFF
--- a/src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/Interaction.vb
+++ b/src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/Interaction.vb
@@ -286,9 +286,9 @@ Namespace Microsoft.VisualBasic
                 ParentWindow = vbhost.GetParentWindow()
             End If
 
-            If String.IsNullOrEmpty(Title) Then
+            If Title Is Nothing Then
                 If vbhost Is Nothing Then
-                    Title = GetTitleFromAssembly(System.Reflection.Assembly.GetCallingAssembly())
+                    Title = GetTitleFromCallingAssembly()
                 Else
                     Title = vbhost.GetWindowTitle()
                 End If
@@ -315,6 +315,28 @@ Namespace Microsoft.VisualBasic
 
         Private Function GetTitleFromAssembly(CallingAssembly As Reflection.Assembly) As String
 
+            'Tries to get the title using the AssemblyProduct attribute, which
+            'can be controlled with the Product property in a project file.
+            Dim attribs = CallingAssembly.GetCustomAttributes(GetType(Reflection.AssemblyProductAttribute), False)
+            If attribs IsNot Nothing AndAlso attribs.Length > 0 Then
+
+                Dim Title = DirectCast(attribs(0), Reflection.AssemblyProductAttribute).Product
+
+                '"Microsoft® .NET" is the default Product name when one is not
+                'explicitly specified, fall back to default case (use assembly name)
+                If Title <> "Microsoft® .NET" Then
+                    Return Title
+                End If
+
+            End If
+
+            'Default case: no assembly product name, fall back to assembly name
+            Return Nothing
+
+        End Function
+
+        Private Function GetNameFromAssembly(CallingAssembly As Reflection.Assembly) As String
+
             Dim Title As String
 
             'Get the Assembly name of the calling assembly 
@@ -338,7 +360,54 @@ Namespace Microsoft.VisualBasic
             End Try
 
             Return Title
+        End Function
 
+        Private Function GetTitleFromCallingAssembly() As String
+            Dim stackTrace As New StackTrace()
+
+            Dim frames = stackTrace.GetFrames()
+
+            Dim Title As String
+
+            'By reading the frames, we maintain the old behavior of using the title / name of the calling
+            'assembly (unless inlined), not necessarily the name of the application as a whole.
+            For i As Integer = 0 To frames.Length - 1
+                Dim frame = frames(i)
+                If frame.HasMethod() Then
+
+                    Dim assembly = frame.GetMethod().DeclaringType.Assembly
+                    Dim name = GetNameFromAssembly(assembly)
+
+                    'The following strings are the assembly names between the public api and here, this
+                    'may break in the future and the list may need to be expanded. This also does not
+                    'account for any of these assemblies using these apis without specifying a title,
+                    'but this it is not likely that the title wouldn't be specified in these calls.
+                    If name <> "Microsoft.VisualBasic.Forms" AndAlso name <> "System.Private.CoreLib" AndAlso name <> "Microsoft.VisualBasic.Core" Then
+                        Title = GetTitleFromAssembly(assembly)
+                        If Title Is Nothing Then
+                            Title = name
+                        End If
+                        Return Title
+                    End If
+
+                End If
+            Next
+
+            'Attempt to fallback to entry assembly, this can fail when GetEntryAssembly() returns
+            'Nothing. This code is only ever called if there is no stack trace outside of the known
+            'excluded assemblies, in which case it fall back to using the entry assembly.
+            Dim entryAssembly = Reflection.Assembly.GetEntryAssembly()
+            If entryAssembly IsNot Nothing Then
+                Title = GetTitleFromAssembly(entryAssembly)
+                If Title Is Nothing Then
+                    Title = GetNameFromAssembly(entryAssembly)
+                End If
+                Return Title
+            End If
+
+            'This can return "Microsoft® .NET" instead of something useful if the product name is not
+            'specified in the project file, this is the fallback for no stack trace and no entry assembly.
+            Return System.Windows.Forms.Application.ProductName
         End Function
 
         Private Function InternalInputBox(Prompt As String, Title As String, DefaultResponse As String, XPos As Integer, YPos As Integer, ParentWindow As Windows.Forms.IWin32Window) As String
@@ -387,7 +456,7 @@ Namespace Microsoft.VisualBasic
             Try
                 If Title Is Nothing Then
                     If vbhost Is Nothing Then
-                        sTitle = GetTitleFromAssembly(System.Reflection.Assembly.GetCallingAssembly())
+                        sTitle = GetTitleFromCallingAssembly()
                     Else
                         sTitle = vbhost.GetWindowTitle()
                     End If
--- a/src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/Interaction.vb
+++ b/src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/Interaction.vb
@@ -317,10 +317,10 @@ Namespace Microsoft.VisualBasic
 
             'Tries to get the title using the AssemblyProduct attribute, which
             'can be controlled with the Product property in a project file.
-            Dim attribs = CallingAssembly.GetCustomAttributes(GetType(Reflection.AssemblyProductAttribute), False)
-            If attribs IsNot Nothing AndAlso attribs.Length > 0 Then
+            Dim attributes = CallingAssembly.GetCustomAttributes(GetType(Reflection.AssemblyProductAttribute), False)
+            If attributes IsNot Nothing AndAlso attributes.Length > 0 Then
 
-                Dim Title = DirectCast(attribs(0), Reflection.AssemblyProductAttribute).Product
+                Dim Title = DirectCast(attributes(0), Reflection.AssemblyProductAttribute).Product
 
                 '"Microsoft® .NET" is the default Product name when one is not
                 'explicitly specified, fall back to default case (use assembly name)
--- a/src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/Interaction.vb
+++ b/src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/Interaction.vb
@@ -286,7 +286,7 @@ Namespace Microsoft.VisualBasic
                 ParentWindow = vbhost.GetParentWindow()
             End If
 
-            If Title Is Nothing Then
+            If String.IsNullOrEmpty(Title) Then
                 If vbhost Is Nothing Then
                     Title = GetTitleFromCallingAssembly()
                 Else
--- a/src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/Interaction.vb
+++ b/src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/Interaction.vb
@@ -369,8 +369,10 @@ Namespace Microsoft.VisualBasic
 
             Dim Title As String
 
-            'By reading the frames, we maintain the old behavior of using the title / name of the calling
-            'assembly (unless inlined), not necessarily the name of the application as a whole.
+            'By reading the frames, we maintain the old behavior of using the name of the calling
+            'assembly (unless inlined), not necessarily the name of the application as a whole. New
+            'behavior is that we try to use an assembly's product name if available, with a fallback
+            'of the dll name (dll name was the old behavior).
             For i As Integer = 0 To frames.Length - 1
                 Dim frame = frames(i)
                 If frame.HasMethod() Then


RussKie:
@KlausLoeffelmann please review

src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/Interaction.vb
```diff
@@ -315,6 +315,28 @@ Namespace Microsoft.VisualBasic
 
         Private Function GetTitleFromAssembly(CallingAssembly As Reflection.Assembly) As String
 
+            'Tries to get the title using the AssemblyProduct attribute, which
+            'can be controlled with the Product property in a project file.
+            Dim attribs = CallingAssembly.GetCustomAttributes(GetType(Reflection.AssemblyProductAttribute), False)
```
Tanya-Solyanik, line 320: Please spell out variable names: `attributes`

src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/Interaction.vb
```diff
@@ -338,7 +360,54 @@ Namespace Microsoft.VisualBasic
...
+
+        Private Function GetTitleFromCallingAssembly() As String
+            Dim stackTrace As New StackTrace()
+
+            Dim frames = stackTrace.GetFrames()
+
+            Dim Title As String
+
+            'By reading the frames, we maintain the old behavior of using the title / name of the calling
+            'assembly (unless inlined), not necessarily the name of the application as a whole.
```
Tanya-Solyanik, line 373: I'm not sure if dll name is useful as a title, why did you prioritize this case over the ProductName? And if product name is not available, then return the dll name.

Tanya-Solyanik:
No special accessibility testing is required in this case.

src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/Interaction.vb
hamarb123: Sorry that it's unclear, it's getting the dll name to check if the frame should be skipped (line 385), and then using `GetTitleFromAssembly` to attempt to read the product name, with a fallback to the dll name (lines 386-390). The comment is supposed to imply that with `using the title / name of the calling assembly`, it is probably a bit unclear, should I change it?

src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/Interaction.vb
hamarb123: done

src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/Interaction.vb
hamarb123: I've updated the comment to make it clearer.

KlausLoeffelmann (changes requested):
I would like to do a Code Review with a fellow team member in a 1:1.
I don't have such a good feeling with this on first glance, so I am blocking this for the time being, until I have time to take a closer look.

KlausLoeffelmann:
@RussKie, if you got the time, let's talk.

RussKie:
Is this not good?
https://github.com/dotnet/winforms/blob/3c0c5925767bcefb5d4eaf2b67dd64121f0bd242/src/System.Windows.Forms/src/System/Windows/Forms/Application.cs#L328-L332

hamarb123:
> Is this not good?
> 
> https://github.com/dotnet/winforms/blob/3c0c5925767bcefb5d4eaf2b67dd64121f0bd242/src/System.Windows.Forms/src/System/Windows/Forms/Application.cs#L328-L332

It should get the product name / assembly name of the calling assembly, otherwise it's a breaking change that realistically is almost as bad as having `Microsoft.VisualBasic.Core`. The breaking change is that it used to get the calling assembly's file name, not the application's name - many people probably relied on this behaviour (including me).
There is also the other issue with the `ProductName` API, it returns `"Microsoft® .NET"` if the user didn't explicitly set a product name in their project, which I don't think is desirable - if we were to use the app product name, that API would either need to be changed, or a new API would need to be created.

Tanya-Solyanik:
@hamarb123  - had you tried to step through the .NET framework implementation or the C# implementation?

hamarb123:
On .NET Framework in Interactive host. The reason it needs to be fixed on  this on is that .NET Core can't inline the `MsgBox` method, whereas .NET Framework can. 
image

RussKie:
> It should get the product name / assembly name of the calling assembly, otherwise it's a breaking change that realistically is almost as bad as having `Microsoft.VisualBasic.Core`. The breaking change is that it used to get the calling assembly's file name, not the application's name - many people probably relied on this behaviour (including me).
> There is also the other issue with the `ProductName` API, it returns `"Microsoft® .NET"` if the user didn't explicitly set a product name in their project, which I don't think is desirable - if we were to use the app product name, that API would either need to be changed, or a new API would need to be created.

I don't see a problem here to be honest. [`MsgBox` API](https://docs.microsoft.com/dotnet/api/microsoft.visualbasic.interaction.msgbox#microsoft-visualbasic-interaction-msgbox(system-object-microsoft-visualbasic-msgboxstyle-system-object)) allows setting the desired dialog title.
It is worth noting that this API is truly legacy, and it is just a convenience wrapper over [`MessageBox` API](https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.messagebox):
https://github.com/dotnet/winforms/blob/3c0c5925767bcefb5d4eaf2b67dd64121f0bd242/src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/Interaction.vb#L407-L412

The suggested fix would be to switch to `MessageBox` and obsolete `MsgBox` API.

KlausLoeffelmann:
> The breaking change is that it used to get the calling assembly's file name, not the application's name - many people probably relied on this behaviour (including me).

But walking the stack trace to increase the probability to get the desired information, which - and granted this is my assumption: - on top is not super-essential in the respective scenarios to begin with, is something we don't really want to consider in a PR. 

My reason: The assembly name is nothing which should be surfaced to the end user in standard scenarios. Surely there are cases where this needs to be done, but then MsgBox is not the right API to _have_ that done. The reason it's doing that lies a little bit in the history of this API, which derives directly from its VB6 predecessor. And one could argue the way it worked back than in the 90s, it made more sense back in the days. But it doesn't now.

So, we discussed this a bit internally, and I agree 100% with @RussKie here. We should:

a) Introduce a breaking change for InputBox as he described.
b) We should obsolete MsgBox, as there are way better modern alternatives, and their adaptation in existing apps is really not a big thing.

Pinging @KathleenDollard for FYI and her take.

hamarb123:
> a) Introduce a breaking change for InputBox as he described.

Not sure what he described. My logic is outlined below for MsgBox though.

b) We should obsolete MsgBox, as there are way better modern alternatives, and their adaptation in existing apps is really not a big thing.

Here's what I think would be reasonable (or anything less breaking): leave current MsgBox and InputBox implementation as-is with an ObsoleteAttribute.
In VB.NET, when using whatever runtime version this is or greater, the MsgBox 'keyword' (if that's what it is) could instead be a shortcut of MessageBox.Show, this will probably stuff a bunch of people using it, but it would give them an error at least.

My reason:
I've currently got old versions of an application that I would like to still work the same in future .NET versions (I'm hoping to be able to just enable roll forward).
I'm happy to deal with a compiler error for future versions, but I don't want to have to modify old versions preferably.

hamarb123:
How long would you approximate it would take to be removed?

KlausLoeffelmann:
I'd go for obsoleting it in .NET 7 first, but not otherwise touch it in .NET 7.

KlausLoeffelmann:
Ugh, well, that changes everything. If it is part of the language specification, we cannot remove it.
@RussKie, FYI.

RussKie:
> MsgBox is part of the VB Language specification 6.1.2.8.1.13 not the .Net Runtime.

Could you please point me to the section you're referring to? I've looked through [Visual Basic language specification](https://www.microsoft.com/en-us/download/details.aspx?id=15039) and was unable to find it.

> Ugh, well, that changes everything. If it is part of the language specification, we cannot remove it.

That's incorrect. Anything can be made obsolete and removed. It is a question of following the prescribed process. 
_If_ `MsgBox` is a part of the language specification (which I need a confirmation of) then here's the process: https://docs.microsoft.com/dotnet/visual-basic/reference/language-specification/introduction#language-deprecation.

> obsoleting MsgBox is going to break a tremendous amount of code

Yes, it will be a breaking change. We do not make any guarantees that major version of .NET are free from breaking changes, though it doesn't mean there's a free reign.
The migration to `MessageBox` API should be relatively straight forward, and [.NET Upgrade Assistant](https://dotnet.microsoft.com/platform/upgrade-assistant) should be able to help with that.

hamarb123:
> MsgBox is part of the VB Language specification 6.1.2.8.1.13 not the .Net Runtime.

Here is what he was referring to: https://docs.microsoft.com/en-us/openspecs/microsoft_general_purpose_programming_languages/ms-vbal/ba0375ee-936a-482c-8a9f-e9fa4515d191
Although it's not a part of the language specification for VB.NET (it is for VBA), I think many people treat it as such, because on VB.NET WinForms project (which is one of the most common uses for VB.NET), it is automatically imported with other functions that are essentially treated as keywords.

I would like to point out that there are approximately 143k uses of `MsgBox` on GitHub API in VB.NET (excluding vba and the likes) using a simple search (https://github.com/search?l=Visual+Basic+.NET&q=msgbox+extension%3Avb&type=Code). I think it would be better to fix it and possibly put a warning on it indefinitely that says something along the lines of `Usage of MessageBox.Show is preferred` - not implying that it will be broken in the future, because I think it shouldn't be, but because `MessageBox.Show` is more powerful (though I might add for most cases, `MsgBox` provides all you need) and has better argument types. Even a warning is probably too far though, due to its frequency of usage.

If any of these projects decide to port to .NET Core they are already going to have to specify a title manually if they want behaviour the same as before (which is what this issue is trying to fix), I think also having to change all of these API calls to a different function with different parameter order and types would turn people with lots of `MsgBox` calls (which would be a decent amount of people who use WinForms with VB.NET) off transitioning, or delay it to when there would be even more breaking changes in other parts of the language and runtime. Personally, I was already upset that all my `MsgBox` calls needed to be fixed, and it almost made me delay transitioning to .NET Core, if I had to change the API as well I would have been much more likely to delay it.

RussKie:
> Here is what he was referring to: [docs.microsoft.com/en-us/openspecs/microsoft_general_purpose_programming_languages/ms-vbal/ba0375ee-936a-482c-8a9f-e9fa4515d191](https://docs.microsoft.com/en-us/openspecs/microsoft_general_purpose_programming_languages/ms-vbal/ba0375ee-936a-482c-8a9f-e9fa4515d191)
> Although it's not a part of the language specification for VB.NET (it is for VBA), I think many people treat it as such, because on VB.NET WinForms project (which is one of the most common uses for VB.NET), it is automatically imported with other functions that are essentially treated as keywords.

Thank you for the link. As you noted, `MsgBox` is part of the VBA language spec, and not VB.NET, which we're discussing here.

> would like to point out that there are approximately 143k uses of `MsgBox` on GitHub API in VB.NET (excluding vba and the likes) using a simple search ([github.com/search?l=Visual+Basic+.NET&q=msgbox+extension%3Avb&type=Code](https://github.com/search?l=Visual+Basic+.NET&q=msgbox+extension%3Avb&type=Code)).

I'm not trying to downplay the number of hits, though on a cursory glance there's a significant portion of those look like "quick" mockup or "school assignment" type of projects. 

To reiterate _my_ position, I fully appreciate that for a lot of folks who use VB and VBA `MsgBox` is omnipresent, and we can't take this API at once. However, given the existing issueswith `MsgBox` API  (such as this one specifically, or lack of full compatibility with the native `MessageBoxW` API) I think marking this API as obsolete would be the right course of action. The obsoletion process entails the following:
- marking the API as obsolete, which will emit a build warning with a link to more information (refer to https://github.com/dotnet/designs/blob/main/accepted/2020/better-obsoletion/better-obsoletion.md),
- documenting the obsoletion,
- enhancing the [.NET Upgrade Assistant](https://dotnet.microsoft.com/platform/upgrade-assistant) to help converting `MsgBox` to `MessageBox` API.

KlausLoeffelmann:
>> Ugh, well, that changes everything. If it is part of the language specification, we cannot remove it.

> That's incorrect. Anything can be made obsolete and removed. It is a question of following the prescribed process.
If MsgBox is a part of the language specification (which I need a confirmation of) then here's the process: https://docs.microsoft.com/dotnet/visual-basic/reference/language-specification/introduction#language-deprecation.

OK, I should have checked @paul1956's statement, and not assume it correct, because I no longer think it is.

@KathleenDollard, @paul1956: This is the last version of the specs, @ljw1004 wrote back in the day. I don't know of a newer one. 

https://www.microsoft.com/en-us/download/details.aspx?id=15039

There is nothing in about `MsgBox`. And think about it: If that would be part of the VB specs, how would it work in a UWP App? Or in a .NET Standard Library? With it, I am back to my previous thoughts (well, almost, keep reading.)

If what Paul quoted were indeed the VBA specs, then know that VBA is not at all relevant for this decision, since it is aiming to be compatible to VB6, which Visual Basic is not. Please remember, we also got rid of the typical VB6 compatible libraries, which included things like `Left$`, `Right$` and such very early in the UWP days, if I remember correctly, so, my feeling is: `MsgBox` is right up in that alley.

(Also, as a side note, I am not so sure it's really still that heavily used. 
image
The stated query for `MsgBox` clearly yield also VBA code, like one of these first hits, so, I'd take the risk to phase it out.)

Apart from that: Walking the stack frame for making this work, is still nothing I would support at this point, except someone with REALLY good arguments changes my mind.

That all said, and this is my "I am almost back to my previous opinion": I think @KathleenDollard should weigh in here, because even if `MsgBox` is surfacing based on WinForms APIs, it is still historically a very VB-sy API, and she should have the final saying in this. 

My final 2 cents on this.

paul1956:
@KlausLoeffelmann If you Google VB MggBox you get sent to VBA and I did miss that, sorry. You would also need to obsolete all the constants associated with MggBox and any Assistant would need to convert those as well. Converting seems straight forward, and I don't see anything what would break. I have run into the .NET Standard issue many times where it is missing and found workarounds. The Title issue has been around at least since the first Release of .NET Core 3.0/WinForms where I had to implement my own version so I would document it and move on.

RussKie:
@hamarb123 thank you for your time and efforts. However, given that a relatively trivial workaround exists (i.e., supply the title parameter), and to avoid perf-related regressions, we can't take your changes. 
Also, as  noted above, both `MsgBox` and `InputBox` API have essentialy reached their EOL, and should be marked as obsolete.

PR #7206, opened 2022-05-23
Refactoring AutomationId at the accessible object classes #6603

### LINKED ISSUE #6603: Refactoring `AutomationId` at the `GetPropertyValue` method

dkazennov:
Fixes #6603
This PR finishes https://github.com/dotnet/winforms/pull/6641 (adds unit tests)

## Proposed changes

- The AutomationId property is moved to the AccessibleObject base class
- This property returns null at AccessibleObject
- It is returned at the GetPropertyValue method of AccessibleObject for the case AutomationIdPropertyId
- The current implementations in the 4 classes derived from AO are kept the same way, but they are marked as overrides
- Returning of the Owner.Name is moved to the ControlAccessibleObjectObject class to make it universal for all its descendants
- All duplicate code with the same logic is cleaned out in the derived classes
- Initial discussion on this topic is here: https://github.com/dotnet/winforms/pull/6429#issuecomment-1009286397

Unit tests:

1) AccessibleObject - AccessibleObject_GetPropertyValue_ReturnsNull (two unit tests combined)
2) DataGridViewCellAccessibleObject - DataGridViewCellAccessibleObject_GetPropertyValue_AutomationId_ReturnsExpected (new)
3) ListViewItemBaseAccessibleObject - ListViewItemBaseAccessibleObject_GetPropertyValue_AutomationId_ReturnsExpected (new)
4) ListViewSubItemAccessibleObject - ListViewSubItemAccessibleObject_GetPropertyValue_returns_correct_values (existing)
5) ControlAccessibleObject - ControlAccessibleObject_GetPropertyValue_AutomationId_ReturnsExpected (new)
6) ButtonBaseAccessibleObject - ButtonBaseAccessibleObject_GetPropertyValue_AutomationId_ReturnsExpected (new)
7) CheckBoxAccessibleObject - CheckBoxAccessibleObject_GetPropertyValue_Invoke_ReturnsExpected (existing)
8) GroupBoxAccessibleObject - GroupBoxAccessibleObject_GetPropertyValue_AutomationId_ReturnsExpected (new)
9) LabelAccessibleObject - LabelAccessibleObject_GetPropertyValue_Invoke_ReturnsExpected (two unit tests combined)
10) ListViewAccessibleObject - ListViewAccessibleObject_GetPropertyValue_returns_correct_values (existing)
11) PanelAccessibleObject - PanelAccessibleObject_GetPropertyValue_Invoke_ReturnsExpected (existing)
12) PictureBoxAccessibleObject - PictureBoxAccessibleObject_GetPropertyValue_Invoke_ReturnsExpected (existing)
13) RadioButtonAccessibleObject - RadioButtonAccessibleObject_GetPropertyValue_Invoke_ReturnsExpected (existing)
14) ScrollBarAccessibleObject - ScrollBarAccessibleObject_GetPropertyValue_Invoke_ReturnsExpected (existing)
15) SplitterAccessibleObject - SplitterAccessibleObject_GetPropertyValue_Invoke_ReturnsExpected (existing)
16) TabControlAccessibleObject - TabControlAccessibleObject_GetPropertyValue_Invoke_ReturnsExpected (two unit tests combined)
17) TabPageAccessibleObject - TabPageAccessibleObject_GetPropertyValue_Invoke_ReturnsExpected (two unit tests combined)
18) TrackBarAccessibleObject - TrackBarAccessibilityObject_GetPropertyValue_Invoke_ReturnsExpected (two unit tests combined)

Combined existed NamePropertyId tests and new AutomationIdPropertyId tests:
{
		TrackBarAccessibilityObject_GetPropertyValue_Invoke_ReturnsExpected
		TabPageAccessibleObject_GetPropertyValue_Invoke_ReturnsExpected
		TabControlAccessibleObject_GetPropertyValue_Invoke_ReturnsExpected
		LabelAccessibleObject_GetPropertyValue_Invoke_ReturnsExpected
}
just like existing:
{
		CheckBoxAccessibleObject_GetPropertyValue_Invoke_ReturnsExpected
		PanelAccessibleObject_GetPropertyValue_Invoke_ReturnsExpected
		PictureBoxAccessibleObject_GetPropertyValue_Invoke_ReturnsExpected
                ListViewSubItemAccessibleObject_GetPropertyValue_returns_correct_values
		ListViewAccessibleObject_GetPropertyValue_returns_correct_values
}

## Customer Impact

- Tests based on the old implementation of AutomationId could become obsolete.

## Regression? 

- No

## Risk

- Minimal

## Test methodology 

- Unit tests
- CTI testing

## Test environment(s) 

Microsoft Windows [Version 120.2212.4170.0]
.NET Core SDK: 7.0.100-preview.3.22179.4

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7206)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/AccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/AccessibleObject.cs
@@ -94,6 +94,8 @@ namespace System.Windows.Forms
             _systemWrapper = true;
         }
 
+        private protected virtual string? AutomationId => null;
+
         /// <summary>
         ///  Gets the bounds of the accessible object, in screen coordinates.
         /// </summary>
@@ -448,6 +450,7 @@ namespace System.Windows.Forms
                 UiaCore.UIA.IsPasswordPropertyId => false,
                 UiaCore.UIA.FrameworkIdPropertyId => NativeMethods.WinFormFrameworkId,
                 UiaCore.UIA.AccessKeyPropertyId => KeyboardShortcut ?? string.Empty,
+                UiaCore.UIA.AutomationIdPropertyId => AutomationId,
                 _ => null
             };
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxChildEditUiaProvider.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxChildEditUiaProvider.cs
@@ -33,6 +33,8 @@ namespace System.Windows.Forms
                 _textProvider = new ComboBoxUiaTextProvider(owner);
             }
 
+            private protected override string AutomationId => COMBO_BOX_EDIT_AUTOMATION_ID;
+
             /// <summary>
             ///  Returns the element in the specified direction.
             /// </summary>
@@ -88,8 +90,6 @@ namespace System.Windows.Forms
                         return (State & AccessibleStates.Focusable) == AccessibleStates.Focusable;
                     case UiaCore.UIA.IsEnabledPropertyId:
                         return _owningComboBox.Enabled;
-                    case UiaCore.UIA.AutomationIdPropertyId:
-                        return COMBO_BOX_EDIT_AUTOMATION_ID;
                     case UiaCore.UIA.NativeWindowHandlePropertyId:
                         return _handle;
                     case UiaCore.UIA.IsOffscreenPropertyId:
--- a/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/GridEntry.GridEntryAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/GridEntry.GridEntryAccessibleObject.cs
@@ -25,6 +25,8 @@ namespace System.Windows.Forms.PropertyGridInternal
                 _owningGridEntry = owner;
             }
 
+            private protected override string AutomationId => GetHashCode().ToString();
+
             public override Rectangle Bounds
             {
                 get => PropertyGridView is not null && PropertyGridView.IsHandleCreated
@@ -337,7 +339,6 @@ namespace System.Windows.Forms.PropertyGridInternal
                     UiaCore.UIA.HasKeyboardFocusPropertyId => _owningGridEntry.HasFocus,
                     UiaCore.UIA.IsKeyboardFocusablePropertyId => (State & AccessibleStates.Focusable) == AccessibleStates.Focusable,
                     UiaCore.UIA.IsEnabledPropertyId => true,
-                    UiaCore.UIA.AutomationIdPropertyId => GetHashCode().ToString(),
                     _ => base.GetPropertyValue(propertyID),
                 };
             }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TabPage.TabAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TabPage.TabAccessibleObject.cs
@@ -12,14 +12,14 @@ namespace System.Windows.Forms
     {
         internal class TabAccessibleObject : AccessibleObject
         {
-            private TabControl? OwningTabControl => _owningTabPage.ParentInternal as TabControl;
-
             private readonly TabPage _owningTabPage;
 
             public TabAccessibleObject(TabPage owningTabPage)
             {
                 _owningTabPage = owningTabPage.OrThrowIfNull();
             }
+
+            private protected override string AutomationId => _owningTabPage.Name;
 
             public override Rectangle Bounds
             {
@@ -47,6 +47,8 @@ namespace System.Windows.Forms
             public override string? DefaultAction => SystemIAccessibleInternal?.get_accDefaultAction(GetChildId());
 
             public override string? Name => _owningTabPage.Text;
+
+            private TabControl? OwningTabControl => _owningTabPage.ParentInternal as TabControl;
 
             public override AccessibleRole Role
                 => SystemIAccessibleInternal?.get_accRole(GetChildId()) is object accRole
@@ -116,7 +118,6 @@ namespace System.Windows.Forms
                 => propertyID switch
                 {
                     UiaCore.UIA.ControlTypePropertyId => UiaCore.UIA.TabItemControlTypeId,
-                    UiaCore.UIA.AutomationIdPropertyId => _owningTabPage.Name,
                     UiaCore.UIA.IsEnabledPropertyId => OwningTabControl?.Enabled ?? false,
                     UiaCore.UIA.HasKeyboardFocusPropertyId => (State & AccessibleStates.Focused) == AccessibleStates.Focused,
                     UiaCore.UIA.IsKeyboardFocusablePropertyId
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjectTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjectTests.cs
@@ -2568,12 +2568,14 @@ namespace System.Windows.Forms.Tests
             Assert.Equal(7, parent.accChildCount);
         }
 
-        [WinFormsFact]
-        public void AccessibleObject_GetPropertyValue_ControlTypeProperty_ReturnsNull()
+        [WinFormsTheory]
+        [InlineData((int)UiaCore.UIA.ControlTypePropertyId)]
+        [InlineData((int)UiaCore.UIA.AutomationIdPropertyId)]
+        public void AccessibleObject_GetPropertyValue_ReturnsNull_IfExpected(int propertyId)
         {
             AccessibleObject accessibleObject = new AccessibleObject();
 
-            Assert.Null(accessibleObject.GetPropertyValue(Interop.UiaCore.UIA.ControlTypePropertyId));
+            Assert.Null(accessibleObject.GetPropertyValue((UiaCore.UIA)propertyId));
         }
 
         public static IEnumerable<object[]> AccessibleObject_RuntimeId_IsOverriden_TestData()
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ButtonBase.ButtonBaseAccessibleObjectTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ButtonBase.ButtonBaseAccessibleObjectTests.cs
@@ -235,6 +235,17 @@ namespace System.Windows.Forms.Tests
             Assert.False(buttonBase.IsHandleCreated);
         }
 
+        [WinFormsFact]
+        public void ButtonBaseAccessibleObject_GetPropertyValue_AutomationId_ReturnsExpected()
+        {
+            using SubButtonBase ownerControl = new() { Name = "test name" };
+            string expected = ownerControl.Name;
+            object actual = ownerControl.AccessibilityObject.GetPropertyValue(UiaCore.UIA.AutomationIdPropertyId);
+
+            Assert.Equal(expected, actual);
+            Assert.False(ownerControl.IsHandleCreated);
+        }
+
         private class SubButtonBase : ButtonBase
         {
             public Action PerformClickAction { get; set; }
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Control.ControlAccessibleObjectTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Control.ControlAccessibleObjectTests.cs
@@ -1568,6 +1568,17 @@ namespace System.Windows.Forms.Tests
             Assert.True(toolStrip.IsHandleCreated);
         }
 
+        [WinFormsFact]
+        public void ControlAccessibleObject_GetPropertyValue_AutomationId_ReturnsExpected()
+        {
+            using Control ownerControl = new() { Name = "test name" };
+            string expected = ownerControl.Name;
+            object actual = ownerControl.AccessibilityObject.GetPropertyValue(UiaCore.UIA.AutomationIdPropertyId);
+
+            Assert.Equal(expected, actual);
+            Assert.False(ownerControl.IsHandleCreated);
+        }
+
         // ContextMenuStrip, From, ToolStripDropDown, ToolStripDropDownMenu
         // are Top level controls that can't be added to a ToolStrip.
         // A TabPage can be added to a TabControl only (see TabPage.AssignParent method).
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewCellAccessibleObjectTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewCellAccessibleObjectTests.cs
@@ -995,6 +995,24 @@ namespace System.Windows.Forms.Tests
             Assert.False(dataGridView.IsHandleCreated);
         }
 
+        [WinFormsFact]
+        public void DataGridViewCellAccessibleObject_GetPropertyValue_AutomationId_ReturnsExpected()
+        {
+            using DataGridView dataGridView = new();
+            dataGridView.Columns.Add(new DataGridViewTextBoxColumn());
+            dataGridView.Rows.Add(new DataGridViewRow());
+
+            DataGridViewCellAccessibleObject dataGridViewCellAccessibleObject = new(dataGridView.Rows[0].Cells[0]);
+            string expected = string.Empty;
+            foreach (int runtimeIdPart in dataGridViewCellAccessibleObject.RuntimeId)
+            {
+                expected += runtimeIdPart.ToString();
+            }
+
+            Assert.Equal(expected, dataGridViewCellAccessibleObject.GetPropertyValue(UiaCore.UIA.AutomationIdPropertyId));
+            Assert.False(dataGridView.IsHandleCreated);
+        }
+
         private class SubDataGridViewCell : DataGridViewCell
         {
         }
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/GroupBoxAccessibleObjectTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/GroupBoxAccessibleObjectTests.cs
@@ -119,5 +119,16 @@ namespace System.Windows.Forms.Tests
             Assert.Equal(expected, actual);
             Assert.False(groupBox.IsHandleCreated);
         }
+
+        [WinFormsFact]
+        public void GroupBoxAccessibleObject_GetPropertyValue_AutomationId_ReturnsExpected()
+        {
+            using GroupBox ownerControl = new() { Name = "test name" };
+            string expected = ownerControl.Name;
+            object actual = ownerControl.AccessibilityObject.GetPropertyValue(UiaCore.UIA.AutomationIdPropertyId);
+
+            Assert.Equal(expected, actual);
+            Assert.False(ownerControl.IsHandleCreated);
+        }
     }
 }
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/LabelAccessibleObjectTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/LabelAccessibleObjectTests.cs
@@ -9,20 +9,23 @@ namespace System.Windows.Forms.Tests
 {
     public class LabelAccessibleObjectTests
     {
-        [WinFormsFact]
-        public void LabelAccessibleObject_GetPropertyValue_Name_ReturnsExpected()
+        [WinFormsTheory]
+        [InlineData((int)UiaCore.UIA.NamePropertyId, "Address")]
+        [InlineData((int)UiaCore.UIA.AutomationIdPropertyId, "Label1")]
+        public void LabelAccessibleObject_GetPropertyValue_Invoke_ReturnsExpected(int propertyID, object expected)
         {
-            string testAccName = "Address";
-            using var label = new Label();
-            label.Text = "Some test label text";
-            label.Name = "Label1";
-            label.AccessibleName = testAccName;
-            AccessibleObject labelAccessibleObject = label.AccessibilityObject;
+            using Label label = new()
+            {
+                Text = "Some test label text",
+                Name = "Label1",
+                AccessibleName = "Address"
+            };
 
+            Label.LabelAccessibleObject accessibilityObject = (Label.LabelAccessibleObject)label.AccessibilityObject;
+
+            object value = accessibilityObject.GetPropertyValue((UiaCore.UIA)propertyID);
+            Assert.Equal(expected, value);
             Assert.False(label.IsHandleCreated);
-
-            var accessibleName = labelAccessibleObject.GetPropertyValue(Interop.UiaCore.UIA.NamePropertyId);
-            Assert.Equal(testAccName, accessibleName);
         }
 
         [WinFormsFact]
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItem.ListViewItemBaseAccessibleObjectTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItem.ListViewItemBaseAccessibleObjectTests.cs
@@ -336,5 +336,18 @@ namespace System.Windows.Forms.Tests
             Assert.Equal((UiaCore.ToggleState)expected, accessibleObject.ToggleState);
             Assert.False(control.IsHandleCreated);
         }
+
+        [WinFormsFact]
+        public void ListViewItemBaseAccessibleObject_GetPropertyValue_AutomationId_ReturnsExpected()
+        {
+            using ListView listView = new();
+            ListViewItem item = new();
+            listView.Items.Add(item);
+            var accessibleObject = (ListViewItemBaseAccessibleObject)item.AccessibilityObject;
+
+            var expected = string.Format("{0}-{1}", typeof(ListViewItem).Name, accessibleObject.CurrentIndex);
+            Assert.Equal(expected, accessibleObject.GetPropertyValue(UiaCore.UIA.AutomationIdPropertyId));
+            Assert.False(listView.IsHandleCreated);
+        }
     }
 }
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PictureBox.PictureBoxAccessibleObjectTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PictureBox.PictureBoxAccessibleObjectTests.cs
@@ -102,8 +102,7 @@ namespace System.Windows.Forms.Tests
                 AccessibleName = "TestName"
             };
 
-            Assert.False(pictureBox.IsHandleCreated);
-            var pictureBoxAccessibleObject = new PictureBox.PictureBoxAccessibleObject(pictureBox);
+            PictureBox.PictureBoxAccessibleObject pictureBoxAccessibleObject = new(pictureBox);
             object value = pictureBoxAccessibleObject.GetPropertyValue((UIA)propertyID);
 
             Assert.Equal(expected, value);
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TabControl.TabControlAccessibleObjectTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TabControl.TabControlAccessibleObjectTests.cs
@@ -744,18 +744,20 @@ namespace System.Windows.Forms.Tests
             Assert.False(tabControl.IsHandleCreated);
         }
 
-        [WinFormsFact]
-        public void TabControlAccessibleObject_GetPropertyValue_Name_ReturnsExpected()
+        [WinFormsTheory]
+        [InlineData((int)UIA.NamePropertyId, "TestName")]
+        [InlineData((int)UIA.AutomationIdPropertyId, "TabControl1")]
+        public void TabControlAccessibleObject_GetPropertyValue_Invoke_ReturnsExpected(int propertyID, object expected)
         {
-            const string name = "Test name";
             using TabControl tabControl = new()
             {
-                AccessibleName = name
+                Name = "TabControl1",
+                AccessibleName = "TestName"
             };
 
-            object actual = tabControl.AccessibilityObject.GetPropertyValue(UiaCore.UIA.NamePropertyId);
+            object actual = tabControl.AccessibilityObject.GetPropertyValue((UIA)propertyID);
 
-            Assert.Equal(name, actual);
+            Assert.Equal(expected, actual);
             Assert.False(tabControl.IsHandleCreated);
         }
 
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TabPage.TabPageAccessibleObjectTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TabPage.TabPageAccessibleObjectTests.cs
@@ -479,18 +479,20 @@ namespace System.Windows.Forms.Tests
             Assert.Equal(createControl, tabPage.IsHandleCreated);
         }
 
-        [WinFormsFact]
-        public void TabPageAccessibleObject_GetPropertyValue_Name_ReturnsExpected()
+        [WinFormsTheory]
+        [InlineData((int)UIA.NamePropertyId, "TestName")]
+        [InlineData((int)UIA.AutomationIdPropertyId, "TabPage1")]
+        public void TabPageAccessibleObject_GetPropertyValue_Invoke_ReturnsExpected(int propertyID, object expected)
         {
-            const string name = "Test name";
             using TabPage tabPage = new()
             {
-                AccessibleName = name
+                Name = "TabPage1",
+                AccessibleName = "TestName"
             };
 
-            object actual = tabPage.AccessibilityObject.GetPropertyValue(UIA.NamePropertyId);
+            object actual = tabPage.AccessibilityObject.GetPropertyValue((UIA)propertyID);
 
-            Assert.Equal(name, actual);
+            Assert.Equal(expected, actual);
             Assert.False(tabPage.IsHandleCreated);
         }
 
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TrackBar.TrackBarAccessibleObjectTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TrackBar.TrackBarAccessibleObjectTests.cs
@@ -413,17 +413,21 @@ namespace System.Windows.Forms.Tests
             Assert.False(trackBar.IsHandleCreated);
         }
 
-        [WinFormsFact]
-        public void TrackBarAccessibilityObject_GetPropertyValue_Name_ReturnsExpected()
+        [WinFormsTheory]
+        [InlineData((int)UiaCore.UIA.NamePropertyId, "TestAccessibleName")]
+        [InlineData((int)UiaCore.UIA.AutomationIdPropertyId, "TestControlName")]
+        public void TrackBarAccessibilityObject_GetPropertyValue_Invoke_ReturnsExpected(int propertyID, object expected)
         {
-            const string name = "Test name";
             using TrackBar trackBar = new()
             {
-                AccessibleName = name
+                Name = expected.ToString(),
+                AccessibleName = expected.ToString()
             };
-            object actual = trackBar.AccessibilityObject.GetPropertyValue(UiaCore.UIA.NamePropertyId);
 
-            Assert.Equal(name, actual);
+            TrackBar.TrackBarAccessibleObject accessibilityObject = (TrackBar.TrackBarAccessibleObject)trackBar.AccessibilityObject;
+
+            object value = accessibilityObject.GetPropertyValue((UiaCore.UIA)propertyID);
+            Assert.Equal(expected, value);
             Assert.False(trackBar.IsHandleCreated);
         }
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Button.ButtonAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Button.ButtonAccessibleObject.cs
@@ -17,8 +17,6 @@ namespace System.Windows.Forms
             internal override object? GetPropertyValue(UIA propertyID)
                 => propertyID switch
                 {
-                    UIA.AutomationIdPropertyId
-                        => Owner.Name,
                     UIA.ControlTypePropertyId
                         // If we don't set a default role for Button and ButtonBase accessible objects
                         // it will be retrieved from Windows.
--- a/src/System.Windows.Forms/src/System/Windows/Forms/CheckBox.CheckBoxAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/CheckBox.CheckBoxAccessibleObject.cs
@@ -70,8 +70,6 @@ namespace System.Windows.Forms
             internal override object? GetPropertyValue(UiaCore.UIA propertyID)
                 => propertyID switch
                 {
-                    UiaCore.UIA.AutomationIdPropertyId
-                        => Owner.Name,
                     UiaCore.UIA.IsKeyboardFocusablePropertyId
                         =>
                         // This is necessary for compatibility with MSAA proxy:
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxChildListUiaProvider.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxChildListUiaProvider.cs
@@ -27,6 +27,8 @@ namespace System.Windows.Forms
                 _childListControlhandle = childListControlhandle;
             }
 
+            private protected override string AutomationId => COMBO_BOX_LIST_AUTOMATION_ID;
+
             internal override Rectangle BoundingRectangle => User32.GetWindowRect(_owningComboBox.GetListNativeWindow());
 
             /// <summary>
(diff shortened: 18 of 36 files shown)


src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ButtonBase.ButtonBaseAccessibleObjectTests.cs
```diff
@@ -235,6 +235,17 @@ public void ButtonBaseAccessibleObject_GetPropertyValue_ControlType_IsExpected_F
             Assert.False(buttonBase.IsHandleCreated);
         }
 
+        [WinFormsFact]
+        public void ButtonBaseAccessibleObject_GetPropertyValue_AutomationId_ReturnsExpected()
+        {
+            using var ownerControl = new SubButtonBase();
```
RussKie, line 241: The codebase was started before `new()` syntax was available, however now this is the preferred way when writing the new code:
```suggestion
            using SubButtonBase ownerControl = new();
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Control.ControlAccessibleObjectTests.cs
```diff
@@ -1555,6 +1555,17 @@ public void ControlAccessibleObject_FragmentNavigate_PreviousSibling_IsPreviousI
             Assert.True(toolStrip.IsHandleCreated);
         }
 
+        [WinFormsFact]
+        public void ControlAccessibleObject_GetPropertyValue_AutomationId_ReturnsExpected()
+        {
+            using Control ownerControl = new Control();
```
RussKie, line 1561:
```suggestion
            using Control ownerControl = new();
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/GroupBoxAccessibleObjectTests.cs
```diff
@@ -119,5 +119,16 @@ public void GroupBoxAccessibleObject_GetPropertyValue_ControlType_IsExpected_For
             Assert.Equal(expected, actual);
             Assert.False(groupBox.IsHandleCreated);
         }
+
+        [WinFormsFact]
+        public void GroupBoxAccessibleObject_GetPropertyValue_AutomationId_ReturnsExpected()
+        {
+            using GroupBox ownerControl = new GroupBox();
```
RussKie, line 126:
```suggestion
            using GroupBox ownerControl = new();
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PictureBox.PictureBoxAccessibleObjectTests.cs
```diff
@@ -107,7 +107,6 @@ public void PictureBoxAccessibleObject_GetPropertyValue_Invoke_ReturnsExpected(i
             object value = pictureBoxAccessibleObject.GetPropertyValue((UIA)propertyID);
 
             Assert.Equal(expected, value);
-            Assert.False(pictureBox.IsHandleCreated);
```
RussKie, line 110: Why was this check deleted? As a general rules accessing an AO must never force a owner control handle creation, and we want to ensure this rule is never broken.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TrackBar.TrackBarAccessibleObjectTests.cs
```diff
@@ -413,18 +413,22 @@ public void TrackBarAccessibilityObject_GetPropertyValue_RuntimeId_ReturnsExpect
...
             {
-                AccessibleName = name
+                Name = "TrackBar1",
+                AccessibleName = "TestName"
             };
-            object actual = trackBar.AccessibilityObject.GetPropertyValue(UiaCore.UIA.NamePropertyId);
 
-            Assert.Equal(name, actual);
             Assert.False(trackBar.IsHandleCreated);
+            TrackBar.TrackBarAccessibleObject accessibilityObject = new TrackBar.TrackBarAccessibleObject(trackBar);
```
RussKie, line 428:
```suggestion
            TrackBar.TrackBarAccessibleObject accessibilityObject = new(trackBar);
```

RussKie (commented):
LGTM

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PictureBox.PictureBoxAccessibleObjectTests.cs
dkazennov: > Why was this check deleted? As a general rules accessing an AO must never force a owner control handle creation, and we want to ensure this rule is never broken.

Assert.False(pictureBox.IsHandleCreated); was being put before var pictureBoxAccessibleObject = new PictureBox.PictureBoxAccessibleObject(pictureBox); (line 105) Should it be put at the end of the test?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ButtonBase.ButtonBaseAccessibleObjectTests.cs
dkazennov: All suggestions changed according to the preferred code style.

Tanya-Solyanik:
>Tests based on the old implementation of AutomationId could become obsolete

Please describe what inspect and AccessibilityInsights show in cases when we return null.

Tanya-Solyanik:
@vladimir-krestov, @dmitrii-drobotov  - please review this change

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/LabelAccessibleObjectTests.cs
```diff
@@ -9,20 +9,23 @@ namespace System.Windows.Forms.Tests
 {
     public class LabelAccessibleObjectTests
     {
-        [WinFormsFact]
-        public void LabelAccessibleObject_GetPropertyValue_Name_ReturnsExpected()
+        [WinFormsTheory]
+        [InlineData((int)Interop.UiaCore.UIA.NamePropertyId, "Address")]
+        [InlineData((int)Interop.UiaCore.UIA.AutomationIdPropertyId, "Label1")]
```
Tanya-Solyanik, line 14: Nice!

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItem.ListViewItemBaseAccessibleObjectTests.cs
```diff
@@ -336,5 +336,18 @@ public void ListViewItemBaseAccessibleObject_Toggle_WorksExpected(bool isChecked
             Assert.Equal((UiaCore.ToggleState)expected, accessibleObject.ToggleState);
             Assert.False(control.IsHandleCreated);
         }
+
+        [WinFormsFact]
+        public void ListViewItemBaseAccessibleObject_GetPropertyValue_AutomationId_ReturnsExpected()
+        {
+            using ListView control = new();
```
Tanya-Solyanik, line 343: Consider `control` -> `listView`

dkazennov:
> > Tests based on the old implementation of AutomationId could become obsolete
> 
> Please describe what inspect and AccessibilityInsights show in cases when we return null for automation ID. This fix changes behavior of ControlAccessibleObject. Previously it was returning a null ID and now it returns Name. I wonder how this is manifested in accessibility tools.

A description was being put into PR text description.
"If AutomationId is null then Inspect will display no AutomationId property at all. It shows the property and the value if the field is overridden and is not null (a normal case)."
I'll check this with AccessibilityInsights as well.

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxChildEditUiaProvider.cs
```diff
@@ -15,6 +15,7 @@ public partial class ComboBox
         internal class ComboBoxChildEditUiaProvider : ChildAccessibleObject
         {
             private const string COMBO_BOX_EDIT_AUTOMATION_ID = "1001";
+            private protected override string AutomationId => COMBO_BOX_EDIT_AUTOMATION_ID;
```
Tanya-Solyanik, line 18: We try to place properties after the constructor, and constructor after the field definitions. Could you please move this property down?

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.TabAccessibleObject.cs
```diff
@@ -12,6 +12,7 @@ public partial class TabPage
     {
         internal class TabAccessibleObject : AccessibleObject
         {
+            private protected override string AutomationId => _owningTabPage.Name;
```
Tanya-Solyanik, line 15: nit: I would put AutomationId after the constructor, and owner too

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/GridEntry.GridEntryAccessibleObject.cs
```diff
@@ -12,6 +12,8 @@ internal abstract partial class GridEntry
     {
         public class GridEntryAccessibleObject : AccessibleObject
         {
+            private protected override string AutomationId => GetHashCode().ToString();
```
Tanya-Solyanik, line 15: nit: move to after the constructor

Tanya-Solyanik (commented):
Looks good!. I added some nits and am waiting for a sign off from @dmitrii-drobotov

dkazennov:
> * You can also remove `UiaCore.UIA.AutomationIdPropertyId => Owner.Name` in `PrintPreviewControlAccessibleObject` and `SplitContainerAccessibleObject`

These classes introduced after this branch were created (next time I will use cherry-pick to take changes of an old commit to a new branch created from most recent main). I propose we will create a new issue regarding these two classes?

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxChildEditUiaProvider.cs
dkazennov: Fixed.

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.TabAccessibleObject.cs
dkazennov: Fixed.

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/GridEntry.GridEntryAccessibleObject.cs
dkazennov: Fixed.

src/System.Windows.Forms/src/System/Windows/Forms/AccessibleObject.cs
```diff
@@ -391,6 +391,8 @@ internal virtual int ProviderOptions
 
         internal virtual UiaCore.IRawElementProviderSimple? HostRawElementProvider => null;
 
+        private protected virtual string? AutomationId => null;
```
vladimir-krestov, line 394: Please move it to be in alphabet order

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxChildEditUiaProvider.cs
```diff
@@ -33,6 +31,9 @@ public ComboBoxChildEditUiaProvider(ComboBox owner, IntPtr childEditControlhandl
                 UseTextProviders(_textProvider, _textProvider);
             }
 
+            private const string COMBO_BOX_EDIT_AUTOMATION_ID = "1001";
```
vladimir-krestov, line 34: I would prefer to keep it as before, because it's a private const, that should be in the beginning of the class

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.TabAccessibleObject.cs
```diff
@@ -12,15 +12,16 @@ public partial class TabPage
...
-
             private readonly TabPage _owningTabPage;
 
             public TabAccessibleObject(TabPage owningTabPage)
             {
                 _owningTabPage = owningTabPage.OrThrowIfNull();
             }
 
+            private protected override string AutomationId => _owningTabPage.Name;
+            private TabControl? OwningTabControl => _owningTabPage.ParentInternal as TabControl;
```
vladimir-krestov, line 23: Please move `OwningTabControl` to be in alphabet order

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjectTests.cs
```diff
@@ -2568,12 +2568,14 @@ public unsafe void TestAccessibleObjectFromPoint_ComboBox()
             Assert.Equal(7, parent.accChildCount);
         }
 
-        [WinFormsFact]
-        public void AccessibleObject_GetPropertyValue_ControlTypeProperty_ReturnsNull()
+        [WinFormsTheory]
+        [InlineData((int)UiaCore.UIA.ControlTypePropertyId)]
+        [InlineData((int)UiaCore.UIA.AutomationIdPropertyId)]
+        public void AccessibleObject_GetPropertyValue_ReturnsNull(int propertyId)
```
vladimir-krestov, line 2574: I don't like this name, because `GetPropertyValue` returns `null` not for all cases. You can rename like that, for example.
```suggestion
        public void AccessibleObject_GetPropertyValue_ReturnsNull_ForSeveralProperties(int propertyId)
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ButtonBase.ButtonBaseAccessibleObjectTests.cs
```diff
@@ -235,6 +235,17 @@ public void ButtonBaseAccessibleObject_GetPropertyValue_ControlType_IsExpected_F
...
 
+        [WinFormsFact]
+        public void ButtonBaseAccessibleObject_GetPropertyValue_AutomationId_ReturnsExpected()
+        {
+            using SubButtonBase ownerControl = new();
+            ButtonBase.ButtonBaseAccessibleObject accessibleObject = new(ownerControl);
+            string expected = accessibleObject.Owner.Name;
+
+            Assert.Equal(accessibleObject.GetPropertyValue(UiaCore.UIA.AutomationIdPropertyId), expected);
+            Assert.False(ownerControl.IsHandleCreated);
```
vladimir-krestov, line 246:
```suggestion
        public void ButtonBaseAccessibleObject_GetPropertyValue_AutomationId_ReturnsExpected()
        {
            using SubButtonBase ownerControl = new() {Name = "test name"};
            string expected = ownerControl.Name;
            object actual = ownerControl.AccessibleObject.GetPropertyValue(UiaCore.UIA.AutomationIdPropertyId);
            
            Assert.Equal(expected, actual);
            Assert.False(ownerControl.IsHandleCreated);
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Control.ControlAccessibleObjectTests.cs
```diff
@@ -1555,6 +1555,17 @@ public void ControlAccessibleObject_FragmentNavigate_PreviousSibling_IsPreviousI
...
+        [WinFormsFact]
+        public void ControlAccessibleObject_GetPropertyValue_AutomationId_ReturnsExpected()
+        {
+            using Control ownerControl = new();
+            Control.ControlAccessibleObject accessibleObject = new Control.ControlAccessibleObject(ownerControl);
+            string expected = accessibleObject.Owner.Name;
+
+            Assert.Equal(accessibleObject.GetPropertyValue(UiaCore.UIA.AutomationIdPropertyId), expected);
+            Assert.False(ownerControl.IsHandleCreated);
+        }
```
vladimir-krestov, line 1567:
```suggestion
        public void ControlAccessibleObject_GetPropertyValue_AutomationId_ReturnsExpected()
        {
            using Control ownerControl = new() {Name = "test name"};
            string expected = ownerControl.Name;
            object actual = ownerControl.AccessibleObject.GetPropertyValue(UiaCore.UIA.AutomationIdPropertyId);
            
            Assert.Equal(expected, actual);
            Assert.False(ownerControl.IsHandleCreated);
        }
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewCellAccessibleObjectTests.cs
```diff
@@ -995,6 +995,24 @@ public void DataGridViewCellAccessibleObject_Column_ReturnExpected_IfCustomOrder
...
+            dataGridView.Rows.Add(new DataGridViewRow());
+
+            DataGridViewCellAccessibleObject dataGridViewCellAccessibleObject = new(dataGridView.Rows[0].Cells[0]);
+            string expected = string.Empty;
+            foreach (int runtimeIdPart in dataGridViewCellAccessibleObject.RuntimeId)
+            {
+                expected += runtimeIdPart.ToString();
+            }
+
+            Assert.Equal(dataGridViewCellAccessibleObject.GetPropertyValue(UiaCore.UIA.AutomationIdPropertyId), expected);
```
vladimir-krestov, line 1012: Replace these arguments, `expected` should be first (check the Equal method inside)

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/GroupBoxAccessibleObjectTests.cs
```diff
@@ -119,5 +119,16 @@ public void GroupBoxAccessibleObject_GetPropertyValue_ControlType_IsExpected_For
             Assert.Equal(expected, actual);
             Assert.False(groupBox.IsHandleCreated);
         }
+
+        [WinFormsFact]
+        public void GroupBoxAccessibleObject_GetPropertyValue_AutomationId_ReturnsExpected()
```
vladimir-krestov, line 124: Rework this test and next like I proposed before as well

vladimir-krestov (commented):
Looks great

vladimir-krestov:
> > > Tests based on the old implementation of AutomationId could become obsolete
> > 
> > 
> > Please describe what inspect and AccessibilityInsights show in cases when we return null for automation ID. This fix changes behavior of ControlAccessibleObject. Previously it was returning a null ID and now it returns Name. I wonder how this is manifested in accessibility tools.
> 
> @Tanya-Solyanik A description was being put into PR text description. "If AutomationId is null then Inspect will display no AutomationId property at all. It shows the property and the value if the field is overridden and is not null (a normal case)." I'll check this with AccessibilityInsights as well.

I'm interesting, what happens with the rest controls, that have "string.Empty" now for `AutomationId`?
What AI says?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjectTests.cs
dkazennov: I believe that AccessibleObject_GetPropertyValue_ReturnsNull_IfExpected should be good.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Control.ControlAccessibleObjectTests.cs
dkazennov: Fixed in ControlAccessibleObjectTest, ButtonBaseAccessibleObject and GroupBoxAccessibleObject.

dkazennov:
This commit was being rebased on the most recent main. Conflicts were being resolved.

dmitrii-drobotov (dismissed):
LGTM👍

dkazennov:
> > > > Tests based on the old implementation of AutomationId could become obsolete
> > > 
> > > 
> > > Please describe what inspect and AccessibilityInsights show in cases when we return null for automation ID. This fix changes behavior of ControlAccessibleObject. Previously it was returning a null ID and now it returns Name. I wonder how this is manifested in accessibility tools.
> > 
> > 
> > @Tanya-Solyanik A description was being put into PR text description. "If AutomationId is null then Inspect will display no AutomationId property at all. It shows the property and the value if the field is overridden and is not null (a normal case)." I'll check this with AccessibilityInsights as well.
> 
> I'm interesting, what happens with the rest controls, that have "string.Empty" now for `AutomationId`? What AI says?

If AutomationId is null is is not being shown by default in AccessibilityInsights. If shown it has description "Property Doesn't exist". Same result if AutomationId is string.Empty.

image

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxChildEditUiaProvider.cs
RussKie: https://docs.microsoft.com/en-us/archive/blogs/brada/internal-coding-guidelines#28-file-organization

RussKie (commented):
👍

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ButtonBase.ButtonBaseAccessibleObjectTests.cs
```diff
@@ -235,6 +235,17 @@ public void ButtonBaseAccessibleObject_GetPropertyValue_ControlType_IsExpected_F
...
         }
 
+        [WinFormsFact]
+        public void ButtonBaseAccessibleObject_GetPropertyValue_AutomationId_ReturnsExpected()
+        {
+            using SubButtonBase ownerControl = new() { Name = "test name" };
+            string expected = ownerControl.Name;
+            object actual = ownerControl.AccessibilityObject.GetPropertyValue(UiaCore.UIA.AutomationIdPropertyId);
+
+            Assert.Equal(actual, expected);
```
Tanya-Solyanik, line 245: If you add parameter name, this would look `Assert.Equal(expected: actual, actual: expected); `  Please swap parameters.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Control.ControlAccessibleObjectTests.cs
```diff
@@ -1568,6 +1568,17 @@ public void ControlAccessibleObject_FragmentNavigate_PreviousSibling_IsPreviousI
...
         }
 
+        [WinFormsFact]
+        public void ControlAccessibleObject_GetPropertyValue_AutomationId_ReturnsExpected()
+        {
+            using Control ownerControl = new() { Name = "test name" };
+            string expected = ownerControl.Name;
+            object actual = ownerControl.AccessibilityObject.GetPropertyValue(UiaCore.UIA.AutomationIdPropertyId);
+
+            Assert.Equal(actual, expected);
```
Tanya-Solyanik, line 1578: Please swap parameters.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewCellAccessibleObjectTests.cs
```diff
@@ -995,6 +995,24 @@ public void DataGridViewCellAccessibleObject_Column_ReturnExpected_IfCustomOrder
...
+            dataGridView.Rows.Add(new DataGridViewRow());
+
+            DataGridViewCellAccessibleObject dataGridViewCellAccessibleObject = new(dataGridView.Rows[0].Cells[0]);
+            string expected = string.Empty;
+            foreach (int runtimeIdPart in dataGridViewCellAccessibleObject.RuntimeId)
+            {
+                expected += runtimeIdPart.ToString();
+            }
+
+            Assert.Equal(expected, dataGridViewCellAccessibleObject.GetPropertyValue(UiaCore.UIA.AutomationIdPropertyId));
```
Tanya-Solyanik, line 1012: Please swap parameters

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/GroupBoxAccessibleObjectTests.cs
```diff
@@ -119,5 +119,16 @@ public void GroupBoxAccessibleObject_GetPropertyValue_ControlType_IsExpected_For
...
+        {
+            using GroupBox ownerControl = new() { Name = "test name" };
+            string expected = ownerControl.Name;
+            object actual = ownerControl.AccessibilityObject.GetPropertyValue(UiaCore.UIA.AutomationIdPropertyId);
+
+            Assert.Equal(actual, expected);
```
Tanya-Solyanik, line 130: Please swap parameters

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItem.ListViewItemBaseAccessibleObjectTests.cs
```diff
@@ -336,5 +336,18 @@ public void ListViewItemBaseAccessibleObject_Toggle_WorksExpected(bool isChecked
...
+        [WinFormsFact]
+        public void ListViewItemBaseAccessibleObject_GetPropertyValue_AutomationId_ReturnsExpected()
+        {
+            using ListView listView = new();
+            ListViewItem item = new();
+            listView.Items.Add(item);
+            ListViewItemBaseAccessibleObject accessibleObject = (ListViewItemBaseAccessibleObject)item.AccessibilityObject;
+
+            var expected = string.Format("{0}-{1}", typeof(ListViewItem).Name, accessibleObject.CurrentIndex);
+            Assert.Equal(accessibleObject.GetPropertyValue(UiaCore.UIA.AutomationIdPropertyId), expected);
```
Tanya-Solyanik, line 349: Please swap parameters

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItem.ListViewItemBaseAccessibleObjectTests.cs
```diff
@@ -336,5 +336,18 @@ public void ListViewItemBaseAccessibleObject_Toggle_WorksExpected(bool isChecked
...
             Assert.False(control.IsHandleCreated);
         }
+
+        [WinFormsFact]
+        public void ListViewItemBaseAccessibleObject_GetPropertyValue_AutomationId_ReturnsExpected()
+        {
+            using ListView listView = new();
+            ListViewItem item = new();
+            listView.Items.Add(item);
+            ListViewItemBaseAccessibleObject accessibleObject = (ListViewItemBaseAccessibleObject)item.AccessibilityObject;
```
Tanya-Solyanik, line 346: It's appropriate to use var in this case because of the explicit cast.
```suggestion
            var accessibleObject = (ListViewItemBaseAccessibleObject)item.AccessibilityObject;
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TabControl.TabControlAccessibleObjectTests.cs
```diff
@@ -744,18 +744,21 @@ public void TabControlAccessibleObject_GetPropertyValue_IsEnabled_ReturnsExpecte
...
-            const string name = "Test name";
             using TabControl tabControl = new()
             {
-                AccessibleName = name
+                Name = "TabControl1",
+                AccessibleName = "TestName"
             };
 
-            object actual = tabControl.AccessibilityObject.GetPropertyValue(UiaCore.UIA.NamePropertyId);
+            TabControlAccessibleObject accessibilityObject = new TabControlAccessibleObject(tabControl);
```
Tanya-Solyanik, line 758: consider using `var` or 
```
TabControlAccessibleObject accessibilityObject = new(tabControl);
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TabPage.TabPageAccessibleObjectTests.cs
```diff
@@ -479,18 +479,21 @@ public void TabPageAccessibleObject_GetPropertyValue_AccessKey_ReturnExpected(bo
...
-            const string name = "Test name";
             using TabPage tabPage = new()
             {
-                AccessibleName = name
+                Name = "TabPage1",
+                AccessibleName = "TestName"
             };
 
-            object actual = tabPage.AccessibilityObject.GetPropertyValue(UIA.NamePropertyId);
+            TabPageAccessibleObject accessibilityObject = new TabPageAccessibleObject(tabPage);
```
Tanya-Solyanik, line 493: Consider using `var` or `new(tabPage);` to match other tests in this file.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TrackBar.TrackBarAccessibleObjectTests.cs
```diff
@@ -413,17 +413,21 @@ public void TrackBarAccessibilityObject_GetPropertyValue_RuntimeId_ReturnsExpect
...
+        [WinFormsTheory]
+        [InlineData((int)UiaCore.UIA.NamePropertyId, "TestName")]
+        [InlineData((int)UiaCore.UIA.AutomationIdPropertyId, "TrackBar1")]
+        public void TrackBarAccessibilityObject_GetPropertyValue_Invoke_ReturnsExpected(int propertyID, object expected)
         {
-            const string name = "Test name";
             using TrackBar trackBar = new()
             {
-                AccessibleName = name
+                Name = "TrackBar1",
```
Tanya-Solyanik, line 423: Consider:
```
Name = "TestControlName",
AccessibleName  = "TestAccessibleName"
```

Tanya-Solyanik (approved):
I added some stylistic nits.

Tanya-Solyanik:
>If AutomationId is null is is not being shown by default in AccessibilityInsights. If shown it has description "Property Doesn't exist". Same result if AutomationId is string.Empty.

@dkazennov  - Could you please double-check this result? I see that  `null` results in Automation ID property being looked up by the native Accessible object, which results in the window text most commonly. Empty string results in "property not defined". So the change is that Inheritors of ControlsAccessibleObjects that were not providing a custom AutomationId property, before the change were seeing ID provided by the MSAA proxy and after the change they will see `Owner.Name`

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewCellAccessibleObjectTests.cs
dkazennov: > Please swap parameters
Is in incorrect? `expected` comes as a first argument, `actual` as a second one.

dkazennov:
> > If AutomationId is null is is not being shown by default in AccessibilityInsights. If shown it has description "Property Doesn't exist". Same result if AutomationId is string.Empty.
> 
> @dkazennov - Could you please double-check this result? I see that `null` results in Automation ID property being looked up by the native Accessible object, which results in the window text most commonly. Empty string results in "property not defined". So the change is that Inheritors of ControlsAccessibleObjects that were not providing a custom AutomationId property, before the change were seeing ID provided by the MSAA proxy and after the change they will see `Owner.Name`

**Case 1. class ListViewItemBaseAccessibleObject : AccessibleObject**

I've tested ListViewItemBaseAccessibleObject with AccessibilityInsight. ListViewItemBaseAccessibleObject has base class AccessibleObject. 
AccessibleObject has property `private protected virtual string? AutomationId => null;`
ListViewItemBaseAccessibleObject has porperty `private protected override string AutomationId => string.Format("{0}-{1}", typeof(ListViewItem).Name, CurrentIndex);`

If I change `ListViewItemBaseAccessibleObject.AutomationId` property to return **null**, AccessibilityInsight will show `Property does not exist`
If I change `ListViewItemBaseAccessibleObject.AutomationId` property to return **string.Empty**, AccessibilityInsight will show `Property does not exist`
If I change `ListViewItemBaseAccessibleObject.AutomationId` property to return **"TEST VALUE STRING"**, AccessibilityInsight will show `"TEST VALUE STRING"`

image

**Case 2. class ButtonBaseAccessibleObject : ControlAccessibleObject**

ControlAccessibleObject has property `private protected override string AutomationId => Owner.Name;`
ButtonBaseAccessibleObject has no `AutomationId` override property.

If tested without override property AutomationId	, AutomationId will show `Property does not exist` (probably because in WinformsConrolsTest `ControlAccessibleObject.Owner.Name` is string.Empty
If tested with override property `Button.AutomationId` = "TEST VALUE STRING", AutomationId will show "TEST VALUE STRING"
If I change ControlAccessibleObject property to `private protected override string AutomationId => "TEST VALUE STRING";` AutomationId shows "TEST VALUE STRING".

image

So classes with base class `ControlAccessibleObject` and without overridden property `AutomationId`  will return ControlAccessibleObject.AutomationId. Which is Owner.Name (wherever it is string.Empty or some value);

null or string.Empty returns  `Property does not exist`

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TabPage.TabPageAccessibleObjectTests.cs
```diff
@@ -479,18 +479,21 @@ public void TabPageAccessibleObject_GetPropertyValue_AccessKey_ReturnExpected(bo
...
-            const string name = "Test name";
             using TabPage tabPage = new()
             {
-                AccessibleName = name
+                Name = "TabPage1",
+                AccessibleName = "TestName"
             };
 
-            object actual = tabPage.AccessibilityObject.GetPropertyValue(UIA.NamePropertyId);
+            TabPageAccessibleObject accessibilityObject = new(tabPage);
```
Tanya-Solyanik, line 493: It will be close to real live scenario to get Accessible object from the control rather than new up one.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewCellAccessibleObjectTests.cs
Tanya-Solyanik: You are right!

Tanya-Solyanik (commented):
Added some cleanup-y comments

dkazennov:
> > null or string.Empty returns Property does not exist
> 
> Had you rebuilt the winforms dll properly when testing this?

I've rebuild the solution, commented out the overridden `AutomationId` property in `ControlAccessibleObject` (`ControlAccessibleObject.AutomationId` returns Owner.Name and it's base `AccessibleObject.AutomationId` returns null) and started the WinformsControlsTest. I checked button control (it has no overridden property and therefore returns null) and AutomationId in both Ispector and AccessibiltyInsights was a number as shown on a screenshot.

image

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/LabelAccessibleObjectTests.cs
```diff
@@ -9,20 +9,23 @@ namespace System.Windows.Forms.Tests
...
-            AccessibleObject labelAccessibleObject = label.AccessibilityObject;
+            using Label label = new()
+            {
+                Text = "Some test label text",
+                Name = "Label1",
+                AccessibleName = "Address"
+            };
 
-            Assert.False(label.IsHandleCreated);
+            Label.LabelAccessibleObject accessibilityObject = new Label.LabelAccessibleObject(label);
```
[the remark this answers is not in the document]
dkazennov: > Why are you not getting Accessible object from your `label` under test? `label.AccessibleObject`?

Oh, I think I've missed it since the last implementation of the test.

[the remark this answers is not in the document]
Tanya-Solyanik: >Btw, all existing tests in LabelAccessibleObjectTests.cs get AccessibleObject thought the constructor LabelAccessibleObject(label); Should it be changed as well?

Great find! yes please do. Feel free to do it either in this or the next PR.

Tanya-Solyanik:
> I checked button control (it has no overridden property and therefore returns null) and AutomationId in both Ispector and AccessibiltyInsights was a number as shown on a screenshot.

Thank you for the experiment. This result matches the pattern that when managed accessible object return null for aq property, UIA reads the value from the default implementation in MSAA proxy.

dkazennov:
> > I checked button control (it has no overridden property and therefore returns null) and AutomationId in both Ispector and AccessibiltyInsights was a number as shown on a screenshot.
> 
> Thank you for the experiment. This result matches the pattern that when managed accessible object return null for aq property, UIA reads the value from the default implementation in MSAA proxy.

@Tanya-Solyanik 
Should there be any changes in this fix then? Or does only concerns documentation?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/LabelAccessibleObjectTests.cs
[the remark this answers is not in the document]
dkazennov: > Great find! yes please do. Feel free to do it either in this or the next PR.
Since there is a CTI testing already requested I believe we do this in next PR.

Tanya-Solyanik:
>Should there be any changes in this fix then? Or does only concerns documentation?

I think this is an acceptable break as long as it is documented.

RussKie (commented):
👍

RussKie:
@Tanya-Solyanik I see you applied "Documentation breaking" label. Could you please elaborate what's breaking here?

dkazennov:
@Tanya-Solyanik
CTI testing confirmed our results regarding  the pattern that when managed accessible object return null for aq property, UIA reads the value from the default implementation in MSAA proxy.

Tanya-Solyanik:
> Could you please elaborate what's breaking here?

We are returning different value for AutomationID property. Exactly how different, is not clear as we are getting conflicting results.

Tanya-Solyanik:
@dkazennov - CTI test results are not quite what I expected. I'd like to understand why ComboBox.Name=null results in AutomationID as "property not available" instead of a numeric value.

dkazennov:
> @dkazennov - CTI test results are not quite what I expected. I'd like to understand why ComboBox.Name=null results in AutomationID as "property not available" instead of a numeric value.

I see. I'll try to figure it out than.

RussKie:
> Winforms code does not expose this property to be easily overridden directly.

Ah, yes, it vaguely rings true for me. We had to subclass a type and its AccessibleObject to override the accessibility values: https://github.com/gitextensions/gitextensions/pull/9887#issuecomment-1063995067

dkazennov:
@Tanya-Solyanik 
I have found that in WinformsControlsTests ComboBoxAccessibleObject (inherits ControlAccessibleObject) and ButtonBaseAccessibleObject (inherits ControlAccessibleObject) behave in the same way.

1. If the property `private protected override string AutomationId => Owner.Name;` is present in `ControlAccessibleObject` and controls have no Names (commented out `this.comboBox2.Name = "comboBox2"` in `ComboBoxes.Designer.cs`) then Insights will show ComboBoxAccessibleObject.AutomationId and ButtonBaseAccessibleObject.AutomationId as 'Property does not exist'.

2.  If the property `private protected override string AutomationId => Owner.Name;` is commented out in `ControlAccessibleObject` and therefore if `private protected virtual string? AutomationId => null` in `AccessibleObject` class plays its role, then Insights will show ComboBoxAccessibleObject.AutomationId and ButtonBaseAccessibleObject.AutomationId as numeric values based on hashcode.

3.  If the property `private protected override string AutomationId => Owner.Name;` is replaced with `private protected override string AutomationId => null;` in `ControlAccessibleObject` then Insights will show ComboBoxAccessibleObject.AutomationId and ButtonBaseAccessibleObject.AutomationId as numeric values based on hashcode.

4. If Owner.Name is set to null or string.Empty in `ControlAccessibleObject.` constructorthen Insights will show ComboBoxAccessibleObject.AutomationId and ButtonBaseAccessibleObject.AutomationId as 'Property does not exist'.

So the exact source of the issue is the new property `private protected override string AutomationId` in `ControlAccessibleObject`. If the property returns null, it is being show as a numeric value based of hash.

image

RussKie:
> 2\. If the property `private protected override string AutomationId => Owner.Name;` is commented out in `ControlAccessibleObject` and therefore if `private protected virtual string? AutomationId => null` in `AccessibleObject` class plays its role, then Insights will show ComboBoxAccessibleObject.AutomationId and ButtonBaseAccessibleObject.AutomationId as numeric values based on hashcode.

Do we have tests for this?

dkazennov:
> > 2. If the property `private protected override string AutomationId => Owner.Name;` is commented out in `ControlAccessibleObject` and therefore if `private protected virtual string? AutomationId => null` in `AccessibleObject` class plays its role, then Insights will show ComboBoxAccessibleObject.AutomationId and ButtonBaseAccessibleObject.AutomationId as numeric values based on hashcode.
> 
> Do we have tests for this?

No, we don't. Is it something to be tested with a unit test?

dkazennov:
**Conclusion:**

If the property `AutomationId` returns `null` (but not `Owner.Name` even if `Owner.Name` is null or empty) then Accessibility Tools will show AutomationId (e.g. ComboBoxAccessibleObject.AutomationId and ButtonBaseAccessibleObject.AutomationId) as numeric values based on a hashcode.

So on the Windows Forms layer the value is null (in all concluded unit tests Assert.Null(actual AutomationId) would be true ), but on the MSAA Proxy level (between MSAA Proxy and Accessibility Tools) this value is being replaced by a native hashcode.

dkazennov:
Rebased on the most recent main branch. 
Implemented changes to WebBrowser.WebBrowserAccessibleObject from recent PR https://github.com/dotnet/winforms/pull/7314

Tanya-Solyanik (approved):
Thank you!

PR #7220, opened 2022-05-26
Removed build warning in VS

Tanya-Solyanik:
Can't build in the latest IntPreview Dev17 without this change.
`unchecked` - fixes a build break, `#pragma` suppression is only a warning.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7220)

### DIFF
--- a/src/System.Windows.Forms.Primitives/tests/UnitTests/Interop/Oleaut32/VARIANTTests.cs
+++ b/src/System.Windows.Forms.Primitives/tests/UnitTests/Interop/Oleaut32/VARIANTTests.cs
@@ -185,161 +185,161 @@ namespace System.Windows.Forms.Tests.Interop.Oleaut32
         {
             if (IntPtr.Size == 8)
             {
-                yield return new object[] { VARENUM.I1, (IntPtr)long.MinValue, (sbyte)0 };
+                yield return new object[] { VARENUM.I1, unchecked((nint)long.MinValue), (sbyte)0 };
             }
 
-            yield return new object[] { VARENUM.I1, (IntPtr)int.MinValue, (sbyte)0 };
-            yield return new object[] { VARENUM.I1, (IntPtr)short.MinValue, (sbyte)0 };
-            yield return new object[] { VARENUM.I1, (IntPtr)sbyte.MinValue, sbyte.MinValue };
-            yield return new object[] { VARENUM.I1, (IntPtr)(-10), (sbyte)(-10) };
-            yield return new object[] { VARENUM.I1, (IntPtr)0, (sbyte)0 };
-            yield return new object[] { VARENUM.I1, (IntPtr)10, (sbyte)10 };
-            yield return new object[] { VARENUM.I1, (IntPtr)sbyte.MaxValue, sbyte.MaxValue };
-            yield return new object[] { VARENUM.I1, (IntPtr)short.MaxValue, (sbyte)(-1) };
-            yield return new object[] { VARENUM.I1, (IntPtr)int.MaxValue, (sbyte)(-1) };
+            yield return new object[] { VARENUM.I1, (nint)int.MinValue, (sbyte)0 };
+            yield return new object[] { VARENUM.I1, (nint)short.MinValue, (sbyte)0 };
+            yield return new object[] { VARENUM.I1, (nint)sbyte.MinValue, sbyte.MinValue };
+            yield return new object[] { VARENUM.I1, (nint)(-10), (sbyte)(-10) };
+            yield return new object[] { VARENUM.I1, (nint)0, (sbyte)0 };
+            yield return new object[] { VARENUM.I1, (nint)10, (sbyte)10 };
+            yield return new object[] { VARENUM.I1, (nint)sbyte.MaxValue, sbyte.MaxValue };
+            yield return new object[] { VARENUM.I1, (nint)short.MaxValue, (sbyte)(-1) };
+            yield return new object[] { VARENUM.I1, (nint)int.MaxValue, (sbyte)(-1) };
             if (IntPtr.Size == 8)
             {
-                yield return new object[] { VARENUM.I1, (IntPtr)long.MaxValue, (sbyte)(-1) };
+                yield return new object[] { VARENUM.I1, unchecked((nint)long.MaxValue), (sbyte)(-1) };
             }
 
-            yield return new object[] { VARENUM.UI1, (IntPtr)(-10), (byte)246 };
-            yield return new object[] { VARENUM.UI1, (IntPtr)0, (byte)0 };
-            yield return new object[] { VARENUM.UI1, (IntPtr)10, (byte)10 };
-            yield return new object[] { VARENUM.UI1, (IntPtr)byte.MaxValue, byte.MaxValue };
-            yield return new object[] { VARENUM.UI1, (IntPtr)ushort.MaxValue, byte.MaxValue };
+            yield return new object[] { VARENUM.UI1, (nint)(-10), (byte)246 };
+            yield return new object[] { VARENUM.UI1, (nint)0, (byte)0 };
+            yield return new object[] { VARENUM.UI1, (nint)10, (byte)10 };
+            yield return new object[] { VARENUM.UI1, (nint)byte.MaxValue, byte.MaxValue };
+            yield return new object[] { VARENUM.UI1, (nint)ushort.MaxValue, byte.MaxValue };
             if (IntPtr.Size == 8)
             {
-                yield return new object[] { VARENUM.UI1, (IntPtr)uint.MaxValue, byte.MaxValue };
+                yield return new object[] { VARENUM.UI1, unchecked((nint)uint.MaxValue), byte.MaxValue };
             }
 
-            yield return new object[] { VARENUM.UI1, (IntPtr)(-1), byte.MaxValue };
-
-            if (IntPtr.Size == 8)
-            {
-                yield return new object[] { VARENUM.I2, (IntPtr)long.MinValue, (short)0 };
-            }
-
-            yield return new object[] { VARENUM.I2, (IntPtr)int.MinValue, (short)0 };
-            yield return new object[] { VARENUM.I2, (IntPtr)short.MinValue, short.MinValue };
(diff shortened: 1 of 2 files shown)


RussKie (dismissed):

Tanya-Solyanik:
Looks like https://github.com/dotnet/arcade/issues/4337 is not as fixed as we expected it to be. Reverting that commit.

src/System.Windows.Forms.Primitives/tests/UnitTests/Interop/Oleaut32/VARIANTTests.cs
```diff
@@ -498,8 +498,10 @@ public static IEnumerable<object[]> ToObject_UI8_TestData()
             yield return new object[] { (IntPtr)ushort.MaxValue, (ulong)ushort.MaxValue };
             if (IntPtr.Size == 8)
             {
+#pragma warning disable CS8778 // Constant value may overflow at runtime (use 'unchecked' syntax to override)
                 yield return new object[] { (IntPtr)uint.MaxValue, (ulong)uint.MaxValue };
-                yield return new object[] { (IntPtr)ulong.MaxValue, ulong.MaxValue };
+#pragma warning restore CS8778 // This code runs on x64 platform only.
+                yield return new object[] { unchecked((IntPtr)ulong.MaxValue), ulong.MaxValue };
```
JeremyKuhne, line 504: Fyi, this doesn't actually do anything. `unchecked` around casts to `IntPtr` don't affect anything as the cast is a method call that does its own `checked`. This particular case is even weirder as `ulong.MaxValue` is a constant that gets implicitly converted to an `int` before it gets passed to the cast method.

Bottom line- don't ever put `checked` or `unchecked` around `IntPtr` casts.

Tanya-Solyanik: This code was generated by the light-bulb suggestion

Tanya-Solyanik: It seems that the latest VS treats IntPtrs as nints. 

>This particular case is even weirder as ulong.MaxValue is a constant that gets implicitly converted to an int before it gets passed to the cast method.

Does this mean that this test input is the same as uint.MaxValue? For .NET Framework we do get to an int64 - conv.i8
```
        IL_0001: ldc.i4.m1
        IL_0002: conv.i8
        IL_0003: call native int [System.Runtime]System.IntPtr::op_Explicit(int64)
        IL_0008: stloc.0
```

Tanya-Solyanik: image

RussKie: I wonder if @tannergooding knows anything about this

tannergooding: This is by design and part of the “Numeric IntPtr” feature: https://github.com/dotnet/csharplang/issues/6065

For .NET 7 and later (or more specifically any compilation that defines `System.Runtime.CompilerServices.RuntimeFeature.NumericIntPtr` in the "core library", that is the library that defines object and has no dependencies) `nint` is to `IntPtr` as `int` is to `Int32`. That is, they are true aliases of each other and are identical to C#. IntPtr has all the operators, literals, and other language support and likewise nint no longer gets the NativeInteger attribute to avoid bloating metadata unnecessarily. Prior to .NET 7 these represented the same type in IL but the language would treat them differently with `IntPtr` not having access to most operators, literal support, constant folding, or other features that built-in integer types had support for.

Because of this change, when targeting .NET 7 or later code like this will now correctly error since the value would be too large to fit into an `nint` on 32-bit platform (it is notably also too large for `nint` on a 64-bit platform since its unsigned and `IntPtr`/`nint` are signed). Users will need to explicitly use `IntPtr p = unchecked((IntPtr)ulong.MaxValue)` or `nint p = unchecked((IntPtr)ulong.MaxValue)` to indicate they are aware of the possible truncation and definite “overflow” that will happen. In this particular case you could just do `IntPtr p = -1` or `nint p = -1` to get “all bits set” without needing `unchecked`.

This is a rare "breaking change" for C# in that it will raise new errors rather than opt-in warning waves or other behavior. However, it makes `IntPtr` and `UIntPtr` significantly more powerful, ensures the functionality is consistent with what is exposed/supported by IL, will help reduce errors in the long term, and ensures that what was a very leaky abstraction no longer leaks.

tannergooding: > If I'm understanding you right, recompiling when using IntPtr casts to variables will no longer be checked.

It will respect the surrounding checked or unchecked context including the `CheckForOverflowUnderflow` MSBuild property or corresponding compiler switch.

> Love that we're doing this, but it is quite a big breaking change.

It is certainly a break for users of `IntPtr` and `UIntPtr`. However, those types tend to be relegated to lowlevel and other interop code so the scope of the break is likely not to be "as broad" overall.

We will be providing an analyzer that helps flag potentially problematic cases and the compiler will error in other potentially problematic cases like the above.

### Used to Overflow

In particular the cases where the compiler will now error are `IntPtr x = (IntPtr)SomeIntegerConstantThatDoesNotFitInInt32` and `UIntPtr x = (UIntPtr)SomeIntegerConstantThatDoesNotFitInUInt32`.

The cases that the compiler won't error but the analyzer will catch are cases such as `IntPtr x = (IntPtr)someNonConstantUInt32Value`, `IntPtr x = (IntPtr)someNonConstantInt64Value`, `IntPtr x = (IntPtr)someNonConstantUInt64Value`. Likewise for `UIntPtr x = (UIntPtr)someNonConstantUInt64Value` and `UIntPtr x = (UIntPtr)someNonConstantSignedValue`.

### Can now Overflow

`IntPtr + int`, `IntPtr - int`, `UIntPtr + int`, and `UIntPtr - int` (yes it is `int`, not `uint`) may likewise now throw in a checked context. Previously it would silently truncate the result.

Tanya-Solyanik: >Switching to use the language keywords nint or nuint directly will allow the same code to be shared between .NET 7+ and .NET 6 and prior
Thank you!

However I don't see the same compile errors in a console application  - 
```
checked
{
    IntPtr p = (IntPtr)(ulong.MaxValue);
}
```
```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net7.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
    <Platforms>AnyCPU</Platforms>
  </PropertyGroup>

</Project>

```
In checked context, this overflows, but compiles successfully. Is there are feature switch?

tannergooding: You might be on an older version of the compiler or SDK. dotnet/arcade and dotnet/runtime tend to be ahead of the latest public previews.

If you pull the latest nightly build from https://github.com/dotnet/installer#installers-and-binaries it should repro locally as well.

src/System.Windows.Forms.Primitives/tests/UnitTests/Interop/PARAMTests.cs
```diff
@@ -40,85 +40,85 @@ public unsafe void FromLowHighUnsigned_x64_Result()
         [ConditionalFact(typeof(ArchitectureDetection), nameof(ArchitectureDetection.Is32bit))]
         public unsafe void LOWORD_x32_Result()
         {
-            Assert.Equal(0x0304, PARAM.LOWORD((IntPtr)unchecked((int)0x01020304)));
-            Assert.Equal(0xF3F4, PARAM.LOWORD((IntPtr)unchecked((int)0xF1F2F3F4)));
+            Assert.Equal(0x0304, PARAM.LOWORD((nint)(int)0x01020304));
+            Assert.Equal(0xF3F4, PARAM.LOWORD((nint)unchecked((int)0xF1F2F3F4)));
```
RussKie, line 44: Should these be?
```suggestion
            Assert.Equal(0x0304, PARAM.LOWORD((nint)0x01020304));
            Assert.Equal(0xF3F4, PARAM.LOWORD(unchecked((nint)0xF1F2F3F4)));
```

src/System.Windows.Forms.Primitives/tests/UnitTests/Interop/PARAMTests.cs
```diff
@@ -40,85 +40,85 @@ public unsafe void FromLowHighUnsigned_x64_Result()
...
         }
 
         [ConditionalFact(typeof(ArchitectureDetection), nameof(ArchitectureDetection.Is64bit))]
         public unsafe void LOWORD_x64_Result()
         {
-            Assert.Equal(0x0304, PARAM.LOWORD((IntPtr)unchecked((long)0x0506070801020304)));
-            Assert.Equal(0xF3F4, PARAM.LOWORD((IntPtr)unchecked((long)0xF5F6F7F8F1F2F3F4)));
+            Assert.Equal(0x0304, PARAM.LOWORD(unchecked((nint)(long)0x0506070801020304)));
+            Assert.Equal(0xF3F4, PARAM.LOWORD(unchecked((nint)unchecked((long)0xF5F6F7F8F1F2F3F4))));
```
RussKie, line 51: This looks like an overkill.... Should these be?
```suggestion
            Assert.Equal(0x0304, PARAM.LOWORD(unchecked((nint)0x0506070801020304)));
            Assert.Equal(0xF3F4, PARAM.LOWORD(unchecked((nint)0xF5F6F7F8F1F2F3F4)));
```

RussKie (commented):
👍 with few questions

RussKie (approved):

RussKie:
Thank you

PR #7223, opened 2022-05-26
revert Exception on accessing accessible object after control is disposed

Tanya-Solyanik:
This prevents exception in control disposal

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7223)


RussKie (commented):
It would be great to understand the scenario that we missed when we took the original change, and add a test for the scenario that led to this revert?

Tanya-Solyanik:
>It would be great to understand the scenario that we missed when we took the original change

On ListView disposal we remove all items from that list view, in response to that the native ListView moves focus from each removed item to the next, at this point items are disconnected thus causing this exception. We should ignore that focus move.

PR #7256, opened 2022-06-03
Add trimming annotations

kant2002:
Mostly make Clone methods works by applying PublicParameterlessConstructor and annotate EditType

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7256)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/System/TrimmingConstants.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/TrimmingConstants.cs
@@ -6,20 +6,20 @@ namespace System
 {
     internal static class TrimmingConstants
     {
-        internal const string FilterRequiresUnreferencedCodeMessage = "The public parameterless constructor or the 'Default' static field may be trimmed from the Attribute's Type.";
-
         internal const string AttributesRequiresUnreferencedCodeMessage = "Generic TypeConverters may require the generic types to be annotated. For example, NullableConverter requires the underlying type to be DynamicallyAccessedMembers All.";
 
-        internal const string PropertyDescriptorPropertyTypeMessage = "PropertyDescriptor's PropertyType cannot be statically discovered.";
+        internal const string BindingListViewFilterMessage = "Members of types used in the filter expression might be trimmed.";
 
         internal const string EditorRequiresUnreferencedCode = "Editors registered in TypeDescriptor.AddEditorTable may be trimmed.";
 
         internal const string EventDescriptorRequiresUnreferencedCodeMessage = "The built-in EventDescriptor implementation uses Reflection which requires unreferenced code.";        
 
-        internal const string TypeConverterGetPropertiesMessage = "The Type of value cannot be statically discovered. The public parameterless constructor or the 'Default' static field may be trimmed from the Attribute's Type.";
+        internal const string FilterRequiresUnreferencedCodeMessage = "The public parameterless constructor or the 'Default' static field may be trimmed from the Attribute's Type.";
 
-        internal const string BindingListViewFilterMessage = "Members of types used in the filter expression might be trimmed.";
+        internal const string PropertyDescriptorPropertyTypeMessage = "PropertyDescriptor's PropertyType cannot be statically discovered.";
 
         internal const string SiteNameMessage = "The Type of components in the container cannot be statically discovered to validate the name.";
+
+        internal const string TypeConverterGetPropertiesMessage = "The Type of value cannot be statically discovered. The public parameterless constructor or the 'Default' static field may be trimmed from the Attribute's Type.";
     }
 }
--- a/src/System.Windows.Forms.Primitives/src/System/TrimmingConstants.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/TrimmingConstants.cs
@@ -6,6 +6,20 @@ namespace System
 {
     internal static class TrimmingConstants
     {
+        internal const string FilterRequiresUnreferencedCodeMessage = "The public parameterless constructor or the 'Default' static field may be trimmed from the Attribute's Type.";
+
+        internal const string AttributesRequiresUnreferencedCodeMessage = "Generic TypeConverters may require the generic types to be annotated. For example, NullableConverter requires the underlying type to be DynamicallyAccessedMembers All.";
+
+        internal const string PropertyDescriptorPropertyTypeMessage = "PropertyDescriptor's PropertyType cannot be statically discovered.";
+
+        internal const string EditorRequiresUnreferencedCode = "Editors registered in TypeDescriptor.AddEditorTable may be trimmed.";
+
+        internal const string EventDescriptorRequiresUnreferencedCodeMessage = "The built-in EventDescriptor implementation uses Reflection which requires unreferenced code.";        
+
         internal const string TypeConverterGetPropertiesMessage = "The Type of value cannot be statically discovered. The public parameterless constructor or the 'Default' static field may be trimmed from the Attribute's Type.";
+
+        internal const string BindingListViewFilterMessage = "Members of types used in the filter expression might be trimmed.";
+
+        internal const string SiteNameMessage = "The Type of components in the container cannot be statically discovered to validate the name.";
     }
 }
--- a/src/System.Windows.Forms.Primitives/src/System/TrimmingConstants.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/TrimmingConstants.cs
@@ -21,5 +21,7 @@ namespace System
         internal const string SiteNameMessage = "The Type of components in the container cannot be statically discovered to validate the name.";
 
         internal const string TypeConverterGetPropertiesMessage = "The Type of value cannot be statically discovered. The public parameterless constructor or the 'Default' static field may be trimmed from the Attribute's Type.";
+
+        internal const string TypeOrValueNotDiscoverableMessage = "The Type of value cannot be statically discovered";
     }
 }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/BindingSource.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/BindingSource.cs
@@ -7,6 +7,7 @@
 using System.Collections;
 using System.ComponentModel;
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using System.Drawing.Design;
 using System.Globalization;
 using System.Reflection;
@@ -1813,6 +1814,8 @@ namespace System.Windows.Forms
         public virtual string Filter
         {
             get => _filter;
+
+            [RequiresUnreferencedCode("Members of types used in the filter expression might be trimmed.")]
             set
             {
                 _filter = value;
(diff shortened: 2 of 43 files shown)


kant2002:
First jab at https://github.com/dotnet/winforms/issues/4649

src/System.Windows.Forms/src/System/Windows/Forms/BindingSource.cs
```diff
@@ -1813,6 +1814,8 @@ public virtual ListSortDescriptionCollection SortDescriptions
         public virtual string Filter
         {
             get => _filter;
+
+            [RequiresUnreferencedCode("Members of types used in the filter expression might be trimmed.")]
```
RussKie, line 1818: We seem to be repeating the same literals over and over. I think it'd make sense to create a dedicated file (e.g., TrimmingConstants.cs) under SWF.Primitives project, and reuse those across the codebase. Similar to what you proposed in ComNativeDescriptor, just scale it out to the solution level.

RussKie (commented):
Could you please describe the developer flow (that is, what steps are to surface the issue) and the testing steps (i.e., the results before and after)?
Thanks

RussKie:
@agocke does it look like we're moving in the right direction?

src/System.Windows.Forms/src/System/Windows/Forms/BindingSource.cs
kant2002: Great. I was looking for proper location of these message!

kant2002:
I think I should start with https://github.com/dotnet/runtime/pull/61874#issuecomment-975553905 suggestions and create baseline first, and then chip-off from that issues in subsequence PR like this one.

RussKie:
> I think I should start with [dotnet/runtime#61874 (comment)](https://github.com/dotnet/runtime/pull/61874#issuecomment-975553905) suggestions and create baseline first, and then chip-off from that issues in subsequence PR like this one.

Thank you for the link. Sounds good👍

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewButtonColumn.cs
```diff
@@ -5,13 +5,15 @@
...
 using System.ComponentModel;
+using System.Diagnostics.CodeAnalysis;
 using System.Drawing;
 using System.Globalization;
 using System.Text;
 
 namespace System.Windows.Forms
 {
     [ToolboxBitmap(typeof(DataGridViewButtonColumn), "DataGridViewButtonColumn")]
+    [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)]
```
RussKie, line 16: This inflames my OCD :) Keep the attributes in A-Z order, if possible

kant2002: Now I understand why feel at home her

src/System.Windows.Forms.Primitives/src/System/TrimmingConstants.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        internal const string EditorRequiresUnreferencedCode = "Editors registered in TypeDescriptor.AddEditorTable may be trimmed.";
+
+        internal const string EventDescriptorRequiresUnreferencedCodeMessage = "The built-in EventDescriptor implementation uses Reflection which requires unreferenced code.";        
+
+        internal const string TypeConverterGetPropertiesMessage = "The Type of value cannot be statically discovered. The public parameterless constructor or the 'Default' static field may be trimmed from the Attribute's Type.";
+
+        internal const string BindingListViewFilterMessage = "Members of types used in the filter expression might be trimmed.";
+
+        internal const string SiteNameMessage = "The Type of components in the container cannot be statically discovered to validate the name.";
```
RussKie, line 23: Please keep these in A-Z order, because once we have dozens of those searching for these outside an IDE may become a royal pain.

RussKie (commented):
Fails to compile

RussKie:
Could you please describe the developer flow what commands you ran, what were the results before this change and with the change? 
I understand you're following the guide provided by @vitek-karas in https://github.com/dotnet/runtime/pull/61874#issuecomment-975553905, however, we can't expect everyone looking at this change to go an reverse engineer the suggested.

src/System.Windows.Forms/src/System/Windows/Forms/ComponentModel/COM2Interop/ComNativeDescriptor.cs
```diff
@@ -516,6 +518,7 @@ string ICustomTypeDescriptor.GetComponentName()
             /// <summary>
             ///  ICustomTypeDescriptor implementation.
             /// </summary>
+            [RequiresUnreferencedCode(AttributesRequiresUnreferencedCodeMessage)]
             TypeConverter ICustomTypeDescriptor.GetConverter()
             {
                 return GetConverter();
```
vitek-karas, line 524: Why isn't the `GetConverter` annotated with RUC?
Typically when we added annotations in runtime we went bottom to top - started with the places which reported warnings and added annotations as necessary. In this case the problematic calls is `TypeDescriptor.GetConverter` - which is in `GetConverter`, so I would expect that to be annotated first - and from there all of these should be annotated as well.

vitek-karas: I see - the interface definition also annotates these with RUC - got it.
You'll need to annotate GetConverter at some point, but I understand how you got here now - sorry.

src/System.Windows.Forms/src/System/Windows/Forms/Control.AxSourcingSite.cs
```diff
@@ -61,6 +62,8 @@ internal AxSourcingSite(IComponent component, Ole32.IOleClientSite clientSite, s
             public string? Name
             {
                 get => _name;
+
+                [RequiresUnreferencedCode(TrimmingConstants.SiteNameMessage)]
```
vitek-karas, line 66: I think this needs some comment to explain what's dangerous about setting the property.
I assume that the name is later used in some reflection stuff, but in that case the .ctor should also be annotated as it can also set the name.

src/System.Windows.Forms/src/System/Windows/Forms/HtmlToClrEventProxy.cs
```diff
@@ -57,51 +58,71 @@ Type IReflect.UnderlyingSystemType
         }
 
         // Methods
+        [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicFields | DynamicallyAccessedMemberTypes.NonPublicFields)]
         FieldInfo IReflect.GetField(string name, BindingFlags bindingAttr)
         {
             return typeIReflectImplementation.GetField(name, bindingAttr);
```
vitek-karas, line 64: You will need to annotate the fields with `All` probably as well for this to work correctly.
Honestly this class is weird - it will end up calling the GetFields on itself - basically `typeof(HtmlToClrEventProxy).GetFields`. I assume it's because it's really only used for the `InvokeMember`, but in that case I would expect everything else to throw, since the way it is right now it doesn't make much sense to me.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewBand.cs
```diff
@@ -12,6 +12,7 @@ namespace System.Windows.Forms
     /// <summary>
     ///  Identifies a band or column in the dataGridView.
     /// </summary>
+    [DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor)]
```
vitek-karas, line 15: Nit: This is the only type in its hierarchy which technically needs the attribute. The attribute is effectively inherited and so you don't need to add it to a derived type again, even if that type "uses" it. But there's no harm in having it on the derived types either.

vitek-karas:
If the effort to make winforms trimmable is serious (as in the repo maintainers want this), then I would probably start with some infra similar to what runtime has:
* Custom step in build which runs the trimmer on all of the assemblies "at once" - @agocke: ideally we would make it easier to set this up, since I assume other repos will want to use this as well - maybe we should make it part of arcade?
* This step uses `ILLink.suppression.xml` files in each project to suppress trim warnings produced - these files effectively act like TODO lists. Once the effort to make winforms trimmable is done, these files should be empty.
* Run it first, populate the suppression files and merge

An example of this in the runtime repo is this target: https://github.com/dotnet/runtime/blob/6c7d9359c271212416865c446ad0167cbdcd17e1/src/libraries/oob.proj#L75
The infra in runtime is probably overly complex (since runtime runs the linker twice, which winforms probably shouldn't do), but the basics can be seen there.

This provides infra which has couple of advantages:
* Provides a repeatable way to get the trim warnings - per @RussKie request
* Should be part of CI, so new code doesn't regress the trimming effort - or at least does it intentionally
* As noted above, acts as a TODO list to track the remaining work (it's not very precise as some warnings are easy while others are hard to fix, but at least something)

src/System.Windows.Forms/src/System/Windows/Forms/Control.AxSourcingSite.cs
kant2002: From what I see from code, I can add supression saying that Name is used only for naming purposes. Even in the designer I think that names used for CodeDom mostly, and never accessed via reflections. 

Okay I found places where reflection used
https://github.com/dotnet/winforms/blob/6e448a3d861c695f3ecf94424ffe79ef5db40fbe/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs#L546-L559

CodeDom
https://github.com/dotnet/winforms/blob/6e448a3d861c695f3ecf94424ffe79ef5db40fbe/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomDesignerLoader.ModifiersExtenderProvider.cs#L188-L199
and 
https://github.com/dotnet/winforms/blob/6e448a3d861c695f3ecf94424ffe79ef5db40fbe/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomDesignerLoader.ModifiersInheritedExtenderProvider.cs#L104-L111

kant2002: And one more I miss to post due to slippery pressing Enter
https://github.com/dotnet/winforms/blob/6e448a3d861c695f3ecf94424ffe79ef5db40fbe/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ToolStripItemDesigner.cs#L711-L720

I'm curious is designers only partially open sourced, or these locations can be rewritten without need for reflection.

RussKie (approved):

RussKie:
Our build agents are currently undergoing some maintenance, so we'll need to re-run the build once the agents are back online.

PR #7269, opened 2022-06-08
Ignore tooltip null-owners

### LINKED ISSUE #7262: DataGridView: sporadic NullReferenceException when trying to display the tooltip
### .NET version

.NET version: 6.0.300
Windows version: Windows 7 using classic theme

I'm not 100% sure whether it works in .NET Framework but I did not face this issue earlier.

### Issue description

```
   at System.Windows.Forms.Control.GetSafeHandle(IWin32Window window)
   at Interop.ComCtl32.ToolInfoWrapper`1..ctor(T handle, TTF flags, String text)
   at System.Windows.Forms.ToolTip.TryGetBubbleSize(IKeyboardToolTip tool, Rectangle toolRectangle, Size& bubbleSize)
   at System.Windows.Forms.ToolTip.ShowKeyboardToolTip(String text, IKeyboardToolTip tool, Int32 duration)
   at System.Windows.Forms.KeyboardToolTipStateMachine.ShowToolTip(IKeyboardToolTip tool, ToolTip toolTip)
   at System.Windows.Forms.KeyboardToolTipStateMachine.Transition(IKeyboardToolTip tool, ToolTip tooltip, SmEvent event)
   at System.Windows.Forms.KeyboardToolTipStateMachine.Transit(SmEvent event, IKeyboardToolTip source)
   at System.Windows.Forms.KeyboardToolTipStateMachine.<SetupInitShowTimer>b__25_0(Timer sender)
   at System.Windows.Forms.KeyboardToolTipStateMachine.<>c__DisplayClass27_0.<GetOneRunTickHandler>g__wrapper|0(Object sender, EventArgs eventArgs)
   at System.Windows.Forms.Timer.TimerNativeWindow.WndProc(Message& m)
```

### Steps to reproduce

It occurs randomly (up to a couple of times per day) at one of my customer's computer who uses Windows 7 with visual styles disabled (classic theme).

The `DataGridView` is rapidly updated via its `DataSource` (in every 100ms). It does not have any tooltip configured so it must be the default tooltip when the cell is too small to be able to display the entire text.

I apologize for not being able to provide a more specific reproduction but please refer to the provided strack trace, which clearly indicates that some handle must be invalid in the `ToolInfoWrapper<T>` constructor when the size of the tooltip is attempted to be determined.

I could never reproduce it locally maybe because I use Windows 10.

RussKie:
Resolves #7262

In some circumstances it's plausible to have a race condition, and the control, for which a tooltip is being created, may have disappeared by the time the tooltip is about to be shown. This may lead to the following:

```
   at System.Windows.Forms.Control.GetSafeHandle(IWin32Window window)
   at Interop.ComCtl32.ToolInfoWrapper`1..ctor(T handle, TTF flags, String text)
   at System.Windows.Forms.ToolTip.TryGetBubbleSize(IKeyboardToolTip tool, Rectangle toolRectangle, Size& bubbleSize)
   at System.Windows.Forms.ToolTip.ShowKeyboardToolTip(String text, IKeyboardToolTip tool, Int32 duration)
   at System.Windows.Forms.KeyboardToolTipStateMachine.ShowToolTip(IKeyboardToolTip tool, ToolTip toolTip)
```

Ignore such situations.

I tried to recreate the issue on W11 with the following tweak to [the existing test app](https://github.com/dotnet/winforms/blob/4223d0a6ff606335e064c378a302a05fa382f0e2/src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/DataGridViewInVirtualModeTest.cs) and had the mouse hover cells in the first column, but similar to [the customer's report](https://github.com/dotnet/winforms/issues/7262#issue-1262896269) was unable to do so. 

```cs
        private Timer _timer = new();
        private Random random = new((int)DateTime.Now.Ticks);

        public DataGridViewInVirtualModeTest()
        {
            InitializeComponent();
            dataGridView1.VirtualMode = false;
            DoubleBuffered = true;

            _timer.Interval = 100;
            _timer.Tick += (s, e) =>
            {
                _customers.Clear();

                // Add some sample entries to the data store
                for (int i = 0; i < random.Next(50, 100); i++)
                {
                    _customers.Add(new TestCustomer("Ann" + new string('0', random.Next(0, 50)), 23, true, "Female"));
                    _customers.Add(new TestCustomer("John" + new string('0', random.Next(0, 50)), 45, true, "Male"));
                    _customers.Add(new TestCustomer("Sarah" + new string('0', random.Next(0, 50)), 67, false, "Female"));
                }

                dataGridView1.DataSource = null;
                dataGridView1.DataSource = _customers;
            };
        }
```

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7269)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/Interop/ComCtl32/Interop.TTOOLINFOW.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/ComCtl32/Interop.TTOOLINFOW.cs
@@ -29,35 +29,48 @@ internal static partial class Interop
             public TTOOLINFOW Info;
             public string? Text { get; set; }
             [MaybeNull]
-            private readonly T _handle;
+            private readonly T? _handle;
 
-            public unsafe ToolInfoWrapper(T handle, TTF flags = default, string? text = null)
+            public unsafe ToolInfoWrapper(T? handle, TTF flags = default, string? text = null)
             {
-                Info = new TTOOLINFOW
-                {
-                    hwnd = handle.Handle,
-                    uId = handle.Handle,
-                    uFlags = flags | TTF.IDISHWND
-                };
-                Text = text;
                 _handle = handle;
+                if (_handle is not null)
+                {
+                    Info = new TTOOLINFOW
+                    {
+                        hwnd = _handle.Handle,
+                        uId = _handle.Handle,
+                        uFlags = flags | TTF.IDISHWND
+                    };
+                }
+
+                Text = text;
             }
 
-            public unsafe ToolInfoWrapper(T handle, IntPtr id, TTF flags = default, string? text = null, RECT rect = default)
+            public unsafe ToolInfoWrapper(T? handle, IntPtr id, TTF flags = default, string? text = null, RECT rect = default)
             {
-                Info = new TTOOLINFOW
-                {
-                    hwnd = handle.Handle,
-                    uId = id,
-                    uFlags = flags,
-                    rect = rect
-                };
-                Text = text;
                 _handle = handle;
+                if (_handle is not null)
+                {
+                    Info = new TTOOLINFOW
+                    {
+                        hwnd = _handle.Handle,
+                        uId = id,
+                        uFlags = flags,
+                        rect = rect
+                    };
+                }
+
+                Text = text;
             }
 
             public unsafe nint SendMessage(IHandle sender, User32.WM message, BOOL state = BOOL.FALSE)
             {
+                if (_handle is null)
+                {
+                    return 0;
+                }
+
                 Info.cbSize = (uint)sizeof(TTOOLINFOW);
                 fixed (char* c = Text)
                 fixed (void* i = &Info)


JeremyKuhne (changes requested):
Making a change to the `ToolInfoWrapper` isn't the right thing to do. Allowing it to be constructed with null will potentially mask other problems.

In addition the guard here is wrong as the handle isn't null, the `IWin32WindowAdapter.Handle` is null ref-ing in `Control.GetSafeHandle`.

https://github.com/dotnet/winforms/blob/05ee2b2105c17ec8182664de8eb9378c2c7fe116/src/System.Windows.Forms/src/System/Windows/Forms/IWin32WindowAdapter.cs#L16

https://github.com/dotnet/winforms/blob/9ebacf6525658bf23d7710cad86c5c8bf1a9c141/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs#L6216

[`DataGridViewCell.GetOwnerWindow()`](https://github.com/dotnet/winforms/blob/9ebacf6525658bf23d7710cad86c5c8bf1a9c141/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCell.cs#L447) is most likely where this is going wrong. There are any number of ways it can be null, we do it temporarily in a few places and the API it hits is public (`DataGridViewElement.DataGridView`).

The better choice is to update: https://github.com/dotnet/winforms/blob/e0e125abe00ae53cd102df47ed9a55deb2d5010d/src/System.Windows.Forms/src/System/Windows/Forms/ToolTip.cs#L1568-L1572

If `.GetOwnerWindow()` returns null set `result` to `IntPtr.Zero` so it returns `false`  and `Size.Empty`.

Once we get fully annotated these things will work their way out naturally.

PR #7279, opened 2022-06-09
Use LibraryImport in some Interop Kernel32

gpetrou:
## Proposed changes

- Use LibraryImport in some Interop Kernel32.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7279)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/Interop/Kernel32/Interop.GetModuleHandle.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/Kernel32/Interop.GetModuleHandle.cs
@@ -8,7 +8,7 @@ internal partial class Interop
 {
     internal partial class Kernel32
     {
-        [DllImport(Libraries.Kernel32, CharSet = CharSet.Unicode, SetLastError = true, ExactSpelling = true)]
-        public static extern IntPtr GetModuleHandleW(string? moduleName);
+        [LibraryImport(Libraries.Kernel32, SetLastError = true, StringMarshalling = StringMarshalling.Utf16)]
+        public static partial IntPtr GetModuleHandleW(string? moduleName);
     }
 }
--- a/src/System.Windows.Forms.Primitives/src/Interop/Kernel32/Interop.MultiByteToWideChar.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/Kernel32/Interop.MultiByteToWideChar.cs
@@ -13,8 +13,8 @@ internal partial class Interop
         //            of WCHARs in the string, while the size of the Out buffer equals the number of bytes. To avoid a buffer overrun, be sure to specify
         //            a buffer size appropriate for the data type the buffer receives. For more information, see Security Considerations: International Features.
         //            Always call these functions passing a null destination buffer to get its size and the create the buffer with the exact size.
-        [DllImport(Libraries.Kernel32, CharSet = CharSet.Unicode, SetLastError = true, ExactSpelling = true)]
-        public unsafe static extern int MultiByteToWideChar(
+        [LibraryImport(Libraries.Kernel32, SetLastError = true)]
+        public static unsafe partial int MultiByteToWideChar(
             CP CodePage,
             uint dwFlags,
             byte* lpMultiByteStr,
--- a/src/System.Windows.Forms.Primitives/src/Interop/Kernel32/Interop.ActivateActCtx.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/Kernel32/Interop.ActivateActCtx.cs
@@ -8,7 +8,7 @@ internal partial class Interop
 {
     internal partial class Kernel32
     {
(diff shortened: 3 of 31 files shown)


src/System.Windows.Forms.Primitives/src/Interop/Kernel32/Interop.MultiByteToWideChar.cs
```diff
@@ -13,8 +13,8 @@ internal partial class Kernel32
         //            of WCHARs in the string, while the size of the Out buffer equals the number of bytes. To avoid a buffer overrun, be sure to specify
         //            a buffer size appropriate for the data type the buffer receives. For more information, see Security Considerations: International Features.
         //            Always call these functions passing a null destination buffer to get its size and the create the buffer with the exact size.
-        [DllImport(Libraries.Kernel32, CharSet = CharSet.Unicode, SetLastError = true, ExactSpelling = true)]
-        public unsafe static extern int MultiByteToWideChar(
+        [LibraryImport(Libraries.Kernel32, SetLastError = true)]
```
gpetrou, line 16: @AaronRobinsonMSFT I used the SYSLIB1054 fixer to do the conversions. Shouldn't it have added `StringMarshalling = StringMarshalling.Utf16` here? It seems that it has added in other places such as in `GetLocaleInfoEx`.

AaronRobinsonMSFT: Not in this case. This is one of those scenarios that is tricky and not obvious what to do so the marshaller dropped the setting. The fixer wants to remove all unnecessary attributes - that is a goal. The next phase is look for types in the signature that we support marshalling using a string marshaller. None of the arguments in this signature are .NET strings. Which means we remove it and putting it back in place isn't going to matter because there is nothing to use it.

The encoding of `byte*` isn't known and the `char*` is just a pointer to a `char`. There is nothing for the string marshaller to do. Now, if you were to convert the `char*` to a `string`, then it would require `StringMarshalling = StringMarshalling.Utf16` or whatever encoding is expected.

gpetrou: I see. OK thanks.

src/System.Windows.Forms.Primitives/src/Interop/Kernel32/Interop.GetModuleHandle.cs
```diff
@@ -8,7 +8,7 @@ internal partial class Interop
 {
     internal partial class Kernel32
     {
-        [DllImport(Libraries.Kernel32, CharSet = CharSet.Unicode, SetLastError = true, ExactSpelling = true)]
-        public static extern IntPtr GetModuleHandleW(string? moduleName);
+        [LibraryImport(Libraries.Kernel32, SetLastError = true, StringMarshalling = StringMarshalling.Utf16)]
```
RussKie, line 11: @AaronRobinsonMSFT just for my own education - we marshal `CharSet.Unicode` as `UTF-16`?

RussKie (approved):

RussKie:
Thank you

src/System.Windows.Forms.Primitives/src/Interop/Kernel32/Interop.GetModuleHandle.cs
AaronRobinsonMSFT: Oh boy.. this is a long standing mess and entirely due to the legacy of .NET Framework and Microsoft being one of the first movers for supporting Unicode.

Microsoft took the bet on UCS-2 and lost. UTF-16 can typically replace UCS-2 as a drop-in so that is what was done. Instead of describing the encoding precisely the generic term "Unicode" was applied in .NET Framework 1.1. This made sense on Windows because the default Windows Unicode encoding was UTF-16, think `wchar_t` on Windows. Aside, `wchar_t` has an implementation defined size and is always 2-bytes on Windows but typically 4-bytes everywhere else.

During the early bring up of .NET Core, a decision needed to be made about what is `CharSet.Unicode`. In an attempt to bring some sanity to the world, we went with `CharSet.Unicode` meaning UTF-16 on all platforms. This helps with consistency but not necessarily with non-Windows interop, because on non-Windows the default Unicode encoding is UTF-8 - [UTF-8 has won](http://utf8everywhere.org/). Users could get the UTF-8 experience on non-Windows by setting `CharSet.Ansi`, but on Windows `CharSet.Ansi` would continue to mean ANSI code-pages. So we brought sanity but nothing intuitive.

The Interop team decided to completely abandon the `CharSet` enumeration because frankly it is broken in a cross-platform world. We introduced the `StringMarshalling` enumeration with precisely defined encodings that are unambiguous. This was our attempt to bring sanity and an intuitive experience.

PR #7284, opened 2022-06-09
Fix ListBox items accessible objects memory leak 

### LINKED ISSUE #7197: ListBox items' accessible objects are leaking in memory
### .NET version

.NET 7.0 Preview 4 

No

No

### Issue description

`ListBox` items' accessible objects are left in memory after the `ListBox` control instance with its items was inspected with accessibility tool (like Accessibility Insights or Inspect) and the form with `ListBox` is closed, garbage is collected, but accessibility tool is still opened. Same applies for case when `ListBox` item is removed from `ListBox`.

Memory snapshot in VS:
Screenshot 2022-05-18 174214

### Steps to reproduce

1. Create a parent form with two buttons: one will be opening a child form with the `ListBox` control and the other will collect the garbage.
2. Press the second button to open the child form with `ListBox`.
4. Run automated checks for the `ListBox` control using AI.
5. Close the child form.
7. DON'T close the AI tool.
8. Look for `ListBoxItemAccessibleObject` accessible objects left in memory (using WinDbg, or dotMemory, or another suitable tool).

Sample app: [ListBoxSample.zip](https://github.com/dotnet/winforms/files/8731986/ListBoxSample.zip)

dmitrii-drobotov:
Fixes #7197.

## Proposed changes

- Add call to `UiaDisconnectProvider` for `ListBoxItemAccessibleObject`s for cases when:
    - `ListBox` is destroyed (now it only releases provider for `ListBoxAccessibleObject`)
    - All items are cleared from `ListBox`
    - An item is removed from `ListBox`
- Add unit tests for above scenarios

## Customer Impact

- `ListBoxItemAccessibleObject`s memory will stop leaking 

## Regression? 

- No

## Risk

- Minimal

### Before

After executing steps from issue description for ListBox with 18 items:

Screenshot 2022-06-09 184103

### After

Screenshot 2022-06-09 180310

After `ListBox.Items.Clear()`:

Screenshot 2022-06-09 183627

One `ListBoxItemAccessibleObject` instance may be held by `ListBox._focusedItem` field, which doesn't create a memory leak, because the AO was disconnected from UIA provider and will be garbage-collected normally once form is closed.

After  `ListBox.Items.RemoveAt(0)` 5 times:

Screenshot 2022-06-09 183839

## Test methodology 

- Unit testing
- Manual testing with WinDbg to check memory leak, Narrator to check for regressions similar to #6230, Inspect/AI to check that removed items are inaccessible there
 

## Test environment(s)

* Windows 11 (Version 10.0.22000)
* .NET 7.0.0-preview.5.22272.3

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7284)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListBox.AccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListBox.AccessibleObject.cs
@@ -177,7 +177,40 @@ namespace System.Windows.Forms
 
             internal void ResetListItemAccessibleObjects()
             {
+                if (OsVersion.IsWindows8OrGreater)
+                { 
+                    foreach (ListBoxItemAccessibleObject itemAccessibleObject in _itemAccessibleObjects.Values)
+                    {
+                        HRESULT result = UiaCore.UiaDisconnectProvider(itemAccessibleObject);
+                        Debug.Assert(result == HRESULT.S_OK);
+                    }
+                }
+
                 _itemAccessibleObjects.Clear();
+            }
+
+            internal void RemoveListItemAccessibleObjectAt(int index)
+            {
+                IReadOnlyList<ItemArray.Entry?> entries = _owningListBox.Items.InnerArray.Entries;
+                if (index >= entries.Count)
+                {
+                    return;
+                }
+
+                ItemArray.Entry? item = entries[index];
+
+                if (item is null || !_itemAccessibleObjects.ContainsKey(item))
+                {
+                    return;
+                }
+
+                if (OsVersion.IsWindows8OrGreater)
+                {
+                    HRESULT result = UiaCore.UiaDisconnectProvider(_itemAccessibleObjects[item]);
+                    Debug.Assert(result == HRESULT.S_OK);
+                }
+
+                _itemAccessibleObjects.Remove(item);
             }
 
             internal override void SelectItem()
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs
@@ -1346,6 +1346,14 @@ namespace System.Windows.Forms
             return Math.Max(oldMax, textSize.Width);
         }
 
+        private void ClearListItemAccessibleObjects()
+        {
+            if (IsAccessibilityObjectCreated && AccessibilityObject is ListBoxAccessibleObject accessibilityObject)
+            {
+                accessibilityObject.ResetListItemAccessibleObjects();
+            }
+        }
+
         /// <summary>
         ///  Deselects all currently selected items.
         /// </summary>
@@ -2089,6 +2097,20 @@ namespace System.Windows.Forms
         protected override void RefreshItem(int index)
         {
             Items.SetItemInternal(index, Items[index]);
+        }
+
+        internal override void ReleaseUiaProvider(IntPtr handle)
+        {
+            base.ReleaseUiaProvider(handle);
+            ClearListItemAccessibleObjects();
+        }
+
+        private void RemoveListItemAccessibleObjectAt(int index)
+        {
+            if (IsAccessibilityObjectCreated && AccessibilityObject is ListBoxAccessibleObject accessibilityObject)
+            {
+                accessibilityObject.RemoveListItemAccessibleObjectAt(index);
+            }
         }
 
         public override void ResetBackColor()
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListBoxAccessibleObjectTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListBoxAccessibleObjectTests.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using Xunit;
+using static System.Windows.Forms.ListBox;
 using static Interop;
 
 namespace System.Windows.Forms.Tests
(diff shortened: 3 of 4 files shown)


RussKie (commented):
👍

src/System.Windows.Forms/src/System/Windows/Forms/ListBox.AccessibleObject.cs
```diff
@@ -177,9 +177,41 @@ internal override bool IsPatternSupported(UiaCore.UIA patternId)
 
             internal void ResetListItemAccessibleObjects()
             {
+                if (OsVersion.IsWindows8OrGreater)
+                { 
+                    foreach (ListBoxItemAccessibleObject itemAccessibleObject in _itemAccessibleObjects.Values)
+                    {
+                        HRESULT result = UiaCore.UiaDisconnectProvider(itemAccessibleObject);
+                        Debug.Assert(result == HRESULT.S_OK);
```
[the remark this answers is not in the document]
dmitrii-drobotov: This extension method would miss `HRESULT.S_FALSE` which has value of 1. See https://github.com/dotnet/winforms/pull/6249#discussion_r870798274

src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs
```diff
@@ -2091,6 +2099,20 @@ protected override void RefreshItem(int index)
             Items.SetItemInternal(index, Items[index]);
         }
 
+        internal override void ReleaseUiaProvider(IntPtr handle)
+        {
+            base.ReleaseUiaProvider(handle);
+            ClearListItemAccessibleObjects();
+        }
+
+        private void RemoveListItemAccessibleObject(int index)
```
vladimir-krestov, line 2108: Consider using `At` in the name end - `RemoveListItemAccessibleObjectAt` like we use for `RemoveAt`

src/System.Windows.Forms/src/System/Windows/Forms/ListBox.AccessibleObject.cs
```diff
@@ -177,9 +177,41 @@ internal override bool IsPatternSupported(UiaCore.UIA patternId)
...
+                    }
+                }
+
                 _itemAccessibleObjects.Clear();
             }
 
+            internal void RemoveListItemAccessibleObject(int index)
+            {
+                if (index >= _owningListBox.Items.InnerArray.Entries.Count)
```
vladimir-krestov, line 194: Extract entries:
```suggestion
            {
                ItemArray.Entry[] entries = _owningListBox.Items.InnerArray.Entries;
                if (index >= entries.Count)
```

src/System.Windows.Forms/src/System/Windows/Forms/ListBox.AccessibleObject.cs
```diff
@@ -177,9 +177,41 @@ internal override bool IsPatternSupported(UiaCore.UIA patternId)
...
+                    {
+                        HRESULT result = UiaCore.UiaDisconnectProvider(itemAccessibleObject);
+                        Debug.Assert(result == HRESULT.S_OK);
+                    }
+                }
+
                 _itemAccessibleObjects.Clear();
             }
 
+            internal void RemoveListItemAccessibleObject(int index)
```
vladimir-krestov, line 192: Consider as well
```suggestion
            internal void RemoveListItemAccessibleObjectAt(int index)
```

src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs
```diff
@@ -2091,6 +2099,20 @@ protected override void RefreshItem(int index)
             Items.SetItemInternal(index, Items[index]);
         }
 
+        internal override void ReleaseUiaProvider(IntPtr handle)
```
vladimir-krestov, line 2102: Please check the RecreateControl case. Are the objects correct after recreating the ListBox control?
For example, open PropertyGrid, use some objects as SelectedObject, and check AccessibleRole property, expand its items and make sure that they are correct in inspect after several expand-collapse actions.
After that also check memory leaks of ListBox after disposing of the PropertyGrid.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListBoxAccessibleObjectTests.cs
```diff
@@ -102,5 +93,72 @@ public void ListBoxItemAccessibleObject_GetPropertyValue_ValueValuePropertyId_Re
...
+                "g"
+            });
+
+            return listBox;
+        }
+
+        private ListBoxAccessibleObject InitListBoxItemsAccessibleObjects(ListBox listBox)
+        {
+            var accessibilityObject = (ListBoxAccessibleObject)listBox.AccessibilityObject;
+            var childCount = accessibilityObject.GetChildCount();
```
vladimir-krestov, line 152: Explicit `var`s

vladimir-krestov (commented):
Looks good! Check recreating case

vladimir-krestov:
PS. Tested the fix for a `PropertyGrid` case (`GridViewListBoxAccessibleObject`). The fix works correctly, 0 ListBox objects are leaking. ✔️

src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs
dmitrii-drobotov: I checked it and can confirm that there are no issues after recreation. `ReleaseUiaProvider` indeed gets called and items become unavailable in Inspect (same as if form was closed) but once you click Refresh, new items will be in place. In AI there is an auto-refresh, so new list simply replaces the old one. Narrator works fine as well. It also doesn't cause memory leak.

RussKie (approved):

RussKie:

vladimir-krestov (approved):

dmitrii-drobotov:
@vladimir-krestov For `CheckedListBox` it also fixes the leak.
Before fix:
image
After fix:
Screenshot 2022-06-27 110051

Tanya-Solyanik (approved):
Looks good!

PR #7290, opened 2022-06-11
Produce trim warning as part of the build

kant2002:
For now I only enable trim warnings as part of the build as per https://docs.microsoft.com/en-us/dotnet/core/deploying/trimming/prepare-libraries-for-trimming#enable-the-roslyn-analyzer
That much simpler setup then proposed by @vitek-karas and I think it
would suffice us.

Related #4649

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7290)

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -23,9 +23,7 @@
                                   >true</IsAnalyzerProject>
 
     <EnableTrimAnalyzer Condition="!$(IsTestProject) and !$(IsAnalyzerProject) and
-                                  !($(MSBuildProjectName.EndsWith('.TestUtilities')) or $(MSBuildProjectName.EndsWith('VisualBasicRuntimeTest'))
-                                    or $(MSBuildProjectName.EndsWith('WinformsControlsTest')) or $(MSBuildProjectName.EndsWith('DesignSurfaceExt'))
-                                    or $(MSBuildProjectName.EndsWith('DemoConsole')))">true</EnableTrimAnalyzer>
+                                  !($(MSBuildProjectName.EndsWith('.TestUtilities'))">true</EnableTrimAnalyzer>
   </PropertyGroup>
 
   <!--
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -21,6 +21,11 @@
     <IsAnalyzerProject Condition="!$(IsTestProject) and
                                   ($(MSBuildProjectName.EndsWith('.Analyzers')) or $(MSBuildProjectName.EndsWith('.Analyzers.CSharp')))"
                                   >true</IsAnalyzerProject>
+
+    <EnableTrimAnalyzer Condition="!$(IsTestProject) and !$(IsAnalyzerProject) and
+                                  !($(MSBuildProjectName.EndsWith('.TestUtilities')) or $(MSBuildProjectName.EndsWith('VisualBasicRuntimeTest'))
+                                    or $(MSBuildProjectName.EndsWith('WinformsControlsTest')) or $(MSBuildProjectName.EndsWith('DesignSurfaceExt'))
+                                    or $(MSBuildProjectName.EndsWith('DemoConsole')))">true</EnableTrimAnalyzer>
   </PropertyGroup>
 
   <!--
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -22,8 +22,8 @@
                                   ($(MSBuildProjectName.EndsWith('.Analyzers')) or $(MSBuildProjectName.EndsWith('.Analyzers.CSharp')))"
                                   >true</IsAnalyzerProject>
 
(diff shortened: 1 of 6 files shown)


Directory.Build.props
```diff
@@ -21,6 +21,11 @@
     <IsAnalyzerProject Condition="!$(IsTestProject) and
                                   ($(MSBuildProjectName.EndsWith('.Analyzers')) or $(MSBuildProjectName.EndsWith('.Analyzers.CSharp')))"
                                   >true</IsAnalyzerProject>
+
+    <EnableTrimAnalyzer Condition="!$(IsTestProject) and !$(IsAnalyzerProject) and
+                                  !($(MSBuildProjectName.EndsWith('.TestUtilities')) or $(MSBuildProjectName.EndsWith('VisualBasicRuntimeTest'))
+                                    or $(MSBuildProjectName.EndsWith('WinformsControlsTest')) or $(MSBuildProjectName.EndsWith('DesignSurfaceExt'))
+                                    or $(MSBuildProjectName.EndsWith('DemoConsole')))">true</EnableTrimAnalyzer>
```
RussKie, line 28: May be use `MSBuildProjectDirectory` (or [other suitable property](https://docs.microsoft.com/visualstudio/msbuild/msbuild-reserved-and-well-known-properties)) and test for `\tests\`?

src/System.Windows.Forms.Design/src/System.Windows.Forms.Design.csproj
```diff
@@ -7,6 +7,11 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <!--the obsolete usage in public surface can't be removed-->
     <NoWarn>$(NoWarn);618</NoWarn>
+    <!-- 
+      IL Trim warnings which should be removed in order to make WinForms trimmable
+      See https://github.com/dotnet/winforms/issues/4649
+    -->
+    <NoWarn>$(NoWarn);IL2026;IL2046;IL2067;IL2070;IL2072;IL2075;IL2094;IL2111</NoWarn>
```
RussKie, line 14: Can we put these in Directory.Build.props, so we have everything in the central place?

RussKie (commented):
👍

src/System.Windows.Forms.Design/src/System.Windows.Forms.Design.csproj
kant2002: No need to central place. I expect to iterate quickly on Primitives. So they would be clean. Having list of issues in central place allow regressions in trim status of individual assembly.

Directory.Build.props
kant2002: I suspect that better would be just fix names of the projects so IsTestProject property would be attached to them

src/System.Windows.Forms.Design/src/System.Windows.Forms.Design.csproj
RussKie: `IsTestProject` et al have specific meanings, i.e., projects that contain runnable tests, https://github.com/dotnet/arcade/blob/main/src/Microsoft.DotNet.Arcade.Sdk/tools/Tests.props. We have other projects located under `tests` folders that do not contain runnable tests.

Directory.Build.props
kant2002: Can I just manually set EnableTrimAnalyzer for these projects? I definitely would like to comeback to some of them, in order to run them with trimming enabled. For example `VisualBasicRuntimeTest` and `WinformsControlsTest` maybe `DesignSurfaceExt`.

Directory.Build.props
```diff
@@ -21,6 +21,9 @@
     <IsAnalyzerProject Condition="!$(IsTestProject) and
                                   ($(MSBuildProjectName.EndsWith('.Analyzers')) or $(MSBuildProjectName.EndsWith('.Analyzers.CSharp')))"
                                   >true</IsAnalyzerProject>
+
+    <EnableTrimAnalyzer Condition="!$(IsTestProject) and !$(IsAnalyzerProject) and
+                                  !($(MSBuildProjectName.EndsWith('.TestUtilities'))">true</EnableTrimAnalyzer>
```
RussKie, line 26: We should always set the default (i.e., without conditions) value, and after that set the value based on a condition. After that you shouldn't need to configure `EnableTrimAnalyzer` in other places....

Would this work?
```suggestion
    <EnableTrimAnalyzer>true</EnableTrimAnalyzer>
    <EnableTrimAnalyzer Condition="$(IsTestProject) or !$(IsAnalyzerProject) or $(MSBuildProjectDirectory.Contains('\tests\'))">false</EnableTrimAnalyzer>
```

RussKie (approved):

RussKie:
Without the suppressions the build fails with the following:
image

PR #7295, opened 2022-06-13
Releasing leak objects of UiaTextRange

### LINKED ISSUE #7248: TextBox TextPattern objects are leaking in memory
### .NET version

.NET 7.0

No

No

### Issue description

If to use TextPattern abilities, UIA creates and keeps several UiaTextRanges, that keep UiaTextProvider and TextBox. So #6886 is not enough for fixing #6885. 
image

### Steps to reproduce

- Create a .NET app
- Add a TextBox there
- Run the app and Narrator/Inspect/AI
- Check TextPattern work, get some text ranges
- Close the test form to dispose the control
- Check how many object are leaking for the TextBox using WinDbg or something like that

vladimir-krestov:
Fixes #7248

## Proposed changes

- Use HandleDestroyed events to release related objects of UiaTextRange objects. We create new instances of the ranges and send them to UIA (eg. Inspect). When TextBox disposing UIA tool keeps several ranges thereby keep the provider and the TextBox accessible object. Due to we have many unregistered text ranges we don't have access to all of them, so I think it would be better to use events and subscription on them for every text range. 

## Customer Impact

- Application slows down or hangs to perform GC 
- This change related to DataGrdiViewTextBoxCell, PropertyGrid edit field, ComboBox, Numeric/DomainUpDown edit field, etc. TextPatterns leaks affect all controls with edit field and keep all their leaked objects, so this fix resolves it.

## Regression? 

- No

## Risk

- We make some fields as null. UIA can call some method or property after disposing, thereby try to get some value of null-object. So we can catch NRE. To avoid it needs to add necessary null checks.

### Before

- UIA keeps text ranges and TextBox accessible object
- TextBox accessible object keeps TextBox control (will be fixed next)
image
image

### After

- TextBox text provider is released
image

## Test methodology 

- Manual testing
- CTI

## Accessibility testing  
-  Using Inspect and Narrator:
     - Focus on a TextBox
     - Navigate through characters
     - Select some range
     - Use Inspect TextPattern explorer to get all possible ranges

## Test environment(s) 

- dotnet 7.0.0-preview.5.22272.3
- Windows 11

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7295)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextProvider.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextProvider.cs
@@ -98,6 +98,13 @@ namespace System.Windows.Forms.Automation
             return doubles;
         }
 
+        internal event EventHandler? OwnerDisposed;
+
+        private protected void RaiseOwnerDisposedEvent()
+        {
+            OwnerDisposed?.Invoke(null, EventArgs.Empty);
+        }
+
         public int SendInput(int inputs, ref INPUT input, int size)
         {
             Span<INPUT> currentInput = stackalloc INPUT[1];
--- a/src/System.Windows.Forms.Primitives/src/Interop/UiaCore/Interop.ITextRangeProvider.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/UiaCore/Interop.ITextRangeProvider.cs
@@ -13,7 +13,7 @@ internal static partial class Interop
         [InterfaceType(ComInterfaceType.InterfaceIsIUnknown)]
         public interface ITextRangeProvider
         {
-            ITextRangeProvider Clone();
+            ITextRangeProvider? Clone();
 
             BOOL Compare(ITextRangeProvider range);
 
@@ -29,7 +29,7 @@ internal static partial class Interop
 
             double[] GetBoundingRectangles();
 
-            IRawElementProviderSimple GetEnclosingElement();
+            IRawElementProviderSimple? GetEnclosingElement();
 
             string GetText(int maxLength);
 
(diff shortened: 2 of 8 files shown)


src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextProvider.cs
```diff
@@ -98,6 +98,13 @@ public static double[] RectListToDoubleArray(List<Rectangle> rectArray)
             return doubles;
         }
 
+        internal event EventHandler? OwnerDisposed;
+
+        private protected void RaiseOwnerDisposedEvent()
```
vladimir-krestov, line 103: FYI: Used this approach instead of call the event inside inherited class because `OwnerDisposed` should be used, otherwise VS throw a warning.

vladimir-krestov:
I investigated if it's possible to use one UiaTextRange instance for all ranges, it breaks TextPattern. I investigated ways to release UiaTextRange objects, but didn't find any. Narrator keeps several ranges.

Tanya-Solyanik:
How many text providers are created per text box?

Consider factoring out cleanup changes in textboxbase and combobox.....cs and submitting a small PR for them

Tanya-Solyanik:
@dmitrii-drobotov  - this is not the highest priority issue, but please factor out changes that test for AcccessibileObject being already created and `IsHandleCreated`, etc, into a smaller PR, if still applicable, and then test what object holds the TextBox, i.e. run !GCRoot on the textbox address. I think it might be OK for the UIAProvider to leak as long as the text box is released.

PR #7296, opened 2022-06-13
Fix TreeNode accessible objects memory leak

### LINKED ISSUE #7196: TreeNode accessible objects are leaking in memory
### .NET version

.NET 7.0 Preview 4 (at least .NET 7.0 Preview 3, where `TreeNodeAccessibleObject` was introduced)

No

No (there was no `TreeNodeAccessibleObject`)

### Issue description

`TreeNode` accessible objects are left in memory after the `TreeView` control instance with its nodes was inspected with accessibility tool (like Accessibility Insights or Inspect) and the form with `TreeView` is closed, garbage is collected, but accessibility tool is still opened. Same applies for case when `TreeNode` is removed from `TreeView`.

Memory snapshot in VS:
Screenshot 2022-05-19 173414

### Steps to reproduce

1. Create a parent form with two buttons: one will be opening a child form with the `TreeView` control and the other will collect the garbage.
2. Press the second button to open the child form with `TreeView`.
4. Run automated checks for the `TreeView` control using AI.
5. Close the child form.
7. DON'T close the AI tool.
8. Look for `TreeNodeAccessibleObject` objects left in memory (using WinDbg, or dotMemory, or another suitable tool).

Sample app: [TreeViewSample.zip](https://github.com/dotnet/winforms/files/8731946/TreeViewSample.zip)

dmitrii-drobotov:
Fixes #7196

## Proposed changes

- Add call to `UiaDisconnectProvider` for `TreeNodeAccessibleObject`s for cases when:
    - `TreeView` is destroyed (now it only releases `TreeViewAccessibleObject`)
    - All nodes are cleared from `TreeView`
    - A node is removed from `TreeView`
- Add unit tests for above scenarios

## Customer Impact

- `TreeNodeAccessibleObject` memory will stop leaking 

## Regression? 

- No

## Risk

- Minimal

### Before

After executing steps from issue description for `TreeView` with 16 nodes:

Screenshot 2022-06-10 161803

### After

Screenshot 2022-06-13 183507

Ideally there should be 0 objects, but two empty `TreeNode[]` are still left. I tried to find the cause of it but couldn't quickly do so. It would also be present even without using accessibility tools, therefore out of scope of this task. Considering it has a very small footprint, I don't think we should raise a separate issue for it.

After `TreeView.Nodes.Clear()` — no `TreeNodeAccessibleObject` instances:

Screenshot 2022-06-13 184741

After  `TreeView.Nodes.RemoveAt()` for several nodes, leaving only 9:

Screenshot 2022-06-13 184641

## Test methodology 

- Unit testing
- Manual testing with WinDbg to check memory leak, Narrator to check for regressions similar to #6230, Inspect/AI to check that removed nodes are inaccessible there
 

## Test environment(s) 

* Windows 11 (Version 10.0.22000)
* .NET 7.0.0-preview.5.22272.3

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7296)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TreeView.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TreeView.cs
@@ -183,6 +183,19 @@ namespace System.Windows.Forms
             SetStyle(ControlStyles.UseTextForAccessibility, false);
         }
 
+        internal override void ReleaseUiaProvider(IntPtr handle)
+        {
+            base.ReleaseUiaProvider(handle);
+
+            foreach (TreeNode rootNode in Nodes)
+            {
+                foreach (TreeNode node in rootNode.GetSelfAndChildNodes())
+                {
+                    node.ReleaseUiaProvider();
+                }
+            }
+        }
+
         /// <summary>
         ///  The background color for this control. Specifying null for
         ///  this parameter sets the
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TreeNode.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TreeNode.cs
@@ -2074,7 +2074,18 @@ namespace System.Windows.Forms
                 _handle = IntPtr.Zero;
             }
 
+            ReleaseUiaProvider();
+
             treeView = null;
+        }
+
+        internal virtual void ReleaseUiaProvider()
+        {
+            if (OsVersion.IsWindows8OrGreater && _accessibleObject is not null)
+            {
+                UiaCore.UiaDisconnectProvider(AccessibilityObject);
+                _accessibleObject = null;
+            }
         }
 
         /// <summary>
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TreeNode.TreeNodeAccessibleObjectTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TreeNode.TreeNodeAccessibleObjectTests.cs
@@ -488,5 +488,95 @@ namespace System.Windows.Forms.Tests
             Assert.True(node.childCount > 0);
             Assert.False(control.IsHandleCreated);
         }
+
+        [WinFormsFact]
+        public void TreeNodeAccessibleObject_IsDisconnected_WhenTreeViewIsReleased()
+        {
+            using TreeView control = new();
+            AccessibilityObjectDisconnectTrackingTreeNode firstLevelNode = new(control);
+            control.Nodes.Add(firstLevelNode);
+            AccessibilityObjectDisconnectTrackingTreeNode secondLevelNode = new(control);
+            firstLevelNode.Nodes.Add(secondLevelNode);
+            AccessibilityObjectDisconnectTrackingTreeNode thirdLevelNode = new(control);
+            secondLevelNode.Nodes.Add(thirdLevelNode);
+            control.CreateControl();
+
+            control.ReleaseUiaProvider(control.Handle);
+
+            Assert.True(firstLevelNode.IsAccessibilityObjectDisconnected);
+            Assert.True(secondLevelNode.IsAccessibilityObjectDisconnected);
+            Assert.True(thirdLevelNode.IsAccessibilityObjectDisconnected);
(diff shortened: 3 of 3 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/TreeView.cs
```diff
@@ -182,6 +182,19 @@ public TreeView()
             SetStyle(ControlStyles.UseTextForAccessibility, false);
         }
 
+        internal override void ReleaseUiaProvider(IntPtr handle)
```
vladimir-krestov, line 185: I see one problem with it. We call it when destroying of the control. But we can destroy it when _recreating_ of the control. So please check if accessible objects work correctly in your implementation when recreating the TreeView control.
PS. We had a bug with it for TextBox in DataGridViewTextBoxCell

vladimir-krestov (commented):
Looks good! But please check RecreateControl case

src/System.Windows.Forms/src/System/Windows/Forms/TreeView.cs
dmitrii-drobotov: It's the same as with ListBox (https://github.com/dotnet/winforms/pull/7284#discussion_r897995390) — no issues on recreation.

RussKie (approved):

RussKie:

vladimir-krestov (dismissed):
👍 
Do we need CTI testing for this change? I don't see risky changes.

dmitrii-drobotov:
Just found a memory leak if node is replaced in runtime, e.g.
```
treeView.Nodes[0] = new TreeNode("...");
```
In that case old node and its children need to be disconnected as well.

dmitrii-drobotov:
> Just found a memory leak if node is replaced in runtime, e.g.
> 
> ```
> treeView.Nodes[0] = new TreeNode("...");
> ```
> 
> In that case old node and its children need to be disconnected as well.

I looked into it and found that replacing a node by index has another bug, and that fixing a memory leak for that scenario wouldn't make a lot sense without fixing the first issue. So I decided to extract it into a separate issue https://github.com/dotnet/winforms/issues/7504 which would cover both problems.

dmitrii-drobotov:
> > Just found a memory leak if node is replaced in runtime, e.g.
> > ```
> > treeView.Nodes[0] = new TreeNode("...");
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In that case old node and its children need to be disconnected as well.
> 
> I looked into it and found that replacing a node by index has another bug, and that fixing a memory leak for that scenario wouldn't make a lot sense without fixing the first issue. So I decided to extract it into a separate issue #7504 which would cover both problems.


Rebased onto the latest main, which now contains fix for #7504, and tested this scenario. As expected, this PR also fixes memory leak for node replacement by index.

Tanya-Solyanik:
> I tried to find the cause of it but couldn't quickly do so. It would also be present even without using accessibility tools, therefore out of scope of this task.

I guess that these nodes are held by a static object, most likely a static comparer for the tree nodes. Take addresses of you arrays - 
image

And call ` !GCRoot <address>` in windbg

dmitrii-drobotov:
> > I tried to find the cause of it but couldn't quickly do so. It would also be present even without using accessibility tools, therefore out of scope of this task.
> 
> I guess that these nodes are held by a static object, most likely a static comparer for the tree nodes. Take addresses of you arrays - image
> 
> And call ` !GCRoot <address>` in windbg. In case you are curious. I agree that there is no reason for hunt these down because they don't hold the Control(i.e. windows handles and other resources).

Yeah, I tried it and went a bit beyond, but at that point I was lacking knowledge on how to further identify this root object, so I decided not to proceed with it, especially that it doesn't seem critical.
image

Tanya-Solyanik:
`HandleTable` means `static object` - https://github.com/dotnet/diagnostics/issues/1922#issuecomment-763079020 and of cause you don't see it in the heap. This looks "by design".  There is no need to instigate further.

Tanya-Solyanik (approved):
Thank you!

PR #7298, opened 2022-06-14
Enable nullability in ToolStripItemInternalLayout

gpetrou:
## Proposed changes

- Enable nullability in ToolStripItemInternalLayout.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7298)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.ToolStripItemInternalLayout.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.ToolStripItemInternalLayout.cs
@@ -170,6 +170,9 @@ namespace System.Windows.Forms
 
                 if (_ownerItem is not null)
                 {
+                    // _currentLayoutOptions will always be initialised if EnsureLayout() is called,
+                    // because it will get called at least once (for _layoutData is null) and, in turn,
+                    // it'll invoke PerformLayout() that'll unconditionally invoke GetLayoutData().
                     _lastPreferredSize = _currentLayoutOptions!.GetPreferredSizeCore(constrainingSize);
                     return _lastPreferredSize;
                 }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.ToolStripItemInternalLayout.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.ToolStripItemInternalLayout.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Diagnostics.CodeAnalysis;
 using System.Drawing;
 using System.Windows.Forms.ButtonInternal;
 
@@ -46,7 +47,7 @@ namespace System.Windows.Forms
                 get
                 {
                     EnsureLayout();
-                    return _layoutData!;
+                    return _layoutData;
                 }
             }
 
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.ToolStripItemInternalLayout.cs
```diff
@@ -169,7 +166,7 @@ public virtual Size GetPreferredSize(Size constrainingSize)
 
                 if (_ownerItem is not null)
                 {
-                    _lastPreferredSize = _currentLayoutOptions.GetPreferredSizeCore(constrainingSize);
+                    _lastPreferredSize = _currentLayoutOptions!.GetPreferredSizeCore(constrainingSize);
```
RussKie, line 169: `GetLayoutData()` must be annotated with `MemeberNotNull`. Though I'm not sure, if the analyzer will see it.

`_currentLayoutOptions` will always be initialised if  `EnsureLayout()` is called, because it will get called at least once (for `_layoutData is null`) and, in turn, it'll invoke `PerformLayout()` that'll unconditionally invoke `GetLayoutData()`. So `EnsureLayout()` must be annotated with `MemeberNotNull(_currentLayoutOptions)`.

Please add a comment above `EnsureLayout` explaining this.

RussKie (changes requested):

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.ToolStripItemInternalLayout.cs
gpetrou: Unfortunately, adding `MemberNotNull` for `_currentLayoutOptions` will not work, so I added a comment instead.

RussKie: Could you please elaborate why it won't work?

image

gpetrou: This is what I see:
image
Perhaps it is because I still use VS 2022 17.2 version? I see lots of other warnings, but the build still works.

RussKie: 👍 

We still need to do this:
```diff
diff --git a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.ToolStripItemInternalLayout.cs b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.ToolStripItemInternalLayout.cs
index 6019222bb..469a5a04c 100644
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.ToolStripItemInternalLayout.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.ToolStripItemInternalLayout.cs
@@ -135,17 +135,15 @@ namespace System.Windows.Forms
             }
 
             [MemberNotNull(nameof(_layoutData))]
-            private bool EnsureLayout()
+            private void EnsureLayout()
             {
                 if (_layoutData is null || _parentLayoutData is null || !_parentLayoutData.IsCurrent(ParentInternal))
                 {
                     PerformLayout();
-                    return true;
                 }
-
-                return false;
             }
 
+            [MemberNotNull(nameof(_currentLayoutOptions))]
             private ButtonBaseAdapter.LayoutData GetLayoutData()
             {
                 _currentLayoutOptions = CommonLayoutOptions();
@@ -181,6 +179,7 @@ namespace System.Windows.Forms
             }
 
             [MemberNotNull(nameof(_layoutData))]
+            [MemberNotNull(nameof(_currentLayoutOptions))]
             internal void PerformLayout()
             {
                 _layoutData = GetLayoutData();
```

RussKie (approved):

RussKie:
Thank you

PR #7316, opened 2022-06-16
Enable nullability in TabControl

gpetrou:
## Proposed changes

- Enable nullability in TabControl.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7316)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
@@ -66,7 +66,7 @@ public partial class TabControl : Control
     private readonly User32.WM _tabBaseReLayoutMessage = User32.RegisterWindowMessageW(Application.WindowMessagesVersion + TabBaseReLayoutMessageName);
 
     // State
-    private List<TabPage>? _tabPages;
+    private readonly List<TabPage> _tabPages = new();
     private int _lastSelection;
     private short _windowId;
 
@@ -726,7 +726,7 @@ public partial class TabControl : Control
         get
         {
             int index = SelectedIndex;
-            if (index == -1 || _tabPages is null)
+            if (index == -1 || _tabPages.Count == 0)
             {
                 return null;
             }
@@ -1048,14 +1048,11 @@ public partial class TabControl : Control
 
     internal int FindTabPage(TabPage? tabPage)
     {
-        if (_tabPages is not null)
+        for (int i = 0; i < _tabPages.Count; i++)
         {
-            for (int i = 0; i < _tabPages.Count; i++)
+            if (_tabPages[i].Equals(tabPage))
             {
-                if (_tabPages[i].Equals(tabPage))
-                {
-                    return i;
-                }
+                return i;
             }
         }
 
@@ -1074,7 +1071,7 @@ public partial class TabControl : Control
             throw new ArgumentOutOfRangeException(nameof(index), index, string.Format(SR.InvalidArgument, nameof(index), index));
         }
 
-        return _tabPages![index];
+        return _tabPages[index];
     }
 
     /// <summary>
@@ -1082,7 +1079,7 @@ public partial class TabControl : Control
     /// </summary>
     protected virtual object[] GetItems()
     {
-        if (_tabPages is not null && _tabPages.Count > 0)
+        if (_tabPages.Count > 0)
         {
             return _tabPages.ToArray();
         }
@@ -1101,7 +1098,7 @@ public partial class TabControl : Control
         {
             for (int i = 0; i < tabPageCount; i++)
             {
-                result[i] = _tabPages![i];
+                result[i] = _tabPages[i];
             }
         }
 
@@ -1159,8 +1156,6 @@ public partial class TabControl : Control
 
     internal void Insert(int index, TabPage tabPage)
     {
-        _tabPages ??= new List<TabPage>();
-
         _tabPages.Insert(index, tabPage);
 
         _cachedDisplayRect = Rectangle.Empty;
@@ -1568,7 +1563,7 @@ public partial class TabControl : Control
             PInvoke.SendMessage(this, (User32.WM)PInvoke.TCM_DELETEALLITEMS);
         }
 
-        _tabPages = null;
+        _tabPages.Clear();
 
         base.RecreateHandleCore();
 
@@ -1601,7 +1596,7 @@ public partial class TabControl : Control
             PInvoke.SendMessage(this, ((User32.WM)PInvoke.TCM_DELETEALLITEMS));
         }
 
-        _tabPages = null;
+        _tabPages.Clear();
     }
 
     private void RemoveTabPage(int index)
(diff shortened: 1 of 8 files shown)


RussKie (dismissed):
I'm going to look into this in more details next week.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
```diff
@@ -71,7 +70,7 @@ public partial class TabControl : Control
         private readonly User32.WM _tabBaseReLayoutMessage = User32.RegisterWindowMessageW(Application.WindowMessagesVersion + TabBaseReLayoutMessageName);
 
         // State
-        private TabPage[] _tabPages;
+        private TabPage?[]? _tabPages;
```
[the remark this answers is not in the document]
gpetrou: I made it like that because we have existing calls like `_tabPages[_tabPageCount] = null;` and `_tabPages = null;`. Some of the `!` usages are to keep the existing tests happy and avoid changing the behavior. For example, `DeselectTab(tabPage!);`. Alternatively, we could change that to Nop or InvalidOperationException.

vladimir-krestov (commented):
1) These changes looks good, but I really don't like how `_tabPages` was implemented before. Its usage looks too confusingly and not obvious. I spent time to understand that we won't catch NRE if try to get `_tabPages`, it will never be null when getting items, but the current implementation too unsafe and unstable. We have 2 related collections: `_tabPages` array (that tries to be as a List) and `_tabCollection` TabPageCollection. It looks strange, unsafe, unstable to me. And I would prefer to rework these collections in the future, not now, because it is risky.
Just my thoughts and FYI @RussKie.

2) Using of "!" operator looks complex and unreadable to me. If we have related checks in methods (like `index < _tabPageCount`, that means that `_tabPages` is not null), why not just to add null-checks (`if (_tabPages is not null) ...`) for `_tabPages` everywhere, where we are sure that it is not null. It would be more clear and readable to me, clear VisualStudio warnings and this way allows us to avoid of using "!" operator. @gpetrou, what do you think about this approach?

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
```diff
@@ -1055,13 +1054,13 @@ internal void EndUpdate(bool invalidate)
...
 
-        internal int FindTabPage(TabPage tabPage)
+        internal int FindTabPage(TabPage? tabPage)
         {
             if (_tabPages is not null)
             {
                 for (int i = 0; i < _tabPageCount; i++)
                 {
-                    if (_tabPages[i].Equals(tabPage))
+                    if (_tabPages[i]!.Equals(tabPage))
```
dreddy-work, line 1063: This doesn't make sense. If input can be null and indexed tab page is not.

dreddy-work (changes requested):
@gpetrou , i see multiple places where `!` operator is not meaningful. A little restructure of existing code would avoid it. Do you mind take a relook and see if we can make changes but not break existing apps.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
gpetrou: Sorry, you mean you want me to do a null check before calling this instead?

gpetrou:
> i see multiple places where `!` operator is not meaningful. A little restructure of existing code would avoid it. Do you mind take a relook and see if we can make changes but not break existing apps.

Can you please give me some examples? As I mentioned in https://github.com/dotnet/winforms/pull/7316#discussion_r899995657 some usages of `!` are because I wanted to avoid changes in behavior.

gpetrou:
> 2\. Using of "!" operator looks complex and unreadable to me. If we have related checks in methods (like `index < _tabPageCount`, that means that `_tabPages` is not null), why not just to add null-checks (`if (_tabPages is not null) ...`) for `_tabPages` everywhere, where we are sure that it is not null. It would be more clear and readable to me, clear VisualStudio warnings and this way allows us to avoid of using "!" operator.

I can try to add a separate commit and see how that looks like.

RussKie:
> I can try to add a separate commit and see how that looks like.

Yes, please

dreddy-work:
By revisiting this PR, it makes more sense to change type for `_tabPages` from array to list. It isn't directly exposed outside of this class.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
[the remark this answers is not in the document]
JeremyKuhne: Why can't this be `TabPage[]?`. Where am I missing the need to have a `null` in the array?

JeremyKuhne:
> By revisiting this PR, it makes more sense to change type for _tabPages from array to list. 

That seems orthogonal. `TabPage[]?` and `List<TabPage>?` are effectively the same from a nullability standpoint. We should first try to get the nulls out of the collection, then see if it is desirable (or even possible) to make the field itself non-nullable. If there is value to making it a `List<TabPage>?` that should be called out separately.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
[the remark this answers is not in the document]
dreddy-work: while removing the tab pages, we are filling the empty entries of array with null: https://github.com/dotnet/winforms/blob/503bff4ae3c967e975836d094e036f08b3429881/src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs#L1644

gpetrou:
@dreddy-work this one should be ready for review as well.

gpetrou:
@dreddy-work do you need any more changes on this one?

dreddy-work:
> @dreddy-work do you need any more changes on this one?

@gpetrou, haven't got chance to go over this again. Will do this week and update if anything needed here.

gpetrou:
@dreddy-work kind reminder on this one.

dreddy-work:
@gpetrou, Can you consider making the `_tabPages `variable read-only and assigning the list inline? This would allow us to check for `Count` where necessary and ignore the possibility of `null`. The `RemoveAll` method could just clear the list. Avoids `!`

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
```diff
@@ -1716,7 +1707,7 @@ public void SelectTab(string tabPageName)
     {
         ArgumentNullException.ThrowIfNull(tabPageName);
 
-        TabPage tabPage = TabPages[tabPageName];
+        TabPage tabPage = TabPages[tabPageName]!;
```
dreddy-work, line 1710:
```suggestion
        TabPage? tabPage = TabPages[tabPageName];
        ArgumentNullException.ThrowIfNull(tabPage);
```

dreddy-work (commented):
Thank you. Seems some unit tests failed.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
gpetrou: This already exists inside `DeselectTab` and seems to be more appropriate there than here.

dreddy-work (approved):

PR #7317, opened 2022-06-16
Enable nullability in RelatedImageListAttribute

gpetrou:
## Proposed changes

- Enable nullability in RelatedImageListAttribute.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7317)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/RelatedImageListAttribute.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/RelatedImageListAttribute.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 namespace System.Windows.Forms
 {
     /// <summary>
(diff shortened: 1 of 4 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/RelatedImageListAttribute.cs
```diff
@@ -13,19 +11,19 @@ namespace System.Windows.Forms
     [AttributeUsage(AttributeTargets.Property, AllowMultiple = false, Inherited = true)]
     public sealed class RelatedImageListAttribute : Attribute
     {
-        private readonly string relatedImageList;
+        private readonly string _relatedImageList;
 
         public RelatedImageListAttribute(string relatedImageList)
```
RussKie, line 16: We don't make any assumptions or assertions about `relatedImageList`
```suggestion
        private readonly string? _relatedImageList;

        public RelatedImageListAttribute(string? relatedImageList)
```

ditto below

RussKie (changes requested):

src/System.Windows.Forms/src/System/Windows/Forms/RelatedImageListAttribute.cs
gpetrou: This is currently used only with strings such as `[RelatedImageList("ListView.GroupImageList")]` in the codebase and the property is called like `RelatedImageList.Split('.');` in a couple of places. So, since there is no null check before `Split` I thought it is better to make it non-nullable as a suggestion.

RussKie: This is used by (at least) the Windows Forms designer, and it's checked for nulls.
This is a public API which doesn't performs any null checks, so it means it can be a null :)

dreddy-work (approved):

RussKie (approved):

PR #7319, opened 2022-06-16
Fixing memory leaks of ComboBox elements and items accessible objects

### LINKED ISSUE #7318: ComboBox items accessible objects are leaking in memory
### .NET version

.NET 7.0 Preview 4

No

No

### Issue description

`ComboBox` items accessible objects are left in memory after the ComboBox control instance with its items was inspected with accessibility tool (like Accessibility Insights or Inspect) and the form with ComboBox is closed, garbage is collected, but accessibility tool is still opened. Same applies for case when ComboBox item is removed from the control's list.

image

### Steps to reproduce

Create a parent form with two buttons: one will be opening a child form with the ComboBox control and the other will collect the garbage.
Press the second button to open the child form with ComboBox.
Start the Accessibility Insights tools or Inspect.
Run automated checks for the ComboBox control using AI/Inspect.
Close the child form.
DON'T close the AI/Inspect tool.
Look for "ComboBox" objects left in memory (using WinDbg, or dotMemory, or another suitable tool).

vladimir-krestov:
Fixes #7318
Has the similar approach like in PR #7284
⚠️Doesn't fix memory leaks of TextPattern. It will be fixed as another fix for TextBox as well.

## Proposed changes

- Add a call of `UiaDisconnectProvider` for `ComboBox` elements (its List, TextBox, Label) and ComboBoxItemAccessibleObjects for cases when:
     - ComboBox is destroyed
     - All items are cleared from a ComboBox
     - An item is removed from a ComboBox
     - Using TextPattern explorer of Inspect

## Customer Impact

- Less memory leaks of ComboBox

## Regression? 

- No

## Risk

- Minimal

### Before
- There are some kept object in the memory after disposing a ComboBox:
image

### After
- These are no ComboBox elements or items objects in the memory after disposing (without TextPattern):
image

## Test methodology

- CTI
- Manually (via WinDbg)

## Accessibility testing 
- Using AI, Inspect, Narrator for cases:
      - Dispose a ComboBox
      - Close a form with a ComboBox (works same as disposing)
      - Remove 1 item then dispose the ComboBox
      - Clear all items then dispose the ComboBox

## Test environment(s)

- .NET 7.0 Preview 4
- Windows 11

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7319)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxAccessibleObject.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Diagnostics;
 using static System.Windows.Forms.ComboBox.ObjectCollection;
 using static Interop;
 
@@ -214,8 +215,47 @@ namespace System.Windows.Forms
                 }
             }
 
+            internal void RemoveListItemAccessibleObjectAt(int index)
+            {
+                IReadOnlyList<Entry> entries = _owningComboBox.Items.InnerList;
+                Debug.Assert(index < entries.Count);
+
+                Entry item = entries[index];
+                if (!ItemAccessibleObjects.ContainsKey(item))
+                {
+                    return;
+                }
+
+                if (OsVersion.IsWindows8OrGreater)
+                {
+                    HRESULT result = UiaCore.UiaDisconnectProvider(ItemAccessibleObjects[item]);
+                    Debug.Assert(result == HRESULT.S_OK);
+                }
+
+                ItemAccessibleObjects.Remove(item);
+            }
+
+            internal void ReleaseDropDownButtonUiaProvider()
+            {
+                if (OsVersion.IsWindows8OrGreater && _dropDownButtonUiaProvider is not null)
+                {
+                    HRESULT result = UiaCore.UiaDisconnectProvider(_dropDownButtonUiaProvider);
+                    Debug.Assert(result == HRESULT.S_OK);
+                    _dropDownButtonUiaProvider = null;
+                }
+            }
+
             internal void ResetListItemAccessibleObjects()
             {
+                if (OsVersion.IsWindows8OrGreater)
+                {
+                    foreach (ComboBoxItemAccessibleObject itemAccessibleObject in ItemAccessibleObjects.Values)
+                    {
+                        HRESULT result = UiaCore.UiaDisconnectProvider(itemAccessibleObject);
+                        Debug.Assert(result == HRESULT.S_OK);
+                    }
+                }
+
                 ItemAccessibleObjects.Clear();
             }
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxChildNativeWindow.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxChildNativeWindow.cs
@@ -42,7 +42,9 @@ namespace System.Windows.Forms
                             // Call the focus event for the new selected item accessible object provided by ComboBoxAccessibleObject.
                             // If the owning ComboBox has a custom accessible object,
                             // it should override the logic and implement setting an item focus by itself.
-                            if (before != after && _owner.IsAccessibilityObjectCreated && _owner.AccessibilityObject is ComboBoxAccessibleObject comboBoxAccessibleObject)
+                            if (before != after &&
+                                _owner.IsAccessibilityObjectCreated &&
+                                _owner.AccessibilityObject is ComboBoxAccessibleObject comboBoxAccessibleObject)
                             {
                                 comboBoxAccessibleObject.SetComboBoxItemFocus();
                             }
@@ -53,10 +55,39 @@ namespace System.Windows.Forms
                         }
 
                         break;
+                    case WM.DESTROY:
+                        AccessibleObject? accessibilityObject = GetChildAccessibleObjectIfCreated();
+
+                        if (accessibilityObject is not null)
+                        {
+                            if (Handle != IntPtr.Zero)
+                            {
+                                UiaCore.UiaReturnRawElementProvider(Handle, wParam: 0, lParam: 0, el: null);
+                            }
+
+                            if (OsVersion.IsWindows8OrGreater)
+                            {
+                                UiaCore.UiaDisconnectProvider(accessibilityObject);
+                            }
+
+                            ClearChildAccessibleObject();
+                        }
+
+                        if (_childWindowType == ChildWindowType.DropDownList)
+                        {
+                            DefWndProc(ref m);
+                        }
+                        else
+                        {
+                            _owner.ChildWndProc(ref m);
+                        }
+
+                        break;
                     default:
                         if (_childWindowType == ChildWindowType.DropDownList)
                         {
-                            DefWndProc(ref m); // Drop Down window should behave by its own.
+                            // Drop Down window should behave by its own.
+                            DefWndProc(ref m);
                         }
                         else
                         {
@@ -67,16 +98,45 @@ namespace System.Windows.Forms
                 }
             }
 
-            private ChildAccessibleObject GetChildAccessibleObject(ChildWindowType childWindowType)
-                => childWindowType switch
+            private ChildAccessibleObject GetChildAccessibleObject()
+                => _childWindowType switch
                 {
                     ChildWindowType.Edit => _owner.ChildEditAccessibleObject,
                     ChildWindowType.ListBox or ChildWindowType.DropDownList => _owner.ChildListAccessibleObject,
-                    _ => throw new ArgumentOutOfRangeException(nameof(childWindowType))
+                    _ => throw new ArgumentOutOfRangeException(nameof(_childWindowType))
                 };
+
+            private ChildAccessibleObject? GetChildAccessibleObjectIfCreated()
+                => _childWindowType switch
+                {
+                    ChildWindowType.Edit => _owner._childEditAccessibleObject,
+                    ChildWindowType.ListBox or ChildWindowType.DropDownList => _owner._childListAccessibleObject,
+                    _ => throw new ArgumentOutOfRangeException(nameof(_childWindowType))
+                };
+
+            private void ClearChildAccessibleObject()
+            {
+                if (_childWindowType == ChildWindowType.Edit)
+                {
+                    _owner.ClearChildEditAccessibleObject();
+                }
+                else if (_childWindowType == ChildWindowType.ListBox || _childWindowType == ChildWindowType.DropDownList)
+                {
+                    _owner.ClearChildListAccessibleObject();
+                }
+            }
 
             private void WmGetObject(ref Message m)
             {
+                if (m.LParamInternal != NativeMethods.UiaRootObjectId && (int)m.LParamInternal != OBJID.CLIENT)
+                {
+                    // Do default message processing.
+                    DefWndProc(ref m);
+                    return;
+                }
+
+                AccessibleObject accessibilityObject = GetChildAccessibleObject();
+
                 if (m.LParamInternal == NativeMethods.UiaRootObjectId)
                 {
                     // If the requested object identifier is UiaRootObjectId,
@@ -85,38 +145,30 @@ namespace System.Windows.Forms
                         this,
                         m.WParamInternal,
                         m.LParamInternal,
-                        GetChildAccessibleObject(_childWindowType));
+                        accessibilityObject);
 
                     return;
                 }
 
-                // See "How to Handle WM_GETOBJECT" in MSDN
-                if ((int)m.LParamInternal == OBJID.CLIENT)
+                try
                 {
-                    // Get an Lresult for the accessibility Object for this control
+                    IntPtr pUnknown = Marshal.GetIUnknownForObject(accessibilityObject!);
+
                     try
                     {
-                        // Obtain the Lresult
-                        IntPtr pUnknown = Marshal.GetIUnknownForObject(GetChildAccessibleObject(_childWindowType));
-
-                        try
-                        {
-                            m.ResultInternal = Oleacc.LresultFromObject(in IID.IAccessible, m.WParamInternal, new HandleRef(this, pUnknown));
-                        }
-                        finally
-                        {
-                            Marshal.Release(pUnknown);
-                        }
+                        m.ResultInternal = Oleacc.LresultFromObject(
+                            in IID.IAccessible,
+                            m.WParamInternal,
+                            new HandleRef(this, pUnknown));
                     }
-                    catch (Exception e)
+                    finally
                     {
-                        throw new InvalidOperationException(SR.RichControlLresult, e);
+                        Marshal.Release(pUnknown);
                     }
                 }
-                else
+                catch (Exception e)
                 {
-                    // m.lparam != OBJID_CLIENT, so do default message processing
-                    DefWndProc(ref m);
+                    throw new InvalidOperationException(SR.RichControlLresult, e);
                 }
             }
         }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ObjectCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ObjectCollection.cs
@@ -253,8 +253,7 @@ namespace System.Windows.Forms
                 }
 
                 InnerList.Clear();
-
-                OwnerComboBoxAccessibleObject?.ItemAccessibleObjects.Clear();
+                OwnerComboBoxAccessibleObject?.ResetListItemAccessibleObjects();
 
                 _owner._selectedIndex = -1;
                 if (_owner.AutoCompleteSource == AutoCompleteSource.ListItems)
@@ -360,7 +359,7 @@ namespace System.Windows.Forms
                             }
                             else
                             {
-                                OwnerComboBoxAccessibleObject?.ItemAccessibleObjects.Remove(InnerList[index]);
+                                OwnerComboBoxAccessibleObject?.RemoveListItemAccessibleObjectAt(index);
                                 InnerList.RemoveAt(index);
                             }
                         }
@@ -385,7 +384,7 @@ namespace System.Windows.Forms
                     _owner.NativeRemoveAt(index);
                 }
 
-                OwnerComboBoxAccessibleObject?.ItemAccessibleObjects.Remove(InnerList[index]);
+                OwnerComboBoxAccessibleObject?.RemoveListItemAccessibleObjectAt(index);
                 InnerList.RemoveAt(index);
 
                 if (!_owner.IsHandleCreated)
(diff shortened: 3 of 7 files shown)


RussKie (commented):
👍 

Could you please confirm that we have tests similar to https://github.com/dotnet/winforms/pull/7284 and https://github.com/dotnet/winforms/pull/7296 already in place? If not, please add those.

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.cs
```diff
@@ -3249,8 +3257,22 @@ internal override void ReleaseUiaProvider(IntPtr handle)
 
             if (IsAccessibilityObjectCreated)
             {
-                var uiaProvider = AccessibilityObject as ComboBoxAccessibleObject;
-                uiaProvider?.ResetListItemAccessibleObjects();
+                if (OsVersion.IsWindows8OrGreater)
+                {
+                    UiaCore.UiaDisconnectProvider(_childListAccessibleObject);
+                    UiaCore.UiaDisconnectProvider(_childEditAccessibleObject);
+                    UiaCore.UiaDisconnectProvider(_childTextAccessibleObject);
```
dmitrii-drobotov, line 3264: Do we need to add a check if each of these objects is not null?

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ObjectCollection.cs
```diff
@@ -253,8 +253,7 @@ internal void ClearInternal()
                 }
 
                 InnerList.Clear();
-
-                OwnerComboBoxAccessibleObject?.ItemAccessibleObjects.Clear();
+                _owner.ClearListItemAccessibleObjects();
```
dmitrii-drobotov, line 256: What do you think about using `OwnerComboBoxAccessibleObject?.ResetListItemAccessibleObjects();` here instead? After that you could remove `ComboBox.ClearListItemAccessibleObjects`.

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ObjectCollection.cs
```diff
@@ -385,7 +384,7 @@ public void RemoveAt(int index)
                     _owner.NativeRemoveAt(index);
                 }
 
-                OwnerComboBoxAccessibleObject?.ItemAccessibleObjects.Remove(InnerList[index]);
+                _owner.RemoveListItemAccessibleObjectAt(index);
```
dmitrii-drobotov, line 387: Similar to clear, I think it would be simpler to do `OwnerComboBoxAccessibleObject?.RemoveListItemAccessibleObjectAt(index);` and then remove `ComboBox.RemoveListItemAccessibleObjectAt` method.

dmitrii-drobotov (commented):
Looks good! Please see if there is a possibility to add unit tests to these changes.

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.cs
vladimir-krestov: I think no, I checked this method doesn't throw any exception for null

RussKie:
> Based on PR https://github.com/dotnet/winforms/pull/7284

Is this still true?

RussKie (commented):
👍

dmitrii-drobotov (dismissed):

vladimir-krestov:
> > Based on PR #7284
> 
> Is this still true?

Actually, I would say, I oriented on this PR implementation. My PR has similar approach. I have reworded the comment in the description.

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxAccessibleObject.cs
```diff
@@ -214,8 +215,48 @@ public override string DefaultAction
...
+                }
+
+                ItemAccessibleObjects.Remove(item);
+            }
+
+            internal void ReleaseChildUiaProvider()
+            {
+                if (OsVersion.IsWindows8OrGreater)
+                {
+                    UiaCore.UiaDisconnectProvider(_dropDownButtonUiaProvider);
```
vladimir-krestov, line 245: Added this because Narrator keeps the drop-down button after navigating in the "Scan" mode and we have to release it as well.

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxAccessibleObject.cs
```diff
@@ -214,8 +215,48 @@ public override string DefaultAction
...
+                }
+
+                ItemAccessibleObjects.Remove(item);
+            }
+
+            internal void ReleaseChildUiaProvider()
+            {
+                if (OsVersion.IsWindows8OrGreater)
+                {
+                    UiaCore.UiaDisconnectProvider(_dropDownButtonUiaProvider!);
```
vladimir-krestov, line 245: Added "!" here, because `UiaDisconnectProvider` can work with null

vladimir-krestov:
CTI approved the fix ✔️

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxAccessibleObject.cs
dreddy-work, line 245: Why are we not asserting here?

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxAccessibleObject.cs
```diff
@@ -214,8 +215,48 @@ public override string DefaultAction
...
+            }
+
             internal void ResetListItemAccessibleObjects()
             {
+                if (OsVersion.IsWindows8OrGreater)
+                {
+                    foreach (ComboBoxItemAccessibleObject itemAccessibleObject in ItemAccessibleObjects.Values)
+                    {
+                        HRESULT result = UiaCore.UiaDisconnectProvider(itemAccessibleObject);
+                        Debug.Assert(result == HRESULT.S_OK);
```
dreddy-work, line 256: nit: May be a private method that tries to disconnect and assert in case of failure. Can be reused.

vladimir-krestov: Make sense, I would like to resolve it separately. Look at #7357

dreddy-work (dismissed):

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxAccessibleObject.cs
RussKie: Do you plan to implement the proposal in #7357 before this is merged, or after? Maybe that change could be done in the context of this PR?

RussKie:
Once #7284 is merged, we can merge this one.

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxAccessibleObject.cs
vladimir-krestov: I would make this change in the context of all memory leak fixes, and I would like to consider this refactoring as a separate fix, that we can discuss separately and close ComboBox issue asap. So I plan to do #7357 after merging of our current memory leaks PRs.

Tanya-Solyanik:
Please let me review this change before merging

dmitrii-drobotov:
@vladimir-krestov can you also test if replacing an item in ComboBox (e.g. calling `comboBox.Items[0] = "123";`) creates a memory leak even with the fix?

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ObjectCollection.cs
```diff
@@ -385,7 +384,7 @@ public void RemoveAt(int index)
                     _owner.NativeRemoveAt(index);
```
Tanya-Solyanik, line 384: Should we disconnect provider on line 362 as well?

Tanya-Solyanik (commented):
Documentation says that UiaReturnRawElementProvider that returns a provider should be followed by the same one that returns null. We are not doing it for ComboBoc child native windows (Edit and list/dropdownlist). 
https://docs.microsoft.com/en-us/windows/win32/api/uiautomationcoreapi/nf-uiautomationcoreapi-uiareturnrawelementprovider#remarks

After disconnecting provider, we should set the field to null.

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxAccessibleObject.cs
```diff
@@ -214,8 +215,49 @@ public override string DefaultAction
...
+                IReadOnlyList<Entry> entries = _owningComboBox.Items.InnerList;
+                if (index >= entries.Count)
+                {
+                    return;
+                }
+
+                Entry item = entries[index];
+                if (item is null || !ItemAccessibleObjects.ContainsKey(item))
+                {
+                    return;
```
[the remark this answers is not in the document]
vladimir-krestov: 1) Actually we have a barbaric approach: `_owningComboBox.Items.InnerList[0] = null;`, it may happen somewhere in the code, some developer can do something trying to reduce some leaks or something like that. Anyway, it's a strange case, moreover, this code place will throw an exception and will show this developer on his mistake.
Otherwise, sure, you are right, there is no way to set it `null`. So, I will remove `null` check from here.
2) `ItemAccessibleObjects` collection might not contain an item, but now it won't happen. See the following case:
   - Run the app
   - Open Inspect/Narrator etc.
   - Navigate through the comboBox
   - **Turn off** the UIA tool
   - Add a new item
   - Remove it back
   
   In this case we **must not create** an accessible object for the new item while any UIA is turned off, so  `ItemAccessibleObjects` collection shouldn't contain this item accessible object. And we would catch an exception when removing this item, because the owning ComboBox has an accessible object created before and we will call `RemoveListItemAccessibleObjectAt`. Make sense, but it doesn't work now. We won't catch an exception, because we create a new item's accessible object anyway even if a UIA tool is turned off (looks incorrect to me). But we can't to fix this behavior, we can't check if a UIA tool is turned off. So we always have item's AO and won't catch the exception. I will try to add a unit test for this case, that may help in the future when we will be able to fix it.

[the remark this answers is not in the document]
Tanya-Solyanik: 1. - not sure if this is possible because InnerList is internal. https://github.com/dotnet/winforms/blob/622c99eb31452bd869d66f50712d57cdfc183b87/src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ObjectCollection.cs#L70. We do not support developers using private reflection because that would prevent us from any future code changes.
2. I agree that accessible object might not exist.

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ObjectCollection.cs
vladimir-krestov: You are right, changed it

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxAccessibleObject.cs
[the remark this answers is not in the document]
vladimir-krestov: I have found a UI case that throws an exception without `!ItemAccessibleObjects.ContainsKey(item)` check, so we must have it. Also, I added a unit test to prevent errors with it in the future.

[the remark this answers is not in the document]
Tanya-Solyanik: My initial comment applied only to `item == null` check. Sorry about the confusion. It is very reasonable and in fact desirable to not have AccessibleObjects that correspond to some of the items because we create AccessibleObjects lazily.  This is what I means when I said `I agree that accessible object might not exist.`

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.cs
```diff
@@ -3227,7 +3227,7 @@ private void ReleaseChildWindow()
             if (_childListBox is not null)
             {
                 // Need to notify UI Automation that it can safely remove all map entries that refer to the specified window.
-                ReleaseUiaProvider(_childListBox.Handle);
+                ReleaseUiaRawElementProvider(_childListBox.Handle);
```
Tanya-Solyanik, line 3230: Did you see any problems in moving this code to the Native windows, where providers are registered in WM_GETOBJECT  handler?

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.cs
```diff
@@ -3639,7 +3665,7 @@ private void WmParentNotify(ref Message m)
                 if (_childDropDown is not null)
                 {
                     // Need to notify UI Automation that it can safely remove all map entries that refer to the specified window.
-                    ReleaseUiaProvider(_childDropDown.Handle);
+                    ReleaseUiaRawElementProvider(_childDropDown.Handle);
```
Tanya-Solyanik, line 3668: Are you guaranteed that Handle exists at this moment? Child windows are destroyed separately from the combobox.

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.cs
```diff
@@ -3227,7 +3227,7 @@ private void ReleaseChildWindow()
             if (_childListBox is not null)
             {
                 // Need to notify UI Automation that it can safely remove all map entries that refer to the specified window.
```
Tanya-Solyanik, line 3229: Note that comment on line 3222 is out of date.
`// We do not use UI Automation provider for child edit, so do not need to release providers.`
we register providers for all children of combo in Core.

vladimir-krestov:
> @vladimir-krestov can you also test if replacing an item in ComboBox (e.g. calling `comboBox.Items[0] = "123";`) creates a memory leak even with the fix?

Checked this case. No memory leaks were found.

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.cs
vladimir-krestov: Reworked it and used your implementation https://github.com/Tanya-Solyanik/winforms/commit/fa2efb8f13cbac9b8b518b5de0443372b0261ec5

vladimir-krestov: Removed all calls of releasing for native elements, because they disconnect in `WM.DESTROY` case

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.cs
vladimir-krestov: No, so took your implementation. Now it has Handle checks

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.cs
vladimir-krestov: Good, but I removed these calls, because now we call them in `WM.DESTROY` case (from https://github.com/Tanya-Solyanik/winforms/commit/fa2efb8f13cbac9b8b518b5de0443372b0261ec5#)

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxChildNativeWindow.cs
```diff
@@ -67,16 +98,46 @@ protected override void WndProc(ref Message m)
                 }
             }
 
-            private ChildAccessibleObject GetChildAccessibleObject(ChildWindowType childWindowType)
-                => childWindowType switch
+            // TODO: these objects sould be fields in this file and accessed through getters on ComboBox class This is rediculous
```
vladimir-krestov, line 101: @Tanya-Solyanik, 
1. I would rework the child native elements in .NET 8.0. I would consider creating separate classes for every element to avoid using `switch` in cases like this. Eg. `ComboBoxDropDownNativeWindow : ComboBoxChildNativeWindow`. I think, we can consider if we need it later.
2. Can we consider reworking of these fields in .NET 8.0 following your TODO comment?

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxChildNativeWindow.cs
```diff
@@ -67,16 +98,46 @@ protected override void WndProc(ref Message m)
...
                     ChildWindowType.Edit => _owner.ChildEditAccessibleObject,
                     ChildWindowType.ListBox or ChildWindowType.DropDownList => _owner.ChildListAccessibleObject,
-                    _ => throw new ArgumentOutOfRangeException(nameof(childWindowType))
+                    _ => throw new ArgumentOutOfRangeException(nameof(_childWindowType))
+                };
+
+            private ChildAccessibleObject? GetChildAccessibleObjectIfCreated()
+                => _childWindowType switch
+                {
+                    ChildWindowType.Edit => _owner._childEditAccessibleObject,
```
vladimir-krestov, line 113: @Tanya-Solyanik, `_childEditAccessibleObject` is accessible in this place, so we can use this way. But I might set null value for it this way too (line 122), but I kept `ClearChildEditAccessibleObject` method call for self-documenting code. Is it ok?

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.cs
```diff
@@ -3219,25 +3223,18 @@ private void ReleaseChildWindow()
...
             {
-                // We do not use UI Automation provider for child edit, so do not need to release providers.
                 _childEdit.ReleaseHandle();
                 _childEdit = null;
             }
 
             if (_childListBox is not null)
             {
-                // Need to notify UI Automation that it can safely remove all map entries that refer to the specified window.
-                ReleaseUiaProvider(_childListBox.Handle);
```
vladimir-krestov, line 3230: FYI: Removed it, because `_childListBox` do it itself when destroying. The same for the rest

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxChildNativeWindow.cs
Tanya-Solyanik: Sure, but I would prefer this done ASAP in NET8 because it's too ugly right now.

Tanya-Solyanik: Please create a work item of a PR targeting 8

Tanya-Solyanik (dismissed):
Please add a link to workitem to the TODO comment

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxChildNativeWindow.cs
Tanya-Solyanik: We are tracking a clean-up item for this code, these methods might change later anyway.

John-Qiao:
@vladimir-krestov we have completed the private testing for this PR, and memory leaks issue is fixed and no new regression issue found.

Tanya-Solyanik (dismissed):

Tanya-Solyanik (approved):

PR #7325, opened 2022-06-21
Over-release now that `VARIANT` is using `IDisposable`.

AaronRobinsonMSFT:
There is now an over-release in the `VARIANT` tests. This is caused by `VARIANT` now implementing `IDisposable`. 

Found while investigating https://github.com/dotnet/runtime/issues/70973.

It was also observed that the use of the `using var ... = new VARIANT(...)` has left a lot of cruft. Most places also continue to call `Close()` or `Dispose()` even though they have the `using var` semantic. I highly recommend someone go through this and clean up this as it is very confusing.

/cc @JeremyKuhne @RussKie 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7325)


Tanya-Solyanik (approved):

RussKie:
Thank you!

AaronRobinsonMSFT:
There are some other issues here. One of them is the following where the assert is causing the test host to crash. The OS is Windows 11.

https://github.com/dotnet/winforms/blob/0aaae6b455fd229bfb073d6c0687bbde20a1620f/src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs#L308-L315

AaronRobinsonMSFT:
There is also still the `IMallocSpy` issue during COM apartment clean-up.

dreddy-work:
> There are some other issues here. One of them is the following where the assert is causing the test host to crash. The OS is Windows 11.
> 
> https://github.com/dotnet/winforms/blob/0aaae6b455fd229bfb073d6c0687bbde20a1620f/src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs#L308-L315

@AaronRobinsonMSFT , which test was this? Did you run the test locally as well?

AaronRobinsonMSFT:
I'd appreciate another look here. I've removed the `IMallocSpy` implementation from the codebase. It was only used to validate two simple internal data types that are trivial to audit visually. The issue here is the impact of `IMallocSpy` is very difficult to address because of how it creeps in. However, I am planning on adding a test in dotnet/runtime to ensure we don't regress this, but for winforms I don't see the value and it is leading to troublesome clean-up during application shutdown.

AaronRobinsonMSFT:
> @AaronRobinsonMSFT can you add some more details on how `IMallocSpy` doesn't play nice?

Sure. `IMallocSpy` isn't really the issue itself. The real issue is that it is implemented in managed code. This turned out to have benefits and substantial issues. A benefit was that it helped track down places in the CLR where we weren't transitioning to preemptive mode (i.e., non-GC controlled) and so I fixed those bugs.

The issues overshadow these wins though. The biggest annoyance is that it is close to impossible to properly detach the `IMallocSpy` from the system when running tests correctly. Once an `IMallocSpy` is registered it can't be revoked until all registered outstanding allocations are cleaned up. This means that even if a test registers and revokes the `IMallocSpy`, it is still global and therefore any other concurrently running tests will be able to retain it while they are being cleaned up. This then becomes a race between attempting to track and untrack allocations across all tests and their associated scenarios. Given the number of tests this becomes difficult.

The part that makes this substantially more difficult is when any cross-apartment marshalling occurs, because that also allocates memory that `IMallocSpy` monitors. The RPC data structures can live far longer than the average test. During process shutdown, long after the CLR has terminated and possibly unloaded, the COM subsystem does a clean-up and, of course, tries to call the registered `IMallocSpy`. This can fail in unpredictable ways because the CLR is shutdown but worse the implemented .NET assembly may have already been unloaded.

The gist here is `IMallocSpy` should not be implemented in managed code in practice. It is helpful for the runtime to find issues but it comes with far too many issues to be used in practice outside of a runtime test.

RussKie:
Thank you @AaronRobinsonMSFT. 

@JeremyKuhne I'm going to merge this to unblock the backlog. Maybe we could get together with @AaronRobinsonMSFT and chat about options to tests
https://github.com/dotnet/winforms/blob/0aaae6b455fd229bfb073d6c0687bbde20a1620f/src/System.Windows.Forms.Primitives/src/Interop/Ole32/Interop.CALPOLESTR.cs#L39-L47
https://github.com/dotnet/winforms/blob/0aaae6b455fd229bfb073d6c0687bbde20a1620f/src/System.Windows.Forms.Primitives/src/Interop/Ole32/Interop.CADWORD.cs#L30

RussKie (approved):

Issue #4207, opened 2020-11-10
[Accessibility] Missing the "Items" list level for DomainUpDown control in the inspect tool


Jessie-Zhang01:
* .NET Core Version: 
.NET SDK 6.0.100-alpha.1.20559.3

* Have you experienced this same bug with .NET Framework?: 
No

**More info**

1. The issue also can reproduce on .NET 5.0 RC2.
2. The issue can't reproduce on .NET SDK 3.1 and .NET 4.8.

**Problem description:**
Missing the "Items" list level for DomainUpDown control in the inspect tool with UI Automation mode.

**Minimal repro:**

1. Create a .NET Core project with a DomainUpDown control.
2. Build and run the core project.
3. Observe the DomainUpDown control in the Inspect tool on UI Automation mode.

Or

1. Open the attached project.
2. Build and run the core project.
3. Observe the domainUpDown1 control in the Inspect tool on UI Automation mode.
[WindowsFormsApp6.zip](https://github.com/dotnet/winforms/files/5516503/WindowsFormsApp6.zip)

**Expected behavior:**
It displays the correct levels for domainUpDown1 control in the Inspect tool.
![image](https://user-images.githubusercontent.com/45864649/98657047-673e8700-237c-11eb-87e1-a87e70597443.png)

**Actual Behavior:**
Missing the "linkLabel1 link" level for domainUpDown1 control in the inspect tool.
![image](https://user-images.githubusercontent.com/45864649/98657393-ca301e00-237c-11eb-9d5f-bb7e24049802.png)

ArtemTatarinov:
After discussion with @merriemcgaw we decided to close this issue. The Items list of the DomainUpDown control couldn't be visible at UI, there's no way to show them as a dropdown list, etc. The user cannot interact with them as a list from UI, and then they aren't needed in the control's hierarchy shown by the Inspect. So, not showing the "Items" list in the Inspect tool is a new, but expected behavior. The current item of DomaunUpDown is still visible at the Inspect:
![image](https://user-images.githubusercontent.com/87859299/137075819-4b6595bc-b44f-467e-83df-b1e92da744f8.png)

PR #7330, opened 2022-06-21
Fixing memory leaks of Numeric/DomainUpDown elements accessible objects

### LINKED ISSUE #7328: DomainUpDown and its elements accessible objects are leaking in memory
### .NET version

.NET 7.0 Preview 4

No

_No response_

### Issue description

DomainUpDown items accessible objects are left in memory after the DomainUpDown control instance with its items was inspected with accessibility tool (like Accessibility Insights or Inspect) and the form with DomainUpDown is closed, garbage is collected, but accessibility tool is still opened. 

image

### Steps to reproduce

Create a parent form with two buttons: one will be opening a child form with the DomainUpDown control and the other will collect the garbage.
Press the second button to open the child form with DomainUpDown .
Start the Accessibility Insights tools or Inspect.
Run automated checks for the DomainUpDown control using AI/Inspect.
Close the child form.
DON'T close the AI/Inspect tool.
Look for "UpDown" objects left in memory (using WinDbg, or dotMemory, or another suitable tool).

vladimir-krestov:
Fixes #7328
⚠️Doesn't fix memory leaks of TextPattern. It will be fixed as another fix for TextBox as well.

## Proposed changes

- Rework and refactor accessible objects implementation of UpDown controls. Removed `ItemList` and Item accessible objects, because they are unnecessary. Moved the implementation to the base class. [This comment](https://github.com/dotnet/winforms/issues/4207#issuecomment-941950437) describes why we don't keep ItemsList anymore. 
- Add a call of `UiaDisconnectProvider` for `Numeric/DomainUpDown` elements (its textbox and buttons) when the control is disposing.

## Customer Impact

- Less memory leaks

## Regression? 

- No

## Risk

- Minimal

### Before

- There are some left objects in memory due to `UpDownEdit` and `UpDownButtons`:
image

### After
- Memory is clear after disposing:
image

## Test methodology
- CTI
- Manual
- Unit tests

## Accessibility testing 
- Using Narrator/Inspect and WinDbg

## Test environment(s) 
- .NET 7.0 Preview 4
- Windows 11

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7330)


src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainItemAccessibleObject.cs
```diff
@@ -1,67 +0,0 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
-// The .NET Foundation licenses this file to you under the MIT license.
-// See the LICENSE file in the project root for more information.
-
-namespace System.Windows.Forms
-{
-    public partial class DomainUpDown
-    {
-        public class DomainItemAccessibleObject : AccessibleObject
```
RussKie, line 9: ❗ We can't remove public API, this is a binary breaking change.
https://docs.microsoft.com/dotnet/api/system.windows.forms.domainupdown.domainitemaccessibleobject#applies-to

I have checked [public GitHub repos](https://github.com/search?l=C%23&p=14&q=DomainItemAccessibleObject&type=Code), and this types doesn't appear to be used. [.NET API catalog](https://apisof.net/catalog/3cc9a46b-db22-be36-1ac4-6f7b88abe571) doesn't appear to have any telemetry for this type either.
Since the docs have a disclaimer that this type shouldn't be used directly I think we _can_ obsolete this API with the error and replace implementations with `NotImplementedExceptions`.
image

Still, we have to follow the obsoletion process https://github.com/dotnet/designs/blob/main/accepted/2020/better-obsoletion/better-obsoletion.md.

In the subsequent release (e.g., in .NET 8 timeframe) we can consider removing the type completely.

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainUpDownAccessibleObject.cs
```diff
@@ -1,86 +0,0 @@
...
-// The .NET Foundation licenses this file to you under the MIT license.
-// See the LICENSE file in the project root for more information.
-
-using static Interop;
-
-namespace System.Windows.Forms
-{
-    public partial class DomainUpDown
-    {
-        public class DomainUpDownAccessibleObject : ControlAccessibleObject
```
RussKie, line 11: ditto

RussKie (changes requested):

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainItemAccessibleObject.cs
vladimir-krestov: Good catch! Will return this type, thank you!

vladimir-krestov:
PS. These controls and their AO already have unit tests. There is no ability to add unit tests for the memory leak fix (for `UiaDisconnectProvider` method).

RussKie (dismissed):
👍 

The team has discussed the obsoletion, and it was decided that we should obsolete the two defunct types with `[ObsoleteAttribute(..., IsError = false)]`. Please raise a new issue to track this work, and I'll work with you on that once this is merged.

RussKie:
Multile test failures:
```

System.Windows.Forms.Tests.DirectionButtonAccessibleObjectTests.DirectionButtonAccessibleObject_Invoke_ChangesValue_InDomainUpDown(Int32 childId, Int32 expectedIndex) in /_/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DirectionButtonAccessibleObjectTests.cs:line 40
```

vladimir-krestov:
> we should obsolete the two defunct types with `[ObsoleteAttribute(..., IsError = false)]`

#7344

src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.UpDownButtons.UpDownButtonsAccessibleObject.cs
```diff
@@ -115,6 +115,15 @@ public override string Name
 
                 public override AccessibleObject Parent => _owner.AccessibilityObject;
 
+                internal void ReleaseChildUiaProviders()
+                {
+                    if (OsVersion.IsWindows8OrGreater)
+                    {
+                        UiaCore.UiaDisconnectProvider(_upButton!);
+                        UiaCore.UiaDisconnectProvider(_downButton!);
```
vladimir-krestov, line 123: Added these changes to release navigation button AOs that Narrator keeps.
Used "!" here, because we don't care if these field are null. `UiaDisconnectProvider` doesn't throw an exception for null.
Moreover, 99% that these field won't be null because they will be created in first call of `FragmentNavigate`.

vladimir-krestov:
CTI approved the fix ✔️

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainUpDownAccessibleObject.cs
```diff
@@ -2,36 +2,18 @@
...
-using static Interop;
-
 namespace System.Windows.Forms
 {
     public partial class DomainUpDown
     {
+        // This class can be removed in the future when it is marked as obsolete.
+        // This class is not needed to be used as an accessible object of DomainUpDown,
+        // UpDownBaseAccessibleObject does it well.
+        // We can't remove this class just like that because it's a public API.
```
RussKie, line 12: 👍

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainUpDownAccessibleObject.cs
```diff
@@ -2,36 +2,18 @@
...
         public class DomainUpDownAccessibleObject : ControlAccessibleObject
         {
-            private DomainItemListAccessibleObject? _domainItemList;
```
RussKie, line 13: I've looked at the dotnet/runtime codebase, and no code is being deleted when types/members are marked as obosolete. We shouldn't do it either.

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.cs
```diff
@@ -215,9 +215,9 @@ public bool Sorted
         ///  should not call base.CreateAccessibilityObject.
         /// </summary>
         protected override AccessibleObject CreateAccessibilityInstance()
-        {
-            return new DomainUpDownAccessibleObject(this);
-        }
+            // Remove this when remove public `DomainUpDownAccessibleObject`.
+            // The base class returns a correct and full object for this control.
+            => new DomainUpDownAccessibleObject(this);
```
RussKie, line 220: Could you please help me understanding this comment?

src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.UpDownButtons.UpDownButtonsAccessibleObject.cs
RussKie: > `UiaDisconnectProvider` doesn't throw an exception for null.

Then we need to update the NRTA for this API, and don't use `!`.

src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.cs
```diff
@@ -214,6 +214,13 @@ public BorderStyle BorderStyle
             }
         }
 
+        /// <summary>
+        ///  Constructs the new instance of the accessibility object for this control. Subclasses
+        ///  should not call base.CreateAccessibilityObject.
+        /// </summary>
```
RussKie, line 220: This is a public API, and it needs to be [documented/annotated correctly](https://docs.microsoft.com/dotnet/api/system.windows.forms.control.createaccessibilityinstance). In this case though, it doesn't look like there are any special conditions to document, and the docs will be inherited from the base. Please use `<inheritdoc />` instead.

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainItemAccessibleObject.cs
```diff
@@ -6,6 +6,9 @@ namespace System.Windows.Forms
 {
     public partial class DomainUpDown
     {
+        // This class can be removed in the future when it is marked as obsolete.
+        // This class is not used anyhow for building of DomainUpDown accessibility tree.
+        // We can't remove this class just like that because it's a public API.
```
RussKie, line 11:
```suggestion
        // This class is not used anyhow for building of DomainUpDown accessibility tree, but
        // we can't remove this class just like that because it's a public API.
        // See https://github.com/dotnet/winforms/issues/7344 for more details.
```

RussKie (changes requested):
A little more work to do

src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.UpDownButtons.UpDownButtonsAccessibleObject.cs
vladimir-krestov: I wasn't right. This API doesn't support null arguments. Yes, we can call it with null and we won't catch an exception, but in this case it returns an [error code](https://docs.microsoft.com/en-us/windows/win32/api/uiautomationcoreapi/nf-uiautomationcoreapi-uiadisconnectprovider#return-value).

src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.cs
vladimir-krestov: Doesn't it inherit by default without `<inheritdoc />`?

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.cs
vladimir-krestov: The base `CreateAccessibilityInstance` method already returns a correct object for `DomainUpDown`, we don't need `DomainUpDownAccessibleObject` for it. I proposed to remove this line when `DomainUpDownAccessibleObject` is removed or marked as "obsolete".

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainUpDownAccessibleObject.cs
vladimir-krestov: Returned all back. Anyway, there are no leaks.

src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.cs
vladimir-krestov: @RussKie, please check it

src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.UpDownButtons.UpDownButtonsAccessibleObject.cs
```diff
@@ -115,6 +116,26 @@ public override string Name
 
                 public override AccessibleObject Parent => _owner.AccessibilityObject;
 
+                internal void ReleaseChildUiaProviders()
+                {
+                    if (OsVersion.IsWindows8OrGreater)
```
dreddy-work, line 121: i see OS check was already made by caller. Do you expect this to be reused elsewhere?

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.cs
vladimir-krestov: Removed this comment, we can't remove public `DomainUpDownAccessibleObject`.

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainUpDownAccessibleObject.cs
```diff
@@ -8,6 +8,10 @@ namespace System.Windows.Forms
 {
     public partial class DomainUpDown
     {
+        // This class can be removed in the future when it is marked as obsolete.
+        // This class is not needed to be used as an accessible object of DomainUpDown,
+        // UpDownBaseAccessibleObject does it well.
+        // We can't remove this class just like that because it's a public API.
         public class DomainUpDownAccessibleObject : ControlAccessibleObject
```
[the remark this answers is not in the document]
vladimir-krestov: No, a public class can't be inherited from a private/internal class (VS throws an error)

src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.UpDownBaseAccessibleObject.cs
```diff
@@ -0,0 +1,30 @@
...
+        private class UpDownBaseAccessibleObject : ControlAccessibleObject
+        {
+            public UpDownBaseAccessibleObject(UpDownBase owner) : base(owner)
+            { }
+
+            public override AccessibleRole Role
+            {
+                get
+                {
+                    AccessibleRole role = Owner.AccessibleRole;
```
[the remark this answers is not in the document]
vladimir-krestov: I wouldn't take this way because we call the complex `AccessibleRole` getter 2 times, that is not productive.
```cs
public override AccessibleRole Role
     => Owner.AccessibleRole != AccessibleRole.Default
          ? Owner.AccessibleRole
          : AccessibleRole.SpinButton;
```

So, I will use this approach:
```cs
public override AccessibleRole Role
{
     get
     {
          AccessibleRole role = Owner.AccessibleRole;

          return role != AccessibleRole.Default
               ? role
               : AccessibleRole.SpinButton;
     }
}
```

src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.UpDownButtons.UpDownButtonsAccessibleObject.cs
```diff
@@ -115,6 +116,26 @@ public override string Name
...
+                    {
+                        HRESULT result;
+
+                        if (_upButton is not null)
+                        {
+                            result = UiaCore.UiaDisconnectProvider(_upButton);
+                            Debug.Assert(result == HRESULT.S_OK);
```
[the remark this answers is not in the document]
vladimir-krestov: @Tanya-Solyanik, I saw your [comment](https://github.com/dotnet/winforms/pull/7319#pullrequestreview-1050060036), but I don't see the reason, why do we need to do it? There are no leaks after disposing, so what is the reason?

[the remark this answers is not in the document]
Tanya-Solyanik: Reason is to not disconnect twice. ReleaseUIAProvider is called from WM_DESTROY message handler. We are usually disposing managed controls in response to this message as well. However in the Dispose sequence we can also reference the _upButton provider. We should not do it because this provider is not in the accessibility client process space anymore. Also this is a guard against a potential double-disconnect.

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainUpDownAccessibleObject.cs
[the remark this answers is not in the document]
Tanya-Solyanik: Good point, but you can have a private field `UpDownBaseAccessibleObject` in this class and redirect all calls to the corresponding calls in _upDownBaseAccessibleObject. 
The reason is to make the code obviously redundant rather than trying to explain that in the comment.

src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.UpDownBaseAccessibleObject.cs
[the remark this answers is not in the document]
Tanya-Solyanik: Sounds good, but I don't see this change applied, thus re-activating the comment.

src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.cs
Tanya-Solyanik: I think that `<inheritdoc />` is redundant in this case, intellisense reads the comment from the base. @RussKie - are you ok with removal of this comment?

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainUpDownAccessibleObject.cs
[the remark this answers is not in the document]
vladimir-krestov: I wouldn't do it.
1. I don't want change "private" to "internal" of `UpDownBaseAccessibleObject` for it only
2. I can create `_upDownBaseAccessibleObject` field, but there is no way to set some instance for it
3. The implementation will be bulky and strange
4. We have no time before RC2

A comment much simpler and faster. Do you agree?

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainItemListAccessibleObject.cs
```diff
@@ -6,6 +6,8 @@ namespace System.Windows.Forms
 {
     public partial class DomainUpDown
     {
+        // This class is not used anyhow for building of DomainUpDown accessibility tree.
+        // But it can't be removed, because it's used as a part of public obsolete DomainUpDownAccessibleObject.
         internal class DomainItemListAccessibleObject : AccessibleObject
```
vladimir-krestov, line 11: I want to delete it at all. It's used for private filed and property of obsolete `DomainUpDownAccessibleObject`, so removing of this file and the private filed and property won't affect users. I don't see risks without cases of using reflection by users (unlikely). So, @Tanya-Solyanik, @RussKie, can I remove this file?

src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.UpDownButtons.UpDownButtonsAccessibleObject.cs
vladimir-krestov: Good point, no, it's won't be reused. Thanks, will remove this extra check

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainUpDownAccessibleObject.cs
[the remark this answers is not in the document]
Tanya-Solyanik: No, I don't agree, comment is not obviously true if looking a this file only - one has to read 2 classes instead of seeing obvious redirect from one to another. Changing private to internal is not a breaking change. I think that number of lines of code will be less than it is now.

However this is not worth a discussion.

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainItemListAccessibleObject.cs
Tanya-Solyanik: I suggest removing this file in NET8, when we properly obsolete the unused objects. Please add it to the follow up work item.

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainItemAccessibleObject.cs
```diff
@@ -6,6 +6,8 @@ namespace System.Windows.Forms
 {
     public partial class DomainUpDown
     {
+        // We can't remove this class just like that because it's a public API.
+        [Obsolete("This public class is not necessary anymore for building of DomainUpDown accessibility tree after https://github.com/dotnet/winforms/issues/4207.", false)]
```
Tanya-Solyanik, line 10: Is text of this warning visible to developer who uses this class? I think the message should call out `nameof(DomainUpDown.DomainItemAccessibleObject)` rather than say "this class".
You should be able to use operator `nameof` in the warning text.

src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.UpDownButtons.UpDownButtonsAccessibleObject.cs
```diff
@@ -115,6 +116,27 @@ public override string Name
...
+                {
+                    HRESULT result;
+
+                    if (_upButton is not null)
+                    {
+                        result = UiaCore.UiaDisconnectProvider(_upButton);
+                        Debug.Assert(result == HRESULT.S_OK);
+                    }
+
+                    _upButton = null;
```
Tanya-Solyanik, line 129: This can be executed only when we know that the provider is not null. I.e. move to after line 126. And the same with _downButton

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainItemAccessibleObject.cs
Tanya-Solyanik: @JeremyKuhne - could you please review this obsoletion?

JeremyKuhne: This was never meant to be used from user code anyway. It should get:

``` C#
[Obsolete($"{nameof(DomainItemAccessibleObject)} is no longer used to provide accessible support for {nameof(DomainUpDown)} items.", error: false)] 
```

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainItemListAccessibleObject.cs
JeremyKuhne: I wouldn't bother keeping this as there is no direct way to create it as an end user. All we do is cast to it (for some reason). Just kill this and change `DomainItemAccessibleObject.Parent` to always return null.

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainUpDownAccessibleObject.cs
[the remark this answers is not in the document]
JeremyKuhne: Same comment as https://github.com/dotnet/winforms/pull/7330/files#r938267721

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainItemAccessibleObject.cs
Tanya-Solyanik: Thank you, Is this a compile time error?

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainUpDownAccessibleObject.cs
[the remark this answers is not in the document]
Tanya-Solyanik: @JeremyKuhne - I don't follow your suggestion. Are you saying we can make `DomainUpDownAccessibleObject` obsolete now?

My comment was about making code in this class look obsolete - for example copy/paste content of UpDownBaseAccessibleObject.cs

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainItemListAccessibleObject.cs
Tanya-Solyanik: Sounds good

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainUpDownAccessibleObject.cs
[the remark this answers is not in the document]
JeremyKuhne: Things that are for internal use only per the docs should be hidden from intellisense.

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainItemListAccessibleObject.cs
vladimir-krestov: Removed the class, but it changed NRTA of `DomainItemAccessibleObject.Parent.get` (see `PublicAPI.Shipped.txt`).
For the rest, everything is ok.

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainUpDownAccessibleObject.cs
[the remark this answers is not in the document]
vladimir-krestov: @Tanya-Solyanik, looks like I see your idea, I have reworked this class to reflect on `UpDownBaseAccessibleObject`, please check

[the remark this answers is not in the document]
Tanya-Solyanik: Hm, they are still not identical.

src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.cs
```diff
@@ -454,6 +457,20 @@ internal override Rectangle ApplyBoundsConstraints(int suggestedX, int suggested
             return base.ApplyBoundsConstraints(suggestedX, suggestedY, proposedWidth, PreferredHeight);
         }
 
+        internal override void ReleaseUiaProvider(IntPtr handle)
+        {
+            if (IsAccessibilityObjectCreated && OsVersion.IsWindows8OrGreater)
+            {
+                (_upDownButtons.AccessibilityObject as UpDownButtons.UpDownButtonsAccessibleObject)?.ReleaseChildUiaProviders();
+                HRESULT result = UiaCore.UiaDisconnectProvider(_upDownButtons.AccessibilityObject);
```
Tanya-Solyanik, line 465: It is possible that we are forcing creation of AOs here, can we delegate disconnect to the `UpDownButtons` class? For example implement internal UpDownButtons.DisconnectProvider method?

Tanya-Solyanik: And the same for UpDownEdit

Tanya-Solyanik (commented):
I added a couple of comments, the rest looks good.

@JeremyKuhne - Since DomainUpDownAccessibleObject class is obsolete now, would it make sense to remove the Create method from DomainUpDown control and rely on the UpDownBase create method?

src/System.Windows.Forms/src/misc/Obsoletions.cs
```diff
@@ -0,0 +1,20 @@
...
+internal static class Obsoletions
+{
+    internal const string SharedUrlFormat = "https://aka.ms/winforms-warnings/{0}";
+
+    // Please see docs\project\list-of-diagnostics.md for instructions on the steps required
+    // to introduce a new obsoletion, apply it to downlevel builds, claim a diagnostic id,
+    // and ensure the "aka.ms/dotnet-warnings/{0}" URL points to documentation for the obsoletion
+    // The diagnostic ids reserved for obsoletions are SYSLIB0### (SYSLIB0001 - SYSLIB0999).
+
+    internal const string DomainUpDownAccessibleObjectMessage = $"{nameof(System.Windows.Forms.DomainUpDown.DomainUpDownAccessibleObject)} is no longer used to provide accessible support for {nameof(System.Windows.Forms.DomainUpDown)} items.";
```
Tanya-Solyanik, line 15:
```suggestion
    internal const string DomainUpDownAccessibleObjectMessage = $"{nameof(System.Windows.Forms.DomainUpDown.DomainUpDownAccessibleObject)} is no longer used to provide accessible support for {nameof(System.Windows.Forms.DomainUpDown)} controls.";
```

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainUpDownAccessibleObject.cs
```diff
@@ -2,85 +2,30 @@
...
 
-using static Interop;
+using System.ComponentModel;
 
 namespace System.Windows.Forms
 {
     public partial class DomainUpDown
     {
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        [Obsolete($"{nameof(DomainUpDownAccessibleObject)} is no longer used to provide accessible support for {nameof(DomainUpDown)} control.", error: false)]
```
Tanya-Solyanik, line 12: Did you intend to use the new message defines in `obsoletions.cs`?

docs/list-of-diagnostics.md
```diff
@@ -0,0 +1,38 @@
...
+* Register the `WINFORMS0###` URL in `aka.ms`
+    * The vanity name will be `winforms-warnings/winforms0###`
+    * Ensure the link's group owner matches the group owner of `winforms-warnings/winforms0001`
+    * Connect with `@igveliko` or `@gewarren` with any questions
+
+### Obsoletion Diagnostics (`WINFORMS0001` - `WINFORMS0999`)
+
+| Diagnostic ID     | Description |
+| :---------------- | :---------- |
+|  __`WINFORMS0001`__ | `DomainUpDown.DomainUpDownAccessibleObject` is no longer used to provide accessible support for `DomainUpDown` items. |
```
Tanya-Solyanik, line 36: We are using a different ID format here - https://github.com/dotnet/winforms/blob/84042a843540a07a6c64fb6fb36006e5f2fe8fd4/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Message.cs#L76

vladimir-krestov: So here are 2 options:
1. Change the new Id names and the new documentation
2. Change `Message.cs` Id naming to the new format

Which one should I use?

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainUpDownAccessibleObject.cs
vladimir-krestov: Implemented, thanks

src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.cs
vladimir-krestov: Great catch! Actually, the buttons and the edit are controls, so they can release their AO themself. Moreover, now I don't do double disconnecting for them.

docs/list-of-diagnostics.md
Tanya-Solyanik: Follow the existing pattern. Also please add message warnings to the .md file

src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.UpDownButtons.cs
```diff
@@ -341,6 +341,18 @@ protected override void OnPaint(PaintEventArgs e)
             protected virtual void OnUpDown(UpDownEventArgs upevent)
                 => _upDownEventHandler?.Invoke(this, upevent);
 
+            internal override void ReleaseUiaProvider(IntPtr handle)
+            {
+                if (IsAccessibilityObjectCreated
```
Tanya-Solyanik, line 346: I think it makes sense to add a `bool TryGetAccessibilityObject(out AccessibleObject accessibeObject)` to control class to handle use cases like this one. It probably logically belongs to your refactoring PR.

src/System.Windows.Forms/src/misc/Obsoletions.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace System;
+
+internal static class Obsoletions
+{
+    internal const string SharedUrlFormat = "https://aka.ms/winforms-warnings/{0}";
+
+    // Please see docs\project\list-of-diagnostics.md for instructions on the steps required
+    // to introduce a new obsoletion, apply it to downlevel builds, claim a diagnostic id,
+    // and ensure the "aka.ms/dotnet-warnings/{0}" URL points to documentation for the obsoletion
+    // The diagnostic ids reserved for obsoletions are SYSLIB0### (SYSLIB0001 - SYSLIB0999).
```
Tanya-Solyanik, line 13: Why does this comment mention SYSLIB? Should it be replaced with WFDEVXXX?

src/System.Windows.Forms/src/misc/Obsoletions.cs
```diff
@@ -0,0 +1,20 @@
...
+
+    internal const string DomainUpDownAccessibleObjectMessage = $"{nameof(System.Windows.Forms.DomainUpDown.DomainUpDownAccessibleObject)} is no longer used to provide accessible support for {nameof(System.Windows.Forms.DomainUpDown)} controls.";
+    internal const string DomainUpDownAccessibleObjectDiagId = "WINFORMS0001";
```
Tanya-Solyanik, line 16: `Diag` -> `Diagnostics` or `Warning`

vladimir-krestov: I like `Diagnostic`

src/System.Windows.Forms/src/misc/Obsoletions.cs
vladimir-krestov: Actually, I don't know where Igor took it. But it should be WFDEV###, I guess.

docs/list-of-diagnostics.md
vladimir-krestov: @Tanya-Solyanik, please see 4th commit, it affects `Message.cs`. Do you like it?

docs/list-of-diagnostics.md
```diff
@@ -0,0 +1,41 @@
...
+* Add the obsoletion to the table below, claiming the next diagnostic id
+    * Ensure the description is meaningful within the context of this table, and without requiring the context of the calling code
+* Add new constants to `src\libraries\Common\src\System\Obsoletions.cs`, following the existing conventions
+    * A `...Message` const using the same description added to the table below
+    * A `...DiagId` const for the `WFDEV###` id
+* Annotate `src` files by referring to the constants defined from `Obsoletions.cs`
+    * Specify the `UrlFormat = Obsoletions.SharedUrlFormat`
+    * Example: `[Obsolete(Obsoletions.CodeAccessSecurityMessage, DiagnosticId = Obsoletions.CodeAccessSecurityDiagId, UrlFormat = Obsoletions.SharedUrlFormat)]`
+    * If the `Obsoletions` type is not available in the project, link it into the project
+        * `<Compile Include="$(CommonPath)System\Obsoletions.cs" Link="Common\System\Obsoletions.cs" />`
```
Tanya-Solyanik, line 20: This is not the right folder, this file is in `misc`

src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.UpDownButtons.cs
vladimir-krestov: Will consider it in scope of #7362

docs/list-of-diagnostics.md
Tanya-Solyanik: Obsoletions should really be moved to a folder that could be linked into all projects

Tanya-Solyanik: For example - C:\WinForms\src\Common\src

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Message.cs
```diff
@@ -62,7 +62,7 @@ public IntPtr WParam
 #if DEBUG
         [Obsolete(
             $"Casting to/from IntPtr is unsafe, use {nameof(LParamInternal)}.",
-            DiagnosticId = "WFDEV001")]
+            DiagnosticId = "WFDEV002")]
```
Tanya-Solyanik, line 65: @JeremyKuhne  - I think you had used the same warning ID intentionally, no?

src/System.Windows.Forms/src/PublicAPI.Unshipped.txt
```diff
@@ -82,4 +83,5 @@ virtual System.Windows.Forms.Control.DataContext.set -> void
 ~virtual System.Windows.Forms.ToolStripItem.OnCommandCanExecuteChanged(System.EventArgs e) -> void
 ~virtual System.Windows.Forms.ToolStripItem.OnCommandChanged(System.EventArgs e) -> void
 ~virtual System.Windows.Forms.ToolStripItem.OnCommandParameterChanged(System.EventArgs e) -> void
-~virtual System.Windows.Forms.ToolStripItem.OnRequestCommandExecute(System.EventArgs e) -> void
\ No newline at end of file
+~virtual System.Windows.Forms.ToolStripItem.OnRequestCommandExecute(System.EventArgs e) -> void
```
Tanya-Solyanik, line 86: @vladimir-krestov - seems like your change collided with Klaus' PR, please re-generate this file

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Message.cs
JeremyKuhne: Yes, there is no need for separate IDs as these are the same root cause.

vladimir-krestov: Which message should I use for the docs?

Tanya-Solyanik: I would say - all of them each on a new line. Ultimately this does not matter.

src/System.Windows.Forms/src/misc/Obsoletions.cs
Tanya-Solyanik: He copied it from the runtime :) But we can modify it to suite our needs

docs/list-of-diagnostics.md
Tanya-Solyanik: I would prefer to have Obsoletions.cs file in a common folder, to be shared among all projects when needed. I expect this to be a quick change. Anyway it might be prudent to send privates to Olina for a quick re-test, so I think you have time to move that file.

docs/list-of-diagnostics.md
vladimir-krestov: If to move it to the common folder, projects will throw build error, because, for example, S.W.F.Primitives project can't link to `System.Windows.Forms.DomainUpDown.DomainUpDownAccessibleObject`. Primitives project doesn't have references to System.Windows.Forms namespace. So, I will move the file, but propose to resolve this issue later.

Tanya-Solyanik (approved):
Looks good, thanks!

RussKie:
For the context, which I failed to provide yesterday to @vladimir-krestov (it was late on my end) - I copied `Obsoletion.cs` and the doc from dotnet/runtime:
* https://github.com/dotnet/runtime/blob/main/src/libraries/Common/src/System/Obsoletions.cs
* https://github.com/dotnet/runtime/blob/main/docs/project/list-of-diagnostics.md

docs/list-of-diagnostics.md
```diff
@@ -0,0 +1,41 @@
...
+    * Ensure the link's group owner matches the group owner of `winforms-warnings/WFDEV001`
+    * Connect with `@igveliko` or `@gewarren` with any questions
+
+### Obsoletion Diagnostics (`WFDEV001` - `WFDEV999`)
+
+| Diagnostic ID     | Description |
+| :---------------- | :---------- |
+|  __`WFDEV001`__ | Casting to/from IntPtr is unsafe, use `WParamInternal`. |
+|  __`WFDEV001`__ | Casting to/from IntPtr is unsafe, use `LParamInternal`. |
+|  __`WFDEV001`__ | Casting to/from IntPtr is unsafe, use `ResultInternal`. |
```
RussKie, line 38: These are all dupes, we should keep only one

vladimir-krestov: Tanya proposed this way: https://github.com/dotnet/winforms/pull/7330#discussion_r940671358
I agree, because they have different messages. Do you think that I should keep only one of them?

src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.cs
```diff
@@ -454,6 +457,19 @@ internal override Rectangle ApplyBoundsConstraints(int suggestedX, int suggested
             return base.ApplyBoundsConstraints(suggestedX, suggestedY, proposedWidth, PreferredHeight);
         }
 
+        internal override void ReleaseUiaProvider(IntPtr handle)
+        {
+            // UpDownEdit as TextBox is a control, that should disconnect its accessible object itself,
+            // but if it supports Uia providers. If no, force disconnecting for UpDownEdit accessible object
+            // as a part of UIA tree of Domain/NumericUpDown controls.
+            if (!_upDownEdit.SupportsUiaProviders)
```
vladimir-krestov, line 465: FYI: `UpDownEdit` inherits from `TextBox` that doesn't support UIA. So `ReleaseUiaProvider` is not call for `UpDownEdit` in `Control.WmDestroy` (line 12326) because it has `SupportsUiaProviders` check. So we need this check to release its objects from memory, otherwise it doesn't work. When we implement UIA supporting for `TextBox`, this check will save us from double disconnection.

vladimir-krestov:
CTI approved the fix ✅

docs/list-of-diagnostics.md
Tanya-Solyanik: Why not document all messages?

Tanya-Solyanik:
I reformatted Obsolete attributes as multilines.

Tanya-Solyanik (approved):
Thank you!

docs/list-of-diagnostics.md
RussKie: Ah, I can now see the types are different `WParamInternal`, `LParamInternal`, `ResultInternal`.... My eyes must have been tired to notice this.
I think thse can be combined, but I don't have a strong opinion.

RussKie:
@vladimir-krestov please raise a breaking docs issue in https://github.com/dotnet/docs/issues/new/choose
You can see examples of how we've done this [here](https://github.com/dotnet/docs/issues?q=label%3Adotnet-winforms%2Ftech+is%3Aclosed+label%3Abreaking-change).

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.cs
```diff
@@ -215,9 +215,7 @@ public bool Sorted
         ///  should not call base.CreateAccessibilityObject.
         /// </summary>
         protected override AccessibleObject CreateAccessibilityInstance()
-        {
-            return new DomainUpDownAccessibleObject(this);
-        }
+            => new DomainUpDownAccessibleObject(this);
```
RussKie, line 218: @vladimir-krestov @Tanya-Solyanik could you please help me understand this? `DomainUpDownAccessibleObject` is now marked as obsolete, shouldn't this be removed completely, so that we would invoke `UpDownBase.CreateAccessibilityInstance()`?

RussKie:
@vladimir-krestov @Tanya-Solyanik I would also apprecite it if you could help me filling the blanks in https://github.com/dotnet/docs/issues/30640. Thank you.

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.cs
vladimir-krestov: I would like to remove it, it looks incorrect when we still use obsolete class, it's a bad example. So, created #7584
Tanya noted about it (https://github.com/dotnet/winforms/pull/7330#pullrequestreview-1064624409), I agree with it.
@JeremyKuhne, could you please give your opinion?

Tanya-Solyanik: Let me sync with Jeremy. Maybe I misinterpreted his comments, I thought we should keep that returning this class from the `create` method.

Tanya-Solyanik: We'll be conservative inNET7 and will remove the create methos in NET8 if compat rules allow us to remove an override

Tanya-Solyanik:
> if you could help me filling the blanks in https://github.com/dotnet/docs/issues/30640.

done

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.cs
RussKie: My recollection of the outcomes of the obsoletion discussion was that we don't remove the obsolete code - i.e., `DomainUpDownAccessibleObject` definition remains in the codebase. 
When we obsolete a type we need to provide an alternative, and I'm struggling to understand what the alternative is. That is, a developer will now get `WFDEV002`, and they want to update the code. IIUIC, the correct behaviour now is to use the accessibility object provided by `UpDownBase`, but a developer using an instance of `DomainUpDown` won't be able to get it because that types overrides the base with the obsolete member.

Perhaps I'm completely missing it, but right now it feels we regressed the devex...

RussKie: The compat rules are [here](https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/breaking-change-rules.md). Adding and removing overrides is completely [acceptable](https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/breaking-change-rules.md#members), and we've been leaning on this rules quite actively for all the accessibility work.

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainUpDownAccessibleObject.cs
```diff
@@ -2,85 +2,34 @@
...
-            private readonly UpDownBase _owningDomainUpDown;
+            private readonly UpDownBaseAccessibleObject _upDownBaseAccessibleObject;
 
-            /// <summary>
-            /// </summary>
             public DomainUpDownAccessibleObject(DomainUpDown owner) : base(owner)
             {
-                _owningDomainUpDown = owner;
+                _upDownBaseAccessibleObject = new(owner);
             }
```
RussKie, line 24: Still trying to figure out....
`DomainUpDownAccessibleObject` now proxies everything to `UpDownBaseAccessibleObject` to facilitate the correct behaviour. So what's the point of obsoleting `DomainUpDownAccessibleObject`?

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.cs
Tanya-Solyanik: >My recollection of the outcomes of the obsoletion discussion was that we don't remove the obsolete code - i.e., DomainUpDownAccessibleObject definition remains in the codebase.

This is exactly what I'm suggesting. Remove the create method, not the type in NET8.
DomainUpDownAccessible object does not have any public methods, only `public override`s, so it is completely replaceable by casting instance created by CreateAccessibilityInstance method to ControlAccessibleObject.

src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainUpDownAccessibleObject.cs
vladimir-krestov: Please see the following thread: https://github.com/dotnet/winforms/pull/7330#discussion_r936067830

PR #4751, opened 2021-03-31
Removed CA1838 warning suppressions on p/invokes for StringBuilders.

AraHaan:
Fixes #3983

## Proposed changes

- Fixes rule violations on CA1838 on the p/invokes.

## Customer Impact

- Minimal, the changed code should be to the p/invokes only.

## Regression? 

- Yes / No
No

## Risk

- Minimal

## Test methodology 

TODO: Test.
- 
- 
- 
 

## Test environment(s) 

- 

```
.NET SDK (reflecting any global.json):
 Version:   6.0.100-preview.2.21155.3
 Commit:    1a9103db2d

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  11.3
 OS Platform: Darwin
 RID:         osx.11.0-x64
 Base Path:   /usr/local/share/dotnet/sdk/6.0.100-preview.2.21155.3/

Host (useful for support):
  Version: 6.0.0-preview.2.21154.6
  Commit:  3eaf1f316b

.NET SDKs installed:
  6.0.100-preview.2.21155.3 [/usr/local/share/dotnet/sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 6.0.0-preview.2.21154.6 [/usr/local/share/dotnet/shared/Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 6.0.0-preview.2.21154.6 [/usr/local/share/dotnet/shared/Microsoft.NETCore.App]

To install additional .NET runtimes or SDKs:
  https://aka.ms/dotnet-download
```
(used my browser on my mac this time to make changes)

Note: the code to the StringBuilders was attempted to try to be a minimal change, however it may require changing all the callsites too and simply not having StringBuilder parameters for the methods that wrap the actual p/invokes, this is due to the fact that StringBuilder was used for the buffers instead of string.

Also VB still needs done on it however.

This is opened as a draft because I think it may require a bit more work and testing on a windows system.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/4751)

PR #7392, opened 2022-07-11
Start using CsWin32

fernandanavamoya:

Converted these threads:
-DuplicateHandle
-GetCurrentProcess
-GetCurrentThread
-GetCurrentThreadId

Fixed the VB files to allow the new Windows.Win32 namespace

Fixed the information messages in Application.ThreadContext

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7392)


fernandanavamoya:
cc: @AArnott

KlausLoeffelmann (approved):
Looks good to me!

src/System.Windows.Forms.Primitives/src/NativeMethods.json
```diff
@@ -0,0 +1,5 @@
+{
+  "$schema": "https://aka.ms/CsWin32.schema.json",
+  "allowMarshaling": false
```
Tanya-Solyanik, line 3: Is this key name spelt correctly? And I'm curious what it is doing?

src/System.Windows.Forms.Primitives/src/System.Windows.Forms.Primitives.csproj
```diff
@@ -23,6 +23,10 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Win32.SystemEvents" Version="$(MicrosoftWin32SystemEventsPackageVersion)" />
+    <PackageReference Include="Microsoft.Windows.CsWin32" Version="0.2.1-beta">
```
RussKie, line 26: We don't specify versions explicitly in proj files, the versions need to be specified in eng/versions.props (writing from memory on a phone)

RussKie:
Could you please elaborate why CsWin32 was chosen? Thank you

src/System.Windows.Forms.Primitives/src/NativeMethods.json
AArnott: Yes, it is. It's description is in [the schema file](https://raw.githubusercontent.com/microsoft/CsWin32/main/src/Microsoft.Windows.CsWin32/settings.schema.json).

> Emit COM interfaces instead of structs, and allow generation of non-blittable structs for the sake of an easier to use API.

Its default value is `true`.

JeremyKuhne:
> Could you please elaborate why CsWin32 was chosen? Thank you

@RussKie this is a prototype effort to examine the viability of using CsWin32. It isn't "chosen" yet. As a general goal, we want to be using the [Win32 metadata](https://github.com/microsoft/win32metadata) for our interop. CsWin32 is the only usable existing projection. We're working in conjunction with the .NET interop team and @AArnott to identify issues and opportunities based on this exploration.

elachlan:
Awesome. Is there an associated issue to keep track of progress and associated issues? Happy to help where I can.

I have made a NativeMethods.txt list of all User32 PInvokes. With a list that long, I think the NativeMethods.txt needs to be split up. 

I have based my effort off of the example given by @AArnott in https://github.com/dotnet/winforms/pull/4751.

RussKie:
Thank you @elachlan for your interest. IIUIC this is an internal experiement run by @fernandanavamoya, and we don't have a definite commitment to this conversion just yet. That said, if you already have made some progress in this area as well and are willing to share it or wish to come along for the ride, we won't say no. With your help we may be able to take the experiement further and/or decide sooner whether the codebase will benefit from this conversion. Though a compulsary disclaimer that this work may get abandoned, if something doesn't pan out or proves to be too difficult.

I'll let @fernandanavamoya and @JeremyKuhne to chime in, but you can open a PR targeting feature/win32 branch we can take it from there.

AArnott:
> I have made a NativeMethods.txt list of all User32 PInvokes.

@elachlan Yes, I imagine that would be a long list. If you really wanted that, you could just do `User32.*` though. Why list them all out? That said, why generate all methods, even if you don't use them? That'll just slow down compilation time and increase the size of the resulting assembly.

elachlan:
@AArnott yes, I should have said "all User32 PInvokes used in winforms".

PR #7362, opened 2022-06-28
Refactoring of `UiaDisconnectProvider` usage implementation

vladimir-krestov:
Fixes #7357

## Proposed changes
- Keep argument and result checks inside a static separate method. It allows to simplify using of `UiaDisconnectProvider`

## Customer Impact

- No

## Regression? 

- No

## Risk

- Minimal

## Test methodology 

- CTI
- Manual testing

## Accessibility testing  
- Using Narrator/Inspect/AI and WinDbg

## Test environment(s) 

- .NET 7.0 Preview 6
- Windows 11

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7362)

PR #7395, opened 2022-07-12
Fixing of ToolStrip elements accessible objects

### LINKED ISSUE #7375: ToolStrip class and it's accessible object are leaking in memory
### .NET version
.NET Version: 7.0

### Issue description
ToolStrip and it's accessible object are left in memory after the ToolStrip control instance was inspected with accessibility tool (like Accessibility Insights or Inspect) and the form with ToolStrip is closed, garbage is collected, but accessibility tool is still opened.

ToolStrip.AccessibilityObject
ToolStripButton.ToolStripButtonAccessibleObject
ToolStripMenuItem.ToolStripMenuItemAccessibleObject
ToolStripSplitButton.ToolStripSplitButtonButton

image

### Expected behavior:

### Steps to reproduce

Create a parent form with two buttons: one will be opening a child form with the ToolStrip control and the other will collect the garbage.
Press the second button to open the child form with ToolStrip.
Start the Inspect tool.
Select the ToolStrip components using the Inspect tool.
Close the child form.
DON'T close the Inspect tool.
Look for ToolStrip accessible objects left in memory (using WinDbg, or dotMemory, or another suitable tool).

dkazennov:
Fixes #7375

## Proposed changes

- Changed ToolStrip class
- Changed ToolStripItem and ToolStripDropDownItem classes

## Customer Impact

- Less memory leaks of ToolStrip

## Regression? 

- No

## Risk

- Minimal

### Before

image

### After

image

## Test methodology 

- CTI
- Manually (via WinDbg)

## Accessibility testing  

- Using AI, Inspect, Narrator

 

## Test environment(s) 

- .NET 7.0 Preview 5
- Windows 10

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7395)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
@@ -4256,6 +4256,35 @@ namespace System.Windows.Forms
             return null;
         }
 
+        internal override void ReleaseUiaProvider(IntPtr handle)
+        {
+            base.ReleaseUiaProvider(handle);
+            if (OsVersion.IsWindows8OrGreater)
+            {
+                ReleaseToolStripItemsProviders(Items);
+            }
+        }
+
+        /// <summary>
+        ///  Call UiaDisconnectProvider for all ToolsStripItem Accessible objects.
+        ///  This method is invoked from ReleaseUiaProvider method.
+        /// </summary>
+        /// <param name="items">contains ToolStrip or ToolStripDropDown items to disconnect</param>
+        internal virtual void ReleaseToolStripItemsProviders(ToolStripItemCollection items)
+        {
+            ToolStripItem[] itemsArray = items.Cast<ToolStripItem>().ToArray();
+            foreach (ToolStripItem toolStripItem in itemsArray)
+            {
+                if (toolStripItem is ToolStripDropDownItem dropDownItem && dropDownItem.DropDownItems.Count > 0)
+                {
+                    ToolStripItemCollection dropDownMenuItems = dropDownItem.DropDownItems;
+                    ReleaseToolStripItemsProviders(dropDownMenuItems);
+                }
+
+                toolStripItem.ReleaseUiaProvider();
+            }
+        }
+
         private void RestoreFocusInternal(bool wasInMenuMode)
         {
             // This is called from the RestoreFocusFilter.  If the state of MenuMode has changed
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
@@ -4258,11 +4258,17 @@ namespace System.Windows.Forms
 
         internal override void ReleaseUiaProvider(IntPtr handle)
         {
-            base.ReleaseUiaProvider(handle);
+            if (!IsAccessibilityObjectCreated)
+            {
+                return;
+            }
+
             if (OsVersion.IsWindows8OrGreater)
             {
                 ReleaseToolStripItemsProviders(Items);
             }
+
+            base.ReleaseUiaProvider(handle);
         }
 
         /// <summary>
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripDropDownMenu.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripDropDownMenu.cs
@@ -598,6 +598,10 @@ namespace System.Windows.Forms
             }
         }
 
+        internal override void ReleaseToolStripItemsProviders(ToolStripItemCollection items)
+        {
+        }
+
         internal override void ResetScaling(int newDpi)
         {
             base.ResetScaling(newDpi);
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
@@ -3196,7 +3196,7 @@ namespace System.Windows.Forms
         internal void RaiseQueryContinueDragEvent(object key, QueryContinueDragEventArgs e)
             => ((QueryContinueDragEventHandler)Events[key])?.Invoke(this, e);
 
-        internal virtual void ReleaseUiaProvider()
+        internal void ReleaseUiaProvider()
         {
             if (TryGetAccessibilityObject(out AccessibleObject accessibleObject))
             {
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
@@ -3196,6 +3196,22 @@ namespace System.Windows.Forms
         internal void RaiseQueryContinueDragEvent(object key, QueryContinueDragEventArgs e)
             => ((QueryContinueDragEventHandler)Events[key])?.Invoke(this, e);
 
+        internal virtual void ReleaseUiaProvider()
+        {
+            if (TryGetAccessibilityObject(out AccessibleObject accessibleObject))
+            {
+                HRESULT result = UiaCore.UiaDisconnectProvider(accessibleObject);
+                Debug.Assert(result == HRESULT.S_OK);
+                Properties.SetObject(s_accessibilityProperty, null);
+            }
+
+            bool TryGetAccessibilityObject(out AccessibleObject accessibleObject)
+            {
+                accessibleObject = Properties.GetObject(s_accessibilityProperty) as AccessibleObject;
+                return accessibleObject is not null;
+            }
+        }
+
         private void ResetToolTipText() => _toolTipText = null;
 
         // This will only be called in PerMonitorV2 scenarios.
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripItemTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripItemTests.cs
@@ -7063,6 +7063,31 @@ namespace System.Windows.Forms.Tests
             Assert.Equal(new Padding(1, 2, 3, 4), item.Margin);
         }
 
+        [WinFormsFact]
+        public void ToolStripItem_Releases_UiaProvider()
+        {
+            using ToolStripWithDisconnectCount toolStrip = new();
+
+            using ToolStripDropDownItemWithDisconnectCount toolStripDropDownItem1 = new();
+            toolStrip.Items.Add(toolStripDropDownItem1);
+
+            using ToolStripDropDownItemWithDisconnectCount toolStripDropDownItem2 = new();
+            toolStripDropDownItem1.DropDownItems.Add(toolStripDropDownItem2);
+
+            _ = toolStrip.AccessibilityObject;
+            Assert.True(toolStrip.IsAccessibilityObjectCreated);
+
+            toolStrip.ReleaseUiaProvider(toolStrip.Handle);
+
+            Assert.True(toolStrip.IsAccessibilityObjectDisconnected);
+            Assert.Equal(1, toolStrip.Disconnects);
+            Assert.True(toolStripDropDownItem1.IsAccessibilityObjectDisconnected);
+            Assert.Equal(1, toolStripDropDownItem1.Disconnects);
+            Assert.True(toolStripDropDownItem2.IsAccessibilityObjectDisconnected);
+            Assert.Equal(1, toolStripDropDownItem2.Disconnects);
+            Assert.True(toolStrip.IsHandleCreated);
+        }
+
         [WinFormsTheory]
         [InlineData(RightToLeft.Inherit, RightToLeft.No)]
         [InlineData(RightToLeft.Yes, RightToLeft.Yes)]
@@ -15545,6 +15570,35 @@ namespace System.Windows.Forms.Tests
             public new void SetBounds(Rectangle bounds) => base.SetBounds(bounds);
 
             public new void SetVisibleCore(bool visible) => base.SetVisibleCore(visible);
+        }
+
+        private class  ToolStripWithDisconnectCount : ToolStrip
+        {
+            public ToolStripWithDisconnectCount() : base() { }
+
+            public int Disconnects { get; private set; }
+            public bool IsAccessibilityObjectDisconnected { get; private set; }
+
+            internal override void ReleaseUiaProvider(IntPtr handle)
+            {
+                base.ReleaseUiaProvider(handle);
+                Disconnects++;
+                IsAccessibilityObjectDisconnected = true;
+            }
+        }
+
+        private class ToolStripDropDownItemWithDisconnectCount : ToolStripDropDownItem
+        {
+            public ToolStripDropDownItemWithDisconnectCount() : base() { }
+            public int Disconnects { get; private set; }
+            public bool IsAccessibilityObjectDisconnected { get; private set; }
+
+            internal override void ReleaseUiaProvider()
+            {
+                base.ReleaseUiaProvider();
+                Disconnects++;
+                IsAccessibilityObjectDisconnected = true;
+            }
         }
     }
 }
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripItemTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripItemTests.cs
@@ -7068,10 +7068,10 @@ namespace System.Windows.Forms.Tests
         {
             using ToolStripWithDisconnectCount toolStrip = new();
 
-            using ToolStripDropDownItemWithDisconnectCount toolStripDropDownItem1 = new();
+            using ToolStripDropDownItemWithAccessibleObjectFieldAccessor toolStripDropDownItem1 = new();
             toolStrip.Items.Add(toolStripDropDownItem1);
 
-            using ToolStripDropDownItemWithDisconnectCount toolStripDropDownItem2 = new();
+            using ToolStripDropDownItemWithAccessibleObjectFieldAccessor toolStripDropDownItem2 = new();
             toolStripDropDownItem1.DropDownItems.Add(toolStripDropDownItem2);
 
             _ = toolStrip.AccessibilityObject;
@@ -7079,12 +7079,9 @@ namespace System.Windows.Forms.Tests
 
             toolStrip.ReleaseUiaProvider(toolStrip.Handle);
 
-            Assert.True(toolStrip.IsAccessibilityObjectDisconnected);
             Assert.Equal(1, toolStrip.Disconnects);
-            Assert.True(toolStripDropDownItem1.IsAccessibilityObjectDisconnected);
-            Assert.Equal(1, toolStripDropDownItem1.Disconnects);
-            Assert.True(toolStripDropDownItem2.IsAccessibilityObjectDisconnected);
-            Assert.Equal(1, toolStripDropDownItem2.Disconnects);
+            Assert.True(toolStripDropDownItem1.IsAccessibleObjectCleared());
+            Assert.True(toolStripDropDownItem2.IsAccessibleObjectCleared());
             Assert.True(toolStrip.IsHandleCreated);
         }
 
@@ -15577,27 +15574,25 @@ namespace System.Windows.Forms.Tests
             public ToolStripWithDisconnectCount() : base() { }
 
             public int Disconnects { get; private set; }
-            public bool IsAccessibilityObjectDisconnected { get; private set; }
 
-            internal override void ReleaseUiaProvider(IntPtr handle)
+            internal new void ReleaseUiaProvider(IntPtr handle)
             {
                 base.ReleaseUiaProvider(handle);
+
                 Disconnects++;
-                IsAccessibilityObjectDisconnected = true;
             }
         }
 
-        private class ToolStripDropDownItemWithDisconnectCount : ToolStripDropDownItem
+        private class ToolStripDropDownItemWithAccessibleObjectFieldAccessor : ToolStripDropDownItem
         {
-            public ToolStripDropDownItemWithDisconnectCount() : base() { }
-            public int Disconnects { get; private set; }
-            public bool IsAccessibilityObjectDisconnected { get; private set; }
+            public ToolStripDropDownItemWithAccessibleObjectFieldAccessor() : base() { }
 
-            internal override void ReleaseUiaProvider()
+            public bool IsAccessibleObjectCleared()
             {
-                base.ReleaseUiaProvider();
-                Disconnects++;
-                IsAccessibilityObjectDisconnected = true;
+                var key = this.TestAccessor().Dynamic.s_accessibilityProperty;
+                var accessibleObject = Properties.GetObject(key) as AccessibleObject;
+
+                return accessibleObject is null;
             }
         }
     }


dkazennov:
On this stage the fix works for all controls except ToolStripDropDownButton and ToolStripSplitButton.
This controls contain child controls. For now UiaCore.UiaDisconnectProvider(childControl.AccessibilityObject) returns false;

dkazennov:
Created CTI test request.

dkazennov:
CTI showed memory leaks. Reproduces in WinformsControlsTest with the form taken from the CTI testing application.

image

dkazennov:
CTI results seems to be connected with the ComboBox+ComboBoxChildEditUiaProvider:

image
image
image

dkazennov:
I have implemented the fix from https://github.com/dotnet/winforms/pull/7319 for ComboBox class.
There is no ComboBox in root now, but still there are leaks with the CTI form being included into WinformsControlsTest.

image

Checking for root objects in memory for some of AO leaked objects:
image

Further investigation needed.

dkazennov:
Last test result: no AccessibilityObject memory leaks.
image

dkazennov:
ToolStrip Control class has ToolStripItemCollection Items collection.
ToolStripItem class is a Component, but has AccessibleObject.
This objects don't disconnect after ToolStrip.AccessibleObject was being disconnected.

Disconnecting ToolStripItem.AccessibleObject for all components of the ToolStrip returns 
- HRESULT.S_OK for ToolStripItem objects that have OnDropDown == false property
- HRESULT.E_FAIL for ToolStripItem objects that have  OnDropDown == true property

e.g.
Case 1.
- Object toolStripButton1 type of ToolStripButton has AccessibilityObject type of ToolStripButtonAccessibleObject
- Object toolStripButton1 (is OnDropDown? False) UiaDisconnectProvider result is S_OK

Case 2.
- Object toolStripMenuItem1 type of ToolStripMenuItem has AccessibilityObject type of ToolStripMenuItemAccessibleObject
- Object toolStripMenuItem1 (is OnDropDown? True) UiaDisconnectProvider result is E_FAIL

dkazennov:
It turned out that when ToolStrip control calls for ReleaseUiaProvider after the form is closed there is also ReleaseUiaProvider call for ToolStripDropDownMenu control (if there is a dropDown component inside the main toolStrip). ToolStripDropDownMenu is a child class for ToolStrip. It is owned by one of ToolStrip.Items type of ToolStripItem and in turn owns its own Items type of ToolStripItem.
Call for disconnection of Items in ToolStrip.ReleaseUiaProvider() repeats when ToolStripDropDownMenu is being closed. But at this moment ToolStripDropDownMenu.Items are already disconnected. Therefore duplication of disconnect calls with FALSE results.

Now all ToolStripItem components are being disconnected in ToolStrip.ReleaseUiaProvider().
ToolStripDropDownMenu.ReleaseUiaProvider() do nothing about AccessibleObjects.

Final result: no leaks.
image

dkazennov:
Rebased on most recent main branch.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
```diff
@@ -4256,6 +4256,29 @@ public ToolStripItem GetItemAt(Point point)
...
+            ReleaseToolStripItemsProviders(Items);
+        }
+
+        internal virtual void ReleaseToolStripItemsProviders(ToolStripItemCollection items)
+        {
+            if (OsVersion.IsWindows8OrGreater)
+            {
+                foreach (ToolStripItem toolStripItem in items)
+                {
+                    if (toolStripItem is ToolStripDropDownItem && ((ToolStripDropDownItem)toolStripItem).DropDownItems.Count > 0)
```
Tanya-Solyanik, line 4271: You can use pattern matching here
```
if (toolStripItem is ToolStripDropDownItem dropDownItem && dropDownItem.DropDownItems.Count > 0)
```

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripDropDownMenu.cs
```diff
@@ -598,6 +598,11 @@ protected override void OnPaintBackground(PaintEventArgs e)
             }
         }
 
+        // All ToolStripItem.AccessibleObjects should be disconnected in ToolStrip.ReleaseUiaProvider()
```
Tanya-Solyanik, line 601: Could you please rephrase this comment using `<summary>`  - [xml docs comments](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/)? Along the lines of ` Call UiaDisconnectProvider for all ToolsStripItem Accessible objects. This method is invoked from ..`

But put that comment on the base class' method definition.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
```diff
@@ -1149,6 +1149,8 @@ public ToolStripItemImageScaling ImageScaling
         internal ToolStripItemInternalLayout InternalLayout
             => _toolStripItemInternalLayout ??= CreateInternalLayout();
 
+        internal bool IsAccessibilityObjectCreated => Properties.GetObject(s_accessibilityProperty) is AccessibleObject;
```
Tanya-Solyanik, line 1152: Looks like this method is used only in this file, please make it private if this is the case.

Tanya-Solyanik: Though for this file a TryGet pattern might be more appropriate
```
        private bool TryGetAccessibilityObject(out AccessibleObject accessibleObject)
        {
            accessibleObject = Properties.GetObject(s_accessibilityProperty) as AccessibleObject;
            return accessibleObject is not null;
        }
```

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
```diff
@@ -3054,6 +3056,16 @@ internal void RaisePaintEvent(object key, PaintEventArgs e)
         internal void RaiseQueryContinueDragEvent(object key, QueryContinueDragEventArgs e)
             => ((QueryContinueDragEventHandler)Events[key])?.Invoke(this, e);
 
+        internal virtual void ReleaseUiaProvider()
+        {
+            if (IsAccessibilityObjectCreated && OsVersion.IsWindows8OrGreater)
```
Tanya-Solyanik, line 3061: I think it will be sufficient to test OS version once from the root call site - in ToolStrip class

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripDropDownMenu.cs
```diff
@@ -598,6 +598,11 @@ protected override void OnPaintBackground(PaintEventArgs e)
...
+        internal override void ReleaseToolStripItemsProviders(ToolStripItemCollection items)
+        {
```
Tanya-Solyanik, line 603: It would also be helpful to add a comment that explains when items placed on the drop down are disconnected since we are not disconnecting them here.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
```diff
@@ -3054,6 +3056,16 @@ internal void RaisePaintEvent(object key, PaintEventArgs e)
...
+            {
+                AccessibleObject accessibleObject = (AccessibleObject)Properties.GetObject(s_accessibilityProperty);
+                HRESULT result = UiaCore.UiaDisconnectProvider(accessibleObject);
+                Debug.Assert(result == HRESULT.S_OK);
```
Tanya-Solyanik, line 3065: Please set AccessibleObject in the properties collection to null after disconnecting it to avoid double-disconnect.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
```diff
@@ -4256,6 +4256,29 @@ public ToolStripItem GetItemAt(Point point)
             return null;
         }
 
+        internal override void ReleaseUiaProvider(IntPtr handle)
+        {
+            base.ReleaseUiaProvider(handle);
+            ReleaseToolStripItemsProviders(Items);
```
Tanya-Solyanik, line 4262: Consider moving OS version test to this call site.

Tanya-Solyanik (commented):
Great job getting rid of the leaks, please see my comments and add a unit test Like @dmitrii-drobotov added in his ListBox PR.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
dkazennov: Thanks! Should use this syntax more often.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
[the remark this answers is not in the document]
dkazennov: Let's implement your advice since it sounds more safe.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
dkazennov: > Looks like this method is used only in this file, please make it private if this is the case.

I thought it is better to have methods like IsAccessibilityObjectCreated internal or even public since we may need those anytime? Same methods in Control classes are usually internal as well.

dkazennov:
To test this Issue with unit test we need to use at least three different controls. While unit tests successfully cover memory leaks in one control, with multiple controls in one test objects often stay connected even after ReleaseUiaProvider call.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
Tanya-Solyanik: We  will change visibility of this method if needed

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
```diff
@@ -4256,6 +4256,34 @@ public ToolStripItem GetItemAt(Point point)
...
+        {
+            base.ReleaseUiaProvider(handle);
+            if (OsVersion.IsWindows8OrGreater)
+            {
+                ReleaseToolStripItemsProviders(Items);
+            }
+        }
+
+        /// <summary>
+        /// Call UiaDisconnectProvider for all ToolsStripItem Accessible objects.
```
Tanya-Solyanik, line 4269: nit: we put 2 whitespaces after `///` in the intellisense comments

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
```diff
@@ -4256,6 +4256,34 @@ public ToolStripItem GetItemAt(Point point)
...
+        /// This method is invoked from ReleaseUiaProvider method.
+        /// </summary>
```
Tanya-Solyanik, line 4271: Please add parameter description to the xml-doc coment

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripDropDownMenu.cs
```diff
@@ -598,6 +598,14 @@ protected override void OnPaintBackground(PaintEventArgs e)
             }
         }
 
+        /// <summary>
+        /// UiaDisconnectProvider for all ToolsStripItem Accessible objects
+        /// is already called for when ToolStrip.ReleaseToolStripItemsProviders method is invoked.
+        /// </summary>
```
Tanya-Solyanik, line 604: Please add parameter desription

Tanya-Solyanik (commented):
I would prefer `private` acessibility modifier on the `is AO created` method. Also please add a unit test

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
dkazennov: 1. Added private accessibility modifier on the is AO created method.
2. There is a problem with unit tests here unlike PR #7453. We can ensure creation of ToolStrip.AccessibilityObject but the Control.ReleaseUiaProvider call doesn't set the ToolStrip value to null.

Tanya-Solyanik: Control.UiaDisconnectProvider should set AO to null as well. However for the purpose of these tests, you can set AOs for ToolStripItems

Tanya-Solyanik (commented):
There are a couple more comments to address and the unit test to finish.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
dkazennov: > Control.UiaDisconnectProvider should set AO to null as well. However for the purpose of these tests, you can set AOs for ToolStripItems

I have found a way around this. I check if ToolStripItem.ReleaseUiaProvider was being called in a way similar to #7296

dkazennov:
I have found a way to detects the call for ToolStripItem.ReleaseUiaProvider without setting AccessibilityObjects to null manual. 
Similar to https://github.com/dotnet/winforms/pull/7296 way.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
```diff
@@ -19,9 +19,9 @@
 
 namespace System.Windows.Forms
 {
-    /// <summary>
+    ///  <summary>
```
Tanya-Solyanik, line 22: Actuali only content is indented by 2 spaces, the `summary` tags are indented by a single space.
```
/// <summary>
///  ToolStrip control.
/// </summary>
```

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
```diff
@@ -3054,6 +3056,17 @@ internal void RaisePaintEvent(object key, PaintEventArgs e)
         internal void RaiseQueryContinueDragEvent(object key, QueryContinueDragEventArgs e)
             => ((QueryContinueDragEventHandler)Events[key])?.Invoke(this, e);
 
+        internal virtual void ReleaseUiaProvider()
+        {
+            if (IsAccessibilityObjectCreated)
+            {
+                AccessibleObject accessibleObject = (AccessibleObject)Properties.GetObject(s_accessibilityProperty);
```
Tanya-Solyanik, line 3063: This way you are calling `GetObject` and casting to AccessibilityObject twice. `TryGet` pattern would have let you do it once.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripItemTests.cs
```diff
@@ -7063,6 +7063,27 @@ public void ToolStripItem_Parent_SetWithMargin_GetReturnsExpected(bool enabled,
             Assert.Equal(new Padding(1, 2, 3, 4), item.Margin);
         }
 
+        [WinFormsFact]
+        public void ToolStripItem_Releases_UiaProvider()
+        {
+            using ToolStrip toolStrip = new();
```
Tanya-Solyanik, line 7069: Please define a test class derived from the ToolStrip that would count Disconnect provider calls in the same way as your test item is counting these calls. and validate that provider was disconnected for this class.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripDropDownMenu.cs
```diff
@@ -598,6 +598,15 @@ protected override void OnPaintBackground(PaintEventArgs e)
             }
         }
 
+        ///  <summary>
+        ///  UiaDisconnectProvider for all ToolsStripItem Accessible objects
+        ///  is already called for when ToolStrip.ReleaseToolStripItemsProviders method is invoked.
+        ///  </summary>
+        ///  <param name="items">contains ToolStrip or ToolStripDropDown items to disconnect</param>
+        internal override void ReleaseToolStripItemsProviders(ToolStripItemCollection items)
```
Tanya-Solyanik, line 606: Intellisense comments are inherited from the base classes for virtual methods. Please remove this comment, as it does not add much content to the what is said in ToolStrip.ReleaseToolStripItemsProviders comment.

Tanya-Solyanik (commented):
There are a couple more comments to address and the unit test to finish.

dkazennov:
I have added a unit test. It seems to work as intended.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
```diff
@@ -1149,6 +1149,8 @@ public ToolStripItemImageScaling ImageScaling
         internal ToolStripItemInternalLayout InternalLayout
             => _toolStripItemInternalLayout ??= CreateInternalLayout();
 
+        private bool IsAccessibilityObjectCreated => Properties.GetObject(s_accessibilityProperty) is AccessibleObject;
```
Tanya-Solyanik, line 1152: I don't think this method is used any more, please remove it.

Tanya-Solyanik (commented):
Looks good, I added a couple of minor comments. Do you think it should be retested by CTIs? I

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
Tanya-Solyanik: We'll track setting Ao to null in Control class separately

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripDropDownMenu.cs
Tanya-Solyanik: Why are menu items' providers not disconnected

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
```diff
@@ -3054,6 +3056,16 @@ internal void RaisePaintEvent(object key, PaintEventArgs e)
         internal void RaiseQueryContinueDragEvent(object key, QueryContinueDragEventArgs e)
             => ((QueryContinueDragEventHandler)Events[key])?.Invoke(this, e);
 
+        internal virtual void ReleaseUiaProvider()
```
Tanya-Solyanik, line 3059: There is no reason to make this method virtual because nobody overrides it

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
```diff
@@ -3590,6 +3602,12 @@ internal virtual void OnKeyboardToolTipUnhook(ToolTip toolTip)
 
         internal virtual bool IsBeingTabbedTo() => ToolStrip.AreCommonNavigationalKeysDown();
 
+        private bool TryGetAccessibilityObject(out AccessibleObject accessibleObject)
```
Tanya-Solyanik, line 3605: This file is poorly alphabetized, so there is no reason to separate two related methods. Also current guidelines are to group related methods together. You might consider making this methos local since it's used only once.

Tanya-Solyanik (commented):
Looks like this change had not been tested by the CTIs, please provide privates to the team.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripItemTests.cs
```diff
@@ -15546,5 +15568,31 @@ public new ToolStrip Parent
 
             public new void SetVisibleCore(bool visible) => base.SetVisibleCore(visible);
         }
+
+        private class  ToolStripWithDisconnectCounter : ToolStrip
```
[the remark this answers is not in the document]
dkazennov: > and have the same implementation of the ReleaseUiaProvider method

To have both classes counting disconnect events AND checking the connection status.

dkazennov:
Regarding the question why are menu items' providers not disconnected (in ToolStripDropDownMenu.ReleaseToolStripItemsProviders): we disconnect all nested menu items in ToolStrip.ReleaseToolStripItemsProviders.

1) There is the hierarchy of component classes:

4.3. ToolStripMenuItem 
4.2. ToolStripDropDownButton
4.1. ToolStripSplitButton
3. ToolStripDropDownItem
2. ToolStripItem <-- This type is being used to iterate though the collections of elements of the toolStrip
1. Component
0. MarshalByRefObject

1.2) Typical button on ToolStrip is ToolStripMenuItem.
1.3) Classes like ToolStripButton, ToolStripControlHost, ToolStripLabel and ToolStripSeparator inherit ToolStripItem too.

2) There is also the hierarchy of control classes:

1. ToolStripDropDownMenu <-- Empty override ReleaseToolStripItemsProviders () to avoid duplication of disconnect calls.
2. ToolStripDropDown
3. ToolStrip <-- Our new method ReleaseToolStripItemsProviders (called from ToolStrip.ReleaseUiaProvider) which disconnects _all_ ToolStripItems in ToolStrip.
4. ScrollableControl
5. Control

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
dkazennov: Oh, good idea about local file, thanks!

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
dkazennov: > There is no reason to make this method virtual because nobody overrides it

We need this method to be virtual because we override it in unit test.

dkazennov:
Requested CTI re-test.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
Tanya-Solyanik: I'd rather not make method virtual for the test purposes, I'll redo the test

Tanya-Solyanik (approved):
Ready to merge after testing.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
```diff
@@ -4256,6 +4256,35 @@ public ToolStripItem GetItemAt(Point point)
...
+        }
+
+        /// <summary>
+        ///  Call UiaDisconnectProvider for all ToolsStripItem Accessible objects.
+        ///  This method is invoked from ReleaseUiaProvider method.
+        /// </summary>
+        /// <param name="items">contains ToolStrip or ToolStripDropDown items to disconnect</param>
+        internal virtual void ReleaseToolStripItemsProviders(ToolStripItemCollection items)
+        {
+            ToolStripItem[] itemsArray = items.Cast<ToolStripItem>().ToArray();
```
dreddy-work, line 4275: Isn't `.ToArray()` create new copy of ToolstripItems array? Why can't we cast on each iteration with no local copy?

dreddy-work (commented):
LGTM.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
Tanya-Solyanik: This handles the case when Items collection is modified on another thread.

vladimir-krestov: Why not to skip this step? `foreach` works for `ToolStripItemCollection` well.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
```diff
@@ -4256,6 +4256,35 @@ public ToolStripItem GetItemAt(Point point)
             return null;
         }
 
+        internal override void ReleaseUiaProvider(IntPtr handle)
+        {
+            base.ReleaseUiaProvider(handle);
+            if (OsVersion.IsWindows8OrGreater)
```
vladimir-krestov, line 4262: We can add `IsAccessibleObjectCreated` check to avoid the extra call of `ReleaseToolStripItemsProviders`. 
@Tanya-Solyanik, will it be helpful or may affect moving items between ToolStrips?

vladimir-krestov (dismissed):
Looks good ✅

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
Tanya-Solyanik: it might not work if this code is called from a finalizer thread

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
Tanya-Solyanik: added in the 3rd commit

Tanya-Solyanik (approved):

dkazennov:
New CTI result shows ComboBox leaks (and only them). 
This branch was created when ComboBox fix was work-in-progress (I was using special testing branch with cherry-picked changes from the ComboBox fix WIP). 
Since the ComboBox fix https://github.com/dotnet/winforms/pull/7319 is now implemented into the main branch there should be no leaks. I can rebase this fix and call for CTI re-test.

dkazennov:
Local test result with the fix rebased on the most recent main branch with the ComboBox fix #7319 included shows no leaks:

image

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
```diff
@@ -4258,11 +4258,17 @@ public ToolStripItem GetItemAt(Point point)
...
+            {
+                return;
+            }
+
             if (OsVersion.IsWindows8OrGreater)
             {
                 ReleaseToolStripItemsProviders(Items);
             }
+
+            base.ReleaseUiaProvider(handle);
```
vladimir-krestov, line 4271: Good catch!

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
```diff
@@ -3196,6 +3196,22 @@ internal void RaisePaintEvent(object key, PaintEventArgs e)
         internal void RaiseQueryContinueDragEvent(object key, QueryContinueDragEventArgs e)
             => ((QueryContinueDragEventHandler)Events[key])?.Invoke(this, e);
 
+        internal void ReleaseUiaProvider()
+        {
+            if (TryGetAccessibilityObject(out AccessibleObject accessibleObject))
+            {
+                HRESULT result = UiaCore.UiaDisconnectProvider(accessibleObject);
```
vladimir-krestov, line 3203: Rework it if #7362 was closed first

Tanya-Solyanik:
I don't think this should be re-tested

Tanya-Solyanik (approved):

PR #7397, opened 2022-07-12
Publish APIs for .NET 7.0 preview 7

Tanya-Solyanik:
standard branding step, this was produced by end/APIConpatibility/mark-shipped.cmd

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7397)


JeremyKuhne:
There isn't anything really new here. We should keep stuff in unpublished with a tracking issue until we figure out the "right" way to deal with the `IntPtr` == `nint` change.

JeremyKuhne:
I see there are a few things that aren't `nint`. We can move those and leave the `nint` stuff unshipped.

JeremyKuhne (approved):

PR #7401, opened 2022-07-13
Fix incorrect WarningsAsErrors element

GrabYourPitchforks:
Using the below syntax performs a complete replacement of warning codes which should be treated as errors.

```xml
<WarningsAsErrors>aaaaa;bbbbb;ccccc;...</WarningsAsErrors>
```

Instead, the following syntax should be used, which inherits the base list coming from the SDK and augments the list with your desired warning codes.

```xml
<WarningsAsErrors>$(WarningsAsErrors);aaaaa;bbbbb;ccccc;...</WarningsAsErrors>
```

> Exception: If you're intentionally trying to suppress the SDK's base list, please ignore this rule.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7401)

### DIFF
--- a/src/System.Windows.Forms/tests/IntegrationTests/VisualBasicTemplate/VbWinForms.vbproj
+++ b/src/System.Windows.Forms/tests/IntegrationTests/VisualBasicTemplate/VbWinForms.vbproj
@@ -7,11 +7,8 @@
     <UseWindowsForms>true</UseWindowsForms>
   </PropertyGroup>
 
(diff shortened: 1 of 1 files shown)


RussKie (approved):

Tanya-Solyanik (approved):
Thank you!

PR #7403, opened 2022-07-13
Make the TreeView respect the DoubleBuffered property

### LINKED ISSUE #4883: TreeView flickering
* .NET Core Version: all up to latest (5.0.5).

* Have you experienced this same bug with .NET Framework?:  yes

**Problem description:**
TreeView is flickering (especially on large amount of data):
Native control has [ex. style](https://docs.microsoft.com/en-us/windows/win32/controls/tree-view-control-window-extended-styles) `TVS_EX_DOUBLEBUFFER` that fix the problem:
```cs
private class TreeViewEx : TreeView
{
    private const int TVM_SETEXTENDEDSTYLE = 0x1100 + 44;
    private const int TVS_EX_DOUBLEBUFFER = 0x0004;

    [System.Runtime.InteropServices.DllImport("user32.dll", SetLastError = true)]
    private static extern IntPtr SendMessage(IntPtr HWnd, int Msg, IntPtr Wp, IntPtr Lp);

    protected override void OnHandleCreated(EventArgs e)
    {
        SendMessage(this.Handle, TVM_SETEXTENDEDSTYLE, (IntPtr)TVS_EX_DOUBLEBUFFER, (IntPtr)TVS_EX_DOUBLEBUFFER);
        base.OnHandleCreated(e);
    }
}
```
Left default `TreeView`, Right with `TVS_EX_DOUBLEBUFFER`:
TreeView
[stackoverflow](https://stackoverflow.com/a/10364283/6075536).

**Expected behavior:**
Ability to avoid flickering without Pinvoke  and deriving own class from `TreeView`.
P.s. protected `DoubleBuffered` doesn't solve the problem.
P.p.s. May be _API request_ is more suitable here?

**Minimal repro:**
[TreeViewFlickering.zip](https://github.com/dotnet/winforms/files/6446248/TreeViewFlickering.zip)

0xC0000054:
Fixes #4883

## Proposed changes

- The TreeView will respect the DoubleBuffered property.

## Customer Impact

- Double buffering can be enabled in a derived class without needing to use P/Invoke.

## Regression? 

- No

## Risk

Probably minimal.

## Test methodology 

- Manually tested the change in a winforms app

## Test environment(s) 

- 7.0.100-preview.5.22307.18 (x64)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7403)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TreeView.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TreeView.cs
@@ -2082,22 +2082,16 @@ namespace System.Windows.Forms
                 return;
             }
 
-            TVS_EX extendedStyles = (TVS_EX)User32.SendMessageW(this, (User32.WM)TVM.GETEXTENDEDSTYLE);
+            TVS_EX extendedStyles = 0;
 
-            SetExtendedStyle(ref extendedStyles, TVS_EX.DOUBLEBUFFER, DoubleBuffered);
-
-            User32.SendMessageW(this, (User32.WM)TVM.SETEXTENDEDSTYLE, (int)extendedStyles, (int)extendedStyles);
-
-            static void SetExtendedStyle(ref TVS_EX extendedStyles, TVS_EX mask, bool value)
+            if (DoubleBuffered)
             {
-                if (value)
-                {
-                    extendedStyles |= mask;
-                }
-                else
-                {
-                    extendedStyles &= ~mask;
-                }
+                extendedStyles |= TVS_EX.DOUBLEBUFFER;
+            }
+
+            if (extendedStyles != 0)
+            {
+                User32.SendMessageW(this, (User32.WM)TVM.SETEXTENDEDSTYLE, (int)extendedStyles, (int)extendedStyles);
             }
         }
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TreeView.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TreeView.cs
@@ -438,10 +438,9 @@ namespace System.Windows.Forms
             get => base.DoubleBuffered;
             set
             {
-                base.DoubleBuffered = value;
-
-                if (IsHandleCreated)
+                if (DoubleBuffered != value)
                 {
+                    base.DoubleBuffered = value;
                     UpdateTreeViewExtendedStyles();
                 }
             }
@@ -2078,6 +2077,11 @@ namespace System.Windows.Forms
 
         private void UpdateTreeViewExtendedStyles()
         {
+            if (!IsHandleCreated)
+            {
+                return;
+            }
+
             TVS_EX extendedStyles = (TVS_EX)User32.SendMessageW(this, (User32.WM)TVM.GETEXTENDEDSTYLE);
 
             SetExtendedStyle(ref extendedStyles, TVS_EX.DOUBLEBUFFER, DoubleBuffered);
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TreeView.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TreeView.cs
@@ -2075,16 +2075,6 @@ namespace System.Windows.Forms
             SelectedNode = savedSelectedNode;
         }
 
-        private void UpdateTreeViewExtendedStyles()
-        {
-            if (!IsHandleCreated)
-            {
-                return;
-            }
-
-            User32.SendMessageW(this, (User32.WM)TVM.SETEXTENDEDSTYLE, (nint)TVS_EX.DOUBLEBUFFER, (nint)(DoubleBuffered ? TVS_EX.DOUBLEBUFFER : 0));
-        }
-
         // Replace the native control's ImageList with our current stateImageList
         // set the value of internalStateImageList to the new list
         private void UpdateNativeStateImageList()
@@ -2748,6 +2738,16 @@ namespace System.Windows.Forms
                     }
                 }
             }
+        }
+
+        private void UpdateTreeViewExtendedStyles()
+        {
+            if (!IsHandleCreated)
+            {
+                return;
+            }
+
+            User32.SendMessageW(this, (User32.WM)TVM.SETEXTENDEDSTYLE, (nint)TVS_EX.DOUBLEBUFFER, (nint)(DoubleBuffered ? TVS_EX.DOUBLEBUFFER : 0));
         }
 
         /// <remarks>
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TreeView.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TreeView.cs
@@ -436,7 +436,15 @@ namespace System.Windows.Forms
         protected override bool DoubleBuffered
         {
             get => base.DoubleBuffered;
-            set => base.DoubleBuffered = value;
+            set
+            {
+                base.DoubleBuffered = value;
+
+                if (IsHandleCreated)
+                {
+                    UpdateTreeViewExtendedStyles();
+                }
+            }
         }
 
         /// <summary>
@@ -1959,6 +1967,9 @@ namespace System.Windows.Forms
 
             base.OnHandleCreated(e);
 
+            // The TreeView extended styles are independent of the window extended styles.
+            UpdateTreeViewExtendedStyles();
+
             int version = (int)User32.SendMessageW(this, (User32.WM)CCM.GETVERSION);
             if (version < 5)
             {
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/TreeView.cs
```diff
@@ -436,7 +436,15 @@ protected override Size DefaultSize
...
         {
             get => base.DoubleBuffered;
-            set => base.DoubleBuffered = value;
+            set
+            {
+                base.DoubleBuffered = value;
+
+                if (IsHandleCreated)
+                {
+                    UpdateTreeViewExtendedStyles();
```
dreddy-work, line 445: why do we have to update every time we access this property?

0xC0000054: I fixed it to only update the styles when the property value has changed.

dreddy-work:
@0xC0000054 , can you share gif of how it looks now after the fix?

src/System.Windows.Forms/src/System/Windows/Forms/TreeView.cs
```diff
@@ -2065,6 +2075,32 @@ protected override void OnHandleCreated(EventArgs e)
             SelectedNode = savedSelectedNode;
         }
 
+        private void UpdateTreeViewExtendedStyles()
+        {
+            if (!IsHandleCreated)
```
dreddy-work, line 2080: Given this is called only on `OnHandleCreated` , i do not see a need to check for `HandleCreated` again.

src/System.Windows.Forms/src/System/Windows/Forms/TreeView.cs
```diff
@@ -2065,6 +2075,32 @@ protected override void OnHandleCreated(EventArgs e)
...
+            {
+                return;
+            }
+
+            TVS_EX extendedStyles = (TVS_EX)User32.SendMessageW(this, (User32.WM)TVM.GETEXTENDEDSTYLE);
+
+            SetExtendedStyle(ref extendedStyles, TVS_EX.DOUBLEBUFFER, DoubleBuffered);
```
dreddy-work, line 2087: What are the default extended styles for TreeView here? why do we have to negate when `DoubleBuffered` is not enabled? It would default go back to current behavior. Did you notice any change in behavior here?

src/System.Windows.Forms/src/System/Windows/Forms/TreeView.cs
0xC0000054: That check is for the DoubleBuffered property, which can be changed before the handle has been created.

src/System.Windows.Forms/src/System/Windows/Forms/TreeView.cs
0xC0000054: The default extended styles appeared to be 0 (None) when I was debugging.

But now that I think about it the code is incorrect and would not clear the style anyway, the first send message parameter is a mask.

0xC0000054:
> can you share gif of how it looks now after the fix?

TreeView

The TreeView on the left uses the code in this PR and sets the DoubleBufferered property to true.
The TreeView on the right overrides the `OnHandleCreated` method to set the `TVS_EX_DOUBLEBUFFER` style.

The test project is attached, it is based on the minimal repo posted in the linked issue.
[TreeViewFlickering.zip](https://github.com/dotnet/winforms/files/9137065/TreeViewFlickering.zip)

src/System.Windows.Forms/src/System/Windows/Forms/TreeView.cs
0xC0000054: I removed the GETEXTENDEDSTYLE call and the negation code.
Negating the extended styles would have been a regression for callers that set the values themselves.

src/System.Windows.Forms/src/System/Windows/Forms/TreeView.cs
```diff
@@ -2065,6 +2075,26 @@ protected override void OnHandleCreated(EventArgs e)
             SelectedNode = savedSelectedNode;
         }
 
+        private void UpdateTreeViewExtendedStyles()
```
RussKie, line 2078: [nit] we strive to keep the members in lexographical order, please move this new method after `UpdateStylesCore`.

RussKie (approved):
LGTM

src/System.Windows.Forms/src/System/Windows/Forms/TreeView.cs
```diff
@@ -2065,6 +2075,26 @@ protected override void OnHandleCreated(EventArgs e)
...
+            TVS_EX extendedStyles = 0;
+
+            if (DoubleBuffered)
+            {
+                extendedStyles |= TVS_EX.DOUBLEBUFFER;
+            }
+
+            if (extendedStyles != 0)
+            {
+                User32.SendMessageW(this, (User32.WM)TVM.SETEXTENDEDSTYLE, (int)extendedStyles, (int)extendedStyles);
```
kirsan31, line 2094:
```suggestion
                User32.SendMessageW(this, (User32.WM)TVM.SETEXTENDEDSTYLE, (nint)TVS_EX.DOUBLEBUFFER, (nint)extendedStyles);
```

src/System.Windows.Forms/src/System/Windows/Forms/TreeView.cs
```diff
@@ -2065,6 +2075,26 @@ protected override void OnHandleCreated(EventArgs e)
...
+            }
+
+            TVS_EX extendedStyles = 0;
+
+            if (DoubleBuffered)
+            {
+                extendedStyles |= TVS_EX.DOUBLEBUFFER;
+            }
+
+            if (extendedStyles != 0)
```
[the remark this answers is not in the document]
0xC0000054: Fixed.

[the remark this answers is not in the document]
kirsan31: I think you missed my update on this :(
For example with this realization my workaround from #4883 will stop working...

kirsan31:
Just in case, so as not to miss, we still have an unresolved issue (possibility to break some users):
https://github.com/dotnet/winforms/pull/7403#discussion_r924015209
https://github.com/dotnet/winforms/pull/7403#discussion_r924176305

kirsan31:
👍
I think it's all we can do with `DoubleBuffered` here. Now we will break ppl only in very _specific_ situations 😁

RussKie:
> Now we will break ppl only in very _specific_ situations 😁

Can you think what these situations may be?

RussKie (approved):
👍

kirsan31:
> Can you think what these situations may be?

Only if some one for any reason will call `DoubleBuffered = true;` and then `DoubleBuffered = false;` before `HandleCreated`.

RussKie:
> Only if some one for any reason will call `DoubleBuffered = true;` and then `DoubleBuffered = false;` before `HandleCreated`.

In that case, `UpdateTreeViewExtendedStyles` would bail in the first `return`, no?
However, this sequence may result in unexpected behaviour **before** handle is created:
```cs
DoubleBuffered = true;
DoubleBuffered = false;
SendMessage(this.Handle, TVM_SETEXTENDEDSTYLE, (IntPtr)TVS_EX_DOUBLEBUFFER, (IntPtr)TVS_EX_DOUBLEBUFFER);
```

kirsan31:
> In that case, `UpdateTreeViewExtendedStyles` would bail in the first `return`, no? However, this sequence may result in unexpected behaviour **before** handle is created:
> 
> ```cs
> DoubleBuffered = true;
> DoubleBuffered = false;
> SendMessage(this.Handle, TVM_SETEXTENDEDSTYLE, (IntPtr)TVS_EX_DOUBLEBUFFER, (IntPtr)TVS_EX_DOUBLEBUFFER);
> ```

Yes, of course, all this makes sense only if we called
```cs
SendMessage(this.Handle, TVM_SETEXTENDEDSTYLE, (IntPtr)TVS_EX_DOUBLEBUFFER, (IntPtr)TVS_EX_DOUBLEBUFFER);
```
before handle is created.

So, if user have this call before handle is created:
```cs
SendMessage(this.Handle, TVM_SETEXTENDEDSTYLE, (IntPtr)TVS_EX_DOUBLEBUFFER, (IntPtr)TVS_EX_DOUBLEBUFFER);
```
and this call before handle is created:

```cs
DoubleBuffered = true;
```
and this call before handle is created and after `DoubleBuffered = true;`:

```cs
DoubleBuffered = false;
```
We will break him.

src/System.Windows.Forms/src/System/Windows/Forms/TreeView.cs
dreddy-work: with realignment of code, this is now required.

dreddy-work:
We are taking this and listen the feedback if we break any existing customers.

PR #7416, opened 2022-07-14

gpetrou:
## Proposed changes

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7416)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
@@ -2210,7 +2210,7 @@ namespace System.Windows.Forms
                 if (needPreProcess)
                 {
                     Control? target = FromChildHandle(lpmsg->hwnd);
-                    if (target is not null && (_control == target || _control.Contains(target)))
+                    if (_control == target || _control.Contains(target))
                     {
                         PreProcessControlState messageState = PreProcessControlMessageInternal(target, ref msg);
                         switch (messageState)
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
@@ -327,7 +327,7 @@ namespace System.Windows.Forms
                                 // check to see if this message should really go to a child
                                 //  control, and if so, map the point into that child's window
                                 //  coordinates
-                                Control realTarget = target.GetChildAtPoint(pt);
+                                Control? realTarget = target.GetChildAtPoint(pt);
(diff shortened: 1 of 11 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
```diff
@@ -327,7 +327,7 @@ internal void Close(OLECLOSE dwSaveOption)
                                 // check to see if this message should really go to a child
                                 //  control, and if so, map the point into that child's window
                                 //  coordinates
-                                Control realTarget = target.GetChildAtPoint(pt);
+                                Control? realTarget = target.GetChildAtPoint(pt);
                                 if (realTarget is not null && realTarget != target)
```
dreddy-work, line 331: line  331 check can be simplified to `if (realTarget != target)` given target can not be null at this line.

src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
```diff
@@ -2209,7 +2209,7 @@ internal unsafe HRESULT TranslateAccelerator(User32.MSG* lpmsg)
                 Message msg = Message.Create(lpmsg->hwnd, lpmsg->message, lpmsg->wParam, lpmsg->lParam);
                 if (needPreProcess)
                 {
-                    Control target = FromChildHandle(lpmsg->hwnd);
+                    Control? target = FromChildHandle(lpmsg->hwnd);
                     if (target is not null && (_control == target || _control.Contains(target)))
```
dreddy-work, line 2213: same here, condition on line 2213 can be simplified.

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -4795,7 +4795,7 @@ internal virtual bool CanSelectCore()
                 return false;
             }
 
-            for (Control ctl = this; ctl is not null; ctl = ctl._parent)
+            for (Control? ctl = this; ctl is not null; ctl = ctl._parent)
```
dreddy-work, line 4798: This look odd and could have been a while loop. I see multiple places in this file that used for loop like this. We may not need to change them in this PR but look odd with nullability.

RussKie (approved):
LGTM

src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
gpetrou: Hmm, since target is not null and realTarget could be null, `realTarget != target` alone would give a different behavior without the null check. I don't think that we should change this.

PR #7417, opened 2022-07-14
Use SequenceEqual instead of a handwritten loop in one case.

teo-tsirpanis:
And simplify some nested ifs.

We save one array allocation (by replacing the static readonly array with the `ReadOnlySpan` property), and code gets shortened.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7417)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataObject.BitmapBinder.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataObject.BitmapBinder.cs
@@ -25,7 +25,7 @@ namespace System.Windows.Forms
             private const string AllowedTypeName = "System.Drawing.Bitmap";
             private const string AllowedAssemblyName = "System.Drawing";
             // PublicKeyToken=b03f5f7f11d50a3a
-            private static readonly byte[] s_allowedToken = new byte[] { 0xB0, 0x3F, 0x5F, 0x7F, 0x11, 0xD5, 0x0A, 0x3A };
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/DataObject.BitmapBinder.cs
```diff
@@ -48,31 +48,11 @@ private class BitmapBinder : SerializationBinder
                     {
                     }
 
-                    if (nameToBind is not null)
+                    if (nameToBind is not null
+                        && string.Equals(nameToBind.Name, AllowedAssemblyName, StringComparison.Ordinal)
+                        && nameToBind.GetPublicKeyToken().AsSpan().SequenceEqual(AllowedToken))
```
RussKie, line 53: `GetPublicKeyToken()` is annotated as nullable, and it was previoulsy checked for null.

RussKie (changes requested):
Nice!

src/System.Windows.Forms/src/System/Windows/Forms/DataObject.BitmapBinder.cs
teo-tsirpanis: `AsSpan` on a `null` array returns an empty span. The comparison would fail.

https://github.com/dotnet/runtime/blob/aafa91036e1efd3d4dcb67eeb261cb6d8f774685/src/libraries/System.Private.CoreLib/src/System/Span.cs#L40-L46

RussKie (approved):

teo-tsirpanis:
@RussKie CI is green.

RussKie:
Thank you!

PR #4836, opened 2021-04-23
Editor Guidelines config

RussKie:
https://marketplace.visualstudio.com/items?itemName=PaulHarrington.EditorGuidelines

image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/4836)

PR #7431, opened 2022-07-16
Workaround for empty finalizer in ApplicationContext.

kirsan31:
Contribute to https://github.com/dotnet/winforms/issues/6858.
This is continuation of https://github.com/dotnet/winforms/pull/6878 based on discussion.
Documentation for calling `GC.SuppressFinalize` on users subclasses also need to be updated some how...

The benefit of this change is not obvious like with `DataGridView`. The one I see is that `ModalApplicationContext` is never disposed after `ShowDialog` call, so it's ending up in Freachable queue.

## Test methodology
Existing UT.
New UT for all subclasses coverage.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7431)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ApplicationContext.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ApplicationContext.cs
@@ -16,6 +16,16 @@ namespace System.Windows.Forms
     /// </summary>
     public class ApplicationContext : IDisposable
     {
+        /// <summary>
+        /// These are subclasses of the <see cref="ApplicationContext"/> for which we don't need to call the finalizer, because it's empty.
+        /// See https://github.com/dotnet/winforms/issues/6858.
+        /// </summary>
+        private static readonly HashSet<Type> s_typesWithEmptyFinalizer = new()
+        {
+            typeof(ApplicationContext),
+            Application.s_typeOfModalApplicationContext
+        };
+
         private Form? _mainForm;
 
         /// <summary>
@@ -27,20 +37,25 @@ namespace System.Windows.Forms
 
         /// <summary>
         ///  Creates a new ApplicationContext with the specified mainForm.
-        ///  If OnMainFormClosed is not overriden, the thread's message
+        ///  If OnMainFormClosed is not overridden, the thread's message
         ///  loop will be terminated when mainForm is closed.
         /// </summary>
         public ApplicationContext(Form? mainForm)
         {
+            if (s_typesWithEmptyFinalizer.Contains(GetType()))
+                GC.SuppressFinalize(this);
+
             MainForm = mainForm;
         }
 
+        // NOTE: currently this finalizer is unneeded (empty). See https://github.com/dotnet/winforms/issues/6858.
+        // All classes that are not need to be finalized contains in ApplicationContext.s_typesWithEmptyFinalizer collection. Consider to modify it if needed.
         ~ApplicationContext() => Dispose(false);
 
         /// <summary>
         ///  Determines the mainForm for this context. This may be changed
         ///  at anytime.
-        ///  If OnMainFormClosed is not overriden, the thread's message
+        ///  If OnMainFormClosed is not overridden, the thread's message
         ///  loop will be terminated when mainForm is closed.
         /// </summary>
         public Form? MainForm
@@ -102,6 +117,11 @@ namespace System.Windows.Forms
                     _mainForm = null;
                 }
             }
+
+            // If you are adding releasing unmanaged resources code here (disposing == false), you need to:
+            // remove GC.SuppressFinalize from constructor of this class and from all of its subclasses
+            // remove ApplicationContext_Subclasses_SuppressFinalizeCall test
+            // modify ~ApplicationContext() description.
         }
 
         /// <summary>
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ApplicationContext.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ApplicationContext.cs
@@ -16,16 +16,6 @@ namespace System.Windows.Forms
     /// </summary>
     public class ApplicationContext : IDisposable
     {
-        /// <summary>
-        ///  These are subclasses of the <see cref="ApplicationContext"/> for which we don't need to call the finalizer,
-        ///  because it's empty. See https://github.com/dotnet/winforms/issues/6858.
-        /// </summary>
-        private static readonly HashSet<Type> s_typesWithEmptyFinalizer = new()
-        {
-            typeof(ApplicationContext),
-            Application.s_typeOfModalApplicationContext
-        };
-
         private Form? _mainForm;
 
         /// <summary>
@@ -42,14 +32,16 @@ namespace System.Windows.Forms
         /// </summary>
         public ApplicationContext(Form? mainForm)
         {
-            if (s_typesWithEmptyFinalizer.Contains(GetType()))
+            //  These are subclasses of the ApplicationContext for which we don't need to call the finalizer,
+            //  because it's empty. See https://github.com/dotnet/winforms/issues/6858.
+            if (GetType() == typeof(ApplicationContext) || GetType() == Application.s_typeOfModalApplicationContext)
                 GC.SuppressFinalize(this);
 
             MainForm = mainForm;
         }
 
         // NOTE: currently this finalizer is unneeded (empty). See https://github.com/dotnet/winforms/issues/6858.
-        // All classes that are not need to be finalized contains in ApplicationContext.s_typesWithEmptyFinalizer collection.
+        // All classes that are not need to be finalized must be checked in ApplicationContext(Form? mainForm) constructor.
         // Consider to modify it if needed.
         ~ApplicationContext() => Dispose(false);
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ApplicationContext.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ApplicationContext.cs
@@ -17,8 +17,8 @@ namespace System.Windows.Forms
     public class ApplicationContext : IDisposable
     {
         /// <summary>
-        /// These are subclasses of the <see cref="ApplicationContext"/> for which we don't need to call the finalizer, because it's empty.
-        /// See https://github.com/dotnet/winforms/issues/6858.
+        ///  These are subclasses of the <see cref="ApplicationContext"/> for which we don't need to call the finalizer,
+        ///  because it's empty. See https://github.com/dotnet/winforms/issues/6858.
         /// </summary>
         private static readonly HashSet<Type> s_typesWithEmptyFinalizer = new()
         {
@@ -49,7 +49,8 @@ namespace System.Windows.Forms
         }
 
         // NOTE: currently this finalizer is unneeded (empty). See https://github.com/dotnet/winforms/issues/6858.
-        // All classes that are not need to be finalized contains in ApplicationContext.s_typesWithEmptyFinalizer collection. Consider to modify it if needed.
+        // All classes that are not need to be finalized contains in ApplicationContext.s_typesWithEmptyFinalizer collection.
+        // Consider to modify it if needed.
         ~ApplicationContext() => Dispose(false);
 
         /// <summary>
@@ -119,9 +120,9 @@ namespace System.Windows.Forms
             }
 
             // If you are adding releasing unmanaged resources code here (disposing == false), you need to:
-            // remove GC.SuppressFinalize from constructor of this class and from all of its subclasses
-            // remove ApplicationContext_Subclasses_SuppressFinalizeCall test
-            // modify ~ApplicationContext() description.
+            // 1. remove GC.SuppressFinalize from constructor of this class and from all of its subclasses
+            // 2. remove ApplicationContext_Subclasses_SuppressFinalizeCall test
+            // 3. modify ~ApplicationContext() description.
         }
 
         /// <summary>
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ApplicationContextTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ApplicationContextTests.cs
@@ -379,15 +379,12 @@ namespace System.Windows.Forms.Tests
         [WinFormsFact]
         public void ApplicationContext_Subclasses_SuppressFinalizeCall()
         {
-            TestAccessor<ApplicationContext> testAccessor = new(null);
-            var typesWithEmptyFinalizer = testAccessor.Dynamic.s_typesWithEmptyFinalizer as HashSet<Type>;
-
             foreach (var type in typeof(ApplicationContext).Assembly.GetTypes().
                 Where(type => type == typeof(ApplicationContext) || type.IsSubclassOf(typeof(ApplicationContext))))
             {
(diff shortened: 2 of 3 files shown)


Tanya-Solyanik (commented):
Looks good, I added some nits.

src/System.Windows.Forms/src/System/Windows/Forms/ApplicationContext.cs
```diff
@@ -27,20 +37,25 @@ public ApplicationContext() : this(null)
...
         public ApplicationContext(Form? mainForm)
         {
+            if (s_typesWithEmptyFinalizer.Contains(GetType()))
+                GC.SuppressFinalize(this);
+
             MainForm = mainForm;
         }
 
+        // NOTE: currently this finalizer is unneeded (empty). See https://github.com/dotnet/winforms/issues/6858.
+        // All classes that are not need to be finalized contains in ApplicationContext.s_typesWithEmptyFinalizer collection. Consider to modify it if needed.
```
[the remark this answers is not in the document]
RussKie: To elaborate: we strive to wrap the lines around 120 mark, and it's acceptable to stretch to no more than 150 chars (with some exceptions being URLs).
image

[EditorGuidelines VS extension](https://marketplace.visualstudio.com/items?itemName=PaulHarrington.EditorGuidelines) makes it easier to visualise (see https://github.com/dotnet/winforms/pull/4836).

src/System.Windows.Forms/src/System/Windows/Forms/ApplicationContext.cs
```diff
@@ -102,6 +117,11 @@ protected virtual void Dispose(bool disposing)
                     _mainForm = null;
                 }
             }
+
+            // If you are adding releasing unmanaged resources code here (disposing == false), you need to:
+            // remove GC.SuppressFinalize from constructor of this class and from all of its subclasses
+            // remove ApplicationContext_Subclasses_SuppressFinalizeCall test
+            // modify ~ApplicationContext() description.
```
RussKie, line 124:
```suggestion
            // If you are adding releasing unmanaged resources code here (disposing == false), you need to:
            // 1. remove GC.SuppressFinalize from constructor of this class and from all of its subclasses
            // 2. remove ApplicationContext_Subclasses_SuppressFinalizeCall test
            // 3, modify ~ApplicationContext() description.
```

RussKie (commented):
👍

kirsan31:
@RussKie 
I've looked at [remaining](https://github.com/dotnet/winforms/issues/6858) `ResXResourceWriter` and `ResXResourceReader` and I think we can leave them as is. They are most using in designer and If used by the user, he _must_ call dispose any way.
What do you think?

Аlso on my machine I constantly see failed tests, details [here](also on my machine I constantly see failed tests, details here.).

RussKie:
> I've looked at https://github.com/dotnet/winforms/issues/6858 ResXResourceWriter and ResXResourceReader and I think we can leave them as is. They are most using in designer and If used by the user, he must call dispose any way.
What do you think?

Works for me, thank you 

> details [here](also on my machine I constantly see failed tests, details here.).

The link didn't seem to go through.

RussKie (dismissed):

Tanya-Solyanik (dismissed):
This change looks good. Could you please elaborate, how it Dispose call enforced in the case of resx reader/writer classes?

src/System.Windows.Forms/src/System/Windows/Forms/ApplicationContext.cs
```diff
@@ -27,20 +37,26 @@ public ApplicationContext() : this(null)
 
         /// <summary>
         ///  Creates a new ApplicationContext with the specified mainForm.
-        ///  If OnMainFormClosed is not overriden, the thread's message
+        ///  If OnMainFormClosed is not overridden, the thread's message
         ///  loop will be terminated when mainForm is closed.
         /// </summary>
         public ApplicationContext(Form? mainForm)
         {
+            if (s_typesWithEmptyFinalizer.Contains(GetType()))
```
JeremyKuhne, line 45: I don't think this will get elided by the JIT. I'd rather see this be `GetType() == typeof(ApplicationContext)`, etc. to allow that to happen.

JeremyKuhne (dismissed):
Comment added on type checking.

src/System.Windows.Forms/src/System/Windows/Forms/ApplicationContext.cs
kirsan31: I don't get your idea - sorry, can you clarify pls?
Also, initially the form of the check was exactly as you wrote `GetType() == typeof(ApplicationContext)`, but the variant with  `HashSet` was suggested [here](https://github.com/dotnet/winforms/pull/6978#discussion_r856565536).

kirsan31: /cc @JeremyKuhne

elachlan:
@kirsan31 were you waiting on anything for this PR?

src/System.Windows.Forms/src/System/Windows/Forms/ApplicationContext.cs
Tanya-Solyanik: @kirsan31  - perf considerations are more important than maintainability ones, sorry about stirring you wrong, please revert my suggestion.

kirsan31: @Tanya-Solyanik 
> perf considerations are more important than maintainability ones, sorry about stirring you wrong, please revert my suggestion.

I defiantly will do, when have a time... Do we need to do same revert with already merged https://github.com/dotnet/winforms/pull/6978?

kirsan31:
@elachlan 

> were you waiting on anything for this PR?

We were waiting for an answer from Jeremy ([dialog](https://github.com/dotnet/winforms/pull/7431#discussion_r925025502) above).

Tanya-Solyanik:
>Do we need to do same revert with already merged https://github.com/dotnet/winforms/pull/6978?

@kirsan31 - depends on whether the JIT would understand all these conditions are a no-op.

@JeremyKuhne 
>I don't think this will get elided by the JIT. I'd rather see this be GetType() == typeof(ApplicationContext), etc. to allow that to happen.

Would a condition that matches this type against multiple other types get elided by the JIT? 

vat type

src/System.Windows.Forms/src/System/Windows/Forms/ApplicationContext.cs
JeremyKuhne: Sorry I missed the tag back. When the JIT sees `typeof() == GetType()` or `typeof(T) == typeof()` it knows that is effectively a constant and will remove blocks that always false and remove the condition check.

dreddy-work:
@kirsan31, hope you got answers to proceed here.

kirsan31:
@dreddy-work 

> hope you got answers to proceed here.

Yes and no.
We still have question by @Tanya-Solyanik 

> Would a condition that matches this type against multiple other types get elided by the JIT?

To which I would also like to receive an answer.

Will this be optimized?
```cs
public ApplicationContext(Form? mainForm)
{
    var tp = GetType();
    if (tp == typeof(ApplicationContext) || tp == s_typeOfModalApplicationContext)
        GC.SuppressFinalize(this);
}
```

And this?

```cs
public DataGridViewElement()
{
    if (NeedSuppressFinalize(GetType()))
        GC.SuppressFinalize(this);
}

private static bool NeedSuppressFinalize(Type type)
{
    return type == typeof(DataGridViewBand) || type == typeof(DataGridViewColumn) || type == typeof(DataGridViewButtonColumn) || type == typeof(DataGridViewCheckBoxColumn) ||
        type == typeof(DataGridViewComboBoxColumn) || type == typeof(DataGridViewImageColumn) || type == typeof(DataGridViewLinkColumn) || type == typeof(DataGridViewTextBoxColumn) ||
        type == typeof(DataGridViewRow) || type == typeof(DataGridViewCell) || type == typeof(DataGridViewButtonCell) || type == typeof(DataGridViewCheckBoxCell) ||
        type == typeof(DataGridViewComboBoxCell) || type == typeof(DataGridViewHeaderCell) || type == typeof(DataGridViewColumnHeaderCell) || type == typeof(DataGridViewTopLeftHeaderCell) ||
        type == typeof(DataGridViewRowHeaderCell) || type == typeof(DataGridViewImageCell) || type == typeof(DataGridViewLinkCell) || type == typeof(DataGridViewTextBoxCell);
}
```

JeremyKuhne:
> Will this be optimized?

You can see some of the behavior in [SharpLab](https://sharplab.io/#v2:EYLgtghglgdgNAFxBAzmAPgAQEwEYCwAUJgMwAEOZAwmQN5FmMXlUAUAlA0/YU32VABmZVgHEApggAqATwAO4jmQC8ysgnniA9oNZV2ZdOjITpmpavWadrACLtOvfox7PnoqgDoAygFc5cgBO4igoAGKwEAA2UABeiggAFlAo7ADcXM4AvplkOU6MuaRkwFpaUWQAskqubpgA7GQAEuJRCoGeAHLi4gAmfgHBoREw0XEJ1rr66bn5fEXkpeVknTW5fA0mkrIKFmoaCjb6GQV5uQslZRVSIQgcubXONADGKmQw4gDuHCduFI3PLo/WZEfJEIjFSi2Mggah0M7EciUACi8PykOwZDCaIhSNwADYKJiWm1xIEiI9mBQCZdlt0+gMgiFwpEYvFWDtxFYFI4+JSNo0DlzLEKjgYjNzhftJnZxcYhSppYddMiZqd8mDTkA).

From that experiment, it appears if you're checking `GetType()` against `typeof()` you need to do that inline rather than passing `GetType()` to a helper. Checking against two types doesn't optimize quite as well from my fiddling around either.

kirsan31:
@JeremyKuhne 

> You can see some of the behavior in [SharpLab](https://sharplab.io/#v2:EYLgtghglgdgNAFxBAzmAPgAQEwEYCwAUJgMwAEOZAwmQN5FmMXlUAUAlA0/YU32VABmZVgHEApggAqATwAO4jmQC8ysgnniA9oNZV2ZdOjITpmpavWadrACLtOvfox7PnoqgDoAygFc5cgBO4igoAGKwEAA2UABeiggAFlAo7ADcXM4AvplkOU6MuaRkwFpaUWQAskqubpgA7GQAEuJRCoGeAHLi4gAmfgHBoREw0XEJ1rr66bn5fEXkpeVknTW5fA0mkrIKFmoaCjb6GQV5uQslZRVSIQgcubXONADGKmQw4gDuHCduFI3PLo/WZEfJEIjFSi2Mggah0M7EciUACi8PykOwZDCaIhSNwADYKJiWm1xIEiI9mBQCZdlt0+gMgiFwpEYvFWDtxFYFI4+JSNo0DlzLEKjgYjNzhftJnZxcYhSppYddMiZqd8mDTkA).

According to this link, even this sentence:

> Sorry I missed the tag back. When the JIT sees typeof() == GetType() or typeof(T) == typeof() it knows that is effectively a constant and will remove blocks that always false and remove the condition check.

Is wrong?

image

JeremyKuhne:
@kirsan31 

Sorry, I should have given more detail. If you look at the `Test()` method, instead of calling `N()` it just returns true. It wouldn't if you called the helper method with `GetType()`.

You make things easier for the JIT by reducing the complexity.  You give the JIT more opportunities to remove things that are effectively constant. It doesn't figure everything out, but it gets better and better with each release.  It doesn't eliminate in the constructor case here, but it likely could in the future. At the very least you avoid a call to a helper, which is significant in this case with the number of types you're checking.

Note that the JIT is particularly good at removing these sorts of checks when generating the code for generics, such as `typeof(T) == typeof(char)`.

kirsan31:
Thanks @JeremyKuhne , I get the point.

Based on the answer above, I think we need to redo this PR and https://github.com/dotnet/winforms/pull/6978 too. To smile type check `if` in the constructor. 
@Tanya-Solyanik  what do you think?

Tanya-Solyanik:
>Based on the answer above, I think we need to redo this PR and https://github.com/dotnet/winforms/pull/6978 too. To smile type check if in the constructor.
@Tanya-Solyanik what do you think?

Makes sense.  We'll prioritize perf in this case.

src/System.Windows.Forms/src/System/Windows/Forms/ApplicationContext.cs
sharwell, line 45: 💭 A more specific test here would be to resolve `Dispose(bool)` for the current instance, and see if the declaring type is `ApplicationContext` (i.e. the type being created does not override the empty `Dispose(bool)` method).

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ApplicationContextTests.cs
```diff
@@ -376,6 +376,21 @@ public void OnMainFormClosed_InvokeWithThreadExit_CallsHandler(object sender, Ev
             Assert.Equal(2, callCount);
         }
 
+        [WinFormsFact]
+        public void ApplicationContext_Subclasses_SuppressFinalizeCall()
```
sharwell, line 380: 💡 If you use a `WinFormsTheory` with member data, you can pass `Type type` as a parameter and it will cause specific instances of the test to fail only for the associated type that wasn't in the collection.

src/System.Windows.Forms/src/System/Windows/Forms/ApplicationContext.cs
kirsan31: I agree, but we decided here that even a hashtable is slow and needs to be changed to a simple type comparison. And for `Dispose(bool)` check we need to use reflection...
But in any case, I think that the number of subclasses here is much less than just base classes (and with overridden `Dispose(bool)` even less).

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ApplicationContextTests.cs
kirsan31: And what benefit from `WinFormsTheory` - we not need to test different inputs here 🤔

kirsan31:
@JeremyKuhne 
I want to say that this result surprised me a lot, I thought that the second entry will be more optimal 😯
image

JeremyKuhne:
> this result surprised me a lot

It doesn't surprise me too much- there is more context the JIT has to track when you're assigning to a variable and using that variable several times. To be fair, I don't _really_ understand how the JIT works under the covers, so this is semi-educated guess work for me. :)

kirsan31:
Are we stuck here again?

dreddy-work:
> Are we stuck here again?
Thank you @kirsan31, @JeremyKuhne is on vacation now and team is going to look into this and update this week if any further feedback.

Tanya-Solyanik (approved):

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ApplicationContextTests.cs
sharwell: > we not need to test different inputs here

We already are testing different inputs here. `ApplicationContext` and every type derived from it is a distinct test.

kirsan31: Oops, I was just about to watch it again when the PR was already merged. 🙄

PR #7444, opened 2022-07-19

fernandanavamoya:
Converted these threads:
-GetCurrentActCtx
-GetExitCodeThread
-GetThreadLocale
-GlobalFree

Fixed the information messages in:
-COM2PropertyDescriptor.cs
-Control.cs

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7444)


fernandanavamoya:
cc: @AArnott

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
```diff
@@ -2299,7 +2300,7 @@ private object GetAmbientProperty(Ole32.DispatchID dispid)
                     return rval;
                 case Ole32.DispatchID.AMBIENT_LOCALEID:
                     Debug.WriteLineIf(s_axHTraceSwitch.TraceVerbose, "asked for localeid");
-                    return Kernel32.GetThreadLocale().RawValue;
+                    return PInvoke.GetThreadLocale();
```
dreddy-work, line 2303: Should this still return RawValue? I can't trace the usage so wanted to understand the change.

dreddy-work: Nevermind, LCID was internal Winforms struct that earlier casted.

dreddy-work (approved):

src/System.Windows.Forms.Primitives/src/NativeMethods.txt
```diff
@@ -8,3 +8,6 @@ GlobalUnlock
 GetCurrentProcessId
 GlobalReAlloc
 GlobalFree
+GetExitCodeThread
```
Tanya-Solyanik, line 11: Is it possible to add a comment to this file that would explain what it is for?

Tanya-Solyanik (approved):

src/System.Windows.Forms.Primitives/src/NativeMethods.txt
```diff
@@ -8,3 +8,6 @@ GlobalUnlock
...
+GetThreadLocale
+GetCurrentActCtx
```
RussKie, line 13: Please have this file sorted in alphabetical order.

I know Notepadd++ can do it easily, but other editors have similar functionality as well
image

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/ThemingScope.cs
```diff
@@ -12,18 +14,19 @@ namespace System.Windows.Forms
...
         ///  really need to activate the activationcontext.  This should be pretty fast.
         /// </summary>
-        private static bool IsContextActive()
+        private unsafe static bool IsContextActive()
         {
+            Foundation.HANDLE current;
             return s_contextCreationSucceeded
-                && Kernel32.GetCurrentActCtx(out IntPtr current).IsTrue()
-                && current == s_hActCtx;
+                && PInvoke.GetCurrentActCtx(&current) == true
```
RussKie, line 28: Does `PInvoke.GetCurrentActCtx(...)` return a bool?

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/ThemingScope.cs
```diff
@@ -2,6 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using Windows.Win32;
+using Foundation = Windows.Win32.Foundation;
```
RussKie, line 6: Why do we need this alias?
image

Ditto below

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/ThemingScope.cs
RussKie: The conversion is implicit
```suggestion
                && PInvoke.GetCurrentActCtx(&current)
```

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/ThemingScope.cs
AArnott: What alias do you mean? I don't see any alias in your screenshot. Do you mean the `using static` line?

src/System.Windows.Forms.Primitives/src/NativeMethods.txt
AArnott: The file format does support lines starting with `//` for comments. That said, I don't recommend it as it wouldn't sort properly and it would probably get outdated. Why not just search for the method name in the code (or use Find References) if you're curious how a method is used?

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
```diff
@@ -3231,21 +3232,21 @@ public unsafe bool HasPropertyPages()
 
         private unsafe void ShowPropertyPageForDispid(Ole32.DispatchID dispid, Guid guid)
         {
-            IntPtr pUnk = Marshal.GetIUnknownForObject(GetOcx());
+            nint pUnk = Marshal.GetIUnknownForObject(GetOcx());
```
AArnott, line 3235: nit: I would leave this at `IntPtr`. You probably don't need the syntax support of `nint` (e.g. arithmetic), and IMO `IntPtr` is better for representing an opaque value like a pointer.

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -3904,23 +3905,23 @@ private void WaitForWaitHandle(WaitHandle waitHandle)
...
-                    returnValue = Kernel32.GetExitCodeThread(threadHandle, out exitCode);
+                    returnValue = PInvoke.GetExitCodeThread((Foundation.HANDLE)threadHandle, &exitCode);
                 }
 
                 //If we didn't find the thread, or if GetExitCodeThread failed, we don't know the thread's state:
                 //if we don't know, we shouldn't throw.
-                if ((returnValue.IsTrue() && exitCode != NativeMethods.STILL_ACTIVE) ||
-                    (returnValue.IsFalse() && Marshal.GetLastWin32Error() == ERROR.INVALID_HANDLE) ||
+                if ((returnValue == true && exitCode != NativeMethods.STILL_ACTIVE) ||
+                    (returnValue == false && Marshal.GetLastWin32Error() == ERROR.INVALID_HANDLE) ||
```
AArnott, line 3924: `returnValue` and `!returnValue` should work. No need to spell out ` == true` and ` == false` unless that's part of your codebase style guide, which at least for `== true` would be weird, IMO.

src/System.Windows.Forms.Primitives/src/NativeMethods.txt
AArnott: Visual Studio's command for sorting is a bit out of the way. I typically type "sort" into the search box on the VS title bar and the "Sort Lines" command comes up.

AArnott (approved):
Change looks good, although I would personally try to avoid mixing the syntax only changes (e.g. `IntPtr` to `nint`) with the CsWin32 changes.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
RussKie: In .NET 7 [`nint` and `IntPtr` will refer to the exact same thing](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-11#numeric-intptr-and-uintptr) so changing between them does not make any difference.

src/System.Windows.Forms.Primitives/src/NativeMethods.txt
RussKie: Ah sweet, I learned something new :)

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/ThemingScope.cs
elachlan: Foundation holds BOOL and Handle as well as other types.

RussKie: Correct, but in this case it doesn't appear necessary as the screenshot shows. I suspect it's probably true for other instances in this PR too.

src/System.Windows.Forms.Primitives/src/NativeMethods.txt
Tanya-Solyanik: I'm curious how this file is used.

dreddy-work: Any public documentation on this? Also, does it require explicit build of the project for source generators to trigger in?

AArnott: The documentation is at https://github.com/microsoft/cswin32 (the readme file). 
This is a standard roslyn source generator. As such, it does _not_ require an explicit build. In fact the generated source code is never saved to disk, but rather produced in memory during compilation.

Tanya-Solyanik: It would be nice to add a link to this readme file to the top of the NativeMethods.txt file. If this is a single line comment, it wouldn't interfere with sorting file lines.

Issue #7445, opened 2022-07-19
Start using CsWin32


elachlan:
### Is your feature request related to a problem? Please describe

This issue is to be used to help centralize information around the change to implement [CsWin32](https://github.com/microsoft/CsWin32).

**Blockers:**
~~https://github.com/microsoft/CsWin32/issues/639~~
~~https://github.com/microsoft/CsWin32/issues/605~~
https://github.com/microsoft/CsWin32/issues/600
https://github.com/microsoft/CsWin32/issues/644
~~https://github.com/microsoft/CsWin32/issues/624~~
~~https://github.com/microsoft/CsWin32/issues/625~~
https://github.com/microsoft/CsWin32/issues/324
https://github.com/dotnet/runtime/issues/59013

**Performance:**
https://github.com/microsoft/CsWin32/issues/595

**Useful:**
https://github.com/microsoft/CsWin32/issues/596
~~https://github.com/microsoft/CsWin32/issues/606~~
https://github.com/microsoft/CsWin32/issues/602
https://github.com/microsoft/CsWin32/issues/610
https://github.com/microsoft/CsWin32/issues/122
~~https://github.com/microsoft/CsWin32/issues/626~~

**Associated Issues/Discussions:**
https://github.com/dotnet/winforms/pull/4751
~~https://github.com/dotnet/winforms/pull/7392~~
~~https://github.com/dotnet/winforms/pull/7444~~
~~https://github.com/dotnet/winforms/pull/7446~~
https://github.com/dotnet/winforms/issues/7468

elachlan:
@JeremyKuhne I have created an issue here to help centralize issues/information around CSWin32. 

Before anything gets too deep, the Winforms team should decide on how they want to split up the NativeMethods.txt file.

JeremyKuhne:
> Before anything gets too deep, the Winforms team should decide on how they want to split up the NativeMethods.txt file.

Not sure how we'd want to split this up in ways that make sense given everything going to one class. Any ideas are certainly welcome.

Note that this is currently a prototype effort targeting .NET 8. We have an intern working on this (@fernandanavamoya) with support from myself and our returning @lonitra.

elachlan:
I have asked the CsWin32 team about the `NativeMethods.txt` in https://github.com/microsoft/CsWin32/issues/606.

[...]

PR #7451, opened 2022-07-19
User32 ActivateKeyBoardLayout

elachlan:
#7445

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7451)


src/System.Windows.Forms.Primitives/src/System.Windows.Forms.Primitives.csproj
```diff
@@ -45,4 +45,8 @@
     </EmbeddedResource>
   </ItemGroup>
 
+  <ItemGroup>
+    <AdditionalFiles Include="Interop\User32\NativeMethods.txt" />
```
RussKie, line 49: Would globing work?
```suggestion
    <AdditionalFiles Include="Interop\**\NativeMethods.txt" />
```

RussKie (approved):

src/System.Windows.Forms.Primitives/src/System.Windows.Forms.Primitives.csproj
elachlan: @RussKie, @AArnott stated that we should use a centralized version of the `NativeMethods.txt` in https://github.com/microsoft/CsWin32/issues/606

> We don't recommend splitting NativeMethods.txt up. At least, I haven't come across any scenario where it would be beneficial. I heard someone suggest a couple weeks ago that perhaps one in each feature directory of a library would make sense, where the APIs used by that feature would be listed. But that's not really sustainable because there is no enforcement, and it would be a lot of extra bookkeeping on the part of the consumer to try to avoid using APIs that weren't asked for in their local file.
The reason we support multiple files is for Shared Project scenarios, where legitimately shared code requires p/invokes and should carry its own NativeMethods.txt file so that anyone who consumes the Shared Project will automatically generate the APIs it needs.

RussKie: >  At least, I haven't come across any scenario where it would be beneficial.

I think in our case it's clearly benefitial - we use a lot of interops 
https://github.com/dotnet/winforms/tree/main/src/System.Windows.Forms.Primitives/src/Interop

JeremyKuhne: I don't want to have multiple NativeMethods.txt as figuring out where to put things and to find existing entries is not clear.

JeremyKuhne (changes requested):
Please keep a single NativeMethods.txt in the root for discoverability.

src/System.Windows.Forms.Primitives/src/System.Windows.Forms.Primitives.csproj
RussKie: @JeremyKuhne how is it different from how we do it now?

JeremyKuhne: No matter where you put these NativeMethods.txt files you'll get the same top level `PInvoke` method call. Other support types are under a structure defined by the Windows team in the metadata (*not* the source dll).

So this isn't a matter of matching code structure, and it is also just a long list of method names, no implementation.

lonitra (changes requested):

src/System.Windows.Forms.Primitives/src/NativeMethods.txt
```diff
@@ -1,9 +1,12 @@
+﻿ActivateKeyboardLayout
```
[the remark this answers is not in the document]
elachlan: It was a zero width space character. I have made a PR to fix the processing in https://github.com/microsoft/CsWin32/pull/609

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
```diff
@@ -51,7 +53,7 @@ public static InputLanguage CurrentInputLanguage
                     value = DefaultInputLanguage;
                 }
 
-                IntPtr handleOld = User32.ActivateKeyboardLayout(value.Handle, 0);
+                HKL handleOld = PInvoke.ActivateKeyboardLayout(new HKL(value.Handle), 0);
                 if (handleOld == IntPtr.Zero)
```
dreddy-work, line 57: should we still comparing it with IntPtr.Zero? can we just check `== 0` here?

elachlan: I tried, I think the compiler doesn't like it because HKL doesn't have an implicit conversion to int? I wasn't sure what to do.

AArnott: I just tried this syntax. It compiled fine when targeting .NET 6 and net472:

```cs
using Windows.Win32.UI.TextServices;

HKL hkl = default;
if (hkl == 0)
{
}
```

This is using CsWin32 0.2.1-beta.

JeremyKuhne: `HKL` should have a `.NULL` on it, if not get the `.Value` and compare to `0`.

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
```diff
@@ -84,12 +86,12 @@ public unsafe static InputLanguageCollection InstalledInputLanguages
         {
             get
             {
-                int size = User32.GetKeyboardLayoutList(0, null);
+                int size = PInvoke.GetKeyboardLayoutList(0, null);
 
-                var handles = new IntPtr[size];
-                fixed (IntPtr* pHandles = handles)
+                var handles = new HKL[size];
+                fixed (HKL* pHandles = handles)
```
AArnott, line 92: No need for `fixed` here any more. With CsWin32 you can do this:

```cs
var handles = new HKL[5];
PInvoke.GetKeyboardLayoutList(handles);
```

AArnott (approved):

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
AArnott: HKL looks like a candidate for the `IsNull` property that we're supposed to generate on pointer-sized handles, per https://github.com/microsoft/CsWin32/issues/206. If HKL is a handle (or a pointer) and we missed that, can you file a new issue for it?

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
```diff
@@ -51,8 +53,8 @@ public static InputLanguage CurrentInputLanguage
                     value = DefaultInputLanguage;
                 }
 
-                IntPtr handleOld = User32.ActivateKeyboardLayout(value.Handle, 0);
-                if (handleOld == IntPtr.Zero)
+                HKL handleOld = PInvoke.ActivateKeyboardLayout(new HKL(value.Handle), 0);
+                if (handleOld.Value == 0)
```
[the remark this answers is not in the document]
AArnott: Weird. Well, I'm glad you have a workaround.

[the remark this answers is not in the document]
elachlan: Doing a cast to nint works and there is no nint.Zero. HKL has an operator for nint, but not int. Should I just do the cast or is there a better way?
```
 HKL handleOld = PInvoke.ActivateKeyboardLayout(new HKL(value.Handle), 0);
                if (handleOld == (nint)0)
                {
                    throw new ArgumentException(SR.ErrorBadInputLanguage, nameof(value));
                }
```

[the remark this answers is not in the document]
AArnott: The cast isn't necessary in my project. Maybe it's an odd C# 11 or .NET 7 targeting thing. You can choose whatever syntax you're happiest with. They're all equivalent in the end.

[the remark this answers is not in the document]
RussKie: Does `if (handleOld == default)` work?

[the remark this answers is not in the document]
elachlan: > Does `if (handleOld == default)` work?

yes!

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
elachlan: done. https://github.com/microsoft/CsWin32/issues/610

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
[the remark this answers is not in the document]
RussKie: for integral types such as `int`, `IntPtr`, etc. `default` == 0. So it's equivalent.

lonitra (approved):

dreddy-work (approved):

dreddy-work:
@elachlan , can you resolve merge conflict?

PR #7314, opened 2022-06-16
Adding UIA support for WebBrowser

dkazennov:
Partially implements https://github.com/dotnet/winforms/issues/3421
For .Net 7

## Proposed changes

- Updated "SupportsUiaProviders" flag to WebBrowser
- Added and implemented WebBrowser.WebBrowserAccessibleObject
- Added unit tests
- Testing application 
[WebBrowserWinFormsApp.zip](https://github.com/dotnet/winforms/files/8919406/WebBrowserWinFormsApp.zip)

### Before
before

ProviderDescription:	"[pid:22832,providerId:0x402AA Main:Nested [pid:27400,providerId:0x402AA Main(parent link):Microsoft: MSAA Proxy (unmanaged:UIAutomationCore.dll)]; Hwnd(parent link):Microsoft: HWND Proxy (unmanaged:uiautomationcore.dll)]"
	
### After
after

ProviderDescription:	"[pid:25792,providerId:0x70BA2 Main:Nested [pid:30696,providerId:0x70BA2 Main(parent link):Unidentified Provider (unmanaged:coreclr.dll)]; Hwnd(parent link):Microsoft: HWND Proxy (unmanaged:uiautomationcore.dll)]"

## Regression? 

- No

## Risk

- Minimal

## Test methodology

- Unit tests
- Manual with custom application

## Accessibility testing

- Inspect
- Accessibility Insights

 

## Test environment

dotnet : 7.0.0-preview.4.22229.4

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7314)

PR #7453, opened 2022-07-19
Fixing TabPage memory leaks

### LINKED ISSUE #7415: TabPage class and it's accessible object are leaking in memory
### .NET version

.NET Version: 7.0

No

_No response_

### Issue description

TabPage and it's accessible object are left in memory after the ToolStrip control instance was selected with Narrator and the form with ToolStrip is closed, garbage is collected, but Narrator is still opened.

### Steps to reproduce

Create a parent form with two buttons: one will be opening a child form with the TabPage control and the other will collect the garbage.
Press the second button to open the child form with TabPage.
Start the Narrator.
Select the TabPage components using the Narrator.
Close the child form.
DON'T close the Narrator.
Look for TabPage accessible objects left in memory (using WinDbg, or dotMemory, or another suitable tool).

dkazennov:
Fixes #7415 

## Proposed changes

- Changed TabPage class
- Changed TabCollection class
- Changed TabPageAccessibleObject class

Unit tests affected:
- TabPage_TabAccessibleObjectTests.TabAccessibleObject_RuntimeId_ReturnsExpected (test changed)
- added a new unit test TabAccessibilityObject_IsDisconnected_WhenTabPageReleasesUiaProvider which checks for _tabAccessibilityObject to be disconnected.

## Customer Impact

- Less memory leaks of TabPage 

## Regression? 

- No

## Risk

- Minimal

### Before

image

### After

image

## Test methodology 

- CTI
- Manually (via WinDbg)

## Accessibility testing  

- Using AI, Inspect, Narrator

 

## Test environment(s) 

- .NET 7.0 Preview 5
- Windows 10

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7453)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TabControl.TabPageCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TabControl.TabPageCollection.cs
@@ -340,6 +340,10 @@ namespace System.Windows.Forms
                 ArgumentNullException.ThrowIfNull(value);
 
                 _owner.Controls.Remove(value);
+                if (value.IsHandleCreated)
+                {
+                    value.ReleaseUiaProvider(value.HandleInternal);
+                }
             }
 
             void IList.Remove(object? value)
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TabPage.TabAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TabPage.TabAccessibleObject.cs
@@ -73,7 +73,7 @@ namespace System.Windows.Forms
                     OwningTabControl is null
                         ? PARAM.ToInt(IntPtr.Zero)
                         : PARAM.ToInt(OwningTabControl.InternalHandle),
-                    GetChildId()
+                    GetHashCode()
                 };
 
             private int CurrentIndex => OwningTabControl?.TabPages.IndexOf(_owningTabPage) ?? -1;
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TabPage.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TabPage.cs
@@ -8,6 +8,7 @@ using System.Diagnostics.CodeAnalysis;
 using System.Drawing;
 using System.Drawing.Design;
 using System.Windows.Forms.Layout;
+using static Interop;
 
 namespace System.Windows.Forms
 {
@@ -619,6 +620,18 @@ namespace System.Windows.Forms
             else
             {
                 base.OnPaintBackground(e);
+            }
+        }
+
+        internal override void ReleaseUiaProvider(IntPtr handle)
+        {
+            base.ReleaseUiaProvider(handle);
+
+            if (_tabAccessibilityObject is not null && OsVersion.IsWindows8OrGreater)
+            {
+                HRESULT result = UiaCore.UiaDisconnectProvider(_tabAccessibilityObject);
+                Debug.Assert(result == HRESULT.S_OK);
+                _tabAccessibilityObject = null;
             }
         }
 
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TabPage.TabAccessibleObjectTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TabPage.TabAccessibleObjectTests.cs
@@ -33,6 +33,24 @@ namespace System.Windows.Forms.Tests
             Assert.Equal(createControl, tabControl.IsHandleCreated);
         }
 
+        [WinFormsFact]
+        public void TabAccessibilityObject_IsDisconnected_WhenTabPageReleasesUiaProvider()
+        {
+            using TabPage tabPage = new();
+            EnforceTabAccessibilityObjectCreation(tabPage);
+
+            tabPage.ReleaseUiaProvider(tabPage.Handle);
+
+            Assert.Null(tabPage.TestAccessor().Dynamic._tabAccessibilityObject);
+            Assert.True(tabPage.IsHandleCreated);
+
+            static void EnforceTabAccessibilityObjectCreation(TabPage tabPage)
+            {
+                _ = tabPage.TabAccessibilityObject;
+                Assert.NotNull(tabPage.TestAccessor().Dynamic._tabAccessibilityObject);
+            }
+        }
+
         [WinFormsTheory]
         [InlineData(null, "")]
         [InlineData("", "")]
@@ -1046,10 +1064,10 @@ namespace System.Windows.Forms.Tests
 
             Assert.NotNull(accessibleObject1.RuntimeId);
             Assert.Equal(tabControl.HandleInternal, (IntPtr)accessibleObject1.RuntimeId[1]);
-            Assert.Equal(accessibleObject1.GetChildId(), accessibleObject1.RuntimeId[2]);
+            Assert.Equal(accessibleObject1.GetHashCode(), accessibleObject1.RuntimeId[2]);
             Assert.NotNull(accessibleObject2.RuntimeId);
             Assert.Equal(tabControl.HandleInternal, (IntPtr)accessibleObject2.RuntimeId[1]);
-            Assert.Equal(accessibleObject2.GetChildId(), accessibleObject2.RuntimeId[2]);
+            Assert.Equal(accessibleObject2.GetHashCode(), accessibleObject2.RuntimeId[2]);
             Assert.Equal(createControl, pages[0].IsHandleCreated);
             Assert.Equal(createControl, pages[1].IsHandleCreated);
             Assert.Equal(createControl, tabControl.IsHandleCreated);


src/System.Windows.Forms/src/System/Windows/Forms/TabPage.cs
```diff
@@ -622,6 +623,21 @@ protected override void OnPaintBackground(PaintEventArgs e)
             }
         }
 
+        internal override void ReleaseUiaProvider(IntPtr handle)
+        {
+            if (OsVersion.IsWindows8OrGreater)
+            {
+                HRESULT result;
```
dreddy-work, line 630: declare inline.

dkazennov:
> Add unit test for the scenario?

As far as I know we don't create unit tests for memory leaks. Memory leaks are being tested manually and with CTI.

dkazennov:
1. Changed TabPageCollection.Remove() -- added call for removed TabPage.ReleaseUiaProvider
2. !gcroot showed AO object to be hold back by TabPage.TabAccessibleObject
3. Changed TabAccessibleObject.RuntimeId to include HashCode, not GetChildId (the index in the collection)
4. TabPage.TabAccessibleObject is fixed, yet AO object are now holded by ButtonAccessibleObject

dkazennov:
The source of the issue discovered by CTI turned out to be hierarchy on nested TabControls, TabPages and elements.
To reproduce it remove the tabPage with nested tabPages.
image

The memory leaks are fixed.
image

dkazennov:
Issues found by CTI are fixed.
CTI re-test will be requested.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.TabPageCollection.cs
```diff
@@ -335,11 +336,35 @@ public IEnumerator GetEnumerator()
...
+            {
+                foreach (Control control in controls)
+                {
+                    Debug.WriteLine($"TabPageItem {control.Name} type of {control.GetType()} (has any Controls? {control.Controls.Count > 0})");
+
+                    if (OsVersion.IsWindows8OrGreater && control.IsAccessibilityObjectCreated)
+                    {
+                        if (control.Controls.Count > 0)
+                        {
+                            ReleaseChildControls(control.Controls);;
```
RussKie, line 349:
```suggestion
                            ReleaseChildControls(control.Controls);
```

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.TabPageCollection.cs
```diff
@@ -335,11 +336,35 @@ public IEnumerator GetEnumerator()
                 return tabPages.GetEnumerator();
             }
 
+            private static void ReleaseChildControls(Control.ControlCollection controls)
+            {
+                foreach (Control control in controls)
+                {
+                    Debug.WriteLine($"TabPageItem {control.Name} type of {control.GetType()} (has any Controls? {control.Controls.Count > 0})");
+
+                    if (OsVersion.IsWindows8OrGreater && control.IsAccessibilityObjectCreated)
```
RussKie, line 345: Invert to reduce nesting. I'd also do this check before the Debug.Writeline

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.TabPageCollection.cs
```diff
@@ -335,11 +336,35 @@ public IEnumerator GetEnumerator()
                 return tabPages.GetEnumerator();
             }
 
+            private static void ReleaseChildControls(Control.ControlCollection controls)
```
RussKie, line 339: If this is the only called from `ReleaseTabPageUiaProviders` then it's worth considering turning this method into a local method.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.TabPageCollection.cs
```diff
@@ -335,11 +336,35 @@ public IEnumerator GetEnumerator()
...
+                        {
+                            ReleaseChildControls(control.Controls);;
+                        }
+
+                        control.ReleaseUiaProvider(control.Handle);
+                    }
+                }
+            }
+
+            private static void ReleaseTabPageUiaProviders(TabPage tabPage)
```
RussKie, line 357: Looks like this method is only called from `Remove`. I don't think there's much value breaking the functionality across multiple methods. Whilst having one large method that does many different things is bad, there's a cost to having multiple methods too. In this case, these costs aren't necessary.
I suggest considering something like:
```cs
            public void Remove(TabPage value)
            {
                ArgumentNullException.ThrowIfNull(value);

                _owner.Controls.Remove(value);

                ReleaseTabPageUiaProviders(value);

                return;

                static void ReleaseTabPageUiaProviders(TabPage tabPage)
                {
                    // all release-related functionality
                }
            }
```

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.TabPageCollection.cs
dkazennov: I believe we can delete Debug.WriteLine now.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.TabPageCollection.cs
dkazennov: Yes, I was thinking that the ReleaseTabPageUiaProviders() method may be redundant as a separate method.

dkazennov:
Fixed failures with unit tests.

dkazennov:
image

dkazennov:
The fix now works both with Narrator and Inspect. Requesting re-test for CTI.

dkazennov:
It seems that some builds are affected by https://github.com/dotnet/winforms/issues/6739

dkazennov:
Relocated static void ReleaseChildrenUiaProviders() from TabCollection.Remove() to TabPage.ReleaseUiaProvider()

dkazennov:
It is hard to pinpoint remaining memory leakages. Most tests shows no leaks.
image

dkazennov:
After consulting with @vladimir-krestov I believe this PR is worthy to be merged into main. I believe I have successfully fixed most memory leaks, including most cases where a TabPage being removed - both with Inspect and Narrator.

There are however some odd and rare case where an empty TabPage is being removed and with the Narrator there are memory leaks (as shown by CTI results). 

@Tanya-Solyanik could we implement what is already achieved by this fix and leave the remaining rare leak case for now? It seems it will take too much time and effort to find where is the exact source of the remaining issue (is possible).

RussKie (dismissed):

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.TabPageCollection.cs
```diff
@@ -338,8 +338,11 @@ public IEnumerator GetEnumerator()
             public void Remove(TabPage value)
             {
                 ArgumentNullException.ThrowIfNull(value);
-
                 _owner.Controls.Remove(value);
+                if (value.IsHandleCreated)
+                {
+                    value.ReleaseUiaProvider(value.Handle);
```
vladimir-krestov, line 344: I'm not sure, that it's right. Control calls ReleaseUiaProvider when destroying. In this case we will call this method twice, that may lead to some unexpected errors, mentioned by @Tanya-Solyanik before, so we should avoid double releasing.
First, please make sure, that a TabPage destroys after removing from the collection or after destroying of the whole TabControl. If it's true, we must not call this method here.

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.TabAccessibleObject.cs
```diff
@@ -73,7 +73,7 @@ internal override int[] RuntimeId
                     OwningTabControl is null
                         ? PARAM.ToInt(IntPtr.Zero)
                         : PARAM.ToInt(OwningTabControl.InternalHandle),
-                    GetChildId()
+                    GetHashCode()
```
vladimir-krestov, line 76: Please describe why did you changed it. It will be helpful in the future to understand reasons of this change and prevent changing back by someone.

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.cs
```diff
@@ -622,6 +623,44 @@ protected override void OnPaintBackground(PaintEventArgs e)
...
+                HRESULT result = UiaCore.UiaDisconnectProvider(TabAccessibilityObject);
+                Debug.Assert(result == HRESULT.S_OK);
+
+                result = UiaCore.UiaDisconnectProvider(AccessibilityObject);
+                Debug.Assert(result == HRESULT.S_OK);
+            }
+
+            base.ReleaseUiaProvider(handle);
+
+            return;
```
vladimir-krestov, line 642: You don't need this

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.cs
```diff
@@ -622,6 +623,44 @@ protected override void OnPaintBackground(PaintEventArgs e)
...
+
+                UiaCore.UiaReturnRawElementProvider(Handle, wParam: IntPtr.Zero, lParam: IntPtr.Zero, el: null);
+                HRESULT result = UiaCore.UiaDisconnectProvider(TabAccessibilityObject);
+                Debug.Assert(result == HRESULT.S_OK);
+
+                result = UiaCore.UiaDisconnectProvider(AccessibilityObject);
+                Debug.Assert(result == HRESULT.S_OK);
+            }
+
+            base.ReleaseUiaProvider(handle);
```
vladimir-krestov, line 640: I usually call it first to release the main object and then its elements

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.cs
```diff
@@ -622,6 +623,44 @@ protected override void OnPaintBackground(PaintEventArgs e)
...
+            if (OsVersion.IsWindows8OrGreater)
+            {
+                ReleaseChildrenUiaProviders(Controls);
+
+                UiaCore.UiaReturnRawElementProvider(Handle, wParam: IntPtr.Zero, lParam: IntPtr.Zero, el: null);
+                HRESULT result = UiaCore.UiaDisconnectProvider(TabAccessibilityObject);
+                Debug.Assert(result == HRESULT.S_OK);
+
+                result = UiaCore.UiaDisconnectProvider(AccessibilityObject);
+                Debug.Assert(result == HRESULT.S_OK);
```
vladimir-krestov, line 637: You should remove this, because the base implementation do it. And we must avoid double releasing of AO

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.cs
```diff
@@ -622,6 +623,44 @@ protected override void OnPaintBackground(PaintEventArgs e)
             }
         }
 
+        internal override void ReleaseUiaProvider(IntPtr handle)
+        {
+            if (OsVersion.IsWindows8OrGreater)
+            {
+                ReleaseChildrenUiaProviders(Controls);
+
+                UiaCore.UiaReturnRawElementProvider(Handle, wParam: IntPtr.Zero, lParam: IntPtr.Zero, el: null);
```
vladimir-krestov, line 632: `Handle` getter forces its value creation. We had some unexpected important bugs with it, this getter may create a Handle in incorrect time, or recreate a control after destroying. It's FYI in the future. But in this case you should remove it, because the base implementation do it well.

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.cs
```diff
@@ -622,6 +623,44 @@ protected override void OnPaintBackground(PaintEventArgs e)
             }
         }
 
+        internal override void ReleaseUiaProvider(IntPtr handle)
+        {
+            if (OsVersion.IsWindows8OrGreater)
```
vladimir-krestov, line 628: Please add IsAOCreated check as well

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.cs
```diff
@@ -622,6 +623,44 @@ protected override void OnPaintBackground(PaintEventArgs e)
...
+
+                result = UiaCore.UiaDisconnectProvider(AccessibilityObject);
+                Debug.Assert(result == HRESULT.S_OK);
+            }
+
+            base.ReleaseUiaProvider(handle);
+
+            return;
+
+            static void ReleaseChildrenUiaProviders(ControlCollection controls)
```
vladimir-krestov, line 644: As I understand, it's really extra. Because as I described above, every control care about its own AO and releasing when destroying. So every child control will call `ReleaseUiaProvider` independently when destroying, we shouldn't force this process and make double releasing, that will lead to unexpected errors. Every control will be destroyed anyway, so they will release their AO themself.
Please:
1) Make sure, that ReleaseUiaProvider is called for them without your changes after destroying of TabControl. (see Control.cs line 12212)
2) Remove this method at all

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.cs
```diff
@@ -622,6 +623,44 @@ protected override void OnPaintBackground(PaintEventArgs e)
...
         }
 
+        internal override void ReleaseUiaProvider(IntPtr handle)
+        {
+            if (OsVersion.IsWindows8OrGreater)
+            {
+                ReleaseChildrenUiaProviders(Controls);
+
+                UiaCore.UiaReturnRawElementProvider(Handle, wParam: IntPtr.Zero, lParam: IntPtr.Zero, el: null);
+                HRESULT result = UiaCore.UiaDisconnectProvider(TabAccessibilityObject);
```
vladimir-krestov, line 633: I see, that it's only one thing that should be implemented here

vladimir-krestov (changes requested):
It should be reworked and retested.
@Tanya-Solyanik, please describe, why should we avoid double releasing of controls?

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.cs
dkazennov: Yes, indeed. So we need to disconnect only TabAccessibilityObject.

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.cs
dkazennov: Removed UiaReturnRawElementProvider.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.TabPageCollection.cs
dkazennov: I believed the issue found by CTI team was because after removing from collection objects still remained in memory and was not properly destroyed. I'll check if objects are destroyed then.

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.cs
dkazennov: image

This way we don't have memory leaks if no TabPage is removed from TabPageCollection. But if TabPage is being removed from TabPageCollection there will be leaks. So I suspect there is no Destroy() for TabPage after it is being removed from collection.

dkazennov: So tests show that there will be memory leaks if TabPage is being removed.

I believe we can safely call Dispose() for a TabPage being removed from TabPageCollection as suggested in https://docs.microsoft.com/en-us/dotnet/desktop/winforms/controls/how-to-add-to-or-remove-from-a-collection-of-controls-at-run-time?view=netframeworkdesktop-4.8&viewFallbackFrom=netdesktop-6.0

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.TabAccessibleObject.cs
dkazennov: If RuntimeId is depended on index in collection, removing or adding elements to collection will affect RuntimeId which will cause objects to not properly disconnected.

dkazennov:
**Conclusion:**

1. On the last meeting we have decided that the most safe way to prevent a memory leak if removing the TabPage is to call ReleaseUiaProvider() from TabPageCollection.Remove() for the removed TabPage and all its controls. Calling for Dispose() leads to undesirable risks. @Tanya-Solyanik 

2. @vladimir-krestov If a TabPage is being removed and right after that added (I've put into test code Add(value) right after Remove(value)) than it will be added with all its elements intact and functional. No memory leaks will be spotted.

image

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.TabAccessibleObject.cs
Tanya-Solyanik: Hash code makes sense to me.

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.cs
Tanya-Solyanik: UiaReturnRawElementProvider(all nulls) is required. If control invokes this API with an actual provider in WM_GETOBJECT handler, it should clean up afterwards with a call with all null values for the same handle.
https://docs.microsoft.com/en-us/windows/win32/api/uiautomationcoreapi/nf-uiautomationcoreapi-uiareturnrawelementprovider#remarks

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.cs
Tanya-Solyanik: This sounds like a reasonable thing to try.

Tanya-Solyanik:
>@Tanya-Solyanik, please describe, why should we avoid double releasing of controls?

Because the second `Disconnect` crashes the application. I think there is more to these crashes than disconnecting twice, but disconnecting twice is contributing.

dkazennov:
@Tanya-Solyanik this issue is ready for review.
We have the final way to fix these leaks.
All cases of memory leaks are covered except for one rare and odd case found by CTI with empty TabPage. Vladimir proposed we implement everything that is already achieved.

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.cs
dkazennov: Both ways (calling for Dispose() in Remove() and calling for disconnection in Remove()) work.
As I understood the second way is more safe.

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.cs
dkazennov: @Tanya-Solyanik in current fix there is a override ReleaseUiaProvider class which calls for base.ReleaseUiaProvider(handle);
There is already the call for UiaCore.UiaReturnRawElementProvider(handle, 0, 0, null); in Control.ReleaseUiaProvider(handle);

Should there be a second call for UiaReturnRawElementProvider in override class of child class?

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.TabPageCollection.cs
Tanya-Solyanik: This is strange because `TabPage` is a Control, thus is backed up by a window handle. And all window handles receive WM_DESTROY, and Control class calls `ReleaseUiaProvider` when handling this message. One possibility is that WM_DESTROY arrived after the native window is un-subclassed, and NativeWindow.cs WinProc receives and handles WM_DESTROY. In this case we might want to do the same trick and combo box's sub-controls do. We create custom Native Windows subclasses for them and have a chance to handle WM_DESTROY in these subclasses. You can use spy++ (distributed with the VS) to double check if this message is sent.

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.cs
```diff
@@ -622,6 +623,18 @@ protected override void OnPaintBackground(PaintEventArgs e)
...
 
+        internal override void ReleaseUiaProvider(IntPtr handle)
+        {
+            base.ReleaseUiaProvider(handle);
+
+            if (_tabAccessibilityObject is not null && OsVersion.IsWindows8OrGreater)
+            {
+                // base.ReleaseUiaProvider only takes care about TabPage.AccessibilityObject type of TabPageAccessibilityObject.
+                HRESULT result = UiaCore.UiaDisconnectProvider(_tabAccessibilityObject);
+                Debug.Assert(result == HRESULT.S_OK);
```
Tanya-Solyanik, line 634: Please set `_tabAccessibilityObject ` to null after it is disconnected.

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.cs
Tanya-Solyanik: No, a single call to `UiaReturnRawElementProvider` is sufficient, I misunderstood Vladimir's comment.

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.cs
Tanya-Solyanik: Yes, Remove is safer because additional Dispose could break code that moves TabPages between different TabControls.

Tanya-Solyanik (commented):
Please understand lifetime of Tab windows in order to find a good spot to Disconnect providers and address all Vladimir's feedback. Also other similar PRs are adding unit tests please see if you can add similar ones.

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.cs
dkazennov: I added the check for value.IsHandleCreated because there are unit tests that check the TabPage class without the handler. 
In all actual cases handle is indeed created, but in unit tests cases the handle should not be created or unit tests will fail.

dkazennov:
> Please understand lifetime of TabPage windows in order to find a good spot to Disconnect providers and address all Vladimir's feedback. Also other similar PRs are adding unit tests please see if you can add similar ones.

As I understand the TabPage in most cases will end it's life if it is removed. But as Vladimir requested I have tested the case when the TabPage is added after it was being removed. That way the TabPage is being added with all it's contents intact, and the accessible objects are being created once more.

I'll try to figure it out if it is possible to test memory leaks with unit tests. I thought there is no way to test a leak with a unit test since we detect leaks with WinDbg?

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.cs
```diff
@@ -622,6 +623,22 @@ protected override void OnPaintBackground(PaintEventArgs e)
...
+                if (result == HRESULT.S_OK)
```
Tanya-Solyanik, line 635: It actually fails to disconnect when there is something wrong with the accessible object, so we don't want to keep a "bad" object around. I would set AO to null regardless of the return code.

Tanya-Solyanik: I guess that the reason why we have to disconnect TabAO manually is that Tab is a fully native window, with no backing managed control, and thus there is no place to put `SupportsUIAProviders == true` and activate release on the Control level.

Tanya-Solyanik (commented):
Remaining issues in this PR:
1. remove `result == S_OK` check
2. add unit tests
@vladimir-krestov  - please re-review, you are blocking this change right now.

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.cs
dkazennov: > It actually fails to disconnect when there is something wrong with the accessible object, so we don't want to keep a "bad" object around. I would set AO to null regardless of the return code.

Oh, I see. Ok.

I have finished the unit test to check the diconnection of _tabAccessibilityObject. But right now the unit test to control the Remove() method in TabPageCollection returns unexpected results.

dkazennov:
I've added a new test TabAccessibilityObject_IsDisconnected_WhenTabPageReleasesUiaProvider which successfully covers the case with _tabAccessibilityObject is being disconnected.

But there is probably no way to cover a case with disconnection of removed from TabControl.TabPageCollection items. While manual testing and CTI show that removed items are being disconnected the call for TabPageCollection .Remove() in a unit test fails to disconnect accessible objects.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.TabPageCollection.cs
Tanya-Solyanik: One way to avoid a double disconnect of provider here is to set Accessible object to null(by calling Properties.SetObject) in Control.cs in ReleaseUiaProvider

Tanya-Solyanik (commented):
Unit test looks good, but please see the comment about the `Remove` method

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.TabPageCollection.cs
```diff
@@ -338,8 +338,32 @@ public IEnumerator GetEnumerator()
...
+                        if (control.IsAccessibilityObjectCreated && OsVersion.IsWindows8OrGreater)
+                        {
+                            if (control.Controls.Count > 0)
+                            {
+                                ReleaseChildrenUiaProviders(control.Controls);
+                            }
+
+                            if (control.IsHandleCreated)
+                            {
+                                control.ReleaseUiaProvider(control.Handle);
```
Tanya-Solyanik, line 362: nit: use `HandleInternal`

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.TabPageCollection.cs
```diff
@@ -338,8 +338,32 @@ public IEnumerator GetEnumerator()
...
+                    foreach (Control control in controls)
+                    {
+                        if (control.IsAccessibilityObjectCreated && OsVersion.IsWindows8OrGreater)
+                        {
+                            if (control.Controls.Count > 0)
+                            {
+                                ReleaseChildrenUiaProviders(control.Controls);
+                            }
+
+                            if (control.IsHandleCreated)
```
Tanya-Solyanik, line 360: please test if this control supports UIA Providers

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.TabPageCollection.cs
Tanya-Solyanik, line 362: `control.ReleaseUiaProvider(control.Handle);` only works correctly if Control.cs sets Accessibility object to null (via Properties.SetObject) This is called out in @vladimir-krestov refactoring PR. Please coordinate who implements this cleanup.

Tanya-Solyanik (commented):
Looks good with minor comments. Should it be re-tested by the CTI team?

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.TabPageCollection.cs
dkazennov: TabPage have 
**ProviderDescription:**
"[pid:13080,providerId:0x10BBA 
Main:Nested [pid:26208,providerId:0x10BBA 
Main(parent link):Unidentified Provider (unmanaged:coreclr.dll)]; 
Hwnd(parent link):Microsoft: HWND Proxy (unmanaged:uiautomationcore.dll)]"

By comparing with the https://github.com/dotnet/winforms/pull/7314 (where I added the UIA Provider support) I can conclude that TabPage Control supports UIA Providers.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.TabPageCollection.cs
dkazennov: This is about https://github.com/dotnet/winforms/pull/7362/files#r939832225, as I understand.
We might need to test this whole issue in separate branch and possibly create a new Issue. This new Issue could probably lead to a separate big PR with huge amount on testing for side effects.

dkazennov:
Probably we don't new CTI since the last one? We haven't changed the fix here since then, only added a unit test and refactored the fix.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.TabPageCollection.cs
Tanya-Solyanik: Please discuss this change with Vladimir, I agree that a separate PR is apropriate.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.TabPageCollection.cs
```diff
@@ -338,8 +338,32 @@ public IEnumerator GetEnumerator()
             public void Remove(TabPage value)
             {
                 ArgumentNullException.ThrowIfNull(value);
-
                 _owner.Controls.Remove(value);
+                if (value.IsHandleCreated)
+                {
+                    value.ReleaseUiaProvider(value.Handle);
```
Tanya-Solyanik, line 344: Please use `HandleInternal`

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.TabPageCollection.cs
Tanya-Solyanik, line 360: I think if control does not have Handle created, there is no reason to iterate its children.

Tanya-Solyanik: I'm not talking about the tabpage control, but about children of this tabpage - i.e. controls placed on the tab page, like buttons. I'm suggesting this:
```
       static void ReleaseChildrenUiaProviders(Control.ControlCollection controls)
        {
            foreach (Control control in controls)
            {
                if (control.IsAccessibilityObjectCreated && control.IsHandleCreated && OsVersion.IsWindows8OrGreater)
                {
                    var children = control.Controls;
                    if (children.Count > 0)
                    {
                        ReleaseChildrenUiaProviders(children);
                    }

                    if (control.SupportsUiaProviders)
                    {
                        control.ReleaseUiaProvider(control.HandleInternal);
                    }
                }
            }

        }
```

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.TabPageCollection.cs
dkazennov: We have decided I should create a separate Issue and it will be fixed in the separate PR.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.TabPageCollection.cs
dkazennov: > I'm not talking about the tabpage control, but about children of this tabpage - i.e. controls placed on the tab page, like buttons. I'm suggesting this:
> 
> ```
>        static void ReleaseChildrenUiaProviders(Control.ControlCollection controls)
>         {
>             foreach (Control control in controls)
>             {
>                 if (control.IsAccessibilityObjectCreated && control.IsHandleCreated && OsVersion.IsWindows8OrGreater)
>                 {
>                     var children = control.Controls;
>                     if (children.Count > 0)
>                     {
>                         ReleaseChildrenUiaProviders(children);
>                     }
> 
>                     if (control.SupportsUiaProviders)
>                     {
>                         control.ReleaseUiaProvider(control.HandleInternal);
>                     }
>                 }
>             }
> 
>         }
> ```

Oh, I got it, thanks. This looks safer.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.TabPageCollection.cs
```diff
@@ -338,8 +338,34 @@ public IEnumerator GetEnumerator()
             public void Remove(TabPage value)
             {
                 ArgumentNullException.ThrowIfNull(value);
-
```
Tanya-Solyanik, line 341: nit: this repo is moving towards the style where argument validation is separated from the rest of the code by a blank line.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.TabPageCollection.cs
```diff
@@ -338,8 +338,34 @@ public IEnumerator GetEnumerator()
...
+                    value.ReleaseUiaProvider(value.HandleInternal);
+                }
+
+                ReleaseChildrenUiaProviders(value.Controls);
+
+                static void ReleaseChildrenUiaProviders(Control.ControlCollection controls)
+                {
+                    foreach (Control control in controls)
+                    {
+                        if (control.IsAccessibilityObjectCreated && control.IsHandleCreated && OsVersion.IsWindows8OrGreater)
```
Tanya-Solyanik, line 353: nit: If handle is not created, then the AO is unlikely to exist as well. So it might be better to test for handle first.

Tanya-Solyanik (approved):
fill free to apply my comments in the next PR.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.TabPageCollection.cs
```diff
@@ -338,8 +338,34 @@ public IEnumerator GetEnumerator()
...
+                            Control.ControlCollection childrenControls = control.Controls;
+
+                            if (childrenControls.Count > 0)
+                            {
+                                ReleaseChildrenUiaProviders(childrenControls);
+                            }
+
+                            if (control.SupportsUiaProviders)
+                            {
+                                control.ReleaseUiaProvider(control.HandleInternal);
```
vladimir-krestov, line 364: I'm not still agree with it. Let me double check
@Tanya-Solyanik, I think inner controls should care about disconnection themself. Right?

vladimir-krestov: Yes, we call `ReleaseUiaProvider` for an inner control twice. So it is incorrect

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.TabPageCollection.cs
```diff
@@ -338,8 +338,34 @@ public IEnumerator GetEnumerator()
...
-
                 _owner.Controls.Remove(value);
+                if (value.IsHandleCreated)
+                {
+                    value.ReleaseUiaProvider(value.HandleInternal);
+                }
+
+                ReleaseChildrenUiaProviders(value.Controls);
+
+                static void ReleaseChildrenUiaProviders(Control.ControlCollection controls)
```
vladimir-krestov, line 349: Actually, if to remove all of this, I don't see leak in case of removing a TabPage form a TabControl. @dkazennov, please describe a case that we need it in.

vladimir-krestov (approved):
I don't like double disconnections, but it should be tested more. Approved the rest ✅

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.TabPageCollection.cs
Tanya-Solyanik: Good point. This will double-disconnect when tage is removed from the control and then its child controls are disposed. @dkazennov - please test if tabpage leaks even after all child controls are disposed bythe application.

vladimir-krestov: Maybe it's try to release TabPage after deletion. But the main reason - `WmDestroy` is not called for deleted TabPage for some reason. I will try to rework it

vladimir-krestov: So `WmDestroy` is not called for a deleted TabPage, and for its Controls as well. I think the right way - destroy TabPage control after deletion, but it's so complex issue with many side effects

vladimir-krestov: Sorry, it was a mistake of testing. This implementation works well, because a deleted TabPage and its inner controls are not destroyed and disconnected, so in the case of deleting a TabPage there are no any double disconnections

Tanya-Solyanik (approved):

src/System.Windows.Forms/src/System/Windows/Forms/TabPage.cs
vladimir-krestov: @Tanya-Solyanik, this dialog.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.TabPageCollection.cs
dkazennov: > so in the case of deleting a TabPage there are no any double disconnections

So everything works as intended and this fix is merged, cool! Thanks!

PR #7458, opened 2022-07-20
Raising AutomationPropertyChanged event on TextChanged for ButtonBase a…

### LINKED ISSUE #7261: Some UIA PropertyChanged events are not being generated
### .NET version

.NET 6.0.5

No

No

### Issue description

In https://github.com/microsoft/accessibility-insights-windows/issues/1346, we received a user report that apps listening for UIA PropertyChanged events on WinForm labels were being notified of changes in the element's BoundingRectangle, but not of changes of the element's Name. At first we thought this might be a problem with Accessibility Insights, but upon further investigation we confirmed that we receive the expected events from WPF applications. There is a sample app in the AIWin bug, but I've also created a gist at https://gist.github.com/DaveTryon/44421194268fe2f5fc6624c13553f117. The test app just uses a timer to update a label on 2 second intervals. 

### Steps to reproduce

1. Clone the gist at https://gist.github.com/DaveTryon/44421194268fe2f5fc6624c13553f117
1. Build the application
1. Run application
1. Using [Application Insights for Windows](https://github.com/microsoft/accessibility-insights-windows/releases/latest), hover over the element that shows the update count
1. In the hierarchy tree, click on the "..." icon next to the element and choose "Listen to Events"
1. Expand the Properties control and check Name (alternatively, just check next to Properties to enable all properties)
1. Click the red dot to begin listening to events
1. Watch for the events--no name events are generated

The events are also not detected by AccEvent.exe. Events from a similar WPF app are detected by both Accessibility Insights and AccEvent. We suspect that this is an issue in the WinForms UIA provider. It may apply more broadly just just label controls, but we made no attempt to identify a full set of impacted control types.

NikitaSemenovAkvelon:
…nd Label

Fixes #7261

## Proposed changes

- Raised AutomationPropertyChanged event with NamePropertyId parameter in ButtonBase class.
- Raised AutomationPropertyChanged event with NamePropertyId parameter in Label class.
- Added unit tests.

## Customer Impact

- User can receives an event of NameProperty changes via AI.

## Regression? 

- No

## Risk

- Minimal

### Before

image

### After

image

## Test methodology 

- Manually tested the change in a winforms app.
- Unit tests.

## Accessibility testing  

- Used AI

 

## Test environment(s) 

- 7.0.100-preview.5.22307.18

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7458)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.cs
@@ -1078,6 +1078,11 @@ namespace System.Windows.Forms
                 base.OnTextChanged(e);
                 Invalidate();
             }
+
+            if (IsAccessibilityObjectCreated)
+            {
+                AccessibilityObject.RaiseAutomationPropertyChangedEvent(UiaCore.UIA.NamePropertyId, Text, Text);
+            }
         }
 
         /// <summary>
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ButtonBase.ButtonBaseAccessibleObjectTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ButtonBase.ButtonBaseAccessibleObjectTests.cs
@@ -246,6 +246,26 @@ namespace System.Windows.Forms.Tests
             Assert.False(ownerControl.IsHandleCreated);
         }
 
+        [WinFormsFact]
+        public void ButtonBaseAccessibleObject_TextChanged_AutomationPropertyChanged_Raised()
+        {
+            const string newText = "New text";
+            using var control = new ButtonWithCustomAccessibleObject(
+                (propertyId, value) => propertyId == UiaCore.UIA.NamePropertyId && ReferenceEquals(value, newText))
+            {
+                Text = "Text"
+            };
+
+            var accessibilityObject = control.AccessibilityObject as ControlAccessibleObjectWithNotificationCounter;
+            Assert.NotNull(accessibilityObject);
+            Assert.True(control.IsAccessibilityObjectCreated);
+            Assert.Equal(0, accessibilityObject.RaiseAutomationNotificationCallCount);
+
+            control.Text = newText;
+
+            Assert.Equal(1, accessibilityObject.RaiseAutomationNotificationCallCount);
+        }
+
         private class SubButtonBase : ButtonBase
         {
             public Action PerformClickAction { get; set; }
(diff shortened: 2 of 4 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.cs
```diff
@@ -1078,6 +1078,11 @@ protected override void OnTextChanged(EventArgs e)
                 base.OnTextChanged(e);
                 Invalidate();
             }
+
+            if (IsAccessibilityObjectCreated)
+            {
+                AccessibilityObject.RaiseAutomationPropertyChangedEvent(UiaCore.UIA.NamePropertyId, Text, Text);
```
Tanya-Solyanik, line 1084: Is there a way to raise this event from a place where both the old name and the new name are available?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ButtonBase.ButtonBaseAccessibleObjectTests.cs
```diff
@@ -256,5 +276,39 @@ private class SubButtonBase : ButtonBase
 
             public void PerformClick() => PerformClickAction();
         }
+
+        private class SubButtonWithCustomAccessibleObject : ButtonBase
```
Tanya-Solyanik, line 280: Consider name with no `sum` prefix - `ButtonWithCustomAccessibleObject`

src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.cs
NikitaSemenovAkvelon: It can be inside setter of the property "Text". Only in that place we contain both names. But in existed cases, RaiseAutomationPropertyChangedEvent with parameter NamePropertyId is raised without old values (e.g. https://github.com/dotnet/winforms/blob/8791f96949b30207bfaf34cc88512b24cad57057/src/System.Windows.Forms/src/System/Windows/Forms/RadioButton.cs#L403 and https://github.com/dotnet/winforms/blob/8791f96949b30207bfaf34cc88512b24cad57057/src/System.Windows.Forms/src/System/Windows/Forms/CheckBox.cs#L455). Also, I don't see previous values in the AI event listener.

Tanya-Solyanik: >Also, I don't see previous values in the AI event listener.
Then we don't need to report it!

Tanya-Solyanik (approved):
Thank you!

dreddy-work (approved):

PR #7467, opened 2022-07-20
Update Form's Minimum and Maximum sizes with DPI_CHANGED event

### LINKED ISSUE #7251: Changing MinimumSize of Form activates the form
### .NET version

6.0

No

_No response_

### Issue description

In Form.MinimumSize setter, `SetWindowPos `is called with `User32.SWP.NOZORDER` flag. I suppose `User32.SWP.NOZORDER | User32.SWP.NOACTIVATE` should be used, since there is no point in activating the form when MinimumSize changes

The problem I was facing was that MinimumSize and MaximumSize are not scaled when dpi changes (not sure why, in my opinion they should be), so I added scaling of them into `OnDpiChanged`. 
An now the magic happened: changing MinSize activates the form, which looks up the ActiveControl to set focus to. But dpi change arrives too early and the control is not visible yet, so ActiveControl remained null on nonprimary screen, but worked fine on the primary one :/

### Steps to reproduce

-

### LINKED ISSUE #7535: Runtime configuration support for Winforms in .NET
# Support for application runtime configuration in WinForms applications

## Overview
\
NET Winforms applications currently have [limited application configurations](https://aka.ms/applicationconfiguration) capabilities that are defined via MSBuils properties and are emitted into source code using source generators at compile time. This document outlines expansion of those application-wide configurations further to cover runtime config options for .NET Windows Forms applications.

## Runtime configuration in .NET Framework applications
\
.NET Framework Windows Forms applications use app.config to define runtime configurations and application-wide settings. Following are the various sections in the app.config that define application's runtime behavior.

### AppContext switches

These settings are used to opt-in or opt-out of a particular feature from WinForms runtime. Please refer to the [AppContext Switches](https://docs.microsoft.com/dotnet/framework/configure-apps/file-schema/runtime/appcontextswitchoverrides-element) documentation for more information

```XML
<configuration>
   <runtime>
      <AppContextSwitchOverrides value="Switch.System.Globalization.NoAsyncCurrentCulture=true" />
   </runtime>
</configuration>
```
### System.Windows.Forms.ApplicationConfigurationSection

This was introduced in .NET Framework 4.7 and is primarily used by Winforms runtime to enable high Dpi and other accessibility improvements. Please refer to the [ApplicationConfigurationSection](https://docs.microsoft.com/dotnet/framework/configure-apps/file-schema/winforms/windows-forms-add-configuration-element) documentation for more information.

```XML
<configuration>
  <System.Windows.Forms.ApplicationConfigurationSection>
  ...
  </System.Windows.Forms.ApplicationConfigurationSection>
</configuration>
```

### App settings from Settings designer/editor page

Unlike above, these settings are used by the user application. These are commonly defined via the Settings designer in Visual Studio, which in turn serialize them into the app.config file. Please refer to the [Application Settings](https://docs.microsoft.com/dotnet/desktop/winforms/advanced/using-application-settings-and-user-settings?view=netframeworkdesktop-4.8) documentation for more information.
```XML
    <userSettings>
        <WinFormsApp2.Properties.Settings>
            <setting name="Settingdfsd" serializeAs="String"> 
                <value>dfds</value>
            </setting>
        </WinFormsApp2.Properties.Settings>
    </userSettings>
    <applicationSettings>
        <WinFormsApp2.Properties.Settings>
            <setting name="dfsd" serializeAs="String">
                <value>sdfsdgs</value>
            </setting>
        </WinFormsApp2.Properties.Settings>
    </applicationSettings>
```

## Winforms runtime configuration in .NET applications
\
app.config has limited support in .NET due to performance and reliability reasons. .NET runtime and other .NET teams use runtimeconfig.json to define .NET runtime configurations and appsettings.json to define application-level settings. In this proposal, we are leveraging runtimeconfig.json to define WinForms runtime configurations.

While this proposal is focusing on providing an alternative solution for existing configuration sections `AppContextSwitchOverrides` and `System.Windows.Forms.ApplicationConfigurationSection` that are primarily used for specifying feature flags impacting winforms runtime behavior, we will be looking into alternatives for `Application Settings` that doesn't require app.config in the upcoming releases of .NET.

### Goals:

-   A replacement for `AppContextSwitchOverrides` and `System.Windows.Forms.ApplicationConfigurationSection` of app.config.
-   Users should be able to update/modify Winforms applications runtime configurations without recompiling the application.
-   The existing applications should be able to seamlessly upgrade to the new configuration model when targeting the latest .NET runtime.
-   The existing [application configuration MSBuild properties](https://aka.ms/applicationconfiguration) continue to work.

### Out of scope:

 App settings that are serialized by the Settings designer/editor page. Applications should continue to use the current model.
-  Dynamic/real-time (re-)loading of configuration values from runtimeconfig.json.
-  Unification of comple/runtime configurations into one place, runtimeconfig.json.

#### Syntax of runtimeConfig.template.Json.

```xml
{
  "configProperties": {
      "System.Globalization.UseNls": true,
      "System.Net.DisableIPv6": true,
      "System.GC.Concurrent": false,
      "System.Threading.ThreadPool.MinThreads": 4,
      "System.Threading.ThreadPool.MaxThreads": 25
    }
}
```

Windows Forms switches will be added to [`configProperties` section](https://docs.microsoft.com/dotnet/core/runtime-config) with the rest of the .NET switches. To avoid name conflics Windows Forms specific switches will be prefixed with `System.Windows.Forms`:

```xml
{
  "configProperties": {
      "System.Globalization.UseNls": true,
      "System.Net.DisableIPv6": true,
      "System.GC.Concurrent": false,
      "System.Threading.ThreadPool.MinThreads": 4,
      "System.Threading.ThreadPool.MaxThreads": 25,
      
      <!-- Windows Forms specific switches -->
      "System.Windows.Forms.ScaleTopLevelFormMinMaxSize": true,
      "System.Windows.Forms.<CustomEnumProperty>": "EnumValue"
    }
}
```

### Reading Winforms runtime configurations:
\
When a project is [built](https://docs.microsoft.com/en-us/dotnet/core/runtime-config/#runtimeconfigjson), an _[appname].runtimeconfig.json_ file is generated in the output directory. If a _runtimeconfig.template.json_ file exists in the same folder as the project file, any configuration options it contains are inserted into the _[appname].runtimeconfig.json_ file.
This proposal focuses primarily on enabling the runtime configurations for WinForms applications. A support for analyzers and source generators may be considered in the future .NET releases. We will revisit this implementation to improve user-experience further, as we make progress.
For example, the content of _[appname].runtimeconfig.json_ generated from above content:

```XML
{
  "runtimeOptions": {
    "tfm": "net7.0",
    "frameworks": [
      {
        "name": "Microsoft.NETCore.App",
        "version": "7.0.0-preview.7.22375.6"
      },
      {
        "name": "Microsoft.WindowsDesktop.App",
        "version": "7.0.0-preview.7.22377.1"
      }
    ],
    "configProperties": {
      "System.Globalization.UseNls": true,
      "System.Net.DisableIPv6": true,
      "System.GC.Concurrent": false,
      "System.Threading.ThreadPool.MinThreads": 4,
      "System.Threading.ThreadPool.MaxThreads": 25,
      "Switch.System.Windows.Forms.ScaleTopLevelFormMinMaxSize": true,
      "System.Windows.Forms.<StringProperty>": "string"
      "System.Windows.Forms.<CustomProperty>": "CustomValue"
    }
  }
}
```

The target framework information added to _[appname].runtimeconfig.json_ file always match with the application target framework irrespective of runtime/SDK installed on the machine or runtime used ([roll-forward scenarios](https://docs.microsoft.com/dotnet/core/versions/selection#framework-dependent-apps-roll-forward)) by the Windows Forms application.

### .NET runtime support in reading runtimeconfig.json:
\
WinForms is leveraging the [support](https://github.com/dotnet/runtime/blob/5098d45cc1bf9649fab5df21f227da4b80daa084/src/native/corehost/runtime_config.cpp) provided by .NET runtime in reading the _[appname].runtimeconfig.json_ file, and thus get all plumbing required for various hosting and environment scenarios. 
WinForms adds the following wrapper on top of .NET runtime implementation to improve the performance by caching configuration options, and also by defining custom defaults values for the WinForms specific feature flags. This wrapper is very similar to the [.NET runtime's implementation](https://github.com/dotnet/runtime/blob/04dac7b0fede29d44f896c5fd793754f83974175/src/libraries/System.Private.CoreLib/src/System/AppContextConfigHelper.cs).
The below example illustrates how to define `ScaleTopLevelFormMinMaxSizeForDpi` feature flag, opt-in by default, if an application is running on Windows 10 and is targeting .NET 8:

```cs
internal static partial class LocalAppContextSwitches
{
    private static readonly FrameworkName? s_targetFrameworkName = GetTargetFrameworkName();
    private static readonly bool s_isNetCoreApp = (s_targetFrameworkName?.Identifier) == ".NETCoreApp";

    private const string SwitchScaleTopLevelFormMinMaxSizeForDpi = "System.Windows.Forms.ScaleTopLevelFormMinMaxSizeForDpi";
    private static int s_scaleTopLevelFormMinMaxSize;

    public static bool ScaleTopLevelFormMinMaxSize
    {
        get => GetCachedSwitchValue(SwitchScaleTopLevelFormMinMaxSizeForDpi, ref s_scaleTopLevelFormMinMaxSize);
    }

    private static FrameworkName? GetTargetFrameworkName()
    {
        string? targetFrameworkName = AppContext.TargetFrameworkName;
        return targetFrameworkName is null ? null : new FrameworkName(targetFrameworkName);
    }

    private static bool GetCachedSwitchValue(string switchName, ref int cachedSwitchValue)
    {
        // The cached switch value has 3 states: 0 - unknown, 1 - true, -1 - false
        if (cachedSwitchValue < 0)
            return false;
        if (cachedSwitchValue > 0)
            return true;

        return GetSwitchValue(switchName, ref cachedSwitchValue);
    }

    private static bool GetSwitchValue(string switchName, ref int cachedSwitchValue)
    {
        bool hasSwitch = AppContext.TryGetSwitch(switchName, out bool isSwitchEnabled);
        if (!hasSwitch)
        {
            isSwitchEnabled = GetSwitchDefaultValue(switchName);
        }

        // Is caching switches disabled?.
        AppContext.TryGetSwitch("TestSwitch.LocalAppContext.DisableCaching", out bool disableCaching);
        if (!disableCaching)
        {
            cachedSwitchValue = isSwitchEnabled ? 1 /*true*/ : -1 /*false*/;
        }

        return isSwitchEnabled;
    }

    // Provides default values for switches based on OS and Targetframework versions.
    private static bool GetSwitchDefaultValue(string switchName)
    {
        if (!s_isNetCoreApp)
        {
            return false;
        }

        if (OsVersion.IsWindows10_1703OrGreater)
        {
            if (s_targetFrameworkName!.Version.CompareTo(new Version("8.0")) >= 0)
            {
                if (switchName == SwitchScaleTopLevelFormMinMaxSizeForDpi)
                {
                    return true;
                }
            }
        }

        return false;
    }
}
```

WinForms runtime then uses the static `LocalAppContextSwitches` class to access the runtime configurations. The below sample demonstrates how to access the feature switch `Switch.System.Windows.Forms.ScaleTopLevelFormMinMaxSize`:

Ex: Use of feature switch in ScaleContainerForDpi() method in ContainerControl.cs to scale Min/Max size of the Container.

```cs
internal void ScaleContainerForDpi(int deviceDpiNew, int deviceDpiOld, Rectangle suggestedRectangle)
{
    CommonProperties.xClearAllPreferredSizeCaches(this);
    SuspendAllLayout(this);
    try
    {
        if (LocalAppContextSwitches.ScaleTopLevelFormMinMaxSize)
        {
            // The suggested rectangle comes from Windows, and it does not match with our calculations for scaling controls by AutoscaleFactor.
            // Hence, we cannot use AutoscaleFactor here for scaling the control properties. See the below description for more details.
            float xScaleFactor = (float)suggestedRectangle.Width / Width;
            float yScaleFactor = (float)suggestedRectangle.Height / Height;
            ScaleMinMaxSize(xScaleFactor, yScaleFactor, updateFormSize: false);
        }

        // If this container is a top-level window, we would receive WM_DPICHANGED message that
        // has SuggestedRectangle for the control. We are forced to use this in such cases to
        // make the control placed in right location with respect to the new monitor that triggered
        // WM_DPICHANGED event. Failing to apply SuggestedRectangle will result in a circular WM_DPICHANGED
        // events on the control.

        // Note: SuggestedRectangle supplied  by WM_DPICHANGED event is Dpi (not Font) scaled. if top-level window is
        // Font scaled, we might see deviations in the expected bounds and may result in adding Scrollbars (horizontal/vertical)
        User32.SetWindowPos(
            new HandleRef(this, HandleInternal),
            User32.HWND_TOP,
            suggestedRectangle.X,
            suggestedRectangle.Y,
            suggestedRectangle.Width,
            suggestedRectangle.Height,
            User32.SWP.NOZORDER | User32.SWP.NOACTIVATE);

        // Bounds are already scaled for the top-level window. We would need to skip scaling of
        // this control further by the 'OnFontChanged' event.
        _isScaledByDpiChangedEvent = true;

        // Factor is used only to scale Font. After that AutoscaleFactor kicks in to scale controls.
        var factor = ((float)deviceDpiNew) / deviceDpiOld;

        // DpiFontscache is available only in PermonitorV2 mode applications.
        if (!TryGetDpiFont(deviceDpiNew, out Font? fontForDpi))
        {
            Font currentFont = Font;
            fontForDpi = currentFont.WithSize(currentFont.Size * factor);
            AddToDpiFonts(deviceDpiNew, fontForDpi);
        }

        ScaledControlFont = fontForDpi;
        if (IsFontSet())
        {
            SetScaledFont(fontForDpi);
        }
        else
        {
            using (new LayoutTransaction(ParentInternal, this, PropertyNames.Font))
            {
                OnFontChanged(EventArgs.Empty);
            }
        }
    }
    finally
    {
        // We want to perform layout for dpi-changed high Dpi improvements - setting the second parameter to 'true'
        ResumeAllLayout(this, true);
        _isScaledByDpiChangedEvent = false;
    }
}
```

OptinSwitch

dreddy-work:
**Description**: 

Currently, Form's `MinimumSize `and `MaximumSize `are not updated when a `DPI_CHNAGED `event received by the Form. This prevents scaling of Form to match with the Dpi when its minimum or maximum size properties are set. Users trying to work around this issue may see problems like #7251. 

Creating this draft PR to assess the risk for .NET 7.0.

Change in behavior:

Application running in PermonitorV2 mode will see following behavioral change when application moved from one monitor to the other that have different Dpi/Scaling settings and Form has either Minimum and/or Maximum size property set.
- Minimum and Maximum size of the Form will be changed and any runtime dependency on these might be impacted. 
- Raising `MinimumSizeChnaged  `and/or `MaximumSizeChnaged` events that were not before. 
- Scaled Form size now have new constraint values for Minimum and Maximum sizes. 

Before fix: ( moving from 100% to 300%)
GetImage (2)

GetImage (3)

After fix:
GetImage
GetImage (1)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7467)

fixes #7251 
fixes #7535


src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
```diff
@@ -1425,6 +1428,13 @@ internal void ScaleContainerForDpi(int deviceDpiNew, int deviceDpiOld, Rectangle
             SuspendAllLayout(this);
             try
             {
+
+                // Suggested rectangle comes from Windows and does not match with our calculations for scaling controls by Autoscale factor.
+                // Hence, we can not use AutoscaleFactor here for scaling control properties. Please below description for more details.
```
RussKie, line 1433: Looks like a bad grammar in "Please below description for more details.". I guess you may have missed a verb here.

src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
```diff
@@ -4901,6 +4908,27 @@ protected override void ScaleCore(float x, float y)
             }
         }
 
+        protected override void ScaleTopLevelWindowProperties(float widthFactor, float hightFactor)
```
RussKie, line 4911:
```suggestion
        protected override void ScaleTopLevelWindowProperties(float widthFactor, float heightFactor)
```
Though I'd suggest the following as we generally use `x` and `y`:
```suggestion
        protected override void ScaleTopLevelWindowProperties(float xScaleFactor, float yScaleFactor)
```

RussKie (commented):
LGTM

dreddy-work:
Tests succeed locally. Seems thread awareness is not being set on lab machines. Looking on how to confirm this theory.

docs/WinformsRuntimeConfiguration.md
```diff
@@ -0,0 +1,238 @@
...
+
+**Goals:**
+
+-   Replacement for `AppContextSwitchOverrides` and `System.Windows.Forms.ApplicationConfigurationSection` of app.config.
+
+-   Users should be able to update/modify Winforms applications runtime configurations without recompiling the application
+
+-   Existing applications should be able to seamlessly upgrade to this
+    new model when targeting to latest .NET.
+-   Existing [build time properties]((https://docs.microsoft.com/en-us/dotnet/desktop/winforms/whats-new/net60?view=netdesktop-6.0#new-application-bootstrap)) defined via project file continue to work.
```
RussKie, line 71:
```suggestion
-   A replacement for `AppContextSwitchOverrides` and `System.Windows.Forms.ApplicationConfigurationSection` of app.config.
-   Users should be able to update/modify Winforms applications runtime configurations without recompiling the application.
-   The existing applications should be able to seamlessly upgrade to the new configuration model when targeting the latest .NET runtiem.
-   The existing [application configuration MSBuild properties](https://aka.ms/applicationconfiguration) continue to work.
```

docs/WinformsRuntimeConfiguration.md
```diff
@@ -0,0 +1,238 @@
...
+
+**Out of scope:**
+
+-  App settings that are serialized from Settings designer/editor page. Applications should continue to use current model until we come up with an alternative solution for this.
+-  Dynamic/real-time loading of configuration values from runtimeconfig.json.
+-  Unification of Build/Runtime configurations into one place, runtimeconfig.json.
+    -   We keep current build time configurations in the project file while adding new runtime configurations into runtimeconfig.json. We may revisit build time properties for unification in the upcoming releases of .NET.
```
RussKie, line 78:
```suggestion
-  App settings that are serialized by the Settings designer/editor page. Applications should continue to use the current model.
-  Dynamic/real-time (re-)loading of configuration values from runtimeconfig.json.
-  Unification of comple/runtime configurations into one place, runtimeconfig.json.
```

The last item is redundant, it's called out as the last option under the Goals section.

docs/WinformsRuntimeConfiguration.md
```diff
@@ -0,0 +1,238 @@
...
+      "System.Globalization.UseNls": true,
+      "System.Net.DisableIPv6": true,
+      "System.GC.Concurrent": false,
+      "System.Threading.ThreadPool.MinThreads": 4,
+      "System.Threading.ThreadPool.MaxThreads": 25,
+      "System.Windows.Forms.ScaleTopLevelFormMinMaxSize": true,
+      "System.Windows.Forms.<CustomEnumProperty>": "EnumValue"
+    }
+}
+```
```
RussKie, line 112:
````suggestion
Windows Forms switches will be added to [`configProperties` section](https://docs.microsoft.com/dotnet/core/runtime-config) with the rest of the .NET switches. To avoid name conflics Windows Forms specific switches will be prefixed with `System.Windows.Forms`:
```xml
{
  "configProperties": {
      "System.Globalization.UseNls": true,
      "System.Net.DisableIPv6": true,
      "System.GC.Concurrent": false,
      "System.Threading.ThreadPool.MinThreads": 4,
      "System.Threading.ThreadPool.MaxThreads": 25,
      
      <!-- Windows Forms specific switches -->
      "System.Windows.Forms.ScaleTopLevelFormMinMaxSize": true,
      "System.Windows.Forms.<CustomEnumProperty>": "EnumValue"
    }
}
```
````

File: docs/WinformsRuntimeConfiguration.md [hunk 6]
```diff
@@ -0,0 +1,238 @@
...
+
+
+**Reading Winforms runtime configurations:**
+
+Build system [generate](https://docs.microsoft.com/en-us/dotnet/core/runtime-config) `<AppName>.runtimeconfig.json` file in the output directory with the content specified in the `runtimeconfig.template.json` file along with application target framework information. .NET runtime recognizes this file and get it loaded while running Winforms applications.
+
+In this proposal, we will be focusing primarily on enabling only runtime configurations for WinForms applications. Support for analyzers and source generators are considered in later .NET releases. We will revisit this implementation to improve user-experience further, as we make progress.
+
+Ex: Content of <AppName>.runtimeconfig.json generated from above content.
```
RussKie, line 121:
```suggestion
When a project is [built](https://docs.microsoft.com/en-us/dotnet/core/runtime-config/#runtimeconfigjson), an _[appname].runtimeconfig.json_ file is generated in the output directory. If a _runtimeconfig.template.json_ file exists in the same folder as the project file, any configuration options it contains are inserted into the _[appname].runtimeconfig.json_ file.

This proposal focuses primarily on enabling the runtime configurations for WinForms applications. A support for analyzers and source generators may be considered in the future .NET releases. We will revisit this implementation to improve user-experience further, as we make progress.

For example, the content of _[appname].runtimeconfig.json generated from above content:
```

RussKie, line 121: The official docs use `[appname].runtimeconfig.json` format, we should use the same for the consistency.

docs/WinformsRuntimeConfiguration.md
```diff
@@ -0,0 +1,238 @@
...
+      "System.Threading.ThreadPool.MaxThreads": 25,
+      "Switch.System.Windows.Forms.ScaleTopLevelFormMinMaxSize": true,
+      "System.Windows.Forms.<StringProperty>": "string"
+      "System.Windows.Forms.<CustomProperty>": "CustomValue"
+    }
+  }
+}
+```
+
+Target framework information added to <AppName>.runtimeconfig.json file always match with application target framework irrespective of runtime/SDK installed on the machine or runtime used (Roll-forward scenarios) by the WinForms application.
```
RussKie, line 151:
```suggestion
The target framework information added to <AppName>.runtimeconfig.json file always match with the application target framework irrespective of runtime/SDK installed on the machine or runtime used ([roll-forward scenarios](https://docs.microsoft.com/dotnet/core/versions/selection#framework-dependent-apps-roll-forward)) by the Windows Forms application.
```

docs/WinformsRuntimeConfiguration.md
```diff
@@ -0,0 +1,238 @@
...
+
+**.NET runtime support in reading runtimeconfig.json:**
+
+WinForms leveraging [the support](https://github.com/dotnet/runtime/blob/5098d45cc1bf9649fab5df21f227da4b80daa084/src/native/corehost/runtime_config.cpp) provided by .NET runtime in reading the `<AppName>.runtimeconfig.json` file and thus get all plumbing required for various hosting and environment scenarios. 
+
+WinForms add following wrapper on top of runtime implementation to improve performance by caching configuration options and also to define custom defaults values for the WinForms specific feature flags. This is very similar to the [implementation](https://github.com/dotnet/runtime/blob/04dac7b0fede29d44f896c5fd793754f83974175/src/libraries/System.Private.CoreLib/src/System/AppContextConfigHelper.cs) used in various .NET runtime components.
+
+Below example illustrate making the feature flag `ScaleTopLevelFormMinMaxSizeForDpi` default opt-in if application is running on Windows 10 and targeting .NET 8.
```
RussKie, line 159:
```suggestion
WinForms is leveraging the [support](https://github.com/dotnet/runtime/blob/5098d45cc1bf9649fab5df21f227da4b80daa084/src/native/corehost/runtime_config.cpp) provided by .NET runtime in reading the `<AppName>.runtimeconfig.json` file, and thus get all plumbing required for various hosting and environment scenarios. 

WinForms adds the following wrapper on top of .NET runtime implementation to improve the performance by caching configuration options, and also by defining custom defaults values for the WinForms specific feature flags. This wrapper is very similar to the [.NET runtime's implementation](https://github.com/dotnet/runtime/blob/04dac7b0fede29d44f896c5fd793754f83974175/src/libraries/System.Private.CoreLib/src/System/AppContextConfigHelper.cs).

The below example illustrates how to define `ScaleTopLevelFormMinMaxSizeForDpi` feature flag, opt-in by default, if an application is running on Windows 10 and is targeting .NET 8:
```

docs/WinformsRuntimeConfiguration.md
```diff
@@ -0,0 +1,238 @@
...
+}
+```
+
+Target framework information added to <AppName>.runtimeconfig.json file always match with application target framework irrespective of runtime/SDK installed on the machine or runtime used (Roll-forward scenarios) by the WinForms application.
+
+**.NET runtime support in reading runtimeconfig.json:**
+
+WinForms leveraging [the support](https://github.com/dotnet/runtime/blob/5098d45cc1bf9649fab5df21f227da4b80daa084/src/native/corehost/runtime_config.cpp) provided by .NET runtime in reading the `<AppName>.runtimeconfig.json` file and thus get all plumbing required for various hosting and environment scenarios. 
+
+WinForms add following wrapper on top of runtime implementation to improve performance by caching configuration options and also to define custom defaults values for the WinForms specific feature flags. This is very similar to the [implementation](https://github.com/dotnet/runtime/blob/04dac7b0fede29d44f896c5fd793754f83974175/src/libraries/System.Private.CoreLib/src/System/AppContextConfigHelper.cs) used in various .NET runtime components.
```
RussKie, line 157: I'm not sure how to read this correctly.

`WinForms adds... wrapper... _to improve_ performance by caching... _and also to define_ custom defaults...` - i.e., add a wrapper to improve and to define

- or -

`WinForms adds... wrapper... to improve performance _by caching_... _and also by defining_ custom defaults...` - i.e., add a wrapper to improve by caching and by defining.

docs/WinformsRuntimeConfiguration.md
```diff
@@ -0,0 +1,238 @@
...
+
+Below example illustrate making the feature flag `ScaleTopLevelFormMinMaxSizeForDpi` default opt-in if application is running on Windows 10 and targeting .NET 8.
+
+```cs
+ internal static partial class LocalAppContextSwitches
+    {
```
RussKie, line 163: Incorrect indentation

docs/WinformsRuntimeConfiguration.md
```diff
@@ -0,0 +1,238 @@
...
+        // Property to cache feature flag "Switch.System.Windows.Forms.ScaleTopLevelFormMinMaxSize" value. 
+        private static int s_scaleTopLevelFormMinMaxSize;
+        public static bool ScaleTopLevelFormMinMaxSize
```
RussKie, line 166: Missing blank line

docs/WinformsRuntimeConfiguration.md
```diff
@@ -0,0 +1,238 @@
...
+                    if (switchName == "Switch.System.Windows.Forms.ScaleTopLevelFormMinMaxSizeForDpi")
+                    {
+                        return true;
+                    }
+                }
+            }
+
+            return false;
+        }
+    }
```
RussKie, line 207: The sample code doesn't look valid (did you manually craft it and didn't update all references correctly?). E.g., `ScaleTopLevelFormMinMaxSize` refers to `GetCachedSwitchValue`, but no such method is present.

nit: "Switch.System.Windows.Forms.ScaleTopLevelFormMinMaxSizeForDpi" should be turned in to a const.

The implementation of `GetCachedSwitchValueInternal` also looks somewhat suboptimal - i.e., we read the switch and only then check if we need to return the cached value of not. I think that should be reversed, i.e. - if we need to return the cached value then there's no need to read the switch at all.

docs/WinformsRuntimeConfiguration.md
```diff
@@ -0,0 +1,238 @@
...
+```
+
+
+WinForms runtime then use the static `LocalAppContextSwitches` class to access runtime configurations. Below is sample accessing feature switch `Switch.System.Windows.Forms.ScaleTopLevelFormMinMaxSize`. 
+
+Ex: Use of feature switch in OnDpiChanged() method in Form.cs to scale Min/Max size of the Form.
+
+```cs
+ protected virtual void OnDpiChanged(DpiChangedEventArgs e)
+        {
```
RussKie, line 217: Incorrect indentation

docs/WinformsRuntimeConfiguration.md
```diff
@@ -0,0 +1,238 @@
...
+                }
+            }
+
+            return false;
+        }
+    }
+```
+
+
+WinForms runtime then use the static `LocalAppContextSwitches` class to access runtime configurations. Below is sample accessing feature switch `Switch.System.Windows.Forms.ScaleTopLevelFormMinMaxSize`.
```
RussKie, line 211:
```suggestion
WinForms runtime then uses the static `LocalAppContextSwitches` class to access the runtime configurations. The below sample demonstrates how to access the feature switch `Switch.System.Windows.Forms.ScaleTopLevelFormMinMaxSize`:
```

RussKie (dismissed):
Pass over the docs only. Will continue to the code later.

RussKie (commented):
The animation is great, but we'll need to redo it. I understand that it's showing two desktops, but it's too wide (a lot of our users still use HD monitors), and it contains information completely unrelated (i.e., the email).
The app could be tweaked to appear significantly closer to the monitor boundary, so that we can clip the view and dragging will require less distance. 
It could be also great to add some test to the animation to indicate each monitor's settings.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.TargetFramework.cs
```diff
@@ -0,0 +1,41 @@
...
+            {
+                if (targetFramework is null)
+                {
+                    return;
+                }
+
+                // Expected string is in the format of ".NETCoreApp,Version=v7.0"
+                var tokens = targetFramework.Split(',');
+
+                if(tokens.Length != 2)
```
RussKie, line 24: [nit] missing space

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -0,0 +1,66 @@
...
+namespace System.Windows.Forms.Primitives
+{
+    // Barrowed from https://github.com/dotnet/runtime/blob/main/src/libraries/Common/src/System/LocalAppContextSwitches.Common.cs
+    internal static partial class LocalAppContextSwitches
+    {
+        private static int s_scaleTopLevelFormMinMaxSize;
+        public static bool ScaleTopLevelFormMinMaxSize
+        {
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            get => GetCachedSwitchValue("Switch.System.Windows.Forms.ScaleTopLevelFormMinMaxSizeForDpi", ref s_scaleTopLevelFormMinMaxSize);
```
RussKie, line 16: ❗ As per the doc - the prefix is "System.Windows.Forms.". There's no need to create additional differentiators. 

Please define a const for `System.Windows.Forms.ScaleTopLevelFormMinMaxSizeForDpi` and use it throughout.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.TargetFramework.cs
```diff
@@ -0,0 +1,41 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Windows.Forms.Primitives
+{
+    internal static partial class LocalAppContextSwitches
+    {
+        private class TargetFramework
```
RussKie, line 9: This looks as though we're reimplementing [`FrameworkName`](https://docs.microsoft.com/dotnet/api/system.runtime.versioning.frameworkname), and it's implementation is optimised.
image

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -0,0 +1,66 @@
...
+
+            return isSwitchEnabled;
+        }
+
+        // Provides default values for switches based on OS and Targetframework versions.
+        private static bool GetSwitchDefaultValue(string switchName)
+        {
+            if (OsVersion.IsWindows10_1703OrGreater)
+            {
+                var tfm = new TargetFramework(AppContext.TargetFrameworkName);
```
RussKie, line 53: Should this be cached? The TFM can't change after the app has been launched, right?

src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
```diff
@@ -1170,6 +1171,9 @@ internal override void Scale(SizeF includedFactor, SizeF excludedFactor, Control
             }
         }
 
+        protected virtual void ScaleTopLevelWindowProperties(float xScaleFactor, float yScaleFactor, bool updateFormSize = true)
```
RussKie, line 1174: All public API must have xml-doc annotations.

src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
```diff
@@ -1423,6 +1427,15 @@ internal void ScaleContainerForDpi(int deviceDpiNew, int deviceDpiOld, Rectangle
             SuspendAllLayout(this);
             try
             {
+                if (LocalAppContextSwitches.ScaleTopLevelFormMinMaxSize)
+                {
+                    // Suggested rectangle comes from Windows operating system and does not match with our calculations for scaling controls by Autoscale factor.
+                    // Hence, we can not use AutoscaleFactor here for scaling control properties. Please see below description for more details.
```
RussKie, line 1433:
```suggestion
                    // The suggested rectangle comes from Windows, and it does not match with our calculations for scaling controls by AutoscaleFactor.
                    // Hence, we cannot use AutoscaleFactor here for scaling the control properties. See the below description for more details.
```

src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
```diff
@@ -1297,35 +1297,45 @@ public override Size MaximumSize
...
+            Properties.SetInteger(PropMaxTrackSizeWidth, value.Width);
+            Properties.SetInteger(PropMaxTrackSizeHeight, value.Height);
 
-                        if (Properties.GetInteger(PropMinTrackSizeHeight) > value.Height)
-                        {
-                            Properties.SetInteger(PropMinTrackSizeHeight, value.Height);
-                        }
-                    }
+            // Bump minimum size if necessary
+            //
```
RussKie, line 1311:
```suggestion
```

src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
```diff
@@ -1297,35 +1297,45 @@ public override Size MaximumSize
...
 
-                    // Bump minimum size if necessary
-                    //
-                    if (!MinimumSize.IsEmpty && !value.IsEmpty)
-                    {
-                        if (Properties.GetInteger(PropMinTrackSizeWidth) > value.Width)
-                        {
-                            Properties.SetInteger(PropMinTrackSizeWidth, value.Width);
-                        }
+        private void UpdateMaximumSize(Size value, bool UpdateFormSize = true)
```
RussKie, line 1305:
```suggestion
        private void UpdateMaximumSize(Size value, bool updateFormSize = true)
```

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -0,0 +1,66 @@
...
+        private static bool GetCachedSwitchValueInternal(string switchName, ref int cachedSwitchValue)
+        {
+            bool hasSwitch = AppContext.TryGetSwitch(switchName, out bool isSwitchEnabled);
+            if (!hasSwitch)
+            {
+                isSwitchEnabled = GetSwitchDefaultValue(switchName);
+            }
+
+            // Is caching switches disabled?.
+            AppContext.TryGetSwitch("TestSwitch.LocalAppContext.DisableCaching", out bool disableCaching);
```
RussKie, line 39: Is this switch our invention or is it used by .NET runtime? The doc must should describe this switch and explain how and when to use it.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -0,0 +1,66 @@
...
+            // The cached switch value has 3 states: 0 - unknown, 1 - true, -1 - false
+            if (cachedSwitchValue < 0)
+                return false;
+            if (cachedSwitchValue > 0)
+                return true;
+
+            return GetCachedSwitchValueInternal(switchName, ref cachedSwitchValue);
+        }
+
+        private static bool GetCachedSwitchValueInternal(string switchName, ref int cachedSwitchValue)
```
RussKie, line 30: I suggest turning this method into a local of `GetCachedSwitchValue` method because on its own this method raises perf-related questions. However, for a local method, when it's only called when a switch value hasn't been read, such questions don't arise.

src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
```diff
@@ -4901,6 +4908,27 @@ protected override void ScaleCore(float x, float y)
             }
         }
 
+        protected override void ScaleTopLevelWindowProperties(float xScaleFactor, float yScaleFactor, bool updateFormSize = true)
```
RussKie, line 4911: Ditto xml-docs missing

src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
```diff
@@ -1297,35 +1297,45 @@ public override Size MaximumSize
...
-                        Size = new Size(Math.Min(size.Width, value.Width), Math.Min(size.Height, value.Height));
-                    }
+                if (Properties.GetInteger(PropMinTrackSizeHeight) > value.Height)
+                {
+                    Properties.SetInteger(PropMinTrackSizeHeight, value.Height);
+                }
+            }
 
-                    OnMaximumSizeChanged(EventArgs.Empty);
+            // UpdateFormSize is set to `False' when updating Minimum/Maximum sizes as a result of DPI_CHANGED message.
```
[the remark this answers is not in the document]
RussKie: With the exception of five instances of bad capitalization we don't use `False` or `'False'` anywhere in the codebase, unless it's in a beginning of a sentence. There are few instances of `FALSE` but it generally is in a context of an interop. Everywhere else it's used as `false` or as `XYZ == false`.

<details>
<summary>Examples</summary>

image
image



</details>

docs/WinformsRuntimeConfiguration.md
dreddy-work: intended to be read as `i.e., add a wrapper to improve and to define`

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
dreddy-work: It is used by runtime.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.TargetFramework.cs
dreddy-work: On a second thought, I decided to go with existing implementation and doesn't introduce this private implementation.

docs/WinformsRuntimeConfiguration.md
```diff
@@ -0,0 +1,238 @@
+**Runtime configuration in legacy .NET framework applications**
```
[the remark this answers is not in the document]
dreddy-work: Lets chat offlien on this.

docs/WinformsRuntimeConfiguration.md
dreddy-work: right. it was altered later and did not update doc with latest source.

dreddy-work:
> The animation is great, but we'll need to redo it. I understand that it's showing two desktops, but it's too wide (a lot of our users still use HD monitors), and it contains information completely unrelated (i.e., the email). The app could be tweaked to appear significantly closer to the monitor boundary, so that we can clip the view and dragging will require less distance. It could be also great to add some test to the animation to indicate each monitor's settings.

I tried but now screen separation isn't clear with latest one. I am sure people reading will understand and demo was just to show how it is picked up. Not planning on spending more time on this for now.

src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
```diff
@@ -1170,6 +1171,15 @@ internal override void Scale(SizeF includedFactor, SizeF excludedFactor, Control
             }
         }
 
+        /// <summary>
+        /// Scales container's properties Min and Max size with the scale factor provided.
+        /// </summary>
+        /// <param name="xScaleFactor">Scale factor to be applied on width of the property being scaled</param>
+        /// <param name="yScaleFactor">Scale factor to be applied on height of the property being scaled</param>
+        /// <param name="updateContainerSize">Specifies if container size need to be updated along with properties being scaled</param>
```
RussKie, line 1179:
```suggestion
        /// <param name="xScaleFactor">The scale factor to be applied on width of the property being scaled.</param>
        /// <param name="yScaleFactor">The scale factor to be applied on height of the property being scaled.</param>
        /// <param name="updateContainerSize"><see langword="true"/> to resize of the container control along with properties being scaled; otherwise, <see langword="false"/>.</param>
```

I think there's a Roslyn-based doc analyzer that we haven't enabled yet to track the full stops :)

src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
```diff
@@ -1297,35 +1297,44 @@ public override Size MaximumSize
...
+                }
+            }
 
-                    OnMaximumSizeChanged(EventArgs.Empty);
+            // UpdateFormSize=false when Minimum/Maximum sizes get updated as a result of DPI_CHANGED message.
+            // DPI_CHANGED message updates the Form size with the SuggestedRectangle provided by Windows.
+            if (updateFormSize)
+            {
+                // Keep form size within new limits
+                //
```
RussKie, line 1329: [nit] this is unnecessary
```suggestion
```

src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
```diff
@@ -4901,6 +4908,31 @@ protected override void ScaleCore(float x, float y)
             }
         }
 
+        /// <summary>
+        /// Scales Form's properties Min and Max size with the scale factor provided.
+        /// </summary>
+        /// <param name="xScaleFactor">Scale factor to be applied on width of the property being scaled</param>
+        /// <param name="yScaleFactor">Scale factor to be applied on height of the property being scaled</param>
+        /// <param name="updateContainerSize">Specifies if Form size need to be updated along with properties being scaled</param>
```
RussKie, line 4916: The overrides inherit the doc from the base, so unless the docs need to differ from the base this is not necessary.
You can see if by removing this doc and hovering the mouse over - you should see the intellisense popup.

src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
```diff
@@ -4866,27 +4885,15 @@ protected override void ScaleCore(float x, float y)
             SuspendLayout();
             try
             {
+                //Get size values in advance to prevent one change from affecting another.
```
RussKie, line 4888:
```suggestion
                // Get size values in advance to prevent one change from affecting another.
```

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -0,0 +1,81 @@
...
+using System.Runtime.Versioning;
+
+namespace System.Windows.Forms.Primitives
+{
+    // Barrowed from https://github.com/dotnet/runtime/blob/main/src/libraries/Common/src/System/LocalAppContextSwitches.Common.cs
+    internal static partial class LocalAppContextSwitches
+    {
+        private const string SwitchScaleTopLevelFormMinMaxSizeForDpi = "System.Windows.Forms.ScaleTopLevelFormMinMaxSizeForDpi";
+        private static int s_scaleTopLevelFormMinMaxSize;
+        public static bool ScaleTopLevelFormMinMaxSize
```
RussKie, line 15: Missing blank line

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        private static bool GetSwitchValue(string switchName, ref int cachedSwitchValue)
+        {
+            bool hasSwitch = AppContext.TryGetSwitch(switchName, out bool isSwitchEnabled);
+            if (!hasSwitch)
+            {
+                isSwitchEnabled = GetSwitchDefaultValue(switchName);
+            }
+
+            // Is caching switches disabled?.
```
RussKie, line 50: [nit]
```suggestion
            // Is caching of the switches disabled?
```

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        private static readonly FrameworkName? s_targetFrameworkName = GetTargetFrameworkName();
+
+        private static readonly bool s_isNetCoreApp = (s_targetFrameworkName?.Identifier) == ".NETCoreApp";
+
+        private static FrameworkName? GetTargetFrameworkName()
+        {
+            string? targetFrameworkName = AppContext.TargetFrameworkName;
+            return targetFrameworkName is null ? null : new FrameworkName(targetFrameworkName);
+        }
```
RussKie, line 29: I'm curious, would this work if put in the static .ctor?

RussKie (commented):
Reviewed just the code-related changes, as this what we'll ship. We can iterate over the docs later.

LGTM with few minor changes.

src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
dreddy-work: small text change from container -> form in the doc here. So i kept it.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
dreddy-work: I don't see why not. I just didn't go that route.

docs/WinformsRuntimeConfiguration.md
```diff
@@ -0,0 +1,238 @@
...
+                if (tfm is not null && tfm.Name == "Microsoft.NETCore.App" && string.Compare(tfm.Version,"8.0", StringComparison.OrdinalIgnoreCase) >=0)
+                {
+                    if (switchName == "Switch.System.Windows.Forms.ScaleTopLevelFormMinMaxSizeForDpi")
+                    {
+                        return true;
+                    }
+                }
+            }
+
+            return false;
```
Tanya-Solyanik: Would it be a problem if this switch is set to true on OS lower than 1703?

docs/WinformsRuntimeConfiguration.md
```diff
@@ -0,0 +1,238 @@
...
+    }
+```
+
+
+WinForms runtime then use the static `LocalAppContextSwitches` class to access runtime configurations. Below is sample accessing feature switch `Switch.System.Windows.Forms.ScaleTopLevelFormMinMaxSize`. 
+
+Ex: Use of feature switch in OnDpiChanged() method in Form.cs to scale Min/Max size of the Form.
+
+```cs
+ protected virtual void OnDpiChanged(DpiChangedEventArgs e)
```
Tanya-Solyanik: nit: formatting

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.TargetFramework.cs
Tanya-Solyanik: reactivating for visibility

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -0,0 +1,66 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.CompilerServices;
+
+namespace System.Windows.Forms.Primitives
+{
+    // Barrowed from https://github.com/dotnet/runtime/blob/main/src/libraries/Common/src/System/LocalAppContextSwitches.Common.cs
```
Tanya-Solyanik, line 10: typo: borrowed

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        // Provides default values for switches based on OS and Targetframework versions.
+        private static bool GetSwitchDefaultValue(string switchName)
+        {
+            if (OsVersion.IsWindows10_1703OrGreater)
+            {
+                var tfm = new TargetFramework(AppContext.TargetFrameworkName);
+                if (tfm.Name == ".NETCoreApp" && string.Compare(tfm.Version, "v8.0", StringComparison.OrdinalIgnoreCase) >= 0)
+                {
+                    if (switchName == "Switch.System.Windows.Forms.ScaleTopLevelFormMinMaxSizeForDpi")
```
Tanya-Solyanik: Are switch names case sensitive in .NET Framework?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Dpi/FormTests.Dpi.cs
```diff
@@ -45,5 +45,76 @@ public void Form_DpiChanged_Bounds(int newDpi)
                 User32.SetThreadDpiAwarenessContext(originalAwarenessContext);
             }
         }
+
+        [WinFormsTheory]
+        [InlineData(3.5 * DpiHelper.LogicalDpi)]
+        public void Form_DpiChanged_MinMaxSizeNotChanged_default(int newDpi)
```
Tanya-Solyanik, line 51: I think the convention is to capitalize each word in the function name

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Dpi/FormTests.Dpi.cs
```diff
@@ -45,5 +45,76 @@ public void Form_DpiChanged_Bounds(int newDpi)
...
+                var minSize = new Drawing.Size(100, 100);
+                var maxSize = new Drawing.Size(500, 500);
+                using var form = new Form();
+                form.MinimumSize = minSize;
+                form.MaximumSize = maxSize;
+                form.AutoScaleMode = AutoScaleMode.Dpi;
+                form.Show();
+
+                // Explicitly opt-in to resize min and max sizes with Dpi changed event.
+                AppContext.SetSwitch("Switch.System.Windows.Forms.ScaleTopLevelFormMinMaxSizeForDpi", true);
```
Tanya-Solyanik, line 106: I think it's more future proof to set the test flag here as well, in case we add another test that sets this flag to false

Tanya-Solyanik (approved):
Looks good.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -0,0 +1,82 @@
...
+using System.Runtime.Versioning;
+
+namespace System.Windows.Forms.Primitives
+{
+    // Barrowed from https://github.com/dotnet/runtime/blob/main/src/libraries/Common/src/System/LocalAppContextSwitches.Common.cs
+    internal static partial class LocalAppContextSwitches
+    {
+        private const string SwitchScaleTopLevelFormMinMaxSizeForDpi = "System.Windows.Forms.ScaleTopLevelFormMinMaxSizeForDpi";
+
+        private static int s_scaleTopLevelFormMinMaxSize;
```
Tanya-Solyanik, line 15: add `ForDpi` suffix to match the text name

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Dpi/FormTests.Dpi.cs
```diff
@@ -45,5 +45,77 @@ public void Form_DpiChanged_Bounds(int newDpi)
...
```
Tanya-Solyanik, line 106: Switch name is incorrect! Remove `Switch` prefix!

docs/WinformsRuntimeConfiguration.md
dreddy-work: This particular switch is for PermonV2 scenarios, and we do not want it to be defaulted to true when running on older OSes. as far as i know, .NET applications are not blocked running on Win7/Win8.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
dreddy-work: yes they are.  Will update this line to match with.

docs/WinformsRuntimeConfiguration.md
Tanya-Solyanik: WIn7/8 will not have PerMonitorV2 mode, so I assume the new code will not be executed at all even if the user manually sets this quirk, right?

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -0,0 +1,82 @@
...
+        private static bool GetSwitchValue(string switchName, ref int cachedSwitchValue)
+        {
+            bool hasSwitch = AppContext.TryGetSwitch(switchName, out bool isSwitchEnabled);
+            if (!hasSwitch)
+            {
+                isSwitchEnabled = GetSwitchDefaultValue(switchName);
+            }
+
+            // Is caching of the switches disabled?
+            AppContext.TryGetSwitch("TestSwitch.LocalAppContext.DisableCaching", out bool disableCaching);
```
Tanya-Solyanik, line 52: I would introduce an internal const for this name to use in tests.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -0,0 +1,82 @@
...
+            return isSwitchEnabled;
+
+            static bool GetSwitchDefaultValue(string switchName)
+            {
+                if (!s_isNetCoreApp)
+                {
+                    return false;
+                }
+
+                if (OsVersion.IsWindows10_1703OrGreater)
```
Tanya-Solyanik, line 67: nit: this could be a single if statement with multiple conditions

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -0,0 +1,82 @@
...
+
+using System.Runtime.CompilerServices;
+using System.Runtime.Versioning;
+
+namespace System.Windows.Forms.Primitives
+{
+    // Borrowed from https://github.com/dotnet/runtime/blob/main/src/libraries/Common/src/System/LocalAppContextSwitches.Common.cs
+    internal static partial class LocalAppContextSwitches
+    {
+        private const string ScaleTopLevelFormMinMaxSizeForDpiSwitchName = "System.Windows.Forms.ScaleTopLevelFormMinMaxSizeForDpi";
```
Tanya-Solyanik, line 13: Make it internal for use in tests or access the private dynamically via the test helper.

Tanya-Solyanik: Also going forward we might want to keep all these names in a special file, potentially in the common folder.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -0,0 +1,82 @@
...
+        private static int s_scaleTopLevelFormMinMaxSizeForDpi;
+        public static bool ScaleTopLevelFormMinMaxSizeForDpi
+        {
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            get => GetCachedSwitchValue(ScaleTopLevelFormMinMaxSizeForDpiSwitchName, ref s_scaleTopLevelFormMinMaxSizeForDpi);
+        }
+
+        private static readonly FrameworkName? s_targetFrameworkName = GetTargetFrameworkName();
+
+        private static readonly bool s_isNetCoreApp = (s_targetFrameworkName?.Identifier) == ".NETCoreApp";
```
Tanya-Solyanik, line 24: Is this supposed to be a case-sensitive comparison?

src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
```diff
@@ -1170,6 +1171,15 @@ internal override void Scale(SizeF includedFactor, SizeF excludedFactor, Control
             }
         }
 
+        /// <summary>
+        /// Scales container's properties Min and Max size with the scale factor provided.
+        /// </summary>
+        /// <param name="xScaleFactor">The scale factor to be applied on width of the property being scaled.</param>
+        /// <param name="yScaleFactor">The scale factor to be applied on height of the property being scaled.</param>
+        /// <param name="updateContainerSize"><see langword="true"/> to resize of the container control along with properties being scaled; otherwise, <see langword="false"/>.</param>
```
Tanya-Solyanik, line 1179: nit: break this line into multiple lines

Tanya-Solyanik (approved):
I added some more nits

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Dpi/FormTests.Dpi.cs
```diff
@@ -45,5 +45,83 @@ public void Form_DpiChanged_Bounds(int newDpi)
...
+                var minSize = new Drawing.Size(100, 100);
+                var maxSize = new Drawing.Size(500, 500);
+                using var form = new Form();
+                form.MinimumSize = minSize;
+                form.MaximumSize = maxSize;
+                form.AutoScaleMode = AutoScaleMode.Dpi;
+                form.Show();
+
+                // Explicitly opt-in to resize min and max sizes with Dpi changed event.
+                const string runtimeSwitchName = "System.Windows.Forms.ScaleTopLevelFormMinMaxSizeForDpi";
```
Tanya-Solyanik, line 106: Use the const from the source code here

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Dpi/FormTests.Dpi.cs
```diff
@@ -45,5 +45,83 @@ public void Form_DpiChanged_Bounds(int newDpi)
...
+                AppContext.SetSwitch(runtimeSwitchName, true);
+
+                DpiMessageHelper.TriggerDpiMessage(User32.WM.DPICHANGED, form, newDpi);
+                var factor = newDpi / DpiHelper.LogicalDpi;
+
+                Assert.NotEqual(form.MinimumSize, minSize);
+                Assert.NotEqual(form.MaximumSize, maxSize);
+
+                // Reset switch.
+                AppContext.SetSwitch(runtimeSwitchName, false);
```
Tanya-Solyanik, line 116: If another tests that relies on this switch runs at the same time, you might have a problem. Perhaps these tests should be moved to the other test project, where they run sequentially.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
dreddy-work: this is a single use. Doesn't add value in making it const.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
dreddy-work: right, as we keep adding switches it comes back to this form anyways. i intentionally kept it this way.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
dreddy-work: Agree on moving to a common file for these string const. Will make that change along with fixing skipped test. I enable auto-merge and it did merge when you approved.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
dreddy-work: Yes. I see that all over runtime and kept it in sync.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
Tanya-Solyanik: But it should be used in tests to simulate different target frameworks. f it is not used in our tests, then we don't need it at all.

RussKie:
@dreddy-work a friendly reminder that this change must be documented, please raise a new "breaking change" issue at https://github.com/dotnet/docs.

RussKie:
@dreddy-work has a breaking doc issue been raised?

dreddy-work:
> @dreddy-work has a breaking doc issue been raised?

Its linked above.

RussKie:
Ah, right, thanks!

The doc: https://github.com/dotnet/docs/issues/31020

PR #7469, opened 2022-07-20
Fully qualify code in `ApplicationConfiguration.Initialize`

### LINKED ISSUE #7346: Generated code should be fully qualified
### .NET version

6.0.202

No

_No response_

### Issue description

If a user's namespace includes `Application`, then the generated code for `ApplicationConfiguration` will result in compile time issues. Controls should be fully qualified to avoid these issues. This is how the [Logging generator](https://github.com/dotnet/runtime/blob/0488519e709811fd6f96aa0aed4c17a72c00898c/src/libraries/Microsoft.Extensions.Logging.Abstractions/tests/Microsoft.Extensions.Logging.Generators.Tests/Baselines/TestWithDefaultValues.generated.txt) works.

**Current:**

```csharp
internal static partial class ApplicationConfiguration
{
    public static void Initialize()
    {
        Application.SetCompatibleTextRenderingDefault(false);
        Application.SetHighDpiMode(HighDpiMode.SystemAware);
    }
}
```

**Desired:**

```csharp
internal static partial class ApplicationConfiguration
{
    public static void Initialize()
    {
        global::System.Windows.Forms.Application.EnableVisualStyles();
        global::System.Windows.Forms.Application.SetCompatibleTextRenderingDefault(false);
        global::System.Windows.Forms.Application.SetHighDpiMode(global::System.Windows.Forms.HighDpiMode.SystemAware);
    }
}
```

### Steps to reproduce

1. Create a new WinForms project `Something.Application`
2. Try to build

fernandanavamoya:
Fixes #7346

## Proposed changes

- Added full qualifier to ApplicationConfiguration generated code by creating a type string variable named 'qualifier' which is equaled to the full qualifier and then adding it to the controls.
-  Changed expected result in test file by adding the full qualifier.

``` C#
static string GenerateCode(ApplicationConfig projectConfig, string? defaultFont, string indent)
            {
                StringBuilder code = new();
                string qualifier = "global::System.Windows.Forms.Application.";
                if (projectConfig.EnableVisualStyles)
                {
                    code.AppendLine($"{indent}{qualifier}EnableVisualStyles();");
                }

                code.AppendLine($"{indent}{qualifier}SetCompatibleTextRenderingDefault({projectConfig.UseCompatibleTextRendering.ToString().ToLowerInvariant()});");

                code.AppendLine($"{indent}{qualifier}SetHighDpiMode(HighDpiMode.{projectConfig.HighDpiMode});");

                // Note: we need to set the font _after_ we set the DPI scaling, as it affects how we scale the font.
                if (!string.IsNullOrWhiteSpace(defaultFont))
                {
                    code.AppendLine($"{indent}{qualifier}SetDefaultFont({defaultFont});");
                }

                // Don't append line as we don't need the trailing \r\n!
                code.Remove(code.Length - Environment.NewLine.Length, Environment.NewLine.Length);

                return code.ToString();
            }
```

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7469)

### DIFF
--- a/src/System.Windows.Forms.Analyzers.CSharp/src/System/Windows/Forms/Generators/ApplicationConfigurationInitializeBuilder.cs
+++ b/src/System.Windows.Forms.Analyzers.CSharp/src/System/Windows/Forms/Generators/ApplicationConfigurationInitializeBuilder.cs
@@ -22,19 +22,20 @@ namespace System.Windows.Forms.Generators
             static string GenerateCode(ApplicationConfig projectConfig, string? defaultFont, string indent)
             {
                 StringBuilder code = new();
+                const string qualifier = "global::System.Windows.Forms.Application.";
                 if (projectConfig.EnableVisualStyles)
                 {
-                    code.AppendLine($"{indent}Application.EnableVisualStyles();");
+                    code.AppendLine($"{indent}{qualifier}EnableVisualStyles();");
                 }
 
-                code.AppendLine($"{indent}Application.SetCompatibleTextRenderingDefault({projectConfig.UseCompatibleTextRendering.ToString().ToLowerInvariant()});");
+                code.AppendLine($"{indent}{qualifier}SetCompatibleTextRenderingDefault({projectConfig.UseCompatibleTextRendering.ToString().ToLowerInvariant()});");
 
-                code.AppendLine($"{indent}Application.SetHighDpiMode(HighDpiMode.{projectConfig.HighDpiMode});");
+                code.AppendLine($"{indent}{qualifier}SetHighDpiMode(HighDpiMode.{projectConfig.HighDpiMode});");
 
                 // Note: we need to set the font _after_ we set the DPI scaling, as it affects how we scale the font.
                 if (!string.IsNullOrWhiteSpace(defaultFont))
                 {
-                    code.AppendLine($"{indent}Application.SetDefaultFont({defaultFont});");
+                    code.AppendLine($"{indent}{qualifier}SetDefaultFont({defaultFont});");
                 }
 
                 // Don't append line as we don't need the trailing \r\n!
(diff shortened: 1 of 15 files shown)


src/System.Windows.Forms.Analyzers.CSharp/src/System/Windows/Forms/Generators/ApplicationConfigurationInitializeBuilder.cs
```diff
@@ -22,19 +22,20 @@ public static string GenerateInitialize(string? projectNamespace, ApplicationCon
             static string GenerateCode(ApplicationConfig projectConfig, string? defaultFont, string indent)
             {
                 StringBuilder code = new();
+                string qualifier = "global::System.Windows.Forms.Application.";
```
KlausLoeffelmann, line 25: This could be a string constant, so we don't allocate anything at runtime.
Take a look here to read up on this:

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/const

Be careful though, when using constants which are visible to other assemblies. It's not good practice, and in any case, should you ever change them, you'd need to recompile everything depending on them, since they are stored in the meta data rather than at runtime. (Which is the whole point! 😄).

KlausLoeffelmann (approved):
Looks good to me!

I'd probably want to change the title of this PR to describe what it fixes rather how it fixes it, though. Like "Fix failing AppConfig code generation for namespaces including 'Application'" or something like that. (The description of the PR is probably derived from a commit description?)

src/System.Windows.Forms.Analyzers.CSharp/src/System/Windows/Forms/Generators/ApplicationConfigurationInitializeBuilder.cs
KlausLoeffelmann: Oh, and take a look why the tests are failing!

KlausLoeffelmann (changes requested):
Please check the failing tests first.

src/System.Windows.Forms.Analyzers.CSharp/src/System/Windows/Forms/Generators/ApplicationConfigurationInitializeBuilder.cs
RussKie:
```suggestion
                const string qualifier = "global::System.Windows.Forms.Application.";
```

RussKie:
@fernandanavamoya there are still [tests failing](https://dev.azure.com/dnceng/public/_build/results?buildId=1893870&view=ms.vss-test-web.build-test-results-tab&runId=49368450&resultId=100127&paneView=debug).

RussKie (approved):

RussKie:
Thank you

PR #7478, opened 2022-07-22
Fix ListView memory leaks

### LINKED ISSUE #7385: ListView accessible objects are leaking in memory
### .NET version

7.0 Preview 6

No

_No response_

### Issue description

`ListView` accessible objects are left in memory after the `ListView` control instance with its items was inspected with accessibility tool (like Accessibility Insights or Inspect) and the form with `ListView` is closed, garbage is collected, but accessibility tool is still opened. 

image

### Steps to reproduce

1. Create a parent form with two buttons: one will be opening a child form with the `ListView` control and the other will collect the garbage.
2. Press the second button to open the child form with `ListView`.
3. Start the Accessibility Insights tools or Inspect.
4. Run automated checks for the `ListView` control using AI/Inspect.
5. Close the child form.
7. DON'T close the AI/Inspect tool.
8. Look for `ListView` objects left in memory (using WinDbg, or dotMemory, or another suitable tool).

dmitrii-drobotov:
Fixes #7385

## Proposed changes

- Add a call of `UiaDisconnectProvider` for `ListView` elements (items, subitems, groups, column headers) for cases when:
    - `ListView` is destroyed
    - Collection is cleared
    - Element is removed from a collection
    - Element is replaced in a collection (where applicable)
- Change `RuntimeId` of some accessible objects to be independent from its index in a collection. When collection is modified, e.g. some elements are removed or inserted at the beginning, indexes (and therefore `RuntimeId`s) of elements change. It then leads to a possible disconnection of wrong items, which had the index initially.

## Customer Impact

- Less memory leaks for `ListView`

## Regression? 

- No

## Risk

- Minimal

### Before

image

### After

image

## Test methodology 

- CTI
- Manually (via WinDbg)
- Unit tests — we don't always add unit tests for memory leaks fixes, but here I wanted to make sure that all disconnects are properly called through the hierarchy, and in addition use unit tests as a form of documentation for main scenarios

## Accessibility testing  

- Using AI, Inspect, Narrator
 

## Test environment(s) 

- .NET 7.0 RC1
- Windows 11

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7478)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListView.ColumnHeaderCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListView.ColumnHeaderCollection.cs
@@ -329,6 +329,7 @@ namespace System.Windows.Forms
                         {
                             int w = _owner._columnHeaders[colIdx].Width; // Update width before detaching from ListView
                             _owner._columnHeaders[colIdx].OwnerListview = null;
+                            _owner._columnHeaders[colIdx].ReleaseUiaProvider();
                         }
 
                         _owner._columnHeaders = null;
@@ -348,6 +349,7 @@ namespace System.Windows.Forms
                             }
 
                             _owner._columnHeaders[colIdx].OwnerListview = null;
+                            _owner._columnHeaders[colIdx].ReleaseUiaProvider();
                         }
 
                         _owner._columnHeaders = null;
@@ -547,6 +549,7 @@ namespace System.Windows.Forms
                 }
 
                 removeHdr.DisplayIndexInternal = -1;
+                removeHdr.ReleaseUiaProvider();
 
                 _owner._columnHeaders[index].OwnerListview = null;
                 int columnCount = _owner._columnHeaders.Length;
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListView.ListViewAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListView.ListViewAccessibleObject.cs
@@ -158,7 +158,7 @@ namespace System.Windows.Forms
                 UiaCore.IRawElementProviderSimple[] columnHeaders = new UiaCore.IRawElementProviderSimple[_owningListView.Columns.Count];
                 for (int i = 0; i < columnHeaders.Length; i++)
                 {
-                    columnHeaders[i] = new ColumnHeader.ListViewColumnHeaderAccessibleObject(_owningListView.Columns[i]);
+                    columnHeaders[i] = _owningListView.Columns[i].AccessibilityObject;
                 }
 
                 return columnHeaders;
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
@@ -5127,6 +5127,36 @@ namespace System.Windows.Forms
             }
         }
 
+        internal override void ReleaseUiaProvider(nint handle)
+        {
+            base.ReleaseUiaProvider(handle);
+
+            if (!OsVersion.IsWindows8OrGreater)
+            {
+                return;
+            }
+
+            for (int i = 0; i < Items.Count; i++)
+            {
+                Items[i].ReleaseUiaProvider();
+            }
+
+            DefaultGroup.ReleaseUiaProvider();
+
+            for (int i = 0; i < Groups.Count; i++)
+            {
+                Groups[i].ReleaseUiaProvider();
+            }
+
+            if (_columnHeaders is not null)
+            {
+                for (int i = 0; i < _columnHeaders.Length; i++)
+                {
+                    _columnHeaders[i].ReleaseUiaProvider();
+                }
+            }
+        }
+
         // makes sure that the list view items which are w/o a listView group are parented to the DefaultGroup - if necessary
         // and then tell win32 to remove this group
         internal void RemoveGroupFromListView(ListViewGroup group)
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
@@ -5131,7 +5131,7 @@ namespace System.Windows.Forms
         {
             base.ReleaseUiaProvider(handle);
 
-            if (!OsVersion.IsWindows8OrGreater)
+            if (!OsVersion.IsWindows8OrGreater || !IsAccessibilityObjectCreated)
             {
                 return;
             }
@@ -5143,17 +5143,14 @@ namespace System.Windows.Forms
 
             DefaultGroup.ReleaseUiaProvider();
 
-            for (int i = 0; i < Groups.Count; i++)
+            foreach (ListViewGroup group in Groups)
             {
-                Groups[i].ReleaseUiaProvider();
+                group.ReleaseUiaProvider();
             }
 
-            if (_columnHeaders is not null)
+            foreach (ColumnHeader columnHeader in Columns)
             {
-                for (int i = 0; i < _columnHeaders.Length; i++)
-                {
-                    _columnHeaders[i].ReleaseUiaProvider();
-                }
+                columnHeader.ReleaseUiaProvider();
             }
         }
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroup.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroup.cs
@@ -20,7 +20,7 @@ namespace System.Windows.Forms
     [DesignTimeVisible(false)]
     [DefaultProperty(nameof(Header))]
     [Serializable] // This type is participating in resx serialization scenarios.
-    public partial class ListViewGroup : ISerializable
+    public sealed partial class ListViewGroup : ISerializable
     {
         private string? _header;
         private HorizontalAlignment _headerAlignment = HorizontalAlignment.Left;
@@ -49,7 +49,7 @@ namespace System.Windows.Forms
         /// <summary>
         ///  Creates a ListViewItem object from an Stream.
         /// </summary>
-        protected ListViewGroup(SerializationInfo info, StreamingContext context) : this()
+        private ListViewGroup(SerializationInfo info, StreamingContext context) : this()
         {
             Deserialize(info);
         }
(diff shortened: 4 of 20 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
```diff
@@ -5127,6 +5127,36 @@ public void RedrawItems(int startIndex, int endIndex, bool invalidateOnly)
             }
         }
 
+        internal override void ReleaseUiaProvider(nint handle)
+        {
+            base.ReleaseUiaProvider(handle);
+
+            if (!OsVersion.IsWindows8OrGreater)
```
vladimir-krestov, line 5134: I would add "IsAccessibleObjectCreated" check

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
```diff
@@ -5127,6 +5127,36 @@ public void RedrawItems(int startIndex, int endIndex, bool invalidateOnly)
...
+            {
+                return;
+            }
+
+            for (int i = 0; i < Items.Count; i++)
```
vladimir-krestov, line 5139: Why not "foreach" for it? Groups as well

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
```diff
@@ -5127,6 +5127,36 @@ public void RedrawItems(int startIndex, int endIndex, bool invalidateOnly)
...
+            DefaultGroup.ReleaseUiaProvider();
+
+            for (int i = 0; i < Groups.Count; i++)
+            {
+                Groups[i].ReleaseUiaProvider();
+            }
+
+            if (_columnHeaders is not null)
+            {
+                for (int i = 0; i < _columnHeaders.Length; i++)
```
vladimir-krestov, line 5153: Ditto

src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroup.cs
```diff
@@ -19,7 +20,7 @@ namespace System.Windows.Forms
     [DesignTimeVisible(false)]
     [DefaultProperty(nameof(Header))]
     [Serializable] // This type is participating in resx serialization scenarios.
-    public sealed partial class ListViewGroup : ISerializable
```
vladimir-krestov, line 22: Why do we need it? If only for tests, so tests are not a reason to change the public API. So please find another way and return it back, Unshipped.txt as well.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.ListViewItemBaseAccessibleObject.cs
```diff
@@ -223,6 +223,14 @@ internal override bool IsPatternSupported(UiaCore.UIA patternId)
                     _ => base.IsPatternSupported(patternId)
                 };
 
+            internal virtual void ReleaseChildUiaProviders()
+            {
+                for (int i = 0; i < _owningItem.SubItemCount; i++)
```
vladimir-krestov, line 228: "foreach" as well

vladimir-krestov (commented):
Looks good!

src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroup.cs
dmitrii-drobotov: Yeah, it was only for testing. I added setting child accessible objects to null after disconnect, so now in tests I can just check for null, and no need to remove `sealed`.

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
dmitrii-drobotov: For `Items` we have to use the `for` loop because `foreach` doesn't work with `VirtualMode`: `ListViewItemCollection.GetEnumerator` throws following exception
> When the ListView is in virtual mode, you cannot enumerate through the ListView items collection using an enumerator or call GetEnumerator. Use the ListView items indexer instead and access an item by index value.

Other collections doesn't have this limitation, so I switched to `foreach`.

dmitrii-drobotov:
CTI approved the fix

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
```diff
@@ -5127,6 +5127,33 @@ public void RedrawItems(int startIndex, int endIndex, bool invalidateOnly)
...
+        internal override void ReleaseUiaProvider(nint handle)
+        {
+            base.ReleaseUiaProvider(handle);
+
+            if (!OsVersion.IsWindows8OrGreater || !IsAccessibilityObjectCreated)
+            {
+                return;
+            }
+
+            for (int i = 0; i < Items.Count; i++)
```
dreddy-work, line 5139: why is this accessed with index while rest with foreach?

dreddy-work (approved):

dreddy-work:
@Tanya-Solyanik  as FYI.

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
dmitrii-drobotov: If we use foreach and the ListView has `VirtualMode` as true, then `ListViewItemCollection.GetEnumerator` will throw following exception

>When the ListView is in virtual mode, you cannot enumerate through the ListView items collection using an enumerator or call GetEnumerator. Use the ListView items indexer instead and access an item by index value.

And it only applies for `Items` collection.

vladimir-krestov (approved):

src/System.Windows.Forms/src/System/Windows/Forms/ListView.ColumnHeaderCollection.cs
```diff
@@ -329,6 +329,7 @@ public virtual void Clear()
                         {
                             int w = _owner._columnHeaders[colIdx].Width; // Update width before detaching from ListView
                             _owner._columnHeaders[colIdx].OwnerListview = null;
+                            _owner._columnHeaders[colIdx].ReleaseUiaProvider();
```
Tanya-Solyanik, line 332: @dmitrii-drobotov  - ColumnHeaders are Components and as such implement IDisposable. We want to dispose each header when clearing the array.

src/System.Windows.Forms/src/System/Windows/Forms/ListView.ColumnHeaderCollection.cs
```diff
@@ -547,6 +549,7 @@ public virtual void RemoveAt(int index)
                 }
 
                 removeHdr.DisplayIndexInternal = -1;
+                removeHdr.ReleaseUiaProvider();
```
Tanya-Solyanik, line 552: I think `removeHdr` should be explicitly disposed as well.

src/System.Windows.Forms/src/System/Windows/Forms/ListView.ListViewAccessibleObject.cs
```diff
@@ -158,7 +158,7 @@ private string GetItemStatus()
                 UiaCore.IRawElementProviderSimple[] columnHeaders = new UiaCore.IRawElementProviderSimple[_owningListView.Columns.Count];
                 for (int i = 0; i < columnHeaders.Length; i++)
                 {
-                    columnHeaders[i] = new ColumnHeader.ListViewColumnHeaderAccessibleObject(_owningListView.Columns[i]);
+                    columnHeaders[i] = _owningListView.Columns[i].AccessibilityObject;
```
Tanya-Solyanik, line 161: nice!

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
```diff
@@ -5127,6 +5127,33 @@ public void RedrawItems(int startIndex, int endIndex, bool invalidateOnly)
...
+            {
+                Items[i].ReleaseUiaProvider();
+            }
+
+            DefaultGroup.ReleaseUiaProvider();
```
Tanya-Solyanik, line 5144: @dmitrii-drobotov - Access to `DefaultGroup` property forces creation of that default group. Should you consult `GroupsEnabled` before accessing the default group?

RussKie: Good catch @Tanya-Solyanik.
@dmitrii-drobotov please use `OwnerHasDefaultGroup` to guard.

dmitrii-drobotov: Fixed in #7524. Decided to directly check `if (_defaultGroup is not null)` because false results of `OwnerHasDefaultGroup` and `GroupsEnabled` don't guarantee that `DefaultGroup` wasn't created before. One can disable groups dynamically (set `listView.ShowGroups` to `false`), then `OwnerHasDefaultGroup`/`GroupsEnabled` will be `false` but the `_defaultGroup` will still be not null and need to be disconnected.

src/System.Windows.Forms/src/System/Windows/Forms/ListView.ColumnHeaderCollection.cs
dmitrii-drobotov: I don't think it needs to be disposed here. In `ColumnHeader.Dispose` there is a call to `ListView.Columns.RemoveAt(index);` which is already being done here. Another point is that removed `ColumnHeader` object can be reused in another `ListView` or readded back, so I'm not sure that disposal wouldn't break those scenarios.

Tanya-Solyanik: ok, makes sense.

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
Tanya-Solyanik: Makes sense

RussKie: > Decided to directly check if (_defaultGroup is not null) because false results of OwnerHasDefaultGroup and GroupsEnabled don't guarantee that DefaultGroup wasn't created before. 

That's a good catch, and we should fix this too.

dmitrii-drobotov: > > Decided to directly check if (_defaultGroup is not null) because false results of OwnerHasDefaultGroup and GroupsEnabled don't guarantee that DefaultGroup wasn't created before.
> 
> That's a good catch, and we should fix this too.

As I can see by usages, those properties should represent if groups are currently shown and don't need to depend on whether DefaultGroup was created before. This seems to be the only case where it matters, so I don't think we should do anything here.

PR #7490, opened 2022-07-25
Adding notification to Narrator about a subtitle.

### LINKED ISSUE #3261: Accessibility: ListViewGroup Subtitle Property
**Problem description:**

The ListViewGroup has a subtitle property that is not being picked up by both inspect and narrator.

**Expected behavior:**

The subtitle property should be read by narrator and it should also be displayed somewhere in inspect

**Minimal repro:**

1. Create a winforms core app
2. Add a ListView and ListViewGroup with a subtitle
4. Run app
3. Start inspect/narrator and click on the ListViewGroup with the subtitle in the app

**Note:**
This issue stems from PR #3234

NikitaSemenovAkvelon:
Fixes #3261

## Proposed changes

- Raised a notificatoin contained a subtitle values on setting focus.
- Added check for focus state in SetState method.
- Added unit test.

## Customer Impact

- User can hear a subtitle of ListViewGroup.

## Regression? 

- No

## Risk

- Minimal

### Before

image

### After

image

## Test methodology 

- Manually
- Narrator
- Unit test

## Accessibility testing  

- Narrator

 

## Test environment(s) 

- 7.0.100-preview.5.22307.18

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7490)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroup.ListViewGroupAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroup.ListViewGroupAccessibleObject.cs
@@ -99,7 +99,9 @@ namespace System.Windows.Forms
                 => _owningListView.AccessibilityObject;
 
             public override string Name
-                => _owningGroup.Header;
+                => !string.IsNullOrEmpty(_owningGroup.Subtitle)
+                    ? $"{_owningGroup.Header}. {_owningGroup.Subtitle}"
+                    : _owningGroup.Header;
 
             public override AccessibleRole Role
                 => AccessibleRole.Grouping;
--- a/src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/ListViewTest.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/ListViewTest.cs
@@ -170,6 +170,7 @@ namespace WinformsControlsTest
         private void AddGroupTasks()
         {
             listView1.Groups[0].TaskLink = "Task";
+            listView1.Groups[0].Subtitle = "Subtitle";
             listView1.GroupTaskLinkClick += listView1_GroupTaskLinkClick;
 
             var lvgroup1 = new ListViewGroup
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewGroup.ListViewGroupAccessibleObjectTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewGroup.ListViewGroupAccessibleObjectTests.cs
@@ -110,6 +110,25 @@ namespace System.Windows.Forms.Tests
             Assert.False(list.IsHandleCreated);
         }
 
+        [WinFormsFact]
+        public void ListViewGroupAccessibleObject_GetPropertyValue_ReturnsExpected_WithSubtitle()
+        {
+            using ListView list = new ListView();
+            const string name = "Group1";
+            const string subtitle = "Subtitle";
+            ListViewGroup listGroup = new ListViewGroup(name) { Subtitle = subtitle };
+            listGroup.Items.Add(new ListViewItem());
+            list.Groups.Add(listGroup);
+
+            AccessibleObject accessibleObject = listGroup.AccessibilityObject;
+            Assert.False(list.IsHandleCreated);
+
+            object accessibleName = accessibleObject.GetPropertyValue(UiaCore.UIA.NamePropertyId);
+            Assert.Equal($"{name}. {subtitle}", accessibleName);
+
+            Assert.False(list.IsHandleCreated);
+        }
+
         [WinFormsTheory]
         [MemberData(nameof(ListViewGroupAccessibleObject_TestData))]
         public void ListViewGroupAccessibleObject_FragmentNavigate_ReturnsExpected_WithDefaultGroup(View view, bool showGroups, bool createHandle)


src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroup.ListViewGroupAccessibleObject.cs
```diff
@@ -296,16 +297,27 @@ internal override bool IsPatternSupported(UiaCore.UIA patternId)
                     _ => base.IsPatternSupported(patternId),
                 };
 
-            internal override unsafe void SetFocus()
+            internal override void SetFocus()
```
Tanya-Solyanik, line 300: I wonder why this had been marked as `unsafe` previously.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroup.ListViewGroupAccessibleObject.cs
```diff
@@ -296,16 +297,27 @@ internal override bool IsPatternSupported(UiaCore.UIA patternId)
...
 
+                if (ReferenceEquals(_owningListView.FocusedGroup, _owningGroup))
+                {
+                    return;
+                }
+
                 _owningListView.FocusedGroup = _owningGroup;
 
                 RaiseAutomationEvent(UiaCore.UIA.AutomationFocusChangedEventId);
+                if (!string.IsNullOrEmpty(_owningGroup.Subtitle))
```
Tanya-Solyanik, line 315: Had you considered overriding AccessibleName for ListViewGroup to include Subtitle? @merriemcgaw  - do you think announcing subtitle every time the title is announced could be too much?

Tanya-Solyanik (commented):
Looks good ut I would like to hear Merrie's opinion on adding subtitle text to AccessibleName

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewGroup.ListViewGroupAccessibleObjectTests.cs
```diff
@@ -1252,6 +1253,34 @@ public static IEnumerable<object[]> ListViewGroupAccessibleObject_IsPatternSuppo
...
+
+            Assert.Equal(0, listViewGroupAccessibleObject.RaiseNotificationCallCount);
+
+            listViewGroupAccessibleObject.SetFocus();
+
+            Assert.Equal(1, listViewGroupAccessibleObject.RaiseNotificationCallCount);
+
+            listViewGroupAccessibleObject.SetFocus();
+
+            Assert.Equal(1, listViewGroupAccessibleObject.RaiseNotificationCallCount);
```
vladimir-krestov, line 1281: Please add a comment, why the count is still "1" after the focus action to make it clear.

vladimir-krestov (commented):
Looks good

vladimir-krestov:
> Looks good ut I would like to hear Merrie's opinion on adding subtitle text to AccessibleName

Actually, adding a subtitle into an AccessibleName makes sense. If to consider some practical case (not abstract text), a subtitle looks like an additional comment to the main title, in my vision. For example:

image

This example sounds good, especially when using "." between the title and the subtitle.

I see the following advantages of the Tanya's approach:
1) Users can change a Narrator's notification, if they don't like it (they can't do it if we raise our notification). They can just implement their own AccessibleName (eg. via reflection)
2) We will avoid out of sync of notifications, the message will be heard the same every time
3) Easily to support and implement
4) Narrator says the message with the same intonation

But the final approach should be based on most common cases of Subtitle usage.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroup.ListViewGroupAccessibleObject.cs
merriemcgaw: Hmmmm, that's a good question. I would almost think that AccessibleDescription might be a better place. So folks can choose whether or not that additional data is read by AT.

merriemcgaw:
> But the final approach should be based on most common cases of Subtitle usage.

That is the best approach. Try looking at web examples of groupings with sub-headings, that may give us the most common approach. I agree with @vladimir-krestov that it does sound good when the title and subtitle make sense together, now that you show me that context. But I'd also like to see how AccessibleDescription shows up to Narrator.

RussKie (dismissed):

vladimir-krestov:
@NikitaSemenovAkvelon, please investigate AccessibleDescription case.
@merriemcgaw, what about "Task" links? As it's a UI element, it should be announced by Narrator as well. Tasks don't look like a part of a title, so I think, we can use the current UIANotification approach for announcement of "tasks". Should we announce them at all in your opinion?

NikitaSemenovAkvelon:
I'd added a description to ListViewGroup and Button and checked how it works. Both of this controls ingore Description. You can see below that the property was filled, but Narrator is still ingoring it.
image
image
image
image

dreddy-work (approved):

dreddy-work:
adding @merriemcgaw to finalize here.

merriemcgaw:
Given that the AccessibleDescription is ignored, I think I'd prefer to add it to AccessibleName as you suggested earlier. I think that makes the most sense here.

dreddy-work:
@NikitaSemenovAkvelon, is there anything else pending here?

vladimir-krestov:
@dreddy-work, it's not ready to merge. The implementation should be reworked, a subtitle should be announced as a part of an AccessibleName.

NikitaSemenovAkvelon:
I've reworked the PR as you suggested, and now the Subtitle property is a part of AccessibleName.
NarratorBuddy:
Screenshot 2022-08-01 093437
AI:
Screenshot 2022-08-01 093521

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/ListViewTest.cs
```diff
@@ -170,6 +170,7 @@ private void listView1_GroupCollapsedStateChanged(object sender, ListViewGroupEv
         private void AddGroupTasks()
         {
             listView1.Groups[0].TaskLink = "Task";
+            listView1.Groups[0].Subtitle = "Subtitle";
```
Tanya-Solyanik, line 173: Thank you for updating the integration test!

Tanya-Solyanik (approved):
Looks good!

PR #7491, opened 2022-07-25

gpetrou:
## Proposed changes

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7491)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ScrollBar.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ScrollBar.cs
@@ -566,7 +566,7 @@ namespace System.Windows.Forms
             // The base implementation should be called before the implementation above,
             // but changing the order in Whidbey would be too much of a breaking change
             // for this particular class.
-            base.OnMouseWheel(e);
+            base.OnMouseWheel(e!);
         }
 
         /// <summary>
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ScrollBar.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ScrollBar.cs
@@ -566,6 +566,7 @@ namespace System.Windows.Forms
             // The base implementation should be called before the implementation above,
             // but changing the order in Whidbey would be too much of a breaking change
             // for this particular class.
+            // Because the code has been like that since long time, we assume that e is not null.
             base.OnMouseWheel(e!);
         }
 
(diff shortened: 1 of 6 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ScrollBar.cs
```diff
@@ -566,7 +566,7 @@ protected override void OnMouseWheel(MouseEventArgs e)
             // The base implementation should be called before the implementation above,
             // but changing the order in Whidbey would be too much of a breaking change
             // for this particular class.
-            base.OnMouseWheel(e);
+            base.OnMouseWheel(e!);
```
gpetrou, line 569: I am not sure what to do about these cases. Should we make the args nullable in all classes that have OnMouseWheel instead?

RussKie (commented):
These three identified methods that check incoming eventargs look very odd, we don't normally null-check `e` (e.g., painting, keyboard or mouse handling).

src/System.Windows.Forms/src/System/Windows/Forms/ScrollBar.cs
gpetrou: Any more thoughts about this? Perhaps the safest thing to do for now is just keep the null forgiving operators we have in the PR in order to avoid any behavioral changes?

gpetrou:
@RussKie any more thoughts about this?

src/System.Windows.Forms/src/System/Windows/Forms/ScrollBar.cs
RussKie: Apologies, it looks like I have failed to respond after this was discussed at a team meeting, and now I'm scratching my head trying to page back the decision.

If I recall correctly, we agreed on the following:
1. We can't assume something we know may be a null as "not null".
2. We can't throw a ANE here as it would be a behaviour breaking change.
3. We should throw a ANE in the base, as it would be within the "acceptable" breaking change rules - i.e., throwing a ANE instead of a NRE:
https://github.com/dotnet/winforms/blob/d121a04d5bce7f376773af64b4151c516b63c837/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs#L8582-L8585

Though, testing it locally it doesn't appear `OnMouseWheel(null!)` throws anything. 🤔 

@JeremyKuhne am I misremembering something?

JeremyKuhne: You have this right, @RussKie. We generally don't want to introduce ANE in existing APIs with the specific exception of when we **know** we're about to throw NRE (in the same method).

gpetrou: Hmm, so what should be the next step? If I change this to be nullable, I suspect that there will be some places where we should add null checks because we override this in classes that inherit from Control. If I change it to throw an ANE, I suspect that there will be changes in tests and therefore behavior. I still think that using `!` is probably the safest option for now.

RussKie: This one is a headscratcher, isn't it? :)

I think it's probably acceptable to use `!` here with an added comment (like we've done in other instances) stating that this has been like this for a very long time, and we believe this should never be null. If/when we get a customer feedback we can review this, if required.

gpetrou: @RussKie I have added the comments. Perhaps we could merge this now?

RussKie (approved):

RussKie:
Thank you

PR #7499, opened 2022-07-27

fernandanavamoya:
Converted these threads:
-GetProcAddress
-GetTickCount
-GetUserDefaultLCID
-GlobalSize

Added PInvoke.GetModuleHandle.cs file

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7499)


src/System.Windows.Forms.Primitives/src/NativeMethods.txt
```diff
@@ -2,15 +2,27 @@ BitBlt
...
+GetProcAddress
 GetThreadLocale
+GetTickCount
 GlobalAlloc
 GlobalFree
 GlobalLock
 GlobalReAlloc
+GlobalSize
 GlobalUnlock
+GetProcAddress
```
dreddy-work, line 23: Sort order missing for new additions here.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleHandle.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        public static unsafe HINSTANCE GetModuleHandle(
+            string lpModuleName)
+        {
+            // We are using GetModuleHandleEx because of the difficulties of the safe handle overload generated when
+            // using GetModuleHandle
+            fixed (char* lpModuleNameLocal = lpModuleName)
+            {
+                HINSTANCE hModule;
+                BOOL result = GetModuleHandleEx(GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT, lpModuleNameLocal, &hModule);
```
dreddy-work, line 19: if you are not using/checking `result`,  you may want to ignore the return value here.

src/System.Windows.Forms/src/System/Windows/Forms/Application.cs
```diff
@@ -118,20 +119,28 @@ private static bool InitializeComCtlSupportsVisualStyles()
             //
             // GetModuleHandle  returns a handle to a mapped module without incrementing its
             // reference count.
-            IntPtr hModule = Kernel32.GetModuleHandleW(Libraries.Comctl32);
-            if (hModule != IntPtr.Zero)
+            var hModule = PInvoke.GetModuleHandle(Libraries.Comctl32);
+            string IpProcName = "ImageList_WriteEx";
```
dreddy-work, line 123: const string?

src/System.Windows.Forms/src/System/Windows/Forms/Application.cs
```diff
@@ -118,20 +119,28 @@ private static bool InitializeComCtlSupportsVisualStyles()
...
+            nint ninthModule = Kernel32.LoadComctl32(StartupPath);
+            if (ninthModule == 0)
             {
                 return false;
             }
 
-            return Kernel32.GetProcAddress(hModule, "ImageList_WriteEx") != IntPtr.Zero;
+            fixed (byte* ptr = &test[0])
+            {
+                return PInvoke.GetProcAddress(hModule, (Foundation.PCSTR)ptr) != 0;
```
dreddy-work, line 142: isn't this `ninthModule`

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
```diff
@@ -3232,7 +3232,7 @@ public unsafe bool HasPropertyPages()
 
         private unsafe void ShowPropertyPageForDispid(Ole32.DispatchID dispid, Guid guid)
         {
-            nint pUnk = Marshal.GetIUnknownForObject(GetOcx());
+            IntPtr pUnk = Marshal.GetIUnknownForObject(GetOcx());
```
dreddy-work, line 3235: why this change?

src/System.Windows.Forms/src/System/Windows/Forms/NativeWindow.cs
```diff
@@ -163,20 +164,24 @@ internal unsafe void ForceExitMessageLoop()
...
         /// </summary>
-        internal static IntPtr DefaultWindowProc
+        internal unsafe static nint DefaultWindowProc
         {
             get
             {
-                if (s_defaultWindowProc == IntPtr.Zero)
+                var hModule = PInvoke.GetModuleHandle(Libraries.Comctl32);
+                string IpProcName = "DefWindowProcW";
+                byte[] test = Encoding.ASCII.GetBytes(IpProcName);
```
dreddy-work, line 173: if `s_defaultWindowProc `is not zero, we should avoid these calls and bail out as it was doing earlier.

src/System.Windows.Forms.Primitives/src/NativeMethods.txt
RussKie: This is a dup - likely a bad merge

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleHandle.cs
```diff
@@ -0,0 +1,24 @@
...
+// See the LICENSE file in the project root for more information.
+
+using Windows.Win32.Foundation;
+
+namespace Windows.Win32
+{
+    internal static partial class PInvoke
+    {
+        public static unsafe HINSTANCE GetModuleHandle(
+            string lpModuleName)
```
RussKie, line 12: Keep on the same line

src/System.Windows.Forms/src/System/Windows/Forms/Application.cs
RussKie: Incorrect casing too.
Ditto below

src/System.Windows.Forms/src/System/Windows/Forms/Application.cs
```diff
@@ -118,20 +119,28 @@ private static bool InitializeComCtlSupportsVisualStyles()
...
-            if (hModule != IntPtr.Zero)
+            var hModule = PInvoke.GetModuleHandle(Libraries.Comctl32);
+            string IpProcName = "ImageList_WriteEx";
+            byte[] test = Encoding.ASCII.GetBytes(IpProcName);
+            fixed (byte* ptr = &test[0])
             {
-                return Kernel32.GetProcAddress(hModule, "ImageList_WriteEx") != IntPtr.Zero;
+                if (hModule != 0)
+                {
+                    return PInvoke.GetProcAddress(hModule,(Foundation.PCSTR)ptr) != 0;
```
RussKie, line 129: Incorrect spacing

src/System.Windows.Forms.Primitives/src/NativeMethods.txt
fernandanavamoya: You're right, already corrected it.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
fernandanavamoya: In the last coding review, it was suggested that `IntPtr` is better for representing an opaque value like a pointer.

src/System.Windows.Forms/src/System/Windows/Forms/Application.cs
fernandanavamoya: `ninthModule` is a `nint` which stores the return value of LoadComctl32, whereas `hModule` is an `HINSTANCE` which stores the return value of GetModuleHandle. The method `GetProcAddres` has as attributes `(HINSTANCE hModule, PCSTR IpProcName)`, this is why in line 142 it's `hModule` instead of `ninthModule`.

[https://docs.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-getprocaddress](url)

src/System.Windows.Forms/src/System/Windows/Forms/Application.cs
fernandanavamoya: By incorrect spacing do you mean adding a space between the `,` and `(Foundation.PCSTR)ptr`?

src/System.Windows.Forms/src/System/Windows/Forms/Application.cs
dreddy-work: Thank you for explanation here.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
JeremyKuhne: I'd leave it as `nint`. When we really want to reflect pointers we should be using `void*`, etc. and unsafe. That isn't necessary here as we're not currently changing it.

Oleaut32.OCPFIPARAMS should be using `void*` for `ppUnk`. Not necessary to fix unless we're actually making changes to it.

src/System.Windows.Forms/src/System/Windows/Forms/NativeWindow.cs
```diff
@@ -7,8 +7,9 @@
 using System.Globalization;
 using System.Runtime.ConstrainedExecution;
 using System.Runtime.InteropServices;
+using System.Text;
```
elachlan, line 10: This using is breaking the test.

src/System.Windows.Forms/src/System/Windows/Forms/Application.cs
RussKie: Yes.
image

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleHandle.cs
RussKie: We must validate the return, and thow 
image

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
RussKie: This is still outstanding

RussKie:
Why was this merged? There are outstanding unresolved feedback issues.

PR #4197, opened 2020-11-06
Enable visual styles

RussKie:
## Proposed changes

- Enable visual styles when running tests to ensure we test the main user scenarios

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/winforms/pull/4197)

PR #7500, opened 2022-07-27
Remove dependency on RemoteExecutor in tests

### LINKED ISSUE #4500: Remove dependency on `RemoteExecutor` in tests
**Problem description:**

`RemoteExecutor` has been used in a number of tests for the sole purpose of enabling visual styles. However `RemoteExecutor` has proven to be rather unstable (see: https://github.com/dotnet/arcade/issues/5325).

The tests have been update to run with visual styles on by default in https://github.com/dotnet/winforms/pull/4197, however `RemoteExecutor` hasn't been cleaned out.

We need to scan the codebase for `RemoteExecutor` (and "_Skip = "Crash with AbandonedMutexException. See: https://github.com/dotnet/arcade/issues/5325"_") and update those tests, removing `RemoteExecutor`.

willibrandon:
Fixes #4500

## Proposed changes

- Remove dependency on `RemoteExecutor` in tests as it has been proven to be unstable.
- Migrate tests that change global state and that rely on `RemoteExecutor` to the UIIntegration test project.

## Customer Impact

- Test methods should not be skipped.

## Regression? 

- No

## Risk

- Minimal

## Test methodology 

- Ran unit and integration tests and observed for any signs of destabilisation. 

## Test environment(s) 

- 7.0.100-preview.6.22352.1

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7500)


src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Application.ParkingWindowTests.cs
```diff
@@ -0,0 +1,60 @@
...
+            : base(testOutputHelper)
+        {
+        }
+
+        [WinFormsFact]
+        public void ParkingWindow_DoesNotThrowOnGarbageCollecting()
+        {
+            Control.CheckForIllegalCrossThreadCalls = true;
+
+            Form form = InitFormWithControlToGarbageCollect();
```
RussKie, line 22:
```suggestion
            using Form form = InitFormWithControlToGarbageCollect();
```

RussKie (approved):

willibrandon:
Since tests have been updated to run with visual styles on by default, are these 2 tests worth keeping, or should they be modified?

https://github.com/dotnet/winforms/blob/55b296932065022705a6af218ab04a9d90049a28/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ApplicationTests.cs#L79

And the first two assertions in this test are no longer true.

https://github.com/dotnet/winforms/blob/55b296932065022705a6af218ab04a9d90049a28/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ApplicationTests.cs#L90

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ApplicationTests.cs
```diff
@@ -75,31 +69,23 @@ public void Application_CurrentCulture_SetNull_ThrowsArgumentNullException()
             Assert.Throws<ArgumentNullException>("value", () => Application.CurrentCulture = null);
         }
 
-        [WinFormsFact(Skip = "Crash with AbandonedMutexException. See: https://github.com/dotnet/arcade/issues/5325")]
+        [WinFormsFact]
```
RussKie, line 72: To set the context: these tests were written before we switched all tests to run with visual styles enabled (https://github.com/dotnet/winforms/pull/4197), and now they need to assert the opposite - i.e., visual styles are disabled.
Because they use the process-wide setting `EnableVisualStyles` these tests can't live here . That's why we attempted to use RE so each of these tests will get spun in isolation.
These tests must be moved to the UI Integration project.

willibrandon: Thank you for the context. Its clear to me now that the goal is to move all of the RE dependent tests to UI Integration. I misunderstood and thought some could stay as unit if they didn't destablise the suite, but clearly they do.

willibrandon: I tested this on both x86 and x64, but not arm64, and sure enough it fails on arm64.

willibrandon: Speaking of which, is there a better way to test the changes? I've been testing the different configurations individually from the command line, which is not ideal.  Is there a way to reproduce the CI build locally in one go?

Sorry if this information is already covered somewhere.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ListViewInsertionMarkTests.cs
```diff
@@ -0,0 +1,245 @@
...
+    {
+        public ListViewInsertionMarkTests(ITestOutputHelper testOutputHelper)
+            : base(testOutputHelper)
+        {
+        }
+
+        [WinFormsFact]
+        public unsafe void ListViewInsertionMark_AppearsAfterItem_GetInsertMark_Success()
+        {
+            Application.EnableVisualStyles();
```
RussKie, line 22: Visual styles are enabled at the base level
https://github.com/dotnet/winforms/blob/c29016421fb3c4bd9253adb8d317b8064a57a972/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/ControlTestBase.cs#L23-L27

For intents and purposes of the tests we're porting in this PR, we'd likely want to be able to toggle visual styles. Perhaps you could do something like this:
```cs
    [UseDefaultXunitCulture]
    public abstract class ControlTestBase : IAsyncLifetime, IDisposable
    {
        protected ControlTestBase(ITestOutputHelper testOutputHelper, bool enableVisualStyles)
        {
            if (enableVisualStyles)
            {
            }
        }
        protected ControlTestBase(ITestOutputHelper testOutputHelper)
            : this(testOutputHelper, enableVisualStyles: true)
        {
        }
```

```cs
    public class ListViewInsertionMarkVisualStylesOnTests : ControlTestBase
    {
        public ListViewInsertionMarkTests(ITestOutputHelper testOutputHelper)
            : base(testOutputHelper, enableVisualStyles: true)
        {
        }

    public class ListViewInsertionMarkVisualStylesOffTests : ControlTestBase
    {
        public ListViewInsertionMarkTests(ITestOutputHelper testOutputHelper)
            : base(testOutputHelper: enableVisualStyles: false)
        {
        }
```

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/CommonDialogTests.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            using var dialog = new SubCommonDialog
+            {
+                RunDialogResult = runDialogResult
+            };
+            var owner = new Mock<IWin32Window>(MockBehavior.Strict);
+            owner
+                .Setup(o => o.Handle)
+                .Returns(IntPtr.Zero);
+            Assert.Equal(expectedDialogResult, dialog.ShowDialog(owner.Object));
```
RussKie, line 32: [nit/readability] A unit test generally has three section, or follows the [AAA pattern](https://methodpoet.com/aaa-in-unit-testing/). Could you please add a blank line between each logic section of a test to make it easier to read.

In a general case, this would look like (adding comments for visualisation purposes only):
```cs
            // arrange
            using var dialog = new SubCommonDialog
            {
                RunDialogResult = runDialogResult
            };
            var owner = new Mock<IWin32Window>(MockBehavior.Strict);
            owner
                .Setup(o => o.Handle)
                .Returns(IntPtr.Zero);

            // act
            DialogResult result = dialog.ShowDialog(owner.Object)

            // assert
            Assert.Equal(expectedDialogResult, result);
```
...but in this case, the following is acceptable too:
```cs
            using var dialog = new SubCommonDialog
            {
                RunDialogResult = runDialogResult
            };
            var owner = new Mock<IWin32Window>(MockBehavior.Strict);
            owner
                .Setup(o => o.Handle)
                .Returns(IntPtr.Zero);

            Assert.Equal(expectedDialogResult, dialog.ShowDialog(owner.Object));
```

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ListViewGroup.ListViewGroupAccessibleObjectTests.cs
```diff
@@ -0,0 +1,59 @@
...
+{
+    public class ListViewGroup_ListViewGroupAccessibleObjectTests : ControlTestBase
+    {
+        public ListViewGroup_ListViewGroupAccessibleObjectTests(ITestOutputHelper testOutputHelper)
+            : base(testOutputHelper)
+        {
+        }
+
+        [WinFormsFact]
+        public void ListViewGroupAccessibleObject_Bounds_ReturnsCorrectValue()
```
RussKie, line 20: This looks like a regular unit test. Could you please move it to the unit test project alongside other similar tests in SWF.Tests\ListViewGroup_ListViewGroupAccessibleObjectTests?

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ListViewTests.cs
```diff
@@ -110,6 +134,79 @@ public async Task ListView_Group_NavigateKeyboard_SucceedsAsync()
             });
         }
 
+        [WinFormsFact]
+        public unsafe void ListView_Handle_GetWithGroups_Success()
```
RussKie, line 138: Ditto here.

I don't have a clear recollection of events, but IIRC originally the test suite was running with visual styles off, and ListViewGroups are only available with the visual styles on (comctlv6). And the tests for the LVG attempted to employ RE to overcome that limitation.
It doesn't look like these tests are doing anything special nor using the mouse/keyboard, and thus don't really look like they belong to the UI project, and could leave in the unit test project.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/TaskDialogTests.cs
```diff
@@ -0,0 +1,53 @@
...
+{
+    public class TaskDialogTests : ControlTestBase
+    {
+        public TaskDialogTests(ITestOutputHelper testOutputHelper)
+            : base(testOutputHelper)
+        {
+        }
+
+        [WinFormsFact]
+        public void TaskDialog_ShowDialog_SetProperty_SameThread_Success()
```
RussKie, line 18: Ditto here. TaskDialog requires comctrlv6 and visual styles, that weren't originall available in the unit test suite.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ApplicationTests.cs
RussKie: Yes, the tests do take a while to run, and they don't allow to use the workstation.

We have the doc detailing testing [here](https://github.com/dotnet/winforms/blob/main/docs/testing.md). Please holler if you see something that could be improved.

The CI agents unlimately run `build.cmd -test` and `build.cmd -integrationtest`:
https://github.com/dotnet/winforms/blob/c29016421fb3c4bd9253adb8d317b8064a57a972/eng/pipelines/build.yml#L81-L121

You can run bitness-specific tests with additional args, e.g.
```
build.cmd -test /p:Platform=x86 /p:TargetArchitecture=x86
```
Though, you'd need to either install VC++/CMake x86 tools, or remove SWF.Interop.Tests project from the solution (which is easier for local testing):
image


And lastly, you can run only specific tests with `dotnet test`, see https://github.com/dotnet/winforms/blob/main/docs/testing.md#adding-new-unit-tests.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ListViewGroup.ListViewGroupAccessibleObjectTests.cs
willibrandon: If I put this test in the unit test project as-is, it causes the following tests to fail.

`ControlCollection_Add_DifferentThreadValueControl_ThrowsArgumentException`
`ControlCollection_Add_DifferentThreadValueOwner_ThrowsArgumentException`

Instead of throwing `ArgumentException`, they then throw `InvalidOperationException`.

It seems that the call to `Control.CheckForIllegalCrossThreadCalls = true` is the line in this tests that happens to affect the `ControlCollection` tests.

RussKie: Ah yes, that is a process-wide singlton, which can affect other tests.
This may warrant a comment, so we don't try to move these tests into unit tests project in the future.

willibrandon: Will do.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ListViewTests.cs
willibrandon: Thank you for the clarification.

What about the calls to `Application.EnableVisualStyles()` in both tests, should they be removed? And if so, should checks be added to the tests that require `Application.UseVisualStyles`?

```
if (!Application.UseVisualStyles)
{
    return;
}
```

RussKie: Yes, that'd be the right move.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/CommonDialogTests.cs
willibrandon: Thank you very much for the feedback. I see a number of other tests that I wrote in `DragDropHelperTests` that are missing a blank line between each section. My plan is to circle back to those and fix the skipped tests and other issues such as this.

RussKie: No worries. This just makes it slightly easier to read the tests.
Thank you

RussKie:
Trying to page back - is there something that you're waiting for me here?

willibrandon:
No. I've been adjusting to a new schedule, pardon me.

I'm close to having the changes ready and will make it a priority now.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ListViewInsertionMarkTests.cs
willibrandon: I've provided a way for the tests to toggle visual styles like you suggested but for this particular test, does it make sense to test with visual styles off?

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/TabPageTests.cs
```diff
@@ -0,0 +1,60 @@
...
+            yield return new object[] { false, TabAppearance.Buttons, Control.DefaultBackColor };
+            yield return new object[] { false, TabAppearance.FlatButtons, Control.DefaultBackColor };
+            yield return new object[] { false, TabAppearance.Normal, Control.DefaultBackColor };
+        }
+
+        [WinFormsTheory]
+        [CommonMemberData(typeof(CommonTestHelper), nameof(CommonTestHelper.GetBoolTheoryData))]
+        public static void TabPage_BackColor_GetVisualStyles_ReturnsExpected(bool useVisualStyleBackColor)
+        {
+            Application.EnableVisualStyles();
```
RussKie, line 33: This is already enabled in `ControlTestBase`
https://github.com/dotnet/winforms/blob/27120230f23120202d1efdf478b738e1f9a654ae/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/ControlTestBase.cs#L27

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ProfessionalColorTableTests.cs
```diff
@@ -0,0 +1,40 @@
...
+        [InlineData(UserPreferenceCategory.Power)]
+        [InlineData(UserPreferenceCategory.Screensaver)]
+        [InlineData(UserPreferenceCategory.Window)]
+        public void ProfessionalColorTable_ChangeUserPreferences_GetColor_ReturnsExpected(UserPreferenceCategory category)
+        {
+            // Simulate a SystemEvents.UserPreferenceChanged event.
+            var table = new ProfessionalColorTable();
+            Color color = table.ButtonSelectedHighlight;
+            SystemEventsHelper.SendMessageOnUserPreferenceChanged(category);
+            Assert.Equal(color, table.ButtonSelectedHighlight);
```
RussKie, line 37:
```suggestion
            ProfessionalColorTable table = new();
            Color color = table.ButtonSelectedHighlight;

            // Simulate a SystemEvents.UserPreferenceChanged event.
            SystemEventsHelper.SendMessageOnUserPreferenceChanged(category);

            Assert.Equal(color, table.ButtonSelectedHighlight);
```

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ProfessionalColorTableTests.cs
```diff
@@ -0,0 +1,40 @@
...
+namespace System.Windows.Forms.UITests
+{
+    public class ProfessionalColorTableTests : ControlTestBase
+    {
+        public ProfessionalColorTableTests(ITestOutputHelper testOutputHelper)
+            : base(testOutputHelper)
+        {
+        }
+
+        [WinFormsTheory(Skip = "Deadlocks under x86, see: https://github.com/dotnet/winforms/issues/3254")]
```
RussKie, line 19: 💭 May interesting to see if this is still an issue with this move

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewTests.cs
```diff
@@ -1878,98 +1877,98 @@ public static IEnumerable<object[]> Handle_GetWithGroups_TestData()
         {
             foreach (bool showGroups in new bool[] { true, false })
             {
-                yield return new object[] { showGroups, null, HorizontalAlignment.Left, null, HorizontalAlignment.Right, string.Empty, string.Empty, 0x00000021 };
-                yield return new object[] { showGroups, null, HorizontalAlignment.Center, null, HorizontalAlignment.Center, string.Empty, string.Empty, 0x00000012 };
-                yield return new object[] { showGroups, null, HorizontalAlignment.Right, null, HorizontalAlignment.Left, string.Empty, string.Empty, 0x0000000C };
+                yield return new object[] { showGroups, null!, HorizontalAlignment.Left, null!, HorizontalAlignment.Right, string.Empty, string.Empty, LVGA.HEADER_LEFT, LVGA.HEADER_LEFT };
+                yield return new object[] { showGroups, null!, HorizontalAlignment.Center, null!, HorizontalAlignment.Center, string.Empty, string.Empty, LVGA.HEADER_LEFT, LVGA.HEADER_CENTER };
+                yield return new object[] { showGroups, null!, HorizontalAlignment.Right, null!, HorizontalAlignment.Left, string.Empty, string.Empty, LVGA.HEADER_LEFT, LVGA.HEADER_RIGHT };
```
RussKie, line 1882: Better to update the signature to `public static IEnumerable<object?[]> Handle_GetWithGroups_TestData()`. And since this method is only used by `ListView_Handle_GetWithGroups_Success` we could modify the signature as required. For example: 
```cs
public static IEnumerable<(bool, string?, HorizontalAlignment, string?, HorizontalAlignment, string, string, LVGA, LVGA)> Handle_GetWithGroups_TestData()
```

Could even give names to each of the tuple items too :)

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ListViewTests.cs
```diff
@@ -21,6 +22,28 @@ public ListViewTests(ITestOutputHelper testOutputHelper)
         {
         }
 
+        public static IEnumerable<object[]> Handle_GetWithGroups_TestData()
```
RussKie, line 25: What uses this?

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ListViewInsertionMarkVisualStylesOffTests.cs
```diff
@@ -0,0 +1,16 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Xunit.Abstractions;
+
+namespace System.Windows.Forms.UITests
+{
+    public class ListViewInsertionMarkVisualStylesOffTests : ControlTestBase
+    {
+        public ListViewInsertionMarkVisualStylesOffTests(ITestOutputHelper testOutputHelper)
```
RussKie, line 11: Any tests for this?

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ListViewGroupTests.cs
```diff
@@ -0,0 +1,385 @@
...
+            yield return new object[] { "header", HorizontalAlignment.Right, 0x00000004 };
+        }
+
+        [WinFormsFact]
+        public unsafe void ListViewGroup_TitleImageIndex_GetGroupInfo_Success()
+        {
+            var data = new (int, int)[] { (-1, -1), (0, 0), (1, 1), (2, 2) };
+            foreach ((int Index, int Expected) value in data)
+            {
+                Application.EnableVisualStyles();
```
RussKie, line 55: This is redundant as noted above

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ListViewGroupTests.cs
```diff
@@ -0,0 +1,385 @@
...
+            };
+
+            Assert.Equal(1, User32.SendMessageW(listView.Handle, (User32.WM)LVM.GETGROUPINFOBYINDEX, 0, ref lvgroup));
+            Assert.Equal(header, new string(lvgroup.pszHeader));
+            Assert.True(lvgroup.iGroupId >= 0);
+            Assert.Equal(expectedAlign, (int)lvgroup.uAlign);
+        }
+
+        [WinFormsFact]
+        public unsafe void ListViewGroup_Collapse_GetGroupInfo_Success()
```
RussKie, line 302: I'm speculating but we likely haven't used theory tests due to RE - that is, spawn an external process once and then iterate over the theory. But since we're in-proc now, we should use theory as designer - e.g., `WinFormsTheory` and pass `CollapsedState_TestData` as theory data. E.g.,
```
        [WinFormsTheory]
        [MemberData(nameof(CollapsedState_TestData))]
```

This applies to other instances.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/CommonDialogTests.cs
```diff
@@ -0,0 +1,94 @@
...
+    {
+        public CommonDialogTests(ITestOutputHelper testOutputHelper)
+            : base(testOutputHelper)
+        {
+        }
+
+        [WinFormsTheory]
+        [InlineData(true, DialogResult.OK)]
+        [InlineData(false, DialogResult.Cancel)]
+        public void ShowDialog_NonControlOwnerWithVisualStyles_ReturnsExpected(bool runDialogResult, DialogResult expectedDialogResult)
```
RussKie, line 22:
```suggestion
        public void ShowDialog_NonControlOwner_ReturnsExpected(bool runDialogResult, DialogResult expectedDialogResult)
```

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/CommonDialogTests.cs
```diff
@@ -0,0 +1,94 @@
...
+                .Setup(o => o.Handle)
+                .Returns(IntPtr.Zero);
+
+            Assert.Equal(expectedDialogResult, dialog.ShowDialog(owner.Object));
+        }
+
+        [WinFormsTheory]
+        [InlineData(true, DialogResult.OK)]
+        [InlineData(false, DialogResult.Cancel)]
+        public void ShowDialog_ControlOwnerWithVisualStyles_ReturnsExpected(bool runDialogResult, DialogResult expectedDialogResult)
```
RussKie, line 39:
```suggestion
        public void ShowDialog_ControlOwner_ReturnsExpected(bool runDialogResult, DialogResult expectedDialogResult)
```

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/CommonDialogTests.cs
```diff
@@ -0,0 +1,94 @@
...
+            };
+            using var owner = new Control();
+
+            Assert.Equal(expectedDialogResult, dialog.ShowDialog(owner));
+        }
+
+        [WinFormsTheory]
+        [InlineData(true, DialogResult.OK)]
+        [InlineData(false, DialogResult.Cancel)]
+        public void ShowDialog_ControlOwnerWithHandleWithVisualStyles_ReturnsExpected(bool runDialogResult, DialogResult expectedDialogResult)
```
RussKie, line 53:
```suggestion
        public void ShowDialog_ControlOwnerWithHandle_ReturnsExpected(bool runDialogResult, DialogResult expectedDialogResult)
```

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Application.ParkingWindowTests.cs
```diff
@@ -0,0 +1,60 @@
...
+            form.Controls.Add(comboBox);
+            form.Show();
+
+            // Park combobox handle in ParkingWindow.
+            comboBox.Parent = null;
+
+            // Recreate combobox handle to set parent to ParkingWindow.
+            comboBox.DropDownStyle = ComboBoxStyle.DropDownList;
+
+            // Lose the reference to combobox to allow Garbage collecting combobox.
```
RussKie, line 54:
```suggestion
            // Lose the reference to the combobox to allow to garbage collect it.
```

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ApplicationTests.cs
```diff
@@ -0,0 +1,57 @@
...
+        [WinFormsFact]
+        public void Application_EnableVisualStyles_InvokeAfterGettingRenderWithVisualStyles_Success()
+        {
+            Assert.True(Application.UseVisualStyles);
+            Assert.True(Application.RenderWithVisualStyles);
+
+            Application.EnableVisualStyles();
+            Assert.True(Application.UseVisualStyles, "New Visual Styles will not be applied on Winforms app. This is a high priority bug and must be looked into");
+            Assert.True(Application.RenderWithVisualStyles);
+        }
```
RussKie, line 36: We need to run with visual styles on and off, e.g. similar to what you've defined in ListViewInsertionMarkVisualStylesOnTests and ListViewInsertionMarkVisualStylesOffTests.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/DataGridViewHeaderCellTests.cs
```diff
@@ -0,0 +1,382 @@
...
+        [WinFormsFact]
+        public void DataGridViewHeaderCell_MouseDownUnsharesRow_InvokeWithDataGridView_ReturnsExpected()
+        {
+            foreach (object[] testData in MouseDownUnsharesRow_WithDataGridView_TestData())
+            {
+                bool enableHeadersVisualStyles = (bool)testData[0];
+                DataGridViewCellMouseEventArgs e = (DataGridViewCellMouseEventArgs)testData[1];
+                bool expected = (bool)testData[2];
+
+                Application.EnableVisualStyles();
```
RussKie, line 53: As noted above

RussKie (commented):
Progressing nicely

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ProfessionalColorTableTests.cs
willibrandon: If I un-skip the test and change `[WinformsTheory]` to `[Theory]`, the test executes and passes successfully. If I leave it as `[WinformsTheory]`, the test hangs on `SystemEventsHelper.SendMessageOnUserPreferenceChanged(category)`.

RussKie: Thank you for testing. At a glance it doesn't look like there's anything that requires an affinitisation to the UI thread, so I think it'd be safe to go with `[Theory]`.

willibrandon:
Sorry for the delay, I'm seeing some issues with the `DataGridViewHeaderCellTests` that I moved into UITests and un-skipped.

The `DataGridViewHeaderCell` class has some `Debug.Asserts` that are (sometimes?) failing, causing the test host process to crash. Which could explain why I saw these tests fail in CI in an earlier push.

Specifically, the test I'm currently looking into is `DataGridViewHeaderCell_OnMouseLeave_InvokeWithDataGridViewMouseDown_Nop`

And the `Debug.Assert` that is failing and causing the test host process to crash is here
https://github.com/dotnet/winforms/blob/c6dda124b462ad5339088e43a1a544c9b420bff5/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewHeaderCell.cs#L586

ColumnIndex is 0 and DataGridView.MouseDownCellAddress.X is -2.

RussKie:
@willibrandon I got a notification of a comment but I can't seem to locate it, sorry.... It looks like we have a lot of ongoing discussions, and I find it hard to keep track of all of them.  Do you think you could rebase and address any remaining pending comments?

willibrandon:
@RussKie - Someone should take the delete button away from me. Sometimes I'm second guessing myself so much that it is too tempting not to click it. With that said, I have been made acutely aware of how confusing this can be and that I should treat this like email. I will do my best to never click the delete button ever again.

RussKie:
Haha, all good. I do post and then delete sometimes, though generally I try to post another message instead.

willibrandon:
I have some feedback left to address on this, and working towards that now.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ListViewInsertionMarkVisualStylesOffTests.cs
```diff
@@ -0,0 +1,19 @@
...
+            : base(testOutputHelper, enableVisualStyles: false)
+        {
+        }
+
+        [WinFormsFact]
+        public unsafe void ListViewInsertionMark_VisualStylesOff_NotAvailable()
+        {
+            Assert.False(Application.UseVisualStyles);
```
willibrandon, line 16: I could use some help understanding why `Application.UseVisualStyles` is true here even though we pass a value of `false` for `enableVisualStyles` to the base constructor. Still affected by other tests enabling visual styles even though it's been moved to `UITests`?

Did I miss something?

RussKie: Gah... In short, it looks like I underappreciated how the tests from this project are executed. All tests from this project are run in a single process, and, because some tests do enable the visual styles, depending on the order of execution tests (that expect the state of visual styles on or off) will fail. If you run this test in isolation it will succeed.

There are multiple ways to resolve this, and after spending several hours exploring some of them, none appears particlualry painless.
E.g., 
1. we can condition the tests using `ConditionalFact`, but it won't work for STA-bound tests, and we'll have to exend `WinFormsFact` (and `WinFormsTheory`) to be conditional. This likely require an update the Xunit.StaFact library, as some infra is private and doesn't allow an easy overriding.
2. We can pepper the code with something like the following:
    ```xml
      <PropertyGroup Condition="'$(TestVisualStylesOn)' == ''">
        <TestVisualStylesOn>true</TestVisualStylesOn>
      </PropertyGroup>

      <PropertyGroup Condition="'$(TestVisualStylesOn)' == 'true'">
        <DefineConstants>$(DefineConstants);VISUAL_STYLES_ON</DefineConstants>
      </PropertyGroup>
     ```
     ```cs
    #if VISUAL_STYLES_ON
        [WinFormsFact(Skip = "Not supported with Visual Styles enabled")]
    #else
        [WinFormsFact]
    #endif
        [WinFormsFact]
        [VisualStylesEnabled(false)]
        public void Application_VisualStylesOn_must_not_be_enabled()
        {
            Assert.False(Application.UseVisualStyles, "Tests in this configuration must not have VisualStyles enabled.");
            Assert.False(Application.RenderWithVisualStyles);
        }
    ```
    We'd also need to wrap all callsites with:
    ```cs
    #if VISUAL_STYLES_ON
    #endif
    ```
    Running from a command line we'd need to add `/p:TestVisualStylesOn=false` to run without the visual styles.
3. We can add traits, and decorate test classes and/or specific tests with traits. This would require finding a way how to pass `-trait VisualStylesEnabled=false` (example) to the test runner (which is part of microsoft.dotnet.arcade.sdk).
4. We can try to update to Xuni v3, which has `Skip` functionality - https://github.com/xunit/xunit/issues/2073#issuecomment-663933140

I also tried this, but it didn't appear to work - instead of skipping the test it would still fail it:
```cs
    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = false, Inherited = true)]
    public class SkipIfVisualStylesAttribute : BeforeAfterTestAttribute
    {
        public bool Enable { get; set; }

        public override unsafe void Before(MethodInfo methodUnderTest)
        {
#if VISUAL_STYLES_ON
            if (Enable)
                throw new SkipException("Not supported with Visual Styles disabled");
#else
            if (!Enable)
                throw new SkipException("Not supported with Visual Styles enabled");
#endif
        }
    }
```

RussKie: So as a quick get out of jail card - we can wrap all tests that expect visual styles to be off with
```cs
#if VISUAL_STYLES_OFF
           // tests
#endif
```
...and then we can finish this work at a later date.

merriemcgaw:
@willibrandon sorry this fell through the cracks! I'm closing this now but if you're willing to pick it back up just holler and we can re-open it for you.

willibrandon:
My apologies. Sometimes I bite off more than I can chew 😩. These tests can be very finicky, and time-consuming.

With that said, I don't want to give up on this, especially if #4500 is still an issue. I just might not be the best person to tackle this one yet. I'll see if I can get back to it, and if so, I will holler.

PR #7505, opened 2022-07-27
Address over release in DragDropFormat tests

AaronRobinsonMSFT:
There is an over release in `DragDropFormatTests.cs`. The main issue is the `DragDropFormat` class has a "copy" semantic that is default in all tests. Since there are no copies being made, ownership of the passed in artifacts is implied so calling [`DragDropFormat.Dispose()`](https://github.com/dotnet/winforms/blob/de1f12d7404f627539ab81cc98a054a12d8020aa/src/System.Windows.Forms/src/System/Windows/Forms/DragDropFormat.cs#L115-L130) means they will be released then and there is no need to call `Ole32.ReleaseStgMedium()` as well.

https://github.com/dotnet/winforms/blob/de1f12d7404f627539ab81cc98a054a12d8020aa/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DragDropFormatTests.cs#L63

https://github.com/dotnet/winforms/blob/de1f12d7404f627539ab81cc98a054a12d8020aa/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DragDropFormatTests.cs#L87

https://github.com/dotnet/winforms/blob/de1f12d7404f627539ab81cc98a054a12d8020aa/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DragDropFormatTests.cs#L130

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7505)

### DIFF
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DragDropFormatTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DragDropFormatTests.cs
@@ -70,7 +70,6 @@ namespace System.Windows.Forms.Tests
             }
             finally
             {
-                Ole32.ReleaseStgMedium(ref medium);
                 dragDropFormat?.Dispose();
             }
         }
@@ -80,12 +79,11 @@ namespace System.Windows.Forms.Tests
         public void DragDropFormat_Set_GetData_ReturnsExpected(FORMATETC formatEtc, STGMEDIUM medium)
         {
             DragDropFormat dragDropFormat = default;
-            STGMEDIUM data = default;
 
             try
             {
                 dragDropFormat = new DragDropFormat(formatEtc.cfFormat, medium, copyData: false);
-                data = dragDropFormat.GetData();
+                STGMEDIUM data = dragDropFormat.GetData();
                 Assert.Equal(medium.pUnkForRelease, data.pUnkForRelease);
                 Assert.Equal(medium.tymed, data.tymed);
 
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DragDropFormatTests.cs
```diff
@@ -168,11 +169,16 @@ public void DragDropFormat_Set_RefreshData_ReturnsExpected(FORMATETC formatEtc,
...
-                Ole32.ReleaseStgMedium(ref medium);
-                Ole32.ReleaseStgMedium(ref data);
-                Ole32.ReleaseStgMedium(ref dataRefresh);
                 dragDropFormat?.Dispose();
             }
+
+            static IntPtr ComPtrType(IntPtr ptr)
+            {
+                if (ptr != IntPtr.Zero)
+                    Marshal.AddRef(ptr);
```
lonitra, line 178: Reading the docs for this method - do we need to do a `Marshal.Release` for this at some point?

AaronRobinsonMSFT: Yes, that does need to happen if we are explicitly managing the lifetime here. However, the `DataDropFormat` class is handling that for us through the call to `Ole32.ReleaseStgMedium(ref _medium);`.

JeremyKuhne (approved):
Thanks for the fix here!

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DragDropFormatTests.cs
lonitra: Ah I see, thank you!

lonitra (approved):
Nice catch!

willibrandon:
Thank you!

PR #7514, opened 2022-07-29
Convert threads to CsWin32 PInvoke

fernandanavamoya:
Converted these threads:
- ActivateActCtx
- DeactivateActCtx
- GetShortPathName
- GetStartupInfo
- GetSystemPowerStatus
- MultiByteToWideChar
- WideChartToMultiByte

Deleted Interop.OpenProcess.cs and Interop.SetThreadLocale.cs, since they had zero references.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7514)


fernandanavamoya:
cc: @AArnott

src/System.Windows.Forms.Primitives/src/NativeMethods.txt
```diff
@@ -1,14 +1,17 @@
 ACTCTX_FLAG_*
+ActivateActCtx
 AdjustWindowRectEx
 AdjustWindowRectExForDpi
 BeginPaint
 BitBlt
 CloseEnhMetaFile
 CombineRgn
+CP*
```
dreddy-work, line 9: Q: does this bring everything starts with `CP`? or something else?

src/System.Windows.Forms/tests/InteropTests/PropertyGridTests.cs
```diff
@@ -99,7 +99,8 @@ private unsafe void ExecuteWithActivationContext(string applicationManifest, Act
 
         try
         {
-            if (Kernel32.ActivateActCtx(handle, out var cookie).IsFalse())
+            nuint cookie;
+            if (PInvoke.ActivateActCtx(handle, &cookie) == false)
```
dreddy-work, line 103:
```suggestion
            if (!PInvoke.ActivateActCtx(handle, &cookie))
```

src/System.Windows.Forms/tests/InteropTests/PropertyGridTests.cs
```diff
@@ -110,7 +111,7 @@ private unsafe void ExecuteWithActivationContext(string applicationManifest, Act
             }
             finally
             {
-                if (Kernel32.DeactivateActCtx(0, cookie).IsFalse())
+                if (PInvoke.DeactivateActCtx(0, cookie) == false)
```
dreddy-work, line 114:
```suggestion
                if (!PInvoke.DeactivateActCtx(0, cookie))
```

src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
```diff
@@ -6057,13 +6058,13 @@ private void WmExitSizeMove()
         private void WmCreate(ref Message m)
         {
             base.WndProc(ref m);
-            var si = new Kernel32.STARTUPINFOW();
-            Kernel32.GetStartupInfoW(ref si);
+            var si = new STARTUPINFOW();
```
AArnott, line 6061: tip: You don't have to initialize the struct any more since it's passed to an `out` parameter. You can just make this `STARTUPINFOW si;`

src/System.Windows.Forms/tests/InteropTests/PropertyGridTests.cs
AArnott: Why not use this syntax?
```cs
if (!PInvoke.ActivateActCtx(handle, out nuint cookie))
```

In particular, you can declare the variable as part of the `out` parameter, and you don't need the pointer syntax.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/ThemingScope.cs
```diff
@@ -36,14 +35,15 @@ private unsafe static bool IsContextActive()
         ///  on the stack even if one is already present. In such cases, this method helps - you get to manage
         ///  the cookie yourself though.
         /// </summary>
-        public static IntPtr Activate(bool useVisualStyles)
+        public unsafe static nint Activate(bool useVisualStyles)
         {
-            if (IsContextActiveButNotCreated(useVisualStyles) && Kernel32.ActivateActCtx(s_hActCtx, out IntPtr userCookie).IsTrue())
+            nuint userCookie;
+            if (IsContextActiveButNotCreated(useVisualStyles) && PInvoke.ActivateActCtx((HANDLE)s_hActCtx, &userCookie))
```
AArnott, line 41: Why use the `&userCookie` argument when you can use `out userCookie`? Or even better, `out nuint userCookie`?

AArnott (approved):

src/System.Windows.Forms.Primitives/src/NativeMethods.txt
RussKie: Is there a risk that'd this import more than we expect?

fernandanavamoya: > Q: does this bring everything starts with `CP`? or something else?

Correct. It calls libraries starting with `CP`. However, looking back, I believe it would be better to call `CP_` instead, this way it only calls the library that's being used: `CP_ACP`. You can also only call `CP_ACP`. However, since I don't know if I'll need to use this library later on, I decided to call `CP_` instead of only `CP_ACP`.

fernandanavamoya: > Is there a risk that'd this import more than we expect?

Since we can specify the imports, I do not believe this should be a problem. Or could there be a risk, @AArnott @JeremyKuhne?

src/System.Windows.Forms/tests/InteropTests/PropertyGridTests.cs
fernandanavamoya: When declaring the variable this way it takes `handle` as a `InteropServices.SafeHandle` instead of `Foundation.HANDLE` and we don't want to use the SafeHandle.

src/System.Windows.Forms.Primitives/src/NativeMethods.txt
JeremyKuhne: `CP_*` would be better. If we aren't using any other definitions the full name would be best: `CP_ACP`.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/ThemingScope.cs
JeremyKuhne: We don't want the `SafeHandle`. https://github.com/microsoft/CsWin32/issues/605

JeremyKuhne (approved):
Looks good!

PR #7515, opened 2022-07-30
Add Tests to StronglyTypedResourceBuilder/ResXDataNode

lonitra:
- Ensure type `byte[]` from `System.Private.Corelib` assembly is handled correctly in `ResXDataNode.GenerateObjectFromDataNodeInfo` and expand its test suite

- Add Bitmap/Icon/Audio/Txt File tests: 

In Framework, the main scenario resources that are added to the .resx file are strings, images, icons, audios, and files. The tests added check if the `StronglyTypedResourceBuilder`, no matter if given a resource list or resx file, is able to spit out the correct resource type. The tests also validate the resulting value grabbed from the property. 

Note that the resource property types (seen in `Resources.designer.cs` file) for images and icons in Core are different from Framework. In Core, image and icon are type `byte[]` while in Framework they are `Bitmap` and `Icon` respectively. Thus, in the image/icon tests there is additional checking for whether the resulting `byte[]` can be converted to type `Icon`/`Bitmap`. Issue will be created for this so that images/icons can actually be strongly typed in core.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7515)


src/System.Windows.Forms.Design/tests/UnitTests/System/Resources/Tools/StronglyTypedResourceBuilderTests.cs
```diff
@@ -338,36 +339,120 @@ public static void StronglyTypedResourceBuilder_Create_BitmapResource_FromFile()
...
+        Bitmap resourceBitmap = Assert.IsType<Bitmap>(converter.ConvertFrom(resourceBytes));
+        Assert.Equal(bitmap.GetPixel(0, 0), resourceBitmap.GetPixel(0, 0));
+        Assert.Equal(bitmap.Size, resourceBitmap.Size);
+    }
+
+    [Fact]
+    public static void StronglyTypedResourceBuilder_Create_BitmapResource_FromFileRef()
+    {
+        ResXFileRef fileRef = new(@"Resources\Image1.png",
+            "System.Byte[], mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089");
```
dreddy-work, line 405: string `mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089` can be made const in the test class and reuse across tests.

src/System.Windows.Forms.Design/tests/UnitTests/System/Resources/Tools/StronglyTypedResourceBuilderTests.cs
```diff
@@ -338,36 +339,120 @@ public static void StronglyTypedResourceBuilder_Create_BitmapResource_FromFile()
...
+        resourceWriter.Generate();
+
+        resourceStream.Position = 0;
+
+        Type type = CodeDomCompileHelper.CompileClass(compileUnit, "Resources", "Namespace", resourceStream);
+        Assert.NotNull(type);
+        var imageProperty = type.GetProperty("Image1");
+        Assert.NotNull(imageProperty);
+        byte[] resourceBytes = Assert.IsType<byte[]>(imageProperty.GetValue(obj: null));
+        Bitmap resourceBitmap = Assert.IsType<Bitmap>(converter.ConvertFrom(resourceBytes));
```
dreddy-work, line 396: if this is a new bitmap instance, we need to dispose before test ends. Can you take a look at other instances in these new tests and see they are disposed as necessary?

src/System.Windows.Forms.Design/tests/UnitTests/System/Resources/Tools/StronglyTypedResourceBuilderTests.cs
```diff
@@ -338,36 +339,120 @@ public static void StronglyTypedResourceBuilder_Create_BitmapResource_FromFile()
...
+        var compileUnit = StronglyTypedResourceBuilder.Create(
+            resourceList: values,
+            baseName: "Resources",
+            generatedCodeNamespace: "Namespace",
+            s_cSharpProvider,
+            internalClass: false,
+            out _);
+
+        MemoryStream resxStream = new();
+        ResXResourceWriter resxWriter = new(resxStream);
```
dreddy-work, line 420: i see some inconsistency in disposing ResXResourcewritters. Some are used with `using `and others are not. Can you cross check  these in new tests added in this PR?

src/System.Windows.Forms.Design/tests/UnitTests/System/Resources/Tools/StronglyTypedResourceBuilderTests.cs
```diff
@@ -382,19 +467,304 @@ public static void StronglyTypedResourceBuilder_Create_BitmapResource_FromMemory
...
+
+        while (enumerator.MoveNext())
+        {
+            resourceWriter.AddResource((string)enumerator.Key, enumerator.Value);
+        }
+
+        resourceWriter.Generate();
+        resourceStream.Position = 0;
+        var converter = TypeDescriptor.GetConverter(typeof(Icon));
+        Icon icon = new(@"Resources\Icon1.ico");
```
dreddy-work, line 573: same here. use `using`  to dispose.

src/System.Windows.Forms.Design/tests/UnitTests/System/Resources/Tools/StronglyTypedResourceBuilderTests.cs
```diff
@@ -338,36 +339,120 @@ public static void StronglyTypedResourceBuilder_Create_BitmapResource_FromFile()
...
+
+        resourceWriter.Generate();
+
+        resourceStream.Position = 0;
+
+        Type type = CodeDomCompileHelper.CompileClass(compileUnit, "Resources", "Namespace", resourceStream);
+        Assert.NotNull(type);
+        var imageProperty = type.GetProperty("Image1");
+        Assert.NotNull(imageProperty);
+        byte[] resourceBytes = Assert.IsType<byte[]>(imageProperty.GetValue(obj: null));
```
RussKie, line 395: This looks like an exact replica of the code above and below. Can you please factor out the repeating [arrangement logic](https://methodpoet.com/aaa-in-unit-testing/#Arrange_or_Preparation) in a helper method to reduce duplication.
Whilst there are different school of thoughts on how to write a good test, in this case duplication of code doesn't carry a great deal of value, but introduces a maintenance overhead.

RussKie (changes requested):

src/System.Windows.Forms.Design/tests/UnitTests/System/Resources/Tools/StronglyTypedResourceBuilderTests.cs
```diff
@@ -338,36 +339,120 @@ public static void StronglyTypedResourceBuilder_Create_BitmapResource_FromFile()
 
         resourceWriter.Generate();
         resourceStream.Position = 0;
+        var converter = TypeDescriptor.GetConverter(typeof(Bitmap));
+        Bitmap bitmap = (Bitmap)Image.FromFile(@"Resources\Image1.png");
```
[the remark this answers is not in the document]
lonitra: Looks like `ResourceWriter`/`ResXResourceWriter`/`ResXResourceReader` dispose of streams that are passed to them upon construction -- do we still need to add `using` statements for these streams?

src/System.Windows.Forms/src/System/Resources/ResXDataNode.cs
```diff
@@ -431,7 +431,8 @@ private object GenerateObjectFromDataNodeInfo(DataNodeInfo dataNodeInfo, ITypeRe
                     {
                         result = null;
                     }
-                    else if (typeName.IndexOf("System.Byte[]") != -1 && typeName.IndexOf("mscorlib") != -1)
+                    else if (type == typeof(byte[]) ||
+                        (typeName.Contains("System.Byte[]") && (typeName.Contains("mscorlib") || typeName.Contains("System.Private.CoreLib"))))
```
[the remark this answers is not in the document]
lonitra: I'm not sure I understand here. We want to do a fast check if `type == typeof(byte[])` , but there is a comment below (line 437) stating that that this may not always work, so in the case that it doesn't for the reason mentioned in the comment, we need to check using the slower way which is where we check that `typeName` contains System.Byte[] and mscorlib or it contains System.Byte[] and private.corlib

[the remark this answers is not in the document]
RussKie: I initially though the expression looked like `A || (B && C || D)`. I had a loooong look again, and I can see that it's actually `A || (B && (C || D))`.
Apologies.

src/System.Windows.Forms.Design/tests/UnitTests/System/Resources/Tools/StronglyTypedResourceBuilderTests.cs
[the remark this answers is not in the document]
RussKie: IMO that's an implementation detail - unless you peak inside you can't be sure. The docs aren't stating that the supplied stream will be disposed of either, so to be on a safe side I suggest always dispose disposable objects.

src/System.Windows.Forms.Design/tests/UnitTests/System/Resources/Tools/StronglyTypedResourceBuilderTests.cs
```diff
@@ -326,52 +359,174 @@ public static void StronglyTypedResourceBuilder_Create_BitmapResource_FromFile()
...
+    }
+
+    private static void validateResultIcon(PropertyInfo iconPropertyInfo, Icon expected, TypeConverter converter)
+    {
+        byte[] resourceByte = Assert.IsType<byte[]>(iconPropertyInfo.GetValue(obj: null));
+        using Icon resourceIcon = Assert.IsType<Icon>(converter.ConvertFrom(resourceByte));
+        Assert.Equal(expected.Size, resourceIcon.Size);
+    }
+
+[Fact]
```
RussKie, line 432: Bad indentation

src/System.Windows.Forms.Design/tests/UnitTests/System/Resources/Tools/StronglyTypedResourceBuilderTests.cs
```diff
@@ -326,52 +359,174 @@ public static void StronglyTypedResourceBuilder_Create_BitmapResource_FromFile()
...
+        resxWriter.AddResource(new ResXDataNode("Image1", fileRef));
+        resxWriter.Generate();
+        resxStream.Position = 0;
+        using ResXResourceReader reader = new(resxStream);
+        var imagePropertyInfo = getPropertyInfo(reader.GetEnumerator(), compileUnit, "Image1");
+        using Bitmap expected = (Bitmap)Image.FromFile(@"Resources\Image1.png");
+        validateResultBitmap(imagePropertyInfo, expected, TypeDescriptor.GetConverter(typeof(Bitmap)));
+    }
+
+    private static void validateResultIcon(PropertyInfo iconPropertyInfo, Icon expected, TypeConverter converter)
```
RussKie, line 425: Incorrect capitalisation

src/System.Windows.Forms.Design/tests/UnitTests/System/Resources/Tools/StronglyTypedResourceBuilderTests.cs
```diff
@@ -307,12 +312,40 @@ public static void StronglyTypedResourceBuilder_Create_StringResource_FromResxDa
         Assert.Equal("TestValue", (string)nameProperty.GetValue(obj: null));
     }
 
+    private static PropertyInfo getPropertyInfo(IDictionaryEnumerator enumerator,
```
RussKie, line 315: Ditto

Can you please move the helper methods after the tests? It _can_ kind of make sense if a method is directly near the callsite, but this (and other) helper method is used multiple times.

src/System.Windows.Forms.Design/tests/UnitTests/System/Resources/Tools/StronglyTypedResourceBuilderTests.cs
```diff
@@ -307,12 +312,40 @@ public static void StronglyTypedResourceBuilder_Create_StringResource_FromResxDa
...
+        CodeCompileUnit compileUnit, string propertyName)
```
RussKie, line 316: Please keep either all args on the same line, or each one on the separate line

src/System.Windows.Forms.Design/tests/UnitTests/System/Resources/Tools/StronglyTypedResourceBuilderTests.cs
```diff
@@ -381,20 +536,100 @@ public static void StronglyTypedResourceBuilder_Create_BitmapResource_FromMemory
...
+
+        using ResXResourceReader reader = new(temp.Path);
+        validateResultAudio(getPropertyInfo(reader.GetEnumerator(), compileUnit, "Audio1"));
+    }
+
+    [Fact]
+    public static void StronglyTypedResourceBuilder_Create_AudioResource_FromFileRef()
+    {
+        ResXFileRef fileRef = new(@"Resources\Audio1.wav",
+            $"System.IO.MemoryStream, {TypeAssembly}");
```
RussKie, line 613: This line is well under 120 chars, please keep it on the same line. Ditto other similar instances.

Please install the [editor guidelines extension](https://github.com/dotnet/winforms/blob/main/docs/developer-guide.md#machine-setup), which will help you in such situations.

src/System.Windows.Forms.Design/tests/UnitTests/System/Resources/Tools/StronglyTypedResourceBuilderTests.cs
```diff
@@ -326,52 +359,174 @@ public static void StronglyTypedResourceBuilder_Create_BitmapResource_FromFile()
...
+
+        MemoryStream resxStream = new();
+        using ResXResourceWriter resxWriter = new(resxStream);
+        resxWriter.AddResource(new ResXDataNode("Icon1", fileRef));
+        resxWriter.Generate();
+        resxStream.Position = 0;
+        using ResXResourceReader reader = new(resxStream);
+        var iconPropertyInfo = getPropertyInfo(reader.GetEnumerator(), compileUnit, "Icon1");
+        using Icon expected = new(@"Resources\Icon1.ico");
+        validateResultIcon(iconPropertyInfo, expected, TypeDescriptor.GetConverter(typeof(Icon)));
```
RussKie, line 510: Following the [AAA pattern](https://methodpoet.com/aaa-in-unit-testing/) (i.e., arrange, act, assert), am I reading this correctly?
```cs
	// arrange
        ResXFileRef fileRef = new(@"Resources\Icon1.ico",
            $"System.Byte[], {TypeAssembly}");
        Hashtable values = new()
        {
            { "Icon1", new ResXDataNode("Icon1", fileRef) }
        };

	// act
        var compileUnit = StronglyTypedResourceBuilder.Create(...);

	// assert
        MemoryStream resxStream = new();
        using ResXResourceWriter resxWriter = new(resxStream);
        // .., the rest
```

RussKie (commented):
Some dispose and styling issues, other than that - 👍

src/System.Windows.Forms.Design/tests/UnitTests/System/Resources/Tools/StronglyTypedResourceBuilderTests.cs
lonitra: Yup, that's correct

RussKie (approved):

PR #7529, opened 2022-08-03
Hiding TabStop property of PrintPreviewControl from Designer and setting the default value as false.

### LINKED ISSUE #3678: The PrintPreviewControl control have no dotted rectangle when using Keyboard focus on it
Repro:
1. create a core winforms application
2. add a text box and a print preview control to a form
3. run the application and TAB through the form
result: TAB key moves focus from the text box to the print preview control even though this control is not editable.

Let's consider a comprehensive fix in Core 6.0 - 
1. set default value for TabStop property to false:
```
        [Browsable(false)]
        [DefaultValue(false)]   << add this line
        new public bool TabStop
        {
            get => base.TabStop;
            set => base.TabStop = value;
        }
```
2. if the developer sets TabStop to true, we should draw the focus rectangle

_Originally posted by @Tanya-Solyanik in https://github.com/dotnet/winforms/issues/3660#issuecomment-666847840_

NikitaSemenovAkvelon:
Fixes #3678

## Proposed changes

- Set the default value of TabStop to false.
- Hided TabStop property from Designer.

## Customer Impact

- Developer won't see the TabStop property in Property view by designer.
- User won't tab to PrintPreviewControl.

## Regression? 

- No

## Risk

- Minimal.

### Before

image
image

### After

MicrosoftTeams-image
MicrosoftTeams-image (1)

## Test methodology 

- Manually via VS and WinForms app.
 

## Test environment(s) 

- .NET 6.0.7

The suggested change in #3678 didn't fix the behavior. DefaultValueAttribute responsible for will be a property set explicitly in Form.Designer.cs or not. I'd looked for how it works in Label and copied a fix from them.
Also, the fix covered case 2 from the issue, because it became unreachable. But if you are strongly sure that we need to add borders, it won't be done easily. Now there is no PrintPreviewControlRenderer class and drawing is processed inside Control.WmPaint(ref Message m) when control draws the first time or resource inside is changed.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7529)


src/System.Windows.Forms/src/System/Windows/Forms/Printing/PrintPreviewControl.cs
```diff
@@ -271,6 +272,15 @@ public int StartPage
             remove => Events.RemoveHandler(EVENT_STARTPAGECHANGED, value);
         }
 
+        [DefaultValue(false)]
+        [Browsable(false)]
+        [EditorBrowsable(EditorBrowsableState.Never)]
```
Tanya-Solyanik, line 277: I think you want to add `[DispId((int)Ole32.DispatchID.TABSTOP)]` attribute. @JeremyKuhne - when do we use that attribute?

src/System.Windows.Forms/src/System/Windows/Forms/Printing/PrintPreviewControl.cs
```diff
@@ -271,6 +272,15 @@ public int StartPage
             remove => Events.RemoveHandler(EVENT_STARTPAGECHANGED, value);
         }
 
+        [DefaultValue(false)]
+        [Browsable(false)]
```
Tanya-Solyanik, line 276: If `TabStop` property is not browsable, it makes sense to make `TabIndex` not browsable to intellisense and property browser as well.

Tanya-Solyanik (commented):
I think we should discuss potential breaks caused by this change within the team and move it to .NET 8 for now.
This fix changes meaning of code-behind generated by the designer. Previously no TabStop property in the .designer.cs file meant `TabStop == true`. Now it would mean `false`.

src/System.Windows.Forms/src/System/Windows/Forms/Printing/PrintPreviewControl.cs
```diff
@@ -56,6 +56,7 @@ public PrintPreviewControl()
             Size = new Size(100, 100);
             SetStyle(ControlStyles.ResizeRedraw, false);
             SetStyle(ControlStyles.Opaque | ControlStyles.OptimizedDoubleBuffer, true);
+            TabStop = false;
```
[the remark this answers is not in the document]
NikitaSemenovAkvelon: The reason why I used TabStop here is that Label did by the same way (https://github.com/NikitaSemenovAkvelon/winforms/blob/main/src/System.Windows.Forms/src/System/Windows/Forms/Label.cs#L84). Moreover, few controls such as PictureBox, Panel, GroupBox and other have identical code.

src/System.Windows.Forms/src/System/Windows/Forms/Printing/PrintPreviewControl.cs
NikitaSemenovAkvelon: I absolutely agree with you, but I've done this like other uneditable controls. For example, Label contains the TabIndex property. However, if you think that we should hide TabIndex property too, I'll be glad to do it.

Tanya-Solyanik: No, let's be consistent with other controls. On the second thought it might be used by the taborder overlay in the VS

src/System.Windows.Forms/src/System/Windows/Forms/Printing/PrintPreviewControl.cs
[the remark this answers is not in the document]
Tanya-Solyanik: This is unfortunate because it raises a change event from the constructor where property did not have a chance to change. Ok, let's be consistent with other controls

Tanya-Solyanik (approved):

Tanya-Solyanik (approved):

PR #7534, opened 2022-08-03
Convert CreatePen to Cswin32

lonitra:
- Add conversions for `Gdi.HPEN` <-> `Interop.HPEN` and `Gdi.HPEN` -> `Interop.HGDIOBJ`
- Some clean up

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7534)


src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DesignerUtils.cs
```diff
@@ -27,12 +27,17 @@ internal static class DesignerUtils
         //brush used to draw a 'hover' state over a designer action glyph
         private static SolidBrush s_hoverBrush = new SolidBrush(Color.FromArgb(50, SystemColors.Highlight));
         //brush used to draw the resizeable selection borders around controls/components
-        private static HatchBrush s_selectionBorderBrush = new HatchBrush(HatchStyle.Percent50, SystemColors.ControlDarkDark, Color.Transparent);
+        private static HatchBrush s_selectionBorderBrush =
+            new HatchBrush(HatchStyle.Percent50, SystemColors.ControlDarkDark, Color.Transparent);
```
dreddy-work, line 31:
```suggestion
        private static HatchBrush s_selectionBorderBrush =
            new (HatchStyle.Percent50, SystemColors.ControlDarkDark, Color.Transparent);
```

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DesignerUtils.cs
```diff
@@ -178,10 +184,10 @@ public static void SyncBrushes()
             s_grabHandleFillBrush = Gdi32.CreateSolidBrush(ColorTranslator.ToWin32(SystemColors.ControlText));
 
             Gdi32.DeleteObject(s_grabHandlePenPrimary);
-            s_grabHandlePenPrimary = Gdi32.CreatePen(Gdi32.PS.SOLID, 1, ColorTranslator.ToWin32(SystemColors.ControlText));
+            s_grabHandlePenPrimary = PInvoke.CreatePen(Gdi.PEN_STYLE.PS_SOLID, 1, (uint)ColorTranslator.ToWin32(SystemColors.ControlText));
```
dreddy-work, line 187: we adapted adding parameter names for constants in the parameter list. It would be nice to add those for the lines you are touching here and other parts of this PR. Or do you see any issues with that in PInvokes? ex:
```suggestion
            s_grabHandlePenPrimary = PInvoke.CreatePen(Gdi.PEN_STYLE.PS_SOLID, cWidth: 1, (uint)ColorTranslator.ToWin32(SystemColors.ControlText));
```

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/OleDragDropHandler.cs
```diff
@@ -537,7 +539,7 @@ private static void DrawReversibleFrame(IntPtr handle, Rectangle rectangle, Colo
             }
 
             using var dc = new User32.GetDcScope(handle);
-            using var pen = new Gdi32.ObjectScope(Gdi32.CreatePen(Gdi32.PS.SOLID, 2, ColorTranslator.ToWin32(backColor)));
+            using var pen = new Gdi32.ObjectScope(PInvoke.CreatePen(Gdi.PEN_STYLE.PS_SOLID, 2, (uint)ColorTranslator.ToWin32(backColor)));
```
dreddy-work, line 542: when return type is not obvious, we adapted declaring it explicitly with type info

JeremyKuhne: `var foo = new SomeType()` is obvious. We're not calling a method here. It is no different than `var foo = new List<T>()`.  We could flip this to the other side now that we have the implied constructor syntax- that feature didn't exist when I wrote this originally.

The only super hard requirement for `var` is that you don't use it for intrinsic types. We don't accept `var foo = 42;`. Every other usage is left to your better judgement on what provides more clarity.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DesignerUtils.cs
JeremyKuhne: We only need to add them where it is not super clear. That typically means where you explicitly pass `true`, `false`, or `null`. Here it is pretty obvious that you're referring to a width.

JeremyKuhne (approved):
Looks good!

JeremyKuhne:
It looks like the x64 run was aborted due to time out? @RussKie I looked at the dump and I didn't see anything hung up.  https://dev.azure.com/dnceng/public/_build/results?buildId=1921188&view=results  What is our upper threshold on the runs?

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DesignerUtils.cs
dreddy-work: Are we saying numeric constants does not need parameter names here? I may not agree on the part that "it is super clear" here.

dreddy-work (approved):

PR #7541, opened 2022-08-04
Convert CloseHandle/GetLocaleInfoEx/LoadLibrary to CsWin32

fernandanavamoya:
Created
- PInvoke.CloseHandle.cs
- PInvoke.LoadLibrary

Use SafeHandle instead of Foundation.HANDLE
- LoadLibraryEx

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7541)


src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.LoadLibrary.cs
```diff
@@ -0,0 +1,76 @@
...
+            {
+                string customPath = Path.Join(startupPath, Libraries.Comctl32);
+                Debug.Assert(Path.IsPathFullyQualified(customPath));
+
+                fixed (char* lpLibFileName = customPath)
+                {
+                    if (Path.IsPathFullyQualified(customPath))
+                    {
+                        // OS will validate the path for us
+                        IntPtr result = LoadLibraryEx(lpLibFileName, (HANDLE)0, 0);
```
JeremyKuhne, line 29: We should pass back `HINSTANCE` from these instead of letting them cast to `IntPtr` or `nint`. Now that we have "strong" handle types, we should try to carry them through the system and only convert to `IntPtr` when absolutely needed.

src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs
```diff
@@ -779,7 +779,7 @@ internal static bool GetAllowQuit()
             /// <summary>
             ///  Retrieves the handle to this thread.
             /// </summary>
-            internal IntPtr GetHandle() => _handle;
+            nint IHandle.Handle => _handle;
```
JeremyKuhne, line 782: This should be `internal Handle => _handle;`. When you do it this way you have to cast it to the interface to use it.

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -3907,7 +3907,7 @@ private unsafe void WaitForWaitHandle(WaitHandle waitHandle)
                 return;
             }
 
-            nint threadHandle = ctx.GetHandle();
+            nint threadHandle = ((IHandle)ctx).Handle;
```
JeremyKuhne, line 3910: See the comment in `ThreadContext`, this cast shouldn't be necessary.

src/System.Windows.Forms/src/System/Windows/Forms/NativeWindow.cs
```diff
@@ -111,7 +111,7 @@ internal unsafe void ForceExitMessageLoop()
                 {
                     uint id = User32.GetWindowThreadProcessId(handle, out uint lpdwProcessId);
                     Application.ThreadContext ctx = Application.ThreadContext.FromId(id);
-                    nint threadHandle = (ctx is null ? 0 : ctx.GetHandle());
+                    nint threadHandle = (ctx is null ? 0 : ((IHandle)ctx).Handle);
```
JeremyKuhne, line 114: Same here- this cast can be removed when you fix `ThreadContext`.

JeremyKuhne (changes requested):
Should try to keep things as their handle type and can skip using explicit interface implementation for .Handle. Otherwise looks good.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.LoadLibrary.cs
fernandanavamoya: By casting it to an `IntPtr` or `nint` only when necessary, do you mean we should also change the return type of the method to `HINSTANCE`?

JeremyKuhne: > do you mean we should also change the return type of the method to HINSTANCE?

Yes

fernandanavamoya:
cc: @AArnott

dreddy-work (approved):
LGTM

JeremyKuhne (approved):

PR #7542, opened 2022-08-04
Convert CreateX to Cswin32

lonitra:
Convert `CreateDIBSection`,`CreateRectRgn`,`CreateCompatibleDC`,`CreateHalftonePalette`
Add conversion for `Gdi.CreatedHDC` -> `HDC` and `Gdi.HPALETTE` <-> `HPALETTE`
Move `RegionScope`, `CreateDCScope`, and `RegionTests` files

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7542)


src/System.Windows.Forms/src/System/Windows/Forms/ControlPaint.cs
```diff
@@ -166,62 +166,65 @@ public unsafe static IntPtr CreateHBitmap16Bit(Bitmap bitmap, Color background)
...
 
             byte[] bitmapInfoBuffer = ArrayPool<byte>.Shared
-                .Rent(checked((int)(sizeof(Gdi32.BITMAPINFOHEADER) + (sizeof(Gdi32.RGBQUAD) * entryCount))));
+                .Rent(checked((int)(sizeof(Gdi.BITMAPINFOHEADER) + (sizeof(Gdi.RGBQUAD) * entryCount))));
 
             // Create a DIB based on the screen DC to write into with a halftone palette
             fixed (byte* bi = bitmapInfoBuffer)
             {
-                *((Gdi32.BITMAPINFOHEADER*)bi) = new Gdi32.BITMAPINFOHEADER
+                fixed (void* ib = imageBuffer)
```
dreddy-work, line 183: can this be moved to where it is used?

dreddy-work (approved):

dreddy-work:
LGTM.

JeremyKuhne (changes requested):
Think there was an existing bug here. :)

src/System.Windows.Forms/src/System/Windows/Forms/ControlPaint.cs
JeremyKuhne: This actually would not be better to move closer. Having it here would allow you to lose a level of nesting because you could use the same block as the other fixed statement. That said, there appears to be a bug in the existing code and we probably don't need the `imageBuffer` array at all.

JeremyKuhne:
ControlPaint tests are failing, I think this might be related to the issue I discussed.

src/System.Windows.Forms.Primitives/tests/UnitTests/System.Windows.Forms.Primitives.Tests.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
RussKie, line 1: This file should be reverted.

PR #7544, opened 2022-08-05
Convert CommDlgExtendedError to CsWin32

elachlan:
Converted CommDlgExtendedError to CsWin32

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7544)


src/System.Windows.Forms/src/System/Windows/Forms/Printing/PrintDialog.cs
```diff
@@ -398,7 +398,7 @@ private unsafe bool ShowPrintDialog(IntPtr hwndOwner)
 
                 if (PrintDlg(ref data).IsFalse())
                 {
-                    var result = CommDlgExtendedError();
+                    var result = PInvoke.CommDlgExtendedError();
```
lonitra, line 401: It looks like this line can actually be removed since we are not using this variable anywhere.

lonitra (approved):
Looks good!

src/System.Windows.Forms/src/System/Windows/Forms/Printing/PrintDialog.cs
```diff
@@ -398,7 +398,6 @@ private unsafe bool ShowPrintDialog(IntPtr hwndOwner)
 
                 if (PrintDlg(ref data).IsFalse())
                 {
-                    var result = CommDlgExtendedError();
```
Tanya-Solyanik, line 401: Would it make sense to keep this method call and print out the result with Debug.WriteLine

elachlan: If that is what the team wants. I will make it so.

RussKie: @Tanya-Solyanik any reason not to `Debug.Assert(result == 0, ....)`?
image

elachlan: @RussKie Happy to change it, probably makes more sense.

elachlan:
build failures are caused by `Using directive is unnecessary.` They are very necessary and "remove and sort usings" doesn't remove any.

RussKie:
The compiler can't lie :) VS is forgiving - i.e., it won't break the build, but if you build from cli the build will fail, if there are any issues.
```
D:\a\_work\1\s\src\System.Windows.Forms\src\System\Windows\Forms\SaveFileDialog.cs(7,1): error IDE0005: Using directive is unnecessary. [D:\a\_work\1\s\src\System.Windows.Forms\src\System.Windows.Forms.csproj]
```

src/System.Windows.Forms/src/System/Windows/Forms/SaveFileDialog.cs
```diff
@@ -4,6 +4,8 @@
 
 using System.ComponentModel;
 using System.Runtime.InteropServices;
+using Windows.Win32;
```
RussKie, line 7: This line the compiler is complaining about

RussKie: https://github.com/dotnet/winforms/blob/247d9cb8dd7aba194cd1a85eeb5e896a0df27ea1/src/System.Windows.Forms.Primitives/src/GlobalUsings.cs#L5

elachlan: Removing that line breaks the PInvoke call @ line 157.

RussKie: @lonitra @fernandanavamoya could you please help?

src/System.Windows.Forms/src/System/Windows/Forms/Printing/PrintDialog.cs
elachlan: As I read it, correct me if I am wrong (it happens a lot). The `if` block above checks the result for `False`. So the call to this doesn't make sense, because the return value from `CommDlgExtendedError` will return `0`. So really the debug.assert should appear outside the if block and check for `!=0`.

Have I misinterpreted the documentation?

RussKie: IIUIC, this function must only be called when the dialog returned FALSE. Anything other than 0 in this case indicate an error of some kind. If the dialog returns TRUE - the call to this function returns undefined.

src/System.Windows.Forms/src/System/Windows/Forms/SaveFileDialog.cs
lonitra: > Removing that line breaks the PInvoke call @ line 157.

Have you pulled recent changes from the feature/win32 branch? The global using was a recent addition so you may not have it locally, which I suspect is the cause of this break.

src/System.Windows.Forms/src/System/Windows/Forms/Printing/PrintDialog.cs
elachlan: Okay, your suggestion makes sense then. since the Assert call only outputs when the condition is `false`. I told you I am wrong a lot!

RussKie: This happens to the best of us :)

lonitra (approved):
Thanks for adding that debug line!

src/System.Windows.Forms/src/System/Windows/Forms/Printing/PrintDialog.cs
Tanya-Solyanik: >@Tanya-Solyanik any reason not to Debug.Assert(result == 0, ....)?

Because Debug.Assert(false,..) looks strange, and Debug.Fail is too much.

And also everything should be wrapped in #if DEBUG because there is no reason to retrieve the error code in retail build

src/System.Windows.Forms/src/System/Windows/Forms/Printing/PrintDialog.cs
```diff
@@ -397,7 +398,8 @@ private unsafe bool ShowPrintDialog(IntPtr hwndOwner)
 
                 if (!PrintDlg(ref data))
                 {
-                    var result = CommDlgExtendedError();
+                    var result = PInvoke.CommDlgExtendedError();
```
Tanya-Solyanik, line 401: Consider wrapping lines 401 and 402 in #if DEBUG - there is no reason to PInvoke in the retail build

Tanya-Solyanik (approved):
Looks good, thanks.

src/System.Windows.Forms/src/System/Windows/Forms/Printing/PrintDialog.cs
```diff
@@ -397,7 +398,10 @@ private unsafe bool ShowPrintDialog(IntPtr hwndOwner)
 
                 if (!PrintDlg(ref data))
                 {
-                    var result = CommDlgExtendedError();
+#if DEBUG
+                    var result = PInvoke.CommDlgExtendedError();
+                    Debug.Assert(result == 0, $"PrintDlg returned non zero error code: {result}");
```
Tanya-Solyanik, line 403: Please remove a `using` directive for System.Diagnostics and add namespace name to the `Debug.Assert` call to fix the build.

Tanya-Solyanik (approved):

src/System.Windows.Forms/src/System/Windows/Forms/Printing/PrintDialog.cs
RussKie: Hahah, good point! Should have picked in my reviews...

PR #7545, opened 2022-08-05
removing unused GDIPlus

elachlan:
removing unused GDIPlus

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7545)


dreddy-work:
@JeremyKuhne as FYI

JeremyKuhne (approved):
This is fine to remove. This was part of investigating ways to avoid allocations with System.Drawing. I have a pending feature to implement to expose the handles in more System.Drawing types that would be necessary to go further with this.

PR #7546, opened 2022-08-05
Introduce IHandle<T>

JeremyKuhne:
Add typed IHandle to keep handles typed. Add HandleRef<T> that derives from IHandle<T> to use where direct binding with IHandle<T> is not always necessary.

Port a number of User32 APIs to cswin32 to vet design.

First commit starts transitioning to global usings for interop to make porting easier.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7546)


src/System.Windows.Forms.Primitives/src/Interop/HandleRef.cs
```diff
@@ -0,0 +1,41 @@
...
+
+    [SetsRequiredMembers]
+    public HandleRef(object? wrapper, THandle handle)
+    {
+        Wrapper = wrapper;
+        Handle = handle;
+    }
+
+    [SetsRequiredMembers]
+    public HandleRef(IHandle<THandle> handle)
```
dreddy-work, line 24: seems handle can be made nullable here unless null check on line 27 is not necessary.

src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
```diff
@@ -666,10 +666,10 @@ internal void FocusActiveControlInternal()
             if (_activeControl is not null && _activeControl.Visible)
             {
                 // Avoid focus loops, especially with ComboBoxes.
-                IntPtr focusHandle = User32.GetFocus();
-                if (focusHandle == IntPtr.Zero || FromChildHandle(focusHandle) != _activeControl)
+                var focusHandle = PInvoke.GetFocus();
```
dreddy-work, line 669: Explicit type here would be consistent.

src/System.Windows.Forms/src/System/Windows/Forms/LinkLabel.cs
```diff
@@ -395,10 +395,9 @@ protected Cursor OverrideCursor
                     {
                         // We want to instantly change the cursor if the mouse is within our bounds.
                         // This includes the case where the mouse is over one of our children
-                        var r = new RECT();
                         User32.GetCursorPos(out Point p);
-                        User32.GetWindowRect(this, ref r);
-                        if ((r.left <= p.X && p.X < r.right && r.top <= p.Y && p.Y < r.bottom) || User32.GetCapture() == Handle)
+                        PInvoke.GetWindowRect(this, out var r);
```
dreddy-work, line 399: out RECT? i see multiple changes like these below but type is not obvious.

JeremyKuhne: It is obvious, this says `GetWindowRect`. It is particularly awkward to be explicit at the moment as there are two `RECT` types as we transition.

Incidentally, our rules are not a hard "is it obvious"? It is a consideration, but readability is also a consideration. The only hard rule is don't use it for intrinsic types (`int`, `bool`, etc.).

src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
JeremyKuhne: Missed this one once I added the global usings.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/Behavior/BehaviorService.cs
```diff
@@ -485,9 +485,9 @@ public void PushCaptureBehavior(Behavior behavior)
             if (uiService != null)
             {
                 IWin32Window hwnd = uiService.GetDialogOwnerWindow();
-                if (hwnd != null && hwnd.Handle != IntPtr.Zero && hwnd.Handle != User32.GetActiveWindow())
+                if (hwnd != null && hwnd.Handle != IntPtr.Zero && hwnd.Handle != PInvoke.GetActiveWindow())
```
lonitra, line 488:
```suggestion
                if (hwnd is not null && hwnd.Handle != 0 && hwnd.Handle != PInvoke.GetActiveWindow())
```

src/System.Windows.Forms.Primitives/src/NativeMethods.txt
```diff
@@ -16,30 +18,57 @@ CreatePen
...
+MOUSEHOOKSTRUCT
+PeekMessage
+PostMessage
+SetActiveWindow
+SetFocus
+SetParent
 SetSuspendState
+SetWindowLong
+SetWindowsHookEx
+MultiByteToWideChar
```
lonitra, line 73: Looks like this got out of alphabetical order

src/System.Windows.Forms.Primitives/src/Interop/UnownedHandle.cs
```diff
@@ -0,0 +1,13 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Diagnostics.CodeAnalysis;
+
+internal readonly struct UnownedHandle<THandle> : IHandle<THandle> where THandle : unmanaged
+{
+    public required THandle Handle { get; init; }
+
+    [SetsRequiredMembers]
+    public UnownedHandle(THandle handle) => Handle = handle;
```
lonitra, line 12: What is this being used for? I didn't see it used anywhere in the files (though I could've easily missed it)

src/System.Windows.Forms/src/GlobalUsings.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
```
RussKie, line 1: Please move these to props/csproj file. There's no need to create a new file for these.

JeremyKuhne: I want them in the source files so it is easy to find / manipulate. We shouldn't have to go to the project file to find a using.

src/System.Windows.Forms.Primitives/src/NativeMethods.txt
JeremyKuhne: Apparently I can't spell when I merge conflicts. :)

src/System.Windows.Forms.Primitives/src/Interop/UnownedHandle.cs
JeremyKuhne: Yeah, that was an experiment that I accidentally left in.

JeremyKuhne:
x64 debug failed once with https://github.com/dotnet/winforms/issues/7547

src/System.Windows.Forms.Primitives/src/Interop/IHandle.cs
```diff
@@ -6,19 +6,49 @@
...
+///   of the classes (such as System.Windows.Forms.Control) that
+///   meet this common pattern in interop and encourages correct alignment
+///   with the proper owner.
+///  </para>
+///  <para>
+///   Note that keeping objects alive is necessary ONLY when the object has
+///   a finalizer that will explicitly close the handle.
+///  </para>
 /// </remarks>
 internal interface IHandle
```
elachlan, line 25: Should this have an obsolete tag?

src/System.Windows.Forms.Primitives/src/Interop/HandleRef.cs
```diff
@@ -0,0 +1,41 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Diagnostics.CodeAnalysis;
+
+/// <summary>
+///  Adapter to use when owning classes cannot directly implement <see cref="IHandle{T}"/>.
+/// </summary>
+internal readonly struct HandleRef<THandle> : IHandle<THandle>, IEquatable<HandleRef<THandle>>
```
Tanya-Solyanik, line 10: We are naming files after the class names, so the right name here would be `HandleRef'1.cs` to indicate that this is a generic class that takes 1 parameter

src/System.Windows.Forms.Primitives/src/Interop/IHandle.cs
JeremyKuhne: If we were keeping it, sure. It will vanish shortly, so it isn't worth doing.

src/System.Windows.Forms.Primitives/src/Interop/HandleRef.cs
JeremyKuhne: I don't think we should be naming things after their generic template unless we absolutely _have_ to. It isn't normal to do so (see the entire .NET runtime repo). `IHandle` is going away shortly and even if it weren't it wouldn't break our reasons for having type to file- it will still be just as easy to find and won't make a giant source file.

lonitra (approved):

PR #7554, opened 2022-08-08
Make wrapper explicit on IHandle<T>

JeremyKuhne:
With an explicit `IHandle<T>.Wrapper` we can avoid boxing when `IHandle<T>` is implmented on a struct and passing `IHandle<T>` to `GC.KeepAlive()`.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7554)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/Interop/IHandle.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/IHandle.cs
@@ -51,4 +51,16 @@ internal interface IHandle
 internal interface IHandle<THandle> where THandle : unmanaged
 {
     THandle Handle { get; }
+
+    /// <summary>
+    ///  Owner of the <see cref="Handle"/> that might close it when finalized. Default is the
+    ///  <see cref="IHandle{THandle}"/> implementer.
+    /// </summary>
+    /// <remarks>
+    ///  <para>
+    ///   This allows decoupling the owner from the <typeparamref name="THandle"/> provider and avoids boxing when
+    ///   <see cref="IHandle{THandle}"/> is on a struct. See <see cref="HandleRef{THandle}"/> for a concrete usage.
+    ///  </para>
+    /// </remarks>
+    object? Wrapper => this;
 }
--- a/src/System.Windows.Forms/src/GlobalUsings.cs
+++ b/src/System.Windows.Forms/src/GlobalUsings.cs
@@ -5,6 +5,7 @@
 global using Windows.Win32;
 global using Windows.Win32.UI.WindowsAndMessaging;
 global using Foundation = Windows.Win32.Foundation;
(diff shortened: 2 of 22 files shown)


dreddy-work (approved):
LGTM.

lonitra (approved):

src/System.Windows.Forms.Primitives/src/Interop/IHandle.cs
```diff
@@ -51,4 +51,16 @@ internal interface IHandle
...
+    ///  Owner of the <see cref="Handle"/> that might close it when finalized. Default is the
+    ///  <see cref="IHandle{THandle}"/> implementer.
+    /// </summary>
+    /// <remarks>
+    ///  <para>
+    ///   This allows decoupling the owner from the <typeparamref name="THandle"/> provider and avoids boxing when
+    ///   <see cref="IHandle{THandle}"/> is on a struct. See <see cref="HandleRef{THandle}"/> for a concrete usage.
+    ///  </para>
+    /// </remarks>
+    object? Wrapper => this;
```
Tanya-Solyanik, line 65: The summary describes this property as `owner`, would `Owner` be a better name than `Wrapper`?

fernandanavamoya (approved):

src/System.Windows.Forms/src/GlobalUsings.cs
```diff
@@ -5,6 +5,7 @@
 global using Windows.Win32;
```
RussKie, line 5: Having GlobalUsing files makes me somewhat uncomfortable. It feels like we're going against the grain of the "implicit usings" feature design - i.e., no need for additional files, all global usings are specified via *.props/*.csproj files. Here's the [doc](https://docs.microsoft.com/dotnet/core/project-sdk/msbuild-props#using).

JeremyKuhne: It is an option, not a mandate to put them in project files. The docs for the actual [language feature](https://docs.microsoft.com/dotnet/csharp/language-reference/keywords/using-directive#global-modifier) say "You can also globally include a namespace by adding a `<Using>` item to your project file", not "you should". The only guidance is that you typically would put them in a single file in a project (for obvious reasons).

Having them in the project file makes discoverability difficult and has the additional detriment of not seeing that we don't need a particular using.

src/System.Windows.Forms.Primitives/src/Interop/IHandle.cs
JeremyKuhne: `Wrapper` is the term `HandleRef` uses, was keeping it consistent.

PR #7561, opened 2022-08-09
Remove conflicting interop structs

JeremyKuhne:
Removed our `RECT`, `HDC`, etc structs in favor of cswin32. This necessitated changing all of the existing P/Invokes back to `[DllImport]` as the `[LibaryImport]` doesn't work with the source generated types.

- Remove all `using foo =` workarounds.
- Push forward to cswin32 daily build to address bool cast issue
- New cswin32 uses System.Drawing types where plausible (`POINT` -> `Point`, as we currently do).

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7561)


src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Message.cs
```diff
@@ -139,5 +144,13 @@ public bool Equals(Message other)
         public override int GetHashCode() => HashCode.Combine(HWnd, Msg);
 
         public override string ToString() => MessageDecoder.ToString(this);
+
+        internal MSG ToMSG() => new()
```
lonitra, line 148: Is this being used? I see we've added an implicit operator from Message -> MSG in MSG.cs

fernandanavamoya (approved):
Reviewed 267 files. Looks good to me.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Message.cs
JeremyKuhne: Yeah, I'll kill it in the next change.

RussKie:
> This necessitated changing all of the existing P/Invokes back to `[DllImport]` as the `[LibaryImport]` doesn't work with the source generated types.

That's a bugger. Should we raise this to the interop team?

PR #7563, opened 2022-08-10
Convert Shell32 structs to CsWin32

fernandanavamoya:
Converted these strucs:
- AUTOCOMPLETEOPTIONS
- DROPDESCRIPTION
- DROPIMAGETYPE

Deleted these files since they have zero references:
- Interop.CDCS.cs

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7563)


src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DragDropHelperTests.cs
```diff
@@ -3,6 +3,7 @@
 using System.Numerics;
 using System.Runtime.InteropServices.ComTypes;
 using Xunit;
+using Windows.Win32.UI.Shell;
```
lonitra, line 6: We might want to consider putting this line into the GlobalUsings.cs file to avoid having to add it to every file we change for Shell32 conversion.

lonitra (approved):
LGTM! :) Just the comment about adding the line to GlobalUsings.cs, but that can happen in a later PR

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DragDropHelperTests.cs
fernandanavamoya: Completely agree

PR #7564, opened 2022-08-10
Convert DeleteX to Cswin32

lonitra:
- Convert `DeleteDC`,`DeleteEnhMetaFile`,`DeleteObject` to Cswin32
- Move `CreateBitmapScope`,`CreateBrushScope`,`CreatePenScope`,`ObjectScope` to PInvoke folder
- General style clean up

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7564)


dreddy-work (approved):

src/System.Windows.Forms.Primitives/tests/UnitTests/Windows/Win32/RegionTests.cs
```diff
@@ -57,12 +57,12 @@ public void GetClipRgn_NoRegion()
         public void RegionScope_NullWithNoClippingRegion()
         {
             // Create a bitmap using the screen's stats
-            HDC hdc = PInvoke.CreateCompatibleDC((HDC)default);
+            HDC hdc = CreateCompatibleDC((HDC)default);
```
KlausLoeffelmann, line 60: So, one question that I have:
What's are the rules with `static using PInvoke` at some places, and not at others?

KlausLoeffelmann (approved):
Just had a question of when to use `static using PInvoke`, and when not.

src/System.Windows.Forms.Primitives/tests/UnitTests/Windows/Win32/RegionTests.cs
lonitra: Originally, we had been using `using static Interop` in a lot of our files up to this point. Since we will eventually be converting all of our interop to cswin32, my thoughts were that we'd eventually replace those old interop usings with `using static PInvoke` for files mostly converted, but now that we have the global usings adding the `static using PInvoke` everywhere is probably not necessary. I will correct this. :)

PR #7576, opened 2022-08-11
Adding contrast borders for a DataGridViewCheckBoxCell in Win11.

### LINKED ISSUE #6657: [Accessibility][Win11] The color contrast of the dataGridViewCheckBox cell is less than 3:1
.NET Core Version: 
* .NET SDK 7.0.100-preview.2.22108.9

- Yes

**More Info:**
1.	This issue can’t reproduce on Windows 10 OS.
2.	This issue can reproduce on Windows 11 OS both for Winforms NetFX and Core projects(can reproduce with 3.1, 5.0, 6.0, 7.0).

**Problem description:**
The color contrast less than 3:1 between background color of checkbox and background color of whole cell.
For Win11:
image

**Expected behavior:**
The color contrast should be greater than 3:1 between background color of checkbox and background color of whole cell.
For Win10:
image

**Minimal repro:**
1.	Create a .NET Core Winforms application with DataGridView control added.
2.	Add a DataGridViewTextBoxColumn to it.
3.	Build and run the project.
4.	Check the checkbox, then launch Accessibility Insights for windows tool and observe the color contrast.
Or download the attached project, then execute above step 4.
[CheckBoxCellColorContrast.zip](https://github.com/dotnet/winforms/files/8031403/CheckBoxCellColorContrast.zip)

NikitaSemenovAkvelon:
Fixes #6657 

## Proposed changes

- Added `RoundedRectangle` class which encapsulates the painting logic for a rounded rectangle.
- Used `RoundedRectangle` inside `DataGridViewCheckBoxCellRenderer` for Win11.

## Customer Impact

- A user can see CheckBox inside DataGridView more clearly.

## Regression? 

- No

## Risk

- Minimal

### Before

image

### After

image

## Test methodology 

- Manually
- AI

## Accessibility testing  

- AI

 

## Test environment(s) 

- 7.0.100-preview.5.22307.18

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7576)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCheckBoxCell.DataGridViewCheckBoxCellRenderer.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCheckBoxCell.DataGridViewCheckBoxCellRenderer.cs
@@ -21,9 +21,9 @@ namespace System.Windows.Forms
                 CheckBoxRenderer.DrawBackground(g, bounds, Rectangle.Truncate(g.ClipBounds));
 
                 // It uses because the checked style of the checkbox in a selected cell has a low contrast ratio with the background. Only for Windows 11.
-                // State 5 is selected, unfocused.
-                // State 6 is selected, focused.
-                if (OsVersion.IsWindows11_OrGreater && (state == 5 || state == 6))
+                const int StateSelectedUnfocused = 5;
+                const int StateSelectedFocused = 6;
+                if (OsVersion.IsWindows11_OrGreater && (state == StateSelectedUnfocused  || state ==StateSelectedFocused))
                 {
                     DrawContrastBorder(g, bounds);
                 }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCheckBoxCell.DataGridViewCheckBoxCellRenderer.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCheckBoxCell.DataGridViewCheckBoxCellRenderer.cs
@@ -13,23 +13,35 @@ namespace System.Windows.Forms
         {
             private static VisualStyleRenderer? s_visualStyleRenderer;
 
-            public static VisualStyleRenderer CheckBoxRenderer
-            {
-                get
-                {
-                    if (s_visualStyleRenderer is null)
-                    {
-                        s_visualStyleRenderer = new VisualStyleRenderer(CheckBoxElement);
-                    }
-
-                    return s_visualStyleRenderer;
-                }
-            }
+            public static VisualStyleRenderer CheckBoxRenderer => s_visualStyleRenderer ??= new VisualStyleRenderer(CheckBoxElement);
 
             public static void DrawCheckBox(Graphics g, Rectangle bounds, int state)
             {
-                CheckBoxRenderer.SetParameters(CheckBoxElement.ClassName, CheckBoxElement.Part, (int)state);
+                CheckBoxRenderer.SetParameters(CheckBoxElement.ClassName, CheckBoxElement.Part, state);
                 CheckBoxRenderer.DrawBackground(g, bounds, Rectangle.Truncate(g.ClipBounds));
+
+                // It uses because the checked style of the checkbox in a selected cell has a low contrast ratio with the background. Only for Windows 11.
+                // State 5 is selected, unfocused.
+                // State 6 is selected, focused.
+                if (OsVersion.IsWindows11_OrGreater && (state == 5 || state == 6))
+                {
+                    DrawContrastBorder(g, bounds);
+                }
+            }
+
+            private static void DrawContrastBorder(Graphics g, Rectangle bounds)
+            {
+                const int penWidth = 1;
+                const int cornerRadius = 3;
+
+                var focusPen = new Pen(SystemColors.ControlText, penWidth);
+                var rectangleBounds = bounds with
+                {
+                    Width = bounds.Width - penWidth,
+                    Height = bounds.Height - penWidth
+                };
+
+                RoundedRectangle.Paint(g, focusPen, rectangleBounds, cornerRadius);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/RoundedRectangle.cs
```diff
@@ -0,0 +1,58 @@
...
+            if (pen == null)
+                throw new ArgumentNullException(nameof(pen));
+
+            using GraphicsPath path = CreatePath(bounds, cornerRadius);
+            graphics.DrawPath(pen, path);
+        }
+
+        private static GraphicsPath CreatePath(Rectangle bounds, int cornerRadius)
+        {
+            int diameter = cornerRadius * 2;
```
Tanya-Solyanik, line 28: Unfortunately, we don't own this radius. Windows will change it in win12 and we will have to add a new case for a new version. I would prefer to not fix this bug at all. We had won't fixed the same bugs related to buttons which are much more common controls, I don't see reason to fix a less common case. @merriemcgaw ? The way to go is to request windows to publish theming APIs.

Tanya-Solyanik (commented):
I would not want to special-case rendering for each OS. Besides, we don't know that this is the right radius.

src/System.Windows.Forms/src/System/Windows/Forms/RoundedRectangle.cs
merriemcgaw: I can go with that.

Tanya-Solyanik: Can we do it when we do other theming changes? This seems to be a relatively minor case given that we have designer work in front of us.

merriemcgaw: I couldn't agree more. We should turn this back into a draft PR and then revisit once we've got the designer where we want it.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCheckBoxCell.DataGridViewCheckBoxCellRenderer.cs
```diff
@@ -13,23 +13,35 @@ private static class DataGridViewCheckBoxCellRenderer
...
                 }
             }
 
-            public static void DrawCheckBox(Graphics g, Rectangle bounds, int state)
+            private static void DrawContrastBorder(Graphics g, Rectangle bounds)
             {
-                CheckBoxRenderer.SetParameters(CheckBoxElement.ClassName, CheckBoxElement.Part, (int)state);
-                CheckBoxRenderer.DrawBackground(g, bounds, Rectangle.Truncate(g.ClipBounds));
+                const int penWidth = 1;
+                const int cornerRadius = 3;
```
Tanya-Solyanik, line 35: We don't know what the algorithm to generate this value is, at this point this is only a guess, which can be broken on the next version of Windows. In order to draw the border, we need ComCtl/UxTheme tesm to publish the API that we can use.

Tanya-Solyanik (commented):
We should delay this change until Windows can provide APIs for us.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCheckBoxCell.DataGridViewCheckBoxCellRenderer.cs
```diff
@@ -13,23 +13,35 @@ private static class DataGridViewCheckBoxCellRenderer
...
-                        s_visualStyleRenderer = new VisualStyleRenderer(CheckBoxElement);
-                    }
+                CheckBoxRenderer.SetParameters(CheckBoxElement.ClassName, CheckBoxElement.Part, state);
+                CheckBoxRenderer.DrawBackground(g, bounds, Rectangle.Truncate(g.ClipBounds));
 
-                    return s_visualStyleRenderer;
+                // It uses because the checked style of the checkbox in a selected cell has a low contrast ratio with the background. Only for Windows 11.
+                // State 5 is selected, unfocused.
+                // State 6 is selected, focused.
+                if (OsVersion.IsWindows11_OrGreater && (state == 5 || state == 6))
```
RussKie, line 26: What do you think about this?
```suggestion
                // It uses because the checked style of the checkbox in a selected cell has a low contrast ratio with the background. Only for Windows 11.
                if (OsVersion.IsWindows11_OrGreater && (state == /* selected, unfocused */5 || state == /* selected, focused */6))
```
...or even better:
```suggestion
                // It uses because the checked style of the checkbox in a selected cell has a low contrast ratio with the background. Only for Windows 11.
                const int StateSelectedUnfocused = 5;
                const int StateSelectedFocused = 6;
                if (OsVersion.IsWindows11_OrGreater && (state == StateSelectedUnfocused  || state ==StateSelectedFocused))
```

src/System.Windows.Forms/src/System/Windows/Forms/RoundedRectangle.cs
```diff
@@ -0,0 +1,58 @@
...
+    ///  This class fully encapsulates the painting logic for a rounded rectangle. (Used by DataGridViewCheckBoxCell)
+    /// </summary>
+    internal static class RoundedRectangle
+    {
+        public static void Paint(Graphics graphics, Pen pen, Rectangle bounds, int cornerRadius)
+        {
+            if (graphics == null)
+                throw new ArgumentNullException(nameof(graphics));
+            if (pen == null)
+                throw new ArgumentNullException(nameof(pen));
```
RussKie, line 20: Throwing exceptions like this hinders possible jit optimisations. We prefer to use helpers instead. 
In this case we can use [`ArgumentNullException.ThrowIfNull`](https://docs.microsoft.com/dotnet/api/system.argumentnullexception.throwifnull). For more exotic cases we have our own helpers in https://github.com/dotnet/winforms/blob/main/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/ArgumentValidation.cs.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCheckBoxCell.DataGridViewCheckBoxCellRenderer.cs
RussKie: Looks like I didn't get the spacing right, as I've written it in the browser.

PR #7580, opened 2022-08-12
Converted FDAP/GETPROPERTYSTOREFLAGS to CsWin32

fernandanavamoya:
Converted these strucst:
- `FDAP`
- `GETPROPERTYSTOREFLAGS`

Added `global using Windows.Win32.UI.Shell` to GlobalUsing.cs of System.Windows.Forms and System.Windows.Forms.Primitives


src/System.Windows.Forms.Primitives/src/Interop/Shell32/Interop.IFileOpenDialog.cs
```diff
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System.Runtime.InteropServices;
+using Shell = Windows.Win32.UI.Shell;
```
JeremyKuhne, line 6: This shouldn't be necessary with the global using, can you double check with the other `using =` in this project that you've added?

JeremyKuhne (changes requested):
Just the comment on possibly unnecessary usings, other than that looks good.

PR #7581, opened 2022-08-12
Convert Gdi32 GetX to Cswin32

lonitra:
- Convert `GetBitmapBits`,`GetBkColor`,`GetBkMode`,`GetBrushOrgEx`,`GetClipBox`,`GetClipRgn`,`GetCurrentObject`,`GetDeviceCaps`,
`GetDIBits`,`GetMapMode`,`GetNearestColor`,`GetObjectType`,`GetRegionData`,`GetRgnBox`,`GetROP2`,`GetStockObject`,
`GetSystemPaletteEntries`,`GetTextAlign`,`GetTextColor`,`GetTextExtentPoint32W`,`GetTextMetrics`,`GetViewportExtEx`,
`GetViewportOrgEx`,`GetWorldTransform`,`StretchDIBits`,`GetObject`,`GetPaletteEntries`,`GetTextExtendPoint32W`
- Move `SelectPaletteScope` to PInvoke folder
- Add `IHandle<HDC>` to `HDC`
- `Control.MetafileDCWrapper.cs`: Change initialization of `BITMAPINFO` used in calls `GetDIBits` and `StretchDIBits`. Cswin32 definition of `BITMAPINFO` contains trailing array thus we now need to manually allocate enough space for the bitmap colors. Also done some re-arrangement of code to try to reduce nesting.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7581)


src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetPaletteEntries.cs
```diff
@@ -0,0 +1,18 @@
...
+
+using System.Runtime.InteropServices;
+
+namespace Windows.Win32;
+
+internal static partial class PInvoke
+{
+    public unsafe static uint GetPaletteEntries(HPALETTE hpal, Span<PALETTEENTRY> entries)
+    {
+        fixed (PALETTEENTRY* entry = &MemoryMarshal.GetReference(entries))
```
JeremyKuhne, line 13: nit: the `&MemoryMarshal.GetReference` is an artifact of early Span work. C# added the ability to provide a method called `GetPinnableReference()` that allows spans (or anything else that implements it) to just get pinned with `&`. So this could be `&entries`, but that can be done in a follow-up.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.SelectPaletteScope.cs
```diff
@@ -2,11 +2,11 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System.Drawing;
```
JeremyKuhne, line 5: Surprised that `System.Drawing` is causing conflicts.

src/System.Windows.Forms.Primitives/tests/TestUtilities/DeviceContextState.cs
```diff
@@ -27,31 +27,35 @@ internal unsafe class DeviceContextState
...
+            BackColor = PInvoke.GetBkColor(hdc);
+            TextColor = PInvoke.GetTextColor(hdc);
+            Rop2Mode = (Gdi32.R2)PInvoke.GetROP2(hdc);
+            TextAlign = (Gdi32.TA)PInvoke.GetTextAlign(hdc);
+            BackgroundMode = (Gdi32.BKMODE)PInvoke.GetBkMode(hdc);
 
-            Matrix3x2 transform = default;
-            Gdi32.GetWorldTransform(hdc, ref transform);
-            Transform = transform;
+            XFORM transform = default;
```
JeremyKuhne, line 37: It would be nice to have `XFORM` mapped to `Matrix3x2`, or at least have the option to. It is supported on .NET Standard, although it does require a package ref.  https://apisof.net/catalog/bb678d8b-2368-b6c1-8f06-e436bb2fa874 Would be worth opening an issue on cswin32.

For now it might be better to just keep it as `Matrix3x2` and pass it as `(XFORM*)(void*)&transform`.

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxUiaTextProvider.cs
```diff
@@ -195,7 +195,7 @@ public override Point GetPositionFromCharForUpperRightCorner(int startCharIndex,
                 pt = GetPositionFromCharIndex(startCharIndex);
 
                 // add the width of the character at that position.
-                if (GetTextExtentPoint32(ch, out Size size))
+                if (GetTextExtentPoint32(ch, out SIZE size))
```
JeremyKuhne, line 198: `SIZE` to `Size` is missing from cswin32 and we should make sure we have an issue on it. For now I'd keep this as `Size` and cast to `SIZE*` after pinning it.

src/System.Windows.Forms/src/System/Windows/Forms/TextExtensions.cs
```diff
@@ -282,17 +282,20 @@ private static (User32.DT Flags, TextPaddingOptions Padding) SplitTextFormatFlag
...
+            SIZE size = default;
             using var selectFont = new Gdi32.SelectObjectScope(hdc, hfont);
 
-            Gdi32.GetTextExtentPoint32W(hdc, text, text.Length, ref size);
+            fixed (char* pText = text)
+            {
+                PInvoke.GetTextExtentPoint32W(hdc, pText, text.Length, &size);
+            }
 
             return new Size(size.Width, size.Height);
```
JeremyKuhne, line 300: Wow, this line is bizarre. Surprised I missed it when I was working on this. When `size` was `Size` it should have just returned `size`.

JeremyKuhne (changes requested):
Awesome work!

Try constraining the `SIZE`/`SIZEF` types in the API surface area. Consider extending `SIZE` and `SIZEF` structs to provide conversion operators to `Size` and `SizeF`. Also it is valid to just force cast when calling with them by reference (`(SIZE*)(void*)&size`). Whichever combination of things you choose I'm fine with.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.SelectPaletteScope.cs
lonitra: On line 54 we had a call `Graphics.GetHalftonePalette` which comes from `System.Drawing`. Since this file has been moved to namespace `Windows.Win32`, the compiler kept trying to find a `GetHalftonePalette` method from `Windows.Win32.Graphics`. Figured since there is only one instance of this call, removing the using and just putting `global::System.Drawing.Graphics.GetHalftonePalette()` would suffice. Thinking again, I could've also just had a static using for `System.Drawing.Graphics`

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxUiaTextProvider.cs
lonitra: I looked again more carefully and cswin32 has implicit conversion for `SIZE` <-> `Size`, just need to cast the pointers. I will fix this up.

JeremyKuhne (approved):
:shipit: 🚀

PR #7584, opened 2022-08-12
Removed obsolete  `DomainUpDownAccessibleObject` usage

### LINKED ISSUE #7587: Complete obsoletion process for DomainUpDownAccessibleObject class.
We'll be conservative in NET7 and will return instance of an obsoleted DomainUpDownAccessibleObject type from DomainUpDown.CreateAccessibilityInstance method. This will prevent failures of the runtime casts in snippets like this one:
```
internal class MySpinner : DomainUpDown
{
<snip>
DomainUpDownAccessibleObject ao = CreateAccessibleObject() as DomainUpDownAccessibleObject;
<snip>
}
```

In .NET 8 we will remove protected override methos `DomainUpDown.CreateAccessibilityInstance`. This kind of change is allowed by the [breaking changes](https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/breaking-change-rules.md#members) rules:
```
Introducing or removing an override

Make note, that introducing an override might cause previous consumers to skip over the override when calling base.
```

Additionally, we should consider promoting build time warning to an error because this class had always been documented as internal use only. Lets discuss this part of the change within the team.

_Originally posted by @Tanya-Solyanik in https://github.com/dotnet/winforms/pull/7330#discussion_r944772133_

vladimir-krestov:
Fixes #7587 

## Proposed changes

- Remove `DomainUpDownAccessibleObject` usage as an accessible object for `DomainUpDown` because this class is obsolete now
- Promote obsoletion from warning to error

A screenshot of Visual Studio editor window demonstrating that using DomainUpDownAccessibleObject is generating a compilation error

## Customer Impact

- If customer used `DomainUpDownAccessibleObject` in their code, they would have to change it to `ControlAccessibleObject` or `AccessibleObject`. Otherwise, no impact.

## Regression? 

- No

## Risk

- Minimal

## Test environment(s) 

- .NET 7.0 RC1
- Windows 11

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7584)

### DIFF
--- a/src/Common/src/Obsoletions.cs
+++ b/src/Common/src/Obsoletions.cs
@@ -14,13 +14,11 @@ internal static class Obsoletions
     // and ensure the "aka.ms/dotnet-warnings/{0}" URL points to documentation for the obsoletion
     // The diagnostic ids reserved for obsoletions are WFDEV### (WFDEV001 - WFDEV999).
 
-#pragma warning disable WFDEV002 // Type or member is obsolete
-    internal const string DomainUpDownAccessibleObjectMessage = $"{nameof(DomainUpDown.DomainUpDownAccessibleObject)} is no longer used to provide accessible support for {nameof(DomainUpDown)} controls.";
-#pragma warning restore WFDEV002 // Type or member is obsolete
+    internal const string DomainUpDownAccessibleObjectMessage = $"DomainUpDownAccessibleObject is no longer used to provide accessible support for {nameof(DomainUpDown)} controls. Use {nameof(Control.ControlAccessibleObject)} instead.";
     internal const string DomainUpDownAccessibleObjectDiagnosticId = "WFDEV002";
 
 #pragma warning disable WFDEV003 // Type or member is obsolete
     internal const string DomainItemAccessibleObjectMessage = $"{nameof(DomainUpDown.DomainItemAccessibleObject)} is no longer used to provide accessible support for {nameof(DomainUpDown)} items.";
-#pragma warning disable WFDEV003 // Type or member is obsolete
+#pragma warning restore WFDEV003 // Type or member is obsolete
     internal const string DomainItemAccessibleObjectDiagnosticId = "WFDEV003";
 }
--- a/docs/list-of-diagnostics.md
+++ b/docs/list-of-diagnostics.md
@@ -36,7 +36,7 @@ The acceptance criteria for adding an obsoletion includes:
 |  __`WFDEV001`__ | Casting to/from IntPtr is unsafe, use `WParamInternal`. |
 |  __`WFDEV001`__ | Casting to/from IntPtr is unsafe, use `LParamInternal`. |
 |  __`WFDEV001`__ | Casting to/from IntPtr is unsafe, use `ResultInternal`. |
-|  __`WFDEV002`__ | `DomainUpDown.DomainUpDownAccessibleObject` is no longer used to provide accessible support for `DomainUpDown` controls. |
+|  __`WFDEV002`__ | `DomainUpDown.DomainUpDownAccessibleObject` is no longer used to provide accessible support for `DomainUpDown` controls. Use `ControlAccessibleObject` instead. |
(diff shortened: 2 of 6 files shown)


dreddy-work:
@vladimir-krestov , may be i missed context. Which class is obsolete now? and how obsolete class makes removal of accessibility object? people may still use obsolete class .

dreddy-work:
> @vladimir-krestov , may be i missed context. Which class is obsolete now? and how obsolete class makes removal of accessibility object? people may still use obsolete class .

nevermind, i am following discussion on original thread.

vladimir-krestov:
Closed this according to https://github.com/dotnet/winforms/pull/7330#discussion_r944772133

Tanya-Solyanik:
Re-opening for .NET 8. I also suggest changing warnings to errors

RussKie:
I think that we either need to take this in .NET 7 or undo this obsoletion
https://github.com/dotnet/winforms/blob/01abac5b35616f182267222da1976d429e1eab19/src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainUpDownAccessibleObject.cs#L11-L17

Tanya-Solyanik:
We obsoleted that type as a compile time warning, applications still can use it. This gives developers time to re-write applications if needed.

merriemcgaw:
Removing this would be binary breaking which we don't want to do. We should mark it as Obsolete as Error.

Tanya-Solyanik:
We still can remove the CreateAccessibleObject override. Error message should specify how to fix the error, i.e. "use the base class instead" I think the public base class is ControlAccessibleObject and AccessibleObject for items, but please verify that.

dmitrii-drobotov:
Once the PR is merged, I will create a request issue in https://github.com/dotnet/docs to add a breaking change and update existing obsoletion WFDEV002 page.

src/Common/src/Obsoletions.cs
```diff
@@ -14,13 +14,11 @@ internal static class Obsoletions
...
-#pragma warning disable WFDEV002 // Type or member is obsolete
-    internal const string DomainUpDownAccessibleObjectMessage = $"{nameof(DomainUpDown.DomainUpDownAccessibleObject)} is no longer used to provide accessible support for {nameof(DomainUpDown)} controls.";
-#pragma warning restore WFDEV002 // Type or member is obsolete
+    internal const string DomainUpDownAccessibleObjectMessage = $"DomainUpDownAccessibleObject is no longer used to provide accessible support for {nameof(DomainUpDown)} controls. Use {nameof(Control.ControlAccessibleObject)} instead.";
     internal const string DomainUpDownAccessibleObjectDiagnosticId = "WFDEV002";
 
 #pragma warning disable WFDEV003 // Type or member is obsolete
     internal const string DomainItemAccessibleObjectMessage = $"{nameof(DomainUpDown.DomainItemAccessibleObject)} is no longer used to provide accessible support for {nameof(DomainUpDown)} items.";
-#pragma warning disable WFDEV003 // Type or member is obsolete
+#pragma warning restore WFDEV003 // Type or member is obsolete
```
Tanya-Solyanik, line 22: Thanks!

Tanya-Solyanik (approved):
Looks good!

MelonWang1:
Verified on latest .NET 8.0.100-alpha.1.22567.28 build + private dll, this issue still reproduces. Error message does not contain this line "Use ControlAccessibleObject instead". Screenshot as below:
Screenshot 2022-11-18 003300

MelonWang1:
@dmitrii-drobotov Thank you, verified on latest .NET 8.0.100-alpha.1.22567.28 build + private dll + replace System.Windows.Forms.dll in C:\Program Files\dotnet\packs\Microsoft.WindowsDesktop.App.Ref\[version]\ref\net8.0 folder, this issue was fixed. Error message contain this line "Use ControlAccessibleObject instead". Screenshot as below:
Screenshot 2022-11-20 174219

PR #7588, opened 2022-08-13
Convert FormatMessageW to CsWin32

fernandanavamoya:
- Converted `FormatMessageW` to CsWin32
- Deleted Kernel32 folder since it has no more files inisde it

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7588)


src/System.Windows.Forms/src/System/Windows/Forms/ComponentModel/COM2Interop/COM2PropertyDescriptor.cs
```diff
@@ -1326,40 +1327,47 @@ public unsafe override void SetValue(object component, object value)
...
+                                255,
+                                null);
+
+                            if (result == 0)
+                            {
+                                errorInfo = string.Format(CultureInfo.CurrentCulture, string.Format(SR.DispInvokeFailed, "SetValue", hr));
+                            }
+                            else
+                            {
+                                errorInfo = TrimNewline(IpBuffer);
```
JeremyKuhne, line 1349: The result is the number of characters in the buffer.  You can create a `ReadOnlySpan<char>` around the buffer with that length, then `.TrimEnd()` the `\n` and `\r` characters off of it. Then simply call `.ToString()` on the span to get your string.

src/System.Windows.Forms/src/System/Windows/Forms/ComponentModel/COM2Interop/COM2PropertyDescriptor.cs
```diff
@@ -1326,40 +1327,47 @@ public unsafe override void SetValue(object component, object value)
...
-                            strMessage,
-                            255,
-                            IntPtr.Zero);
-
-                        if (result == 0)
-                        {
-                            errorInfo = string.Format(CultureInfo.CurrentCulture, string.Format(SR.DispInvokeFailed, "SetValue", hr));
-                        }
-                        else
+                        char[] IpBuffer = ArrayPool<char>.Shared.Rent(256 + 1);
```
JeremyKuhne, line 1330: Local variables should be camel cased. `buffer` would be a good name here. When you are done with the array you need to return it to the `ArrayPool`.

JeremyKuhne (changes requested):

src/System.Windows.Forms/src/System/Windows/Forms/ComponentModel/COM2Interop/COM2PropertyDescriptor.cs
```diff
@@ -1326,40 +1327,47 @@ public unsafe override void SetValue(object component, object value)
...
+                                null,
+                                (uint)hr,
+                                PInvoke.GetThreadLocale(),
+                                b,
+                                255,
+                                null);
+
+                            if (result == 0)
+                            {
+                                errorInfo = string.Format(CultureInfo.CurrentCulture, string.Format(SR.DispInvokeFailed, "SetValue", hr));
```
Tanya-Solyanik, line 1345: You don't need to nest string.Format, I think this override will work - https://docs.microsoft.com/en-us/dotnet/api/system.string.format?view=net-6.0#system-string-format(system-iformatprovider-system-string-system-object-system-object). Though this is not part of your change, so take it or leave it. This would be a candidate for a second commit within the PR.

src/System.Windows.Forms/src/System/Windows/Forms/ComponentModel/COM2Interop/COM2PropertyDescriptor.cs
```diff
@@ -1326,40 +1327,53 @@ public unsafe override void SetValue(object component, object value)
...
+                            {
+                                errorInfo = string.Format(CultureInfo.CurrentCulture, SR.DispInvokeFailed, "SetValue", hr);
+                            }
+                            else
+                            {
+                                ReadOnlySpan<char> buff = new(buffer);
+                                buff.TrimEnd('\n');
+                                buff.TrimEnd('\r');
+                                errorInfo = buff.ToString();
+                                //errorInfo = TrimNewline(buffer);
```
lonitra, line 1353: Let's go ahead and remove this line :)

src/System.Windows.Forms/src/System/Windows/Forms/ComponentModel/COM2Interop/COM2PropertyDescriptor.cs
```diff
@@ -1326,40 +1327,53 @@ public unsafe override void SetValue(object component, object value)
...
+
+                        ArrayPool<char>.Shared.Return(buffer);
                     }
 
                     throw new ExternalException(errorInfo, (int)hr);
             }
         }
 
-        private static string TrimNewline(StringBuilder errorInfo)
+        private static string TrimNewline(char[] errorInfo)
```
lonitra, line 1364: Is this method still in use? If not, we can remove this as well.

src/System.Windows.Forms/src/System/Windows/Forms/ComponentModel/COM2Interop/COM2PropertyDescriptor.cs
```diff
@@ -1346,9 +1346,15 @@ public unsafe override void SetValue(object component, object value)
                             }
                             else
                             {
-                                errorInfo = TrimNewline(IpBuffer);
+                                ReadOnlySpan<char> buff = new(buffer);
```
Tanya-Solyanik, line 1349: Please use full words for variable names

PR #7590, opened 2022-08-14
Temporarily fix VB App Framework Logging.

KlausLoeffelmann:
The runtime PR https://github.com/dotnet/runtime/pull/73087 broke the Visual Basic Application Framework Logging.

https://source.dot.net/#Microsoft.VisualBasic.Forms/Microsoft/VisualBasic/Logging/Log.vb,170

image

The original assumption that this would work was, that when `GetSupportedAttributes` gets called in a class derived from `TraceSource` that then would tell that the trace source was configured from a config file. This is not only what `DefaultTraceSource` does, it seems to be the only reason for it existence. Now, as far as I understand it, after the change, `GetSupportAttributes` gets called unconditionally, so `HasBeenConfigured` in the VB AppFramework's `DefaultTraceSource` returns always `true`, and so a FileLogTraceListener gets never added, since the Log class assumes, the trace source was configured by a config file. And that's why there will never be a value other than `nothing` for the `DefaultFileLogWriter` property.

After discussing this with the WinForms Team and the VB PM, we're fixing this temporary in the runtime, by returning `false` unconditionally for `Logging.Log.HasBeenConfigured`. This assumes, that the trace source has NOT been configured by a config file. Configuring by a config file remains a broken scenario for VB My.Logging, until we will be getting a fix from the Runtime.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7590)


src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -11,45 +11,53 @@ namespace System.Windows.Forms.Primitives
...
     {
         private const string ScaleTopLevelFormMinMaxSizeForDpiSwitchName = "System.Windows.Forms.ScaleTopLevelFormMinMaxSizeForDpi";
-
-        private static int s_scaleTopLevelFormMinMaxSizeForDpi;
-        public static bool ScaleTopLevelFormMinMaxSizeForDpi
-        {
-            [MethodImpl(MethodImplOptions.AggressiveInlining)]
-            get => GetCachedSwitchValue(ScaleTopLevelFormMinMaxSizeForDpiSwitchName, ref s_scaleTopLevelFormMinMaxSizeForDpi);
-        }
+        private const string AssumeVbLogClassWasConfiguredByConfigFileName = "System.Windows.Forms.AssumeVbLogClassWasConfiguredByConfigFile";
```
Tanya-Solyanik, line 14: CS applications don't need this switch, why not put it into the VB assembly?

src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/Logging/Log.vb
```diff
@@ -163,20 +164,10 @@ Namespace Microsoft.VisualBasic.Logging
                     If _listenerAttributes Is Nothing Then
                         _listenerAttributes = Attributes
                     End If
-                    Return _hasBeenInitializedFromConfigFile
```
Tanya-Solyanik, line 166: I would had returned `false` here for the RC and would have worked on the proper fix, i.e. how to get if this was initialized from the config file.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -11,45 +11,53 @@ namespace System.Windows.Forms.Primitives
...
 
+        public static bool AssumeVbLogClassWasConfiguredByConfigFile
+            => GetSwitchValue(AssumeVbLogClassWasConfiguredByConfigFileName, ref s_assumeVbLogClassWasConfiguredByConfigFile);
+
+        private static bool GetCachedSwitchValue(string switchName, ref int cachedSwitchValue)
+            => cachedSwitchValue switch
+            {
+                // The cached switch value has 3 states: 0 - unknown, 1 - true, -1 - false
+
+                < 0 => false,
```
Tanya-Solyanik, line 44: Please compare IL generated for this code and for the original one, The original pattern was chosen because it was generating the tightest IL.

KlausLoeffelmann: I strongly suggest such thing should be documented then.
I'll revert this then, because this is not the right time for investigations for this.

src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/Logging/Log.vb
KlausLoeffelmann: I thought about this first. But then I thought, there is a good chance it will stay like this, and then we could have it right away.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
KlausLoeffelmann: I wouldn't say this is language related? 
It's not that the Application Framework can only be used by VB. So, for example, take a look at this:

https://stackoverflow.com/questions/59268557/where-is-the-winforms-application-framework-in-c-sharp-that-exists-in-vb-net

Also, I often saw even C# WPF apps using a few features from the VB App Framework to ease their life. It's a valid approach!

Tanya-Solyanik: Application that benefits from this quirk, references AppFramework, right? Why doesn't this quirk belong to Applicationframework then?

src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/Logging/Log.vb
Tanya-Solyanik: >there is a good chance it will stay like this,

Maybe runtime can add a protected property for us to look up?

src/System.Windows.Forms.Primitives/src/System.Windows.Forms.Primitives.csproj
```diff
@@ -35,6 +35,10 @@
     <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="4.5.1" />
   </ItemGroup>
 
+  <ItemGroup>
+    <InternalsVisibleTo Include="Microsoft.VisualBasic.Forms"/>
+  </ItemGroup>
```
RussKie, line 40: We expose the internals here:
https://github.com/dotnet/winforms/blob/01abac5b35616f182267222da1976d429e1eab19/src/System.Windows.Forms.Primitives/src/Properties/AssemblyInfo.cs#L9-L11

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -11,45 +11,53 @@ namespace System.Windows.Forms.Primitives
...
-                return false;
-            if (cachedSwitchValue > 0)
-                return true;
-
-            return GetSwitchValue(switchName, ref cachedSwitchValue);
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            get => GetCachedSwitchValue(ScaleTopLevelFormMinMaxSizeForDpiSwitchName, ref s_scaleTopLevelFormMinMaxSizeForDpi);
         }
 
+        public static bool AssumeVbLogClassWasConfiguredByConfigFile
```
RussKie, line 36: Is this an opt-in or opt-out switch?

RussKie:
I don't quite understand the use cases, developer and user experiences. It would be great if you could provide more details explaining those. E.g., this is how a .NET Framework app worked, now ported to .NET 7 it'd work like this. Would a developer or use need to opt-in or opt-out? 
Thank you

src/System.Windows.Forms.Primitives/src/System.Windows.Forms.Primitives.csproj
RussKie: I ran a little experiment, and ultimately both ways achieve the same result:
image
So for consistency, everything should be in the AssemblyInfo.


That said, it looks as though there's a bug in VB compiler as it doesn't appear to honour the attribute...
```
D:\Development\dotnet-winforms\src\Microsoft.VisualBasic.Forms\src\Microsoft\VisualBasic\Logging\Log.vb(167,28): error BC30389: 'System.Windows.Forms.Primitives.LocalAppContextSwitches' is not accessible in this context because it is 'Friend'. [D:\Development\dotnet-winforms\src\Microsoft.VisualBasic.Forms\src\Microsoft.VisualBasic.Forms.vbproj]
    0 Warning(s)
    1 Error(s)
```

KlausLoeffelmann: Well, Roslyn does make it work, so, it must be possible.
Will investigate, but for now, we have decided to skip the whole quirking-effort for RC1 and will try to concentrate on fixing the issue properly after it. @merriemcgaw, @KathleenDollard  for FYI.

For now, we will return `false` for `Log.DefaultTraceSource.HasBeenConfigured` unconditionally, and then we'll fix this for RC2.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
KlausLoeffelmann: Depends on the point of view. We are opting out of the assumption that the trace source was configured by code and opting into the assumption that the trace source was configured by a config file. But let's shelve this for the time being, until we know what we are doing with this, anyway.

KlausLoeffelmann:
I pretty much explained the breaking change in the runtime which led to this.
I added a few lines of explanation in the introduction, what we did.
This needs to go in quick, to unblock us.

Tanya-Solyanik:
I would prefer returning a false unconditionally because:
1 quirks are public APIs and it will be difficult to remove them when we have the fix
2. if an application enables logging in the config file, it has also to define a quirk in json file, this is inconvenient
3. we had not exhausted possibilities of a correct fix,

src/Microsoft.VisualBasic.Forms/src/Microsoft.VisualBasic.Forms.vbproj
```diff
@@ -1,4 +1,4 @@
-<Project>
```
dreddy-work, line 1: Undo this file?

dreddy-work (approved):

dreddy-work:
Squash merge please.

src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/Logging/Log.vb
```diff
@@ -163,20 +163,12 @@ Namespace Microsoft.VisualBasic.Logging
...
+
+                    ' TODO: This is a tempory fix, which will break configuring logging via file for the time being. See: https://github.com/dotnet/winforms/pull/7590
```
Tanya-Solyanik, line 167: Did you open a follow up bug to get this fixed?

RussKie:
@KlausLoeffelmann is this a breaking change that we need to document?

merriemcgaw:
@RussKie yeah, probably. It would be a minute number of customers impacted but it's good to document it of course.

Tanya-Solyanik:
app config was not enabled in the previous versions, so this is a regression from .NET framework but not from the previous release of Core

PR #7594, opened 2022-08-15
Update loc

dreddy-work:
Disable loc

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7594)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -57,7 +57,7 @@ variables:
   - name: PostBuildSign
     value: true
   - name: EnableLoc
-    value: true
+    value: false
   - name: isPublic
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -57,7 +57,7 @@ variables:
   - name: PostBuildSign
     value: true
   - name: EnableLoc
-    value: true
+    value: false
```
RussKie, line 60: We can probably automate this with something like (pseudocode) 
```
    value: ${{ contains(variables['target branch'], 'release/') }}
```

PR #7596, opened 2022-08-15
Convert Last of Gdi32 CreateX to Cswin32

lonitra:
- Convert `CreateDC`,`CreateEnhMetaFile`,`CreateICW`,`CreatePatternBrush`,`CreateFontIndirect` to Cswin32
- Add implicit conversion `HdcMetdataEnhFileHandle` > `HDC`
- Temporarily added `using LOGFONTW = Windows.Win32.Graphics.Gdi.LOGFONTW` for files with ambiguous `LOGFONTW` reference. Will replace all `LOGFONTW` references to use Cswin32 `LOGFONTW` in next PR

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7596)


RussKie:
Looks like this change has broken several tests.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextRange.cs
```diff
@@ -8,6 +8,7 @@
 using static Interop;
 using static Interop.UiaCore;
 using static Interop.User32;
+using LOGFONTW = Windows.Win32.Graphics.Gdi.LOGFONTW;
```
dreddy-work, line 11: Q: why do you have to define it like this? Is there any ambiguous use?

lonitra: Yes, there is an ambiguous use right now. We have our own definition of `LOGFONTW` and Cswin32 does as well. They are currently clashing. My plan is to temporarily resolve the ambiguous use this way for this PR and then get rid of our `LOGFONTW` definition and use Cswin32's `LOGFONTW` definition throughout our codebase in a separate PR. We have quite a few `LOGFONTW` usages so I thought it would be better suited for a separate PR.

dreddy-work: make sense. thank you.

dreddy-work (approved):
LGTM.

PR #7600, opened 2022-08-16
Turn Off Safe Handle Friendly Overloads

lonitra:
- Updated Cswin32 version
- Removed temporary usage of `GetModuleHandleEx` and replace with `GetModuleHandle`. We used `GetModuleHandleEx` because of the difficulties with the safe handle overload generated by `GetModuleHandle`, but now we have ability to turn off safe handles.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7600)


lonitra:
cc: @AArnott

fernandanavamoya (approved):

PR #7674, opened 2022-08-26
Convert threads to CsWin32

fernandanavamoya:
- Converted `COMDLG_FILTERSPEC`, `DragAcceptFiles`, `SHCreateItemFromParsingName`, `SIATTRIBFLAGS` and `SIGDN` into CsWin32.
- Added `DragAcceptFiles` and `SHCreateItemFromParsingName` to our PInvoke library.
- Added `Windows.Win32.UI.Shell.Common` to Glabal Usings

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7674)


src/System.Windows.Forms.Primitives/src/GlobalUsings.cs
```diff
@@ -9,3 +9,6 @@
 global using Windows.Win32.UI.Shell;
 global using Windows.Win32.UI.Shell.PropertiesSystem;
 global using Windows.Win32.UI.WindowsAndMessaging;
+global using Windows.Win32.UI.Shell.Common;
+global using Shell = Windows.Win32.UI.Shell;
+global using Common = Windows.Win32.UI.Shell.Common;
```
Tanya-Solyanik, line 14: This name might collide with other `Common` namespaces in the future. How about `ShellCommon`?
Also why did you need this alias given line 12 uses this namespace?

src/System.Windows.Forms/src/GlobalUsings.cs
```diff
@@ -9,3 +9,5 @@
 global using Windows.Win32.UI.Controls;
 global using Windows.Win32.UI.Shell;
 global using Windows.Win32.UI.WindowsAndMessaging;
+global using SIGDN = Windows.Win32.UI.Shell.SIGDN;
```
Tanya-Solyanik, line 12: Why do you need these aliases, if `Windows.Win32.UI.Shell` namespace is already referenced?

src/System.Windows.Forms/src/System/Windows/Forms/FolderBrowserDialog.cs
```diff
@@ -393,7 +393,7 @@ private void GetResult(IFileDialog dialog)
             dialog.GetResult(out IShellItem? item);
             if (item is not null)
             {
-                HRESULT hr = item.GetDisplayName(SIGDN.FILESYSPATH, out _selectedPath!);
+                HRESULT hr = item.GetDisplayName(SIGDN.SIGDN_FILESYSPATH, out _selectedPath!);
```
Tanya-Solyanik, line 396: `SIGDN.SIGDN_FILESYSPATH` - in this case `SIGDN.` namespace name does not add context, and thus makes code less readable.

src/System.Windows.Forms.Primitives/src/GlobalUsings.cs
fernandanavamoya: This was my mistake. While building it there were some ambiguous using between CsWin32's and Interop's `COMLG_FILTERSPEC` struct even with the global using in line 12 because I hadn't deleted `Interop.COMLG_FILTERSPEC`. However, now that I have, I can delete line 13 and 14 and delete all `Shell.struct` and `Common.struct`.

src/System.Windows.Forms/src/GlobalUsings.cs
fernandanavamoya: Same case here as the one above.

src/System.Windows.Forms/src/System/Windows/Forms/RichTextBox.cs
```diff
@@ -3237,7 +3237,7 @@ private void UpdateOleCallback()
                     }
                 }
 
-                Shell32.DragAcceptFiles(this, false);
+                PInvoke.DragAcceptFiles(this, false);
```
lonitra, line 3240: nit: add param names when passing true/false as param values.

lonitra (approved):
LGTM

src/System.Windows.Forms.Primitives/src/GlobalUsings.cs
Tanya-Solyanik: Perfect!

Tanya-Solyanik (approved):
Looks good, but please see Loni's comment.

PR #7679, opened 2022-08-27
CsWin32 Remove Interop.WindowClasses

elachlan:
Removed Interop.WindowClasses and used PInvoke constants

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7679)


elachlan:
As a part of this change, I ran a sort lines on the `NativeMethods.txt`. That is why there are more changes in that file. Hope this is okay.

elachlan:
@RussKie the test failures are strange here.

dreddy-work (approved):

dreddy-work:
> @RussKie the test failures are strange here.

Was failing on arm64 arch only. We do some intermittent failures on ARM. i reset it for now. lets see if it fails consistently.

PR #7685, opened 2022-08-29
Update TargetingRefPack to MicrosoftNETCoreAppRuntimewinx64PackageVer…

dkurepa:
Prepare branch for release.
This change was tested on https://dev.azure.com/dnceng/internal/_git/dotnet-winforms?version=GB7.0.0-test, in the .NET 7 stable test build

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -65,14 +65,13 @@
   <ItemGroup>
     <!-- Set TargetingPackVersion -->
     <FrameworkReference Update="Microsoft.NETCore.App"
-      Condition=" '$(VSRedistCommonNetCoreSharedFrameworkx6480PackageVersion)'!='' And $(TargetFramework.StartsWith('net7.')) ">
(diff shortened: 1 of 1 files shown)


dreddy-work (commented):
@RussKie , this you just modified to get the flow unblocking where build was failing on VS redist package versions.

@dkurepa , can you share context on this change? what is making you to do this?

RussKie:
A bit more background info on this change would be great.

But conceptually I don't see a problem - there's [Microsoft.NETCore.App.Runtime.win-x64 package](https://dev.azure.com/dnceng/public/_artifacts/feed/dotnet8/NuGet/Microsoft.NETCore.App.Runtime.win-x64/overview/) targeting 8.0.0-alpha.1 on the dotnet8 feed.

RussKie (approved):

dkurepa:
@dreddy-work I'm working on a test stable build for .NET 7, this was a fix that was needed, so I'm just making the same changes in release/7.0 and main. I'm not a 100% sure if it's needed in main tho

RussKie:
@dkurepa is this something we want to have merged?

PR #7698, opened 2022-08-31
Update build images (release/7.0)

RussKie:
Per the offline email directions

Tested successfully:
* https://dev.azure.com/dnceng/public/_build/results?buildId=1975549&view=results
* https://dev.azure.com/dnceng/internal/_build/results?buildId=1975548&view=results

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7698)

PR #7709, opened 2022-09-01
Update public pool names

alexperovich:
This change is required for builds to continue working in the new org, dev.azure.com/dnceng-public.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7709)

### DIFF
--- a/azure-pipelines-richnav.yml
+++ b/azure-pipelines-richnav.yml
@@ -16,7 +16,7 @@ stages:
         enableRichCodeNavigation: true
         richCodeNavigationEnvironment: 'production'
         pool:
-          name: NetCore1ESPool-Svc-Public
+          name: NetCore-Svc-Public
           demands: ImageOverride -equals Build.Windows.Amd64.VS2022.Pre.Open
 
         steps:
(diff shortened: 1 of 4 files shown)


azure-pipelines-richnav.yml
```diff
@@ -16,7 +16,7 @@ stages:
         enableRichCodeNavigation: true
         richCodeNavigationEnvironment: 'production'
         pool:
-          name: NetCore1ESPool-Svc-Public
+          name: NetCore-Svc-Public
           demands: ImageOverride -equals Build.Windows.Amd64.VS2022.Pre.Open
```
RussKie, line 20: If we're servicing rc1 branch, then we should use this opportunity to update the image names as well - similar to https://github.com/dotnet/winforms/pull/7698
```suggestion
          demands: ImageOverride -equals windows.vs2022preview.amd64.open
```

RussKie (changes requested):

RussKie:
@mmitche @MattGal are we taking changes into rc1 branch?

RussKie:
Ah, that makes sense. Thank you for clarifying it.

PR #7741, opened 2022-09-06
Implementing UIA provider for TextBox and MaskedTextBox controls

### LINKED ISSUE #3421: Epic - [Accessibility] Implement UIA provider for all controls

NikitaSemenovAkvelon:
Fixes #3421 

## Proposed changes

- Enabled UIA provider for `TextBox` with nested classes and `MaskedTextBox`.
- Refactored `TextBox's` child classes to reduce extra code.
- Implemented `IValueProvider` for `TextBoxBaseAccessibleObject`.
- Raised `AutomationFocusChangedEvent` in `OnGotFocus` methods.
- Added few unit tests.

## Customer Impact

- Improving development experience for `TextBox` with nested classes and `MaskedTextBox` control accessibility.

## Regression? 

- No

## Risk

- Minimal

### Before

Screenshot 2022-09-06 090441
_TextBox_
Screenshot 2022-09-06 090506
_MaskedTextBox_

### After

Screenshot 2022-09-06 085747
Screenshot 2022-09-06 090029
_TextBox_
Screenshot 2022-09-06 085948
Screenshot 2022-09-06 090049
_MaskedTextBox_

## Test methodology 

- Manually
- Accessibility tools
- Unit tests
- CTI

## Accessibility testing  

- Inspect
- AI
- Narrator

 

## Test environment(s) 

- 8.0.100-alpha.1.22423.9

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7741)

### DIFF
--- a/src/System.Windows.Forms/src/PublicAPI.Unshipped.txt
+++ b/src/System.Windows.Forms/src/PublicAPI.Unshipped.txt
@@ -1,4 +1,7 @@
 System.Windows.Forms.PrintPreviewControl.TabStop.get -> bool
 System.Windows.Forms.PrintPreviewControl.TabStop.set -> void
 System.Windows.Forms.PrintPreviewControl.TabStopChanged -> System.EventHandler?
-*REMOVED* override System.Windows.Forms.ToolStripButton.CanSelect.get -> bool
+*REMOVED*override System.Windows.Forms.ToolStripButton.CanSelect.get -> bool
+override System.Windows.Forms.MaskedTextBox.CreateAccessibilityInstance() -> System.Windows.Forms.AccessibleObject!
+override System.Windows.Forms.MaskedTextBox.OnGotFocus(System.EventArgs! e) -> void
+override System.Windows.Forms.MaskedTextBox.OnMouseDown(System.Windows.Forms.MouseEventArgs! e) -> void
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewTextBoxEditingControl.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewTextBoxEditingControl.cs
@@ -24,9 +24,7 @@ namespace System.Windows.Forms
         }
 
         protected override AccessibleObject CreateAccessibilityInstance()
-        {
-            return new DataGridViewTextBoxEditingControlAccessibleObject(this);
-        }
+            => new DataGridViewTextBoxEditingControlAccessibleObject(this);
 
         public virtual DataGridView? EditingControlDataGridView
         {
@@ -235,15 +233,6 @@ namespace System.Windows.Forms
         {
             _valueChanged = true;
             _dataGridView?.NotifyCurrentCellDirty(true);
-        }
-
-        protected override void OnGotFocus(EventArgs e)
-        {
-            base.OnGotFocus(e);
-            if (IsAccessibilityObjectCreated)
-            {
-                AccessibilityObject.RaiseAutomationEvent(UiaCore.UIA.AutomationFocusChangedEventId);
-            }
         }
 
         protected override void OnMouseWheel(MouseEventArgs e)
--- a/src/System.Windows.Forms/src/System/Windows/Forms/MaskedTextBox.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/MaskedTextBox.cs
@@ -27,7 +27,7 @@ namespace System.Windows.Forms
     [DefaultProperty(nameof(Mask))]
     [Designer("System.Windows.Forms.Design.MaskedTextBoxDesigner, " + AssemblyRef.SystemDesign)]
     [SRDescription(nameof(SR.DescriptionMaskedTextBox))]
-    public class MaskedTextBox : TextBoxBase
+    public partial class MaskedTextBox : TextBoxBase
     {
         // Consider: The MaskedTextBox control, when initialized with a non-null/empty mask, processes all
         // WM_CHAR messages and always sets the text using the SetWindowText Windows function in the furthest base
@@ -1136,6 +1136,8 @@ namespace System.Windows.Forms
             }
         }
 
+        internal override bool SupportsUiaProviders => true;
+        
         /// <summary>
         ///  The Text setter validates the input char by char, raising the MaskInputRejected event for invalid chars.
         ///  The Text getter returns the formatted text according to the IncludeLiterals and IncludePrompt properties.
@@ -1481,6 +1483,8 @@ namespace System.Windows.Forms
         {
         }
 
+        protected override AccessibleObject CreateAccessibilityInstance() => new MaskedTextBoxAccessibleObject(this);
+
         /// <summary>
         ///  Creates a handle for this control. This method is called by the framework, this should
         ///  not be called directly. Inheriting classes should always call <c>base.CreateHandle</c> when overriding this method.
@@ -1804,6 +1808,16 @@ namespace System.Windows.Forms
             }
         }
 
+        protected override void OnGotFocus(EventArgs e)
+        {
+            base.OnGotFocus(e);
+
+            if (IsAccessibilityObjectCreated)
+            {
+                AccessibilityObject.SetFocus();
+            }
+        }
+
         /// <summary>
         ///  Overridden to update the newly created handle with the settings of the PasswordChar properties
         ///  if no mask has been set.
@@ -2030,6 +2044,11 @@ namespace System.Windows.Forms
             {
                 _flagState[IME_ENDING_COMPOSITION] = false;
             }
+
+            if (IsHandleCreated && IsAccessibilityObjectCreated && ContainsNavigationKeyCode(e.KeyCode))
+            {
+                AccessibilityObject?.RaiseAutomationEvent(UiaCore.UIA.Text_TextSelectionChangedEventId);
+            }
         }
 
         /// <summary>
@@ -2060,6 +2079,20 @@ namespace System.Windows.Forms
             if (Events[EVENT_MASKINPUTREJECTED] is MaskInputRejectedEventHandler eh)
             {
                 eh(this, e);
+            }
+        }
+
+        protected override void OnMouseDown(MouseEventArgs e)
+        {
+            base.OnMouseDown(e);
+
+            if (IsHandleCreated && IsAccessibilityObjectCreated)
+            {
+                // As there is no corresponding windows notification
+                // about text selection changed for TextBox assuming
+                // that any mouse down on textbox leads to change of
+                // the caret position and thereby change the selection.
+                AccessibilityObject.RaiseAutomationEvent(UiaCore.UIA.Text_TextSelectionChangedEventId);
             }
         }
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyGridView.GridViewTextBox.GridViewTextBoxAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyGridView.GridViewTextBox.GridViewTextBoxAccessibleObject.cs
@@ -75,12 +75,6 @@ namespace System.Windows.Forms.PropertyGridInternal
                     _ => base.GetPropertyValue(propertyID)
                 };
 
-                internal override bool IsPatternSupported(UiaCore.UIA patternId) => patternId switch
-                {
-                    UiaCore.UIA.ValuePatternId => true,
-                    _ => base.IsPatternSupported(patternId)
-                };
-
                 internal override UiaCore.IRawElementProviderSimple? HostRawElementProvider
                 {
                     get
@@ -102,18 +96,9 @@ namespace System.Windows.Forms.PropertyGridInternal
                         {
                             return name;
                         }
-                        else
-                        {
-                            GridEntry selectedGridEntry = _owningPropertyGridView.SelectedGridEntry;
-                            if (selectedGridEntry is not null)
-                            {
-                                return selectedGridEntry.AccessibilityObject.Name;
-                            }
-                        }
 
-                        return base.Name;
+                        return _owningPropertyGridView.SelectedGridEntry?.AccessibilityObject.Name ?? base.Name;
                     }
-                    set => base.Name = value;
                 }
 
                 internal override int[] RuntimeId
@@ -127,12 +112,6 @@ namespace System.Windows.Forms.PropertyGridInternal
                 internal override bool IsReadOnly
                     => _owningPropertyGridView.SelectedGridEntry is not PropertyDescriptorGridEntry propertyDescriptorGridEntry
                         || propertyDescriptorGridEntry.IsPropertyReadOnly;
-
-                internal override void SetFocus()
-                {
-                    RaiseAutomationEvent(UiaCore.UIA.AutomationFocusChangedEventId);
-                    base.SetFocus();
-                }
             }
         }
     }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TextBox.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TextBox.cs
@@ -397,6 +397,8 @@ namespace System.Windows.Forms
             }
         }
 
+        internal override bool SupportsUiaProviders => true;
+
         internal override Size GetPreferredSizeCore(Size proposedConstraints)
         {
             Size scrollBarPadding = Size.Empty;
@@ -590,6 +592,11 @@ namespace System.Windows.Forms
                     SelectAll();
                 }
             }
+
+            if (IsAccessibilityObjectCreated)
+            {
+                AccessibilityObject.SetFocus();
+            }
         }
 
         /// <summary>
@@ -651,24 +658,6 @@ namespace System.Windows.Forms
             }
         }
 
-        private static bool ContainsNavigationKeyCode(Keys keyCode)
-        {
-            switch (keyCode)
-            {
-                case Keys.Up:
-                case Keys.Down:
-                case Keys.PageUp:
-                case Keys.PageDown:
-                case Keys.Home:
-                case Keys.End:
-                case Keys.Left:
-                case Keys.Right:
-                    return true;
-                default:
-                    return false;
-            }
-        }
-
         protected override void OnMouseDown(MouseEventArgs e)
         {
             base.OnMouseDown(e);
@@ -679,7 +668,7 @@ namespace System.Windows.Forms
                 // about text selection changed for TextBox assuming
                 // that any mouse down on textbox leads to change of
                 // the caret position and thereby change the selection.
-                AccessibilityObject?.RaiseAutomationEvent(UiaCore.UIA.Text_TextSelectionChangedEventId);
+                AccessibilityObject.RaiseAutomationEvent(UiaCore.UIA.Text_TextSelectionChangedEventId);
             }
         }
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.TextBoxBaseAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.TextBoxBaseAccessibleObject.cs
@@ -58,11 +58,43 @@ namespace System.Windows.Forms
                 {
                     UIA.TextPatternId => true,
                     UIA.TextPattern2Id => true,
+                    UIA.ValuePatternId => true,
                     _ => base.IsPatternSupported(patternId)
                 };
 
-            internal override bool IsReadOnly
-                => Owner is TextBoxBase textBoxBase && textBoxBase.ReadOnly;
+            internal override bool IsReadOnly => _owningTextBoxBase.ReadOnly;
+
+            public override string? Name
+            {
+                get
+                {
+                    var name = base.Name;
+                    return name != null || !_owningTextBoxBase.PasswordProtect ? name : string.Empty;
+                }
+                set => base.Name = value;
+            }
+
+            public override string? Value => !_owningTextBoxBase.PasswordProtect ? ValueInternal : SR.AccessDenied;
+
+            protected virtual string ValueInternal => Owner.Text;
+
+            internal override void SetFocus()
+            {
+                if (!Owner.IsHandleCreated)
+                {
+                    return;
+                }
+
+                base.SetFocus();
+
+                RaiseAutomationEvent(UIA.AutomationFocusChangedEventId);
+            }
+
+            internal override void SetValue(string? newValue)
+            {
+                Owner.Text = newValue;
+                base.SetValue(newValue);
+            }
 
             internal override ITextRangeProvider? DocumentRangeInternal
                 => _textProvider?.DocumentRange;
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.cs
@@ -1341,6 +1341,24 @@ namespace System.Windows.Forms
             }
         }
 
+        protected bool ContainsNavigationKeyCode(Keys keyCode)
+        {
+            switch (keyCode)
+            {
+                case Keys.Up:
+                case Keys.Down:
+                case Keys.PageUp:
+                case Keys.PageDown:
+                case Keys.Home:
+                case Keys.End:
+                case Keys.Left:
+                case Keys.Right:
+                    return true;
+                default:
+                    return false;
+            }
+        }
+
         /// <summary>
         ///  Copies the current selection in the text box to the Clipboard.
         /// </summary>
@@ -2204,23 +2222,13 @@ namespace System.Windows.Forms
 
                     break;
                 case WM.DESTROY:
-                    base.WndProc(ref m);
-
-                    if (TryGetAccessibilityObject(out AccessibleObject? @object) && !RecreatingHandle)
+                    if (TryGetAccessibilityObject(out AccessibleObject? @object) && @object is TextBoxBaseAccessibleObject accessibleObject &&
+                        !RecreatingHandle)
                     {
-                        // The SupportsUiaProviders check prevents double disconnection after Control.ReleaseUiaProvider.
-                        // ReleaseUiaProvider works only for controls that support UIA, so this check allows to avoid
-                        // double disconnection when TextBoxBase will start UIA supporting.
-                        if (OsVersion.IsWindows8OrGreater && !SupportsUiaProviders)
-                        {
-                            UiaCore.UiaDisconnectProvider(@object);
-                        }
-
-                        if (@object is TextBoxBaseAccessibleObject accessibleObject)
-                        {
-                            accessibleObject.ClearObjects();
-                        }
+                        accessibleObject.ClearObjects();
                     }
+
+                    base.WndProc(ref m);
 
                     break;
                 default:
--- a/src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.UpDownEdit.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.UpDownEdit.cs
@@ -15,10 +15,9 @@ namespace System.Windows.Forms
             private readonly UpDownBase _parent;
             private bool _doubleClickFired;
 
-            internal UpDownEdit(UpDownBase parent) : base()
+            internal UpDownEdit(UpDownBase parent)
             {
                 SetStyle(ControlStyles.FixedHeight | ControlStyles.FixedWidth, true);
-                SetStyle(ControlStyles.Selectable, false);
 
                 _parent = parent;
             }
@@ -51,6 +50,15 @@ namespace System.Windows.Forms
                 }
 
                 _parent.OnMouseDown(_parent.TranslateMouseEvent(this, e));
+
+                if (IsHandleCreated && IsAccessibilityObjectCreated)
+                {
+                    // As there is no corresponding windows notification
+                    // about text selection changed for TextBox assuming
+                    // that any mouse down on textbox leads to change of
+                    // the caret position and thereby change the selection.
+                    AccessibilityObject.RaiseAutomationEvent(UiaCore.UIA.Text_TextSelectionChangedEventId);
+                }
             }
 
             protected override void OnMouseUp(MouseEventArgs e)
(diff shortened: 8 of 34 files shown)


dkazennov (dismissed):
As I understand in it's principle it is similar to other PRs to add UIA provider to controls (like #7314 I did).
Looks good to me.

src/System.Windows.Forms/src/System/Windows/Forms/MaskedTextBox.cs
```diff
@@ -1802,6 +1804,16 @@ protected unsafe override void OnBackColorChanged(EventArgs e)
             }
         }
 
+        protected override void OnGotFocus(EventArgs e)
+        {
+            base.OnGotFocus(e);
+
+            if (IsAccessibilityObjectCreated)
+            {
+                AccessibilityObject.SetFocus();
```
vladimir-krestov, line 1813: Why do we need it?

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyGridView.GridViewTextBox.GridViewTextBoxAccessibleObject.cs
```diff
@@ -102,18 +96,15 @@ public override AccessibleStates State
...
+
+                        GridEntry selectedGridEntry = _owningPropertyGridView.SelectedGridEntry;
+                        if (selectedGridEntry?.AccessibilityObject.Name is not null)
                         {
-                            GridEntry selectedGridEntry = _owningPropertyGridView.SelectedGridEntry;
-                            if (selectedGridEntry is not null)
-                            {
-                                return selectedGridEntry.AccessibilityObject.Name;
-                            }
+                            return selectedGridEntry.AccessibilityObject.Name;
```
vladimir-krestov, line 103: It duplicates. Try this:
```cs
return _owningPropertyGridView.SelectedGridEntry?.AccessibilityObject.Name ?? base.Name;
```

src/System.Windows.Forms/src/System/Windows/Forms/TextBox.cs
```diff
@@ -590,6 +592,11 @@ protected override void OnGotFocus(EventArgs e)
                     SelectAll();
                 }
             }
+
+            if (IsAccessibilityObjectCreated)
+            {
+                AccessibilityObject.SetFocus();
```
vladimir-krestov, line 598: What is the reasons for this?

src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.TextBoxBaseAccessibleObject.cs
```diff
@@ -50,11 +50,28 @@ internal override bool IsPatternSupported(UIA patternId)
                 {
                     UIA.TextPatternId => true,
                     UIA.TextPattern2Id => true,
+                    UIA.ValuePatternId => true,
```
vladimir-krestov, line 53: 1) Did it work before? For .NET Framework for example? Or in Explorer?
2) Please check related methods of this pattern, they should be implemented as well. (like `Expand` method of expand-collapse pattern)

src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.TextBoxBaseAccessibleObject.cs
```diff
@@ -50,11 +50,28 @@ internal override bool IsPatternSupported(UIA patternId)
...
                     _ => base.IsPatternSupported(patternId)
                 };
 
             internal override bool IsReadOnly
-                => Owner is TextBoxBase textBoxBase && textBoxBase.ReadOnly;
+                => Owner is TextBoxBase { ReadOnly: true };
```
vladimir-krestov, line 58: Great! 👍

src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.cs
```diff
@@ -2206,20 +2206,9 @@ protected override void WndProc(ref Message m)
...
-                        // double disconnection when TextBoxBase will start UIA supporting.
-                        if (OsVersion.IsWindows8OrGreater && !SupportsUiaProviders)
-                        {
-                            UiaCore.UiaDisconnectProvider(@object);
-                        }
-
-                        if (@object is TextBoxBaseAccessibleObject accessibleObject)
-                        {
-                            accessibleObject.ClearObjects();
-                        }
```
vladimir-krestov, line 2222: It was necessary for RichTextBox AO, find the commit in history of this change to read the reason.
As I remember, you should keep it

src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.UpDownEdit.UpDownEditAccessibleObject.cs
```diff
@@ -17,16 +17,15 @@ public UpDownEditAccessibleObject(UpDownEdit owner, UpDownBase parent) : base(ow
...
                 {
                     get => _parent.AccessibilityObject.Name
                         ?? _parent switch
                         {
                             NumericUpDown _ => SR.EditDefaultAccessibleName,
                             DomainUpDown _ => SR.EditDefaultAccessibleName,
-                            _ => null
+                            _ => string.Empty
                         };
-                    set => _parent.AccessibilityObject.Name = value;
```
vladimir-krestov, line 29: I think, you should return it, because we don't set it via `base.Name`, we use the another logic

src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.UpDownEdit.cs
```diff
@@ -15,10 +15,9 @@ internal partial class UpDownEdit : TextBox
             private readonly UpDownBase _parent;
             private bool _doubleClickFired;
 
-            internal UpDownEdit(UpDownBase parent) : base()
+            internal UpDownEdit(UpDownBase parent)
             {
                 SetStyle(ControlStyles.FixedHeight | ControlStyles.FixedWidth, true);
-                SetStyle(ControlStyles.Selectable, false);
```
vladimir-krestov, line 21: Please describe in details here and in the PR description why should we remove this.
It's necessary now and in the future to understand reasons of this change and make it clear if we will catch some possible future regressions

src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.UpDownEdit.cs
```diff
@@ -107,6 +106,11 @@ protected override void OnGotFocus(EventArgs e)
             {
                 _parent.SetActiveControl(this);
                 _parent.InvokeGotFocus(_parent, e);
+
+                if (IsAccessibilityObjectCreated)
+                {
+                    AccessibilityObject.SetFocus();
```
vladimir-krestov, line 112: Doesn't the base implementation work?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TextBoxBaseAccessibleObjectTests.cs
```diff
@@ -63,13 +63,13 @@ public void TextBoxBaseAccessibleObject_IsReadOnly_ReturnsCorrectValue(bool read
         }
 
         [WinFormsFact]
-        public void TextBoxBaseAccessibleObject_Value_ReturnsEmpty_WithoutHandle()
```
vladimir-krestov, line 66: Please keep this test and create new one for `Name`

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TextBoxBaseAccessibleObjectTests.cs
```diff
@@ -63,13 +63,13 @@ public void TextBoxBaseAccessibleObject_IsReadOnly_ReturnsCorrectValue(bool read
...
+        public void TextBoxBaseAccessibleObject_Name_Default()
```
vladimir-krestov, line 66: This test name is not clear enough

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/UpDownBase.UpDownEdit.UpDownEditAccessibleObjectTests.cs
```diff
@@ -60,12 +60,12 @@ public void UpDownEditAccessibleObject_Custom_Name_ReturnsCorrectValue()
...
 
         [WinFormsFact]
-        public void UpDownEditAccessibleObject_Default_Name_ReturnsNull()
+        public void UpDownEditAccessibleObject_Default_Name_ReturnsEmpty()
         {
             using UpDownBase upDown = new SubUpDownBase();
             using UpDownBase.UpDownEdit upDownEdit = new UpDownBase.UpDownEdit(upDown);
             AccessibleObject accessibleObject = upDownEdit.AccessibilityObject;
-            Assert.Null(accessibleObject.Name);
+            Assert.Equal(string.Empty, accessibleObject.Name);
```
vladimir-krestov, line 68: Will `Assert.Empty` work?

vladimir-krestov (commented):
Looks good, but should be carefully tested

src/System.Windows.Forms/src/System/Windows/Forms/MaskedTextBox.cs
NikitaSemenovAkvelon: It is for Narrator announcements. In AO.SetFocus method notifications are raised. The answer is also relevant to the same questions in other classes.

src/System.Windows.Forms/src/System/Windows/Forms/TextBox.cs
NikitaSemenovAkvelon: Answered here: https://github.com/dotnet/winforms/pull/7741#discussion_r968320837

src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.cs
NikitaSemenovAkvelon: The first case was removed because of `UiaCore.UiaDisconnectProvider` is called from another place. Moreover, it is unreachable now.
But the second case is still here, below.

src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.UpDownEdit.cs
NikitaSemenovAkvelon: Because it is a very old line of code. Without it the control works identically. But the line breaks the logic of calculating IsKeyboardFocusableProperty value. 
Moreover, via MSAA it always returns as true, even if control is disabled. But currently, it depends on its styles.
Do you think that is a good idea to put a comment about the removed code? I think it may be confusing.

src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.UpDownEdit.cs
NikitaSemenovAkvelon: The base method contains additional logic which may be unnecessary. Therefore we don't call it here.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TextBoxBaseAccessibleObjectTests.cs
NikitaSemenovAkvelon: Currently, the test doesn't make sense. `AO.Value` strongly referenced to `Owner.Text`, as in MSAA.

src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.TextBoxBaseAccessibleObject.cs
NikitaSemenovAkvelon: 1. Yes, it works in Explorer and via MSAA.
2. Fixed and added a unit test for the SetValue method.

src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.cs
dmitrii-drobotov: > The first case was removed because of `UiaCore.UiaDisconnectProvider` is called from another place. 

It's called only when control has `SupportsUiaProviders` as `true`, but there is `RichTextBox` for which it's `false`. RichTextBox could still somehow have an active `AccessibilityObject` for which we need to call `UiaCore.UiaDisconnectProvider`, so you definitely should keep it. Although maybe it would make sense to move it to `RichTextBox` class if it's the only affected control now.

dmitrii-drobotov (commented):
Nice work! Did you also test that there are no memory leaks for accessible objects after your changes?

src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.cs
NikitaSemenovAkvelon: Oh, it's true. I've moved this logic to `RichTextBox`.

Tanya-Solyanik: >RichTextBox could still somehow have an active AccessibilityObject

RichTextBox should not register a `IRawElementProviderSimple` (because `SupportsUiaProviders` is false) and the MSAA Proxy provider should be ref-counted by the proxy. So we should not need to disconnect in this case and our disconnect call should not do anything.....

src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.UpDownEdit.cs
Tanya-Solyanik: Sorry, I don't follow why this line was removed. ControlStyles are part of the control. Provider should represent this control. Mayve providers should not take this style into account?

src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.TextBoxBaseAccessibleObject.cs
```diff
@@ -50,11 +50,34 @@ internal override bool IsPatternSupported(UIA patternId)
...
+
+            public override string? Name => Owner.AccessibleName ?? string.Empty;
```
Tanya-Solyanik, line 60: Why does the base class implementation not work? TextBox should get AccessibleName from the label placed before it in the tab order. You dropped that part of the implementation. Maybe it does not make sense for the DGV  cells, but it makes sense for other TextBoxes.

src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.TextBoxBaseAccessibleObject.cs
```diff
@@ -50,11 +50,34 @@ internal override bool IsPatternSupported(UIA patternId)
...
+            internal override void SetFocus()
+            {
+                if (!Owner.IsHandleCreated)
+                {
+                    return;
+                }
+
+                base.SetFocus();
+
+                RaiseAutomationEvent(UIA.AutomationFocusChangedEventId);
```
Tanya-Solyanik, line 73: This implementation probably belongs to the ControlAccessibleObject. But not in this PR.

src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.UpDownEdit.UpDownEditAccessibleObject.cs
```diff
@@ -24,7 +24,7 @@ public UpDownEditAccessibleObject(UpDownEdit owner, UpDownBase parent) : base(ow
                         {
                             NumericUpDown _ => SR.EditDefaultAccessibleName,
                             DomainUpDown _ => SR.EditDefaultAccessibleName,
-                            _ => null
+                            _ => string.Empty
```
Tanya-Solyanik, line 27: Why did you change that? null and "" are interpreted differently by UIAutomation

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/ToolStripTests.Designer.cs
```diff
@@ -201,6 +203,12 @@ private void InitializeComponent()
             this.toolStripLabel1.Size = new System.Drawing.Size(86, 22);
             this.toolStripLabel1.Text = "ToolStripLabel";
             // 
+            // toolStripTextBox1
```
Tanya-Solyanik, line 206: 👍

Tanya-Solyanik (commented):
Looks good, I added some comments, please see them. Had you compared the resulting provider set and properties to what is required by the docs https://docs.microsoft.com/windows/win32/winauto/uiauto-supporteditcontroltype?

src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.TextBoxBaseAccessibleObject.cs
NikitaSemenovAkvelon: When I compared my implementation with MSAA behavior, the `AccessibleName` was empty. It was a reason why I implemented it that way. It was probably wrong and I've reworked it through the documentation above.

src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.UpDownEdit.cs
NikitaSemenovAkvelon: Vladimir tried to investigate why the line was put here. He looked into branches that were created in 2010 and the line was still here. So, we didn't find the reason for that code and we decided that it may be obsolete.
However, it doesn't make sense, because the control is selectable. The property `IsKeyboardFocusableProperty` is depending on `Control.CanSelect` which is calculated here and uses this style (https://github.com/dotnet/winforms/blob/main/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs#L4862).

src/System.Windows.Forms/src/PublicAPI.Unshipped.txt
```diff
@@ -2,3 +2,5 @@ System.Windows.Forms.PrintPreviewControl.TabStop.get -> bool
 System.Windows.Forms.PrintPreviewControl.TabStop.set -> void
 System.Windows.Forms.PrintPreviewControl.TabStopChanged -> System.EventHandler?
 *REMOVED* override System.Windows.Forms.ToolStripButton.CanSelect.get -> bool
+override System.Windows.Forms.MaskedTextBox.OnGotFocus(System.EventArgs! e) -> void
+*REMOVED*override System.Windows.Forms.DataGridViewTextBoxEditingControl.OnGotFocus(System.EventArgs! e) -> void
```
Tanya-Solyanik, line 6: Add space after the `removed` prefix

Tanya-Solyanik: And remove this API from the `shipped` file.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewTextBoxEditingControl.cs
```diff
@@ -23,10 +23,8 @@ public DataGridViewTextBoxEditingControl() : base()
             TabStop = false;
         }
 
-        protected override AccessibleObject CreateAccessibilityInstance()
-        {
-            return new DataGridViewTextBoxEditingControlAccessibleObject(this);
-        }
+        protected override AccessibleObject CreateAccessibilityInstance() =>
```
Tanya-Solyanik, line 26: I think overall it's more common to start the line with `=>` rather than to leave it on the previous line in this codebase. My reasoning is that CS is read left to right, so the more important statements should be on the left.

src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.UpDownEdit.cs
Tanya-Solyanik: I suppose this control being an internal one, we can change this property, as no one should be calling `CanSelect` on it.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TextBoxBaseAccessibleObjectTests.cs
```diff
@@ -49,6 +51,18 @@ public void TextBoxBaseAccessibleObject_TextPatternSupported(int patternId)
             Assert.False(textBoxBase.IsHandleCreated);
         }
 
+        [WinFormsFact]
+        public void TextBoxBaseAccessibleObject_SetValue_OwnerTextChanged()
+        {
+            const string text = "some text";
+            using TextBoxBase textBoxBase = new SubTextBoxBase();
+            textBoxBase.CreateControl();
+            AccessibleObject accessibleObject = textBoxBase.AccessibilityObject;
```
Tanya-Solyanik, line 60: Consider using a discard
`_ = textBoxBase.AccessibilityObject;`

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TextBoxBaseAccessibleObjectTests.cs
```diff
@@ -49,6 +51,18 @@ public void TextBoxBaseAccessibleObject_TextPatternSupported(int patternId)
...
+            accessibleObject.SetValue(text);
+            Assert.Equal(text, textBoxBase.Text);
```
Tanya-Solyanik, line 62: Please insert a blank line before the validation.

Tanya-Solyanik (commented):
Please see the question about the RichTextBox

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TextBoxBaseAccessibleObjectTests.cs
NikitaSemenovAkvelon: I'm using the variable in a next line.

src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.TextBoxBaseAccessibleObject.cs
dmitrii-drobotov: I see that now accessible names match the TextBox content:
Accessibility Insights shows accessible names of TextBoxes which match its content

I don't think this is expected behavior because it makes Narrator announce the content twice:

> TextBoxes window, Some long long text for a multiline TextBox that contains several lines with withspaces,tabs and numbers 12345, edit, Some long long text for a multiline TextBox that contains several lines with withspaces,tabs and numbers 12345, 
> Some long long text for a RTL TextBox, edit, Some long long text for a RTL TextBox, 
> Some long long text for a TextBox, edit, Some long long text for a TextBox, 

If there is a label before TextBox then name will be filled from it, but I think the default name should be empty.

dmitrii-drobotov (commented):
Looks good! I also checked the [accessibility docs](https://learn.microsoft.com/en-us/windows/win32/winauto/uiauto-supporteditcontroltype) for edit control and found some inconsistencies. For example:

> If an edit control contains placeholder text (for example, a cue banner), the text should be used as the HelpText property unless the text can be edited by the user and then reused as placeholder text. For example, the Windows Internet Explorer address bar contains the text "about:Tabs" when a new tab is opened. This is not HelpText because it is a programmatic address that can be used or edited by the user.

We have the `PlaceholderText` property, so it's possible to set a placeholder, but the `HelpText` property won't be set, because it's not implemented in this PR. But what I found is that this functionality didn't work before, so this PR won't be a regression.

I didn't go through all the docs content but there is at least `UIA_IsPasswordPropertyId` with the same problem — it's not implemented in this PR but wasn't working before either. Should we fix all these inconsistencies in this PR or later as separate issues?

Tanya-Solyanik:
@dmitrii-drobotov  - great findings! Yes, please open a new issue.

src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.cs
Tanya-Solyanik: Had you tested rich text boxes after this change?

src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.TextBoxBaseAccessibleObject.cs
Tanya-Solyanik: I see this as an application problem that should be raised by the AccessibilityInsights for the app developer to fix. Framework does not know the business logic behind this text box and can't assign a meaningful name. As long as tooling raises the issue, we are good. 

If there is a code path to keep Accessible Name empty , it would be a better fix, but I wouldn't spend much time on it (I think we have a property that controls it)

CodeDummy-03 (dismissed):
Looks good to me! ... Good work :)

Tanya-Solyanik:
@NikitaSemenovAkvelon - please confirm if you had test rich textbox for leaks and send this to the CTI team

NikitaSemenovAkvelon:
> @NikitaSemenovAkvelon - please confirm if you had test rich textbox for leaks and send this to the CTI team

Confirmed. There are no leaks of TextBox*AccessibleObject after closing a window with RichTextBox.
image

Tanya-Solyanik:
@NikitaSemenovAkvelon  - is this the same set of objects as we have before the fix? BTW, I don't find the screenshot useful because it's not clear what code you executed to  get there. Specifically, did you run GC?

NikitaSemenovAkvelon:
> @NikitaSemenovAkvelon - is this the same set of objects as we have before the fix? BTW, I don't find the screenshot useful because it's not clear what code you executed to get there. Specifically, did you run GC?

Oh, sorry. I meant the RichTextBoxes form. The version without a fix has the same set of objects. Also, I didn't call GC.
image

Tanya-Solyanik (approved):
Thank you! Please sent to the CTI team for testing if you had not done already.

NikitaSemenovAkvelon:
Fixed cases that CTI found.

src/System.Windows.Forms/src/System/Windows/Forms/MaskedTextBox.cs
```diff
@@ -2030,6 +2044,11 @@ protected override void OnKeyUp(KeyEventArgs e)
             {
                 _flagState[IME_ENDING_COMPOSITION] = false;
             }
+
+            if (IsHandleCreated && ContainsNavigationKeyCode(e.KeyCode))
+            {
+                AccessibilityObject?.RaiseAutomationEvent(UiaCore.UIA.Text_TextSelectionChangedEventId);
```
dmitrii-drobotov, line 2050: Please add a check for `IsAccessibilityObjectCreated`, because `AccessibilityObject` accessor creates an accessible object even if user doesn't use any accessibility tools.

dmitrii-drobotov (changes requested):

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/MaskedTextBox.MaskedTextBoxAccessibleObjectTests.cs
```diff
@@ -74,6 +74,35 @@ public void MaskedTextBoxAccessibleObject_GetPropertyValue_ControlType_IsExpecte
...
+
+            var actual = (string)maskedTextBox.AccessibilityObject.GetPropertyValue(UiaCore.UIA.ValueValuePropertyId);
+
+            Assert.Equal(maskedTextBox.WindowText, actual);
+            Assert.Equal(useMask, maskedTextBox.Mask?.Length == actual.Length);
+            Assert.False(maskedTextBox.IsHandleCreated);
+        }
+
+        [WinFormsFact]
+        public void MaskedTextBoxAccessibleObject_GetPropertyValue_Value_AccessDenied_WithUseSystemPasswordChar()
```
dmitrii-drobotov, line 94: I think it would make sense to add a similar test for regular `TextBox` to avoid possible regressions for it as well.

dmitrii-drobotov (approved):

Tanya-Solyanik (approved):

MelonWang1:
Verified this issue with latest .NET 8.0.100-alpha.1.22512.27, Add support UIA provider for TextBox and MaskedTextBox controls, screenshort as below:
Textbox:
inspect-textbox
ai-textbox

MaskedTextbox:
inspect-masktextbox
ai-masktextbox

PR #7788, opened 2022-09-16
Released AccessibleObjects of ToolStripItems

### LINKED ISSUE #7713: ToolStrip class and it's accessible object are leaking in memory
### .NET version

.NET 7.0

No

No, it can repro in .NET 6.0.

### Issue description
Focused on ToolStrip controls with items in Narrator, Accessibility Insight and Inspect, have a relatively large number of memory leaks.

https://user-images.githubusercontent.com/56664234/187878642-e986b0c4-562c-49e3-a7bd-e4ef6b98c9c0.mp4

### Expected behavior:

### Steps to reproduce

1.Create a parent form with two buttons: one will be opening a child form with the ToolStrip control and the other will collect the garbage.
2. Press the second button to open the child form with ToolStrip.
3. Start project without debugging.
4. Attaching project to WinDbg and adding symbols folder.
5. Start the Narrator tool.
3. Automatic reading for the ToolStrip control using Narrator.
6. Close the child form.
7. Clear memory with the first button of the parent form.
8. Don't close the Narrator tool.
9. Look for ToolStrip accessible objects left in memory (using WinDbg).
TestApp: [WinFormsApp5_ToolStrip.zip](https://github.com/dotnet/winforms/files/9468756/WinFormsApp5_ToolStrip.zip)

NikitaSemenovAkvelon:
Fixes #7713 

## Proposed changes

- Added a condition to access `AccessibleObject` via `FragmentNavigate` for `ComboBoxChildEditUiaProvider` and `ComboBoxChildListUiaProvider` if an owning `ComboBox` wasn't created.
- Released `Grip` and `OverflowButton` with `ToolStrip` releasing.
- Released `Control` with `ToolStripControlHost` releasing.
- Released `DropDown` with `ToolStripDropDownItem` releasing.
- Added unit tests.

## Customer Impact

- No memory leaks from `ToolStrip`.

## Regression? 

- No

## Risk

- Minimal

### Before

image

### After

image

## Test methodology 

- Manually
- WinDbg

## Accessibility testing  

- Inspect

 

## Test environment(s) 

- 8.0.0-alpha.1.22464.17

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7788)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxChildEditUiaProvider.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxChildEditUiaProvider.cs
@@ -42,7 +42,10 @@ namespace System.Windows.Forms
             /// <returns>Returns the element in the specified direction.</returns>
             internal override UiaCore.IRawElementProviderFragment? FragmentNavigate(UiaCore.NavigateDirection direction)
             {
-                if (!_owningComboBox.IsHandleCreated)
+                if (!_owningComboBox.IsHandleCreated ||
+                    // Created is set to false in WM_DESTROY, but the window Handle is released on NCDESTROY, which comes after DESTROY.
+                    // But between these calls, AccessibleObject can be recreated which is going to memory leaking.
+                    !_owningComboBox.Created)
                 {
                     return null;
                 }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxChildEditUiaProvider.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxChildEditUiaProvider.cs
@@ -44,7 +44,7 @@ namespace System.Windows.Forms
             {
                 if (!_owningComboBox.IsHandleCreated ||
                     // Created is set to false in WM_DESTROY, but the window Handle is released on NCDESTROY, which comes after DESTROY.
-                    // But between these calls, AccessibleObject can be recreated which is going to memory leaking.
+                    // But between these calls, AccessibleObject can be recreated and might cause memory leaks.
                     !_owningComboBox.Created)
                 {
                     return null;
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxChildListUiaProvider.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxChildListUiaProvider.cs
@@ -65,7 +65,7 @@ namespace System.Windows.Forms
             {
                 if (!_owningComboBox.IsHandleCreated ||
                     // Created is set to false in WM_DESTROY, but the window Handle is released on NCDESTROY, which comes after DESTROY.
-                    // But between these calls, AccessibleObject can be recreated which is going to memory leaking.
+                    // But between these calls, AccessibleObject can be recreated and might cause memory leaks.
                     !_owningComboBox.Created)
                 {
                     return null;
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxChildListUiaProvider.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxChildListUiaProvider.cs
@@ -63,7 +63,10 @@ namespace System.Windows.Forms
             /// <returns>Returns the element in the specified direction.</returns>
             internal override UiaCore.IRawElementProviderFragment? FragmentNavigate(UiaCore.NavigateDirection direction)
             {
-                if (!_owningComboBox.IsHandleCreated)
+                if (!_owningComboBox.IsHandleCreated ||
+                    // Created is set to false in WM_DESTROY, but the window Handle is released on NCDESTROY, which comes after DESTROY.
+                    // But between these calls, AccessibleObject can be recreated which is going to memory leaking.
+                    !_owningComboBox.Created)
                 {
                     return null;
                 }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.cs
@@ -876,6 +876,12 @@ namespace System.Windows.Forms
 
         private void SuspendSizeSync() => _suspendSyncSizeCount++;
 
+        internal override void ReleaseUiaProvider()
+        {
+            _control?.ReleaseUiaProvider(IntPtr.Zero);
+            base.ReleaseUiaProvider();
+        }
+
         private void ResumeSizeSync() => _suspendSyncSizeCount--;
 
         internal override bool ShouldSerializeBackColor()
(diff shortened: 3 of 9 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxChildEditUiaProvider.cs
```diff
@@ -42,7 +42,7 @@ public ComboBoxChildEditUiaProvider(ComboBox owner, IntPtr childEditControlhandl
             /// <returns>Returns the element in the specified direction.</returns>
             internal override UiaCore.IRawElementProviderFragment? FragmentNavigate(UiaCore.NavigateDirection direction)
             {
-                if (!_owningComboBox.IsHandleCreated)
+                if (!_owningComboBox.IsHandleCreated || !_owningComboBox.Created)
```
Tanya-Solyanik, line 45: Could you please explain the added test? Are you testing Visibility?

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.cs
```diff
@@ -876,6 +876,12 @@ protected override void SetVisibleCore(bool visible)
 
         private void SuspendSizeSync() => _suspendSyncSizeCount++;
 
+        internal override void ReleaseUiaProvider()
```
Tanya-Solyanik, line 879: Could you please add unit test for this one too?

Tanya-Solyanik (commented):
Looks good, please see my question and send to the CTI team.

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxChildEditUiaProvider.cs
NikitaSemenovAkvelon: Which test do you mean? I've not seen any difference between `Visibility` and without it. 
This change should be here because the `ComboBoxChild*UiaProvider` was destroyed after owning `ComboBox` and native libraries call `FragmentNavigate` with `Parent` direction. After that `ComboBoxAccessibleObject` was recreated. I've fixed it by checking that `ComboBox` wasn't created.

Tanya-Solyanik: I mean this test - `!_owningComboBox.Created`.

WM_DESTROY handler for the combo box disposes and unregisters the UIA providers for child controls. But the accessibility tool might still have the provider proxy and might navigate using it. `Create` is set to false in WM_DESTROY, but the window Handle is released on NCDESTROY, which comes after DESTROY. So you are covering the interval between these messages?
We might have the same problem in other Accessible objects, please open a work item to review them.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripComboBox.ToolStripComboBoxAccessibleObjectTests.cs
```diff
@@ -0,0 +1,31 @@
...
+// See the LICENSE file in the project root for more information.
+
+using Xunit;
+
+namespace System.Windows.Forms.Tests
+{
+    public class ToolStripComboBox_ToolStripComboBoxAccessibleObjectTests
+    {
+        [WinFormsFact]
+        public void ToolStripComboBoxAccessibleObject_ReleaseUiaProvider_ClearToolStripComboBoxControl()
```
Tanya-Solyanik, line 12: nit: remove word `Clear` because we already say `ReleaseUIAProvider`.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripComboBox.ToolStripComboBoxAccessibleObjectTests.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            using var toolStrip = new ToolStrip();
+            using var toolStripComboBox = new ToolStripComboBox();
+            toolStrip.Items.Add(toolStripComboBox);
+            toolStripComboBox.Parent = toolStrip;
+            toolStrip.CreateControl();
+
+            // Just create accessible objects.
+            var toolStripComboBoxAccessibilityObject = toolStripComboBox.AccessibilityObject;
```
Tanya-Solyanik, line 21: `_ =  toolStripComboBox.AccessibilityObject;` and then comment is not needed because it's clear that we are not using the variable directly.
The same applies to other tests in this PR as well.

Tanya-Solyanik (approved):
Good job finding the `!Created` condition, please see if it applies to other UIA providers as well and open a follow up item.

Tanya-Solyanik (approved):

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxChildEditUiaProvider.cs
NikitaSemenovAkvelon: > Create is set to false in WM_DESTROY, but the window Handle is released on NCDESTROY, which comes after DESTROY. 
Yes, I've retested it, and it is the true reason.

dmitrii-drobotov (dismissed):
Looks good!

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxChildEditUiaProvider.cs
Tanya-Solyanik: BTW, it might be useful to add an explanation comment about it. In fact whenever reviewer asks  - why is this done this way,   it means a comment is missing.

Ashley-Li:
Tested private binaries based on the latest .NET 8.0 SDK build for this PR, focused on ToolStrip controls with items in Accessibility Insight, Inspect, and Narrator tools with WinDbg, there's no toolStrip items memory leak about AccessibleObjects.

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxChildListUiaProvider.cs
```diff
@@ -63,7 +63,10 @@ public ComboBoxChildListUiaProvider(ComboBox owningComboBox, IntPtr childListCon
             /// <returns>Returns the element in the specified direction.</returns>
             internal override UiaCore.IRawElementProviderFragment? FragmentNavigate(UiaCore.NavigateDirection direction)
             {
-                if (!_owningComboBox.IsHandleCreated)
+                if (!_owningComboBox.IsHandleCreated ||
+                    // Created is set to false in WM_DESTROY, but the window Handle is released on NCDESTROY, which comes after DESTROY.
+                    // But between these calls, AccessibleObject can be recreated which is going to memory leaking.
```
Tanya-Solyanik, line 68:
```suggestion
                    // But between these calls, AccessibleObject can be recreated and might cause memory leaks.
```

Tanya-Solyanik (approved):

PR #7793, opened 2022-09-17
Explicitly declare HighDpiMode in the project file for integration test

Tanya-Solyanik:
... for easy experiments with DPI awareness.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7793)

### DIFF
--- a/src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/WinformsControlsTest.csproj
+++ b/src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/WinformsControlsTest.csproj
@@ -12,6 +12,13 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
 
+  <!--ApplicationConfiguration specific settings.-->
+  <!--https://github.com/dotnet/designs/blob/main/accepted/2021/winforms/streamline-application-bootstrap.md-->
+  <PropertyGroup>
+    <ApplicationDefaultFont>Calibri, 11pt, style=regular</ApplicationDefaultFont>
+    <ApplicationHighDpiMode>SystemAware</ApplicationHighDpiMode>
+  </PropertyGroup>
+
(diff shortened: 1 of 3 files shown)


src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/WinformsControlsTest.csproj
```diff
@@ -10,6 +10,7 @@
     <EnableXlfLocalization>false</EnableXlfLocalization>
     <UpdateXlfOnBuild>false</UpdateXlfOnBuild>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <ApplicationHighDpiMode>SystemAware</ApplicationHighDpiMode>
```
RussKie, line 13: We declare the settings globally here:
https://github.com/dotnet/winforms/blob/1b17d5b15bebc394af0a697c3ee0dffe285ae3b5/Directory.Build.props#L47-L50

So we should either declare the settings there, or move here.

Tanya-Solyanik: I added Dpi mode to the test project in order to make it easier to experiment with this particular application. Global settings can stay unchanged.

RussKie: ApplicationDefaultFont is there for the exact same purpose. So it's worth having all related options in the same place. I have no strong preference which place though.

Tanya-Solyanik: I see, the root of confusion is that I have multiple test exes in my branch

RussKie (approved):

PR #7865, opened 2022-09-29
Convert some Ole32 functions to CsWin32

elachlan:
Related: #7445 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7865)


src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs
```diff
@@ -848,7 +848,11 @@ internal ApartmentState OleRequired()
                 _ = Thread.CurrentThread;
                 if (!GetState(STATE_OLEINITIALIZED))
                 {
-                    HRESULT ret = Ole32.OleInitialize(IntPtr.Zero);
+                    HRESULT ret;
+                    unsafe
+                    {
+                        ret = PInvoke.OleInitialize((void*)IntPtr.Zero);
```
lonitra, line 854: [docs](https://learn.microsoft.com/en-us/windows/win32/api/ole2/nf-ole2-oleinitialize) mention that the param must be `null` so can we just pass `null` here?

lonitra (commented):
Looks like there are still some remaining references to `Interop.Ole32.RevokeDragDrop`. Otherwise changes so far looks good.

src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs
elachlan: @AArnott there are lots of reserved parameters and often they should just be null. Should we add a friendly overrides to handle them? Would we need to tag them in metadata?

RussKie (approved):

src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs
AArnott: Yes, friendly overloads are probably a good choice for dropping such parameters. Even the `extern` methods themselves could set the ones that are at the tail end of the parameter list to have default arguments so they are optional as well. And yes, we'd look for the metadata to designate which parameters are expected to always be set to their default values.

elachlan: Raised an issue here: https://github.com/microsoft/win32metadata/issues/1288

lonitra (approved):
Looks good, thank you

PR #7867, opened 2022-09-29
Convert window positioning APIs to CsWin32

JeremyKuhne:
Also convert a few dialog related APIs.
Some minor comment and code cleanup as well.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7867)


lonitra (commented):
LGTM, just nits

JeremyKuhne:
Ugh. Something is up with https://github.com/dotnet/winforms/pull/6368's test. Still trying to figure out precisely what is going wrong.

JeremyKuhne:
> Ugh. Something is up with https://github.com/dotnet/winforms/pull/6368 test. Still trying to figure out precisely what is going wrong.

Well, `WINDOWPLACEMENT` was defined wrong in our existing code (wrong size). As it was too big I don't think it surfaced anywhere. Windows doesn't check the length field when getting, and sets it to the correct size. Setting does check, but we're doing it with a returned value so I think we were always good.

lonitra (approved):

Issue #7518, opened 2022-07-29
.NET 7.0 preview DragEnter bug after update to latest versions. 


c-ohle:
#### When trying to drag and drop files from folders or other sources into a Windows.Forms.Controls:

NotImplementedException in [IDropTargetVtbl.cs](https://source.dot.net/#System.Windows.Forms.Primitives/Interop/WinFormsComWrappers.IDropTargetVtbl.cs,77)
VS2022 preview, .NET 7.0 preview, SDK, all latest versions.

It works compiled for  .NET 6.0, it worked for all .NET 7.0 previews before.
Example project is available at [GitHub](https://github.com/c-ohle/RationalNumerics)

![image](https://user-images.githubusercontent.com/63112403/181831855-a223c635-d765-4857-883a-3a398c4a61d4.png)

### Reproduction Steps

Should be easy to reproduce: 
VS2022 - New Project - Windows Forms App - .NET 7.0 preview
Form1 - AllowDrop = true; - override OnDragEnter
on file drop -> exception

### Expected behavior

Should work or message/exception if there is a security problem or what ever. 

### Actual behavior

NotImplementedException

### Known Workarounds

no

### Configuration

Processor: Intel(R) Core(TM) i5-10210U CPU @ 1.60GHz (2112 MHz) × 8
System: Microsoft Windows 10.0.22622 X64 Runtime: .NET 7.0.0-preview.6.22324.4 X64
BigRational: 1.0.0.5 Runtime: v4.0.30319 .NETCoreApp,Version=v7.0 

For Test project on: [GitHub sample](https://github.com/c-ohle/RationalNumerics)

dreddy-work:
@kant2002, can you take a quick look to see if your latest changes in anyway causing this?

kant2002:
I think issue was introduced in https://github.com/dotnet/winforms/pull/6976

c-ohle:
@kant2002 Yes, this seems to be the same problem. But there is something more. 

Since an earlier NET7 version (3 or so) something is wrong with COM refcounts.
For example in my test app: Previously it was possible to drag and drop 3D objects from the 3D view of one app instance to the 3D view of another app instance.
Now the source app instance hangs after a some deep internal COM exceptions.
Compiled with Target NET 6.0 it works.

So it's easy to reproduce with the app:
[https://github.com/c-ohle/RationalNumerics](https://github.com/c-ohle/RationalNumerics)

c-ohle:
[...]

PR #7869, opened 2022-09-30
Restore back ComWrappers for DoDragDrop

### LINKED ISSUE #7864: IDataObject.GetData with non-serializable type returns null
### .NET version

.NET 7 RC1

Yes

Yes, up to and including .NET 6.

### Issue description

Hello,

My name is Norton, and I'm writing on behalf of the WinForms team at DevExpress.

Our customers who were testing our controls with the .NET 7 RC1 runtime reported issues when using our controls' drag-drop capabilities. Upon further research, we found what seems to be a breaking change in .NET 7: whenever an instance of a non-serializable class is used with the [`Control.DoDragDrop`](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.control.dodragdrop) method, calling [`IDataObject.GetData`](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.idataobject.getdata) with that class as a parameter afterwards returns `null` rather than the dragged object.

We could not find any issues/PRs or MSDN articles documenting this change. Could you have a look and let us know if this change is intended?

### Steps to reproduce

I attached a sample application. To replicate this behavior, run the project and start dragging anywhere on the form. The [`Debugger.Break`](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.debugger.break) method is called to indicate that the `GetData` method returned `null`.

If the type of the dragged object is made serializable (e.g. by modifying the `using` statement in `Form1.cs`), this behavior no longer takes place.

[DXApplication1.zip](https://github.com/dotnet/winforms/files/9674453/DXApplication1.zip)

kant2002:
Fixes #7864

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7869)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/Interop/Ole32/Interop.DoDragDrop.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/Ole32/Interop.DoDragDrop.cs
@@ -9,36 +9,11 @@ internal static partial class Interop
 {
     internal static partial class Ole32
     {
-        [DllImport(Libraries.Ole32)]
-        private static extern HRESULT DoDragDrop(
-            IntPtr pDataObj,
-            IntPtr pDropSource,
-            DROPEFFECT dwOKEffects,
-            out DROPEFFECT pdwEffect);
-
-        public static HRESULT DoDragDrop(
+        [DllImport(Libraries.Ole32, ExactSpelling = true)]
+        public static extern HRESULT DoDragDrop(
             IDataObject pDataObj,
             IDropSource pDropSource,
             DROPEFFECT dwOKEffects,
-            out DROPEFFECT pdwEffect)
-        {
-            var result = WinFormsComWrappers.Instance.TryGetComPointer(pDataObj, IID.IDataObject, out var dataObjectPtr);
-            if (result.Failed)
-            {
-                pdwEffect = DROPEFFECT.NONE;
-                return result;
-            }
-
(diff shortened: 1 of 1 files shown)


kant2002:
Also test with project mentioned here https://github.com/dotnet/winforms/issues/7518#issuecomment-1203643595 no regressions

willibrandon:
I tested this PR using the test provided in the issue description and confirmed it fixes the issue.

Before
https://user-images.githubusercontent.com/5017479/193437723-0525b074-007e-4415-8d51-c1133d4cb8f8.mp4

After
https://user-images.githubusercontent.com/5017479/193437729-a14c2fab-7ec1-4dfb-8a34-10b4f6e653a8.mp4

I can write the regression test if you would like.

kant2002:
@willibrandon if you can, please write regression test. We definitely need this kind of things

willibrandon:
@kant2002 - I'm not sure your current project status, but I can provide test support, and sooner or later we'll have all of the cases covered :smile: As we have found, drag and drop could use more tests, and I'm in a good spot to write those now.

willibrandon:
I’ve written a UI regression test that reproduces the issue and confirms the fix.

kant2002:
Send PR with test. let it fail, and you can cherry pick this commit so we would check that it fixed your new test.

JeremyKuhne (approved):
Thanks for the quick turnaround here.

RussKie:
Thank you @kant2002!

PR #7876, opened 2022-10-02
Add serialized and non-serialized object drag and drop UI tests

willibrandon:
Fixes #7875 

## Proposed changes

- Add UI tests which drag and drop serialized and non-serialized data to the application.
- I've written 2 tests covering both the serialized and non-serialized scenarios, with the serialized case working as expected, and the non-serialized case failing, as reported by the customer.
- Initial results show that the fix in https://github.com/dotnet/winforms/pull/7869 alleviates the issue described in https://github.com/dotnet/winforms/issues/7864.

## Customer Impact

- Supports the work being done in #5163 
- Regression test for #7864 

## Regression? 

- Yes

## Risk

- Minimal

## Test methodology 

- Ran unit and integration tests.

## Test environment(s) 

- .NET 7 RC1

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7876)

PR #7878, opened 2022-10-03
Port fix for IDataObject.GetData with non-serializable type returns null

### LINKED ISSUE #7864: IDataObject.GetData with non-serializable type returns null

dreddy-work:
Port #7869 and #7876 

Fixes #7864

## Proposed changes

Revert ComWrapper usage to allow managed types to cycle through drag and drop.

## Customer Impact

Without this change, drag and drop does not work in many (most?) customer applications.

## Regression? 

- Yes

## Risk

- Low. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7878)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/Interop/Ole32/Interop.DoDragDrop.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/Ole32/Interop.DoDragDrop.cs
@@ -9,36 +9,11 @@ internal static partial class Interop
 {
     internal static partial class Ole32
     {
-        [LibraryImport(Libraries.Ole32)]
-        private static partial HRESULT DoDragDrop(
-            IntPtr pDataObj,
-            IntPtr pDropSource,
-            DROPEFFECT dwOKEffects,
-            out DROPEFFECT pdwEffect);
-
-        public static HRESULT DoDragDrop(
+        [DllImport(Libraries.Ole32, ExactSpelling = true)]
+        public static extern HRESULT DoDragDrop(
             IDataObject pDataObj,
             IDropSource pDropSource,
             DROPEFFECT dwOKEffects,
-            out DROPEFFECT pdwEffect)
-        {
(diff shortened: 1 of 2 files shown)


RussKie (approved):

RussKie:
Thank you for porting it

noxe:
will that also be available in the upcoming RC2?

dreddy-work:
@noxe , no, RC2 is already locked. This is for RTM/GA .

Issue #7887, opened 2022-10-05
Enable Analyzer disabled as a part of globalconfig merge from runtime


elachlan:
### Is your feature request related to a problem? Please describe

List of diagnostics that need to be investigated, fixed, and enabled.
```
Line  260: dotnet_diagnostic.CA1727.severity = suggestion # TODO: warning
Line  273: dotnet_diagnostic.CA1810.severity = none # TODO: warning
Line  297: dotnet_diagnostic.CA1821.severity = none # TODO: warning
Line  394: dotnet_diagnostic.CA1853.severity =  none # TODO: warning
Line  397: dotnet_diagnostic.CA1854.severity =  none # TODO: warning
Line  406: dotnet_diagnostic.CA2007.severity = none # TODO: warning
Line  439: dotnet_diagnostic.CA2019.severity = none # TODO: warning
Line  466: dotnet_diagnostic.CA2208.severity = none # TODO: warning
Line  533: dotnet_diagnostic.CA2245.severity = none # TODO: warning
Line  557: dotnet_diagnostic.CA2253.severity = none # TODO: warning
Line  851: dotnet_diagnostic.IL3000.severity = none # TODO: warning
Line  873: dotnet_diagnostic.SA1001.severity = none # TODO: warning
Line  912: dotnet_diagnostic.SA1014.severity = none # TODO: warning
Line  924: dotnet_diagnostic.SA1018.severity = none # TODO: warning
Line  930: dotnet_diagnostic.SA1020.severity = none # TODO: warning
Line  948: dotnet_diagnostic.SA1026.severity = none # TODO: warning
Line  951: dotnet_diagnostic.SA1027.severity = none # TODO: warning
Line  954: dotnet_diagnostic.SA1028.severity = none # TODO: warning
Line  963: dotnet_diagnostic.SA1102.severity = none # TODO: warning
Line  966: dotnet_diagnostic.SA1105.severity = none # TODO: warning
Line  996: dotnet_diagnostic.SA1113.severity = none # TODO: warning
Line 1002: dotnet_diagnostic.SA1115.severity = none # TODO: warning
Line 1020: dotnet_diagnostic.SA1121.severity = none # TODO: warning
Line 1044: dotnet_diagnostic.SA1129.severity = none # TODO: warning
Line 1074: dotnet_diagnostic.SA1141.severity = none # TODO: warning
Line 1077: dotnet_diagnostic.SA1142.severity = none # TODO: warning
[...]

PR #7889, opened 2022-10-05
Enabling analyzers that don't cause build failures and fix CA1853

elachlan:
Fix CA1853: Unnecessary call to 'Dictionary.ContainsKey(key)'
Enables analyzers that don't cause build failures that were tagged with TODO.

Related: #7887

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7889)

### DIFF
--- a/eng/CodeAnalysis.src.globalconfig
+++ b/eng/CodeAnalysis.src.globalconfig
@@ -257,7 +257,7 @@ dotnet_diagnostic.CA1724.severity = none
 dotnet_diagnostic.CA1725.severity = suggestion
 
 # CA1727: Use PascalCase for named placeholders
-dotnet_diagnostic.CA1727.severity = suggestion # TODO: warning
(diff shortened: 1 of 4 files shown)


eng/CodeAnalysis.src.globalconfig
```diff
@@ -257,7 +257,7 @@ dotnet_diagnostic.CA1724.severity = none
 dotnet_diagnostic.CA1725.severity = suggestion
```
RussKie, line 257: Do we need to apply the same to the test config as well?

elachlan: Sure, I will add it to the PR.

RussKie (approved):

RussKie:
Thank you

PR #7890, opened 2022-10-05
Fixing the AutomationId property for AccessibleObject of options in ContentUI and DockUI forms

### LINKED ISSUE #7830: [Accessibility] AutomationID property should not be empty for dropdown items under ImageAlign/TextAlign/CheckAlign/Dock in PropertyGrid
### .NET version

.Net 8.0 SDK: 8.0.100-alpha.1.22469.15

Yes

In .Net 3.1, this issue cannot repro, and .Net 6.0 and 7.0 can repro

### Issue description

AutomationID property should not be empty for dropdown items under ImageAlign/TextAlign/CheckAlign/Anchor/Dock property in PropertyGrid

Inspect:
image

### Steps to reproduce

**Minimal repro:**
1. Create a winforms core app with PropertyGrid and RadioButton control.
2. Associate the Button with PropertyGrid as a selected object.
3. Build and run the core project.
4. Start Inspect tool.
5. Open the drop down list next to ImageAlign.
6. Focus on "Top Center" and observe AutomationID property in Inspect.

Tested app: 
[WinFormsApp13.zip](https://github.com/dotnet/winforms/files/9632177/WinFormsApp13.zip)

NikitaSemenovAkvelon:
Fixes #7830 

## Proposed changes

- Added `Name` to buttons in `DockUI` and `ContentUI`.

## Customer Impact

- The `AutomationId` property will be filled.

## Regression? 

- No

## Risk

- Minimal

### Before

Screenshot 2022-10-05 150607
_ContentUI_
Screenshot 2022-10-05 150630
_DockUI_

### After

Screenshot 2022-10-05 145718
_ContentUI_
Screenshot 2022-10-05 145751
_DockUI_

## Test methodology 

- Manually
- Inspect

## Accessibility testing  

- Inspect

 

## Test environment(s) 

.NET SDK:
 Version:   8.0.100-alpha.1.22423.9
 Commit:    b9635390c8

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.22621

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7890)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Drawing/Design/ContentAlignmentEditor.ContentUI.cs
+++ b/src/System.Windows.Forms.Design/src/System/Drawing/Design/ContentAlignmentEditor.ContentUI.cs
@@ -117,46 +117,55 @@ namespace System.Drawing.Design
 
                 _topLeft.TabIndex = 8;
                 _topLeft.Text = string.Empty;
+                _topLeft.Name = "_topLeft";
                 _topLeft.Appearance = Appearance.Button;
                 _topLeft.AccessibleName = SR.ContentAlignmentEditorTopLeftAccName;
 
                 _topCenter.TabIndex = 0;
                 _topCenter.Text = string.Empty;
+                _topCenter.Name = "_topCenter";
                 _topCenter.Appearance = Appearance.Button;
                 _topCenter.AccessibleName = SR.ContentAlignmentEditorTopCenterAccName;
 
                 _topRight.TabIndex = 1;
                 _topRight.Text = string.Empty;
+                _topRight.Name = "_topRight";
                 _topRight.Appearance = Appearance.Button;
                 _topRight.AccessibleName = SR.ContentAlignmentEditorTopRightAccName;
 
                 _middleLeft.TabIndex = 2;
                 _middleLeft.Text = string.Empty;
+                _middleLeft.Name = "_middleLeft";
                 _middleLeft.Appearance = Appearance.Button;
                 _middleLeft.AccessibleName = SR.ContentAlignmentEditorMiddleLeftAccName;
 
                 _middleCenter.TabIndex = 3;
                 _middleCenter.Text = string.Empty;
+                _middleCenter.Name = "_middleCenter";
                 _middleCenter.Appearance = Appearance.Button;
                 _middleCenter.AccessibleName = SR.ContentAlignmentEditorMiddleCenterAccName;
 
                 _middleRight.TabIndex = 4;
                 _middleRight.Text = string.Empty;
+                _middleRight.Name = "_middleRight";
                 _middleRight.Appearance = Appearance.Button;
                 _middleRight.AccessibleName = SR.ContentAlignmentEditorMiddleRightAccName;
 
                 _bottomLeft.TabIndex = 5;
                 _bottomLeft.Text = string.Empty;
+                _bottomLeft.Name = "_bottomLeft";
                 _bottomLeft.Appearance = Appearance.Button;
                 _bottomLeft.AccessibleName = SR.ContentAlignmentEditorBottomLeftAccName;
 
                 _bottomCenter.TabIndex = 6;
                 _bottomCenter.Text = string.Empty;
+                _bottomCenter.Name = "_bottomCenter";
                 _bottomCenter.Appearance = Appearance.Button;
                 _bottomCenter.AccessibleName = SR.ContentAlignmentEditorBottomCenterAccName;
 
                 _bottomRight.TabIndex = 7;
                 _bottomRight.Text = string.Empty;
+                _bottomRight.Name = "_bottomRight";
                 _bottomRight.Appearance = Appearance.Button;
                 _bottomRight.AccessibleName = SR.ContentAlignmentEditorBottomRightAccName;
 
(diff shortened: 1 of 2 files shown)


v-elnovikova (dismissed):

Tanya-Solyanik:
Name properly value should match the control name, ie it needs the unders core prefix. Maybe it makes sense to set automation id?

NikitaSemenovAkvelon:
> Name properly value should match the control name, ie it needs the unders core prefix. 

Hmm, I'd checked how it works in other forms, and names were set without the underscore:
https://github.com/dotnet/winforms/blob/7cdd8e390b2b6a03b2c6d2e4aab063a690cc1791/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewCellStyleBuilder.cs#L166
https://github.com/dotnet/winforms/blob/7cdd8e390b2b6a03b2c6d2e4aab063a690cc1791/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/CollectionEditor.CollectionEditorCollectionForm.cs#L431
https://github.com/dotnet/winforms/blob/7cdd8e390b2b6a03b2c6d2e4aab063a690cc1791/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/BinaryEditor.BinaryUI.cs#L155

> Maybe it makes sense to set automation id?
You mean use `RuntimeId` like in `DataGridViewCellAccessibleObject`?

NikitaSemenovAkvelon:
> How was this working in Framework? If we just need that string, i see it is already available in the `Name` property of the Inspect tool. where was it coming from before this fix?

It is a number.
image

dmitrii-drobotov:
> > How was this working in Framework? If we just need that string, i see it is already available in the `Name` property of the Inspect tool. where was it coming from before this fix?
> 
> It is a number. image

This value looks like its window handle (and also second part of Runtime Id) converted from hex to decimal (0x10F68 == 69480). Now that we usually use control's `Name` as AutomationId, I don't think it's worth restoring the old behavior, so setting `Name` looks good to me.

I also agree that it should match the variable name, i.e. contain underscores. In designer when you change "(Name)" field of a control, it changes both variable name and `Name` property, so they're synchronized. Although I guess keeping it without underscores would be also fine for consistency 🙂

dmitrii-drobotov (approved):

dreddy-work:
I was talkign bout following. where is it coming from before this change?

image

Tanya-Solyanik:
@NikitaSemenovAkvelon 
>I'd checked how it works in other forms, and names were set without the underscore:

This reflects that we were not thorough when we renamed fields in these pieces of UI. Create a new application in designer and look up Name property on a button. `Name` property is used by the design time serialization to generate fields within the piece of UI. We are not generating InitializeComponent code for this control, we wrote that code manually, thus the value of `.Name` property does not matter. If we wanted to make this UI compatible with the designer, we would have had a problem with underscores.

I'm suggesting removing `private` here - 
https://github.com/dotnet/winforms/blob/766c440e9c209194e930216f7d7811405a44a9a0/src/System.Windows.Forms/src/System/Windows/Forms/AccessibleObject.cs#L97

Tanya-Solyanik:
@dreddy-work - `Name` property that you highlighted is the accessible name, different from the design time name

dkazennov (dismissed):
Looks good.

Tanya-Solyanik:
@NikitaSemenovAkvelon  - sorry, I did not finish the thought. Next step is to implement custom accessible object for each button in this UI, inherit from RadioButton's AO, since it's public,  and implement custom AutomationID in these AccessibleObjects(perhaps as simple as programmatically remove the underscore prefix from the Name)

NikitaSemenovAkvelon:
> @NikitaSemenovAkvelon - sorry, I did not finish the thought. Next step is to implement custom accessible object for each button in this UI, inherit from RadioButton's AO, since it's public, and implement custom AutomationID in these AccessibleObjects(perhaps as simple as programmatically remove the underscore prefix from the Name)

What value should be in the `AutomationID`? Currently, it is `Owner.Name`, like a `ControlAccessibleObject`

Tanya-Solyanik:
>What value should be in the AutomationID? Currently, it is Owner.Name, like a ControlAccessibleObject

When I wrote this comment , I wanted to preserve AutometionID that we shipped in NET7 in order to not break automation tests. I.e. Control.Name with no underscore or `null` to let MSAA proxy generate something(handle).  

Now i don't think that it's worth the extra AccObject, let's use Control.Name, this will be consistent with other controls. So we don't need to open up AutomationID right now.

src/System.Windows.Forms/src/PublicAPI.Unshipped.txt
```diff
@@ -4,4 +4,4 @@ System.Windows.Forms.PrintPreviewControl.TabStopChanged -> System.EventHandler?
 *REMOVED*override System.Windows.Forms.ToolStripButton.CanSelect.get -> bool
 override System.Windows.Forms.MaskedTextBox.CreateAccessibilityInstance() -> System.Windows.Forms.AccessibleObject!
 override System.Windows.Forms.MaskedTextBox.OnGotFocus(System.EventArgs! e) -> void
-override System.Windows.Forms.MaskedTextBox.OnMouseDown(System.Windows.Forms.MouseEventArgs! e) -> void
+override System.Windows.Forms.MaskedTextBox.OnMouseDown(System.Windows.Forms.MouseEventArgs! e) -> void
```
Tanya-Solyanik, line 7: Please revert this change

dmitrii-drobotov (approved):

Tanya-Solyanik (approved):

PR #7892, opened 2022-10-05
Clean up PInvoke Definitions

lonitra:
- Removed `in` from PInvoke method signatures with generics. Using sharplab.io, found that `in` is causing some extra work to be done to make copies of the parameter
- Some other minor code clean up

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7892)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.BitBlt.cs
+++ b/src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.BitBlt.cs
@@ -7,7 +7,7 @@ namespace Windows.Win32
     internal static partial class PInvoke
     {
         public static BOOL BitBlt<T>(
-            in T hdc,
+            T hdc,
             int x,
             int y,
             int cx,
@@ -27,7 +27,7 @@ namespace Windows.Win32
                 x1,
                 y1,
                 rop);
-            GC.KeepAlive(hdc.Handle);
+            GC.KeepAlive(hdc.Wrapper);
             return result;
         }
 
@@ -37,7 +37,7 @@ namespace Windows.Win32
             int y,
             int cx,
             int cy,
-            in T hdcSrc,
+            T hdcSrc,
             int x1,
             int y1,
             ROP_CODE rop) where T : IHandle<HDC>
@@ -52,7 +52,7 @@ namespace Windows.Win32
                 x1,
                 y1,
                 rop);
-            GC.KeepAlive(hdcSrc.Handle);
+            GC.KeepAlive(hdcSrc.Wrapper);
             return result;
         }
     }
--- a/src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.CloseHandle.cs
+++ b/src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.CloseHandle.cs
@@ -6,10 +6,10 @@ namespace Windows.Win32
 {
     internal static partial class PInvoke
     {
-        public static BOOL CloseHandle(IHandle handle)
+        public static BOOL CloseHandle<T>(T handle) where T : IHandle<HANDLE>
         {
-            BOOL result = CloseHandle((HANDLE)handle.Handle);
-            GC.KeepAlive(handle);
+            BOOL result = CloseHandle(handle.Handle);
+            GC.KeepAlive(handle.Wrapper);
             return result;
         }
     }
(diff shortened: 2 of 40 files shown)


src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.BitBlt.cs
```diff
@@ -27,7 +27,7 @@ internal static partial class PInvoke
                 x1,
                 y1,
                 rop);
-            GC.KeepAlive(hdc.Handle);
+            GC.KeepAlive(hdc.Wrapper);
```
JeremyKuhne, line 30: Nice catch!

JeremyKuhne (approved):
Looks good!

Tanya-Solyanik:
Does 'out' cause any problems?

Tanya-Solyanik (approved):

JeremyKuhne:
> Does 'out' cause any problems?

Made it a recursive call.

JeremyKuhne (approved):

dreddy-work:
> Made it a recursive call.

Apart from that, are not these method signatures generated from the cswin32 tool?

elachlan:
> > Made it a recursive call.
> 
> Apart from that, are not these method signatures generated from the cswin32 tool?

No, most of them are written as wrappers for the CsWin32 PInvoke calls to avoid GC.

If there are similar issues in CsWin32, it would be fantastic to raise issues there so we can fix them as well so everyone gets the performance improvements.

RussKie:
> * Using sharplab.io, found that `in` is causing some extra work to be done to make copies of the parameter

Could you share some examples please?

lonitra:
> Could you share some examples please?

[This ](https://sharplab.io/#v2:C4LghgzgtgNAJiA1AHwAICYCMBYAUKgZgAIMiBhIgbzyNqJrsJMwDYiAjAe04BsiBJCAHUAlgDs4nAO4BRAB4AKABJCAcgBEiACykSAlFQZ1GAdm264AOkGqArjx4K9AbiNEAvm7cBtAMpawACcABwAZMHZLACVbMWARKABTSwApEWAAcUSxRMCRAGMFYABPYMTOADNlNXU9PQBdNyZUVg5uPkFRCWlMAB4AFQA+BX7tIX0iKS1cxKJRkAElMAkeRN6VDUG3alxjYy5eIkDEiHtgIgBeAWFxSVlFLXGrJZXElzdjDLJLAGlExOCAEEeCIAG6JBSPCSWISBMDBMqBd67Pa0VBmY6nHjAVwo2iePH0Ql+AIhcKRGJxBLJNKZbK5ApFUrlKobWoNJrEFpsA4dG7dKToAbDcRzMYTKYzMULfgvOCrdY1LaEnao2i8o4nM6Xa5dO7ySFPSxy1bItVEL6/f5AkHgw3Q2HwxFmtXozVYnFuAnGTlEcTAXIVMD5Way5bytb9E2JQaTabHObRogLWJQZZgADmiTg2w+dCj4dWRCTlCIWZxHjztE47AAVol8sAAPxER0I3KXWPALQiCC44zeui+45gSRiHjFIjBILxMB8CDAQK2RvFmrJxaFtZs5XGVVq4J5UFgAOa0ecceTsT+ogAfSPPFsiX7ruIGps9kcBgusbvc8flyuAAGZ9USYNli03TsiG7XsQL2Jga3rRsWzbREoLED84IJPB3CAA=) is the experiment with `in T` vs. `T`

RussKie:
`in` helps with readonly structs, otherwise it causes additional copies to be made, https://devblogs.microsoft.com/premier-developer/the-in-modifier-and-the-readonly-structs-in-c/#the-in-modifier.

lonitra:
> Does `out T` cause extra IL lines too?

I don't see it causing any significant extra IL lines -- only lines for initializing the out param

lonitra:
> `in` helps with readonly structs, otherwise it causes additional copies to be made, https://devblogs.microsoft.com/premier-developer/the-in-modifier-and-the-readonly-structs-in-c/#the-in-modifier.

We are passing readonly structs with the `in T` param, but seems like it cannot figure this out with generics and ends up making additional copies unfortunately

RussKie:
> We are passing readonly structs with the `in T` param, but seems like it cannot figure this out with generics and ends up making additional copies unfortunately

I may be worth while chatting with the Roslyn team to find our whether it's a bug or a feature :)

PR #7894, opened 2022-10-06
Convert Cursor PInvokes to CsWin32

JeremyKuhne:
Converts a number of PInvokes, mostly with the Cursor class.

Reduce some nesting in affected areas.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7894)


src/System.Windows.Forms/src/System/Windows/Forms/Cursor.cs
```diff
@@ -246,120 +232,111 @@ private void DrawImageCore(Graphics graphics, Rectangle imageRect, Rectangle tar
...
+
+                drawWidth = cursorSize.Width;
+                drawHeight = cursorSize.Height;
+                clipWidth = targetWidth < imageWidth ? targetWidth : imageWidth;
+                clipHeight = targetHeight < imageHeight ? targetHeight : imageHeight;
             }
+
+            // The ROP is SRCCOPY, so we can be simple here and take advantage of clipping regions.
+            // Drawing the cursor is merely a matter of offsetting and clipping.
+            PInvoke.IntersectClipRect(dc, targetX, targetY, targetX + clipWidth, targetY + clipHeight);
```
lonitra, line 313: We have a couple of spots in codebase where we would pass `new HandleRef<HDC>(graphics, dc)`. Would we rather want to pass a `DeviceContextHdcScope` instead of the handleref?

lonitra (commented):
LGTM, just a couple questions

src/System.Windows.Forms.Primitives/src/Windows/Win32/Graphics/Gdi/HBITMAP.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Windows.Win32.Graphics.Gdi
+{
+    internal readonly partial struct HBITMAP : IDisposable
```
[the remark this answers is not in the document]
JeremyKuhne: In the case I added this as we're just getting back an HBITMAP. I think we'll probably iterate a bit on patterns here...

src/System.Windows.Forms/src/System/Windows/Forms/Cursor.cs
JeremyKuhne: Probably, would need to look at each case to be sure.

lonitra (approved):

PR #7911, opened 2022-10-10
Fix bug in TabPageCollection.RemoveAt (#7837)

### LINKED ISSUE #7837: TabControl.TabPageCollection.RemoveAt(int) removes wrong item!
### .NET version

All .NET versions to .NET 6

No

Same bug in all previous .NET versions & .NET Framework

### Issue description

This RemoveAt method in System.Windows.Forms.TabControl.TabPageCollection has a horrible bug that can cause it to remove the wrong page from the TabControl.

There's only one line of code in the method, so the bug is obvious:

`owner.Controls.RemoveAt(index);`

You can see that it's applying the index to the **Controls** collection rather than the **TabPages** collection. These two collections are not the same; they can sometimes get out of sync when the SelectedTab changes. You used to be able to work around the bug by setting the [AllowUpdateChildControlIndexForTabControls application switch](https://learn.microsoft.com/en-us/dotnet/core/compatibility/winforms#allowupdatechildcontrolindexfortabcontrols-compatibility-switch-not-supported) to false, but this switch is no longer supported.

You can fix the bug by changing that line of code to:

`owner.Controls.Remove(this[index]);`

### Steps to reproduce

I don't think this is needed - the bug is obvious from looking at the one line of code.

albahari:
Fixes #7837

## Proposed changes

- Fix logic in TabPageCollection.RemoveAt to avoid removing incorrect TabPage when TabControl.Controls collection is out of sync with TabControls._tabPages. This can happen when Control.WmWindowPosChanged calls _parent.UpdateChildControlIndex, which calls Controls.SetChildIndex.

## Customer Impact

- None

## Regression? 

- No

## Risk

- Low

## Test methodology 

- Adhoc testing. There's just one of code, and the new implementation calls only public methods.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7911)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
@@ -11787,9 +11787,10 @@ namespace System.Windows.Forms
         ///  reflect its index.
         /// </summary>
         private void UpdateChildControlIndex(Control control)
-        {            
-            if (// Don't reorder the child control array for tab controls. Implemented as a special case
-                // in order to keep the method private.
+        {
+            // Don't reorder the child control array for tab controls. Implemented as a special case
+            // in order to keep the method private.
+            if (
                 this is TabControl ||
                 // Also short-circuit when the Control class is instantiated directly. This is to provide
                 // consistency with the behavior prior to bug fix https://github.com/dotnet/winforms/issues/7837
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
@@ -11790,7 +11790,13 @@ namespace System.Windows.Forms
         {
             // Don't reorder the child control array for tab controls. Implemented as a special case
             // in order to keep the method private.
-            if (GetType().IsAssignableFrom(typeof(TabControl)))
+            if (typeof(TabControl).IsAssignableFrom(GetType()))
+            {
+                return;
+            }
+
+            // The following check is to mitigate a potentially disruptive effect of bug fix #7837 / PR 7911
+            if (GetType() == typeof(Control))
             {
                 return;
             }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
@@ -11787,16 +11787,13 @@ namespace System.Windows.Forms
         ///  reflect its index.
         /// </summary>
         private void UpdateChildControlIndex(Control control)
-        {
-            // Don't reorder the child control array for tab controls. Implemented as a special case
-            // in order to keep the method private.
-            if (typeof(TabControl).IsAssignableFrom(GetType()))
-            {
-                return;
-            }
-
-            // The following check is to mitigate a potentially disruptive effect of bug fix #7837 / PR 7911
-            if (GetType() == typeof(Control))
+        {            
+            if (// Don't reorder the child control array for tab controls. Implemented as a special case
+                // in order to keep the method private.
+                this is TabControl ||
+                // Also short-circuit when the Control class is instantiated directly. This is to provide
+                // consistency with the behavior prior to bug fix https://github.com/dotnet/winforms/issues/7837
+                GetType() == typeof(Control))
             {
                 return;
             }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
@@ -11790,11 +11790,9 @@ namespace System.Windows.Forms
         {
             // Don't reorder the child control array for tab controls. Implemented as a special case
             // in order to keep the method private.
-            if (
-                this is TabControl ||
-                // Also short-circuit when the Control class is instantiated directly. This is to provide
-                // consistency with the behavior prior to bug fix https://github.com/dotnet/winforms/issues/7837
-                GetType() == typeof(Control))
+            // Also short-circuit when the Control class is instantiated directly. This is to provide
+            // consistency with the behavior prior to bug fix https://github.com/dotnet/winforms/issues/7837
+            if (this is TabControl || GetType() == typeof(Control))
             {
                 return;
             }
--- a/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/TabControlTests.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/TabControlTests.cs
@@ -96,5 +96,63 @@ namespace System.Windows.Forms.UITests
                     return tabControl;
                 });
         }
+
+        [WinFormsFact]
+        public async Task TabControl_ControlsDoNotReorderWhenSelectedIndexChanges()
+        {
+            await RunSingleControlTestAsync(
+                testDriverAsync: async (form, tabControl) =>
+                {
+                    var originalPages = tabControl.TabPages.Cast<TabPage>().ToArray();
+
+                    tabControl.SelectedIndex--;
+                    await Task.Yield();
+
+                    // The changes we made to SelectedIndex will result in a WM_WINDOWPOSCHANGED message which will 
+                    // fire TabPage.WmWindowPosChanged, which will call TabControl.UpdateChildControlIndex.
+
+                    // This test ensures that the latter method short-circuits and does not call Controls.SetChildIndex
+                    // which would inappropriately reorder the Controls collection, resulting in TabPages and Controls
+                    // becoming out of sync.
+
+                    for (int i = 0; i < originalPages.Length; i++)
+                    {
+                        Assert.Equal(tabControl.Controls[i], tabControl.TabPages[i]);
+                    }
+
+                    // The following test proves that it *matters* that the Controls and TabPages collection remain in sync.
+
+                    // Remove the first page by position:
+                    tabControl.TabPages.RemoveAt(0);
+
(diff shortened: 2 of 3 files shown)


RussKie (commented):
Looks sensible.

Do you think it's possible to come up with a regression test that would simulate the original issue, so that we can verify the fix?

albahari:
> * Do you think it's possible to come up with a regression test that would simulate the original issue, so that we can verify the fix?

It's tricky - the sequence of events to get the collections out of sync is complicated.

RussKie:
How did you initially observe the issue? How can we be sure we've fixed it?

dreddy-work:
@albahari ,  Can you respond to the questions above?  PRs with inactivity will be closed automatically after 14days.

albahari:
Sorry for the delay - it took a while to get to a minimum repro. It's not until you subclass TabControl that the condition becomes apparent (or at least easy to reproduce). This discovery in itself has lead to finding a second bug. This second bug may in fact be a root cause (rendering the first bug fix redundant). 

Control.UpdateChildControlIndex starts with the following code:
 
```
if (GetType().IsAssignableFrom(typeof(TabControl)))
{
   return;
}
```

This check is the wrong way around. It should be:

```
if (typeof(TabControl).IsAssignableFrom (GetType()))
{
   return;
}
```

Fixing this bug means that the tabControl.Controls collection (appears to) stay in sync with TabControls._tabPages. Assuming that the two collections are *intended* to always stay in sync (and there are no other conditions that could result in these collections being ordered differently), the first bug fix becomes redundant.

Here is a minimum repro. Note that it requires a running message loop, so TabPage can receive a WM_WINDOWPOSCHANGED message:

```
// Instantiate a subclassed TabControl with 5 pages:
var tc = new MyTabControl ();
var pages = Enumerable.Range (0, 5).Select (i => new TabPage ($"Page {i}")).ToArray();
tc.TabPages.AddRange (pages);	

// Select the last page:
tc.SelectedIndex = tc.TabPages.Count - 1;

// Show the TabControl on a form:
var form = new Form();
form.Controls.Add (tc);
form.Show();

// The change we made to SelectedIndex will result in a WM_WINDOWPOSCHANGED message which will 
// fire TabPage.WmWindowPosChanged, which will call TabControl.UpdateChildControlIndex, which will
// call Controls.SetChildIndex. This will reorder Controls without reordering TabPages.

// Notice that the TabPages and Controls collections are out of sync:
for (int i = 0; i < tc.TabCount; i++)
    Console.WriteLine ($"TabPage {tc.TabPages [i].Text} => {tc.Controls [i].Text}");
    
// I don't know whether having these collections out of sync in itself violates an invariant,
// or whether these collections are allowed to be differently-ordered. In any case, calling
// RemoveAt(int index) fails when the collections are not in sync:

// Remove the first page by position:
tc.TabPages.RemoveAt (0);

// We should be left with pages[1] and pages[2]
Debug.Assert (tc.TabPages[0] == pages [1]);      // Fails
Debug.Assert (tc.TabPages[1] == pages [2]);      // Fails

class MyTabControl : TabControl
{	
}
```
How should I proceed with this? Create a new PR for the second bug fix?

Note that while the original bug fix has virtually zero risk, the second bug fix could conceivably result in changed behavior in an (arguably obscure) scenario. Consider that someone instantiates the Control class directly: right now, the bug inappropriately short-circuits UpdateChildControlIndex when WM_WINDOWPOSCHANGED events are received.

dreddy-work:
Thank you @albahari for making time and sharing the findings here. We will review this and update you on next steps.

dreddy-work:
@albahari , lets go ahead and update this PR with the new fix given current one is redundant after that. Also, make sure you add some unit tests / integration tests to cover the code path.

albahari:
I've updated the PR and added an integration test that fails without the bugfix.

It doesn't make sense that the CI test would fail for X86 release on ActiveXFontMarshaler_RoundTripManagedFont() - perhaps this test needs to be repeated?

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -11790,7 +11790,13 @@ private void UpdateChildControlIndex(Control control)
...
             // Don't reorder the child control array for tab controls. Implemented as a special case
             // in order to keep the method private.
-            if (GetType().IsAssignableFrom(typeof(TabControl)))
+            if (typeof(TabControl).IsAssignableFrom(GetType()))
+            {
+                return;
+            }
+
+            // The following check is to mitigate a potentially disruptive effect of bug fix #7837 / PR 7911
+            if (GetType() == typeof(Control))
```
lonitra, line 11799: Should this be `GetType() != typeof(Control)` if we do not want it to short circuit in this scenario? Please also combine this with the if statement above.

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -11790,7 +11790,13 @@ private void UpdateChildControlIndex(Control control)
         {
             // Don't reorder the child control array for tab controls. Implemented as a special case
             // in order to keep the method private.
-            if (GetType().IsAssignableFrom(typeof(TabControl)))
+            if (typeof(TabControl).IsAssignableFrom(GetType()))
+            {
+                return;
+            }
+
+            // The following check is to mitigate a potentially disruptive effect of bug fix #7837 / PR 7911
```
lonitra, line 11798: Let's directly state the scenario we are trying to avoid with this if statement. Something like "Ensure we do not short-circuit if we have a Control class directly"

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/TabControlTests.cs
```diff
@@ -96,5 +96,63 @@ private async Task RunTestAsync(Func<Form, TabControl, Task> runTest)
...
+                    tabControl.TabPages.Add(tabPage3);
+
+                    // Start with the Selected Page at the end:
+                    tabControl.SelectedIndex = tabControl.TabPages.Count - 1;
+
+                    return tabControl;
+                });
+        }
+
+        // Bug #7837 occured only when TabControl was subclassed.
```
lonitra, line 155: Let's link this bug.

lonitra (commented):
Test looks great! Just a few comments.

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
albahari: No. The formerly incorrect predicate `GetType().IsAssignableFrom(typeof(TabControl))` could fail in two ways:

1. A false negative, when GetType() returns a subclassed TabControl
2. A false positive, when GetType() == typeof(Control)

We want to fix the first, because it's causing the bug demonstrated by the newly added integration test. However, we want to avoid fixing the second, because this is more of a quirk than a bug (and a quirk that consumers could possibly have become dependent on).

Note that I've simplified the predicate to "this is TabControl" - we don't need reflection now that we've got it the right way round.

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
albahari: Done

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/TabControlTests.cs
albahari: Done

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -11787,10 +11787,13 @@ private void UpdateBindings()
...
-            // Don't reorder the child control array for tab controls. Implemented as a special case
-            // in order to keep the method private.
-            if (GetType().IsAssignableFrom(typeof(TabControl)))
+        {            
+            if (// Don't reorder the child control array for tab controls. Implemented as a special case
+                // in order to keep the method private.
+                this is TabControl ||
+                // Also short-circuit when the Control class is instantiated directly. This is to provide
+                // consistency with the behavior prior to bug fix https://github.com/dotnet/winforms/issues/7837
+                GetType() == typeof(Control))
```
lonitra, line 11796: Please move comments above the if
```suggestion
            // Don't reorder the child control array for tab controls. Implemented as a special case
            // in order to keep the method private.
            // Also short-circuit when the Control class is instantiated directly. This is to provide
            // consistency with the behavior prior to bug fix https://github.com/dotnet/winforms/issues/7837
            if (this is TabControl || GetType() == typeof(Control))
```

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
lonitra: I see, thank you for the explanation here!

lonitra:
Looks like there is some trailing whitespace that is causing build to complain. It is difficult for me to see where on web, but if you build on command line (cd to winforms repo and type 'build'), the same errors should show and help point you to where the trailing whitespace is.

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
albahari: Done

lonitra: This is still outstanding. Let's move both comments above the if and format if statement like suggestion. Everything else looks good so once that is fixed we can get this merged :)

lonitra (approved):
🚀

PR #7918, opened 2022-10-11
Enable IDE0110 (Remove unnecessary discard)

elachlan:
Enable IDE0110 Remove unnecessary discard

Related: #7887

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7918)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.UpDownEdit.UpDownEditAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.UpDownEdit.UpDownEditAccessibleObject.cs
@@ -22,8 +22,7 @@ namespace System.Windows.Forms
                     get => _parent.AccessibilityObject.Name
                         ?? _parent switch
                         {
-                            NumericUpDown _ => SR.EditDefaultAccessibleName,
-                            DomainUpDown _ => SR.EditDefaultAccessibleName,
+                            NumericUpDown or DomainUpDown => SR.EditDefaultAccessibleName,
                             _ => null
                         };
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.UpDownEdit.UpDownEditAccessibleObject.cs
```diff
@@ -22,8 +22,8 @@ public UpDownEditAccessibleObject(UpDownEdit owner, UpDownBase parent) : base(ow
                     get => _parent.AccessibilityObject.Name
                         ?? _parent switch
                         {
-                            NumericUpDown _ => SR.EditDefaultAccessibleName,
-                            DomainUpDown _ => SR.EditDefaultAccessibleName,
+                            NumericUpDown => SR.EditDefaultAccessibleName,
```
dreddy-work, line 25: what exactly those discards were doing early? 

nit: we can combine these two cases.

elachlan: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0110

> This rule flags unnecessary [discard patterns](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/patterns#discard-pattern). A discard pattern is unnecessary when used in a context where it has no effect.

RussKie (approved):

PR #7928, opened 2022-10-11
Update TFM for build targets

dreddy-work:
Updating build target for Winforms binaries to .NET 8.0.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7928)

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -7,7 +7,9 @@
 
   <PropertyGroup>
     <TargetFrameworkName>net</TargetFrameworkName>
-    <TargetFrameworkVersion>8.0</TargetFrameworkVersion>
+	<TargetFrameworkMajorVersion>8</TargetFrameworkMajorVersion>
+	<TargetFrameworkMinorVersion>0</TargetFrameworkMinorVersion>
+    <TargetFrameworkVersion>$(TargetFrameworkMajorVersion)$(TargetFrameworkMinorVersion)</TargetFrameworkVersion>
     <TargetFramework>$(TargetFrameworkName)$(TargetFrameworkVersion)</TargetFramework>
     <Product>Microsoft&#xAE; .NET</Product>
     <Copyright>$(CopyrightNetFoundation)</Copyright>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -7,7 +7,7 @@
 
   <PropertyGroup>
     <TargetFrameworkName>net</TargetFrameworkName>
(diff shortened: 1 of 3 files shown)


Directory.Build.props
```diff
@@ -7,7 +7,7 @@
 
   <PropertyGroup>
     <TargetFrameworkName>net</TargetFrameworkName>
-    <TargetFrameworkVersion>7.0</TargetFrameworkVersion>
+    <TargetFrameworkVersion>8.0</TargetFrameworkVersion>
     <TargetFramework>$(TargetFrameworkName)$(TargetFrameworkVersion)</TargetFramework>
```
RussKie, line 11: We can probably introduce a property to contain `net8.`, and reuse the variable in the targets file, so that we don't have to update multiple files in the future.

dreddy-work:
@marcpopMSFT / @mmitche, we need to get this resolved as soon as we can. Until then, any testing we are doing against main branch, or their insertions are not valid. Both WinForms and WPF are assuming the work needed to target .NET8.0 TFM is done in those repos but please advise if anything else needs to be done. We cannot create or target applications to .NET 8.0. 

@singhashish-wpf  as FYI.

Directory.Build.props
dreddy-work: make sense. Will push new commit.

dreddy-work:
@singhashish-wpf , you may need similar change in WPF.

Directory.Build.targets
```diff
@@ -34,6 +34,7 @@
     <_ProjectDirectoryRelativePath>$([MSBuild]::MakeRelative($(RepoRoot), $(MSBuildProjectDirectory)))</_ProjectDirectoryRelativePath>
     <_IsUnderTestsFolder>$(_ProjectDirectoryRelativePath.Contains("\tests\"))</_IsUnderTestsFolder>
     <IsShipping Condition="$(_IsUnderTestsFolder)">false</IsShipping>
+	<RedistTargetFrameworkName>$(TargetFrameworkName)$(TargetFrameworkMajorVersion).</RedistTargetFrameworkName>
```
RussKie, line 37: Any reason not to put it in the props file? This value isn't supposed to be changing during the build.

RussKie (approved):

Directory.Build.targets
dreddy-work: No specific. I noticed other properties related are in this file and decided to keep it here.

PR #7933, opened 2022-10-12
Convert file dialog COM usage to CsWin32

JeremyKuhne:
Directly uses COM pointers when accessing the native dialog.

This also rewrites the CCW for the dialog options to use the CsWin32 generated interface and moves the statics into the CsWin32 COM struct (which is partial).

Adds better helpers to initialize vtables and find the CCW for a given object.

Removes our DECIMAL and BSTR definitions.

Adds an enum helper for setting/clearing flags based on a boolean value.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7933)


elachlan:
CC: @kant2002

JeremyKuhne:
FYI: @AArnott, @AaronRobinsonMSFT, @jkoritzinsky, @jacdavis, @elinor-fung, @vitek-karas, @LakshanF, @tannergooding

We're getting enough in place from CsWin32 to start basing our ComWrappers on it. Putting our statics in the partial structs generated by CsWin32, `IFileDialogEvents` is the example here. The implementation is pretty simple- hopefully we can move more to the code generators. We'll continue to convert things to build confidence in our patterns.

src/System.Windows.Forms.Primitives/src/Windows/Win32/UI/Shell/IFileDialogEvents.cs
```diff
@@ -0,0 +1,53 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using ComWrappers = Interop.WinFormsComWrappers;
+
+namespace Windows.Win32.UI.Shell;
+
+internal unsafe partial struct IFileDialogEvents : INativeGuid, IPopulateVTable<IFileDialogEvents.Vtbl>
```
JeremyKuhne, line 11: This is our ComWrappers "CCW" implementation. Pretty boilerplate stuff now.

src/System.Windows.Forms.Primitives/src/Windows/Win32/System/Com/CLSID.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Windows.Win32.System.Com
+{
+    internal static class CLSID
```
elachlan, line 7: These are in `ShlGuid.h` Win32Metadata outs them with CLSID_* constants. Does CsWin32 not output them?

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.IDataObjectVtbl.cs
```diff
@@ -51,23 +54,32 @@ private static HRESULT GetData(IntPtr thisPtr, FORMATETC* format, STGMEDIUM_Raw*
                 }
             }
 
-            [UnmanagedCallersOnly]
-            private static unsafe HRESULT GetDataHere(IntPtr thisPtr, FORMATETC* format, STGMEDIUM_Raw* pMedium)
+            [UnmanagedCallersOnly(CallConvs = new[] { typeof(CallConvStdcall) })]
+            private static unsafe HRESULT GetDataHere(IDataObject* @this, FORMATETC* format, STGMEDIUM* pMedium)
             {
-                var instance = ComInterfaceDispatch.GetInstance<IDataObject>((ComInterfaceDispatch*)thisPtr);
+                if (pMedium is null)
```
AaronRobinsonMSFT, line 60: Do we need to check for `format` here too? It is dereferenced below.

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.IDataObjectVtbl.cs
```diff
@@ -77,34 +89,41 @@ private static unsafe HRESULT GetDataHere(IntPtr thisPtr, FORMATETC* format, STG
...
             }
 
-            [UnmanagedCallersOnly]
-            private static unsafe HRESULT QueryGetData(IntPtr thisPtr, FORMATETC* format)
+            [UnmanagedCallersOnly(CallConvs = new[] { typeof(CallConvStdcall) })]
+            private static unsafe HRESULT QueryGetData(IDataObject* @this, FORMATETC* format)
             {
-                var instance = ComInterfaceDispatch.GetInstance<IDataObject>((ComInterfaceDispatch*)thisPtr);
-                return (HRESULT)instance.QueryGetData(ref *format);
+                var instance = ComInterfaceDispatch.GetInstance<ComTypes.IDataObject>((ComInterfaceDispatch*)@this);
```
AaronRobinsonMSFT, line 95: Check `format` is `null`?

AaronRobinsonMSFT: I'll stop commenting.

src/System.Windows.Forms/src/System/Windows/Forms/ComponentModel/COM2Interop/COM2TypeInfoProcessor.cs
```diff
@@ -528,14 +530,14 @@ private unsafe static PropInfo ProcessDataCore(ITypeInfo typeInfo, IDictionary p
                 pi.DispId = dispid;
                 pi.Attributes.Add(new DispIdAttribute((int)pi.DispId));
             }
-
-            string helpString = helpStringBstr.String.ToString();
-            if (!string.IsNullOrEmpty(helpString))
+
```
lonitra, line 533: extra spacing

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.IEnumFORMATETCVtbl.cs
```diff
@@ -93,8 +94,16 @@ private static HRESULT Clone(Com.IEnumFORMATETC* @this, Com.IEnumFORMATETC** ppe
                 {
                     IEnumFORMATETC instance = ComInterfaceDispatch.GetInstance<IEnumFORMATETC>((ComInterfaceDispatch*)@this);
                     instance.Clone(out var cloned);
```
AaronRobinsonMSFT, line 96: Does this return an `HRESULT` or is this using the built-in exception throwing?

src/System.Windows.Forms.Primitives/src/Windows/Win32/System/Com/CLSID.cs
JeremyKuhne: Not sure. I'll follow up with my next change.

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.cs
```diff
@@ -256,38 +236,62 @@ protected override void ReleaseObjects(IEnumerable objects)
...
+        ///   Handles exceptions and converts to <see cref="HRESULT"/>.
+        ///  </para>
+        /// </remarks>
+        internal static HRESULT UnwrapAndInvoke<TThis, TInterface>(TThis* @this, Func<TInterface, HRESULT> func)
+            where TThis : unmanaged
+            where TInterface : class
+        {
+            try
+            {
+                TInterface? @object = ComInterfaceDispatch.GetInstance<TInterface>((ComInterfaceDispatch*)@this);
```
AaronRobinsonMSFT, line 288: This will never return `null`.

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.IEnumFORMATETCVtbl.cs
```diff
@@ -93,8 +94,16 @@ private static HRESULT Clone(Com.IEnumFORMATETC* @this, Com.IEnumFORMATETC** ppe
...
-                    *ppenum = (Com.IEnumFORMATETC*)WinFormsComWrappers.Instance.GetComPointer(cloned, IID.IEnumFORMATETC);
-                    return HRESULT.S_OK;
+                    *ppenum = null;
```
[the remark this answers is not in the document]
JeremyKuhne: We can add that when we convert this one over.

src/System.Windows.Forms/src/System/Windows/Forms/FolderBrowserDialog.cs
```diff
@@ -253,37 +254,26 @@ public override void Reset()
         /// <returns>
         ///  <see langword="true" /> if the folder browser dialog was successfully run; otherwise, <see langword="false" />.
         /// </returns>
-        protected override bool RunDialog(IntPtr hWndOwner)
-        {
+        protected override bool RunDialog(IntPtr hWndOwner) =>
```
elachlan, line 257: class is sealed and function is protected. Should we use `HWND` here instead of `IntPtr`?

src/System.Windows.Forms/src/System/Windows/Forms/ComponentModel/COM2Interop/COM2TypeInfoProcessor.cs
JeremyKuhne: I'll grab that in the next PR so you can build on this one.

lonitra (approved):
🥳

src/System.Windows.Forms/src/System/Windows/Forms/FolderBrowserDialog.cs
JeremyKuhne: `CommonDialog` isn't sealed so we can't change this.

JeremyKuhne:
Thanks to @lonitra for sitting through a good chunk of this with me.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetOpenFileName.cs
```diff
@@ -0,0 +1,19 @@
...
+
+using System.Runtime.InteropServices;
+using Windows.Win32.UI.Controls.Dialogs;
+
+namespace Windows.Win32
+{
+    // https://github.com/microsoft/win32metadata/issues/1300
+    internal static partial class PInvoke
+    {
+        [DllImport("COMDLG32.dll", EntryPoint = "GetOpenFileNameW", ExactSpelling = true, PreserveSig = false)]
```
AaronRobinsonMSFT, line 13: I think the more common pattern would be naming the enclosing class `COMDLG32` and then specifying the file name using `nameof(COMDLG32)`. I'd also consider using `LibraryImport` for these definitions.

JeremyKuhne:
If there are any other comments please leave them. I'll be keeping an eye on this. Getting it in as we have further changes backed behind it. :)

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetOpenFileName.cs
```diff
@@ -0,0 +1,19 @@
...
+        public static unsafe extern BOOL GetOpenFileName([In][Out] OPENFILENAME* param0);
```
AaronRobinsonMSFT, line 14: The `[In][Out]` have no functional value here. If they are being used to express indent, that seems fine, but I would avoid them and rely on a comment instead.

src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.cs
```diff
@@ -408,52 +386,48 @@ public bool ValidateNames
         /// <summary>
         ///  Processes the CDN_FILEOK notification.
         /// </summary>
-        private bool DoFileOk(IntPtr lpOFN)
+        private unsafe bool DoFileOk(OPENFILENAME* lpOFN)
```
jacdavis, line 389: >OPENFILENAME* [](http://example.com/codeflow?start=37&length=13)

Does it matter that this is no longer verifiably safe? Was this already compiling with AllowUnsafeBlocks?

src/System.Windows.Forms.Primitives/src/Windows/Win32/UI/Shell/IFileDialogEvents.cs
```diff
@@ -0,0 +1,53 @@
...
+        vtable->OnFolderChange_6 = &OnFolderChange;
+        vtable->OnSelectionChange_7 = &OnSelectionChange;
+        vtable->OnShareViolation_8 = &OnShareViolation;
+        vtable->OnTypeChange_9 = &OnTypeChange;
+        vtable->OnOverwrite_10 = &OnOverwrite;
+    }
+
+    [UnmanagedCallersOnly(CallConvs = new[] { typeof(CallConvStdcall) })]
+    public static HRESULT OnFileOk(IFileDialogEvents* @this, IFileDialog* pfd)
+        => ComWrappers.UnwrapAndInvoke<IFileDialogEvents, Interface>(@this, o => o.OnFileOk(pfd));
```
AaronRobinsonMSFT, line 28: I'm not sure this is a good idea. A delegate allocation is occurring for each invocation. Based on the pointer usage and all the other unsafe C# going on here, this is out of place.

src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.cs
jacdavis: Nevermind. I see plenty of other unsafe contexts in this file.

src/System.Windows.Forms/src/System/Windows/Forms/OpenFileDialog.cs
```diff
@@ -136,84 +133,55 @@ private protected override bool RunFileDialog(NativeMethods.OPENFILENAME_I ofn)
...
+            {
+                return Array.Empty<string>();
             }
-            else
+
+            try
             {
-                openDialog.GetResult(out IShellItem? item);
-                if (item is null)
+                items->GetCount(out uint count);
```
AaronRobinsonMSFT, line 154: Does this throw or do we need to check the `HRESULT`?

src/System.Windows.Forms/src/System/Windows/Forms/FolderBrowserDialog.cs
```diff
@@ -312,46 +302,48 @@ private bool TryRunDialogVista(IntPtr owner, out bool returnValue)
...
                 else
                 {
-                    var customize = (WinFormsComWrappers.FileOpenDialogWrapper)dialog;
-                    customize.AddText(0, _descriptionText);
+                    IFileDialogCustomize* customize;
+                    Guid iid = IID.IFileDialogCustomize;
+                    if (dialog->QueryInterface(&iid, (void**)&customize) == HRESULT.S_OK)
+                    {
+                        customize->AddText(0, _descriptionText).ThrowOnFailure();
+                        customize->Release();
```
jacdavis, line 332: >customize->Release(); [](http://example.com/codeflow?start=24&length=21)

The release call should probably be within a finally block.

src/System.Windows.Forms/src/System/Windows/Forms/FolderBrowserDialog.cs
```diff
@@ -312,46 +302,48 @@ private bool TryRunDialogVista(IntPtr owner, out bool returnValue)
...
-                    dialog.SetDefaultFolder(initialDirectory);
-                    dialog.SetFolder(initialDirectory);
-                }
-                catch (FileNotFoundException)
+                IShellItem* initialDirectory = PInvoke.SHCreateShellItem(_initialDirectory);
+                if (initialDirectory is not null)
                 {
+                    dialog->SetDefaultFolder(initialDirectory).ThrowOnFailure();
+                    dialog->SetFolder(initialDirectory).ThrowOnFailure();
+                    initialDirectory->Release();
```
jacdavis, line 346: > initialDirectory->Release(); [](http://example.com/codeflow?start=19&length=29)

Same here. This will leak if SetDefaultFolder or SetFolder fail.

src/System.Windows.Forms/src/System/Windows/Forms/FolderBrowserDialog.cs
```diff
@@ -388,13 +380,16 @@ private void SetOption(int option, bool value)
...
+            IShellItem* item;
+            dialog->GetResult(&item);
             if (item is not null)
             {
-                HRESULT hr = item.GetDisplayName(SIGDN.SIGDN_FILESYSPATH, out _selectedPath!);
-                hr.ThrowOnFailure();
+                item->GetDisplayName(SIGDN.SIGDN_FILESYSPATH, out PWSTR ppszName).ThrowOnFailure();
+
+                _selectedPath = new(ppszName);
+                Marshal.FreeCoTaskMem((nint)(void*)ppszName);
```
jacdavis, line 392: >FreeCoTaskMem [](http://example.com/codeflow?start=24&length=13)

This one is more of a nit as the code is currently written, but the GetDisplayName out param will leak if this function exits early. The only way that is likely to happen as written is if the new allocation allocation fails. But, if someone adds any more code in the future, this may become more likely. Putting the Marshal.FreeCoTaskMem in a finally block will resolve this.

src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.cs
```diff
@@ -52,11 +54,7 @@ internal FileDialog()
         [SRCategory(nameof(SR.CatBehavior))]
         [DefaultValue(true)]
         [SRDescription(nameof(SR.FDaddExtensionDescr))]
-        public bool AddExtension
-        {
-            get => GetOption(AddExtensionOption);
-            set => SetOption(AddExtensionOption, value);
-        }
+        public bool AddExtension { get; set; }
```
jacdavis, line 57: >AddExtension { get; set; } [](http://example.com/codeflow?start=20&length=26)

I'm a little confused about this one. Is this no longer stored as an option in the like the other properties?

src/System.Windows.Forms/src/System/Windows/Forms/OpenFileDialog.cs
```diff
@@ -136,84 +133,55 @@ private protected override bool RunFileDialog(NativeMethods.OPENFILENAME_I ofn)
...
-            var openDialog = (WinFormsComWrappers.FileOpenDialogWrapper)dialog;
-            if (Multiselect)
+            if (!Multiselect)
             {
-                openDialog.GetResults(out WinFormsComWrappers.ShellItemArrayWrapper? results);
-                if (results is null)
-                {
-                    return Array.Empty<string>();
-                }
+                IShellItem* item;
```
jacdavis, line 140: >item [](http://example.com/codeflow?start=28&length=4)

Where is this released?

src/System.Windows.Forms/src/System/Windows/Forms/OpenFileDialog.cs
```diff
@@ -136,84 +133,55 @@ private protected override bool RunFileDialog(NativeMethods.OPENFILENAME_I ofn)
...
+            try
             {
-                openDialog.GetResult(out IShellItem? item);
-                if (item is null)
+                items->GetCount(out uint count);
+                string[] files = new string[count];
+                for (uint i = 0; i < count; ++i)
                 {
-                    return Array.Empty<string>();
+                    IShellItem* item;
```
jacdavis, line 158: >item [](http://example.com/codeflow?start=32&length=4)

Same as above. Where is the release on this?

src/System.Windows.Forms/src/System/Windows/Forms/SaveFileDialog.cs
```diff
@@ -81,74 +81,66 @@ public Stream OpenFile()
...
         private protected override bool PromptUserIfAppropriate(string fileName)
         {
             if (!base.PromptUserIfAppropriate(fileName))
             {
                 return false;
             }
 
-            //Note: When we are using the Vista dialog mode we get two prompts (one from us and one from the OS) if we do this
-            if ((_options & (int)Comdlg32.OFN.OVERWRITEPROMPT) != 0 && FileExists(fileName) && !UseVistaDialogInternal)
+            // Note: Vista dialog mode automatically promtps for overwrite.
```
jacdavis, line 114: >promtps [](http://example.com/codeflow?start=53&length=7)

nit: spelling

src/System.Windows.Forms/src/System/Windows/Forms/ComponentModel/COM2Interop/COM2TypeInfoProcessor.cs
```diff
@@ -495,7 +497,7 @@ private static PropertyDescriptor[] InternalGetProperties(object obj, ITypeInfo
             return props;
         }
 
-        private unsafe static PropInfo ProcessDataCore(ITypeInfo typeInfo, IDictionary propInfoList, DispatchID dispid, DispatchID nameDispID, in TYPEDESC typeDesc, VARFLAGS flags)
+        private static unsafe PropInfo ProcessDataCore(ITypeInfo typeInfo, IDictionary propInfoList, DispatchID dispid, DispatchID nameDispID, in TYPEDESC typeDesc, VARFLAGS flags)
         {
             // get the name and the helpstring
             using var nameBstr = new BSTR();
```
jacdavis, line 503: >= new BSTR(); [](http://example.com/codeflow?start=31&length=13)

Is this initialization required? These are out params in teh GetDocumentation call ([https://learn.microsoft.com/en-us/windows/win32/api/oaidl/nf-oaidl-itypeinfo-getdocumentation](https://learn.microsoft.com/en-us/windows/win32/api/oaidl/nf-oaidl-itypeinfo-getdocumentation)) so these are simply overwritten.

src/System.Windows.Forms/src/System/Windows/Forms/ComponentModel/COM2Interop/COM2TypeInfoProcessor.cs
```diff
@@ -507,14 +509,14 @@ private unsafe static PropInfo ProcessDataCore(ITypeInfo typeInfo, IDictionary p
                 throw new COMException(string.Format(SR.TYPEINFOPROCESSORGetDocumentationFailed, dispid, hr, ComNativeDescriptor.GetClassName(typeInfo)), (int)hr);
```
jacdavis, line 509: >(int)h [](http://example.com/codeflow?start=154&length=6)

Is this definition of HRESULT defined as unsigned? If so, this can throw an overflow exception.

src/System.Windows.Forms/src/System/Windows/Forms/ComponentModel/COM2Interop/COM2TypeInfoProcessor.cs
```diff
@@ -753,45 +755,53 @@ private unsafe static Type ProcessTypeInfoEnum(ITypeInfo enumTypeInfo)
                             varValue = null;
 
                             // get the name and the helpstring
-                            using var nameBstr = new BSTR();
-                            using var helpBstr = new BSTR();
+                            using BSTR nameBstr = new();
+                            using BSTR helpBstr = new();
```
jacdavis, line 759: Same as above. Why are these out params initialized? Is this to satisfy the using? Isn't this call SysAllocString for each instance?

src/System.Windows.Forms.Primitives/src/Interop/Ole32/Interop.RegisterDragDrop.cs
```diff
@@ -14,16 +15,15 @@ internal static partial class Ole32
...
+            if (!WinFormsComWrappers.Instance.TryGetComPointer(pDropTarget, IID.IDropTarget, out IUnknown* dropTargetPtr))
             {
-                return result;
+                return HRESULT.E_NOINTERFACE;
             }
 
-            result = RegisterDragDrop(hwnd.Handle, dropTargetPtr);
-            Marshal.Release(dropTargetPtr);
+            HRESULT result = RegisterDragDrop(hwnd.Handle, (nint)(void*)dropTargetPtr);
+            Marshal.Release((nint)(void*)dropTargetPtr);
```
jacdavis, line 26: >Marshal.Release((nint)(void*)dropTargetPtr); [](http://example.com/codeflow?start=12&length=44)

this leaks if RegisterDragDrop fails. Put in finally?

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/FileDialogCustomPlacesCollection.cs
```diff
@@ -3,30 +3,20 @@
...
-                try
-                {
-                    IShellItem? shellItem = customPlace.GetNativePath();
-                    if (shellItem is not null)
-                    {
-                        dialog.AddPlace(shellItem, 0);
-                    }
-                }
-                catch (FileNotFoundException)
+                IShellItem* shellItem = customPlace.GetNativePath();
```
jacdavis, line 16: >shellItem [](http://example.com/codeflow?start=28&length=9)

Is this an addref'd reference? If so, does it need a release call?

src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/BSTR.cs
```diff
@@ -0,0 +1,28 @@
...
+internal unsafe readonly partial struct BSTR : IDisposable
+{
+    public BSTR(string value) : this((char*)Marshal.StringToBSTR(value))
+    {
+    }
+
+    /// <summary>
+    ///  Returns the length of the native BSTR.
+    /// </summary>
+    public unsafe uint Length => Value is null ? 0 : *(((uint*)Value) - 1) / 2;
```
jacdavis, line 18: >/ 2; [](http://example.com/codeflow?start=75&length=4)

This assumption is only correct if the BSTR is utf16. It is ALWAYS utf16 on 32 bit windows, but if we were ever to do a cross plat version of winforms this would be a bug (most posix os's use utf8). In native code SysStringLen abstracts out this dependency and is a fast invocation. Would it be better to pinvoke that with a BSTR*?

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.IDataObjectVtbl.cs
```diff
@@ -15,33 +16,35 @@ internal static class IDataObjectVtbl
...
-                vtblRaw[3] = (IntPtr)(delegate* unmanaged<IntPtr, FORMATETC*, STGMEDIUM_Raw*, HRESULT>)&GetData;
-                vtblRaw[4] = (IntPtr)(delegate* unmanaged<IntPtr, FORMATETC*, STGMEDIUM_Raw*, HRESULT>)&GetDataHere;
-                vtblRaw[5] = (IntPtr)(delegate* unmanaged<IntPtr, FORMATETC*, HRESULT>)&QueryGetData;
-                vtblRaw[6] = (IntPtr)(delegate* unmanaged<IntPtr, FORMATETC*, FORMATETC*, HRESULT>)&GetCanonicalFormatEtc;
-                vtblRaw[7] = (IntPtr)(delegate* unmanaged<IntPtr, FORMATETC*, STGMEDIUM_Raw*, int, HRESULT>)&SetData;
-                vtblRaw[8] = (IntPtr)(delegate* unmanaged<IntPtr, DATADIR, IntPtr*, HRESULT>)&EnumFormatEtc;
-                vtblRaw[9] = (IntPtr)(delegate* unmanaged<IntPtr, FORMATETC*, ADVF, IntPtr, int*, HRESULT>)&DAdvise;
-                vtblRaw[10] = (IntPtr)(delegate* unmanaged<IntPtr, int, HRESULT>)&DUnadvise;
-                vtblRaw[11] = (IntPtr)(delegate* unmanaged<IntPtr, IntPtr*, HRESULT>)&EnumDAdvise;
+                IDataObject.Vtbl* vtblRaw = (IDataObject.Vtbl*)RuntimeHelpers.AllocateTypeAssociatedMemory(typeof(IDataObject.Vtbl), sizeof(IDataObject.Vtbl));
```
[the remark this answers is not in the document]
JeremyKuhne: This is only done for projecting to native from managed. In that case we need to keep all of our callbacks so I'm not sure what value it adds in this case?

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.IEnumFORMATETCVtbl.cs
JeremyKuhne: We'll be porting this one over to the new format shortly.

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.IDataObjectVtbl.cs
JeremyKuhne: Yeah, when we convert this one we're going to make sure we validate all of these out dereferences.

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.cs
JeremyKuhne: What does it return when something has been deref'ed to 0?

AaronRobinsonMSFT: Calling a function on a zero ref CCW is a bug and undefined.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetOpenFileName.cs
JeremyKuhne: I can't use `LibraryImport` in many cases as we reference types from CsWin32. This one is a temporary hack as the Windows metadata doesn't project a platform agnostic type (the issue is linked above it). It will be moved to CsWin32 once the issue is resolved.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetOpenFileName.cs
JeremyKuhne: Noted. This is another temporary because of the Win32 metadata issue linked.

src/System.Windows.Forms.Primitives/src/Windows/Win32/UI/Shell/IFileDialogEvents.cs
JeremyKuhne: I thought there were improvements made here regarding caching these? In any case, the hope is to move this to source generation (in CsWin32) so we might be able to entirely avoid it. I knew this wasn't as efficient but decided that it is more important (for our usages) that we're easily made consistent and the code is easy to maintain.

src/System.Windows.Forms/src/System/Windows/Forms/OpenFileDialog.cs
JeremyKuhne: Ah, missed this one. Thanks!

src/System.Windows.Forms/src/System/Windows/Forms/FolderBrowserDialog.cs
JeremyKuhne: I'll create a cleanup bit for this. I was riding the fence on this one as it really shouldn't fail.

src/System.Windows.Forms/src/System/Windows/Forms/FileDialog.cs
JeremyKuhne: It was our own option that was hacked into the other flags and filtered out when actually using the flags. Not too concerned about an extra field in this class.  :)

src/System.Windows.Forms.Primitives/src/Windows/Win32/UI/Shell/IFileDialogEvents.cs
AArnott: The compiler cannot cache a delegate that captures local variables. This one captures `pfd`.

src/System.Windows.Forms.Primitives/tests/UnitTests/Interop/Oleaut32/ITypeInfoTests.cs
```diff
@@ -92,8 +91,8 @@ public unsafe void ITypeInfo_GetDllEntry_Invoke_Success()
             ushort wOrdinal = ushort.MaxValue;
             hr = typeInfo.GetDllEntry((DispatchID)6, INVOKEKIND.FUNC, &dllName, &name, &wOrdinal);
             Assert.Equal(HRESULT.TYPE_E_BADMODULEKIND, hr);
-            Assert.Empty(dllName.String.ToString());
-            Assert.Empty(name.String.ToString());
+            Assert.True(dllName.Length == 0);
```
AArnott, line 94: tip: using `Assert.Equal(0, dllName.Length)` is better than using `True == ` because when it fails, the exception will tell you more about what the actual value was. In fact if you use `Assert.Equal(string.Empty, dllName)` here, you'd get the actual *string* in a failure case, which would be optimal.

src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/BSTR.cs
JeremyKuhne: The runtime does this directly as well: https://github.com/dotnet/runtime/blob/c6c8bb8b73ff4f7a686ca5a350fbb0093b3b708b/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.cs#L1263

The generated code also does it, but we can get that changed. If we do we'd probably want to get CsWin32 to add the `[SuppressGCTransition]` attribute to make it actually speedy.

Fwiw, the odds of doing an x-plat WinForms are astronomically low so we take Win32 as a given.

@AaronRobinsonMSFT, any thoughts on this from the `Marshal` perspective? Would exposing a `GetBSTRLength()` method on `Marshal` be plausible?

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetOpenFileName.cs
AArnott: You shouldn't need to declare this method. If you just declare the struct (which is the arch-specific API), then CsWin32 should be willing to generate the method that depends on it.

src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/BSTR.cs
jacdavis: Fair enough, but I mention it because this has been a very costly issue for me in the past. Using the length prefix ties to what is essentially an implementation detail. In native code, invoking SysStringLen is only a few instructions longer where the abstraction exists and avoids the problem. For native, avoiding the pinvoke may be reason enough to accept the risk here.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/FileDialogCustomPlacesCollection.cs
JeremyKuhne: Added a helper to scope these.

src/System.Windows.Forms.Primitives/src/Interop/Ole32/Interop.RegisterDragDrop.cs
JeremyKuhne: This one should only fail if there was something truly catastrophic (AV or OOM).

src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/BSTR.cs
AArnott: > Using the length prefix ties to what is essentially an implementation detail

BSTR's length prefix is not only documented, but hard-coded in several runtimes. It couldn't possibly change, nor would I feel bad about depending on it given it's documented. Why would you hesitate? Or did I misunderstand what the impl detail was that you were talking about, @jacdavis ?

The pointer arithmetic is complex, but if it works, fine. I wonder though why the hesitation to just reuse `Marshal.PtrToStringBSTR`? 
Also, what is `[SuppressGCTransition]` and why would its use be important here? Our own method would just be calling the `Marshal` method, which itself is managed.

src/System.Windows.Forms/src/System/Windows/Forms/ComponentModel/COM2Interop/COM2TypeInfoProcessor.cs
JeremyKuhne: Yes, this is to satisfy the using. They're empty and populated in the call.

src/System.Windows.Forms/src/System/Windows/Forms/ComponentModel/COM2Interop/COM2TypeInfoProcessor.cs
JeremyKuhne: It's an int. Or more specifically a struct around an `int`. There is an implicit converter on the struct. None of the operators do `checked`.

src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/BSTR.cs
jacdavis: It is only documented as being the length of the string in bytes. If the internal representation of a character changes as it can do on posix based systems that use UTF8 (1-4 byte variable length chars), then the conversion from bytes to characters changes becomes much more complex. I believe we finally standardized on utf16 for bstrs in the PAL even cross plat, but this means we constantly have to convert to and from utf8 when calling apis. I have no issue with leaving this as-is if we think we'll never hit this. However, I will say that I never expected to be porting the debugger cross plat either where I have hit this.

src/System.Windows.Forms/src/System/Windows/Forms/ComponentModel/COM2Interop/COM2TypeInfoProcessor.cs
jacdavis: Fair enough if the struct handles it. I didn't realize there was a struct wrapping it. I do see several other cases where int based hresults are checked for this reason.

src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/BSTR.cs
AArnott: Interesting. Well, CsWin32 is only generating code for Windows so far, so I don't anticipate any problem there. 
That's very interesting that BSTR is such a common exchange type but doesn't clearly define the width of a character. 
Length is hardly the worst problem here in that case. We may use direct BSTR pointers to create strings out of BSTRs, and vice versa, since their string representation is the same. Indeed, I believe the CLR itself pulls this trick and may be the reason System.String is both length prefixed and null terminated (at least IIRC). So if .NET programs ever run on a system where BSTR isn't UTF-16, those programs will have a harder time. I guess that's cause for using the Marshal methods to do the string/BSTR conversions and get the length though, which if we add this method to CsWin32-generated code I'm totally good to do. There's no (or little) reason to reimplement complex math or logic when a Marshal method does the same thing where we can leave it to the runtime to worry about interop. :)

src/System.Windows.Forms/src/System/Windows/Forms/FolderBrowserDialog.cs
JeremyKuhne: Here I think the risk of that being missed is pretty low. I did scope the item though.

src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/BSTR.cs
RussKie: @jacdavis what does PAL stand for?

src/System.Windows.Forms.Primitives/src/Windows/Win32/UI/Shell/IFileDialogEvents.cs
RussKie: A bit of related info can be found here: https://devblogs.microsoft.com/dotnet/performance_improvements_in_net_7/ in the paragraph that starts with "_Let’s start with IDE0200_".

src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/BSTR.cs
AaronRobinsonMSFT: > @AaronRobinsonMSFT, any thoughts on this from the Marshal perspective? Would exposing a GetBSTRLength() method on Marshal be plausible?

We could do it. I am dubious it is worth the effort though. Feel free to submit an api-proposal. It is likely this has already come up before though. I think the current approach is likely fine. My only minor quibble would be to use `sizeof(char)`.

> @jacdavis what does PAL stand for?

"Platform Abstraction Layer" or "Platform Adaptation Layer". The .NET runtime has one to help paper over platform differences. Unfortunately ours is sort of an Win32 emulation layer instead of a nicely abstracted PAL.

AaronRobinsonMSFT: Another data point that might be helpful is that in .NET Core/.NET 5+, the following marshalling scenarios have been marked obsolete: `UnmanagedType.AnsiBStr` and `UnmanagedType.TBStr`. We've not removed them yet, but since they've been marked this way for so long with perhaps one or two questions, I think we are close to a point where they could be removed entirely. For the new `LibraryImport` source generator, they aren't supported in any form.

PR #7936, opened 2022-10-13
Add ref struct for scoping COM objects

JeremyKuhne:
I'll continue to fiddle with the struct to see if I can make these cleaner, but this gets something in place to ensure we Release in all cases.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7936)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/System.Windows.Forms.Primitives.csproj
+++ b/src/System.Windows.Forms.Primitives/src/System.Windows.Forms.Primitives.csproj
@@ -24,6 +24,7 @@
     <Deterministic>true</Deterministic>
     <ProduceReferenceAssembly>true</ProduceReferenceAssembly>
     <UsePublicApiAnalyzers>true</UsePublicApiAnalyzers>
+    <RootNamespace/>
   </PropertyGroup>
 
   <ItemGroup>
@@ -42,6 +43,7 @@
 
   <ItemGroup>
     <EmbeddedResource Update="Resources\SR.resx">
+      <ManifestResourceName>System.Windows.Forms.Primitives.Resources.SR</ManifestResourceName>
       <GenerateSource>true</GenerateSource>
     </EmbeddedResource>
   </ItemGroup>
--- a/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/FileDialogCustomPlacesCollection.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/FileDialogCustomPlacesCollection.cs
@@ -13,8 +13,8 @@ namespace System.Windows.Forms
             for (int i = Items.Count - 1; i >= 0; --i)
             {
                 FileDialogCustomPlace customPlace = Items[i];
-                IShellItem* shellItem = customPlace.GetNativePath();
-                if (shellItem is not null)
+                using var shellItem = new ComScope<IShellItem>(customPlace.GetNativePath());
+                if (!shellItem.IsNull)
                 {
                     dialog->AddPlace(shellItem, 0).ThrowOnFailure();
                 }
--- a/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/FileDialogCustomPlacesCollection.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/FileDialogCustomPlacesCollection.cs
@@ -13,7 +13,7 @@ namespace System.Windows.Forms
             for (int i = Items.Count - 1; i >= 0; --i)
             {
                 FileDialogCustomPlace customPlace = Items[i];
-                using var shellItem = new ComScope<IShellItem>(customPlace.GetNativePath());
+                using ComScope<IShellItem> shellItem = new(customPlace.GetNativePath());
                 if (!shellItem.IsNull)
                 {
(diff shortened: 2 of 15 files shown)


JeremyKuhne:
@jacdavis this ensures the Release() calls.

JeremyKuhne:
Whoops- missed the magic sauce for the project file. I removed the default namespace as we're getting that completely from the folder structure. Necessitated manually setting the .resx location.

src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/ComScope.cs
```diff
@@ -0,0 +1,30 @@
...
+        public ComScope(T* value) => Value = value;
+
+        public static implicit operator T*(ComScope<T> scope) => scope.Value;
+
+        public bool IsNull => Value is null;
+
+        public void Dispose()
+        {
+            IUnknown* unknown = (IUnknown*)Value;
+            if (unknown is not null)
```
Tanya-Solyanik, line 24: Isn't `unknown is null` an error condition?

src/System.Windows.Forms/src/System/Windows/Forms/SaveFileDialog.cs
```diff
@@ -157,9 +157,9 @@ private protected unsafe override bool RunFileDialog(OPENFILENAME* ofn)
 
         private protected unsafe override string[] ProcessVistaFiles(IFileDialog* dialog)
         {
-            IShellItem* item;
-            dialog->GetResult(&item);
-            return item is null ? Array.Empty<string>() : new string[] { GetFilePathFromShellItem(item) };
+            using var item = new ComScope<IShellItem>(null);
+            dialog->GetResult((IShellItem**)&item);
```
lonitra, line 161: Do we want to call `ThrowOnFailure()` here?

src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/ComScope.cs
JeremyKuhne: No, as these are used as out parameters as well.

src/System.Windows.Forms.Primitives/src/System.Windows.Forms.Primitives.csproj
```diff
@@ -42,6 +43,7 @@
 
   <ItemGroup>
     <EmbeddedResource Update="Resources\SR.resx">
+      <ManifestResourceName>System.Windows.Forms.Primitives.Resources.SR</ManifestResourceName>
```
dreddy-work, line 46: nit: alignment. Q: Isn't embedded resource map to this name now?

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/FileDialogCustomPlacesCollection.cs
```diff
@@ -13,8 +13,8 @@ internal unsafe void Apply(IFileDialog* dialog)
             for (int i = Items.Count - 1; i >= 0; --i)
             {
                 FileDialogCustomPlace customPlace = Items[i];
-                IShellItem* shellItem = customPlace.GetNativePath();
-                if (shellItem is not null)
+                using var shellItem = new ComScope<IShellItem>(customPlace.GetNativePath());
```
dreddy-work, line 16: Loni has been using following syntax and would be consistent  here and other places in this file and avoids 'var'.
```suggestion
                using ComScope<IShellItem> shellItem = new (customPlace.GetNativePath());
```

src/System.Windows.Forms/src/System/Windows/Forms/FolderBrowserDialog.cs
```diff
@@ -383,9 +385,10 @@ private void SetOption(FILEOPENDIALOGOPTIONS option, bool value)
         private unsafe void GetResult(IFileOpenDialog* dialog)
         {
             IShellItem* item;
-            dialog->GetResult(&item);
+            dialog->GetResult(&item).ThrowOnFailure();
```
dreddy-work, line 388: would `item` be null in case of success? earlier condition wasn't checking return value and was relying on this.

src/System.Windows.Forms.Primitives/src/System.Windows.Forms.Primitives.csproj
JeremyKuhne: It did, but I deliberately turned off the default namespace as it was never correct for our code files. This necessitated manually setting this.

src/System.Windows.Forms/src/System/Windows/Forms/SaveFileDialog.cs
JeremyKuhne: Added.

src/System.Windows.Forms/src/System/Windows/Forms/FolderBrowserDialog.cs
JeremyKuhne: Reworked

lonitra (approved):

PR #7954, opened 2022-10-17
Issue 7815 Fixing HelpText property for AccessibleObject of TextBox

### LINKED ISSUE #7815: TextBox should set HelpText accessible property when it contains placeholder text
### .NET version

8.0.100-alpha.1.22423.9

No

No

### Issue description

From [Edit docs](https://learn.microsoft.com/en-us/windows/win32/winauto/uiauto-supporteditcontroltype):

> If an edit control contains placeholder text (for example, a cue banner), the text should be used as the HelpText property unless the text can be edited by the user and then reused as placeholder text. For example, the Windows Internet Explorer address bar contains the text "about:Tabs" when a new tab is opened. This is not HelpText because it is a programmatic address that can be used or edited by the user.

There is a `PlaceholderText` property on a `TextBox` which acts as a placeholder. It cannot be edited by the user and disappears when `TextBox` gets focus, thus needs to be exposed as a `HelpText`.

### Current behavior:

When setting `PlaceholderText` property on a `TextBox`, the `HelpText` property stays empty, and placeholder isn't announced by screen readers. 

A screenshot of a form with 3 text boxes, where the second TextBox has a placeholder text. Accessibility Insights app shows that HelpText property is empty on a second text box

Narrator Buddy output when switching between these TextBoxes:

> edit, Text box 1, 
> edit, 
> edit, Text box 3, 

Placeholder text is not announced.

### Expected behavior:
When `PlaceholderText` is non-empty, `HelpText` accessible property should have value of `PlaceholderText`. As a result, it should also be announced by screen readers. 

### Steps to reproduce

Sample application:
[TextBoxPlaceholderSample.zip](https://github.com/dotnet/winforms/files/9617884/TextBoxPlaceholderSample.zip)

v-elnovikova:
Fixes #7815

## Proposed changes

- Created TextBoxAccessibleObject class with overridden GetPropertyValue method.

## Customer Impact

- The HelpText property will be filled.

## Regression? 

- No

## Risk

- Minimal

### Before

before

### After

after

## Test methodology 

- Added unit test for HelpTextPropertyId case.

## Accessibility testing  

- Inspect
 

## Test environment(s) 

- .NET SDK 7.0.100-rc.2.22477.23
- Windows 11

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7954)


src/System.Windows.Forms/src/System/Windows/Forms/TextBox.TextBoxAccessibleObject.cs
```diff
@@ -0,0 +1,30 @@
+﻿using static Interop;
+
+namespace System.Windows.Forms
+{
+    public partial class TextBox
+    {
+        internal class TextBoxAccessibleObject : TextBoxBaseAccessibleObject
+        {
+            private readonly TextBox _owningTextBox;
```
Tanya-Solyanik, line 9: This field is not used, we can add it later when it's needed.

Tanya-Solyanik (commented):
Looks good!

src/System.Windows.Forms/src/System/Windows/Forms/TextBox.TextBoxAccessibleObject.cs
```diff
@@ -0,0 +1,30 @@
...
+                _owningTextBox = owner;
+            }
+
+            internal override object? GetPropertyValue(UiaCore.UIA propertyID)
+            {
+                switch (propertyID)
+                {
+                    case UiaCore.UIA.HelpTextPropertyId:
+                        string? placeholderText = _owningTextBox?.PlaceholderText;
+                        return string.IsNullOrEmpty(placeholderText) ? base.GetPropertyValue(propertyID) : placeholderText;
```
RussKie, line 22: What about whitespaces? E.g., a space (` `) or a tab (`\t`)?

v-elnovikova: When TextBox.PlaceholderText is not set, it has a value of empty string. So here IsNullOrEmpty checks if user has set any value. If a user explicitly uses whitespaces as placeholder, should it be considered as a valid HelpText?

I tried it on Windows 11: if HelpText has whitespace value, for example `<U+200B><U+200A><U+2009>`, then Narrator does not voice it, and Narrator Buddy shows it like `edit, Text box text, ​  ⠀, `

src/System.Windows.Forms/src/System/Windows/Forms/TextBox.TextBoxAccessibleObject.cs
v-elnovikova: Removed the field, now getting PlaceholderText value via the Owner property from base class.

Now TextBoxAccessibleObject class is not keeping any extra refs to TextBox, so no memory leaks are possible here. Thanks for pointing this out!

dmitrii-drobotov: @Tanya-Solyanik it was used for getting the `PlaceholderText` of a `TextBox`, and in my opinion `_owningTextBox.PlaceholderText` is better than `(Owner as TextBox)?.PlaceholderText`. It also shouldn't create a memory leak, so it's safe to use it.

Tanya-Solyanik: I see, I missed the use case, thanks for pointing it out.

RussKie: > `_owningTextBox.PlaceholderText` is better than `(Owner as TextBox)?.PlaceholderText`

Since C# 9 we can use [covariant returns](https://learn.microsoft.com/dotnet/csharp/language-reference/proposals/csharp-9.0/covariant-returns), though we never got around enhancing the code. Updating the code will allow us to remove the duplicate fields and use `Owner` property throughout.

NikitaSemenovAkvelon (approved):
LGTM

src/System.Windows.Forms/src/System/Windows/Forms/TextBox.TextBoxAccessibleObject.cs
RussKie: It could be a good exercise, if you have spare cycles.

RussKie (approved):

RussKie:
Thank you

PR #7973, opened 2022-10-19
Fix top-level window size for non-linear dimensions based on AutoScaleMode

dreddy-work:
In PermonV2 mode applications, We have been using linear sizes for top-level windows that are provided by the Windows irrespective of the AutoScaleMode.  This has been problematic for AutoScaleMode.Font, where Form scaling should be non-linear given its child controls are scaled non-linearly and depend on the Font that was assigned to Form/child controls.

We are making changes to utilize [GetDpiScaledSize](https://learn.microsoft.com/en-us/windows/win32/hidpi/wm-getdpiscaledsize) message and let Windows know that Form may need non-linear sizes depending on AutoScaleMode.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7973)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
@@ -143,11 +143,14 @@ namespace System.Windows.Forms
 
                 // If no one has configured auto scale dimensions yet, the scaling factor
                 // is the unit scale.
-                _currentAutoScaleFactor = saved.IsEmpty ? new SizeF(1.0F, 1.0F) : new SizeF(current.Width / saved.Width, current.Height / saved.Height);
+                _currentAutoScaleFactor = GetCurrentAutoScaleFactor(current, saved);
 
                 return _currentAutoScaleFactor;
             }
         }
+
+        internal static SizeF GetCurrentAutoScaleFactor(SizeF currentAutoScaleDimensions, SizeF savedAutoScaleDimensions)
+            => savedAutoScaleDimensions.IsEmpty ? new SizeF(1.0F, 1.0F) : new SizeF(currentAutoScaleDimensions.Width / savedAutoScaleDimensions.Width, currentAutoScaleDimensions.Height / savedAutoScaleDimensions.Height);
 
         /// <summary>
         ///  Determines the scaling mode of this control. The default is no scaling.
@@ -316,34 +319,42 @@ namespace System.Windows.Forms
             {
                 if (_currentAutoScaleDimensions.IsEmpty)
                 {
-                    switch (AutoScaleMode)
-                    {
-                        case AutoScaleMode.Font:
-                            _currentAutoScaleDimensions = GetFontAutoScaleDimensions();
-                            break;
-
-                        case AutoScaleMode.Dpi:
-                            // Screen Dpi
-                            if (DpiHelper.IsPerMonitorV2Awareness)
-                            {
-                                _currentAutoScaleDimensions = new SizeF(_deviceDpi, _deviceDpi);
-                            }
-                            else
-                            {
-                                // This Dpi value comes from the primary monitor.
-                                _currentAutoScaleDimensions = new SizeF(DpiHelper.DeviceDpi, DpiHelper.DeviceDpi);
-                            }
-
-                            break;
-
-                        default:
-                            _currentAutoScaleDimensions = AutoScaleDimensions;
-                            break;
-                    }
+                    _currentAutoScaleDimensions = GetCurrentAutoScaleDimensions(FontHandle);
                 }
 
                 return _currentAutoScaleDimensions;
             }
+        }
+
+        internal SizeF GetCurrentAutoScaleDimensions(HFONT fontHandle)
+        {
+            var currentAutoScaleDimensions = SizeF.Empty;
+            switch (AutoScaleMode)
+            {
+                case AutoScaleMode.Font:
+                    currentAutoScaleDimensions = GetFontAutoScaleDimensions(fontHandle);
+                    break;
+
+                case AutoScaleMode.Dpi:
+                    // Screen Dpi
+                    if (DpiHelper.IsPerMonitorV2Awareness)
+                    {
+                        currentAutoScaleDimensions = new SizeF(_deviceDpi, _deviceDpi);
+                    }
+                    else
+                    {
+                        // This Dpi value comes from the primary monitor.
+                        currentAutoScaleDimensions = new SizeF(DpiHelper.DeviceDpi, DpiHelper.DeviceDpi);
+                    }
+
+                    break;
+
+                default:
+                    currentAutoScaleDimensions = AutoScaleDimensions;
+                    break;
+            }
+
+            return currentAutoScaleDimensions;
         }
 
         /// <summary>
@@ -730,7 +741,7 @@ namespace System.Windows.Forms
         /// <summary>
         ///  This method calculates the auto scale dimensions based on the control's current font.
         /// </summary>
-        private unsafe SizeF GetFontAutoScaleDimensions()
+        private unsafe SizeF GetFontAutoScaleDimensions(HFONT fontHandle)
         {
             SizeF retval = SizeF.Empty;
 
@@ -750,7 +761,7 @@ namespace System.Windows.Forms
             // We must do the same here if our dialogs are to scale in a
             // similar fashion.
 
-            using PInvoke.SelectObjectScope fontSelection = new(dc, FontHandle);
+            using PInvoke.SelectObjectScope fontSelection = new(dc, fontHandle);
 
             TEXTMETRICW tm = default;
             PInvoke.GetTextMetrics(dc, &tm);
@@ -1422,7 +1433,7 @@ namespace System.Windows.Forms
             }
         }
 
-        internal void ScaleContainerForDpi(int deviceDpiNew, int deviceDpiOld, Rectangle suggestedRectangle)
+        internal void ScaleContainerForDpi(int deviceDpiNew, Rectangle suggestedRectangle)
         {
             CommonProperties.xClearAllPreferredSizeCaches(this);
             SuspendAllLayout(this);
@@ -1432,9 +1443,7 @@ namespace System.Windows.Forms
                 {
                     // The suggested rectangle comes from Windows, and it does not match with our calculations for scaling controls by AutoscaleFactor.
                     // Hence, we cannot use AutoscaleFactor here for scaling the control properties. See the below description for more details.
-                    float xScaleFactor = (float)suggestedRectangle.Width / Width;
-                    float yScaleFactor = (float)suggestedRectangle.Height / Height;
-                    ScaleMinMaxSize(xScaleFactor, yScaleFactor, updateContainerSize: false);
+                    ScaleMinMaxSize(AutoScaleFactor.Width, AutoScaleFactor.Height, updateContainerSize: false);
                 }
 
                 // If this container is a top-level window, we would receive WM_DPICHANGED message that
@@ -1461,16 +1470,8 @@ namespace System.Windows.Forms
                 // this control further by the 'OnFontChanged' event.
                 _isScaledByDpiChangedEvent = true;
 
-                // Factor is used only to scale Font. After that AutoscaleFactor kicks in to scale controls.
-                var factor = ((float)deviceDpiNew) / deviceDpiOld;
-
-                // DpiFontscache is available only in PermonitorV2 mode applications.
-                if (!TryGetDpiFont(deviceDpiNew, out Font? fontForDpi))
-                {
-                    Font currentFont = Font;
-                    fontForDpi = currentFont.WithSize(currentFont.Size * factor);
-                    AddToDpiFonts(deviceDpiNew, fontForDpi);
-                }
+                TryGetDpiFont(deviceDpiNew, out Font? fontForDpi);
+                Debug.Assert(fontForDpi != null, "We should have Font updated for Dpi from WM_GETDPISCALEDSIZE message");
 
                 ScaledControlFont = fontForDpi;
                 if (IsFontSet())
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
@@ -150,7 +150,9 @@ namespace System.Windows.Forms
         }
 
         internal static SizeF GetCurrentAutoScaleFactor(SizeF currentAutoScaleDimensions, SizeF savedAutoScaleDimensions)
-            => savedAutoScaleDimensions.IsEmpty ? new SizeF(1.0F, 1.0F) : new SizeF(currentAutoScaleDimensions.Width / savedAutoScaleDimensions.Width, currentAutoScaleDimensions.Height / savedAutoScaleDimensions.Height);
+            => savedAutoScaleDimensions.IsEmpty ?
+            new SizeF(1.0F, 1.0F)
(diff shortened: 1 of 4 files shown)


dreddy-work:
Tests pass locally. Need to investigate.

src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
```diff
@@ -143,12 +143,15 @@ protected SizeF AutoScaleFactor
...
                 // is the unit scale.
-                _currentAutoScaleFactor = saved.IsEmpty ? new SizeF(1.0F, 1.0F) : new SizeF(current.Width / saved.Width, current.Height / saved.Height);
+                _currentAutoScaleFactor = GetCurrentAutoScaleFactor(current, saved);
 
                 return _currentAutoScaleFactor;
             }
         }
 
+        internal static SizeF GetCurrentAutoScaleFactor(SizeF currentAutoScaleDimensions, SizeF savedAutoScaleDimensions)
+            => savedAutoScaleDimensions.IsEmpty ? new SizeF(1.0F, 1.0F) : new SizeF(currentAutoScaleDimensions.Width / savedAutoScaleDimensions.Width, currentAutoScaleDimensions.Height / savedAutoScaleDimensions.Height);
```
RussKie, line 153: [nit] this looks like a very long line (even by my standards, which are lax). Do you have [Editor guidelines](https://marketplace.visualstudio.com/items?itemName=PaulHarrington.EditorGuidelines) extension installed?

src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
```diff
@@ -1461,16 +1470,8 @@ internal void ScaleContainerForDpi(int deviceDpiNew, int deviceDpiOld, Rectangle
...
-
-                // DpiFontscache is available only in PermonitorV2 mode applications.
-                if (!TryGetDpiFont(deviceDpiNew, out Font? fontForDpi))
-                {
-                    Font currentFont = Font;
-                    fontForDpi = currentFont.WithSize(currentFont.Size * factor);
-                    AddToDpiFonts(deviceDpiNew, fontForDpi);
-                }
+                TryGetDpiFont(deviceDpiNew, out Font? fontForDpi);
+                Debug.Assert(fontForDpi != null, "We should have Font updated for Dpi from WM_GETDPISCALEDSIZE message");
```
RussKie, line 1474:
```suggestion
                Debug.Assert(fontForDpi is not null, "Font should have been updated for DPI from WM_GETDPISCALEDSIZE message");
```

RussKie: `TryGetDpiFont(..)` indicates whether it's successful or not. Why aren't we using the result and throwing, if it's unexpected?

src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
```diff
@@ -4321,7 +4321,35 @@ private void WmDpiChanged(ref Message m)
         [EditorBrowsable(EditorBrowsableState.Advanced)]
         protected virtual bool OnGetDpiScaledSize(int deviceDpiOld, int deviceDpiNew, ref Size desiredSize)
         {
-            return false; // scale linearly
+            // Compute and update Font for the current Dpi.
+            var factor = ((float)deviceDpiNew) / deviceDpiOld;
```
RussKie, line 4325: Please use the type here

src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
```diff
@@ -4321,7 +4321,35 @@ private void WmDpiChanged(ref Message m)
         [EditorBrowsable(EditorBrowsableState.Advanced)]
         protected virtual bool OnGetDpiScaledSize(int deviceDpiOld, int deviceDpiNew, ref Size desiredSize)
         {
-            return false; // scale linearly
+            // Compute and update Font for the current Dpi.
```
RussKie, line 4324: This comment is either misplaces or inaccurate. The adjacent code calculates the scaling factor.

RussKie, line 4324: Here and other comments - the casing appears rather arbitrary. DPI is an acronym and should be capitalized. The "font" doesn't appear to carry any special meaning that could warrant capitalization.

src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
```diff
@@ -4321,7 +4321,35 @@ private void WmDpiChanged(ref Message m)
...
+
+            // Dpi specific fonts cache is available only in PermonitorV2 mode applications.
+            if (!TryGetDpiFont(deviceDpiNew, out Font? fontForDpi))
+            {
+                Font currentFont = Font;
+                fontForDpi = currentFont.WithSize(currentFont.Size * factor);
+                AddToDpiFonts(deviceDpiNew, fontForDpi);
+            }
+
+            // If AutoScaleMode is Dpi, We continue with the linear size we get from Windows for the top-level window.
```
RussKie, line 4335:
```suggestion
            // If we DPI-scaling then we continue with the linear size we get from Windows for the top-level window.
```

src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
```diff
@@ -4321,7 +4321,35 @@ private void WmDpiChanged(ref Message m)
...
+            // need non-linear size for top-level window based on AutoScaleMode property.
+            FontHandleWrapper fontwrapper = new FontHandleWrapper(fontForDpi);
+            SizeF currentAutoScaleDimensions = GetCurrentAutoScaleDimensions(fontwrapper.Handle);
+            SizeF autoScaleFactor = GetCurrentAutoScaleFactor(currentAutoScaleDimensions, AutoScaleDimensions);
+
+            desiredSize.Width = (int)(Size.Width * autoScaleFactor.Width);
+            desiredSize.Height = (int)(Size.Height * autoScaleFactor.Height);
+            Debug.WriteLine($"AutoScaleFactor computed for new Dpi = {autoScaleFactor.Width} - {autoScaleFactor.Height}");
+
+            return true; // Notifying Windows that the top-level window size should be based on AutoScale mode.
```
RussKie, line 4352:
```suggestion
            // Notify Windows that the top-level window size should be based on AutoScaleMode value.
            return true; 
```

src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
dreddy-work: We won't see it null in any practical/real-time scenarios but when writing tests, if you don't send WM_GETDPISCALEDSIZE message but send WM_DPICHNAGED directly, you will end up in this state. Debug assertion is to help those test scenarios.

src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
dreddy-work: i don't insist but as per our guidelines, type is clear here to use `var`.

src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
dreddy-work: I still feel my sentence is better describing it.  `we DPI-scaling` is ambiguous and doesn't tell how that happen. Any better way to include `AutoScaleMode` property in the description to make it clear on what type of scaling is coming form?

src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
RussKie: Right, makes sense. Perhaps it's worth adding a comment here.

src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
RussKie: Makes sense. Perhaps then `If AutoScaleMode=AutoScaleMode.Dpi...`

src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
```diff
@@ -4321,7 +4321,35 @@ private void WmDpiChanged(ref Message m)
         [EditorBrowsable(EditorBrowsableState.Advanced)]
         protected virtual bool OnGetDpiScaledSize(int deviceDpiOld, int deviceDpiNew, ref Size desiredSize)
         {
-            return false; // scale linearly
+            float factor = ((float)deviceDpiNew) / deviceDpiOld;
```
KlausLoeffelmann, line 4324: I am wondering:
When a Form is a) used as a control or b) is an MDI-Form, does that also work the intended way?

src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
```diff
@@ -4321,7 +4321,35 @@ private void WmDpiChanged(ref Message m)
...
+
+            // Compute font for the current DPI and cache it. DPI specific fonts cache is available only in PermonitorV2 mode applications.
+            if (!TryGetDpiFont(deviceDpiNew, out Font? fontForDpi))
+            {
+                Font currentFont = Font;
+                fontForDpi = currentFont.WithSize(currentFont.Size * factor);
+                AddToDpiFonts(deviceDpiNew, fontForDpi);
+            }
+
+            // If AutoScaleMode=AutoScaleMode.Dpi, We continue with the linear size we get from Windows for the top-level window.
```
KlausLoeffelmann, line 4334:
```suggestion
            // If AutoScaleMode=AutoScaleMode.Dpi then we continue with the linear size we get from Windows for the top-level window.
```

src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
```diff
@@ -1461,16 +1472,10 @@ internal void ScaleContainerForDpi(int deviceDpiNew, int deviceDpiOld, Rectangle
...
+                TryGetDpiFont(deviceDpiNew, out Font? fontForDpi);
 
-                // DpiFontscache is available only in PermonitorV2 mode applications.
-                if (!TryGetDpiFont(deviceDpiNew, out Font? fontForDpi))
-                {
-                    Font currentFont = Font;
-                    fontForDpi = currentFont.WithSize(currentFont.Size * factor);
-                    AddToDpiFonts(deviceDpiNew, fontForDpi);
-                }
+                // For test scenarios, if we send WM_DPICHNAGED message but not WM_GETDPISCALEDSIZE message, fontForDpi may be null.
```
KlausLoeffelmann, line 1477:
```suggestion
                // For test scenarios, if we send WM_DPICHANGED message but not WM_GETDPISCALEDSIZE message, fontForDpi may be null.
```

KlausLoeffelmann:
> We have been using linear sizes for top-level windows that are provided by the Windows irrespective of the AutoScaleMode.

I don't understand what that means. Do you have a screenshot or a GIF which could illustrate that?

KlausLoeffelmann (commented):
Have a question and a few typo nits.

src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
dreddy-work: In those cases, it is not top-level window and does not receive these messages. They only get DPI_CHANGED_BEEFORE and AFTER events.

dreddy-work:
> > We have been using linear sizes for top-level windows that are provided by the Windows irrespective of the AutoScaleMode.
> 
> I don't understand what that means. Do you have a screenshot or a GIF which could illustrate that?

Gif may not justify but let me try describing this:

When you move Form from 100% monitor to 200% monitor,  factor become 2 for both `width` and `height`. Windows use this factor and scale the Form. However, when AutoScaleMode is `Font`, for the same move operation, Factors will not be linear and have separate factors for `Width` and `Height` (i.e  width factor - 1.1 and height factor 0.95 etc) .

src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
```diff
@@ -143,12 +143,17 @@ protected SizeF AutoScaleFactor
...
+                _currentAutoScaleFactor = GetCurrentAutoScaleFactor(current, saved);
 
                 return _currentAutoScaleFactor;
             }
         }
 
+        internal static SizeF GetCurrentAutoScaleFactor(SizeF currentAutoScaleDimensions, SizeF savedAutoScaleDimensions)
+            => savedAutoScaleDimensions.IsEmpty ?
+            new SizeF(1.0F, 1.0F)
+            : new SizeF(currentAutoScaleDimensions.Width / savedAutoScaleDimensions.Width, currentAutoScaleDimensions.Height / savedAutoScaleDimensions.Height);
```
RussKie, line 155: [nit]
```suggestion
            => savedAutoScaleDimensions.IsEmpty
                ? new(1.0F, 1.0F)
                : new(currentAutoScaleDimensions.Width / savedAutoScaleDimensions.Width, currentAutoScaleDimensions.Height / savedAutoScaleDimensions.Height);
```

src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
```diff
@@ -1461,16 +1472,10 @@ internal void ScaleContainerForDpi(int deviceDpiNew, int deviceDpiOld, Rectangle
...
 
-                // DpiFontscache is available only in PermonitorV2 mode applications.
-                if (!TryGetDpiFont(deviceDpiNew, out Font? fontForDpi))
-                {
-                    Font currentFont = Font;
-                    fontForDpi = currentFont.WithSize(currentFont.Size * factor);
-                    AddToDpiFonts(deviceDpiNew, fontForDpi);
-                }
+                // For test scenarios, if we send WM_DPICHNAGED message but not WM_GETDPISCALEDSIZE message, fontForDpi may be null.
+                Debug.Assert(fontForDpi != null, "We should have Font updated for Dpi from WM_GETDPISCALEDSIZE message");
```
RussKie, line 1478:
```suggestion
                Debug.Assert(fontForDpi is not null, "Font should have been updated for DPI from WM_GETDPISCALEDSIZE message");```

File: src/System.Windows.Forms/src/System/Windows/Forms/Form.cs [hunk 12]
```diff
@@ -4321,7 +4321,35 @@ private void WmDpiChanged(ref Message m)
...
+            // Calculate AutoscaleFactor for AutoScaleMode.Font. We will be using this factor to scale child controls
+            // and use same factor to compute desired size for top-level windows for the current DPI.
+            // This desired size is then used to notify Windows that we need non-linear size for top-level window.
+            FontHandleWrapper fontwrapper = new FontHandleWrapper(fontForDpi);
+            SizeF currentAutoScaleDimensions = GetCurrentAutoScaleDimensions(fontwrapper.Handle);
+            SizeF autoScaleFactor = GetCurrentAutoScaleFactor(currentAutoScaleDimensions, AutoScaleDimensions);
+
+            desiredSize.Width = (int)(Size.Width * autoScaleFactor.Width);
+            desiredSize.Height = (int)(Size.Height * autoScaleFactor.Height);
+            Debug.WriteLine($"AutoScaleFactor computed for new Dpi = {autoScaleFactor.Width} - {autoScaleFactor.Height}");
```
RussKie, line 4349:
```suggestion
            Debug.WriteLine($"AutoScaleFactor computed for new DPI = {autoScaleFactor.Width} - {autoScaleFactor.Height}");
```

RussKie (approved):
👍 with few nits

RussKie (approved):

dreddy-work:
Thank you.

dreddy-work:
https://github.com/dotnet/docs/issues/33696

PR #7982, opened 2022-10-21
Fix `FolderBrowserDialog.ShowDialog`

### LINKED ISSUE #7981: The original FolderBrowserDialog cannot be opened
### .NET version

8.0.100-alpha.1.22514.14 

Yes

yes, cannot repro in .Net 7.0 

### Issue description

The FolderBrowserDialog cannot be opened when setting AutoUpgradeEnabled as false. This issue cannot be resolved after the PR fixing: https://github.com/dotnet/winforms/pull/7978
**.Net 8.0 result:**
image
_System.AccessViolationException
  HResult=0x80004003
  Message=Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
  Source=Cannot evaluate the exception source
  StackTrace:
Cannot evaluate the exception stack trace_

**.Net 7.0 result:**
image

### Steps to reproduce

1. Open the [TestApp.zip](https://github.com/dotnet/winforms/files/9835559/TestApp.zip) project in VS
2. Build and run it
3. Click the button in the form and observe the result

elachlan:
Fixes #7981

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7982)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/FolderBrowserDialog.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/FolderBrowserDialog.cs
@@ -494,7 +494,7 @@ namespace System.Windows.Forms
                     {
                         // Try to retrieve the path from the IDList
                         char* buffer = stackalloc char[PInvoke.MAX_PATH + 1];
-                        bool isFileSystemFolder = PInvoke.SHGetPathFromIDList(selectedPidl, (PWSTR)b);
+                        bool isFileSystemFolder = PInvoke.SHGetPathFromIDList(selectedPidl, (PWSTR)buffer);
                         PInvoke.SendMessage(hwnd, (User32.WM)PInvoke.BFFM_ENABLEOK, 0, (nint)(BOOL)isFileSystemFolder);
                     }
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/FolderBrowserDialog.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/FolderBrowserDialog.cs
@@ -473,8 +473,7 @@ namespace System.Windows.Forms
             {
                 case PInvoke.BFFM_INITIALIZED:
                     // Indicates the browse dialog box has finished initializing. The lpData value is zero.
-
-                    var instance = (FolderBrowserDialog)GCHandle.FromIntPtr(lParam).Target!;
+                    var instance = (FolderBrowserDialog)GCHandle.FromIntPtr(lpData).Target!;
                     if (instance._initialDirectory.Length != 0)
                     {
                         // Try to expand the folder specified by initialDir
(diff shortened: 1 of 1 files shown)


elachlan:
CC: @lonitra @JeremyKuhne

RussKie:
We can probably enumerate open windows and search for the dialog window.
In UIIntegration project we have methods to search for controls, maybe something like that can be used...
Just random thoughts

src/System.Windows.Forms/src/System/Windows/Forms/FolderBrowserDialog.cs
```diff
@@ -491,11 +490,15 @@ private static unsafe int FolderBrowserDialog_BrowseCallbackProc(HWND hwnd, uint
                 case PInvoke.BFFM_SELCHANGED:
                     // Indicates the selection has changed. The lpData parameter points to the item identifier list for the newly selected item.
                     ITEMIDLIST* selectedPidl = (ITEMIDLIST*)lParam;
-                    if (selectedPidl is not null)
+                    Span<char> buffer = stackalloc char[PInvoke.MAX_PATH + 1];
```
JeremyKuhne, line 493: This should be inside the `if (selectedPidl is not null)` and should assign to `char* buffer` so you can just pass it as the argument. Then you can avoid the unnecessary pin syntax.

JeremyKuhne (changes requested):
Thanks for looking into this! Just one comment.

elachlan:
The other issue was we were passing lParam to `GCHanle.FromIntPtr` which caused another exception since its `null` when `msg == 'BFFM_INITIALIZED'`.

elachlan:
With regards to the tests. I looked at adding a `RunCommonDialogAsync` to `ControlTestBase` Or making a `ComponentTestBase`. Issue is that `CommonDialog` doesn't have a `Close()` method, since its always called via `ShowDialog`. I think someone more skilled will need to put together a plan so we can test the folder/save/open dialogs.

RussKie:
> I have no idea how to write a test for this since `showdialog()` will block.

Could probably try something like this: https://stackoverflow.com/a/42405020/2338036.

elachlan:
> > I have no idea how to write a test for this since `showdialog()` will block.
> 
> Could probably try something like this: https://stackoverflow.com/a/42405020/2338036.

Thanks, So I'd have to call `ShowDialog` in a different thread and then call to window via that code?

RussKie:
I think - the other way around. Here's a raw braindump of how I'd try it - spawn a new thread (e.g., a Windows Forms timer) that would search for a window and cap the lifetime of that thread too (e.g., give it like 3 seconds, else fail with timeout). If the window is found - close it - send WM_DESTROY message or ESC key...

JeremyKuhne (approved):
Thanks! Going to put this in to get manual coverage of the scenario. You can follow up with a unit test for it on another PR.

RussKie:
@JeremyKuhne we shouldn't be merging fixes without tests, as then it makes it harder to validate. There was nothing pressing about this regression and the test was already being developed.

Issue #7974, opened 2022-10-20
Use covariant returns for `Owner` property in accessible objects


dmitrii-drobotov:
Since C# 9 we can use [covariant returns](https://learn.microsoft.com/dotnet/csharp/language-reference/proposals/csharp-9.0/covariant-returns), though we never got around enhancing the code. Updating the code will allow us to remove the duplicate fields and use `Owner` property throughout.

_Originally posted by @RussKie in https://github.com/dotnet/winforms/pull/7954#discussion_r998992435_

Tanya-Solyanik:
https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/breaking-change-rules.md#source-and-binary-compatibility-changes:

```
Adding virtual to a member

While this change would often work without breaking too many scenarios because C# compiler tends to emit callvirt IL instructions to call non-virtual methods (callvirt performs a null check, while a normal call won't), we can't rely on it. C# is not the only language we target and the C# compiler increasingly tries to optimize callvirt to a normal call whenever the target method is non-virtual and the this is provably not null (such as a method accessed through the ?. null propagation operator). Making a method virtual would mean that consumer code would often end up calling it non-virtually.
```

merriemcgaw:
We need to start with a new internal property on `ControlAccessibleObject` and then as we work on others we pick them up.

NikitaSemenovAkvelon:
Tried to use covariant return for new property `OwnerInternal`, but got a `TypeLoadException` in `Marshal.GetIUnknownForObject` method:

> System.TypeLoadException
  HResult=0x80131522
  Message=Signature of the body and declaration in a method implementation do not match.  Type: 'ButtonBaseAccessibleObject'.  Assembly: 'System.Windows.Forms, Version=8.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.
  Source=System.Private.CoreLib
  StackTrace:
   at System.Runtime.InteropServices.Marshal.GetIUnknownForObjectNative(Object o)
   at System.Windows.Forms.Control.WmGetObject(Message& m) in C:\Projects\winforms.main\src\System.Windows.Forms\src\System\Windows\Forms\Control.cs:line 12152

https://github.com/dotnet/winforms/blob/1b8ac24a2b13f78eae47e177abfe8a59658372de/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs#L12152
[...]

PR #7983, opened 2022-10-21
Fixed MaskedTextBoxAccessibleObject.Name if Owner.Mask is filled

### LINKED ISSUE #7938: For MaskedTextBox with Mask setting, the Name property in Inspect should be empty
### .NET version

8.0.100-alpha.1.22512.32

Yes

Yes, cannot repro in previous .Net versions: 7.0/6.0

### Issue description

For MaskedTextBox with Mask setting, the Name property in Inspect should be empty. This issue is regressed from PR fixing: https://github.com/dotnet/winforms/pull/7741

**Actual result:**
image

**Expected result:**
image

### Steps to reproduce

**Repro steps:**
1.	Create a Winforms .Net 8.0 application
2.	Add several maskedTextBox controls
3.	Set them with different mask: PhoneNumeber, ShortDate
4.	Build and run application, use Inspect to check these controls' Name property

or use this attached sample application:
[WinFormsApp8.zip](https://github.com/dotnet/winforms/files/9774999/WinFormsApp8.zip)

NikitaSemenovAkvelon:
Fixes #7938

## Proposed changes

- Overriden `Name` property in `MaskedTextBoxAccessibleObject`.
- Added new unit tests. 

## Customer Impact

- The `Name` of `MaskedTextBoxAccessibleObject` is more clarify.

## Regression? 

- Yes

## Risk

- Minimal

### Before

Screenshot 2022-10-21 121439

### After

Screenshot 2022-10-21 120955

## Test methodology 

- Manually
- Unit testing 

## Accessibility testing  

- Inspect

 

## Test environment(s) 

.NET SDK:
 Commit:    1b80461e45

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.22621
 OS Platform: Windows

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7983)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/MaskedTextBox.MaskedTextBoxAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/MaskedTextBox.MaskedTextBoxAccessibleObject.cs
@@ -12,6 +12,16 @@ namespace System.Windows.Forms
             {
             }
 
+            public override string? Name
+            {
+                get => string.IsNullOrEmpty((Owner as MaskedTextBox)?.Mask)
+                    ? base.Name
+                    // If base.Name is null mask template will be used as a name, which is not descriptive for users.
+                    // Instead, we want to show an empty string to signal developers to set an appropriate name.
+                    : base.Name ?? string.Empty;
+                set => base.Name = value;
+            }
+
             protected override string ValueInternal => Owner.WindowText;
         }
     }
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/MaskedTextBox.MaskedTextBoxAccessibleObjectTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/MaskedTextBox.MaskedTextBoxAccessibleObjectTests.cs
@@ -75,6 +75,37 @@ namespace System.Windows.Forms.Tests
         }
 
         [WinFormsTheory]
+        [InlineData(null, null)]
+        [InlineData("Test", "Test")]
+        public void MaskedTextBoxAccessibleObject_Name_IsExpected_WithoutMask(string accessibleName, string expectedAccessibleName)
+        {
+            using MaskedTextBox maskedTextBox = new MaskedTextBox();
+            maskedTextBox.Text = "000000";
+            maskedTextBox.AccessibleName = accessibleName;
+
+            var actual = (string)maskedTextBox.AccessibilityObject.GetPropertyValue(UiaCore.UIA.NamePropertyId);
+
+            Assert.Equal(expectedAccessibleName, actual);
+            Assert.False(maskedTextBox.IsHandleCreated);
+        }
+
+        [WinFormsTheory]
+        [InlineData(null, "")]
+        [InlineData("Test", "Test")]
+        public void MaskedTextBoxAccessibleObject_Name_IsExpected_WithMask(string accessibleName, string expectedAccessibleName)
+        {
+            using MaskedTextBox maskedTextBox = new MaskedTextBox();
+            maskedTextBox.Text = "000000";
+            maskedTextBox.Mask = "00/00/0000";
+            maskedTextBox.AccessibleName = accessibleName;
+
+            var actual = (string)maskedTextBox.AccessibilityObject.GetPropertyValue(UiaCore.UIA.NamePropertyId);
+
+            Assert.Equal(expectedAccessibleName, actual);
+            Assert.False(maskedTextBox.IsHandleCreated);
+        }
+
+        [WinFormsTheory]
         [InlineData(true)]
         [InlineData(false)]
         public void MaskedTextBoxAccessibleObject_GetPropertyValue_Value_IsExpected_WithMask(bool useMask)


dmitrii-drobotov (approved):
Looks good, but I would add a comment to clarify the change

src/System.Windows.Forms/src/System/Windows/Forms/MaskedTextBox.MaskedTextBoxAccessibleObject.cs
```diff
@@ -12,6 +12,14 @@ public MaskedTextBoxAccessibleObject(MaskedTextBox owner) : base(owner)
             {
             }
 
+            public override string? Name
+            {
+                get => string.IsNullOrEmpty((Owner as MaskedTextBox)!.Mask)
+                    ? base.Name
```
Tanya-Solyanik, line 18: Do we have a test in WinformsControlsApp that covers MaskedTextBox? Could you please add one if not.

src/System.Windows.Forms/src/System/Windows/Forms/MaskedTextBox.MaskedTextBoxAccessibleObject.cs
```diff
@@ -12,6 +12,14 @@ public MaskedTextBoxAccessibleObject(MaskedTextBox owner) : base(owner)
             {
             }
 
+            public override string? Name
```
Tanya-Solyanik, line 15: I expected `UseTextForAccessibility` style to handle this case and not read the text box text.  https://github.com/dotnet/winforms/blob/55b05866b52ec11ce1eeedb89b6ab8d6b1ef9571/src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.cs#L98-L102 

and
https://github.com/dotnet/winforms/blob/55b05866b52ec11ce1eeedb89b6ab8d6b1ef9571/src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlAccessibleObject.cs#L311-L323

Do you see why this didn't work?

src/System.Windows.Forms/src/System/Windows/Forms/MaskedTextBox.MaskedTextBoxAccessibleObject.cs
```diff
@@ -12,6 +12,14 @@ public MaskedTextBoxAccessibleObject(MaskedTextBox owner) : base(owner)
...
+            {
+                get => string.IsNullOrEmpty((Owner as MaskedTextBox)!.Mask)
+                    ? base.Name
+                    : base.Name ?? string.Empty;
```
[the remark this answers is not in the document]
RussKie: I suggest against using `!` unless absolultely necessary, and this doesn't look like that case.
`Owner as MaskedTextBox` is superflous, see https://github.com/dotnet/winforms/issues/7974.

src/System.Windows.Forms/src/System/Windows/Forms/MaskedTextBox.MaskedTextBoxAccessibleObject.cs
NikitaSemenovAkvelon: CollectionEditors form contains MaskedTextBox. This form is on my screenshots above.

src/System.Windows.Forms/src/System/Windows/Forms/MaskedTextBox.MaskedTextBoxAccessibleObject.cs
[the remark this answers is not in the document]
NikitaSemenovAkvelon: Maybe the entire refactoring should be located in a single PR? Also, #7974 is not in progress yet.

src/System.Windows.Forms/src/System/Windows/Forms/MaskedTextBox.MaskedTextBoxAccessibleObject.cs
NikitaSemenovAkvelon: It's working. But in this case, it hasn't any Label.

- Without any changes, the `Name` gets a value from a previous `Label`:
Screenshot 2022-10-24 100512
- Set `UseTextForAccessibility` style to `true`. The name gets a value from the owner's `Text`:
Screenshot 2022-10-24 100620

The issue occurs because the `Name` property is changed to `Value` somewhere outside if we return null.

src/System.Windows.Forms/src/System/Windows/Forms/MaskedTextBox.MaskedTextBoxAccessibleObject.cs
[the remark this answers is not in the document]
Tanya-Solyanik: >Maybe the entire refactoring should be located in a single PR?

not necessarily, fill free to start here.

src/System.Windows.Forms/src/System/Windows/Forms/MaskedTextBox.MaskedTextBoxAccessibleObject.cs
Tanya-Solyanik: >The issue occurs because the Name property is changed to Value somewhere outside if we return null.

This is probably happening in the system IAccessible.

src/System.Windows.Forms/src/System/Windows/Forms/MaskedTextBox.MaskedTextBoxAccessibleObject.cs
Tanya-Solyanik, line 19: Please add a comment that we want to avoid the use of `Mask` as an AccessibleName

src/System.Windows.Forms/src/System/Windows/Forms/MaskedTextBox.MaskedTextBoxAccessibleObject.cs
```diff
@@ -10,8 +10,21 @@ internal class MaskedTextBoxAccessibleObject : TextBoxBaseAccessibleObject
         {
             public MaskedTextBoxAccessibleObject(MaskedTextBox owner) : base(owner)
             {
+                Owner = owner;
             }
 
+            public override string? Name
+            {
+                get => string.IsNullOrEmpty((Owner as MaskedTextBox)!.Mask)
```
dmitrii-drobotov, line 18: It should work now
```suggestion
                get => string.IsNullOrEmpty(Owner.Mask)
```

dmitrii-drobotov (dismissed):

src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlAccessibleObject.cs
```diff
@@ -338,7 +338,7 @@ public IntPtr Handle
                 }
             }
 
-            public Control Owner { get; }
+            public virtual Control Owner { get; }
```
Tanya-Solyanik, line 341: Sorry about stirring you into a wrong direction. This is a breaking change - 
```
Adding virtual to a member

While this change would often work without breaking too many scenarios because C# compiler tends to emit callvirt IL instructions to call non-virtual methods (callvirt performs a null check, while a normal call won't), we can't rely on it. C# is not the only language we target and the C# compiler increasingly tries to optimize callvirt to a normal call whenever the target method is non-virtual and the this is provably not null (such as a method accessed through the ?. null propagation operator). Making a method virtual would mean that consumer code would often end up calling it non-virtually.
```
https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/breaking-change-rules.md#source-and-binary-compatibility-changes

We could add a second property with the same content which would be internal virtual. But it's really best done in a separate PR.

RussKie: Good call, @Tanya-Solyanik.

We should add a new internal property:
```cs
internal virtual Control OwnerInternal { get; }
```

dmitrii-drobotov (approved):

Tanya-Solyanik (approved):

PR #7986, opened 2022-10-21
Remove BinaryFormatter support from Control's ActiveX projection.

JeremyKuhne:
In order for properties to be exposed to `IPropertyBag` they must have a TypeConverter to and from either string (preferred) or `byte[]`.

These were always the first choice, `ISerializable` was the final fallback. This matches what we've done elsewhere when we've removed `BinaryFormatter`.

cc: @GrabYourPitchforks, @merriemcgaw 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7986)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
@@ -11,7 +11,6 @@ using System.Drawing;
 using System.Globalization;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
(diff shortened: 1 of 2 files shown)


lonitra (approved):

JeremyKuhne:
There will be a follow up change to deal with the `IStream` persistence using `BinaryFormatter` in this class.

JeremyKuhne:
I put this back with #8597 as I mistakenly thought the plan was to completely remove the `BinaryFormatter` from the runtime when I made this change.

PR #7987, opened 2022-10-21
Break out nested classes in Control.ActiveXImpl.

JeremyKuhne:
Also do some code cleanup in the main class.

Added a couple sanity tests on VARIANT behavior with the .NET runtime.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7987)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
@@ -2,16 +2,12 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System.Collections;
 using System.Collections.Specialized;
 using System.ComponentModel;
 using System.Diagnostics;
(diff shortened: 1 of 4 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
```diff
@@ -1768,7 +1457,9 @@ internal void Save(Oleaut32.IPropertyBag pPropBag, BOOL fClearDirty, BOOL fSaveA
 
                     if (converter.CanConvertFrom(typeof(string)))
                     {
+                        VARIANT variant = new();
```
lonitra, line 1460: Should this be removed from this change since it's not being used currently?

JeremyKuhne: We're taking the address of it below.

lonitra (approved):

PR #7992, opened 2022-10-22
Enable nullability in Link class

gpetrou:
## Proposed changes

- Enable nullability in Link class.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7992)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/LinkLabel.Link.LinkAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/LinkLabel.Link.LinkAccessibleObject.cs
@@ -33,11 +33,10 @@ namespace System.Windows.Forms
                             return Rectangle.Empty;
                         }
 
-                        Region region = _owningLink.VisualRegion;
-                        using Graphics graphics = Graphics.FromHwnd(_owningLink.Owner.Handle);
+                        Region? region = _owningLink.VisualRegion;
+                        using Graphics graphics = Graphics.FromHwnd(_owningLink.Owner?.Handle ?? IntPtr.Zero);
 
                         // Make sure we have a region for this link
-                        //
                         if (region is null)
                         {
                             _owningLinkLabel.EnsureRun(graphics);
@@ -52,7 +51,6 @@ namespace System.Windows.Forms
                         rect = Rectangle.Ceiling(region.GetBounds(graphics));
 
                         // Translate rect to screen coordinates
-                        //
                         return _owningLinkLabel.RectangleToScreen(rect);
                     }
                 }
@@ -61,7 +59,7 @@ namespace System.Windows.Forms
 
                 public override string DefaultAction => SR.AccessibleActionClick;
 
-                public override string Description => _owningLink.Description;
+                public override string? Description => _owningLink.Description;
 
                 public override void DoDefaultAction()
                 {
--- a/src/System.Windows.Forms/src/System/Windows/Forms/LinkLabel.Link.LinkAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/LinkLabel.Link.LinkAccessibleObject.cs
@@ -28,13 +28,13 @@ namespace System.Windows.Forms
                 {
                     get
                     {
-                        if (!_owningLinkLabel.IsHandleCreated)
+                        if (!_owningLinkLabel.IsHandleCreated || _owningLink.Owner is null)
                         {
                             return Rectangle.Empty;
                         }
 
                         Region? region = _owningLink.VisualRegion;
-                        using Graphics graphics = Graphics.FromHwnd(_owningLink.Owner?.Handle ?? IntPtr.Zero);
+                        using Graphics graphics = Graphics.FromHwnd(_owningLink.Owner.Handle);
 
                         // Make sure we have a region for this link
                         if (region is null)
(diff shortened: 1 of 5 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/LinkLabel.Link.LinkAccessibleObject.cs
```diff
@@ -33,11 +33,10 @@ public override Rectangle Bounds
                             return Rectangle.Empty;
                         }
 
-                        Region region = _owningLink.VisualRegion;
-                        using Graphics graphics = Graphics.FromHwnd(_owningLink.Owner.Handle);
+                        Region? region = _owningLink.VisualRegion;
+                        using Graphics graphics = Graphics.FromHwnd(_owningLink.Owner?.Handle ?? IntPtr.Zero);
```
dreddy-work, line 37: Are you forced to pass `IntPtr.Zero` here?

src/System.Windows.Forms/src/System/Windows/Forms/LinkLabel.cs
```diff
@@ -1195,7 +1195,7 @@ protected override void OnPaint(PaintEventArgs e)
                             {
                                 //exclude the area to draw the focus rectangle
                                 g.Clip = originalClip;
-                                RectangleF[] rects = _links[0].VisualRegion.GetRegionScans(e.GraphicsInternal.Transform);
+                                RectangleF[]? rects = _links[0].VisualRegion?.GetRegionScans(e.GraphicsInternal.Transform);
```
dreddy-work, line 1198: We were not expecting `VisualRegion` to be null at this point in earlier code. Do you have a scenario that it could?

src/System.Windows.Forms/src/System/Windows/Forms/LinkLabel.Link.LinkAccessibleObject.cs
RussKie: Yes, if `Owner` is null (which it can be) then we have no handle.
I would, however, move the check up to line:31 - if we don't have a valid handle, then we create `Graphics` object for the desktop area, which feels wrong. In this case it probably better to return an empty rectangle.

@Tanya-Solyanik @dmitrii-drobotov what do you think?

src/System.Windows.Forms/src/System/Windows/Forms/LinkLabel.cs
RussKie: @dreddy-work this looks like a missing null-check, because everywhere else `VisualRegion` is checked for null. We can't guarantee that `VisualRegion` isn't null here.

RussKie (commented):
👍 modulo the discussion in LAO.

src/System.Windows.Forms/src/System/Windows/Forms/LinkLabel.cs
dreddy-work: wouldn't we hear by now if there was a scenario for this to be null in this event?

src/System.Windows.Forms/src/System/Windows/Forms/LinkLabel.Link.LinkAccessibleObject.cs
dreddy-work: I know owner can be null. I mean, is null not accepted for this call? what is forcing to explicit pass of `IntPtr.Zero`?

RussKie: I'm not sure I understand your questions... Please help me.
IntPtr is a struct, [Graphics.FromHwnd(IntPtr)](https://learn.microsoft.com/dotnet/api/system.drawing.graphics.fromhwnd) doesn't accept null.

dreddy-work: my mistake. i thought that method accepts `null`.

dmitrii-drobotov: > Yes, if `Owner` is null (which it can be) then we have no handle. I would, however, move the check up to line:31 - if we don't have a valid handle, then we create `Graphics` object for the desktop area, which feels wrong. In this case it probably better to return an empty rectangle.
> 
> @Tanya-Solyanik @dmitrii-drobotov what do you think?

I agree that returning an empty rectangle is better in this case. But this is probably an extremely rare scenario since `_owningLink.Owner` should be the same object as `_owningLinkLabel`:
https://github.com/dotnet/winforms/blob/9538742d94b721c7af4b1b88e343eab7934d9d16/src/System.Windows.Forms/src/System/Windows/Forms/LinkLabel.Link.cs#L46-L47
https://github.com/dotnet/winforms/blob/9538742d94b721c7af4b1b88e343eab7934d9d16/src/System.Windows.Forms/src/System/Windows/Forms/LinkLabel.Link.LinkAccessibleObject.cs#L20-L23

RussKie (approved):

src/System.Windows.Forms/src/System/Windows/Forms/LinkLabel.Link.LinkAccessibleObject.cs
RussKie: Thank you

RussKie:
Thank you

PR #8000, opened 2022-10-22
Enable nullability in ToolStripLabel

gpetrou:
## Proposed changes

- Enable nullability in ToolStripLabel.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8000)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripLabel.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripLabel.cs
@@ -33,27 +33,27 @@ namespace System.Windows.Forms
         }
 
         public ToolStripLabel(string? text)
-            : base(text, null, null)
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ToolStripLabel.cs
```diff
@@ -33,27 +32,33 @@ public ToolStripLabel()
         {
         }
 
-        public ToolStripLabel(string text) : base(text, null, null)
+        public ToolStripLabel(string? text)
+            : base(text, null, null)
```
dreddy-work, line 36: nit: would be great if you add parameter names for null constants.

dreddy-work (approved):

RussKie (approved):

PR #8018, opened 2022-10-24
Fix WM_GETDPISCALEDSIZE message handling.

dreddy-work:
WM_GETDPISCALEDSIZE is sent once per device and when moved back to same device, we only get DPI_CHANGED message but not WM_GETDPISCALEDSIZE. Windows cache the result from the first time and reuse it. This isn't very clear from the [documentation](https://learn.microsoft.com/en-us/windows/win32/hidpi/wm-getdpiscaledsize) and is found while experimenting and validating the scenarios.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8018)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
@@ -2196,7 +2196,8 @@ namespace System.Windows.Forms
             Debug.Assert(PInvoke.AreDpiAwarenessContextsEqualInternal(DpiAwarenessContext, DPI_AWARENESS_CONTEXT.DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2),
                 $"Fonts need to be cached only for PermonitorV2 mode applications : {DpiHelper.IsPerMonitorV2Awareness} : {DpiAwarenessContext}");
 
-            if(_dpiFonts is not null && _dpiFonts.TryGetValue(dpiNew, out Font? scaledFont))
+            _dpiFonts ??= new Dictionary<int, Font>();
+            if (_dpiFonts.TryGetValue(dpiNew, out Font? scaledFont))
             {
                 return scaledFont!;
             }
@@ -2204,7 +2205,6 @@ namespace System.Windows.Forms
             float factor = ((float)newDpi / oldDpi);
             scaledFont = font.WithSize(font.Size * factor);
 
-            _dpiFonts ??= new Dictionary<int, Font>();
             _dpiFonts.Add(dpiNew, scaledFont);
 
             return scaledFont;
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
@@ -2197,7 +2197,7 @@ namespace System.Windows.Forms
                 $"Fonts need to be cached only for PermonitorV2 mode applications : {DpiHelper.IsPerMonitorV2Awareness} : {DpiAwarenessContext}");
 
             _dpiFonts ??= new Dictionary<int, Font>();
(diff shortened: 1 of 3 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -2191,28 +2191,23 @@ public virtual Font Font
...
 
-        private protected bool TryGetDpiFont(int dpi, [NotNullWhen(true)] out Font? font)
-        {
-            font = null;
-            if (!PInvoke.AreDpiAwarenessContextsEqualInternal(DpiAwarenessContext, DPI_AWARENESS_CONTEXT.DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2))
-            {
-                Debug.Assert(false, $"Fonts need to be cached only for PermonitorV2 mode applications : {DpiHelper.IsPerMonitorV2Awareness} : {DpiAwarenessContext}");
-                return false;
-            }
+            _dpiFonts ??= new Dictionary<int, Font>();
```
JeremyKuhne, line 2207: As you always create this, you might as well move it before the check for null above.

JeremyKuhne (changes requested):
One minor comment

JeremyKuhne (dismissed):

lonitra (approved):
LGTM

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -2191,28 +2191,23 @@ public virtual Font Font
             }
         }
 
-        private protected void AddToDpiFonts(int dpi, Font font)
+        internal Font GetScaledFont(Font font, int newDpi, int oldDpi)
```
Tanya-Solyanik, line 2194: Is it possible to use Control.DeviceDpi instead of oldDpi?

Tanya-Solyanik (commented):
I'm not sure if Control.DeviceDpi can be used instead of the old DPI from the message

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
dreddy-work: in some cases, DeviceDPi is updated to the new DPI before calling this method. This new DPI is part of the message we are handling and should be more reliable if we ever reorder/restructure the code.

Tanya-Solyanik (approved):

dreddy-work:
Thank you.

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -2191,28 +2191,23 @@ public virtual Font Font
...
-
-        private protected bool TryGetDpiFont(int dpi, [NotNullWhen(true)] out Font? font)
-        {
-            font = null;
-            if (!PInvoke.AreDpiAwarenessContextsEqualInternal(DpiAwarenessContext, DPI_AWARENESS_CONTEXT.DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2))
+            if (_dpiFonts.TryGetValue(newDpi, out Font? scaledFont))
             {
-                Debug.Assert(false, $"Fonts need to be cached only for PermonitorV2 mode applications : {DpiHelper.IsPerMonitorV2Awareness} : {DpiAwarenessContext}");
-                return false;
+                return scaledFont!;
```
RussKie, line 2202: `!` shouldn't be required here

dreddy-work: Thanks. Check that in next commit.

PR #8021, opened 2022-10-24
Add Test for FolderBrowserDialog

### LINKED ISSUE #79: Add more tests to FolderBrowserDialog
Once more advanced testing infrastructure is in-place, `FolderBrowseDialog` can use additional tests.

elachlan:
Fixes #79
Related: #7981

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8021)

### DIFF
--- a/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FolderBrowserDialogTests.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FolderBrowserDialogTests.cs
@@ -2,12 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System.Diagnostics;
-using WindowsInput;
-using WindowsInput.Native;
 using Xunit;
 using Xunit.Abstractions;
-using Xunit.Sdk;
 using static Interop;
 
 namespace System.Windows.Forms.UITests
--- a/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FolderBrowserDialogTests.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FolderBrowserDialogTests.cs
@@ -16,12 +16,14 @@ namespace System.Windows.Forms.UITests
         {
         }
 
-        [WinFormsFact]
-        public void FolderBrowserDialog_ShowDialog()
+        [WinFormsTheory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public void FolderBrowserDialog_ShowDialog(bool autoUpgradeEnabled)
         {
             FolderBrowserDialog form = new()
             {
-                AutoUpgradeEnabled = false,
+                AutoUpgradeEnabled = autoUpgradeEnabled,
             };
 
             Timer timer = new();
--- a/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FolderBrowserDialogTests.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FolderBrowserDialogTests.cs
@@ -6,6 +6,7 @@ using WindowsInput;
 using WindowsInput.Native;
 using Xunit;
 using Xunit.Abstractions;
+using static Interop.Ole32;
 
 namespace System.Windows.Forms.UITests
 {
@@ -29,14 +30,15 @@ namespace System.Windows.Forms.UITests
             using Timer timer = new();
             timer.Interval = 1_000;
             int counter = 0;
+            bool failedClose = false;
             timer.Tick += (s, e) =>
             {
                 counter++;
                 if (counter > 2)
                 {
                     timer.Stop();
+                    failedClose=true;
                     Application.Exit();
-                    Assert.Fail("Failed to close the dialog");
                 }
 
                 // Close the dialog
@@ -47,7 +49,7 @@ namespace System.Windows.Forms.UITests
             dialog.ShowDialog();
 
             // The dialog has opened and closed successfully
-            Assert.True(true);
+            Assert.False(failedClose, "Failed to close the dialog");
         }
     }
 }
--- a/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FolderBrowserDialogTests.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FolderBrowserDialogTests.cs
@@ -25,15 +25,22 @@ namespace System.Windows.Forms.UITests
             };
 
             Timer timer = new();
-            timer.Interval = 3000;
+            timer.Interval = 1_000;
+            int counter = 0;
             timer.Tick += (s, e) =>
             {
-                timer.Stop();
+                counter++;
+                if (counter > 2)
+                    throw new TimeoutException("Failed to close the dialog");
+
                 new InputSimulator().Keyboard.KeyPress(VirtualKeyCode.ESCAPE);
             };
 
             timer.Start();
             form.ShowDialog();
+
+            // The dialog has opened and closed successfully
+            Assert.True(true);
         }
     }
 }
--- a/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FolderBrowserDialogTests.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FolderBrowserDialogTests.cs
@@ -28,13 +28,12 @@ namespace System.Windows.Forms.UITests
             timer.Interval = 3000;
             timer.Tick += (s, e) =>
             {
-                ((Timer)s).Stop();
+                timer.Stop();
                 new InputSimulator().Keyboard.KeyPress(VirtualKeyCode.ESCAPE);
             };
 
             timer.Start();
             form.ShowDialog();
         }
-
     }
 }
--- a/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FolderBrowserDialogTests.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FolderBrowserDialogTests.cs
@@ -27,7 +27,7 @@ namespace System.Windows.Forms.UITests
             };
 
             using Timer timer = new();
-            timer.Interval = 1_000;
+            timer.Interval = 2_000;
             int counter = 0;
             bool failedClose = false;
             timer.Tick += (s, e) =>
--- a/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FolderBrowserDialogTests.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FolderBrowserDialogTests.cs
@@ -49,7 +49,7 @@ namespace System.Windows.Forms.UITests
                         return;
                     }
 
-                    TestOutputHelper?.WriteLine($"Couldn't find the dialog ({counter}/3). Retrying...");
+                    TestOutputHelper.WriteLine($"Couldn't find the dialog ({counter}/3). Retrying...");
                     timer.Start();
                 }
             };
@@ -67,8 +67,8 @@ namespace System.Windows.Forms.UITests
                 PInvoke.GetWindowThreadProcessId(hwnd, &processId);
                 if (processId == PInvoke.GetCurrentProcessId() && PInvoke.IsWindowVisible(hwnd))
                 {
-                    TestOutputHelper?.WriteLine($"Process ID: {processId}");
-                    TestOutputHelper?.WriteLine($"Dialog window found: 0x{hwnd.Value:X8}");
+                    TestOutputHelper.WriteLine($"Process ID: {processId}");
+                    TestOutputHelper.WriteLine($"Dialog window found: 0x{hwnd.Value:X8}");
 
                     PInvoke.SendMessage(hwnd, User32.WM.CLOSE);
 
--- a/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FolderBrowserDialogTests.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FolderBrowserDialogTests.cs
@@ -6,7 +6,6 @@ using WindowsInput;
 using WindowsInput.Native;
 using Xunit;
 using Xunit.Abstractions;
-using static Interop.Ole32;
 
 namespace System.Windows.Forms.UITests
 {
--- a/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FolderBrowserDialogTests.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FolderBrowserDialogTests.cs
@@ -2,10 +2,13 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Diagnostics;
 using WindowsInput;
 using WindowsInput.Native;
 using Xunit;
 using Xunit.Abstractions;
+using Xunit.Sdk;
+using static Interop;
 
 namespace System.Windows.Forms.UITests
 {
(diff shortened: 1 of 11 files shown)


src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FolderBrowserDialogTests.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        public FolderBrowserDialogTests(ITestOutputHelper testOutputHelper)
+            : base(testOutputHelper)
+        {
+        }
+
+        [WinFormsFact]
+        public void FolderBrowserDialog_ShowDialog()
+        {
+            FolderBrowserDialog form = new()
```
RussKie, line 22:
```suggestion
            using FolderBrowserDialog form = new()
```

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FolderBrowserDialogTests.cs
```diff
@@ -0,0 +1,39 @@
...
+            {
+                AutoUpgradeEnabled = false,
+            };
+
+            Timer timer = new();
```
RussKie, line 27:
```suggestion
            using Timer timer = new();
```

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FolderBrowserDialogTests.cs
```diff
@@ -0,0 +1,39 @@
...
+            timer.Interval = 3000;
```
RussKie, line 28: This makes the timer fires every three seconds. We can probably reduce this value to 1,000.
Also we need to introduce a timeout mechanism - i.e., if the dialog remains open after, say, three seconds - fail the test.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FolderBrowserDialogTests.cs
```diff
@@ -0,0 +1,39 @@
...
+            timer.Tick += (s, e) =>
+            {
+                timer.Stop();
+                new InputSimulator().Keyboard.KeyPress(VirtualKeyCode.ESCAPE);
+            };
+
+            timer.Start();
+            form.ShowDialog();
+        }
```
RussKie, line 37: Psuedocode
```suggestion
            Timer timer = new();
            timer.Interval = 1_000;
            
            int counter = 0;
            timer.Tick += (s, e) =>
            {
                counter++;                
                if (counter > 2)
                    throw new TimeoutException("Failed to close the dialog");
                    
                new InputSimulator().Keyboard.KeyPress(VirtualKeyCode.ESCAPE);
            };

            timer.Start();
            form.ShowDialog();
            
            // The dialog has opened and closed successfully
            Assert.True(true);
        }
```

RussKie (changes requested):
I think you're on the right track, though a little more work is needed.

elachlan:
Sorry for the force push, I rebased to main.

RussKie:
No worries at all, the diff is simple :)

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FolderBrowserDialogTests.cs
```diff
@@ -0,0 +1,46 @@
...
+            : base(testOutputHelper)
+        {
+        }
+
+        [WinFormsFact]
+        public void FolderBrowserDialog_ShowDialog()
+        {
+            FolderBrowserDialog form = new()
+            {
+                AutoUpgradeEnabled = false,
```
RussKie, line 24: Also we'd want to test both cases, right? Suggest turning the test into a theory.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FolderBrowserDialogTests.cs
```diff
@@ -0,0 +1,46 @@
...
+            Timer timer = new();
+            timer.Interval = 1_000;
+            int counter = 0;
+            timer.Tick += (s, e) =>
+            {
+                counter++;
+                if (counter > 2)
+                    throw new TimeoutException("Failed to close the dialog");
+
+                new InputSimulator().Keyboard.KeyPress(VirtualKeyCode.ESCAPE);
```
RussKie, line 36: Have you checked the timeout approach is working? E.g., comment out line:36 and see if the test'd fail.

elachlan: It works! 

I get an exception here though, which I am unsure if its a bug:
https://github.com/dotnet/winforms/blob/7ba5f72a2698b3b2476a016f48f7066af8158c89/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/ThemingScope.cs#L55

```
System.Runtime.InteropServices.SEHException: 'External component has thrown an exception.'
System.Windows.Forms.Primitives.dll!System.Windows.Forms.ThemingScope.Deactivate(nint userCookie) Line 55	C#
System.Windows.Forms.dll!System.Windows.Forms.CommonDialog.ShowDialog(System.Windows.Forms.IWin32Window owner) Line 236	C#
System.Windows.Forms.dll!System.Windows.Forms.CommonDialog.ShowDialog() Line 158	C#
System.Windows.Forms.UI.IntegrationTests.dll!System.Windows.Forms.UITests.FolderBrowserDialogTests.FolderBrowserDialog_ShowDialog() Line 40	C#
[External Code]	
[Resuming Async Method]	
[External Code]	
System.Windows.Forms.dll!System.Windows.Forms.Control.InvokeMarshaledCallbackDo(System.Windows.Forms.Control.ThreadMethodEntry tme) Line 6652	C#
System.Windows.Forms.dll!System.Windows.Forms.Control.InvokeMarshaledCallbackHelper(object obj) Line 6600	C#
[External Code]	
System.Windows.Forms.dll!System.Windows.Forms.Control.InvokeMarshaledCallback(System.Windows.Forms.Control.ThreadMethodEntry tme) Line 6577	C#
System.Windows.Forms.dll!System.Windows.Forms.Control.InvokeMarshaledCallbacks() Line 6687	C#
System.Windows.Forms.dll!System.Windows.Forms.Control.WndProc(ref System.Windows.Forms.Message m) Line 13438	C#
System.Windows.Forms.dll!System.Windows.Forms.Control.ControlNativeWindow.OnMessage(ref System.Windows.Forms.Message m) Line 65	C#
System.Windows.Forms.dll!System.Windows.Forms.Control.ControlNativeWindow.WndProc(ref System.Windows.Forms.Message m) Line 114	C#
System.Windows.Forms.dll!System.Windows.Forms.NativeWindow.Callback(Windows.Win32.Foundation.HWND hWnd, Interop.User32.WM msg, Windows.Win32.Foundation.WPARAM wparam, Windows.Win32.Foundation.LPARAM lparam) Line 376	C#
[External Code]	
System.Windows.Forms.dll!System.Windows.Forms.Application.ComponentManager.Interop.Mso.IMsoComponentManager.FPushMessageLoop(nuint dwComponentID, Interop.Mso.msoloop uReason, void* pvLoopData) Line 340	C#
System.Windows.Forms.dll!System.Windows.Forms.Application.ThreadContext.RunMessageLoopInner(Interop.Mso.msoloop reason, System.Windows.Forms.ApplicationContext context) Line 1116	C#
System.Windows.Forms.dll!System.Windows.Forms.Application.ThreadContext.RunMessageLoop(Interop.Mso.msoloop reason, System.Windows.Forms.ApplicationContext context) Line 980	C#
System.Windows.Forms.dll!System.Windows.Forms.Application.DoEvents() Line 807	C#
[External Code]
```

RussKie: Do you think you could run your test against release/7.0 branch? This way we could establish a baseline whether this is something we're doing wrong in the test, or whether it's an indication of a regression of some kind in the main branch.
Thank you

elachlan: The failure is when I remove the "escape key press". I will test this scenario on 7.0 branch.

elachlan: 7.0 Branch also throws this error when we comment out the line:
`new InputSimulator().Keyboard.KeyPress(VirtualKeyCode.ESCAPE)`

So I think its good to go. Although I am unsure on the test failures here.

elachlan: @RussKie are you happy with this?

RussKie: I experimented with it, and I don't get the same error. I made few discoveries though.
1. We need to stop the timer upon the timeout
1. If we fail by timeout - we lock the test process with a popup dialog, which is quite bad:
image
2. We don't need the send any keys, and can terminate the application instead. Here's the demo code:
```cs
            using FolderBrowserDialog dialog = new()
            {
                AutoUpgradeEnabled = autoUpgradeEnabled,
            };

            using Timer timer = new();
            timer.Interval = 1_000;
            int counter = 0;

            bool dialogDismissed = true;

            timer.Tick += (s, e) =>
            {
                counter++;
                TestOutputHelper.WriteLine($"Counter {counter}");

                if (counter > 2)
                {
                    dialogDismissed = false;

                    TestOutputHelper.WriteLine($"Timeout");

                    timer.Stop();
                    TestOutputHelper.WriteLine($"timer.Stop");

                    Application.Exit();
                    TestOutputHelper.WriteLine($"Application.Exit");
                }

                //new InputSimulator().Keyboard.KeyPress(VirtualKeyCode.ESCAPE);
            };

            timer.Start();
            dialog.ShowDialog();

            // The dialog has opened and closed successfully
            TestOutputHelper.WriteLine($"End");
            Assert.True(dialogDismissed);
        }
```

But now it looks like we don't need the input simulator as we can dismiss the dialog ourselves (though in a brutal kind of way).

I have locally undone the fix and, having played with it, I think we can reduce the test down to the following:
```cs
            using FolderBrowserDialog dialog = new()
            {
                AutoUpgradeEnabled = autoUpgradeEnabled,
            };

            bool dialogDismissed = false;

            using Timer timer = new();
            timer.Interval = 1_000;

            timer.Tick += (s, e) =>
            {
                dialogDismissed = true;
                timer.Stop();

                // Forcefully close the dialog
                Application.Exit();
            };

            timer.Start();
            dialog.ShowDialog();

            // The dialog has opened and closed successfully
            Assert.True(dialogDismissed);
```

If the dialog has opned - we can force-close it. If it failed to open - it'd have brought down the test process.

elachlan: would the code ever hit this?
```csharp
// The dialog has opened and closed successfully
Assert.True(dialogDismissed);
```

Doesn't application.exit close the whole app and not just the dialog? Or are the tests treated as individual apps?

elachlan: With this code I get the following exception when running with autoUpgradeEnabled=true
```
System.Runtime.InteropServices.COMException: 'Catastrophic failure (0x8000FFFF (E_UNEXPECTED))'
System.Windows.Forms.Primitives.dll!Windows.Win32.Foundation.HRESULT.ThrowOnFailure(nint errorInfo) Line 63	C#
System.Windows.Forms.dll!System.Windows.Forms.FolderBrowserDialog.GetResult(Windows.Win32.UI.Shell.IFileOpenDialog* dialog) Line 383	C#
System.Windows.Forms.dll!System.Windows.Forms.FolderBrowserDialog.TryRunDialogVista(Windows.Win32.Foundation.HWND owner, out bool returnValue) Line 299	C#
System.Windows.Forms.dll!System.Windows.Forms.FolderBrowserDialog.RunDialog(nint hWndOwner) Line 261	C#
System.Windows.Forms.dll!System.Windows.Forms.CommonDialog.ShowDialog(System.Windows.Forms.IWin32Window owner) Line 221	C#
System.Windows.Forms.dll!System.Windows.Forms.CommonDialog.ShowDialog() Line 158	C#
System.Windows.Forms.UI.IntegrationTests.dll!System.Windows.Forms.UITests.FolderBrowserDialogTests.FolderBrowserDialog_ShowDialog(bool autoUpgradeEnabled) Line 42	C#
```

RussKie: I got this too in the trimmed down version. I didn't have time to investigate it though... The question whether this work on .NET 7 (i.e., pror to cswin32 work) or not.
Do you get the same with the input simulator? If not, we can probably retain that iteration.

elachlan: .NET7 has the same error. Might be best to use the input simulator to avoid the error.

elachlan: OK. I think this latest push is the best of both worlds. I imagine there are some minor improvements there.

RussKie: > would the code ever hit this?
> 
> ```cs
> // The dialog has opened and closed successfully
> Assert.True(dialogDismissed);
> ```
> 
> Doesn't application.exit close the whole app and not just the dialog? Or are the tests treated as individual apps?

Yes, that code is hit - if you run my sample you'd see the debug output.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FolderBrowserDialogTests.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            using Timer timer = new();
+            timer.Interval = 1_000;
+            int counter = 0;
+            timer.Tick += (s, e) =>
+            {
+                counter++;
+                if (counter > 2)
+                {
+                    timer.Stop();
```
RussKie, line 37: If we end up here - we need to fail the test. Right now, I don't think it's happening, is it?

elachlan: I think I need to add back in `dialogDismissed`. I'll test and adjust accordingly.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FolderBrowserDialogTests.cs
```diff
@@ -0,0 +1,53 @@
...
+                    Application.Exit();
+                    Assert.Fail("Failed to close the dialog");
```
RussKie, line 39: My local tests didn't indicate this will  be hit or result in a failed test.

elachlan: It works if I use only `Assert.Fail()`. 

I was looking into the exception caused by `ThemingScope.Deactivate(userCookie);` in `CommonDialog.cs`. The call to `PInvoke.DeactivateActCtx` causes the exception only if `Application.Exit()` or `Assert.Fail()` is called from the timer.tick event. I think this is threading related but I am not 100% sure.

https://stackoverflow.com/questions/26514954/registration-free-com-interop-deactivating-activation-context-in-finalizer-thro

elachlan: With this latest code, the vista version throws an exception because of an `E_UNEXPECTED` HRESULT returned from GetResult in this call chain. `dialog->GetResult(item).ThrowOnFailure();`.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FolderBrowserDialogTests.cs
```diff
@@ -0,0 +1,54 @@
...
+            timer.Interval = 1_000;
+            int counter = 0;
+            bool failedClose = false;
+            timer.Tick += (s, e) =>
+            {
+                counter++;
+                if (counter > 2)
+                {
+                    timer.Stop();
+                    failedClose=true;
```
RussKie, line 39:
```suggestion
                    failedClose = true;
```

RussKie (commented):
I think we're almost there :)

RussKie (approved):

elachlan:
Test is failing. Maybe its flaky? :(

RussKie:
image

elachlan:
Are the UI Integration tests run concurrently or do they run serially?

RussKie:
> Are the UI Integration tests run concurrently or do they run serially?

All tests in UI Integration project should be running serially:
https://github.com/dotnet/winforms/blob/8f1be5d8453763501aea20070bb6cabf97a49b34/eng/pipelines/build.yml#L108

RussKie (dismissed):

elachlan:
No other test uses the `InputSimulator` directly like this. It wraps everything to make sure the Form/Control are focused and the input is done on the UI thread.

elachlan:
@RussKie I think you did it! Can you please rerun the pipeline?

RussKie:
Indeed :)

RussKie:
I'm going to run this few times to see if there are any flakiness to it.

RussKie (approved):

RussKie:
Thank you!

PR #8032, opened 2022-10-26
Fixing incorrect Column value in ListViewSubItemAccessibleObject

### LINKED ISSUE #8025: [Accessibility] ArgumentOutOfRange exception is raised on selecting the last SubItem of ListViewItem in Details view if it has an image.
### .NET version

8.0.100-alpha.1.22512.5

Yes

Yes, reproduces only in .NET 8.

### Issue description

If select or just move cursor over the last `SubItem` of `ListViewItem` which has an image in `Details` view an `ArgumentOutOfRange` exception will be raised.
Screenshot 2022-10-25 160736
It is a regression bug from #7672

### Steps to reproduce

1. Create a new WinForms application.
2. Add ListView.
3. Set View to Details.
4. Add a new item with an image.
5. Select the last SubItem while some accessibility tool (e.g. AI) is working.

NikitaSemenovAkvelon:
Fixes #8025

## Proposed changes

- Fixed value in the `Column` property of `ListViewSubItemAccessibleObject` class.
- Covered `ListViewSubItemAccessibleObject.GetColumnHeaderItems` method with tests.

## Customer Impact

- Customer will see correct header info from a sub-item.

## Regression? 

- Yes

## Risk

- Minimal

### Before

Screenshot 2022-10-26 084435
Screenshot 2022-10-26 084406

### After

Screenshot 2022-10-26 084114

## Test methodology 

- Manually
- Unit tests 

## Accessibility testing  

- Inspect

 

## Test environment(s) 

.NET SDK:
 Commit:    1b80461e45

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.22621

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8032)

### DIFF
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItem.ListViewSubItem.ListViewSubItemAccessibleObjectTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItem.ListViewSubItem.ListViewSubItemAccessibleObjectTests.cs
@@ -974,6 +974,72 @@ namespace System.Windows.Forms.Tests
         }
 
         [WinFormsFact]
+        public void ListViewSubItemAccessibleObject_GetColumnHeaderItems_IsExpected()
+        {
+            using ListView control = GetListViewWithSubItemData(
+                View.Details,
+                createControl: false,
+                virtualMode: false,
+                showGroups: false,
+                columnCount: 3,
+                subItemCount: 2);
+
+            ListViewSubItemAccessibleObject_GetColumnHeaderItems_IsExpected_Internal(control);
+        }
+
+        [WinFormsFact]
+        public void ListViewSubItemAccessibleObject_GetColumnHeaderItems_IsExpected_WithImage()
+        {
+            using ImageList imageCollection = new();
+            imageCollection.Images.Add(Form.DefaultIcon);
+            using ListView control = GetListViewWithSubItemData(
+                View.Details,
+                createControl: false,
+                virtualMode: false,
+                showGroups: false,
+                columnCount: 3,
+                subItemCount: 2);
+            control.SmallImageList = imageCollection;
+            control.Items[0].ImageIndex = 0;
+
+            ListViewSubItemAccessibleObject_GetColumnHeaderItems_IsExpected_Internal(control);
+        }
+
+        private void ListViewSubItemAccessibleObject_GetColumnHeaderItems_IsExpected_Internal(ListView control)
+        {
+            ListView.ColumnHeaderCollection columns = control.Columns;
+            for (int i = 0; i < columns.Count; i++)
+            {
+                var subItemAccessibleObject = (ListViewSubItemAccessibleObject)control.Items[0].SubItems[i].AccessibilityObject;
+                AccessibleObject columnHeaderAccessibleObject = columns[i].AccessibilityObject;
+                UiaCore.IRawElementProviderSimple[] columnHeaderItems = subItemAccessibleObject.GetColumnHeaderItems();
+
+                Assert.Equal(1, columnHeaderItems.Length);
+                Assert.Same(columnHeaderAccessibleObject, columnHeaderItems[0]);
+            }
+        }
+
+        [WinFormsTheory]
+        [InlineData(View.LargeIcon)]
+        [InlineData(View.List)]
+        [InlineData(View.SmallIcon)]
+        [InlineData(View.Tile)]
+        public void ListViewSubItemAccessibleObject_GetColumnHeaderItems_ReturnsNull_NoColumsViews(View view)
+        {
+            using ListView control = GetListViewWithSubItemData(
+                view,
+                createControl: false,
+                virtualMode: false,
+                showGroups: false,
+                columnCount: 3,
+                subItemCount: 2);
+
+            var subItemAccessibleObject = (ListViewSubItemAccessibleObject)control.Items[0].SubItems[1].AccessibilityObject;
+
+            Assert.Null(subItemAccessibleObject.GetColumnHeaderItems());
+        }
+
+        [WinFormsFact]
         public void ListViewSubItemAccessibleObject_State_ReturnsFocusable()
         {
             using ListView listview = new();
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.ListViewSubItem.ListViewSubItemAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.ListViewSubItem.ListViewSubItemAccessibleObject.cs
@@ -83,7 +83,7 @@ namespace System.Windows.Forms
 
                 internal override int Column
                     => _owningListView.View == View.Details
-                        ? ParentInternal.GetChildIndex(this)
+                        ? ParentInternal.GetChildIndex(this) - ParentInternal.FirstSubItemIndex
                         : InvalidIndex;
 
                 /// <summary>
@@ -146,7 +146,9 @@ namespace System.Windows.Forms
                 internal override int Row => _owningItem.Index;
 
                 internal override UiaCore.IRawElementProviderSimple[]? GetColumnHeaderItems()
-                    => new UiaCore.IRawElementProviderSimple[] { _owningListView.Columns[Column].AccessibilityObject };
+                    => _owningListView.View == View.Details
+                        ? new UiaCore.IRawElementProviderSimple[] { _owningListView.Columns[Column].AccessibilityObject }
+                        : null;
 
                 internal override bool IsPatternSupported(UiaCore.UIA patternId)
                 {


src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItem.ListViewSubItem.ListViewSubItemAccessibleObjectTests.cs
```diff
@@ -973,6 +973,56 @@ public void ListViewSubItemAccessibleObject_GetPropertyValue_Pattern_ReturnsExpe
             Assert.False(listView.IsHandleCreated);
         }
 
+        [WinFormsFact]
+        public void ListViewSubItemAccessibleObject_GetColumnHeaderItems_AsExpected()
```
dmitrii-drobotov, line 977:
```suggestion
        public void ListViewSubItemAccessibleObject_GetColumnHeaderItems_IsExpected()
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItem.ListViewSubItem.ListViewSubItemAccessibleObjectTests.cs
```diff
@@ -973,6 +973,56 @@ public void ListViewSubItemAccessibleObject_GetPropertyValue_Pattern_ReturnsExpe
...
+        {
+            using var control = GetListViewWithSubItemData(View.Details, false, false, false, 3, 2);
+
+            for(int i = 0; i < control.Columns.Count; i++)
```
dmitrii-drobotov, line 981:
```suggestion
            for (int i = 0; i < control.Columns.Count; i++)
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItem.ListViewSubItem.ListViewSubItemAccessibleObjectTests.cs
```diff
@@ -973,6 +973,56 @@ public void ListViewSubItemAccessibleObject_GetPropertyValue_Pattern_ReturnsExpe
...
+                var columnHeaderAccessibleObject = control.Columns[i].AccessibilityObject;
+                var columnHeaderItems = subItemAccessibleObject.GetColumnHeaderItems();
+
+                Assert.Equal(1, columnHeaderItems.Length);
+                Assert.Same(columnHeaderAccessibleObject, columnHeaderItems[0]);
+            }
+        }
+
+        [WinFormsFact]
+        public void ListViewSubItemAccessibleObject_GetColumnHeaderItems_AsExpected_WithImage()
```
dmitrii-drobotov, line 993:
```suggestion
        public void ListViewSubItemAccessibleObject_GetColumnHeaderItems_IsExpected_WithImage()
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItem.ListViewSubItem.ListViewSubItemAccessibleObjectTests.cs
```diff
@@ -973,6 +973,56 @@ public void ListViewSubItemAccessibleObject_GetPropertyValue_Pattern_ReturnsExpe
...
+        [InlineData(View.List)]
+        [InlineData(View.SmallIcon)]
+        [InlineData(View.Tile)]
+        public void ListViewSubItemAccessibleObject_GetColumnHeaderItems_ThrowsException_NotSupportedViews(View view)
+        {
+            using var control = GetListViewWithSubItemData(view, false, false, false, 3, 2);
+
+            var subItemAccessibleObject = (ListViewSubItemAccessibleObject)control.Items[0].SubItems[1].AccessibilityObject;
+
+            Assert.Throws<ArgumentOutOfRangeException>(subItemAccessibleObject.GetColumnHeaderItems);
```
dmitrii-drobotov, line 1023: I would expect that `GetColumnHeaderItems` simply returned `null` if columns are not supported, like in `DataGridViewCellAccessibleObject`. Is it something that changed recently or we were always throwing this exception?

NikitaSemenovAkvelon: Only in the `Details` view TableItemPattern is supported. Therefore you can't reproduce this case in runtime.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItem.ListViewSubItem.ListViewSubItemAccessibleObjectTests.cs
```diff
@@ -973,6 +973,56 @@ public void ListViewSubItemAccessibleObject_GetPropertyValue_Pattern_ReturnsExpe
             Assert.False(listView.IsHandleCreated);
         }
 
+        [WinFormsFact]
+        public void ListViewSubItemAccessibleObject_GetColumnHeaderItems_IsExpected()
+        {
+            using var control = GetListViewWithSubItemData(View.Details, false, false, false, 3, 2);
+
+            for (int i = 0; i < control.Columns.Count; i++)
```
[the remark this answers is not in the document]
Tanya-Solyanik: Also consider factoring out validation loop into a helper function.

Tanya-Solyanik (commented):
Looks good!

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItem.ListViewSubItem.ListViewSubItemAccessibleObjectTests.cs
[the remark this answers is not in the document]
RussKie: I'd like to caution against too much of optimisation of tests - there should be a balance between code optimisation and self-containment - i.e., it should be easy to lift a test and move it, ...or remove it, if the test is no longer applicable. 
When a test's implementation is spread over multiple methods, it often inhibits the ability to reason about the test. Also, not seldomly such organisation leads to helper methods turning in "jack of all trades" kinds of method catering for different tests. Last but not least, when test suites evolve and tests change and/or get removed, the helper methods remain in the codebase. And with tens of thousands of tests, it's humanly impossible to check.

dmitrii-drobotov (approved):
LGTM

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItem.ListViewSubItem.ListViewSubItemAccessibleObjectTests.cs
```diff
@@ -973,6 +973,86 @@ public void ListViewSubItemAccessibleObject_GetPropertyValue_Pattern_ReturnsExpe
...
+            for (int i = 0; i < columns.Count; i++)
+            {
+                var subItemAccessibleObject = (ListViewSubItemAccessibleObject)control.Items[0].SubItems[i].AccessibilityObject;
+                AccessibleObject columnHeaderAccessibleObject = columns[i].AccessibilityObject;
+                UiaCore.IRawElementProviderSimple[] columnHeaderItems = subItemAccessibleObject.GetColumnHeaderItems();
+
+                ValidateItem(columnHeaderAccessibleObject, columnHeaderItems);
+            }
+
+            void ValidateItem(AccessibleObject columnHeaderAccessibleObject, UiaCore.IRawElementProviderSimple[] actualColumnHeaderItems)
```
Tanya-Solyanik, line 997: Local functions do not improve readability. Previous version was at least shorter.
A common validation function between 2 tests indicates that with images and with no images, we get the same results. As the tests are written, each line of one test has to be matched against a line of another test to understand that the difference is a single line setting the images.

dkazennov (commented):
Looks good.
I believe that @Tanya-Solyanik has the point. ValidateItem() method can be separate and perhaps being reused later?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItem.ListViewSubItem.ListViewSubItemAccessibleObjectTests.cs
RussKie: > Local functions do not improve readability. Previous version was at least shorter.

I differ here, I don't see an issue with this.
Equally I don't see an issue with reusing `ValidateItem` method.

>  As the tests are written, each line of one test has to be matched against a line of another test to understand that the difference is a single line setting the images.

I agree. I'm having difficulties seeing what's materially different apart from using `imageCollection`. Return types appear the same, so does the validation logic....
What's the difference?

NikitaSemenovAkvelon: Setting the images is the most critical difference between these two methods. It was the reason of the bug. 
And I thought that the names of the methods described the difference:

> ListViewSubItemAccessibleObject_GetColumnHeaderItems_IsExpected
> ListViewSubItemAccessibleObject_GetColumnHeaderItems_IsExpected_**WithImage**

RussKie: Have you given any thoughts to combining the two under a theory, and updating the test signature to `ListViewSubItemAccessibleObject_GetColumnHeaderItems_IsExpected(bool itemHasImage)`?

NikitaSemenovAkvelon: Yes, I decided that it would be less readable. In that case, `itemHasImage` is not a simply `?:` expression.
But if you disagree, I can combinate these. It's not a problem.

RussKie: Do you think you could show a diff (\`\`\`diff) of a consolidated version?

NikitaSemenovAkvelon: It would be like this:
image

Tanya-Solyanik: >Yes, I decided that it would be less readable. In that case, itemHasImage is not a simply ?: expression.

If you want a one line expression, you can add a local function that creates imagelist and sets up the ListView to use it.

The fact that a test is a theory clearly shows that after initialization, control is exercised in the same way and results are the same. As a result, unit test clearly documents the expected behavior. When you have 2 facts, it's not obvious that they perform the same steps, or what is expected. 
 I would put all code after the initialization into a helper function:
```
test1
    custom initialization
    CommonHelper

test2
    custom initialization
    CommonHelper

CommonHelper
```

RussKie: > It would be like this:

To me _personally_ this looks great, this is how I would have structured the test, if I'd have written it:
* The theory indicates that the input argument is the key.
* The validator is test-specific and, thus, local.

Tanya-Solyanik (approved):

RussKie (approved):

PR #8049, opened 2022-10-27

gpetrou:
## Proposed changes

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8049)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
@@ -2,11 +2,10 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.Collections.Specialized;
 using System.ComponentModel;
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using System.Drawing;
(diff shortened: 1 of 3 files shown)


dreddy-work (approved):

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
```diff
@@ -522,7 +523,7 @@ public virtual ImageLayout BackgroundImageLayout
                     return ImageLayout.Tile;
                 }
 
-                return ((ImageLayout)Properties.GetObject(s_backgroundImageLayoutProperty));
+                return (ImageLayout)Properties.GetObject(s_backgroundImageLayoutProperty)!;
```
RussKie, line 526: How can we be sure this is never a null?

gpetrou: This is an enum, so since the value exists because we passed the previous if statement it should not be null. We have the same in Control class.

RussKie: Makes sense, thank you.

RussKie (approved):

PR #8054, opened 2022-10-27
Enable CodeQL in WinForms repo

dreddy-work:
- Enabling CodeQL for WinForms repo by adding new YML
- CodeQL is increasing the build time hence, scheduled to run weekly and comply with the policy.
- Will be adding new pipeline, new pipeline is going to run on this new YML, scheduled weekly, and separate from regular pipeline. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8054)

### DIFF
--- a/eng/common/native/CommonLibrary.psm1
+++ b/eng/common/native/CommonLibrary.psm1
@@ -170,7 +170,7 @@ function Get-File {
         $Attempt++
         if ($Attempt -Lt $DownloadRetries) {
           $AttemptsLeft = $DownloadRetries - $Attempt
-          Write-Warning "Download failed, $AttemptsLeft attempts remaining, will retry in $RetryWaitTimeInSeconds seconds"
+          Write-Warning "'$Path' does not exist, download `$Uri` to `$TempPath` failed, , $AttemptsLeft attempts remaining, will retry in $RetryWaitTimeInSeconds seconds"
           Start-Sleep -Seconds $RetryWaitTimeInSeconds
         }
         else {
--- a/eng/common/native/CommonLibrary.psm1
+++ b/eng/common/native/CommonLibrary.psm1
@@ -170,7 +170,7 @@ function Get-File {
         $Attempt++
         if ($Attempt -Lt $DownloadRetries) {
           $AttemptsLeft = $DownloadRetries - $Attempt
-          Write-Warning "'$Path' does not exist, download `$Uri` to `$TempPath` failed, , $AttemptsLeft attempts remaining, will retry in $RetryWaitTimeInSeconds seconds"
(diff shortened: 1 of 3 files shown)


eng/pipelines/azure-pipelines-codeql.yml
```diff
@@ -0,0 +1,58 @@
...
+trigger: none
+
+schedules:
+  - cron: 0 12 * * 1
+    displayName: Weekly Monday CodeQL run
+    branches:
+      include:
+      - main
+      - release/6.0
+      - release/7.0
```
RussKie, line 31: Would this work?
```suggestion
      - release/*
```

If the older release branches aren't needed, we should be able to exclude those via `exclude:` node

garath: The syntax is valid, but be aware there are currently 12 branches to exclude versus only 2 to include. It seems like more work to maintain an explicit exclude list.

dreddy-work: blanket release\* is always prone to unexpected runs if there is ever a test release branch ( we did create one recently) or something of sort. it would be easy to find something is not running than running more than needed. Keeping current set for now and revisit if needed.

eng/common/native/CommonLibrary.psm1
```diff
@@ -170,7 +170,7 @@ function Get-File {
         $Attempt++
         if ($Attempt -Lt $DownloadRetries) {
           $AttemptsLeft = $DownloadRetries - $Attempt
-          Write-Warning "Download failed, $AttemptsLeft attempts remaining, will retry in $RetryWaitTimeInSeconds seconds"
+          Write-Warning "'$Path' does not exist, download `$Uri` to `$TempPath` failed, , $AttemptsLeft attempts remaining, will retry in $RetryWaitTimeInSeconds seconds"
```
garath, line 173: Everything under the `eng/common/` directory will be overwritten whenever updated Arcade versions flow to the winforms repository. This change should instead be made as a PR to Arcade.

dreddy-work: Can you take care of it from Arcade side? if it doesn't create merge conflicts, i will leave it until that happen.

dreddy-work: On a second thought, I remove this to make it clean but please update this on Arcade. It would be useful.

dreddy-work:
Test pipeline succeed: https://dev.azure.com/dnceng/internal/_build/results?buildId=2031918&view=logs&j=eb2c8817-e906-5459-8080-1a4c8ed1e2e2

garath (approved):

PR #8060, opened 2022-10-28
Avoid excessive memory usage and slow `TrackBar` control 

### LINKED ISSUE #329: Setting the Maximum value on a Trackbar to a very large number results in excessive memory usage
* .NET Core Version: 3.0 Preview1

**Problem description:**  
Setting the MaxValue on the Trackbar control to a very high number results in excessive ram usage (in the gigabytes for a very large number).  

This appears to be because on initialization of the underlying windows control, the maximum is set before the tick mark frequency, resulting in it initializing excessive amounts of tickmarks, although I do not have the build configuration to verify this theory myself at this time as if I did, I would just verify and submit this as a PR.
https://github.com/dotnet/winforms/blob/01eef03d8e0a6630076716b6cb08ed0a9def611a/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs#L893-L894

**Actual behavior:** 
Excessive ram is used (multiple gigabytes if compiled in x64)

**Expected behavior:**
A reasonable amount of ram is used (assuming you set tickmarks to a reasonable number).

**Minimal repro:**  
Within a form
```
TrackBar tb = new TrackBar();
tb.TickFrequency = 0; //Ensure no tickmarks should be drawn
tb.Maximum = 1000000000;
this.Controls.Add(tb);
```

conorgee:
Allow users to set large a range of values on the trackbar without excessive memory consumption.

Fixes #329

## Proposed changes
- Solving the issue of Memory consumption for the trackbar for large range of values on trackbar.
- Turning `TBS_AUTOTICKS` off when the the tick range is larger then the physical limitations of the trackbar by checking this with the Boolean method `ShouldAutoDrawTicks()`. This sets the flag `_autoDrawTicks` in `CreateParams`.
- The fix will be put behind a feature switch called `TrackBarModernRendering` in `LocalAppContextSwitches ` because of the significant risk - primarily due to the use of `RecreateHandle()` method. If they opt out of the feature the `ShouldAutoDrawTicks()` then it will always return true. This means that the way of rendering ticks will be reverted back to the old way e.g. using `TBS_AUTOTICKS` and setting tick frequency with the `TBM_SETTICFREQ` message.

   ```C#
        /// <summary>
        /// This checks all the use cases that we potentially might want to keep `TBS_AUTOTICKS`.
        /// </summary>
        private bool ShouldAutoDrawTicks()
        {
            // If the user decides to opt out of TrackBarModernRendering for drawing ticks,
            // by returning true autoticks will be set to true which will
            // use the old way of rendering ticks on the trackbar.
            // TBS_AUTOTICKS will be enabled and sending
            // the TBM_SETTICFREQ message to set tick frequency.
            if (!LocalAppContextSwitches.TrackBarModernRendering)
            {
                return true;
            }

            if (TickStyle == TickStyle.None)
            {
                return true;
            }

            int size = Orientation == Orientation.Horizontal ? Size.Width : Size.Height;
            if (size == 0)
            {
                return true;
            }

            uint range = (uint)(_maximum - _minimum);
            return range <= (size / 2);
        }
    ```
- The method   `DrawTicks()` replaces the message `TBM_SETTICFREQ` It gets called everywhere  `TBM_SETTICFREQ`  would have been used.
- It checks whether the tickstyle is None (return) or `_autoDrawTicks` is true which sets the `_tickFrequency` using the  `TBM_SETTICFREQ` message then return .
- When the tick range is greater then physical limitaions of the size of the trackbar it calculates if the users set `_tickFrequency` will fit on the trackbar.
- If the value is greater then the physical limiations we have then we set `drawnTickFrequency ` to `range / maxTickCount`.
- Using a for loop we maually draw the ticks based on the  `drawnTickFrequency ` set.
 ```C#
        private void DrawTicks()
        {
            if (_tickStyle == TickStyle.None)
            {
                return;
            }

            int drawnTickFrequency = _tickFrequency;
            // Will be true if they opt out TrackBarModernRendering.
            if (_autoDrawTicks)
            {
                PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETTICFREQ, (WPARAM)drawnTickFrequency);
                return;
            }

            // Divide by 2 because otherwise the ticks appear as a solid line.
            int maxTickCount = (Orientation == Orientation.Horizontal ? Size.Width : Size.Height) / 2;
            uint range = (uint)(_maximum - _minimum);
            if (range > maxTickCount && maxTickCount != 0)
            {
                int calculatedTickFrequency = (int)(range / maxTickCount);
                if (calculatedTickFrequency > drawnTickFrequency)
                {
                    drawnTickFrequency = calculatedTickFrequency;
                }
            }

            PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_CLEARTICS, (WPARAM)1, (LPARAM)0);
            for (int i = _minimum + drawnTickFrequency; i < _maximum - drawnTickFrequency; i += drawnTickFrequency)
            {
                LRESULT lresult = PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETTIC, lParam: (IntPtr)i);
                Debug.Assert((bool)(BOOL)lresult);
            }
        }
```

- Before Recreating Handle check to see If it needs to be recreated by using `ShouldRecreateHandle()` to set the `recreateHandle` flag in places where  `DrawTicks()` gets called

 ```C#
        /// <summary>
        /// Determine if the decision of whether the ticks drawing,
        /// is performed by the native control or the Windows Forms runtime
        /// is still valid. If it's no longer valid, we'll need to recreate the native control.
        /// If user opts out of TrackBarModernRendering then this will always return false.
        /// </summary>
        private bool ShouldRecreateHandle() => IsHandleCreated && _autoDrawTicks != ShouldAutoDrawTicks();
 ```

- `DrawTicks()` Is refrenced in three places first is when the user wants to change the `TickFrequency` and the second place is `SetRange()` when max or min is changed and `OnHandleCreated()` .
- If `recreateHandle` is true then  `DrawTicks()` will not be called in either of the first two places it will be called in  `OnHandleCreated()`

 ```C#
        /// <summary>
        ///  Indicates just how many ticks will be drawn. For a TrackBar with a
        ///  range of 0..100, it might be impractical to draw all 100 ticks for a
        ///  very small control. Passing in a value of 5 here would only draw
        ///  20 ticks -- i.e. each tick would represent 5 units in the TrackBars
        ///  range of values.
        /// </summary>
       public int TickFrequency
        {
            ....

                // Determine if the decision of whether the ticks drawing,
                // is performed by the native control or the Windows Forms runtime
                // is still valid. If it's no longer valid, we'll need to recreate the native control.
                bool recreateHandle = ShouldRecreateHandle();
                if (IsHandleCreated)
                {
                    PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETTICFREQ, (WPARAM)value);
                    // If user opts out of TrackBarModernRendering then recreateHandle
                    // will always be false.
                    if (recreateHandle)
                    {
                        RecreateHandle();
                    }
                    else
                    {
                        DrawTicks();
                        Invalidate();
                    }
                }
            }
        }
```

 ```C#
        /// <summary>
        ///  Lets you set the entire range for the TrackBar control at once.
        ///  The values passed are both the lower and upper limits to the range
        ///  with which the control will work.
        /// </summary>
        public void SetRange(int minValue, int maxValue)
        {
            ....

                // Determine if the decision of whether the ticks drawing,
                // is performed by the native control or the Windows Forms runtime
                // is still valid. If it's no longer valid, we'll need to recreate the native control.
                bool recreateHandle = ShouldRecreateHandle();
                // If user opts out of TrackBarModernRendering then recreateHandle
                // will always be false.
                if (IsHandleCreated && !recreateHandle)
                {
                    PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMIN, (WPARAM)(BOOL)false, (LPARAM)_minimum);
                    PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMAX, (WPARAM)(BOOL)true, (LPARAM)_maximum);
                    DrawTicks();
                    Invalidate();
                }

               ....

                // If user opts out of TrackBarModernRendering then recreateHandle
                // will always be false.
                if (recreateHandle)
                {
                    RecreateHandle();
                }
                else
                {
                    SetTrackBarPosition();
                }
            }
        }
```

- If the trackbar has been created or `RecreateHandle()` is excuted then we excute `DrawTicks();` here

```C#
    protected override void OnHandleCreated(EventArgs e)
        {
            base.OnHandleCreated(e);

            if (!IsHandleCreated)
            {
                return;
            }

            Debug.Assert(_autoDrawTicks == ShouldAutoDrawTicks());
            PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMIN, (WPARAM)(BOOL)false, (LPARAM)_minimum);
            PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMAX, (WPARAM)(BOOL)false, (LPARAM)_maximum);
            DrawTicks();
            PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETPAGESIZE, (WPARAM)0, (LPARAM)_largeChange);
            PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETLINESIZE, (WPARAM)0, (LPARAM)_smallChange);
            SetTrackBarPosition();
            AdjustSize();
        }
```

## Customer Impact

- Solves the issue of ticks slowing down performance via large memory consumption.

## Regression? 

-  No.

## Risk

- Significant risk - primarily due to the use of `RecreateHandle()` method.

### Before

image

### After

image

### Test Feature Switch 
- With `TrackBarModernRendering` set to false

image

- With `TrackBarModernRendering` set to true

image

### Test  int.MinValue 

- For min int value – trackbar breaks and is unuseable seems to be issue with the trackbar range cannot be greater then the positive value for an integer e.g. -2,147,483,648 would create a range larger then 2,147,483,647 . 
-  This is for both horizontal and vertical orientation
- set a trackbar width to 588px, orientation horizontal, min=int.MinValue, max=0, tick freq=1 -> expect memory consumption to be low. 0  ticks drawn on trackbar. Autoticks disabled

image

image

- set a trackbar width to 364px, orientation vertical, min=-int.MinValue, max=0, tick freq=1 -> expect memory consumption to be low. 0 ticks drawn on trackbar. Autoticks disabled

image
### Test int.MaxValue
-  For max int value – trackbar works for both horizontal and vertical orientation
- set a trackbar width to 588px, orientation horizontal, min=0, max=int.MaxValue, tick freq=1 -> expect memory consumption to be low. 294 ticks drawn on trackbar. Autoticks disabled

image

- set a trackbar width to 364px, orientation vertical, min=0, max=int.MaxValue, tick freq=1 -> expect memory consumption to be low. 182 ticks drawn on trackbar. Autoticks disabled

image

### Basic Tests
- Testing for a large value range to turn off  `TBS_AUTOTICKS`  then setting to a small range for  `TBS_AUTOTICKS`  to be turned on , works for both `horizontal `and `vertical `orientation.

- set a trackbar width to 588px, orientation horizontal, min=-100, max=1000, tick freq=1 -> expect memory consumption to be low e.g 20mb. expect 294 ticks, autoTicks disabled ticks manually drawn 

image

- set a trackbar width to 588px, orientation horizontal, min=-100, max=100, tick freq=1 -> expect memory consumption to be low e.g 20mb. expect 200 ticks, autoTicks enabled 

image

- set a trackbar width to 588px, orientation horizontal, min=-100, max=100, tick freq=1 , Tickstyle set to none, -> expect memory consumption to be low e.g 20mb. expect 0 ticks, autoTicks disabled

image

### Trackbar Size tests
- works the same for both `horizontal `and `vertical `orientation.
- set a trackbar width to 0px, orientation vertical, min=-100, max=100, tick freq=10 -> expect memory consumption to be low e.g 20mb.expect 0 ticks, autoTicks disabled
image 
- set a trackbar width to 1,000px, orientation horizontal, min=-100, max=100, tick freq=1 -> expect memory consumption to be low e.g 20mb. expect 200 ticks, autoTicks enabled

image

- set a trackbar width to 1,000px, orientation horizontal, min=-1000, max=1000, tick freq=1 -> expect memory consumption to be low e.g 20mb. expect 500 ticks, autoTicks disabled
image

- set a trackbar width to 1,000px, orientation horizontal, min=-1000, max=1000, tick freq=100 -> expect memory consumption to be low e.g 20mb. expect 20 ticks, autoTicks disabled
image

### Tickfrequency tests
- works the same for both `horizontal `and `vertical `orientation.

- set a trackbar width to 588px, orientation horizontal, min=-100, max=100, tick freq=100 -> expect memory consumption to be low e.g 20mb. expect 3 ticks, autoTicks enabled

image

- set a trackbar width to 364px, orientation vertical, min=-10000, max=10000, tick freq=1000 -> expect memory consumption to be low e.g 20mb. expect 20 ticks, autoTicks disabled 

image

- TODO: Add unit tests.
- TODO: Add UIIntegration tests.

## Accessibility testing  
TODO

 

## Test environment(s) 

.NET 8.0.0 - alpha.1.22605.1

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8060)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
@@ -868,7 +868,7 @@ namespace System.Windows.Forms
             int maxTickCount = (Orientation == Orientation.Horizontal ? Size.Width : Size.Height) / 2;
             int ticksDrawn;
             uint range = (uint)(_maximum - _minimum);
-            if (_tickFrequency > 1)
+            if (_tickFrequency!=1)
             {
                ticksDrawn = (int)(range / _tickFrequency);
                 if(ticksDrawn<maxTickCount)
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
@@ -39,8 +39,7 @@ namespace System.Windows.Forms
         private int _tickFrequency = 1;
         private TickStyle _tickStyle = TickStyle.BottomRight;
         private int _requestedDim;
-        private bool _AutoTicksRecreate;
-
+        private bool _autoDrawTicks;
         // Mouse wheel movement
         private int _cumulativeWheelData;
 
@@ -186,8 +185,10 @@ namespace System.Windows.Forms
 
                 void EnableAutoTicksIfRequired()
                 {
-                     if (AutoDrawTicks)
+                    _autoDrawTicks = false;
+                    if (ShouldAutoDrawTicks)
                      {
+                        _autoDrawTicks = true;
                         cp.Style |= (int)PInvoke.TBS_AUTOTICKS;
                      }
                 }
@@ -567,10 +568,20 @@ namespace System.Windows.Forms
                 }
 
                 _tickFrequency = value;
-                if (IsHandleCreated)
+                bool recreateHandle = false;
+                if (IsHandleCreated && _autoDrawTicks != ShouldAutoDrawTicks)
                 {
-                    CreateTicks();
+                    recreateHandle = true;
+                }
+
+                if (!recreateHandle && IsHandleCreated)
+                {
+                    DrawTicks();
                     Invalidate();
+                }
+                else
+                {
+                    RecreateHandle();
                 }
             }
         }
@@ -817,18 +828,17 @@ namespace System.Windows.Forms
                 return;
             }
 
-            _AutoTicksRecreate = true;
+            Debug.Assert(_autoDrawTicks == ShouldAutoDrawTicks);
             PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMIN, (WPARAM)(BOOL)false, (LPARAM)_minimum);
             PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMAX, (WPARAM)(BOOL)false, (LPARAM)_maximum);
-            CreateTicks();
+            DrawTicks();
             PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETPAGESIZE, (WPARAM)0, (LPARAM)_largeChange);
             PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETLINESIZE, (WPARAM)0, (LPARAM)_smallChange);
             SetTrackBarPosition();
             AdjustSize();
-            _AutoTicksRecreate = false;
         }
 
-        private bool AutoDrawTicks
+        private bool ShouldAutoDrawTicks
         {
             get
             {
@@ -848,37 +858,23 @@ namespace System.Windows.Forms
             }
         }
 
-        private void CreateTicks()
+        private void DrawTicks()
         {
             // Check if the value of the max is greater then the taskbar size.
             // If so then we divide the value by size and only that many ticks to be drawn on the screen.
 
-            if (!_AutoTicksRecreate)
-            {
-                RecreateHandle();
-            }
-
-            if (AutoDrawTicks)
+            if (_autoDrawTicks)
             {
                 PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETTICFREQ, (WPARAM)_tickFrequency);
                 return;
             }
 
-            bool tickFreqFine = true;
             int maxTickCount = (Orientation == Orientation.Horizontal ? Size.Width : Size.Height) / 2;
-            int ticksDrawn;
             uint range = (uint)(_maximum - _minimum);
-            if (_tickFrequency!=1)
-            {
-               ticksDrawn = (int)(range / _tickFrequency);
-                if(ticksDrawn<maxTickCount)
-                {
-                    tickFreqFine = false;
-                }
-            }
-
             int ticksDrawnFrequency = _tickFrequency;
-            if (range > maxTickCount && maxTickCount != 0 && tickFreqFine)
+            int ticksDrawn = (int)(range / ticksDrawnFrequency);
+
+            if (maxTickCount != 0 && ticksDrawn > maxTickCount)
             {
                 ticksDrawnFrequency = (int)(range / maxTickCount);
             }
@@ -1069,8 +1065,21 @@ namespace System.Windows.Forms
                 {
                     PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMIN, (WPARAM)(BOOL)false, (LPARAM)_minimum);
                     PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMAX, (WPARAM)(BOOL)true, (LPARAM)_maximum);
-                    CreateTicks();
-                    Invalidate();
+                    bool recreateHandle = false;
+                    if (_autoDrawTicks != ShouldAutoDrawTicks)
+                    {
+                        recreateHandle = true;
+                    }
+
+                    if (!recreateHandle && IsHandleCreated)
+                    {
+                        DrawTicks();
+                        Invalidate();
+                    }
+                    else
+                    {
+                        RecreateHandle();
+                    }
                 }
 
                 // When we change the range, the comctl32 trackbar's internal position can change
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
@@ -158,27 +158,15 @@ namespace System.Windows.Forms
                         break;
                     case TickStyle.TopLeft:
                         cp.Style |= (int)(PInvoke.TBS_TOP);
-                        if (_autoTicks)
-                        {
-                            cp.Style |= (int)PInvoke.TBS_AUTOTICKS;
-                        }
-
+                        EnableAutoTicksIfRequired();
                         break;
                     case TickStyle.BottomRight:
                         cp.Style |= (int)(PInvoke.TBS_BOTTOM);
-                        if (_autoTicks)
-                        {
-                            cp.Style |= (int)PInvoke.TBS_AUTOTICKS;
-                        }
-
+                        EnableAutoTicksIfRequired();
                         break;
                     case TickStyle.Both:
                         cp.Style |= (int)(PInvoke.TBS_BOTH);
-                        if (_autoTicks)
-                        {
-                            cp.Style |= (int)PInvoke.TBS_AUTOTICKS;
-                        }
-
+                        EnableAutoTicksIfRequired();
                         break;
                 }
 
@@ -196,6 +184,14 @@ namespace System.Windows.Forms
                 }
 
                 return cp;
+
+                void EnableAutoTicksIfRequired()
+                {
+                     if (_autoTicks)
+                     {
+                        cp.Style |= (int)PInvoke.TBS_AUTOTICKS;
+                     }
+                }
             }
         }
 
@@ -831,30 +827,47 @@ namespace System.Windows.Forms
             AdjustSize();
         }
 
+        private bool AutoDrawTicks
+        {
+            get
+            {
+                if (TickStyle == TickStyle.None)
+                {
+                    return true;
+                }
+
+                int tickFrequency = _tickFrequency;
+                int maxTickCount = (Orientation == Orientation.Horizontal ? Size.Width : Size.Height) / 2;
+                int range = _maximum - _minimum;
+
+                return range <= maxTickCount || maxTickCount == 0;
+            }
+        }
+
         private void SetTickFrequency()
         {
-            // check if the value of the max is greater then the taskbar size
-            // if so then we divide the value by size and only that many ticks to be drawn on the screen
+            // Check if the value of the max is greater then the taskbar size.
+            // If so then we divide the value by size and only that many ticks to be drawn on the screen.
 
-            if (TickStyle == System.Windows.Forms.TickStyle.None)
+            if (TickStyle == TickStyle.None || AutoDrawTicks)
             {
                 return;
             }
 
             int tickFrequency = _tickFrequency;
-            int trackbarSize = (Orientation == Orientation.Horizontal ? Size.Width : Size.Height) / 2;
-            uint maxValue = (uint)(-_minimum+_maximum);
+            int maxTickCount = (Orientation == Orientation.Horizontal ? Size.Width : Size.Height) / 2;
+            int range = _maximum - _minimum;
 
-            if (maxValue > trackbarSize && trackbarSize != 0)
+            if (range > maxTickCount && maxTickCount != 0)
             {
-                tickFrequency = ((int)(maxValue / trackbarSize));
-                _autoTicks = false;
+                tickFrequency = range / maxTickCount;
             }
 
             PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_CLEARTICS, (WPARAM)1, (LPARAM)0);
-            for (int i = tickFrequency; i < maxValue; i += tickFrequency)
+            for (int i = _minimum + tickFrequency; i < _maximum - tickFrequency; i += tickFrequency)
             {
-                PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETTIC, lParam: (LPARAM)i);
+                LRESULT lresult = PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETTIC, lParam: (IntPtr)i);
+                Debug.Assert((bool)(BOOL)lresult);
             }
         }
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
@@ -29,7 +29,7 @@ namespace System.Windows.Forms
         private static readonly object s_scrollEvent = new object();
         private static readonly object s_valueChangedEvent = new object();
         private static readonly object s_rightToLeftChangedEvent = new object();
-
+        private bool _autoTicks = true;
         private bool _autoSize = true;
         private int _largeChange = 5;
         private int _maximum = 10;
@@ -158,12 +158,27 @@ namespace System.Windows.Forms
                         break;
                     case TickStyle.TopLeft:
                         cp.Style |= (int)(PInvoke.TBS_TOP);
+                        if (_autoTicks)
+                        {
+                            cp.Style |= (int)PInvoke.TBS_AUTOTICKS;
+                        }
+
                         break;
                     case TickStyle.BottomRight:
                         cp.Style |= (int)(PInvoke.TBS_BOTTOM);
+                        if (_autoTicks)
+                        {
+                            cp.Style |= (int)PInvoke.TBS_AUTOTICKS;
+                        }
+
                         break;
                     case TickStyle.Both:
                         cp.Style |= (int)(PInvoke.TBS_BOTH);
+                        if (_autoTicks)
+                        {
+                            cp.Style |= (int)PInvoke.TBS_AUTOTICKS;
+                        }
+
                         break;
                 }
 
@@ -807,9 +822,9 @@ namespace System.Windows.Forms
                 return;
             }
 
-            SetTickFrequency();
             PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMIN, (WPARAM)(BOOL)false, (LPARAM)_minimum);
             PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMAX, (WPARAM)(BOOL)false, (LPARAM)_maximum);
+            SetTickFrequency();
             PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETPAGESIZE, (WPARAM)0, (LPARAM)_largeChange);
             PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETLINESIZE, (WPARAM)0, (LPARAM)_smallChange);
             SetTrackBarPosition();
@@ -828,11 +843,12 @@ namespace System.Windows.Forms
 
             int tickFrequency = _tickFrequency;
             int trackbarSize = (Orientation == Orientation.Horizontal ? Size.Width : Size.Height) / 2;
-            uint maxValue = (uint)(_minimum + _maximum);
+            uint maxValue = (uint)(System.Math.Abs(_minimum) + _maximum);
 
             if (maxValue > trackbarSize && trackbarSize != 0)
             {
                 tickFrequency = ((int)(maxValue / trackbarSize));
+                _autoTicks = false;
             }
 
             PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_CLEARTICS, (WPARAM)1, (LPARAM)0);
@@ -1018,9 +1034,9 @@ namespace System.Windows.Forms
 
                 if (IsHandleCreated)
                 {
-                    SetTickFrequency();
                     PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMIN, (WPARAM)(BOOL)false, (LPARAM)_minimum);
                     PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMAX, (WPARAM)(BOOL)true, (LPARAM)_maximum);
+                    SetTickFrequency();
                     Invalidate();
                 }
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
@@ -853,26 +853,38 @@ namespace System.Windows.Forms
             // Check if the value of the max is greater then the taskbar size.
             // If so then we divide the value by size and only that many ticks to be drawn on the screen.
 
-            if (AutoDrawTicks)
-            {
-                return;
-            }
-
-            if (_AutoTicksRecreate!)
+            if (!_AutoTicksRecreate)
             {
                 RecreateHandle();
             }
 
-            int maxTickCount = (Orientation == Orientation.Horizontal ? Size.Width : Size.Height) / 2;
-            uint range = (uint)(_maximum - _minimum);
-
-            if (range > maxTickCount && maxTickCount != 0)
+            if (AutoDrawTicks)
             {
-                _tickFrequency = (int)(range / maxTickCount);
+                PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETTICFREQ, (WPARAM)_tickFrequency);
+                return;
+            }
+
+            bool tickFreqFine = true;
+            int maxTickCount = (Orientation == Orientation.Horizontal ? Size.Width : Size.Height) / 2;
+            int ticksDrawn;
+            uint range = (uint)(_maximum - _minimum);
+            if (_tickFrequency > 1)
+            {
+               ticksDrawn = (int)(range / _tickFrequency);
+                if(ticksDrawn<maxTickCount)
+                {
+                    tickFreqFine = false;
+                }
+            }
+
+            int ticksDrawnFrequency = _tickFrequency;
+            if (range > maxTickCount && maxTickCount != 0 && tickFreqFine)
+            {
+                ticksDrawnFrequency = (int)(range / maxTickCount);
             }
 
             PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_CLEARTICS, (WPARAM)1, (LPARAM)0);
-            for (int i = (_minimum + _tickFrequency); i < _maximum - _tickFrequency; i += _tickFrequency)
+            for (int i = (_minimum + ticksDrawnFrequency); i < _maximum - ticksDrawnFrequency; i += ticksDrawnFrequency)
             {
                 LRESULT lresult = PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETTIC, lParam: (IntPtr)i);
                 Debug.Assert((bool)(BOOL)lresult);
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
@@ -148,8 +148,12 @@ namespace System.Windows.Forms
             {
                 CreateParams cp = base.CreateParams;
                 cp.ClassName = PInvoke.TRACKBAR_CLASS;
+                if (ShouldAutoDrawTicks())
+                {
+                    _autoDrawTicks = true;
+                }
 
-                switch (_tickStyle)
+                    switch (_tickStyle)
                 {
                     case TickStyle.None:
                         cp.Style |= (int)PInvoke.TBS_NOTICKS;
@@ -853,14 +857,19 @@ namespace System.Windows.Forms
 
         uint range = (uint)(_maximum - _minimum);
         return range <= (size / 2);
-
         }
 
-
+        /// <summary>
+        ///  Check if the value of the max is greater then the taskbar size.
+        ///  If so then we divide the value by size and only that many ticks to be drawn on the screen.
+        /// </summary>
         private void DrawTicks()
         {
-            // Check if the value of the max is greater then the taskbar size.
-            // If so then we divide the value by size and only that many ticks to be drawn on the screen.
+            if (_tickStyle == TickStyle.None)
+            {
+                return;
+            }
+
             int ticksDrawnFrequency = _tickFrequency;
             if (_autoDrawTicks)
             {
@@ -871,7 +880,6 @@ namespace System.Windows.Forms
             int maxTickCount = (Orientation == Orientation.Horizontal ? Size.Width : Size.Height) / 2;
             uint range = (uint)(_maximum - _minimum);
             int ticksDrawn = 1;
-
             if (ticksDrawnFrequency != 0)
             {
                 ticksDrawn = (int)(range / ticksDrawnFrequency);
@@ -883,7 +891,7 @@ namespace System.Windows.Forms
             }
 
             PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_CLEARTICS, (WPARAM)1, (LPARAM)0);
-            for (int i = (_minimum + ticksDrawnFrequency); i < _maximum - ticksDrawnFrequency; i += ticksDrawnFrequency)
+            for (int i = _minimum + ticksDrawnFrequency; i < _maximum - ticksDrawnFrequency; i += ticksDrawnFrequency)
             {
                 LRESULT lresult = PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETTIC, lParam: (IntPtr)i);
                 Debug.Assert((bool)(BOOL)lresult);
@@ -1072,15 +1080,17 @@ namespace System.Windows.Forms
                         recreateHandle = true;
                     }
 
-                    if (recreateHandle)
+                    if (!recreateHandle)
+                    {
+                        PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMIN, (WPARAM)(BOOL)false, (LPARAM)_minimum);
+                        PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMAX, (WPARAM)(BOOL)true, (LPARAM)_maximum);
+                        DrawTicks();
+                        Invalidate();
+                    }
+                    else
                     {
                         RecreateHandle();
                     }
-
-                    PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMIN, (WPARAM)(BOOL)false, (LPARAM)_minimum);
-                    PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMAX, (WPARAM)(BOOL)true, (LPARAM)_maximum);
-                    DrawTicks();
-                    Invalidate();
                 }
 
                 // When we change the range, the comctl32 trackbar's internal position can change
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
@@ -147,10 +147,8 @@ namespace System.Windows.Forms
             get
             {
                 _autoDrawTicks = ShouldAutoDrawTicks();
-
                 CreateParams cp = base.CreateParams;
                 cp.ClassName = PInvoke.TRACKBAR_CLASS;
-
                 switch (_tickStyle)
                 {
                     case TickStyle.None:
@@ -568,6 +566,11 @@ namespace System.Windows.Forms
                 }
 
                 _tickFrequency = value;
+                // set the recreateHandle flag to null,
+                // if the _autoDrawTicks flag does not equal what ShouldAutoDrawTicks()
+                // we set the recreateHandle to true,
+                // then we skip sending the messages and,
+                // recreateHandle will be called at the end of the method.
                 bool recreateHandle = false;
                 if (IsHandleCreated && _autoDrawTicks != ShouldAutoDrawTicks())
                 {
@@ -765,6 +768,10 @@ namespace System.Windows.Forms
             base.CreateHandle();
         }
 
+        /// <summary>
+        ///  Check if the value of the max is greater then the taskbar size.
+        ///  If so then we divide the value by size and only that many ticks to be drawn on the screen.
+        /// </summary>
         private void DrawTicks()
         {
             if (_tickStyle == TickStyle.None)
@@ -870,11 +877,6 @@ namespace System.Windows.Forms
             SetTrackBarPosition();
             AdjustSize();
         }
-
-        /// <summary>
-        ///  Check if the value of the max is greater then the taskbar size.
-        ///  If so then we divide the value by size and only that many ticks to be drawn on the screen.
-        /// </summary>
 
         [EditorBrowsable(EditorBrowsableState.Advanced)]
         protected virtual void OnRightToLeftLayoutChanged(EventArgs e)
@@ -1049,7 +1051,11 @@ namespace System.Windows.Forms
 
                 _minimum = minValue;
                 _maximum = maxValue;
-
+                // set the recreateHandle flag to null,
+                // if the _autoDrawTicks flag does not equal what ShouldAutoDrawTicks()
+                // we set the recreateHandle to true,
+                // then we skip sending the messages and,
+                // recreateHandle will be called at the end of the method,
                 bool recreateHandle = false;
                 if (_autoDrawTicks != ShouldAutoDrawTicks())
                 {
@@ -1115,6 +1121,9 @@ namespace System.Windows.Forms
             }
         }
 
+        /// <summary>
+        /// This checks all the use cases that we potentially might want to keep `TBS_AUTOTICKS`.
+        /// </summary>
         private bool ShouldAutoDrawTicks()
         {
             if (TickStyle == TickStyle.None)
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
@@ -856,21 +856,26 @@ namespace System.Windows.Forms
 
         }
 
+
         private void DrawTicks()
         {
             // Check if the value of the max is greater then the taskbar size.
             // If so then we divide the value by size and only that many ticks to be drawn on the screen.
-
+            int ticksDrawnFrequency = _tickFrequency;
             if (_autoDrawTicks)
             {
-                PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETTICFREQ, (WPARAM)_tickFrequency);
+                PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETTICFREQ, (WPARAM)ticksDrawnFrequency);
                 return;
             }
 
             int maxTickCount = (Orientation == Orientation.Horizontal ? Size.Width : Size.Height) / 2;
             uint range = (uint)(_maximum - _minimum);
-            int ticksDrawnFrequency = _tickFrequency;
-            int ticksDrawn = (int)(range / ticksDrawnFrequency);
+            int ticksDrawn=1;
+
+            if (ticksDrawnFrequency != 0)
+            {
+                ticksDrawn = (int)(range / ticksDrawnFrequency);
+            }
 
             if (maxTickCount != 0 && ticksDrawn > maxTickCount)
             {
@@ -1061,23 +1066,21 @@ namespace System.Windows.Forms
 
                 if (IsHandleCreated)
                 {
-                    PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMIN, (WPARAM)(BOOL)false, (LPARAM)_minimum);
-                    PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMAX, (WPARAM)(BOOL)true, (LPARAM)_maximum);
                     bool recreateHandle = false;
                     if (_autoDrawTicks != ShouldAutoDrawTicks())
                     {
                         recreateHandle = true;
                     }
 
-                    if (!recreateHandle && IsHandleCreated)
-                    {
-                        DrawTicks();
-                        Invalidate();
-                    }
-                    else
+                    if (recreateHandle)
                     {
                         RecreateHandle();
                     }
+
+                    PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMIN, (WPARAM)(BOOL)false, (LPARAM)_minimum);
+                    PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMAX, (WPARAM)(BOOL)true, (LPARAM)_maximum);
+                    DrawTicks();
+                    Invalidate();
                 }
 
                 // When we change the range, the comctl32 trackbar's internal position can change
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
@@ -38,8 +38,8 @@ namespace System.Windows.Forms
         private int _smallChange = 1;
         private int _tickFrequency = 1;
         private TickStyle _tickStyle = TickStyle.BottomRight;
-
         private int _requestedDim;
+        private bool _AutoTicksRecreate;
 
         // Mouse wheel movement
         private int _cumulativeWheelData;
@@ -569,7 +569,7 @@ namespace System.Windows.Forms
                 _tickFrequency = value;
                 if (IsHandleCreated)
                 {
-                    SetTickFrequency();
+                    CreateTicks();
                     Invalidate();
                 }
             }
@@ -817,13 +817,15 @@ namespace System.Windows.Forms
                 return;
             }
 
+            _AutoTicksRecreate = true;
             PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMIN, (WPARAM)(BOOL)false, (LPARAM)_minimum);
             PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMAX, (WPARAM)(BOOL)false, (LPARAM)_maximum);
-            SetTickFrequency();
+            CreateTicks();
             PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETPAGESIZE, (WPARAM)0, (LPARAM)_largeChange);
             PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETLINESIZE, (WPARAM)0, (LPARAM)_smallChange);
             SetTrackBarPosition();
             AdjustSize();
+            _AutoTicksRecreate = false;
         }
 
         private bool AutoDrawTicks
@@ -835,34 +837,42 @@ namespace System.Windows.Forms
                     return true;
                 }
 
-                int tickFrequency = _tickFrequency;
-                int maxTickCount = (Orientation == Orientation.Horizontal ? Size.Width : Size.Height) / 2;
-                uint range = (uint)(_maximum - _minimum);
+                int size = Orientation == Orientation.Horizontal ? Size.Width : Size.Height;
+                if (size == 0)
+                {
+                    return true;
+                }
 
-                return range <= maxTickCount || maxTickCount == 0;
+                uint range = (uint)(_maximum - _minimum);
+                return range <= (size / 2);
             }
         }
 
-        private void SetTickFrequency()
+        private void CreateTicks()
         {
             // Check if the value of the max is greater then the taskbar size.
             // If so then we divide the value by size and only that many ticks to be drawn on the screen.
 
-            if (TickStyle == TickStyle.None || AutoDrawTicks)
+            if (AutoDrawTicks)
             {
                 return;
             }
 
-            int tickFrequency = _tickFrequency;
+            if (_AutoTicksRecreate!)
+            {
+                RecreateHandle();
+            }
+
             int maxTickCount = (Orientation == Orientation.Horizontal ? Size.Width : Size.Height) / 2;
             uint range = (uint)(_maximum - _minimum);
+
             if (range > maxTickCount && maxTickCount != 0)
             {
-                tickFrequency = (int)(range / maxTickCount);
+                _tickFrequency = (int)(range / maxTickCount);
             }
 
             PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_CLEARTICS, (WPARAM)1, (LPARAM)0);
-            for (int i = (_minimum + tickFrequency); i < _maximum - tickFrequency; i += tickFrequency)
+            for (int i = (_minimum + _tickFrequency); i < _maximum - _tickFrequency; i += _tickFrequency)
             {
                 LRESULT lresult = PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETTIC, lParam: (IntPtr)i);
                 Debug.Assert((bool)(BOOL)lresult);
@@ -1047,7 +1057,7 @@ namespace System.Windows.Forms
                 {
                     PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMIN, (WPARAM)(BOOL)false, (LPARAM)_minimum);
                     PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMAX, (WPARAM)(BOOL)true, (LPARAM)_maximum);
-                    SetTickFrequency();
+                    CreateTicks();
                     Invalidate();
                 }
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
@@ -7,7 +7,6 @@ using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Drawing;
 using System.Windows.Forms.Layout;
-using static System.Windows.Forms.VisualStyles.VisualStyleElement;
 using static Interop;
 using static Interop.ComCtl32;
 
@@ -308,7 +307,6 @@ namespace System.Windows.Forms
                 }
 
                 SetRange(_minimum, value);
-                SetTickFrequency();
             }
         }
 
@@ -335,7 +333,6 @@ namespace System.Windows.Forms
                 }
 
                 SetRange(value, _maximum);
-                SetTickFrequency();
             }
         }
 
@@ -539,8 +536,6 @@ namespace System.Windows.Forms
             }
         }
 
-
-
         /// <summary>
         ///  Indicates just how many ticks will be drawn. For a TrackBar with a
         ///  range of 0..100, it might be impractical to draw all 100 ticks for a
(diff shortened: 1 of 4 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/winforms?pullRequest=8060) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/winforms?pullRequest=8060) before we can accept your contribution.<br/><br/>:x: conorgee [sign now](https://cla.dotnetfoundation.org/dotnet/winforms?pullRequest=8060)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/winforms?pullRequest=8060) it.</sub>

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/TrackBars.cs
```diff
@@ -3,11 +3,17 @@
...
+using static Interop.ComCtl32;
+using static Interop;
+using static System.Windows.Forms.VisualStyles.VisualStyleElement;
+using Newtonsoft.Json.Linq;
 
 namespace WinformsControlsTest
 {
     public partial class TrackBars : Form
     {
+        private int V;
```
elachlan, line 15: Never used.

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/TrackBars.cs
```diff
@@ -29,7 +35,18 @@ private void chbRightToLeft_CheckedChanged(object sender, EventArgs e)
...
+        private void tickstyleNone_CheckedChanged(object sender, EventArgs e)
+        {
+            if (trackBar1.TickStyle.Equals(System.Windows.Forms.TickStyle.BottomRight))
+            {
+                trackBar1.TickStyle = System.Windows.Forms.TickStyle.None;
+            }
+            else
+            { trackBar1.TickStyle = System.Windows.Forms.TickStyle.BottomRight; }
+        }
+
```
elachlan, line 47: extra blank line.

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/TrackBars.cs
```diff
@@ -38,9 +55,11 @@ private void numericMinimum_ValueChanged(object sender, EventArgs e)
 
         private void numericMaximum_ValueChanged(object sender, EventArgs e)
         {
+
             trackBar1.Maximum = Decimal.ToInt32(numericMaximum.Value);
             numericMinimum.Maximum = numericMaximum.Value;
             UpdateValueLabel();
+
```
elachlan, line 62: blank line.

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/TrackBars.cs
```diff
@@ -38,9 +55,11 @@ private void numericMinimum_ValueChanged(object sender, EventArgs e)
 
         private void numericMaximum_ValueChanged(object sender, EventArgs e)
         {
+
```
elachlan, line 58: blank line.

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/TrackBars.cs
```diff
@@ -3,11 +3,17 @@
 // See the LICENSE file in the project root for more information.
 
 using System.Windows.Forms;
+using static Interop.ComCtl32;
+using static Interop;
+using static System.Windows.Forms.VisualStyles.VisualStyleElement;
+using Newtonsoft.Json.Linq;
```
elachlan, line 9: run a "remove & sort usings".

RussKie: You can do this by right mouse click on the body of the document and invoke the clean action from the context menu:
image

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -809,13 +809,43 @@ protected override void OnHandleCreated(EventArgs e)
...
+                trackbarSize = Size.Width;
+            }
+            else
+            {
+                trackbarSize = Size.Height;
+            }
+
+            if (maxValue > trackbarSize)
+            {
+                tickFrequency = (int)(maxValue / trackbarSize);
```
hughbe, line 843: is there a risk of a div-by-zero here?

elachlan: It looks like it. Either of the dimensions of size can be zero.

elachlan:
You probably should have rebased this instead of pulling all the changes in from main. It makes it very difficult for the reviewers if you don't.

RussKie:
Hi @conorgee, it looks like instead of rebasing you pulled completely unrelated changes in.
I suggest you do the following to fix this:
1. Hard reset your branch back to fc60e82 commit. There are few ways of  doing this - either via the command line (`git reset --hard fc60e82`) or via git UI (in Git Extensions right click on fc60e82 commit in the revision grid > "Reset current branch to here..." > Hard reset),
2. Fetch the upstream trunk (`git fetch -all` from cli) or via git UI
3. Rebase your branch on top of the latest upstream/main (something like `git -c rebase.autoSquash=false rebase upstream/main`) or via in git UI right click on the upstream/main commit in the revision grid > "Rebase current branch on" > "Selected commit"
4. After that's done (and any merge conflicts resolved, if any), you'll need to re-apply these changes - either manually, or by cherry-picking those in:
image

elachlan:
What if you set the frequency prior to the min/max values via PInvoke? Did you try that? Does it improve the memory usage?

conorgee:
Changing the the frequency before changing max or min has no effect on the memory consumption , The memory just shoots up as soon as you excute PInvoke.

image

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -809,13 +809,43 @@ protected override void OnHandleCreated(EventArgs e)
 
             PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMIN, (WPARAM)(BOOL)false, (LPARAM)_minimum);
             PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMAX, (WPARAM)(BOOL)false, (LPARAM)_maximum);
-            PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETTICFREQ, (WPARAM)_tickFrequency);
+            SetTickFrequency();
```
elachlan, line 812: Even though this didn't make a difference, can you please move the call to `SetTickFrequency` to above the min/max pinvoke calls?

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -807,15 +807,41 @@ protected override void OnHandleCreated(EventArgs e)
                 return;
             }
 
+            SetTickFrequency();
             PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMIN, (WPARAM)(BOOL)false, (LPARAM)_minimum);
             PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMAX, (WPARAM)(BOOL)false, (LPARAM)_maximum);
-            PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETTICFREQ, (WPARAM)_tickFrequency);
```
elachlan, line 812: move `SetTickFrequency` to after the set min/max

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -807,15 +807,41 @@ protected override void OnHandleCreated(EventArgs e)
...
+            // if so then we divide the value by size and only that many ticks to be drawn on the screen
+
+            if (TickStyle == System.Windows.Forms.TickStyle.None)
+            {
+                return;
+            }
+
+            int tickFrequency = _tickFrequency;
+            int trackbarSize = (Orientation == Orientation.Horizontal ? Size.Width : Size.Height) / 2;
+            uint maxValue = (uint)(_minimum + _maximum);
```
RussKie, line 831: What are we calculating here?

elachlan: `/2` to allow a pixel space per tick. The rest is done by Conorgee.

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/TrackBars.cs
```diff
@@ -29,7 +29,17 @@ private void chbRightToLeft_CheckedChanged(object sender, EventArgs e)
             trackBar1.RightToLeft = chbRightToLeft.Checked ? RightToLeft.Yes : RightToLeft.No;
         }
 
-        private void numericMinimum_ValueChanged(object sender, EventArgs e)
+        private void tickstyleNone_CheckedChanged(object sender, EventArgs e)
+        {
+            if (trackBar1.TickStyle.Equals(System.Windows.Forms.TickStyle.BottomRight))
```
RussKie, line 34: A more C#-like would be this:
```suggestion
            if (trackBar1.TickStyle == TickStyle.BottomRight)
```

...and use `using System.Windows.Forms;` added at the top of the document.

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -809,13 +824,40 @@ protected override void OnHandleCreated(EventArgs e)
...
+            int trackbarSize = (Orientation == Orientation.Horizontal ? Size.Width : Size.Height) / 2;
+            uint maxValue = (uint)(-_minimum+_maximum);
+
+            if (maxValue > trackbarSize && trackbarSize != 0)
+            {
+                tickFrequency = ((int)(maxValue / trackbarSize));
+                _autoTicks = false;
+            }
+
+            PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_CLEARTICS, (WPARAM)1, (LPARAM)0);
```
elachlan, line 854: Don't call clear ticks when using `TBS_AUTOTICKS`, for whatever reason its not needed.

RussKie:
I found a bug :)
image

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
RussKie: There are few issues here.
1. Variable name - what's "max value"? Either the name is incorrect, or the way it is used is inappropriate.
2. What are we getting by adding min and max together? The logic here is flawed. What will we get if the range is [-10, 10]?

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
RussKie: That's a very good point.

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -157,13 +157,28 @@ protected override CreateParams CreateParams
...
                         break;
                     case TickStyle.Both:
-                        cp.Style |= (int)(PInvoke.TBS_AUTOTICKS | PInvoke.TBS_BOTH);
+                        cp.Style |= (int)(PInvoke.TBS_BOTH);
+                        if (_autoTicks)
+                        {
+                            cp.Style |= (int)PInvoke.TBS_AUTOTICKS;
+                        }
+
                         break;
```
RussKie, line 182: This can be optimised to remove code duplication, e.g.
```diff
diff --git a/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs b/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
index fb7dec797..7a56f8a10 100644
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
@@ -156,14 +156,20 @@ namespace System.Windows.Forms
                     case TickStyle.None:
                         cp.Style |= (int)PInvoke.TBS_NOTICKS;
                         break;
+
                     case TickStyle.TopLeft:
-                        cp.Style |= (int)(PInvoke.TBS_AUTOTICKS | PInvoke.TBS_TOP);
+                        cp.Style |= (int)(PInvoke.TBS_TOP);
+                        EnableAutoTicksIfRequired();
                         break;
+
                     case TickStyle.BottomRight:
-                        cp.Style |= (int)(PInvoke.TBS_AUTOTICKS | PInvoke.TBS_BOTTOM);
+                        cp.Style |= (int)(PInvoke.TBS_BOTTOM);
+                        EnableAutoTicksIfRequired();
                         break;
+
                     case TickStyle.Both:
-                        cp.Style |= (int)(PInvoke.TBS_AUTOTICKS | PInvoke.TBS_BOTH);
+                        cp.Style |= (int)(PInvoke.TBS_BOTH);
+                        EnableAutoTicksIfRequired();
                         break;
                 }
 
@@ -181,6 +187,14 @@ namespace System.Windows.Forms
                 }
 
                 return cp;
+
+                void EnableAutoTicksIfRequired()
+                {
+                    if (_autoTicks)
+                    {
+                        cp.Style |= (int)PInvoke.TBS_AUTOTICKS;
+                    }
+                }
             }
         }
```

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -809,13 +824,40 @@ protected override void OnHandleCreated(EventArgs e)
...
+            // check if the value of the max is greater then the taskbar size
+            // if so then we divide the value by size and only that many ticks to be drawn on the screen
+
+            if (TickStyle == System.Windows.Forms.TickStyle.None)
+            {
+                return;
+            }
+
+            int tickFrequency = _tickFrequency;
+            int trackbarSize = (Orientation == Orientation.Horizontal ? Size.Width : Size.Height) / 2;
```
RussKie, line 845: The variable name is misleading. The trackbar size is represented by either `Size.Width` or `Size.Height` depending on the orientation. What are calculating here?

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -809,13 +824,40 @@ protected override void OnHandleCreated(EventArgs e)
...
+            uint maxValue = (uint)(-_minimum+_maximum);
```
RussKie, line 846: Ditto here

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -809,13 +824,40 @@ protected override void OnHandleCreated(EventArgs e)
...
+
+            if (maxValue > trackbarSize && trackbarSize != 0)
+            {
+                tickFrequency = ((int)(maxValue / trackbarSize));
+                _autoTicks = false;
+            }
+
+            PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_CLEARTICS, (WPARAM)1, (LPARAM)0);
+            for (int i = tickFrequency; i < maxValue; i += tickFrequency)
```
elachlan, line 855: The bug @RussKie found is caused by the for loop here. 

I suggest we got with a reverse for loop starting from the max value and working our way down to min value, incrementing by the minus tickfrequency (make sure tick frequency is positive int).

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -809,13 +824,40 @@ protected override void OnHandleCreated(EventArgs e)
...
+            {
+                PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETTIC, lParam: (LPARAM)i);
+            }
```
RussKie, line 858: This results in this:
image

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/TrackBars.cs
```diff
@@ -29,7 +29,17 @@ private void chbRightToLeft_CheckedChanged(object sender, EventArgs e)
...
 
-        private void numericMinimum_ValueChanged(object sender, EventArgs e)
+        private void tickstyleNone_CheckedChanged(object sender, EventArgs e)
+        {
+            if (trackBar1.TickStyle == TickStyle.BottomRight)
+            {
+                trackBar1.TickStyle = System.Windows.Forms.TickStyle.None;
+            }
+            else
+            { trackBar1.TickStyle = System.Windows.Forms.TickStyle.BottomRight; }
```
RussKie, line 39: [styling] incorrect formatting
```suggestion
            {
                trackBar1.TickStyle = System.Windows.Forms.TickStyle.BottomRight;
            }
```

RussKie, line 39: We do not generally use fully qualified names except in auto-generated code
```suggestion
            { trackBar1.TickStyle = TickStyle.BottomRight; }
```

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/TrackBars.cs
```diff
@@ -29,7 +29,17 @@ private void chbRightToLeft_CheckedChanged(object sender, EventArgs e)
...
+        }
```
RussKie, line 40: [styling] stray spaces

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -809,13 +824,40 @@ protected override void OnHandleCreated(EventArgs e)
...
             SetTrackBarPosition();
             AdjustSize();
         }
 
+        private void SetTickFrequency()
+        {
+            // check if the value of the max is greater then the taskbar size
+            // if so then we divide the value by size and only that many ticks to be drawn on the screen
+
+            if (TickStyle == System.Windows.Forms.TickStyle.None)
```
RussKie, line 839:
```suggestion
            if (TickStyle == TickStyle.None)
```

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -809,13 +824,40 @@ protected override void OnHandleCreated(EventArgs e)
...
             PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETPAGESIZE, (WPARAM)0, (LPARAM)_largeChange);
             PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETLINESIZE, (WPARAM)0, (LPARAM)_smallChange);
             SetTrackBarPosition();
             AdjustSize();
         }
 
+        private void SetTickFrequency()
+        {
+            // check if the value of the max is greater then the taskbar size
+            // if so then we divide the value by size and only that many ticks to be drawn on the screen
```
RussKie, line 837: Please use sentence casing for comments with full stop at the end.

RussKie:
I've spent some time playing with the code, and I don't think it'll work if we continue to follow this path.

The control has to be re-created whenever `_autoTicks` changes. However, `SetTickFrequency()` API is getting invoked when a handle is created (i.e., from `OnHandleCreated()`), which will lead to a terminal failure - we can't recreate a control while creating the control. To avoid this, the flag has to be determined at the time control is created, i.e., when `CreateParams` property is invoked.
The flag must be re-determined whenever `Minimum` or `Maximum` values change.

Also, the ticks are drawn by the native control - i.e., the value are set via `SendMessage(TBM_SETTIC)` - it must be invoked _after_ `Minimum` or `Maximum` values are set.

RussKie:
Something like this:
```cs
        private bool AutoDrawTicks
        {
            get
            {
                if (TickStyle == TickStyle.None)
                {
                    return true;
                }

                int tickFrequency = _tickFrequency;
                int maxTickCount = (Orientation == Orientation.Horizontal ? Size.Width : Size.Height) / 2;
                int range = _maximum - _minimum;

                return range <= maxTickCount || maxTickCount == 0;
            }
        }
```

```diff
diff --git a/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs b/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
index 964c4d52e..97d4b5ad1 100644
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
@@ -833,28 +831,28 @@ namespace System.Windows.Forms
 
         private void SetTickFrequency()
         {
             // check if the value of the max is greater then the taskbar size
             // if so then we divide the value by size and only that many ticks to be drawn on the screen
 
-            if (TickStyle == System.Windows.Forms.TickStyle.None)
+            if (TickStyle == TickStyle.None || AutoDrawTicks)
             {
                 return;
             }
 
             int tickFrequency = _tickFrequency;
-            int trackbarSize = (Orientation == Orientation.Horizontal ? Size.Width : Size.Height) / 2;
-            uint maxValue = (uint)(-_minimum+_maximum);
+            int maxTickCount = (Orientation == Orientation.Horizontal ? Size.Width : Size.Height) / 2;
+            int range = _maximum - _minimum;
 
-            if (maxValue > trackbarSize && trackbarSize != 0)
+            if (range > maxTickCount && maxTickCount != 0)
             {
-                tickFrequency = ((int)(maxValue / trackbarSize));
-                _autoTicks = false;
+                tickFrequency = range / maxTickCount;
             }
 
             PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_CLEARTICS, (WPARAM)1, (LPARAM)0);
-            for (int i = tickFrequency; i < maxValue; i += tickFrequency)
+
+            for (int i = _minimum + tickFrequency; i < _maximum - tickFrequency; i += tickFrequency)
             {
-                PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETTIC, lParam: (LPARAM)i);
+                LRESULT lresult = PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETTIC, lParam: (IntPtr)i);
+                Debug.Assert((bool)(BOOL)lresult);
             }
         }
```

conorgee:
> Something like this:
> 
> ```cs
>         private bool AutoDrawTicks
>         {
>             get
>             {
>                 if (TickStyle == TickStyle.None)
>                 {
>                     return true;
>                 }
> 
>                 int tickFrequency = _tickFrequency;
>                 int maxTickCount = (Orientation == Orientation.Horizontal ? Size.Width : Size.Height) / 2;
>                 int range = _maximum - _minimum;
> 
>                 return range <= maxTickCount || maxTickCount == 0;
>             }
>         }
> ```
> 
> ```diff
> diff --git a/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs b/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
> index 964c4d52e..97d4b5ad1 100644
> --- a/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
> +++ b/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
> @@ -833,28 +831,28 @@ namespace System.Windows.Forms
>  
>          private void SetTickFrequency()
>          {
>              // check if the value of the max is greater then the taskbar size
>              // if so then we divide the value by size and only that many ticks to be drawn on the screen
>  
> -            if (TickStyle == System.Windows.Forms.TickStyle.None)
> +            if (TickStyle == TickStyle.None || AutoDrawTicks)
>              {
>                  return;
>              }
>  
>              int tickFrequency = _tickFrequency;
> -            int trackbarSize = (Orientation == Orientation.Horizontal ? Size.Width : Size.Height) / 2;
> -            uint maxValue = (uint)(-_minimum+_maximum);
> +            int maxTickCount = (Orientation == Orientation.Horizontal ? Size.Width : Size.Height) / 2;
> +            int range = _maximum - _minimum;
>  
> -            if (maxValue > trackbarSize && trackbarSize != 0)
> +            if (range > maxTickCount && maxTickCount != 0)
>              {
> -                tickFrequency = ((int)(maxValue / trackbarSize));
> -                _autoTicks = false;
> +                tickFrequency = range / maxTickCount;
>              }
>  
>              PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_CLEARTICS, (WPARAM)1, (LPARAM)0);
> -            for (int i = tickFrequency; i < maxValue; i += tickFrequency)
> +
> +            for (int i = _minimum + tickFrequency; i < _maximum - tickFrequency; i += tickFrequency)
>              {
> -                PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETTIC, lParam: (LPARAM)i);
> +                LRESULT lresult = PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETTIC, lParam: (IntPtr)i);
> +                Debug.Assert((bool)(BOOL)lresult);
>              }
>          }
> ```

This works but the memory consumption shoots up due to max and min message being set before setTickFrequency. The memory consumption goes down once it sends these messages but it can take a bit of time to load. Is their a way around this?

https://github.com/conorgee/winforms/blob/6053465ed068be7bc9a967c63710458fc3c70a1b/src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs#L1048-L1054
```
              
  if (IsHandleCreated)
                {
                    PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMIN, (WPARAM)(BOOL)false, (LPARAM)_minimum);
                    PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMAX, (WPARAM)(BOOL)true, (LPARAM)_maximum);
                    SetTickFrequency();
                    Invalidate();
                }


```

elachlan:
> Is their a way around this?

Yes, You have to conditionally call `SetTickFrequency()` based on `AutoDrawTicks`. Otherwise separate out the logic for the Auto vs Manual Ticks.

elachlan:
@merriemcgaw/ @dreddy-work RussKie was reviewing this, could we get another team member to review it?

dreddy-work:
> @merriemcgaw/ @dreddy-work RussKie was reviewing this, could we get another team member to review it?

Thanks for the ping @elachlan. Draft PRs often overlooked. I will go over these changes. Given this is short week, expect an update sometime next week.

elachlan:
@conorgee if you feel like you have sorted all the feedback raised and the trackbar is functioning as expected. Consider unmarking the pr as draft.

conorgee:
The control needs to be recreated (i.e., call RecreateHandle()) whenever AutoDrawTicks changes when not invoked during handle creation sequence. I however am having trouble implementing this any suggestions ?

elachlan:
`CreateHandle` uses `RecreatingHandle`, so maybe check that? Or create your own bool which is toggled when the handle is being created, then check its values to see if you recreate?

RussKie:
There are several places that invoke ticks drawing method. That method is condition on the flag indicating whether ticks are auto-drawn or not.
Whenever the value of the flag toggles we need to recreate the handle.
However, and this is important, we must not try to recreate the handle while we are creating the handle (that is called from OnHandleCreated method). We mustn't set minvalue or maxvalue on the native control if we need to recreate the control either.

So we'll probably need few bools (bit vector as optimisation,  but we can worry about it a little later) that track the current value of the auto-draw flag and whether ticks drawing was called from OnHandleCreated.

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -809,13 +819,56 @@ protected override void OnHandleCreated(EventArgs e)
...
+
+                int tickFrequency = _tickFrequency;
+                int maxTickCount = (Orientation == Orientation.Horizontal ? Size.Width : Size.Height) / 2;
+                uint range = (uint)(_maximum - _minimum);
+
+                return range <= maxTickCount || maxTickCount == 0;
+            }
+        }
+
+        private void SetTickFrequency()
```
RussKie, line 846: As discussed offline, this name does not reflect the implementation.

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -809,13 +819,56 @@ protected override void OnHandleCreated(EventArgs e)
...
+            {
+                tickFrequency = (int)(range / maxTickCount);
+            }
+
+            PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_CLEARTICS, (WPARAM)1, (LPARAM)0);
+            for (int i = (_minimum + tickFrequency); i < _maximum - tickFrequency; i += tickFrequency)
+            {
+                LRESULT lresult = PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETTIC, lParam: (IntPtr)i);
+                Debug.Assert((bool)(BOOL)lresult);
+            }
```
RussKie, line 869: How does it behave when `trackbar.Minimum = int.MinValue` and `trackbar.Maximum = int.MaxValue`?

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -809,13 +819,56 @@ protected override void OnHandleCreated(EventArgs e)
...
+        private bool AutoDrawTicks
+        {
+            get
+            {
+                if (TickStyle == TickStyle.None)
+                {
+                    return true;
+                }
+
+                int tickFrequency = _tickFrequency;
```
RussKie, line 838: Why do we need this?

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -809,13 +819,56 @@ protected override void OnHandleCreated(EventArgs e)
...
+                int maxTickCount = (Orientation == Orientation.Horizontal ? Size.Width : Size.Height) / 2;
+                uint range = (uint)(_maximum - _minimum);
+
+                return range <= maxTickCount || maxTickCount == 0;
```
RussKie, line 842:
```suggestion
                if (TickStyle == TickStyle.None)
                {
                    return true;
                }

                int size = Orientation == Orientation.Horizontal ? Size.Width : Size.Height;
                if (size == 0)
                {
                    return true;
                }

                uint range = (uint)(_maximum - _minimum);
                return range <= (size / 2);
```

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
conorgee: image

The trackbar itself doesnt work when the value of min and max have a range greater then int.maxValue. So Im not too sure as to what I can implement as its a problem as soon as you send the message to windows.

conorgee:
> 



> So we'll probably need few bools (bit vector as optimisation, but we can worry about it a little later) that track the current value of the auto-draw flag and whether ticks drawing was called from OnHandleCreated.

Attempted a version of this but still stuck figuring out where to set the flags

RussKie:
> > So we'll probably need few bools (bit vector as optimisation, but we can worry about it a little later) that track the current value of the auto-draw flag and whether ticks drawing was called from OnHandleCreated.
> 
> Attempted a version of this but still stuck figuring out where to set the flags

As usual I suggest starting with a block flow diagram, and try to answer the question - "_Under which conditions should a control be recreated?_" (hint: this involves `AutoDrawTicks`). Once this is established, you should be able to follow the three paths in which `CreateTicks` method is called and conditionalise based on the answer to the question.

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -807,13 +817,66 @@ protected override void OnHandleCreated(EventArgs e)
...
+        {
+            // Check if the value of the max is greater then the taskbar size.
+            // If so then we divide the value by size and only that many ticks to be drawn on the screen.
+
+            if (AutoDrawTicks)
+            {
+                return;
+            }
+
+            if (_AutoTicksRecreate!)
```
RussKie, line 861: This looks like a typo. In this case, `!` is used as a [_null-forgiving operator_](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-forgiving) (aka the _"bang" operator_).
For boolean operations `!` has to be placed at the start.

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -807,13 +817,66 @@ protected override void OnHandleCreated(EventArgs e)
...
+            {
+                RecreateHandle();
+            }
+
+            int maxTickCount = (Orientation == Orientation.Horizontal ? Size.Width : Size.Height) / 2;
+            uint range = (uint)(_maximum - _minimum);
+
+            if (range > maxTickCount && maxTickCount != 0)
+            {
+                _tickFrequency = (int)(range / maxTickCount);
```
RussKie, line 871: We can't do this - we'll be altering the developer-supplied value.

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
RussKie: We can't completely remove this line.
For example, the desired tick frequency is not reflected:
image

conorgee:
> Once this is established, you should be able to follow the three paths in which `CreateTicks` method is called and conditionalise based on the answer to the question

I have came up with a hypothesis to solve this. By just setting the flag within `OnHandleCreated()` . This seems to have solved the auto draw ticks being set when coming from a large range to a range the size of the trackbar.

conorgee:
@RussKie 
 Apologies for the handwriting.
image

Flow 
Start- within oncreatehandle set _AutoTicksRecreate to true – if Auto drawticks false then recreatehandle if true then skip e.g when creatinghandle – if autodrawticks returns true then we set tick freq by sending message to windows and returning – bool tickFreqFine is set to true, if the tickfrequency has not equal to one 1 e.g altered by devolper. Then we check to see if the amount ticks they are wanting to draw is less then the space we have to draw ticks on the bar if true set tickFreqFine to false - if (range > maxTickCount && maxTickCount != 0 && tickFreqFine) then we know the amount of ticks being drawn will not work so we must set calculate the tickfrequency e.g ticksDrawnFrequency = (int)(range / maxTickCount)- then we send message to windows to clear ticks and have a for loop that indivdually draws the ticks on the track bar in correlation to the frequency.

Different conditions
For min int value – trackbar breaks and is unuseable this is a native windows issue
If range is greater then the max int value- the trackbar again breaks native windows issue
For max max int value – trackbar works perfect so long as min isnt less then zero
When vertical or horizontal – it works as the size of the trackbar is calculated based on its current orientation
If going from large value to small value – the trackbar recreates handle which then sets auto ticks if amount of ticks is smaller then trackbar size
If trackbar size is very large – then autoticks will still be on this is a very small use case that will not be accounted for

conorgee:
When you change maximum to int.maxValue the memory consumption sky rockets due to auto ticks being still enabled before it excutes recreate handle. 
              
    if (IsHandleCreated)
          {
              PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMIN, (WPARAM)(BOOL)false, (LPARAM)_minimum);
              PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMAX, (WPARAM)(BOOL)true, (LPARAM)_maximum);
              bool recreateHandle = false;
              if (_autoDrawTicks != ShouldAutoDrawTicks())
              {
                  recreateHandle = true;
              }


              if (!recreateHandle && IsHandleCreated)
              {
                  DrawTicks();
                  Invalidate();
              }
              else
              {
                  RecreateHandle();
              }
           }

This memory goes down once it excutes this but this is still a noticeable issue as it may take a long time to reach recreate handle.

RussKie:
> When you change maximum to int.maxValue the memory consumption sky rockets due to auto ticks being still enabled before it excutes recreate handle.
> 
> ```cs
>       if (IsHandleCreated)
>       {
>           PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMIN, (WPARAM)(BOOL)false, (LPARAM)_minimum);
>           PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMAX, (WPARAM)(BOOL)true, (LPARAM)_maximum);
>           bool recreateHandle = false;
>           if (_autoDrawTicks != ShouldAutoDrawTicks())
>           {
>               recreateHandle = true;
>           }
> 
> 
>           if (!recreateHandle && IsHandleCreated)
>           {
>               DrawTicks();
>               Invalidate();
>           }
>           else
>           {
>               RecreateHandle();
>           }
>        }
> ```
> 
> This memory goes down once it excutes this but this is still a noticeable issue as it may take a long time to reach recreate handle.

We shouldn't be sending any messages if we know that we have to recreate the control, so the check determining that has to be hoisted up.

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -807,15 +828,68 @@ protected override void OnHandleCreated(EventArgs e)
...
+
+            if (ticksDrawnFrequency != 0)
+            {
+                ticksDrawn = (int)(range / ticksDrawnFrequency);
+            }
+
+            if (maxTickCount != 0 && ticksDrawn > maxTickCount)
+            {
+                ticksDrawnFrequency = (int)(range / maxTickCount);
+            }
```
RussKie, line 883: 🤔 I've been looking at this code for some time, and I'm having troubles fully understanding it. Why do we need this: `ticksDrawn = (int)(range / ticksDrawnFrequency)`? 
We need to figure out the tick frequency - i.e., the increment between each drawn tick.

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -181,6 +182,16 @@ protected override CreateParams CreateParams
...
+
+                void EnableAutoTicksIfRequired()
+                {
+                    _autoDrawTicks = false;
+                    if (ShouldAutoDrawTicks())
+                     {
+                        _autoDrawTicks = true;
+                        cp.Style |= (int)PInvoke.TBS_AUTOTICKS;
+                     }
+                }
```
RussKie, line 194: I think this can be simplified:
1. Call `_autoDrawTicks = ShouldAutoDrawTicks();` at the top of the property.
2. Update this as follows:
    ```suggestion

                void EnableAutoTicksIfRequired()
                {
                     if (_autoDrawTicks)
                     {
                        cp.Style |= (int)PInvoke.TBS_AUTOTICKS;
                     }
                }
    ```

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -992,11 +1066,20 @@ public void SetRange(int minValue, int maxValue)
...
+                    if (recreateHandle)
+                    {
+                        RecreateHandle();
+                    }
 
+                    PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMIN, (WPARAM)(BOOL)false, (LPARAM)_minimum);
+                    PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMAX, (WPARAM)(BOOL)true, (LPARAM)_maximum);
+                    DrawTicks();
                     Invalidate();
                 }
```
RussKie, line 1084: I don't think this is correct. 
`RecreateHandle()` is kind of "end of the line" method, meaning we shouldn't be calling anything after that as the control will be recreated from a scratch and reinitialised with the configured values. This means we can't just `return` after calling  `RecreateHandle()` as we need to set `_value`.
The logic should be similar to how `TickFrequency` is implemented.

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
conorgee: if the user wants a range of 2000 we and tick frequency set to 500 , this calculates that this frequency is ok e.g only 4 ticks drawn and the users input will be drawn on the screen rather then trackbar size / 2 .  This stops if the users input range is larger then trackbar size we dont just overide the frequency they set if it will fit on the bar . The reason Im not using auto ticks for this is because the memory consumption could be huge even if they only want 5 ticks on screen if the range is large.

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
conorgee: where do we set the  `_autoDrawTicks` to true or false ? . If you change it in `ShouldAutoDrawTicks()` would the values not always be the same when comparing ?. e.g. `if (_autoDrawTicks != ShouldAutoDrawTicks())`

RussKie: At the top of the property:
```cs
        protected override CreateParams CreateParams
        {
            get
            {
                _autoDrawTicks = ShouldAutoDrawTicks();

                CreateParams cp = base.CreateParams;
                ...
```

This is the time when the native control is being created.

> ...would the values not always be the same when comparing ?. e.g. `if (_autoDrawTicks != ShouldAutoDrawTicks())`

No, because `ShouldAutoDrawTicks()` can change based on several conditions.

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -807,15 +832,72 @@ protected override void OnHandleCreated(EventArgs e)
...
+
+        int size = Orientation == Orientation.Horizontal ? Size.Width : Size.Height;
+        if (size == 0)
+        {
+            return true;
+        }
+
+        uint range = (uint)(_maximum - _minimum);
+        return range <= (size / 2);
+        }
```
RussKie, line 860: [nit] incorrect indentation
[file organisation] new API must be placed in a way to maintain the lexicographical order

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -807,15 +832,72 @@ protected override void OnHandleCreated(EventArgs e)
...
+        ///  If so then we divide the value by size and only that many ticks to be drawn on the screen.
+        /// </summary>
+        private void DrawTicks()
+        {
+            if (_tickStyle == TickStyle.None)
+            {
+                return;
+            }
+
+            int ticksDrawnFrequency = _tickFrequency;
```
RussKie, line 873:
```suggestion
            int drawnTickFrequency = _tickFrequency;
```
...or just
```suggestion
            int tickFrequency = _tickFrequency;
```

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
RussKie: Any reason we can't have it like this?
```cs
            if (range > maxTickCount && maxTickCount != 0)
            {
                int calculatedTickFrequency = (int)(range / maxTickCount);
                if (calculatedTickFrequency > tickFrequency)
                {
                    tickFrequency = calculatedTickFrequency;
                }
            }
```

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
RussKie: This is still outstanding.

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
RussKie: still outstanding

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -557,11 +566,26 @@ public int TickFrequency
                 }
 
                 _tickFrequency = value;
-                if (IsHandleCreated)
+                // set the recreateHandle flag to null,
+                // if the _autoDrawTicks flag does not equal what ShouldAutoDrawTicks()
+                // we set the recreateHandle to true,
+                // then we skip sending the messages and,
+                // recreateHandle will be called at the end of the method.
```
RussKie, line 573: A good comment is the one explaining "why" or "how" rather than restating what the code does.
I would rephase it to something like the following:
```suggestion
                // Determine if the decision of whether the ticks drawing is performed by the native control or the Windows Forms runtime
                // is still valid. If it's no longer valid, we'll need to recreate the native control.
```

Also, this repo strives to wrap lines between 120-150 chars, so the text should be reflowed.

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -557,11 +566,26 @@ public int TickFrequency
...
+                bool recreateHandle = false;
+                if (IsHandleCreated && _autoDrawTicks != ShouldAutoDrawTicks())
                 {
-                    PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETTICFREQ, (WPARAM)value);
+                    recreateHandle = true;
+                }
```
RussKie, line 578: Since we use the identical code in two places, perhaps it makes sense to extract it into a new method, e.g.:
```cs
        /// <summary>
        ///  Indicates if the native control needs to be recreated because the decision of whether the ticks drawing is
        ///  performed by the native control or the Windows Forms runtime is no longer valid.
        /// </summary>
        private bool ShouldRecreateHandle() => IsHandleCreated && _autoDrawTicks != ShouldAutoDrawTicks();
```

```suggestion
                bool recreateHandle = ShouldRecreateHandle();
```

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -557,11 +566,26 @@ public int TickFrequency
...
+
+                if (!recreateHandle && IsHandleCreated)
+                {
+                    DrawTicks();
                     Invalidate();
                 }
+                else
+                {
+                    RecreateHandle();
+                }
```
RussKie, line 588: I don't think this is correct - i.e., this can get invoked when `!recreateHandle && IsHandleCreated` does not equate to `true`. However, the control must only be recreated when `recreateHandle == true`.

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -744,6 +768,43 @@ protected override void CreateHandle()
...
+            }
+
+            int drawnTickFrequency = _tickFrequency;
+            if (_autoDrawTicks)
+            {
+                PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETTICFREQ, (WPARAM)drawnTickFrequency);
+                return;
+            }
+
+            int maxTickCount = (Orientation == Orientation.Horizontal ? Size.Width : Size.Height) / 2;
```
RussKie, line 789: This probably warrants a comment explaining why we're dividing by 2 here.

RussKie (commented):
Looks like we're getting close. 👍

RussKie:
@dotnet/dotnet-winforms please start the reviews, thanks.

RussKie:
@conorgee trackbar related tests are failing, which indicate there's an undesirable behaviour change:
image

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -148,22 +146,25 @@ protected override CreateParams CreateParams
         {
             get
             {
+                _autoDrawTicks = ShouldAutoDrawTicks();
```
RussKie, line 149:
```suggestion
                if (LocalAppContextSwitches.TrackBarLegacyRendering)
                {
                    // Navite control always renders
                    _autoDrawTicks = true;
                }
                else 
                {
                    _autoDrawTicks = ShouldAutoDrawTicks();
                }
```

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -1044,6 +1112,32 @@ private void SetTrackBarPosition()
             }
         }
 
+        /// <summary>
+        /// This checks all the use cases that we potentially might want to keep `TBS_AUTOTICKS`.
+        /// </summary>
+        private bool ShouldAutoDrawTicks()
+        {
+            if (TickStyle == TickStyle.None)
```
RussKie, line 1120:
```suggestion
            if (!LocalAppContextSwitches.TrackBarModernRendering)
            {
                return true;
            }

            if (TickStyle == TickStyle.None)
```

RussKie (commented):
Here are some ideas how the feature switch can be incorporated.

RussKie (approved):

elachlan:
@conorgee Thank you for your persistence on this.

lonitra (commented):
Thank you for adding your changes under a switch! Just a couple comments.

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -989,14 +1055,17 @@ public void SetRange(int minValue, int maxValue)
...
+                // If user opts out of TrackBarModernRendering then recreateHandle
+                // will always be false.
+                if (IsHandleCreated && !recreateHandle)
                 {
                     PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMIN, (WPARAM)(BOOL)false, (LPARAM)_minimum);
-
-                    // We must repaint the trackbar after changing the range.
                     PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMAX, (WPARAM)(BOOL)true, (LPARAM)_maximum);
-
+                    DrawTicks();
```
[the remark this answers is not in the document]
conorgee: Very true,  so when `TBS_AUTOTICKS ` is used the range is automatically changed on the trackbar however when turned off we need to set the ticks manually. I feel this would be an appropriate solution, if they opt out of the fix `_autoDrawTicks` will always be true.
```
                    if (!_autoDrawTicks)
                    {
                        DrawTicks();
                    }
```

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -557,10 +569,24 @@ public int TickFrequency
                 }
 
                 _tickFrequency = value;
+                // Determine if the decision of whether the ticks drawing,
+                // is performed by the native control or the Windows Forms runtime
+                // is still valid. If it's no longer valid, we'll need to recreate the native control.
+                bool recreateHandle = ShouldRecreateHandle();
                 if (IsHandleCreated)
                 {
                     PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETTICFREQ, (WPARAM)value);
```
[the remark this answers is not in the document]
conorgee: Good point definitely a oversight on my part , as a solution I decided to remove the `TBM_SETTICFREQ ` message from within `DrawTicks()`.

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -989,13 +1049,20 @@ public void SetRange(int minValue, int maxValue)
...
+                // will always be false.
+                if (IsHandleCreated && !recreateHandle)
                 {
                     PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMIN, (WPARAM)(BOOL)false, (LPARAM)_minimum);
-
-                    // We must repaint the trackbar after changing the range.
                     PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMAX, (WPARAM)(BOOL)true, (LPARAM)_maximum);
+                    if (!_autoDrawTicks)
+                    {
+                        DrawTicks();
```
lonitra, line 1064: Since your other change removed `PInvoke.SendMessage(...,TBM_SETTICFREQ,...)` from `DrawTicks()`, we should add that call before `DrawTicks()`

conorgee: I dont believe we should, when the switch is opted out of `PInvoke.SendMessage(...,TBM_SETTICFREQ,...) ` doesnt need to be sent unless the user specifys a change which will be handled in the `TickFrequency()` method or when handle is created.

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
```diff
@@ -807,9 +866,10 @@ protected override void OnHandleCreated(EventArgs e)
                 return;
             }
 
+            Debug.Assert(_autoDrawTicks == ShouldAutoDrawTicks());
             PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMIN, (WPARAM)(BOOL)false, (LPARAM)_minimum);
             PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETRANGEMAX, (WPARAM)(BOOL)false, (LPARAM)_maximum);
-            PInvoke.SendMessage(this, (User32.WM)PInvoke.TBM_SETTICFREQ, (WPARAM)_tickFrequency);
```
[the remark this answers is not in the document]
conorgee: done.

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.cs
lonitra: Ah, I see. Thanks for the explanation!

lonitra (approved):
LGTM, thank you so much for all your effort 🥳

conorgee:
No problem at all, thank you for the opportunity to work on this. It has been a truly remarkable learning experience for me and everyone has been very helpful . At the beginining the issue looked alot simpler then it turned out to be but luckily with the help of @RussKie and @elachlan we could figure this one out. Hats off to the WinForms team. This has been an extraordinary experience for me, thank you.

elachlan:
Thanks for persisting with this.

lonitra:
@merriemcgaw my assumption is that this fix will need to be documented somewhere so that users are aware of the switch?

dreddy-work:
> @merriemcgaw my assumption is that this fix will need to be documented somewhere so that users are aware of the switch?

This should go with other switches we are introducing/introduced. https://github.com/dotnet/winforms/issues/8442

PR #8063, opened 2022-10-28
Reorder Message fields

JeremyKuhne:
Historically our first four fields matched the layout of MSG. Keeping it that way for better back compat.

As Message doesn't match MSG, they shouldn't be blitted over each other, but for those that might be doing this we can keep the order. Note that other future changes to the struct could break doing this sort of thing even if the first four fields stay the same.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8063)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Message.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Message.cs
@@ -20,16 +20,20 @@ namespace System.Windows.Forms
         private static readonly TraceSwitch s_allWinMessages = new("AllWinMessages", "Output every received message");
 #endif
 
-        // Using prefixed variants of the property names for easier diffing.
-#pragma warning disable IDE1006 // Naming Styles
-        internal LRESULT ResultInternal;
-        internal LPARAM LParamInternal;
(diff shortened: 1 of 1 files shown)


dreddy-work (approved):

dreddy-work:
I should have put a note to wait on merge. We should be good given branding is already made. TimeLine was from 1st NOV. Will keep an eye if we need to do somethign about this.

merriemcgaw:
This one went into .NET 8. We'll need to cherry pick the changes for the 7.0.1 branch and add servicing consider tag to the actual bug.

RussKie:
@JeremyKuhne in the future please use [GitHub keywords](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) to establish cross-links between issues and pull requests.

JeremyKuhne:
@RussKie I deliberately did not use the keyword in this case as I didn't want the issue automatically closed.

RussKie:
Closed issues can be reopened, The lack of cross-referencing affects ability to data-mine via Kusto queries.

PR #8065, opened 2022-10-29
Enable nullability in ToolStripDropDownItem

gpetrou:
## Proposed changes

- Enable nullability in ToolStripDropDownItem.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8065)


src/System.Windows.Forms/src/System/Windows/Forms/ToolStripDropDownItem.cs
```diff
@@ -56,55 +58,56 @@ protected ToolStripDropDownItem(string text, Image image, params ToolStripItem[]
...
             {
-                if (dropDown is null)
+                if (_dropDown is null)
                 {
                     DropDown = CreateDefaultDropDown();
-                    if (!(this is ToolStripOverflowButton))
+                    if (this is not ToolStripOverflowButton)
                     {
-                        dropDown.SetAutoGeneratedInternal(true);
+                        _dropDown!.SetAutoGeneratedInternal(true);
```
dreddy-work, line 71: Where is `_dropDown` set in this block? it entered when it is null.

RussKie: Here: `DropDown = CreateDefaultDropDown();`

RussKie: I wonder if we can apply `MemberNotNull(nameof(_dropDown))` to the setter... 💭

RussKie (commented):
Looks good modulo `DropDown` API questions

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripDropDownItem.cs
gpetrou: The problem is that DropDown can also be set to `null`. So, we cannot use `MemberNotNull(nameof(_dropDown))`. There is a pending TODO for CA2245 warnings in CodeAnalysis.src.globalconfig that if I am not mistaken should address such cases.

RussKie: Ah, yes. Thank you for confirming it.

RussKie (approved):

PR #8067, opened 2022-10-29
Enable nullability in Help

gpetrou:
## Proposed changes

- Enable nullability in Help.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8067)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/Interop/Hhctrl/Interop.HtmlHelpW.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/Hhctrl/Interop.HtmlHelpW.cs
@@ -9,16 +9,16 @@ internal partial class Interop
     internal static partial class Hhctl
     {
         [DllImport(Libraries.Hhctrl, CharSet = CharSet.Unicode, ExactSpelling = true)]
-        public static extern int HtmlHelpW(IntPtr hwndCaller, string pszFile, HH uCommand, IntPtr dwData);
+        public static extern int HtmlHelpW(IntPtr hwndCaller, string? pszFile, HH uCommand, IntPtr dwData);
 
-        public static int HtmlHelpW(HandleRef<HWND> hwndCaller, string pszFile, HH uCommand, IntPtr dwData)
+        public static int HtmlHelpW(HandleRef<HWND> hwndCaller, string? pszFile, HH uCommand, IntPtr dwData)
         {
             int result = HtmlHelpW(hwndCaller.Handle, pszFile, uCommand, dwData);
             GC.KeepAlive(hwndCaller.Wrapper);
             return result;
         }
 
-        public static unsafe int HtmlHelpW(HandleRef<HWND> hwndCaller, string pszFile, HH uCommand, string data)
+        public static unsafe int HtmlHelpW(HandleRef<HWND> hwndCaller, string? pszFile, HH uCommand, string data)
         {
             fixed (char* dwData = data)
             {
@@ -26,7 +26,7 @@ internal partial class Interop
             }
         }
 
-        public static unsafe int HtmlHelpW<T>(HandleRef<HWND> hwndCaller, string pszFile, HH uCommand, ref T data) where T : unmanaged
+        public static unsafe int HtmlHelpW<T>(HandleRef<HWND> hwndCaller, string? pszFile, HH uCommand, ref T data) where T : unmanaged
         {
             fixed (void* dwData = &data)
             {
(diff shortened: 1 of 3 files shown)


src/System.Windows.Forms.Primitives/src/Interop/Hhctrl/Interop.HtmlHelpW.cs
```diff
@@ -9,24 +9,24 @@ internal partial class Interop
     internal static partial class Hhctl
     {
         [DllImport(Libraries.Hhctrl, CharSet = CharSet.Unicode, ExactSpelling = true)]
-        public static extern int HtmlHelpW(IntPtr hwndCaller, string pszFile, HH uCommand, IntPtr dwData);
+        public static extern int HtmlHelpW(IntPtr hwndCaller, string? pszFile, HH uCommand, IntPtr dwData);
```
elachlan, line 12: Raised an issue for win32metadata so we can use cswin32 for these PInvokes.
https://github.com/microsoft/win32metadata/issues/1345

I suspect the change would mean the `pszFile` would change from a `string` to `LPCWSTR` (struct) and therefore no longer be nullable.

RussKie: Thank you @elachlan. How would it reconcile with the docs?
image

src/System.Windows.Forms/src/System/Windows/Forms/Help.cs
```diff
@@ -396,7 +392,7 @@ private static unsafe HH MapCommandToHTMLCommand(HelpNavigator command, string p
                     {
                         try
                         {
-                            htmlParam = int.Parse(param, CultureInfo.InvariantCulture);
+                            htmlParam = int.Parse(param!, CultureInfo.InvariantCulture);
```
RussKie, line 395: I can't see any reason to do this. If we know `param` is null - we can just return `HH.HELP_CONTEXT`. Exceptions are expensive, and if we should avoid them if we can.

src/System.Windows.Forms/src/System/Windows/Forms/Help.cs
```diff
@@ -118,26 +116,27 @@ public static unsafe void ShowPopup(Control parent, string caption, Point locati
         /// <summary>
         ///  Displays HTML 1.0 Help with the specified parameters
         /// </summary>
-        private static unsafe void ShowHTML10Help(Control parent, string url, HelpNavigator command, object param)
+        private static unsafe void ShowHTML10Help(Control? parent, string? url, HelpNavigator command, object? param)
         {
-            Debug.WriteLineIf(WindowsFormsHelpTrace.TraceVerbose, "Help:: ShowHTML10Help:: " + url + ", " + command.ToString("G") + ", " + param);
+            Debug.WriteLineIf(WindowsFormsHelpTrace!.TraceVerbose, "Help:: ShowHTML10Help:: " + url + ", " + command.ToString("G") + ", " + param);
```
RussKie, line 121: Since we're touching these, could you please use string interpolation syntax?

src/System.Windows.Forms/src/System/Windows/Forms/Help.cs
```diff
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
```
RussKie, line 2: Could you please reformat the document too? looks like there are few styling imperfections.
image

RussKie (changes requested):

src/System.Windows.Forms.Primitives/src/Interop/Hhctrl/Interop.HtmlHelpW.cs
elachlan: Yes, I saw that. I am not sure how C# will work with LPCWSTR. I think you will just be able to pass null without issues.

dreddy-work (approved):

src/System.Windows.Forms/src/System/Windows/Forms/Help.cs
```diff
@@ -394,16 +390,14 @@ private static unsafe HH MapCommandToHTMLCommand(HelpNavigator command, string p
 
                 case HelpNavigator.TopicId:
                     {
-                        try
+                        if (int.TryParse(param, out int htmlParamAsInt))
```
RussKie, line 393: Nice!

RussKie (approved):

src/System.Windows.Forms.Primitives/src/Interop/Hhctrl/Interop.HtmlHelpW.cs
RussKie: We can iterate on this later when CsWin32 is ready for this.

PR #8072, opened 2022-10-31

gpetrou:
## Proposed changes

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8072)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/MenuTimer.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/MenuTimer.cs
@@ -38,7 +38,7 @@ namespace System.Windows.Forms
             }
             set
             {
-                Debug.WriteLineIf(ToolStrip.s_menuAutoExpandDebug!.TraceVerbose && currentItem != value, "[MenuTimer.CurrentItem] changed: " + ((value is null) ? "null" : value.ToString()));
+                Debug.WriteLineIf(ToolStrip.s_menuAutoExpandDebug!.TraceVerbose && currentItem != value, $"[MenuTimer.CurrentItem] changed: {(value is null ? "null" : value.ToString())}");
                 currentItem = value;
             }
         }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/MenuTimer.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/MenuTimer.cs
@@ -38,7 +38,7 @@ namespace System.Windows.Forms
             }
             set
             {
-                Debug.WriteLineIf(ToolStrip.s_menuAutoExpandDebug.TraceVerbose && currentItem != value, "[MenuTimer.CurrentItem] changed: " + ((value is null) ? "null" : value.ToString()));
+                Debug.WriteLineIf(ToolStrip.s_menuAutoExpandDebug!.TraceVerbose && currentItem != value, "[MenuTimer.CurrentItem] changed: " + ((value is null) ? "null" : value.ToString()));
                 currentItem = value;
             }
         }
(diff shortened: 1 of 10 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/MenuTimer.cs
```diff
@@ -38,7 +38,7 @@ public MenuTimer()
             }
             set
             {
-                Debug.WriteLineIf(ToolStrip.s_menuAutoExpandDebug.TraceVerbose && currentItem != value, "[MenuTimer.CurrentItem] changed: " + ((value is null) ? "null" : value.ToString()));
+                Debug.WriteLineIf(ToolStrip.s_menuAutoExpandDebug!.TraceVerbose && currentItem != value, "[MenuTimer.CurrentItem] changed: " + ((value is null) ? "null" : value.ToString()));
```
lonitra, line 41:
```suggestion
                Debug.WriteLineIf(ToolStrip.s_menuAutoExpandDebug!.TraceVerbose && currentItem != value, $"[MenuTimer.CurrentItem] changed: {(value is null ? "null" : value.ToString())}");
```

lonitra (commented):
One suggestion, otherwise LGTM!

lonitra (approved):

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
```diff
@@ -125,21 +124,21 @@ public partial class ToolStrip : ScrollableControl, IArrangedElement, ISupportTo
...
-        internal static readonly TraceSwitch s_layoutDebugSwitch;
-        internal static readonly TraceSwitch s_mouseActivateDebug;
-        internal static readonly TraceSwitch s_mergeDebug;
-        internal static readonly TraceSwitch s_snapFocusDebug;
-        internal static readonly TraceSwitch s_flickerDebug;
-        internal static readonly TraceSwitch s_itemReorderDebug;
-        internal static readonly TraceSwitch s_mdiMergeDebug;
-        internal static readonly TraceSwitch s_menuAutoExpandDebug;
-        internal static readonly TraceSwitch s_controlTabDebug;
+        internal static readonly TraceSwitch? s_selectionDebug;
```
dreddy-work, line 127: These are not used outside of DEBUG.  Any other way to avoid `!` usage?

RussKie: I've been thinking of this another day... An option would be `internal static readonly TraceSwitch s_selectionDebug = null!;`, but it means that if we were to accidentally use `s_selectionDebug` in the RELEASE code, it could result in NRE...

RussKie: We could also consider adding an extension method that would allow us writing something like this:
```cs
ToolStrip.s_snapFocusDebug.TraceVerbose("[ToolStripDropDown.OnMouseUp] mouse up outside of the toolstrip - this should dismiss the entire chain");
```
And the extension looking something like:
```cs
public static void TraceVerbose(this TraceSwitch? traceSwitch, string message)
{
    if (traceSwitch is not null && traceSwitch.TraceVerbose)
         Debug.WriteLine(message);
}
```

RussKie (approved):

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
gpetrou: I could probably add this in a separate PR.

RussKie: That'd be great, though we should do it across the codebase (at least the annotated portion of it).

PR #8081, opened 2022-10-31
Removed a duplicate WIN32_ERROR from  NativeMethods.txt 

Tanya-Solyanik:
Removed a duplicate instance of WIN32_ERROR line in NativeMethods.txt file.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8081)


dreddy-work (approved):

elachlan:
Should we also order the file as well? Or is there a method to the randomness?

dreddy-work:
> Should we also order the file as well? Or is there a method to the randomness?

We should order the file. I remember we are following the order. This seems to be one off. ?

PR #7956, opened 2022-10-17
Update Anchor Layout to support high DPI machines.

dreddy-work:
We have multiple [issues](https://github.com/dotnet/winforms/issues?q=is%3Aissue+is%3Aopen+anchor+-label%3Atest-bug+-label%3A%22%3Aconstruction%3A+work+in+progress%22+-label%3Aapi-suggestion+-label%3Aapi-approved+-label%3Atenet-localization) reported in WinForms around the anchor layout being problematic on higher DPI scale monitors, irrespective of application DPI mode. This document outlines the changes being made in .NET 8.0 to address these issues along with setting the goal to support all supported application DPI modes in WinForms.

**Problem in Scope:**
	Anchored control’s position with respect to its parent should be able to determine at the design time and would only need to be changed if there were explicit changes in the control’s Bounds or when the control is scaled in response to a DPI changed event. Bounds changes as result of Parent’s bounds change shouldn’t alter control’s relative position in the parent’s rectangle. However, layout in WinForms computes the anchored control’s position every time there are changes to control’s bounds or control’s property change that may impact its position. This is leading to the [issues](https://github.com/dotnet/winforms/issues?q=is%3Aissue+is%3Aopen+anchor+-label%3Atest-bug+-label%3A%22%3Aconstruction%3A+work+in+progress%22+-label%3Aapi-suggestion+-label%3Aapi-approved+-label%3Atenet-localization) we have been seeing. The following is a source snippet that is being serialized in WinForms designer and added comments show the various events that trigger anchor computations and why they could be wrong or unnecessary.

```CS
this.button14.Anchor = (System.Windows.Forms.AnchorStyles.Bottom | System.Windows.Forms.AnchorStyles.Left) // Layout tries to compute anchors with default button14 size and without parent. 
this.button14.Location = new System.Drawing.Point(431, 110); // Layout updates computed anchors for this new bounds, still with default size and no parent.
this.button14.Name = "button14";
this.button14.Size = new System.Drawing.Size(111, 23); // Layout updates computed anchors for this new bounds, still with no parent.
this.Controls.Add(this.button14); // Layout updates computed anchors with parents default size. Which may still be wrong.
this.Size = new System.Drawing.Size(828, 146); // Layout updates computed anchors with parents new size but this size could change based on the Dpi when parent Handle is created.
this.ResumeLayout(false) // Layout updates computed anchors with parents size. Dpi of the screen is still not considered (Handle creations or delayed until control is visible)

```

[...]

PR #8086, opened 2022-11-01
Fix Anchors in PermonitorV2 mode applications.

dreddy-work:
fixes: #8087

In PermonitorV2 mode applications, when moving the Form from one monitor to the other, Form gets an invalid display rectangle for a brief period during the scaling of its child controls. Updating anchors, because of layout, during this period would end up invalid anchor calculations. For all scaling scenarios due to DPI changed event, we will be scaling anchors and avoid/skip recalculation of anchors to prevent using this invalid display rect.

Following is the snippet where we see invalid DisplayRect of the Form for a brief period while moving from one monitor to the other.

**Invalid state:**
image

**Final state:**

image

Amending #7956 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8086)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
@@ -48,8 +48,10 @@ namespace System.Windows.Forms
         /// </summary>
         private bool _isScaledByDpiChangedEvent;
 
-        // Indicates scaling, due to DPI changed event, of the container control is in progress.
-        internal bool ContainersDpiScalingInProgress;
+        /// <summary>
+        ///  Indicates scaling, due to DPI changed event, of the container control is in progress.
+        /// </summary>
+        internal bool _dpiScalingInProgress;
 
         private BitVector32 _state;
 
@@ -1444,7 +1446,7 @@ namespace System.Windows.Forms
             SuspendAllLayout(this);
             try
             {
-                ContainersDpiScalingInProgress = true;
+                _dpiScalingInProgress = true;
 
                 if (LocalAppContextSwitches.ScaleTopLevelFormMinMaxSizeForDpi)
                 {
@@ -1498,7 +1500,7 @@ namespace System.Windows.Forms
                 _isScaledByDpiChangedEvent = false;
 
                 // Scaling and ResumeLayout, due to DPI changed event, should be finished by now for this container.
-                ContainersDpiScalingInProgress = false;
+                _dpiScalingInProgress = false;
             }
         }
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
@@ -48,6 +48,9 @@ namespace System.Windows.Forms
         /// </summary>
         private bool _isScaledByDpiChangedEvent;
 
+        // Indicates scaling, due to DPI changed event, of the container control is in progress.
+        internal bool ContainersDpiScalingInProgress;
+
         private BitVector32 _state;
 
         /// <summary>
@@ -1441,6 +1444,8 @@ namespace System.Windows.Forms
             SuspendAllLayout(this);
             try
             {
+                ContainersDpiScalingInProgress = true;
+
                 if (LocalAppContextSwitches.ScaleTopLevelFormMinMaxSizeForDpi)
                 {
                     // The suggested rectangle comes from Windows, and it does not match with our calculations for scaling controls by AutoscaleFactor.
@@ -1491,6 +1496,9 @@ namespace System.Windows.Forms
                 // We want to perform layout for dpi-changed high Dpi improvements - setting the second parameter to 'true'
                 ResumeAllLayout(this, true);
                 _isScaledByDpiChangedEvent = false;
+
+                // Scaling and ResumeLayout, due to DPI changed event, should be finished by now for this container.
+                ContainersDpiScalingInProgress = false;
             }
         }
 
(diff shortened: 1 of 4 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
```diff
@@ -48,6 +48,9 @@ public class ContainerControl : ScrollableControl, IContainerControl
         /// </summary>
         private bool _isScaledByDpiChangedEvent;
 
+        // Indicates scaling, due to DPI changed event, of the container control is in progress.
```
Tanya-Solyanik, line 51: Please use XML doc comments

src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
```diff
@@ -48,6 +48,9 @@ public class ContainerControl : ScrollableControl, IContainerControl
...
+        internal bool ContainersDpiScalingInProgress;
```
Tanya-Solyanik, line 52: nit: you can remove the word `Container` because this is the ContainerControl class, so container is implied.

src/System.Windows.Forms/src/System/Windows/Forms/Layout/DefaultLayout.cs
```diff
@@ -896,6 +896,12 @@ internal static void UpdateAnchorInfoV2(Control control)
                 return;
             }
 
+            // Anchors are scaled for the new DPI and doesn't need to be recalculated.
```
Tanya-Solyanik, line 899: The container is being scaled, not the anchors. Maybe say something like this:
> Delay anchor calculation until container is rescaled.

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -10589,9 +10589,6 @@ protected virtual void ScaleControl(SizeF factor, BoundsSpecified specified)
             scaledSize = LayoutUtils.UnionSizes(scaledSize, minSize);
 
             if (DpiHelper.IsScalingRequirementMet
-                // In the v2 layout, anchors are updated/computed after the controls bounds changed
-                // and, thus, don't need scaling.
-                && !DefaultLayout.UseAnchorLayoutV2(this)
```
Tanya-Solyanik, line 10594: Had youremeasured how many anchor calculations we are getting rid of now?

src/System.Windows.Forms/src/System/Windows/Forms/Layout/DefaultLayout.cs
```diff
@@ -913,6 +919,27 @@ internal static void UpdateAnchorInfoV2(Control control)
 
             anchorInfo.Right = displayRect.Width - (x + elementBounds.Width);
             anchorInfo.Bottom = displayRect.Height - (y + elementBounds.Height);
+
+            // Walk through parent hierarchy and check if scaling due to DPI change is in progress.
```
Tanya-Solyanik, line 923: Please use XML doc comments.

src/System.Windows.Forms/src/System/Windows/Forms/Layout/DefaultLayout.cs
```diff
@@ -913,6 +919,27 @@ internal static void UpdateAnchorInfoV2(Control control)
...
+            static bool IsDPIScalingInProgress(Control control, Control parent)
+            {
+                if (control.IsCurrentlyBeingScaled)
+                {
+                    return true;
+                }
+
+                while (parent is not null)
+                {
+                    if (parent is ContainerControl container && container.ContainersDpiScalingInProgress)
```
Tanya-Solyanik, line 933: `ContainerDpiScalingInProgress` and `IsCurrentlyBeingScaled` are similar. One is set only when scaling for DPI and the other during any scaling. Could you think of similar-sounding names for them? For example:
`ScalingInProgress`
`DpiScalingInProgress`

Tanya-Solyanik (commented):
Looks good.
How does this change affect your perf results?
Had you tested this change in apps where the anchored control is a container itself?

src/System.Windows.Forms/src/System/Windows/Forms/Layout/DefaultLayout.cs
```diff
@@ -913,6 +919,27 @@ internal static void UpdateAnchorInfoV2(Control control)
 
             anchorInfo.Right = displayRect.Width - (x + elementBounds.Width);
             anchorInfo.Bottom = displayRect.Height - (y + elementBounds.Height);
+
+            // Walk through parent hierarchy and check if scaling due to DPI change is in progress.
+            static bool IsDPIScalingInProgress(Control control, Control parent)
```
RussKie, line 924: capitalisation
```suggestion
            static bool IsDpiScalingInProgress(Control control, Control parent)
```

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
dreddy-work: I see at least 5 computations reduced now for each control anchored in the Form.

src/System.Windows.Forms/src/System/Windows/Forms/Layout/DefaultLayout.cs
dreddy-work: Rephrased but wanted to keep why we skip. `We already scaled for the new DPI.`. I think that is sufficient here.

dreddy-work:
> Had you tested this change in apps where the anchored control is a container itself?

Yes. Form with Panel with anchored controls. And panel is anchored to Form. Works as expected.

src/System.Windows.Forms/src/System/Windows/Forms/Layout/DefaultLayout.cs
dreddy-work: Cannot use XML comment on local static methods.

Tanya-Solyanik (approved):

PR #8091, opened 2022-11-01
Update branding to 6.0.12

vseanreesermsft:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8091)

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <MajorVersion>6</MajorVersion>
     <MinorVersion>0</MinorVersion>
-    <PatchVersion>11</PatchVersion>
+    <PatchVersion>12</PatchVersion>
     <!-- version in our package name #.#.#-below.#####.## -->
     <PreReleaseVersionLabel>servicing</PreReleaseVersionLabel>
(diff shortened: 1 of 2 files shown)


dreddy-work:
cmake version being used is not compatible with VS 2019?  @vseanreesermsft , do you see similar failure in any other repo?

RussKie:
@jonfortescue @MattGal any guidance?

jonfortescue:
looks like the cmake version is pinned in the branch that this is building from (see [here](https://github.com/vseanreesermsft/winforms/blob/branding-6.0.12-2022-11-01-1420/global.json#L24)). That needs to be changed to `latest` or `3`; we took cmake 3.22.6 in October which means 3.22.5 is no longer available on the machines.

dreddy-work:
> looks like the cmake version is pinned in the branch that this is building from (see [here](https://github.com/vseanreesermsft/winforms/blob/branding-6.0.12-2022-11-01-1420/global.json#L24)). That needs to be changed to `latest` or `3`; we took cmake 3.22.6 in October which means 3.22.5 is no longer available on the machines.

@jonfortescue , do you mind pushing that change? Seems i do not have permissions to push your branch.

dreddy-work (approved):

PR #8096, opened 2022-11-02
Replace Hashtable with generic dictionary in ToolStrip

gpetrou:
## Proposed changes

- Replace Hashtable with generic dictionary in ToolStrip.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8096)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
@@ -2,7 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System.Collections;
 using System.ComponentModel;
 using System.ComponentModel.Design.Serialization;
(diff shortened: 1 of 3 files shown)


elachlan:
src\System.Windows.Forms\src\System\Windows\Forms\ToolStrip.cs(5,1): error IDE0005: (NETCORE_ENGINEERING_TELEMETRY=Build) Using directive is unnecessary.

elachlan:
Should we be doing this everywhere that isn't a public API?

RussKie (approved):

RussKie:
> Should we be doing this everywhere that isn't a public API?

We can certainly entertain such possibilities :) 
Though as past experience shows, and @gpetrou can confirm, we need to be careful when going from untyped to strongly typed collections - some API may look the same, but they don't always work the same. Automated tests help reducing the risk, obviously. But there are still many places where we don't have test coverage.

PR #8105, opened 2022-11-02
Use IStorage*/IStream* Instead of Interface

lonitra:
- Use IStream*/IStorage* instead of interface where appropriate
- Convert `CreateILockBytesOnHGlobal`,`GetHGlobalFromILockBytes`,`StgCreateDocfileOnILockBytes`,`StgOpenStorageOnILockBytes` to cswin32
- Remove LockBytesWrapper
- Implement basic dispose pattern in AxHost.State to release unmanaged resources
- Update IPersistStorage::Load/Save methods to return correct HRESULT value

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8105)


src/System.Windows.Forms.Design/src/System/ComponentModel/Design/MultilineStringEditor.OleCallback.cs
```diff
@@ -24,19 +25,24 @@ private static TraceSwitch RichTextDebug
...
+                {
+                    HRESULT result = PInvoke.StgCreateDocfileOnILockBytes(
                     pLockBytes,
-                    Ole32.STGM.SHARE_EXCLUSIVE | Ole32.STGM.CREATE | Ole32.STGM.READWRITE);
+                    STGM.STGM_SHARE_EXCLUSIVE | STGM.STGM_CREATE | STGM.STGM_READWRITE,
+                    reserved: 0,
+                    pStorage);
 
-                Debug.Assert(storage is not null, "storage is NULL!");
+                    Debug.Assert(result.Succeeded, "storage is NULL!");
```
Tanya-Solyanik, line 42: Should the error message be updated?

JeremyKuhne (approved):
Nice!

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/MultilineStringEditor.OleCallback.cs
lonitra: This error message should still be accurate. Storage is the newly created storage object from `StgCreateDocfileOnILockBytes()` so if that operation failed, storage should've been set to null.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.State.cs
```diff
@@ -376,6 +375,33 @@ void ISerializable.GetObjectData(SerializationInfo si, StreamingContext context)
...
+            protected virtual void Dispose(bool disposing)
+            {
+                ReleaseResources();
+            }
+
+            public void Dispose()
+            {
+                Dispose(true);
+                GC.SuppressFinalize(this);
+            }
```
RussKie, line 404: Do we also need a finalizer?

lonitra: I talked with Jeremy Barton and it was recommended that public types should not have a finalizer.

RussKie: ...but if a consumer, who (according to the doc above) read an instance of this object via `getOcxState` or by reading in from a stream, would not call `Dispose` - we'd leak native resources (i.e., `ReleaseResources` wouldn't get called).
Or am I missing something?

RussKie: /cc: @JeremyKuhne @bartonjs

JeremyKuhne: Yes we do need a finalizer here as we have native resources and this was never disposed of before. I glossed over there not being a finalizer.

bartonjs: The official guidance/recommendation is to encapsulate the resource(s) that require finalization into a SafeHandle type, and not add a finalizer to the public type.

If that's unworkable for some reason, then, yes, add a finalizer here, which should only have a body of `Dispose(false)`.

lonitra: Ah, I misunderstood here. Thank you for clarifying!

RussKie (commented):
👍

RussKie (approved):

RussKie:
Looks like tests failing across multiple agents all with the same error, e.g.: https://dev.azure.com/dnceng-public/public/_build/results?buildId=77220&view=results (https://github.com/dotnet/winforms/pull/8142)

```
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at Windows.Win32.System.Com.IStream.Release() in /_/src/System.Windows.Forms.Primitives/src/Microsoft.Windows.CsWin32/Microsoft.Windows.CsWin32.SourceGenerator/Windows.Win32.IStream.g.cs:line 55
   at System.Windows.Forms.DataStreamFromComStream.Dispose(Boolean disposing) in /_/src/System.Windows.Forms/src/System/Windows/Forms/DataStreamFromComStream.cs:line 158
   at System.Windows.Forms.DataStreamFromComStream.Finalize() in /_/src/System.Windows.Forms/src/System/Windows/Forms/DataStreamFromComStream.cs:line 167
```

image

src/System.Windows.Forms/src/System/Windows/Forms/DataStreamFromComStream.cs
```diff
@@ -171,20 +147,19 @@ public override unsafe void Write(ReadOnlySpan<byte> buffer)
...
+                    _comStream->Commit(STGC.STGC_DEFAULT);
                 }
 
                 // Can't release a COM stream from the finalizer thread.
-                _comStream = null!;
-            }
-            finally
-            {
-                base.Dispose(disposing);
+                _comStream->Release();
```
RussKie, line 158: ❗  This is causing the NRE in the finalizer. The comment above states that.

PR #8113, opened 2022-11-04
CA1838 Avoid 'StringBuilder' parameters for P/Invokes

### LINKED ISSUE #3983: Review use of `CA1838` warning - Avoid 'StringBuilder' parameters for P/Invokes
To update the underlying SDK/toolsets to 5.0-RC1 we had to suppress `A1838 // Avoid 'StringBuilder' parameters for P/Invokes` in https://github.com/dotnet/winforms/pull/3959/commits/9dfaaa152b889f2040f646697dfb31c77d7058e8.

https://github.com/dotnet/winforms/search?q=CA1838

The suppression needs to be reviewed, and corrected.

elachlan:
Fixes #3983

Analyzer:
- https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1838

Work based on:
- https://github.com/dotnet/msbuild/pull/7186
- #4751

TODO:
https://github.com/dotnet/winforms/blob/15ab89b1017e0da7e7448118b3d0a80632c9f146/src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/Helpers/NativeMethods.vb#L30-L33

Related: #7887

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8113)


src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.MAX_PATH.cs
```diff
@@ -2,10 +2,65 @@
...
+            }
+        }
+
+        private static void SetMaxPath()
+        {
+            lock (MaxPathLock)
+            {
+                if (!IsMaxPathSet)
+                {
+                    bool isMaxPathRestricted = IsLongPathsEnabledRegistry();
```
elachlan, line 45: In MSBuild there is an escape hatch for long paths. I am unsure if winforms needs an escape hatch here.
https://github.com/dotnet/msbuild/blob/207f71329700f2b178a832bb1f0930c107fda0e9/src/Framework/NativeMethods.cs#L600-L611

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -2556,22 +2555,19 @@ internal bool HostedInWin32DialogManager
...
-                                sb.Capacity = len + 5;
+                                length = PInvoke.GetClassName(lastParentHandle, lpClassName, buffer.Length);
                             }
 
-                            UnsafeNativeMethods.GetClassName(new HandleRef(null, lastParentHandle), sb, sb.Capacity);
-
-                            if (sb.ToString() == "#32770")
+                            // class name #32770
+                            ReadOnlySpan<char> className = "#32770";
+                            if (MemoryExtensions.Equals(className, buffer.SliceAtFirstNull(), StringComparison.Ordinal))
```
elachlan, line 2570: The idea here was to avoid a string allocation. But I am unsure if this is correct.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RichTextBoxTests.cs
```diff
@@ -10791,12 +10791,15 @@ public new bool ResizeRedraw
             public new void WndProc(ref Message m) => base.WndProc(ref m);
         }
 
-        private static string GetClassName(IntPtr hWnd)
+        private static unsafe string GetClassName(HWND hWnd)
```
elachlan, line 10794: should this move to the PInvoke namespace?

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
RussKie: In the tests we aren't as strict about memory allocations (though we shouldn't go bonkers either :))

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.MAX_CLASS_NAME.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Windows.Win32
+{
+    internal static partial class PInvoke
+    {
+        public const int MAX_CLASS_NAME = 256;
```
RussKie, line 9: How is this used?

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
RussKie: Oh, this is actually the shippable code... 👀

RussKie: Oh, this is actually the shippable code... 👀

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.MAX_CLASS_NAME.cs
elachlan: RichTextBoxTests.cs `GetClassName`

elachlan:
I don't think the vb.net PInvokes are able to avoid stringbuilder. We would have to pull it into the PInvoke class and then have a wrapper that returns the needed information.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.MAX_CLASS_NAME.cs
RussKie: If it's test-specific, then it should reside in the test part of the solution.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.MAX_PATH.cs
RussKie: Tbh, I'm not currently convinced we need this... I'll leave it to @JeremyKuhne and @lonitra to decide though.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.MAX_CLASS_NAME.cs
elachlan: Do we want to centralise the PInvokes though? It would reduce the chance of code duplication.

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
lonitra: This LGTM! Just a nit that this can be rewritten as 
`className.Equals(buffer.SliceAtFirstNull(), StringComparison.Ordinal)`

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.MAX_PATH.cs
lonitra: I also don't see that it is needed, but waiting for Jeremy's thoughts.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.MAX_PATH.cs
```diff
@@ -2,10 +2,65 @@
...
+                    _maxPath = isMaxPathRestricted ? MAX_PATH : int.MaxValue;
+                    IsMaxPathSet = true;
+                }
+            }
+        }
+
+        private static bool IsLongPathsEnabledRegistry()
```
JeremyKuhne, line 52: This has nothing to do with whether or not long paths can be used in general, this is whether or not you can opt-in to using long paths without using `\\?\` as a path prefix. You also have to actually have a proper manifest for the current executable. The only way to really know is to call an undocumented API. In any case, there is no need to do this.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.MAX_PATH.cs
```diff
@@ -2,10 +2,65 @@
...
+        private static int _maxPath;
+        private static bool IsMaxPathSet { get; set; }
+        private static readonly object MaxPathLock = new object();
+        internal static bool HasMaxPath => MaxPath == MAX_PATH;
         public const int MAX_PATH = 260;
+
+        /// <summary>
+        /// Gets the windows max path limit.
+        /// </summary>
+        internal static int MaxPath
```
JeremyKuhne, line 26: This is not going to be correct and you shouldn't try to implement it. Fundamentally the maximum path is the size of UNICODE_STRING, which is a ushort of bytes in length. Even knowing that isn't sufficient as the path gets manipulated as it passes through the system to whichever driver is involved. You can think of the path max as being roughly 32K characters, but that is just the ballpark. It will never be over, however.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleFileNameLongPath.cs
```diff
@@ -0,0 +1,48 @@
...
+{
+    internal static partial class PInvoke
+    {
+        public static unsafe string GetModuleFileNameLongPath(HINSTANCE hModule)
+        {
+            const int INSUFFICIENT_BUFFER = 0x007A;
+            char[] buffer;
+
+            // Try increased buffer sizes if on longpath-enabled Windows
+            for (int bufferSize = MAX_PATH; bufferSize <= MaxPath; bufferSize *= 2)
```
JeremyKuhne, line 19: I wouldn't double as it is a bit too aggressive for those that would have more common long paths. You also won't be correct here as you'll never actually include the longest paths by doubling this way (presuming you had MaxPath defined correctly, but you don't). If performance was super critical here you could:

1. Attempt a stackalloc buffer of MAX_PATH
2. Fall back to a loop of ArrayPool buffers, starting at 4096 characters and adding 4096 every time until you pass short.MaxValue.

But you're probably just fine doing `4096 * i` buffer rents from the start here. We don't call this that often and the 4K buffer will get reused for sure. I'd actually recommend it as it is simpler, logic-wise.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleFileNameLongPath.cs
```diff
@@ -0,0 +1,48 @@
...
+            {
+                buffer = ArrayPool<char>.Shared.Rent(bufferSize);
+                try
```
JeremyKuhne, line 22: You don't need a try block here. You shouldn't be throwing out of here. Generally speaking, if exceptions are expected to be rare it is better to drop an ArrayPool buffer than take the hit of the try block.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleFileNameLongPath.cs
```diff
@@ -0,0 +1,48 @@
...
+                {
+                    uint pathLength;
+                    fixed (char* lpFilename = buffer)
+                    {
+                        pathLength = GetModuleFileName(hModule, lpFilename, (uint)bufferSize);
+                    }
+
+                    bool isBufferTooSmall = Marshal.GetLastWin32Error() == INSUFFICIENT_BUFFER;
+                    if (pathLength != 0 && !isBufferTooSmall && bufferSize <= int.MaxValue)
```
JeremyKuhne, line 31: `bufferSize <= int.MaxValue` is _always_ true.

JeremyKuhne, line 31: The right check is `(pathLength > 0 && (pathLength < buffer.Length || Marshal.GetLastWin32Error() != INSUFFICENT_BUFFER))`

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleFileNameLongPath.cs
```diff
@@ -0,0 +1,48 @@
...
+                    {
```
JeremyKuhne, line 32: You'll keep increasing the size of the buffer in the error case (returns a length of 0).

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleFileNameLongPath.cs
```diff
@@ -0,0 +1,48 @@
...
+            // Try increased buffer sizes if on longpath-enabled Windows
+            for (int bufferSize = MAX_PATH; bufferSize <= MaxPath; bufferSize *= 2)
+            {
+                buffer = ArrayPool<char>.Shared.Rent(bufferSize);
+                try
+                {
+                    uint pathLength;
+                    fixed (char* lpFilename = buffer)
+                    {
+                        pathLength = GetModuleFileName(hModule, lpFilename, (uint)bufferSize);
```
JeremyKuhne, line 27: Use `buffer.Length`, never an intermediary value.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleFileNameLongPath.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+
+namespace Windows.Win32
+{
+    internal static partial class PInvoke
+    {
+        public static unsafe string GetModuleFileNameLongPath(HINSTANCE hModule)
+        {
+            const int INSUFFICIENT_BUFFER = 0x007A;
```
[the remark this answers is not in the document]
JeremyKuhne: This is a standard error so we should be pulling it from CSWin32. `WIN32_ERROR.ERROR_INSUFFICENT_BUFFER`

src/System.Windows.Forms/src/System/Windows/Forms/Application.cs
```diff
@@ -292,12 +292,7 @@ public static string ExecutablePath
         {
             get
             {
-                if (s_executablePath is null)
-                {
-                    StringBuilder sb = UnsafeNativeMethods.GetModuleFileNameLongPath(NativeMethods.NullHandleRef);
-                    s_executablePath = Path.GetFullPath(sb.ToString());
-                }
-
+                s_executablePath ??= Path.GetFullPath(PInvoke.GetModuleFileNameLongPath(default));
```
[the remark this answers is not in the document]
JeremyKuhne: Calling `GetFullPath` will do nothing here as the API claims it always returns a fully qualified path.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RichTextBoxTests.cs
```diff
@@ -10791,12 +10791,15 @@ public new bool ResizeRedraw
...
         {
-            const int MaxClassName = 256;
-            StringBuilder sb = new StringBuilder(MaxClassName);
-            UnsafeNativeMethods.GetClassName(new HandleRef(null, hWnd), sb, MaxClassName);
-            return sb.ToString();
+            Span<char> buf = stackalloc char[PInvoke.MAX_CLASS_NAME];
```
JeremyKuhne, line 10796: Don't abbreviate here- use `buffer`

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RichTextBoxTests.cs
```diff
@@ -10791,12 +10791,15 @@ public new bool ResizeRedraw
...
+            fixed (char* lpClassName = buf)
+            {
+                _ = PInvoke.GetClassName(hWnd, lpClassName, buf.Length);
```
JeremyKuhne, line 10799: This API is clearly documented to return the count of characters. We should capture that and slice at that length, not scan for the null.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RichTextBoxTests.cs
```diff
@@ -10584,7 +10584,7 @@ public void RichTextBox_CheckDefaultNativeControlVersions()
             using var control = new RichTextBox();
             control.CreateControl();
 
-            Assert.Contains("RICHEDIT50W", GetClassName(control.Handle), StringComparison.InvariantCultureIgnoreCase);
+            Assert.Contains("RICHEDIT50W", GetClassName(control.HWND), StringComparison.InvariantCultureIgnoreCase);
```
JeremyKuhne, line 10587: The comparison here should be ordinal (best) or ordinal ignore case. Class names don't have a culture, they're an ID.

JeremyKuhne (changes requested):
Left a number of comments. Biggest thing is not to try and guess what the exact max allowable path will be (outside of constraining to not go over 32K characters for sanity).

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RichTextBoxTests.cs
elachlan: I didn't move this because its a wrapper and isn't used in the other calls. I am happy to still move it. The other uses avoid string allocations.

elachlan:
Any advice on the VB PInvokes? Should I write a wrapper in C# and call it from the VB code? I think that is most efficient to avoid a string builder.
https://github.com/dotnet/winforms/blob/15ab89b1017e0da7e7448118b3d0a80632c9f146/src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/Helpers/NativeMethods.vb#L30-L33

elachlan:
@AArnott Would you want GetModuleFileNameLongPath in CsWin32 as a friendly wrapper/overloads?

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleFileNameLongPath.cs
```diff
@@ -0,0 +1,42 @@
...
+            for (int i = 1; bufferSize <= short.MaxValue; i++, bufferSize = 4096 * i)
+            {
+                buffer = ArrayPool<char>.Shared.Rent(bufferSize);
+                uint pathLength;
+                fixed (char* lpFilename = buffer)
+                {
+                    pathLength = GetModuleFileName(hModule, lpFilename, (uint)buffer.Length);
+                }
+
+                if (pathLength > 0 && (pathLength < buffer.Length || (WIN32_ERROR)Marshal.GetLastWin32Error() == WIN32_ERROR.ERROR_INSUFFICIENT_BUFFER))
```
JeremyKuhne, line 27: You're still not exiting out in the error case ('0')

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleFileNameLongPath.cs
```diff
@@ -0,0 +1,42 @@
...
+                {
+                    return new string(buffer, 0, (int)pathLength);
+                }
+
+                // Return to array pool
+                ArrayPool<char>.Shared.Return(buffer);
```
JeremyKuhne, line 33: You need to be returning this to the pool when you succeed. It is ok to not return them in super rare cases, but you defeat the purpose of using the pool if you don't normally return them.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleFileNameLongPath.cs
```diff
@@ -0,0 +1,42 @@
...
+
+                // Double check that the buffer is not insanely big
+                Debug.Assert(bufferSize <= int.MaxValue / 2, "Buffer size approaching int.MaxValue");
```
JeremyKuhne, line 36: This assert isn't useful, I'd just remove it. Additionally, in its current state it will always be true. If we thought large paths were interesting to know about in our debug builds we could put a `Debug.WriteLine` in, but I don't think that it actually adds any real value.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.MAX_CLASS_NAME.cs
JeremyKuhne, line 9: I don't think this is a real Windows define. Presuming it isn't we should use normal casing `MaxClassName` and document where this number comes from. The simplest way is probably to `<see cref>` the `WNDCLASS.lpszClassName` definition as it should have the docs on it. (Or WNDCLASSEX, I'm not remembering off the top of my head what we actually use.)

src/System.Windows.Forms/src/System/Windows/Forms/Application.cs
```diff
@@ -292,12 +292,7 @@ public static string ExecutablePath
         {
             get
             {
-                if (s_executablePath is null)
-                {
-                    StringBuilder sb = UnsafeNativeMethods.GetModuleFileNameLongPath(NativeMethods.NullHandleRef);
-                    s_executablePath = Path.GetFullPath(sb.ToString());
-                }
-
+                s_executablePath ??= Path.GetFullPath(PInvoke.GetModuleFileNameLongPath(HINSTANCE.Null));
```
JeremyKuhne, line 295: nit: combine with the next line and make this `ExecutablePath => s_...`  Also, no need for `Path.GetFullPath` as I mentioned before.

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -2556,22 +2555,19 @@ internal bool HostedInWin32DialogManager
...
+                            fixed (char* lpClassName = buffer)
                             {
-                                sb.Capacity = len + 5;
+                                length = PInvoke.GetClassName(lastParentHandle, lpClassName, buffer.Length);
                             }
 
-                            UnsafeNativeMethods.GetClassName(new HandleRef(null, lastParentHandle), sb, sb.Capacity);
-
-                            if (sb.ToString() == "#32770")
+                            // class name #32770
```
JeremyKuhne, line 2568: This comment isn't useful. What would be is specifying that this is the class name for dialogs and linking to https://learn.microsoft.com/windows/win32/winmsg/about-window-classes#system-classes.

JeremyKuhne (changes requested):
Added a few more comments

JeremyKuhne:
> Any advice on the VB PInvokes? Should I write a wrapper in C# and call it from the VB code? I think that is most efficient to avoid a string builder.

I would presume it is better to keep this in VB. @KlausLoeffelmann do you have any opinions here?

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.MAX_CLASS_NAME.cs
elachlan: https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-wndclassexw

> The maximum length for lpszClassName is 256. If lpszClassName is greater than the maximum length, the [RegisterClassEx](https://learn.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-registerclassexa) function will fail.

I was unable to find a related constant in the metadata project in `winuser.h`.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleFileNameLongPath.cs
```diff
@@ -0,0 +1,47 @@
...
+    internal static partial class PInvoke
+    {
+        public static unsafe string GetModuleFileNameLongPath(HINSTANCE hModule)
+        {
+            char[] buffer;
+            int bufferSize = 4096;
+            // Allocate increasingly larger portions of memory until successful or we hit short.maxvalue.
+            for (int i = 1; bufferSize <= short.MaxValue; i++, bufferSize = 4096 * i)
+            {
+                buffer = ArrayPool<char>.Shared.Rent(bufferSize);
```
AArnott, line 19: Array pools can return longer arrays than requested. If for example it gave you 16KB instead of 4KB, you may as well skip the 8, 12, and 16KB attempts that you would otherwise try. 
But that makes the logic more complex, and unless it's thoroughly tested, that could lead to bugs.
As an alternative, considering a reasonable max size like 32KB, you could use stackalloc for the whole 32KB and lose the arraypool and loop altogether. Super simple and better performing.

AArnott:
> Would you want GetModuleFileNameLongPath in CsWin32 as a friendly wrapper/overloads?

@elachlan not at this point. The implementation has to be opinionated because there are tradeoffs to be made, so unless there was evidence of some consensus around which implementation was best 90% of the time, I'd say it's better for folks to write their own (as you do here).

AArnott:
> I don't think the vb.net PInvokes are able to avoid stringbuilder. We would have to pull it into the PInvoke class and then have a wrapper that returns the needed information.

VB.NET should be able to invoke IntPtr methods after allocating memory using the `Marshal` class, then again use that class to convert the data into a `String`. It certainly makes things harder to not have pointers, ironically.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleFileNameLongPath.cs
elachlan: In dotnet runtime they try to keep stackalocs to 1024 or less. Is there a general rule to avoid an overflow?

elachlan: I managed to find where I remembered reading it.
https://github.com/dotnet/msbuild/pull/7186#discussion_r796202803

I think its more, is the  stackalloc likely to push the stack over the standard 1mb windows stack limit. Causing a stack overflow. So the guidance is to be as conservative as possible when using stackalloc.

AArnott: Sure, but you can also think about the scenario that your code will execute in. For the .NET runtime itself, almost anything in the BCL is fair game for an app to invoke anywhere on a stack, so they want to be conservative. This is especially important when the method may be called multiple times on a thread recursively, such that what you allocate may be multiplied. But at the app level, you can decide how likely your call stacks are to approach the 1MB (default) limit and make other decisions. In WinForms, if you have an idea about the scenario you're called from, such that you would know you'll likely be on a 'short' callstack, then a 32KB frame size on a scale of 1MB seems small, IMO. 

Alternatively, if you're going to use a heap array, which is totally fine, I would _not_ try 'n' times. I would try once or twice. If once, I would just borrow an array that is at least as large as the largest value I would consider allowable. 
If twice, I would borrow an array that should fit 95% of the cases (say, 500 bytes), and if that fails, retry with the largest allowable length to scoop up the last 5%. The loop approach is just getting the worst of all worlds, IMO. You'll be slow, all in the name of trying to be efficient by not borrowing a longer array, but borrowing an array is pretty cheap. And borrowing _several_ arrays in a loop is more expensive.

elachlan: Yeah, I feel the same way. I'll investigate the call stack to see how short it is. I like the idea of avoiding the heap. If we use is a loop is most likely a horrible idea compared to just asking for a bit more memory upfront.

AArnott: Another option: use `stackalloc` with 1KB, and fallback to a 32KB array from the pool if that fails. You get the best of all worlds that way.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.MaxClassName.cs
```diff
@@ -0,0 +1,15 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Windows.Win32
+{
+    internal static partial class PInvoke
+    {
+        /// <summary>
+        /// <para>The maximum length for lpszClassName is 256. If lpszClassName is greater than the maximum length, the RegisterClassEx function will fail.</para>
+        /// <para><see href="https://learn.microsoft.com/en-us/windows/win32/api/winuser/ns-winuser-wndclassexw#members">Read more on docs.microsoft.com</see>.</para>
```
RussKie, line 11:
```suggestion
        /// <para><see href="https://learn.microsoft.com/windows/win32/api/winuser/ns-winuser-wndclassexw#members">Read more on docs.microsoft.com</see>.</para>
```

https://review.learn.microsoft.com/help/platform/links-how-to?branch=main#locale-codes-in-external-links

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -2556,22 +2555,21 @@ internal bool HostedInWin32DialogManager
...
                             {
-                                sb.Capacity = len + 5;
+                                length = PInvoke.GetClassName(lastParentHandle, lpClassName, buffer.Length);
                             }
 
-                            UnsafeNativeMethods.GetClassName(new HandleRef(null, lastParentHandle), sb, sb.Capacity);
-
-                            if (sb.ToString() == "#32770")
+                            // #32770 is the standard windows dialog class name
+                            // https://learn.microsoft.com/en-au/windows/win32/winmsg/about-window-classes#system-classes
```
RussKie, line 2570: Ditto

src/System.Windows.Forms/src/System/Windows/Forms/TaskDialog.cs
```diff
@@ -777,9 +777,7 @@ internal void UpdateCaption(string? caption)
             // executable's name as title if the string is null or empty.
             if (TaskDialogPage.IsNativeStringNullOrEmpty(caption))
             {
-                caption = Path.GetFileName(
-                    UnsafeNativeMethods.GetModuleFileNameLongPath(NativeMethods.NullHandleRef)
-                                       .ToString());
+                caption = Path.GetFileName(PInvoke.GetModuleFileNameLongPath(HINSTANCE.Null));
```
elachlan, line 780: Instead of calling out to GetModuleFileNameLongPath, can we just use `AppDomain.CurrentDomain.FriendlyName`?
https://learn.microsoft.com/en-us/dotnet/api/system.appdomain.friendlyname?view=net-7.0

https://github.com/dotnet/runtime/blob/215b39abf947da7a40b0cb137eab4bceb24ad3e3/src/libraries/System.Private.CoreLib/src/System/AppDomain.cs#L52-L59

Or is that a lot slower of a code path?

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleFileNameLongPath.cs
JeremyKuhne: Yes, we definitely don't want to stack alloc more than 1KB and I generally tried to keep under .5K on the runtime team (well, up to `MAX_PATH` chars (520 bytes)). In this case, looping is fine, and you could double I suppose to accommodate the internals of `ArrayPool`, but I wouldn't bother stressing too much about it. 4K (of bytes or chars) are super common buffer sizes taken from the pool and would cover 99.999% of path cases. While you could have paths that are longer than 2,000 characters, it would be terrible practice. In Unix based implementations the most you're going to see is 4KB of bytes (UTF-8 typically), but if I recall correctly Linux is 1K? In any case, we shouldn't optimize for extreme cases. :)

When running as 32bit (which WinForms users do, sometimes out of necessity) large allocations can have more of an impact. If we grab 32K buffers from the `ArrayPool` while under memory pressure they are much more likely to get aggressively trimmed from the `ArrayPool`, which would cause a lot of extra work for cases we will almost never see.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleFileNameLongPath.cs
```diff
@@ -0,0 +1,47 @@
...
+                    pathLength = GetModuleFileName(hModule, lpFilename, (uint)buffer.Length);
+                }
+
+                if (pathLength == 0)
+                {
+                    ArrayPool<char>.Shared.Return(buffer);
+                    return string.Empty;
+                }
+
+                if (pathLength > 0 && (pathLength < buffer.Length || (WIN32_ERROR)Marshal.GetLastWin32Error() != WIN32_ERROR.ERROR_INSUFFICIENT_BUFFER))
```
JeremyKuhne, line 32: Probably need a comment here `If the length equals the buffer size we need to check to see if we were told the buffer was insufficient (it was trimmed).`.

Additionally, as it is a `uint` we don't need to check for `> 0`.  Finally, it should be `pathLength < buffer.Length - 1` (for the null).

KlausLoeffelmann:
> I would presume it is better to keep this in VB. @KlausLoeffelmann do you have any opinions here?

Not sure. Honestly, I am tending to go for a C# mapper. I guess it seems more consistent, and it might be even less error prone and more efficient.

elachlan:
The code has a `MAX_TITLE_LENGTH = 511`. Should I use this inside the PInvoke as the buffer constant? At the moment I have it written to call `GetWindowTextLength` and then either use `stackalloc` or `ArrayPool` depending on the size returned.
https://github.com/dotnet/winforms/blob/2479775c33fa9196e6717770e852b663aba283a7/src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/Interaction.vb#L136-L146

Secondly, VB is complaining because the `PInvoke` class is internal. How do I expose it to `Microsoft.VisualBasic.Forms`? Is it possible? Does `[assembly: InternalsVisibleTo("Microsoft.VisualBasic.Forms, PublicKey=00000000000000000400000000000000")]` in `AssemblyInfo.cs` work with VB.Net?

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleFileNameLongPath.cs
```diff
@@ -43,5 +49,29 @@ public static unsafe string GetModuleFileNameLongPath(HINSTANCE hModule)
...
+            }
+
+            if (pathLength == 0 || (WIN32_ERROR)Marshal.GetLastWin32Error() == WIN32_ERROR.ERROR_INSUFFICIENT_BUFFER)
+            {
+                return false;
+            }
+
+            if (pathLength < buffer.Length)
+            {
+                bufferOut = buffer[..(int)pathLength].ToArray();
```
AArnott, line 70: nit: You're allocating a `char[]` only to pass it to a `string` that is also allocated. What if we move the `stackalloc` to the caller and pass in `ref Span<char>` as a parameter instead. This method could slice the span to fit the length and the caller can then initialize the string based on the `Span<char>` without ever allocating a `char[]`.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleFileNameLongPath.cs
```diff
@@ -11,67 +11,57 @@ internal static partial class PInvoke
...
         public static unsafe string GetModuleFileNameLongPath(HINSTANCE hModule)
         {
-            if (TryGetModuleFileName(hModule, out char[] path))
+            Span<char> buffer = stackalloc char[MAX_PATH];
+            uint pathLength;
+            bool isBufferTooSmall = false;
+            fixed (char* lpFilename = buffer)
+            {
+                pathLength = GetModuleFileName(hModule, lpFilename, (uint)buffer.Length);
+                isBufferTooSmall = (WIN32_ERROR)Marshal.GetLastWin32Error() == WIN32_ERROR.ERROR_INSUFFICIENT_BUFFER;
```
AArnott, line 20: You should _not_ use `GetLastWin32Error` to determine *if* there was a problem. Per the docs:

> If the function succeeds, the return value is the length of the string that is copied to the buffer, in characters, not including the terminating null character. If the buffer is too small to hold the module name, the string is truncated to nSize characters including the terminating null character, the function returns nSize, and the function sets the last error to ERROR_INSUFFICIENT_BUFFER.

Note that in a success case, 'last error' is undefined. This is a common mistake in calling win32 APIs. You first must ascertain whether an error occurred, then if it has, you can call `GetLastWin32Error` to obtain the reason for failure. 
The bug here is that many Win32 APIs (including this one, we must assume based on the docs) do not call `SetLastError` in their success paths, so by calling `GetLastError` unconditionally here, you run the risk of getting an error code set by some earlier Win32 call, which may or may not randomly match the value you're checking for here.

Based on the docs, you can detect an error by comparing `pathLength == buffer.Length`. That can *only* be true in an error case, because in a success case, `pathLength < buffer.Length` since `pathLength` will exclude the null terminator.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleFileNameLongPath.cs
```diff
@@ -11,67 +11,57 @@ internal static partial class PInvoke
...
+                    ArrayPool<char>.Shared.Return(lbuffer);
                     return returnValue;
                 }
 
                 // buffer was too small, return to array pool.
-                ArrayPool<char>.Shared.Return(buffer);
+                ArrayPool<char>.Shared.Return(lbuffer);
             }
 
             return string.Empty;
```
AArnott, line 61: Is quietly returning an empty string here really the best thing to do here, or should we throw to call attention to the bizarre long path case?

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleFileNameLongPath.cs
```diff
@@ -11,67 +11,57 @@ internal static partial class PInvoke
...
                 if (pathLength == 0)
                 {
-                    ArrayPool<char>.Shared.Return(buffer);
+                    ArrayPool<char>.Shared.Return(lbuffer);
                     return string.Empty;
                 }
 
                 // If the length equals the buffer size we need to check to see if we were told the buffer was insufficient (it was trimmed)
-                if (pathLength < buffer.Length - 1 || (WIN32_ERROR)Marshal.GetLastWin32Error() != WIN32_ERROR.ERROR_INSUFFICIENT_BUFFER)
+                if (pathLength < lbuffer.Length - 1 && !isBufferTooSmall)
```
AArnott:
```suggestion
                if (pathLength < lbuffer.Length)
```

I don't know why you had the `- 1` as part of the expression. That looks like a bug.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleFileNameLongPath.cs
elachlan: Most usages will error out anyway with an empty string. So I guess it might make sense to throw here so its obvious what is happening.

elachlan: I will wait on the winforms team to advise what they want.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetWindowText.cs
```diff
@@ -0,0 +1,62 @@
...
+                // or if the window or control handle is invalid, the return value is zero
+                if (length == 0)
+                {
+                    return string.Empty;
+                }
+            }
+
+            // Use arraypool for larger than 1024 characters.
+            char[] lBuffer;
+            lBuffer = ArrayPool<char>.Shared.Rent(length + 1);
```
AArnott, line 41: You're missing a `return` above this point in the happy path on a non-empty window text that is 1KB or less. :)

AArnott (changes requested):

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetWindowText.cs
```diff
@@ -0,0 +1,62 @@
...
+            int pathLength;
+            fixed (char* lpString = lBuffer)
+            {
+                pathLength = GetWindowText(hWnd, (PWSTR)lpString, lBuffer.Length);
```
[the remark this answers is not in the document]
elachlan: That is why I asked about `MAX_TITLE_LENGTH`, If we use that, then the size doesn't matter because we will truncate it to the `MAX_TITLE_LENGTH`.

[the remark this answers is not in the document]
elachlan: Also, if the max length is used, then we can avoid the whole > 1024 scenario.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleFileNameLongPath.cs
elachlan: What exception would you use here?

elachlan: I used `InvalidOperationException`, its seems most appropriate.

src/System.Windows.Forms.Primitives/src/Properties/AssemblyInfo.cs
```diff
@@ -6,6 +6,7 @@
 
 [assembly: System.Runtime.InteropServices.ComVisible(false)]
 
+[assembly: InternalsVisibleTo("Microsoft.VisualBasic.Forms, PublicKey=00000000000000000400000000000000")]
```
elachlan, line 9: This should have worked but we are still getting build errors. I don't know why.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RichTextBoxTests.cs
RussKie: If you think this is a dup, feel free to remove this. I can't recall exact reasons why I copied it here.

src/System.Windows.Forms.Primitives/src/Properties/AssemblyInfo.cs
RussKie: The build is failing for other reasons:
```
##[error]src\System.Windows.Forms.Primitives\src\Windows\Win32\PInvoke.GetModuleFileNameLongPath.cs(6,1): error IDE0005: (NETCORE_ENGINEERING_TELEMETRY=Build) Using directive is unnecessary.
```

AArnott (approved):

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RichTextBoxTests.cs
elachlan: Since its a wrapper of the pinvoke and we don't reuse it, I think it can stay here.

src/System.Windows.Forms.Primitives/src/Properties/AssemblyInfo.cs
elachlan: Now it has the failures:
```
error BC30389: (NETCORE_ENGINEERING_TELEMETRY=Build) 'Windows.Win32.PInvoke' is not accessible in this context because it is 'Friend'.
```

RussKie: Microsoft.VisualBasic.Forms doesn't reference System.Windows.Forms.Primitives project directly.

RussKie: ...and I _personally_ don't think it should.

elachlan: It references `System.Windows.Forms`, which gives it a transitive reference to `System.Windows.Forms.Primitives` doesn't it?

AArnott: IVT is not transitive.

elachlan: @RussKie Did you want me to remove the `Microsoft.VisualBasic.Forms` changes and we can work out what to do for it in a separate PR?

RussKie: I defer the decision to @lonitra and @JeremyKuhne.

JeremyKuhne: @elachan it's probably better to do the VB stuff separately for clarity.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleFileNameLongPath.cs
```diff
@@ -0,0 +1,63 @@
...
+            {
+                pathLength = GetModuleFileName(hModule, lpFilename, (uint)buffer.Length);
+            }
+
+            if (pathLength == 0)
+            {
+                return string.Empty;
+            }
+
+            if (pathLength < buffer.Length)
```
JeremyKuhne, line 25: You need to check the error result if the `pathLength` is less than `buffer.Length - 1`. Remember that it will truncate at `buffer.Length - 1` if it can't fit. It is `-1` as it always null terminates. Same goes for below. Documentation is off here (and conflicting). I've looked at the actual implementation.

Try to test this code out by manually setting smaller buffer sizes to make sure your logic is working correctly.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleFileNameLongPath.cs
```diff
@@ -0,0 +1,63 @@
...
+
+using System.Buffers;
+
+namespace Windows.Win32
+{
+    internal static partial class PInvoke
+    {
+        public static unsafe string GetModuleFileNameLongPath(HINSTANCE hModule)
+        {
+            Span<char> buffer = stackalloc char[MAX_PATH];
```
JeremyKuhne, line 13: I'd call this `stackBuffer` and the other `buffer`

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleFileNameLongPath.cs
```diff
@@ -0,0 +1,63 @@
...
+                }
+
+                if (pathLength == 0)
+                {
+                    ArrayPool<char>.Shared.Return(lbuffer);
+                    return string.Empty;
+                }
+
+                // If the length equals the buffer size we need to check to see if we were told the buffer was insufficient (it was trimmed)
+                if (pathLength < lbuffer.Length)
```
JeremyKuhne, line 48: As above, this needs to be -1 and check for the buffer too small error.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleFileNameLongPath.cs
```diff
@@ -0,0 +1,63 @@
...
+                    string returnValue = new string(lbuffer, 0, (int)pathLength);
+                    ArrayPool<char>.Shared.Return(lbuffer);
+                    return returnValue;
+                }
+
+                // buffer was too small, return to array pool.
+                ArrayPool<char>.Shared.Return(lbuffer);
+            }
+
+            throw new InvalidOperationException();
```
JeremyKuhne, line 60: Don't introduce a throw here. Debug.Fail is more appropriate.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleFileNameLongPath.cs
AArnott: As above, I believe `-1` would introduce a bug where a file name that is exactly `lBuffer.Length - 1` in length would be misinterpreted as a 'too small' buffer.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleFileNameLongPath.cs
AArnott: `Debug.Fail` is meaningless in the release bits, as you know. Would you really rather return `string.Empty` as if nothing went wrong?

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleFileNameLongPath.cs
JeremyKuhne: As I mentioned, the docs are conflicting. It also states that it always reserves space for the null, which isn't included in the returned size. The returns `nSize` is a bug in the docs- it will _never_ return `nSize`. I've validated that this is the case by reviewing the Windows sources, but this can also be confirmed by running some tests. I'll see about creating an issue on the docs.

JeremyKuhne: Never mind, I'm wrong. Looked again and it does return nSize (well as long as it isn't larger than the maximum unicode string)- the docs could stand for some clarification in any case.

JeremyKuhne: These truncations are terribly inconsistent in the Win32 API. So in this case:

If you give a buffer `nSize` of 100 and the path is 100 characters long, you will get 100 back and it will _include_ the null. You actually have a 99 character string.

If you give a buffer `nSize` of 100 characters and the path is 99 characters long you will get 99 back.

So you can presume that if you get `nSize` back you didn't have a buffer that was big enough. Sorry for the confusion.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleFileNameLongPath.cs
JeremyKuhne: You're right.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.GetModuleFileNameLongPath.cs
JeremyKuhne: Yes, as that was our "normal" error case.

src/System.Windows.Forms.Primitives/src/Properties/AssemblyInfo.cs
elachlan: Removed it.

JeremyKuhne (approved):
Thanks for your effort and patience here!

PR #8100, opened 2022-11-02
Enable nullability in ToolStripMenuItem

gpetrou:
## Proposed changes

- Enable nullability in ToolStripMenuItem.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8100)

PR #8117, opened 2022-11-05
Make image nullable in ImageRectangle property of ToolStripMenuItem

gpetrou:
## Proposed changes

- Make image nullable in ImageRectangle property of ToolStripMenuItem.
- Addresses https://github.com/dotnet/winforms/pull/8100#discussion_r1012382425

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8117)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripMenuItem.ToolStripMenuItemInternalLayout.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripMenuItem.ToolStripMenuItemInternalLayout.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Diagnostics;
 using System.Drawing;
 using System.Windows.Forms.Layout;
 
@@ -122,9 +123,10 @@ namespace System.Windows.Forms
                             }
                             else
                             {
-                                //If we don't have an image, use the CheckedImage
-                                Image image = _ownerItem.Image ?? _ownerItem.CheckedImage!;
-                                imageRect.Size = image.Size;
+                                // If we don't have an image, use the CheckedImage.
+                                Image? image = _ownerItem.Image ?? _ownerItem.CheckedImage;
+                                Debug.Assert(image is not null);
+                                imageRect.Size = image?.Size ?? menu.ImageScalingSize;
                             }
 
                             imageRect.Y = LayoutUtils.VAlign(imageRect.Size, _ownerItem.ClientBounds, ContentAlignment.MiddleCenter).Y;


lonitra (approved):
LGTM!

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripMenuItem.ToolStripMenuItemInternalLayout.cs
```diff
@@ -122,9 +122,9 @@ public override Rectangle ImageRectangle
                             }
                             else
                             {
-                                //If we don't have an image, use the CheckedImage
-                                Image image = _ownerItem.Image ?? _ownerItem.CheckedImage!;
-                                imageRect.Size = image.Size;
+                                // If we don't have an image, use the CheckedImage.
```
dreddy-work, line 125: do you have any test or scenario where you end up with null for `CheckedImage`? can you add a test around it?

RussKie: We do check `_ownerItem.CheckedImage` for null in `CheckRectangle` API above, so it can be null.

RussKie (approved):

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripMenuItem.ToolStripMenuItemInternalLayout.cs
dreddy-work: Yes i see. In this case, we are setting size to empty when it is null and would be nice to capture scenario in test for reference.

gpetrou: Hmm, looking at the code and the various checks in place, I am not sure how we could end up with an exception. I ended up adding a small test using a TestAccessor to reproduce the exception which is obviously not a typical use case. So, perhaps I should remove it?

dreddy-work: if we don't find a scenario where it could be null and previously wasn't expecting it to be null, I would rather go with `bang` than modifying the functionality here. @RussKie.

RussKie: IMO since we _expect_ `_ownerItem.CheckedImage` is not null here, but we have no guarantee it isn't, it's better to play it safe and do what we're currently doing:
```cs
                                Image? image = _ownerItem.Image ?? _ownerItem.CheckedImage;
                                imageRect.Size = image?.Size ?? Size.Empty;
```

RussKie (dismissed):

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripMenuItem.ToolStripMenuItemInternalLayout.cs
dreddy-work: That would be propagating error to somewhere else down the line. I will be convinced if the scenario proved to be safe here. If  you don't have a scenario where you could see this null, I still feel we shouldn't.

gpetrou: In general, as we have progressed with nullable annotations, we have added in some places extra checks for null with fallback values or return/break statements based on the warnings we have. That does not mean that there were failing tests or that the code could be reached in its current state and throw exceptions. For example, recently we did a similar change in [ToolStripItem](https://github.com/dotnet/winforms/pull/8118/files#diff-e1c1c5ca80cf52bfb52d45d7a5a27651d4e431a1fab20cb57932bd8ca5981e90R1643). So, if we don't want to do that, then I guess we should revert some of these changes for consistency. But I thought that adding such checks could avoid potential problems that we are not aware of, so it seemed better to have them.

RussKie: > That would be propagating error to somewhere else down the line. I will be convinced if the scenario proved to be safe here. If you don't have a scenario where you could see this null, I still feel we shouldn't.

@dreddy-work so, between "incorrectly determine an icon size" and "crash a user app" you chose the latter?
This is not the first time we're finding gaps in the original implementations, where members were incorrectly assumed to be non-nullable.

dreddy-work: We are speaking hypothetically without a practical scenario here. I prefer to catch early. If this happened to be null, we would have heard those crash reports.

Tanya-Solyanik: For early in the release cycle for a new release, I would prefer a crash in this scenario. Crash is much easier to debug and find during testing than a missing or an incorrectly sized image. For servicing it would be a different story. In this case it seems that there was a crash in the past, and we are keeping it. We can assume that apps already hit this case and supplied an image as needed.
We can come up with a more descriptive exception than a null reference to get rid of the `!`.

JeremyKuhne: I believe a deliberate crash here is not a good idea. If we want to catch an odd state here add a debug assert (which I think is a good idea).

In this particular case I think a better choice would be to fall back to `menu.ImageScalingSize` to follow the other block in this condition.

The question we need to ask is how dependent we think people will be on our code doing bad things. Generally speaking, I would want clear justification for continuing to fall over with NREs (e.g., we think people are really dependent on it). Pushing stronger argument validation is good- if we can root-cause the bad state coming in we can add an argument exception at the source. If it isn't clear how things would get into that state, the assert can help flush that out.

In new code we should absolutely lean towards fail hard and fail fast. Null annotating and strong input validation will help that happen closer to the source of the "state corruption".

@gpetrou if you can add an assert here and make the change to `ImageScalingSize` and tag me directly I'd appreciate it.

Tanya-Solyanik: @JeremyKuhne  - Falling back to `ImageScalingSize` has potential of introducing a couple pixels off errors which will be very hard to pin point for third party developers. This is a public API and we don't know when it is called. I feel strongly that we should preserve an exception. Perhaps ArgumentException saying that both Image and CheckedImage are null?

This case is more akin to the new code case becasue this is  a new not a long term release. Most likely users are the new applicaiton devs. I don't think that an explicit exception woud affect migrating applications becasue this code used to throw inthe past and existing applications had already handled it, i.e. set an image.

JeremyKuhne: @Tanya-Solyanik It threw a `NullReferenceException` they may be catching that specifically, not `Exception` (which is usually a bad idea).

> which will be very hard to pin point for third party developers

Sure, but suppose we do change it to `ArgumentException`. Now existing apps that did catch that specific error are likely to fall over (see above) and our answer to end users on why their app doesn't work is that we wanted to make it easier for the developers? That isn't going to be as palatable as "the icon looks odd, but my app still works". We shouldn't punish end-users for developer convenience.

> This case is more akin to the new code case becasue this is a new not a long term release.

That isn't really a good analogy. .NET 7 targeted apps will get the same failure on .NET 9. To say that we'll take more risk on non-LTS releases is fair, but our design principles should not change between LTS and non LTS.


As this is contentious, I looked at this further to see how you get into this state. This occurs if you have a `CheckedState` of `Unchecked` and you hit that code path. It's crazy to throw for that. In other places `CheckedState` is checked for null and we do fallbacks that seem appropriate. This is bottom-line a bug in _our_ code that needs fixed.

Sumary:

- NREs are terrible, and we should do something about them
- When we have an NRE from parameter input we should convert them to `ArgumentException` (of some sort)
- We should be very cautious about throwing in other cases and have strong justification that it is the best option
  - We _know_ it will fall over later is one such case
- We should consider the impact of changes to end users
- We don't want to push our bugs onto developers
- Brand new API should default to failing fast and have strong argument validation (and be null annotated)

Tanya-Solyanik: This change replaces a NullReferenceException with a misaligned menu item. This is a breaking change and should be documented as such.

dreddy-work: > This is bottom-line a bug in our code that needs fixed
> This change replaces a NullReferenceException with a misaligned menu item

My concern is, fixing such things as part of nullability changes makes it very hard to review the change and also to track any regressions that it may cause. Separate PR would be good to track and document.

JeremyKuhne: @gpetrou We discussed offline as a team. If you can just make the change requested here https://github.com/dotnet/winforms/pull/8117#discussion_r1034064894 I'll get this approved and merged. Thanks for your patience!

gpetrou: @JeremyKuhne hopefully this is what you had in mind? I removed the test since it was obviously hitting the assert.

dreddy-work (approved):

PR #8118, opened 2022-11-05

gpetrou:
## Proposed changes

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8118)


src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
```diff
@@ -1640,7 +1640,7 @@ internal Size PreferredImageSize
                 Size imageSize = Size.Empty;
                 if (usingImageList)
                 {
-                    imageSize = Owner.ImageList.ImageSize;
+                    imageSize = Owner!.ImageList!.ImageSize;
```
dreddy-work, line 1643: online 1633, we may expect `Owner` to be null and here we say NOT. I do not see anything else assigning it inbetween.

RussKie:
```suggestion
                    imageSize = Owner?.ImageList!?ImageSize ?? Size.Empty;
```

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripSeparator.cs
```diff
@@ -170,15 +170,15 @@ private bool IsVertical
         {
             get
             {
-                ToolStrip parent = ParentInternal;
+                ToolStrip? parent = ParentInternal;
```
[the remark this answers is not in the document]
RussKie:
```suggestion
                ToolStrip? parent = ParentInternal ?? Owner;
```

Not sure if `Owner` can be null here or not.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripSeparator.cs
```diff
@@ -170,15 +170,15 @@ private bool IsVertical
...
                 parent ??= Owner;
 
                 if (parent is ToolStripDropDownMenu dropDownMenu)
                 {
                     return false;
                 }
 
-                switch (parent.LayoutStyle)
+                switch (parent!.LayoutStyle)
```
RussKie, line 181: Whether this is safe depends on line:173

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripButton.cs
```diff
@@ -197,7 +197,7 @@ protected override void OnPaint(PaintEventArgs e)
                 return;
             }
 
-            ToolStripRenderer renderer = Renderer;
+            ToolStripRenderer renderer = Renderer!;
```
RussKie, line 200: Could you please clarify why are we making this assumption?

gpetrou: The implementation is a bit hard to go through, but inside `Renderer` property we check if `Owner is not null` and return a non-nullable `Renderer`. Otherwise, we return a nullable renderer. Because here and in other similar cases we check if Owner is null in the previous lines it seems that we can use the null-forgiving operator.

RussKie: Makes sense, thank you.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
```diff
@@ -1888,20 +1888,20 @@ internal bool SupportsDisabledHotTracking
...
         [TypeConverter(typeof(StringConverter))]
-        public object Tag
+        public object? Tag
         {
             get
             {
                 if (Properties.ContainsObject(ToolStripItem.s_tagProperty))
                 {
-                    return _propertyStore.GetObject(ToolStripItem.s_tagProperty);
+                    return _propertyStore!.GetObject(ToolStripItem.s_tagProperty);
```
RussKie, line 1897: 1. I think `Properties` should be decorated with `MemberNotNull`, and
2. we should use `Properties` here instead of the field.

RussKie (changes requested):

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
gpetrou: Just using `Properties` seems enough. `_propertyStore` is not used in other places besides inside the property.

RussKie (approved):

PR #8119, opened 2022-11-05
Introduce and use TraceVerbose extension method

gpetrou:
## Proposed changes

- Introduce and use TraceVerbose extension method
- Addresses https://github.com/dotnet/winforms/pull/8072#discussion_r1009826425
- Since this involves many changes, we can use the method in System.Windows.Forms.Design in a separate PR.
- We can also introduce TraceInfo and other extension methods in a separate PR.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8119)

### DIFF
--- a/src/Common/src/TraceSwitchExtensions.cs
+++ b/src/Common/src/TraceSwitchExtensions.cs
@@ -1,0 +1,19 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Diagnostics;
+
+namespace System.Windows.Forms
+{
+    internal static class TraceSwitchExtensions
+    {
+        public static void TraceVerbose(this TraceSwitch? traceSwitch, string message)
+        {
+            if (traceSwitch is not null && traceSwitch.TraceVerbose)
+            {
+                Debug.WriteLine(message);
+            }
+        }
+    }
+}
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripPanelRow.VerticalRowManager.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripPanelRow.VerticalRowManager.cs
@@ -3,7 +3,9 @@
 // See the LICENSE file in the project root for more information.
 
 using System.Collections;
+#if DEBUG
 using System.Diagnostics;
+#endif
 using System.Drawing;
 using System.Globalization;
 using System.Windows.Forms.Layout;
@@ -97,7 +99,7 @@ namespace System.Windows.Forms
                     return totalSize.Height < DisplayRectangle.Height;
                 }
 
-                Debug.WriteLineIf(ToolStripPanelRow.s_toolStripPanelRowCreationDebug.TraceVerbose, "VerticalRM.CanMove returns false - not enough room");
+                ToolStripPanelRow.s_toolStripPanelRowCreationDebug.TraceVerbose("VerticalRM.CanMove returns false - not enough room");
                 return false;
             }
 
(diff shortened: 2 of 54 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ToolStripPanelRow.VerticalRowManager.cs
```diff
@@ -3,7 +3,9 @@
 // See the LICENSE file in the project root for more information.
 
 using System.Collections;
+#if DEBUG
 using System.Diagnostics;
+#endif
```
RussKie, line 8: Why is this necessary?

RussKie (commented):
👍

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripPanelRow.VerticalRowManager.cs
gpetrou: We have extra code running inside a `DEBUG` pre-processor check, so this namespace is only needed there. Otherwise, there is a warning in Release.

RussKie (approved):

src/Common/src/TraceSwitchExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Diagnostics;
+
+namespace System.Windows.Forms
+{
+    internal static class TraceSwitchExtensions
+    {
+        public static void TraceVerbose(this TraceSwitch? traceSwitch, string message)
```
ThomasGoulet73, line 11: Sorry for jumping in after this PR was merged but I think you should add the attribute `[Conditional("DEBUG")]` to this method. Otherwise calls to `TraceVerbose` will be included in Release dlls (Even string concatenation needed for the `message` parameter). This is a perf/size regression because calls to `Debug.WriteLineIf` were not included in Release dlls before this PR.

See here (`C.M1` is before this PR, `C.M2` is after this PR and `C.M3` is with my suggestion):
https://sharplab.io/#v2:D4AQTAjAsAUCAMACEEB0ARAlgQwOYDsB7AZwBdMBjYgbllhAGZkxEBhRAb1kR8QAcATpgBu2UgFNkEAGyIB47ABNC+ADYBPRABUB2CuIDKAd0ykKAC0TEA+vOIA1bKoCu4gAoDCwzIvEDjphaIALyI+OJG2rr6AWbmABQARABK4sQAGokANIiJ6OIARs64iKTmknaEzgL6iKIukoJePn6JAJS0MNy8jMgALIgAshDxKEj42AC24m3dPFwwvEuI+UW4qADqQhIAMpjhAJIAZvE2do4NHs2+/iZxqDp64vZ+BSTiOYk8dgAeAFyIJSKfYlSrVWpfADUYSmM06SwAvnNkEwQANBmBRhBxrDZoteAtlrwzmkLq4rt4brELA9os9Xu94gASL5yNL/QGKYH4UFpKo1SQcCbTBHteG8JH4ni9NFDBhYnHTPFLQlEqy2UlOcmeSl+anmWlPF4CN7EcSYlnfdkAoEgtnEfm1IWw0UdZGSyWwfYSAQTVRSWTgKJPfUAUR+EnwxEwKmIsFV0tRMn6wf0xtN4niZUwxFThjuFgA/KU6fqcmNENNiMQ8DNkQnlpgjogs6WC5Yc2FCKQws5VP6AGQDksh9uGtMMs3KtUNtUrQrFTbbcR7cLxKs13Bw5GI93IgDarBUwPIKicSXQoYAQgBVADi7QAusiZcnZY8Jyb3pjs7mP/nAnMYtSDbQDy2xSs0k3OspU4HdeCbFsQNHQDEE7Ige3wPtB2HZCYjHf903eaciVnNVVkXLZTBXfZMw3Ws3Vg3dYI9IA===

gpetrou: Oh, yes. I missed that. I created https://github.com/dotnet/winforms/pull/8138 about it. Thanks.

RussKie: Good catch!

PR #8121, opened 2022-11-05
Enable nullability in CategoryGridEntry

gpetrou:
## Proposed changes

- Enable nullability in CategoryGridEntry.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8121)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/CategoryGridEntry.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/CategoryGridEntry.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.Drawing;
 
 namespace System.Windows.Forms.PropertyGridInternal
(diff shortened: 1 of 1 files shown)


dreddy-work (approved):

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/CategoryGridEntry.cs
```diff
@@ -85,14 +83,14 @@ internal override bool InternalExpanded
                 base.InternalExpanded = value;
                 lock (s_lock)
                 {
-                    s_categoryStates[_name] = value;
+                    s_categoryStates![_name] = value;
```
RussKie, line 86: How do we guarantee this isn't null?

gpetrou: `s_categoryStates` is initialized in the constructor and is never set to null after that.

RussKie: In that case we shouldn't need the bang operator, right?

gpetrou: Without it there is a warning. Perhaps because it is static?

RussKie: 🤕

PR #8123, opened 2022-11-05
Enable nullability in some ListViewItem members

gpetrou:
## Proposed changes

- Enable nullability in some ListViewItem members.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8123)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroup.ListViewGroupAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroup.ListViewGroupAccessibleObject.cs
@@ -23,8 +23,8 @@ public partial class ListViewGroup
 
             // Using item from group for getting of ListView is a workaround for https://github.com/dotnet/winforms/issues/4019
             _owningListView = owningGroup.ListView
-                ?? (owningGroup.Items.Count > 0 && _owningGroup.Items[0].ListView is not null
-                    ? _owningGroup.Items[0].ListView!
+                ?? (owningGroup.Items.Count > 0 && _owningGroup.Items[0].ListView is ListView listView
+                    ? listView
                     : throw new InvalidOperationException(nameof(owningGroup.ListView)));
 
             _owningListViewAccessibilityObject = _owningListView.AccessibilityObject as ListView.ListViewAccessibleObject
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroup.ListViewGroupAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroup.ListViewGroupAccessibleObject.cs
@@ -24,7 +24,7 @@ public partial class ListViewGroup
             // Using item from group for getting of ListView is a workaround for https://github.com/dotnet/winforms/issues/4019
             _owningListView = owningGroup.ListView
                 ?? (owningGroup.Items.Count > 0 && _owningGroup.Items[0].ListView is not null
-                    ? _owningGroup.Items[0].ListView
+                    ? _owningGroup.Items[0].ListView!
                     : throw new InvalidOperationException(nameof(owningGroup.ListView)));
 
             _owningListViewAccessibilityObject = _owningListView.AccessibilityObject as ListView.ListViewAccessibleObject
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.ListViewSubItemCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.ListViewSubItemCollection.cs
@@ -50,7 +50,7 @@ public partial class ListViewItem
                     throw new ArgumentOutOfRangeException(nameof(index), index, string.Format(SR.InvalidArgument, nameof(index), index));
                 }
 
-                return _owner.subItems![index];
+                return _owner.subItems[index];
             }
             set
             {
@@ -59,7 +59,7 @@ public partial class ListViewItem
                     throw new ArgumentOutOfRangeException(nameof(index), index, string.Format(SR.InvalidArgument, nameof(index), index));
                 }
 
-                ListViewSubItem oldSubItem = _owner.subItems![index];
+                ListViewSubItem oldSubItem = _owner.subItems[index];
 
                 _owner.subItems[index] = value.OrThrowIfNull();
                 value._owner = _owner;
@@ -113,9 +113,9 @@ public partial class ListViewItem
         {
             ArgumentNullException.ThrowIfNull(item);
 
-            EnsureSubItemSpace(1, -1);
+            EnsureSubItemsCapacity(1);
             item._owner = _owner;
-            _owner.subItems![_owner.SubItemCount] = item;
+            _owner.subItems.Add(item);
             _owner.UpdateSubItems(_owner.SubItemCount++);
             return item;
         }
@@ -138,13 +138,14 @@ public partial class ListViewItem
         {
             ArgumentNullException.ThrowIfNull(items);
 
-            EnsureSubItemSpace(items.Length, -1);
+            EnsureSubItemsCapacity(items.Length);
             foreach (ListViewSubItem item in items)
             {
                 if (item is not null)
                 {
                     item._owner = _owner;
-                    _owner.subItems![_owner.SubItemCount++] = item;
+                    _owner.subItems.Add(item);
+                    _owner.SubItemCount++;
                 }
             }
 
@@ -155,12 +156,13 @@ public partial class ListViewItem
         {
             ArgumentNullException.ThrowIfNull(items);
 
-            EnsureSubItemSpace(items.Length, -1);
+            EnsureSubItemsCapacity(items.Length);
             foreach (string item in items)
             {
                 if (item is not null)
                 {
-                    _owner.subItems![_owner.SubItemCount++] = new ListViewSubItem(_owner, item);
+                    _owner.subItems.Add(new ListViewSubItem(_owner, item));
+                    _owner.SubItemCount++;
                 }
             }
 
@@ -171,12 +173,13 @@ public partial class ListViewItem
         {
             ArgumentNullException.ThrowIfNull(items);
 
-            EnsureSubItemSpace(items.Length, -1);
+            EnsureSubItemsCapacity(items.Length);
             foreach (string item in items)
             {
                 if (item is not null)
                 {
-                    _owner.subItems![_owner.SubItemCount++] = new ListViewSubItem(_owner, item, foreColor, backColor, font);
+                    _owner.subItems.Add(new ListViewSubItem(_owner, item, foreColor, backColor, font));
+                    _owner.SubItemCount++;
                 }
             }
 
@@ -201,9 +204,10 @@ public partial class ListViewItem
                 for (int i = 0; i < oldCount; i++)
                 {
                     _owner.SubItems[i]._owner = null;
-                    _owner.subItems![i].ReleaseUiaProvider();
+                    _owner.subItems[i].ReleaseUiaProvider();
                 }
 
+                _owner.subItems.Clear();
                 _owner.SubItemCount = 0;
                 _owner.UpdateSubItems(-1, oldCount);
             }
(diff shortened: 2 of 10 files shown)


RussKie:
We're adding a lot new bang operators, and I find it concerning... Did you check if the code could be restructured in some way that could help avoiding these?

gpetrou:
> We're adding a lot new bang operators, and I find it concerning... Did you check if the code could be restructured in some way that could help avoiding these?

It's the same problem we have in other classes where there is an existing check about the number of items (for example we have `SubItemCount` here) or a method from another class that initializes the collection (for example we have `EnsureSubItemSpace` here) or checking of index and throwing `ArgumentOutOfRangeException`, etc... I have not found a good alternative to avoid these.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.cs
```diff
@@ -36,29 +34,29 @@ public partial class ListViewItem : ICloneable, ISerializable
         private static readonly BitVector32.Section s_savedStateImageIndexSection = BitVector32.CreateSection(15, s_stateWholeRowOneStyleSection);
         private static readonly BitVector32.Section s_subItemCountSection = BitVector32.CreateSection(MaxSubItems, s_savedStateImageIndexSection);
 
-        private int indentCount;
```
JeremyKuhne, line 39: Fyi: normally you cannot change field names for classes that have `[Serializable]`. You can in this case as serialization is customized with `ISerializable` and the name changes won't impact that method.

JeremyKuhne (changes requested):
I would harden the deserialization to not allow nulls in as a first step, then iterate from that point. One should presume everything is well-formed after calling Deserialize. Don't allow any nulls into the `_subItems` array.

dreddy-work:
@gpetrou, Can you see if you can spend some time to clean this and get it ready for review again?

gpetrou:
> Can you see if you can spend some time to clean this and get it ready for review again?

@dreddy-work I updated this one and https://github.com/dotnet/winforms/pull/7316.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.ListViewSubItemCollection.cs
```diff
@@ -408,7 +408,7 @@ public void RemoveAt(int index)
 
             int oldCount = _owner.SubItemCount;
             _owner.SubItemCount--;
-            _owner.subItems[_owner.SubItemCount] = null;
+            _owner.subItems[_owner.SubItemCount] = null!;
```
dreddy-work, line 411: on one end, we are ensuring we are not inserting null into `subItems` and throw right away but here in the remove method, we are inserting null. Given `subItem` is private, can you try moving it away from array?

dreddy-work: that should help prevent use of `!` and also ensure `subItem` never contain `null`

gpetrou: Hmm, I assume you mean to use a generic List instead? What should we do in this case with `public ListViewItem(ListViewSubItem[] subItems, int imageIndex)` constructor for example? Is it OK to allocate more by converting to a list?

dreddy-work: > What should we do in this case with public ListViewItem(ListViewSubItem[] subItems, int imageIndex) constructor for example? 

I missed that constructor. Allocating memory is not the only issue in this case. `ListViewSubItem[]` is by reference and any state changes like `owner` or `Name` or other properties assigned after wouldn't be in sync.

dreddy-work: I will take another look this week and let you know if any other feedback.

JeremyKuhne: As there is some risk of breaking here you could just put `Array.Empty<ListViewSubItem>()` into the field. While it seems unlikely that someone would be dependent on the passed in array being the initial array, it's probably not worth changing this without significantly more documented value.

As to using `List<T>`, it doesn't allocate an array until you add something to it, so it isn't a huge hit and could be considered where we don't have additional side effects like this classes' constructor.

gpetrou: If I understand correctly your comments, I should just change `null!` to `Array.Empty<ListViewSubItem>()`.

dreddy-work: @gpetrou, discussed offline with Jeremy. In this case, it is more appropriate to go with `List<ListViewSubItem>` and prevent inserting null values into it. Also, on public constructor that passes `ListViewSubItem[]` , let's check the values of the array being non-null before assigning/inserting them to `subItems` to make it consistent.

gpetrou: @dreddy-work and @JeremyKuhne I added a separate commit. Is this what you had in mind?

src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.cs
```diff
@@ -913,7 +915,7 @@ public virtual object Clone()
             newItem = (ListViewItem)Activator.CreateInstance(clonedType);
         }
 
-        newItem.subItems = clonedSubItems;
+        newItem.subItems = clonedSubItems.ToList();
```
JeremyKuhne, line 918: This code is a little suboptimal- we're passing it in when we're not creating a derived instance above. I'd just do `new ListViewItem()` above and manually copy the items into the new `subItems` _here_ to avoid the extra unnecessary array creations.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.cs
```diff
@@ -1257,7 +1259,7 @@ protected virtual void Deserialize(SerializationInfo info, StreamingContext cont
             }
 
             newItems[0] = subItems[0];
-            subItems = newItems;
+            subItems = newItems.ToList();
```
JeremyKuhne, line 1262: This again is a place you could just ensure capacity and add the rehydrated subitems directly to the list.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.ListViewSubItemCollection.cs
```diff
@@ -227,61 +230,16 @@ bool IList.Contains(object? item)
...
-                // We had plenty of room. Just move the items if we need to
-                if (index != -1)
-                {
-                    for (int i = _owner.SubItemCount - 1; i >= index; i--)
-                    {
-                        _owner.subItems[i + size] = _owner.subItems[i];
-                    }
-                }
-            }
+            _owner.subItems ??= new List<ListViewSubItem>();
```
JeremyKuhne, line 242: This shouldn't be necessary at all. If there is always a list you can just add, insert, etc.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.ListViewSubItemCollection.cs
```diff
@@ -428,15 +382,15 @@ void ICollection.CopyTo(Array dest, int index)
...
+                Array.Copy(_owner.subItems!.ToArray(), 0, dest, index, Count);
             }
         }
 
         public IEnumerator GetEnumerator()
         {
             if (_owner.subItems is not null)
             {
-                return new ArraySubsetEnumerator(_owner.subItems, _owner.SubItemCount);
+                return new ArraySubsetEnumerator(_owner.subItems.ToArray(), _owner.SubItemCount);
```
JeremyKuhne, line 393: This can all be collapsed to `public IEnumerator GetEnumerator() => _owner.subItems;`

src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.ListViewSubItemCollection.cs
```diff
@@ -428,15 +382,15 @@ void ICollection.CopyTo(Array dest, int index)
         {
             if (Count > 0)
             {
-                Array.Copy(_owner.subItems!, 0, dest, index, Count);
+                Array.Copy(_owner.subItems!.ToArray(), 0, dest, index, Count);
```
JeremyKuhne, line 385: You can skip an unnecessary copy here by `((ICollection)_owner.subItems).CopyTo(index, Count)`

JeremyKuhne (changes requested):
This is in the right direction, changing to `List` can greatly simplify things. I've left a number of comments in this regard.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.ListViewSubItemCollection.cs
```diff
@@ -155,12 +156,13 @@ public void AddRange(string[] items)
         {
             ArgumentNullException.ThrowIfNull(items);
 
-            EnsureSubItemSpace(items.Length, -1);
+            EnsureSubItemsCapacity(items.Length);
```
JeremyKuhne, line 159: These aren't quite right. The method would ensure that there is (_owner._subItems.Count + items.Length) capacity. There is also a MaxItems check, which I can't find the reason for the limitation in the docs.

The check isn't quite right anyway, as it only ever checks if the capacity is exactly at the limit, it doesn't check when you're adding things if they go over. I'm not sure that the check is worth having without a documented reason.

Consider changing the method to `EnsureAdditionalCapacity` and having it calculate Length + additional.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.cs
```diff
@@ -362,22 +386,23 @@ public bool Focused
...
 
                 return Control.DefaultFont;
             }
             else
             {
-                return subItems[0].Font;
+                return _subItems[0].Font;
             }
         }
         set => SubItems[0].Font = value;
```
JeremyKuhne, line 408: This seems odd. Why is there no check for SubItemsCount when there is above?

src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.cs
```diff
@@ -662,34 +688,34 @@ public bool Selected
...
                 {
                     // Set the selected state on the list view item only if the list view's Handle is already created.
-                    listView.CacheSelectedStateForItem(this, value);
+                    _listView.CacheSelectedStateForItem(this, value);
                 }
             }
         }
     }
-
+#nullable disable
```
JeremyKuhne, line 718: Why disable again?

JeremyKuhne (changes requested):
Looking better, added a few more comments

src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.cs
gpetrou: When files contain many lines and require lots of changes, I split the annotation changes in many PRs to help reviewers check them.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.ListViewSubItemCollection.cs
gpetrou: I changed it to `EnsureAdditionalCapacity`. Hopefully, this is what you had in mind. With regards to the limit, I wasn't sure why is it there, so I kept it.

gpetrou:
@JeremyKuhne, @dreddy-work any more feedback on this one?

JeremyKuhne (changes requested):
@gpetrou just one last minor thing. This is looking great! I've been really busy, so please tag me when you've addressed it.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroup.ListViewGroupAccessibleObject.cs
```diff
@@ -24,7 +24,7 @@ public ListViewGroupAccessibleObject(ListViewGroup owningGroup, bool owningGroup
             // Using item from group for getting of ListView is a workaround for https://github.com/dotnet/winforms/issues/4019
             _owningListView = owningGroup.ListView
                 ?? (owningGroup.Items.Count > 0 && _owningGroup.Items[0].ListView is not null
-                    ? _owningGroup.Items[0].ListView
+                    ? _owningGroup.Items[0].ListView!
```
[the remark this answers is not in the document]
gpetrou: @JeremyKuhne I updated this.

JeremyKuhne (approved):
Looks great!

PR #8126, opened 2022-11-06
Enable nullability in some TreeNode members

gpetrou:
## Proposed changes

- Enable nullability in some TreeNode members.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8126)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TreeNode.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TreeNode.cs
@@ -1,8 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-#nullable disable
-
 using System.Collections;
 using System.ComponentModel;
 using System.Drawing;
@@ -36,9 +34,9 @@ public partial class TreeNode : MarshalByRefObject, ICloneable, ISerializable
     // we use it to store font and color data in a minimal-memory-cost manner
     // ie. nodes which don't use fancy fonts or colors (ie. that use the TreeView settings for these)
     //     will take up less memory than those that do.
-    internal OwnerDrawPropertyBag propBag;
-    internal string text;
-    internal string name;
+    internal OwnerDrawPropertyBag? propBag;
+    internal string? text;
+    internal string? name;
 
     // note: as the checked state of a node is user controlled, and this variable is simply for
     // state caching when a node hasn't yet been realized, you should use the Checked property to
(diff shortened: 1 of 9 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/TreeNode.cs
```diff
@@ -331,9 +334,9 @@ public bool Checked
             get
             {
 #if DEBUG
-                if (_handle != IntPtr.Zero && !treeView.IsDisposed)
+                if (_handle != IntPtr.Zero && treeView is not null && !treeView.IsDisposed)
                 {
-                    TreeView tv = TreeView;
+                    TreeView tv = TreeView!;
```
dreddy-work, line 339: is `TreeView` and t`reeView` same here?

src/System.Windows.Forms/src/System/Windows/Forms/TreeNode.cs
```diff
@@ -674,7 +678,7 @@ public TreeNode LastNode
                     return null;
                 }
 
-                return children[childCount - 1];
+                return children![childCount - 1];
```
JeremyKuhne, line 681: Why can't we tie `childCount` and `children` together? Can you try to remove `childCount` and let `children` be the source of truth?

JeremyKuhne (changes requested):
It's worth trying to fix up the disconnect between the array and the count as it's really dangerous to have them separate. Additionally please make sure the `Deserialize` doesn't let garbage (nulls, etc.) back in where we would never have serialized garbage in the first place. That will put us in a better position to null annotate.

src/System.Windows.Forms/src/System/Windows/Forms/TreeNode.cs
gpetrou: Hmm, `childCount` according to `TreeNodeAndPropertyBag_RoundTripAndExchangeWithNet` participates in serialization/deserialization. So, I assume that we can't just remove it, right? Or you mean something else?

RussKie: I don't think it "participates" in the serialization, it just gets validated after the deserialization is complete to contain the expected value.

RussKie: The field is read here, but the value can be equally read from another source:
https://github.com/dotnet/winforms/blob/852e29689aca09c81bba1c73eade41a96e13704b/src/System.Windows.Forms/src/System/Windows/Forms/TreeNode.cs#L2152

Here we probably don't technically need to assign the field, but use it as a local variable to initialise the underlying value collection:
https://github.com/dotnet/winforms/blob/852e29689aca09c81bba1c73eade41a96e13704b/src/System.Windows.Forms/src/System/Windows/Forms/TreeNode.cs#L1597-L1598
https://github.com/dotnet/winforms/blob/852e29689aca09c81bba1c73eade41a96e13704b/src/System.Windows.Forms/src/System/Windows/Forms/TreeNode.cs#L1636-L1646

dreddy-work:
@gpetrou, checking back on this long pending PR. Let us know if you are waiting on us.

gpetrou:
@dreddy-work any feedback on this one?

gpetrou:
@dreddy-work maybe you could also take a look at this one?

JeremyKuhne (approved):
Just one nit, otherwise looks good.

PR #5976, opened 2021-10-14
Fixing too large up and down arrows at ToolStripScrollButton

ArtemTatarinov:
Fixes #5413

## Proposed changes

- Images for these buttons are created from the icons by the `GetBitmapFromIcon` method from the `DpiHelper` class. Icon `ScrollButtonUp.ico`, for example, has different sizes (16x16, 24x24, 32x32, etc). The icon constructor is choosing the size that is the closest match to the default size and that default size for 100% DPI scaling is 32x32 (returned by the `GetSystemMetrics` method with parameters `SM_CXICON` and `SM_CYICON`). That's the reason why up and down images are so big.
- Also, the icons are square, and up and down images have a lot of empty space above and below the arrow, like this:
image
- We can change the way of creating these pictures by setting the default start size and scaling it for every DPI mode (see the screenshots below for different modes, starting size is 16x16), which will give us a smooth transition between different scaling factors.
- We can also set the default start size a bit larger to make the buttons bigger (proportionally at every DPI scaling mode):
Start sizes 20x20 & 24x24 (I haven't tried 28x28 and 30x30, because they're looking rather close to the button size of the original issue)
startSize20 startSize24

## Customer Impact

**Before the fix**
beforeTheFix

**After the fix**
afterTheFix

## Regression? 
- No

## Risk
- Minimal

## Test methodology 

- Manual testing (tested 100%, 125%, 150%, 175% scaling modes - see the screenshots above)
- CTI team

## Test environment(s) 
Microsoft Windows [Version 10.0.19043.1237]
.NET SDK 7.0.100-alpha.1.21512.13

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/5976)

PR #8145, opened 2022-11-09
Using GetScaledBitmapFromIcon instead of GetBitmapFromIcon

### LINKED ISSUE #6137: Use getting scaled bitmap method for icons
Use the `GetScaledBitmapFromIcon` from the `DpiHelper` class for getting scaled bitmaps from icons the same way it is used at the `ToolStripScrollButton` class.

@Can you move this resizing Bitmap logic to `DpiHelper` class and create a new issue to track it for remaining Icons in winforms?

_Originally posted by @dreddy-work in https://github.com/dotnet/winforms/pull/5976#discussion_r742237740_

NikitaSemenovAkvelon:
Fixes #6137

## Proposed changes

- Added new overload `Bitmap GetScaledBitmapFromIcon(Type type, string name)` with default size 16x16 which uses in most cases.
- Replaced `GetBitmapFromIcon` with `GetScaledBitmapFromIcon` in cases where resize isn't applied and we know the size of the icon precisely. 

## Customer Impact

- Icons will be more appropriate and their size will be depended on the current DPI.

## Regression? 

- No

## Risk

- Minimal.

## Test methodology 

- Manually (using different screen scales 100%-175%)

 

## Test environment(s) 

.NET SDK:
 Commit:    1b80461e45

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.22621

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8145)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.cs
@@ -318,15 +318,25 @@ namespace System.Windows.Forms
 
         /// <summary>
         ///  Creates a new bitmap scaled to the closest size from the icon set according to the current DPI mode.
+        ///  Default size for 100% DPI is set to 16x16.
         /// </summary>
-        /// <param name="type">Resource type</param>
-        /// <param name="name">Resource name</param>
-        /// <param name="defaultSize">Default size for 100% DPI</param>
-        /// <returns>New scaled bitmap</returns>
-        internal static Bitmap GetScaledBitmapFromIcon(Type type, string name, Size defaultSize)
+        /// <param name="type">A <see cref="Type"/> that specifies the assembly in which to look for the resource.</param>
+        /// <param name="name">The resource name to load.</param>
+        /// <returns>New scaled bitmap.</returns>
+        internal static Bitmap GetScaledBitmapFromIcon(Type type, string name)
+            => GetScaledBitmapFromIcon(type, name, new(16, 16));
+
+        /// <summary>
+        ///  Creates a new bitmap scaled to the closest size from the icon set according to the current DPI mode.
+        /// </summary>
+        /// <param name="type">A <see cref="Type"/> that specifies the assembly in which to look for the resource.</param>
+        /// <param name="name">The resource name to load.</param>
+        /// <param name="unscaledSize">Default size for 100% DPI</param>
+        /// <returns>New scaled bitmap.</returns>
+        internal static Bitmap GetScaledBitmapFromIcon(Type type, string name, Size unscaledSize)
         {
             using Icon icon = new(type, name);
-            Size scaledSize = LogicalToDeviceUnits(defaultSize);
+            Size scaledSize = LogicalToDeviceUnits(unscaledSize);
             Size deltaSize = icon.Size - scaledSize;
             if (Math.Abs(deltaSize.Height) <= 2 && Math.Abs(deltaSize.Width) <= 2)
             {
--- a/src/System.Windows.Forms/src/System/Windows/Forms/BindingNavigator.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/BindingNavigator.cs
@@ -179,19 +179,12 @@ namespace System.Windows.Forms
             // Set up images
             //
 
-            Bitmap moveFirstImage = DpiHelper.GetBitmapFromIcon(typeof(BindingNavigator), "BindingNavigator.MoveFirst");
-            Bitmap movePreviousImage = DpiHelper.GetBitmapFromIcon(typeof(BindingNavigator), "BindingNavigator.MovePrevious");
-            Bitmap moveNextImage = DpiHelper.GetBitmapFromIcon(typeof(BindingNavigator), "BindingNavigator.MoveNext");
-            Bitmap moveLastImage = DpiHelper.GetBitmapFromIcon(typeof(BindingNavigator), "BindingNavigator.MoveLast");
-            Bitmap addNewImage = DpiHelper.GetBitmapFromIcon(typeof(BindingNavigator), "BindingNavigator.AddNew");
-            Bitmap deleteImage = DpiHelper.GetBitmapFromIcon(typeof(BindingNavigator), "BindingNavigator.Delete");
-
-            MoveFirstItem.Image = moveFirstImage;
-            MovePreviousItem.Image = movePreviousImage;
-            MoveNextItem.Image = moveNextImage;
-            MoveLastItem.Image = moveLastImage;
-            AddNewItem.Image = addNewImage;
-            DeleteItem.Image = deleteImage;
+            MoveFirstItem.Image = DpiHelper.GetScaledBitmapFromIcon(typeof(BindingNavigator), "BindingNavigator.MoveFirst");
+            MovePreviousItem.Image = DpiHelper.GetScaledBitmapFromIcon(typeof(BindingNavigator), "BindingNavigator.MovePrevious");
+            MoveNextItem.Image = DpiHelper.GetScaledBitmapFromIcon(typeof(BindingNavigator), "BindingNavigator.MoveNext");
+            MoveLastItem.Image = DpiHelper.GetScaledBitmapFromIcon(typeof(BindingNavigator), "BindingNavigator.MoveLast");
+            AddNewItem.Image = DpiHelper.GetScaledBitmapFromIcon(typeof(BindingNavigator), "BindingNavigator.AddNew");
+            DeleteItem.Image = DpiHelper.GetScaledBitmapFromIcon(typeof(BindingNavigator), "BindingNavigator.Delete");
 
             MoveFirstItem.RightToLeftAutoMirrorImage = true;
             MovePreviousItem.RightToLeftAutoMirrorImage = true;
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Printing/PrintPreviewDialog.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Printing/PrintPreviewDialog.cs
@@ -57,7 +57,7 @@ namespace System.Windows.Forms
 
             _previewControl = new PrintPreviewControl();
             _imageList = new ImageList();
-            _imageList.Images.AddStrip(DpiHelper.GetBitmapFromIcon(typeof(PrintPreviewDialog), "PrintPreviewStrip"));
+            _imageList.Images.AddStrip(DpiHelper.GetScaledBitmapFromIcon(typeof(PrintPreviewDialog), "PrintPreviewStrip", new(112, 16)));
             InitForm();
         }
 
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripTests.cs
@@ -7218,7 +7218,6 @@ namespace System.Windows.Forms.Tests
         {
             Type toolStripScrollButtonType = typeof(ToolStripScrollButton);
             var accessor = typeof(DpiHelper).TestAccessor();
-            Size defaultSize = new(16, 16);
             int oldDeviceDpi = DpiHelper.DeviceDpi;
             DpiTestData dpiTestData = new()
             {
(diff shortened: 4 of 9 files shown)


src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.cs
```diff
@@ -316,6 +316,16 @@ public static Bitmap GetBitmapFromIcon(Type t, string name)
             return bitmap;
         }
 
+        /// <summary>
+        ///  Creates a new bitmap scaled to the closest size from the icon set according to the current DPI mode.
+        ///  Syze by default is set to 16x16.
```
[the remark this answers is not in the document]
Tanya-Solyanik: I would say something like " most common icon size in WinForms controls is 16x16, use this size by default."

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.cs
```diff
@@ -316,6 +316,16 @@ public static Bitmap GetBitmapFromIcon(Type t, string name)
...
+        /// </summary>
+        /// <param name="type">Resource type</param>
```
Tanya-Solyanik, line 323: This comment was there before your change, so take it or leave it. But this explanation is not helpful, and we are trying to improve the files we are touching.  MSDN says:
https://learn.microsoft.com/en-us/dotnet/api/system.drawing.icon.-ctor?view=windowsdesktop-7.0#system-drawing-icon-ctor(system-type-system-string)
> A [Type](https://learn.microsoft.com/en-us/dotnet/api/system.type?view=windowsdesktop-7.0) that specifies the assembly in which to look for the resource.

Tanya-Solyanik: Also `defaultSize` is not an accurate argument name, this is an unscaled size

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.cs
```diff
@@ -316,6 +316,16 @@ public static Bitmap GetBitmapFromIcon(Type t, string name)
...
+        /// <param name="name">Resource name</param>
+        /// <returns>New scaled bitmap</returns>
+        internal static Bitmap GetScaledBitmapFromIcon(Type type, string name)
```
Tanya-Solyanik, line 326: `GetScaledBitmapFromIcon` does not reflect what this function is doing. Function is creating a bitmap of an unscaled size actually. At most callsites, these images are scaled by controls later (by buttons)

src/System.Windows.Forms/src/System/Windows/Forms/BindingNavigator.cs
```diff
@@ -179,12 +179,12 @@ public virtual void AddStandardItems()
             // Set up images
             //
 
-            Bitmap moveFirstImage = DpiHelper.GetBitmapFromIcon(typeof(BindingNavigator), "BindingNavigator.MoveFirst");
-            Bitmap movePreviousImage = DpiHelper.GetBitmapFromIcon(typeof(BindingNavigator), "BindingNavigator.MovePrevious");
-            Bitmap moveNextImage = DpiHelper.GetBitmapFromIcon(typeof(BindingNavigator), "BindingNavigator.MoveNext");
-            Bitmap moveLastImage = DpiHelper.GetBitmapFromIcon(typeof(BindingNavigator), "BindingNavigator.MoveLast");
-            Bitmap addNewImage = DpiHelper.GetBitmapFromIcon(typeof(BindingNavigator), "BindingNavigator.AddNew");
-            Bitmap deleteImage = DpiHelper.GetBitmapFromIcon(typeof(BindingNavigator), "BindingNavigator.Delete");
+            Bitmap moveFirstImage = DpiHelper.GetScaledBitmapFromIcon(typeof(BindingNavigator), "BindingNavigator.MoveFirst");
```
Tanya-Solyanik, line 182: nitpick: these local variables are used in a single place, they could be removed.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.cs
Tanya-Solyanik: Maybe: `Get16by16BitmapFromIcon`? And then you don't need the `<summary>` comment, as they are not required for internals and only are confusing.

dreddy-work (changes requested):
Before i go into the changes, I wanted to get clarification. Winforms uses various sizes for their default size. we use `SM_CYSMICON` and `SM_CXICON` as required. Defaulting to 16*16 will not work for all of them. Also, are you certain we never needed 32*32 as default? 

Also, we already have a method called `GetScaledBitmapFromIcon`. Can you please elaborate your changes and intention here more clearly and how complete is your testing with respect to all WinForms controls? Are you addressing a specific section here and we still have more instances? These changes need deeper analysis.

RussKie:
@NikitaSemenovAkvelon could you please provide some images showing the before and the after states?

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.cs
```diff
@@ -316,6 +316,16 @@ public static Bitmap GetBitmapFromIcon(Type t, string name)
             return bitmap;
         }
 
+        /// <summary>
+        ///  Creates a new bitmap scaled to the closest size from the icon set according to the current DPI mode.
+        ///  Syze by default is set to 16x16.
+        /// </summary>
+        /// <param name="type">Resource type</param>
+        /// <param name="name">Resource name</param>
+        /// <returns>New scaled bitmap</returns>
```
RussKie, line 325: Generally, non public API aren't required to have xml-docs (though very much encouraged). However, the xml-docs for non-public API must follow the same [rules and standards](https://learn.microsoft.comdotnet/csharp/language-reference/xmldoc/recommended-tags) as their public counterparts. That is:
- follow the wording patterns - use MSDN as a source of examples,
- follow "Sentence case rules." - i.e., capitalise the first word and end with a full stop.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.cs
RussKie: > Maybe: `Get16by16BitmapFromIcon`? And then you don't need the `<summary>` comment, as they are not required for internals and only are confusing.

`Get16x16BitmapFromIcon

src/System.Windows.Forms/src/System/Windows/Forms/Printing/PrintPreviewDialog.cs
```diff
@@ -57,7 +57,7 @@ public PrintPreviewDialog()
 
             _previewControl = new PrintPreviewControl();
             _imageList = new ImageList();
-            _imageList.Images.AddStrip(DpiHelper.GetBitmapFromIcon(typeof(PrintPreviewDialog), "PrintPreviewStrip"));
+            _imageList.Images.AddStrip(DpiHelper.GetScaledBitmapFromIcon(typeof(PrintPreviewDialog), "PrintPreviewStrip", new(112, 16)));
```
RussKie, line 60: `112 x 16`? Is this expected? If so, this deserves a comment.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripTests.cs
```diff
@@ -7230,7 +7229,7 @@ public void ToolStripScrollButton_Arrows_Size_ReturnsExpected(string resourceNam
             try
             {
                 accessor.Dynamic.DeviceDpi = dpiTestData.Dpi;
-                Bitmap bitmap = DpiHelper.GetScaledBitmapFromIcon(toolStripScrollButtonType, dpiTestData.ResourceName, defaultSize);
+                Bitmap bitmap = DpiHelper.GetScaledBitmapFromIcon(toolStripScrollButtonType, dpiTestData.ResourceName);
```
RussKie, line 7232:
```suggestion
                using Bitmap bitmap = DpiHelper.GetScaledBitmapFromIcon(toolStripScrollButtonType, dpiTestData.ResourceName);
```

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.cs
NikitaSemenovAkvelon: This method was added just for ease of use due to the default size of most icons is 16x16. By the method, we can avoid creating instances of default unscaled size if it 16x16.
Anyway, it calls `GetScaledBitmapFromIcon(Type type, string name, Size unscaledSize)` overload and then compares unscaled and scaled sizes, and if needed load icon in different sizes (20x20, 24x24, 32x32, etc).

src/System.Windows.Forms/src/System/Windows/Forms/Printing/PrintPreviewDialog.cs
NikitaSemenovAkvelon: It is the size of the icon in recourses.

NikitaSemenovAkvelon:
> Before i go into the changes, I wanted to get clarification. Winforms uses various sizes for their default size. we use `SM_CYSMICON` and `SM_CXICON` as required. Defaulting to 16_16 will not work for all of them. Also, are you certain we never needed 32_32 as default?
> 
> Also, we already have a method called `GetScaledBitmapFromIcon`. Can you please elaborate your changes and intention here more clearly and how complete is your testing with respect to all WinForms controls? Are you addressing a specific section here and we still have more instances? These changes need deeper analysis.

As I mentioned above, the new method `Bitmap GetScaledBitmapFromIcon(Type type, string name)` was added just for ease of use. A method with `Size` is still here, you might provide any custom value. And the resulting size is still calculated and might be different from the default unscaled if a user uses a custom DPI.
As the default unscaled size, I used the size of icons in resources.

NikitaSemenovAkvelon:
> @NikitaSemenovAkvelon could you please provide some images showing the before and the after states?

In UI no changes now, therefore I haven't used Before/After section. Currently, all of these icons, which I used in the PR, have small sizes and can't be scaled bigger. As I see, previous changes (#5976) work due to the size of icons being 64x64 and they can be scaled to smaller sizes.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.cs
Tanya-Solyanik: Well, the callers expect unscaled bitmap because it will be scaled later in the process and control has to store an unscaled bitmap as a reference to use for scaling on each DPI change.
```
        public static Bitmap GetBitmapFromIcon(Type t, string name)
        {
            Icon b = new Icon(t, name);
            Bitmap bitmap = b.ToBitmap();
            b.Dispose();
            return bitmap;
        }
```

dreddy-work:
> As I mentioned above, the new method `Bitmap GetScaledBitmapFromIcon(Type type, string name)` was added just for ease of use. A method with `Size` is still here, you might provide any custom value. And the resulting size is still calculated and might be different from the default unscaled if a user uses a custom DPI. As the default unscaled size, I used the size of icons in resources.

I do not see a value add with this additional method other than adding more confusion with existing similar method names.  You have not clarified if 16*16 size is specific to Toolstrip's Icons or applicable across all WinForms controls.  Also, this change still doesn't solve the original issue of showing difference by scaling bitmaps between 126% and 150%. Happy to chat offline if you like to.

elachlan:
BindingNavigator.cs(8,1): error IDE0005: (NETCORE_ENGINEERING_TELEMETRY=Build) Using directive is unnecessary.

Tanya-Solyanik:
@dreddy-work  - I agree that `Bitmap GetScaledBitmapFromIcon(Type type, string name)` is error-prone, and that that call should always pass in the actual resource size.

>Also, this change still doesn't solve the original issue of showing difference by scaling bitmaps between 126% and 150%.

Please clarify this for me. I thought the issue was about always scaling using the icon class methods because they pick the frame that is closes to the desired size, so we do less "stretching" on our own.

elachlan:
Could the DPIHelper functions be moved to System.Drawing? I imagine that means an API Proposal/Review?

Tanya-Solyanik:
>Could the DPIHelper functions be moved to System.Drawing? I imagine that means an API Proposal/Review?

This would be out of scope for this change.

lonitra:
Closing given that issue has no customer impact and this PR is outdated with main. If anyone feels strongly, they are welcome to give this another go in a newer PR.

PR #8148, opened 2022-11-09
Prevent Over Releasing of IStream

lonitra:
Addresses https://github.com/dotnet/winforms/pull/8105#discussion_r1017451431
~~Update `Dispose(bool disposing)` methods for `AxHost.State` and `DataStreamFromComStream` to avoid releasing COM references on finalizer thread.~~

Ensure `_comStream` in `DataStreamFromComStream` does not get over released
- ~~Add field indicating whether `DataStreamFromComStream` takes ownership of the stream passed into constructor~~
- ~~Check if `DataStreamFromComStream` owns the stream before releasing~~
- Do not take ownership of `_comStream` in `DataStreamFromComStream` and remove its finalizer

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8148)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.PropertyBagStream.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.PropertyBagStream.cs
@@ -24,26 +24,27 @@ namespace System.Windows.Forms
                 {
                     // Visual Basic's memory streams don't support seeking, so we have to work around this limitation
                     // here. We do this by copying the contents of the stream into a MemoryStream object.
-                    Stream stream = new DataStreamFromComStream(istream);
                     const int PAGE_SIZE = 0x1000; // one page (4096b)
                     byte[] streamData = new byte[PAGE_SIZE];
-                    int offset = 0;
-
-                    int count = stream.Read(streamData, offset, PAGE_SIZE);
-                    int totalCount = count;
-
-                    while (count == PAGE_SIZE)
+                    using (DataStreamFromComStream comStream = new(istream, ownsHandle: false))
                     {
-                        byte[] newChunk = new byte[streamData.Length + PAGE_SIZE];
-                        Array.Copy(streamData, newChunk, streamData.Length);
-                        streamData = newChunk;
+                        int offset = 0;
+                        int count = comStream.Read(streamData, offset, PAGE_SIZE);
+                        int totalCount = count;
 
-                        offset += PAGE_SIZE;
-                        count = stream.Read(streamData, offset, PAGE_SIZE);
-                        totalCount += count;
+                        while (count == PAGE_SIZE)
+                        {
+                            byte[] newChunk = new byte[streamData.Length + PAGE_SIZE];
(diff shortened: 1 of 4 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/DataStreamFromComStream.cs
```diff
@@ -154,8 +156,10 @@ protected override void Dispose(bool disposing)
                     _comStream->Commit(STGC.STGC_DEFAULT);
                 }
 
-                // Can't release a COM stream from the finalizer thread.
-                _comStream->Release();
+                if (_ownsHandle)
+                {
+                    _comStream->Release();
```
Tanya-Solyanik, line 161: This code seems to be releasing COM stream from the finalizer thread , since `disposing` is not neccesarily true here. Is this stream an STA or an MTA object, i.e. can it be accessed from another(finalizer) thread?

src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.PropertyBagStream.cs
```diff
@@ -94,8 +95,8 @@ HRESULT IPropertyBag.Interface.Write(PCWSTR pszPropName, VARIANT* pVar)
 
                 internal void Write(IStream* istream)
                 {
-                    Stream stream = new DataStreamFromComStream(istream);
-                    BinaryFormatter formatter = new BinaryFormatter();
+                    using DataStreamFromComStream stream = new(istream, ownsHandle: false);
```
Tanya-Solyanik, line 98: I did not see where `ownsHandle` is `true`.  Do we really need this argument?

JeremyKuhne: Better to have it to force the concept in front of any other potential users.

src/System.Windows.Forms/src/System/Windows/Forms/DataStreamFromComStream.cs
JeremyKuhne: No idea on the incoming stream, would depend. @lonitra if you want to remove this code altogether, remove the finalizer, and just state that it doesn't take ownership for the constructor docs I'm fine with that. We need to tackle the [`IAgileReference`](https://devblogs.microsoft.com/oldnewthing/20151020-00/?p=91321) bit to truly make this safe.

Tanya-Solyanik: Yeah, a comment that it is not ours to release and we should not touch it from the finalizer thread seems to be more suitable here.

lonitra: If that's the case then should `_ownsHandle` field be removed until we address `IAgileReference`?

JeremyKuhne (approved):

PR #8149, opened 2022-11-09
Scrollbar scaling according to Dpi is controllable via public property in Framework

### LINKED ISSUE #2742: Possible porting bug: ScrollBar.ScaleScrollBarForDpiChange is unused in .NET Core
In .NET Framework, if `ScaleScrollBarForDpiChange` then the scroll bar's size does not change if DPI is rescaled.

E.g. https://referencesource.microsoft.com/#System.Windows.Forms/winforms/Managed/System/WinForms/ScrollBar.cs,218

We're missing this check in .NET Core

dreddy-work:
When original fix to scale scrollbars was introduced, we hide change under a switch and introduced public property to help developers control the scaling. In .NET, we removed all switches and accidentally removed developer control via public property `ScaleScrollBarForDpiChange`.

Bringing parity with Framework here.

Fixes #2742

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8149)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ScrollBar.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ScrollBar.cs
@@ -26,7 +26,6 @@ namespace System.Windows.Forms
         private int _value;
         private readonly ScrollOrientation _scrollOrientation;
         private int _wheelDelta;
-        private bool _scaleScrollBarForDpiChange = true;
 
         /// <summary>
         ///  Initializes a new instance of the <see cref="ScrollBar"/> class.
@@ -136,7 +135,7 @@ namespace System.Windows.Forms
         {
             base.RescaleConstantsForDpi(deviceDpiOld, deviceDpiNew);
 
-            if (_scaleScrollBarForDpiChange)
+            if (ScaleScrollBarForDpiChange)
             {
                 Scale((float)deviceDpiNew / deviceDpiOld);
             }
@@ -381,11 +380,7 @@ namespace System.Windows.Forms
         [Browsable(true)]
         [EditorBrowsable(EditorBrowsableState.Always)]
         [SRDescription(nameof(SR.ControlDpiChangeScale))]
-        public bool ScaleScrollBarForDpiChange
-        {
-            get => _scaleScrollBarForDpiChange;
-            set => _scaleScrollBarForDpiChange = value;
-        }
+        public bool ScaleScrollBarForDpiChange { get; set; } = true;
 
         [Browsable(false)]
         [EditorBrowsable(EditorBrowsableState.Never)]
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ScrollBar.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ScrollBar.cs
@@ -135,7 +135,11 @@ namespace System.Windows.Forms
         protected override void RescaleConstantsForDpi(int deviceDpiOld, int deviceDpiNew)
         {
             base.RescaleConstantsForDpi(deviceDpiOld, deviceDpiNew);
-            Scale((float)deviceDpiNew / deviceDpiOld);
+
+            if (_scaleScrollBarForDpiChange)
+            {
+                Scale((float)deviceDpiNew / deviceDpiOld);
+            }
         }
 
         /// <summary>
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ScrollBarTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ScrollBarTests.cs
@@ -2623,11 +2623,6 @@ namespace System.Windows.Forms.Tests
             control.RescaleConstantsForDpi(deviceDpiOld, deviceDpiNew);
             Assert.Equal(expected, control.Size);
             Assert.False(control.IsHandleCreated);
-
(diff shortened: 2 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ScrollBar.cs
```diff
@@ -135,7 +135,11 @@ protected override CreateParams CreateParams
         protected override void RescaleConstantsForDpi(int deviceDpiOld, int deviceDpiNew)
         {
             base.RescaleConstantsForDpi(deviceDpiOld, deviceDpiNew);
-            Scale((float)deviceDpiNew / deviceDpiOld);
+
+            if (_scaleScrollBarForDpiChange)
```
Tanya-Solyanik, line 139: Consider changing `ScaleScrollBarForDpiChange` into auto property initialized to true.

RussKie:
Could you please add screenshots of how it looks with and without the change?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ScrollBarTests.cs
```diff
@@ -2605,33 +2603,30 @@ public void ScrollBar_RescaleConstantsForDpi_Invoke_Nop(bool scaleScrollBarForDp
...
             };
             control.RescaleConstantsForDpi(deviceDpiOld, deviceDpiNew);
-            Assert.Equal(expected1, control.Size);
+            Assert.Equal(expected, control.Size);
             Assert.False(control.IsHandleCreated);
 
             // Call again.
             control.RescaleConstantsForDpi(deviceDpiOld, deviceDpiNew);
-            Assert.Equal(expected2, control.Size);
+            Assert.Equal(expected, control.Size);
```
Tanya-Solyanik: Please help me undertand why the scroll bar is not re-scaled again. Do we short-circuit in Scale method?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ScrollBarTests.cs
```diff
@@ -2578,10 +2578,8 @@ public static IEnumerable<object[]> RescaleConstantsForDpi_TestData()
         {
             foreach (bool scaleScrollBarForDpiChange in new bool[] { true, false })
             {
-                yield return new object[] { scaleScrollBarForDpiChange, 1, 2 };
-                yield return new object[] { scaleScrollBarForDpiChange, 1, 1 };
-                yield return new object[] { scaleScrollBarForDpiChange, 0, 0 };
```
Tanya-Solyanik, line 2583: Testing invalid input is important to maintain the same failure handling. Is it possible to put zeroes and negative values back?

dreddy-work: We can but tbh, that's not a practical scenario. We were not throwing any failures or validating for invalid inputs in the Rescale Constants Method. Sending (-1, -1) is same as sending (1,1).  I do not see a value there.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ScrollBarTests.cs
dreddy-work: No. We scale every time we ask for rescale. It's unnecessary step i removed. I think you were reviewing earlier commit here.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ScrollBarTests.cs
Tanya-Solyanik: We are not handling invalid values even down the line, in the core scaling methods? Don't we short-circuit scaling? I think this is important to document in the unit test in case someone counts on that. I would keep at least one invalid case.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ScrollBarTests.cs
Tanya-Solyanik: Please point me to that earlier commit.

dreddy-work: @Tanya-Solyanik , here is the earlier commit :https://github.com/dotnet/winforms/pull/8149/commits/ec87261b399f3052536004dc81c4ae93906d8908

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ScrollBarTests.cs
dreddy-work: Addressing concern by adding invalid test input values. Also added comment.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ScrollBarTests.cs
Tanya-Solyanik: >We scale every time we ask for rescale. It's unnecessary step i removed.

I'm missing this point, let's chat.

Tanya-Solyanik (approved):

PR #8115, opened 2022-11-04
Prevent OpenFileDialog From Throwing When Opening With Non-Existing Initial Directory

lonitra:
Fixes #8108 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8115)

PR #8150, opened 2022-11-09
Update FolderBrowserDialogTest

lonitra:
Addresses https://github.com/dotnet/winforms/pull/8115#pullrequestreview-1171260833
- Update FolderBrowserDialogTest to match OpenFileDialogTest
- Move form that is hosting dialogs in the tests to TestUtilities

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8150)


src/System.Windows.Forms/tests/TestUtilities/DialogForm.cs
```diff
@@ -0,0 +1,47 @@
...
+///  Helper class to host dialogs for testing.
+///  This class is typically passed as a parameter to a call to <see cref="CommonDialog.ShowDialog(IWin32Window?)"/>.
+/// </summary>
+internal class DialogForm : Form
+{
+    private User32.WM _message;
+    private User32.WM _action;
+
+    /// <summary>
+    ///  Constructs a DialogForm that performs <paramref name="action"/> to
```
Tanya-Solyanik, line 22:
```suggestion
    ///  Constructs a form that performs <paramref name="action"/> to
```

src/System.Windows.Forms/tests/TestUtilities/DialogForm.cs
```diff
@@ -0,0 +1,47 @@
...
+    ///  the dialog it is hosting once the <paramref name="message"/> has been received.
+    /// </summary>
+    /// <param name="message">
+    ///  The message the form should be listening for.
+    /// </param>
+    /// <param name="action">
+    ///  The action to be done to the dialog once the form has received <paramref name="message"/>.
+    /// </param>
+    public DialogForm(User32.WM message, User32.WM action)
```
Tanya-Solyanik, line 31: Consider: 
`action` -> `messageToPost`

src/System.Windows.Forms/tests/TestUtilities/DialogForm.cs
```diff
@@ -0,0 +1,47 @@
...
+using Windows.Win32.Foundation;
+using static Interop;
+
+namespace System;
+
+/// <summary>
+///  Helper class to host dialogs for testing.
+///  This class is typically passed as a parameter to a call to <see cref="CommonDialog.ShowDialog(IWin32Window?)"/>.
+/// </summary>
+internal class DialogForm : Form
```
Tanya-Solyanik, line 16: Consider: `DialogForm` -> `Host`

src/System.Windows.Forms/tests/TestUtilities/DialogForm.cs
```diff
@@ -0,0 +1,47 @@
...
+    /// </param>
+    public DialogForm(User32.WM message, User32.WM action)
+    {
+        _message = message;
+        _action = action;
+    }
+
+    protected override void WndProc(ref Message m)
+    {
+        if (m.MsgInternal == _message && m.WParamInternal == (uint)MSGF.DIALOGBOX)
```
Tanya-Solyanik, line 39: `_message` can't really be anything other than WM_ENTERIDLE, different message would interpret WParam == 0 differently... I don't think this class can be that generic, it can only work with a single message that indicates that dialog is up.

src/System.Windows.Forms/tests/TestUtilities/DialogForm.cs
RussKie, line 31: Further to the @Tanya-Solyanik feedback, if you want to allow this form to handle other messages, then you should do something like
```suggestion
    public DialogForm(Func<Message, bool> predicate, User32.WM messageToSend)
```
...where `predicate` for the above test would look something like:
```cs
using DialogForm dialogOwnerForm = new(m => m.MsgInternal == User32.WM.ENTERIDLE && m.WParamInternal == (uint)MSGF.DIALOGBOX, User32.WM.CLOSE);
```

src/System.Windows.Forms/tests/TestUtilities/DialogForm.cs
lonitra: Ah yes you're right I overlooked this initially🤦‍♀️  - thanks for the catch!

src/System.Windows.Forms/tests/TestUtilities/DialogForm.cs
lonitra: For now, I will leave it to only handle `WM_ENTERIDLE` as I'm currently not seeing other things we might want to test aside from confirming the dialog has displayed successfully. In the future we can make this more generic if needed, but wow this is very cool! Definitely keeping this in mind for the future. 😄

src/System.Windows.Forms/tests/TestUtilities/Host.cs
```diff
@@ -0,0 +1,42 @@
...
+using Windows.Win32.Foundation;
+using static Interop.User32;
+
+namespace System;
+
+/// <summary>
+///  Helper class to host dialogs for testing.
+///  This class is typically passed as a parameter to a call to <see cref="CommonDialog.ShowDialog(IWin32Window?)"/>.
+/// </summary>
+internal class Host : Form
```
RussKie, line 16: This name is lacking substance, suggest changing it to provide clatity. E.g., `DialogHostForm` or `DialogAutoCloseForm`...

RussKie: ...with that WM_CLOSE can be hardcoded in the form until such time we have a need to pass another message.

Tanya-Solyanik (approved):

RussKie:
In the future you could even take it a step further:
```diff
diff --git a/src/System.Windows.Forms/tests/TestUtilities/DialogHostForm.cs b/src/System.Windows.Forms/tests/TestUtilities/DialogHostForm.cs
index 5c1d5d8d8..a85adf276 100644
--- a/src/System.Windows.Forms/tests/TestUtilities/DialogHostForm.cs
+++ b/src/System.Windows.Forms/tests/TestUtilities/DialogHostForm.cs
@@ -13,14 +13,39 @@ namespace System;
 ///  Helper class to host dialogs for testing.
 ///  This class is typically passed as a parameter to a call to <see cref="CommonDialog.ShowDialog(IWin32Window?)"/>.
 /// </summary>
-internal class DialogHostForm : Form
+internal class DialogHostForm<T> : Form where T : CommonDialog, new()
 {
+    private readonly T _dialog;
+    private readonly Action<T> _onDialogClosed;
+
+    public DialogHostForm(Action<T> onDialogCreated, Action<T> onDialogClosed) 
+    {
+        _onDialogClosed = onDialogClosed;
+
+        _dialog = new();
+        onDialogCreated(_dialog);
+    }
+
+    public T Dialog => _dialog;
+
+    protected override void Dispose(bool disposing)
+    {
+        if (disposing)
+        {
+            _dialog.Dispose();
+        }
+
+        base.Dispose(disposing);
+    }
+
     protected override void WndProc(ref Message m)
     {
         if (m.MsgInternal == WM.ENTERIDLE && m.WParamInternal == (uint)MSGF.DIALOGBOX)
         {
             HWND dialogHandle = (HWND)m.LParamInternal;
             PInvoke.PostMessage(dialogHandle, WM.CLOSE);
+
+            _onDialogClosed(_dialog);
         }
 
         base.WndProc(ref m);
```

Which makes tests look like this:
```cs
    public void FolderBrowserDialog_ShowDialog(bool autoUpgradeEnabled)
    {
        using DialogHostForm<FolderBrowserDialog> dialogOwnerForm = new(
            dialog =>
            {
                dialog.AutoUpgradeEnabled = autoUpgradeEnabled;
            },
            dialog =>
            {
                // Assert dialog's state after it was closed
            });

        Assert.Equal(DialogResult.Cancel, dialogOwnerForm.Dialog.ShowDialog(dialogOwnerForm));
    }
```

RussKie (approved):

Issue #8140, opened 2022-11-09
Convert `ArrayList` usage to `List<T>` where possible


elachlan:
### Is your feature request related to a problem? Please describe

Winforms uses `Arraylist` extensively.

### Describe the solution you'd like and alternatives you've considered

Convert `Arraylist` usage to `List<T>` to reduce boxing and improve memory utilization.

### Will this feature affect UI controls?

N/A

Related: #2644

Remaining occurrences in `System.Windows.Forms.Design`
|File|Line Count|
|-----------------------------------------------------------------------------------------------------------------------------------------------------|----|
|src\System\ComponentModel\Design\ExceptionCollection.cs                                      |33  |
|src\System\Windows\Forms\Design\ToolStripItemDataObject.cs                                   |34  |
|src\System\Windows\Forms\Design\ControlDesigner.DesignerControlCollectionCodeDomSerializer.cs|38  |
|src\System\Windows\Forms\Design\UpDownBaseDesigner.cs                                        |63  |
|src\System\Windows\Forms\Design\DesignBindingValueUIHandler.cs                               |66  |
|src\System\Windows\Forms\Design\ImageListImageEditor.cs                                      |108 |
|src\System\Windows\Forms\Design\ImageCollectionEditor.cs                                     |113 |
|src\System\Windows\Forms\Design\GroupBoxDesigner.cs                                          |124 |
|src\System\Windows\Forms\Design\ComboBoxDesigner.cs                                          |128 |
|src\System\Windows\Forms\Design\LabelDesigner.cs                                             |146 |
|src\System\Windows\Forms\Design\TextBoxBaseDesigner.cs                                       |148 |
|src\System\Windows\Forms\Design\OleDragDropHandler.CfCodeToolboxItem.cs                      |162 |
|src\System\Windows\Forms\Design\ButtonBaseDesigner.cs                                        |218 |
|src\System\Windows\Forms\Design\Behavior\ContainerSelectorBehavior.cs                        |230 |
|src\System\Windows\Forms\Design\ToolStripAdornerWindowService.cs                             |256 |
|src\System\Windows\Forms\Design\BaseContextMenuStrip.cs                                      |290 |
|src\System\Windows\Forms\Design\OleDragDropHandler.ComponentDataObject.cs                    |302 |
|src\System\ComponentModel\Design\InheritedPropertyDescriptor.cs                              |308 |
[...]

PR #8151, opened 2022-11-09
Refactor TableLayout to replace ArrayList

elachlan:
Refactored TableLayout to replace `ArrayList` with `List<T>`, then removed unnecessary casts via code analyzer.

Related: #8140


src/System.Windows.Forms/src/System/Windows/Forms/Layout/TableLayout.cs
```diff
@@ -1487,7 +1487,7 @@ private static void Debug_VerifyNoOverlapping(IArrangedElement container)
             // this code may be useful for debugging, but doesnt work well with
             // row styles
 
-            ArrayList layoutInfos = new ArrayList(container.Children.Count);
+            List<LayoutInfo> layoutInfos = new List<LayoutInfo>(container.Children.Count);
```
dreddy-work, line 1490:
```suggestion
            List<LayoutInfo> layoutInfos = new(container.Children.Count);
```

PR #8152, opened 2022-11-09
Refactor `AxPropertyDescriptor` to replace `ArrayList`

elachlan:
Refactored `AxPropertyDescriptor` to replace `ArrayList` with `List<T>`

Related: #8140

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8152)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/AxHost.AxPropertyDescriptor.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/AxHost.AxPropertyDescriptor.cs
@@ -26,7 +26,7 @@ namespace System.Windows.Forms
 
             private TypeConverter _converter;
             private UITypeEditor _editor;
-            private readonly ArrayList _updateAttributes = new();
+            private readonly List<Attribute> _updateAttributes = new();
             private int _flags;
 
             private const int FlagUpdatedEditorAndConverter = 0x00000001;
@@ -268,7 +268,7 @@ namespace System.Windows.Forms
                     return;
                 }
 
-                ArrayList attributes = new ArrayList(AttributeArray);
+                List<Attribute> attributes = new List<Attribute>(AttributeArray);
                 foreach (Attribute attr in _updateAttributes)
                 {
                     attributes.Add(attr);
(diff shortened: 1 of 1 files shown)


elachlan:
There is a section where there is a `CopyTo` call:
https://github.com/dotnet/winforms/blob/70a0680cb90e48502a2a119c4df6be1649c41f90/src/System.Windows.Forms/src/System/Windows/Forms/AxHost.AxPropertyDescriptor.cs#L276-L278

Should I use ToArray() on the List instead? I wanted to keep the changes to a minimum hence why I didn't do it.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.AxPropertyDescriptor.cs
```diff
@@ -268,7 +267,7 @@ internal void UpdateAttributes()
                     return;
                 }
 
-                ArrayList attributes = new ArrayList(AttributeArray);
+                List<Attribute> attributes = new List<Attribute>(AttributeArray);
```
dreddy-work, line 270:
```suggestion
                List<Attribute> attributes = new(AttributeArray);
```

dreddy-work:
> Should I use ToArray() on the List instead? I wanted to keep the changes to a minimum hence why I didn't do it.

I don't see why not. it creates a new copy of array.

RussKie (approved):

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.AxPropertyDescriptor.cs
```diff
@@ -268,16 +267,13 @@ internal void UpdateAttributes()
                     return;
                 }
 
-                ArrayList attributes = new ArrayList(AttributeArray);
+                List<Attribute> attributes = new(AttributeArray);
                 foreach (Attribute attr in _updateAttributes)
                 {
                     attributes.Add(attr);
```
KindDragon, line 273: I think this foreach equal to 
attributes.AddRange(_updateAttributes);

elachlan: Good idea. Could you put in a PR?

PR #8155, opened 2022-11-10
Add support for the COM global interface table

JeremyKuhne:
When we're holding onto COM pointers that need to be released in a finalizer we need to ensure that we're hitting an agile reference to get any needed apartment marshalling. The Global Interface Table provides a mechanism for this down to our lowest supported platforms.

This creates a static wrapper for the GIT and a disposable wrapper class `AgileComPointer` for managing access to the GIT.

This moves AxHost.State to utilize the new `AgileComPointer`.

Also add a global using for XUnit for the interop unit tests.

cc: @lonitra, @AaronRobinsonMSFT

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8155)


JeremyKuhne:
@jacdavis, you might be interested in looking at this as well.

src/System.Windows.Forms.Primitives/tests/UnitTests/Windows/Win32/GlobalInterfaceTableTests.cs
```diff
@@ -0,0 +1,92 @@
...
+        [Fact]
+        public void GlobalInterfaceTable_CheckRefCounts()
+        {
+            using var stream = ComHelpers.GetComScope<IStream>(new MyStream(), out bool success);
+            Assert.True(success);
+            uint count = stream.Value->AddRef();
+            Assert.Equal(2u, count);
+            count = stream.Value->Release();
+            Assert.Equal(1u, count);
+            uint cookie = GlobalInterfaceTable.RegisterInterface(stream.Value);
```
lonitra, line 22: Can we also add a call to `GlobalInterfaceTable.GetInterface()` and check the count so we are checking all of the methods work as expected?

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.State.cs
```diff
@@ -376,31 +375,22 @@ void ISerializable.GetObjectData(SerializationInfo si, StreamingContext context)
...
 
-            protected virtual void Dispose(bool disposing) => ReleaseResources();
-
             public void Dispose()
             {
                 Dispose(true);
                 GC.SuppressFinalize(this);
             }
-
-            ~State() => Dispose(false);
```
lonitra, line 403: We might want to also remove this from PublicAPI.Unshipped.txt

src/System.Windows.Forms.Primitives/tests/UnitTests/Windows/Win32/GlobalInterfaceTableTests.cs
JeremyKuhne: I'm indirectly doing that in the AgileComPointer

lonitra: Ah, I missed that.

src/System.Windows.Forms.Primitives/tests/UnitTests/Windows/Win32/GlobalInterfaceTableTests.cs
```diff
@@ -0,0 +1,91 @@
...
+            count = stream.Value->AddRef();
+            // +1 for the AddRef, and +1 for the git registration
+            Assert.Equal(3u, count);
+            count = stream.Value->Release();
+            Assert.Equal(2u, count);
+            GlobalInterfaceTable.RevokeInterface(cookie);
+            count = stream.Value->AddRef();
+            Assert.Equal(2u, count);
+            stream.Value->Release();
+        }
```
RussKie, line 31: Please consider adding blank lines to help with readability
```suggestion
        public void GlobalInterfaceTable_CheckRefCounts()
        {
            using var stream = ComHelpers.GetComScope<IStream>(new MyStream(), out bool success);
            Assert.True(success);

            uint count = stream.Value->AddRef();
            Assert.Equal(2u, count);

            count = stream.Value->Release();
            Assert.Equal(1u, count);

            uint cookie = GlobalInterfaceTable.RegisterInterface(stream.Value);
            count = stream.Value->AddRef();
            // +1 for the AddRef, and +1 for the git registration
            Assert.Equal(3u, count);

            count = stream.Value->Release();
            Assert.Equal(2u, count);

            GlobalInterfaceTable.RevokeInterface(cookie);
            count = stream.Value->AddRef();
            Assert.Equal(2u, count);
            stream.Value->Release();
        }
```

JeremyKuhne:
ARM64 release failed with #8156

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.State.cs
```diff
@@ -163,55 +163,52 @@ private unsafe void CreateStorage()
...
+                        &storage);
+
                 if (hr.Failed)
                 {
+                    lockBytes->Release();
                     PInvoke.GlobalFree(hglobal);
-                    ReleaseResources();
                 }
 
-                fixed (IStorage** pStorage = &_storage)
```
Tanya-Solyanik, line 178: Could you please explain why `fixed` is no longer needed?

AaronRobinsonMSFT: It is no longer needed because it is now on the stack as a local instead of a field on a reference type (i.e., `State`).

AaronRobinsonMSFT: It is a good question and indicative of something we need to seriously consider in some of this work. This is very low-level interop and easy to get wrong and cause inscrutable issues. I'm personally struggling with some of this design because it is going to create a large burden with future maintenance. It is pushing the boundary of where the interop team is with delivering source generators.

Please keep these questions coming by the way.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.State.cs
```diff
@@ -163,55 +163,52 @@ private unsafe void CreateStorage()
...
+                        lockBytes,
+                        pstgPriority: null,
+                        STGM.STGM_READWRITE | STGM.STGM_SHARE_EXCLUSIVE,
+                        snbExclude: null,
+                        reserved: 0,
+                        &storage);
+
                 if (hr.Failed)
                 {
+                    lockBytes->Release();
```
AaronRobinsonMSFT, line 191: This is potentially leaking. The runtime can throw an exception during the loading/discovery of a P/Invoke and so these resources would leak.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.State.cs
```diff
@@ -274,23 +271,25 @@ private void InitializeBufferFromStream(Stream ids)
...
+                if (_storage is null || _lockBytes is null)
                 {
                     return null;
                 }
 
-                iPersistStorage.Save(_storage, fSameAsLoad: true);
-                _storage->Commit(0);
+                using var storage = _storage.GetInterface();
+                iPersistStorage.Save(storage, fSameAsLoad: true);
+                storage.Value->Commit(0);
```
AaronRobinsonMSFT, line 282: This returns an `HRESULT`. Does it need to be checked?

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.State.cs
```diff
@@ -274,23 +271,25 @@ private void InitializeBufferFromStream(Stream ids)
...
                 iPersistStorage.HandsOffStorage();
                 try
                 {
                     _buffer = null;
                     _memoryStream = null;
-                    _iLockBytes->Stat(out STATSTG stat, STATFLAG.STATFLAG_NONAME);
+                    using var lockBytes = _lockBytes.GetInterface();
+                    lockBytes.Value->Stat(out STATSTG stat, STATFLAG.STATFLAG_NONAME);
```
AaronRobinsonMSFT, line 289: Check `HRESULT`.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.State.cs
```diff
@@ -274,23 +271,25 @@ private void InitializeBufferFromStream(Stream ids)
...
                     _length = (int)stat.cbSize;
                     _buffer = new byte[_length];
-                    PInvoke.GetHGlobalFromILockBytes(_iLockBytes, out nint hglobal).ThrowOnFailure();
+                    PInvoke.GetHGlobalFromILockBytes(lockBytes, out nint hglobal).ThrowOnFailure();
                     void* pointer = PInvoke.GlobalLock(hglobal);
```
AaronRobinsonMSFT, line 293: A `NULL` return value is a condition that warrants a check of `GetLastError()` typically. Is there something here that makes that unnecessary?

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.State.cs
JeremyKuhne: Yes, these were missed when this was converted. We'll make sure to do another full audit compared to Framework to keep throw scenarios the same. This is a little more complicated than "throw everywhere" as some things had `PreserveSig` and there are numerous places where we swallow all exceptions. While it is a bit of a pain, it's nice to see the logic explicitly now.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.State.cs
JeremyKuhne: Good catch. The original code was not the best at error checking and we're a bit hamstrung by having to maintain compatibility. The big issue is that we can't report errors that we didn't report before. We can, however, optimize things sometimes by short circuiting, but we have to be pretty careful to maintain the same exceptions if there are any.

In this case I'm pulling the Unlock try into the `not null` case.

For those following along, a notable exception that we _have_ been changing is throwing `ArgumentException` (of some sort) instead of `NullReferenceException`. The justification there is that while that _might_ break some, the odds are slim and the gain in diagnosability is worth it.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.State.cs
JeremyKuhne: Anything you can offer to improve the patterns we have is always appreciated. We're leveraging all the tools C# is giving us and will continue to iterate as we learn more and new C# features come online (expressing copy semantics for structs being one example).

lonitra (approved):
🚀

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.State.cs
RussKie: It may worth adding a comment here noting it's a possible leak.

Issue #8143, opened 2022-11-09
Convert `Hashtable` usage to `Dictionary<TKey, TValue`


elachlan:
### Is your feature request related to a problem? Please describe

Hashtable has been used all over the winforms code base. 

### Describe the solution you'd like and alternatives you've considered

Refactoring to `Dictionary<TKey, TValue` is explicit in its typing and avoids boxing.

### Will this feature affect UI controls?

N/A

elachlan:
`HelpProvider` has a few `Hashtables`. `_boundControls` a bit weird since the key and value are the same and of type `Control`. 
https://github.com/dotnet/winforms/blob/ba551ea8cb9a804e43b0031b964a45be7298d09b/src/System.Windows.Forms/src/System/Windows/Forms/HelpProvider.cs#L26

I thought this might be a good use of `Hashset<T>`, but this suggests the performance isn't as good as List<T> for objects until after 20 items. 

Does the team have a suggestion here?

elachlan:
@Jericho, you might want to jump in on this issue as well.

Jericho:
@elachlan sure. As soon as I'm done working on #8140, I'll inventory the remaining HashTable and create a list to track progress (like I did for the ArrayList).

elachlan:
I attempted [WeakHashtable](https://github.com/dotnet/winforms/blob/bc73fc2790fdce8ac7399eddf76824161bdd35cc/src/System.Windows.Forms/src/misc/WeakHashtable.cs). It is a complicated one. I am unsure how to proceed. 

Maybe add a type parameter and implement `IDictionary<object, TValue>`? That would allow using a private `Dictionary<EqualityWeakReference, TValue>`, thus hiding the implementation and avoiding boxing.

Jericho:
I'm done with #8140 for the time being so I'll start contributing here. First priority: create a list of the classes that need to be refactored (similar to what I did [here](https://github.com/dotnet/winforms/issues/8140#issuecomment-1310654366)). This will enable us to track our progress.

Jericho:
Here's the list of classes that contains at least one private property of type `HashTable`:

- [ ] ComponentDesigner
- [ ] ComponentDesigner.ShadowPropertyCollection
- [ ] DesignerHost
- [ ] InheritanceService
- [ ] CodeDomLocalizationProvider
- [ ] DesignerSerializationManager
- [ ] ExpressionTable
- [ ] ResourceCodeDomSerializer
[...]

PR #8160, opened 2022-11-10
Refactor `ResXResourceWriter` to use `Dictionary<string, string>` instead of `Hashtable`

elachlan:
Refactor `ResXResourceWriter` to use `Dictionary<string, string>` instead of `Hashtable`.

Related: #8143

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8160)

### DIFF
--- a/src/System.Windows.Forms/src/System/Resources/ResXResourceWriter.cs
+++ b/src/System.Windows.Forms/src/System/Resources/ResXResourceWriter.cs
@@ -511,7 +511,7 @@ namespace System.Resources
         private string GetAliasFromName(AssemblyName assemblyName)
         {
             _cachedAliases ??= new();
-            if (!_cachedAliases.TryGetValue(assemblyName.FullName, out string alias))
+            if (!_cachedAliases.TryGetValue(assemblyName.FullName, out string alias) || string.IsNullOrEmpty(alias))
             {
                 alias = assemblyName.Name;
                 AddAlias(alias, assemblyName);
--- a/src/System.Windows.Forms/src/System/Resources/ResXResourceWriter.cs
+++ b/src/System.Windows.Forms/src/System/Resources/ResXResourceWriter.cs
@@ -4,7 +4,6 @@
 
 #nullable disable
 
-using System.Collections;
 using System.Diagnostics;
 using System.Reflection;
 using System.Text;
--- a/src/System.Windows.Forms/src/System/Resources/ResXResourceWriter.cs
+++ b/src/System.Windows.Forms/src/System/Resources/ResXResourceWriter.cs
@@ -34,7 +34,7 @@ namespace System.Resources
         internal const string AssemblyStr = "assembly";
         internal const string AliasStr = "alias";
 
-        private Hashtable _cachedAliases;
+        private Dictionary<string, string> _cachedAliases;
 
         private static readonly TraceSwitch s_resValueProviderSwitch = new TraceSwitch("ResX", "Debug the resource value provider");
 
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/src/System/Resources/ResXResourceWriter.cs
```diff
@@ -513,10 +510,8 @@ private void AddAssemblyRow(string elementName, string alias, string name)
 
         private string GetAliasFromName(AssemblyName assemblyName)
         {
-            _cachedAliases ??= new Hashtable();
-
-            string alias = (string)_cachedAliases[assemblyName.FullName];
-
+            _cachedAliases ??= new();
+            string alias = _cachedAliases[assemblyName.FullName];
```
dreddy-work, line 514: wouldn't this result in key not found exception when `_cachedAliases` is assigned with `new` ?  Did you see a scenario where we are calling this with `_cachedAliases` being null?

elachlan: Ahh, I see. I will sort it. We will need to use TryGetValue.

src/System.Windows.Forms/src/System/Resources/ResXResourceWriter.cs
```diff
@@ -513,11 +510,8 @@ private void AddAssemblyRow(string elementName, string alias, string name)
 
         private string GetAliasFromName(AssemblyName assemblyName)
         {
-            _cachedAliases ??= new Hashtable();
-
-            string alias = (string)_cachedAliases[assemblyName.FullName];
-
-            if (string.IsNullOrEmpty(alias))
+            _cachedAliases ??= new();
+            if (!_cachedAliases.TryGetValue(assemblyName.FullName, out string alias))
```
dreddy-work, line 514: I know this is existing code but it is flawed.  It was okay with `Hashtable` but replacing it with Dictionary prompts to refactor this code. Assigning an empty dictionary and immediately trying to access the index is not efficient. 
Is assignment necessary in this method? can we simply do   ` if (_cachedAliases == null ) return null ;`

elachlan: No. If `GetAliasFromName` is called, it needs to return an alias string. 

The logic is basically saying, create a cache if we haven't created one already. Check if the value exists in the cache, if it doesn't exist in the cache, add the alias to the cache and return said alias.

dreddy-work: :). you are right. i read next line checking as non null.

dreddy-work (dismissed):
LGTM.

src/System.Windows.Forms/src/System/Resources/ResXResourceWriter.cs
elachlan: Thank you for being a sanity check!

JeremyKuhne, line 514: This isn't exactly the same. Without null annotations turned on this is a dictionary of `<string?, string?>`. Please check that the returned value isn't null or empty as before.

JeremyKuhne (changes requested):
See my comment on checking for null or empty.

src/System.Windows.Forms/src/System/Resources/ResXResourceWriter.cs
elachlan: done.

RussKie (approved):

elachlan:
@JeremyKuhne Could you please look over this one as well.

JeremyKuhne (approved):

PR #8163, opened 2022-11-10
Refactor `AxHost` to use `Dictionary<VSSDK.PROPCAT, CategoryAttribute>` instead of `Hastable`

elachlan:
Related: #8143

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8163)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
@@ -2509,8 +2509,8 @@ namespace System.Windows.Forms
                 return s_categoryNames[index];
             }
 
-            if (_objectDefinedCategoryNames?.TryGetValue(propcat, out CategoryAttribute category)
-                ?? false && category is not null)
+            if (_objectDefinedCategoryNames?.TryGetValue(propcat, out CategoryAttribute category) ?? false
+                && category is not null)
             {
                 return category;
             }


src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
```diff
@@ -2511,8 +2511,7 @@ private unsafe CategoryAttribute GetCategoryForDispid(Ole32.DispatchID dispid)
 
             if (_objectDefinedCategoryNames is not null)
             {
-                CategoryAttribute rval = (CategoryAttribute)_objectDefinedCategoryNames[propcat];
-                if (rval is not null)
+                if (_objectDefinedCategoryNames.TryGetValue(propcat, out CategoryAttribute rval))
```
dreddy-work, line 2514: nit: merge with above `if` and try simplify.

elachlan: I tried this, this solution seemed cleaner because it avoids a clash with the local further down and the null check cannot be simplified for whatever reason.

dreddy-work: hmm.

dreddy-work (dismissed):

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
lonitra: The two local `rval` seem unnecessary here. We should create just one of these (`CategoryAttribute rval = null`) before the `if` and reuse this same variable to get/return the `CategoryAttribute` throughout. We would be able to merge the `if` statements if we restructure it this way I believe. We should also just rename this variable to `category` :).

lonitra (changes requested):
Just a suggestion to restructure code a little bit

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
elachlan: I think that does it.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
```diff
@@ -2509,22 +2509,19 @@ private unsafe CategoryAttribute GetCategoryForDispid(Ole32.DispatchID dispid)
                 return s_categoryNames[index];
             }
 
-            if (_objectDefinedCategoryNames is not null)
+            if (_objectDefinedCategoryNames?.TryGetValue(propcat, out CategoryAttribute category)
+                ?? false && category is not null)
```
RussKie, line 2513:
```suggestion
            if (_objectDefinedCategoryNames?.TryGetValue(propcat, out CategoryAttribute category) ?? false
                && category is not null)
```

RussKie (approved):

lonitra (approved):
LGTM, thank you!

PR #8164, opened 2022-11-11
Refactor `HelpProvider` to use `Dictionary` instead of `Hashtable`

elachlan:
Refactor `HelpProvider` to use `Dictionary` instead of `Hashtable`

Related: #8143

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8164)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/HelpProvider.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/HelpProvider.cs
@@ -67,7 +67,7 @@ namespace System.Windows.Forms
         public virtual string GetHelpKeyword(Control ctl)
         {
             ArgumentNullException.ThrowIfNull(ctl);
-            return _keywords.TryGetValue(ctl, out string value) ? value : string.Empty;
+            return _keywords.TryGetValue(ctl, out string value) ? value : null;
         }
 
         /// <summary>
@@ -91,7 +91,7 @@ namespace System.Windows.Forms
         public virtual string GetHelpString(Control ctl)
         {
             ArgumentNullException.ThrowIfNull(ctl);
-            return _helpStrings.TryGetValue(ctl, out string value) ? value : string.Empty;
+            return _helpStrings.TryGetValue(ctl, out string value) ? value : null;
         }
 
         /// <summary>
--- a/src/System.Windows.Forms/src/System/Windows/Forms/HelpProvider.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/HelpProvider.cs
@@ -4,7 +4,6 @@
 
 #nullable disable
 
-using System.Collections;
 using System.ComponentModel;
 using System.Drawing.Design;
 
@@ -23,7 +22,7 @@ namespace System.Windows.Forms
     {
         private readonly Dictionary<Control, string> _helpStrings = new();
         private readonly Dictionary<Control, bool> _showHelp = new();
-        private readonly Hashtable _boundControls = new Hashtable();
+        private readonly List<Control> _boundControls = new();
         private readonly Dictionary<Control, string> _keywords = new();
         private readonly Dictionary<Control, HelpNavigator> _navigators = new();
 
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/HelpProvider.cs
```diff
@@ -21,11 +21,11 @@ namespace System.Windows.Forms
     [SRDescription(nameof(SR.DescriptionHelpProvider))]
     public class HelpProvider : Component, IExtenderProvider
     {
-        private readonly Hashtable _helpStrings = new Hashtable();
-        private readonly Hashtable _showHelp = new Hashtable();
+        private readonly Dictionary<Control, string> _helpStrings = new();
+        private readonly Dictionary<Control, bool> _showHelp = new();
         private readonly Hashtable _boundControls = new Hashtable();
```
elachlan, line 26: I thought this might be a good use of `Hashset<T>`, but stackoverflow suggests the performance isn't as good as `List<T>` for objects until after 20 items. https://stackoverflow.com/a/10762995/908889

src/System.Windows.Forms/src/System/Windows/Forms/HelpProvider.cs
```diff
@@ -80,9 +79,7 @@ public virtual string GetHelpKeyword(Control ctl)
         public virtual HelpNavigator GetHelpNavigator(Control ctl)
         {
             ArgumentNullException.ThrowIfNull(ctl);
-
-            object nav = _navigators[ctl];
-            return nav is null ? HelpNavigator.AssociateIndex : (HelpNavigator)nav;
+            return _navigators.TryGetValue(ctl, out HelpNavigator value) ? value : HelpNavigator.AssociateIndex;
```
JeremyKuhne, line 82: This needs to allow for a null `HelpNavigator`. Without null annotations to ensure that the `Dictionary` doesn't allow nulls, play safe.

```suggestion
            return _navigators.TryGetValue(ctl, out HelpNavigator value) && value is not null ? value : HelpNavigator.AssociateIndex;
```

src/System.Windows.Forms/src/System/Windows/Forms/HelpProvider.cs
JeremyKuhne: I'd still make this something of `<Control>`. Pick `List<T>` or `HashSet<T>`, I'm fine with either if we think there will be few items.

JeremyKuhne (changes requested):
A couple of comments.

src/System.Windows.Forms/src/System/Windows/Forms/HelpProvider.cs
elachlan: Maybe `Dictionary<HWND, Control>`, that way we can quickly lookup the control without doing a large computation on the hashcode? `Control.HWND` should be unique enough for this scenario right?

src/System.Windows.Forms/src/System/Windows/Forms/HelpProvider.cs
elachlan: If I add that I get the following error: `Cannot convert null to 'HelpNavigator' because it is a non-nullable value type`.

JeremyKuhne: Sorry, thought it was a class, my mistake. :)

src/System.Windows.Forms/src/System/Windows/Forms/HelpProvider.cs
JeremyKuhne: Using the HWND as a key is risky (and wrong) as Control can have its handle recreated. Hash codes for objects shouldn't be a concern, they're static for a given instance, have nothing to do with instance fields, and are likely to have already been generated. If you want a hashed lookup, just use HashSet<T>.

src/System.Windows.Forms/src/System/Windows/Forms/HelpProvider.cs
```diff
@@ -21,11 +20,11 @@ namespace System.Windows.Forms
...
     public class HelpProvider : Component, IExtenderProvider
     {
-        private readonly Hashtable _helpStrings = new Hashtable();
-        private readonly Hashtable _showHelp = new Hashtable();
-        private readonly Hashtable _boundControls = new Hashtable();
-        private readonly Hashtable _keywords = new Hashtable();
-        private readonly Hashtable _navigators = new Hashtable();
+        private readonly Dictionary<Control, string> _helpStrings = new();
+        private readonly Dictionary<Control, bool> _showHelp = new();
+        private readonly List<Control> _boundControls = new();
```
elachlan, line 25: I think instead of `List<Control>` we should use `Dictionary<HWND, Control>` as it would reduce the lookup time.

src/System.Windows.Forms/src/System/Windows/Forms/HelpProvider.cs
elachlan: Thanks! I think List<Control> should be okay then. Since its performance is better than HashSet for Count < 20.

elachlan:
@RussKie flaky test, can you please re-run pipeline?

RussKie (approved):

elachlan:
@JeremyKuhne could you please check over this again?

JeremyKuhne (approved):

PR #8166, opened 2022-11-11
Refactor `VisualStyleRenderer` to use `Dictionary` instead of `Hashtable`

elachlan:
Related: #8143

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8166)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/VisualStyles/VisualStyleRenderer.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/VisualStyles/VisualStyleRenderer.cs
@@ -2,7 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System.Collections;
 using System.Drawing;
 using Microsoft.Win32;
 
@@ -17,7 +16,7 @@ namespace System.Windows.Forms.VisualStyles
         private const int NumberOfPossibleClasses = VisualStyleElement.Count; //used as size for themeHandles
 
         [ThreadStatic]
-        private static Hashtable? t_themeHandles; // per-thread cache of ThemeHandle objects.
+        private static Dictionary<string, ThemeHandle>? t_themeHandles; // per-thread cache of ThemeHandle objects.
 
         [ThreadStatic]
         private static long t_threadCacheVersion;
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/VisualStyles/VisualStyleRenderer.cs
```diff
@@ -849,31 +840,27 @@ private static IntPtr GetHandle(string className)
...
-                ThemeHandle? tHandle = ThemeHandle.Create(className, throwExceptionOnFail);
+                tHandle = ThemeHandle.Create(className, throwExceptionOnFail);
                 if (tHandle is null)
                 {
                     return IntPtr.Zero;
                 }
 
-                t_themeHandles.Add(className, tHandle);
+                t_themeHandles[className] = tHandle;
                 return tHandle.Handle;
```
lonitra, line 860: Let's remove this line as this is the same return statement as below.

src/System.Windows.Forms/src/System/Windows/Forms/VisualStyles/VisualStyleRenderer.cs
```diff
@@ -849,31 +840,27 @@ private static IntPtr GetHandle(string className)
...
-
+            t_themeHandles ??= new(NumberOfPossibleClasses);
             if (t_threadCacheVersion != s_globalCacheVersion)
             {
                 RefreshCache();
                 t_threadCacheVersion = s_globalCacheVersion;
             }
 
-            if (!t_themeHandles!.Contains(className))
+            if (!t_themeHandles.TryGetValue(className, out ThemeHandle? tHandle))
```
lonitra, line 850: nit:
```suggestion
            if (!t_themeHandles.TryGetValue(className, out ThemeHandle? themeHandle))
```

lonitra (changes requested):
Some small fixes, otherwise looks good! :)

lonitra (approved):
Thank you!

src/System.Windows.Forms/src/System/Windows/Forms/VisualStyles/VisualStyleRenderer.cs
```diff
@@ -822,17 +813,17 @@ private static void RefreshCache()
 
             foreach (string className in classNames)
             {
-                var tHandle = (ThemeHandle?)t_themeHandles[className];
-                tHandle?.Dispose();
+                ThemeHandle? themeHandle = t_themeHandles[className];
```
RussKie, line 816: Should this use `TryGet` method?

elachlan: It might be a good idea to do so. But not totally necessary, since we do a null check when adding the themeHandle to the dictionary.

Issue #7089, opened 2022-04-27
Uncaught exception or assertion failure: System.Collections.Generic.KeyNotFoundException for listView's setSorting test scenario


Olina-Zhang:
### .NET version

**.Net 7.0 SDK:** 7.0.100-preview.5.22226.4

### Did it work in .NET Framework?

No

### Did it work in any of the earlier releases of .NET Core or .NET 5+?

Yes, it is a regression issue in recent .Net 7.0 build from main branch, cannot repro in .Net 7.0 SDK build: 7.0.100-preview.5.22220.23, it may be caused by PR: https://github.com/dotnet/winforms/pull/7029

### Issue description

Uncaught exception or assertion failure: System.Collections.Generic.KeyNotFoundException when running automation XListView case for setSorting scenario

### Steps to reproduce

We cannot get a manual repro, just met it when running our internal automation case about ListView with setSorting scenario, get following calll stack for product code:

_Uncaught exception or assertion failure: System.Collections.Generic.KeyNotFoundException
System.Collections.Generic.KeyNotFoundException: The given key '0' was not present in the dictionary.
   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
   at System.Windows.Forms.ListView.ListViewNativeItemCollection.get_Item(Int32 displayIndex) in D:\WinForms\Fork\winforms\src\System.Windows.Forms\src\System\Windows\Forms\ListView.ListViewNativeItemCollection.cs:line 83
   at System.Windows.Forms.ListView.ListViewItemCollection.get_Item(Int32 index) in D:\WinForms\Fork\winforms\src\System.Windows.Forms\src\System\Windows\Forms\ListView.ListViewItemCollection.cs:line 107
   at System.Windows.Forms.ListView.WmReflectNotify(Message&amp; m) in D:\WinForms\Fork\winforms\src\System.Windows.Forms\src\System\Windows\Forms\ListView.cs:line 6614
   at System.Windows.Forms.ListView.WndProc(Message&amp; m) in D:\WinForms\Fork\winforms\src\System.Windows.Forms\src\System\Windows\Forms\ListView.cs:line 6908
   at System.Windows.Forms.Control.ControlNativeWindow.OnMessage(Message&amp; m) in D:\WinForms\Fork\winforms\src\System.Windows.Forms\src\System\Windows\Forms\Control.ControlNativeWindow.cs:line 65
[...]

PR #8168, opened 2022-11-11
Refactor `ListView` to use `Dictionary` instead of `Hashtable`

elachlan:
Related: #8143

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8168)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListView.ListViewNativeItemCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListView.ListViewNativeItemCollection.cs
@@ -61,7 +61,8 @@ namespace System.Windows.Forms
 
                         if (_owner.IsHandleCreated && !_owner.ListViewHandleDestroyed)
                         {
-                            return _owner._listItemsTable[DisplayIndexToID(displayIndex)];
+                            _owner._listItemsTable.TryGetValue(DisplayIndexToID(displayIndex), out ListViewItem item);
+                            return item;
                         }
                         else
                         {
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ListView.ListViewNativeItemCollection.cs
```diff
@@ -80,7 +61,7 @@ public bool OwnerIsDesignMode
 
                         if (_owner.IsHandleCreated && !_owner.ListViewHandleDestroyed)
                         {
-                            return (ListViewItem)_owner._listItemsTable[DisplayIndexToID(displayIndex)];
+                            return _owner._listItemsTable[DisplayIndexToID(displayIndex)];
```
gpetrou, line 64: This will lead to a regression. Check https://github.com/dotnet/winforms/issues/7089 and related PRs. We could use TryGetValue here, but I was never able to reproduce the problem with a test.

elachlan: @RussKie When you investigated this originally, was TryGetValue evaluated as a solution? Or was the issue more sinister than that?

RussKie: IIRC, whilst the signatures are the same, the behaviours and null value handling is different between `Hashtable` and `Dictionary<T,V>` types. `TryGetValue` was found to be the solution.

RussKie:
Could you please confirm we have some test coverage for the changed code?

RussKie (approved):

elachlan:
> Could you please confirm we have some test coverage for the changed code?

It seems like the changes are indirectly covered by tests. But there is no direct testing of the code.

RussKie (approved):

PR #8171, opened 2022-11-11
GridError dialog missing AutoScaleDimensions property

### LINKED ISSUE #5111: [HDPI] The warning properties window is not scaled properly when running on 300% DPI
* .NET Core Version: 
.NET SDK 6.0.100-preview.6.21313.2

Yes

**Problem description:**

On 300% DPI machine, the warning properties window is not scaled properly, see screenshot as below:
Actual

**Expected behavior:**

The warning properties window should be scaled properly as 100% DPI.
Expected

**Minimal repro:**

1. Set Windows DPI as 300%, create a .NET Core Winforms application.
2. Focus on form designer, and go to **Properties** window.
3. In **Properties** window, go to **BackColor** property, and type a string "fgjj" to it, and press **Enter**.
4. Related VSO bug: [389044](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/389044)

dreddy-work:
GridError dialog is missing `AutoScaleDimension` property resulting in default scale factor `1`. Thus, not scaled according to the DPI of the primary monitor. This property is auto generated if dialog was designed in the designer.

Fixes #5111 .
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8171)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/GridErrorDialog.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/GridErrorDialog.cs
@@ -272,6 +272,7 @@ namespace System.Windows.Forms.PropertyGridInternal
 
             AcceptButton = _okButton;
             AutoSize = true;
+            AutoScaleDimensions = new System.Drawing.SizeF(7F, 15F);
             AutoScaleMode = AutoScaleMode.Font;
             AutoSizeMode = AutoSizeMode.GrowAndShrink;
             CancelButton = _cancelButton;


src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/GridErrorDialog.cs
```diff
@@ -272,6 +272,7 @@ private void InitializeComponent()
 
             AcceptButton = _okButton;
             AutoSize = true;
+            AutoScaleDimensions = new System.Drawing.SizeF(7F, 15F);
```
JeremyKuhne, line 275: Where do these numbers come from?

dreddy-work: They are from 100% DPI machine.

Tanya-Solyanik (approved):

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/GridErrorDialog.cs
Tanya-Solyanik: This is the "font size" at the design time. i.e. at 100%.

PR #8173, opened 2022-11-12
Remove ANSI methods from the message loop

JeremyKuhne:
While fiddling with hosting VB6 controls I hit asserts regarding the message window changing unicode/ansi state. Investigating it was a result of the HWND changing as we make various peek/get calls for messages. I could track this and use the various APIs, but this flipping back and forth is most likely an artifact of Win9x (guessing hat is on here). Windows does the translation automatically:

https://learn.microsoft.com/windows/win32/api/winuser/nf-winuser-iswindowunicode#remarks

The VB6 control (which has an ANSI window) works just fine with these changes. It simplifies the logic and will speed up the message loop at least a little.

Removes the old GetMessage/PeekMessage defines in favor of CsWin32.

Also remove redundant </inheritdoc> labels- docs are automatically inherited from the interface when explicitly implementing.

cc: @lonitra 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8173)


elachlan:
`PInvoke.PeekMessage.cs(5,1): error IDE0005: (NETCORE_ENGINEERING_TELEMETRY=Build) Using directive is unnecessary.`

RussKie (approved):

lonitra (approved):
👍

JeremyKuhne:
> Why is this applicable only to the message loop, can't we send messages to the hosted VB6 controls?

@Tanya-Solyanik I'm not sure what you're asking. You never have to translate between different windows, Windows does that for you. This is the one place where we were trying to do it ourselves afaik.

PR #8174, opened 2022-11-12

gpetrou:
## Proposed changes

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8174)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.ToolStripItemAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.ToolStripItemAccessibleObject.cs
@@ -312,7 +312,7 @@ namespace System.Windows.Forms
                     if (Owner.IsOnDropDown)
                     {
                         // Return the owner item as the accessible parent.
-                        ToolStripDropDown dropDown = Owner.GetCurrentParentDropDown();
+                        ToolStripDropDown dropDown = Owner.GetCurrentParentDropDown()!;
                         return dropDown.AccessibilityObject;
                     }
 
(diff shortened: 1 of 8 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.ToolStripItemAccessibleObject.cs
```diff
@@ -312,7 +312,7 @@ public override Rectangle Bounds
                     if (Owner.IsOnDropDown)
                     {
                         // Return the owner item as the accessible parent.
-                        ToolStripDropDown dropDown = Owner.GetCurrentParentDropDown();
+                        ToolStripDropDown dropDown = Owner.GetCurrentParentDropDown()!;
```
RussKie, line 315: How do we know this isn't null?

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripMenuItem.cs
```diff
@@ -988,7 +988,7 @@ private void OnMouseButtonStateChange(MouseEventArgs e, bool isMouseDown)
             bool showDropDown = true;
             if (IsOnDropDown)
             {
-                ToolStripDropDown dropDown = GetCurrentParentDropDown() as ToolStripDropDown;
+                ToolStripDropDown dropDown = GetCurrentParentDropDown()!;
```
RussKie, line 991: ditto

RussKie (commented):
👍

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripMenuItem.cs
gpetrou: Inside `IsDropDown` which is called before this line, there are some checks about the parent.

dreddy-work (approved):

PR #8188, opened 2022-11-14
Refactor several data grid view classes to use List<T> instead of ArrayList

Jericho:
Related: https://github.com/dotnet/winforms/issues/8140

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8188)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCellCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCellCollection.cs
@@ -84,7 +84,7 @@ namespace System.Windows.Forms
 
         void ICollection.CopyTo(Array array, int index)
         {
-            _items.ToArray().CopyTo(array, index);
+            ((ICollection)_items).CopyTo(array, index);
         }
 
         int ICollection.Count
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCellCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCellCollection.cs
@@ -18,7 +18,7 @@ namespace System.Windows.Forms
     public class DataGridViewCellCollection : BaseCollection, IList
     {
         private CollectionChangeEventHandler _onCollectionChanged;
-        private readonly ArrayList _items = new ArrayList();
+        private readonly List<DataGridViewCell> _items = new List<DataGridViewCell>();
         private readonly DataGridViewRow _owner;
 
         int IList.Add(object value)
@@ -33,12 +33,22 @@ namespace System.Windows.Forms
 
         bool IList.Contains(object value)
         {
-            return _items.Contains(value);
+            return value switch
+            {
+                null => false,
+                DataGridViewCell dataGridViewCell => Contains(dataGridViewCell),
+                _ => false,
+            };
         }
 
         int IList.IndexOf(object value)
         {
-            return _items.IndexOf(value);
+            return value switch
+            {
+                null => -1,
+                DataGridViewCell dataGridViewCell => _items.IndexOf(dataGridViewCell),
+                _ => -1,
+            };
         }
 
         void IList.Insert(int index, object value)
@@ -74,7 +84,7 @@ namespace System.Windows.Forms
 
         void ICollection.CopyTo(Array array, int index)
         {
-            _items.CopyTo(array, index);
+            _items.ToArray().CopyTo(array, index);
         }
 
         int ICollection.Count
@@ -107,7 +117,7 @@ namespace System.Windows.Forms
         {
             get
             {
-                return _items;
+                return ArrayList.Adapter(_items);
             }
         }
 
@@ -118,7 +128,7 @@ namespace System.Windows.Forms
         {
             get
             {
-                return (DataGridViewCell)_items[index];
+                return _items[index];
             }
             set
             {
@@ -136,7 +146,7 @@ namespace System.Windows.Forms
 
                 _owner.DataGridView?.OnReplacingCell(_owner, index);
 
-                DataGridViewCell oldDataGridViewCell = (DataGridViewCell)_items[index];
+                DataGridViewCell oldDataGridViewCell = _items[index];
                 _items[index] = dataGridViewCell;
                 dataGridViewCell.OwningRow = _owner;
                 dataGridViewCell.State = oldDataGridViewCell.State;
@@ -180,7 +190,7 @@ namespace System.Windows.Forms
                     throw new ArgumentException(string.Format(SR.DataGridViewColumnCollection_ColumnNotFound, columnName), nameof(columnName));
                 }
 
-                return (DataGridViewCell)_items[dataGridViewColumn.Index];
+                return _items[dataGridViewColumn.Index];
             }
             set
             {
@@ -226,7 +236,10 @@ namespace System.Windows.Forms
         internal int AddInternal(DataGridViewCell dataGridViewCell)
         {
             Debug.Assert(!dataGridViewCell.Selected);
-            int index = _items.Add(dataGridViewCell);
+
+            _items.Add(dataGridViewCell);
+            int index = _items.LastIndexOf(dataGridViewCell);
+
             dataGridViewCell.OwningRow = _owner;
             DataGridView dataGridView = _owner.DataGridView;
             if (dataGridView is not null && dataGridView.Columns.Count > index)
@@ -397,7 +410,7 @@ namespace System.Windows.Forms
 
         internal void RemoveAtInternal(int index)
         {
-            DataGridViewCell dataGridViewCell = (DataGridViewCell)_items[index];
+            DataGridViewCell dataGridViewCell = _items[index];
             _items.RemoveAt(index);
             dataGridViewCell.DataGridView = null;
             dataGridViewCell.OwningRow = null;
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCellCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCellCollection.cs
@@ -35,7 +35,6 @@ namespace System.Windows.Forms
         {
             return value switch
             {
-                null => false,
                 DataGridViewCell dataGridViewCell => Contains(dataGridViewCell),
                 _ => false,
             };
@@ -45,7 +44,6 @@ namespace System.Windows.Forms
         {
             return value switch
             {
-                null => -1,
                 DataGridViewCell dataGridViewCell => _items.IndexOf(dataGridViewCell),
                 _ => -1,
             };
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCellCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCellCollection.cs
@@ -18,7 +18,7 @@ namespace System.Windows.Forms
     public class DataGridViewCellCollection : BaseCollection, IList
     {
         private CollectionChangeEventHandler _onCollectionChanged;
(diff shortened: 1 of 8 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewSelectedCellCollection.cs
```diff
@@ -72,7 +82,7 @@ bool IList.IsReadOnly
 
         void ICollection.CopyTo(Array array, int index)
         {
-            _items.CopyTo(array, index);
+            _items.ToArray().CopyTo(array, index);
```
Jericho, line 85: FYI: I would have prefered `_items.CopyTo((DataGridViewCell[])array, index);` but the `DataGridViewSelectedCellCollection_IListCopyTo_InvokeEmpty_Success` unit test verifies that the collection, when empty, can be copied to an array of objects. This array of objects cannot be cast to `.DataGridViewCell[]` which leads to the failed unit test with this exception: `System.InvalidCastException : Unable to cast object of type 'System.Object[]' to type 'System.Windows.Forms.DataGridViewCell[]'.`

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewSelectedRowCollection.cs
```diff
@@ -71,7 +81,7 @@ bool IList.IsReadOnly
 
         void ICollection.CopyTo(Array array, int index)
         {
-            _items.CopyTo(array, index);
+            _items.ToArray().CopyTo(array, index);
```
Jericho, line 84: FYI: I would have preferred `_items.CopyTo((DataGridViewRow[])array, index);` but the `DataGridViewSelectedRowCollection_IListCopyTo_InvokeEmpty_Success` and `DataGridViewSelectedRowCollection_IListCopyTo_InvokeNotEmpty_ReturnsExpected` unit tests verify that the collection can be copied to an array of objects. This array of objects cannot be cast to `DataGridViewRow[]` which leads to failed unit tests with this exception: `System.InvalidCastException : Unable to cast object of type 'System.Object[]' to type 'System.Windows.Forms.DataGridViewRow[]'.`

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCellCollection.cs
```diff
@@ -18,7 +18,7 @@ namespace System.Windows.Forms
     public class DataGridViewCellCollection : BaseCollection, IList
     {
         private CollectionChangeEventHandler _onCollectionChanged;
-        private readonly ArrayList _items = new ArrayList();
+        private readonly List<DataGridViewCell> _items = new List<DataGridViewCell>();
```
RussKie, line 21: [nit]
```suggestion
        private readonly List<DataGridViewCell> _items = new();
```

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCellCollection.cs
```diff
@@ -33,12 +33,22 @@ void IList.Clear()
 
         bool IList.Contains(object value)
         {
-            return _items.Contains(value);
+            return value switch
+            {
+                null => false,
+                DataGridViewCell dataGridViewCell => Contains(dataGridViewCell),
+                _ => false,
+            };
```
RussKie, line 41: I'm not an expert on pattern matching, do we need `null => false` clause here? Isn't this sufficient?
```cs
            return value switch
            {
                DataGridViewCell dataGridViewCell => Contains(dataGridViewCell),
                _ => false,
            };
```

Do we have a test coverage for this method?

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewSelectedCellCollection.cs
RussKie: Since we're changing the underlying types, it's appropriate and acceptable to update tests.
I think we've dealt with a similar issue in the past, see if [this discussion](https://github.com/dotnet/winforms/pull/6909#discussion_r837008443) helps.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCellCollection.cs
```diff
@@ -74,7 +84,7 @@ bool IList.IsReadOnly
 
         void ICollection.CopyTo(Array array, int index)
         {
-            _items.CopyTo(array, index);
+            _items.ToArray().CopyTo(array, index);
```
RussKie, line 87: On the surface this seems wasteful - `_items.ToArray()` allocates one array, then we copy from that array into another array...
I think we've dealt with a similar issue in the past, see if [this discussion](https://github.com/dotnet/winforms/pull/6909#discussion_r837008443) helps.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewSelectedCellCollection.cs
```diff
@@ -17,7 +17,7 @@ namespace System.Windows.Forms
     [ListBindable(false)]
     public class DataGridViewSelectedCellCollection : BaseCollection, IList
     {
-        private readonly ArrayList _items = new ArrayList();
+        private readonly List<DataGridViewCell> _items = new List<DataGridViewCell>();
```
RussKie, line 20:
```suggestion
        private readonly List<DataGridViewCell> _items = new();
```

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCellCollection.cs
Jericho: Thanks for the link. I'll implement Stephen's suggestion.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCellCollection.cs
Jericho: Your suggestion is indeed sufficient. I'll make the change.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCellCollection.cs
```diff
@@ -226,7 +234,10 @@ public virtual int Add(DataGridViewCell dataGridViewCell)
         internal int AddInternal(DataGridViewCell dataGridViewCell)
         {
             Debug.Assert(!dataGridViewCell.Selected);
-            int index = _items.Add(dataGridViewCell);
+
+            _items.Add(dataGridViewCell);
+            int index = _items.LastIndexOf(dataGridViewCell);
```
dreddy-work, line 239: can it not be just `_items.Count -1` ?

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewSelectedCellCollection.cs
```diff
@@ -121,7 +129,10 @@ protected override ArrayList List
         internal int Add(DataGridViewCell dataGridViewCell)
         {
             Debug.Assert(!Contains(dataGridViewCell));
-            return _items.Add(dataGridViewCell);
+            _items.Add(dataGridViewCell);
+
+            var index = _items.LastIndexOf(dataGridViewCell);
```
dreddy-work, line 134: same here.

Jericho: yes it can

RussKie (commented):
👍

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewSelectedColumnCollection.cs
```diff
@@ -12,7 +12,7 @@ namespace System.Windows.Forms
     [ListBindable(false)]
     public class DataGridViewSelectedColumnCollection : BaseCollection, IList
     {
-        private readonly ArrayList _items = new ArrayList();
+        private readonly List<DataGridViewColumn> _items = new List<DataGridViewColumn>();
```
RussKie, line 15:
```suggestion
        private readonly List<DataGridViewColumn> _items = new();
```

RussKie (approved):

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewSelectedCellCollection.cs
```diff
@@ -121,7 +129,10 @@ protected override ArrayList List
         internal int Add(DataGridViewCell dataGridViewCell)
         {
             Debug.Assert(!Contains(dataGridViewCell));
-            return _items.Add(dataGridViewCell);
+            _items.Add(dataGridViewCell);
+
+            int index = _items.Count - 1;
+            return index;
```
dreddy-work, line 135:
```suggestion
            return _items.Count - 1;
```

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewSelectedColumnCollection.cs
```diff
@@ -115,7 +123,8 @@ protected override ArrayList List
         /// </summary>
         internal int Add(DataGridViewColumn dataGridViewColumn)
         {
-            return _items.Add(dataGridViewColumn);
+            _items.Add(dataGridViewColumn);
+            return _items.LastIndexOf(dataGridViewColumn);
```
dreddy-work, line 127: same here. `return _items.Count - 1;`

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewSelectedRowCollection.cs
```diff
@@ -119,7 +127,10 @@ protected override ArrayList List
         /// </summary>
         internal int Add(DataGridViewRow dataGridViewRow)
         {
-            return _items.Add(dataGridViewRow);
+            _items.Add(dataGridViewRow);
+
+            int index = _items.Count - 1;
+            return index;
```
dreddy-work, line 133:
```suggestion
            return _items.Count - 1;
```

dreddy-work (changes requested):
Added comments

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewSelectedCellCollection.cs
Jericho: I was trying to be consistent with other instances where the index of the new item is returned by respecting the syntax that was used in those cases but I'm happy to switch to your proposed syntax.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewSelectedColumnCollection.cs
Jericho: Same explanation too!

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewSelectedRowCollection.cs
Jericho: I was trying to be consistent with other instances where the index of the new item is returned by respecting the syntax that was used in those cases but I'm happy to switch to your proposed syntax.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewSelectedColumnCollection.cs
Jericho: Change made.

Jericho:
@dreddy-work thanks. I addressed all your comments

dreddy-work (approved):

PR #8189, opened 2022-11-14
Refactor WeakRefCollection to use List<WeakRefObject?> instead of ArrayList 

Jericho:
Related: #8140 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8189)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/WeakRefCollection.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/WeakRefCollection.cs
@@ -92,7 +92,7 @@ namespace System.Windows.Forms
 
             for (int i = 0; i < Count; i++)
             {
-                WeakRefObject? item = (WeakRefObject?)InnerList[i];
+                WeakRefObject? item = InnerList[i];
                 if (item != other.InnerList[i])
                 {
                     if (item is null || !item.Equals(other.InnerList[i]))
--- a/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/WeakRefCollection.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/WeakRefCollection.cs
@@ -30,10 +30,10 @@ namespace System.Windows.Forms
 
         public WeakRefCollection(int size)
         {
-            InnerList = new ArrayList(size);
+            InnerList = new List<WeakRefObject?>(size);
         }
 
-        public ArrayList InnerList { get; }
+        public List<WeakRefObject?> InnerList { get; }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/WeakRefCollection.cs
```diff
@@ -147,18 +147,22 @@ public int Add(object? value)
...
+            InnerList.Add(weakRefObject);
+
+            int index = InnerList.LastIndexOf(weakRefObject);
+            return index;
         }
 
         public int Count => InnerList.Count;
 
-        object ICollection.SyncRoot => InnerList.SyncRoot;
+        object ICollection.SyncRoot => this;
```
elachlan, line 159: SyncRoot and IsSynchronized, are they used by the code base at all? They could be problematic.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/WeakRefCollection.cs
```diff
@@ -30,10 +30,10 @@ public WeakRefCollection() : this(4)
```
RussKie, line 30: As a thought exercise - what if changed to this?
```cs
internal class WeakRefCollection : List<WeakRefObject?>
{
    // ...
}
```

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/WeakRefCollection.cs
RussKie: @JeremyKuhne you may have some thoughts on this.

elachlan: These properties one are a blocker for many of the collections being moved to `List<T>`.

Jericho: I'm going to go with `((ICollection)InnerList).SyncRoot`

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/WeakRefCollection.cs
Jericho: I don't think we can do this easily and within the scope of this PR but you should raise a new issue where this can be debated. Changing the interface (or the class derived from) most likely would mean overriding several new methods.

JeremyKuhne (approved):
Thanks!

PR #8198, opened 2022-11-15
Refactor DesignerActionService to replace Hashtable and ArrayList

elachlan:
Related: #8140, #8143

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8198)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionService.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionService.cs
@@ -187,11 +187,7 @@ namespace System.ComponentModel.Design
                         if (verbs is not null && verbs.Count != 0)
                         {
                             List<DesignerVerb> verbsArray = new();
-                            bool hookupEvents = !_componentToVerbsEventHookedUp.Contains(component);
-                            if (hookupEvents)
-                            {
-                                _componentToVerbsEventHookedUp.Add(component);
-                            }
+                            bool hookupEvents = _componentToVerbsEventHookedUp.Add(component);
 
                             foreach (DesignerVerb verb in verbs)
                             {
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionService.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionService.cs
@@ -2,7 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System.Collections;
 using System.Windows.Forms.Design;
 
 namespace System.ComponentModel.Design
@@ -12,11 +11,11 @@ namespace System.ComponentModel.Design
     /// </summary>
     public class DesignerActionService : IDisposable
     {
-        private readonly Hashtable _designerActionLists; // this is how we store 'em.  Syntax: key = object, value = DesignerActionListCollection
+        private readonly Dictionary<IComponent, DesignerActionListCollection> _designerActionLists; // this is how we store 'em.  Syntax: key = object, value = DesignerActionListCollection
         private DesignerActionListsChangedEventHandler _designerActionListsChanged;
         private readonly IServiceProvider _serviceProvider; // standard service provider
         private readonly ISelectionService _selSvc; // selection service
-        private readonly Hashtable _componentToVerbsEventHookedUp; //table component true/false
+        private readonly HashSet<IComponent> _componentToVerbsEventHookedUp; //Hashset of components which have events hooked up.
         // Guard against ReEntrant Code. The Infragistics TabControlDesigner, Sets the Commands Status when the Verbs property is accessed. This property is used in the OnVerbStatusChanged code here and hence causes recursion leading to Stack Overflow Exception.
         private bool _reEntrantCode;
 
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionService.cs
```diff
@@ -192,11 +186,11 @@ protected virtual void GetComponentDesignerActions(IComponent component, Designe
                         DesignerVerbCollection verbs = dcs.Verbs;
                         if (verbs is not null && verbs.Count != 0)
                         {
-                            ArrayList verbsArray = new ArrayList();
-                            bool hookupEvents = _componentToVerbsEventHookedUp[component] is null;
+                            List<DesignerVerb> verbsArray = new();
+                            bool hookupEvents = !_componentToVerbsEventHookedUp.Contains(component);
```
halgab, line 190:
```suggestion
                            bool hookupEvents = _componentToVerbsEventHookedUp.Add(component);
```
and delete `if` below?

elachlan: Thanks, done.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionService.cs
```diff
@@ -192,12 +186,8 @@ protected virtual void GetComponentDesignerActions(IComponent component, Designe
...
                         if (verbs is not null && verbs.Count != 0)
                         {
-                            ArrayList verbsArray = new ArrayList();
-                            bool hookupEvents = _componentToVerbsEventHookedUp[component] is null;
-                            if (hookupEvents)
-                            {
-                                _componentToVerbsEventHookedUp[component] = true;
-                            }
+                            List<DesignerVerb> verbsArray = new();
+                            bool hookupEvents = _componentToVerbsEventHookedUp.Add(component);
```
dreddy-work, line 190: can you explain how this is equivalent? Also is `hookupEvents` used anywhere else?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionService.cs
```diff
@@ -370,21 +349,15 @@ public void Remove(IComponent comp, DesignerActionList actionList)
...
-                ArrayList actionListsToRemove = new ArrayList(1);
-                foreach (DesignerActionList t in actionLists)
+                for (int i = actionLists.Count - 1; i >= 0; i--)
                 {
-                    if (actionList.Equals(t))
+                    if (actionList.Equals(actionLists[i]))
                     {
                         //found one to remove
-                        actionListsToRemove.Add(t);
+                        actionLists.RemoveAt(i);
```
dreddy-work, line 357: I don't think this is right way to remove element. Once you remove, its elements indexes changed. It would be nice if you can add a unit test as well here to make sure we don't do this in the future accidentally.

dreddy-work (changes requested):
Added comments

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionService.cs
elachlan: `hookupEvents` is a local. `HashSet.Add()` returns a bool depending on the existence of the item in the HashSet.

> true if the element is added to the [HashSet<T>](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.hashset-1?view=net-7.0) object; false if the element is already present.
https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.hashset-1.add?view=net-7.0

It simplifies the whole thing since the old version only had items in the collection when their value was true.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionService.cs
elachlan: We are doing a reverse loop. Example: https://stackoverflow.com/questions/1582285/how-to-remove-elements-from-a-generic-list-while-iterating-over-it

I think the other option is to do a `List<T>.RemoveAll()` like in the link. That would be even less verbose.

dreddy-work: Thank you. I should have paid more attention. Names confused me.

dreddy-work (approved):

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionService.cs
elachlan: Do you want me to use RemoveAll? it would remove the loop and make it less verbose and easier to understand.

dreddy-work: Sure. I do not see collection list doing anything else when removed item individually.

elachlan: Actually its of type `DesignerActionListCollection`, so `RemoveAll` isn't available :(

PR #8201, opened 2022-11-15
Fix anchors in nested UserControls

### LINKED ISSUE #6381: Anchors of user control don't work when scaling on high DPI

dreddy-work:
This is not a regression, but the root cause is different in Framework vs .NET 8.0. In .NET 8.0 Anchor calculations are delayed until after the Handles are created. This is problematic in the scenarios where UserControls containing anchored controls are nested and resized. See below example.

ex:
-  User control has a TextBox anchored in all directions.
-  User control is then added to Form and resized, Textbox anchors should have been calculated based on the UserControls original size and grow/shrink on this resize event. However, in runtime while deserializing this, none of the Handles have been created yet and thus anchor calculations computed with the new size for UserControl, when their Handles are created.

Only reason we delayed anchors calculation in .NET8.0 is to accommodate initialization to complete, and all parenting and sizes are applied on the controls. We can achieve this by delaying anchors calculation until after the parent layout is resumed. When handles are created, DPI is applied to handles and WinForms runtime scales controls and anchors (if applicable).

Fixes #6381.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8201)


src/System.Windows.Forms/src/System/Windows/Forms/Layout/DefaultLayout.cs
```diff
@@ -204,13 +204,13 @@ private static Rectangle ComputeAnchoredBoundsV2(IArrangedElement element, Recta
                 // to the parent's width.
                 if (IsAnchored(anchors, AnchorStyles.Right))
                 {
-                    anchorInfo.Left = displayRect.Width - elementBounds.Width - anchorInfo.Right;
+                    anchorInfo.Left = displayRect.Width - bounds.Width - anchorInfo.Right;
```
RussKie, line 207: We created `width` variable above. I would strongly suggest using one or the other, otherwise it makes it much harder to reason about the code, and could lead to inadvertent issues.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Layout/AnchorLayoutTests.cs
```diff
@@ -70,10 +75,9 @@ public void ConfigSwitch_Disabled_HanldeNotCreated_AnchorsComputed()
 
         private static (Form, Button) GetFormWithAnchoredButton()
         {
-            Form form = new()
-            {
-                Size = new Size(200, 300)
-            };
+            Form form = new();
+            form.SuspendLayout();
+            form.Size = new Size(200, 300);
```
RussKie, line 80:
```suggestion
            form.Size = new(200, 300);
```

src/System.Windows.Forms/src/System/Windows/Forms/Layout/DefaultLayout.cs
```diff
@@ -909,8 +882,15 @@ internal static void UpdateAnchorInfoV2(Control control)
                 SetAnchorInfo(control, anchorInfo);
             }
 
+            // Only control's Size or Parent change, prompts recalculation of anchors. Otherwise,
```
JeremyKuhne, line 885: nit: it is better to put this in the block and keep the `else` connected to the `{` above it.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Layout/AnchorLayoutTests.cs
```diff
@@ -66,25 +95,57 @@ public void ConfigSwitch_Disabled_HanldeNotCreated_AnchorsComputed()
...
+        ///  Sets <see cref="LocalAppContextSwitches"/> switch value via reflection.
+        /// </summary>
+        /// <param name="switchName"> AppContext switch name to be set</param>
+        /// <param name="value">AppContext switch value to be set</param>
+        private static void SetSwitch(string switchName, LocalAppContextSwitchValue value)
+        {
+            switch (switchName)
+            {
+                case LocalAppContextSwitches.AnchorLayoutV2SwitchName:
+                    FieldInfo anchorLayoutV2 = typeof(LocalAppContextSwitches).GetField("s_AnchorLayoutV2",
```
JeremyKuhne, line 128: You can do this with the `TestAccessor`

dreddy-work: Do we have something similar for static objects? It seems TestAccessor is for objects that can be instantiated.

RussKie: Yes, there are a few examples throughout the codebase. Perhaps, start here: https://github.com/dotnet/winforms/blob/main/src/System.Windows.Forms/tests/UnitTests/TestAccessorTests.cs
https://github.com/dotnet/winforms/blob/4f4829de9a3b4c56496a60ec18d774ca40956b53/src/System.Windows.Forms/tests/UnitTests/TestAccessorTests.cs#L21-L40

JeremyKuhne (approved):
:shipit:

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Layout/AnchorLayoutTests.cs
```diff
@@ -90,5 +126,12 @@ private static void Dispose(Form form, Button button)
             button?.Dispose();
             form?.Dispose();
         }
+
+        private static void SetAnchorLayoutV2Switch(int value)
```
Tanya-Solyanik, line 130: Do we need the same in the UI unit test file?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Layout/AnchorLayoutTests.cs
```diff
@@ -3,82 +3,118 @@
 // See the LICENSE file in the project root for more information.
 
 using Xunit;
```
Tanya-Solyanik, line 5: COnsider adding `#nullable enable` to this file

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Layout/AnchorLayoutTests.cs
```diff
@@ -90,5 +126,12 @@ private static void Dispose(Form form, Button button)
             button?.Dispose();
```
Tanya-Solyanik, line 126: I don't think these arguments can be `null`.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Layout/AnchorLayoutTests.cs
```diff
@@ -3,82 +3,118 @@
...
 
         [WinFormsFact]
-        public void Control_ParentHandleNotCreated_AnchorsNotComputed()
+        public void Control_ResumedLayout_AnchorsComputed()
         {
+            // TargetFramework on the testhost.exe is NetCoreApp2.1. AppContext.TargetFrameworkName return this value
+            // while running unit tests. To avoid using this invalid target framework for unit tests, we are
+            // explicitly setting and unsetting the switch.
+            // Switch value has 3 states: 0 - unknown, 1 - true, -1 - false
+            SetAnchorLayoutV2Switch(1);
```
Tanya-Solyanik, line 68: Consider introducing several helper methods with more specific names and with no arguments, i.e. `SetUnknownAnchorLayout`, `SetAncorLayoutV2`, `SetAnchorLayoutV1`.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Layout/AnchorLayoutTests.cs
```diff
@@ -3,82 +3,118 @@
...
 
             DefaultLayout.AnchorInfo anchorInfo = DefaultLayout.GetAnchorInfo(button);
             Assert.Null(anchorInfo);
 
             form.Controls.Add(button);
+            anchorInfo = DefaultLayout.GetAnchorInfo(button);
             Assert.Null(anchorInfo);
 
+            // Reset switch to Unknown.
+            SetAnchorLayoutV2Switch(0);
```
Tanya-Solyanik, line 57: Original state is not necessarily `Unknown`. At least it might change in the future, is it possible to cache original value to set the cached value back?

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -348,12 +348,12 @@ private protected void TraceCanProcessMnemonic()
...
         private Dictionary<int, Font>? _dpiFonts;
 
-#if DEBUG
         internal int LayoutSuspendCount
         {
             get { return _layoutSuspendCount; }
         }
 
+#if DEBUG
         internal void AssertLayoutSuspendCount(int value)
```
Tanya-Solyanik, line 357: Please put `example usage` below into a `devdoc` comment in [example ](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/recommended-tags#example)tag.

src/System.Windows.Forms/src/System/Windows/Forms/Layout/DefaultLayout.cs
```diff
@@ -170,20 +170,20 @@ private static Rectangle GetAnchorDestination(IArrangedElement element, Rectangl
 
         private static Rectangle ComputeAnchoredBoundsV2(IArrangedElement element, Rectangle displayRect)
         {
+            Rectangle bounds = GetCachedBounds(element);
```
Tanya-Solyanik, line 173: What is the difference between the cached bounds and the element bounds?

Tanya-Solyanik (dismissed):
Looks good, I added some minor comments. I assume this had been signed off by Olina's team?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Layout/AnchorLayoutTests.cs
dreddy-work: nullable changes are handled separately and being done in chunks. I will leave it for now.

src/System.Windows.Forms/src/System/Windows/Forms/Layout/DefaultLayout.cs
dreddy-work: When you resize Form with anchored controls, control size as result of resize of the Form, are stored in cache and reused. At this point, elements bounds are not applied with these new bounds.  I am preserving the existing approach here.

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
dreddy-work: I will leave this to cleanup on this file.

src/System.Windows.Forms/src/System/Windows/Forms/Layout/DefaultLayout.cs
```diff
@@ -870,28 +841,30 @@ private static void UpdateAnchorInfo(IArrangedElement element)
...
             Control parent = control.Parent;
 
             // Check if control is ready for anchors calculation.
             if (parent is null
-                || !control.IsHandleCreated
-                || !parent.IsHandleCreated)
+                // Design time scenarios suspend layout while deserializing the designer. This is an extra suspension
+                // outside of serialized source and happen only in design-time scenario. Hence, checking for
+                // LayoutSuspendCount > 1.
+                || (control.IsAncestorSiteInDesignMode && parent.LayoutSuspendCount > 1)
```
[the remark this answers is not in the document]
dreddy-work: Maybe, but at this point i am not so clear on how/when to clear to handle popup dialogs in designer. We can chat about it further for this PR, i am leaving it to local.

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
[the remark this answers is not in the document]
dreddy-work: There are variations of assertions checking for `SuspendCount.` Will leave this out of this PR.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Layout/AnchorLayoutTests.cs
dreddy-work: Switch related tests are grouped into unit tests only.

Tanya-Solyanik: Please clarify. Are you changing it soon?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Layout/AnchorLayoutTests.cs
Tanya-Solyanik: This is a new file (and a short file too). So it's easy enough to have new code follow the newest guidelines.

dreddy-work: Why do you say its new file?

Tanya-Solyanik: You added it for this feature work :)

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Layout/AnchorLayoutTests.cs
```diff
@@ -3,82 +3,119 @@
...
+                form.Controls.Add(button);
+                anchorInfo = DefaultLayout.GetAnchorInfo(button);
+                Assert.Null(anchorInfo);
+
+                form.ResumeLayout(performLayout: false);
+                anchorInfo = DefaultLayout.GetAnchorInfo(button);
+                Assert.NotNull(anchorInfo);
+            }
+            finally
+            {
```
Tanya-Solyanik, line 81: Would be nice to refactor these tests so that switches are reset to original values automatically. Maybe Fixtures can help?

dreddy-work: Moved tests to integration suite to run them sequentially. Regarding the refactor suggestion, nothing comes to my mind at this point on how after moving tests. Its a nice to have but doesn't require in this PR.

Tanya-Solyanik (approved):
Loog good, but please investigate if fixtures can be used to unify switch cleanup.

Tanya-Solyanik (approved):

PR #8153, opened 2022-11-09
Refactor DataGridViewColumnCollection to replace ArrayList

elachlan:
Refactored `DataGridViewColumnCollection` to replace `ArrayList` with `List<T>`

Related: #8140

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8153)

PR #8142, opened 2022-11-09
Refactor `AutoCompleteStringCollection` to use `List<string>` instead of `ArrayList`

elachlan:
Work related to #8140

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8142)

PR #8207, opened 2022-11-16
Refactor ArrangedElementCollection to use List<IArrangedElement> instead of ArrayList

Jericho:
Related: #8140

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8207)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlCollection.cs
@@ -188,10 +188,7 @@ namespace System.Windows.Forms
                 return ccOther;
             }
 
-            public bool Contains(Control? control)
-            {
-                return control is null ? false : InnerList.Contains(control);
-            }
+            public bool Contains(Control? control) => control is null ? false : InnerList.Contains(control);
 
             /// <summary>
             ///  Searches for Controls by their Name property, builds up an array
@@ -257,10 +254,7 @@ namespace System.Windows.Forms
                 return new ControlCollectionEnumerator(this);
             }
 
-            public int IndexOf(Control? control)
-            {
-                return control is null ? -1 : InnerList.IndexOf(control);
-            }
+            public int IndexOf(Control? control) => control is null ? -1 : InnerList.IndexOf(control);
 
             /// <summary>
             ///  The zero-based index of the first occurrence of value within the entire CollectionBase, if found; otherwise, -1.
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlCollection.cs
@@ -188,7 +188,7 @@ namespace System.Windows.Forms
                 return ccOther;
             }
 
-            public bool Contains(Control? control) => control is null ? false : InnerList.Contains(control);
+            public bool Contains(Control? control) => ((IList)InnerList).Contains(control);
 
             /// <summary>
             ///  Searches for Controls by their Name property, builds up an array
@@ -254,7 +254,7 @@ namespace System.Windows.Forms
                 return new ControlCollectionEnumerator(this);
             }
 
-            public int IndexOf(Control? control) => control is null ? -1 : InnerList.IndexOf(control);
+            public int IndexOf(Control? control) => ((IList)InnerList).IndexOf(control);
 
             /// <summary>
             ///  The zero-based index of the first occurrence of value within the entire CollectionBase, if found; otherwise, -1.
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlCollection.cs
@@ -184,7 +184,7 @@ namespace System.Windows.Forms
                 ControlCollection ccOther = Owner.CreateControlsInstance();
 
                 // We add using InnerList to prevent unnecessary parent cycle checks, etc.
-                ccOther.InnerList.AddRange(this);
+                ccOther.InnerList.AddRange(InnerList);
                 return ccOther;
             }
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlCollection.cs
@@ -190,7 +190,7 @@ namespace System.Windows.Forms
 
             public bool Contains(Control? control)
             {
-                return InnerList.Contains(control);
+                return control is null ? false : InnerList.Contains(control);
             }
 
             /// <summary>
@@ -259,7 +259,7 @@ namespace System.Windows.Forms
 
             public int IndexOf(Control? control)
             {
-                return InnerList.IndexOf(control);
+                return control is null ? -1 : InnerList.IndexOf(control);
             }
 
             /// <summary>
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Layout/ArrangedElementCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Layout/ArrangedElementCollection.cs
@@ -27,7 +27,7 @@ namespace System.Windows.Forms.Layout
 
         private protected List<IArrangedElement> InnerList { get; }
 
-        internal virtual IArrangedElement this[int index] => InnerList[index]!;
+        internal virtual IArrangedElement this[int index] => InnerList[index];
 
         public override bool Equals(object? obj)
         {
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Layout/ArrangedElementCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Layout/ArrangedElementCollection.cs
@@ -125,47 +125,21 @@ namespace System.Windows.Forms.Layout
             }
         }
 
-        void IList.Clear() => InnerList.Clear();
+        void IList.Clear() => ((IList)InnerList).Clear();
 
-        bool IList.IsFixedSize => false;
+        bool IList.IsFixedSize => ((IList)InnerList).IsFixedSize;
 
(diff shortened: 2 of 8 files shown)


elachlan:
```
src\System.Windows.Forms\src\System\Windows\Forms\Control.ControlCollection.cs(193,43): error CS8604: (NETCORE_ENGINEERING_TELEMETRY=Build) Possible null reference argument for parameter 'item' in 'bool List<IArrangedElement>.Contains(IArrangedElement item)'.
```

Jericho:
@elachlan Thanks for pointing this out. Fixed.

src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlCollection.cs
```diff
@@ -184,13 +184,13 @@ object ICloneable.Clone()
                 ControlCollection ccOther = Owner.CreateControlsInstance();
 
                 // We add using InnerList to prevent unnecessary parent cycle checks, etc.
-                ccOther.InnerList.AddRange(this);
```
dreddy-work, line 187: Little confusing here. fist, why this change? second, InnerList is coming from parent whereas this is current control. Can you explain these two are same here?

src/System.Windows.Forms/src/System/Windows/Forms/Layout/ArrangedElementCollection.cs
```diff
@@ -127,19 +127,45 @@ private static void Copy(ArrangedElementCollection sourceList, int sourceIndex,
 
         void IList.Clear() => InnerList.Clear();
 
-        bool IList.IsFixedSize => InnerList.IsFixedSize;
+        bool IList.IsFixedSize => false;
 
-        bool IList.Contains(object? value) => InnerList.Contains(value);
+        bool IList.Contains(object? value) => (value is IArrangedElement element) ? InnerList.Contains(element) : false;
 
-        public virtual bool IsReadOnly => InnerList.IsReadOnly;
```
dreddy-work, line 134: Changes like these makes it hard to review. Can you make sure you add detailed description of your changes ? this is just not replacing `Arraylaist` with generic Collections.

dreddy-work (dismissed):
Please add description with details for each change types to help review better.

src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlCollection.cs
Jericho: Without this change you would get the following compile this exception:
```
Error	CS1503	Argument 1: cannot convert from 'System.Windows.Forms.Control.ControlCollection' to 'System.Collections.Generic.IEnumerable<System.Windows.Forms.Layout.IArrangedElement>'	System.Windows.Forms	D:\_build\winforms\src\System.Windows.Forms\src\System\Windows\Forms\Control.ControlCollection.cs	187	Active	Normal

File: src/System.Windows.Forms/src/System/Windows/Forms/Layout/ArrangedElementCollection.cs (hunk 2 above)
Comment (Jericho) [reply src/System.Windows.Forms/src/System/Windows/Forms/Layout/ArrangedElementCollection.cs:134]: The data type of 'InnerList' used to be `ArrayList` which has a `IsReadOnly` property. That's why, before this change, you could write ` public virtual bool IsReadOnly => InnerList.IsReadOnly;`. But, after that change, this is no longer possible because `List<T>` does not have a `IsReadOnly` property. I argue that this change in the code is absolutely related to replacing `ArrayList` with `List<T>`. We have made this exact same change in several other places as well such as [here](https://github.com/dotnet/winforms/pull/8142/files#diff-42a2342eabb487120f84d2a4d87df7c6e59b47b2f10c2529eb725f25a4e610b7R47), [here](https://github.com/dotnet/winforms/pull/8212/files#diff-94be5b9664d883143a3192afd1664637a0945d3b3dd3eaee693ab63d1e82319aR116) and [here](https://github.com/dotnet/winforms/pull/8153/files#diff-0a1ebad9dd3c77e05a4f6a7bcaf4c7e72a9a0521b799b6f6a6d9908621c9bcdbR34).

Comment (dreddy-work) [reply src/System.Windows.Forms/src/System/Windows/Forms/Layout/ArrangedElementCollection.cs:134]: Thank you @Jericho for pointing those. Will go over the PR once again.

**Comment by elachlan** [comment]
@dreddy-work this looks like flaky tests, can you please refresh the pipeline?

File: src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlCollection.cs [hunk 3]
```diff
@@ -184,13 +184,13 @@ object ICloneable.Clone()
...
+                ccOther.InnerList.AddRange(InnerList);
                 return ccOther;
             }
 
             public bool Contains(Control? control)
             {
-                return InnerList.Contains(control);
+                return control is null ? false : InnerList.Contains(control);
```
elachlan, line 193: use expression body?

src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlCollection.cs
```diff
@@ -259,7 +259,7 @@ public override IEnumerator GetEnumerator()
 
             public int IndexOf(Control? control)
             {
-                return InnerList.IndexOf(control);
+                return control is null ? -1 : InnerList.IndexOf(control);
```
elachlan, line 262: use expression body?

src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlCollection.cs
Jericho: I was trying to respect the current syntax but happy to change it.

src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlCollection.cs
Jericho: I was trying to respect the current syntax but happy to change it.

src/System.Windows.Forms/src/System/Windows/Forms/Layout/ArrangedElementCollection.cs
```diff
@@ -11,23 +11,23 @@ public class ArrangedElementCollection : IList
...
         {
-            InnerList = new ArrayList(size);
+            InnerList = new List<IArrangedElement>(size);
         }
 
-        private protected ArrayList InnerList { get; }
+        private protected List<IArrangedElement> InnerList { get; }
 
-        internal virtual IArrangedElement this[int index] => (IArrangedElement)InnerList[index]!;
+        internal virtual IArrangedElement this[int index] => InnerList[index]!;
```
JeremyKuhne, line 30: Shouldn't the bang be able to be removed now?

src/System.Windows.Forms/src/System/Windows/Forms/Layout/ArrangedElementCollection.cs
```diff
@@ -127,19 +127,45 @@ private static void Copy(ArrangedElementCollection sourceList, int sourceIndex,
...
+            if (value is IArrangedElement element)
+            {
+                InnerList.Remove(element);
+            }
+        }
 
-        int IList.Add(object? value) => InnerList.Add(value);
+        int IList.Add(object? value)
+        {
+            if (value is IArrangedElement element)
```
JeremyKuhne, line 148: Typing this collection is a breaking change so it is probably better to throw an argument exception here and change the documentation to make it clear that the collection will only take `IArrangedElement` now. `-1` is not something you would ever get from `ArrayList.Add`.

JeremyKuhne (changes requested):
Thanks for doing this! I've got a couple of comments to address, but otherwise looks good!

src/System.Windows.Forms/src/System/Windows/Forms/Layout/ArrangedElementCollection.cs
JeremyKuhne: One option here is to cast the inner list to `IList` so you get the same exact throw it would do (`ArgumentException`). You'll also get the count out that way.

src/System.Windows.Forms/src/System/Windows/Forms/Layout/ArrangedElementCollection.cs
Jericho: I'll remove it.

src/System.Windows.Forms/src/System/Windows/Forms/Layout/ArrangedElementCollection.cs
Jericho: I like your idea to cast to `IList`. I'll make the change in `Add` and `IndexOf` as well.

FYI: as part of #8140, we modified several `Add`, `Contains`, `IndexOf`, etc. methods that could probably have benefited from your suggestion. Here are just a few examples: [here](https://github.com/dotnet/winforms/pull/8188/files#diff-0b3e6d3de1bcf8ce19e1a6bf0f7c6c4a4de8edcde488a696471cc55fb6470abeR36-R40), [here](https://github.com/dotnet/winforms/pull/8142/files#diff-42a2342eabb487120f84d2a4d87df7c6e59b47b2f10c2529eb725f25a4e610b7R65-R66), [here](https://github.com/dotnet/winforms/pull/8153/files#diff-0a1ebad9dd3c77e05a4f6a7bcaf4c7e72a9a0521b799b6f6a6d9908621c9bcdbR198-R199) and [here](https://github.com/dotnet/winforms/pull/8188/files#diff-0b3e6d3de1bcf8ce19e1a6bf0f7c6c4a4de8edcde488a696471cc55fb6470abeR45-R49).

JeremyKuhne (commented):
:shipit:

JeremyKuhne (approved):

src/System.Windows.Forms/src/System/Windows/Forms/Layout/ArrangedElementCollection.cs
Jericho: Absolutely. I'll go back to all PRS where we refactored classes to use List instead if Array List.

Issue #7308, opened 2022-06-15
Remove obsolete API usage from Winforms


dreddy-work:
### .NET version

.NET 7.0

### Did it work in .NET Framework?

Yes

### Did it work in any of the earlier releases of .NET Core or .NET 5+?

Yes

### Issue description

Latest .NET runtime enforcing to discourage use of obsolete APIs in WinForms. This issue is to track obsolete APIs and replace them.

- [ ] AssemblyName.Codebase
- [ ] AssemblyNAme.EscapedCodebase

### Steps to reproduce
refer to work around commit in https://github.com/dotnet/winforms/pull/7306,  https://github.com/dotnet/winforms/pull/7306/commits/564f1c14c355352c598fbec2c63d1fbf25c457f4

elachlan:
Ran into this issue when looking at enabling `IL3000` analyzer.https://github.com/dotnet/winforms/blob/466d7b9dcd22225249e03f1e1e65df8aebc1f3ae/eng/CodeAnalysis.src.globalconfig#L850-L851

elachlan:
@RussKie AFAIK, you can't get `Assembly` from `AssemblyName` directly. Is this correct?

For `AssemblyName.Codebase` replacement, we would have to do:
```CSharp
System.Reflection.Assembly.Load(AssemblyName).Location;
```

dreddy-work:
> @RussKie AFAIK, you can't get `Assembly` from `AssemblyName` directly. Is this correct?

Yes. Other way is fine though. 

> For `AssemblyName.Codebase` replacement, we would have to do:
> 
> ```cs
> System.Reflection.Assembly.Load(AssemblyName).Location;
> ```

Unfortunately, primary scenario where we use `codebase` is a fall back from assembly load failure. `System.Reflection.Assembly.Load(AssemblyName)` failed and WinForms is trying to load assembly from codebase.

dreddy-work:
[commit](https://github.com/dotnet/winforms/commit/564f1c14c355352c598fbec2c63d1fbf25c457f4) has those instances.

elachlan:
@buyaa-n you were involved in the `AssemblyName.Codebase` obsoleting in https://github.com/dotnet/runtime/issues/42857. Do you have any suggestions on how to get a filepath from `AssemblyName`? We usually do this as a fallback from  `Assembly.Load(assemblyName)` in Winforms.

buyaa-n:
[...]

PR #8213, opened 2022-11-17
Refactor AssemblyNamesTypeResolutionService to replace HashTable

elachlan:
Replaced the Hashtables, enabled nullability, removed `AssemblyName.CodeBase` call.

Related: #8143, #7308

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8213)

### DIFF
--- a/src/System.Windows.Forms/src/System/Resources/AssemblyNamesTypeResolutionService.cs
+++ b/src/System.Windows.Forms/src/System/Resources/AssemblyNamesTypeResolutionService.cs
@@ -2,9 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
-using System.Collections;
+using System.Collections.Concurrent;
 using System.ComponentModel.Design;
 using System.Diagnostics.CodeAnalysis;
 using System.Reflection;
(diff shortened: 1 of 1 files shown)


elachlan:
@dreddy-work I think in this case, the fallback on Assembly.Load() failure looks into the names array for a matching entry and then attempts to load from the name located there.

elachlan:
@dreddy-work Flaky tests?

dreddy-work:
> @dreddy-work Flaky tests?

@elachlan , seems so. reset it for now

src/System.Windows.Forms/src/System/Resources/AssemblyNamesTypeResolutionService.cs
```diff
@@ -78,31 +65,26 @@ public Assembly GetAssembly(AssemblyName name, bool throwOnError)
         [UnconditionalSuppressMessage("SingleFile", "IL3002", Justification = "Returns null if in a single file")]
         public string GetPathOfAssembly(AssemblyName name)
         {
-#pragma warning disable SYSLIB0044 // Type or member is obsolete. Ref https://github.com/dotnet/winforms/issues/7308
-            return name.CodeBase;
-#pragma warning restore SYSLIB0044 // Type or member is obsolete
+            try
+            {
+                return Assembly.Load(name).Location;
```
dreddy-work, line 70: Public API earlier was not loading assembly but now it does. I do not know repercussions here.  @JeremyKuhne / @Tanya-Solyanik  any inputs/comments on this change?

dreddy-work: Change is helping remove obsolete API.

dreddy-work:
> @dreddy-work I think in this case, the fallback on Assembly.Load() failure looks into the names array for a matching entry and then attempts to load from the name located there.

My only concern is the change in public API now loads assembly earlier not.

src/System.Windows.Forms/src/System/Resources/AssemblyNamesTypeResolutionService.cs
elachlan: Yes, its an internal class.

dreddy-work: Assembly.Load and `LoadFrom` use different load context. In this call hierarchy, direct load failed and ending up calling with `LoadFrom.`  Accessing Location this way throw exception at best.  @elachlan did you run any scenario here?

elachlan: I did not. I'd love to find an alternative. I couldn't find a better one. Happy to leave it as it was originally, as it wasn't the main focus of the PR.

dreddy-work: Agree. Lets deal with obsolete APIs separately.

elachlan: Done.

dreddy-work (approved):

src/System.Windows.Forms/src/System/Resources/AssemblyNamesTypeResolutionService.cs
```diff
@@ -196,13 +173,10 @@ public Type GetType(string name, bool throwOnError, bool ignoreCase)
         ///  This is matching %windir%\Microsoft.NET\Framework*, so both 32bit and 64bit framework will be covered.
```
Tanya-Solyanik, line 173: @elachlan  - could you please fix this comment in a follow up PR?

src/System.Windows.Forms/src/System/Resources/AssemblyNamesTypeResolutionService.cs
Tanya-Solyanik: @elachlan . @dreddy-work  - I don't think that this use case of CodeBase is obsoleted. Docs say, don't load assembly from the Codebase path and `Do not use an AssemblyName with only the CodeBase property set.`. The harm from a new Assembly.Load is higher than that from a suppressed warning.

dreddy-work: agree and for that reason, we reverted that change.

PR #8216, opened 2022-11-17
Update MediaPlayer Test

lonitra:
Remove manually added dlls and add core project that targets Framework to reference MediaPlayer dlls. This added project will be used to add in and write tests other ActiveX controls.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8216)


src/System.Windows.Forms/tests/AxHosts/AxHosts.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System.ComponentModel;
```
dreddy-work, line 1: License?

lonitra: Thanks for the catch!

src/System.Windows.Forms/tests/AxHosts/AxHosts.cs
```diff
@@ -0,0 +1,29 @@
...
+
+using System.ComponentModel;
+using System.IO;
+using AxWMPLib;
+
+namespace AxHosts
+{
+    public partial class AxHosts : Form
+    {
+        private AxWindowsMediaPlayer _mediaPlayer;
```
RussKie, line 13:
```suggestion
        private readonly AxWindowsMediaPlayer _mediaPlayer;
```

src/System.Windows.Forms/tests/AxHosts/AxHosts.cs
```diff
@@ -0,0 +1,29 @@
...
+        {
+            InitializeComponent();
+            _mediaPlayer = new()
+            {
+                Dock = DockStyle.Fill,
+            };
+
+            ((ISupportInitialize)_mediaPlayer).BeginInit();
+            Controls.Add(_mediaPlayer);
+            ((ISupportInitialize)_mediaPlayer).EndInit();
```
RussKie, line 25: Any reason not to have this inside `InitializeComponent`?

src/System.Windows.Forms/tests/AxHosts/AxHosts.csproj
```diff
@@ -0,0 +1,36 @@
...
+    <OutputType>WinExe</OutputType>
+    <TargetFramework>net472</TargetFramework>
+    <Nullable>enable</Nullable>
+    <UseWindowsForms>true</UseWindowsForms>
+    <LangVersion>10</LangVersion>
+    <SignAssembly>false</SignAssembly>
+    <ImplicitUsings>disable</ImplicitUsings>
+    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+    <PlatformTarget>AnyCPU</PlatformTarget>
+  </PropertyGroup>
```
RussKie, line 13: I think you'd want to add these properties to avoid xlf (and remove those from the PR):
https://github.com/dotnet/winforms/blob/14e550693bc3afce8440e18136f58c01a76f11da/src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/WinformsControlsTest.csproj#L10-L11

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/WinformsControlsTest.csproj
```diff
@@ -121,12 +121,11 @@
   </ItemGroup>
 
   <ItemGroup>
-    <!-- Generated by adding a COM reference to Windows Media Player in a .NET Framework project -->
     <Reference Include="AxInterop.WMPLib">
-      <HintPath>Resources\AxInterop.WMPLib.dll</HintPath>
+      <HintPath>..\..\..\..\..\artifacts\bin\AxHosts\$(Configuration)\net472\AxInterop.WMPLib.dll</HintPath>
```
RussKie, line 125: This is rather brittle. We have MSBuild properties for these situations, e.g.:
```suggestion
      <HintPath>$(ArtifactsBinDir)\AxHosts\$(Configuration)\net472\AxInterop.WMPLib.dll</HintPath>
```

image

You can see other properties defined here: https://github.com/dotnet/arcade/blob/main/src/Microsoft.DotNet.Arcade.Sdk/tools/RepoLayout.props

JeremyKuhne:
Ugh- the [task](https://learn.microsoft.com/visualstudio/msbuild/resolvecomreference-task?view=vs-2022#msb4803-error) doesn't work from `dotnet build`. @lonitra I'll work with you tomorrow to work around this.

RussKie:
You could add a custom task that'd kick off an [MSBuild task](https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-task?view=vs-2022) or an [Exec task](https://learn.microsoft.com/en-us/visualstudio/msbuild/exec-task?view=vs-2022).
Here's an [example](https://github.com/gitextensions/gitextensions/blob/master/scripts/native.proj#L45).

JeremyKuhne:
> You could add a custom task that'd kick off an [MSBuild task](https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-task?view=vs-2022) or an [Exec task](https://learn.microsoft.com/en-us/visualstudio/msbuild/exec-task?view=vs-2022).
Here's an [example](https://github.com/gitextensions/gitextensions/blob/master/scripts/native.proj#L45).

Thanks for the links. That was one of the options I was thinking of. Would have to be the exec task to get the Framework version of MSBuild obviously. Presuming I'd have to take the Framework projects out of the solution build for the Debug/Release configurations and include this project (or something similar) to get dependencies working in VS.

JeremyKuhne:
Another, longer term solution would be to see if we can stand up a `.target` file that would do the work. I don't know that the task has anything Framework specific- it might have been excluded on Core as it has no relevance outside of Windows builds (and there was no COM support for the first few versions). Will follow up with the MsBuild folks.

src/System.Windows.Forms/tests/AxHosts/AxHosts.cs
lonitra: Adding the media player to the designer was giving me errors before, but it curiously works now so I will move it into `InitializeComponent`

RussKie (commented):
🚀

RussKie:
@lonitra this is the tool you want to have and use dealing with MSBuild related work - https://msbuildlog.com/.

RussKie:
This is still failing to build https://dev.azure.com/dnceng-public/public/_build/results?buildId=87338&view=results

JeremyKuhne (approved):
🥳

RussKie:
```
  'msbuild.exe' is not recognized as an internal or external command,
  operable program or batch file.
D:\a\_work\1\s\src\BuildAssist\BuildAssist.msbuildproj(11,5): error MSB3073: The command "msbuild.exe D:\a\_work\1\s\src\System.Windows.Forms\tests\AxHosts\AxHosts.csproj /p:Configuration=Debug /p:Platform=AnyCPU" exited with code 9009.
##[error]src\BuildAssist\BuildAssist.msbuildproj(11,5): error MSB3073: (NETCORE_ENGINEERING_TELEMETRY=Build) The command "msbuild.exe D:\a\_work\1\s\src\System.Windows.Forms\tests\AxHosts\AxHosts.csproj /p:Configuration=Debug /p:Platform=AnyCPU" exited with code 9009.
```

You need to supply the full path to msbuild. Here's an example of how to do it: https://github.com/gitextensions/gitextensions/blob/6f9591c6ad0d0d4ab6970fa1c837560a9e743e3c/scripts/native.proj#L21-L35

src/BuildAssist/BuildAssist.msbuildproj
```diff
@@ -0,0 +1,44 @@
+﻿<Project Sdk="Microsoft.Build.NoTargets/3.6.0">
+  <ItemGroup>
+    <PackageReference Include="vswhere" Version="3.1.1" IsImplicitlyDefined="true" GeneratePathProperty="true" />
```
RussKie, line 3: Please follow the way we specify versions everywhere - https://github.com/dotnet/winforms/blob/main/eng/Versions.props

src/BuildAssist/BuildAssist.msbuildproj
```diff
@@ -0,0 +1,44 @@
...
+  </ItemGroup>
+  
+  <!--
+    We cannot build Framework projects using dotnet build when there are unsupported tasks involved such as ResolveComReferences.
+    This project is an adapter to use the Framework version of msbuild from dotnet build.
+  -->
```
RussKie, line 9: This comment should be at the top. I clearly didn't have a chance to read it when I suggested adding the itemgroup at the top.

RussKie (commented):
👍

JeremyKuhne (approved):
Whew!

PR #8222, opened 2022-11-17
Refactor ComponentDesigner.ShadowPropertyCollection to use Dictionary<TKey, TValue> instead of Hashtable

Jericho:
Related: #8143

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8222)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.ShadowPropertyCollection.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.ShadowPropertyCollection.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System.Collections;
-
 namespace System.ComponentModel.Design
 {
     public partial class ComponentDesigner
@@ -14,8 +12,8 @@ namespace System.ComponentModel.Design
         protected sealed class ShadowPropertyCollection
         {
             private readonly ComponentDesigner _designer;
-            private Hashtable _properties;
-            private Hashtable _descriptors;
+            private Dictionary<string, PropertyDescriptor> _properties;
+            private Dictionary<string, PropertyDescriptor> _descriptors;
 
             internal ShadowPropertyCollection(ComponentDesigner designer) => _designer = designer;
 
@@ -30,9 +28,9 @@ namespace System.ComponentModel.Design
                     ArgumentNullException.ThrowIfNull(propertyName);
 
                     // First, check to see if the name is in the given properties table
-                    if (_properties is not null && _properties.ContainsKey(propertyName))
+                    if (_properties is not null && _properties.TryGetValue(propertyName, out PropertyDescriptor property))
                     {
-                        return _properties[propertyName];
+                        return property;
                     }
 
                     // Next, check to see if the name is in the descriptors table.  If it isn't, we will search the
@@ -43,8 +41,8 @@ namespace System.ComponentModel.Design
                 }
                 set
                 {
-                    _properties ??= new Hashtable();
-                    _properties[propertyName] = value;
+                    _properties ??= new();
+                    _properties[propertyName] = (PropertyDescriptor)value;
                 }
             }
 
@@ -58,12 +56,11 @@ namespace System.ComponentModel.Design
             /// </summary>
             private PropertyDescriptor GetShadowedPropertyDescriptor(string propertyName)
             {
-                _descriptors ??= new Hashtable();
+                _descriptors ??= new();
 
-                PropertyDescriptor property = (PropertyDescriptor)_descriptors[propertyName];
-                property ??= TypeDescriptor.GetProperties(_designer.Component.GetType())[propertyName];
-
-                return property;
+                return _descriptors.TryGetValue(propertyName, out PropertyDescriptor property)
+                    ? property
+                    : TypeDescriptor.GetProperties(_designer.Component)[propertyName];
             }
 
             /// <summary>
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.ShadowPropertyCollection.cs
```diff
@@ -58,12 +56,14 @@ protected sealed class ShadowPropertyCollection
...
             private PropertyDescriptor GetShadowedPropertyDescriptor(string propertyName)
             {
-                _descriptors ??= new Hashtable();
+                _descriptors ??= new();
 
-                PropertyDescriptor property = (PropertyDescriptor)_descriptors[propertyName];
-                property ??= TypeDescriptor.GetProperties(_designer.Component.GetType())[propertyName];
+                if (_descriptors.TryGetValue(propertyName, out PropertyDescriptor property))
+                {
+                    return property;
```
dreddy-work, line 63: could `property `be null here?

elachlan: `_descriptors` is never added to and seems redundant? Maybe one of the @dotnet/dotnet-winforms team members knows whats going on there.

dreddy-work: Given setter is public and may allow null, What do we lose if we like to keep existing functionality?

elachlan: I think its designed as a cache, so it should probably add to it further down. But I am not 100% sure of that.

Jericho: The values stored in the Hashtable (or Dictionary) are not null and this particular dictionary is not publicly exposed (the other dictionary is exposed though).

I was about to point out that the `_descriptors` Hashtable (or Dictionary) is never assigned to but @elachlan beat me to it! This means that it's always empty! So the current logic is flawed. My guess is the original intent was to use `_descriptors` as a cache to store the values returned by `TypeDescriptor.GetProperties(_designer.Component.GetType())[propertyName]` so that it would be much faster to subsequently retrieve the values. Again, I'm just guessing, but my guess is the original author of this code forgot to add the values to the Hashtable and nobody ever noticed this omission.

I'll be happy to fix this omission by caching the values into the Dictionary.

Jericho: I propose fixing the omission like so:

```csharp
if (!_descriptors.TryGetValue(propertyName, out PropertyDescriptor property))
{
    property = TypeDescriptor.GetProperties(_designer.Component.GetType())[propertyName];
    _descriptors.Add(propertyName, property);
}

return property;
```

elachlan: I think it might be better if we create an issue for that specific change and keep the existing logic. I think we would need a bit of investigation to see if this would interfere with object lifecycles.

dreddy-work: @Jericho / @elachlan , looking at the various references, `_descriptors `could be populated with [null values](https://github.com/dotnet/winforms/blob/7d1bd879fe2dd030f5d36a8cb25ef1b352e90c60/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/FormDocumentDesigner.cs#L341) and [here](https://github.com/dotnet/winforms/blob/7d1bd879fe2dd030f5d36a8cb25ef1b352e90c60/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ToolStripDropDownDesigner.cs#L576) not just missing descriptors. Call sites of this method was not expecting null in all cases though. Needs further investigation before updating functionality.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.ShadowPropertyCollection.cs
```diff
@@ -43,7 +41,7 @@ protected sealed class ShadowPropertyCollection
                 }
                 set
                 {
-                    _properties ??= new Hashtable();
+                    _properties ??= new();
                     _properties[propertyName] = value;
```
elachlan, line 45: @dreddy-work Should we add a null check on value here to avoid the null problem?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.ShadowPropertyCollection.cs
```diff
@@ -30,9 +28,9 @@ protected sealed class ShadowPropertyCollection
                     ArgumentNullException.ThrowIfNull(propertyName);
 
                     // First, check to see if the name is in the given properties table
-                    if (_properties is not null && _properties.ContainsKey(propertyName))
+                    if (_properties is not null && _properties.TryGetValue(propertyName, out object prop))
                     {
-                        return _properties[propertyName];
+                        return prop;
```
JeremyKuhne, line 33: You could make this `property` by getting rid of the other local or use a descriptive term like `existing`.

JeremyKuhne (changes requested):
Just a minor nit.

JeremyKuhne (approved):

PR #8226, opened 2022-11-17
Refactor WebBrowserContainer to replace HashTable and Enable Nullability

elachlan:
Refactored WebBrowserContainer to replace HashTable and Enable Nullability.

Related to #8143

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8226)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/WebBrowserContainer.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/WebBrowserContainer.cs
@@ -392,7 +392,7 @@ namespace System.Windows.Forms
         }
 
         internal static string GetNameForControl(Control ctl)
-            => ctl.Site is not null ? ctl.Site.Name ?? string.Empty : ctl.Name;
+            => ctl.Site is not null ? ctl.Site.Name ?? string.Empty : ctl.Name ?? string.Empty;
 
         internal void OnUIActivate(WebBrowserBase site)
         {
--- a/src/System.Windows.Forms/src/System/Windows/Forms/WebBrowserContainer.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/WebBrowserContainer.cs
@@ -198,7 +198,7 @@ namespace System.Windows.Forms
                 return;
             }
 
-            foreach (Control ctl in components)
+            foreach (Control ctl in components.ToArray())
             {
                 if (ctl is WebBrowserBase webBrowserBase)
                 {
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/WebBrowserContainer.cs
```diff
@@ -193,84 +190,67 @@ unsafe HRESULT Ole32.IOleInPlaceFrame.TranslateAccelerator(MSG* lpmsg, ushort wI
...
         {
-            Hashtable components = GetComponents();
+            HashSet<Control>? components = GetComponents();
             if (components is null)
             {
                 return;
             }
 
-            Control[] ctls = new Control[components.Keys.Count];
-            components.Keys.CopyTo(ctls, 0);
```
dreddy-work, line 205: Would there be any reason that `components `could be changed while we are processing here? It seems intentional copy .

src/System.Windows.Forms/src/System/Windows/Forms/WebBrowserContainer.cs
```diff
@@ -418,10 +392,7 @@ internal static WebBrowserContainer FindContainerForControl(WebBrowserBase ctl)
         }
 
         internal static string GetNameForControl(Control ctl)
-        {
-            string name = (ctl.Site is not null) ? ctl.Site.Name : ctl.Name;
-            return name ?? "";
-        }
+            => ctl.Site is not null ? ctl.Site.Name ?? string.Empty : ctl.Name;
```
dreddy-work, line 395: `ctl.Name` can not be null?

elachlan: This just condenses the existing logic. `ctl.Name` is only used when `ctl.Site` is null. Otherwise `ctl.Site.Name` is used and that can be `null`, so it returns `string.Empy` in that case.

dreddy-work: right, earlier, we never return null. now we may be.

dreddy-work (approved):
LGTM.

src/System.Windows.Forms/src/System/Windows/Forms/WebBrowserContainer.cs
elachlan: I've updated it to check if ctl.Name is null and return string.empty. That should handle it.

PR #8230, opened 2022-11-18
Enable nullability in remaining ToolStripItem members

gpetrou:
## Proposed changes

- Enable nullability in remaining ToolStripItem members.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8230)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
@@ -2981,11 +2981,11 @@ namespace System.Windows.Forms
                 OnBackColorChanged(e);
             }
         }
-#nullable disable
+
         /// <summary>
         ///  Inheriting classes should override this method to handle this event.
         /// </summary>
-        protected virtual void OnParentChanged(ToolStrip oldParent, ToolStrip newParent)
+        protected virtual void OnParentChanged(ToolStrip? oldParent, ToolStrip? newParent)
         {
             SetAmbientMargin();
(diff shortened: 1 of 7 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
```diff
@@ -3068,7 +3068,7 @@ internal void OnOwnerTextDirectionChanged()
             ToolStripTextDirection textDirection = ToolStripTextDirection.Inherit;
             if (Properties.ContainsObject(ToolStripItem.s_textDirectionProperty))
             {
-                textDirection = (ToolStripTextDirection)Properties.GetObject(ToolStripItem.s_textDirectionProperty);
+                textDirection = (ToolStripTextDirection)Properties.GetObject(ToolStripItem.s_textDirectionProperty)!;
```
dreddy-work, line 3071: we were not checking for non-null on `textDirection `earlier. What causes you to use `!` here?

gpetrou: `Properties.GetObject` returns a nullable object, but `ToolStripTextDirection` is an enum, so we can use the null-forgiving operator here. We have used the same approach in other places.

dreddy-work (approved):

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
dreddy-work: Thank you!.

PR #8235, opened 2022-11-19
Refactor BindingContext to use Dictionary<TKey, TValue> instead of Hashtable and List<T> instead of ArrayList

Jericho:
Related: #8143 and #8140


elachlan:
Looks like tests are checking for `DictionaryEntry` instead of `KeyValuePair`. 

```
System.Windows.Forms.Tests.BindingContextTests.BindingContext_CopyTo_WithNullWeakReferenceTarget_ScrubsWeakRefs
Assert.IsType() Failure\r\nExpected: System.Collections.DictionaryEntry\r\nActual:   System.Collections.Generic.KeyValuePair`2[[System.Windows.Forms.BindingContext+HashKey, System.Windows.Forms, Version=8.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089],[System.WeakReference, System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]]
```

Can you also move `HashKey` to a partial class file?

@dreddy-work, this might be a public API change? The associated test is testing `ICollection.CopyTo`. The test code is expecting a `DictionaryEntry` but getting a `KeyValuePair`.

elachlan:
> > @dreddy-work, this might be a public API change? The associated test is testing `ICollection.CopyTo`. The test code is expecting a `DictionaryEntry` but getting a `KeyValuePair`.
> 
> At first glance, i do not see any public API surface change here. Need to check tests on why specific types are checked. Will take a look later.

The API design doesn't guarantee a type because the collection before wasn't typed. Even though it was only storing a particular set of values. Now generics are being used underneath, so the values in the collection have changed.

So in my mind its okay. Since no gaurantees were made, but it would cause issue for people relying on code returning a dictionary entry. This will be a problem for most collections we convert I think.

dreddy-work:
> This will be a problem for most collections we convert I think.

Agree. Tests need to be updated to account this change. @Jericho , can you make necessary changes in tests?

Jericho:
I'm trying to replace `DictionaryEntry` in this line:
```csharp
Assert.NotNull(Assert.IsType<DictionaryEntry>(array[1]).Key);
```
with the following:
```csharp
Assert.NotNull(Assert.IsType<KeyValuePair<HashKey, WeakReference>>(array[1]).Key);
```

but the `HashKey` type is a private class, therefore not accessible. My change results in the following exception: `CS0246 The type or namespace name 'HashKey' could not be found (are you missing a using directive or an assembly reference?)`.

To be honest I don't understand the logic of `BindingContext` being a public class but the data type of its enumerator key being a private type? To be clear: this problem existed before I changed anything, but it was obscured by the fact the data type of a `DictionaryEntry` key is `object` and the fact that none of the unit tests attempt to verify the data type of the key.

To further illustrate what I'm talking about, let's add the following assertion to the `BindingContext_CopyTo_Invoke_Success` unit test (in a branch that does not include my PR):
```csharp
Assert.NotNull(Assert.IsType<HashKey>(Assert.IsType<DictionaryEntry>(array[1]).Key));
```
You'll get the same `CS0246 The type or namespace name 'HashKey' could not be found (are you missing a using directive or an assembly reference?)` exception.

dreddy-work:
> ```cs
> Assert.NotNull(Assert.IsType<KeyValuePair<HashKey, WeakReference>>(array[1]).Key);
> ```

Tests seems simulating GC collection and Key was never checked for type in these tests. Did you try replacing  `HashKey `with object in `KeyValuePair` as work around here?  @Tanya-Solyanik / @KlausLoeffelmann  do you guys have any suggestion here?

elachlan:
Tests are supposed to have visibility via [InternalsVisibleToAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.internalsvisibletoattribute?view=net-7.0)

dreddy-work:
> Tests are supposed to have visibility via [InternalsVisibleToAttribute](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.internalsvisibletoattribute?view=net-7.0)

`HashKey `has been private.

Jericho:
@dreddy-work : I replaced `Hashkey` with `object` as you suggested but unit test fails:
```
Message: 
Assert.IsType() Failure
Expected: System.Collections.Generic.KeyValuePair`2[[System.Object, System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.WeakReference, System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]]
Actual:   System.Collections.Generic.KeyValuePair`2[[System.Windows.Forms.BindingContext+HashKey, System.Windows.Forms, Version=8.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089],[System.WeakReference, System.Private.CoreLib, Version=7.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]]
```

@elachlan : The class in question is defined like this: `private class HashKey` and as far as I know `InternalsVisibleTo` only helps with `internal` members.

Even if we figured out a way to fix the unit test, I think the larger question remains : does it make sense for the public member of a public class to contain objects of a private type? I have to admit that I have never seen this scenario, and I fail to understand how this make sense.

elachlan:
Its exposed either way.
In the test:
image

In a test app using .NET 7:
image

Developer experience and the API are already partially "broken", because you won't be able to use anything in `Key`.

Can we cast the `KeyValuePair` to `DictionaryEntry` to solve the test issue? Otherwise we'd have to expose HashKey as Internal.

Jericho:
> Can we cast the `KeyValuePair` to `DictionaryEntry` to solve the test issue? Otherwise we'd have to expose HashKey as Internal.

How about this:

```csharp
void ICollection.CopyTo(Array ar, int index)
{
    ScrubWeakRefs();

    int i = index;
    foreach (KeyValuePair<HashKey, WeakReference> kvp in _listManagers)
    {
        ar.SetValue(new DictionaryEntry(kvp.Key, kvp.Value), i++);
    }
}

IEnumerator IEnumerable.GetEnumerator()
{
    ScrubWeakRefs();
    return _listManagers
        .Select(kvp => new DictionaryEntry(kvp.Key, kvp.Value))
        .GetEnumerator();
}
```

I tested it and it resolves all the failing unit tests.

KlausLoeffelmann:
So, what's the current state here:
Are collection items exposed publicly, and would this PR change the type of one item?

Jericho:
> So, what's the current state here: Are collection items exposed publicly, and would this PR change the type of one item?

The status is that [I made a suggestion](https://github.com/dotnet/winforms/pull/8235#issuecomment-1326623180) that would ensure the type of the items publicly exposed would be preserved.

Having said that, I want to be clear that the situation we are trying to preserve is that instances of a private type are being publicly exposed (which is very weird to me!), but I understand that fixing this weird situation would be a breaking change and I don't know if the team has any appetite for such a change.

At this point I'm waiting for feedback on the solution I proposed and/or guidance on what the team wants to do.

KlausLoeffelmann:
>  I don't know if the team has any appetite for such a change.

The team does absolutely not (me as part of the team and focusing on Data Binding both in the Designer and in the runtime, know that pretty well!). The reason is: Especially around data binding, it would be not only a breaking change, but also such a subtle one, that it would become a potential migration blocker, and it would be particularly bad for large LOB apps.

Jericho:
> The team does absolutely not

Understood!

For my own education, do you know why instances of the private `HashKey` type are being publicly exposed? The fact that the type is private, as far as I understand, means that these instances are unusable so this whole situation does not make a lot of sense to me. I am assuming there is some legacy reason and, as you pointed out, the cost of "fixing" it is simply too high.

KlausLoeffelmann:
> For my own education, do you know why instances of the private HashKey type are being publicly exposed? The fact that the type is private, as far as I understand, means that these instances are unusable so this whole situation does not make a lot of sense to me. I am assuming there is some legacy reason and, as you pointed out, the cost of "fixing" it is simply too high.

Welcome to the club! 😄

The "original-original" code base is over 20 years old, and Data Binding is one of the earliest features, because it was important to have it in for (at least concept-) backwards compatibility to VB6 design scenarios. So, when we have head-scratchers like this (and I agree with you here, no doubt), the first thing I do is try to put this in the context of the time: First version (maybe first _beta_ version) of C#, lack of experience, lack of ready NetFX APIs - maybe? But, yes there might be even more to it - but we wouldn't necessarily know. Sure, we have internal design documents from the time, but they don't have the granularity needed to explain every single implementation-detail-design decision. So, on that level, some things will remain a mystery when it comes to the motivation behind a technical implementation decision. 

Now, that all said, the moment an implementation status-quo could become a potential security risk (which I - on first glance here - do not think it does), our decision process certainly must completely change, and at that point that wins over avoiding breaking changes. And I am also not per se against breaking changes, as long as they can be easily discovered and fixed. But that exactly is my concern in this scenario. Would you agree on that?

Jericho:
Yes, I completely agree. I was just trying to understand the "reasoning" (if you can call it that!) behind what appears to be a non-sensical situation. I appreciate the importance of preserving (as much as reasonably possible) the status-quo unless there's a security concern.

By the way, thank you for taking the time to explain, I really appreciate it. Many lifetimes ago I wrote VB6 LOB apps for a living and also taught VB at a local college, and I do remember how smooth the transition was when I switched to c# and winform in 2001. So, if some of the "quirks" we find in the codebase today are due to an effort to preserve backward compatibility to VB6 back then, my "20-year-younger-self" is grateful!

KlausLoeffelmann:
> So, if some of the "quirks" we find...

While you mentioned quirks (even though in a different context): 

When a new feature would have the potential of (somewhat noticeable at least) breaking changes but is still too useful to _not_ make it, 'quirks' are always something, we can and should consider. Folks could then with a switch opt in or out of a new feature to preserve the old behavior. These switches we call quirks.

elachlan:
I mean, technically its not breaking the public contract. Since `ICollection` has no type guarantees. But the traditionally returned types from `CopyTo` will change.

I wrote a test to double check, but going from a constrained kyvaluepair to DictionaryEntry is fine. Because the types are becoming more lax (unsure on formal name).

This test worked:
```csharp
Dictionary<int, string> test = new();
test.Add(1, "test1");
test.Add(2, "test2");
test.Add(3, "test3");

DictionaryEntry[] items = new DictionaryEntry[test.Count];
((ICollection)test).CopyTo(items, 0);

foreach (var item in items)
{
    Console.WriteLine(item.Value);
}
```

The only issue is if they are checking the type. As our tests are doing for whatever reason.

KlausLoeffelmann:
So, just to understand correctly, because I haven't really caught up with reading the whole thread, yet:
`CopyTo` aside, because that's not soooo much my concern:

Iterating over a `BindingContext` instance used to yield `DictionaryEntry` items before and would now yield `KeyValuePair` items, is that correct? 

To change that, you would convert the whole dictionary in `GetEnumerator` to return `DictionaryEntry` items?

What actual benefit would this PR then leave us with?

elachlan:
Just tested `GetEnumerator`, it will break existing code if we expose it like we are with `InvalidCastException`. So we can't make this change.

Jericho:
@elachlan I'll go ahead and commit [my proposed changes](https://github.com/dotnet/winforms/pull/8235#issuecomment-1326623180) which resolve the problems with `CopyTo` and `GetEnumerator`.

elachlan:
Errors in build: `Converting null literal or possible null value to non-nullable type.`

elachlan:
More build errors.

lonitra (dismissed):
This LGTM!

KlausLoeffelmann:
Well, I would like to have an answer to my question?
What is the actual benefit of the change?

KlausLoeffelmann:
To be more precise:

image

So, it seems, when we're actually iterating over `BindingContext` we're allocating because we convert?
Did I get this correctly?

dreddy-work:
> What is the actual benefit of the change?

@KlausLoeffelmann, we are replacing `HashTable` and `ArrayLists` with generics across code base and this change is inspired from that goal. Do you have any suggestions here? I am looking if there is a way to avoid extra allocation or its impact on perf.

elachlan:
I just test this:
image

I think instead of projecting using linq, its better to cast to `IDictionary`. I am unsure if it just does the same thing though.

src/System.Windows.Forms/src/System/Windows/Forms/BindingContext.cs
```diff
@@ -53,7 +58,9 @@ void ICollection.CopyTo(Array ar, int index)
         IEnumerator IEnumerable.GetEnumerator()
         {
             ScrubWeakRefs();
-            return _listManagers.GetEnumerator();
+            return _listManagers
```
JeremyKuhne, line 61: If you cast to `IDictionary` to get the enumerator you'll get an enumerator that works in `DictionaryEntry`.

src/System.Windows.Forms/src/System/Windows/Forms/BindingContext.cs
```diff
@@ -44,7 +44,12 @@ int ICollection.Count
         void ICollection.CopyTo(Array ar, int index)
         {
             ScrubWeakRefs();
-            _listManagers.CopyTo(ar, index);
+
+            int i = index;
```
JeremyKuhne, line 48: As below, if you cast to `IDictionary` you'll get `DictionaryEntry`.

You'll also need to do parameter validation to get the same behavior as `Hashtable` and write tests to validate.  https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Collections/Hashtable.cs,568

JeremyKuhne:
Presumably we have the same compat related issues with anything we're changing from `HashTable` to `Dictionary<TKey, TValue>` and exposing access to. This adapting should be done in centralized helper code that is thoroughly tested to be compatible with the behavior of `HashTable`. Exceptions should be the same, types should be the same.

When we expose the actual `HashTable` as a modifiable interface we can not change from it to `Dictionary<>`. For example, if we give the nested type back as `ICollection` we cannot have a copy of the collection.

JeremyKuhne (changes requested):
The public interfaces must be fully compatible with prior behavior. See my comments.

src/System.Windows.Forms/src/System/Windows/Forms/BindingContext.cs
Jericho: Done.

dreddy-work (dismissed):

src/System.Windows.Forms/src/System/Windows/Forms/BindingContext.cs
Jericho: Done.

src/System.Windows.Forms/src/System/Windows/Forms/BindingContext.cs
```diff
@@ -44,7 +44,7 @@ int ICollection.Count
         void ICollection.CopyTo(Array ar, int index)
         {
             ScrubWeakRefs();
-            _listManagers.CopyTo(ar, index);
+            ((IDictionary)_listManagers).CopyTo(ar, index);
```
JeremyKuhne, line 47: This is not sufficient. If the array is `object` you will now get different behavior.

JeremyKuhne (changes requested):
See my comment on `object[]`

src/System.Windows.Forms/src/System/Windows/Forms/BindingContext.cs
dreddy-work, line 47: @Jericho , as Jermey pointed out above, it seems HashSet CopyTo is different to Icollection.CopyT0
https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/Dictionary.cs,1034

For this PR, we may want to copy the logic from HashSet and override this method.

We should do something like below here.
```suggestion
            if (array == null)
            {
                ThrowHelper.ThrowArgumentNullException(ExceptionArgument.array);
            }
 
            if (array.Rank != 1)
            {
                ThrowHelper.ThrowArgumentException(ExceptionResource.Arg_RankMultiDimNotSupported);
            }
 
            if (array.GetLowerBound(0) != 0)
            {
                ThrowHelper.ThrowArgumentException(ExceptionResource.Arg_NonZeroLowerBound);
            }
 
            if ((uint)index > (uint)array.Length)
            {
                ThrowHelper.ThrowIndexArgumentOutOfRange_NeedNonNegNumException();
            }
 
            if (array.Length - index < Count)
            {
                ThrowHelper.ThrowArgumentException(ExceptionResource.Arg_ArrayPlusOffTooSmall);
            }
			
			var enumerator = _listManagers.GetEnumerator();
			int arraPosition = 0;
			while (enumerator.MoveNext())
			{
			     if (arraPosition < index)
			     {
			         arraPosition ++;
			         continue;
			     }
			     
			     KeyValuePair keyValuePair = enumerator.Current;
			     if (keyValuePair.Key is not null)
			     {
			          ar.SetValue(keyValuePair.Value, arraPosition )
			          arraPosition ++;
			      }
			   }
```

https://github.com/dotnet/winforms/issues/8357

elachlan: Is performance effected if we make it an internal extension method? Also did you mean `HashTable` not `HashSet`?

dreddy-work: it was HashTable reference.

dreddy-work: Nope. Not necessary. Its good that we identified this now. It is going to help with guidance we plan to provide further.

Jericho: @dreddy-work with the exception of parameter validation, the code you just pointed me to is conceptually identical to what I [proposed](https://github.com/dotnet/winforms/pull/8235#issuecomment-1326623180) two weeks ago and [committed](https://github.com/dotnet/winforms/pull/8235/commits/5c1a6cfcbe46aa987fcb1672c9d7c449951c908f) a few days later. To be clear, @JeremyKuhne asked me to change this code during his review and now you are asking me to go back to this logic?

I'll add the parameter validation which is a nice improvement over what I had proposed but other than that, we are essentially going back to:
- looping through the enumeration
- instantiating a new DictionaryEntry for each item
- updating the Array with the newly instantiated dictionary entry

Are we all in agreement?

Jericho: @dreddy-work I just realized that I could simply accept your proposed code change. Thanks for doing that, makes it easy for me!

The gist of my previous comment remains though: I want to make sure everyone is onboard with going back to "looping, instantiating, updating" especially since that's what I had proposed some time ago and I was asked to change this logic.

JeremyKuhne: @Jericho

To be clear, casting to `IDictionary` for the enumerator is fine, as the behavior matches what `HashTable` does. That's the fundamental piece here- We should match `HashTable`'s behavior on things we were getting out directly from `HashTable` and returning through public API.

As far as `CopyTo` is concerned, the minbar is that it does at least as much parameter validation and matches what happens when you pass in `object[]` and `DictionaryEntry[]`. We _can_ also add the ability to handle `KeyValuePair<TKey, TValue>[]` and probably should (even though here that isn't terribly useful as `TKey` is not public). What we don't want is a currently working positive scenario to no longer work.

This code should be in a static helper class. Maybe `HashTableAdapter` or something like that. We could potentially derive our own `Dictionary` but that doesn't seem like a good idea. Changing the behavior of existing methods undoes some of the benefits of having a typed collection and having additional methods doesn't add as much clarity in the source as to where we've done something different.

I'm sorry I didn't catch these earlier- I've been pretty heavily invested in establishing our CsWin32 COM patterns.

Jericho: > To be clear, casting to `IDictionary` for the enumerator is fine, 

Excellent, thanks for confirming.

> As far as `CopyTo` is concerned, the minbar is that it does at least as much parameter validation and matches what happens when you pass in `object[]` and `DictionaryEntry[]`. 

Looking at the source for `HashTable.CopyTo`, the parameter validations are:
1. `ArgumentNullException.ThrowIfNull(array);`
2. `if (array.Rank != 1)`
3. `if (arrayIndex < 0)`
4. `if (array.Length - arrayIndex < Count)`

The code suggested by @dreddy-work has two additional checks:
5. if (array.GetLowerBound(0) != 0)
6. if ((uint)index > (uint)array.Length)

Are you suggesting some aditional checks? Such as checking the data type in `Array` for instance?

> This code should be in a static helper class. Maybe `HashTableAdapter` or something like that. 

Makes sense, I'll create a new class.

Speaking of argument validation, I am assuming that hard coding exception messages is frowned upon. I'm guessing I should add the message to the `SR` resource, am I right?

src/System.Windows.Forms/src/System/Windows/Forms/BindingContext.cs
```diff
@@ -44,7 +44,37 @@ int ICollection.Count
         void ICollection.CopyTo(Array ar, int index)
         {
             ScrubWeakRefs();
-            _listManagers.CopyTo(ar, index);
+
+            ArgumentNullException.ThrowIfNull(ar);
```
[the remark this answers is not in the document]
Jericho: I agree. I'm on it...

src/System.Windows.Forms/src/System/Windows/Forms/BindingContext.cs
```diff
@@ -44,7 +44,37 @@ int ICollection.Count
...
+
+            if (ar.GetLowerBound(0) != 0)
+                throw new ArgumentException("The lower bound of target array must be zero.", nameof(ar));
+
+            ArgumentOutOfRangeException.ThrowIfGreaterThan((uint)index, (uint)ar.Length, nameof(index) );
+
+            if (ar.Length - index < _listManagers.Count)
+                throw new ArgumentException("Destination array is not long enough to copy all the items in the collection. Check array index and length.");
+
+            var enumerator = _listManagers.GetEnumerator();
```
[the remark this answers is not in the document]
Jericho: I found the relevant `Dictionary<TKey, TValue>` source, now I understand what you mean! I'm on it...

src/System.Windows.Forms/src/misc/CollectionHelper.cs
```diff
@@ -0,0 +1,66 @@
+﻿namespace System.Collections.Generic
```
dreddy-work, line 1: license?

src/System.Windows.Forms/tests/UnitTests/misc/CollectionHelperTests.cs
```diff
@@ -0,0 +1,154 @@
...
+    public class CollectionHelperTests
+    {
+        [Fact]
+        public void HashTableCopyTo_throws_when_target_is_null()
+        {
+            var source = new Dictionary<string, string>();
+            var target = (object[])null;
+            var index = 0;
+
+            var ex = Assert.Throws<ArgumentNullException>(() => source.HashTableCopyTo(target, index));
```
dreddy-work, line 15: just Assert no assignment to local

JeremyKuhne, line 15: These tests are great, but I'd add checks for an actual `HashTable` in each of these methods so the relationship is fully expressed in code.

JeremyKuhne (changes requested):
This is looking great! The only thing I'd add is an actual HashTable in the individual tests to assert the same things against it.

src/System.Windows.Forms/tests/UnitTests/misc/CollectionHelperTests.cs
Jericho: do you mean that you would like to see unit test similar to this:
```csharp
var sourceDictionary = new Dictionary<string, string>();
var sourceHashtable = new Hashtable();
var target = (object[])null;
var index = 0;

Assert.Throws<ArgumentNullException>(() => sourceDictionary.HashtableCopyTo(target, index));
Assert.Throws<ArgumentNullException>(() => sourceHashtable.CopyTo(target, index));
```
In order to prove that my `HashtableCopyTo` extension method behaves similarly to `Hashtable.CopyTo`?

src/System.Windows.Forms/tests/UnitTests/misc/CollectionHelperTests.cs
```diff
@@ -0,0 +1,158 @@
...
+    public class CollectionHelperTests
+    {
+        [Fact]
+        public void HashtableCopyTo_throws_when_target_is_null()
+        {
+            var source = new Dictionary<string, string>();
+            var target = (object[])null;
+            var index = 0;
+
+            Assert.Throws<ArgumentNullException>(() => source.HashtableCopyTo(target, index));
```
dreddy-work, line 19: We could avoid local and follow below across these tests.

```suggestion
            Assert.Throws<ArgumentNullException>(() => source.HashtableCopyTo(target, index: 0));
```

src/System.Windows.Forms/tests/UnitTests/misc/CollectionHelperTests.cs
```diff
@@ -0,0 +1,158 @@
...
+using Xunit;
+
+namespace System.Windows.Forms.misc.Tests
+{
+    public class CollectionHelperTests
+    {
+        [Fact]
+        public void HashtableCopyTo_throws_when_target_is_null()
+        {
+            var source = new Dictionary<string, string>();
```
dreddy-work, line 15: WinForms is unifying declarations as below across repo. Please update tests to this syntax wherever possible.

```suggestion
            Dictionary<string, string> source = new();
```

src/System.Windows.Forms/tests/UnitTests/misc/CollectionHelperTests.cs
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        [Fact]
+        public void HashtableCopyTo_throws_when_target_is_null_object_array()
+        {
+            var source = new Dictionary<string, string>();
+            var target = (object[])null;
+            var index = 0;
+
+            Assert.Throws<ArgumentNullException>(() => source.HashtableCopyTo(target, index));
```
dreddy-work, line 83:
```suggestion
            Assert.Throws<ArgumentNullException>(() => source.HashtableCopyTo(target: null, index: 0));
```

src/System.Windows.Forms/src/misc/CollectionHelper.cs
```diff
@@ -0,0 +1,70 @@
...
+                    if (kvp.Key is not null)
+                    {
+                        dictionaryTarget[index++] = new DictionaryEntry(kvp.Key, kvp.Value);
+                    }
+                }
+            }
+            else
+            {
+                object[]? objects = target as object[];
+                if (objects is null)
```
dreddy-work, line 48:
```suggestion
                if (target is not object[] objects)
```

src/System.Windows.Forms/src/misc/CollectionHelper.cs
```diff
@@ -0,0 +1,70 @@
...
+            if (target.GetLowerBound(0) != 0)
+                throw new ArgumentException(SR.Arg_NonZeroLowerBound, nameof(target));
+            if (target.Length - index < source.Count)
+                throw new ArgumentException(SR.Arg_ArrayPlusOffTooSmall);
+
+            if (target is KeyValuePair<TKey, TValue>[] pairsTarget)
+            {
+                foreach (KeyValuePair<TKey, TValue> kvp in source)
+                {
+                    pairsTarget[index++] = kvp;
```
dreddy-work, line 32: why are we not checking for `key `being null here while we do check for other types? @JeremyKuhne.

src/System.Windows.Forms/tests/UnitTests/misc/CollectionHelperTests.cs
```diff
@@ -0,0 +1,149 @@
...
+        public void HashtableCopyTo_successfully_copies_to_Object_array()
+        {
+            var source = new Dictionary<string, string>()
+            {
+                { "key-one", "value-one" },
+                { "key-two", "value-two" },
+            };
+            var target = new DictionaryEntry[2];
+
+            source.HashtableCopyTo(target, 0);
```
dreddy-work, line 136:
```suggestion
            source.HashtableCopyTo(target, index: 0);
```
constants should always carry param names for readability.

src/System.Windows.Forms/src/misc/CollectionHelper.cs
Jericho: We check if `key` is null on lines 39 and 57 because the key is used to instantiate a new DictionaryEntry. Without the check for null, Visual Studio would issue the following complaint: `CS8604 Possible null reference argument for parameter 'key' in 'DictionaryEntry.DictionaryEntry(object key, object? value)'.`

src/System.Windows.Forms/tests/UnitTests/misc/CollectionHelperTests.cs
```diff
@@ -0,0 +1,146 @@
...
+{
+    public class CollectionHelperTests
+    {
+        [Fact]
+        public void HashtableCopyTo_throws_when_target_is_null()
+        {
+            Dictionary<string, string> source = new();
+            object[] target = null;
+
+            Assert.Throws<ArgumentNullException>(() => source.HashtableCopyTo(target, index: 0));
```
dreddy-work, line 18:
```suggestion
            Assert.Throws<ArgumentNullException>(() => source.HashtableCopyTo(target: null, index: 0));
```

src/System.Windows.Forms/tests/UnitTests/misc/CollectionHelperTests.cs
```diff
@@ -0,0 +1,146 @@
...
+            Assert.Throws<ArgumentException>(() => source.HashtableCopyTo(target, index: 2));
+        }
+
+        [Fact]
+        public void HashtableCopyTo_throws_when_target_is_null_object_array()
+        {
+            Dictionary<string, string> source = new();
+            var target = (object[])null;
+
+            Assert.Throws<ArgumentNullException>(() => source.HashtableCopyTo(target, index: 0));
```
dreddy-work, line 76:
```suggestion
            Assert.Throws<ArgumentNullException>(() => source.HashtableCopyTo(target: null, index: 0));
```

src/System.Windows.Forms/tests/UnitTests/misc/CollectionHelperTests.cs
```diff
@@ -0,0 +1,146 @@
...
+                { "key-two", "value-two" },
+            };
+            var target = new KeyValuePair<string, string>[2];
+
+            source.HashtableCopyTo(target, index: 0);
+
+            var firstTargetItem = target[0];
+            Assert.Equal(typeof(KeyValuePair<string, string>), firstTargetItem.GetType());
+            Assert.Equal("key-one", firstTargetItem.Key);
+            Assert.Equal("value-one", firstTargetItem.Value);
```
dreddy-work, line 94:
```suggestion
            var firstSourceItem = source[0];
            Assert.Equal(firstSourceItem.Key, firstTargetItem.Key);
            Assert.Equal(firstSourceItem.Value, firstTargetItem.Value);
```

src/System.Windows.Forms/tests/UnitTests/misc/CollectionHelperTests.cs
```diff
@@ -0,0 +1,146 @@
...
+
+            var secondTargetItem = target[1];
+            Assert.Equal(typeof(KeyValuePair<string, string>), secondTargetItem.GetType());
+            Assert.Equal("key-two", secondTargetItem.Key);
```
dreddy-work, line 98: Same here.

src/System.Windows.Forms/src/System/Windows/Forms/BindingContext.cs
```diff
@@ -327,21 +321,20 @@ private static void CheckPropertyBindingCycles(BindingContext newBindingContext,
...
                 }
             }
 
             if (cleanupList is not null)
             {
-                foreach (object o in cleanupList)
+                foreach (HashKey o in cleanupList)
                 {
                     _listManagers.Remove(o);
                 }
```
dreddy-work, line 340:
```suggestion
                foreach (HashKey key in cleanupList)
                {
                    _listManagers.Remove(key);
                }
```

src/System.Windows.Forms/src/System/Windows/Forms/BindingContext.cs
```diff
@@ -327,21 +321,20 @@ private static void CheckPropertyBindingCycles(BindingContext newBindingContext,
 
         private void ScrubWeakRefs()
         {
-            ArrayList? cleanupList = null;
-            foreach (DictionaryEntry de in _listManagers)
+            List<HashKey>? cleanupList = null;
+            foreach (KeyValuePair<HashKey, WeakReference> de in _listManagers)
```
dreddy-work, line 325: de ->keyValuePair or something that is not abbreviated.

dreddy-work: We are avoiding cut shorts/abbreviations in the sources/files we are changing across code base.

src/System.Windows.Forms/tests/UnitTests/misc/CollectionHelperTests.cs
Jericho: You can't reference dictionary items by numerical index.

dreddy-work: my bad. We could use element but i don't think we need to do that here. Thanks for correcting me.

src/System.Windows.Forms/tests/UnitTests/misc/CollectionHelperTests.cs
```diff
@@ -0,0 +1,143 @@
...
+                { "key-one", "value-one" },
+                { "key-two", "value-two" },
+            };
+            var target = new KeyValuePair<string, string>[2];
+
+            source.HashtableCopyTo(target, index: 0);
+
+            var firstTargetItem = target[0];
+            Assert.Equal(typeof(KeyValuePair<string, string>), firstTargetItem.GetType());
+            var firstSourceItem = source[0];
```
dreddy-work, line 89: You need to revert this as indexing dictionary is not straight.

JeremyKuhne (approved):

dreddy-work:
@Jericho , can you revisit this PR and finish final pending review comments?

dreddy-work:
Merging the pull request while there is still pending feedback that consists of minor nits and can be taken care of in later PRs that touching this file.

src/System.Windows.Forms/src/System/Windows/Forms/BindingContext.cs
Jericho: The name of the KVP was not changed. I preserved the name that was previously in use.

src/System.Windows.Forms/src/misc/CollectionHelper.cs
Jericho: I trust that the explanation I provided was satisfactory. Therefore I am marking it as resolved...

src/System.Windows.Forms/tests/UnitTests/misc/CollectionHelperTests.cs
Jericho: Not sure what you mean by "straight". This unit test completes successfully and demonstrates that `HashtableCopyTo` can copy the elements from the source dictionary to the target.

What is it exactly that you want me to revert?

src/System.Windows.Forms/src/System/Windows/Forms/BindingContext.cs
dreddy-work: We are moving away from shorts/abbreviations in Winforms code base and renaming local variable to a more meaningful.  Just to keep in mind for the next PRs.

Jericho: Is there a list of the changes you expect me to make in the existing code base if/when I submit PRs in the future?

dreddy-work: When you work on your next pull request, please try to rename any variables that are abbreviations or shortened versions. For example, you could change "de" to "keyValuePair" in this case. I don't expect you to create a pull request just for this change, but in general, if you notice any variables that could be clearer if they were renamed, feel free to do so as long as they are not public.

src/System.Windows.Forms/tests/UnitTests/misc/CollectionHelperTests.cs
dreddy-work: now, i am trying to recollect :)

src/System.Windows.Forms/src/System/Windows/Forms/BindingContext.cs
Jericho: What I was trying to understand is: what are the changes I am expected to make whenever I submit PRs, in addition to improving naming of local variables as you pointed out. It would be great to know so I can be proactive and make the changes you expect and avoid nit comments during code review.

My personal guidelines whenever submitting PRs in someone else's project is to be as respectful as possible of the existing code and modify only the code necessary to achieve the stated objective. That's why I try not to change variable names, code formatting, spacing, etc. But in this case it sounds like you expect me to make such changes, I'm trying to figure out what they are.

dreddy-work: Our coding guidelines can be found in [here](https://github.com/dotnet/winforms/blob/main/CONTRIBUTING.md). If you haven't had a chance to review them yet, please do so. We will try to update the guidelines to include any practices that we follow but may be missing there.

PR #8237, opened 2022-11-20
Refactor MergedAttributeCollection to replace HashTable

elachlan:
Refactor MergedAttributeCollection to replace HashTable and enable nullability.

Related: #8143

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8237)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/MergePropertyDescriptor.MergedAttributeCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/MergePropertyDescriptor.MergedAttributeCollection.cs
@@ -2,9 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
-using System.Collections;
 using System.ComponentModel;
 using System.Diagnostics.CodeAnalysis;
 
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/MergePropertyDescriptor.MergedAttributeCollection.cs
```diff
@@ -61,16 +55,20 @@ private Attribute GetCommonAttribute([DynamicallyAccessedMembers(DynamicallyAcce
...
                     {
                         value = GetDefaultAttribute(attributeType);
                         break;
                     }
                 }
 
-                _foundAttributes ??= new Hashtable();
-                _foundAttributes[attributeType] = value;
+                _foundAttributes ??= new();
+                if (value is Attribute)
```
dreddy-work, line 67: Do you expect non-Attribute types here?

elachlan: This is effectively a null check. I will change it.

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/MergePropertyDescriptor.MergedAttributeCollection.cs
```diff
@@ -61,16 +55,20 @@ private Attribute GetCommonAttribute([DynamicallyAccessedMembers(DynamicallyAcce
...
                     {
                         value = GetDefaultAttribute(attributeType);
                         break;
                     }
                 }
 
-                _foundAttributes ??= new Hashtable();
-                _foundAttributes[attributeType] = value;
+                _foundAttributes ??= new();
+                if (value is not null)
```
dreddy-work, line 67: you won't get here if it was null.

dreddy-work: you can remove this condition altogether

elachlan: I disagree. There is a possibility it will be null. `GetDefaultAttribute()` can return a null value.

dreddy-work: Okay. that make sense. However, if we are making sure `null` is not added to `_foundAttributes`, why do you need a check on line 44?

elachlan: Good Catch, I feel like this is to preserve the previous logic. Even though our dictionary's value isn't nullable. I am happy to remove it, its redundant now based on the design.

dreddy-work:
Thank you.

dreddy-work (approved):

PR #8251, opened 2022-11-22
Add Serialization Tests for AxHost

lonitra:
- Minor clean ups for `State` and `StateConverter`
- Add serialization test for `AxHost` using `BinaryFormatter`
- Add test for serializing `AxHost` using `StronglyTypedResourceBuilder`
    - `StateConverter` was never hooked up properly so manually register `StateConverter` as the `TypeConverter` for `State` using `CustomConverter` class
    - Fix `State` constructor utilized by `StateConverter` to not save the passed in stream since extra information to build the State is contained in it.
- Add test for `ResXResourceReader` verifying AxHosts can be deserialized from resx file

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8251)

### DIFF
--- a/src/Common/tests/TestUtilities/CustomConverter.cs
+++ b/src/Common/tests/TestUtilities/CustomConverter.cs
@@ -15,7 +15,7 @@ public static class CustomConverter
     public static RegistrationScope RegisterConverter(Type type, TypeConverter converter)
     {
         TypeDescriptionProvider parentProvider = TypeDescriptor.GetProvider(type);
-        TypeDescriptionProvider newProvider = new CustomTypeDescriptionProvider(parentProvider, converter);
+        CustomTypeDescriptionProvider newProvider = new(parentProvider, converter);
         TypeDescriptor.AddProvider(newProvider, type);
         return new RegistrationScope(type, newProvider);
     }
@@ -25,26 +25,18 @@ public static class CustomConverter
     ///  This is meant to be utilized in a <see langword="using"/> statement to ensure
     ///  <see cref="TypeDescriptor.RemoveProvider(TypeDescriptionProvider, Type)"/> is called when going out of scope with the using.
     /// </summary>
-    public ref struct RegistrationScope
+    public readonly ref struct RegistrationScope
     {
         private readonly Type _type;
         private readonly TypeDescriptionProvider _provider;
-        private TypeConverter _converter;
 
         public RegistrationScope(Type type, TypeDescriptionProvider provider)
         {
             _type = type;
             _provider = provider;
-            _converter = TypeDescriptor.GetConverter(type);
         }
 
-        public TypeConverter Converter => _converter;
-
-        public void Dispose()
-        {
-            TypeDescriptor.RemoveProvider(_provider, _type);
-            _converter = null;
-        }
+        public void Dispose() => TypeDescriptor.RemoveProvider(_provider, _type);
     }
 
     public class CustomTypeDescriptionProvider : TypeDescriptionProvider
--- a/src/Common/tests/TestUtilities/CustomConverter.cs
+++ b/src/Common/tests/TestUtilities/CustomConverter.cs
@@ -1,0 +1,67 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.ComponentModel;
+using System.Diagnostics.CodeAnalysis;
+
+namespace System;
+
+public static class CustomConverter
+{
+    /// <summary>
+    ///  Registers a TypeConverter with the specified type.
+    /// </summary>
+    public static RegistrationScope RegisterConverter(Type type, TypeConverter converter)
+    {
+        TypeDescriptionProvider parentProvider = TypeDescriptor.GetProvider(type);
+        TypeDescriptionProvider newProvider = new CustomTypeDescriptionProvider(parentProvider, converter);
+        TypeDescriptor.AddProvider(newProvider, type);
+        return new RegistrationScope(type, newProvider);
+    }
+
+    /// <summary>
+    ///  Manages the registration lifetime of a TypeConverter to a type.
+    ///  This is meant to be utilized in a <see langword="using"/> statement to ensure
+    ///  <see cref="TypeDescriptor.RemoveProvider(TypeDescriptionProvider, Type)"/> is called when going out of scope with the using.
+    /// </summary>
+    public readonly ref struct RegistrationScope
+    {
+        private readonly Type _type;
+        private readonly TypeDescriptionProvider _provider;
+
+        public RegistrationScope(Type type, TypeDescriptionProvider provider)
+        {
+            _type = type;
+            _provider = provider;
+        }
+
+        public void Dispose() => TypeDescriptor.RemoveProvider(_provider, _type);
+    }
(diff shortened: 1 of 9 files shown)


src/System.Windows.Forms.Design/tests/UnitTests/System/Resources/Tools/StronglyTypedResourceBuilderTests.cs
```diff
@@ -576,6 +579,81 @@ public static void StronglyTypedResourceBuilder_Create_AudioResource_FromFileRef
...
+            mediaPlayer.OcxState = state;
+            Assert.Equal(expectedUrl, mediaPlayer.URL);
+        }
+        finally
+        {
+            TypeDescriptor.RemoveProvider(newProvider, typeof(AxHost.State));
+        }
+    }
+
+    public class AxHostStateTypeDescriptionProvider : TypeDescriptionProvider
```
JeremyKuhne, line 639: It would probably be worth factoring this into a helper in the Common test utilities.

I'd consider doing it like this:

``` C#
public static class CustomConverter
{
    public RegistrationScope RegisterConverter(Type type, TypeConverter converter)
    // Get existing converter, create CustomTypeConverterTypeDescriptionProvider, and wrap in a RegistrationScope
    public ref struct RegistrationScope
    {
        public void Dispose() // Unregister the provider
    }
}

public class CustomTypeConverterTypeDescriptionProvider : TypeDescriptionProvider
{
   CustomTypeConverterTypeDescriptionProvider(TypeDecriptionProvider parent, TypeConverter converter)
   // ...
}
```

src/System.Windows.Forms.Design/tests/UnitTests/System/Resources/Tools/StronglyTypedResourceBuilderTests.cs
```diff
@@ -576,6 +579,81 @@ public static void StronglyTypedResourceBuilder_Create_AudioResource_FromFileRef
         ValidateResultAudio(GetPropertyInfo(reader.GetEnumerator(), compileUnit, "Audio1"));
     }
 
+    [WinFormsFact]
+    public static void StronglyTypedResourceBuilder_Create_OcxState_FromMemory()
+    {
+        TypeConverter converter = TypeDescriptor.GetConverter(typeof(AxHost.State));
+        Assert.Equal(typeof(TypeConverter), converter.GetType());
+
+        TypeDescriptionProvider parentProvider = TypeDescriptor.GetProvider(typeof(AxHost.State));
```
JeremyKuhne, line 588: A comment describing that we're testing serializing with the AxHostStateConverter, which isn't actually registered against the State would be good. Could also add to the test name.

src/Common/tests/TestUtilities/CustomConverter.cs
```diff
@@ -0,0 +1,71 @@
...
+
+public static class CustomConverter
+{
+    /// <summary>
+    ///  Registers a TypeConverter with the specified type.
+    /// </summary>
+    public static RegistrationScope RegisterConverter(Type type, TypeConverter converter)
+    {
+        TypeDescriptionProvider parentProvider = TypeDescriptor.GetProvider(type);
+        TypeDescriptionProvider newProvider = new CustomTypeDescriptionProvider(parentProvider, converter);
```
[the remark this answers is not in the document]
dreddy-work: nit: do you want to enable nullability in the new files being added?

src/System.Windows.Forms.Design/tests/UnitTests/System/Resources/Tools/StronglyTypedResourceBuilderTests.cs
```diff
@@ -573,19 +576,68 @@ public static void StronglyTypedResourceBuilder_Create_AudioResource_FromFileRef
...
+
+        ResXDataNode node = new("MediaPlayer1", scope.Converter.ConvertTo(mediaPlayer.OcxState, typeof(byte[])));
+        using var temp = TempFile.Create();
+        using (ResXResourceWriter resxWriter = new(temp.Path))
+        {
+            resxWriter.AddResource(node);
+            resxWriter.Generate();
+        }
+
+        var compileUnit = StronglyTypedResourceBuilder.Create(
```
dreddy-work, line 608: nit:  we are adapting explicit types if type is not evident from reading the code. I think some places in this test could use explicit types.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.StateConverter.cs
```diff
@@ -58,30 +35,24 @@ public override bool CanConvertTo(ITypeDescriptorContext? context, Type? destina
...
-                        using MemoryStream ms = new MemoryStream();
-                        State state = (State)value;
-                        state.Save(ms);
-                        ms.Close();
-                        return ms.ToArray();
+                        return Array.Empty<byte>();
                     }
-                    else
+
+                    if (value is State state)
```
dreddy-work, line 50: Public class and public type, this is changing behavior.

dreddy-work (commented):
RLGTM.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.StateConverter.cs
dreddy-work: I would leave that to you and @JeremyKuhne  to decide but we should consider a breaking change documentation for this if we continue.

src/Common/tests/TestUtilities/CustomConverter.cs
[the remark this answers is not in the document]
lonitra: I will go ahead and enable nullability for the TestUtility project in a follow-up

src/System.Windows.Forms.Design/tests/UnitTests/System/Resources/Tools/StronglyTypedResourceBuilderTests.cs
JeremyKuhne: @dreddy-work That is a consideration, not a rule. If it isn't easily understood what the return type is you should _consider_ being explicit. Sometimes being explicit adds more noise than value (inhibiting comprehension).

In this case, the variable name conveys all you need to know to understand the logic. Having the type is redundant.

It's everyone's best judgement for whatever makes the code the easiest to understand. The only hard rules we have are don't use `var` for intrinsic types (`bool`, `int`, `string`, etc.).

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.StateConverter.cs
JeremyKuhne: In this case this is really a bug.

lonitra: `base.ConvertTo` will ultimately throw a `NotSupportedException`. This is the better exception to throw since it is documented rather than a casting exception. I can go ahead and throw within this `if` statement instead of having to go to the base method.

JeremyKuhne: Throwing in the base is fine here.

JeremyKuhne (approved):
🚢 it

PR #8253, opened 2022-11-23
Add initial VB6 tests

JeremyKuhne:
Adds a really simple VB6 control to write tests against.

Adds utility code for handling COM classes straight from their host assembly.

Add a simple DynamicAxHost class to utilize these ComClass wrappers.

Also skip needless throwing and catching exceptions from the connection cookie setup in AxHost.

Note that I've renamed the control from .ocx to .vb6 to avoid having it accidentally picked up by anything.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8253)


lonitra (approved):
LGTM

JeremyKuhne:
Well, this is failing spectacularly. Trying to narrow it down.

JeremyKuhne:
@AaronRobinsonMSFT apparently it isn't my hack to create it directly. I've registered and used CoCreateInstance and I still get the corruption.

JeremyKuhne:
> Do we have a DMP? Can I reproduce this locally? I'll take a look later today and reach out offline.

It can be reproduced locally. With this branch run the following from the root:

`build -platform x86 -test /p:TargetArchitecture=x86`

The test run will abort with no details. In VS you can get it to repro by setting the Processor Architecture for Any CPU Projects to x86 in the Test menu and selecting all of the tests in System.Windows.Forms.Tests except for the System.Windows.Forms.Tests and running them. I've never been able to capture it as it AVs or OOMs (with plenty of memory left).

image

I will check in the code eventually. Was trying to get a better handle on the project structure to know what I should include. It is an ActiveX Control project with a text box and a button that sets the text in the text box when clicked. Just dragged and dropped and set a click handler through the UI.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AxHost.VisualBasic6Tests.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Runtime.InteropServices;
+using System.Windows.Forms.Tests.TestResources;
+using Xunit;
+
+namespace System.Windows.Forms.Tests
+{
+    public class AxHostVisualBasic6Tests
+    {
+        [WinFormsFact]
+        public void AxHost_SimpleControl_Create()
```
RussKie, line 15: Any specific reason to run this as a unit test and not as UI integration test? The latter are run serially.

JeremyKuhne:
This fails in the same way on .NET 7. :/

AaronRobinsonMSFT:
@JeremyKuhne Is it worth checking how this behaves on .NET Framework in the current test harness? I'd like to know how much effort I should put into looking into this as it is very hard to untangle these sorts of bugs when they only occur in while running in a process with many threads, the VB6 runtime just wasn't designed for that environment.

JeremyKuhne:
> Is it worth checking how this behaves on .NET Framework in the current test harness?

@AaronRobinsonMSFT I'll keep trying to narrow down a repro for it. These tests are all on the same thread- which is what worries me a bit. It's quite possible it is some side effect of us never quite doing the COM right (thinking of how we don't handle casting to base interfaces for example). I'll keep you posted.

JeremyKuhne:
This was merged disabled in #8322.

PR #8255, opened 2022-11-23
Refactor `KeysConverter` to remove `ArrayList`

elachlan:
Related #8140

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8255)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
@@ -276,7 +276,6 @@ namespace System.Windows.Forms
 
             string GetTermsString(Keys key)
             {
-                bool added = false;
                 StringBuilder termStrings = new(32);
                 Keys modifiers = (key & Keys.Modifiers);
 
@@ -288,13 +287,7 @@ namespace System.Windows.Forms
                     Keys keyValue = _keyNames[keyString];
                     if (keyValue.HasFlag(modifiers))
                     {
-                        if (added)
-                        {
-                            termStrings.Append('+');
-                        }
-
-                        termStrings.Append(keyString);
-                        added = true;
+                        termStrings.Append(keyString).Append('+');
                     }
                 }
 
@@ -302,11 +295,6 @@ namespace System.Windows.Forms
                 // we find a match.
                 Keys keyOnly = key & Keys.KeyCode;
                 bool foundKey = false;
-
-                if (added)
-                {
-                    termStrings.Append('+');
-                }
 
                 for (int i = 0; i < DisplayOrder.Count; i++)
                 {
--- a/src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
@@ -119,12 +119,7 @@ namespace System.Windows.Forms
         /// </summary>
         public override bool CanConvertFrom(ITypeDescriptorContext? context, Type sourceType)
         {
-            if (sourceType == typeof(string) || sourceType == typeof(Enum[]))
-            {
-                return true;
-            }
-
-            return base.CanConvertFrom(context, sourceType);
+            return sourceType == typeof(string) || sourceType == typeof(Enum[]) ? true : base.CanConvertFrom(context, sourceType);
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


dreddy-work (dismissed):

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
```diff
@@ -243,7 +243,8 @@ public int Compare(object? a, object? b)
                 {
                     Keys key = (Keys)value;
                     bool added = false;
-                    ArrayList terms = new ArrayList();
+                    List<string> termStrings = asString ? new() : null;
```
halgab, line 246: You could turn `termStrings` into a StringBuilder directly

elachlan: I think it might be faster to do a string.join instead. I'll have to do a test.

dreddy-work:
@elachlan , multiple build failures here.
src\System.Windows.Forms\src\System\Windows\Forms\KeysConverter.cs(246,48): error CS8600: (NETCORE_ENGINEERING_TELEMETRY=Build) Converting null literal or possible null value to non-nullable type.

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
```diff
@@ -243,7 +243,8 @@ public int Compare(object? a, object? b)
                 {
                     Keys key = (Keys)value;
                     bool added = false;
-                    ArrayList terms = new ArrayList();
```
lonitra, line 246: Build is complaining about possible null references. We can potentially do the following instead of avoid having to null check everywhere and avoid having two variables that essentially go through the same code:
``` c#
bool asString = destinationType == typeof(string);
bool asEnum = !asString && destinationType == typeof(Enum[]);
IList? terms = asString ? new List<string>() :
                asEnum ? new List<Enum>() : null;
if (terms is null)
{
    return base.ConvertTo(context, culture, value, destinationType);
}

// continue with rest of code..
```

though I am currently not sure if there are performance drawbacks with `IList`..

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
```diff
@@ -230,7 +230,7 @@ public int Compare(object? a, object? b)
         {
             ArgumentNullException.ThrowIfNull(destinationType);
 
-            if (value is Keys || value is int)
+            if (value is Keys or int)
```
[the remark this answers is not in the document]
elachlan:
```csharp
if (value is not Keys or not int)
{
    return base.ConvertTo(context, culture, value, destinationType);
}
```

I get the following warning:
```
Warning	CS8794	An expression of type 'object' always matches the provided pattern.
```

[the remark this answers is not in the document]
lonitra: If we invert we should get 
```c#
if (value is not Keys and not int)
{
    return base.ConvertTo(context, culture, value, destinationType);
}
```

Do you still get a warning with this?

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
elachlan: I tried a few different things here, the current solution seems the most simple whilst keeping with generics. 

The option I didn't try was refactoring out the string vs enum solution into their own functions. I think that might be the best solution, since it would remove most of the branching.

elachlan: I've pushed through a refactor using local functions, which simplifies the logic. but duplicates the code a bit.

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
```diff
@@ -230,112 +220,124 @@ public int Compare(object? a, object? b)
...
                     }
-                    else
+                }
+
+                // Now reset and do the key values. Here, we quit if
+                // we find a match.
+                Keys keyOnly = key & Keys.KeyCode;
+                bool foundKey = false;
+
+                if (added && asString)
```
lonitra, line 312: Now that this is refactored into its own method I don't think checking `asString` is necessary anymore.
```suggestion
                if (added)
```

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
```diff
@@ -230,112 +220,124 @@ public int Compare(object? a, object? b)
...
+                Keys modifiers = (key & Keys.Modifiers);
 
-                    if (asString)
+                // First, iterate through and do the modifiers. These are
+                // additive, so we support things like Ctrl + Alt
+                for (int i = 0; i < DisplayOrder.Count; i++)
+                {
+                    string keyString = DisplayOrder[i];
+                    Keys keyValue = _keyNames[keyString];
+                    if (((int)keyValue & (int)modifiers) != 0)
```
lonitra, line 295:
```suggestion
                    if (keyValue.HasFlag(modifiers))
```

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
```diff
@@ -230,112 +220,124 @@ public int Compare(object? a, object? b)
...
+                List<Enum> termKeys = new();
+                Keys modifiers = (key & Keys.Modifiers);
+
+                // First, iterate through and do the modifiers. These are
+                // additive, so we support things like Ctrl + Alt
+                for (int i = 0; i < DisplayOrder.Count; i++)
+                {
+                    string keyString = DisplayOrder[i];
+                    Keys keyValue = _keyNames[keyString];
+                    if (((int)keyValue & (int)modifiers) != 0)
```
lonitra, line 249:
```suggestion
                    if (keyValue.HasFlag(modifiers))
```

lonitra (changes requested):
Just a few comments, otherwise this is looking much cleaner! :)

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
```diff
@@ -230,112 +220,124 @@ public int Compare(object? a, object? b)
...
+                for (int i = 0; i < DisplayOrder.Count; i++)
+                {
+                    string keyString = DisplayOrder[i];
+                    Keys keyValue = _keyNames[keyString];
+                    if (keyValue.Equals(keyOnly))
                     {
-                        return (Enum[])terms.ToArray(typeof(Enum));
+                        termStrings.Append(keyString);
+
+                        added = true;
```
dreddy-work, line 325: what is this assignment for?

elachlan: I have removed it. It looked redundant.

dreddy-work (approved):

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
```diff
@@ -230,112 +214,122 @@ public int Compare(object? a, object? b)
...
+                    string keyString = DisplayOrder[i];
+                    Keys keyValue = _keyNames[keyString];
+                    if (keyValue.HasFlag(modifiers))
                     {
-                        StringBuilder b = new StringBuilder(32);
-                        foreach (string t in terms)
+                        if (added)
                         {
-                            b.Append(t);
+                            termStrings.Append('+');
```
halgab, line 293: Isn't this equivalent to directly doing `termStrings.Append(keyString).Append('+')` on line 296 and removing the `added` logic entirely?

elachlan: Good spot. It took me a bit, but it makes sense.

PR #8256, opened 2022-11-23
Refactor `UndoEngine` to remove `ArrayList`/`HashTable`

elachlan:
Related #8140

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8256)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.cs
@@ -1012,13 +1012,10 @@ namespace System.ComponentModel.Design
                                 List<IComponent> list = new(names.Length);
                                 foreach (string name in names)
                                 {
-                                    if (name is not null)
+                                    IComponent comp = _lastSelection[name].Components[name];
+                                    if (comp is not null)
                                     {
-                                        IComponent comp = _lastSelection[name].Components[name];
-                                        if (comp is not null)
-                                        {
-                                            list.Add(comp);
-                                        }
+                                        list.Add(comp);
                                     }
                                 }
 
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.cs
@@ -577,7 +577,7 @@ namespace System.ComponentModel.Design
             private List<IComponent> _ignoreAddingList; // the list of objects that are currently being added.  We ignore change events between adding and added.
             private List<IComponent> _ignoreAddedList; // the list of objects that are added. We do not serialize before state for change events that happen in the same transaction
             private bool _reverse; // if true, we walk the events list from the bottom up
-            private readonly Hashtable _lastSelection; // the selection as it was before we gathered undo info
+            private readonly Dictionary<string, IContainer> _lastSelection; // the selection as it was before we gathered undo info
 
             public UndoUnit(UndoEngine engine, string name)
             {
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.cs
```diff
@@ -1008,14 +1008,13 @@ protected virtual void UndoCore()
...
                             if (UndoEngine.GetService(typeof(ISelectionService)) is ISelectionService ss)
                             {
-                                string[] names = new string[_lastSelection.Keys.Count];
-                                _lastSelection.Keys.CopyTo(names, 0);
-                                ArrayList list = new ArrayList(names.Length);
+                                string[] names = _lastSelection.Keys.ToArray();
+                                List<IComponent> list = new(names.Length);
                                 foreach (string name in names)
                                 {
                                     if (name is not null)
```
halgab, line 1015: `name` is a dictionary key, and those can't be null right?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.cs
```diff
@@ -1008,14 +1008,13 @@ protected virtual void UndoCore()
                         {
                             if (UndoEngine.GetService(typeof(ISelectionService)) is ISelectionService ss)
                             {
-                                string[] names = new string[_lastSelection.Keys.Count];
-                                _lastSelection.Keys.CopyTo(names, 0);
-                                ArrayList list = new ArrayList(names.Length);
+                                string[] names = _lastSelection.Keys.ToArray();
```
halgab, line 1011: Couldn't you directly loop through the dictionary KeyValuePairs?

elachlan: In a similar issue I was told that a copy is usually made in case changes to the collection happen.

dreddy-work: in this case, if `_lastSelection `changes during loop, we could get `KeyNotFound` exception.

elachlan: Done.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.cs
```diff
@@ -1008,18 +1008,13 @@ protected virtual void UndoCore()
                         {
                             if (UndoEngine.GetService(typeof(ISelectionService)) is ISelectionService ss)
                             {
-                                string[] names = new string[_lastSelection.Keys.Count];
-                                _lastSelection.Keys.CopyTo(names, 0);
-                                ArrayList list = new ArrayList(names.Length);
-                                foreach (string name in names)
+                                List<IComponent> list = new(_lastSelection.Keys.Count);
+                                foreach (string name in _lastSelection.Keys)
```
halgab, line 1012: Reading it once again, you could actually loop only through the values of `_lastSelection`. The keys are really just used to get the values on the line below

elachlan: I explored that briefly. It doesn't work because we index into `Components` using `name` as well.

halgab: That's true, I missed that, sorry

elachlan: No problem. I wish we could simplify it more.

dreddy-work (dismissed):

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.cs
halgab: The only tiny thing I can see is not looping through the Keys collection but directly the dictionary to save one lookup, but that's it

elachlan: Done.

dreddy-work (approved):

PR #5791, opened 2021-09-17
Move from IntPtr to nint to avoid runtime casting errors.

JeremyKuhne:
We still had a number of potential and actual overflow scenarios. In order to make this more resilient, move SendMessage, PostMessage, and Get/SetWindowLong from IntPtr to nint.

This also makes things slightly faster as we aren't creating IntPtr structs and invoking the cast methods on it.

Note that IntPtr implicitly converts to nint. I've removed most of the actual IntPtr usages. I've also removed most of the HandleRef overloads and vetted the callers to do the right thing in regards to avoiding finalizer issues.

There is more we can do here, but this is a big first step.

Also clean up DateTimePicker and move SYSTEMTIME converters to SYSTEMTIME.

I did some style cleanup, but not exhaustively. I'll fix anything I changed if you catch anything I messed up.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/5791)

PR #8283, opened 2022-11-25
Fix setting Form.Owner in Form.Show/ShowDialog(IWin32Window owner) if owner is a Control

### LINKED ISSUE #8280: Major regression in Form.ShowDialog(IWin32Window) in .NET 7 -- ship blocker for Paint.NET 5.0
### .NET version

7.0.0

Yes

Yes, it worked fine in .NET 5 and .NET 6

### Issue description

When `Form.ShowDialog(IWin32Window)` is called with a `Control` instead of a `Form`, the dialog's parenting/ownership is not set correctly, causing the dialog to not "track" with the parent form correctly.

The most visible symptom of this is that if you switch between virtual desktops, the parent form is correctly hidden but the dialog is not. And, it is possible to interact with the main window in certain ways. In my app (Paint.NET), it is possible to switch between images by using the taskbar (kind of like how IE/Edge used to let you switch between tabs). This is normally prevented when a modal dialog is active, but right now it's not. This can result in actions affecting the wrong image, which then corrupts the state of the image's change history and causes data loss.

Another symptom is that the `Owner` for the dialog is `null`, even though it should probably be `control.FindForm()`

**This is currently a ship-blocker for Paint.NET 5.0 on .NET 7**. I cannot release the new version with this bug.

Known workaround: use `dialog.ShowDialog(control.FindForm()` instead of `dialog.ShowDialog(control)`. However, this requires changes to _many_ places in the code, and is not a fix I can apply to libraries or plugins.

I believe this regressed due to https://github.com/dotnet/winforms/pull/3394 cc @dreddy-work which changed whether `Owner` was set _or_ `SetWindowLong(..., GWL.HWNDPARENT, ...)` was called. Previously, `SetWindowLong` was always called.

### Steps to reproduce

Create a Form and do `Application.Run(form)`. Make sure the Form has at least 1 child control, like a `button`.
In some code within the form, like in the `button`'s click handler, create another Form and call `dialog.ShowDialog(button)`
While the dialog is open, switch between virtual desktops using WinKey+Shift+Left/Right

dmitrii-drobotov:
This restores a behavior until #5791, where this condition was inverted as a part of refactoring.

Fix #8280 on .NET 8

## Proposed changes

- Check if owner doesn't have a `TOPMOST` style to correctly initialize it with a top-level Control in this case

## Customer Impact

- Owner property is correctly set for a form if it was shown with a Control as owner

## Regression? 

- Yes 

## Risk

- Minimal - restores previous behavior

## Test methodology 

- Unit test
- Manual test with sample app from https://github.com/dotnet/winforms/issues/8280#issuecomment-1326776047
- CTI
 

## Test environment(s) 

- .NET 8.0.100-alpha.1.22565.7

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8283)

### DIFF
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
@@ -2572,6 +2572,54 @@ namespace System.Windows.Forms.Tests
             Assert.True(control.IsHandleCreated);
         }
 
+        public static IEnumerable<object[]> FormShow_TestData()
+        {
+            // Test with control passed as owner.
+            yield return new Action<Form, Control>[]
+            {
+                (Form form, Control control) => form.Show(owner: control)
+            };
+            yield return new Action<Form, Control>[]
+            {
+                (Form form, Control control) => form.ShowDialog(owner: control)
+            };
+
+            // Test with parent form passed as owner.
+            yield return new Action<Form, Control>[]
+            {
+                (Form form, Control control) => form.Show(owner: control.FindForm())
+            };
+            yield return new Action<Form, Control>[]
+            {
+                (Form form, Control control) => form.ShowDialog(owner: control.FindForm())
+            };
+        }
+
+        [WinFormsTheory]
+        [MemberData(nameof(FormShow_TestData))]
+        public void Form_Show_SetsOwnerToTopLevelForm_WhenShownWithOwner(Action<Form, Control> showAction)
+        {
+            // Regression test for https://github.com/dotnet/winforms/issues/8280
+
+            using Form parent = new();
+            using Control control = new();
+            parent.Controls.Add(control);
+            parent.Show();
+
+            using Form child = new();
+            Form owner = null;
+            child.Load += (object sender, EventArgs e) =>
+            {
+                owner = ((Form)sender).Owner;
+                child.Close();
+            };
+
+            showAction(child, control);
+
+            Assert.Same(parent, owner);
+            Assert.False(child.IsHandleCreated);
+        }
+
         public class SubForm : Form
         {
             public new const int ScrollStateAutoScrolling = Form.ScrollStateAutoScrolling;
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
@@ -5181,7 +5181,7 @@ namespace System.Windows.Forms
                 throw new InvalidOperationException(SR.CantShowModalOnNonInteractive);
             }
 
-            if ((owner is not null) && owner.GetExtendedStyle().HasFlag(WINDOW_EX_STYLE.WS_EX_TOPMOST))
+            if ((owner is not null) && !owner.GetExtendedStyle().HasFlag(WINDOW_EX_STYLE.WS_EX_TOPMOST))
             {
                 // It's not the top-most window
                 if (owner is Control ownerControl)
@@ -5254,7 +5254,7 @@ namespace System.Windows.Forms
                 throw new InvalidOperationException(SR.CantShowModalOnNonInteractive);
             }
 
-            if ((owner is not null) && owner.GetExtendedStyle().HasFlag(WINDOW_EX_STYLE.WS_EX_TOPMOST))
+            if ((owner is not null) && !owner.GetExtendedStyle().HasFlag(WINDOW_EX_STYLE.WS_EX_TOPMOST))
             {
                 // It's not the top-most window
                 if (owner is Control ownerControl)


merriemcgaw:
@dmitrii-drobotov can you add some tests around this?

@JeremyKuhne, @lonitra, @dreddy-work FYI, we need this approved ASAP and a servicing template filled out for the cherry-pick to the servicing branch.

gpetrou:
Perhaps something like the following should be enough?
```csharp
using Form mainForm = new();
using Button button = new();
mainForm.Controls.Add(button);
button.Click += (object sender, EventArgs e) =>
{
    using (Form form = new Form())
    {
        form.Load += (object sender, EventArgs e) =>
        {
            Assert.NotNull(((Form)sender).Owner);
        };
        form.ShowDialog(button);
    }
};

mainForm.Show();
button.PerformClick();
```

rickbrew:
I built my own version of System.Windows.Forms.dll with the fix, and I can thankfully confirm that it does fix the issues I was seeing in Paint.NET

Looks like I'll have to ship my own S.W.F.dll for the time being, but this is preferable to rolling back to .NET 6. phew!

Tanya-Solyanik:
@gpetrou  - thank you for the test. However, assert in the Load event handler is not handled by Xunit because it is [caught ](https://github.com/dotnet/winforms/blob/459b2713917a9337c1e81b693629a7495c81db3d/src/System.Windows.Forms/src/System/Windows/Forms/NativeWindow.cs#L382)by your secondary form's NativeWindow wndProc and treated as a thread exception with a popup dialog. The result should be saved in a local variable and verified after the form is closed.

dreddy-work (dismissed):
LGTM.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
```diff
@@ -2572,6 +2572,49 @@ public void Form_OnHandleDestroyed_InvokeWithHandle_CallsHandleDestroyed(EventAr
             Assert.True(control.IsHandleCreated);
         }
 
+        [WinFormsFact]
+        public void Form_Show_SetsOwnerToTopLevelForm_WhenShownWithControlAsOwner()
```
Tanya-Solyanik, line 2576: I would prefer a theory test in this case because the code is identical and it's hard to read it to understand what the difference is.

```
        public static IEnumerable<object[]> FormShow_TestData()
        {
            yield return new Action<Form, Control>[]
            {
                (Form form, Control control) =>
                {
                    form.Show(owner : control);
                }
            };
            yield return new Action<Form, Control>[]
            {
                (Form form, Control control) =>
                {
                    form.ShowDialog(owner : control);
                }
            };
        }
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
```diff
@@ -2572,6 +2572,49 @@ public void Form_OnHandleDestroyed_InvokeWithHandle_CallsHandleDestroyed(EventAr
...
+        {
+            using Form parent = new();
+            using Control control = new();
+            parent.Controls.Add(control);
+            parent.Show();
+            using Form form = new();
```
Tanya-Solyanik, line 2582: Please add an empty line after `Show`

Tanya-Solyanik, line 2582: Consider renaming the second form to `child`.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
```diff
@@ -2572,6 +2572,49 @@ public void Form_OnHandleDestroyed_InvokeWithHandle_CallsHandleDestroyed(EventAr
...
+
+            using Form form = new();
+            Form owner = null;
+            form.Load += (object sender, EventArgs e) =>
+            {
+                owner = ((Form)sender).Owner;
+                form.Close();
+            };
+
+            form.ShowDialog(owner: control);
```
Tanya-Solyanik, line 2612: Could you please also add a test where `parent` forms is the owner so that the scenario is better covered.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
```diff
@@ -2572,6 +2572,49 @@ public void Form_OnHandleDestroyed_InvokeWithHandle_CallsHandleDestroyed(EventAr
...
+            };
+
+            form.Show(owner: control);
+
+            Assert.Same(owner, parent);
+            Assert.False(form.IsHandleCreated);
+        }
+
+        [WinFormsFact]
+        public void Form_ShowDialog_SetsOwnerToTopLevelForm_WhenShownWithControlAsOwner()
```
Tanya-Solyanik, line 2597: Please reference the bug you are regressing here.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
```diff
@@ -2572,6 +2572,64 @@ public void Form_OnHandleDestroyed_InvokeWithHandle_CallsHandleDestroyed(EventAr
...
+        public static IEnumerable<object[]> FormShow_TestData()
+        {
+            yield return new Action<Form, Control>[]
+            {
+                (Form form, Control control) => form.Show(owner: control)
+            };
+            yield return new Action<Form, Control>[]
+            {
+                (Form form, Control control) => form.ShowDialog(owner: control)
+            };
```
Tanya-Solyanik, line 2584: Is it possible to craft such theory data that could be used with a single test? Perhaps 
`(Form form, Control control) => form.ShowDialog(owner: control.FindForm())`
or change the action to take 3 parameters, (parentForm, childForm, control)?

dmitrii-drobotov: Good idea, I added actions with `control.FindForm()`

Tanya-Solyanik (approved):

dreddy-work (approved):

PR #8309, opened 2022-12-01
Convert `StubSite` to use Dictionary instead of HashTable and Enable Nullability

elachlan:
Related: #8143, #6793

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8309)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.StubSite.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.StubSite.cs
@@ -10,17 +10,17 @@ namespace System.Windows.Forms
 {
     public partial class ToolStripControlHost
     {
-        /// <remarks>
-        ///  <para>Our implementation of ISite:
+        /// <devdoc>
+        ///  Our implementation of ISite:
         ///  Since the Control which is wrapped by ToolStripControlHost is a runtime instance, there is no way of knowing
         ///  whether the control is in runtime or designtime.
         ///  This implementation of ISite would be set to Control.Site when ToolStripControlHost.Site is set at DesignTime. (Refer to Site property on ToolStripControlHost)
         ///  This implementation just returns the DesignMode property to be ToolStripControlHost's DesignMode property.
-        ///  Everything else is pretty much default implementation.</para>
-        /// </remarks>
+        ///  Everything else is pretty much default implementation.
+        /// </devdoc>
         private class StubSite : ISite, IDictionaryService
         {
-            private Dictionary<string, object>? _dictionary;
+            private Dictionary<object, object>? _dictionary;
             private readonly IComponent _comp;
             private readonly IComponent _owner;
 
@@ -73,7 +73,7 @@ namespace System.Windows.Forms
                 // We have to implement our own dictionary service. If we don't,
                 // the properties of the underlying component will end up being
                 // overwritten by our own properties when GetProperties is called
-                return service == typeof(IDictionaryService) ? this : (_owner.Site?.GetService(service));
+                return service == typeof(IDictionaryService) ? this : _owner.Site?.GetService(service);
             }
 
             /// <summary>
@@ -81,15 +81,17 @@ namespace System.Windows.Forms
             /// </summary>
             object? IDictionaryService.GetKey(object? value)
             {
-                if (_dictionary is not null)
+                if (value is null || _dictionary is null)
                 {
-                    foreach (var item in _dictionary)
+                    return null;
+                }
+
+                foreach (var item in _dictionary)
+                {
+                    object o = item.Value;
+                    if (value.Equals(o))
                     {
-                        object o = item.Value;
-                        if (value is not null && value.Equals(o))
-                        {
-                            return item.Key;
-                        }
+                        return item.Key;
                     }
                 }
 
@@ -101,10 +103,9 @@ namespace System.Windows.Forms
             /// </summary>
             object? IDictionaryService.GetValue(object key)
             {
-                return key is not string keyString
-                    ? null
-                    : _dictionary is not null
-                      && _dictionary.TryGetValue(keyString, out object? value) ? value : null;
+                object? value = null;
+                _dictionary?.TryGetValue(key, out value);
+                return value;
             }
 
             /// <summary>
@@ -115,19 +116,15 @@ namespace System.Windows.Forms
             void IDictionaryService.SetValue(object key, object? value)
             {
                 _dictionary ??= new();
-
-                if (key is not string keyString)
-                {
-                    return;
-                }
+                ArgumentNullException.ThrowIfNull(key);
 
                 if (value is not object)
                 {
-                    _dictionary.Remove(keyString);
+                    _dictionary.Remove(key);
                 }
                 else
                 {
-                    _dictionary[keyString] = value;
+                    _dictionary[key] = value;
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


elachlan:
@gpetrou, your changes in #6793 didn't convert the HashTable. So I have done so here and enabled nullability.

gpetrou:
> your changes in #6793 didn't convert the HashTable. So I have done so here and enabled nullability.

OK. But please keep in mind that I am planning to update my pending PRs. If you want to work on nullability, perhaps work on DataGridView-related classes which I do not intent do spend time on.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.StubSite.cs
```diff
@@ -41,56 +38,57 @@ public StubSite(IComponent control, IComponent host)
...
+            string? ISite.Name
             {
-                get => _owner.Site.Name;
+                get => _owner.Site?.Name;
 
                 [RequiresUnreferencedCode(TrimmingConstants.SiteNameMessage)]
-                set => _owner.Site.Name = value;
+                set
+                {
+                    if (_owner.Site is not null)
```
dreddy-work, line 59: this change is causing test failures.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.StubSite.cs
```diff
@@ -101,32 +99,35 @@ object IDictionaryService.GetKey(object value)
...
-            object IDictionaryService.GetValue(object key)
+            object? IDictionaryService.GetValue(object key)
             {
-                if (_dictionary is not null)
-                {
-                    return _dictionary[key];
-                }
-
-                return null;
+                return key is not string keyString
```
dreddy-work, line 104: I am trying to find a scenario if key can be non-string.  Can you add comment on why you do not expect non-strings here?

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.StubSite.cs
```diff
@@ -14,18 +11,18 @@ namespace System.Windows.Forms
     public partial class ToolStripControlHost
     {
         /// <remarks>
-        ///  Our implementation of ISite:
+        ///  <para>Our implementation of ISite:
```
JeremyKuhne, line 14: This isn't [public documentation](https://learn.microsoft.com/dotnet/api/system.windows.forms.toolstripcontrolhost?view=windowsdesktop-7.0) and as such shouldn't get a tag that will get picked up. Ultimately we want to have the xml comments synced with the docs. `<DevDoc>` is one typical alternative.

elachlan: is this instead of `<remarks>` or `<para>`?

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.StubSite.cs
JeremyKuhne: It takes `object` on this public interface so we have to allow for it. We can add an assert if we expect it never to be (but don't add one unless you're confident that it shouldn't fire).

nit: In this case I believe it would better to separate and early out for clarity:

``` C#
if (key is not string keyString)
{
    return null;
}

// ....
```

Also consider being safe with the null check by either:

``` C#
return _dictionary is { } dictionary
    ? dictionary.TryGetValue(stringKey, out object? value) ? value : null
    : null;

// or (better imho)

object? value = null;
_dictionary?.TryGetValue(stringKey, out value);
return value;
```

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.StubSite.cs
```diff
@@ -41,56 +38,57 @@ public StubSite(IComponent control, IComponent host)
...
+            object? IDictionaryService.GetKey(object? value)
             {
                 if (_dictionary is not null)
                 {
-                    foreach (DictionaryEntry de in _dictionary)
+                    foreach (var item in _dictionary)
                     {
-                        object o = de.Value;
+                        object o = item.Value;
                         if (value is not null && value.Equals(o))
```
halgab, line 89: you could check `value` for null before the loop

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.StubSite.cs
elachlan: Also do we want to add a useful documentation comment here?

JeremyKuhne: Instead of `<remarks>`. And I should have said `<devdoc>`. :)

You're welcome to copy stuff from the docs site if you want. It's a fair amount of work, however, and we'll probably do some sort of import in the future.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.StubSite.cs
elachlan: You are correct, it might be a safer choice to make it a `Dictionary<object, object>`?

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.StubSite.cs
JeremyKuhne: And I don't think the info here is useful for external docs.

elachlan: We can do docs in another PR.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.StubSite.cs
elachlan: I've converted it to `Dictionary<object, object>`. I am unsure if this is correct, but its closer to the original behaviour.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripControlHostTests.cs
```diff
@@ -1812,69 +1812,6 @@ public void ToolStripControlHost_Site_Set_GetReturnsExpected(IContainer containe
...
-                .Setup(s => s.Container)
-                .Returns(container);
-            mockSite
-                .Setup(s => s.GetService(typeof(AmbientProperties)))
-                .Returns(null);
-            item.Site = mockSite.Object;
-            ISite oldSite = c.Site;
-
-            item.Site = null;
-            Assert.Throws<NullReferenceException>(() => oldSite.Container);
```
gpetrou, line 1833: Perhaps update the asserts and rename the test methods instead?

elachlan: Ahh, so a test to make sure it doesn't throw?

elachlan: Done.

lonitra (approved):
LGTM :)

dreddy-work (approved):

Issue #8291, opened 2022-11-28
ComWrapper needs to return vtable based on types, not interfaces


JeremyKuhne:
``` C#
protected override unsafe ComInterfaceEntry* ComputeVtables(object obj, CreateComInterfaceFlags flags, out int count)
```

The implementation here needs to go off of specific class types, not interfaces. Not all classes that implement an interface will only implement a single interface (many won't).

The best place to start is by hard tying specific class types to entries. We can iterate with additional flexibility as warranted.

I plan to tackle this imminently.

elachlan:
@kant2002 things are happening!

kant2002:
Honestly I did not get the idea. Any example from existing COM patterns?

JeremyKuhne:
This has been completely rewritten.

PR #8311, opened 2022-12-01
Introduce IManagedWrapper/IVTable For WinFormsComWrapper Logic

lonitra:
Related: #8291 

Currently, `WinFormsComWrappers.ComputeVtables()` goes through and returns the first VTable for the COM interface the object implements but this only captures a single COM interface the object implements when many implement multiple. As a result, calling `QueryInterface()` on the CCW will not have all the interfaces the object actually implements. This PR introduces a new pattern as a starting point to be able to account for objects that implement multiple interfaces. 

- New APIs:
    - `IVTable` - Replaces `IPopulateVTable` and is used to get the populated vtable.
    - `IVTable<TComInterface, TVTable>` - Allocates space for the vtable and populates the `IUnknown` methods. CCWs implement this and populate the rest of the vtable with its own specific methods.
    - `ComInterfaceTable` - Struct that ties together the number of COM interfaces an object implements and its `ComInterfaceEntry*`.
    - `IManagedWrapper` - This is used to get a populated `ComInterfaceTable` for `WinFormsComWrappers` to use.
    - `IManagedWrapper<TComInterface, ...>` - To be implemented by managed objects. No actual implementation needs to be done as the interface does all the work of setting up the `ComInterfaceTable`.
- Additionally I have also moved `InitializeXEntry()` logic for our COM Interfaces that have not yet been converted to the new pattern to their respective implementing classes and implemented the non generic `IManagedWrapper`. As a follow up I will convert these interfaces over to the new pattern.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8311)


src/System.Windows.Forms.Primitives/src/Interop/Ole32/Interop.GPStream.cs
```diff
@@ -5,13 +5,13 @@
...
 using System.Runtime.InteropServices;
-using Com = Windows.Win32.System.Com;
+using Windows.Win32.System.Com;
 
 internal static partial class Interop
 {
     internal static partial class Ole32
     {
-        public sealed unsafe class GPStream : Com.IStream.Interface
+        public sealed unsafe class GPStream : IStream.Interface, IManagedWrapper<IStream>
```
JeremyKuhne, line 14: This is one of the broken ones- it should technically be `IManagedWrapper<IStream, ISequentialStream>`.

src/System.Windows.Forms.Primitives/tests/UnitTests/Windows/Win32/GlobalInterfaceTableTests.cs
```diff
@@ -35,7 +35,7 @@ public void GlobalInterfaceTable_CheckRefCounts()
             stream.Value->Release();
         }
 
-        internal class MyStream : IStream.Interface
+        internal class MyStream : IStream.Interface, IManagedWrapper<IStream>
```
JeremyKuhne, line 38: Again, should be `<IStream, ISequentialStream>`

JeremyKuhne (changes requested):
Just need to update the IStream usages to also explicitly call out the base class. Looks good otherwise!

JeremyKuhne (approved):
🚀

PR #8313, opened 2022-12-02
Refactoring to replace usages of HashTable

elachlan:
Related: #8143


src/System.Windows.Forms.Design/src/System/ComponentModel/Design/InheritanceService.cs
```diff
@@ -185,12 +184,19 @@ protected virtual void AddInheritedComponents(Type type, IComponent component, I
...
-                        bool notPresent = (_inheritedComponents[value] is null);
-                        _inheritedComponents[value] = attr;
+                        // We only get values via IComponent Keys and don't expose any other behaviors.
+                        // So only use IComponent as a key.
+                        bool notPresent = false;
+                        if (value is IComponent compValue)
+                        {
+                            notPresent = !_inheritedComponents.TryGetValue(compValue, out InheritanceAttribute oldAttr)
+                                || oldAttr is null;
+                            _inheritedComponents[compValue] = attr;
```
lonitra, line 194: Is there a reason why we would want `attr` to be potentially null for a `IComponent` key?

elachlan: Good point.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/InheritanceService.cs
```diff
@@ -255,18 +260,16 @@ protected virtual bool IgnoreInheritedMember(MemberInfo member, IComponent compo
...
         /// </summary>
         public InheritanceAttribute GetInheritanceAttribute(IComponent component)
-        {
-            InheritanceAttribute attr = (InheritanceAttribute)_inheritedComponents[component];
-            attr ??= InheritanceAttribute.Default;
-
-            return attr;
-        }
+            => _inheritedComponents.TryGetValue(component, out InheritanceAttribute attr)
+                && attr is not null
```
lonitra, line 264: Let's also remove this null check then if we do not expect `attr` to be null

lonitra (dismissed):

dreddy-work (approved):

PR #8324, opened 2022-12-04
Enable nullability in Application

gpetrou:
## Proposed changes

- Enable nullability in Application.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8324)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs
@@ -385,7 +385,7 @@ namespace System.Windows.Forms
             ///  Allows you to setup a message filter for the application's message pump.  This
             ///  installs the filter on the current thread.
             /// </summary>
-            internal void AddMessageFilter(IMessageFilter f)
+            internal void AddMessageFilter(IMessageFilter? f)
             {
                 _messageFilters ??= new List<IMessageFilter>();
 
@@ -935,7 +935,7 @@ namespace System.Windows.Forms
             /// <summary>
             ///  Allows the hosting environment to register a callback
             /// </summary>
-            internal void RegisterMessageLoop(MessageLoopCallback callback)
+            internal void RegisterMessageLoop(MessageLoopCallback? callback)
             {
                 _messageLoopCallback = callback;
             }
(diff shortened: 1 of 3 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs
```diff
@@ -1013,7 +1013,7 @@ private unsafe void RunMessageLoopInner(msoloop reason, ApplicationContext conte
 
                     ApplicationContext = context;
 
-                    ApplicationContext.ThreadExit += new EventHandler(OnAppThreadExit);
+                    ApplicationContext!.ThreadExit += new EventHandler(OnAppThreadExit);
```
dreddy-work, line 1016: Looks like this method need to be cleaned but can go in a separate PR.

src/System.Windows.Forms/src/System/Windows/Forms/Application.cs
```diff
@@ -230,11 +228,11 @@ public static string CompanyName
                         // won't work with MC++ see GetAppMainType.
                         if (s_companyName is null || s_companyName.Length == 0)
                         {
-                            Type t = GetAppMainType();
+                            Type? t = GetAppMainType();
```
dreddy-work, line 231: nit:  can you change `initials` to full name  t->type  wherever we touch the source?

src/System.Windows.Forms/src/System/Windows/Forms/Application.cs
```diff
@@ -839,7 +840,7 @@ internal static void EndModalMessageLoop()
         /// <summary>
         ///  Overload of <see cref="Exit(CancelEventArgs)"/> that does not care about e.Cancel.
         /// </summary>
-        public static void Exit() => Exit(null);
+        public static void Exit() => Exit(null!);
```
dreddy-work, line 843: Any reason to use `!` ?  instead of changing the method and it was already expecting `null` events.

src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs
gpetrou: It is a bit confusing, but when we have `reason` equal to `msoloop.Main` this should not be null.

dreddy-work (approved):

PR #8325, opened 2022-12-04
Enable nullability in ControlItem

gpetrou:
## Proposed changes

- Enable nullability in ControlItem.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8325)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ErrorProvider.ControlItem.ControlItemAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ErrorProvider.ControlItem.ControlItemAccessibleObject.cs
@@ -139,7 +139,15 @@ namespace System.Windows.Forms
                 {
                     get
                     {
-                        Debug.Assert(_window!.AccessibilityObject.RuntimeId.Length >= 3);
+                        if (_window is null)
+                        {
+                            return new int[]
+                            {
+                                _controlItem.GetHashCode()
+                            };
+                        }
+
+                        Debug.Assert(_window.AccessibilityObject.RuntimeId.Length >= 3);
 
                         return new int[]
                         {
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ErrorProvider.ControlItem.ControlItemAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ErrorProvider.ControlItem.ControlItemAccessibleObject.cs
@@ -16,11 +16,15 @@ namespace System.Windows.Forms
             private class ControlItemAccessibleObject : AccessibleObject
             {
                 private readonly ControlItem _controlItem;
-                private readonly ErrorWindow _window;
-                private readonly Control _control;
+                private readonly ErrorWindow? _window;
+                private readonly Control? _control;
                 private readonly ErrorProvider _provider;
 
-                public ControlItemAccessibleObject(ControlItem controlItem, ErrorWindow window, Control control, ErrorProvider provider)
+                public ControlItemAccessibleObject(
+                    ControlItem controlItem,
+                    ErrorWindow? window,
+                    Control? control,
+                    ErrorProvider provider)
                 {
                     _controlItem = controlItem;
                     _window = window;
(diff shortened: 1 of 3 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ErrorProvider.ControlItem.ControlItemAccessibleObject.cs
```diff
@@ -16,11 +16,15 @@ internal partial class ControlItem
...
-                private readonly Control _control;
+                private readonly ErrorWindow? _window;
+                private readonly Control? _control;
                 private readonly ErrorProvider _provider;
 
-                public ControlItemAccessibleObject(ControlItem controlItem, ErrorWindow window, Control control, ErrorProvider provider)
+                public ControlItemAccessibleObject(
+                    ControlItem controlItem,
+                    ErrorWindow? window,
+                    Control? control,
```
gpetrou, line 26: In some existing tests such as `ControlItemAccessibleObjectTests_FragmentNavigate_NextSibling_ReturnsExpected` these  are null.

dreddy-work (dismissed):

src/System.Windows.Forms/src/System/Windows/Forms/ErrorProvider.ControlItem.ControlItemAccessibleObject.cs
```diff
@@ -135,7 +139,7 @@ internal override int[] RuntimeId
                 {
                     get
                     {
-                        Debug.Assert(_window.AccessibilityObject.RuntimeId.Length >= 3);
+                        Debug.Assert(_window!.AccessibilityObject.RuntimeId.Length >= 3);
```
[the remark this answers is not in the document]
gpetrou: I added a null check instead. Is this better?

src/System.Windows.Forms/src/System/Windows/Forms/ErrorProvider.ControlItem.ControlItemAccessibleObject.cs
Tanya-Solyanik: Would it make sense to fix the test?

Tanya-Solyanik: In what tests is `_control` null? This does not look right. `_window` is null until an error is encountered. But why would control be null?

src/System.Windows.Forms/src/System/Windows/Forms/ErrorProvider.ControlItem.ControlItemAccessibleObject.cs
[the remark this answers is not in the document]
Tanya-Solyanik: I guess it's ok, because accessible object with a windows will behave differently from the one with no window.

src/System.Windows.Forms/src/System/Windows/Forms/ErrorProvider.ControlItem.ControlItemAccessibleObject.cs
gpetrou: `ControlItemAccessibleObjectTests_ControlType_ReturnsExpected`
`ControlItemAccessibleObjectTests_Ctor_Default`
`ControlItemAccessibleObjectTests_FragmentNavigate_NextSibling_ReturnsExpected`
`ControlItemAccessibleObjectTests_FragmentNavigate_PreviousSibling_ReturnsExpected`
`ErrorWindowAccessibleObject_FragmentNavigate_ReturnsExpected`
`ErrorWindowAccessibleObject_GetChild_ReturnsExpected`
etc...
There are 12 tests in total.

gpetrou: @Tanya-Solyanik could we merge this now?

Tanya-Solyanik: I think this code was broken before you started working on it. And tests are using `null` because it was faster than creating and disposing an unused control.  I see 2 possible solutions, both requiring test fixes:

* `_control` is used to calculate accessible bounds. However.  this control is a [Parent ](https://github.com/dotnet/winforms/blob/15c4fc827e978603cea1daf19ea22394c3cbb281/src/System.Windows.Forms/src/System/Windows/Forms/ErrorProvider.ControlItem.cs#L75) of the control the error is associated with. Thus bounds will be incorrect.

* Assuming it should be the parent, code is still incorrect because `_control` is not updated after the accessible object is created and might be out of date (due to reparenting). We can instead access the parent when it's actually consumed. I.e. accessible item constructor takes ControlItem's control, not its parent and we access parent when calculating the bounds.

I suggest the second way for the nullability change, and opening a work item to investigate if bouonds are accurate.

gpetrou: @Tanya-Solyanik  I added a new commit. Is this what you had in mind?

Tanya-Solyanik: Yes, the change looks good.

Tanya-Solyanik (approved):

PR #8331, opened 2022-12-05
Enable nullability for TestUtilities Project

lonitra:
Follow up to https://github.com/dotnet/winforms/pull/8251#discussion_r1030797441

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8331)

### DIFF
--- a/src/Common/tests/TestUtilities/CommonMemberDataAttribute.cs
+++ b/src/Common/tests/TestUtilities/CommonMemberDataAttribute.cs
@@ -20,7 +20,7 @@ namespace System.Windows.Forms.TestUtilities
             MemberType = memberType;
         }
 
-        protected override object[] ConvertDataItem(MethodInfo testMethod, object item)
+        protected override object[]? ConvertDataItem(MethodInfo testMethod, object item)
         {
             if (item is null)
             {
--- a/src/Common/tests/TestUtilities/CommonMemberDataAttribute.cs
+++ b/src/Common/tests/TestUtilities/CommonMemberDataAttribute.cs
@@ -20,7 +20,7 @@ namespace System.Windows.Forms.TestUtilities
             MemberType = memberType;
         }
 
-        protected override object[]? ConvertDataItem(MethodInfo testMethod, object item)
+        protected override object[]? ConvertDataItem(MethodInfo testMethod, object? item)
         {
             if (item is null)
             {
--- a/src/Common/tests/TestUtilities/CommonTestHelper.cs
+++ b/src/Common/tests/TestUtilities/CommonTestHelper.cs
@@ -12,7 +12,7 @@ namespace System.Windows.Forms.TestUtilities
     {
         public static TheoryData GetEnumTypeTheoryData(Type enumType)
         {
-            var data = new TheoryData<Enum>();
+            TheoryData<Enum> data = new();
             foreach (Enum item in Enum.GetValues(enumType))
             {
                 data.Add(item);
@@ -24,7 +24,7 @@ namespace System.Windows.Forms.TestUtilities
         public static TheoryData GetEnumTypeTheoryDataWitIdentity(Type enumType)
         {
             int counter = 0;
-            var data = new TheoryData<string, int>();
+            TheoryData<string, int> data = new();
             foreach (Enum item in Enum.GetValues(enumType))
             {
                 data.Add(item.ToString(), counter++);
@@ -36,7 +36,7 @@ namespace System.Windows.Forms.TestUtilities
 
         public static TheoryData<Enum> GetEnumTypeTheoryDataInvalid(Type enumType)
         {
-            var data = new TheoryData<Enum>();
+            TheoryData<Enum> data = new();
             Enum[] values = Enum.GetValues(enumType).Cast<Enum>().OrderBy(p => p).Distinct().ToArray();
 
             for (int i = 0; i < values.Length - 2; i++)
@@ -62,7 +62,7 @@ namespace System.Windows.Forms.TestUtilities
 
         public static TheoryData<Enum> GetEnumTypeTheoryDataInvalidMasked(Type enumType)
         {
-            var data = new TheoryData<Enum>();
+            TheoryData<Enum> data = new();
             IEnumerable<Enum> values = Enum.GetValues(enumType).Cast<Enum>().OrderBy(p => p);
 
             long allMasked = 0;
@@ -80,19 +80,15 @@ namespace System.Windows.Forms.TestUtilities
 
         // helper method to generate theory data for all values of a boolean
         public static TheoryData<bool> GetBoolTheoryData()
-        {
-            var data = new TheoryData<bool>
+            => new()
             {
                 true,
                 false
             };
-            return data;
-        }
 
         // helper method to generate theory data for some values of a int
         public static TheoryData<int> GetIntTheoryData()
-        {
-            var data = new TheoryData<int>
+            => new()
             {
                 int.MinValue,
                 int.MaxValue,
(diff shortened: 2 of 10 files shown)


src/Common/tests/TestUtilities/XUnit/UseCultureAttribute.cs
```diff
@@ -92,12 +92,12 @@ public override unsafe void Before(MethodInfo methodUnderTest)
         /// <param name="methodUnderTest">The method under test</param>
         public override void After(MethodInfo methodUnderTest)
         {
-            Thread.CurrentThread.CurrentCulture = _originalCulture;
-            Thread.CurrentThread.CurrentUICulture = _originalUICulture;
+            Thread.CurrentThread.CurrentCulture = _originalCulture!;
```
Tanya-Solyanik, line 95: Is it possible to put a [MemberNotNull](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.membernotnullattribute?view=net-7.0) attribute on the `Before` method instead of the `!`?

src/Common/tests/TestUtilities/CommonMemberDataAttribute.cs
```diff
@@ -20,7 +20,7 @@ public CommonMemberDataAttribute(Type memberType, string memberName, params obje
             MemberType = memberType;
         }
 
-        protected override object[] ConvertDataItem(MethodInfo testMethod, object item)
+        protected override object[]? ConvertDataItem(MethodInfo testMethod, object item)
```
dreddy-work, line 23: and why item not marked nullable here?

src/Common/tests/TestUtilities/CommonTestHelper.cs
```diff
@@ -188,9 +188,9 @@ public static TheoryData<string> GetStringTheoryData()
             return data;
         }
 
-        public static TheoryData<string> GetStringWithNullTheoryData()
+        public static TheoryData<string?> GetStringWithNullTheoryData()
         {
-            var data = new TheoryData<string>
+            var data = new TheoryData<string?>
```
dreddy-work, line 193: as we are adapting following notion, you may want to make those changes as well for the lines/files your touch.
```suggestion
            TheoryData<string?>data = new
```

src/Common/tests/TestUtilities/XUnit/UseCultureAttribute.cs
```diff
@@ -21,8 +21,8 @@ public class UseCultureAttribute : BeforeAfterTestAttribute
     {
         private readonly Lazy<CultureInfo> _culture;
         private readonly Lazy<CultureInfo> _uiCulture;
-        private CultureInfo _originalCulture;
-        private CultureInfo _originalUICulture;
+        private CultureInfo? _originalCulture;
+        private CultureInfo? _originalUICulture;
```
RussKie, line 25: I don't believe either of these can ever be null:
```cs
            _originalCulture = Thread.CurrentThread.CurrentCulture;
            _originalUICulture = Thread.CurrentThread.CurrentUICulture;
```
```suggestion
        private CultureInfo _originalCulture = null!;
        private CultureInfo _originalUICulture = null!;
```

RussKie: https://learn.microsoft.com/en-us/dotnet/api/system.threading.thread.currentuiculture?view=net-7.0#exceptions

src/Common/tests/TestUtilities/XUnit/UseCultureAttribute.cs
lonitra: I tried this and still ended up needing to put a `!` but I will go with [Igor's suggestion](https://github.com/dotnet/winforms/pull/8331#discussion_r1040191364) to get around the `!` here.

src/Common/tests/TestUtilities/XUnit/UseCultureAttribute.cs
lonitra: Thank you! I wasn't quite sure the best way to get around the fact that these are not set in the constructor. It makes more sense that these do not have `?`

src/Common/tests/TestUtilities/CustomConverter.cs
```diff
@@ -14,6 +14,11 @@ public static class CustomConverter
     /// </summary>
     public static RegistrationScope RegisterConverter(Type type, TypeConverter converter)
     {
+        if (type is null || converter is null)
```
JeremyKuhne, line 17: nit: it's better to throw for which one is null, but that isn't critical as this is test code `ArgumentNullException.ThrowIfNull(type)`

src/Common/tests/TestUtilities/TestAccessor.cs
```diff
@@ -77,42 +78,42 @@ public TDelegate CreateDelegate<TDelegate>(string methodName = null)
...
                     catch (AmbiguousMatchException)
                     {
                         // More than one match for the name, specify the arguments
-                        methodInfo = type.GetMethod(
+                        methodInfo = type?.GetMethod(
                             binder.Name,
                             BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.Static,
                             binder: null,
-                            args.Select(a => a.GetType()).ToArray(),
+                            args is null ? Array.Empty<Type>() : args.Where(a => a is not null).Select(a => a!.GetType()).ToArray(),
```
JeremyKuhne, line 116: The right answer here is to throw if any of the parameters is null- we don't want to match random signatures.

src/Common/tests/TestUtilities/TestAccessors.cs
```diff
@@ -38,18 +38,20 @@ public static partial class TestAccessors
...
+            : (ITestAccessor?)Activator.CreateInstance(
                     typeof(TestAccessor<>).MakeGenericType(instanceOrType.GetType()),
                     instanceOrType);
+
+            if (testAccessor is null)
+            {
+                throw new ArgumentException("Cannot create TestAccessor for Nullable<T> instances with no value.");
             }
+
+            return testAccessor!;
```
JeremyKuhne, line 54: Why does this need a bang after you checked it?

src/Common/tests/TestUtilities/XUnit/UseCultureAttribute.cs
```diff
@@ -21,8 +21,8 @@ public class UseCultureAttribute : BeforeAfterTestAttribute
     {
         private readonly Lazy<CultureInfo> _culture;
         private readonly Lazy<CultureInfo> _uiCulture;
-        private CultureInfo _originalCulture;
-        private CultureInfo _originalUICulture;
+        private CultureInfo _originalCulture = null!;
```
JeremyKuhne, line 24: A comment would be good

JeremyKuhne (changes requested):
Yay! Only one non-nit comment.

src/Common/tests/TestUtilities/TestAccessors.cs
```diff
@@ -38,18 +38,20 @@ public static partial class TestAccessors
         /// </example>
         public static ITestAccessor TestAccessor(this object instanceOrType)
         {
-            if (instanceOrType is Type type)
-            {
-                return (ITestAccessor)Activator.CreateInstance(
+            ITestAccessor? testAccessor = instanceOrType is Type type
```
RussKie, line 41: Looks like broken indentation

src/Common/tests/TestUtilities/TestAccessor.cs
```diff
@@ -77,42 +78,44 @@ public TDelegate CreateDelegate<TDelegate>(string methodName = null)
...
                     catch (AmbiguousMatchException)
                     {
                         // More than one match for the name, specify the arguments
-                        methodInfo = type.GetMethod(
+                        methodInfo = type?.GetMethod(
                             binder.Name,
                             BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.Static,
                             binder: null,
-                            args.Select(a => a.GetType()).ToArray(),
+                            args.Where(a => a is not null).Select(a => a!.GetType()).ToArray(),
```
JeremyKuhne, line 118: You can't have the `Where` condition here.  We're trying to find a matching signature with `args.Length` arguments. So far we haven't hit a situation where we're trying to pass null to an overloaded method. When we do we'll have to make this more sophisticated to find a single match for all of the non-null arguments.

JeremyKuhne (approved):
🚀

PR #8332, opened 2022-12-06
Refactor `SerializationResourceManager` to use `Dictionary` instead of `HashTable`

elachlan:
- Refactored `ResourceCodeDomSerializer.SerializationResourceManager` to replace usages of `Hashtable`. 
- Refactored `ResXResourceReader` to use `Dictionary` instead of `ListDictionary`

Related: #8143

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8332)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ResourceCodeDomSerializer.SerializationResourceManager.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ResourceCodeDomSerializer.SerializationResourceManager.cs
@@ -24,7 +24,7 @@ namespace System.ComponentModel.Design.Serialization
             private CultureInfo _localizationLanguage;
             private IResourceWriter _writer;
             private CultureInfo _readCulture;
-            private readonly HashSet<string> _nameTable;
+            private readonly Dictionary<string, int> _nameTable;
             private Dictionary<CultureInfo, Dictionary<string, object>> _resourceSets;
             private Dictionary<string, object> _metadata;
             private Dictionary<string, object> _mergedMetadata;
@@ -207,54 +207,35 @@ namespace System.ComponentModel.Design.Serialization
             }
 
             /// <summary>
-            ///  This determines if the given resource name/value pair can be retrieved from a parent culture.  We don't want to write duplicate resources for each language, so we do a check of the parent culture.
+            ///  This determines if the given resource name/value pair can be retrieved from a parent culture.
+            ///  We don't want to write duplicate resources for each language, so we do a check of the parent culture.
             /// </summary>
             private CompareValue CompareWithParentValue(string name, object value)
             {
                 Debug.Assert(name is not null, "name is null");
-                // If there is no parent culture, treat that as being different from the parent's resource, which results in the "normal" code path for the caller.
-                if (ReadCulture.Equals(CultureInfo.InvariantCulture))
+                // If there is no parent culture, treat that as being different from the parent's resource.
+                // which results in the "normal" code path for the caller.
+                return ReadCulture.Equals(CultureInfo.InvariantCulture)
+                    ? CompareValue.Different
+                    : CompareWithParentValue(ReadCulture, name, value);
+            }
+
+            private CompareValue CompareWithParentValue(CultureInfo culture, string name, object value)
+            {
+                Debug.Assert(culture.Parent != culture, "should have returned when culture = InvariantCulture");
+                CultureInfo parent = culture.Parent;
+                Dictionary<string, object> resourceSet = GetResourceSet(culture);
+                bool contains = (resourceSet is null) ? false : resourceSet.ContainsKey(name);
+                if (resourceSet is not null && resourceSet.TryGetValue(name, out object parentValue))
                 {
-                    return CompareValue.Different;
+                    return !parentValue.Equals(value) || parentValue is null ? CompareValue.Different : CompareValue.Same;
+                }
+                else if (culture.Equals(CultureInfo.InvariantCulture))
+                {
+                    return CompareValue.New;
                 }
 
-                CultureInfo culture = ReadCulture;
-#pragma warning disable SA1009 // Closing parenthesis should be spaced correctly
-                for (; ; )
-#pragma warning restore SA1009 // Closing parenthesis should be spaced correctly
-                {
-                    Debug.Assert(culture.Parent != culture, "should have exited loop when culture = InvariantCulture");
-                    culture = culture.Parent;
-                    Dictionary<string, object> rs = GetResourceSet(culture);
-                    bool contains = (rs is null) ? false : rs.ContainsKey(name);
-                    if (contains)
-                    {
-                        object parentValue = rs?[name];
-                        if (parentValue == value)
-                        {
-                            return CompareValue.Same;
-                        }
-                        else if (parentValue is not null)
-                        {
-                            if (parentValue.Equals(value))
-                            {
-                                return CompareValue.Same;
-                            }
-                            else
-                            {
-                                return CompareValue.Different;
-                            }
-                        }
-                        else
-                        {
-                            return CompareValue.Different;
-                        }
-                    }
-                    else if (culture.Equals(CultureInfo.InvariantCulture))
-                    {
-                        return CompareValue.New;
-                    }
-                }
+                return CompareWithParentValue(parent, name, value);
             }
 
             /// <summary>
@@ -732,7 +713,11 @@ namespace System.ComponentModel.Design.Serialization
             }
 
             /// <summary>
-            ///  Writes the given resource value under the given name. This checks the parent resource to see if the values are the same.  If they are, the resource is not written.  If not, then the resource is written.  We always write using the resource language we read in with, so we don't stomp on the wrong resource data in the event that someone changes the language.
+            ///  Writes the given resource value under the given name.
+            ///  This checks the parent resource to see if the values are the same.
+            ///  If they are, the resource is not written. If not, then the resource is written.
+            ///  We always write using the resource language we read in with,
+            ///  so we don't stomp on the wrong resource data in the event that someone changes the language.
             /// </summary>
             public string SetValue(IDesignerSerializationManager manager, ExpressionContext tree, object value, bool forceInvariant, bool shouldSerializeInvariant, bool ensureInvariant, bool applyingCachedResources)
             {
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Resources/ResXResourceReader.cs
```diff
@@ -314,7 +313,7 @@ public IDictionaryEnumerator GetEnumerator()
         {
             _isReaderDirty = true;
             EnsureResData();
-            return _resData.GetEnumerator();
+            return ((IDictionary)_resData).GetEnumerator();
```
elachlan, line 316: This part was required, which might make the whole thing pointless if there is a performance cost for the cast? 

Ultimately it would be nice if these were updated to be properly typed, but that is a breaking change.

elachlan: `IDictionary` is one of the last implemented interfaces, so performance won't be so great.
https://www.thomasclaudiushuber.com/2020/03/05/c-the-order-of-interfaces-is-important-for-casting-performance/

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ResourceCodeDomSerializer.SerializationResourceManager.cs
```diff
@@ -333,52 +300,39 @@ public IDictionaryEnumerator GetMetadataEnumerator()
...
             public IDictionaryEnumerator GetEnumerator(CultureInfo culture)
             {
-                Hashtable ht = GetResourceSet(culture);
-                if (ht is not null)
-                {
-                    return ht.GetEnumerator();
-                }
-
-                return null;
+                Dictionary<string, object> ht = GetResourceSet(culture);
```
dreddy-work, line 328: nullable declaration?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ResourceCodeDomSerializer.SerializationResourceManager.cs
```diff
@@ -333,52 +300,39 @@ public IDictionaryEnumerator GetMetadataEnumerator()
             {
                 if (_mergedMetadata is null)
                 {
-                    Hashtable t = GetMetadata();
+                    Dictionary<string, object> t = GetMetadata();
                     if (t is not null)
                     {
                         // This is for backwards compatibility and also for the case when our reader/writer don't support metadata.  We must merge the original enumeration data in here or  else existing design time properties won't show up.  That would be really bad for things like Localizable.
-                        Hashtable it = GetResourceSet(CultureInfo.InvariantCulture);
+                        Dictionary<string, object> it = GetResourceSet(CultureInfo.InvariantCulture);
```
dreddy-work, line 307: nullable?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ResourceCodeDomSerializer.SerializationResourceManager.cs
```diff
@@ -333,52 +300,39 @@ public IDictionaryEnumerator GetMetadataEnumerator()
...
+                return ht?.GetEnumerator();
             }
 
             /// <summary>
             ///  Loads the metadata table
             /// </summary>
-            private Hashtable GetMetadata()
+            private Dictionary<string, object> GetMetadata()
             {
                 if (_metadata is null)
```
dreddy-work, line 337: invert this if block.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ResourceCodeDomSerializer.SerializationResourceManager.cs
```diff
@@ -333,52 +300,39 @@ public IDictionaryEnumerator GetMetadataEnumerator()
...
+                    Dictionary<string, object> t = GetMetadata();
                     if (t is not null)
                     {
                         // This is for backwards compatibility and also for the case when our reader/writer don't support metadata.  We must merge the original enumeration data in here or  else existing design time properties won't show up.  That would be really bad for things like Localizable.
-                        Hashtable it = GetResourceSet(CultureInfo.InvariantCulture);
+                        Dictionary<string, object> it = GetResourceSet(CultureInfo.InvariantCulture);
                         if (it is not null)
                         {
-                            foreach (DictionaryEntry de in it)
+                            foreach (KeyValuePair<string, object> de in it)
```
dreddy-work, line 310: nit: replace short forms for the sources you are touching.

dreddy-work:
@elachlan, I see partial nullability in these files. There are multiple instances that needs to be declared as nullable.

src/System.Windows.Forms/src/System/Resources/ResXResourceReader.cs
elachlan: @JeremyKuhne I'd love your input on this. The move to generics might be a performance regression here.

JeremyKuhne: Casting is not going to measure in a scenario like this. While there is a difference, it is incredibly small. A few nanoseconds are not going to matter in something that won't run more than a few times.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ResourceCodeDomSerializer.SerializationResourceManager.cs
```diff
@@ -426,25 +384,13 @@ public object GetObject(string resourceName, bool forceInvariant)
...
 
                 object value = null;
                 while (value is null)
                 {
-                    Hashtable rs = GetResourceSet(culture);
-                    if (rs is not null)
-                    {
-                        value = rs[resourceName];
-                    }
+                    Dictionary<string, object> rs = GetResourceSet(culture);
```
dreddy-work, line 392: @elachlan, can you take care of short names as well and renaming them for the sources you are touching?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ResourceCodeDomSerializer.SerializationResourceManager.cs
```diff
@@ -874,23 +803,23 @@ public string SetValue(IDesignerSerializationManager manager, ExpressionContext
                 {
```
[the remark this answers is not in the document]
elachlan: I think I can write this into a `Dictionary<string, int>` and use the baseName as the key. Therefore keeping track of the count. I'll push through once I work out the proper logic.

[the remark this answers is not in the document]
elachlan: @dreddy-work done! Both should be easier to understand now I think.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ResourceCodeDomSerializer.SerializationResourceManager.cs
```diff
@@ -231,62 +207,43 @@ public override void ApplyResources(object value, string objectName, CultureInfo
...
+                    ? CompareValue.Different
+                    : CompareWithParentValue(ReadCulture, name, value);
+            }
+
+            private CompareValue CompareWithParentValue(CultureInfo culture, string name, object value)
+            {
+                Debug.Assert(culture.Parent != culture, "should have returned when culture = InvariantCulture");
+                CultureInfo parent = culture.Parent;
+                Dictionary<string, object> resourceSet = GetResourceSet(culture);
+                bool contains = (resourceSet is null) ? false : resourceSet.ContainsKey(name);
```
dreddy-work, line 228: you don't need this anymore.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ResourceCodeDomSerializer.SerializationResourceManager.cs
```diff
@@ -231,62 +207,43 @@ public override void ApplyResources(object value, string objectName, CultureInfo
...
+                if (resourceSet is not null && resourceSet.TryGetValue(name, out object parentValue))
                 {
-                    return CompareValue.Different;
+                    return !parentValue.Equals(value) || parentValue is null ? CompareValue.Different : CompareValue.Same;
```
dreddy-work, line 231: what if both `parentValue` and `value` is null?

dreddy-work: @elachlan , did you run any tests here? This is core for resource serialization and like to do some unit testing performed with this change and record results before we merge it.

elachlan: I did not do any manual testing, but did run the Design Unit Tests.

The ordering seemed important here. Because previously it was checking if both were equal before checking for null. Meaning that its designed to allow both to be null for `Same` to be returned. So I replicated that logic here.

dreddy-work: can you try a simple application with multiple sets of controls and localize the application?

elachlan: I have no idea how to do this. You mean as a part of one of the test projects?

dreddy-work: No, by replacing binaries locally on your machine and create .NET 8.0 targeted application.  Here are instructions on private testing in case you have not used them before: https://github.com/dotnet/winforms/blob/main/docs/testing.md

elachlan: I was playing around with the tests, the code is only hit in `SerializationResourceManager` when the `PrimitiveCodeDomSerializer` is serializing a string with a char count > 200. This isn't tested for.

I tried testing this scenario and writing a separate test class for `ResourceCodeDomSerializer`. Both resulted in exceptions due to either a missing session. If I manually created a session, then the exception would be on `SerializationResourceManager.Writer` with a missing `ResourceWriter` service.

If I could work out how to add a writer service in the test, then I can write tests for it.

dreddy-work: good to know that. Logically, change is no-op here. Lets take it in after addressing rest of the comments.

dreddy-work (approved):

PR #8333, opened 2022-12-06
Cache Top-level Window Form's size in PMV2 mode applications

### LINKED ISSUE #8341: PMv2 mode application shrink in Size when moved between monitors
### .NET version

All .NET versions including .NET framework

No

No

### Issue description

In a PermonitorV2 mode applications, Windows Form when moved between monitors of different DPI settings repetitively, shrinks the Form size and eventually reaches to Zero after few iterations. Please see below gif on how Form's size is impacted when moved between monitors.

### Steps to reproduce

Clean simple Winforms application running in PermonitorV2 mode. 
Multimonitor setup where monitors have different DPI scale settings on them ( i.e 100% and 200%)
Run the app and move between monitors couple of times as shown in gif.

FormSizeCache

dreddy-work:
fixes #8341 
When `AutoScaleMode` is `Font`, for PMV2 mode applications, and top-level `Form` is moved from one monitor to the other with different DPI settings, we compute the new size based on the scale factor we compute for the `Font` on the new monitor. This sometimes, especially when the DPI setting on those monitors is not multiple of 100s, causes rounding issues in the size computations. Rounding issue gets worse when user move top-level `Form` between monitors multiple times. Caching size for the Form, prevents rounding issues in this scenario.

Also, moved high DPI tests into integration tests to run them synchronously and be able to set DpiAwareness on the thread. 

Before: 
FormSizeCache

After: 
FormSizeCache_2

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8333)


elachlan:
@dreddy-work flaky test?

dreddy-work:
> @dreddy-work flaky test?

Yep. Also, i am still working on it. Will reset when this is ready for review.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.cs
```diff
@@ -34,7 +34,7 @@ internal static partial class DpiHelper
 
         static DpiHelper() => Initialize();
 
-        private static void Initialize()
+        internal static void Initialize()
```
JeremyKuhne, line 37: If you're only doing this for tests, use the TestAccessor.

src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
```diff
@@ -1446,9 +1446,12 @@ internal void ScaleContainerForDpi(int deviceDpiNew, int deviceDpiOld, Rectangle
 
                 if (LocalAppContextSwitches.ScaleTopLevelFormMinMaxSizeForDpi)
                 {
-                    // The suggested rectangle comes from Windows, and it does not match with our calculations for scaling controls by AutoscaleFactor.
-                    // Hence, we cannot use AutoscaleFactor here for scaling the control properties. See the below description for more details.
-                    ScaleMinMaxSize(AutoScaleFactor.Width, AutoScaleFactor.Height, updateContainerSize: false);
+                    // AutoscaleFactor is not updated until after OnFontChanged event is raised. Hence, computing
+                    // factor based on the change in bounds of the Form, which aligns with AutoscaleFactor for both
+                    // AutoscaleMode is Font and/or Dpi. Especially after adding support for non-leaner Form size in PMv2.
```
JeremyKuhne, line 1451:
```suggestion
                    // AutoscaleMode is Font and/or Dpi. Especially after adding support for non-linear Form size in PMv2.
```

src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
```diff
@@ -1446,9 +1446,12 @@ internal void ScaleContainerForDpi(int deviceDpiNew, int deviceDpiOld, Rectangle
 
                 if (LocalAppContextSwitches.ScaleTopLevelFormMinMaxSizeForDpi)
                 {
-                    // The suggested rectangle comes from Windows, and it does not match with our calculations for scaling controls by AutoscaleFactor.
-                    // Hence, we cannot use AutoscaleFactor here for scaling the control properties. See the below description for more details.
-                    ScaleMinMaxSize(AutoScaleFactor.Width, AutoScaleFactor.Height, updateContainerSize: false);
+                    // AutoscaleFactor is not updated until after OnFontChanged event is raised. Hence, computing
```
JeremyKuhne, line 1449:
```suggestion
                    // AutoscaleFactor is not updated until after the OnFontChanged event is raised. Hence, computing
```

src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
```diff
@@ -156,6 +156,13 @@ public partial class Form : ContainerControl
 
         private VisualStyleRenderer? _sizeGripRenderer;
 
+        // Cache Form's size for the DPI. When Form is moved between the monitors with
+        // different DPI settings, we use cached values to set the size matching the DPI
+        // on the Form instead of recalculating the size again. This help preventing rounding
+        // error in size calculations with float DPI factor and rounding it to nearest integer.
+        internal Dictionary<int, Size>? FormSizeCache;
```
JeremyKuhne, line 163: `DpiFormSizes` would be clearer. And again, don't make things internal just for tests. Use the `TestAccessor`.

src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
```diff
@@ -156,6 +156,13 @@ public partial class Form : ContainerControl
...
+        private bool _clearFormSizeCache = true;
```
JeremyKuhne, line 164: This would be much better tied to the scenario by `_inDpiChanged`, `_processingDpiChanged` or something similar.

JeremyKuhne (changes requested):
Added a few comments

src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
```diff
@@ -4282,10 +4294,35 @@ protected virtual void OnDpiChanged(DpiChangedEventArgs e)
...
+                        FormSizeCache.Add(e.DeviceDpiOld, Size);
+                    }
+
+                    // Prevent clearing Form's size cache while applying bounds from DPI change.
+                    // Any other events that cause Form's size change should clear the cache.
+                    _clearFormSizeCache = false;
+                }
+
+                ScaleContainerForDpi(e.DeviceDpiNew, e.DeviceDpiOld, e.SuggestedRectangle);
+                _clearFormSizeCache = true;
```
JeremyKuhne, line 4325: Should probably use a try finally to clear the flag.

src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
RussKie, line 163: [naming]
```suggestion
        internal Dictionary<int, Size>? _formSizeCache;
```

src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
```diff
@@ -156,6 +156,13 @@ public partial class Form : ContainerControl
 
         private VisualStyleRenderer? _sizeGripRenderer;
 
+        // Cache Form's size for the DPI. When Form is moved between the monitors with
+        // different DPI settings, we use cached values to set the size matching the DPI
+        // on the Form instead of recalculating the size again. This help preventing rounding
+        // error in size calculations with float DPI factor and rounding it to nearest integer.
```
RussKie, line 162: This can be reflowed to 120 chars.

src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
```diff
@@ -4282,10 +4293,41 @@ protected virtual void OnDpiChanged(DpiChangedEventArgs e)
...
+                    return;
+                }
+
+                try
                 {
+                    // Cache Form's size for the current and new DPI if not already done. We do this only
+                    // for AutoScaleMode is Font. In other modes, Windows OS will compute Form's size.
+                    if (AutoScaleMode == AutoScaleMode.Font)
+                    {
+                        _dpiFormSizes ??= new Dictionary<int, Size>();
```
Tanya-Solyanik, line 4307: This dictionary should be cleared if form's `AutoScaleMode` or Form's font is changed dynamically.

dreddy-work: Thanks @Tanya-Solyanik. When do you expect `AutoScaleMode` change dynamically? That would break lot more things and i haven't come across any such scenario so far.

`FontChange `make sense. I will update it in the next PR.

Tanya-Solyanik: >When do you expect AutoScaleMode change dynamically?

This is a public property, it can change any time.

PR #8337, opened 2022-12-06
Refactor `CodeDomDesignerLoader` to replace `HashTable` and `ArrayList` usages

elachlan:
Related: #8143, #8140

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomDesignerLoader.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomDesignerLoader.cs
@@ -213,7 +213,7 @@ namespace System.ComponentModel.Design.Serialization
             {
                 // We keep track of any failures here.  If we failed to find a type this
                 // array list will contain a list of strings listing what we have tried.
-                ArrayList failures = null;
+                List<string> failures = null;
                 bool firstClass = true;
 
                 if (_documentCompileUnit.UserData[typeof(InvalidOperationException)] is not null)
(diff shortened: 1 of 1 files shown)


elachlan:
flaky test

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomDesignerLoader.cs
```diff
@@ -423,7 +422,9 @@ private bool IntegrateSerializedTree(IDesignerSerializationManager manager, Code
             bool lockField = false;
             int methodInsertLocation = 0;
             bool lockMethod = false;
-            IDictionary docMemberHash = new HybridDictionary(docDecl.Members.Count, caseInsensitive);
+            Dictionary<string, int> docMemberHash = new(docDecl.Members.Count, caseInsensitive
```
dreddy-work, line 425: why do you want to change this from `HybridDictionary`?

elachlan: The main thought process was to move to generics. Is there an idea of how large the dictionary will get?

dreddy-work: That depends on application size.  It was documented to gain perf if count is small otherwise no-op.

elachlan: I think there will be gains from avoiding unboxing compared to using a hybrid dictionary. In the original code, we grab the value, check its not null, then cast it to int. The new code avoids the cast by using generics and does the other two ops via TryGetValue.

dreddy-work: make sense.!.

dreddy-work (approved):

lonitra (approved):
🚀

PR #8351, opened 2022-12-08
Changed type of the `handle` parameter from `IntPtr` to  `HWND` in the `Control.ReleaseUiaProvider` method

### LINKED ISSUE #8217: Unifying an argument in the `ReleaseUiaProvider` method to `HWND`.
Replace the type of the `handle` parameter from `IntPtr` to `IHandle<HWND>` in the Control.ReleaseUiaProvider method and its overrides. 

        Suggestion - consider unifying argument in this method to `nint` in the next PR.

_Originally posted by @Tanya-Solyanik in https://github.com/dotnet/winforms/pull/7349#discussion_r1023472589_
 📓 - modified bug title and body per @elachlan suggestion

NikitaSemenovAkvelon:
Fixes #8217

## Proposed changes

- Changed type to `HWND`.
- Refactored `Control.HWNDInternal` property to remove extra convert.
- Used `HWND` value instead of `Handle`.
- Used `HWND.Null` value instead of `IntPtr.Zero`.

## Customer Impact

- Method becomes more clear and an argument type is unified.

## Regression? 

- No

## Risk

- Minimal

## Test methodology 

- Manual
- Unit tests

## Accessibility testing  

- Inspect 

## Test environment(s) 

.NET SDK:
 Commit:    1b80461e45

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.22621

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8351)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/Interop/UiaCore/Interop.UiaReturnRawElementProvider.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/UiaCore/Interop.UiaReturnRawElementProvider.cs
@@ -9,7 +9,7 @@ internal partial class Interop
     internal static partial class UiaCore
     {
         [DllImport(Libraries.UiaCore, CharSet = CharSet.Unicode)]
-        public static extern nint UiaReturnRawElementProvider(IntPtr hwnd, nint wParam, nint lParam, IRawElementProviderSimple? el);
+        public static extern nint UiaReturnRawElementProvider(HWND hwnd, nint wParam, nint lParam, IRawElementProviderSimple? el);
 
         public static nint UiaReturnRawElementProvider(IHandle<HWND> hwnd, WPARAM wParam, LPARAM lParam, IRawElementProviderSimple? el)
         {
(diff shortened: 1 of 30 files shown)


elachlan:
LGTM 🚀

My only nit would be to have two methods, one for the HWNDInternal release and the other for the HWND.Null release. That way there is no way to call a release with anything else.

dmitrii-drobotov (approved):

NikitaSemenovAkvelon:
> My only nit would be to have two methods, one for the HWNDInternal release and the other for the HWND.Null release. That way there is no way to call a release with anything else.

I'm not sure if it is a good idea because we have overriding of this method in many classes. 
If we split it into two parameterless methods we will have to have a main with the parameter for overriding.

lonitra (approved):
LGTM!

Tanya-Solyanik (approved):

Issue #8342, opened 2022-12-07
Enable nullable reference types for `System.Windows.Forms.Design`


elachlan:
### Is your feature request related to a problem? Please describe

`System.Windows.Forms.Design` does not have `<Nullable>enable</Nullable>` in its project file and the individual files aren't marked to enable/disable NRTs.

I am creating this issue to help track the progress

I suggest the following plan which has been implemented previously with other projects:
1. Enable Nullable in the project file and marking each file with `#nullable disable` where needed.
2. Community/team PRs to slowly enable nullability in the code base.

Related: #1107

### Describe the solution you'd like and alternatives you've considered

N/A

### Will this feature affect UI controls?

N/A

elachlan:
I have started work on this. I immediately ran into interfaces in runtime which aren't annotated. Should interfaces be annotated with NRTs?

halgab:
I think we're going to have issues with all `CodeDom` related code. Apparently the upstream System.CodeDom dll is not annotated: https://github.com/dotnet/runtime/issues/41720, https://github.com/dotnet/runtime/issues/78036.

lonitra:
The team has decided to prioritize nullability changes to files in `System.Windows.Forms` as there are a greater number of users and we have much fuller test coverage here. Current PRs that are open that change nullability in `System.Windows.Forms.Design` space will be taken to completion, however we will not be prioritizing any new nullability changes in the `System.Windows.Forms.Design` area. Thank you so much for the amazing work community has done in this space! The team will revisit enabling nullability for files in `System.Windows.Forms.Design` once enabling nullability is complete in `System.Windows.Forms`.

FYI @halgab , @gpetrou

lonitra:
Thank you for your patience on this! We've revisited this and decided we will start taking a look at new nullability changes in `System.Windows.Forms.Design` area again. However, when enabling nullability for classes that reside in the`System.Windows.Forms.Design`, test coverage for public methods on the class should be added if it is lacking coverage. These tests should pass before and after nullability changes here to ensure that unintentional behavior changes are not made. 

FYI @halgab , @gpetrou

PR #8353, opened 2022-12-09
Refactor `DesignerSerializationManager` to replace `HashTable` and enable Nullability

elachlan:
Refactored to replace HashTable
Refactored for null-ability annotations
General code cleanup

Dependent on runtime changes in: https://github.com/dotnet/runtime/pull/79429

Related: #8342, #8143


src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/DesignerSerializationManager.cs
```diff
@@ -17,27 +15,27 @@ namespace System.ComponentModel.Design.Serialization
     /// </summary>
     public class DesignerSerializationManager : IDesignerSerializationManager
     {
-        private readonly IServiceProvider provider;
-        private ITypeResolutionService typeResolver;
+        private readonly IServiceProvider? provider;
```
Tanya-Solyanik, line 18: This readonly field is assigned in the constructor, why is it nullable?

elachlan: Because there is a constructor without any parameters, which does not set provider.

lonitra (changes requested):
There seems to be a bug in the original code. Otherwise, mostly about cleaning up some old comments

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/DesignerSerializationManager.cs
```diff
@@ -485,11 +496,14 @@ public object GetSerializer(Type objectType, Type serializerType)
                     continueLoop = false;
                     foreach (IDesignerSerializationProvider provider in designerSerializationProviders)
                     {
-                        object newSerializer = provider.GetSerializer(this, serializer, objectType, serializerType);
-                        if (newSerializer is not null)
+                        if (serializer is not null && objectType is not null)
```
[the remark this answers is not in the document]
elachlan: `IDesignerSerializationProvider.GetSerializer` doesn't allow nulls in its `currentSerializer` and `objectType` parameters. Nullability on that interface could be wrong.

elachlan:
I think that `TypeDescriptor.CreateInstance` should change the `args` parameter from `object[]?` to `object?[]?`.
https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.typedescriptor.createinstance?view=net-7.0

lonitra (commented):
Small comment on whether a null check can be removed, otherwise LGTM!

lonitra:
> I think that `TypeDescriptor.CreateInstance` should change the `args` parameter from `object[]?` to `object?[]?`. https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.typedescriptor.createinstance?view=net-7.0

Looking at [source code](https://source.dot.net/#System.ComponentModel.TypeConverter/System/ComponentModel/TypeDescriptor.cs,394) it may need to be addressed in TypeDescriptorProvider. I believe [runtime](https://github.com/dotnet/runtime/issues) owns TypeDescriptor. Feel free to raise in issue.

elachlan:
@lonitra I'm digging through runtime now. `Activator.CreateInstance()` has `object?[]?` for `args`. So just need to update all the annotations up through the stack. I'll put together a PR.

lonitra:
> @lonitra I'm digging through runtime now. `Activator.CreateInstance()` has `object?[]?` for `args`. So just need to update all the annotations up through the stack. I'll put together a PR.

Thanks for doing that!

lonitra (commented):
This LGTM :) but looks like we have some build errors. If you build on command line (cd to winforms repo on command line and type 'build') these errors should show up

elachlan:
Thanks, some of the build errors are related to the winforms PRs. I'll check over it again.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/DesignerSerializationManager.cs
```diff
@@ -192,18 +189,23 @@ private void CheckSession()
...
+        ///  Creates an instance of the specified type.
+        ///  The default implementation will create the object.
+        ///  If the object implements IComponent and only requires an empty or IContainer style constructor,
+        ///  this will search for IDesignerHost and create through the host. Otherwise it will use reflection.
+        ///  If addToContainer is true, this will add to the container using this class's Container property,
+        ///  using the name provided if it is not null.
         /// </summary>
-        protected virtual object CreateInstance(Type type, ICollection arguments, string name, bool addToContainer)
+        protected virtual object CreateInstance(Type type, ICollection? arguments, string? name, bool addToContainer)
```
elachlan, line 199: We get a `CS8603: Possible null reference return.` on instance at the end of this function. I am unsure how to resolve this. Should I be throwing an exception if the instance is null?

`TypeDescriptor.CreateInstance` can return null, I think this code is expecting it to return a `MissingMethodException` on failing to create an instance?
https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.typedescriptor.createinstance?view=net-7.0

If we are not going to change the interface for `IDesignerSerializationManager.CreateInstance` and add a nullability annotation, then I think the only way around it is to throw an exception.

lonitra: > I think this code is expecting it to return a `MissingMethodException` on failing to create an instance? 

I agree here. [Docs for this method](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.design.serialization.designerserializationmanager.createinstance?view=windowsdesktop-7.0) also seem to indicate we should not be returning a null instance. Let's do a check for null before returning and throw `SerializationException` if it is. You may need to factor out the code in the catch where we create the `SerializationException`.

elachlan:
Do we add exception information to XML comments?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/DesignerSerializationManager.cs
```diff
@@ -263,66 +262,67 @@ protected virtual object CreateInstance(Type type, ICollection arguments, string
                     try
                     {
                         // First, just try to create the object directly with the arguments.  generally this should work.
-                        instance = TypeDescriptor.CreateInstance(provider, type, null, argArray);
+                        instance = TypeDescriptor.CreateInstance(provider, type, argTypes: null, argArray);
```
lonitra, line 265: Build is complaining that the nullability does not match.. I wonder if we can get away with casting `argArray` to `object[]?` and add a comment to remove cast when the nullability mismatch is fixed, linking the issue/PR

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/DesignerSerializationManager.cs
```diff
@@ -263,66 +262,67 @@ protected virtual object CreateInstance(Type type, ICollection arguments, string
...
 
-                                // all of the parameters were converted or matched, so try the creation again. if that works, we're in the money.
-                                if (match)
-                                {
-                                    instance = TypeDescriptor.CreateInstance(provider, type, null, tempArgs);
-                                    break;
+                                    // All of the parameters were converted or matched, so try the creation again.
+                                    if (match)
+                                    {
+                                        instance = TypeDescriptor.CreateInstance(provider, type, argTypes: null, tempArgs);
```
lonitra, line 318: Ditto here.

lonitra:
> Do we add exception information to XML comments?

Sorry, I'm not sure I follow. Which XML comments?

elachlan:
> > Do we add exception information to XML comments?
> 
> Sorry, I'm not sure I follow. Which XML comments?

Just as a general rule for Winforms code base. Do we document in the XML comments what exceptions can be thrown from a function?

lonitra:
> Just as a general rule for Winforms code base. Do we document in the XML comments what exceptions can be thrown from a function?

Yes, we should document what exceptions and what scenarios they will be thrown.

elachlan:
Just waiting on the runtime PR to be merged.

elachlan:
@lonitra I have rebased on main, which has resolved the nullability mismatch.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/DesignerSerializationManager.cs
```diff
@@ -538,31 +552,22 @@ protected virtual Type GetType(string typeName)
...
-            {
-                type = Type.GetType(typeName);
-            }
-            else
-            {
-                type = typeResolver.GetType(typeName);
-            }
-
-            return type;
+            return typeName is null ? null : typeResolver is null ? Type.GetType(typeName) : typeResolver.GetType(typeName);
```
lonitra, line 566: There are some tests failing and I think some of it might be because of this change. I believe this used to throw an ArgumentNullException if `typeName` was `null`, why did we change to return null right away when `typeName` is null?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/DesignerSerializationManager.cs
```diff
@@ -371,16 +349,46 @@ protected virtual object CreateInstance(Type type, ICollection arguments, string
...
                     }
                 }
             }
 
-            return instance;
+            return instance is null ? throw GetSerializationException() : instance;
+
+            SerializationException GetSerializationException()
+            {
+                StringBuilder argTypes = new StringBuilder();
```
lonitra, line 365: nit
```suggestion
                StringBuilder argTypes = new();
```

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/DesignerSerializationManager.cs
elachlan: This was a code fix refactor. It should be the same as before. I can add a null check to typename if you want. But it didn't have one before.

lonitra:
```c#
typeResolver is null ? Type.GetType(typeName) : typeResolver.GetType(typeName);
```
Looks to be a refactor, but before this we are doing:
```c#
typeName is null ? null : ....
```
which I don't think was happening before unless I am missing something?

elachlan: Apologies, I missed that. You are correct. So it was done to fix the null annotations. Because `GetType` doesn't have null annotations. But we can't throw an `ArgumentNullException` because the tests fail. 

I am not sure what to do.

elachlan: I think either `ITypeResolutionService.GetType` needs to have a null annotation on its `name` parameter or the test needs to be changed to not pass null to a non-null annotated interface. 

I am leaning towards the test, since its a moq test. But I need someone more knowledgeable to confirm this.

elachlan: The test is inputting null value to `GetType` and expecting a Integer Type. Which makes zero sense. But an implementation might resolve a null value to an integer type for whatever reason.

elachlan: This test fails when we return `null`.
https://github.com/dotnet/winforms/blob/f6eaa898b9f130b99b6249161e8133a6b91f8f8b/src/System.Windows.Forms.Design/tests/UnitTests/System/ComponentModel/Design/Serialization/DesignerSerializationManagerTests.cs#L1416-L1473

elachlan: `DesignerSerializationManager_AddSerializationProvider_NonNullProvider_GetSerializerReturnsExpected` also fails because of changes in `GetSerializer`

elachlan: Looks like there are a bunch of failing tests for a few different problems. I will investigate and fix as many as I can and get back to you. It might be sometime next week.

elachlan: The current failing tests:
image

lonitra: Thank you for all your effort on this!

elachlan: I have solved the test failures. There are now compile warnings because of nullability.

The issues are around these two interface methods:

- [IDesignerSerializationProvider.GetSerializer](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.design.serialization.idesignerserializationprovider.getserializer?view=net-7.0)
- [IServiceProvider.GetService](https://learn.microsoft.com/en-us/dotnet/api/system.iserviceprovider.getservice?view=net-7.0#system-iserviceprovider-getservice(system-type))

- GetSerializer should allow nulls for `currentSerializer` and `objectType`.
- GetService should allow null for `serviceType`.

elachlan: Alternatively we use the null-forgiving operator where changing nullability doesn't make sense. Which would be on `objectType` and `serviceType`, since the docs don't say they are nullable.

lonitra: I agree that we should allow null for `currentSerializer` since the doc you provided mentions it can be null. I'm seeing [here](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.design.serialization.designerserializationmanager.getserializer?view=windowsdesktop-7.0) that it actually calls out that we throw if `objectType` is null, but we don't seem to be doing that. For `serviceType` it doesn't make sense to me that we would allow null. Let's keep the nullability for those two as is and adjust tests if needed.

elachlan: The only reason I can think of to allow a null for `objectType` is when a serializer would return a default value for it. Which is what the tests are doing.

lonitra: Apologies, I mean keep the nullability as is (do not allow null) for the two parameters in `GetSerializer` and the parameter in `GetService` as well. 
For `GetService` it seems sensible to fix the compile warnings by just returning null if `serviceType` is null. 
For `GetSerializer` , we should throw if either of the parameters are null like the docs mentioned.

elachlan: I will need to update tests to facilitate the new behavior. I take it that using a null forgiving operator and passing it through is a no go?

lonitra: Investigating further I am seeing quite a bit of discrepancies between docs and what it going on in codebase so I am unsure if making these params nonnullable is the right thing to do.. I will bring this to someone in team who has more history/knowledge on this to give guidance.

lonitra: Thanks for bearing with me, I discussed with @KlausLoeffelmann and concluded that for `DesignerSerializationManager.GetSerializer()`, it looks like we have instances in our codebase where we are passing null for `objectType`, notably [here](https://github.com/dotnet/winforms/blob/00e6b62faf223b3cf2c2417b888de71c1648fbee/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomSerializerBase.cs#L2267-L2273) and we've done this since Framework, so `objectType` should be made nullable. The annotations for `IDesignerSerializationProvider.GetSerializer()`  need to be updated to reflect nullability for params `currentSerializer` and `objectType` as these parameters should allow null.
For `DesignerSerializationManager.GetService()`, not seeing any places where we pass null to this method aside from tests, could you verify if we do anywhere? If we don't, we can leave this parameter as not nullable.

elachlan: https://github.com/dotnet/runtime/pull/81203

elachlan:
@dreddy-work could you please trigger the pipeline again to see if we have pulled in the required runtime version?

dreddy-work:
> @dreddy-work could you please trigger the pipeline again to see if we have pulled in the required runtime version?

Sure. I see following build failure. Doesn't it repro locally?

src\System.Windows.Forms.Design\src\System\ComponentModel\Design\Serialization\DesignerSerializationManager.cs(511,78): error CS8604: (NETCORE_ENGINEERING_TELEMETRY=Build) Possible null reference argument for parameter 'currentSerializer' in 'object? IDesignerSerializationProvider.GetSerializer(IDesignerSerializationManager manager, object currentSerializer, Type objectType, Type serializerType)'.

elachlan:
The build error is caused by the fact https://github.com/dotnet/runtime/pull/81203 hadn't been merged to runtime until recently. So locally I do have the issue, but I'd have to rebase to test it. Which would reset the reviews.

dreddy-work:
> The build error is caused by the fact [dotnet/runtime#81203](https://github.com/dotnet/runtime/pull/81203) hadn't been merged to runtime until recently. So locally I do have the issue, but I'd have to rebase to test it. Which would reset the reviews.

If you rebase and build is resolved, i would force push rebase version if failure persists.

dreddy-work:
Error message


Stack trace
   at System.ComponentModel.Design.Serialization.CodeDomComponentSerializationService.CodeDomSerializationStore.Close() in /_/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs:line 323
   at System.ComponentModel.Design.Serialization.Tests.CodeDomComponentSerializationServiceTests.CreateStore_CloseSerializeWithValidProvider_Success(Object componentChangeService, Object designerOptionService) in /_/src/System.Windows.Forms.Design/tests/UnitTests/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationServiceTests.cs:line 327
   at InvokeStub_CodeDomComponentSerializationServiceTests.CreateStore_CloseSerializeWithValidProvider_Success(Object, Object, IntPtr*)
   at System.Reflection.MethodInvoker.Invoke(Object obj, IntPtr* args, BindingFlags invokeAttr)

elachlan:
I apologize for the rebase/force push. Test should be fixed now.

lonitra (approved):
Thank you for your persistence on this!

elachlan:
Thanks, I hope that the rest of the code base doesn't take me 2 months per code file.

elachlan:
I'll try and put in a pr to fix it soon.

dreddy-work:
> The same may apply to other hashtables such as namesByInstance.

Thank you @KalleOlaviNiemitalo for pointing this out. @elachlan, lets take a look at these on case by case.

PR #8362, opened 2022-12-10
Update to the latest CsWin32

JeremyKuhne:
This drop adds CCW generation, which removes a good chunk of our boilerplate and will facilitate additional progress on moving to ComWrappers.


AArnott (approved):

lonitra (approved):
🥳

PR #8364, opened 2022-12-11
Refactor BindingCollection to replace ArrayList with List<Binding>

Jericho:
Related: #8140

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8364)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/BindingsCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/BindingsCollection.cs
@@ -33,7 +33,7 @@ namespace System.Windows.Forms
         /// <summary>
         ///  Gets the <see cref="Binding"/> at the specified index.
         /// </summary>
-        public Binding this[int index] =>_list[index];
+        public Binding this[int index] => (_list ??= new())[index];
 
         protected internal void Add(Binding binding)
         {
@@ -49,8 +49,8 @@ namespace System.Windows.Forms
         protected virtual void AddCore(Binding dataBinding)
         {
             ArgumentNullException.ThrowIfNull(dataBinding);
-
-            _list?.Add(dataBinding);
+            _list ??= new();
+            _list.Add(dataBinding);
         }
 
         /// <summary>
--- a/src/System.Windows.Forms/src/System/Windows/Forms/BindingsCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/BindingsCollection.cs
@@ -15,7 +15,7 @@ namespace System.Windows.Forms
     [DefaultEvent(nameof(CollectionChanged))]
     public class BindingsCollection : BaseCollection
     {
-        private ArrayList _list;
+        private List<Binding> _list;
         private CollectionChangeEventHandler _onCollectionChanging;
         private CollectionChangeEventHandler _onCollectionChanged;
 
@@ -23,17 +23,17 @@ namespace System.Windows.Forms
         {
         }
 
-        public override int Count => _list is null ? 0 : base.Count;
+        public override int Count => _list?.Count ?? 0;
 
         /// <summary>
         ///  Gets the bindings in the collection as an object.
         /// </summary>
-        protected override ArrayList List => _list ??= new ArrayList();
+        protected override ArrayList List => ArrayList.Adapter(_list ??= new());
 
         /// <summary>
         ///  Gets the <see cref="Binding"/> at the specified index.
         /// </summary>
-        public Binding this[int index] => (Binding)List[index];
+        public Binding this[int index] =>_list[index];
 
         protected internal void Add(Binding binding)
         {
@@ -50,7 +50,7 @@ namespace System.Windows.Forms
         {
             ArgumentNullException.ThrowIfNull(dataBinding);
 
-            List.Add(dataBinding);
+            _list?.Add(dataBinding);
         }
 
         /// <summary>
@@ -84,7 +84,7 @@ namespace System.Windows.Forms
         /// <summary>
         ///  Clears the collection of any members.
         /// </summary>
-        protected virtual void ClearCore() => List.Clear();
+        protected virtual void ClearCore() => _list?.Clear();
 
         /// <summary>
         ///  Raises the <see cref="CollectionChanging"/> event.
@@ -115,7 +115,7 @@ namespace System.Windows.Forms
         /// <summary>
         ///  Removes the specified <see cref="Binding"/> from the collection.
         /// </summary>
-        protected virtual void RemoveCore(Binding dataBinding) => List.Remove(dataBinding);
+        protected virtual void RemoveCore(Binding dataBinding) => _list?.Remove(dataBinding);
 
         protected internal bool ShouldSerializeMyAll() => Count > 0;
     }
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/BindingsCollection.cs
```diff
@@ -15,25 +15,25 @@ namespace System.Windows.Forms
     [DefaultEvent(nameof(CollectionChanged))]
     public class BindingsCollection : BaseCollection
     {
-        private ArrayList _list;
+        private List<Binding> _list;
```
lonitra, line 18: Let's go ahead and enable nullability here and change this to `List<Binding>?`

lonitra (commented):
Just a comment about enabling nullability since we seem to be doing those checks in the file anyways.

src/System.Windows.Forms/src/System/Windows/Forms/BindingsCollection.cs
Jericho: sure.

src/System.Windows.Forms/src/System/Windows/Forms/BindingsCollection.cs
```diff
@@ -15,25 +13,25 @@ namespace System.Windows.Forms
     [DefaultEvent(nameof(CollectionChanged))]
     public class BindingsCollection : BaseCollection
     {
-        private ArrayList _list;
-        private CollectionChangeEventHandler _onCollectionChanging;
-        private CollectionChangeEventHandler _onCollectionChanged;
+        private List<Binding>? _list;
```
dreddy-work, line 16: This could be assigned as `private List<Binding> _list = new();` and remove nullable from it.

Jericho: The change you are proposing would allow removing a lot of null checking code (for example: `_list ??= new()`). However, it seems to me there was a deliberate decision to delay private members instantiation as much as possible throughout this project (presumably to ensure instantiation of the class in question is as quick as possible) so I'm reluctant to make this change here unless you tell me that's exactly what you want.

Here are a few examples taken at random from other classes where nullability has been enabled but private member instantiation has not been changed like you are suggesting: [here](https://github.com/dotnet/winforms/blob/079b548eea0799bd64f21ecf93771e7e317a5cb5/src/System.Windows.Forms/src/System/Windows/Forms/Application.ThreadContext.cs#L52), [here](https://github.com/dotnet/winforms/blob/079b548eea0799bd64f21ecf93771e7e317a5cb5/src/System.Windows.Forms/src/System/Windows/Forms/Form.EnumThreadWindowsCallback.cs#L15), [here](https://github.com/dotnet/winforms/blob/079b548eea0799bd64f21ecf93771e7e317a5cb5/src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs#L21), [here](https://github.com/dotnet/winforms/blob/079b548eea0799bd64f21ecf93771e7e317a5cb5/src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs#L172) and [here](https://github.com/dotnet/winforms/blob/079b548eea0799bd64f21ecf93771e7e317a5cb5/src/System.Windows.Forms/src/System/Windows/Forms/TabPage.cs#L31).

Playing devil's advocate, I was able to find a handful of examples where private members instantiation is not delayed: [here](https://github.com/dotnet/winforms/blob/079b548eea0799bd64f21ecf93771e7e317a5cb5/src/System.Windows.Forms/src/System/Windows/Forms/ImageList.cs#L49) and [here](https://github.com/dotnet/winforms/blob/079b548eea0799bd64f21ecf93771e7e317a5cb5/src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs#L138). 

The bottom line is that I can make the change you mentioned but let me know if that's what you want (in your comment, you said this `could` be done so I wasn't sure if that meant you wanted me to make the change or not).

dreddy-work: There is a difference between this initialization and some of the others you pointed out. In this case, any instance accessing list is either allocating it or returning `0` count. Same is not true for others ( ex : message filters, in which you do not allocate always when you access it and there are few code paths that made sense to check for null). 

>you said this could be done so I wasn't sure if that meant you wanted me to make the change or not

Given i was reviewing in the browser, wasn't sure if any compile-time/runtime errors prevented you from doing it. So wanted you to double check if it is safe to do so.

src/System.Windows.Forms/src/System/Windows/Forms/BindingsCollection.cs
```diff
@@ -15,25 +13,25 @@ namespace System.Windows.Forms
...
         }
 
-        public override int Count => _list is null ? 0 : base.Count;
+        public override int Count => _list.Count;
 
         /// <summary>
         ///  Gets the bindings in the collection as an object.
         /// </summary>
-        protected override ArrayList List => _list ??= new ArrayList();
+        protected override ArrayList List => ArrayList.Adapter(_list);
```
hughbe, line 29: is this breaking? Previously ArrayList and List would return the same instance, now a new instance is created each time

elachlan:
bumping this so it doesn't get closed.

elachlan:
@Jericho if its okay with you, I can possibly take over this?

Jericho:
@elachlan sure but all feedback was addressed and just waiting for merge.

elachlan:
I'll leave it to the team to finish the review.

elachlan:
@dreddy-work I think this might be waiting on the team. Could someone please review it and see what needs to be done?

src/System.Windows.Forms/src/System/Windows/Forms/BindingsCollection.cs
dreddy-work: @hughbe, It doesn't creates new list but just wraps around it.

dreddy-work:
> @dreddy-work I think this might be waiting on the team. Could someone please review it and see what needs to be done?

I looked into it again and i feel my comment above about initializing list still make sense in this scenario. If @Jericho / @elachlan push a change with respect to that, I will merge it. Rest all look good to me.

Jericho:
@dreddy-work your feedback was addressed on Dec 13 2022 in commit 8a7327d0feabc8c4d4089e8e194027cd528422e7

dreddy-work:
> @dreddy-work your feedback was addressed on Dec 13 2022 in commit [8a7327d](https://github.com/dotnet/winforms/commit/8a7327d0feabc8c4d4089e8e194027cd528422e7)

Thanks @Jericho, I saw comment was not resolved but didn't cross check the commit. Thank you.

dreddy-work (dismissed):

dreddy-work:
@Jericho , can you rebase this to get it merged?

dreddy-work:
src\System.Windows.Forms\src\System\Windows\Forms\BindingsCollection.cs(67,52): error RS0036: (NETCORE_ENGINEERING_TELEMETRY=Build) Symbol 'CollectionChanged' is missing nullability annotations in the declared API.

dreddy-work (approved):

PR #8367, opened 2022-12-12
Fix `Cursor(Stream)` to avoid incomplete reads and possible exception

### LINKED ISSUE #3492: Cursor(Stream) bugs
**.NET Core Version:**
master

**Have you experienced this same bug with .NET Framework?:**
probably

**Problem description:**
As noted [here in a PR](https://github.com/dotnet/winforms/pull/3476#discussion_r443216116) the constrcutor of `Cursor` taking a `Stream` is broken

```cs
public Cursor(Stream stream)
{
    if (stream == null)
    {
        throw new ArgumentNullException(nameof(stream));
    }

    int length = checked((int)stream.Length);
    _cursorData = new byte[length];
    stream.Read(_cursorData, 0, length);
    LoadPicture(
        new Ole32.GPStream(new MemoryStream(_cursorData)),
        nameof(stream));
}
```

- if the stream position isn't at zero the buffer would be allocated too large (suboptimal)
- if the stream doesn't know its length this would throw even though it doesn't have to (bug)
- `Read` is allowed to read less then `length` bytes so cursor data may be incomplete (bug)

Users can work around those bugs by wrapping their stream, but really the code should do a `stream.CopyTo(MemoryStream)` to let it do the heavy lifting and be more compatible to passing in arbitrary streams.

Note the current code doesn't throw, it just over-allocates and then reads less bytes than requested, which is fine if the cursor data is still represented completely at the current position.

**Expected behavior:**
- streams which don't expose their length can still be read
- streams which provide data in chunks still can be loaded

**Minimal repro:**
provide a custom stream

elachlan:
Fixes #3492

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8367)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Cursor.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Cursor.cs
@@ -81,12 +81,12 @@ namespace System.Windows.Forms
         public Cursor(Stream stream)
         {
             ArgumentNullException.ThrowIfNull(stream);
-
-            int length = checked((int)stream.Length);
-            _cursorData = new byte[length];
-            stream.Read(_cursorData, 0, length);
+            MemoryStream memoryStream = new();
+            stream.CopyTo(memoryStream);
+            memoryStream.Position = 0;
+            _cursorData = memoryStream.ToArray();
             LoadPicture(
-                new Ole32.GPStream(new MemoryStream(_cursorData)),
+                new Ole32.GPStream(memoryStream),
                 nameof(stream));
         }
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Cursor.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Cursor.cs
@@ -83,8 +83,10 @@ namespace System.Windows.Forms
             ArgumentNullException.ThrowIfNull(stream);
             MemoryStream memoryStream = new();
             stream.CopyTo(memoryStream);
-            memoryStream.Position = 0;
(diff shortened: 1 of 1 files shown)


elachlan:
CC: @weltkante

weltkante:
I don't really remember why I didn't make a PR back then, as the fix itself is trivial, maybe because it was unclear how much unit testing was wanted. As far as I can tell this looks good 👍

lonitra (approved):
LGTM

src/System.Windows.Forms/src/System/Windows/Forms/Cursor.cs
```diff
@@ -81,12 +81,12 @@ public Cursor(Type type, string resource)
         public Cursor(Stream stream)
         {
             ArgumentNullException.ThrowIfNull(stream);
-
-            int length = checked((int)stream.Length);
-            _cursorData = new byte[length];
-            stream.Read(_cursorData, 0, length);
+            MemoryStream memoryStream = new();
+            stream.CopyTo(memoryStream);
+            memoryStream.Position = 0;
```
dreddy-work, line 86: Is Position reset necessary here? `ToArray` doesn't care. Or was it for its use on line 89?

elachlan: stream.CopyTo advances all streams involved. I have updated the code for more clarity.

dreddy-work (approved):

Issue #3780, opened 2020-08-26
LinkLabel cut off


ahdung:
* .NET Version: 
3.1.301

* Have you experienced this same bug with .NET Framework?: 
Yes. .NET Framework 2.0-4.8

**Environment:**
- VS2019 16.6.3
- Windows Server 2008 R2 x64 Chinese lang

**Problem description:**

LinkLabel width not enough.
![image](https://user-images.githubusercontent.com/12094247/91275592-f6db8d00-e7b2-11ea-9614-829d32c61a5a.png)

**Expected behavior:**

like last one.

**Minimal repro:**

1. create a winform project
2. drag a `CheckBox` on `Form1`
3. add a new `Form` (auto name to `Form2`)
4. drag two `LinkLabel`s and a `Label` on `Form2`, make sure the z-order is(top to bottom): `linkLabel1`, `linkLabel2`, `label1`
5. set `linkLabel1` and `linkLabel2` font to `Arial, 10.5pt`
6. register the `Load` event for `Form2`, write codes:
    ```c#
    linkLabel2.Text = string.Empty;
    linkLabel1.Text = "<未选择>";
    linkLabel2.Text = "<未选择>";
    ```
7. modify `Main` method contents to follow codes in `program.cs` :
    ```c#
    Application.EnableVisualStyles();
    Application.SetCompatibleTextRenderingDefault(false);
    
    new Form1().ShowDialog();
    Application.Run(new Form2());
    ```
8. F5, close `Form1`, now you can see the `linkLabel1` width is not enough

**Demo.zip**
[LinkLabelWidthBugDemo.zip](https://github.com/dotnet/winforms/files/5128326/LinkLabelWidthBugDemo.zip)

**Reference:**
- [https://developercommunity.visualstudio.com/content/problem/1159811/linklabel-width-not-enough-in-particular-cases.html](https://developercommunity.visualstudio.com/content/problem/1159811/linklabel-width-not-enough-in-particular-cases.html)
- [https://stackoverflow.com/questions/63554856/linklabel-width-not-enough-in-particular-cases](https://stackoverflow.com/questions/63554856/linklabel-width-not-enough-in-particular-cases)

Zheng-Li01:
**More info**
Try to reproduce the issue using created project. get the below info.
1. The issue can't reproduce on English Lang.
2. Run the project calling the form2, can reproduce the issue; calling the form3, can't reproduce the issue, the difference is that no label controls on the Form3.
[...]

Issue #6229, opened 2021-11-25
The TextRenderer.MeasureText has different return between .Net FrameWork and .Net 6


Maspapa:
* .NET Core Version:  
.Net6

* Have you experienced this same bug with .NET Framework?: 
NO

**Problem description:**
The TextRenderer.MeasureText has different return between .Net FrameWork and .Net 6
`The Code below return different between .Net6 and .NetFrameWork

            Font font2 = new Font("Segoe UI", 9);
            Font font = new Font("Arial", 20F, FontStyle.Italic | FontStyle.Bold, GraphicsUnit.Point, ((byte)(128)));
            Size size = TextRenderer.MeasureText("Item77777", font, new Size(int.MaxValue, int.MaxValue), TextFormatFlags.NoPadding | TextFormatFlags.NoPrefix | TextFormatFlags.SingleLine);

            MessageBox.Show(size.Width.ToString());
`

**Expected behavior:**
In FrameWork it return 141
In Net6 it return 129

JeremyKuhne:
Sorry this took so long to get to. This is a side effect of fixing the `TextRenderer` to respect `TextFormatFlags.NoPadding`. On .NET Framework it would be ignored and `TextFormatFlags.GlyphOverhangPadding` would be applied. If you want to get the same behavior on both versions explicitly call out `TextFormatFlags.GlyphOverhangPadding`.

PR #8368, opened 2022-12-13
Fix `LinkLabel` truncate when programmatically disabled and `UseCompatibleTextRendering=false`

### LINKED ISSUE #7341: LinkLabels truncate when programmically disabled
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/t/LinkLabels-truncate-when-programmically-/10078592)._

---
[severity:It bothers me. A fix would be nice]
The link label truncates when it's Enabled property is set to false. Does not happen in the designer, just the program. I have set the backgound colour to highlight that it is within the control - the control itself remains the same size when disabled/enabled. Note the words "Artist" and "Album" are not complete. The controls work fine when Enabled = true;

---
#### Feedback Bot on 6/23/2022, 02:09 PM: 

(private comment, text removed)
#### Melon Wang [MSFT] on 6/24/2022, 09:33 AM: 

(private comment, text removed)
#### David Bailey on 6/24/2022, 09:55 AM: 

(private comment, text removed)

---
### Original Solutions
(no solutions)

elachlan:
Fixes #7341

Before:
image

After:
image

Original comment: https://github.com/dotnet/winforms/issues/7341#issuecomment-1345829400

> This comment pointed me in the right direction.
> https://github.com/dotnet/winforms/issues/3780#issuecomment-855463125
> 
> When setting `UseCompatibleTextRendering=false`, it causes the issue.
> 
> The most basic reproduction is:
> ```csharp
> using Form form = new();
> using LinkLabel linkLabel = new()
> {
>     Text = "linkLabel12345",
>     Enabled = false,
>     UseCompatibleTextRendering = false,
>     AutoSize = true
> };
> form.Controls.Add(linkLabel);
> form.ShowDialog();
> ```
> 
> The issue stems from how `OnPaint` functions when `Enabled` and `UseCompatibleTextRendering`. The key here is around `Graphics.ExcludeClip` vs `Graphics.IntersectClip` and how Margins are included in the `_textRegion` in `EnsureRun`.
> 
> When `Enabled` is `true`, a manual paint is done via `PaintLink` and then `Graphics.ExcludeClip(_textRegion)` is called to avoid rendering that part of the graphics.
> 
> When `Enabled` is `false`, `Graphics.IntersectClip(_textRegion)` is called and then depending on `UseCompatibleTextRendering` we call a different overload of `ControlPaint.DrawStringDisabled`. `UseCompatibleTextRendering=true` will result in a `graphics.DrawString` call, where as `UseCompatibleTextRendering=false` results in a `User32.DrawTextExW` call.
> 
> When `UseCompatibleTextRendering=true` it uses `Graphics.MeasureCharacterRanges` to set the `_textRegion`. When `UseCompatibleTextRendering=false`, it uses the `ClientRectWithPadding` and then the margins returned from `hfont.GetTextMargins()`. The margins returned are then **SUBTRACTED** from the width to create a new rectangle. This is the bug, they should be ADDED to the width to account for the extra space required for the text.
> 
> https://github.com/dotnet/winforms/blob/b0e5f5b7c4698d3b39424c22d09dbce01e0ce544/src/System.Windows.Forms/src/System/Windows/Forms/LinkLabel.cs#L704-L707
> 
> By changing the code to the following the behaviour was fixed in all scenarios.
> ```csharp
> Rectangle visualRectangle = new Rectangle(clientRectWithPadding.X + iLeftMargin,
>                                                                   clientRectWithPadding.Y,
>                                                                   textSize.Width + iRightMargin + iLeftMargin,
>                                                                   textSize.Height);
> ```
> Calling _textRegion.GetBounds(g); when UseCompatibleTextRendering=true gets Width = 81 Height = 16.
> 
> When UseCompatibleTextRendering=false
> With Changes:
> visualRectangle has Width = 91 Height = 15. Which is then changed to Width = 84 Height = 15 after the call to CalcTextRenderBounds which narrows it to the clientRectWithPadding.
> 
> Without Changes:
> visualRectangle has Width = 77 Height = 15.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8368)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/LinkLabel.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/LinkLabel.cs
@@ -703,7 +703,7 @@ namespace System.Windows.Forms
 
                         Rectangle visualRectangle = new Rectangle(clientRectWithPadding.X + iLeftMargin,
                                                                   clientRectWithPadding.Y,
-                                                                  textSize.Width - iRightMargin - iLeftMargin,
+                                                                  textSize.Width + iRightMargin + iLeftMargin,
                                                                   textSize.Height);
                         visualRectangle = CalcTextRenderBounds(visualRectangle /*textRect*/, clientRectWithPadding /*clientRect*/, RtlTranslateContent(TextAlign));
 


elachlan:
CC: @merriemcgaw

elachlan:
#6229 seems to be related and would explain the difference. If `TextRenderer.MeasureText` is returning a smaller size that .NET framework, then the code her would truncate.

dreddy-work:
> #6229 seems to be related and would explain the difference. If `TextRenderer.MeasureText` is returning a smaller size that .NET framework, then the code her would truncate.

LinkLabel issue is marked to reproduce even in .NET framework.  @elachlan you do not see this issue in .NET framework?

elachlan:
@dreddy-work No, it works in .NET Framework 4.8.

JeremyKuhne:
Thanks for looking into this @elachlan. I'm a little concerned that the root issue lies a bit further out as this code looks identical to Framework. Can you try stepping through the scenario in a Framework app to see why subtracting works there? If you haven't done this before you can walk through the Framework code by checking "Enable .NET Framework Source Stepping" in the debug options in VS.

Wondering if this is:

- A side effect of the new font caching
- A potential bug in MeasureText
- Different defaults in the Framework app
- Some other fallout of a layout change

In any case, we need to pin down why it is different for what looks like similar code.

elachlan:
@JeremyKuhne I think this PR doesn't fix the underlying cause. I think its actually in MeasureText based on the bug report at https://github.com/dotnet/winforms/issues/6229.

I'll look into that specifically now.

elachlan:
I'll close this an continue discussion in the issue.

PR #8358, opened 2022-12-10
Cleanup explicit `EmbeddedResource` items in `System.Windows.Forms.Design.csproj`

elachlan:
For `System.Windows.Forms.Design` this uses globbing to add the Embedded Resources instead of explicit items.

If this goes okay, I will look into the rest of the projects in the solution.

Related: #49

Reference:
https://stackoverflow.com/questions/46584499/how-to-add-a-glob-for-resx-files-for-new-sdk-csproj-file

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8358)

PR #8371, opened 2022-12-13
Cleanup explicit `EmbeddedResource` items from `System.Windows.Forms.csproj`

### LINKED ISSUE #49: Cleanup explicit EmbeddedResource items from System.Windows.Forms.csproj
Feedback from EricStj:

Replace all these explicit items with a convention.  Seems that you should be able to come up with a globbing convention for these and update the reading code where it doesn't align with the convention (or update items when the convention is insufficient).

elachlan:
- Glob EmbeddedResources in Project file
- Rename icon files to help with globbing
- delete redundant icons
- Add tests for EmbeddedResources

Fixes #49
Related #8358

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8371)


elachlan:
CC: @dreddy-work @Tanya-Solyanik

src/System.Windows.Forms/src/System.Windows.Forms.csproj
```diff
@@ -66,589 +66,50 @@
...
-      <LogicalName>System.Windows.Forms.BindingNavigator</LogicalName>
+    <EmbeddedResource Include="Resources\System\Windows\Forms\PropertyGrid\*.ico" 
+                      Exclude="Resources\System\Windows\Forms\PropertyGrid\PropertiesTab.ico;
+                               Resources\System\Windows\Forms\PropertyGrid\EventsTab.ico">
+      <LogicalName>System.Windows.Forms.$([System.String]::Copy('%(FileName)'))</LogicalName>
     </EmbeddedResource>
-    <EmbeddedResource Include="Resources\System\Windows\Forms\DataGridViewButtonColumn.ico">
-      <LogicalName>System.Windows.Forms.DataGridViewButtonColumn</LogicalName>
+    <EmbeddedResource Include="Resources\System\Windows\Forms\*.cur">
+      <LogicalName>System.Windows.Forms.$([System.String]::Copy('%(FileName)')).cur</LogicalName>
```
lonitra, line 83: Should .cur be removed from logical name?

dreddy-work: looks logical name was carrying .cur earlier too.

src/System.Windows.Forms/src/System.Windows.Forms.csproj
```diff
@@ -66,589 +66,50 @@
...
     </EmbeddedResource>
-    <EmbeddedResource Include="Resources\System\Windows\Forms\ActiveDocumentHost.ico">
-      <LogicalName>System.Windows.Forms.ActiveDocumentHost</LogicalName>
+    <EmbeddedResource Include="Resources\System\Windows\Forms\Printing\*.ico">
+      <LogicalName>System.Windows.Forms.$([System.String]::Copy('%(FileName)'))</LogicalName>
     </EmbeddedResource>
-    <EmbeddedResource Include="Resources\System\Windows\Forms\BindingNavigator.ico">
-      <LogicalName>System.Windows.Forms.BindingNavigator</LogicalName>
+    <EmbeddedResource Include="Resources\System\Windows\Forms\PropertyGrid\*.ico" 
+                      Exclude="Resources\System\Windows\Forms\PropertyGrid\PropertiesTab.ico;
```
dreddy-work, line 78: We should revisit usage on these. If they are internal only ( as name suggests), we should be able to rearrange them.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/EmbeddedResourceTests.cs
```diff
@@ -0,0 +1,236 @@
...
+
+using System.Drawing;
+using System.Reflection;
+using Xunit;
+
+namespace System.Windows.Forms.Tests
+{
+    public class EmbeddedResourceTests : IClassFixture<ThreadExceptionFixture>
+    {
+        // Get System.Windows.Forms assembly to verify that it contains all the icons that the code uses.
```
dreddy-work, line 13:
```suggestion
        // Get System.Windows.Forms assembly to verify that it contains all the resources that the code uses.
```

dreddy-work (approved):
LGTM

src/System.Windows.Forms/src/System.Windows.Forms.csproj
```diff
@@ -66,589 +66,50 @@
     </EmbeddedResource>
   </ItemGroup>
 
+  <!-- Embedded Resources are globbed based on their location. -->
```
Tanya-Solyanik, line 69: Globbing is based on the fact that location reflects the namespace name, consider adding something about this to this comment.

src/System.Windows.Forms/src/System.Windows.Forms.csproj
Tanya-Solyanik: Sorry, I missed this comment. Could you please clarify?

src/System.Windows.Forms/src/System.Windows.Forms.csproj
Tanya-Solyanik: Actually I don't think any comment adds clarity here.

src/System.Windows.Forms/src/System.Windows.Forms.csproj
dreddy-work: Logical names for these ( i.e Resources\System\Windows\Forms\PropertyGrid\PropertiesTab.ico) are not matching the conventional hence excluded here and added separately. I think if we could capture usage of these icons and if they happened to be internal only, we could align/rename them and then should be able to remove these exclusion list.

elachlan: I think I can make this better. Leave it with me.

src/System.Windows.Forms/src/System.Windows.Forms.csproj
```diff
@@ -66,589 +66,50 @@
...
-      <LogicalName>System.Windows.Forms.VScrollBar</LogicalName>
-    </EmbeddedResource>
-    <EmbeddedResource Include="Resources\System\Windows\Forms\wfc.ico">
-      <LogicalName>System.Windows.Forms.wfc</LogicalName>
+    <EmbeddedResource Include="Resources\System\Windows\Forms\ToolStrip\*.ico">
+      <LogicalName>System.Windows.Forms.$([System.String]::Copy('%(FileName)'))</LogicalName>
     </EmbeddedResource>
+  </ItemGroup>
+
+  <!-- Embedded Resources which are unable to be globbed. -->
```
Tanya-Solyanik, line 90: Timer.ico could be moved one level up in the directory tree and globbed with other icons.

I'm not sure that the comment is needed to explain the one off resource.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/EmbeddedResourceTests.cs
```diff
@@ -0,0 +1,236 @@
...
+        [InlineData("System.Windows.Forms.printPreview")]
+        [InlineData("System.Windows.Forms.save")]
+        [InlineData("System.Windows.Forms.lockControls")]
+        [InlineData("System.Windows.Forms.alignToGrid")]
+        [InlineData("System.Windows.Forms.bringToFront")]
+        [InlineData("System.Windows.Forms.sendToBack")]
+        [InlineData("System.Windows.Forms.blank")]
+        public void EmbeddedResource_ResourcesExist_Icon(string resourceName)
+        {
+            using Stream stream = assembly.GetManifestResourceStream(resourceName);
```
Tanya-Solyanik, line 200: All resource names start with the same prefix, so you could make this theory data more readable by omitting the common part and re-adding it here - 
```
assembly.GetManifestResourceStream($"System.Windows.Forms.{resourceName}");
```

src/System.Windows.Forms/src/System.Windows.Forms.csproj
elachlan: We can glob everything if Printing/PropertyGrid/ToolStrip icons were moved up one directory and I worked out a way to parse the path into a namespace.

elachlan: For future reference: https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-well-known-item-metadata?view=vs-2022

src/System.Windows.Forms/src/System.Windows.Forms.csproj
```diff
@@ -66,589 +66,35 @@
     </EmbeddedResource>
   </ItemGroup>
 
+  <!-- Embedded Resources are globbed based on their location. -->
   <ItemGroup>
-    <EmbeddedResource Include="Resources\System\Windows\Forms\Animation.ico">
-      <LogicalName>System.Windows.Forms.Animation</LogicalName>
+    <EmbeddedResource Include="Resources\System\Windows\Forms\*.ico;
```
Tanya-Solyanik, line 71: Would something like this work:
```  
 <EmbeddedResource Include="Resources\**\*.ico>
      <LogicalName>System.Windows.Forms.%(FileName)</LogicalName>
```

Tanya-Solyanik: Why do you need the copy command?

Tanya-Solyanik: `**` does not work because of the `Design` namespace. But the copy command is not needed.

dreddy-work: Same goes to tests.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/EmbeddedResourceTests.cs
```diff
@@ -0,0 +1,236 @@
...
+            Assert.NotNull(stream);
+            using Icon icon = new(stream);
+            Assert.NotNull(icon);
+        }
```
Tanya-Solyanik, line 204: Could you please add test for ComponentModel icons?

Tanya-Solyanik: Resources folder contains 189 icons, this test covers 181. the remainder is in 
ComponentModel - 1
Server - 1
Picturebox - 1
2 removed - arrow and ToolStripContainer
Where is the remainder?

How do we proof that it's safe to remove duplicates?

elachlan: They exist, I have sorted the InlineData to make the tests easier to find.

elachlan: Those have test:
- ComponentModel.OrderImages
- Server.Arrow
- PictureBox.Loading

The two files deleted:
- src/System.Windows.Forms/src/Resources/System/Windows/Forms/PropertyGrid/arrow.ico wasn't embedded and was duplicate of Server.Arrow.
- src/System.Windows.Forms/src/Resources/System/Windows/Forms/ToolStripContainer.ico has been added back in when I moved the file from ToolStrip folder

Tanya-Solyanik: trustmanager icons are no longer included in the assembly and can be deleted.
 sm_TrustManagerGlobe.ico
TrustManagerGlobe.ico
TrustManagerHighRisk.ico
TrustManagerOK.ico
 TrustManagerWarning.ico

Tanya-Solyanik: In fact if you don't delete them, they will be included by mistake.

elachlan: Thanks for tracking those down.

elachlan:
@Tanya-Solyanik are you happy with it all now?

Tanya-Solyanik:
>@Tanya-Solyanik are you happy with it all now?

@elachlan - I'll play with this change a bit tonight

src/System.Windows.Forms/src/System.Windows.Forms.csproj
```diff
@@ -66,589 +66,17 @@
     </EmbeddedResource>
   </ItemGroup>
 
+  <!-- Embedded Resources are globbed based on their location. -->
   <ItemGroup>
-    <EmbeddedResource Include="Resources\System\Windows\Forms\Animation.ico">
-      <LogicalName>System.Windows.Forms.Animation</LogicalName>
+    <EmbeddedResource Include="Resources\System\Windows\Forms\**\*.ico;">
+      <LogicalName>$([System.String]::Copy('%(RelativeDir)').Replace("Resources\","").Replace("\",".").TrimEnd('.')).$([System.String]::Copy('%(FileName)'))</LogicalName>
```
Tanya-Solyanik, line 72: I think the copy operation is not needed here  - `$([System.String]::Copy('%(FileName)'))` - because file name is not modified, like the directory name is.
And the same applies to the cursor files.

elachlan: I have pushed it through. Tests are passing.

elachlan: Also pushed through the string.copy removal for the `System.Windows.Forms.Design.csproj` as well.

elachlan: This doesn't work for Design because the folder structure doesn't reflect the namespace. I'll do a separate PR for that as it requires renaming lots of files.

```xml
<EmbeddedResource Include="Resources\System\Windows\Forms\**\*.ico;">
      <LogicalName>$([System.String]::Copy('%(RelativeDir)').Replace("Resources\","").Replace("\",".").TrimEnd('.')).%(FileName)</LogicalName>
    </EmbeddedResource>
```

Tanya-Solyanik: I'm not sure if saving a couple of lines in csproj is worth the renaming. Would we get a cleaner folder structure in .Design project?

Tanya-Solyanik:
I had [dumped out resource names](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.getmanifestresourcenames?view=net-7.0#system-reflection-assembly-getmanifestresourcenames) from the main version of System.Windows.Forms.dll and your PR version, and found one new resource in your version - `System.Windows.Forms.ImageEditor`.

Tanya-Solyanik:
Please delete ImageEditor.ico and this PR is good to go!

src/System.Windows.Forms/src/System.Windows.Forms.csproj
elachlan: `Resources\System\Winforms\Design` => `Resources\System\Windows\Forms\Design`. Its not a big change, but it means that all resources will be automatically added when places in `Resources\System\`.

Only problem with the whole setup is we aren't doing negative testing.

Tanya-Solyanik (approved):
Thank you!

PR #8389, opened 2022-12-15
Replace Workspaces nuget packages with compiler only packages

### LINKED ISSUE #8387: Analyzer projects shouldn't have a dependency on Workspaces
They shouldn't need it, and it's problematic if types from Workspaces are used in these projects.

elachlan:
Fixes #8387 - Replace Workspaces nuget packages with compiler only packages

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8389)


elachlan:
CC: @Youssef1313

elachlan:
We should probably move to the compilers nuget package. But I figure the team might want the flexibility to update each on individually.

Youssef1313 (approved):
LGTM

dreddy-work (approved):

dreddy-work:
> We should probably move to the compilers nuget package. But I figure the team might want the flexibility to update each one individually.

I am not sure how consistent they are with versioning. For now, we can keep it this way.

elachlan:
@dreddy-work sounds good. The tests failures here are flaky.

PR #8390, opened 2022-12-15
Enable Nullability in multiple `Design` classes

elachlan:
Enable Nullability in multiple Design classes

Related: #8342

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8390)

### DIFF
--- a/src/System.Windows.Forms.Design/src/PublicAPI.Shipped.txt
+++ b/src/System.Windows.Forms.Design/src/PublicAPI.Shipped.txt
@@ -71,10 +71,10 @@ override System.Drawing.Design.ColorEditor.PaintValue(System.Drawing.Design.Pain
 ~override System.Drawing.Design.FontEditor.GetEditStyle(System.ComponentModel.ITypeDescriptorContext context) -> System.Drawing.Design.UITypeEditorEditStyle
 ~override System.Drawing.Design.FontNameEditor.GetPaintValueSupported(System.ComponentModel.ITypeDescriptorContext context) -> bool
 ~override System.Drawing.Design.FontNameEditor.PaintValue(System.Drawing.Design.PaintValueEventArgs e) -> void
-~override System.Drawing.Design.IconEditor.EditValue(System.ComponentModel.ITypeDescriptorContext context, System.IServiceProvider provider, object value) -> object
-~override System.Drawing.Design.IconEditor.GetEditStyle(System.ComponentModel.ITypeDescriptorContext context) -> System.Drawing.Design.UITypeEditorEditStyle
-~override System.Drawing.Design.IconEditor.GetPaintValueSupported(System.ComponentModel.ITypeDescriptorContext context) -> bool
-~override System.Drawing.Design.IconEditor.PaintValue(System.Drawing.Design.PaintValueEventArgs e) -> void
+override System.Drawing.Design.IconEditor.EditValue(System.ComponentModel.ITypeDescriptorContext? context, System.IServiceProvider! provider, object? value) -> object?
+override System.Drawing.Design.IconEditor.GetEditStyle(System.ComponentModel.ITypeDescriptorContext? context) -> System.Drawing.Design.UITypeEditorEditStyle
+override System.Drawing.Design.IconEditor.GetPaintValueSupported(System.ComponentModel.ITypeDescriptorContext? context) -> bool
+override System.Drawing.Design.IconEditor.PaintValue(System.Drawing.Design.PaintValueEventArgs! e) -> void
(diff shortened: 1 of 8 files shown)


src/System.Windows.Forms.Design/src/PublicAPI.Shipped.txt
```diff
@@ -659,10 +659,10 @@ System.Windows.Forms.Design.Behavior.SnapLine.SnapLine(System.Windows.Forms.Desi
...
 ~virtual System.Drawing.Design.IconEditor.GetFileDialogDescription() -> string
 ~virtual System.Drawing.Design.IconEditor.LoadFromStream(System.IO.Stream stream) -> System.Drawing.Icon
-~virtual System.Drawing.Design.ImageEditor.GetExtensions() -> string[]
-~virtual System.Drawing.Design.ImageEditor.GetFileDialogDescription() -> string
-~virtual System.Drawing.Design.ImageEditor.GetImageExtenders() -> System.Type[]
-~virtual System.Drawing.Design.ImageEditor.LoadFromStream(System.IO.Stream stream) -> System.Drawing.Image
+virtual System.Drawing.Design.ImageEditor.GetExtensions() -> string![]!
+virtual System.Drawing.Design.ImageEditor.GetFileDialogDescription() -> string!
+virtual System.Drawing.Design.ImageEditor.GetImageExtenders() -> System.Type![]!
+virtual System.Drawing.Design.ImageEditor.LoadFromStream(System.IO.Stream! stream) -> System.Drawing.Image!
```
elachlan, line 665: `LoadFromStream` does a null check `ArgumentNullException.ThrowIfNull(stream);`. I am unsure what to do with the annotation here?

elachlan: Remembered I got told about this before:
> They stay annotated as not-null and we leave the null check in place. Being annotated as not-null doesn't change behavior at run-time, so the null check is there to validate null isn't being passed in, just as before.

https://github.com/dotnet/runtime/pull/79429#discussion_r1044033195

elachlan:
@lonitra I think the test failures are unrelated?

src/System.Windows.Forms.Design/src/System/Drawing/Design/IconEditor.cs
```diff
@@ -48,12 +46,12 @@ protected static string CreateExtensionsString(string[] extensions, string sep)
         protected static string CreateFilterEntry(IconEditor editor)
         {
             string description = editor.GetFileDialogDescription();
-            string extensions = CreateExtensionsString(editor.GetExtensions(), ",");
-            string extensionsWithSemicolons = CreateExtensionsString(editor.GetExtensions(), ";");
+            string? extensions = CreateExtensionsString(editor.GetExtensions(), ",");
```
lonitra, line 49: I think there is some inconsistency here. The first parameter in `CreateExtensionsString()` is nullable, but it looks like every time `CreateExtensionsString()` is called we are passing in the return value of `GetExtensions()` which is currently nonnullable. Could you verify whether or not `GetExtensions()` ever returns a `null` value and adjust accordingly?

lonitra (changes requested):

src/System.Windows.Forms.Design/src/System/Drawing/Design/IconEditor.cs
elachlan: `GetExtensions` always returns a non-null value. `CreateExtensionsString` is `protected` so it is externally accessible. So a null value is possible. 

I've removed the nullability annotations on `CreateExtensionsString` parameter `extensions`. I think its best to let the user know that we expect a value. Since if you pass null, we return null.

elachlan: I think I might have been wrong here. In a stricter sense, if we throw a `ArguementNullException` or `ArguementException`, then it would make more sense to annotate that we don't want nulls. But because we handle it in code, I think we should have the nullability.

I think the idea is that if a null input is going to cause an error or invalid state, then we annotate it not null. This isn't the case here.

dreddy-work: Correct, while we like to make it explicit with respect to nullability, in this case, we are handling null values and it's been exposed publicly. We should mark it nullable for compatibility even when we internally never pass null. Our documentation is also doesn't enforce this.

dreddy-work (approved):

elachlan:
I rebased due to a clash. Apologies for that. The last commit is really the only change.

dreddy-work (approved):

PR #8393, opened 2022-12-15
added validation of the resource list for winforms and .Design assemblies

Tanya-Solyanik:
Ensure that we do not add any unexpected resources.

Removed test that validated presence of manifest because it's superseded by the test that matches available resources against the expected list.
Expected lists are dumped out from assemblies before the change that introduced regular expressions to the project files.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8393)


elachlan:
LGTM! Only suggestion would be if we can maintain the lists of icons, cursors, and manifest separately. Then use them in the other tests as well as this new test. That way we reduce the duplicated data.

Tanya-Solyanik:
>Then use them in the other tests as well as this new test. 

Thank you, good idea! This list is easier to dump and update than to match inline data to file names! This is how I tested your fix. I was going add the same to .Design assembly, where the list is generated off version before your change.

elachlan:
LGTM! This is great!

lonitra (approved):
🚀

Tanya-Solyanik:
Thank you @elachlan @lonitra!

PR #8394, opened 2022-12-15
Enable Nullability on a couple of Interfaces

elachlan:
Enable Nullability on a couple of Interfaces

Related: #8342

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8394)

### DIFF
--- a/src/System.Windows.Forms.Design/src/PublicAPI.Shipped.txt
+++ b/src/System.Windows.Forms.Design/src/PublicAPI.Shipped.txt
@@ -459,9 +459,9 @@ System.ComponentModel.Design.MenuCommandsChangedEventArgs.MenuCommandsChangedEve
 ~System.Drawing.Design.ToolboxItemCollection.this[int index].get -> System.Drawing.Design.ToolboxItem
 ~System.Drawing.Design.ToolboxItemCollection.ToolboxItemCollection(System.Drawing.Design.ToolboxItem[] value) -> void
 ~System.Drawing.Design.ToolboxItemCollection.ToolboxItemCollection(System.Drawing.Design.ToolboxItemCollection value) -> void
-~System.Resources.Tools.ITargetAwareCodeDomProvider.SupportsProperty(System.Type type, string propertyName, bool isWritable) -> bool
-~System.Runtime.InteropServices.UCOMITypeLib.GetDocumentation(int index, out string strName, out string strDocString, out int dwHelpContext, out string strHelpFile) -> void
-~System.Runtime.InteropServices.UCOMITypeLib.IsName(string szNameBuf, int lHashVal) -> bool
+System.Resources.Tools.ITargetAwareCodeDomProvider.SupportsProperty(System.Type! type, string! propertyName, bool isWritable) -> bool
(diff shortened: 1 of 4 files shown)


src/System.Windows.Forms.Design/src/PublicAPI.Shipped.txt
```diff
@@ -459,9 +459,9 @@ System.ComponentModel.Design.MenuCommandsChangedEventArgs.MenuCommandsChangedEve
 ~System.Drawing.Design.ToolboxItemCollection.this[int index].get -> System.Drawing.Design.ToolboxItem
 ~System.Drawing.Design.ToolboxItemCollection.ToolboxItemCollection(System.Drawing.Design.ToolboxItem[] value) -> void
 ~System.Drawing.Design.ToolboxItemCollection.ToolboxItemCollection(System.Drawing.Design.ToolboxItemCollection value) -> void
-~System.Resources.Tools.ITargetAwareCodeDomProvider.SupportsProperty(System.Type type, string propertyName, bool isWritable) -> bool
-~System.Runtime.InteropServices.UCOMITypeLib.GetDocumentation(int index, out string strName, out string strDocString, out int dwHelpContext, out string strHelpFile) -> void
-~System.Runtime.InteropServices.UCOMITypeLib.IsName(string szNameBuf, int lHashVal) -> bool
+System.Resources.Tools.ITargetAwareCodeDomProvider.SupportsProperty(System.Type! type, string! propertyName, bool isWritable) -> bool
+System.Runtime.InteropServices.UCOMITypeLib.GetDocumentation(int index, out string! strName, out string! strDocString, out int dwHelpContext, out string! strHelpFile) -> void
+System.Runtime.InteropServices.UCOMITypeLib.IsName(string! szNameBuf, int lHashVal) -> bool
```
lonitra, line 464: Curious, how did you verify that these should be nonnull?

lonitra (commented):
Just a question, but otherwise LGTM

src/System.Windows.Forms.Design/src/PublicAPI.Shipped.txt
elachlan: Good catch, I checked the first link. But neglected to check the second. It looks like it can be null.

- https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.ucomitypelib.getdocumentation?view=windowsdesktop-7.0#system-runtime-interopservices-ucomitypelib-getdocumentation(system-int32-system-string@-system-string@-system-int32@-system-string@)
- https://learn.microsoft.com/en-us/windows/win32/api/oaidl/nf-oaidl-itypelib-getdocumentation

elachlan: For the ITypeLib it should in-fact allow null. So I am wrong here.

elachlan: `If the caller does not need the item name, then pBstrName can be null.` So this would be true if we were using it as a standard interop. We could pass null if we didn't want to get the strName. But since this interface has out parameters we aren't providing a pointer to a string. So it will always return the value.



Does that make sense?

lonitra (approved):
🚀

PR #8398, opened 2022-12-17
Null annotate AxContainer

JeremyKuhne:
Null annotate AxContainer and clean up some related AxHost/Control bits.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8398)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/AxHost.AxContainer.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/AxHost.AxContainer.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.ComponentModel;
 using System.ComponentModel.Design;
 using System.Diagnostics;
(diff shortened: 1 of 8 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/AxHost.AxContainer.cs
```diff
@@ -14,1287 +12,1228 @@
...
+
+                bool selected = dwWhich.HasFlag(GC_WCH.FSELECTED);
+                bool reverse = dwWhich.HasFlag(GC_WCH.FREVERSEDIR);
+
+                // Note that Visual Basic actually ignores the next/prev flags. We will not.
+                bool onlyNext = dwWhich.HasFlag(GC_WCH.FONLYNEXT);
+                bool onlyPrevious = dwWhich.HasFlag(GC_WCH.FONLYPREV);
+
+                dwWhich &= ~(GC_WCH.FSELECTED | GC_WCH.FREVERSEDIR | GC_WCH.FONLYNEXT | GC_WCH.FONLYPREV);
+                if (onlyNext && onlyPrevious)
```
Tanya-Solyanik, line 214: This test can be moved up, right after these variables are set.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.AxContainer.cs
```diff
@@ -14,1287 +12,1228 @@
...
-                    // Note that visual basic actually ignores the next/prev flags... we will not
-                    bool onlyNext = (dwWhich & GC_WCH.FONLYNEXT) != 0;
-                    bool onlyPrev = (dwWhich & GC_WCH.FONLYPREV) != 0;
-                    dwWhich &= ~(GC_WCH.FSELECTED | GC_WCH.FREVERSEDIR |
-                                          GC_WCH.FONLYNEXT | GC_WCH.FONLYPREV);
-                    if (onlyNext && onlyPrev)
-                    {
-                        Debug.Fail("onlyNext && onlyPrev are both set!");
+                    default:
+                        Debug.Fail("Bad GC_WCH");
```
Tanya-Solyanik, line 232: Why not add this validation to line 220? Or at least move it to validation "section" explicitly.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.AxContainer.cs
```diff
@@ -14,1287 +12,1228 @@
...
+            }
+        }
 
-                if (ctl is AxHost hostctl && (dwOleContF & OLECONTF.EMBEDDINGS) != 0)
-                {
-                    list.Add(hostctl.GetOcx());
-                }
-                else if ((dwOleContF & OLECONTF.OTHERS) != 0)
+        [MemberNotNull(nameof(_components))]
+        private void FillComponentsTable(IContainer? container)
```
Tanya-Solyanik, line 349: This method can be renamed to match the field type (PopulateControlSet), or to reflect business logic. Note that this method is referred to on line 366, update both places if you take this suggestion.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.AxContainer.cs
```diff
@@ -14,1287 +12,1228 @@
...
+            return _components!;
+        }
 
-            private bool GetControlBelongs(Control ctl)
-            {
-                HashSet<Control> comps = GetComponents();
-                return comps.Contains(ctl);
-            }
+        private IContainer? GetParentIsDesigned()
+            => _parent.Site is { } site && site.DesignMode ? site.Container : null;
```
Tanya-Solyanik, line 416: We could move to the new recursive version of `DesignMode` property.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.AxContainer.cs
```diff
@@ -14,1287 +12,1228 @@
...
+                foreach (Control child in control.Controls)
                 {
-                    GetAllChildren(c);
+                    GetAllChildren(child);
                 }
             }
+        }
 
-            private HashSet<Control> GetComponents()
+        private HashSet<Control> GetComponents()
```
Tanya-Solyanik, line 403: Should this method be renamed to `GetControls`?

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.AxContainer.cs
```diff
@@ -14,1287 +12,1228 @@
...
+            s_axHTraceSwitch.TraceVerbose("in IOleInPlaceFrame.TranslateAccelerator");
+            return HRESULT.S_FALSE;
+        }
 
-            HRESULT IOleInPlaceFrame.SetMenu(IntPtr hmenuShared, IntPtr holemenu, IntPtr hwndActiveObject)
-            {
-                s_axHTraceSwitch.TraceVerbose("in SetMenu");
-                return HRESULT.E_NOTIMPL;
-            }
+        private class ExtenderProxy :
```
Tanya-Solyanik, line 799: Please more to a new file.

Tanya-Solyanik (dismissed):
Looks good, I added some suggestions.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.AxContainer.cs
```diff
@@ -14,1287 +12,1228 @@
...
+                        _associatedContainer = site.Container;
+                        if (site.TryGetService(out IComponentChangeService? changeService))
                         {
-                            _associatedContainer = site.Container;
-                            IComponentChangeService ccs = (IComponentChangeService)site.GetService(typeof(IComponentChangeService));
-                            if (ccs is not null)
-                            {
-                                ccs.ComponentRemoved += new ComponentEventHandler(OnComponentRemoved);
-                            }
+                            changeService.ComponentRemoved += OnComponentRemoved;
```
lonitra, line 170: Should there be a null check for `changeService`?

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.AxContainer.cs
```diff
@@ -14,1287 +12,1228 @@
...
+            {
+                get
                 {
-                    get
-                    {
-                        s_axHTraceSwitch.TraceVerbose("In UnderlyingSystemType");
-                        return null;
-                    }
+                    s_axHTraceSwitch.TraceVerbose("In UnderlyingSystemType");
+                    return null!;
```
lonitra, line 1236: Why don't we want to change return type to `Type?` here?

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
```diff
@@ -2102,55 +2094,42 @@ public ContainerControl ContainingControl
...
         {
-            bool ret = false;
             try
             {
-                ret = (Text.Length != 0);
+                return Text.Length != 0;
             }
             catch (COMException)
             {
             }
```
lonitra, line 2103: Why does this need to be in a try-catch?

JeremyKuhne: Because `Length` is virtual. I'd rather not have this here, but it is technically possible someone calls an OCX property in an override.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.AxContainer.cs
JeremyKuhne: Explicitly didn't in this change to have a diff.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.AxContainer.cs
Tanya-Solyanik: IReflect belongs to System.Runtime

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.AxContainer.cs
JeremyKuhne: No, `TryGetService` does that and establishes that it isn't null if it returns true. The out has to be nullable for the false case.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.AxContainer.cs
JeremyKuhne: Because I was trying to limit the logic changes to things that de-nested. This change is pretty big- and not likely to be the last. An additional pass will be easier to follow.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.AxContainer.cs
JeremyKuhne: Would be a good follow up

JeremyKuhne:
I'll do a follow up to pull the nested class out.

Tanya-Solyanik (approved):

PR #8400, opened 2022-12-18
Refactor `TabOrder` to enable Nullability and replace `ArrayList`/`HashTable`

elachlan:
There are several places where I had to do some unusual null checks where I think the code previously just assumed the they were not null. Such as any of the collections (`tabComplete`, ect), or `Control.Parent`.

Related: #8342, #8143, #8140

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8400)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/TabOrder.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/TabOrder.cs
@@ -758,21 +758,13 @@ namespace System.Windows.Forms.Design
         {
             if (ctlHover != ctl)
             {
-                if (ctlHover is not null)
-                {
-                    if (region is not null)
-                    {
-                        region.Dispose();
-                        region = null;
-                    }
-
-                    Rectangle rc = GetConvertedBounds(ctlHover);
-                    rc.Inflate(selSize, selSize);
-                    Invalidate(rc);
-                }
-
+                InvalidateHoverRegion();
                 ctlHover = ctl;
+                InvalidateHoverRegion();
+            }
 
+            void InvalidateHoverRegion()
+            {
                 if (ctlHover is not null)
                 {
                     if (region is not null)
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/TabOrder.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/TabOrder.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.Collections;
 using System.ComponentModel;
 using System.ComponentModel.Design;
@@ -22,19 +20,19 @@ namespace System.Windows.Forms.Design
     [ToolboxItem(false)]
     internal class TabOrder : Control, IMouseHandler, IMenuStatusHandler
     {
-        private IDesignerHost host;
-        private Control ctlHover;
-        private ArrayList tabControls;
-        private Rectangle[] tabGlyphs;
-        private ArrayList tabComplete;
-        private Hashtable tabNext;
+        private IDesignerHost? host;
+        private Control? ctlHover;
+        private List<Control>? tabControls;
+        private Rectangle[]? tabGlyphs;
+        private HashSet<Control>? tabComplete;
+        private Dictionary<Control, int>? tabNext;
         private readonly Font tabFont;
         private readonly StringBuilder drawString;
         private readonly Brush highlightTextBrush;
         private readonly Pen highlightPen;
         private readonly int selSize;
-        private readonly Hashtable tabProperties;
-        private Region region;
+        private readonly Dictionary<Control, PropertyDescriptor> tabProperties;
+        private Region? region;
         private readonly MenuCommand[] commands;
         private readonly MenuCommand[] newCommands;
         private readonly string decimalSep;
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/TabOrder.cs
```diff
@@ -22,19 +20,19 @@ namespace System.Windows.Forms.Design
     [ToolboxItem(false)]
     internal class TabOrder : Control, IMouseHandler, IMenuStatusHandler
     {
-        private IDesignerHost host;
```
elachlan, line 25: `host` was made nullable because of the `Dispose`, do I assign `null!` to `host` there instead?

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/TabOrder.cs
```diff
@@ -294,18 +265,19 @@ private void DrawTabs(IList tabs, Graphics gr, bool fRegion)
...
                 Color textColor;
                 if (fRegion)
                 {
-                    region.Union(rc);
+                    region?.Union(rc);
                 }
                 else
                 {
-                    if (tabComplete.IndexOf(ctl) != -1)
+                    if (tabComplete is not null && !tabComplete.Contains(ctl))
```
halgab, line 280: I think you reversed the condition here
```suggestion
                    if (tabComplete is not null && tabComplete.Contains(ctl))
```

elachlan: Good Catch!

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/TabOrder.cs
```diff
@@ -245,31 +211,36 @@ private void DrawTabs(IList tabs, Graphics gr, bool fRegion)
...
-
-                    Region clip = gr.Clip;
-                    gr.ExcludeClip(ctlInner);
-                    using (SolidBrush brush = new SolidBrush(backColor))
+                    if (ctlHover.Parent is Control hoverParent)
                     {
-                        gr.FillRectangle(brush, ctlOuter);
-                    }
+                        Color backColor;
+                        backColor = hoverParent.BackColor;
```
lonitra, line 217:
```suggestion
                        Color backColor = hoverParent.BackColor;
```

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/TabOrder.cs
```diff
@@ -245,31 +211,36 @@ private void DrawTabs(IList tabs, Graphics gr, bool fRegion)
...
+                        Region clip = gr.Clip;
+                        gr.ExcludeClip(ctlInner);
+                        using (SolidBrush brush = new SolidBrush(backColor))
+                        {
+                            gr.FillRectangle(brush, ctlOuter);
+                        }
 
-                    ControlPaint.DrawSelectionFrame(gr, false, ctlOuter, ctlInner, backColor);
-                    gr.Clip = clip;
+                        ControlPaint.DrawSelectionFrame(gr, false, ctlOuter, ctlInner, backColor);
```
lonitra, line 226: nit: param name

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/TabOrder.cs
```diff
@@ -245,31 +211,36 @@ private void DrawTabs(IList tabs, Graphics gr, bool fRegion)
...
+                        gr.Clip = clip;
+                    }
                 }
             }
 
+            int iCtl = 0;
+            string str;
             Rectangle rc;
-            while (e.MoveNext())
+            Control? parent;
```
lonitra, line 235: Could these be moved and assigned where appropriate inside the loop? I don't see it being used anywhere else.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/TabOrder.cs
```diff
@@ -329,9 +301,10 @@ private void DrawTabs(IList tabs, Graphics gr, bool fRegion)
 
             if (fRegion)
             {
-                ctl = (Control)host.RootComponent;
-                rc = GetConvertedBounds(ctl);
-                region.Intersect(rc);
+                Debug.Assert(host is not null, "host is set in the contructor should not be null");
+                Control rootControl = (Control)host.RootComponent;
+                rc = GetConvertedBounds(rootControl);
+                region?.Intersect(rc);
```
lonitra, line 307: This can be changed to the following to accomodate for my above comment about moving those variables inside the loop
```suggestion
                region?.Intersect(GetConvertedBounds(rootControl));
```

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/TabOrder.cs
```diff
@@ -22,19 +20,19 @@ namespace System.Windows.Forms.Design
...
         private readonly Brush highlightTextBrush;
         private readonly Pen highlightPen;
         private readonly int selSize;
-        private readonly Hashtable tabProperties;
-        private Region region;
+        private readonly Dictionary<Control, PropertyDescriptor> tabProperties;
+        private Region? region;
         private readonly MenuCommand[] commands;
         private readonly MenuCommand[] newCommands;
         private readonly string decimalSep;
```
lonitra, line 38: Let's rename these fields to use appropriate naming style for fields e.g. `_host`,`_region`,etc. Preferrably in separate commit for reviewability.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/TabOrder.cs
```diff
@@ -245,31 +211,36 @@ private void DrawTabs(IList tabs, Graphics gr, bool fRegion)
...
+                    }
                 }
             }
 
+            int iCtl = 0;
+            string str;
             Rectangle rc;
-            while (e.MoveNext())
+            Control? parent;
+            foreach (Control ctl in tabs)
```
lonitra, line 236: nit: we are avoiding using abbreviated names.
```suggestion
            foreach (Control control in tabs)
```

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/TabOrder.cs
```diff
@@ -245,31 +211,36 @@ private void DrawTabs(IList tabs, Graphics gr, bool fRegion)
...
             {
-                ctl = (Control)e.Current;
                 rc = GetConvertedBounds(ctl);
```
lonitra, line 238:
```suggestion
                Rectangle convertedRectangle = GetConvertedBounds(ctl);
```

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/TabOrder.cs
```diff
@@ -340,29 +313,30 @@ private void DrawTabs(IList tabs, Graphics gr, bool fRegion)
         ///  returns a control in the given tab vector that is at the given point, in
         ///  screen coords.
         /// </summary>
-        private Control GetControlAtPoint(IList tabs, int x, int y)
+        private Control? GetControlAtPoint(List<Control> tabs, int x, int y)
         {
             IEnumerator e = tabs.GetEnumerator();
```
lonitra, line 318: I don't think this is being used anymore, we can remove this

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/TabOrder.cs
```diff
@@ -340,29 +313,30 @@ private void DrawTabs(IList tabs, Graphics gr, bool fRegion)
...
+                // We do not break if we find it here. The vector is already setup
                 // to have all controls in the current tabbing order, and child controls
-                // are always after their parents.  If we broke, we wouldn't necessarily
+                // are always after their parents. If we broke, we wouldn't necessarily
                 // find the appropriate child.
-                //
                 if (rc.Contains(x, y))
                 {
-                    ctlFound = ctl;
+                    ctlFound = control;
```
lonitra, line 339: Hm, I'm having a hard time seeing why we are continuing this loop once the control has been found. Perhaps we should return once it has been found.

lonitra (commented):
👍 mostly nits

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/TabOrder.cs
lonitra: If `host` is not null at any other point aside from `Dispose`, we can just make `host` not nullable and assign `null!` during disposal.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/TabOrder.cs
elachlan: The comment above explains it. I figure since someone went to the effort to explain that it was on purpose, then there is probably something not obvious happening.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/TabOrder.cs
elachlan: Its used in `OnMouseMove`.

lonitra: Am i missing something here? I don't see an `OnMouseMove` call in this method or this enumerator being utilized in this method.

elachlan: ohh I see, sorry. I thought you meant `GetControlAtPoint` wasn't used.

lonitra (approved):
👍

Issue #5036, opened 2021-06-03
Multilanguage .NET 5 Winforms app crashes on French Windows, related to Del shortcut key


xpouyat:
* .NET Core Version:
.Net 5.0.6

* Have you experienced this same bug with .NET Framework?: 
No

**Problem description:**

There an issue with a new multilanguage .NET 5 winforms app when deployed and run on a French Windows OS. The published App crashes at start on the French Windows with error:
Description: The process was terminated due to an unhandled exception.
Exception Info: System.ArgumentException: Requested value ‘Del’ was not found.

The issue seems related to the shortcut key “Del”. On the French Windows, this key is named differently (“Suppr”) and that may be the root of the issue.
Strangely, other keys like “End” are rejected by designer in Visual Studio, telling “Property value is not valid”. But the no such message is displayed for the “Del” key.

**Expected behavior:**
Expected: No crash on French Windows when running the app.

**Minimal repro:**

- On a English Windows, create a new .NET 5 Winforms app with latest Visual Studio 2019 (Version 16.10.0)
- Select the mainform and set Localizable property to True
- Add a ContextMenuStrip control and add an entry “test” in it. Associate this entry to the “Del” key, by setting the ShortcutKeys to “Del”.
- Publish the app to a local folder
- Run the published app on a French Windows.
- The UI does not show, no error is displayed on screen.
- In the event viewer of the French Windows, there is an error :
Application: WinFormsApp9.exe
CoreCLR Version: 5.0.621.22011
.NET Version: 5.0.6
Description: The process was terminated due to an unhandled exception.
Exception Info: System.ArgumentException: Requested value ‘Del’ was not found.
at System.Enum.TryParseByName(RuntimeType enumType, String originalValueString, ReadOnlySpan1 value, Boolean ignoreCase, Boolean throwOnFailure, UInt64& result)    at System.Enum.TryParseInt32Enum(RuntimeType enumType, String originalValueString, ReadOnlySpan1 value, Int32 minInclusive, Int32 maxInclusive, Boolean ignoreCase, Boolean throwOnFailure, TypeCode type, Int32& result)
at System.Enum.TryParse(Type enumType, String value, Boolean ignoreCase, Boolean throwOnFailure, Object& result)
at System.Enum.Parse(Type enumType, String value)
at System.Windows.Forms.KeysConverter.ConvertFrom(ITypeDescriptorContext context, CultureInfo culture, Object value)
[...]

PR #8401, opened 2022-12-19
Fixing runtime crash if an application contains `ToolStripMenuItem` with configured shortcut and launched in other locales.

### LINKED ISSUE #2886: Runtime crash due to localization and the way designer serializes keyboard shortcuts
* .NET Core Version: 3.1.2

**Problem description:**

Steps to reproduce:
- Create a blank .NET Framework WinForms application with a Form.
- Set the Form to `Localizable = true` in the designer.
- Add context menu with single item.
- Set the `ShortcutKeys` on the item to `Control+C`.
- Observe that the application works just fine on .NET Framework 4.8.
- Switch locale at startup to German by `System.Threading.Thread.CurrentThread.CurrentUICulture = new System.Globalization.CultureInfo("de-DE");`
- Observe that the application still starts just fine on .NET Framework 4.8.
- Change the application to target `netcoreapp31` and it immediately crashes on startup.

The reason for the crash is that the .resx file contains the shortcut key serialized as
```
  <data name="booToolStripMenuItem.ShortcutKeys" type="System.Windows.Forms.Keys, System.Windows.Forms">
    <value>Ctrl+C</value>
  </data>
```

Notice that the serialized value starts with `Ctrl`, not `Control`. It is produced and consumed by `KeysConverter` class that is [locale dependent](https://github.com/dotnet/winforms/blob/df89eb965dd59fd15341972e3a6ce33e51a86317/src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs#L44-L66). While the value `Control` would work in any locale due to the fallback to names in the `Keys` enum the value `Ctrl` is taken from a dictionary that contains localized names. On .NET Core 3.1 in the German localization the name for `Control` key is `Strg`. Thus the value `Ctrl+C` in the resource file is invalid and causes conversion error at runtime.

Notably we migrated a large project from .NET Framework to .NET Core where we hit this issue. The .NET Framework also has similar issue in code but the localizations may differ which would explain why the problem is not hit.

Manually fixing the value in the .resx file to `Control+C` makes the application work in any locale but any further editing in the designer breaks it again.

**Expected behavior:**

No crash. The resources generated by the designer should be locale independent.

**Minimal repro:**

Here's the sample app created using the steps above. It is multi-targeting both .NET Framework and .NET Core so it can run in both variation from inside Visual Studio.

[WindowsFormsApp2.zip](https://github.com/dotnet/winforms/files/4226867/WindowsFormsApp2.zip)

NikitaSemenovAkvelon:
Fixes #2886 

## Proposed changes

- Added support for the `culture` parameter in `ConvertFrom` and `ConvertTo` methods.
- Dictionaries with localization are loaded per request.
- `KeysConverter` becomes more independent and doesn't contain the logic of choosing the "right" locale. 

## Customer Impact

- No crashes when an application is run in different locales.
- A developer's locale doesn't impact values in the .resx file.
- The placeholder when the shortcut isn't set becomes more apparent.

## Regression? 

- Yes (after migration to .net core).

## Risk

- Old applications that contain localized values of `ShortcutKeys` (e.g. Strg for de-DE) are crashed.
- `ShortcutKeysEditor` seems inconsistent due to `KeysConverter` receiving `CultureInfo.CurrentCulture` while `ShortcutKeysUI` is localized with `CultureInfo.CurrentUICulture`.

### Before

Screenshot 2022-12-19 142918
Screenshot 2022-12-19 142011
image

### After

image
image
Screenshot 2022-12-22 171839

## Test methodology 

- Manually 
 

## Test environment(s) 

.NET SDK:
 Commit:    1b80461e45

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.22621

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8401)

### DIFF
--- a/src/System.Windows.Forms/src/Resources/SR.resx
+++ b/src/System.Windows.Forms/src/Resources/SR.resx
@@ -6963,4 +6963,7 @@ Stack trace where the illegal operation occurred was:
   <data name="FailedToLoadCursor" xml:space="preserve">
     <value>Failed to load cursor. Error '{1}'.</value>
   </data>
+  <data name="ResourceValueNotFound" xml:space="preserve">
+    <value>The resource value '{0}' was not found.</value>
+  </data>
 </root>
--- a/src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
@@ -15,34 +15,38 @@ namespace System.Windows.Forms;
 /// </summary>
 public class KeysConverter : TypeConverter, IComparer
 {
-    private IDictionary<string, Keys>? _keyNames;
-    private List<string>? _displayOrder;
+    private Dictionary<CultureInfo, List<string>>? _cultureToDisplayOrder;
+    private Dictionary<CultureInfo, Dictionary<string, Keys>>? _cultureToKeyName;
     private StandardValuesCollection? _values;
 
-    [MemberNotNull(nameof(_keyNames))]
-    [MemberNotNull(nameof(_displayOrder))]
+    [MemberNotNull(nameof(_cultureToDisplayOrder))]
+    [MemberNotNull(nameof(_cultureToKeyName))]
     private void Initialize()
     {
-        if (_keyNames is not null && _displayOrder is not null)
+        _cultureToDisplayOrder = new();
+        _cultureToKeyName = new();
+        AddLocalizedKeyNames(CultureInfo.InvariantCulture);
+    }
+
+    private void AddLocalizedKeyNames(CultureInfo cultureInfo)
+    {
+        if (CultureToDisplayOrder.ContainsKey(cultureInfo) && CultureToDisplayOrder.ContainsKey(cultureInfo))
         {
             return;
         }
 
-        Debug.Assert(_displayOrder is null);
-        Debug.Assert(_keyNames is null);
+        var localizedOrder = new List<string>(34);
+        var localizedNames = new Dictionary<string, Keys>(34);
 
-        _keyNames = new Dictionary<string, Keys>(34);
-        _displayOrder = new List<string>(34);
-
-        AddKey(SR.toStringEnter, Keys.Return);
+        AddLocalizedKey(nameof(SR.toStringEnter), Keys.Return);
         AddKey("F12", Keys.F12);
         AddKey("F11", Keys.F11);
         AddKey("F10", Keys.F10);
-        AddKey(SR.toStringEnd, Keys.End);
-        AddKey(SR.toStringControl, Keys.Control);
+        AddLocalizedKey(nameof(SR.toStringEnd), Keys.End);
+        AddLocalizedKey(nameof(SR.toStringControl), Keys.Control);
         AddKey("F8", Keys.F8);
         AddKey("F9", Keys.F9);
-        AddKey(SR.toStringAlt, Keys.Alt);
+        AddLocalizedKey(nameof(SR.toStringAlt), Keys.Alt);
         AddKey("F4", Keys.F4);
         AddKey("F5", Keys.F5);
         AddKey("F6", Keys.F6);
@@ -50,15 +54,16 @@ public class KeysConverter : TypeConverter, IComparer
         AddKey("F1", Keys.F1);
         AddKey("F2", Keys.F2);
         AddKey("F3", Keys.F3);
-        AddKey(SR.toStringPageDown, Keys.Next);
-        AddKey(SR.toStringInsert, Keys.Insert);
-        AddKey(SR.toStringHome, Keys.Home);
-        AddKey(SR.toStringDelete, Keys.Delete);
-        AddKey(SR.toStringShift, Keys.Shift);
-        AddKey(SR.toStringPageUp, Keys.Prior);
-        AddKey(SR.toStringBack, Keys.Back);
+        AddLocalizedKey(nameof(SR.toStringPageDown), Keys.Next);
+        AddLocalizedKey(nameof(SR.toStringInsert), Keys.Insert);
+        AddLocalizedKey(nameof(SR.toStringHome), Keys.Home);
+        AddLocalizedKey(nameof(SR.toStringDelete), Keys.Delete);
+        AddLocalizedKey(nameof(SR.toStringShift), Keys.Shift);
+        AddLocalizedKey(nameof(SR.toStringPageUp), Keys.Prior);
+        AddLocalizedKey(nameof(SR.toStringBack), Keys.Back);
+        AddLocalizedKey(nameof(SR.toStringNone), Keys.None);
 
-        //new whidbey keys follow here...
+        // new whidbey keys follow here...
         // Add string mappings for these values (numbers 0-9) so that the keyboard shortcuts
         // will be displayed properly in menus.
         AddKey("0", Keys.D0);
@@ -72,10 +77,20 @@ public class KeysConverter : TypeConverter, IComparer
         AddKey("8", Keys.D8);
         AddKey("9", Keys.D9);
 
+        CultureToKeyName.Add(cultureInfo, localizedNames);
+        CultureToDisplayOrder.Add(cultureInfo, localizedOrder);
+
         void AddKey(string key, Keys value)
         {
-            _keyNames[key] = value;
-            _displayOrder.Add(key);
+            localizedNames[key] = value;
+            localizedOrder.Add(key);
+        }
+
+        void AddLocalizedKey(string keyName, Keys value)
+        {
+            var key = SR.ResourceManager.GetString(keyName, cultureInfo)
+                      ?? throw new InvalidOperationException(string.Format(SR.ResourceValueNotFound, keyName));
+            AddKey(key, value);
         }
     }
 
@@ -83,25 +98,29 @@ public class KeysConverter : TypeConverter, IComparer
     ///  Access to a lookup table of name/value pairs for keys.  These are localized
     ///  names.
     /// </summary>
-    [MemberNotNull(nameof(_keyNames))]
-    [MemberNotNull(nameof(_displayOrder))]
-    private IDictionary<string, Keys> KeyNames
+    private Dictionary<CultureInfo, Dictionary<string, Keys>> CultureToKeyName
     {
         get
         {
-            Initialize();
-            return _keyNames;
+            if (_cultureToKeyName is null)
+            {
+                Initialize();
+            }
+
+            return _cultureToKeyName;
         }
     }
 
-    [MemberNotNull(nameof(_keyNames))]
-    [MemberNotNull(nameof(_displayOrder))]
-    private List<string> DisplayOrder
+    private Dictionary<CultureInfo, List<string>> CultureToDisplayOrder
     {
         get
         {
-            Initialize();
-            return _displayOrder;
+            if (_cultureToDisplayOrder is null)
+            {
+                Initialize();
+            }
+
+            return _cultureToDisplayOrder;
         }
     }
 
@@ -110,18 +129,14 @@ public class KeysConverter : TypeConverter, IComparer
     ///  type to the native type of the converter.
     /// </summary>
     public override bool CanConvertFrom(ITypeDescriptorContext? context, Type sourceType)
-    {
-        return sourceType == typeof(string) || sourceType == typeof(Enum[]) ? true : base.CanConvertFrom(context, sourceType);
-    }
+        => sourceType == typeof(string) || sourceType == typeof(Enum[]) || base.CanConvertFrom(context, sourceType);
 
     /// <summary>
     ///  Gets a value indicating whether this converter can
     ///  convert an object to the given destination type using the context.
     /// </summary>
     public override bool CanConvertTo(ITypeDescriptorContext? context, Type? destinationType)
-    {
-        return destinationType == typeof(Enum[]) ? true : base.CanConvertTo(context, destinationType);
-    }
+        => destinationType == typeof(Enum[]) || base.CanConvertTo(context, destinationType);
 
     /// <summary>
     ///  Compares two key values for equivalence.
@@ -144,50 +159,51 @@ public class KeysConverter : TypeConverter, IComparer
             {
                 return null;
             }
-            else
+
+            IDictionary<string, Keys> keyNames = GetKeyNames(culture);
+
+            // Parse an array of key tokens.
+            string[] tokens = text.Split(new char[] { '+' });
+            for (int i = 0; i < tokens.Length; i++)
             {
-                // Parse an array of key tokens.
-                string[] tokens = text.Split(new char[] { '+' });
-                for (int i = 0; i < tokens.Length; i++)
-                {
-                    tokens[i] = tokens[i].Trim();
-                }
-
-                // Now lookup each key token in our key hashtable.
-                Keys key = (Keys)0;
-                bool foundKeyCode = false;
-
-                for (int i = 0; i < tokens.Length; i++)
-                {
-                    if (!KeyNames.TryGetValue(tokens[i], out Keys currentKey))
-                    {
-                        // Key was not found in our dictionary.  See if it is a valid value in
-                        // the Keys enum.
-                        currentKey = (Keys)Enum.Parse(typeof(Keys), tokens[i]);
-                    }
-
-                    if ((currentKey & Keys.KeyCode) != 0)
-                    {
-                        // We found a match.  If we have previously found a
-                        // key code, then check to see that this guy
-                        // isn't a key code (it is illegal to have, say,
-                        // "A + B"
-                        if (foundKeyCode)
-                        {
-                            throw new FormatException(SR.KeysConverterInvalidKeyCombination);
-                        }
-
-                        foundKeyCode = true;
-                    }
-
-                    // Now OR the key into our current key
-                    key |= currentKey;
-                }
-
-                return (object)key;
+                tokens[i] = tokens[i].Trim();
             }
+
+            // Now lookup each key token in our key hashtable.
+            Keys key = 0;
+            bool foundKeyCode = false;
+
+            foreach (string token in tokens)
+            {
+                if (!keyNames.TryGetValue(token, out Keys currentKey))
+                {
+                    // Key was not found in our dictionary.  See if it is a valid value in
+                    // the Keys enum.
+                    currentKey = (Keys)Enum.Parse(typeof(Keys), token);
+                }
+
+                if ((currentKey & Keys.KeyCode) != 0)
+                {
+                    // We found a match.  If we have previously found a
+                    // key code, then check to see that this guy
+                    // isn't a key code (it is illegal to have, say,
+                    // "A + B"
+                    if (foundKeyCode)
+                    {
+                        throw new FormatException(SR.KeysConverterInvalidKeyCombination);
+                    }
+
+                    foundKeyCode = true;
+                }
+
+                // Now OR the key into our current key
+                key |= currentKey;
+            }
+
+            return key;
         }
-        else if (value is Enum[] valueAsEnumArray)
+
+        if (value is Enum[] valueAsEnumArray)
         {
             long finalValue = 0;
             foreach (Enum e in valueAsEnumArray)
@@ -217,30 +233,30 @@ public class KeysConverter : TypeConverter, IComparer
             return base.ConvertTo(context, culture, value, destinationType);
         }
 
-        bool asString = destinationType == typeof(string);
-        bool asEnum = !asString && destinationType == typeof(Enum[]);
-        if (asString || asEnum)
-        {
-            Keys key = (Keys)value;
-            return asString ? GetTermsString(key) : GetTermKeys(key);
-        }
-
-        return base.ConvertTo(context, culture, value, destinationType);
+        return destinationType == typeof(string)
+            ? GetTermsString((Keys)value)
+            : destinationType == typeof(Enum[])
+                ? GetTermKeys((Keys)value)
+                : base.ConvertTo(context, culture, value, destinationType);
 
         Enum[] GetTermKeys(Keys key)
         {
             List<Enum> termKeys = new();
-            Keys modifiers = (key & Keys.Modifiers);
+            Keys modifiers = key & Keys.Modifiers;
+            IDictionary<string, Keys> keyNames = GetKeyNames(culture);
+            IList<string> displayOrder = GetDisplayOrder(culture);
 
-            // First, iterate through and do the modifiers. These are
-            // additive, so we support things like Ctrl + Alt
-            for (int i = 0; i < DisplayOrder.Count; i++)
+            if (key != Keys.None)
             {
-                string keyString = DisplayOrder[i];
-                Keys keyValue = _keyNames[keyString];
-                if (modifiers.HasFlag(keyValue))
+                // First, iterate through and do the modifiers. These are
+                // additive, so we support things like Ctrl + Alt
+                foreach (var keyString in displayOrder)
                 {
-                    termKeys.Add(keyValue);
+                    Keys keyValue = keyNames[keyString];
+                    if (keyValue != Keys.None && modifiers.HasFlag(keyValue))
+                    {
+                        termKeys.Add(keyValue);
+                    }
                 }
             }
 
@@ -249,10 +265,9 @@ public class KeysConverter : TypeConverter, IComparer
             Keys keyOnly = key & Keys.KeyCode;
             bool foundKey = false;
 
-            for (int i = 0; i < DisplayOrder.Count; i++)
+            foreach (var keyString in displayOrder)
             {
-                string keyString = DisplayOrder[i];
-                Keys keyValue = _keyNames[keyString];
+                Keys keyValue = keyNames[keyString];
                 if (keyValue.Equals(keyOnly))
                 {
                     termKeys.Add(keyValue);
@@ -275,17 +290,21 @@ public class KeysConverter : TypeConverter, IComparer
         string GetTermsString(Keys key)
         {
             StringBuilder termStrings = new(32);
-            Keys modifiers = (key & Keys.Modifiers);
+            Keys modifiers = key & Keys.Modifiers;
+            IDictionary<string, Keys> keyNames = GetKeyNames(culture);
+            IList<string> displayOrder = GetDisplayOrder(culture);
 
-            // First, iterate through and do the modifiers. These are
-            // additive, so we support things like Ctrl + Alt
-            for (int i = 0; i < DisplayOrder.Count; i++)
+            if (key != Keys.None)
             {
-                string keyString = DisplayOrder[i];
-                Keys keyValue = _keyNames[keyString];
-                if (modifiers.HasFlag(keyValue))
+                // First, iterate through and do the modifiers. These are
+                // additive, so we support things like Ctrl + Alt
+                foreach (var keyString in displayOrder)
                 {
-                    termStrings.Append(keyString).Append('+');
+                    Keys keyValue = keyNames[keyString];
+                    if (keyValue != Keys.None && modifiers.HasFlag(keyValue))
+                    {
+                        termStrings.Append(keyString).Append('+');
+                    }
                 }
             }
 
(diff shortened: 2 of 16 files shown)


NikitaSemenovAkvelon:
**Important notes:**

- In `net481` it looks this way:
Screenshot 2022-12-19 141734
Screenshot 2022-12-19 140313

- Resource files contain value depending on version:
net481: Ctrl+D
net8 (before the fix): Strg+D
net8 (after the fix): Ctrl+D

**A few suggestions:**

1. To fix the crushes of old applications (the risk mentioned in the PR's description), add a workaround in `ShortcutKeysEditor` that would also check a parameter in the current locale in the `ConvertFrom` method.
2. Make a difference in resources of `ShortcutKeysEditor` for looking as in the `net481` version.

v-elnovikova (approved):
LGTM

dmitrii-drobotov (commented):
1. Let's say there is a .NET 7 app with "Strg+D" shortcut in resources, which is later updated to .NET 8 with this fix. As I understand the app will crash on runtime and designer won't load the form. I'm not sure it will be clear for developers what exactly needs to be changed in order to fix it. Is it possible to provide more details in this case?
2. Does it also fix scenario from https://github.com/dotnet/winforms/issues/5036?

Tanya-Solyanik:
What happens if a NET 8application is retargetted to NET7 in VS?

Tanya-Solyanik:
What happens if a NET 8application is retargettted to NET7 in VS?

NikitaSemenovAkvelon:
> Let's say there is a .NET 7 app with "Strg+D" shortcut in resources, which is later updated to .NET 8 with this fix. As I understand the app will crash on runtime and designer won't load the form.

Yes, an error would look like this:
image

> Is it possible to provide more details in this case? Maybe a breaking change page in docs could be useful?

Yes, we could handle this case and throw an exception with a detailed message. Also, we can modify the docs.

> Does it also fix scenario from https://github.com/dotnet/winforms/issues/5036?

I guess, it looks like the reason is the same.

> What happens if a NET 8application is retargetted to NET7 in VS?

It depends on the user's locale. For example, if the user has `de-DE`, the behavior will be exactly the same as described in this issue. But in other locales localized value of `toStringControl` is the same as in invariant culture.
I published this fix because in my humble opinion it covered most of the cases without workarounds. Even if we add a second dictionary that contains localized values it will crash when the resource file will have localized value in one locale while the application is opened in another non-invariant locale (e.g. `toStringBack` has different values in each locale). 
To prevent all crashes we have to check an incoming value in all locales and return only in the invariant culture. Whereas, it would a workaround for supporting previous versions and it also would reduce speed a bit.

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
```diff
@@ -79,6 +79,10 @@ void AddKey(string key, Keys value)
                 _keyNames[key] = value;
                 _displayOrder.Add(key);
             }
+
+            string GetResourceValueAsInvariant(string name)
+                => SR.ResourceManager.GetString(name, CultureInfo.InvariantCulture)
+                   ?? throw new InvalidOperationException($"The resource value hasn't been found for '{name}'.");
```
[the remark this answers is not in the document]
NikitaSemenovAkvelon: In this method, it doesn't make sense. It is just a double-check validation due to `ResourceManager.GetString` returning `nullable-string`.

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
```diff
@@ -36,29 +36,29 @@ private void Initialize()
             _keyNames = new Dictionary<string, Keys>(34);
             _displayOrder = new List<string>(34);
 
-            AddKey(SR.toStringEnter, Keys.Return);
+            AddKey(GetResourceValueAsInvariant(nameof(SR.toStringEnter)), Keys.Return);
```
Tanya-Solyanik, line 39: How did this code work in the .NET Framework designer? These strings were localized previously as well.

Tanya-Solyanik (commented):
In addition to this change we could add a custom TypeConverter to the designer to handle localization for downlevel versions of the framework.

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
Tanya-Solyanik, line 39: Please use collection initializer instead of `AddKey`.

NikitaSemenovAkvelon:
I've decided to rework PR to support provided culture in ConvertFrom and ConvertTo methods. According to a comment (https://github.com/dotnet/winforms/issues/2886#issuecomment-590344662) it was really broken. And it works as expected, the bug has gone but localization in runtime is still working.

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
```diff
@@ -330,7 +337,7 @@ public override StandardValuesCollection GetStandardValues(ITypeDescriptorContex
         {
             if (_values is null)
             {
-                Keys[] list = KeyNames.Values.ToArray();
+                Keys[] list = KeyNames[CultureInfo.InvariantCulture].Values.ToArray();
```
dmitrii-drobotov, line 340: Can you check in which culture `GetStandardValues` was returning in .NET Framework and .NET 7? It may have been returning values in current culture instead of invariant, and in that case I would keep existing behavior.

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
```diff
@@ -17,77 +17,87 @@ namespace System.Windows.Forms
...
+        {
+            Debug.Assert(_keyNames is not null);
+            if (_keyNames.ContainsKey(cultureInfo))
             {
-                _keyNames[key] = value;
-                _displayOrder.Add(key);
+                return;
             }
+
+            _keyNames = new Dictionary<CultureInfo, IDictionary<string, Keys>>();
```
dreddy-work, line 45: `_keyNames` should have already been assigned by this time. Also, if this dictionary is expected to have more than one culture stored in this, assigning like this will reset it.

elachlan:
Maybe we should merge PR #8485 with this?

elachlan:
Are the keys serialized to their non-localized values in the designer code? What happens if the translation changes? Can a form be designed okay between culture changes?

dreddy-work:
> Are the keys serialized to their non-localized values in the designer code? What happens if the translation changes? Can a form be designed okay between culture changes?

Localized Forms serialize into resx for the properties that are language dependent.  May be i didn't get your question.

NikitaSemenovAkvelon:
> Maybe we should merge PR #8485 with this?

It's not a problem to merge these separately, because these fix different cases and complement each other well.

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
weltkante: Not exactly, both the editor and msbuild were running in VS which used the VS installations locale (CurrentUICulture was fixed to be the VS installation language), however this bug canceled itself out because all the tools during the build ran under that locale. If you tried to open a project made in a german VS inside an english VS you would have run into problems though.

See my analysis [here in the issue](https://github.com/dotnet/winforms/issues/2886#issuecomment-590486228)

dreddy-work:
@NikitaSemenovAkvelon , can you rebase this and resolve conflicts?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/KeysConverterTests.cs
```diff
@@ -23,7 +24,7 @@ public void ConvertFrom_ShouldConvertKeys(string input, Keys keys)
         }
 
         [Theory]
-        [InlineData(Keys.None, "None")]
+        [InlineData(Keys.None, "(none)")]
```
dmitrii-drobotov, line 27: Changing `"None"` to `"(none)"` may be a breaking change and I don't see why it's needed. I think after https://github.com/dotnet/winforms/issues/8456 was fixed, adding `Keys.None` to `localizedNames` is not needed anymore.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/KeysConverterTests.cs
```diff
@@ -32,6 +33,7 @@ public void ConvertFrom_ShouldConvertKeys(string input, Keys keys)
         [InlineData(Keys.Control | Keys.Alt | Keys.Shift | Keys.F1, "Ctrl+Alt+Shift+F1")]
         public void ConvertToString_ShouldConvertKeys(Keys keys, string expectedResult)
         {
+            Thread.CurrentThread.CurrentCulture = CultureInfo.InvariantCulture;
```
dmitrii-drobotov, line 36: Consider passing `CultureInfo.InvariantCulture` as a parameter to `ConvertToString` call, changing thread culture may have side effects.

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
```diff
@@ -277,17 +270,20 @@ Enum[] GetTermKeys(Keys key)
...
-                // additive, so we support things like Ctrl + Alt
-                for (int i = 0; i < DisplayOrder.Count; i++)
+                if (key != Keys.None)
                 {
-                    string keyString = DisplayOrder[i];
-                    Keys keyValue = _keyNames[keyString];
-                    if (modifiers.HasFlag(keyValue))
+                    // First, iterate through and do the modifiers. These are
+                    // additive, so we support things like Ctrl + Alt
+                    foreach (var keyString in keyNames.Keys)
```
dmitrii-drobotov, line 280: With `DisplayOrder` list there was a guaranteed order of modifiers, for example for `Keys.Control | Keys.Shift | Keys.Alt` we would always get `"Ctrl+Alt+Shift"`. But when iterating `keyNames.Keys` this order is not guaranteed, so we might get `"Alt+Shift+Ctrl"` or something else. Is there a good way to fix it with current data structures?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/KeysConverterTests.cs
NikitaSemenovAkvelon: Almost everywhere we use `(none)` for an empty value. So, in addition to supporting provided language in `KeysConverter,` these changes also lead the "none" string value to the default value. 
image

elachlan: That is a very good point. Consistency is important.

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
NikitaSemenovAkvelon: Hmm... Yeah, due to the order isn't guaranteed (https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.keycollection?view=net-7.0#:~:text=order%20of%20the%20keys%20in%20the%20Dictionary%3CTKey%2CTValue%3E.KeyCollection%20is%20unspecified), it makes sense.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/KeysConverterTests.cs
dmitrii-drobotov: Ok, sounds good. Also given that "None" string is still correctly converted to `Keys.None`, I think it should be fine.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/KeysConverterTests.cs
```diff
@@ -23,17 +25,18 @@ public void ConvertFrom_ShouldConvertKeys(string input, Keys keys)
...
         [InlineData(Keys.Control | Keys.Add, "Ctrl+Add")]
         [InlineData(Keys.Control | Keys.Alt | Keys.D, "Ctrl+Alt+D")]
         [InlineData(Keys.Control | Keys.Alt | Keys.Shift | Keys.A, "Ctrl+Alt+Shift+A")]
         [InlineData(Keys.Control | Keys.Alt | Keys.Shift | Keys.F1, "Ctrl+Alt+Shift+F1")]
+        [InlineData(Keys.F2 | Keys.Shift | Keys.Alt | Keys.Control, "Ctrl+Alt+Shift+F2")]
         public void ConvertToString_ShouldConvertKeys(Keys keys, string expectedResult)
         {
             KeysConverter converter = new();
-            var result = converter.ConvertToString(keys);
+            var result = converter.ConvertToString(new MyTypeDescriptorContext(), CultureInfo.InvariantCulture, keys);
```
dmitrii-drobotov, line 39: I think you can just pass null as TypeDescriptorContext with the same result and without needing to create a new type.

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
```diff
@@ -219,30 +238,30 @@ public int Compare(object? a, object? b)
...
                 List<Enum> termKeys = new();
-                Keys modifiers = (key & Keys.Modifiers);
+                Keys modifiers = key & Keys.Modifiers;
+                IDictionary<string, Keys> keyNames = GetKeyNames(culture);
+                IList<string> displayOrder = GetDisplayOrder(culture);
 
-                // First, iterate through and do the modifiers. These are
-                // additive, so we support things like Ctrl + Alt
-                for (int i = 0; i < DisplayOrder.Count; i++)
+                if (key != Keys.None)
```
dmitrii-drobotov, line 254: Is this condition still needed? If `key` is `Keys.None` then it won't be added anyway because there is another check for it later: `if (keyValue != Keys.None && modifiers.HasFlag(keyValue))`. Same for `GetTermsString`

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/KeysConverterTests.cs
RussKie: Unless there's an breaking scenario you can think of, this change could be accepted and  documented as a "breaking change".

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
NikitaSemenovAkvelon: It's a small optimization to avoid extra iteration.

Tanya-Solyanik (commented):
Please describe scenarios you tested in greater details.

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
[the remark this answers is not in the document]
NikitaSemenovAkvelon: There would be no exceptions. Behavior doesn't change if the culture is installed or not. It depends only on Winforms resources. If provided culture is absent in our resources, the invariant culture will be used. A few examples below:

1. Culture is installed on the machine and has a translation in resources:
Screenshot 2023-01-23 092328
2. Culture isn't installed on the machine and doesn't have a translation in resources:
Screenshot 2023-01-23 091953
3. Culture isn't installed on the machine and has a translation in resources:
Screenshot 2023-01-23 091726

NikitaSemenovAkvelon:
> Please describe scenarios you tested in greater details.

Tested changes in different combinations of locales. All of them lead to store in resources values only in the invariant culture.

1. OS language: En, VS language: En
yriY834AcD

2. OS language: En, VS language: Ge
q816z9A3Ro

3. OS language: Ge, VS language: Ge
8ujk6bmUcJ

4. OS language: Ge, VS language: En
957rQ2ucdG

Also, here test results of running the application with different locales: https://github.com/dotnet/winforms/pull/8401#discussion_r1083679714

elachlan:
Great work! It is looking good!

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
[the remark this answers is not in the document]
Tanya-Solyanik: So we'll have problems when/if we add resources for currently unsupported languages and try to serialize the old resx file. This is reasonable. I verified that self-contained assemblies publish all resources.

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
```diff
@@ -74,36 +89,43 @@ private void Initialize()
             AddKey("8", Keys.D8);
             AddKey("9", Keys.D9);
 
+            _cultureToKeyName.Add(cultureInfo, localizedNames);
+            _cultureToDisplayOrder.Add(cultureInfo, localizedOrder);
+
             void AddKey(string key, Keys value)
```
Tanya-Solyanik, line 95: Please add explicit access modifier to AddKey and GetResourceValue.

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
```diff
@@ -17,50 +17,65 @@ namespace System.Windows.Forms
...
-
-            //new whidbey keys follow here...
+            AddKey(GetResourceValue(nameof(SR.toStringPageDown)), Keys.Next);
+            AddKey(GetResourceValue(nameof(SR.toStringInsert)), Keys.Insert);
+            AddKey(GetResourceValue(nameof(SR.toStringHome)), Keys.Home);
+            AddKey(GetResourceValue(nameof(SR.toStringDelete)), Keys.Delete);
+            AddKey(GetResourceValue(nameof(SR.toStringShift)), Keys.Shift);
+            AddKey(GetResourceValue(nameof(SR.toStringPageUp)), Keys.Prior);
+            AddKey(GetResourceValue(nameof(SR.toStringBack)), Keys.Back);
+            AddKey(GetResourceValue(nameof(SR.toStringNone)), Keys.None);
```
Tanya-Solyanik, line 76: Consider adding a helper method `AddLocalizedKey` that takes resource name as an argument.

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
```diff
@@ -74,36 +89,43 @@ private void Initialize()
...
             {
-                _keyNames[key] = value;
-                _displayOrder.Add(key);
+                localizedNames[key] = value;
+                localizedOrder.Add(key);
             }
+
+            string GetResourceValue(string name)
+                => SR.ResourceManager.GetString(name, cultureInfo)
+                   ?? throw new InvalidOperationException($"The resource value hasn't been found for '{name}'.");
```
Tanya-Solyanik, line 103: Please localize this exception message.

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
```diff
@@ -17,50 +17,65 @@ namespace System.Windows.Forms
...
-            if (_keyNames is not null && _displayOrder is not null)
+            if (_cultureToDisplayOrder is not null && _cultureToKeyName is not null)
             {
                 return;
             }
 
-            Debug.Assert(_displayOrder is null);
-            Debug.Assert(_keyNames is null);
+            Debug.Assert(_cultureToDisplayOrder is null);
+            Debug.Assert(_cultureToKeyName is null);
```
Tanya-Solyanik, line 34: These Asserts will never be raised because of the condition on line 28, please remove them.

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
```diff
@@ -17,50 +17,65 @@ namespace System.Windows.Forms
...
-            _keyNames = new Dictionary<string, Keys>(34);
-            _displayOrder = new List<string>(34);
+            _cultureToDisplayOrder = new Dictionary<CultureInfo, IList<string>>();
+            _cultureToKeyName = new Dictionary<CultureInfo, IDictionary<string, Keys>>();
+            AddLocalizedKeyNames(CultureInfo.InvariantCulture);
+        }
+
+        private void AddLocalizedKeyNames(CultureInfo cultureInfo)
+        {
+            Debug.Assert(_cultureToDisplayOrder is not null);
```
Tanya-Solyanik, line 43: Is `MemberNotNull` attribute not sufficient to raise warnings if the `initialize` method is not called?

Tanya-Solyanik (commented):
I would prefer to have this fixed in the designer, then the fix would apply to NET7.

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
NikitaSemenovAkvelon: These are local functions. Access modifiers aren't available here.

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
NikitaSemenovAkvelon: Yes, without `Debug.Assert` it has a warning.

RussKie: `Debug` isn't available in Release builds, so this won't work.

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
```diff
@@ -17,50 +17,62 @@ namespace System.Windows.Forms
...
-            if (_keyNames is not null && _displayOrder is not null)
+            if (_cultureToDisplayOrder is not null && _cultureToKeyName is not null)
             {
                 return;
             }
 
-            Debug.Assert(_displayOrder is null);
-            Debug.Assert(_keyNames is null);
+            _cultureToDisplayOrder = new Dictionary<CultureInfo, IList<string>>();
+            _cultureToKeyName = new Dictionary<CultureInfo, IDictionary<string, Keys>>();
```
RussKie, line 34:
```suggestion
            _cultureToDisplayOrder = new();
            _cultureToKeyName = new();
```

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
```diff
@@ -17,50 +17,62 @@ namespace System.Windows.Forms
...
+            AddLocalizedKeyNames(CultureInfo.InvariantCulture);
+        }
+
+        private void AddLocalizedKeyNames(CultureInfo cultureInfo)
+        {
+            Debug.Assert(_cultureToDisplayOrder is not null);
+            Debug.Assert(_cultureToKeyName is not null);
+            if (_cultureToDisplayOrder.ContainsKey(cultureInfo) && _cultureToKeyName.ContainsKey(cultureInfo))
```
RussKie, line 42: Any reason not to use `CultureToKeyName` and `CultureToDisplayOrder`?

NikitaSemenovAkvelon: Refactored this part

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/KeysConverterTests.cs
```diff
@@ -22,21 +23,44 @@ public void ConvertFrom_ShouldConvertKeys(string input, Keys keys)
             Assert.Equal(keys, result);
         }
 
+        [Fact]
+        public void ConvertFrom_ShouldConvertKeys_Localization()
+        {
```
riverar, line 28: I suggest adding a test for `None` to ensure that conversion to Keys.None continues to be supported in all languages. (I tested it locally and it works.)

riverar:
Thanks for fixing this! Just a minor nit but looks good otherwise.

src/System.Windows.Forms/src/Resources/SR.resx
```diff
@@ -6961,4 +6961,7 @@ Stack trace where the illegal operation occurred was:
   <data name="InvalidArgumentType" xml:space="preserve">
     <value>Argument '{0}' must be of type '{1}'.</value>
   </data>
+  <data name="ResourceValueHasntBeenFound" xml:space="preserve">
```
riverar, line 6964:
```suggestion
  <data name="ResourceValueNotFound" xml:space="preserve">
```

riverar (dismissed):

dkazennov (approved):
Great job. LGTM.

dreddy-work:
@NikitaSemenovAkvelon, please rebase this PR.

dmitrii-drobotov:
I will take over this PR, since Nikita isn't on the team anymore.

Tanya-Solyanik:
@dmitrii-drobotov - please rebase this PR

Tanya-Solyanik (dismissed):

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
```diff
@@ -318,6 +336,28 @@ string GetTermsString(Keys key)
...
+        {
+            AddLocalizedKeyNames(culture);
+        }
+
+        return CultureToDisplayOrder[culture];
+    }
+
+    private IDictionary<string, Keys> GetKeyNames(CultureInfo? culture)
+    {
+        culture ??= CultureInfo.CurrentCulture;
```
Tanya-Solyanik, line 352: Why is this not a CurrentUICulture? 
What are other TypeConverters using?

Tanya-Solyanik:
What controls besides MenuItems should be tested(i.e. have shortcut keys)?

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
riverar: Other type converters are also using `CurrentCulture`, which is strange. I tend to agree that this should be `CurrentUICulture` as I may have Windows display language set to Spanish but use English regional dates, currency, etc.

riverar:
> What controls besides MenuItems should be tested(i.e. have shortcut keys)?

Our app uses the `KeysConverter` directly and round trips text in the control. We are satisified that the tests in this PR cover our scenario.

src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
dmitrii-drobotov: Yeah, in winforms repo we use `CurrentCulture` in converters and it's also what base `TypeConverter` uses: https://github.com/dotnet/runtime/blob/9f8f6531b72c44da09d5ffcadf638f11cf51d227/src/libraries/System.ComponentModel.TypeConverter/src/System/ComponentModel/TypeConverter.cs#L49

Tanya-Solyanik: @dmitrii-drobotov - please add a comment that other converters use CurrentCulture. I guess when developers need a specific culture, for example the UI one, they specify it explicitly.

riverar: It matches what other converters use in the codebase, I don't think a comment is worth anything here.

dmitrii-drobotov:
> What controls besides MenuItems should be tested(i.e. have shortcut keys)?

I don't see any other controls with `Keys` properties, other than `ToolStripMenuItem`. But I guess it's possible to create a custom control with `Keys` property, and it would be used there. There's also [`ShortcutKeysEditor.ShortcutKeysUI`](https://github.com/dotnet/winforms/blob/2ad2bc990c7a3d874e48a38322ae0dc35322defd/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ShortcutKeysEditor.ShortcutKeysUI.cs#L63) which will be opened in a PropertyGrid entry for a Keys property:
image

dmitrii-drobotov:
> 2.Scenario:
> 
> * on EN OS and EN VS, create a form with toolstrip
> * add a menu item with a shortcut key  >>> Form1.resx contains Ctrl+A for a shortcut
> * Make the form localizable, set language to DE-DE
> * Change the shortcut key to Ctrl+B->from1.de-DE.resx contains Ctrl+B as a shortcut, i.e. localized resource file has an invariant key name - this matches the .NET FX
>   Question: run that app on a DE machine with this change - which resource file is the shortcut read from in a Core app and in a FX app?

@Tanya-Solyanik 
The shortcut will be read from DE resource file both in .NET Fx and .NET 8 with or without this PR. But Form1.de-DE.resx needs to be created manually, for me it wasn't created automatically even on all DE settings. And the DE shortcut value also needs to be set manually, but it will override invariant value correctly.

Tanya-Solyanik:
>But Form1.de-DE.resx needs to be created manually, for me it wasn't created automatically even on all DE settings. 

This is strange. When I set Localizable property on the form to true and Language to de-DE, Form1.de-DE.resx is created

Tanya-Solyanik (approved):

PR #8413, opened 2022-12-22
Enabling nullability annotations to `System.Windows.Forms.Design` classes

elachlan:
Enable Nullability in multiple Design classes.

This PR focuses on classes with low line counts to reduce the review burden.

Related: #8342

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8413)

### DIFF
--- a/src/System.Windows.Forms.Design/src/Resources/SR.resx
+++ b/src/System.Windows.Forms.Design/src/Resources/SR.resx
@@ -1314,4 +1314,7 @@ Press Ctrl+Enter to accept Text.</value>
   <data name="StringPropertyTruncatedComment" xml:space="preserve">
     <value> {0} [rest of string was truncated]";</value>
   </data>
+  <data name="ExceptionCollectionInvalidArgument" xml:space="preserve">
+    <value>Argument must only contain objects of type Exception.</value>
+  </data>
 </root>
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionPropertyItem.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionPropertyItem.cs
@@ -6,18 +6,18 @@ namespace System.ComponentModel.Design
 {
     public sealed class DesignerActionPropertyItem : DesignerActionItem
     {
-        public DesignerActionPropertyItem(string memberName, string displayName, string? category, string? description)
+        public DesignerActionPropertyItem(string memberName, string? displayName, string? category, string? description)
             : base(displayName, category, description)
         {
             MemberName = memberName;
         }
 
-        public DesignerActionPropertyItem(string memberName, string displayName)
+        public DesignerActionPropertyItem(string memberName, string? displayName)
             : this(memberName, displayName, category: null, description: null)
         {
         }
 
-        public DesignerActionPropertyItem(string memberName, string displayName, string? category)
+        public DesignerActionPropertyItem(string memberName, string? displayName, string? category)
             : this(memberName, displayName, category, description: null)
         {
         }
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionPropertyItem.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionPropertyItem.cs
@@ -2,13 +2,11 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 namespace System.ComponentModel.Design
 {
     public sealed class DesignerActionPropertyItem : DesignerActionItem
     {
-        public DesignerActionPropertyItem(string memberName, string displayName, string category, string description) : base(displayName, category, description)
+        public DesignerActionPropertyItem(string memberName, string displayName, string? category, string? description) : base(displayName, category, description)
         {
             MemberName = memberName;
         }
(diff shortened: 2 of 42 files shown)


src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionPropertyItem.cs
```diff
@@ -17,12 +15,12 @@ public DesignerActionPropertyItem(string memberName, string displayName) : this(
         {
         }
 
-        public DesignerActionPropertyItem(string memberName, string displayName, string category) : this(memberName, displayName, category, null)
+        public DesignerActionPropertyItem(string memberName, string displayName, string? category) : this(memberName, displayName, category, null)
```
gpetrou, line 18: Maybe add the parameter name `description: null` and move `: this(etc...` to a new line? You could do similar changes in other places.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ObjectSelectorEditor.SelectorNode.cs
```diff
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.Windows.Forms;
 
 namespace System.ComponentModel.Design
```
gpetrou, line 7: I guess the constructor parameters could be nullable?

src/System.Windows.Forms.Design/src/System/Drawing/Design/ColorEditor.SystemColorComparer.cs
```diff
@@ -16,10 +14,12 @@ public partial class ColorEditor
         /// </summary>
         private class SystemColorComparer : IComparer
```
gpetrou, line 15: Since this is private, maybe change it to `IComparer<Color>`?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ObjectSelectorEditor.SelectorNode.cs
elachlan: I wasn't sure. `value` could probably be nullable, but `label` isn't nullable in `TreeNode`. Documentation doesn't say much either.

src/System.Windows.Forms.Design/src/System/Drawing/Design/ColorEditor.SystemColorComparer.cs
elachlan: That ended up causing compile errors in `ColorEditor.ColorUI`. So enabled nullability there and refactored it to fix the errors.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ObjectSelectorEditor.SelectorNode.cs
lonitra: Hm, I think `value` can be annotated as nullable since it is public and there is no indicators via doc or codebase that is cannot be null.

elachlan:
The test failures are due to the `ExceptionCollection_Ctor_ArrayList` test passing in `new ArrayList { 1, 2, 3 }` to the constructor. The documentation says the ArrayList should be an array of exceptions: https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.design.exceptioncollection.-ctor

@lonitra This is technically a behaviour change. Should I revert the change, or update the test to use exceptions and add a arguement exception if there are any non-exception typed objects in the ArrayList?
@lonitra what would the team like to do there.

dreddy-work:
> The test failures are due to the `ExceptionCollection_Ctor_ArrayList` test passing in `new ArrayList { 1, 2, 3 }` to the constructor. The documentation says the ArrayList should be an array of exceptions: https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.design.exceptioncollection.-ctor

Given the documentation is clear about this, I think we should be good to go with test changes here.  @JeremyKuhne / @Tanya-Solyanik FYI.

elachlan:
@dreddy-work is it okay to add a check in the constructor to throw an argument exception?

Also why don't we use `AggregateException` instead? (https://learn.microsoft.com/en-us/dotnet/api/system.aggregateexception?view=net-7.0) This would remove the need for Winforms to maintain its own implementation.

dreddy-work:
> @dreddy-work is it okay to add a check in the constructor to throw an argument exception?

That make sense. Failing early. 

> Also why don't we use AggregateException instead?

This, I need to look further. If you make a draft PR around that, we could discuss separately.

elachlan:
> This, I need to look further. If you make a draft PR around that, we could discuss separately.

Done. #8449

dreddy-work (commented):
RLGTM.

dreddy-work (approved):
LGTM.

dmitrii-drobotov (approved):

JeremyKuhne:
> Given that the documentation clearly explains this, I believe we can proceed with making changes to the tests. @JeremyKuhne / @Tanya-Solyanik FYI.

Given that these are effectively supposed to be typed collections I'm fine with the breaking change of actually enforcing that via new exceptions. We need to document that we've updated our collections in general to enforce correct usage as part of our breaking change docs.

src/System.Windows.Forms.Design/src/Resources/SR.resx
```diff
@@ -1314,4 +1314,7 @@ Press Ctrl+Enter to accept Text.</value>
   <data name="StringPropertyTruncatedComment" xml:space="preserve">
     <value> {0} [rest of string was truncated]";</value>
   </data>
-</root>
+  <data name="ExceptionCollectionInvalidArgument" xml:space="preserve">
+    <value>Argument must only contain objects of type Exception.</value>
```
Tanya-Solyanik, line 1318: @elachlan  - could you please change this string to a format where the argument is `nameof(Exception)` and add quotation marks around that argument in a follow up PR?

dreddy-work:
https://github.com/dotnet/docs/issues/33695

PR #8420, opened 2022-12-23
ActiveX Code Shuffling

JeremyKuhne:
Move the Control ActiveX interfaces to their own file and move the remaining interfaces to CsWin32.

Pull out the nested AxContainer.ExtenderProxy class.

Remove unused Interop code.

Some sporadic cleanup in code I've been reviewing.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8420)


src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
```diff
@@ -1598,75 +1664,70 @@ internal void SetClientSite(IOleClientSite* value)
...
-                    _activeXState[s_changingExtents] = true;
+                _activeXState[s_changingExtents] = true;
 
-                    try
-                    {
-                        Size size = new Size(HiMetricToPixel(pSizel->Width, pSizel->Height));
-                        Debug.WriteLineIf(CompModSwitches.ActiveX.TraceInfo, "SetExtent : new size:" + size.ToString());
+                try
+                {
+                    Size size = new Size(HiMetricToPixel(pSizel->Width, pSizel->Height));
```
lonitra, line 1684: Is there a chance `pSizel` can be null? We do not check before we dereference it here.

JeremyKuhne: We check on both entry points

lonitra (approved):
🚀

PR #8429, opened 2022-12-28

gpetrou:
## Proposed changes

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8429)


src/System.Windows.Forms/src/System/Windows/Forms/StatusStrip.cs
```diff
@@ -235,7 +235,7 @@ public new bool Stretch
 
         private TableLayoutSettings TableLayoutSettings
         {
-            get { return (TableLayoutSettings)LayoutSettings; }
+            get { return (TableLayoutSettings)LayoutSettings!; }
```
lonitra, line 238: I see the get for LayoutSettings is nullable, how do we know it's not null here?

gpetrou: This private property seems to be always called when we have `LayoutStyle == ToolStripLayoutStyle.Table`. So, I assumed that this is not null when used.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
```diff
@@ -1577,15 +1573,16 @@ public bool Stretch
...
         {
             get
             {
                 if (IsDropDown)
                 {
                     // PERF: since this is called a lot we don't want to make it virtual
-                    ToolStripDropDown dropDown = this as ToolStripDropDown;
-                    if (dropDown is ToolStripOverflow || dropDown.IsAutoGenerated)
+                    ToolStripDropDown? dropDown = this as ToolStripDropDown;
+                    if (dropDown is ToolStripOverflow || (dropDown is not null && dropDown.IsAutoGenerated))
```
dreddy-work, line 1585: It is a dropDown and to keep the behavior unchanged,  I would change it with direct cast.
```suggestion
                    var dropDown = (ToolStripDropDown)this;
                    if (dropDown is ToolStripOverflow || dropDown.IsAutoGenerated)
```

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
```diff
@@ -1608,7 +1605,7 @@ public ToolStripRenderer Renderer
                     Renderer = ToolStripManager.CreateRenderer(RenderMode);
```
dreddy-work, line 1605: We can replace this with.
```suggestion
                    _renderer = value;
                    _currentRendererType = _renderer.GetType();
                    OnRendererChanged(EventArgs.Empty);
```

dreddy-work (approved):

PR #8435, opened 2022-12-29
Enable nullability in Formatter

gpetrou:
## Proposed changes

- Enable nullability in Formatter.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8435)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Formatter.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Formatter.cs
@@ -436,37 +436,37 @@ namespace System.Windows.Forms
         {
             try
             {
-                MethodInfo? mi = targetType.GetMethod(
+                MethodInfo? methodInfo = targetType.GetMethod(
                     "Parse",
                     BindingFlags.Public | BindingFlags.Static,
                     null,
                     new Type[] { stringType, typeof(NumberStyles), typeof(IFormatProvider) },
                     null);
-                if (mi is not null)
+                if (methodInfo is not null)
                 {
-                    return mi.Invoke(obj: null, new object?[] { (string?)value, NumberStyles.Any, formatInfo });
+                    return methodInfo.Invoke(obj: null, new object?[] { (string?)value, NumberStyles.Any, formatInfo });
                 }
 
-                mi = targetType.GetMethod(
+                methodInfo = targetType.GetMethod(
                     "Parse",
                     BindingFlags.Public | BindingFlags.Static,
                     null,
                     new Type[] { stringType, typeof(IFormatProvider) },
                     null);
-                if (mi is not null)
+                if (methodInfo is not null)
                 {
-                    return mi.Invoke(obj: null, new object?[] { (string?)value, formatInfo });
+                    return methodInfo.Invoke(obj: null, new object?[] { (string?)value, formatInfo });
                 }
 
-                mi = targetType.GetMethod(
+                methodInfo = targetType.GetMethod(
                     "Parse",
                     BindingFlags.Public | BindingFlags.Static,
                     null,
                     new Type[] { stringType },
                     null);
-                if (mi is not null)
+                if (methodInfo is not null)
                 {
-                    return mi.Invoke(obj: null, new object?[] { (string?)value });
+                    return methodInfo.Invoke(obj: null, new object?[] { (string?)value });
                 }
 
                 return parseMethodNotFound;
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Formatter.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Formatter.cs
@@ -138,13 +138,13 @@ namespace System.Windows.Forms
             TypeConverter sourceTypeTypeConverter = TypeDescriptor.GetConverter(sourceType);
             if (sourceConverter is not null && sourceConverter != sourceTypeTypeConverter && sourceConverter.CanConvertTo(targetType))
             {
-                return sourceConverter.ConvertTo(null, GetFormatterCulture(formatInfo), value, targetType);
+                return sourceConverter.ConvertTo(context: null, GetFormatterCulture(formatInfo), value, targetType);
             }
 
             TypeConverter targetTypeTypeConverter = TypeDescriptor.GetConverter(targetType);
             if (targetConverter is not null && targetConverter != targetTypeTypeConverter && targetConverter.CanConvertFrom(sourceType))
             {
-                return targetConverter.ConvertFrom(null, GetFormatterCulture(formatInfo), value);
+                return targetConverter.ConvertFrom(context: null, GetFormatterCulture(formatInfo), value);
             }
 
             if (targetType == checkStateType)
@@ -159,7 +159,7 @@ namespace System.Windows.Forms
 
                     if (sourceConverter is not null && sourceConverter.CanConvertTo(booleanType))
                     {
-                        return (bool)sourceConverter.ConvertTo(null, GetFormatterCulture(formatInfo), value, booleanType)!
+                        return (bool)sourceConverter.ConvertTo(context: null, GetFormatterCulture(formatInfo), value, booleanType)!
                             ? CheckState.Checked
                             : CheckState.Unchecked;
                     }
@@ -185,11 +185,11 @@ namespace System.Windows.Forms
 
             if (sourceConverter is not null && sourceConverter.CanConvertTo(targetType))
             {
-                return sourceConverter.ConvertTo(null, GetFormatterCulture(formatInfo), value, targetType);
+                return sourceConverter.ConvertTo(context: null, GetFormatterCulture(formatInfo), value, targetType);
             }
             else if (targetConverter is not null && targetConverter.CanConvertFrom(sourceType))
             {
-                return targetConverter.ConvertFrom(null, GetFormatterCulture(formatInfo), value);
+                return targetConverter.ConvertFrom(context: null, GetFormatterCulture(formatInfo), value);
             }
             else if (value is IConvertible)
             {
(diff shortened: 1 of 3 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Formatter.cs
```diff
@@ -286,7 +289,7 @@ internal static class Formatter
             TypeConverter targetTypeTypeConverter = TypeDescriptor.GetConverter(targetType);
             if (targetConverter is not null && targetTypeTypeConverter != targetConverter && targetConverter.CanConvertFrom(sourceType))
             {
-                return targetConverter.ConvertFrom(null, GetFormatterCulture(formatInfo), value);
+                return targetConverter.ConvertFrom(null, GetFormatterCulture(formatInfo), value!);
```
dreddy-work, line 292: param names where constants/nulls passed. 

Also, you could check for value being bull as well on line 280 and avoid `!` here.

src/System.Windows.Forms/src/System/Windows/Forms/Formatter.cs
```diff
@@ -344,7 +347,7 @@ internal static class Formatter
 
             if (targetConverter is not null && targetConverter.CanConvertFrom(sourceType))
             {
-                return targetConverter.ConvertFrom(null, GetFormatterCulture(formatInfo), value);
+                return targetConverter.ConvertFrom(null, GetFormatterCulture(formatInfo), value!);
```
dreddy-work, line 350: param names for `null` . We have multiple instances like this in this file.

src/System.Windows.Forms/src/System/Windows/Forms/Formatter.cs
```diff
@@ -429,64 +432,65 @@ private static CultureInfo GetFormatterCulture(IFormatProvider formatInfo)
...
             try
             {
-                MethodInfo mi;
-
-                mi = targetType.GetMethod("Parse",
-                                        BindingFlags.Public | BindingFlags.Static,
-                                        null,
-                                        new Type[] { stringType, typeof(NumberStyles), typeof(IFormatProvider) },
-                                        null);
+                MethodInfo? mi = targetType.GetMethod(
```
dreddy-work, line 439: mi => methodInfo

src/System.Windows.Forms/src/System/Windows/Forms/Formatter.cs
```diff
@@ -429,64 +432,65 @@ private static CultureInfo GetFormatterCulture(IFormatProvider formatInfo)
...
-                    return mi.Invoke(null, new object[] { (string)value });
+                    return mi.Invoke(null, new object?[] { (string?)value });
                 }
 
                 return parseMethodNotFound;
             }
             catch (TargetInvocationException ex)
             {
-                throw new FormatException(ex.InnerException.Message, ex.InnerException);
+                throw new FormatException(ex.InnerException!.Message, ex.InnerException);
```
dreddy-work, line 476: I do not see any reason why we can not make this change here. 

```suggestion
                throw new FormatException(ex.InnerException?.Message, ex.InnerException);
```

dreddy-work (changes requested):

src/System.Windows.Forms/src/System/Windows/Forms/Formatter.cs
dreddy-work: @gpetrou , you missed this suggestion?

src/System.Windows.Forms/src/System/Windows/Forms/Formatter.cs
```diff
@@ -262,21 +264,22 @@ internal static class Formatter
...
+            TypeConverter? sourceConverter,
+            IFormatProvider? formatInfo,
+            object? formattedNullValue)
         {
             //
             // Convert the formatted representation of 'null' to DBNull (if possible)
             //
 
-            if (EqualsFormattedNullValue(value, formattedNullValue, formatInfo) || value == System.DBNull.Value)
+            if (EqualsFormattedNullValue(value, formattedNullValue, formatInfo) || value == DBNull.Value)
```
dreddy-work, line 280: Checking `value` to null in this block make sense based on description and remove `!` from rest of the instances.  @Tanya-Solyanik / @KlausLoeffelmann any suggestions here?

src/System.Windows.Forms/src/System/Windows/Forms/Formatter.cs
```diff
@@ -344,11 +347,11 @@ internal static class Formatter
 
             if (targetConverter is not null && targetConverter.CanConvertFrom(sourceType))
             {
-                return targetConverter.ConvertFrom(null, GetFormatterCulture(formatInfo), value);
+                return targetConverter.ConvertFrom(context: null, GetFormatterCulture(formatInfo), value!);
```
dreddy-work, line 350: Based on my understanding of the code, it seems that it is possible for the value to be null in this situation, which supports the reasoning for the change i proposed above..

src/System.Windows.Forms/src/System/Windows/Forms/Formatter.cs
dreddy-work: Only if `formattedNullValue` is null.

src/System.Windows.Forms/src/System/Windows/Forms/Formatter.cs
KlausLoeffelmann: The way I understand it is: We're parsing `value`, which could be any representation which would lead to `dbNull`, so I guess that also includes a "real" `null`. In that case, we're indeed returning `dbNull` - so at this point every `null`, `""`, `"{NULL}"` case (I am making the null-versions up for an easier explanation) has been handled. Since we are not changing the type of the variable just because of its nullability, I'd assume a) `value` cannot be `null` at this point and b) using `!` in this context is one of the cases where we have `!` for.

dreddy-work: > so I guess that also includes a "real" `null`. In that case, we're indeed returning `dbNull` - so at this point every `null`, `""`, `"{NULL}"` case (I am making the null-versions up for an easier explanation) has been handled

Based on `null` check on line 330, i assumed it is possible for `value ` to be null here. Where do you see `value ` being `null` handled? Are you referring to callers?

KlausLoeffelmann: I wasn't, but that should be checked. Correct, the "real" `null` isn't checked at that point. The question is, would `null` ever been passed, but then again if that is the case, we wouldn't need nullability for `value` in that signature to begin with.

dreddy-work: I have noticed that there is at least one situation where a `null `value (in a `DataGridViewCell`) may be passed. I am not sure if it is safe to return dbNull if the value is `null`. @Tanya-Solyanik ?

src/System.Windows.Forms/src/System/Windows/Forms/Formatter.cs
elachlan: @Youssef1313 sorry to ping you all the time, but do you know of an analyzer for using a parameter name when the value is null?

src/System.Windows.Forms/src/System/Windows/Forms/Formatter.cs
```diff
@@ -262,21 +264,22 @@ internal static class Formatter
         ///  - Uses TypeConverters or IConvertible where appropriate
         ///  - Throws a FormatException is no suitable conversion can be found
```
Tanya-Solyanik, line 265:
```suggestion
        ///  - Throws a FormatException if no suitable conversion can be found
```

src/System.Windows.Forms/src/System/Windows/Forms/Formatter.cs
Tanya-Solyanik: Hm, `value==null` means that value had not been entered, not that a `null` value had been entered...

Tanya-Solyanik: Before these changes `value==null` was processed by the type converter or resulted in a FormatException. I don't think that returning DBNull.Value is accurate.

src/System.Windows.Forms/src/System/Windows/Forms/Formatter.cs
```diff
@@ -286,19 +289,19 @@ internal static class Formatter
             TypeConverter targetTypeTypeConverter = TypeDescriptor.GetConverter(targetType);
             if (targetConverter is not null && targetTypeTypeConverter != targetConverter && targetConverter.CanConvertFrom(sourceType))
             {
-                return targetConverter.ConvertFrom(null, GetFormatterCulture(formatInfo), value);
+                return targetConverter.ConvertFrom(context: null, GetFormatterCulture(formatInfo), value!);
```
Tanya-Solyanik, line 292: I'm not sure that it's correct that TypeConverter does not accept a `null` value in ConvertFrom but accepts it in  ConvertTo.

src/System.Windows.Forms/src/System/Windows/Forms/Formatter.cs
Tanya-Solyanik: I don't think we should return dbNull in the case of `value==null`.  Nor should we skip TypeConverter - it could be user-defined and it could handle nulls. It's unfortunate that the runtime team already marked ConvertFrom as not taking nulls. I think it's appropriate to throw an exception inthis case in the `Parse...` method as this should be something that developer finds while porting the app.

src/System.Windows.Forms/src/System/Windows/Forms/Formatter.cs
Youssef1313: @elachlan https://github.com/dotnet/roslyn/issues/43988

dreddy-work (approved):

src/System.Windows.Forms/src/System/Windows/Forms/Formatter.cs
elachlan: I would, but the analyzer hasn't been written yet. I'll add it once it's been written.

PR #8439, opened 2022-12-31
Fix keyboard layout name returned from System.Windows.Forms.InputLanguage.LayoutName

### LINKED ISSUE #4345: InputLanguage.LayoutName can return the wrong value when multiple layouts for a single language exist
* .NET Core Version: 5

**Problem description:**
When there are multiple keyboard layouts of the same language for a single language installed, InputLanguage.LayoutName returns the same value for all of them.

**Expected behavior:**
The correct name of each keyboard layout should be returned

**Minimal repro:**
In Windows 10, add the Bulgarian language (just basic typing), then add both the Bulgarian and Bulgarian (Phonetic Traditional) keyboard layouts. Use InputLanguage.InstalledInputLanguages to iterate over both layouts, note that the InputLanguage for both returns the same value of "Bulgarian"

**Analysis**
Looking at the code, this is because it's looking up substitutions in HKCU\Substitutes but matching on language alone: https://github.com/dotnet/winforms/blob/a07615a219d84f7af24f639dc84f18427083f754/src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs#L180

If you look in the registry key, you can see values for both 00000402 and d0010402 which would both match Bulgarian language (0402), and therefore the same substitution is used for both layouts. I believe the fix is probably to search all substitutions and then check the Layout Id value of all target substitutes to see which one matches the correct layout Id.

DJm00n:
Properly convert HKL to KLID string according to the comment from Windows team at https://github.com/dotnet/winforms/issues/4345#issuecomment-759161693

Don't need to look at `HKEY_CURRENT_USER\Keyboard Layout\Substitutes` at all since HKL is already have all needed information.

Fixes #4345 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8439)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
@@ -104,14 +104,8 @@ namespace System.Windows.Forms
         {
             get
             {
-                // There is no good way to do this in Windows.
-                // GetKeyboardLayoutName does what we want, but only for the current input
-                // language; setting and resetting the current input language would generate
-                // spurious InputLanguageChanged events.
-                // Try to extract needed information manually from the HKL.
                 string layoutName = GetKeyboardLayoutNameForHKL(_handle);
-
-                return GetLocalizedKeyboardLayoutName(layoutName);
+                return GetKeyboardLayoutDisplayName(layoutName);
             }
         }
 
@@ -125,24 +119,39 @@ namespace System.Windows.Forms
         /// </summary>
         internal static string GetKeyboardLayoutNameForHKL(IntPtr hkl)
         {
-            // HKL is a 32 bit value.
-            // HIWORD is a Device Handle.
-            // LOWORD is Language Id.
-            int language = unchecked((int)(long)hkl) & 0xFFFF;
-            int device = (unchecked((int)(long)hkl) >> 16) & 0xFFFF;
+            // There is no good way to do this in Windows.
+            // GetKeyboardLayoutName does what we want, but only for the current input
+            // language; setting and resetting the current input language would generate
+            // spurious InputLanguageChanged events.
+            //
+            // According to the docs of GetKeyboardLayout API function high word of HKL
+            // contains a device handle to the physical layout of the keyboard
+            // but exact format of this handle is not documented.
+            //
+            // We may try to use low word of HKL which contains input language but same
+            // keyboard can be installed more than once under different languages in Windows
+            // and it would give us wrong results in cases where French keyboard is
+            // installed under US input language etc.
+            //
+            // For older keyboards layouts device handle seems contain keyboard layout
+            // language which we can use as KLID without issues.
+            int device = PARAM.HIWORD(hkl);
 
-            // Device Handle contains Layout Id if its high nibble is 0xF
+            // But for newer keyboards layouts device handle contains layout id
+            // if its high nibble is 0xF. This id may be used to search for keyboard layout
+            // under registry.
+            // NOTE: this logic may break in future versions of Windows since its not documented.
             if ((device & 0xF000) == 0xF000)
             {
-                // Extract Layout Id from the Device Handle
+                // Extract layout id from the device handle
                 int layoutId = device & 0x0FFF;
 
-                if (KeyboardLayoutsRegistryKey is RegistryKey subKeys)
+                if (KeyboardLayoutsRegistryKey is RegistryKey key)
                 {
-                    // Match keyboard layout by Layout Id
-                    foreach (string subKeyName in subKeys.GetSubKeyNames())
+                    // Match keyboard layout by layout id
+                    foreach (string subKeyName in key.GetSubKeyNames())
                     {
-                        if (subKeys.OpenSubKey(subKeyName) is not RegistryKey subKey)
+                        if (key.OpenSubKey(subKeyName) is not RegistryKey subKey)
                         {
                             continue;
                         }
@@ -160,25 +169,22 @@ namespace System.Windows.Forms
                     }
                 }
             }
-            else
+
+            // Fallback to input language if keyboard language is not available
+            if (device == 0)
             {
-                // Device Handle contains keyboard layout Language Id.
-                // This is crucial when keyboard layout is installed more than once under different languages.
-                if (device != 0)
-                {
-                    language = device;
-                }
+                device = PARAM.LOWORD(hkl);
             }
 
-            // Pad with zeros to its left to produce arbitrary length string
-            return language.ToString("x8");
+            // Pad with zeros to its left to produce arbitrary length KLID string
+            return device.ToString("x8");
         }
 
         /// <summary>
         ///  Return localized keyboard layout name for provided KLID string.
         ///  See https://learn.microsoft.com/windows/win32/intl/using-registry-string-redirection#create-resources-for-keyboard-layout-strings
         /// </summary>
-        private static string GetLocalizedKeyboardLayoutName(string layoutName)
+        private static string GetKeyboardLayoutDisplayName(string layoutName)
         {
             if (KeyboardLayoutsRegistryKey?.OpenSubKey(layoutName) is RegistryKey key)
             {
--- a/src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
@@ -116,6 +116,7 @@ namespace System.Windows.Forms
         ///  Returns the KLID string for provided HKL handle.
         ///  Same as GetKeyboardLayoutName API but for any HKL.
         ///  See https://learn.microsoft.com/windows-hardware/manufacture/desktop/windows-language-pack-default-values
+        ///  for a list of possible KLID values.
         /// </summary>
         internal static string GetKeyboardLayoutNameForHKL(IntPtr hkl)
         {
@@ -124,20 +125,20 @@ namespace System.Windows.Forms
             // language; setting and resetting the current input language would generate
             // spurious InputLanguageChanged events.
             //
-            // According to the docs of GetKeyboardLayout API function high word of HKL
-            // contains a device handle to the physical layout of the keyboard
-            // but exact format of this handle is not documented.
+            // According to the GetKeyboardLayout API function docs low word of HKL
+            // contains input language but same keyboard can be installed more than once
+            // under different languages in Windows and it would give us wrong results
+            // in cases where for example French keyboard is installed under US input
+            // language etc.
             //
-            // We may try to use low word of HKL which contains input language but same
-            // keyboard can be installed more than once under different languages in Windows
-            // and it would give us wrong results in cases where French keyboard is
-            // installed under US input language etc.
+            // High word of HKL contains a device handle to the physical layout of
+            // the keyboard but exact format of this handle is not documented.
             //
-            // For older keyboards layouts device handle seems contain keyboard layout
-            // language which we can use as KLID without issues.
+            // For older keyboard layouts device handle seems contain keyboard layout
+            // language which we can use as KLID without any issues.
             int device = PARAM.HIWORD(hkl);
 
-            // But for newer keyboards layouts device handle contains layout id
+            // But for newer keyboard layouts device handle contains layout id
             // if its high nibble is 0xF. This id may be used to search for keyboard layout
             // under registry.
             // NOTE: this logic may break in future versions of Windows since its not documented.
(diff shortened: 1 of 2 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/winforms?pullRequest=8439) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/winforms?pullRequest=8439) before we can accept your contribution.<br/><br/>:x: DJm00n [sign now](https://cla.dotnetfoundation.org/dotnet/winforms?pullRequest=8439)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/winforms?pullRequest=8439) it.</sub>

elachlan:
@dreddy-work I think this is a flaky test.

elachlan:
@DJm00n would it be possible to write a test for this?

DJm00n:
@elachlan Good idea. I'll add some HKL examples to tests.

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
```diff
@@ -108,187 +108,107 @@ public string LayoutName
                 // GetKeyboardLayoutName does what we want, but only for the current input
                 // language; setting and resetting the current input language would generate
                 // spurious InputLanguageChanged events.
+                // Try to extract needed information manually from the HKL.
+                string layoutName = GetKeyboardLayoutNameForHKL(_handle);
 
-                /*
-                            HKL is a 32 bit value. HIWORD is a Device Handle. LOWORD is Language ID.
```
RussKie, line 113: Is this comment no longer valid or relevant?

DJm00n: There was too many irrelevant info.
I just left part of it in GetKeyboardLayoutNameForHKL:
```
// HKL is a 32 bit value.
// HIWORD is a Device Handle.
// LOWORD is Language Id.
int language = unchecked((int)(long)hkl) & 0xFFFF;
int device = (unchecked((int)(long)hkl) >> 16) & 0xFFFF;
```
Relevant claim about `Layout Id` is there too.

DJm00n:
@elachlan I have added test with some example HKL corner cases.

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
```diff
@@ -108,187 +108,107 @@ public string LayoutName
...
-                Name="4"  (may vary)
-                Value="d0000409"  -> Language ID = 0409
-                Look in HKEY_CURRENT_USER\Keyboard Layout\Substitutes
-                Name="d0000409"
-                Value="00010409"
-                Look in HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Keyboard Layouts\00010409
-                "Layout File": name of keyboard layout DLL (KBDDV.DLL)
-                "Layout Id": ID of this layout (0002)
-                Windows will change the top nibble of layout ID to F, which makes F002.
-                Combined with Language ID, the final HKL is F0020409.
```
RussKie, line 141: Why is this irrelevant? And where else can this information be found?

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
```diff
@@ -108,187 +108,107 @@ public string LayoutName
...
+        ///  Same as GetKeyboardLayoutName API but for any HKL.
+        ///  See https://learn.microsoft.com/windows-hardware/manufacture/desktop/windows-language-pack-default-values
+        /// </summary>
+        internal static string GetKeyboardLayoutNameForHKL(IntPtr hkl)
+        {
+            // HKL is a 32 bit value.
+            // HIWORD is a Device Handle.
+            // LOWORD is Language Id.
+            int language = unchecked((int)(long)hkl) & 0xFFFF;
+            int device = (unchecked((int)(long)hkl) >> 16) & 0xFFFF;
```
RussKie, line 132: I think these can be replaced with helpers from https://github.com/dotnet/winforms/blob/main/src/System.Windows.Forms.Primitives/src/Interop/Interop.PARAM.cs

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
DJm00n: In current version of code we don't look at `HKEY_CURRENT_USER\Keyboard Layout\Preload` or `HKEY_CURRENT_USER\Keyboard Layout\Substitutes` at all. `Layout Id` part is relevant but I have rewritten it a bit in a new code comment.

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
DJm00n: done

dreddy-work:
@tarekgh / @JeremyKuhne , can you help take a look at this change and see we are in right direction here?

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
```diff
@@ -104,191 +104,115 @@ public string LayoutName
...
-                        }
+                        return layoutText;
                     }
                 }
 
                 return SR.UnknownInputLanguageLayout;
             }
         }
 
+        private static RegistryKey? KeyboardLayoutsRegistryKey
```
JeremyKuhne, line 141: Don't put keys in a static field, they need closed immediately after use.

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
```diff
@@ -104,191 +104,115 @@ public string LayoutName
...
-                        ?? SR.UnknownInputLanguageLayout;
-                }
+                string layoutName = GetKeyboardLayoutNameForHKL(_handle);
 
-                // Look for a substitution
-                RegistryKey? substitutions = Registry.CurrentUser.OpenSubKey("Keyboard Layout\\Substitutes");
-                string[]? encodings = null;
-                if (substitutions is not null)
+                // https://learn.microsoft.com/windows/win32/intl/using-registry-string-redirection#create-resources-for-keyboard-layout-strings
+                if (KeyboardLayoutsRegistryKey?.OpenSubKey(layoutName) is RegistryKey key)
```
JeremyKuhne, line 110: All keys need put back into using statements so the handles will be closed. Right now this is leaking.

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
```diff
@@ -104,191 +104,115 @@ public string LayoutName
...
+
+                        if (subKey.GetValue("Layout Id") is not string subKeyLayoutId)
                         {
-                            return buffer.SliceAtFirstNull().ToString();
+                            continue;
+                        }
+
+                        if (layoutId == Convert.ToInt32(subKeyLayoutId, 16))
+                        {
+                            Debug.Assert(subKeyName.Length == 8, "unexpected key length in registry: " + subKey.Name);
```
JeremyKuhne, line 194: Please use interpolated strings as that will prevent formatting if the assert is valid.

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
DJm00n: done

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
DJm00n: done

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
DJm00n: done

DJm00n:
@JeremyKuhne any objections?

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
```diff
@@ -104,158 +106,35 @@ public string LayoutName
...
-                        int encodingValue = Convert.ToInt32(encoding, 16);
-                        if (encodingValue == unchecked((int)(long)currentHandle) ||
-                            (encodingValue & 0x0FFFFFFF) == (unchecked((int)(long)currentHandle) & 0x0FFFFFFF) ||
-                            (encodingValue & 0xFFFF) == language)
+                        // Localizable string resource associated with the keyboard layout
+                        if (subKey.GetValue("Layout Display Name") is string layoutDisplayName)
                         {
-                            string? encodingSubstitution = (string?)substitutions.GetValue(encoding);
-                            if (encodingSubstitution is null)
+                            unsafe
```
JeremyKuhne, line 119: nit: We prefer keeping the unsafe on the method to reduce nesting.

JeremyKuhne:
> @JeremyKuhne any objections?

@DJm00n I still don't see usings for all of the `RegistryKey` usages

DJm00n:
>@DJm00n I still don't see usings for all of the RegistryKey usages

Oh, I see what you mean. Sorry, it's been a long time since I wrote in C#. Is it better now?

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
DJm00n: I can do `public unsafe string LayoutName` but not sure about it...

RussKie: It aligns with how the rest of the code is done, so that's totally ok.

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
```diff
@@ -265,30 +142,79 @@ Language ID
         }
 
         /// <summary>
-        ///  Attempts to extract the localized keyboard layout name using the SHLoadIndirectString API.
-        ///  Returning null from this method will force us to use the legacy codepath (pulling the text
-        ///  directly from the registry).
+        ///  Returns the KLID string for provided HKL handle.
```
RussKie, line 145: What's "KLID" and "HKL"?

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
```diff
@@ -265,30 +142,79 @@ Language ID
...
+        ///  for a list of possible KLID values.
         /// </summary>
-        private static string? GetLocalizedKeyboardLayoutName(string? layoutDisplayName)
+        internal static string GetKeyboardLayoutNameForHKL(IntPtr hkl)
         {
-            if (layoutDisplayName is not null)
+            // There is no good way to do this in Windows.
+            // GetKeyboardLayoutName does what we want, but only for the current
+            // input language; setting and resetting the current input language
+            // would generate spurious InputLanguageChanged events.
```
RussKie, line 155: Please wrap in between 120-150 chars range, see "Editor guidelines" in https://github.com/dotnet/winforms/blob/main/docs/developer-guide.md.

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
DJm00n: KLID - keyboard layout ID string returned by [GetKeyboardLayoutName](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getkeyboardlayoutnamew) API. We cannot use this API directly because it returns only active KLID of the current thread but we need KLID of the particular HKL (that is contained in the InputLanguage object).
HKL - handle returned by [GetKeyboardLayout](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getkeyboardlayout) and [GetKeyboardLayoutList](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getkeyboardlayoutlist) APIs.

elachlan: Could you please update the comment to clarify this? That is what Russkie was implying.

DJm00n: I hope that it is better now

dreddy-work:
@xadxura, Can you take a look or share pointers on this change from windows OS perspective?

dreddy-work:
@DJm00n , just FYI, I am still waiting on Windows OS team to get some confirmation here. Will get this merged irrespective of response from windows by next week.

dreddy-work:
Taking this in Preview 1 as-is.  Have not heard any objections from Windows yet but we will keep an eye in Preview 1 usage to see how this change goes with LOC.

dreddy-work (approved):

PR #8449, opened 2023-01-04
Obsolete `ExceptionCollection` and replace with `AggregateException`

elachlan:
Based on feedback in #8413, this PR marks the `ExceptionCollection` as obsolete and replaces winforms usage with `AggregateException`.

## Customer Impact

- Anyone relying on catching `ExceptionCollection` will run into compile warnings and told to replace their usage
- 

## Regression? 

- Yes / No

## Risk

- Low risk, diagnostic should help customers migrate quite easily and provide better usability.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8449)

### DIFF
--- a/docs/list-of-diagnostics.md
+++ b/docs/list-of-diagnostics.md
@@ -38,7 +38,7 @@ The acceptance criteria for adding an obsoletion includes:
 |  __`WFDEV001`__ | Casting to/from IntPtr is unsafe, use `ResultInternal`. |
 |  __`WFDEV002`__ | `DomainUpDown.DomainUpDownAccessibleObject` is no longer used to provide accessible support for `DomainUpDown` controls. Use `ControlAccessibleObject` instead. |
 |  __`WFDEV003`__ | `DomainUpDown.DomainItemAccessibleObject` is no longer used to provide accessible support for `DomainUpDown` items. |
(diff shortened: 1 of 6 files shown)


elachlan:
image
I will change the message as its a bit redundant.

elachlan:
@dreddy-work this is what I was thinking for replacing `ExceptionCollection`

dreddy-work:
> @dreddy-work this is what I was thinking for replacing `ExceptionCollection`

Thank you @elachlan. The `ExceptionCollection `is exposed through the `DesignerHost`. I am not certain if custom designer developers handle this exception and parse the collection. I will discuss this in an internal meeting this week and provide an update on the next steps here.

merriemcgaw:
We've discussed this again as a team, and I'm so glad we brought it as its own PR to discuss separate from the other PR - thank you for the additional effort of separating it out! We looked at this PR and how the original code was used, and it turned out to be a bigger breaking change than we had anticipated. Given how breaking this change turned out to be, we're inclined to reject the proposal. The main reason here is that we don't want 3rd party ISVs to have to change how they approach managing exceptions in their designers (they've had to do enough with the move to .NET 😉). When we write new features in other contexts, they should totally use the AggregateException.

PR #8462, opened 2023-01-06
Move most of OLE code to CsWin32

JeremyKuhne:
This change removes a significant amount of our OLE wrappers to CsWin32. It also provides an IDispatch and IDispatchEx solution with adpaters for `IReflect` usages. Additionally it moves the VS interfaces to match the current interop patterns and fully documents them.

For consistency the various COM helper APIs have been renamed to `Try*` and non-Try variants. Try* variants will not throw- the equivalents without "Try" will validate `HRESULT`s and throw.

Some classes were completely moved to ComWrappers, notably `AxHost` nested classes: `AxHost.AxContainer`, `AxHost.AxContainer.ExtenderProxy`, `AxHost.EnumUnknown`, `AxHost.OleInterfaces`, `AxHost.VBFormat`, and `Control.ActiveXVerbEnum`

`StandardDispatch` is a new base class that you can derive from to provide `IDispatch` support for a class by utilizing predefined typelibs. This is our normal case when we're exposing `IDispatch`. It lets Windows create the `IDispatch` implementation (which is significantly complicated). It also allows for providing `IDispatchEx` support, but that must be manually provided. Derive from this base class and tag `IDispatch` and (optionally) `IDispatchEx` in `IManagedWrapper<>` to expose these interfaces. Currently we're just exposing `IUnknown`. This also will work with `IAccessible` and most other Win32 `IDispatch` interfaces by providing the typelib Guid.

`ClassPropertyDispatchAdapter` is a new helper class that facilitates exposing class public properties through `IDispatchEx`. It's used to replace `IReflect` usages on our COM visible classes. COM Interop uses `IReflect` to generate `IDispatchEx` when applied to non-public classes.

`using ComScope<T>` is the prescribed way to manage scoping COM pointers. Notes on their usage:

- When going out of scope, the held pointer, if any, will be `.Release()`ed.
- They are implicitly convertable to their pointers as well as `void*` and `void**` (for passing to APIs).
- To access the pointer you need to use `scope.Value->` (the `->` operator cannot be overloaded unfortunately).
- Like other wrappers you can use `.IsNull` to quickly check for null. Do not directly compare to `null`.

When you have to hold a COM pointer in a field, you should *always* use `AgileComPointer`. This provides a thread-safe wrapper with a finalizer if Dispose is not called correctly. The class now has additional helpers and debug disposal tracking.

Other work:

- Broke out a small number of classes into individual files (where there was no diff).
- Changed some files to file scoped namespaces to reduce nesting.
- `AxHost.ConnectionPointCookie` moved to `AgileComPointer` for safe finalization.
- Tweaked disposal tracking.

Notes:

The key in this PR was to remove most of the OLE defines and establish the patterns for exposing IDispatch/IDispatchEx. It is an important first step in continuing to convert off of built-in COM interop where possible.

We will never be able to be completely built-in COM compatible or remove every single usage of built-in COM. What we can do, however, is gracefully handle when built-in COM is not available and provide a way to opt in to "no built-in COM interop usage" from WinForms. We'll try to maintain compatibility with "normal" mainstream usages when built-in COM is available (for example, we'll continue to expose the OCX in AxHost as a built-in RCW). We won't, however, maintain compatibility for "technicalities". You could, for example, get a pointer we gave out to native through some other wrapper and expect to cast it back to the original managed object. This sort of thing is expected to be exceptionally rare and would now require using new features in the runtime such as:

https://github.com/dotnet/runtime/issues/79674

### Get a particular native interface from a managed object

``` C#
// ComScope variants
// -----------------

// This will throw when it fails.
using var unknown = ComHelpers.GetComScope<IUnknown>(obj);

// Does not throw.
using var unknown = ComHelpers.TryGetComScope<IUnknown>(obj, out HRESULT hr);
if (hr.Succeeded)
// ...

// Raw pointer variants
// --------------------

// This will throw when it fails.
*ppvObj = ComHelpers.GetComPointer<IVBFormat>(new VBFormat());

*ppvObj = ComHelpers.TryGetComPointer<IVBFormat>(new VBFormat(), out HRESULT hr);
```

### Get a ComScope from a native API

``` C#
using ComScope<IOleClientSite> clientSite = new(null);
HRESULT hr = oleObject.Value->GetClientSite(clientSite);
```

### Query another interface from an existing ComScope

```
using var oleObject = ComScope<IOleObject>.TryQueryFrom(pActiveObject, out HRESULT hr);
if (hr.Failed)
{
    // ...
}
```

### Get a managed wrapper for a COM interface

``` C#
// Currently this uses built-in COM, but will be tied into our ComWrappers implementation.
// When you can avoid it, try not to create a managed wrapper (RCW).

// IUnknown* unknown;
if (!ComHelpers.TryGetManagedInterface(
    unknown,
    takeOwnership: true, // Releases unknown
    out IDispatch.Interface dispatch)
// ...
```

### Pass a BSTR to/from a native API

``` C#
string value = "Foo";
BSTR bstrValue = new(value);
// Caller frees the BSTR
PInvoke.SomeApi(bstrValue);

// Using will free
using BSTR outBstrValue = default;
PInvoke.SomeOtherApi(&outBstrValue);
string outValue = outBstrValue.ToString();
```

### Enlist a class for a ComWrappers CCW

``` C#
internal class MyOleWindow : IOleWindow.Interface, IManagedWrapper<IOleWindow>
{
    // ...
}

// Any exposed CsWin32 interface needs tagged with the following for now.
// https://github.com/microsoft/CsWin32/issues/831 tracks getting CsWin32 to do this.
internal unsafe partial struct IOleWindow : IVTable<IOleWindow, IOleWindow.Vtbl> {}
```

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8462)


JeremyKuhne:
cc: @AArnott @AaronRobinsonMSFT @jkoritzinsky

src/System.Windows.Forms.Primitives/src/Interop/WinFormsComWrappers.cs
```diff
@@ -95,5 +95,35 @@ protected override void ReleaseObjects(IEnumerable objects)
...
+                {
+                    return HRESULT.COR_E_OBJECTDISPOSED;
+                }
+
+                action(@object);
+                return HRESULT.S_OK;
+            }
+            catch (Exception ex)
+            {
+                return ex;
```
AaronRobinsonMSFT, line 125: Shouldn't this be `ex.HResult`? I don't understnad how this is compiling.

JeremyKuhne: Implict conversion.

AArnott:
It's interesting that with all the deleted code, the PR still *adds* more lines than it deletes.

JeremyKuhne:
> It's interesting that with all the deleted code, the PR still adds more lines than it deletes.

That is partially due to replacement of COM Interop functionality (notably IDispatch/IDispatchEx). I think the need to expose Windows interfaces with IDispatch will be similar to what we're doing- where the Windows implementation is all that is needed.

Note that the really tough part is trying to implement ITypeInfo for IDispatch. I wasn't looking forward to trying to write that and was quite thrilled to find I could rely on Windows to do the heavy lifting.

src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/GlobalInterfaceTable.cs
```diff
@@ -5,66 +5,65 @@
...
-    /// <summary>
-    ///  Wrapper for the COM global interface table.
-    /// </summary>
-    internal static unsafe class GlobalInterfaceTable
-    {
-        private static readonly IGlobalInterfaceTable* s_globalInterfaceTable;
+    private static readonly IGlobalInterfaceTable* s_globalInterfaceTable;
 
-        static GlobalInterfaceTable()
+    static GlobalInterfaceTable()
```
kant2002, line 17: Question. Do this still needed given that Win7 is not supported anymore?

JeremyKuhne: Yes, because we're still not allowed to deliberately break Win7. We could conditionalize, but I don't think there is any benefit to the newer method (which just uses the GIT) for our purposes.

dreddy-work (commented):
Still trying to understand the change. Partially reviewed. Will go over rest tomorrow.

src/System.Windows.Forms/src/System/Windows/Forms/ComponentModel/COM2Interop/COM2IPerPropertyBrowsingHandler.cs
```diff
@@ -7,160 +7,159 @@
...
-                Debug.Fail($"An exception occurred inside IPerPropertyBrowsing::GetPredefinedStrings(dispid={sender.DISPID}), object type={ComNativeDescriptor.GetClassName(ppb)}");
-            }
+        HRESULT hr;
+        try
+        {
+            hr = ppb.GetPredefinedStrings((int)sender.DISPID, &caStrings, &caCookies);
+        }
+        catch (ExternalException ex)
+        {
+            hr = (HRESULT)ex.ErrorCode;
```
dreddy-work, line 124: nit: I don't see this assignment necessary.

src/System.Windows.Forms/src/System/Windows/Forms/ComponentModel/COM2Interop/COM2IManagedPerPropertyBrowsingHandler.cs
```diff
@@ -151,121 +124,129 @@ internal static unsafe Attribute[] GetComponentAttributes(VSSDK.IVSMDPerProperty
...
+
+        string[] strings = new string[count];
+        for (int i = 0; i < count; i++)
+        {
+            strings[i] = bstrs[i].ToStringAndFree();
+        }
 
+        try
+        {
+            Marshal.FreeCoTaskMem((nint)(void*)bstrs);
```
dreddy-work, line 215: Q: can this method be simplified with `using` on the caller input? Or this free is doing more than that?

dreddy-work (commented):
RLGTM.

src/System.Windows.Forms/src/System/Windows/Forms/ComponentModel/COM2Interop/COM2IManagedPerPropertyBrowsingHandler.cs
JeremyKuhne: This is an array of BSTRs. Each needs freed individually, then the array itself needs freed.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
```diff
@@ -3125,31 +3102,32 @@ public unsafe bool HasPropertyPages()
             }
         }
 
-        private unsafe void ShowPropertyPageForDispid(Ole32.DispatchID dispid, Guid guid)
+        private unsafe void ShowPropertyPageForDispid(int dispid, Guid guid)
         {
-            IntPtr pUnk = Marshal.GetIUnknownForObject(GetOcx());
-            try
+            ComScope<IUnknown> unknown = ComHelpers.TryGetComScope<IUnknown>(_instance, out HRESULT hr);
```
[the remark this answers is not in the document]
JeremyKuhne: Yes, I missed the using here. Good catch.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
```diff
@@ -3092,29 +3056,42 @@ private void StopEvents()
...
 
-        private bool CanShowPropertyPages() => GetOcState() >= OC_RUNNING && GetOcx() is Ole32.ISpecifyPropertyPages;
+        private bool CanShowPropertyPages()
+        {
+            if (GetOcState() < OC_RUNNING)
+            {
+                return false;
+            }
+
+            using var pages = ComHelpers.TryGetComScope<ISpecifyPropertyPages>(_instance, out HRESULT hr);
```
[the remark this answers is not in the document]
JeremyKuhne: I have such a method in my follow up change.

src/System.Windows.Forms/src/System/Windows/Forms/ComponentModel/COM2Interop/COM2IPerPropertyBrowsingHandler.cs
JeremyKuhne: We check hr against S_OK a few lines down.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/IExtender.cs
```diff
@@ -0,0 +1,249 @@
...
+
+        BSTR Name { get; }
+
+        IUnknown* Parent { get; }
+
+        HWND Hwnd { get; }
+
+        IUnknown* Container { get; }
+
+        // This isn't even utilized
```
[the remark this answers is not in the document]
JeremyKuhne: Yes.

dreddy-work (approved):

PR #8467, opened 2023-01-06
Fix read only static member evaluation order

dreddy-work:
There is a problem with the parsing of the TargetFramework, which is causing an invalid value to be set. static nullable private property needs to be assigned before its use.

Because the AppContext flags are read-only for .NET apps, and the incorrect TargetFramework is bypassing the logic behind these flags.

Found while verifying #6381. 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8467)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
@@ -31,9 +31,6 @@ namespace System.Windows.Forms.Primitives
                 s_targetFrameworkName ??= AppContext.TargetFrameworkName is { } name ? new(name) : null;
                 return s_targetFrameworkName;
             }
-
-            // Used in tests.
-            private set => s_targetFrameworkName = value;
         }
 
         /// <summary>
--- a/src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
@@ -21,6 +21,9 @@ namespace System.Windows.Forms.Primitives
         private static int s_scaleTopLevelFormMinMaxSizeForDpi;
         private static int s_trackBarModernRendering;
 
+        private static readonly FrameworkName? s_targetFrameworkName = GetTargetFrameworkName();
+        private static readonly bool s_isNetCoreApp = (s_targetFrameworkName?.Identifier) == ".NETCoreApp";
+
         /// <summary>
         ///  Indicates whether AnchorLayoutV2 feature is enabled.
         /// </summary>
@@ -113,10 +116,6 @@ namespace System.Windows.Forms.Primitives
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
             get => GetCachedSwitchValue(ScaleTopLevelFormMinMaxSizeForDpiSwitchName, ref s_scaleTopLevelFormMinMaxSizeForDpi);
         }
-
-        private static readonly bool s_isNetCoreApp = (s_targetFrameworkName?.Identifier) == ".NETCoreApp";
-
-        private static readonly FrameworkName? s_targetFrameworkName = GetTargetFrameworkName();
 
         public static bool TrackBarModernRendering
         {
--- a/src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
@@ -16,13 +16,30 @@ namespace System.Windows.Forms.Primitives
         private const string ScaleTopLevelFormMinMaxSizeForDpiSwitchName = "System.Windows.Forms.ScaleTopLevelFormMinMaxSizeForDpi";
         internal const string AnchorLayoutV2SwitchName = "System.Windows.Forms.AnchorLayoutV2";
         internal const string TrackBarModernRenderingSwitchName = "System.Windows.Forms.TrackBarModernRendering";
-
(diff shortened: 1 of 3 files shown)


dreddy-work:
@Amy-Li03 , can you cross verify issue with these privates?

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -21,6 +21,9 @@ internal static partial class LocalAppContextSwitches
         private static int s_scaleTopLevelFormMinMaxSizeForDpi;
         private static int s_trackBarModernRendering;
 
+        private static readonly FrameworkName? s_targetFrameworkName = GetTargetFrameworkName();
```
JeremyKuhne, line 24: This could stand some comments.

``` C#
/// <summary>
///  The <see cref="TargetFrameworkAttribute"/> value for the entry assembly, if any.
/// </summary>
public static FrameworkName? TargetFrameworkName { get; }
    = AppContext.TargetFrameworkName is { } name ? new(name) : null;

// IsNetCoreApp *must* come after TargetFrameworkName as it depends on it.

/// <summary>
///  Returns <see langword="true"/> if the <see cref="TargetFrameworkAttribute"/> value for the entry assembly specifies .NET Core.
/// </summary>
public static bool IsNetCoreApp { get; } = string.Equals(TargetFrameworkName?.Identifier, ".NETCoreApp");
```

JeremyKuhne:
Can you unit test that this comes out true in tests?

RussKie:
Why not use a static .ctor? That will ensure statics are initialised in the right order, instead of relying on comments.

dreddy-work:
> Why not use a static .ctor? That will ensure statics are initialised in the right order, instead of relying on comments.

We could. Moved the initialization to a property in order to avoid reliance on the order of initialization, even if it causes a slight performance hit with string comparison once per each switch.

Amy-Li03:
@dreddy-work verified issue: [6381](https://github.com/dotnet/winforms/issues/6381) with these privates, issue was fixed, the Anchors of user control works well when scaling on high DPI.
Fixed

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -16,10 +16,27 @@ internal static partial class LocalAppContextSwitches
...
-
         private static int s_AnchorLayoutV2;
         private static int s_scaleTopLevelFormMinMaxSizeForDpi;
         private static int s_trackBarModernRendering;
+        private static FrameworkName? s_targetFrameworkName;
+
+        /// <summary>
+        ///  The <see cref="TargetFrameworkAttribute"/> value for the entry assembly, if any.
+        /// </summary>
+        public static FrameworkName? TargetFrameworkName
```
JeremyKuhne, line 27: Why have a backing field? This is only one line, you can do `{ get; } =` as I described.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -16,10 +16,27 @@ internal static partial class LocalAppContextSwitches
...
+            {
+                s_targetFrameworkName ??= AppContext.TargetFrameworkName is { } name ? new(name) : null;
+                return s_targetFrameworkName;
+            }
+        }
+
+        /// <summary>
+        ///  Returns <see langword="true"/> if the <see cref="TargetFrameworkAttribute"/> value for the entry assembly specifies .NET Core.
+        /// </summary>
+        public static bool IsNetCoreApp => string.Equals(TargetFrameworkName?.Identifier, ".NETCoreApp");
```
JeremyKuhne, line 39: This is fine I suppose. Was trying to avoid doing this string comparison for every switch population, but we only have one right now.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
dreddy-work: I am using backing field in tests. Test accessor doesn't get propertie's private set if I go that route. Do you have any sggestion on how i can set this  in tests?

JeremyKuhne: You should be able to get the generated backing field with the test accessor. The name should be `<TargetFrameworkName>k__BackingField` I believe.

src/System.Windows.Forms.Primitives/tests/UnitTests/System/Windows/Forms/LocalAppContextSwitchesTest.cs
```diff
@@ -0,0 +1,66 @@
+﻿// Licensed to the.NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.Versioning;
+using System.Windows.Forms.Primitives;
+
+namespace System.Windows.Forms.Tests
+{
+    public class LocalAppContextSwitchesTest
```
JeremyKuhne, line 10: This is going to need to be put into a collection to avoid colliding with other tests.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
dreddy-work: Did not make any change as its going to be cleaned up with this: https://github.com/dotnet/winforms/issues/8476.

JeremyKuhne (approved):

PR #8470, opened 2023-01-09
Refactor collection remove methods to fix ownership issues

### LINKED ISSUE #2750: Collection `Remove` methods are sometimes missing ownership checks.
* .NET Core Version: master
* Have you experienced this same bug with .NET Framework?: didn't check but probably has the same bugs

**Problem description:**
Some custom `IList` implementations don't check in their `Remove` implementations whether the removed object is actually part of the list. They just go ahead and start disconnecting the object passed as argument, even if it belongs to a different container. The object will probably end up in a partially disconnected state without its actual owner being notified.

This was discovered while trying to implement generic `IList<T>` on those collections, where `Remove` is required to report success. It is immediately noticeable if a method doesn't have a branch that returns false (#2644 and #2749)

**Expected behavior:**
`Remove` methods should check if the object to be removed is actually part of the collection before they start disconnecting the object.

**Minimal repro:**
discovered on following types via code review while implementing generic `Remove` for generic `IList<T>`:
* [ListViewGroupCollection](https://source.dot.net/#System.Windows.Forms/System/Windows/Forms/ListViewGroupCollection.cs,295)
* [TableLayoutStyleCollection](https://source.dot.net/#System.Windows.Forms/System/Windows/Forms/TableLayoutStyleCollection.cs,90)
* [ToolStripItemCollection](https://source.dot.net/#System.Windows.Forms/System/Windows/Forms/ToolStripItemCollection.cs,441)

There might be more, will do a more thorough review later while working on the generic IList implementation, just want to document the bugs for now.

elachlan:
Use List<T>.Remove return value to confirm ownership and then return if false to avoid any clean-up that should not take place.

Fixes #2750

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8470)

### DIFF
--- a/src/System.Windows.Forms/src/PublicAPI.Shipped.txt
+++ b/src/System.Windows.Forms/src/PublicAPI.Shipped.txt
@@ -1388,9 +1388,9 @@ System.Windows.Forms.MonthCalendar.SelectionRange.set -> void
 ~System.Windows.Forms.TableLayoutRowStyleCollection.Remove(System.Windows.Forms.RowStyle rowStyle) -> void
 ~System.Windows.Forms.TableLayoutRowStyleCollection.this[int index].get -> System.Windows.Forms.RowStyle
 ~System.Windows.Forms.TableLayoutRowStyleCollection.this[int index].set -> void
-~System.Windows.Forms.TableLayoutStyleCollection.Add(System.Windows.Forms.TableLayoutStyle style) -> int
-~System.Windows.Forms.TableLayoutStyleCollection.this[int index].get -> System.Windows.Forms.TableLayoutStyle
-~System.Windows.Forms.TableLayoutStyleCollection.this[int index].set -> void
+System.Windows.Forms.TableLayoutStyleCollection.Add(System.Windows.Forms.TableLayoutStyle! style) -> int
+System.Windows.Forms.TableLayoutStyleCollection.this[int index].get -> System.Windows.Forms.TableLayoutStyle!
+System.Windows.Forms.TableLayoutStyleCollection.this[int index].set -> void
 System.Windows.Forms.ToolStrip.Controls.get -> System.Windows.Forms.Control.ControlCollection!
 ~System.Windows.Forms.ToolStrip.GetChildAtPoint(System.Drawing.Point point) -> System.Windows.Forms.Control
 ~System.Windows.Forms.ToolStrip.GetChildAtPoint(System.Drawing.Point pt, System.Windows.Forms.GetChildAtPointSkip skipValue) -> System.Windows.Forms.Control
--- a/src/System.Windows.Forms/src/Resources/SR.resx
+++ b/src/System.Windows.Forms/src/Resources/SR.resx
@@ -6958,4 +6958,7 @@ Stack trace where the illegal operation occurred was:
   <data name="Arg_RankMultiDimNotSupported" xml:space="preserve">
     <value>Only single dimensional arrays are supported for the requested action.</value>
   </data>
+  <data name="InvalidArgumentType" xml:space="preserve">
+    <value>Argument '{0}' must be of type '{1}'.</value>
+  </data>
 </root>
--- a/src/System.Windows.Forms/src/Resources/xlf/SR.cs.xlf
+++ b/src/System.Windows.Forms/src/Resources/xlf/SR.cs.xlf
@@ -5831,6 +5831,11 @@ Trasování zásobníku, kde došlo k neplatné operaci:
         <target state="translated">InvalidArgument=Hodnota {1} není platnou hodnotou pro argument {0}.</target>
         <note />
       </trans-unit>
+      <trans-unit id="InvalidArgumentType">
+        <source>Argument '{0}' must be of type '{1}'.</source>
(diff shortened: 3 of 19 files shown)


elachlan:
@merriemcgaw Passing an non `TableLayoutStyle` now doesn't cause an `InvalidCastException`, its just ignored. Did we want to throw an `ArgumentException` instead of ignoring it?

dreddy-work:
/cc: @weltkante. 

The removal process in question does not indicate whether it was successful or not and as a result, it is now placing the responsibility on the caller to determine if the operation was successful. At this point, i am not sure how this is going to impact existing applications but seems reasonable to go in this direction.

Are there any other situations in WinForms where the success or failure of an operation is not explicitly indicated?

/cc: @Tanya-Solyanik / @JeremyKuhne

weltkante:
> it is now placing the responsibility on the caller to determine if the operation was successful

It always was the responsibility of the caller to remove elements only from collections they are contained in, but not in all cases the caller may have noticed if they were doing something incorrectly.

So while this could affect applications that were accidently using the API incorrectly and didn't notice, they were already broken and now remain broken in a different, possibly more visible way. Documenting this as "breaking change" should be sufficient. I don't see a simple way to retain full compatibility with this kind of broken callers.

Applications that were using the API correctly shouldn't be affected.

> Are there any other instances like these in WinForms?

I did a fairly complete review of all internal collections and the ones listed in the issue were the only ones I found that behaved incorrectly.

dreddy-work:
> @merriemcgaw Passing an non `TableLayoutStyle` now doesn't cause an `InvalidCastException`, its just ignored. Did we want to throw an `ArgumentException` instead of ignoring it?

We have the option to update our tests, but thinking again, it raises the question of whether exceptions should be thrown in other cases as well in order to make the outcome of the operation more explicit to the caller?

elachlan:
> We have the option to update our tests, but thinking again, it raises the question of whether exceptions should be thrown in other cases as well in order to make the outcome of the operation more explicit to the caller?

I'd defer to the team for guidance on this. 

When these were written .NET lacked generics, so its just following the interface. The next best thing after a typed API is throwing an exception where you expect a certain type. If we don't throw an exception, someone might be confused as to why the call to Remove works. 

If they had an object they passed to remove, it would have errored if they manually added it. So it would have to be a mistake on their part to call remove on an object that cannot be removed. It might be a bad idea to be silent in this case. Because the developer might have passed in the wrong object.

Bugs like that are a pain because you don't have a compile error or a runtime error, its just silent.

JeremyKuhne:
In every case we have we should individually consider the expected impact of throwing where we didn't throw before. For these "typed" collections I think it is fair to introduce argument exceptions, particularly if having the bad value would have caused an exception further down the line (by said bad value in the collection). That leaves us with trying to remove an item that isn't in the collection- if that didn't throw before I think it is reasonable to still have it not throw (but we should have debug asserts).

Adding the exceptions here is in line with the sort of breaking changes we were trying to enable in .NET Core.

elachlan:
@JeremyKuhne If the remove was previously throwing an `InvalidCastException` would it be correct to introduce an `ArguementException` as a replacement? 

For cases where the remove wasn't throwing previously, I agree that a debug.assert would be a good idea.

dreddy-work:
> @JeremyKuhne If the remove was previously throwing an `InvalidCastException` would it be correct to introduce an `ArguementException` as a replacement?
> 

`ArguementException` is more suitable her and that should be fine. We can document it.

JeremyKuhne:
> ArguementException is more suitable her and that should be fine. We can document it.

Agreed. It is completely legitimate to change invalid input data exceptions to `ArgumentException` (or a derivative).

elachlan:
`ListView.SelectedIndexCollection` Remove/Add both raise an `ArgumentException`
https://github.com/dotnet/winforms/blob/94df74558df86435dccc2d3c302422ca6759eb65/src/System.Windows.Forms/src/System/Windows/Forms/ListView.SelectedIndexCollection.cs#L221-L231

An InvalidCastException tells us what type it should be, the new `ArgumentException` uses `SR.InvalidArgument` which might not be as explicit.

dreddy-work:
> An InvalidCastException tells us what type it should be, the new `ArgumentException` uses `SR.InvalidArgument` which might not be as explicit.

Feel free to add a new string `SR.InvalidArgumentType` that can be used for the cases like these.

elachlan:
Thanks, I think that is the best way forward if we are not converting to generics.

elachlan:
image
Would you like the fully qualified type name, or just the type name?

dreddy-work:
>  Would you like the fully qualified type name, or just the type name?

Qualified type name would unambiguous.

dreddy-work (approved):

elachlan:
Test run is stuck.

dmitrii-drobotov (approved):
LGTM! Should this PR also have a 'documentation: breaking' label?

dreddy-work:
> LGTM! Should this PR also have a 'documentation: breaking' label?

yep.

dreddy-work:
https://github.com/dotnet/docs/issues/33694

Issue #4463, opened 2021-01-13
The code generated by the ListBox control is not consistent with Framework


Tanya-Solyanik:
* .NET Core Version: 
3.1+

* Have you experienced this same bug with .NET Framework?: 
No

**Problem description:**

The generated code for ListBox1 have extra code: this.listBox1.ItemHeight = 15;
![image](https://user-images.githubusercontent.com/15823268/104525385-e63a8080-55b4-11eb-8018-1cd886b658a0.png)

**Expected behavior:**
The generated code for ListBox1 should not have this code: this.listBox1.ItemHeight = 15;
![image](https://user-images.githubusercontent.com/15823268/104525352-d1f68380-55b4-11eb-9bf4-fafbf11c85f4.png)

**Minimal repro:**
1. Create a Core winforms application.
2. Add a **ListBox** control to **form**.
3. Double click the **Form1.Designer.cs** file in Solution Explorer.
4. Observe the generated code of the **ListBox**.
Result: The generated code for ListBox1 should not have this code: this.listBox1.ItemHeight = 15;

Note from Dustin: ListBox.ItemHeight property. 
dotnet/winforms@4bca5b3/src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs#L596-L616
It has a `[DefaultValue(DefaultItemHeight)]` attribute and DefaultItemHeight is 13.
Note from Tanya: Our default value is not accurate anymore due to the default font change. 13 is the default for Microsoft Sans Serif, 8.25pt, 15 is what comctl32 listbox calculates for Segoe UI, 9pt. To confirm - change the parent control's font.

RussKie:
@DustinCampbell @Tanya-Solyanik  I don't think there's a fix for this.

We can't change `DefaultItemHeight` because it is a public const, any library/app that has referenced it must be recompiled.
https://github.com/dotnet/winforms/blob/4bca5b30aea871a2c1761b39f8ff8128cb155119/src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs#L52-L56

I don't think we can divorce `ItemHeight` from `DefaultItemHeight` either, as it would mess with the designer, and create a mental disconnect for a developer.
https://github.com/dotnet/winforms/blob/4bca5b30aea871a2c1761b39f8ff8128cb155119/src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs#L601-L605

weltkante:
[...]

PR #8474, opened 2023-01-09
Removed DefaultValue attribute from the RowTemplate.Height property

### LINKED ISSUE #4461: The code generated by the DataGridView control has extra code for the RowTemplate.Height property
* .NET Core Version: 
3.1+

No

**Problem description:**

The generated code for dataGirdView1 has extra code for the RowTemplate.Height property. For the newly created control *at 100% scaling*, designer should generate the same value as is defined by the DefaultValueAttribute

image

**Expected behavior:**
RowTemplate.Height line should not be generated
 
image

**Minimal repro:**
1. Create a Winforms .Net Core application.
2. Add a DataGridView to from designer.
3. Double click to open Form1.Designer.cs file in Solution Explorer.
4. Observe the generated code of the dataGridView1.

Note from Dustin: DataGridViewRow.Height has a [DefaultValue(22)] attribute, so the code-gen above is expected if the value is 25.

dkazennov:
Fixes #4461

## Proposed changes

- Introduces a new private property `DefaultHeight` with **the intended value** (by default in Winforms Control.DefaultFont is **Segoe UI, Size 9, Height 16** which gives us **default Row height = 25**).
- Removes the DefaultValue attribute from the `DataGridViewRow.Height` property;
- Implements `ResetHeight` and `ShouldSerializeHeight` methods ([see more](https://learn.microsoft.com/en-us/dotnet/desktop/winforms/controls/defining-default-values-with-the-shouldserialize-and-reset-methods?view=netframeworkdesktop-4.8)).
- The `DataGridViewRow.Height` property still returns the value of the  `Thickness` property.
- Existing unit tests weren't changed.
- New unit test `DataGridViewRow_Height_NotEqualDefaultHeight_IfDefaultFontIsChanged` is added.

## Customer Impact

- Fixes the Issue which impacts user experience,  breaks DPI and Font scaling. For the _newly created_ control at 100% scaling, designer should generate the same value as is defined by the Default. There is no extra code.

## Regression? 

- No

## Risk

- Minimal

### Before

image

### After

image

## Test methodology

- Manual (Issue can be tested via substitution of dlls, complete restart of the VS recommended)
- Unit test

 

## Test environment(s) 

- dotnet 8.0.100-alpha.1.22607.6

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8474)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRow.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRow.cs
@@ -19,6 +19,7 @@ namespace System.Windows.Forms
     public partial class DataGridViewRow : DataGridViewBand
     {
         private static readonly Type s_rowType = typeof(DataGridViewRow);
+        private static int s_defaultHeight = -1;
         private static readonly int s_propRowErrorText = PropertyStore.CreateKey();
         private static readonly int s_propRowAccessibilityObject = PropertyStore.CreateKey();
 
@@ -34,7 +35,7 @@ namespace System.Windows.Forms
         public DataGridViewRow() : base()
         {
             MinimumThickness = DefaultMinRowThickness;
-            Thickness = Control.DefaultFont.Height + 9;
+            Thickness = DefaultHeight;
         }
 
         [Browsable(false)]
@@ -151,6 +152,19 @@ namespace System.Windows.Forms
             set => ErrorTextInternal = value;
         }
 
+        private static int DefaultHeight
+        {
+            get
+            {
+                if (s_defaultHeight == -1)
+                {
+                    s_defaultHeight = Control.DefaultFont.Height + 9;
+                }
+
+                return s_defaultHeight;
+            }
+        }
+
         private string ErrorTextInternal
         {
             get
@@ -209,7 +223,6 @@ namespace System.Windows.Forms
             set => base.HeaderCellCore = value;
         }
 
-        [DefaultValue(22)]
         [NotifyParentProperty(true)]
         [SRCategory(nameof(SR.CatAppearance))]
         [SRDescription(nameof(SR.DataGridView_RowHeightDescr))]
@@ -1730,6 +1743,12 @@ namespace System.Windows.Forms
             }
         }
 
+        // ShouldSerialize and Reset Methods are being used by Designer via reflection.
+        private void ResetHeight()
+        {
+            Height = DefaultHeight;
+        }
+
         internal void ReleaseUiaProvider()
         {
             if (!IsAccessibilityObjectCreated)
@@ -1807,6 +1826,12 @@ namespace System.Windows.Forms
             }
 
             return setResult && values.Length <= cellCount;
+        }
+
+        // ShouldSerialize and Reset Methods are being used by Designer via reflection.
+        private bool ShouldSerializeHeight()
+        {
+            return Height != DefaultHeight;
         }
 
         public override string ToString()
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewRowTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewRowTests.cs
@@ -2020,6 +2020,30 @@ namespace System.Windows.Forms.Tests
         }
 
         [WinFormsFact]
+        public void DataGridViewRow_Height_IsEqualDefaultHeight_IfDefaultFontIsChanged()
+        {
+            var oldApplicationDefaultFont = Application.DefaultFont;
+            using var font = new Font("Times New Roman", 12);
+            using var row = new TestDataGridViewRow();
+            var applicationTestAccessor = typeof(Application).TestAccessor().Dynamic;
+
+            try
+            {
+                applicationTestAccessor.s_defaultFont = font;
+                Assert.NotEqual(oldApplicationDefaultFont, Application.DefaultFont);
+
+                var rowDefaultHeight = row.GetDefaultHeight();
+                Assert.Equal(Control.DefaultFont.Height + 9, rowDefaultHeight);
+                Assert.Equal(Control.DefaultFont.Height + 9, row.Height);
+            }
+            finally
+            {
+                applicationTestAccessor.s_defaultFont = null;
+                applicationTestAccessor.s_defaultFontScaled = null;
+            }
+        }
+
+        [WinFormsFact]
         public void DataGridViewRow_Height_SetShared_ThrowsInvalidOperationException()
         {
             using var control = new DataGridView
@@ -5791,6 +5815,14 @@ namespace System.Windows.Forms.Tests
                 DataGridViewRowAccessibleObject accessibilityObject = Assert.IsType<DataGridViewRowAccessibleObject>(CreateAccessibilityInstance());
                 Assert.Equal(this, accessibilityObject.Owner);
             }
+        }
+    }
+
+    internal class TestDataGridViewRow : DataGridViewRow
+    {
+        internal int GetDefaultHeight()
+        {
+            return this.TestAccessor().Dynamic.DefaultHeight;
         }
     }
 }


dkazennov:
If this approach suits us we can fix https://github.com/dotnet/winforms/issues/4463 the same way.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRow.cs
```diff
@@ -25,8 +25,10 @@ public partial class DataGridViewRow : DataGridViewBand
         private const DataGridViewAutoSizeRowCriteriaInternal InvalidDataGridViewAutoSizeRowCriteriaInternalMask = ~(DataGridViewAutoSizeRowCriteriaInternal.Header | DataGridViewAutoSizeRowCriteriaInternal.AllColumns);
 
         private const int DefaultMinRowThickness = 3;
+        private const int DefaultHeight = 22;
```
merriemcgaw, line 28: Is this 22 the old default height from Framework, or is it taking into consideration the new default font and the height that should be associated with that?

merriemcgaw (changes requested):

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRow.cs
dkazennov: > Is this 22 the old default height from Framework, or is it taking into consideration the new default font and the height that should be associated with that?

This is default value which was hardcoded in the DefaultValue attribute. Given Issue description and the way to fix the similar problem proposed by @RussKie I thought we need to preserve this value as a private constant and avoid rely on default font height.

dmitrii-drobotov: I would go with 25 to keep the appropriate height for current font and to not change .NET Core+ apps which already had the default height as 25. And also I'm not sure that using a fixed value is the best approach, because if there is a [custom DefaultFont](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props-desktop#applicationdefaultfont), then current implementation would adjust height accordingly, but with this change it'd break and require setting a new height.

dkazennov: > I would go with 25 to keep the appropriate height for current font and to not change .NET Core+ apps which already had the default height as 25. And also I'm not sure that using a fixed value is the best approach, because if there is a [custom DefaultFont](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props-desktop#applicationdefaultfont), then current implementation would adjust height accordingly, but with this change it'd break and require setting a new height.

Yes, indeed. But there are also some other problems here.
Existing realization uses DataGridViewBand.Thickness as a value for DataGridViewRow.Height, and DataGridViewBand.Thickness (the property of the base class) has complex set and get logic.  This is why 21 unit tests failed with this fix.

dkazennov:
This issue raises the question : should we even made the Height property of the DVG row a constant as @RussKie [proposed](https://github.com/dotnet/winforms/issues/4463#issuecomment-768831519) or should we stick to font height value? 
Given the task in the Issue I believed we should made the Property value to be equal the constant by default.

Failing unit tests made me believe that instead of changing the DataGridViewRow.Height we should probably change DataGridViewBand.Thickness (but this will affect many other components of DataGridView I'm afraid).

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRow.cs
dkazennov: > I would go with 25 to keep the appropriate height for current font and to not change .NET Core+ apps which already had the default height as 25. And also I'm not sure that using a fixed value is the best approach, because if there is a [custom DefaultFont](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props-desktop#applicationdefaultfont), then current implementation would adjust height accordingly, but with this change it'd break and require setting a new height.

The Control.DefaultFont is either Application.DefaultFont or SystemFonts.MessageBoxFont. I have Application.DefaultFont "Calibri" (Size=11 & **Height = 23**), so in my case DataGridViewBand.Thickness should be 32. The Idea with private constant value in a similar situation was initially proposed by @RussKie [here](https://github.com/dotnet/winforms/issues/4463#issuecomment-768831519). 

What bothers me here is the value returned (by Application.DefaultFont) will vary depending on the user's operating system the local culture setting of their system.

dkazennov:
This way we:
1) check for changes in Control.DefaultFont.Height;
2) get rid of extra code as intended by the issue;
3) all unit tests passes (therefore we make sure the Height has indented value);

This may looks not exactly the perfect solutions but it is best we can get - as @RussKie [pointed out](https://github.com/dotnet/winforms/issues/4463#issuecomment-768831519).

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRow.cs
```diff
@@ -1809,6 +1820,11 @@ internal bool SetValuesInternal(params object[] values)
             return setResult && values.Length <= cellCount;
         }
 
+        private bool ShouldSerializeHeight()
+        {
+            return _height != DefaultHeight;
```
dmitrii-drobotov, line 1825: Can it be just `Height != DefaultHeight;`? That way `_height` variable could be removed.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRow.cs
```diff
@@ -1730,6 +1736,11 @@ internal void OnSharedStateChanging(int sharedRowIndex, DataGridViewElementState
             }
         }
 
+        private void ResetHeight()
```
dmitrii-drobotov, line 1739: Since this is a private method without any explicit usages, I think someone could remove it by mistake. Consider adding a comment that this method is used by the designer via reflection. And the same for `ShouldSerializeHeight`

dmitrii-drobotov (commented):
Looks good!

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRow.cs
dkazennov: > Can it be just `Height != DefaultHeight;`? That way `_height` variable could be removed.

The trick here is that without this assignment 

> Can it be just `Height != DefaultHeight;`? That way `_height` variable could be removed.

The trick here is that without this assignment there would be extra code `dataGridView1.RowTemplate.Height`

dkazennov:
Some builds in the pipeline has faulty tests (and those tests have nothing to do with this fix or the `DataGridViewRow.Height` property in particular):

    DataGridView_ColumnHeadersHeight_SetWithHandle_GetReturnsExpected
    DataGridView_ColumnHeadersHeight_SetWithParentWithHandle_GetReturnsExpected
    DataGridView_RowHeadersWidthSizeMode_SetWithHandle_GetReturnsExpected
    DataGridView_ColumnHeadersHeightSizeMode_SetNonResizeThenResize_RestoresOldValue
    DataGridView_RowHeadersWidthSizeMode_SetNonResizeThenResize_RestoresOldValue
    
According to [Runfo application](https://runfo.azurewebsites.net/search/tests/?q=started%3A%7E100+definition%3A76+name%3AataGridView_ColumnHeadersHeight_SetWithHandle_GetReturnsExpected) those tests fail in different PRs like #8500

I'll do rebase for Issue_4461_ExtraCode_RowTemplateHeight on the most recent main to see if it fixes the issue.

dmitrii-drobotov (commented):
LGTM! Can you also add unit tests, so that if the default font changes in the future, we would notice it?

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRow.cs
```diff
@@ -35,6 +36,7 @@ public DataGridViewRow() : base()
         {
             MinimumThickness = DefaultMinRowThickness;
             Thickness = Control.DefaultFont.Height + 9;
+            Height = Control.DefaultFont.Height + 9 == DefaultHeight ? DefaultHeight : Thickness;
```
Tanya-Solyanik, line 39:
```suggestion
            Height = Thickness;
```

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRow.cs
Tanya-Solyanik: >because if there is a [custom DefaultFont](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props-desktop#applicationdefaultfont), then current implementation would adjust height accordingly, but with this change it'd break and require setting a new height.

@dkazennov  - had you investigated this idea? It would handle the case when control is defined in the library were default font is not modified and used in an all with a custom default font.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRow.cs
```diff
@@ -1730,6 +1731,12 @@ internal void OnSharedStateChanging(int sharedRowIndex, DataGridViewElementState
             }
         }
 
+        // ShouldSerialize and Reset Methods are being used by Designer via reflection.
+        private void ResetHeight()
+        {
+            Height = DefaultHeight;
```
Tanya-Solyanik, line 1737: This implementation runs the same logic as  `[DefaultValue(25)]` would had run, i.e. a comparison with a constant. Dynamic methods had an advantage if you adjust `DefaultValue` according to DPI and custom default font. @dreddy-work  -is default font adjusted for the DPI?

Tanya-Solyanik:
Unit test that compares `DefaultHeight` field to the actual app default font height(with appropriate adjustments) would be nice.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRow.cs
dreddy-work: Default Font if it is not SystemFont, is scaled to DPI settings on primary monitor. But it is not updated at runtime when moving between monitors.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRow.cs
dkazennov: @Tanya-Solyanik 
> isn't this condition equivalent to -

If (Control.DefaultFont.Height + 9 == DefaultHeight) then the default font is not changed.
This way we can use the constant DefaultHeight.

But if (Control.DefaultFont.Height + 9 != DefaultHeight) then the default font is changed and so we should use Thickness (as was before this fix).

dkazennov:
> Unit test that compares `DefaultHeight` field to the actual app default font height(with appropriate adjustments) would be nice.

Yes! I will make tests to cover this fix as @dmitrii-drobotov also proposed.

dkazennov:
@Tanya-Solyanik 
Regarding custom default font and `Height = Control.DefaultFont.Height + 9 == DefaultHeight ? DefaultHeight : Thickness;`
In WinformsControlsTest application we have custom default font of Calibri-11 with height 23. So Height and Thickness are 32 instead on 25.

image

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRow.cs
dkazennov: > isn't this condition equivalent to -

Yes, you are correct. I'm not getting anything by adding this assignment.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRow.cs
```diff
@@ -25,6 +25,7 @@ public partial class DataGridViewRow : DataGridViewBand
         private const DataGridViewAutoSizeRowCriteriaInternal InvalidDataGridViewAutoSizeRowCriteriaInternalMask = ~(DataGridViewAutoSizeRowCriteriaInternal.Header | DataGridViewAutoSizeRowCriteriaInternal.AllColumns);
 
         private const int DefaultMinRowThickness = 3;
+        private const int DefaultHeight = 25;
```
Tanya-Solyanik, line 28: I would investigate into this direction:

private int DefaultFont
{
    return Control.DefaultFont.Height + 9;
}

Tanya-Solyanik:
>In WinformsControlsTest application we have custom default font of Calibri-11 with height 23. So Height and Thickness are 32 instead on 25.

I think that row height value of 32 should not be serialized in this case, is it?

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRow.cs
dkazennov: This way the unit test implemented earlier fails (since in is designer to make sure that custom default font's height is not equal to the constant AND that actual row height is equal to Control.DefaultFont.Height + 9). 

I believed that we should stick to the initial idea of the fix which works as far as I understand. Or if there is something wrong with the current fix?

dkazennov:
Builds fail because new unit test. But it passes locally both in Test Explorer and with cmd script replicating CI\CD builds.
The fix with the property instead of the constant seems to work as well (with a unit test changed). I will commit changed code and see if the test will pass or fail.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRow.cs
dkazennov: > I would investigate into this direction:
> 
> private int DefaultHeight { return Control.DefaultFont.Height + 9; } This way height serialized for one DefaultFont will pick up a different default font at runtime.

Actually this works too and looks more consistent than the initial idea.
I've changed the PR and adjusted the unit tests as well.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewRowTests.cs
```diff
@@ -2019,6 +2019,20 @@ public void DataGridViewRow_Height_SetInvalid_ThrowsArgumentOutOfRangeException(
             Assert.Throws<ArgumentOutOfRangeException>("value", () => row.Height = value);
         }
 
+        [WinFormsFact]
+        public void DataGridViewRow_Height_NotEqualDefaultHeight_IfDefaultFontIsChanged()
+        {
+            var oldApplicationDefaultFont = Application.DefaultFont;
+            var applicationTestAccessor = typeof(Application).TestAccessor().Dynamic;
+            applicationTestAccessor.s_defaultFont = new Font("Times New Roman", 12);
```
dmitrii-drobotov, line 2027: Can `Application.SetDefaultFont` be used here instead?

dkazennov: > Can `Application.SetDefaultFont` be used here instead?

It gives `SetDefaultFont must be called before the first IWin32Window object is created in the application.` which is probably why TestAccessor is being used in all similar cases.

Tanya-Solyanik (dismissed):

NikitaSemenovAkvelon:
Is the default value should be recalculated if `DefaultCellStyle.Font` is changed?
image
image

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRow.cs
```diff
@@ -151,6 +151,11 @@ public string ErrorText
             set => ErrorTextInternal = value;
         }
 
+        private static int DefaultHeight
+        {
+            get => Control.DefaultFont.Height + 9;
```
NikitaSemenovAkvelon, line 156: In line 37, `Thickness` is set to the same value `Control.DefaultFont.Height + 9`. I suggest replacing it with `DefaultHeight`.

NikitaSemenovAkvelon, line 156: I would like to move `9` to a `const` with an appropriate name. Now, at least for me, it seems like a magic value.

dkazennov:
> Is the default value should be recalculated if `DefaultCellStyle.Font` is changed?
In the most recent revision we don't recalculate the value. We use a property instead of a constant, so we always have the actual value.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRow.cs
dkazennov: > I would like to move `9` to a `const` with an appropriate name. Now, at least for me, it seems like a magic value.

It looks like a loop. We add a property (or a constant) to store some value. Then we add a constant to store the addendum value.

NikitaSemenovAkvelon: In my humble opinion, it is still better than a magic value

dkazennov: We have to store a literal value one way or another. We don't have any kind of enum with offset constants, but we now have the new property DefaultHeight which already serves exactly this purpose. I believe it is fine to use DefaultHeight to store the value.

dmitrii-drobotov (approved):
LGTM

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRow.cs
dkazennov: This marked as resolved for now the DefaultHeight property has no constant value. It depends on the actual Control.DefaultFont.Height.

dkazennov:
@merriemcgaw could you please approve this PR regarding the changes you have requested? (Since then this fix have changed quite a lot).

merriemcgaw (approved):
Approving to unblock merging based on the new approach.

dreddy-work:
@Tanya-Solyanik, can you take a final look at this?

Tanya-Solyanik:
>@Tanya-Solyanik, can you take a final look at this?

@dreddy-work - I want to discuss this pattern with the team. Attribute removal is a breaking change - https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/breaking-change-rules.md#attributes. But I consider this case a `grey area` because functionality is preserved.

merriemcgaw:
@Tanya-Solyanik, @dreddy-work Let's put this on the agenda for tomorrow's meeting.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRow.cs
Tanya-Solyanik: Please store the height in a static field in the same way as you did in the list box PR.

Tanya-Solyanik:
breaking change doc that covers all attributes removals that follow this pattern - https://github.com/dotnet/docs/issues/33954

Tanya-Solyanik (approved):
Thank you!

Tanya-Solyanik:
@Nora-Zhou01  - yes, this is expected because we had changed the font height in .NET

PR #8477, opened 2023-01-10
Improve `Cursor` `Stream` constructor to be more robust

### LINKED ISSUE #8373: Flaky test: cursorconverter_convertto_unknowntoinstancedescriptor_throwsnotsupportedexception
Runfo Tracking Issue: [Flaky test: cursorconverter_convertto_unknowntoinstancedescriptor_throwsnotsupportedexception](https://runfo.azurewebsites.net/tracking/issue/445)
|Build|Definition|Kind|Run Name|
|---|---|---|---|
|[207837](https://dev.azure.com/dnceng-public/public/_build/results?buildId=207837)|[dotnet-winforms CI](https://dnceng-public.visualstudio.com/public/_build?definitionId=76)|[PR 8865](https://github.com/dotnet/winforms/pull/8865)|Windows_x64-xunit|
|[207837](https://dev.azure.com/dnceng-public/public/_build/results?buildId=207837)|[dotnet-winforms CI](https://dnceng-public.visualstudio.com/public/_build?definitionId=76)|[PR 8865](https://github.com/dotnet/winforms/pull/8865)|Windows_arm64-xunit|
|[207837](https://dev.azure.com/dnceng-public/public/_build/results?buildId=207837)|[dotnet-winforms CI](https://dnceng-public.visualstudio.com/public/_build?definitionId=76)|[PR 8865](https://github.com/dotnet/winforms/pull/8865)|Windows_x64-xunit|
|[207837](https://dev.azure.com/dnceng-public/public/_build/results?buildId=207837)|[dotnet-winforms CI](https://dnceng-public.visualstudio.com/public/_build?definitionId=76)|[PR 8865](https://github.com/dotnet/winforms/pull/8865)|Windows_x86-xunit|

Build Result Summary
|Day Hit Count|Week Hit Count|Month Hit Count|
|---|---|---|
|1|1|1|

elachlan:
Change Cursor ctor to reset the stream parameters position ready for the copyTo call.

This is a guess on the fix, but would make sense.

Related: #8367, #8373

> Copying begins at the current position in the current stream, and does not reset the position of the destination stream after the copy operation is complete.
https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.copyto?view=net-7.0

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8477)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Cursor.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Cursor.cs
@@ -82,6 +82,12 @@ namespace System.Windows.Forms
         {
             ArgumentNullException.ThrowIfNull(stream);
             MemoryStream memoryStream = new();
+            // reset stream position to start, there are no gaurantees it is at the start.
+            if (stream.CanSeek)
+            {
+                stream.Position = 0;
+            }
+
             stream.CopyTo(memoryStream);
             _cursorData = memoryStream.ToArray();
 
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CursorTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CursorTests.cs
@@ -58,6 +58,21 @@ namespace System.Windows.Forms.Tests
         }
 
         [Fact]
+        public void Cursor_Ctor_Stream_NonStartPosition()
+        {
+            using var stream = new MemoryStream(File.ReadAllBytes(Path.Combine("bitmaps", "cursor.cur")));
+            stream.Position = 5;
+            using var cursor = new Cursor(stream);
+            Assert.NotNull(cursor);
+        }
+
+        [Fact]
+        public void Cursor_Ctor_EmptyStream_ThrowsArgumentException()
+        {
+            Assert.Throws<ArgumentException>("stream", () => new Cursor(new MemoryStream()));
+        }
+
+        [Fact]
         public void Cursor_Ctor_NullStream_ThrowsArgumentNullException()
         {
             Assert.Throws<ArgumentNullException>("stream", () => new Cursor((Stream)null));


src/System.Windows.Forms/src/System/Windows/Forms/Cursor.cs
```diff
@@ -82,6 +82,12 @@ public Cursor(Stream stream)
         {
             ArgumentNullException.ThrowIfNull(stream);
             MemoryStream memoryStream = new();
+            // reset stream position to start, there are no gaurantees it is at the start.
+            if (stream.CanSeek)
```
elachlan, line 86: I've used `CanSeek` here because it still allows for streams which aren't seekable.

src/System.Windows.Forms/src/System/Windows/Forms/Cursor.cs
```diff
@@ -82,6 +82,12 @@ public Cursor(Stream stream)
...
+            {
+                stream.Position = 0;
+            }
+
             stream.CopyTo(memoryStream);
```
elachlan, line 91: We could add a check here for `length==0` and throw a ArguementException early. That way we avoid calling into interop.

elachlan:
I can't for sure say it fixes the flaky test, but it might help. It would be great if we can rerun the test a dozen times and see if the flaky tests surfaces.

src/System.Windows.Forms/src/System/Windows/Forms/Cursor.cs
dreddy-work: If `Stream` is not seekable, does `CopyTo` succeed?

src/System.Windows.Forms/src/System/Windows/Forms/Cursor.cs
dreddy-work: An empty but not `null `_cursorData is a valid scenario, i do not think we should change that behavior.  Do you think this would not be a use case?

src/System.Windows.Forms/src/System/Windows/Forms/Cursor.cs
dreddy-work: if not, we don't need this check

src/System.Windows.Forms/src/System/Windows/Forms/Cursor.cs
dreddy-work: Given its public, we don't control input. Unless we are failing down the line, i wouldn't fail here either.

src/System.Windows.Forms/src/System/Windows/Forms/Cursor.cs
elachlan: > Copying begins at the current position in the current stream.

If the stream is not seekable, it will just copy from whatever the current position of the stream is to the end. I can't see why it would "fail" on a seek-less stream. Only failure case (which I also think is valid) is if the passed in stream was seek-less and already had been read. We can't do anything about that and should just fail.

src/System.Windows.Forms/src/System/Windows/Forms/Cursor.cs
elachlan: Actually, if you pass an empty `MemoryStream` you get an `ArgumentException` bubbled up from an image load failure in interop layer.

So its currently not a valid scenario to pass an empty stream, which is why I suggested a length check. BUT not all streams can check the length. So I say we leave it as is, since its handled in interop when read.

src/System.Windows.Forms/src/System/Windows/Forms/Cursor.cs
dreddy-work: According to the documentation, the CopyStream method should be able to adjust the position of the stream. I was curious to know if it is possible to perform the copy operation without not being able to seek.

elachlan: I do not know the answer to that.

dreddy-work (approved):

PR #8478, opened 2023-01-10
Fixes `Form.Show(IWin32Window)` causing memory leaks by removing setting of `PropDialogOwner`

### LINKED ISSUE #530: Form.Show/PropDialogOwner causing memory leaks
* .NET Core Version: (e.g. 3.0 Preview1, or daily build number, use `dotnet --info`)

**Problem description:**
Form.Show(IWin32Window) causes memory leaks because of references to the Owner form held within the PropertyStore of the form.

problematic line:
https://github.com/dotnet/winforms/blob/master/src/System.Windows.Forms/src/System/Windows/Forms/Form.cs#L5929

**Actual behavior:** 
In ShowDialog(IWin32Window) I see the property PropDialogOwner is nulled in the finally block. Using Show(...) method it never gets nulled. In Dispose only PropOwner is nulled. So the reference to the provided Owner-Form is held the PropertyStore and child forms will not GCed.

**Expected behavior:**
A: The said property is cleared in Dispose like PropOwner
B: PropDialogOwner is not used at all because it is not ShowDialog

**Minimal repro:**
* open a Form
* open some forms which consume some memory
* let them open child forms with said method 
* and close them
* profile memory for parent forms

```
    static class Program
    {
        [STAThread]
        static void Main()
        {
            Application.SetCompatibleTextRenderingDefault(false);
            Application.Run(new MainForm());
        }
    }

    public class MainForm : Form
    {
        public MainForm()
        {
            long a = Process.GetCurrentProcess().WorkingSet64;

            var children = new List<Form>();
            for (int i = 0; i < 50; i++)
            {
                ParentForm parent = new ParentForm();
                parent.Show();
                children.Add(parent.Yum());
                parent.Close();
            }
            
            children.ForEach(cf => cf.Close());

            long b = Process.GetCurrentProcess().WorkingSet64;
            GC.Collect();
            GC.Collect();
            long c = Process.GetCurrentProcess().WorkingSet64;
            
            this.Controls.Add(new Label() { Text = $"{a}\n{b}\n{c}", Dock = DockStyle.Fill, TextAlign = ContentAlignment.MiddleCenter});
        }
    }

    public class ParentForm : Form
    {
        private Byte[] array;

        public Form Yum()
        {
            this.array = new Byte[1024 * 1024];
            Array.Clear(array, 0, this.array.Length);
            
            Form child = new Form();
            child.Show(this); // <-----

            return child;
        }
    }
```

**Workaround**
```
child.Owner = this;
child.Show();
```

elachlan:
Fixes #530

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8478)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
@@ -5235,7 +5235,6 @@ namespace System.Windows.Forms
 
             HWND activeHwnd = PInvoke.GetActiveWindow();
             HandleRef<HWND> ownerHwnd = owner is null ? GetHandleRef(activeHwnd) : GetSafeHandle(owner);
-            Properties.SetObject(PropDialogOwner, owner);
             Form? oldOwner = OwnerInternal;
             if (owner is Form ownerForm && owner != oldOwner)
             {
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
@@ -109,6 +109,7 @@ namespace System.Windows.Forms
         private static readonly int PropAcceptButton = PropertyStore.CreateKey();
         private static readonly int PropCancelButton = PropertyStore.CreateKey();
         private static readonly int PropDefaultButton = PropertyStore.CreateKey();
+        private static readonly int PropDialogOwner = PropertyStore.CreateKey();
 
         private static readonly int PropOwner = PropertyStore.CreateKey();
         private static readonly int PropOwnedForms = PropertyStore.CreateKey();
@@ -796,9 +797,10 @@ namespace System.Windows.Forms
                     cp.ExStyle |= (int)WINDOW_EX_STYLE.WS_EX_LAYERED;
                 }
 
-                if (OwnerInternal is not null)
+                IWin32Window? dialogOwner = (IWin32Window?)Properties.GetObject(PropDialogOwner);
+                if (dialogOwner is not null)
                 {
-                    cp.Parent = GetSafeHandle(OwnerInternal!).Handle;
+                    cp.Parent = GetSafeHandle(dialogOwner).Handle;
                 }
 
                 FillInCreateParamsBorderStyles(cp);
@@ -3378,6 +3380,8 @@ namespace System.Windows.Forms
                     Properties.SetObject(PropOwner, null);
                 }
 
+                Properties.SetObject(PropDialogOwner, null);
+
                 Form?[]? ownedForms = (Form?[]?)Properties.GetObject(PropOwnedForms);
                 int ownedFormsCount = Properties.GetInteger(PropOwnedFormsCount);
 
@@ -3544,9 +3548,12 @@ namespace System.Windows.Forms
                     else
                     {
                         Screen desktop;
-                        if (OwnerInternal is not null)
+                        IWin32Window? dialogOwner = (IWin32Window?)Properties.GetObject(PropDialogOwner);
+                        if ((OwnerInternal is not null) || (dialogOwner is not null))
                         {
-                            HandleRef<HWND> ownerHandle = new(OwnerInternal!);
+                            HandleRef<HWND> ownerHandle = dialogOwner is not null
+                                ? GetSafeHandle(dialogOwner)
+                                : new(OwnerInternal!);
                             desktop = Screen.FromHandle(ownerHandle.Handle);
                             GC.KeepAlive(ownerHandle.Wrapper);
                         }
@@ -5230,6 +5237,7 @@ namespace System.Windows.Forms
 
             HWND activeHwnd = PInvoke.GetActiveWindow();
             HandleRef<HWND> ownerHwnd = owner is null ? GetHandleRef(activeHwnd) : GetSafeHandle(owner);
+            Properties.SetObject(PropDialogOwner, owner);
             Form? oldOwner = OwnerInternal;
             if (owner is Form ownerForm && owner != oldOwner)
             {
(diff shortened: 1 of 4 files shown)


elachlan:
It was suggested that `PropOwner` be used instead of `PropDialogOwner`, but `PropOwner` is set anyway via the `Owner` property setter.

elachlan:
@Olina-Zhang could your team please test this fix?

Olina-Zhang:
@elachlan, tested this PR based on the latest .Net 8.0 SDK build, after GC, the process is still higher than before showing forms when showing forms 50 times and 100 times, is it an acceptable result?
image

elachlan:
Thanks Olina! I'll wait for a review and see how we go.

elachlan:
@dmitrii-drobotov its showing an improvement with the PR, but there are still excess memory usage. Do you think its an accessible object issue?

dmitrii-drobotov:
> @dmitrii-drobotov its showing an improvement with the PR, but there are still excess memory usage. Do you think its an accessible object issue?

@elachlan It's probably something else because accessible objects shouldn't be created at all without using assistive tools like screen readers. You can try to see which objects stays in memory by taking [memory snapshots in VS](https://learn.microsoft.com/en-us/visualstudio/profiling/memory-usage?view=vs-2022) or other tools like WinDbg.

elachlan:
Ok, I think this is solved. I used a variation on the test with 500 forms. There is some inconsistency in testing where the difference after isn't as pronounced.
image

But some runs its a good difference:
image

Comparing it to without the fix:
image

So I don't think there is a leak anymore. Snapshots in VS show lower heap usage each time.

elachlan:
@dreddy-work Can you have a quick look?

dreddy-work:
Apart from the memory usage, Did you perform any testing on the flow of these new changes, specifically regarding when and how `PropDialogOwner `is set and if all code paths correctly retrieve it through `GetObject`?

dreddy-work:
And the reason for leak? I do see it is being set when handle is created. What makes it to not leak now vs earlier?

elachlan:
I reviewed this again.

`PropDialogOwner` is only ever disposed by `ShowDialog`. Not on `Form.Dispose`.

The usage that concerned me after I looked deeper into it is in `CreateParams`.
https://github.com/dotnet/winforms/blob/7b9fe1a1abc0302f4eeef2f7bab778db34e12426/src/System.Windows.Forms/src/System/Windows/Forms/Form.cs#L800-L804

I am not sure how this is used. It does not set the Form.Parent property. The solution to keep this is to add the nulling of `PropDialogOwner` to `Form.Dispose`.

The other usage is in the start position, which uses the `OwnerInternal` if `PropDialogOwner` isn't present.

My comment previously explained why the code change is just a line removal:
> It was suggested that PropOwner be used instead of PropDialogOwner, but PropOwner is set anyway via the Owner property setter.

Here is the usage I was talking about:
https://github.com/dotnet/winforms/blob/7b9fe1a1abc0302f4eeef2f7bab778db34e12426/src/System.Windows.Forms/src/System/Windows/Forms/Form.cs#L1833-L1839

and where the owner is set:
https://github.com/dotnet/winforms/blob/7b9fe1a1abc0302f4eeef2f7bab778db34e12426/src/System.Windows.Forms/src/System/Windows/Forms/Form.cs#L5238-L5243

and the owner property:
https://github.com/dotnet/winforms/blob/7b9fe1a1abc0302f4eeef2f7bab778db34e12426/src/System.Windows.Forms/src/System/Windows/Forms/Form.cs#L1799-L1831

elachlan:
@dreddy-work Is there a difference between an "Owned form" and a "Child Form"?

Eventually CreateParams.Parent gets used in the `PInvoke.CreateWindowEx` call in `NativeWindow`. So Removing `PropDialogOwner` isn't good because the native owner isn't set.
https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-createwindowexa

BUT we can follow the pattern used elsewhere in `Form`:
```csharp
IWin32Window? dialogOwner = (IWin32Window?)Properties.GetObject(PropDialogOwner);
if ((OwnerInternal is not null) || (dialogOwner is not null))
{
    cp.Parent = dialogOwner is not null
        ? GetSafeHandle(dialogOwner).Handle
        : GetSafeHandle(OwnerInternal!).Handle;
}
```

Thus we aren't duplicating a value for no reason. 

But then it begs the question, if we set the `Owner` property in `ShowDialog(IWin32Window? owner)`, why do we need `PropDialogOwner` at all? Can't we just use `OwnerInternal` all the time anyway?

elachlan:
@JeremyKuhne You might want to chime in here, I might be missing something obvious.

Tldr; I think in `Forms.cs` `PropDialogOwner` is redundant since we can use `OwnerInternal` which uses `PropOwner`. Removing it removes a duplicate reference.

JeremyKuhne:
> Tldr; I think in Forms.cs PropDialogOwner is redundant since we can use OwnerInternal which uses PropOwner. Removing it removes a duplicate reference.

I think it would be safer to just make sure it is cleared on dispose. You'd be changing the semantics when `Show(IWin32Window? owner)` is not a `Form`, you wouldn't get the `IWin32Window`'s handle in other places.

elachlan:
Results after change:
image

elachlan:
@Olina-Zhang could you please test this again with the latest changes to confirm its fixed?

Olina-Zhang:
@elachlan, tested this PR change for 500 forms, but the result doesn't look good after GC, here is the testing app and testing result, please take a look.
[TestSample.zip](https://github.com/dotnet/winforms/files/10430512/TestSample.zip)
image

elachlan:
Thanks, I think I ran into a similar problem previously. You need to call `children.clear();` after `children.ForEach(cf => cf.Close());`.

Olina-Zhang:
Added `children.Clear()` after `children.ForEach(cf => cf.Close())`, it still shows:
image

elachlan:
I've added a test to WinformsControlsTest to help reproduce it easily.

Before:
image

After:
image

Olina-Zhang:
My testing result from WinformsControlsTest app by `start-vs.cmd` , it is with your PR change by default, right? But it is not consistent with yours.
image

elachlan:
The memory usage will vary, but as long as the during isn't growing to 500mb now and the After is lower than the during it should be fixed.

dreddy-work (approved):

PR #8484, opened 2023-01-11
Add `ToolStripSeparatorPreferredSize` form in `WinformsControlsTest` for `ToolStripSeparator.GetPreferredSize`

### LINKED ISSUE #2576: ToolStripSeparator.GetPreferredSize has a suspicious always-true if check
The implementation of `ToolStripSeparator.GetPreferredSize` is as follows:

```cs
public override Size GetPreferredSize(Size constrainingSize)
{
    ToolStrip parent = ParentInternal;

    if (parent == null)
    {
        parent = Owner;
    }
    if (parent == null)
    {
        return new Size(ToolStrip_SEPARATORTHICKNESS, ToolStrip_SEPARATORTHICKNESS);
    }

    if (parent is ToolStripDropDownMenu dropDownMenu)
    {
        return new Size(parent.Width - (parent.Padding.Horizontal - dropDownMenu.ImageMargin.Width), ToolStrip_SEPARATORTHICKNESS);
    }
    else
    {
        if (parent.LayoutStyle != ToolStripLayoutStyle.HorizontalStackWithOverflow || parent.LayoutStyle != ToolStripLayoutStyle.VerticalStackWithOverflow)
        {
            // we dont actually know what size to make it, so just keep it a stock size.
            constrainingSize.Width = ToolStrip_SEPARATORHEIGHT;
            constrainingSize.Height = ToolStrip_SEPARATORHEIGHT;
        }
        if (IsVertical)
        {
            return new Size(ToolStrip_SEPARATORTHICKNESS, constrainingSize.Height);
        }
        else
        {
            return new Size(constrainingSize.Width, ToolStrip_SEPARATORTHICKNESS);
        }
    }

}
```

Notice this case: `if (parent.LayoutStyle != ToolStripLayoutStyle.HorizontalStackWithOverflow || parent.LayoutStyle != ToolStripLayoutStyle.VerticalStackWithOverflow)`

This appears to always be true, as `parent.LayoutStyle` can't be both `HorizontalStackWithOverflow` and `VerticalStackWithOverflow`. 

Should this actually be an `&&` check rather than `||`

elachlan:
Changed to pattern matching and did minor refactor for conditional expression

Related: #2576

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8484)

### DIFF
--- a/src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/MainForm.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/MainForm.cs
@@ -209,6 +209,11 @@ public partial class MainForm : Form
         {
             MainFormControlsTabOrder.PasswordButton,
             new InitInfo("Password", (obj, e) => new Password().Show(this))
+        },
+        {
+            // Helps test the output of GetPreferredSize output https://github.com/dotnet/winforms/issues/2576
+            MainFormControlsTabOrder.ToolStripSeparatorPreferredSize,
+            new InitInfo("ToolStripSeparatorPreferredSize", (obj, e) => new ToolStripSeparatorPreferredSize().Show(this))
         }
     };
 
--- a/src/System.Windows.Forms/tests/IntegrationTests/System.Windows.Forms.IntegrationTests.Common/MainFormControlsTabOrder.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/System.Windows.Forms.IntegrationTests.Common/MainFormControlsTabOrder.cs
@@ -43,5 +43,6 @@ public enum MainFormControlsTabOrder
     MediaPlayerButton,
     FormOwnerTestButton,
     ListBoxTestButton,
-    PasswordButton
+    PasswordButton,
+    ToolStripSeparatorPreferredSize
 }
--- a/src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/ToolStripSeparatorPreferredSize.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/ToolStripSeparatorPreferredSize.cs
@@ -1,0 +1,105 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace WinformsControlsTest;
+
+internal class ToolStripSeparatorPreferredSize : Form
+{
+    public ToolStripSeparatorPreferredSize()
+    {
+        InitializeComponent();
+        this.buttonChangeOrientation.Click += ChangeOrientation;
+        SetLabelText();
+    }
+
+    private void ChangeOrientation(object sender, EventArgs e)
+    {
+        toolStrip1.LayoutStyle = toolStrip1.LayoutStyle == ToolStripLayoutStyle.VerticalStackWithOverflow
+            ? ToolStripLayoutStyle.HorizontalStackWithOverflow
+            : ToolStripLayoutStyle.VerticalStackWithOverflow;
+        SetLabelText();
+    }
+
+    private void SetLabelText()
+    {
+        label1.Text = $"Layout: {toolStrip1.LayoutStyle}, Size: {toolStripSeparator1.Size}, GetPreferredSize: {toolStripSeparator1.GetPreferredSize(toolStripSeparator1.Size)}";
+    }
+
+    private void InitializeComponent()
+    {
+        this.toolStrip1 = new System.Windows.Forms.ToolStrip();
+        this.toolStripLabel1 = new System.Windows.Forms.ToolStripLabel();
+        this.toolStripLabel2 = new System.Windows.Forms.ToolStripLabel();
+        this.toolStripSeparator1 = new System.Windows.Forms.ToolStripSeparator();
+        this.buttonChangeOrientation = new System.Windows.Forms.Button();
+        this.label1 = new System.Windows.Forms.Label();
+        this.toolStrip1.SuspendLayout();
+        this.SuspendLayout();
+        // 
+        // toolStrip1
+        // 
+        this.toolStrip1.Items.AddRange(new System.Windows.Forms.ToolStripItem[]
+        {
+            this.toolStripLabel1,
+            this.toolStripSeparator1,
+            this.toolStripLabel2
+        });
+        this.toolStrip1.Location = new System.Drawing.Point(0, 0);
+        this.toolStrip1.Name = "toolStrip1";
+        this.toolStrip1.Size = new System.Drawing.Size(481, 22);
+        this.toolStrip1.TabIndex = 0;
+        this.toolStrip1.Text = "toolStrip1";
+        this.toolStrip1.TabStop = true;
+        // 
+        // toolStripLabel1
+        // 
+        this.toolStripLabel1.Name = "toolStripLabel1";
+        this.toolStripLabel1.Size = new System.Drawing.Size(86, 22);
+        this.toolStripLabel1.Text = "ToolStripLabel";
+        // 
+        // toolStripLabel2
+        // 
+        this.toolStripLabel2.Name = "toolStripLabel2";
+        this.toolStripLabel2.Size = new System.Drawing.Size(86, 22);
+        this.toolStripLabel2.Text = "ToolStripLabel";
+        // 
+        // buttonChangeOrientation
+        // 
+        this.buttonChangeOrientation.AutoSizeMode = AutoSizeMode.GrowAndShrink;
+        this.buttonChangeOrientation.Name = "buttonChangeOrientation";
+        this.buttonChangeOrientation.Location = new System.Drawing.Point(10, 75);
+        this.buttonChangeOrientation.Size = new System.Drawing.Size(150, 25);
+        this.buttonChangeOrientation.TabIndex = 0;
+        this.buttonChangeOrientation.Text = "Change Orientation";
+        this.buttonChangeOrientation.UseVisualStyleBackColor = true;
+        // 
+        // label1
+        //
+        this.label1.Name = "label1";
+        this.label1.Location = new System.Drawing.Point(10, 50);
+        this.label1.AutoSize = true;
+        // 
+        // Form1
+        // 
+        this.AutoScaleDimensions = new System.Drawing.SizeF(6F, 13F);
+        this.AutoScaleMode = System.Windows.Forms.AutoScaleMode.Font;
+        this.Size = new System.Drawing.Size(881, 150);
+        this.Controls.Add(this.toolStrip1);
+        this.Controls.Add(this.label1);
+        this.Controls.Add(this.buttonChangeOrientation);
+        this.Name = "ToolStripSeparatorPreferredSize";
+        this.Text = "ToolStripSeparator GetPreferredSize";
+        this.toolStrip1.ResumeLayout(false);
+        this.toolStrip1.PerformLayout();
+        this.ResumeLayout(false);
+        this.PerformLayout();
+    }
+
+    private System.Windows.Forms.ToolStrip toolStrip1;
+    private System.Windows.Forms.ToolStripLabel toolStripLabel1;
+    private System.Windows.Forms.ToolStripLabel toolStripLabel2;
+    private System.Windows.Forms.ToolStripSeparator toolStripSeparator1;
+    private System.Windows.Forms.Label label1;
+    private System.Windows.Forms.Button buttonChangeOrientation;
+}


dreddy-work:
@elachlan , can you add a unit test around this?

dreddy-work:
Looks like some tests are already covering this code path :)

elachlan:
I thought that maybe the values would be slightly different. But there were some extreme changes, such as the test using `int.MaxValue`.

So to keep existing functionality, which seems to be working, I removed the always true condition and moved it into the conditional expression.

dreddy-work:
> So to keep existing functionality, which seems to be working, I removed the always true condition and moved it into the conditional expression.

This change is no-op and only serves as a cleanup. Have you tested it with different layout settings configurations? I am hesitant to take it without fully understanding why the initial commit was incorrect.

elachlan:
> This change is no-op and only serves as a cleanup. 

Yes. It just reduces the complexity.

> Have you tested it with different layout settings configurations? I am hesitant to take it without fully understanding why the initial commit was incorrect.

There is unit test coverage. The "mistake" has been there since the beginning as far as I can tell.

When I changed the code to "fix" the logic. It caused the `constrainingSize` width or height settings to be kept based on the style. This failed the tests which pass in unusual width and height values. I do not know if this is correct behaviour or not, but the values looked wrong.

I think the whole point of the original code is do one thing when parent is ToolStripDropDownMenu otherwise just return a standard size for the separator depending if it horizontal or vertical (`23,6` or `6,23`)

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripSeparator.cs
```diff
@@ -275,19 +275,8 @@ public override Size GetPreferredSize(Size constrainingSize)
...
-                constrainingSize.Height = SeparatorHeight;
-            }
-
-            if (IsVertical)
-            {
-                return new Size(SeparatorThickness, constrainingSize.Height);
-            }
-
-            return new Size(constrainingSize.Width, SeparatorThickness);
+            // we don't actually know what size to make it, so just keep it a stock size.
```
Tanya-Solyanik, line 278: Please capitalize the comment

Tanya-Solyanik (approved):
Nice cleanup!

dreddy-work:
@elachlan,  I need some time to consider this. I still believe that the original change (first commit) is the better option. I will review the codebase at some time this week and see if we coudl merge this as-is.

elachlan:
@dreddy-work, I think ultimately you are correct. I don't think it will cause problems either, because the size is basically a suggestion and the layout will ultimately decide when it happens.

This documentation helped: https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.toolstriplayoutstyle?view=netframework-4.8

I think depending on IsVertical we need to constrain the values based on the Parent width and height. Those extreme values we test for should not be returned.

I'll have a think and update the PR.

dkazennov (commented):
So with the initial conditional statement with `||` we have no reported issues but a suspicious always-true result. 
With the first commit (conditional statement with `&&` which is similar with other such statements elsewhere) we have some unit tests failing.
With the final version we have no useless always-true conditional statement, no other reported bugs with this class and no unit tests failing. 
It looks like we are not loosing anything this final way. I believe we can keep the final result.

Tanya-Solyanik:
I agree with @dkazennov  - the "always true" test is not needed because it is superseded by IsVertical. Default initial size of the separator depends not only on the layout style but on more complex conditions outlined here - https://github.com/dotnet/winforms/blob/75dee0b5ba4ce9f1f433598c2aca7b1701f16893/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripSeparator.cs#L168

elachlan:
@Tanya-Solyanik I think you are correct. The "always true" was really to reset to a default value and if it was properly used it would be a way to test for when the code couldn't decide if the Separator should be vertical or not. But since we have a vertical check, it doesn't matter.

The issue I see is that the height is always being set to the default and doesn't check the parent when is something other than `ToolStripDropDownMenu`. Its how it has worked worked since the history began in this repo, so I don't think we need to change it now unless it fixes an incorrect behaviour of somekind.

NikitaSemenovAkvelon:
I have doubts about this refactoring.
First of all, `LayoutStyle` is a property of type `ToolStripLayoutStyle` and this `enum` has 5 values, but the condition, which is being refactored, checks only 2 values of the `enum`. I do believe that we shouldn't remove it if we aren't sure that it does not affect something. And technically even in that case, it wouldn't just refactoring, because it may change the resulting values.

elachlan:
@NikitaSemenovAkvelon If you look at the original code, it was an always true condition. If we undo that always true to "fix" it. This causes the tests to fail. 

The docs on the enum explain it pretty well:
https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.toolstriplayoutstyle?view=netframework-4.8

`VerticalStackWithOverflow`/`HorizontalStackWithOverflow` can be used to get the separators orientation, but the rest of them need to check the parents orientation to be correct.

That doesn't solve the test failure problem, because the inputs values for the size are reset to defaults each time.

`IsVertical` will need to be refactored to check the parent value for each of the LayoutStyle's we are unsure on. We also need to add additional type checks on parent in `GetPreferredSize`.

Being absolutely correct here isn't necessary though, the code currently works and a massive refactor to "be more correct" might actually break stuff.

NikitaSemenovAkvelon:
Oh sorry, I've got the wrong idea about these changes. Of course, if we have condition like `(x != a || x != b)`, where `a != b` we always get true.
But I think the more important thing is that mentioned in the original issue about replacing `or` condition with `and`.
The documentation explains the `GetPreferredSize` method as "[Retrieves the size of a rectangular area into which a ToolStripSeparator can be fitted.](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.toolstripseparator.getpreferredsize?view=netframework-4.8)". Now, and after your changes, the resulting area will be always equal to (23;6) or (6;23). But for me, it seems wrong, because height or width should depend on provided `constrainingSize`.
A few examples:
image
image
If I correctly understand the meaning of the `GetPreferredSize` method, the resulting values in these cases should be different.

Tanya-Solyanik:
>If I correctly understand the meaning of the GetPreferredSize method, the resulting values in these cases should be different.

Well, we don't have concrete bugs to show it.

elachlan:
When I get a free moment. I will put together a test form into the test project so we can test all combinations.

dreddy-work:
@elachlan, keeping this on your plate for final test addition and then take it in.

elachlan:
> @elachlan, keeping this on your plate for final test addition and then take it in.

Thanks for the reminder. I had forgotten about this one.

elachlan:
rebased, starting work on the tests now.

elachlan:
Visual studio isn't letting me open the form for designing again. Related: #8484 

```
[10:38:28.022232] info: [WinformsControlsTest]: Launching design tools server process...
[10:38:28.667790] fail: [WinformsControlsTest]: Unhandled exception. System.TypeInitializationException: The type initializer for 'Microsoft.VisualStudio.WinForms.Server.WinFormsDesignTimeAssemblies' threw an exception.
[10:38:28.667790] fail: [WinformsControlsTest]:  ---> System.IO.FileNotFoundException: Could not load file or assembly 'System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'. The system cannot find the file specified.
[10:38:28.667790] fail: [WinformsControlsTest]: File name: 'System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'
[10:38:28.667790] fail: [WinformsControlsTest]:    at Microsoft.VisualStudio.WinForms.Server.WinFormsDesignTimeAssemblies..cctor()
[10:38:28.667790] fail: [WinformsControlsTest]:    --- End of inner exception stack trace ---
[10:38:28.667790] fail: [WinformsControlsTest]:    at Microsoft.VisualStudio.WinForms.Server.WinFormsDesignTimeAssemblies.get_All()
[10:38:28.667790] fail: [WinformsControlsTest]:    at DesignToolsServerX64.Program.Main(String[] args)
```

I guess I will just write the form code directly.

dreddy-work:
Checking if this is still blocked for you or you can add code manually to unblock here.

dreddy-work:
> Apologies, I have been busy with life and work. I will try and make some time to look at this soon.

np. just want to make sure you are not waiting on us.

dreddy-work:
waiting on author.

dreddy-work:
still.

dreddy-work:
ping @elachlan.

elachlan:
@dreddy-work re-reading this now. Do you just want the tests and no change to `GetPreferredSize`?

With the "corrected" logic, the returned size will honor the `constrainingSize` input and so inputs that provide `zero` for either value will be honored (we pass Size.Empty in a few places). 

Basically the logic at the moment effectively just ignores `constrainingSize` by overwriting both values. The unit tests reflect this as well.

The places that might override it are:
https://github.com/dotnet/winforms/blob/e517336a469381c4e5ae926280bf5b347d3846a2/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripSystemRenderer.cs#L679-L732

https://github.com/dotnet/winforms/blob/e517336a469381c4e5ae926280bf5b347d3846a2/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs#L2465-L2526

https://github.com/dotnet/winforms/blob/e517336a469381c4e5ae926280bf5b347d3846a2/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs#L2402-L2463

dreddy-work:
> @dreddy-work re-reading this now. Do you just want the tests and no change to `GetPreferredSize`?

Just the tests that you planned to add. If you think existing tests are covering it,  just sync it with main and i will help review and merge it.

elachlan:
To be honest, I think the existing tests are wrong. Which is why we were going to add a demo to the test app. So we can see the input, output, and resulting visuals.

elachlan:
If I make the changes I am suggesting then we get sane values from `GetPreferredSize` when using `HorizontalStackWithOverflow` or `VerticalStackWithOverflow`.
image
image

elachlan:
https://stackoverflow.com/questions/6496160/what-is-the-purpose-of-control-getpreferredsize-method

Based on this I think we should fix the mistake. Instead of keeping it.

dreddy-work:
@elachlan, i think we discussed to take the change where you were removing always true condition ( in a way no-op). And wanted to add a test around this. But latest change i see is only test but original change removed? Sorry if i missed context. its a long thread.

elachlan:
Yes, but if we change it to be correct then the output of the function is more "sane" for its purpose.

dreddy-work:
@elachlan, apology for this long-standing PR. I took time to go over conversation. Given the commits were overridden, couldn't connect conversation exactly. It does seem original source is incorrect to be always true. But, given the functionality has been there and no active bugs reported on this, We could just document and comment in the source (and in tests) for now. 

I will then quickly help merge it to main.

elachlan:
sorry about this. I'll get it sorted sometime today.

lonitra:
@elachlan do you have a cycle to get this updated with main, remove the check in no-op way, and perhaps add comments here?

elachlan:
It has been rebased. The test kind of demonstrates the reasoning for making the change I suggested:
https://github.com/dotnet/winforms/pull/8484#issuecomment-1636633932

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/MainForm.cs
```diff
@@ -209,6 +209,10 @@ public MainForm()
         {
             MainFormControlsTabOrder.PasswordButton,
             new InitInfo("Password", (obj, e) => new Password().Show(this))
+        },
+        {
+            MainFormControlsTabOrder.ToolStripSeparatorPreferredSize,
```
lonitra, line 214: Could you add a short description of this test in comments here so that others looking at this down the road can better understand the intent for this?

lonitra (commented):
Talked with Jeremy offline and we decided it is best to just take the tests here. We will re-evaluate making the change to `&&` once we have a report. Could you address the comment and then also update PR name to reflect this change more accurately. Let's also remove that it fixes the issue from the description.

lonitra (approved):
Thank you!

John-Qiao:
Verified this PR on latest main 9.0 branch of winforms repo, the ToolStripSeparatorPreferredSize form works as below screenshot now, is that the expected result?

testresult

elachlan:
Yes, We did not change the behavior of the the `ToolStripSeparatorPreferredSize` in this PR.

PR #8485, opened 2023-01-11
Fix issues in `KeysConverter.ConvertTo` introduced in #8255

### LINKED ISSUE #8456: Error in Menustrip's shortcutkey property
### .NET version

8.0.100-alpha.1.23054.7

Yes

Yes,it cannot repro on .NET 6.0 & 7.0 SDK builds.

### Issue description

Error in MenuStrip's shortcutkey property, default is "none"

### Steps to reproduce

1. Create a .NET 8.0 WinForms project.
2. Add **MenuStrip** control.
3. Add MenuItem control inside MenuStrip.
4. Click "**Type here**" to set the text property to "TEST".
5. Observe MenuItem's **Shortcutkey** property value in Properties Window.

Or 

1. Extract and open [WinFormsApp28.zip](https://github.com/dotnet/winforms/files/10350587/WinFormsApp28.zip) use VS.
2. Observe MenuItem's Shortcutkey property in Properties Window.

**Actual result:**
Error in MenuStrip's shortcutkey property, displaying "Enter+F12+F11+F10+End+Ctrl+F8+F9+Alt+F4+F5+F6+F7+F1+F2+F3+PgDn+Ins+Home+Del+Shift+PgUp+Back+0+1+2+3+4+5+6+7+8+9+0"
T1
TEST

**Expected result:**
The shortcutkey property in MenuStrip should default to "none".
T2
**More Info:**

1. This issue doesn't repro on .NET Framework.
3. This issue doesn't repro on .NET 7.0, .NET 6.0.
4. VS Version:The latest main VS build.
5. This is a regression issue, it cannot reproduce on .NET version 8.0.100-alpha.1.22601.15. build.
6. Reproduced in version 8.0.100-alpha.1.22607.6 and above.

elachlan:
Also added some tests to better cover the KeysConverter.

Fixes #8456

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8485)

### DIFF
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/KeysConverterTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/KeysConverterTests.cs
@@ -19,5 +19,17 @@ namespace System.Windows.Forms.Tests
 
             Assert.Equal(keys, result);
         }
+
+        [Theory]
+        [InlineData(Keys.None, "None")]
+        [InlineData(Keys.S, "S")]
+        [InlineData(Keys.Control | Keys.C, "Ctrl+C")]
+        [InlineData(Keys.Control | Keys.Alt | Keys.D, "Ctrl+Alt+D")]
+        public void ConvertToString_ShouldConvertKeys(Keys keys, string expectedResult)
+        {
+            KeysConverter converter = new();
+            var result = converter.ConvertToString(keys);
+            Assert.Equal(expectedResult, result);
+        }
     }
 }
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/KeysConverterTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/KeysConverterTests.cs
@@ -9,14 +9,16 @@ namespace System.Windows.Forms.Tests
     public class KeysConverterTests
     {
         [Theory]
+        [InlineData("Ctrl+Alt+Shift+A", Keys.Control | Keys.Alt | Keys.Shift | Keys.A)]
+        [InlineData("Ctrl+Alt+Shift+F1", Keys.Control | Keys.Alt | Keys.Shift | Keys.F1)]
+        [InlineData("Ctrl+Alt+D", Keys.Control | Keys.Alt | Keys.D)]
         [InlineData("Ctrl + N", Keys.Control | Keys.N)]
         [InlineData("G", Keys.G)]
+        [InlineData("None", Keys.None)]
         public void ConvertFrom_ShouldConvertKeys(string input, Keys keys)
         {
             KeysConverter converter = new();
-
             var result = (Keys)converter.ConvertFrom(input);
-
             Assert.Equal(keys, result);
         }
 
@@ -24,7 +26,10 @@ namespace System.Windows.Forms.Tests
         [InlineData(Keys.None, "None")]
         [InlineData(Keys.S, "S")]
         [InlineData(Keys.Control | Keys.C, "Ctrl+C")]
+        [InlineData(Keys.Control | Keys.Add, "Ctrl+Add")]
         [InlineData(Keys.Control | Keys.Alt | Keys.D, "Ctrl+Alt+D")]
+        [InlineData(Keys.Control | Keys.Alt | Keys.Shift | Keys.A, "Ctrl+Alt+Shift+A")]
+        [InlineData(Keys.Control | Keys.Alt | Keys.Shift | Keys.F1, "Ctrl+Alt+Shift+F1")]
         public void ConvertToString_ShouldConvertKeys(Keys keys, string expectedResult)
         {
             KeysConverter converter = new();
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/KeysConverterTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/KeysConverterTests.cs
@@ -2,7 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System.Globalization;
 using Xunit;
 
 namespace System.Windows.Forms.Tests
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/KeysConverterTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/KeysConverterTests.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Globalization;
 using Xunit;
 
 namespace System.Windows.Forms.Tests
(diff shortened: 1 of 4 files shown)


elachlan:
CC: @merriemcgaw here is the fix for the shortcutkeys.

@Nora-Zhou01 could you please test this to confirm the issue is resolved? Unit tests seem to say so.

Nora-Zhou01:
Tested this PR was fixed that the shortcutkey property in MenuStrip should default to "none".
menu

Olina-Zhang:
@elachlan  In addition, found 2 regression issues. Please take a look.
**Issue1:** Some ShotcutKeys setting causes runtime  unhandled exception, for example: Ctrl+Alt+Shift+A
image
**Issue2**: The runtime display of .Net Core and .net framework are not consistent when setting ShortcutKeys as "Ctrl+Alt+Shift+F1"
image

elachlan:
Okay, for what ever reason the designer has serialized `Ctrl+Alt+Shift+A` as `Keys.A`. This fails because the shortcut isn't valid, it fails the validation logic and throws `InvalidEnumArgumentException`.

I can't see why that is happening.

elachlan:
I've added test cases to prove the the new regressions aren't caused by the `KeysConverter` or `ToolStripMenuItem`. I think the issue outlined above is caused by the serializer or designer.

elachlan:
@Olina-Zhang were the regressions caused by the PR or already present?

Olina-Zhang:
@elachlan **Issue1** and **Issue2** don't reproduce in .Net 7.0 SDK build:
image
image
image
image

And .Net 8.0 without your this PR change has this issue https://github.com/dotnet/winforms/issues/8456 itself and also repro above issue1 with setting ShotcutKeys as Ctrl+Alt+Shift+A, issue2 with setting "Ctrl+Alt+Shift+F1" changes "Enter+F12+F11+F10+End+Ctrl+F8+F9+Alt+F4+F5+F6+F7+F1+F2+F3+PgDn+Ins+Home+Del+Shift+PgUp+Back+0+1+2+3+4+5+6+7+8+9+0" automatically

elachlan:
So this PR has not caused the additional regressions that were found. Which is good to know.

This will require some guidance from the team.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripMenuItemTests.cs
```diff
@@ -120,6 +120,26 @@ public void ToolStripMenuItem_GetNativeMenuItemImage_ReturnsExpected(int nativeM
...
 
+        [WinFormsTheory]
+        [InlineData(Keys.None)]
+        [InlineData(Keys.Control | Keys.Alt | Keys.Shift | Keys.A)]
+        [InlineData(Keys.Control | Keys.Alt | Keys.Shift | Keys.F1)]
+        [InlineData(Keys.F1)]
+        public void ToolStripMenuItem_SetShortcutKeys(Keys keys)
+        {
+            using var item = new SubToolStripMenuItem();
+            item.ShortcutKeys=keys;
```
dmitrii-drobotov, line 131:
```suggestion
            item.ShortcutKeys = keys;
```
Same in line 140

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripMenuItemTests.cs
```diff
@@ -120,6 +120,26 @@ public void ToolStripMenuItem_GetNativeMenuItemImage_ReturnsExpected(int nativeM
...
+            Assert.Equal(keys, item.ShortcutKeys);
+        }
+
+		[WinFormsTheory]
+        [InlineData(Keys.A)]
```
dmitrii-drobotov, line 136: Consider adding more test cases, e.g. just modifier keys or `Keys.None`.

dreddy-work:
> Okay, for what ever reason the designer has serialized `Ctrl+Alt+Shift+A` as `Keys.A`. This fails because the shortcut isn't valid, it fails the validation logic and throws `InvalidEnumArgumentException`.

@Olina-Zhang , can you open an issue in the designer repo for this. We will investigate this in the designer serialization scope.

dreddy-work:
I will take a look at serialization part later this week and update.

dmitrii-drobotov (commented):
ConvertTo fix looks good to me. Regarding missing modifiers issue I can see it reproduces only with 2 or more modifiers, so for example setting "Ctrl+A" works well, but "Ctrl+Shift+A" doesn't.

dmitrii-drobotov:
I think this line also need to be replaced by `if (modifiers.HasFlag(keyValue))` and it should resolve the issue:
https://github.com/dotnet/winforms/blob/1be4ac0524287c704cc14b4f28fb580649bf0f1e/src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs#L243

@elachlan - I assume it would reproduce in a unit test, e.g. check that `converter.ConvertTo(Keys.Control | Keys.Shift | Keys.A, typeof(Enum[]))` converts to `[Keys.Control, Keys.Shift, Keys.A]`.

elachlan:
@Olina-Zhang could your team re-test this please?

Nora-Zhou01:
Tested private binaries built by us for PR: 8485 based on the latest .NET 8.0 SDK build, GH issue 8456 is fixed，and the two regressions caused have been resolved.
**Regressions Issue1:**
R
**Regressions Issue2:**
F1

dmitrii-drobotov (approved):

dreddy-work (approved):

dreddy-work:
Thank you @elachlan

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/KeysConverterTests.cs
```diff
@@ -9,15 +9,52 @@ namespace System.Windows.Forms.Tests
...
+
+        [Theory]
+        [InlineData(Keys.None, "None")]
+        [InlineData(Keys.S, "S")]
+        [InlineData(Keys.Control | Keys.C, "Ctrl+C")]
+        [InlineData(Keys.Control | Keys.Add, "Ctrl+Add")]
+        [InlineData(Keys.Control | Keys.Alt | Keys.D, "Ctrl+Alt+D")]
+        [InlineData(Keys.Control | Keys.Alt | Keys.Shift | Keys.A, "Ctrl+Alt+Shift+A")]
+        [InlineData(Keys.Control | Keys.Alt | Keys.Shift | Keys.F1, "Ctrl+Alt+Shift+F1")]
+        public void ConvertToString_ShouldConvertKeys(Keys keys, string expectedResult)
```
RussKie, line 33: Would `Keys.Alt | Keys.Control | Keys.Shift` still result in "Ctrl+Alt+Shift"?

elachlan: I didn't touch the display order here, but future PRs could cause issues. So it is a good idea for tests. Dmitrii raised this concern in a related PR: https://github.com/dotnet/winforms/pull/8401#discussion_r1071256450

elachlan: On further thought, what I noticed with the keys enum is that the enum value order was always the same after assignment. So the order of the key values doesn't matter. The ordering is set via the `DisplayOrder` dictionary, so it needs to be preserved for the order to also be preserved.

PR #8488, opened 2023-01-11
[RUS] The ‘Details’ button text overlaps its drop down arrow in unhandled exception dialog

### LINKED ISSUE #6490: [RUS] The ‘Details’ button text overlaps its drop down arrow in unhandled exception dialog
.NET Core Version: 

- .NET SDK 7.0.100-alpha.1.22060.9

- No

**Problem description:**
The text overlaps the drop down arrow in ‘Details’ button of Unhandled exception window.
RUS-Core

**Expected behavior:**
There should be no overlapping.
NetFX:
RUS-NetFX

.NET Core 3.1:
RUS-Core31

**Minimal repro:**
1.	Create a Winforms core project.
2.	Add a button control on Form designer.
3.	Double click the button and add following code: `throw new Exception("")`; in `button_click` event handler.
4.	Run the project via pressing `Ctrl + F5`.
Or use the sample project with changing the target to 7.0: 
[UnhandledExceptionDialog.zip](https://github.com/dotnet/winforms/files/7859422/UnhandledExceptionDialog.zip)

**More Info:**
1. This is a regression issue, it can’t reproduce when targeting .net core 3.1, it reproduce from .net 5.0 to latest .net 7.0.
2. This issue can’t reproduce with .NET Framework project.

elachlan:
Fixes #6490

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8488)

### DIFF
--- a/src/System.Windows.Forms/src/Resources/xlf/SR.ru.xlf
+++ b/src/System.Windows.Forms/src/Resources/xlf/SR.ru.xlf
@@ -4914,7 +4914,7 @@ If you click Continue, the application will ignore this error and attempt to con
       </trans-unit>
       <trans-unit id="ExDlgShowDetails">
         <source>&amp;Details</source>
-        <target state="translated">&amp;Подробности</target>
+        <target state="translated">&amp;сведения</target>
         <note />
       </trans-unit>
       <trans-unit id="ExDlgWarningText">
--- a/src/System.Windows.Forms/src/Resources/xlf/SR.ru.xlf
+++ b/src/System.Windows.Forms/src/Resources/xlf/SR.ru.xlf
@@ -4914,7 +4914,7 @@ If you click Continue, the application will ignore this error and attempt to con
       </trans-unit>
       <trans-unit id="ExDlgShowDetails">
         <source>&amp;Details</source>
-        <target state="translated">&amp;сведения</target>
+        <target state="translated">&amp;Подробности</target>
         <note />
       </trans-unit>
       <trans-unit id="ExDlgWarningText">
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ThreadExceptionDialog.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ThreadExceptionDialog.cs
@@ -30,9 +30,9 @@ namespace System.Windows.Forms
         private const int BUTTONDETAILS_LEFTPADDING = 8;
         private const int MESSAGE_TOPPADDING = 8;
         private const int HEIGHTPADDING = 8;
-        private const int BUTTONWIDTH = 100;
+        private const int BUTTONWIDTH = 130;
         private const int BUTTONHEIGHT = 23;
-        private const int BUTTONALIGNMENTWIDTH = 105;
+        private const int BUTTONALIGNMENTWIDTH = 135;
         private const int BUTTONALIGNMENTPADDING = 5;
         private const int DETAILSWIDTHPADDING = 16;
         private const int DETAILSHEIGHT = 154;
@@ -307,7 +307,7 @@ namespace System.Windows.Forms
             _helpButton.FlatStyle = FlatStyle.Standard;
             _helpButton.DialogResult = DialogResult.Yes;
 
-            _detailsButton.Text = SR.ExDlgShowDetails;
+            _detailsButton.Text = "&Подробности";
             _detailsButton.FlatStyle = FlatStyle.Standard;
             _detailsButton.Click += new EventHandler(DetailsClick);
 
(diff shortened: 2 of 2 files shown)


src/System.Windows.Forms/src/Resources/xlf/SR.ru.xlf
```diff
@@ -4914,7 +4914,7 @@ If you click Continue, the application will ignore this error and attempt to con
       </trans-unit>
       <trans-unit id="ExDlgShowDetails">
         <source>&amp;Details</source>
-        <target state="translated">&amp;Подробности</target>
+        <target state="translated">&amp;сведения</target>
```
RussKie, line 4917: The translations are performed by an external system, so it should be updated there rather than here.

elachlan: Sorry about that. I did not know. Should I close this PR? or should it pivot to increase the size of the button?

Tanya-Solyanik: Resizing the button sounds great!

elachlan: Do all the buttons need to be the same size? I was just going to enable auto-size on the details button.

dreddy-work: They are all same size now. If we can make them same size after this change and doesn't impact any other languages, we should go that direction.

elachlan:
image

src/System.Windows.Forms/src/System/Windows/Forms/ThreadExceptionDialog.cs
```diff
@@ -307,7 +307,7 @@ public ThreadExceptionDialog(Exception t)
             _helpButton.FlatStyle = FlatStyle.Standard;
             _helpButton.DialogResult = DialogResult.Yes;
 
-            _detailsButton.Text = SR.ExDlgShowDetails;
+            _detailsButton.Text = "&Подробности";
```
dreddy-work, line 310: forgot to revert back?

src/System.Windows.Forms/src/Resources/xlf/SR.ru.xlf
```diff
@@ -4914,7 +4914,7 @@ If you click Continue, the application will ignore this error and attempt to con
       </trans-unit>
       <trans-unit id="ExDlgShowDetails">
         <source>&amp;Details</source>
-        <target state="translated">&amp;сведения</target>
```
dreddy-work: and, we shouldn't change loc string. Is this wrongly translated?

src/System.Windows.Forms/src/System/Windows/Forms/ThreadExceptionDialog.cs
elachlan: Sorry, yes.

src/System.Windows.Forms/src/Resources/xlf/SR.ru.xlf
dreddy-work: nevermid. this was from earlier commit.

elachlan:
I thought maybe we could do something "Smart" like measure the text length for each button and use the max. But I think measure text isn't working exactly and its easier just to have slightly bigger buttons.

dreddy-work (approved):

src/System.Windows.Forms/src/System/Windows/Forms/ThreadExceptionDialog.cs
```diff
@@ -30,9 +30,9 @@ public class ThreadExceptionDialog : Form
         private const int BUTTONDETAILS_LEFTPADDING = 8;
         private const int MESSAGE_TOPPADDING = 8;
         private const int HEIGHTPADDING = 8;
-        private const int BUTTONWIDTH = 100;
+        private const int BUTTONWIDTH = 130;
         private const int BUTTONHEIGHT = 23;
-        private const int BUTTONALIGNMENTWIDTH = 105;
+        private const int BUTTONALIGNMENTWIDTH = 135;
```
RussKie, line 35: I would caution against hardcoding widths as different languages have different lengths for strings. It's best to use `AutoSize=true`.

elachlan: I thought about this, but it was hardcoded to begin with and `AutoSize=true` means the buttons aren't synced.

Issue #8414, opened 2022-12-22
[dotnet-sdk-8.0.100-alpha.1.22619.1] System.IndexOutOfRangeException: 'Index was outside the bounds of the array.' exception thrown when OpenFileDialog.Multiselect=true


Junjun-zhao:
### .NET version

```
.NET SDK:
Version:   8.0.100-alpha.1.22619.1
Commit:    8750a90545
Runtime Environment:
OS Name:     Windows
OS Version:  10.0.19045
OS Platform: Windows
RID:         win10-x64
Base Path:   C:\Program Files\dotnet\sdk\8.0.100-alpha.1.22619.1\

Host:
 Version:      8.0.0-alpha.1.22605.1
 Architecture: x64
 Commit:       1a37caf773

.NET SDKs installed:
  8.0.100-alpha.1.22619.1 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 8.0.0-alpha.1.22615.1 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 8.0.0-alpha.1.22605.1 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 8.0.0-alpha.1.22614.1 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
```

### Did it work in .NET Framework?

Yes

### Did it work in any of the earlier releases of .NET Core or .NET 5+?

**Verify Scenarios:**
1). Windows10 21H2 x64+dotnet-sdk-8.0.100-alpha.1.22621.5: Fail
2). Windows10 21H2 x64+dotnet-sdk-8.0.100-alpha.1.22619.1: Fail
3). Windows10 21H2 x64+dotnet-sdk-8.0.100-alpha.1.22616.7: Fail
4). Windows10 21H2 x64+dotnet-sdk-7.0.101: Pass

### Issue description

When testing the 3rd Party Apps with .NET 8 runtime, System.IndexOutOfRangeException: 'Index was outside the bounds of the array.' exception thrown when OpenFileDialog.Multiselect=true

**Application Name:**  paint.netcore, XamlViewer
**OS:** Windows 10 21H2
**CPU:** X64
**.NET Build Number:** dotnet-sdk-8.0.100-alpha.1.22619.1

[...]

PR #8500, opened 2023-01-13
Fix multiselect in OpenFileDialog

JeremyKuhne:
Query for the right interface before getting results and guard correctly.

Adds regression test and fixes a mistake in ComScope.

Adds a little cleanup for easier debugging and I left in a few defines that are difficult to find and we'll want to use soon. (I tried a variety of approaches to automating, including sending key messages and using accessibility. Sending WM_COMMAND is far simpler.)

fixes: #8414

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8500)


elachlan:
test failures due to usings: `Using directive is unnecessary.`

src/System.Windows.Forms/tests/TestUtilities/DialogHostForm.cs
```diff
@@ -19,10 +20,19 @@ protected override void WndProc(ref Message m)
...
+
+    protected virtual void OnDialogIdle(HWND dialogHandle)
+    {
+        PInvoke.PostMessage(dialogHandle, WM.CLOSE);
+    }
+
+    protected unsafe void Accept(HWND handle)
+    {
+        PInvoke.SendMessage(handle, WM.COMMAND, (WPARAM)(nint)MESSAGEBOX_RESULT.IDOK);
+    }
```
elachlan, line 37: I love these changes! They are so simple!

dreddy-work (approved):

PR #8509, opened 2023-01-16
Refactor `ToolStripControlHost.Control` property to throw `ObjectDisposedException` if null

### LINKED ISSUE #2466: NullReferenceException in ToolStripControlHost
* .NET Core Version: 3.0.100

**Problem description:**
NullReferenceException while calling various methods on ToolStripControlHost after disposal.

**Actual behavior:** 
`NullReferenceException` thrown
> System.NullReferenceException: Object reference not set to an instance of an object.
>   \+ System.Windows.Forms.ToolStripControlHost.SetVisibleCore(bool)

**Expected behavior:**
`ObjectDisposedException` thrown

**Minimal repro:**
The following code was used in Roslin to reproduce the issue
> #r "System.Windows.Forms"
> using System.Windows.Forms;
> Button btnTest = new Button();
> ToolStripControlHost host = new ToolStripControlHost(btnTest);
> host.Dispose();
> host.Visible = true;

Note: From what I see in the source code, it does not need to be on a visible form to reproduce.

**Suspected Cause**
`Dispose` sets `control` to `null`

**Suggested Solution**
Check for null in the Control property and throw an `ObjectDisposedException` if it is null

elachlan:
Refactored `ToolStripControlHost.Control` property to throw `ObjectDisposedException` if null.
Also some minor refactoring using expression bodies

Fixes #2466

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8509)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.cs
@@ -86,7 +86,7 @@ namespace System.Windows.Forms
         /// <summary>
         ///  Overridden to return value from Control.CanSelect.
         /// </summary>
-        public override bool CanSelect => _control is not null ? DesignMode || Control.CanSelect : false;
+        public override bool CanSelect => _control is not null && (DesignMode || _control.CanSelect);
 
         [SRCategory(nameof(SR.CatFocus))]
         [DefaultValue(true)]
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.cs
@@ -6,6 +6,7 @@
 
 using System.ComponentModel;
 using System.Drawing;
+using System.Runtime.CompilerServices;
 using System.Windows.Forms.Layout;
 
 namespace System.Windows.Forms
@@ -126,6 +127,7 @@ namespace System.Windows.Forms
         /// </summary>
         private Control ControlInternal
         {
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
             get
             {
                 ObjectDisposedException.ThrowIf(_control is null, _control);
@@ -806,7 +808,7 @@ namespace System.Windows.Forms
         private void SyncControlParent()
         {
             ReadOnlyControlCollection newControls = GetControlCollection(ParentInternal);
-            newControls?.AddInternal(ControlInternal);
+            newControls?.AddInternal(_control);
         }
 
         protected virtual void OnHostedControlResize(EventArgs e)
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.cs
@@ -123,7 +123,7 @@ namespace System.Windows.Forms
         {
             get
             {
-                ObjectDisposedException.ThrowIf(_control is null, nameof(Control));
+                ObjectDisposedException.ThrowIf(_control is null, Control);
                 return _control;
             }
         }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.cs
@@ -86,18 +86,7 @@ namespace System.Windows.Forms
         /// <summary>
         ///  Overriden to return value from Control.CanSelect.
         /// </summary>
-        public override bool CanSelect
-        {
-            get
-            {
-                if (_control is not null)
-                {
-                    return (DesignMode || Control.CanSelect);
-                }
-
-                return false;
-            }
-        }
+        public override bool CanSelect => _control is not null ? DesignMode || Control.CanSelect : false;
 
         [SRCategory(nameof(SR.CatFocus))]
         [DefaultValue(true)]
@@ -130,7 +119,14 @@ namespace System.Windows.Forms
         /// </summary>
         [Browsable(false)]
         [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
-        public Control Control => _control;
+        public Control Control
+        {
+            get
+            {
+                ObjectDisposedException.ThrowIf(_control is null, nameof(Control));
+                return _control;
+            }
+        }
 
         internal AccessibleObject ControlAccessibilityObject => Control?.AccessibilityObject;
 
@@ -323,19 +319,16 @@ namespace System.Windows.Forms
         {
             get
             {
-                if (_control is not null)
-                {
-                    return _control.RightToLeft;
-                }
-
-                return base.RightToLeft;
+                return _control is not null ? _control.RightToLeft : base.RightToLeft;
             }
             set
             {
-                if (_control is not null)
+                if (_control is null)
                 {
-                    _control.RightToLeft = value;
+                    return;
                 }
+
+                _control.RightToLeft = value;
             }
         }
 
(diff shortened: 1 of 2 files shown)


elachlan:
Looking at the test stack trace, it does not make sense to me. I can't see how the `ToolStripControlHost` calls to its base constructor (`ToolStripItem`).

elachlan:
Debugging tests is still broken for me. It's slowed me down a bit.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.cs
```diff
@@ -130,7 +119,14 @@ public ContentAlignment ControlAlign
         /// </summary>
         [Browsable(false)]
         [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
-        public Control Control => _control;
+        public Control Control
+        {
+            get
+            {
+                ObjectDisposedException.ThrowIf(_control is null, _control);
```
Tanya-Solyanik, line 126: This code will throw when we are invoking getter to test the result for null. See line 131 below. Please make sure that we are not using this property internally for null-tests. I would add ControlInternal that throws ObjectDisposed to be used where previously we had NullRefs, and keep `Control` as is, because 3rd parties would rely on Control property to skip event that get raised after the host is disposed.

@dreddy-work - what is the most consistent approach to align with the nullability work? We could get rid of the nullref and just ignore the invocation after the disposal.

elachlan: is `ObjectDisposedException` correct, would `InvalidOperationException` make more sense? Is `ObjectDisposedException` only supposed to happen after it has been assigned and subsequently nulled?

Maybe where ever we used the `Control` property internally, we just check `_control` instead and add `ObjectDisposedException.ThrowIf(_control is null, _control);` to check it.

Tanya-Solyanik: ObjectDisposed is the correct exception to throw, if we want to throw in this case. In many cases just ignoring a disposed control would work, this is why it's best to be consistent with the direction nullability work takes us.

>Maybe where ever we used the Control property internally, we just check _control instead and add ObjectDisposedException.ThrowIf(_control is null, _control); to check it.

This is not sufficient, the 3rd party users of this property might be doing `Control?.` as well, this is valid and should not throw.

elachlan:
The refactor uses the private field `_control` where a null check is made and the `ControlInternal` property when an exception is expected.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.cs
```diff
@@ -119,7 +119,14 @@ public ContentAlignment ControlAlign
...
         [Browsable(false)]
         [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
-        public Control Control
+        public Control Control => _control;
+
+        /// <summary>
+        ///  The control that this item is hosting.
+        /// </summary>
+        [Browsable(false)]
+        [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
```
Tanya-Solyanik, line 128: These attributes are not actually needed because this property is private and it does not have a set method. Only public or protected properties that can be set are serialized into `InitilizeComponent` And only public or protected properties show up in the property browser.

Create a UserControl with these properties:
```
 private int _field;
        public int AProperty
        {
            get { return _field; }
            set { _field = value; }
        }
        public int APropertyNoSetter
        {
            get { return _field; }
        }
        private int APrivateProperty
        {
            get { return _field; }
            set { _field = value; }
        }
        private int APrivatePropertyNoSetter
        {
            get { return _field; }
        }
```

Drop it on the form:

```
            // 
            // userControl11
            // 
            this.userControl11.AProperty = 0;
            this.userControl11.Location = new System.Drawing.Point(0, 0);
            this.userControl11.Name = "userControl11";
            this.userControl11.Size = new System.Drawing.Size(150, 150);
            this.userControl11.TabIndex = 0;

```
Look it up in the property browser:
image

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.cs
```diff
@@ -132,7 +121,21 @@ public ContentAlignment ControlAlign
...
+        private Control ControlInternal
+        {
+            get
+            {
+                ObjectDisposedException.ThrowIf(_control is null, _control);
+                return _control;
+            }
+        }
+
+        internal AccessibleObject ControlAccessibilityObject => _control?.AccessibilityObject;
```
dreddy-work, line 138:
```suggestion
        internal AccessibleObject? ControlAccessibilityObject => _control?.AccessibilityObject;
```

or 

```suggestion
        internal AccessibleObject? ControlAccessibilityObject => ControlInternal.AccessibilityObject;
```

Tanya-Solyanik: This property should not throw. we are getting unexpected calls into it from the accessibility tools which we don't know how to get rid of. Accessible objects are not purged from the clients when the control is disposed.

elachlan: There is test coverage for that. So when I tried, there were test failures. I'll mark it as nullable.

elachlan: We haven't enabled nullability on this class yet. When we do it will change to `AccessibleObject?`

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.cs
```diff
@@ -79,33 +79,22 @@ public override Image BackgroundImage
...
-            {
-                if (_control is not null)
-                {
-                    return (DesignMode || Control.CanSelect);
-                }
-
-                return false;
-            }
-        }
+        public override bool CanSelect => _control is not null ? DesignMode || Control.CanSelect : false;
```
Tanya-Solyanik, line 89: Missed it the first time around - there is no reason to use backing field in one spot and property in another, please use field in both. Also consider reformatting the ternary conditional expression so that it stands out.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.cs
```diff
@@ -132,7 +121,19 @@ public ContentAlignment ControlAlign
         [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
         public Control Control => _control;
 
-        internal AccessibleObject ControlAccessibilityObject => Control?.AccessibilityObject;
+        /// <summary>
+        ///  The control that this item is hosting.
+        /// </summary>
+        private Control ControlInternal
```
Tanya-Solyanik, line 127: This property might get inlined if we add [MethodImpl(MethodImplOptions.AggressiveInlining)] attributebecause it uses exception builder.

Tanya-Solyanik:
Adding `documentation: breaking` because this fix changed exception type.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripControlHost.cs
elachlan: You can only use `MethodImplAttribute` on methods or constructors.
https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.methodimplattribute?view=net-6.0

Tanya-Solyanik: You can put directly above the `get` method name.

dmitrii-drobotov (approved):

Tanya-Solyanik (approved):

elachlan:
@dreddy-work is this okay to merge?

Tanya-Solyanik:
The NullReferenceException had never been documented in the changed APIs, we assume that the developers had not taken a dependency on it, thus taking the documentation breaking label off.

PR #8518, opened 2023-01-24
Removed the DefaultValue attribute from ListBox.ItemHeight property

### LINKED ISSUE #4463: The code generated by the ListBox control is not consistent with Framework

dkazennov:
Fixes #4463 (partially, see also #8640)

## Proposed changes

- Introduces a new private property `DefaultHeight` with **the intended value** equals Winforms Control.DefaultFont.Height (for example in Windows.Forms it is **Segoe UI, Size 9, Height 16** by default).
- Removes the DefaultValue attribute from the `ListBox.ItemHeight` property;
- Implements `ResetItemHeight` and `ShouldSerializeItemHeight` methods ([see more](https://learn.microsoft.com/en-us/dotnet/desktop/winforms/controls/defining-default-values-with-the-shouldserialize-and-reset-methods?view=netframeworkdesktop-4.8)).
- The ListBox.DefaultItemHeight constant marked as obsolete. The `ListBox.ItemHeight` is now depended on actual `DefaultFont.Height` value.
- The `ListBox.ItemHeight` property still returns the value of the  `_itemHeight` property as usual.
- Changed 9 ListBox unit tests

## Customer Impact

- Fixes the Issue which impacts user experience,  breaks DPI and Font scaling. For the _newly created_ control at 100% scaling, designer should generate the same value as is defined by the Default. There is no extra code.

## Regression? 

- No

## Risk

- Minimal

### Before

image

### After

image

## Test methodology

- Manual (Issue can be tested via substitution of dlls, complete restart of the VS recommended)
- Unit test

 

## Test environment(s) 

- dotnet 8.0.100-alpha.1.22607.6

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8518)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs
@@ -62,8 +62,9 @@ namespace System.Windows.Forms
         private SelectedIndexCollection? _selectedIndices;
         private ObjectCollection? _itemsCollection;
 
-        private int _itemHeight = DefaultItemHeight;
+        private int _itemHeight = DefaultListBoxItemHeight;
         private int _columnWidth;
+        private static int s_defaultListBoxItemHeight = -1;
         private int _requestedHeight;
         private int _topIndex;
         private int _horizontalExtent;
@@ -389,6 +390,19 @@ namespace System.Windows.Forms
             }
         }
 
+        private static int DefaultListBoxItemHeight
+        {
+            get
+            {
+                if (s_defaultListBoxItemHeight == -1)
+                {
+                    s_defaultListBoxItemHeight = DefaultFont.Height;
+                }
+
+                return s_defaultListBoxItemHeight;
+            }
+        }
+
         protected override Size DefaultSize
         {
             get
@@ -586,7 +600,6 @@ namespace System.Windows.Forms
         ///  the height of an item in an owner-draw list box.
         /// </summary>
         [SRCategory(nameof(SR.CatBehavior))]
-        [DefaultValue(DefaultItemHeight)]
         [Localizable(true)]
         [SRDescription(nameof(SR.ListBoxItemHeightDescr))]
         [RefreshProperties(RefreshProperties.Repaint)]
@@ -2107,9 +2120,10 @@ namespace System.Windows.Forms
             base.ResetForeColor();
         }
 
+        // ShouldSerialize and Reset Methods are being used by Designer via reflection.
         private void ResetItemHeight()
         {
-            _itemHeight = DefaultItemHeight;
+            _itemHeight = DefaultListBoxItemHeight;
         }
 
         protected override void ScaleControl(SizeF factor, BoundsSpecified specified)
@@ -2206,6 +2220,12 @@ namespace System.Windows.Forms
 
             SelectedItems.Dirty();
             OnSelectedIndexChanged(EventArgs.Empty);
+        }
+
+        // ShouldSerialize and Reset Methods are being used by Designer via reflection.
+        private bool ShouldSerializeItemHeight()
+        {
+            return ItemHeight != DefaultListBoxItemHeight;
         }
 
         /// <summary>
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListBoxTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListBoxTests.cs
@@ -86,7 +86,7 @@ namespace System.Windows.Forms.Tests
             Assert.True(control.IntegralHeight);
             Assert.False(control.IsAccessible);
             Assert.False(control.IsMirrored);
-            Assert.Equal(13, control.ItemHeight);
+            Assert.Equal(Control.DefaultFont.Height, control.ItemHeight);
             Assert.Empty(control.Items);
             Assert.Same(control.Items, control.Items);
             Assert.NotNull(control.LayoutEngine);
@@ -100,7 +100,7 @@ namespace System.Windows.Forms.Tests
             Assert.Equal(Padding.Empty, control.Padding);
             Assert.Null(control.Parent);
             Assert.Equal("Microsoft\u00AE .NET", control.ProductName);
-            Assert.Equal(13 + SystemInformation.BorderSize.Height * 4 + 3, control.PreferredHeight);
+            Assert.Equal(Control.DefaultFont.Height + SystemInformation.BorderSize.Height * 4 + 3, control.PreferredHeight);
             Assert.Equal(new Size(120, 96), control.PreferredSize);
             Assert.False(control.RecreatingHandle);
             Assert.Null(control.Region);
@@ -1742,13 +1742,13 @@ namespace System.Windows.Forms.Tests
             foreach (bool integralHeight in new bool[] { true, false })
             {
                 yield return new object[] { DrawMode.Normal, integralHeight, 1, 0 };
-                yield return new object[] { DrawMode.Normal, integralHeight, 13, 0 };
+                yield return new object[] { DrawMode.Normal, integralHeight, Control.DefaultFont.Height, 0 };
                 yield return new object[] { DrawMode.Normal, integralHeight, 255, 0 };
                 yield return new object[] { DrawMode.OwnerDrawFixed, integralHeight, 1, 1 };
-                yield return new object[] { DrawMode.OwnerDrawFixed, integralHeight, 13, 0 };
+                yield return new object[] { DrawMode.OwnerDrawFixed, integralHeight, Control.DefaultFont.Height, 0 };
                 yield return new object[] { DrawMode.OwnerDrawFixed, integralHeight, 255, 1 };
                 yield return new object[] { DrawMode.OwnerDrawVariable, integralHeight, 1, 0 };
-                yield return new object[] { DrawMode.OwnerDrawVariable, integralHeight, 13, 0 };
+                yield return new object[] { DrawMode.OwnerDrawVariable, integralHeight, Control.DefaultFont.Height, 0 };
                 yield return new object[] { DrawMode.OwnerDrawVariable, integralHeight, 255, 0 };
             }
         }
@@ -1825,7 +1825,7 @@ namespace System.Windows.Forms.Tests
             Assert.True(property.CanResetValue(control));
 
             property.ResetValue(control);
-            Assert.Equal(13, control.ItemHeight);
+            Assert.Equal(Control.DefaultFont.Height, control.ItemHeight);
             Assert.False(property.CanResetValue(control));
         }
 
@@ -1841,7 +1841,7 @@ namespace System.Windows.Forms.Tests
             Assert.True(property.ShouldSerializeValue(control));
 
             property.ResetValue(control);
-            Assert.Equal(13, control.ItemHeight);
+            Assert.Equal(Control.DefaultFont.Height, control.ItemHeight);
             Assert.False(property.ShouldSerializeValue(control));
         }
 
@@ -2050,13 +2050,13 @@ namespace System.Windows.Forms.Tests
         public static IEnumerable<object[]> PreferredHeight_GetEmpty_TestData()
         {
             int extra = SystemInformation.BorderSize.Height * 4 + 3;
-            yield return new object[] { DrawMode.Normal, BorderStyle.Fixed3D, 13 + extra };
-            yield return new object[] { DrawMode.Normal, BorderStyle.FixedSingle, 13 + extra };
-            yield return new object[] { DrawMode.Normal, BorderStyle.None, 13 };
+            yield return new object[] { DrawMode.Normal, BorderStyle.Fixed3D, Control.DefaultFont.Height + extra };
+            yield return new object[] { DrawMode.Normal, BorderStyle.FixedSingle, Control.DefaultFont.Height + extra };
+            yield return new object[] { DrawMode.Normal, BorderStyle.None, Control.DefaultFont.Height };
 
-            yield return new object[] { DrawMode.OwnerDrawFixed, BorderStyle.Fixed3D, 13 + extra };
-            yield return new object[] { DrawMode.OwnerDrawFixed, BorderStyle.FixedSingle, 13 + extra };
-            yield return new object[] { DrawMode.OwnerDrawFixed, BorderStyle.None, 13 };
+            yield return new object[] { DrawMode.OwnerDrawFixed, BorderStyle.Fixed3D, Control.DefaultFont.Height + extra };
+            yield return new object[] { DrawMode.OwnerDrawFixed, BorderStyle.FixedSingle, Control.DefaultFont.Height + extra };
+            yield return new object[] { DrawMode.OwnerDrawFixed, BorderStyle.None, Control.DefaultFont.Height };
 
             yield return new object[] { DrawMode.OwnerDrawVariable, BorderStyle.Fixed3D, extra };
             yield return new object[] { DrawMode.OwnerDrawVariable, BorderStyle.FixedSingle, extra };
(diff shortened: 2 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs
```diff
@@ -51,6 +51,7 @@ public partial class ListBox : ListControl
         /// <summary>
         ///  The default item height for an owner-draw ListBox. The ListBox's non-ownerdraw
         ///  item height is 13 for the default font on Windows.
+        ///  This constant is obsolete.
```
dmitrii-drobotov, line 54: To properly obsolete a member, you need to use `ObsoleteAttribute`, for example: https://github.com/dotnet/winforms/blob/37e19459f4554e4b69de9cb5f95d623d272c899d/src/System.Windows.Forms/src/System/Windows/Forms/DomainUpDown.DomainItemAccessibleObject.cs#L12-L17
And follow the process described in https://github.com/dotnet/winforms/blob/main/docs/list-of-diagnostics.md

src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs
```diff
@@ -389,6 +390,11 @@ public bool UseCustomTabOffsets
             }
         }
 
+        private static int DefaultHeight
```
dmitrii-drobotov, line 393: Name `ListBox.DefaultHeight` sounds like it refers to the whole `ListBox`, rather than to the height of a single item. What do you think about `DefaultItemHeightInternal`?

Tanya-Solyanik: `Internal` implies that there is another property for external use. We are actually deprecating that other `const`.  Though I can't come up with a great name... `DefaultListBoxItemHeight`?

NikitaSemenovAkvelon (approved):
LGTM

dkazennov:
Some builds failed due to issues with integration tests unrelated to this fix (and those test are not shown in the list of tests in the pipeline description).

src/Common/src/Obsoletions.cs
```diff
@@ -21,4 +21,9 @@ internal static class Obsoletions
     internal const string DomainItemAccessibleObjectMessage = $"{nameof(DomainUpDown.DomainItemAccessibleObject)} is no longer used to provide accessible support for {nameof(DomainUpDown)} items.";
 #pragma warning restore WFDEV003 // Type or member is obsolete
     internal const string DomainItemAccessibleObjectDiagnosticId = "WFDEV003";
+
+#pragma warning disable WFDEV004
+    internal const string ListBoxDefaultItemHeightMessage = $"The {nameof(ListBox.DefaultItemHeight)} const is replaced with the ListBox.DefaultListBoxItemHeight property.";
```
dmitrii-drobotov, line 26: Referring to a private property may not be helpful for those who will want to replace the const in their code. I suggest mentioning what can be used instead and also saying that default height is now dependent on application's font.

```suggestion
    internal const string ListBoxDefaultItemHeightMessage = $"{nameof(ListBox.DefaultItemHeight)} is no longer used as default item height, which now depends on application's default font. Use {nameof(ListBox.ItemHeight)} instead.";
```

src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs
```diff
@@ -389,6 +394,11 @@ public bool UseCustomTabOffsets
             }
         }
 
+        private static int DefaultListBoxItemHeight
+        {
+            get => DefaultFont.Height;
+        }
```
dmitrii-drobotov, line 400: Nit:
```suggestion
        private static int DefaultListBoxItemHeight => DefaultFont.Height;
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListBoxTests.cs
```diff
@@ -1382,6 +1382,30 @@ public static IEnumerable<object[]> HorizontalExtent_Set_TestData()
             }
         }
 
+        [WinFormsFact]
+        public void ListBox_Height_IsEqualDefaultHeight_IfDefaultFontIsChanged()
+        {
+            var oldApplicationDefaultFont = Application.DefaultFont;
+            using var font = new Font("Times New Roman", 12);
+            using var listBox = new TestListBox();
```
dmitrii-drobotov, line 1390:
```suggestion
            using Font font = new("Times New Roman", 12);
            using TestListBox listBox = new();
```

dmitrii-drobotov (commented):
Can you also update the table in https://github.com/dotnet/winforms/blob/main/docs/list-of-diagnostics.md in this PR?

src/Common/src/Obsoletions.cs
dkazennov: Yes, that's a good point here.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListBoxTests.cs
```diff
@@ -1382,6 +1382,30 @@ public static IEnumerable<object[]> HorizontalExtent_Set_TestData()
...
+            using Font font = new Font("Times New Roman", 12);
+            using TestListBox listBox = new TestListBox();
+            var applicationTestAccessor = typeof(Application).TestAccessor().Dynamic;
+
+            try
+            {
+                applicationTestAccessor.s_defaultFont = font;
+                Assert.NotEqual(oldApplicationDefaultFont, Application.DefaultFont);
+
+                var rowDefaultHeight = listBox.GetDefaultListBoxItemHeight();
```
dmitrii-drobotov, line 1398: Since `GetDefaultListBoxItemHeight` is used only here, perhaps it can be inlined and then TestListBox can be removed?
```suggestion
                int rowDefaultHeight = listBox.TestAccessor().Dynamic.DefaultListBoxItemHeight;
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListBoxTests.cs
```diff
@@ -1382,6 +1382,30 @@ public static IEnumerable<object[]> HorizontalExtent_Set_TestData()
             }
         }
 
+        [WinFormsFact]
+        public void ListBox_Height_IsEqualDefaultHeight_IfDefaultFontIsChanged()
+        {
+            var oldApplicationDefaultFont = Application.DefaultFont;
+            using Font font = new Font("Times New Roman", 12);
+            using TestListBox listBox = new TestListBox();
```
dmitrii-drobotov, line 1390:
```suggestion
            using Font font = new("Times New Roman", 12);
            using TestListBox listBox = new();
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListBoxTests.cs
```diff
@@ -1382,6 +1382,30 @@ public static IEnumerable<object[]> HorizontalExtent_Set_TestData()
             }
         }
 
+        [WinFormsFact]
+        public void ListBox_Height_IsEqualDefaultHeight_IfDefaultFontIsChanged()
```
dmitrii-drobotov, line 1386:
```suggestion
        public void ListBox_ItemHeight_EqualsDefaultFontHeight_IfDefaultFontIsChanged()
```

dmitrii-drobotov (commented):
A few more nits, otherwise looks good!

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListBoxTests.cs
dkazennov: Yes. Now than you've pointed it out, I understood that I expected this class to be more useful.

src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs
```diff
@@ -586,7 +593,6 @@ public bool IntegralHeight
         ///  the height of an item in an owner-draw list box.
         /// </summary>
         [SRCategory(nameof(SR.CatBehavior))]
-        [DefaultValue(DefaultItemHeight)]
```
Tanya-Solyanik, line 589: I see this as a not observable change because we replace functionality with SholdSerialize and reset methods. https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/breaking-change-rules.md#attributes
@dreddy-work @JeremyKuhne  - thoughts?

src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs
```diff
@@ -389,6 +395,8 @@ public bool UseCustomTabOffsets
             }
         }
 
+        private static int DefaultListBoxItemHeight => _defaultListBoxItemHeight;
```
dmitrii-drobotov, line 398: I think the property can be removed now, and every usage updated to just `_defaultListBoxItemHeight`, I don't see a reason to have both.

docs/list-of-diagnostics.md
```diff
@@ -38,6 +38,8 @@ The acceptance criteria for adding an obsoletion includes:
 |  __`WFDEV001`__ | Casting to/from IntPtr is unsafe, use `ResultInternal`. |
 |  __`WFDEV002`__ | `DomainUpDown.DomainUpDownAccessibleObject` is no longer used to provide accessible support for `DomainUpDown` controls. Use `ControlAccessibleObject` instead. |
 |  __`WFDEV003`__ | `DomainUpDown.DomainItemAccessibleObject` is no longer used to provide accessible support for `DomainUpDown` items. |
+|  __`WFDEV004`__ | `ListBox.DefaultItemHeight` constant is no longer used as the default item height. 
+Default item height is now scaled according to the application default font. |
```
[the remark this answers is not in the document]
dkazennov: Yes, thanks a lot! Missed this issue.

src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs
```diff
@@ -62,8 +67,9 @@ public partial class ListBox : ListControl
         private SelectedIndexCollection? _selectedIndices;
         private ObjectCollection? _itemsCollection;
 
-        private int _itemHeight = DefaultItemHeight;
+        private int _itemHeight = DefaultListBoxItemHeight;
         private int _columnWidth;
+        private static int _defaultListBoxItemHeight = DefaultFont.Height;
```
Tanya-Solyanik, line 72: Are we guaranteed that `DefaultFont` is already initialized at this point?
It would be safer to do
```
private static int _defaultListBoxItemHeight = -1;

private static int DefaultListBoxItemHeight
{
    get
    {
        if (_defaultListBoxItemHeight == -1)
        {
            _defaultListBoxItemHeight = DefaultFont.Height;
        }
        return _defaultListBoxItemHeight;    
    }
}

Comment (dkazennov) [reply src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs:72]: Yes, that's a good point!

File: src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs (hunk 10 above)
Comment (dkazennov) [reply src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs]: So I've decided to keep the property (see below).

File: src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs [hunk 13]
```diff
@@ -62,8 +67,9 @@ public partial class ListBox : ListControl
         private SelectedIndexCollection? _selectedIndices;
         private ObjectCollection? _itemsCollection;
 
-        private int _itemHeight = DefaultItemHeight;
+        private int _itemHeight = DefaultListBoxItemHeight;
         private int _columnWidth;
+        private static int _defaultListBoxItemHeight = -1;
```
Tanya-Solyanik, line 72: Please add a `s_` prefix per .editorconfig rules to this static field name.

dkazennov:
Implemented final nit and rebased this fix on the most recent main as well.

dmitrii-drobotov (approved):

Tanya-Solyanik (approved):

src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs
dreddy-work: I am not completely sure how attribute default value is accessed without accessing `Get` property. But given the backing field is replaced with new default, looks safe to me.

dreddy-work (approved):

dreddy-work:
@dkazennov / @Tanya-Solyanik , can you please create a doc bug here: https://github.com/dotnet/docs/issues, before merging this?

Tanya-Solyanik:
@dreddy-work  - this should not be merged, it's waiting on the API review

src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs
Tanya-Solyanik: Here is the get property - https://github.com/dotnet/runtime/blob/57bfe474518ab5b7cfe6bf7424a79ce3af9d6657/src/libraries/System.Private.CoreLib/src/System/ComponentModel/DefaultValueAttribute.cs#L232

Tanya-Solyanik: The change will be visible in application that are enumerating attributes on properties of this class. The purpose of this attribute is to be consumed by the TypeDescriptor, but since this is a public surface, anybody can read it.

Tanya-Solyanik:
Breaking change doc for the obsoletion of the DefaultItemHeight contant https://github.com/dotnet/docs/issues/33953

Tanya-Solyanik:
breaking change doc that covers all attributes removals that follow this pattern - https://github.com/dotnet/docs/issues/33954

src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs
```diff
@@ -52,6 +52,11 @@ public partial class ListBox : ListControl
         ///  The default item height for an owner-draw ListBox. The ListBox's non-ownerdraw
         ///  item height is 13 for the default font on Windows.
         /// </summary>
+        [Obsolete(
```
Tanya-Solyanik, line 55: @dkazennov  - could you please move this obsoletion into a dedicated PR, We need to provide a replacement API. THe rest seems to be ready to go.

dkazennov:
Obsoletion of the ListBox.DefaultItemHeight was moved to #8640.

src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs
dkazennov: > @dkazennov - could you please move this obsoletion into a dedicated PR, We need to provide a replacement API. THe rest seems to be ready to go.

The new dedicated PR is #8640

Tanya-Solyanik (approved):

PR #8519, opened 2023-01-24
Enable Thread's DPI hosting behavior for mixed mode DPI applications.

### LINKED ISSUE #8144: Form designer failed to load after setting ApplicationHighDpiMode as PerMonitorV2
### .NET version

8.0.100-alpha.1.22556.1

Yes

No, it also repro in .Net 7.0 and .Net 6.0 SDK builds

### Issue description

On 100%DPI or High DPI environment, after setting ApplicationHighDpiMode to PerMonitorV2, form designer failed to load:

Here is the logs in Output window: 
_[00:53:41.632681] warn: DPI mode PerMonitorV2 specified in the project is not compatible with Visual Studio process DPI mode SystemAware. Winforms designer will have DPI mode SystemAware 
[00:53:47.223103] fail: Request failures: Sessions/EndLoad.
                        Microsoft.DotNet.DesignTools.Client.DesignToolsServerException: Failed to set Win32 parent window of the Control._

image
designerloadissue

**Call stack:**
************** Exception Text **************
System.ComponentModel.Win32Exception (5023): Failed to set Win32 parent window of the Control.
   at System.Windows.Forms.Control.SetParentHandle(HWND value) in C:\Users\xxx\Documents\Winforms\src\System.Windows.Forms\src\System\Windows\Forms\Control.cs:line 11054
   at System.Windows.Forms.Control.CreateControl(Boolean fIgnoreVisible) in C:\Users\xxx\Documents\Winforms\src\System.Windows.Forms\src\System\Windows\Forms\Control.cs:line 5082
   at System.Windows.Forms.Control.CreateControl(Boolean fIgnoreVisible) in C:\Users\xxx\Documents\Winforms\src\System.Windows.Forms\src\System\Windows\Forms\Control.cs:line 5085
   at System.Windows.Forms.Control.CreateControl() in C:\Users\xxx\Documents\Winforms\src\System.Windows.Forms\src\System\Windows\Forms\Control.cs:line 5028
   at System.Windows.Forms.Control.OnVisibleChanged(EventArgs e) in C:\Users\xxx\Documents\Winforms\src\System.Windows.Forms\src\System\Windows\Forms\Control.cs:line 7810
   at System.Windows.Forms.Control.OnParentVisibleChanged(EventArgs e) in C:\Users\xxx\Documents\Winforms\src\System.Windows.Forms\src\System\Windows\Forms\Control.cs:line 7707
   at System.Windows.Forms.Control.OnVisibleChanged(EventArgs e) in C:\Users\xxx\Documents\Winforms\src\System.Windows.Forms\src\System\Windows\Forms\Control.cs:line 7830
   at System.Windows.Forms.Control.WmShowWindow(Message& m) in C:\Users\xxx\Documents\Winforms\src\System.Windows.Forms\src\System\Windows\Forms\Control.cs:line 12974
   at System.Windows.Forms.Control.WndProc(Message& m) in C:\Users\xxx\Documents\Winforms\src\System.Windows.Forms\src\System\Windows\Forms\Control.cs:line 13382
   at Microsoft.DotNet.DesignTools.Frame.DesignerWindowFrame.WndProc(Message& m) in C:\Users\xxx\source\repos\winforms-designer\src\Microsoft.DotNet.DesignTools.Server\Frame\DesignerWindowFrame.cs:line 123
   at System.Windows.Forms.Control.ControlNativeWindow.WndProc(Message& m) in C:\Users\xxx\Documents\Winforms\src\System.Windows.Forms\src\System\Windows\Forms\Control.ControlNativeWindow.cs:line 114
   at System.Windows.Forms.NativeWindow.Callback(HWND hWnd, WM msg, WPARAM wparam, LPARAM lparam) in C:\Users\xxx\Documents\Winforms\src\System.Windows.Forms\src\System\Windows\Forms\NativeWindow.cs:line 375

### Steps to reproduce

1. Open this application: [WinFormsApp1.zip](https://github.com/dotnet/winforms/files/9969279/WinFormsApp1.zip)
2. Go to csproj file, set ApplicationHighDpiMode as PerMonitorV2
3. Clean and build application
4. Open form designer

dreddy-work:
fixes #8144 

WinForms now supports mixed mode DPI scenarios for applications hosted in Visual Studio. This requires the explicit alteration of default [hosting behavior](https://learn.microsoft.com/en-us/windows/win32/api/windef/ne-windef-dpi_hosting_behavior) on the threads creating the main UI, which Visual Studio has done to allow for mixed mode DPI child elements while in PermonitorV2 DPI mode.

 However, when investigating issue #8144, it was found that WinForms control parenting was failing in the designer within Visual Studio due to threads creating WinForms controls not enabling mixed mode hosting, resulting in a failure to build the parenting hierarchy when the DpiAwarenessContext differs, and not yet parented to Visual Studio. 

To support mixed mode DPI outside of Visual Studio hosting or WinForms standalone applications, mixed mode DPI hosting needs to be enabled on the threads that creating windows in WinForms.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8519)


src/System.Windows.Forms/src/System/Windows/Forms/NativeWindow.cs
```diff
@@ -497,6 +497,16 @@ public virtual unsafe void CreateHandle(CreateParams cp)
                         throw new Win32Exception(lastWin32Error, SR.ErrorCreatingHandle);
                     }
 
+                    if (OsVersion.IsWindows10_18030rGreater())
+                    {
+                        // In a mixed DPI hosting environment, the DPI settings for child windows can be determined by either the parent window or the thread hosting it,
+                        // based on the window properties and the behavior of the thread. For additional information, please refer to
+                        // https://microsoft.visualstudio.com/OS/_search?action=contents&text=ref%3AIsValidKernelDpiAwarenessContext&type=code&filters=ProjectFilters%7BOS%7DRepositoryFilters%7Bos.2020%7D&pageSize=25&includeFacets=false&result=DefaultCollection/OS/os.2020/GBofficial/main//clientcore/windows/Core/ntuser/kernel/windows/createw.cxx
```
RussKie, line 504: I don't think this is the correct link - it points to a callsite, whereas you probably want to reference the source (see w32dpi.h). Also, not sure whether the internal link is acceptable here.

dreddy-work: Thanks @RussKie for pointing the link. I am correcting it. Yes this is internal and is very useful for reference for any future changes in this space. If i had that handy, it could have saved me couple of hours with my experiments. Unless there is a hard NO on internal links in open source, I see value in keeping it.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessContext.cs
```diff
@@ -42,56 +42,69 @@ public static T CreateInstanceInSystemAwareContext<T>(Func<T> createInstance)
         private class DpiAwarenessScope : IDisposable
         {
             private bool dpiAwarenessScopeIsSet;
+            private DPI_HOSTING_BEHAVIOR currentThreadHostingBehavior = DPI_HOSTING_BEHAVIOR.DPI_HOSTING_BEHAVIOR_INVALID;
```
JeremyKuhne, line 45: Can we please fix these to be to standard? Notably they should have `_`.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessContext.cs
```diff
@@ -42,56 +42,69 @@ public static T CreateInstanceInSystemAwareContext<T>(Func<T> createInstance)
...
             private readonly DPI_AWARENESS_CONTEXT originalAwareness = DPI_AWARENESS_CONTEXT.UNSPECIFIED_DPI_AWARENESS_CONTEXT;
 
             /// <summary>
-            ///  Enters given Dpi awareness scope
+            ///  Sets DpiAwarenessContext for the thread.
             /// </summary>
-            public DpiAwarenessScope(DPI_AWARENESS_CONTEXT awareness)
+            /// <param name="awarenessContext">DpiawarenessContext to be set on the thread.</param>
```
Tanya-Solyanik, line 51:
```suggestion
            /// <param name="awarenessContext"><see cref="DPI_AWARENESS_CONTEXT" /> to be set on the thread.</param>
```

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessContext.cs
```diff
@@ -42,56 +42,69 @@ public static T CreateInstanceInSystemAwareContext<T>(Func<T> createInstance)
...
+            /// <param name="dpiHostingBehavior">DPI hosting behavior to be set on the thread.</param>
```
Tanya-Solyanik, line 52:
```suggestion
            /// <param name="dpiHostingBehavior"><see cref="DPI_HOSTING_BEHAVIOR" /> to be set on the thread.</param>
```

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessContext.cs
```diff
@@ -42,56 +42,69 @@ public static T CreateInstanceInSystemAwareContext<T>(Func<T> createInstance)
         private class DpiAwarenessScope : IDisposable
         {
             private bool dpiAwarenessScopeIsSet;
+            private DPI_HOSTING_BEHAVIOR currentThreadHostingBehavior = DPI_HOSTING_BEHAVIOR.DPI_HOSTING_BEHAVIOR_INVALID;
             private readonly DPI_AWARENESS_CONTEXT originalAwareness = DPI_AWARENESS_CONTEXT.UNSPECIFIED_DPI_AWARENESS_CONTEXT;
 
             /// <summary>
-            ///  Enters given Dpi awareness scope
+            ///  Sets DpiAwarenessContext for the thread.
```
Tanya-Solyanik, line 49:
```suggestion
            ///  Sets <see cref="DPI_AWARENESS_CONTEXT" /> for the thread.
```

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessContext.cs
```diff
@@ -42,56 +42,69 @@ public static T CreateInstanceInSystemAwareContext<T>(Func<T> createInstance)
...
             /// </summary>
-            public DpiAwarenessScope(DPI_AWARENESS_CONTEXT awareness)
+            /// <param name="awarenessContext">DpiawarenessContext to be set on the thread.</param>
+            /// <param name="dpiHostingBehavior">DPI hosting behavior to be set on the thread.</param>
+            public DpiAwarenessScope(DPI_AWARENESS_CONTEXT awarenessContext, DPI_HOSTING_BEHAVIOR dpiHostingBehavior)
             {
-                try
+                // Full support for DpiawarenessContext and mixed mode DPI hosting behavior on the thread is only available after the RS4 OS release.
```
Tanya-Solyanik, line 55:
```suggestion
                // Full support for DPI_AWARENESS_CONTEXT and mixed mode DPI hosting behavior on the thread is only available after the RS4 OS release.
```

src/System.Windows.Forms/src/System/Windows/Forms/NativeWindow.cs
Tanya-Solyanik: @terrajobst - Is it allowed to have links to the OS repo in source code?

Tanya-Solyanik: Either way, it would be more helpful to summarize how child windows get DPI awareness rather than to point user elsewhere.

Tanya-Solyanik:
>WinForms control parenting was failing in the designer within Visual Studio due to threads creating WinForms controls not enabling mixed mode hosting, 

@dreddy-work - could you please clarify what windows were failing and what DPI awareness context for each was. What was VS context? I assume server context was PMv2?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/MixedDpiHostingTests.cs
```diff
@@ -0,0 +1,52 @@
...
+        [WinFormsFact]
+        public void MixedHosting_Default()
+        {
+            // Run tests only on Windows 10 versions that support thread dpi awareness.
+            if (!PlatformDetection.IsWindows10Version1803OrGreater)
+            {
+                return;
+            }
+
+            // set thread awareness context to PermonitorV2(PMv2).
```
Tanya-Solyanik, line 20: Please capitalize sentences

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/MixedDpiHostingTests.cs
```diff
@@ -0,0 +1,52 @@
...
+            // if process/thread is not in PMv2, calling 'EnterDpiAwarenessScope' is a no-op and that is by design.
+            // In this case, we will be setting thread to PMv2 mode and then scope to UNAWARE
+            DPI_AWARENESS_CONTEXT originalAwarenessContext = PInvoke.SetThreadDpiAwarenessContextInternal(DPI_AWARENESS_CONTEXT.DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2);
+
+            try
+            {
+                using Form form = new Form();
+                using (DpiHelper.EnterDpiAwarenessScope(DPI_AWARENESS_CONTEXT.DPI_AWARENESS_CONTEXT_SYSTEM_AWARE))
```
Tanya-Solyanik, line 28: What awareness context is the test process running in?

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessContext.cs
```diff
@@ -42,56 +42,69 @@ public static T CreateInstanceInSystemAwareContext<T>(Func<T> createInstance)
         private class DpiAwarenessScope : IDisposable
```
Tanya-Solyanik, line 42: Please move to a separate file in another PR.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessContext.cs
```diff
@@ -42,56 +42,69 @@ public static T CreateInstanceInSystemAwareContext<T>(Func<T> createInstance)
...
         {
             private bool dpiAwarenessScopeIsSet;
+            private DPI_HOSTING_BEHAVIOR currentThreadHostingBehavior = DPI_HOSTING_BEHAVIOR.DPI_HOSTING_BEHAVIOR_INVALID;
             private readonly DPI_AWARENESS_CONTEXT originalAwareness = DPI_AWARENESS_CONTEXT.UNSPECIFIED_DPI_AWARENESS_CONTEXT;
```
Tanya-Solyanik, line 46: Consider: `_originalThreadContext`

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessContext.cs
```diff
@@ -42,56 +42,69 @@ public static T CreateInstanceInSystemAwareContext<T>(Func<T> createInstance)
...
 
             /// <summary>
-            ///  Enters given Dpi awareness scope
+            ///  Sets DpiAwarenessContext for the thread.
             /// </summary>
-            public DpiAwarenessScope(DPI_AWARENESS_CONTEXT awareness)
+            /// <param name="awarenessContext">DpiawarenessContext to be set on the thread.</param>
+            /// <param name="dpiHostingBehavior">DPI hosting behavior to be set on the thread.</param>
+            public DpiAwarenessScope(DPI_AWARENESS_CONTEXT awarenessContext, DPI_HOSTING_BEHAVIOR dpiHostingBehavior)
```
Tanya-Solyanik, line 53: Consider: `awarenessContext` -> `context`

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessContext.cs
```diff
@@ -42,56 +42,69 @@ public static T CreateInstanceInSystemAwareContext<T>(Func<T> createInstance)
...
-                        originalAwareness = PInvoke.GetThreadDpiAwarenessContextInternal();
+                    return;
+                }
 
-                        // If current process dpiawareness is SYSTEM_UNAWARE or SYSTEM_AWARE (must be equal to awareness), calling this method will be a no-op.
-                        if (!PInvoke.AreDpiAwarenessContextsEqualInternal(originalAwareness, awareness) &&
-                            !PInvoke.AreDpiAwarenessContextsEqualInternal(originalAwareness, DPI_AWARENESS_CONTEXT.DPI_AWARENESS_CONTEXT_UNAWARE))
-                        {
-                            originalAwareness = PInvoke.SetThreadDpiAwarenessContextInternal(awareness);
+                // Unsupported DpiawarenessContext result in no-op.
```
Tanya-Solyanik, line 61: Please capitalize this property name correctly in the comments - `DpiAwarenessContext`

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessContext.cs
```diff
@@ -42,56 +42,69 @@ public static T CreateInstanceInSystemAwareContext<T>(Func<T> createInstance)
...
+                    originalAwareness = PInvoke.SetThreadDpiAwarenessContext(awarenessContext);
+
+                    // As reported in https://github.com/dotnet/winforms/issues/2969
+                    // under unknown conditions originalAwareness may remain 0 (which means the call failed)
+                    // causing us to set dpiAwarenessScopeIsSet=true, which would lead to a crash when we attempt to dispose the scope.
+                    dpiAwarenessScopeIsSet = originalAwareness != IntPtr.Zero;
                 }
-                catch (EntryPointNotFoundException)
+
+                // Ignore thread's hosting behaviour if thread's DpiawarenessContext can not be set.
```
Tanya-Solyanik, line 79: This comment is a bit misleading.  This is also the case when the child has the same context as the parent and no changes are needed.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessContext.cs
```diff
@@ -42,56 +42,69 @@ public static T CreateInstanceInSystemAwareContext<T>(Func<T> createInstance)
...
 
-            /// <summary>
-            ///  resetting dpiawareness of the thread.
-            /// </summary>
-            private void ResetDpiAwarenessContextChanges()
-            {
-                if (dpiAwarenessScopeIsSet)
+#pragma warning disable CA1416 // OS version check is already covered when `dpiAwarenessScopeIsSet` flag is set to `true`.
+                PInvoke.SetThreadDpiAwarenessContext(originalAwareness);
+                dpiAwarenessScopeIsSet = false;
```
Tanya-Solyanik, line 100: Should we test if the `set` operation succedded?

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessContext.cs
Tanya-Solyanik, line 45: Isn't this the `original` behavior? To match the naming for context field.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessContext.cs
```diff
@@ -42,56 +42,69 @@ public static T CreateInstanceInSystemAwareContext<T>(Func<T> createInstance)
...
+
+                if (currentThreadHostingBehavior != DPI_HOSTING_BEHAVIOR.DPI_HOSTING_BEHAVIOR_INVALID)
                 {
-                    PInvoke.SetThreadDpiAwarenessContextInternal(originalAwareness);
-                    dpiAwarenessScopeIsSet = false;
+                    PInvoke.SetThreadDpiHostingBehavior(currentThreadHostingBehavior);
+                    currentThreadHostingBehavior = DPI_HOSTING_BEHAVIOR.DPI_HOSTING_BEHAVIOR_INVALID;
```
Tanya-Solyanik, line 105: Line 105 is not really needed, right? And line 100 either?

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/OsVersion.cs
```diff
@@ -30,6 +30,13 @@ internal static class OsVersion
         [SupportedOSPlatformGuard("windows10.0.15063")]
         public static bool IsWindows10_1703OrGreater() => OperatingSystem.IsWindowsVersionAtLeast(major: 10, build: 15063);
 
+        /// <summary>
+        ///  Is Windows 10 Creators Update or later. (Redstone 4, build 17134, version 1803)
```
Tanya-Solyanik, line 34: I think it's spelt `RedStone`.

src/System.Windows.Forms/src/System/Windows/Forms/NativeWindow.cs
```diff
@@ -445,7 +445,7 @@ public virtual unsafe void CreateHandle(CreateParams cp)
 
                         // Parking window dpi awareness context need to match with dpi awareness context of control being
                         // parented to this parking window. Otherwise, reparenting of control will fail.
-                        using (DpiHelper.EnterDpiAwarenessScope(DpiAwarenessContext))
+                        using (DpiHelper.EnterDpiAwarenessScope(DpiAwarenessContext, DPI_HOSTING_BEHAVIOR.DPI_HOSTING_BEHAVIOR_MIXED))
```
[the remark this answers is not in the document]
dreddy-work: These are all the scenarios covered in the EnterDpiAwarenessScope. It will be no-op in those cases.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/MixedDpiHostingTests.cs
dreddy-work: Nothing is set. Default to UnAware. However, if the process awareness is not set yet, you could change Threads awareness.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/OsVersion.cs
dreddy-work: I am not certain but leaving it as-is for consistency with the rest of the file.

src/System.Windows.Forms/src/System/Windows/Forms/NativeWindow.cs
dreddy-work: internal links are accepted (confirmed offline). I gave high level description that describes primary case. If needed link provided will help navigate.

src/System.Windows.Forms.Primitives/src/Resources/SR.resx
```diff
@@ -128,5 +128,14 @@
   </data>
   <data name="General_MissingService" xml:space="preserve">
     <value>The service '{0}' must be installed for this feature to work.  Ensure that this service is available.</value>
+  </data>  
+  <data name="Win32SetThreadsDpiContextFailed" xml:space="preserve">
+    <value>Failed to set threads DpiAwareness context {0}.</value>
```
Tanya-Solyanik, line 133:
```suggestion
    <value>Failed to set thread's DpiAwareness context {0}.</value>
```

src/System.Windows.Forms.Primitives/src/Resources/SR.resx
```diff
@@ -128,5 +128,14 @@
...
+  </data>
+  <data name="Win32GetThreadsDpiContextFailed" xml:space="preserve">
+    <value>Failed to get threads DpiAwareness context.</value>
```
Tanya-Solyanik, line 136:
```suggestion
    <value>Failed to get thread's DpiAwareness context.</value>
```

src/System.Windows.Forms.Primitives/src/Resources/SR.resx
```diff
@@ -128,5 +128,14 @@
...
+  </data>
+  <data name="Win32SetThreadsDpiHostingBehaviorFailed" xml:space="preserve">
+    <value>Failed to set threads DPI hosting behavior {0}.</value>
```
Tanya-Solyanik, line 139:
```suggestion
    <value>Failed to set thread's DPI hosting behavior {0}.</value>
```

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessScope.cs
```diff
@@ -0,0 +1,106 @@
...
+namespace System.Windows.Forms
+{
+    internal static partial class DpiHelper
+    {
+        /// <summary>
+        ///  Class that help setting <see cref="DPI_AWARENESS_CONTEXT"/> scope
+        /// </summary>
+        private class DpiAwarenessScope : IDisposable
+        {
+            private bool dpiAwarenessScopeIsSet;
```
Tanya-Solyanik, line 18: Private field name should start with an underscore.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessScope.cs
```diff
@@ -0,0 +1,106 @@
...
+using System.ComponentModel;
+using System.Runtime.InteropServices;
+using System.Windows.Forms.Primitives.Resources;
+
+namespace System.Windows.Forms
+{
+    internal static partial class DpiHelper
+    {
+        /// <summary>
+        ///  Class that help setting <see cref="DPI_AWARENESS_CONTEXT"/> scope
```
Tanya-Solyanik, line 14:
```suggestion
        ///  Class that help setting <see cref="DPI_AWARENESS_CONTEXT"/> scope.
```

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessScope.cs
```diff
@@ -0,0 +1,106 @@
...
+                // Full support for DPI_AWARENESS_CONTEXT and mixed mode DPI_HOSTING_BEHAVIOR on the thread is only available after the RS4 OS release.
+                if (!OsVersion.IsWindows10_18030rGreater())
+                {
+                    throw new PlatformNotSupportedException();
+                }
+
+                // Unsupported DPI_AWARENESS_CONTEXT result in no-op.
+                if (PInvoke.AreDpiAwarenessContextsEqualInternal(context, DPI_AWARENESS_CONTEXT.UNSPECIFIED_DPI_AWARENESS_CONTEXT))
+                {
+                    // DpiContext specified for the scope is not supported.
```
Tanya-Solyanik, line 38: this comment is not adding any new information to the exception we are throwing.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessScope.cs
```diff
@@ -0,0 +1,106 @@
...
+            /// <param name="behavior"><see cref="DPI_HOSTING_BEHAVIOR"/> to be set on the thread.</param>
+            public DpiAwarenessScope(DPI_AWARENESS_CONTEXT context, DPI_HOSTING_BEHAVIOR behavior)
+            {
+                // Full support for DPI_AWARENESS_CONTEXT and mixed mode DPI_HOSTING_BEHAVIOR on the thread is only available after the RS4 OS release.
+                if (!OsVersion.IsWindows10_18030rGreater())
+                {
+                    throw new PlatformNotSupportedException();
+                }
+
+                // Unsupported DPI_AWARENESS_CONTEXT result in no-op.
```
Tanya-Solyanik, line 35: Not a "no-op", in an exception...

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessScope.cs
```diff
@@ -0,0 +1,106 @@
...
+                if (PInvoke.AreDpiAwarenessContextsEqualInternal(context, DPI_AWARENESS_CONTEXT.UNSPECIFIED_DPI_AWARENESS_CONTEXT))
+                {
+                    // DpiContext specified for the scope is not supported.
+                    throw new NotSupportedException();
+                }
+
+                _originalDpiAwarenessContext = PInvoke.GetThreadDpiAwarenessContext();
+                if (_originalDpiAwarenessContext == 0)
```
Tanya-Solyanik, line 43: `0` is actually `UNSPECIFIED_DPI_AWARENESS_CONTEXT`,

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessScope.cs
```diff
@@ -0,0 +1,106 @@
...
+                {
+                    throw new Win32Exception(Marshal.GetLastWin32Error(), SR.Win32GetThreadsDpiContextFailed);
+                }
+
+                // No-op when requested DpiContext and current threads context is same.
```
Tanya-Solyanik, line 48:
```suggestion
                // No-op when requested DPI_AWARENESS_CONTEXT and current thread's context is the same.
```

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessScope.cs
```diff
@@ -0,0 +1,106 @@
...
+                if (!PInvoke.AreDpiAwarenessContextsEqual(_originalDpiAwarenessContext, context))
+                {
+                    if (PInvoke.AreDpiAwarenessContextsEqual(_originalDpiAwarenessContext, DPI_AWARENESS_CONTEXT.DPI_AWARENESS_CONTEXT_UNAWARE))
+                    {
+                        // It is not appropriate to establish the DPI scope for a process or thread that is unaware.
+                        // Any windows that are created within this scope cannot be parented to windows created in unaware processes or threads.
+                        throw new NotSupportedException();
```
Tanya-Solyanik, line 55: I would rather let Win32 report that error when the window is created. Maybe the scope is entered for reasons other than window creation? TO recalculate coordinates for example?

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessScope.cs
```diff
@@ -0,0 +1,106 @@
...
+                    }
+
+                    if (PInvoke.SetThreadDpiAwarenessContext(context) == 0)
```
Tanya-Solyanik, line 58: You have to compare against the DPI_AWARENESS_CONTEXT struct fields because it's not an int, it's a handle

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessScope.cs
```diff
@@ -0,0 +1,106 @@
...
+            private bool dpiAwarenessScopeIsSet;
+            private DPI_HOSTING_BEHAVIOR _originalDpiHostingBehavior = DPI_HOSTING_BEHAVIOR.DPI_HOSTING_BEHAVIOR_INVALID;
+            private readonly DPI_AWARENESS_CONTEXT _originalDpiAwarenessContext;
+
+            /// <summary>
+            ///  Sets <see cref="DPI_AWARENESS_CONTEXT"/> for the thread.
+            /// </summary>
+            /// <param name="context"><see cref="DPI_AWARENESS_CONTEXT"/> to be set on the thread.</param>
+            /// <param name="behavior"><see cref="DPI_HOSTING_BEHAVIOR"/> to be set on the thread.</param>
+            public DpiAwarenessScope(DPI_AWARENESS_CONTEXT context, DPI_HOSTING_BEHAVIOR behavior)
```
Tanya-Solyanik, line 27: Should the second argument be a bool instead? You are not allowing any value other than mixed. So it could be `bool setMixed`.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessScope.cs
```diff
@@ -0,0 +1,106 @@
...
+
+                    dpiAwarenessScopeIsSet = true;
+                }
+
+                if (behavior != DPI_HOSTING_BEHAVIOR.DPI_HOSTING_BEHAVIOR_MIXED)
+                {
+                    return;
+                }
+
+                _originalDpiHostingBehavior = PInvoke.SetThreadDpiHostingBehavior(behavior);
```
Tanya-Solyanik, line 71: Should you run GetThreadDpiHostingBehavior first to see if you need to set/unset anything?

Tanya-Solyanik: Then you would need a bool `_isHostingBehaviorSet` to properly restore hosting behavior in dispose

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessScope.cs
```diff
@@ -0,0 +1,106 @@
...
+                if (_originalDpiHostingBehavior == DPI_HOSTING_BEHAVIOR.DPI_HOSTING_BEHAVIOR_INVALID)
+                {
+                    throw new Win32Exception(Marshal.GetLastWin32Error(), string.Format(SR.Win32SetThreadsDpiHostingBehaviorFailed, behavior));
+                }
+            }
+
+            /// <summary>
+            ///  Reset thread's <see cref="DPI_AWARENESS_CONTEXT"/> and <see cref="DPI_HOSTING_BEHAVIOR"/>.
+            /// </summary>
+            public void Dispose()
```
Tanya-Solyanik, line 81: I think dispose is called when "open scope" throws an exception, but I had not verified. Seems to be safer to test the OS version again.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessScope.cs
```diff
@@ -0,0 +1,106 @@
...
+            ///  Sets <see cref="DPI_AWARENESS_CONTEXT"/> for the thread.
+            /// </summary>
+            /// <param name="context"><see cref="DPI_AWARENESS_CONTEXT"/> to be set on the thread.</param>
+            /// <param name="behavior"><see cref="DPI_HOSTING_BEHAVIOR"/> to be set on the thread.</param>
+            public DpiAwarenessScope(DPI_AWARENESS_CONTEXT context, DPI_HOSTING_BEHAVIOR behavior)
+            {
+                // Full support for DPI_AWARENESS_CONTEXT and mixed mode DPI_HOSTING_BEHAVIOR on the thread is only available after the RS4 OS release.
+                if (!OsVersion.IsWindows10_18030rGreater())
+                {
+                    throw new PlatformNotSupportedException();
```
Tanya-Solyanik, line 32: I think it's safer to fail silently because the UI might or might not be usable, but that's better than a crash. This situation is unlikely to be discovered at development time, so failing fast might not be helpful to developer. But add a `Debug.Assert`

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessScope.cs
dreddy-work: I am seeing this in tests. Removing this condition.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessScope.cs
dreddy-work: Mixed is default for compatibility. I am inclining to make this generic and allow other hosting behaviors if users/tests like to exercise it.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessScope.cs
dreddy-work: Given the APIs are conditioned now, we don't need OS version check anymore.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessScope.cs
dreddy-work: Agree. Given I am not certain if DpiScope that we used since. NET Framework 4.7 may not be guarded against OS version check.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.SetThreadDpiHostingBehavior.cs
```diff
@@ -0,0 +1,34 @@
...
+    internal static partial class PInvoke
+    {
+        /// <summary>
+        ///  Sets the thread DPI hosting behavior.
+        /// </summary>
+        /// <returns>
+        ///  The old thread DPI hosting behavior if the API is available in this version of OS.
+        ///  Otherwise, <see cref="DPI_HOSTING_BEHAVIOR.DPI_HOSTING_BEHAVIOR_INVALID"/>.
+        /// </returns>
+        public static DPI_HOSTING_BEHAVIOR SetThreadDpiHostingBehaviorInternal(DPI_HOSTING_BEHAVIOR dpiHostingBehavior)
```
Tanya-Solyanik, line 18: I was suggesting a helper with no parameters:
public static DPI_HOSTING_BEHAVIOR SetMixedDpiHostingBehaviorInternal()

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.SetThreadDpiHostingBehavior.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            if (dpiHostingBehavior != DPI_HOSTING_BEHAVIOR.DPI_HOSTING_BEHAVIOR_MIXED)
+            {
+                // Winforms only affects the behavior of thread hosting in mixed DPI mode situations, and does not apply to any other scenarios.
+                throw new ArgumentException(nameof(dpiHostingBehavior), dpiHostingBehavior.ToString());
+            }
+
+            if (!OsVersion.IsWindows10_18030rGreater())
+            {
+                throw new PlatformNotSupportedException();
```
Tanya-Solyanik, line 28: Are we throwing in other places? I thought we wanted to fail silently.

Tanya-Solyanik: We were catching `EntryPointNotFoundException` before.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.SetThreadDpiHostingBehavior.cs
dreddy-work: Right. But this is generic enough and can be used in tests so i inclined to go as is here.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.SetThreadDpiHostingBehavior.cs
dreddy-work: There is no scenario at this point that tries to set behavior that is not mixed and calling this explicitly on legacy OS and running PermonV2.  Calling into Win32 knowingly it is not available isn't helpful.

Tanya-Solyanik: We can just skip the call when it's no available, the app is SA or UA anyway.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.SetThreadDpiHostingBehavior.cs
Tanya-Solyanik: It's a bad API that takes an enum type but really accepts only a single value. Maybe you can define different helper methods for tests?

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.SetThreadDpiHostingBehavior.cs
dreddy-work: Sorry i wasn't right. Looking changes in Github browsers didn't give clear picture.  This method isn't used currently and is not needed any more after refactored code addressing your original review.  We are returning silently when running on legacy OS and never set hosting behavior or call it.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.SetThreadDpiHostingBehavior.cs
dreddy-work: Sorry i was referring to method `DpiAwarenessScope` in my earlier comment. This method isn't used and removed in latest commit.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessScope.cs
```diff
@@ -0,0 +1,97 @@
...
+                if (!OsVersion.IsWindows10_18030rGreater())
+                {
+                    Debug.Assert(false, "Full support for DPI_AWARENESS_CONTEXT and mixed mode DPI_HOSTING_BEHAVIOR on the thread is only available after the RS4 OS release");
+                    return;
+                }
+
+                // Unsupported DPI_AWARENESS_CONTEXT result in NotSupportedException.
+                if (PInvoke.AreDpiAwarenessContextsEqualInternal(context, DPI_AWARENESS_CONTEXT.UNSPECIFIED_DPI_AWARENESS_CONTEXT))
+                {
+                    throw new NotSupportedException();
```
Tanya-Solyanik, line 41: Do you mean that the windows will return error HR when "Unspecified" context is passed in? Why not bubble up the original windows error code to the users?

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessScope.cs
```diff
@@ -0,0 +1,97 @@
...
+                }
+
+                _originalDpiAwarenessContext = PInvoke.GetThreadDpiAwarenessContext();
+
+                // No-op when requested DPI_AWARENESS_CONTEXT and current thread's context is the same.
+                if (!PInvoke.AreDpiAwarenessContextsEqual(_originalDpiAwarenessContext, context))
+                {
+                    if (PInvoke.SetThreadDpiAwarenessContext(context) == IntPtr.Zero)
+                    {
+                        throw new Win32Exception(Marshal.GetLastWin32Error(), SR.Win32GetThreadsDpiContextFailed);
```
Tanya-Solyanik, line 51: Is CS Win32 actually passing the last error to us?

Tanya-Solyanik (commented):
Looks good, I added a question, and disabled the autocomplete because your description is out of date now, I think.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessScope.cs
dreddy-work: `UNSPECIFIED_DPI_AWARENESS_CONTEXT` is not known to windows. We defined it in Winforms as default value. Windows cannot set it and return NULL if called with it.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.DpiAwarenessScope.cs
dreddy-work: CS Win32 generated code should be responsible to set it. I see `SetLastError` being set on some methods and not on others. For now, i am leaving as - is this failed scenario.

Tanya-Solyanik (approved):

PR #8528, opened 2023-01-27
Refactor `SelectionService` to replace `ArrayList` and enable Nullability

elachlan:
Refactored `ISelectionService` to replace `ArrayList` and enable Nullability.

I had to submit a PR to Runtime to fix the nullability on `ISelectionService.PrimarySelection`, this PR will need to be rebased to main once that change has been made available. Until then this will have build errors.
https://github.com/dotnet/runtime/pull/81254

Related: #8140

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8528)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/SelectionService.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/SelectionService.cs
@@ -227,7 +227,7 @@ namespace System.ComponentModel.Design
                 }
             }
 
-            _contextAttributes = new(_selection.Count);
+            _contextAttributes = new();
 
             for (int i = 0; i < _selection.Count; i++)
             {
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/SelectionService.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/SelectionService.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.Collections;
 using System.Collections.Specialized;
 using System.Diagnostics;
@@ -14,7 +12,10 @@ using System.Windows.Forms.Design;
 namespace System.ComponentModel.Design
 {
     /// <summary>
-    ///  The selection service handles selection within a form.  There is one selection service for each designer host. A selection consists of an array of objects.  One objects is designated the "primary" selection.
+    ///  The selection service handles selection within a form.
+    ///  There is one selection service for each designer host.
+    ///  A selection consists of an array of objects.
+    ///  One objects is designated the "primary" selection.
     /// </summary>
     internal sealed class SelectionService : ISelectionService, IDisposable
     {
@@ -33,18 +34,19 @@ namespace System.ComponentModel.Design
         private IServiceProvider _provider; // The service provider
         private BitVector32 _state; // state of the selection service
         private readonly EventHandlerList _events; // the events we raise
-        private ArrayList _selection; // list of selected objects
-        private string[] _contextAttributes; // help context information we have pushed to the help service.
+        private List<IComponent>? _selection; // list of selected objects
+        private string[]? _contextAttributes; // help context information we have pushed to the help service.
         private short _contextKeyword; // the offset into the selection keywords for the current selection.
         private StatusCommandUI _statusCommandUI; // UI for setting the StatusBar Information..
 
         /// <summary>
-        ///  Creates a new selection manager object.  The selection manager manages all selection of all designers under the current form file.
+        ///  Creates a new selection manager object.
+        ///  The selection manager manages all selection of all designers under the current form file.
         /// </summary>
         internal SelectionService(IServiceProvider provider) : base()
         {
             _provider = provider;
-            _state = default(BitVector32);
+            _state = default;
             _events = new EventHandlerList();
             _statusCommandUI = new StatusCommandUI(provider);
         }
@@ -52,11 +54,11 @@ namespace System.ComponentModel.Design
         /// <summary>
         ///  Adds the given selection to our selection list.
         /// </summary>
-        internal void AddSelection(object sel)
+        internal void AddSelection(IComponent sel)
         {
             if (_selection is null)
             {
-                _selection = new ArrayList();
+                _selection = new();
                 // Now is the opportune time to hook up all of our events
                 if (GetService(typeof(IComponentChangeService)) is IComponentChangeService cs)
                 {
@@ -81,7 +83,8 @@ namespace System.ComponentModel.Design
         }
 
         /// <summary>
-        ///  Called when our visibility or batch mode changes.  Flushes any pending notifications or updates if possible.
+        ///  Called when our visibility or batch mode changes.
+        ///  Flushes any pending notifications or updates if possible.
         /// </summary>
         private void FlushSelectionChanges()
         {
@@ -95,22 +98,15 @@ namespace System.ComponentModel.Design
         /// <summary>
         ///  Helper function to retrieve services.
         /// </summary>
-        private object GetService(Type serviceType)
-        {
-            if (_provider is not null)
-            {
-                return _provider.GetService(serviceType);
-            }
-
-            return null;
-        }
+        private object? GetService(Type serviceType) => _provider?.GetService(serviceType);
 
         /// <summary>
-        ///  called by the formcore when someone has removed a component.  This will remove any selection on the component without disturbing the rest of the selection
+        ///  called by the formcore when someone has removed a component.
+        ///  This will remove any selection on the component without disturbing the rest of the selection
         /// </summary>
-        private void OnComponentRemove(object sender, ComponentEventArgs ce)
+        private void OnComponentRemove(object? sender, ComponentEventArgs ce)
         {
-            if (_selection is not null && _selection.Contains(ce.Component))
+            if (_selection is not null && ce.Component is not null && _selection.Contains(ce.Component))
             {
                 RemoveSelection(ce.Component);
                 OnSelectionChanged();
@@ -118,7 +114,8 @@ namespace System.ComponentModel.Design
         }
 
         /// <summary>
-        ///  called anytime the selection has changed.  We update our UI for the selection, and then we fire a juicy change event.
+        ///  called anytime the selection has changed.
+        ///  We update our UI for the selection, and then we fire a juicy change event.
         /// </summary>
         private void OnSelectionChanged()
         {
@@ -128,7 +125,7 @@ namespace System.ComponentModel.Design
             }
             else
             {
-                EventHandler eh = _events[s_eventSelectionChanging] as EventHandler;
+                EventHandler? eh = _events[s_eventSelectionChanging] as EventHandler;
                 eh?.Invoke(this, EventArgs.Empty);
 
                 UpdateHelpKeyword(true);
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms.Design/src/System/ComponentModel/Design/SelectionService.cs
```diff
@@ -237,15 +234,18 @@ private void UpdateHelpKeyword(bool tryLater)
 
             for (int i = 0; i < _selection.Count; i++)
             {
-                object s = _selection[i];
-                string helpContext = TypeDescriptor.GetClassName(s);
-                HelpKeywordAttribute contextAttr = (HelpKeywordAttribute)TypeDescriptor.GetAttributes(s)[typeof(HelpKeywordAttribute)];
-                if (contextAttr is not null && !contextAttr.IsDefaultAttribute())
+                object component = _selection[i];
```
lonitra, line 237: In `AddSelection` , it doesn't seem like

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/SelectionService.cs
```diff
@@ -291,41 +292,21 @@ void IDisposable.Dispose()
...
-            {
-                if (_selection is not null && _selection.Count > 0)
-                {
-                    return _selection[0];
-                }
-
-                return null;
-            }
-        }
+        object? ISelectionService.PrimarySelection => _selection?.Count > 0 ? _selection[0] : null;
```
lonitra, line 304: Consider calling to our internal `PrimarySelection` method so we don't have the same code in different places.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/SelectionService.cs
```diff
@@ -192,13 +185,17 @@ private void ApplicationIdle(object source, EventArgs args)
...
                 if (tryLater)
                 {
                     // we don't have an help service YET, we need to wait for it...
                     // hook up to the application.idle event
-                    // yes this is UGLY but we don't have a choice, vs is always returning a UserContext, so even if we manage to instantiate the HelpService beforehand and class pushcontext on it (trying to stack up help context in the helpservice to be flushed when we get the documentactivation event we just don't know if that's going to work or not... so we just wait...) :(((
+                    // yes this is UGLY but we don't have a choice.
+                    // vs is always returning a UserContext, so even if we manage to instantiate the HelpService
+                    // beforehand and class pushcontext on it
+                    // (trying to stack up help context in the helpservice to be flushed when we get the
+                    // documentactivation event we just don't know if that's going to work or not... so we just wait...) :(((
```
lonitra, line 198: Let's go ahead and fix this comment. Perhaps something like this?
```c#
// We don't have a HelpService yet, hook up to the ApplicationIdle event.
// VS is always returning a UserContext, so instantiating the HelpService
// beforehand and doing class pushcontext on it to try to
// stack up help context in the HelpService to be flushed when we get the
// documentation event may not work, so we need to wait for a HelpService instead.
```

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/SelectionService.cs
elachlan: Good catch, I have added a null check to `AddSelection` to make it a no-op.

elachlan:
https://github.com/dotnet/runtime/pull/81254 is done, so waiting on winforms to update main with latest build and then I can rebase this.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/SelectionService.cs
```diff
@@ -161,27 +159,22 @@ private void OnTransactionClosed(object sender, DesignerTransactionCloseEventArg
...
         }
 
         /// <summary>
         ///  Removes the given selection from the selection list.
         /// </summary>
-        internal void RemoveSelection(object sel)
-        {
-            _selection?.Remove(sel);
-        }
+        internal void RemoveSelection(IComponent sel) => _selection?.Remove(sel);
```
lonitra, line 175:
```suggestion
        internal void RemoveSelection(IComponent selection) => _selection?.Remove(selection);
```

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/SelectionService.cs
```diff
@@ -74,14 +76,15 @@ internal void AddSelection(object sel)
                 }
             }
 
-            if (!_selection.Contains(sel))
+            if (!_selection.Contains(selection))
```
lonitra, line 79: Am I missing something? I do not see the null check on `selection` to make this method a no-op for null param.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/SelectionService.cs
elachlan: Done

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/SelectionService.cs
```diff
@@ -33,30 +34,36 @@ internal sealed class SelectionService : ISelectionService, IDisposable
         private IServiceProvider _provider; // The service provider
         private BitVector32 _state; // state of the selection service
         private readonly EventHandlerList _events; // the events we raise
-        private ArrayList _selection; // list of selected objects
-        private string[] _contextAttributes; // help context information we have pushed to the help service.
+        private List<IComponent>? _selection; // list of selected objects
+        private string[]? _contextAttributes; // help context information we have pushed to the help service.
```
dmitrii-drobotov, line 38: Should the type be `string?[]?`? It seems possible that some elements won't be initialized and stay null due to following condition: https://github.com/dotnet/winforms/blob/c7e036d2c2545869e7ff4e33d6ae89a9ad89c8a9/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/SelectionService.cs#L245-L248

elachlan: I refactored it to List<string>, but then thought that might be overkill. So I switched back to `string?[]?`.

lonitra (approved):
LGTM

dmitrii-drobotov (dismissed):

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/SelectionService.cs
```diff
@@ -33,30 +34,36 @@ internal sealed class SelectionService : ISelectionService, IDisposable
...
+            _state = default;
             _events = new EventHandlerList();
             _statusCommandUI = new StatusCommandUI(provider);
         }
 
         /// <summary>
         ///  Adds the given selection to our selection list.
         /// </summary>
-        internal void AddSelection(object sel)
+        internal void AddSelection(IComponent selection)
```
dreddy-work, line 57: why not

```suggestion
        internal void AddSelection(IComponent? selection)
```

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/SelectionService.cs
```diff
@@ -74,14 +81,15 @@ internal void AddSelection(object sel)
                 }
             }
 
-            if (!_selection.Contains(sel))
+            if (!_selection.Contains(selection))
```
dreddy-work, line 84: nit:  you could avoid checking this incase if you revert condition on line 64.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/SelectionService.cs
```diff
@@ -233,29 +233,36 @@ private void UpdateHelpKeyword(bool tryLater)
...
+                if (contextAttribute is not null && !contextAttribute.IsDefaultAttribute())
                 {
-                    helpContext = contextAttr.HelpKeyword;
+                    helpContext = contextAttribute.HelpKeyword;
                 }
 
-                _contextAttributes[i] = helpContext;
+                if (helpContext is not null)
+                {
+                    _contextAttributes[i] = helpContext;
```
dreddy-work, line 250: So, where do you insert null values into `_contextAttributes`?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/SelectionService.cs
elachlan: basically, we don't want to allow nulls. So its a no-op, returns straight away. Ideally we would add a `ArgumentNullException` but I think that is a breaking change?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/SelectionService.cs
elachlan: I am unsure how that would work. The section above sets up the list. But it has to happen inside the if statement because of the other code that is there.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/SelectionService.cs
elachlan: Here: `_contextAttributes = new string?[_selection.Count];`

When we create the array, we initialize it to a size, so all the positions in the array start as null.

dreddy-work: in the old code, `_selection` contained nulls so is `_contextAttributes`. Do you see any reason for the size of `_contextAttributes` should match with size of `_selection` ? If not, why can't we go with dynamic size here instead of preallocating?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/SelectionService.cs
dreddy-work: Having null selection would throw eventually in `TypeDescriptor.GetClassName`.  It shouldn't be breaking change and it is internal.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/SelectionService.cs
elachlan: It did this previously, we just changed the nullability on _contextAttributes.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/SelectionService.cs
dreddy-work: Actually leave it as-is. Not much gain in restructuring this block.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/SelectionService.cs
dreddy-work: right. given we do not insert nulls into `_selection `and we don't insert nulls into `_contextAttributes`, we can avoid pre-allocation with null values. am i missing something?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/SelectionService.cs
dreddy-work: sorry for confusion. I meant, throw right away if null and not allow nulls.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/SelectionService.cs
elachlan: I believe I did this previous and felt it was slower using a List<string> vs string[]. But happy to go with List<string> if you would like.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/SelectionService.cs
elachlan: Ok, I'll sort it out.

dmitrii-drobotov (approved):

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/SelectionService.cs
```diff
@@ -233,19 +227,22 @@ private void UpdateHelpKeyword(bool tryLater)
                 }
             }
 
-            _contextAttributes = new string[_selection.Count];
+            _contextAttributes = new(_selection.Count);
```
dreddy-work, line 230: why do you need initial capacity here?

dreddy-work (approved):

PR #8556, opened 2023-02-02
Enable CheckCertificateRevocationList Under a Switch

lonitra:
Related: [CodeQL](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1666879/) , https://github.com/dotnet/winforms/issues/8561
- Introduces a switch for turning on `ServicePointManager.CheckCertificateRevocationList`
- Fixes tests and adds test to check for same behavior when switch is both on and off
- Add CodeQL suppression in case it tries to reflag this because of the switch
- General clean up

**Background:**
Before creating a `WebClient` or `WebRequest`, it is best practice to set `ServicePointManager.CheckCertificateRevocationList` to `true` beforehand so that revoked certificates are not accepted by `WebClient`s or `WebRequest`s as valid. 
Now that we are checking certificate revocation list, the behavior should remain the same as before, however there is a rare case when it could fail which is if:
a) the certificate authority provides revocation only through a technology called CRL, 
b) the client has not recently enough cached the relevant CRL, and 
c) the client is operating on an egress-restricted network and the CRL location isn't on the egress allow-list.  
There may be some other corner cases, and a mild performance hit, hence why this is placed under a switch for user to have opportunity to opt out if they run into this.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8556)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
@@ -15,13 +15,14 @@ namespace System.Windows.Forms.Primitives
         // for more details on how to enable these switches in the application.
         private const string ScaleTopLevelFormMinMaxSizeForDpiSwitchName = "System.Windows.Forms.ScaleTopLevelFormMinMaxSizeForDpi";
         internal const string AnchorLayoutV2SwitchName = "System.Windows.Forms.AnchorLayoutV2";
-        internal const string ServicePointManagerCheckCRLSwitchName = "System.Windows.Forms.ServicePointManagerCheckCRL";
+        internal const string ServicePointManagerCheckCrlSwitchName = "System.Windows.Forms.ServicePointManagerCheckCrl";
         internal const string TrackBarModernRenderingSwitchName = "System.Windows.Forms.TrackBarModernRendering";
         private static int s_AnchorLayoutV2;
         private static int s_scaleTopLevelFormMinMaxSizeForDpi;
-        private static int s_servicePointManagerCheckCRL;
+        private static int s_servicePointManagerCheckCrl;
         private static int s_trackBarModernRendering;
         private static FrameworkName? s_targetFrameworkName;
+        private static Version? s_minimumSupportVersion;
 
         /// <summary>
         ///  The <see cref="TargetFrameworkAttribute"/> value for the entry assembly, if any.
@@ -32,6 +33,18 @@ namespace System.Windows.Forms.Primitives
             {
                 s_targetFrameworkName ??= AppContext.TargetFrameworkName is { } name ? new(name) : null;
                 return s_targetFrameworkName;
+            }
+        }
+
+        /// <summary>
+        ///  The earliest <see cref="Version"/> switches are supported in.
+        /// </summary>
+        private static Version MinimumSupportVersion
+        {
+            get
+            {
+                s_minimumSupportVersion ??= new("8.0");
+                return s_minimumSupportVersion;
             }
         }
 
@@ -82,7 +95,7 @@ namespace System.Windows.Forms.Primitives
 
                 // We are introducing switch defaults in .NET 8.0+ and support matrix for this product is
                 // limited to Windows 10 and above versions.
-                if (OsVersion.IsWindows10_1703OrGreater() && TargetFrameworkName!.Version.CompareTo(new Version("8.0")) >= 0)
+                if (OsVersion.IsWindows10_1703OrGreater() && TargetFrameworkName?.Version.CompareTo(MinimumSupportVersion) >= 0)
                 {
                     if (switchName == ScaleTopLevelFormMinMaxSizeForDpiSwitchName)
                     {
@@ -99,7 +112,7 @@ namespace System.Windows.Forms.Primitives
                         return true;
                     }
 
-                    if (switchName == ServicePointManagerCheckCRLSwitchName)
+                    if (switchName == ServicePointManagerCheckCrlSwitchName)
                     {
                         return true;
                     }
@@ -137,13 +150,13 @@ namespace System.Windows.Forms.Primitives
 
         /// <summary>
         ///  Indicates whether certificates are checked against the certificate authority revocation list.
-        ///  If true, revoked certificates will be accepted by WebRequests and WebClients as invalid.
+        ///  If true, revoked certificates will not be accepted by WebRequests and WebClients as valid.
         ///  Otherwise, revoked certificates will be accepted as valid.
         /// </summary>
-        public static bool ServicePointManagerCheckCRL
+        public static bool ServicePointManagerCheckCrl
         {
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
-            get => GetCachedSwitchValue(ServicePointManagerCheckCRLSwitchName, ref s_servicePointManagerCheckCRL);
+            get => GetCachedSwitchValue(ServicePointManagerCheckCrlSwitchName, ref s_servicePointManagerCheckCrl);
         }
     }
 }
--- a/src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
@@ -15,9 +15,11 @@ namespace System.Windows.Forms.Primitives
         // for more details on how to enable these switches in the application.
         private const string ScaleTopLevelFormMinMaxSizeForDpiSwitchName = "System.Windows.Forms.ScaleTopLevelFormMinMaxSizeForDpi";
         internal const string AnchorLayoutV2SwitchName = "System.Windows.Forms.AnchorLayoutV2";
+        internal const string ServicePointManagerCheckCRLSwitchName = "System.Windows.Forms.ServicePointManagerCheckCRL";
         internal const string TrackBarModernRenderingSwitchName = "System.Windows.Forms.TrackBarModernRendering";
         private static int s_AnchorLayoutV2;
         private static int s_scaleTopLevelFormMinMaxSizeForDpi;
+        private static int s_servicePointManagerCheckCRL;
         private static int s_trackBarModernRendering;
         private static FrameworkName? s_targetFrameworkName;
 
@@ -112,6 +114,11 @@ namespace System.Windows.Forms.Primitives
                         {
                             return true;
                         }
+
+                        if (switchName == ServicePointManagerCheckCRLSwitchName)
+                        {
+                            return true;
+                        }
                     }
                 }
 
@@ -129,6 +136,17 @@ namespace System.Windows.Forms.Primitives
         {
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
             get => GetCachedSwitchValue(TrackBarModernRenderingSwitchName, ref s_trackBarModernRendering);
+        }
+
+        /// <summary>
+        ///  Indicates whether certificates are checked against the certificate authority revocation list.
+        ///  If true, revoked certificates will be accepted by WebRequests and WebClients as invalid.
+        ///  Otherwise, revoked certificates will be accepted as valid.
+        /// </summary>
+        public static bool ServicePointManagerCheckCRL
+        {
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            get => GetCachedSwitchValue(ServicePointManagerCheckCRLSwitchName, ref s_servicePointManagerCheckCRL);
         }
     }
 }
--- a/src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
@@ -40,20 +40,6 @@ namespace System.Windows.Forms.Primitives
         /// </summary>
         public static bool IsNetCoreApp => string.Equals(TargetFrameworkName?.Identifier, ".NETCoreApp");
 
-        /// <summary>
-        ///  Indicates whether AnchorLayoutV2 feature is enabled.
-        /// </summary>
-        /// <devdoc>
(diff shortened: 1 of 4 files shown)


src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -130,5 +134,16 @@ public static bool TrackBarModernRendering
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
             get => GetCachedSwitchValue(TrackBarModernRenderingSwitchName, ref s_trackBarModernRendering);
         }
+
+        /// <summary>
+        ///  Indicates whether certificates are checked against the certificate authority revocation list.
+        ///  If true, revoked certificates will be accepted by WebRequests and WebClients as invalid.
```
dreddy-work, line 140: is this line should be  
```suggestion
        ///  If true, revoked certificates will not be accepted by WebRequests and WebClients as invalid.
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PictureBoxTests.cs
```diff
@@ -741,13 +744,44 @@ public void PictureBox_ImageLocation_SetValidWithWaitOnLoadTrueUri_GetReturnsExp
...
+        public void PictureBox_ImageLocation_SetValidWithWaitOnLoadTrueUri_ConfigSwitch_CheckCRL_Enabled_GetReturnsExpected()
+        {
+            try
+            {
+                using var pictureBox = new PictureBox
+                {
+                    WaitOnLoad = true
+                };
+
+                AppContext.SetSwitch(LocalAppContextSwitches.ServicePointManagerCheckCRLSwitchName, true);
```
dreddy-work, line 775: Q: i do not see diff between setting switch `false` or `true` in these tests.

lonitra: This test checks if we will still get the same behavior if the switch is set to `true` as when the switch is set to `false`. I wasn't sure how to make this test into a `[WinFormsTheory]` because it caches the value the switch is set to the first time it is set so I've created separate tests for them. Do you have suggestions on how I could make it into a `[WinFormsTheory]`?

The only time the behavior should be different between when the switch is set to `true` vs `false` is when client is using certificate that is on the revocation list, but I am not sure how I'd be able to put that situation into a test and also think that this is not something for us to spend time testing.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
lonitra: Is this clearer?
"If true, revoked certificates will not be accepted by WebRequests and WebClients as valid."

dreddy-work: Yep.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PictureBoxTests.cs
```diff
@@ -1999,14 +2033,45 @@ public void PictureBox_Load_UrlValidWithWaitOnLoadTrueUri_GetReturnsExpected()
                     WaitOnLoad = true
                 };
 
-                pictureBox.ImageLocation = UrlImageLocation;
+                AppContext.SetSwitch(LocalAppContextSwitches.ServicePointManagerCheckCRLSwitchName, false);
```
dreddy-work, line 2036: this switch needs to be reset. Otherwise any future tests that may rely on this switch will be impacted.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PictureBoxTests.cs
dreddy-work: Oh okay. If you want to convert this into theory,  you can look at `LocalAppContextSwitchesTest` on how we are resetting the switch.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -94,31 +82,47 @@ static bool GetSwitchDefaultValue(string switchName)
 
                 // We are introducing switch defaults in .NET 8.0+ and support matrix for this product is
                 // limited to Windows 10 and above versions.
-                if (OsVersion.IsWindows10_1703OrGreater())
+                if (OsVersion.IsWindows10_1703OrGreater() && TargetFrameworkName!.Version.CompareTo(new Version("8.0")) >= 0)
```
JeremyKuhne, line 85: I don't know why we are gating on Windows 10. That doesn't make any sense to me. Also:

- `TargetFrameworkName` can be null
- We shouldn't be regenerating `Version` every time we come in this method. This should be cached.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PictureBoxTests.cs
JeremyKuhne: Yes, this should use a scope for messing with the switch.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
JeremyKuhne: One other comment. We're introducing _these_ switches for .NET 8. It might be clearer to say "Switches added in .NET 8".

dreddy-work: The two switches related to scaling and anchors require Windows 10 and since our testing coverage is limited and we don't prevent installation on lower operating systems, it's safe to implement these breaking changes (hence, switches) only on Windows 10 and higher.

lonitra: > One other comment. We're introducing _these_ switches for .NET 8. It might be clearer to say "Switches added in .NET 8".

Are you referring to the comment on line 84/96?

JeremyKuhne: > The two switches related to scaling and anchors require Windows 10 and since our testing coverage is limited to > windows 10 and we don't prevent installation on lower operating systems, it's safe to implement these breaking changes (hence, switches) only on Windows 10 and higher.

@dreddy-work ultimately it makes no sense to have that check duplicated. You'll have to check for the OS anyway at the usage site as people can manually set this switch. Even if we did keep it (I would not), it should have been documented, as you're seeing we're risking accruing things that have no relation to the OS version.

> Are you referring to the comment on line 84/96?

@lonitra Yes. The current comment infers that everything is gated on .NET 8. We will end up having switches for .NET 9, 10, etc. The current comment risks switches added in .NET 9 being lumped into .NET 8 as the intent isn't super clear.

dreddy-work: > @dreddy-work ultimately it makes no sense to have that check duplicated. You'll have to check for the OS anyway at the usage site as people can manually set this switch. Even if we did keep it (I would not), it should have been documented, as you're seeing we're risking accruing things that have no relation to the OS version.

I need to review the use cases before making a decision on the Windows 10 restriction. @lonitra, since this isn't added in the current PR, please continue with the rest of the review feedback and create a tracking issue for this comment, assigning it to me.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -34,23 +37,21 @@ internal static partial class LocalAppContextSwitches
         }
 
         /// <summary>
-        ///  Returns <see langword="true"/> if the <see cref="TargetFrameworkAttribute"/> value for the entry assembly specifies .NET Core.
+        ///  The earliest <see cref="Version"/> switches are supported in.
         /// </summary>
-        public static bool IsNetCoreApp => string.Equals(TargetFrameworkName?.Identifier, ".NETCoreApp");
+        private static Version MinimumSupportVersion
```
JeremyKuhne, line 42: What I meant is the creation and comparison on every check. Additionally, you should not be using strings to create the `Version` as you're forcing an unnecessary parse. You should use the overloads that take a numerical value.

In any case, we're losing the intent here. This is not about `MinimumSupportVersion`, this is about "version 8.0" which only applies to our switches we're adding now. In .NET 9 we'll probably need a "version 9.0", etc. Additionally, for our purposes we don't ever expect to care about anything other than the major version. This can be simplified to simply caching the major version as an int (with a -1 if we don't have a `FrameworkName`) and comparing against that in our logic. If we ever find that we need to care about the minor we can add the additional logic when as needed.

Reminder that with all of this being statics, be careful about putting them in dependency order.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -94,31 +95,47 @@ static bool GetSwitchDefaultValue(string switchName)
 
                 // We are introducing switch defaults in .NET 8.0+ and support matrix for this product is
                 // limited to Windows 10 and above versions.
-                if (OsVersion.IsWindows10_1703OrGreater())
+                if (OsVersion.IsWindows10_1703OrGreater() && TargetFrameworkName?.Version.CompareTo(MinimumSupportVersion) >= 0)
```
JeremyKuhne, line 98: As mentioned in my other comment, the OS check really shouldn't be done here period. At the very least it has nothing to do with your switch. This logic should be something like:

``` C#
if (s_frameworkVersion >= 8)
{
    // .NET 8 switches
    if (switchName == ...
}

// In the future...
if (s_frameworkVersion >= 9)
```

JeremyKuhne, line 98: As mentioned in my other comment, the OS check really shouldn't be done here period. At the very least it has nothing to do with your switch. This logic should be something like:

``` C#
if (s_frameworkVersion >= 8)
{
    // .NET 8 switches
    if (switchName == ...
}

// In the future...
if (s_frameworkVersion >= 9)
```

JeremyKuhne (approved):

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -113,12 +106,31 @@ static bool GetSwitchDefaultValue(string switchName)
                             return true;
```
Tanya-Solyanik, line 106: Not related to this PR, but we adopted a pattern where the default that we assume going forward (NET9+) is `true`, i.e. not a default value and requires an assignment. So each switch will contribute 3 IL lines to do an assignment even when developer had not set any switches. Of cause otherwise the code is not as readable as it is currently. Had that been discussed?

lonitra: I'm not sure I follow here. What do you mean by it is not a default value and requires an assigment with our assumption going forward the default is `true`?

JeremyKuhne: I'm not sure why we would change the established pattern. The IL you're talking about (setting a bool) is trivial in the grand scheme of things. This is only called once for each switch and a far larger impact would be to try and get the string comparisons out for things the user doesn't define. For example: could this be turned into an internal enum and have the matching only done to convert a string to an enum value? That seems relatively easy to accomplish.

Tanya-Solyanik: @JeremyKuhne - we don't even need enums. We have a very specific (per switch string) get property - 
```
        public static bool AnchorLayoutV2
        {
            get => GetCachedSwitchValue(AnchorLayoutV2SwitchName, ref s_AnchorLayoutV2);
        }
```
from that property, we call a generic method that should handle multiple switches. Instead, we can handle each switch directly in the get method.

@lonitra -  `private static int s_AnchorLayoutV2;` - this field will be `false` by default. in NET9 we would want layout to follow the new code path unless explicitly opted out because we want to support PMv2 out of the box. So for each application that does not opt out of V2 layout, we will have 3 IL lines that do `s_AnchorLayoutV2=true;`. This is negligible.

JeremyKuhne: > we don't even need enums. We have a very specific (per switch string) get property 

@Tanya-Solyanik Yes, and those are passing strings that will do a string comparison. Comparing two ints for equality is far, far faster (which is what an enum equality check is).

> Instead, we can handle each switch directly in the get method.

You do that by `[MethodImpl(MethodImplOptions.AggressiveInlining)]` on `GetCachedSwitchValue()`, which is missing from this code.

In any case the only thing that truly matters is how fast the switch access is once it is set. As mentioned, we're missing an `AggressiveInlining` and that should be fixed. This might be able to be further optimized by having a single return statement- I'll take a look.

Once we accumulate a number of switches, then we can look at reducing the overhead in the default handling by doing something like using enums.

JeremyKuhne: I did look at the perf here. The way it is written is really about as fast as you can get this (about .1ns to check a cached value). Using a single return statement doesn't help in this case, and `GetCachedSwitchValue()` is already inlining in simple perf tests, but should probably be added anyway when we have the chance.

lonitra:
https://github.com/dotnet/docs/issues/37551

PR #8570, opened 2023-02-05

gpetrou:
## Proposed changes

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8570)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
@@ -2203,7 +2203,7 @@ namespace System.Windows.Forms
                 return null;
             }
 
-            Debug.WriteLineIf(s_selectionDebug!.TraceVerbose && (current != -1), "[SelectDBG GetNextToolStripItem] Last selected item was " + ((current != -1) ? DisplayedItems[current].Text : ""));
+            Debug.WriteLineIf(s_selectionDebug!.TraceVerbose && (current != -1), $"[SelectDBG GetNextToolStripItem] Last selected item was {DisplayedItems[current].Text}");
             Debug.WriteLineIf(s_selectionDebug!.TraceVerbose && (current == -1), "[SelectDBG GetNextToolStripItem] Last selected item was null");
 
             int count = DisplayedItems.Count;
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
@@ -2407,14 +2407,7 @@ namespace System.Windows.Forms
         internal static Size GetPreferredSizeHorizontal(IArrangedElement container, Size proposedConstraints)
         {
             Size maxSize = Size.Empty;
-            ToolStrip? toolStrip = container as ToolStrip;
-
-            Debug.Assert(toolStrip is not null);
-
-            if (toolStrip is null)
-            {
-                return maxSize;
-            }
+            ToolStrip toolStrip = (container as ToolStrip)!;
 
             // ensure preferred size respects default size as a minimum.
             Size defaultSize = toolStrip.DefaultSize - toolStrip.Padding.Size;
@@ -2478,14 +2471,7 @@ namespace System.Windows.Forms
         {
             Size maxSize = Size.Empty;
             bool requiresOverflow = false;
-            ToolStrip? toolStrip = container as ToolStrip;
-
-            Debug.Assert(toolStrip is not null);
-
-            if (toolStrip is null)
-            {
-                return maxSize;
-            }
+            ToolStrip toolStrip = (container as ToolStrip)!;
 
             bool foundItemParticipatingInLayout = false;
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
@@ -2407,7 +2407,7 @@ namespace System.Windows.Forms
         internal static Size GetPreferredSizeHorizontal(IArrangedElement container, Size proposedConstraints)
         {
             Size maxSize = Size.Empty;
-            ToolStrip toolStrip = (container as ToolStrip)!;
+            ToolStrip toolStrip = (ToolStrip)container;
 
             // ensure preferred size respects default size as a minimum.
             Size defaultSize = toolStrip.DefaultSize - toolStrip.Padding.Size;
@@ -2471,7 +2471,7 @@ namespace System.Windows.Forms
         {
             Size maxSize = Size.Empty;
             bool requiresOverflow = false;
-            ToolStrip toolStrip = (container as ToolStrip)!;
+            ToolStrip toolStrip = (ToolStrip)container;
 
             bool foundItemParticipatingInLayout = false;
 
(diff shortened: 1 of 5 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
```diff
@@ -1779,7 +1779,7 @@ internal ToolTip ToolTip
                 }
                 else
                 {
-                    toolTip = (ToolTip)Properties.GetObject(ToolStrip.s_propToolTip);
+                    toolTip = (ToolTip)Properties.GetObject(ToolStrip.s_propToolTip)!;
```
Tanya-Solyanik, line 1782: Consider using `TryGet` pattern instead of using `Contains`.
https://github.com/dotnet/winforms/blob/766c440e9c209194e930216f7d7811405a44a9a0/src/System.Windows.Forms/src/System/Windows/Forms/PropertyStore.cs#L190. `TryGet` is annotated so that we can get rid of the `!`.

gpetrou: I created https://github.com/dotnet/winforms/pull/8575 to apply this suggestion in more places.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
```diff
@@ -2217,8 +2217,8 @@ private ToolStripItem GetNextItemHorizontal(ToolStripItem start, bool forward)
                 return null;
             }
 
-            Debug.WriteLineIf(s_selectionDebug.TraceVerbose && (current != -1), "[SelectDBG GetNextToolStripItem] Last selected item was " + ((current != -1) ? DisplayedItems[current].Text : ""));
-            Debug.WriteLineIf(s_selectionDebug.TraceVerbose && (current == -1), "[SelectDBG GetNextToolStripItem] Last selected item was null");
+            Debug.WriteLineIf(s_selectionDebug!.TraceVerbose && (current != -1), "[SelectDBG GetNextToolStripItem] Last selected item was " + ((current != -1) ? DisplayedItems[current].Text : ""));
```
lonitra, line 2220: We are trying to use string interpolation throughout our codebase instead now
```suggestion
            Debug.WriteLineIf(s_selectionDebug!.TraceVerbose && (current != -1), $"[SelectDBG GetNextToolStripItem] Last selected item was {((current != -1) ? DisplayedItems[current].Text : "")}");
```

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
Tanya-Solyanik: @gpetrou  - thank you!

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
```diff
@@ -2208,8 +2208,8 @@ private ToolStripItem GetNextItemHorizontal(ToolStripItem start, bool forward)
                 return null;
             }
 
-            Debug.WriteLineIf(s_selectionDebug.TraceVerbose && (current != -1), "[SelectDBG GetNextToolStripItem] Last selected item was " + ((current != -1) ? DisplayedItems[current].Text : ""));
-            Debug.WriteLineIf(s_selectionDebug.TraceVerbose && (current == -1), "[SelectDBG GetNextToolStripItem] Last selected item was null");
+            Debug.WriteLineIf(s_selectionDebug!.TraceVerbose && (current != -1), $"[SelectDBG GetNextToolStripItem] Last selected item was {((current != -1) ? DisplayedItems[current].Text : string.Empty)}");
```
lonitra, line 2211: I don't think we need to do another check for `current != -1` in the string. We are already checking that in the first param of the `Debug.WriteLineIf`.
```suggestion
            Debug.WriteLineIf(s_selectionDebug!.TraceVerbose && (current != -1), $"[SelectDBG GetNextToolStripItem] Last selected item was {DisplayedItems[current].Text}");
```

lonitra (commented):
Small comment on an extra check that we do not need. Otherwise looks good!

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
```diff
@@ -2413,7 +2412,14 @@ internal override Size GetPreferredSizeCore(Size proposedSize)
         internal static Size GetPreferredSizeHorizontal(IArrangedElement container, Size proposedConstraints)
         {
             Size maxSize = Size.Empty;
-            ToolStrip toolStrip = container as ToolStrip;
+            ToolStrip? toolStrip = container as ToolStrip;
+
+            Debug.Assert(toolStrip is not null);
```
dreddy-work, line 2417: why this change? earlier we were throwing null reference and now not. If we do not expect Toolstrip to be null at this place, it would be better to throw it right away.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
```diff
@@ -2565,7 +2578,7 @@ internal bool GetToolStripState(int flag)
             return (_toolStripState & flag) != 0;
         }
 
-        internal virtual ToolStrip GetToplevelOwnerToolStrip()
+        internal virtual ToolStrip? GetToplevelOwnerToolStrip()
```
dreddy-work, line 2581: why nullable?

gpetrou: We override this in `ToolStripDropDown` and there it is nullable.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
gpetrou: I applied a similar change to what we did in https://github.com/dotnet/winforms/pull/8117/files.

gpetrou:
@dreddy-work could we merge this?

dreddy-work:
> @dreddy-work could we merge this?

Not yet. We discussed this. I will be adding details on what we should be doing in the scenarios like this.

gpetrou:
> Not yet. We discussed this. I will be adding details on what we should be doing in the scenarios like this.

@dreddy-work could you please share these details?

gpetrou:
@dreddy-work do you want any more changes to this one?

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
```diff
@@ -2472,7 +2471,7 @@ internal static Size GetPreferredSizeVertical(IArrangedElement container)
         {
             Size maxSize = Size.Empty;
             bool requiresOverflow = false;
-            ToolStrip toolStrip = container as ToolStrip;
+            ToolStrip toolStrip = (container as ToolStrip)!;
```
dreddy-work, line 2474: why not 
```suggestion
            ToolStrip toolStrip = (ToolStrip)container;
```

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
```diff
@@ -2407,14 +2407,7 @@ internal override Size GetPreferredSizeCore(Size proposedSize)
...
             Size maxSize = Size.Empty;
-            ToolStrip? toolStrip = container as ToolStrip;
-
-            Debug.Assert(toolStrip is not null);
-
-            if (toolStrip is null)
-            {
-                return maxSize;
-            }
+            ToolStrip toolStrip = (container as ToolStrip)!;
```
dreddy-work, line 2410: same here.

dreddy-work:
> @dreddy-work do you want any more changes to this one?

Thanks @gpetrou. Rest look good to me.

dreddy-work (approved):

PR #8571, opened 2023-02-05
Use TabPage generic list instead of array in TabControl

gpetrou:
## Proposed changes

- Use TabPage generic list instead of array in TabControl.
- Addresses some of the feedback in https://github.com/dotnet/winforms/pull/7316.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8571)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
@@ -802,10 +802,7 @@ namespace System.Windows.Forms
         [Browsable(false)]
         [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
         [SRDescription(nameof(SR.TabBaseTabCountDescr))]
-        public int TabCount
-        {
-            get { return _tabPages?.Count ?? 0; }
-        }
+        public int TabCount => _tabPages?.Count ?? 0;
 
         /// <summary>
         ///  Returns the Collection of TabPages.
@@ -933,7 +930,7 @@ namespace System.Windows.Forms
 
         private int AddNativeTabPage(TabPage tabPage)
         {
-            int index = SendMessage(PInvoke.TCM_INSERTITEMW, (_tabPages?.Count ?? 0) + 1, tabPage);
+            int index = SendMessage(PInvoke.TCM_INSERTITEMW, TabCount + 1, tabPage);
             User32.PostMessageW(this, _tabBaseReLayoutMessage);
             return index;
         }
@@ -1074,7 +1071,7 @@ namespace System.Windows.Forms
 
         internal TabPage GetTabPage(int index)
         {
-            if (index < 0 || index >= (_tabPages?.Count ?? 0))
+            if (index < 0 || index >= TabCount)
             {
                 throw new ArgumentOutOfRangeException(nameof(index), index, string.Format(SR.InvalidArgument, nameof(index), index));
             }
@@ -1100,7 +1097,7 @@ namespace System.Windows.Forms
         /// </summary>
         protected virtual object[] GetItems(Type baseType)
         {
-            int tabPageCount = _tabPages?.Count ?? 0;
+            int tabPageCount = TabCount;
             object[] result = (object[])Array.CreateInstance(baseType, tabPageCount);
             if (tabPageCount > 0)
             {
@@ -1123,7 +1120,7 @@ namespace System.Windows.Forms
         /// </summary>
         public Rectangle GetTabRect(int index)
         {
-            if (index < 0 || (index >= (_tabPages?.Count ?? 0) && !GetState(State.GetTabRectfromItemSize)))
+            if (index < 0 || (index >= TabCount && !GetState(State.GetTabRectfromItemSize)))
             {
                 throw new ArgumentOutOfRangeException(nameof(index), index, string.Format(SR.InvalidArgument, nameof(index), index));
             }
@@ -1181,7 +1178,7 @@ namespace System.Windows.Forms
         /// </summary>
         private void InsertItem(int index, TabPage tabPage)
         {
-            if (index < 0 || index > (_tabPages?.Count ?? 0))
+            if (index < 0 || index > TabCount)
             {
                 throw new ArgumentOutOfRangeException(nameof(index), index, string.Format(SR.InvalidArgument, nameof(index), index));
             }
@@ -1614,7 +1611,7 @@ namespace System.Windows.Forms
 
         private void RemoveTabPage(int index)
         {
-            if (index < 0 || index >= (_tabPages?.Count ?? 0))
+            if (index < 0 || index >= TabCount)
             {
                 throw new ArgumentOutOfRangeException(nameof(index), index, string.Format(SR.InvalidArgument, nameof(index), index));
             }
@@ -1669,7 +1666,7 @@ namespace System.Windows.Forms
 
         private void SetTabPage(int index, TabPage value)
         {
-            if (index < 0 || index >= (_tabPages?.Count ?? 0))
+            if (index < 0 || index >= TabCount)
             {
                 throw new ArgumentOutOfRangeException(nameof(index), index, string.Format(SR.InvalidArgument, nameof(index), index));
             }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
@@ -1621,7 +1621,6 @@ namespace System.Windows.Forms
 
             if (index < _tabPages.Count)
             {
-                _tabPages[index] = null;
                 _tabPages.RemoveAt(index);
             }
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
@@ -729,7 +729,7 @@ namespace System.Windows.Forms
             get
             {
                 int index = SelectedIndex;
-                if (index == -1 || _tabPages is null || index > _tabPages.Count - 1)
+                if (index == -1 || _tabPages is null)
                 {
                     return null;
                 }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
@@ -6,7 +6,6 @@
 
 using System.Collections.Specialized;
 using System.ComponentModel;
-using System.Diagnostics;
 using System.Drawing;
 using System.Drawing.Design;
 using System.Globalization;
@@ -70,8 +69,7 @@ namespace System.Windows.Forms
         private readonly User32.WM _tabBaseReLayoutMessage = User32.RegisterWindowMessageW(Application.WindowMessagesVersion + TabBaseReLayoutMessageName);
 
         // State
-        private TabPage[] _tabPages;
-        private int _tabPageCount;
+        private List<TabPage> _tabPages;
         private int _lastSelection;
         private short _windowId;
 
@@ -731,12 +729,11 @@ namespace System.Windows.Forms
             get
             {
                 int index = SelectedIndex;
-                if (index == -1 || _tabPages is null)
+                if (index == -1 || _tabPages is null || index > _tabPages.Count - 1)
                 {
                     return null;
                 }
 
-                Debug.Assert(index >= 0 && index < _tabPages.Length, "SelectedIndex returned an invalid index");
                 return _tabPages[index];
             }
             set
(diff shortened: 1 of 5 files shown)


src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TabControlTests.cs
```diff
@@ -3907,17 +3907,26 @@ public void TabControl_GetItems_InvokeInvalidTypeWithoutPages_ThrowsInvalidCastE
...
+
+        [WinFormsFact]
+        public void TabControl_GetItems_InvokeInvalidInheritedTypeWithPages_ThrowsArrayTypeMismatchException()
+        {
+            using var control = new SubTabControl();
+            using var page1 = new TabPage();
+            using var page2 = new SubTabPage();
+            control.TabPages.Add(page1);
+            control.TabPages.Add(page2);
+            Assert.Throws<ArrayTypeMismatchException>(() => control.GetItems(typeof(SubTabPage)));
```
dmitrii-drobotov, line 3929: I'm not sure if it's allowed to change the exception type this way per https://github.com/dotnet/runtime/blob/main/docs/coding-guidelines/breaking-change-rules.md#exceptions. On the other hand, [the documentation](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.tabcontrol.getitems?view=windowsdesktop-7.0#system-windows-forms-tabcontrol-getitems(system-type)) already says that `ArrayTypeMismatchException` is thrown in this case, so this is probably a good change to do. Although I would like to hear another opinion on this.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
```diff
@@ -1625,23 +1610,21 @@ protected void RemoveAll()
...
             {
                 throw new ArgumentOutOfRangeException(nameof(index), index, string.Format(SR.InvalidArgument, nameof(index), index));
             }
 
-            _tabPageCount--;
-            if (index < _tabPageCount)
+            if (index < _tabPages.Count)
             {
-                Array.Copy(_tabPages, index + 1, _tabPages, index, _tabPageCount - index);
+                _tabPages[index] = null;
```
dmitrii-drobotov, line 1624: Nit: the line `_tabPages[index] = null;` can be removed, it's sufficient to just call `RemoveAt`.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
```diff
@@ -1164,24 +1164,10 @@ private void ImageListRecreateHandle(object sender, EventArgs e)
...
-            {
-                _tabPages = new TabPage[4];
-            }
-            else if (_tabPages.Length == _tabPageCount)
-            {
-                TabPage[] newTabPages = new TabPage[_tabPageCount * 2];
-                Array.Copy(_tabPages, 0, newTabPages, 0, _tabPageCount);
-                _tabPages = newTabPages;
-            }
+            _tabPages ??= new List<TabPage>(4);
```
dmitrii-drobotov, line 1167: Nit: consider removing the initial size argument, I don't think it's needed now that we use `List`.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TabControlTests.cs
dreddy-work: Correcting code to align with already documented information should not result in a breaking change, but it may cause some existing applications to break if they are retargeted. That should be acceptable when retargeting.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
```diff
@@ -731,12 +729,11 @@ public TabPage SelectedTab
             get
             {
                 int index = SelectedIndex;
-                if (index == -1 || _tabPages is null)
+                if (index == -1 || _tabPages is null || index > _tabPages.Count - 1)
```
dreddy-work, line 732: This is public property. Why are you suppressing index out of range exception here?

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
```diff
@@ -807,7 +804,7 @@ public bool ShowToolTips
         [SRDescription(nameof(SR.TabBaseTabCountDescr))]
         public int TabCount
         {
-            get { return _tabPageCount; }
+            get { return _tabPages?.Count ?? 0; }
```
dreddy-work, line 807: q: Is it worth caching count? and synching it with `insert` and `remove` operations?

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
gpetrou: There were some failing tests without this.

dreddy-work: i see now why. it might be returning null or `IndexOutOfRange` exception earlier depending on the size. Given that now we are not pre allocating memory, throwing `indexOutOfRange` is reasonable and we can document it.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
```diff
@@ -1090,24 +1087,27 @@ internal TabPage GetTabPage(int index)
...
-            object[] result = (object[])Array.CreateInstance(baseType, _tabPageCount);
-            if (_tabPageCount > 0)
+            int tabPageCount = _tabPages?.Count ?? 0;
+            object[] result = (object[])Array.CreateInstance(baseType, tabPageCount);
+            if (tabPageCount > 0)
             {
-                Array.Copy(_tabPages, 0, result, 0, _tabPageCount);
+                for (int i = 0; i < tabPageCount; i++)
+                {
+                    result[i] = _tabPages[i];
```
dreddy-work, line 1109: isn't ToArray() doing implicit cast here?

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
gpetrou: I added a separate commit. Is this what you had in mind?

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
gpetrou: Hmm, calling this seems to take less than half nanosecond. So, I would say no.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
gpetrou: Using `ToArray` will break other tests such as `TabControl_GetItems_InvokeTypeBaseTypeWithPages_ReturnsExpected`.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
RussKie: This piece of code is repeated over and over, which is a maintenance overhead. Suggest abstracting it behind a private property.

Tanya-Solyanik: Caching might not be a good idea because this collection might change dynamically.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
dreddy-work: yep. Thanks.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
dreddy-work: Curious, @gpetrou, can you share error info on these tests failure?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TabControlTests.cs
```diff
@@ -3836,7 +3835,7 @@ public void TabControl_GetItems_InvokeWithoutPages_ReturnsExpected()
             using var control = new SubTabControl();
             object[] result = Assert.IsType<TabPage[]>(control.GetItems());
             Assert.Empty(result);
-            Assert.NotSame(result, control.GetItems());
+            Assert.Same(result, control.GetItems());
```
dreddy-work, line 3838: q: what was changing here?

gpetrou: We now use `Array.Empty<TabPage>()` in `GetItems`. We have done similar changes in the tests in the past because of the usage of `Array.Empty`.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
gpetrou: I had just used `result = _tabPages.ToArray();`. I have changed that to `Array.Copy(_tabPages.ToArray(), 0, result, 0, TabCount);`. So, these tests behave the same as in main branch.

gpetrou:
```
Message: 
    Assert.IsType() Failure
    Expected: System.Windows.Forms.Tests.TabControlTests+SubTabPage[]
    Actual:   System.Windows.Forms.TabPage[]

  Stack Trace: 
    TabControlTests.TabControl_GetItems_InvokeTypeSubTypeWithPages_ReturnsExpected() line 3889
    RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
    MethodInvoker.Invoke(Object obj, IntPtr* args, BindingFlags invokeAttr)
```
```
  Message: 
    Assert.IsType() Failure
    Expected: System.Windows.Forms.Control[]
    Actual:   System.Windows.Forms.TabPage[]

  Stack Trace: 
    TabControlTests.TabControl_GetItems_InvokeTypeBaseTypeWithPages_ReturnsExpected(Type baseType) line 3878
    InvokeStub_TabControlTests.TabControl_GetItems_InvokeTypeBaseTypeWithPages_ReturnsExpected(Object, Object, IntPtr*)
    MethodInvoker.Invoke(Object obj, IntPtr* args, BindingFlags invokeAttr)
```
```
Message: 
    Assert.IsType() Failure
    Expected: System.Object[]
    Actual:   System.Windows.Forms.TabPage[]

  Stack Trace: 
    TabControlTests.TabControl_GetItems_InvokeTypeBaseTypeWithPages_ReturnsExpected(Type baseType) line 3878
    RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)
    MethodInvoker.Invoke(Object obj, IntPtr* args, BindingFlags invokeAttr)
```
```
Message: 
    Assert.Throws() Failure
    Expected: typeof(System.InvalidCastException)
    Actual:   (No exception was thrown)

  Stack Trace: 
    TabControlTests.TabControl_GetItems_InvokeInvalidTypeWithPages_ThrowsInvalidCastException(Type baseType) line 3919
    InvokeStub_TabControlTests.TabControl_GetItems_InvokeInvalidTypeWithPages_ThrowsInvalidCastException(Object, Object, IntPtr*)
    MethodInvoker.Invoke(Object obj, IntPtr* args, BindingFlags invokeAttr)
```

Perhaps I should remove my last commit and keep that version instead?

gpetrou: @dreddy-work what should we do here?

dreddy-work: @gpetrou, my initial thought is, we should update tests but let me go over these one more time for final suggestion. I will be able to look into this Sometime today or early next week. Will update this thread.

gpetrou: Wouldn't changing some of this tests, be heading towards the wrong direction. Expecting `System.Windows.Forms.Tests.TabControlTests+SubTabPage[]` but getting `System.Windows.Forms.TabPage[]` seems like a valid problem to me.

gpetrou: @dreddy-work any final decision on this?

dreddy-work: @gpetrou, haven't got chance to go over this yet. Will look into it today.

dreddy-work: @gpetrou, you are right. This override method seems specifically designed to return original type back. Please revert your last commit.

dreddy-work (approved):

PR #8573, opened 2023-02-06
Fix System.Windows.Forms.InputLanguage.FromCulture() for languages without LANGID value

DJm00n:
Some Windows input languages do not have a unique pre-assigned LANGID to identify them and instead use a value of `0x0c00` (LOCALE_CUSTOM_DEFAULT) or `0x1000` (LOCALE_CUSTOM_UNSPECIFIED).

Windows tries to provide a unique transient value in these cases in the HKL lowerword, but these values can't be used to identify the keyboard layout, and we can't get the InputLanguage correctly in the FromCulture method.

More info:
https://learn.microsoft.com/windows/win32/winmsg/wm-inputlangchange 
https://learn.microsoft.com/globalization/locale/locale-names#the-deprecation-of-lcids

## Proposed changes

- Compare locale names instead of LANGID to fix these corner cases.

## Customer Impact

- InputLanguage.FromCulture() now should properly work for these locales:

```
[localeName] [lcId]=[langId]:[keyboardId]
jv 0x1000=0c00:00000409
jv-Java 0x1000=0c00:00110c00
jv-Java-ID 0x1000=0c00:00110c00
jv-Latn 0x1000=0c00:00000409
jv-Latn-ID 0x1000=0c00:00000409
mg 0x1000=0c00:0000040c
mg-MG 0x1000=0c00:0000040c
nqo 0x1000=0c00:00090c00
nqo-GN 0x1000=0c00:00090c00
sn 0x1000=0c00:00000409
sn-Latn 0x1000=0c00:00000409
sn-Latn-ZW 0x1000=0c00:00000409
zgh 0x1000=0c00:0000105f
zgh-Tfng 0x1000=0c00:0000105f
zgh-Tfng-Ma 0x1000=0c00:0000105f
```

## Regression? 

- No

## Risk

-

### Before

image

### After

image

## Test methodology 

- Install "N’Ko" input language (or any other from the list above)
- Set it as default input language
- Try to run `InputLanguage_FromCulture_Roundtrip_Success` test.

## Test environment(s)

```
dotnet --info
.NET SDK:
 Version:   8.0.100-alpha.1.23061.8
 Commit:    c8d103ed3c

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.19044
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   G:\project\winforms\.dotnet\sdk\8.0.100-alpha.1.23061.8\

Host:
  Version:      8.0.0-alpha.1.23079.4
  Architecture: x64
  Commit:       a34291586e

.NET SDKs installed:
  8.0.100-alpha.1.22512.5 [G:\project\winforms\.dotnet\sdk]
  8.0.100-alpha.1.22622.3 [G:\project\winforms\.dotnet\sdk]
  8.0.100-alpha.1.23061.8 [G:\project\winforms\.dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 8.0.0-alpha.1.22510.12 [G:\project\winforms\.dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 8.0.0-alpha.1.22615.1 [G:\project\winforms\.dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 8.0.0-alpha.1.23058.7 [G:\project\winforms\.dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 8.0.0-alpha.1.22507.5 [G:\project\winforms\.dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 8.0.0-alpha.1.22605.1 [G:\project\winforms\.dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 8.0.0-alpha.1.22627.5 [G:\project\winforms\.dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 8.0.0-alpha.1.23057.5 [G:\project\winforms\.dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 8.0.0-alpha.1.23058.2 [G:\project\winforms\.dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 8.0.0-alpha.1.23079.4 [G:\project\winforms\.dotnet\shared\Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 8.0.0-alpha.1.22507.3 [G:\project\winforms\.dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 8.0.0-alpha.1.22614.1 [G:\project\winforms\.dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 8.0.0-alpha.1.23057.1 [G:\project\winforms\.dotnet\shared\Microsoft.WindowsDesktop.App]

Other architectures found:
  x86   [C:\Program Files (x86)\dotnet]
    registered at [HKLM\SOFTWARE\dotnet\Setup\InstalledVersions\x86\InstallLocation]

Environment variables:
  DOTNET_ROOT       [G:\project\winforms\.dotnet]

global.json file:
  G:\project\winforms\global.json
```


elachlan:
There are a couple of test failures. Great job in tracking this down!

DJm00n:
It turns out that the LCIDToLocaleName API, which is used inside CultureInfo, can return incorrect names for transient language identifiers. For example, it returns "nqo-GN" and "jv-Java-ID" instead of the language tags "nqo" and "jv-Java" (as seen in the [Get-WinUserLanguageList](https://learn.microsoft.com/powershell/module/international/get-winuserlanguagelist) PowerShell cmdlet). Also in this case, both languages mapped to the same transient language identifier.

So I have to look at the parent language to see if it has the same language identifier and use it in such cases.

DJm00n:
@elachlan can you please help me with this issue:
> ##[error]src\System.Windows.Forms\src\System\Windows\Forms\InputLanguage.cs(227,24): error CA1416: (NETCORE_ENGINEERING_TELEMETRY=Build) This call site is reachable on: 'windows' 6.1 and later. 'PInvoke.Bcp47FromHkl(nint)' is only supported on: 'windows' 8.0 and later.

I have added `if (OsVersion.IsWindows8OrGreater())` check. But it didn't help.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.Bcp47FromHkl.cs
```diff
@@ -0,0 +1,28 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.InteropServices;
+using System.Runtime.Versioning;
+
+namespace Windows.Win32
+{
+    internal static partial class PInvoke
+    {
+        [DllImport("bcp47langs.dll", ExactSpelling = true, SetLastError = true, CharSet = CharSet.Unicode)]
```
elachlan, line 12: `"bcsp47langs.dll"` needs to be added to `Interop.Libraries`.

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
```diff
@@ -135,71 +129,103 @@ public string LayoutName
...
-                if (device != 0)
+                // We need to convert the language identifier to a language tag as soon as possible, because they are
+                // obsolete and may have a transient value - 0x2000, 0x2400, 0x2800, 0x2C00.
+                // https://learn.microsoft.com/globalization/locale/locale-names#the-deprecation-of-lcids
+                //
+                // It turns out that the LCIDToLocaleName API, which is used inside CultureInfo, may return incorrect
+                // language tags for transient language identifiers. For example, it returns "nqo-GN" and "jv-Java-ID"
+                // instead of the "nqo" and "jv-Java" (as seen in the Get-WinUserLanguageList PowerShell cmdlet).
+                // Bcp47FromHkl API returns valid values in these cases.
+                if (!OsVersion.IsWindows8OrGreater())
```
elachlan, line 220: Try flipping the condition to see if that helps with the guarding?

elachlan:
@merriemcgaw I think @JeremyKuhne will need to review this because of the addition of a DLL Import?

Does Microsoft have some documentation around `Bcp47FromHkl`? If its undocumented, then it might not be a great idea to use it.

As far as I can tell `bcp47langs.dll` is a kernel DLL and wouldn't be included in Win32Metadata for inclusion in CsWin32. Although it might make it into [wdkmetadata](https://github.com/microsoft/wdkmetadata), which might mean CsWin32 could include it if it decided to support that as well.

@mikebattista please correct me if I way off track here and it isn't a kernel DLL and just missing from Win32Metadata.

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
elachlan: This might be a bug in `IsWindows8OrGreater` it might be that `[SupportedOSPlatformGuard("windows8")]` should be `[SupportedOSPlatformGuard("windows8.0")]`. Make a change there and it should fix it I think.

DJm00n:
@dotnet-policy-service agree

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
DJm00n: This helped

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.Bcp47FromHkl.cs
DJm00n: done

DJm00n:
@elachlan can you please take a look what’s wrong with arm tests? I don’t see logs and don’t have arm to run these locally.

elachlan:
Thanks for the heads up. We had something similar in other PRs. 

@dmitrii-drobotov there is another ARM test failure in this issue. Is it related to https://github.com/dotnet/winforms/pull/8672#issuecomment-1441888398?

src/System.Windows.Forms.Primitives/src/Interop/Interop.Libraries.cs
```diff
@@ -6,6 +6,7 @@ internal static partial class Interop
 {
     public static class Libraries
     {
+        public const string BCP47Langs = "BCP47Langs.dll";
```
elachlan, line 9: Are the capital letters necessary?

DJm00n: I don’t know. Just looked how file is named on my Windows 11 system…

elachlan (dismissed):
Great work! I imagine the failures are flaky tests. But the team will be able to tell you after reviewing logs.

src/System.Windows.Forms.Primitives/src/Interop/Interop.Libraries.cs
elachlan: Yeah, confirmed with mine as well. Thanks!

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.Bcp47FromHkl.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Runtime.InteropServices;
+using System.Runtime.Versioning;
+
+namespace Windows.Win32
+{
+    internal static partial class PInvoke
+    {
+        [DllImport(Interop.Libraries.BCP47Langs, ExactSpelling = true, SetLastError = true, CharSet = CharSet.Unicode)]
+        [SupportedOSPlatform("windows8.0")]
+        public static extern HRESULT Bcp47FromHkl(IntPtr hkl, out System.WinRT.HSTRING hstring);
```
elachlan, line 14: nit: Maybe add a comment above explaining what Bcp47FromHkl is supposed to do?

RussKie: The main question here is whether we can use this API. Is it public?

DJm00n: The one usage of `bcp47langs.dll` I have found is in [MRT Core](https://github.com/microsoft/WindowsAppSDK/blob/403ed3884ae50176601ac630d5af297fb32c648b/dev/MRTCore/mrt/mrm/mrmmin/Platform.cpp#L1245-L1340) module of Windows App SDK. Comment there giving a clue that it is planned some kind of SDK with corresponding headers.

DJm00n: Turns out that those `GetDistanceOfClosestLanguageInList`/`IsWellFormedTag` methods are documented already: https://learn.microsoft.com/en-us/windows/win32/api/bcp47mrm/

DJm00n: @jwmsft @Karl-Bridge-Microsoft @GrantMeStrength @stevewhims Can you please clarify the possibility of using `Bcp47FromHkl` API from `bcp47langs.dll` in WinForms code? Or maybe there are some alternative API avalible?

RussKie: > Turns out that those `GetDistanceOfClosestLanguageInList`/`IsWellFormedTag` methods that are used in MRT Core are documented already: [learn.microsoft.com/en-us/windows/win32/api/bcp47mrm](https://learn.microsoft.com/en-us/windows/win32/api/bcp47mrm/)

Thank you. I think this is an indication that we can use it :)

DJm00n: @RussKie I'm not so sure. There is no any info on `Bcp47FromHkl`, only other two methods. ((

RussKie: Oh sorry... It's late here, my eyes are tired. I'll see myself out :)

stevewhims: > @jwmsft @Karl-Bridge-Microsoft @GrantMeStrength @stevewhims Can you please clarify the possibility of using `Bcp47FromHkl` API from `bcp47langs.dll` in WinForms code? Or maybe there are some alternative API [available]?

I've forwarded the question on (via internal email) to our colleague who's the writer-owner for MRT Core, and for MRT in general (@hickeys). If he has contacts in the product team, then he might be able to pursue an answer that way. As for my personal feeling about what the answer might be, I would expect that we'd discourage using that API. I could find no hits for the name of the API nor the dll in either the documentation (that my team looks after) nor in the platform header files. Since the API isn't doc'd, it's possible that it'll change or disappear in future. Anyway, like I say, that's just my preliminary feeling; let's see what the official answer is, and any potential documented alternative. -Steve

DJm00n:
I found another way to convert the transient LANGID into a language tag, which does not depend on the undocumented `Bcp47FromHkl` method, but depends on an undocumented registry key:

```cs
int langId = PARAM.LOWORD(_handle);

// NOTE: this logic may break in future versions of Windows since it is not documented.
using RegistryKey? key = Registry.LocalMachine.OpenSubKey(s_userProfileRegistryPath);
if (key is not null)
{
	foreach (string subKeyName in key.GetSubKeyNames())
	{
		using RegistryKey? subKey = key.OpenSubKey(subKeyName);
		if (subKey is null)
		{
			continue;
		}

		if (subKey.GetValue("TransientLangId") is not int transientLangId)
		{
			continue;
		}

		if (transientLangId == langId)
		{
			return subKeyName;
		}
	}
}

return CultureInfo.GetCultureInfo(langId).Name;
```

I have tested this locally and everything looks good.
I can push this version if using `Bcp47FromHkl` is not allowed.

@elachlan @RussKie what do you think on this approach?

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.Bcp47FromHkl.cs
hickeys: I've reached out to the MRT Core team to see if they have any insight based on their use of bcp47langs.dll. If I learn anything useful, I'll update here.

elachlan:
> I found another way to convert the transient LANGID into a language tag (without a bug for transient LANGID language tags), which does not depend on the undocumented `Bcp47FromHkl` method, but depends on an undocumented registry key:
> 
> ```cs
> int langId = PARAM.LOWORD(_handle);
> using RegistryKey? key = Registry.CurrentUser.OpenSubKey(@"Control Panel\International\User Profile");
> if (key is not null && key.GetValue("Languages") is string[] languages)
> {
>     foreach (string language in languages)
>     {
>         using RegistryKey? subKey = key.OpenSubKey(language);
>         if (subKey is not null && subKey.GetValue("TransientLangId") is int transientLangId &&
>             transientLangId == langId)
>         {
>             return language;
>         }
>     }
> }
> 
> return CultureInfo.GetCultureInfo(langId).Name;
> ```
> 
> I have tested this locally and everything looks good. I can push this version if using `Bcp47FromHkl` is not allowed.
> 
> @elachlan @RussKie what do you think about this approach?

Anything undocumented is probably a no-go. Since I believe Microsoft wouldn't consider it a "public API". We will have to wait for the response from someone in Microsoft working on Windows to let us know what is possible. It might just mean `Bcp47FromHkl` gets documented.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.Bcp47FromHkl.cs
DJm00n: Hi @hickeys!
Do you have any updates on the topic?

hickeys: Yes. I heard back from the MRT Core team. They had to go through an internal process to get `GetDistanceOfClosestLanguageInList` and `IsWellFormedTag` approved for use in their libraries. If the WinForms team would like to use this method, we could go through the same process. However, I see that you found another way to convert the langid to a language tag without using this method. 

I'd like someone from the WinForms team to weigh in on their preferred solution. If they would like to get `Bcp47FromHkl` approved, I can help start that process.

hickeys: @dmitrii-drobotov - forgot to tag you in my above comment

DJm00n: @elachlan @RussKie please share your opinion on how we can proceed.

elachlan: I am not in the winforms team so I can't really comment. But I don't mind either method as long as it works and is supported going forward.

RussKie: Ditto here. I'll leave the decision to @dreddy-work and @JeremyKuhne.
The salient point here is that we shouldn't use the non-public API. If you found a public (meaning, documented) Win32 API, and it's providing the expected results - 👍.

JeremyKuhne: @hickeys we do want to get things documented so we don't try to replicate the Windows logic. We'd prefer the existing internal API to be documented. Note that it doesn't have to be in the public headers for our purposes.

@tarekgh I don't know that you've been tagged on this, but you should be aware as this is culture related.

cc: @merriemcgaw @dreddy-work

hickeys: @JeremyKuhne Great. I'll follow up with you via email to get the approval process started.

JeremyKuhne:
@elachlan yes, we cannot use undocumented Windows APIs in the shipping binaries. There are long standing legal reasons, so there isn't a lot of flexibility. We'll try to get things [documented](https://github.com/dotnet/winforms/pull/8573#discussion_r1152322856) so we can close this gap properly.

This sort of thing does come up from time to time, but it is infrequent enough that it is always a little rocky. Thanks for your patience on this everyone.

DJm00n:
Update: I have added proper tests for supplemental languages that doesn't have LCID and so use a transient LCID values when installed in Windows.

src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.Bcp47FromHkl.cs
DJm00n: Hi @hickeys @JeremyKuhne! Has the process of releasing Bcp47FromHkl to the public been started and is there any approximate timeframe for such changes?

dreddy-work:
@DJm00n, can you help rebase this PR? I will checkback on documentation and see if we can merge this for upcoming preview after.

dreddy-work:
> @dreddy-work rebased. If we can't use `Bcp47FromHkl` then [#8573 (comment)](https://github.com/dotnet/winforms/pull/8573#issuecomment-1451934938) is working workaround.

Thank you @DJm00n. We pinged doc team and hoping to hear soon.

DJm00n:
Thank you @hickeys for a foolow up!

@dreddy-work I removed use of Bcp47FromHkl and rebased my changes.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/InputLanguageTests.cs
```diff
@@ -161,12 +189,45 @@ private static void VerifyInputLanguage(InputLanguage language)
...
+        process.StartInfo.FileName = "powershell.exe";
+        process.StartInfo.Arguments = $"-NoProfile -ExecutionPolicy ByPass -File \"{path}\"";
+
+        process.Start();
+        process.WaitForExit();
+    }
+
+    private static void InstallUserLanguage(string languageTag)
+    {
+        var path = Path.Combine(
```
JeremyKuhne, line 205: Use `TempFile` with a `using` to ensure you get this file cleaned up.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/InputLanguageTests.cs
```diff
@@ -161,12 +189,45 @@ private static void VerifyInputLanguage(InputLanguage language)
...
+            Directory.GetCurrentDirectory(),
+            $"install-language-{languageTag}.ps1");
+
+        var sb = new StringBuilder();
+        sb.AppendLine("$list = Get-WinUserLanguageList");
```
JeremyKuhne, line 210: A [raw string literal](https://learn.microsoft.com/dotnet/csharp/programming-guide/strings/#raw-string-literals) is a better choice here.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/InputLanguageTests.cs
```diff
@@ -127,10 +128,28 @@ public static IEnumerable<object[]> InputLanguageLayoutId_TestData()
...
+        yield return new object[] { "zgh-Tfng", "0000105F", "Tifinagh (Basic)" };
+    }
+
+    [Theory]
+    [MemberData(nameof(SupplementalInputLanguages_TestData))]
+    public void InputLanguage_FromCulture_SupplementalInputLanguages_Expected(string languageTag, string layoutId, string layoutName)
+    {
+        InstallUserLanguage(languageTag);
+
+        var culture = new CultureInfo(languageTag);
```
JeremyKuhne, line 148: Prefer `CultureInfo culture = new()`.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/InputLanguageTests.cs
```diff
@@ -127,10 +128,28 @@ public static IEnumerable<object[]> InputLanguageLayoutId_TestData()
...
+        InputLanguage language = InputLanguage.FromCulture(culture);
+        VerifyInputLanguage(language, languageTag, layoutId, layoutName);
+
+        UninstallUserLanguage(languageTag);
```
JeremyKuhne, line 152: This needs to be in a finally block to make sure we put the machine back to its original state. The try block should cover everything after InstallUserLanguage.

I think it is fair to expect that the languages in question here are not going to be installed so I don't think you need to check.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/InputLanguageTests.cs
```diff
@@ -161,12 +189,45 @@ private static void VerifyInputLanguage(InputLanguage language)
...
+        sb.AppendLine($"$list.Add(\"{languageTag}\")");
+        sb.AppendLine("Set-WinUserLanguageList $list -force");
+        File.WriteAllText(path, sb.ToString());
+
+        RunPowerShellScript(path);
+    }
+
+    private static void UninstallUserLanguage(string languageTag)
+    {
+        var path = Path.Combine(
```
JeremyKuhne, line 220: Same comments as the Install method.

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
```diff
@@ -180,6 +172,61 @@ internal string LayoutId
...
+            // https://learn.microsoft.com/globalization/locale/other-locale-names#lcid
+            // https://learn.microsoft.com/windows/win32/winmsg/wm-inputlangchange#remarks
+            //
+            // It turns out that the LCIDToLocaleName API, which is used inside CultureInfo, may return incorrect
+            // language tags for transient language identifiers. For example, it returns "nqo-GN" and "jv-Java-ID"
+            // instead of the "nqo" and "jv-Java" (as seen in the Get-WinUserLanguageList PowerShell cmdlet).
+            //
+            // Try to extract proper language tag from registry as a workaround.
+            //
+            // NOTE: this logic may break in future versions of Windows since it is not documented.
```
JeremyKuhne, line 207: This should have a link to the comment from the Windows team that says that we should use this as a workaround.

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
```diff
@@ -180,6 +172,61 @@ internal string LayoutId
...
+        0x2000, // LOCALE_TRANSIENT_KEYBOARD1
+        0x2400, // LOCALE_TRANSIENT_KEYBOARD2
+        0x2800, // LOCALE_TRANSIENT_KEYBOARD3
+        0x2C00  // LOCALE_TRANSIENT_KEYBOARD4
+    };
+
+    private static string s_userProfileRegistryPath => @"Control Panel\International\User Profile";
+
+    /// <summary>
+    /// Returns the <see href="https://learn.microsoft.com/globalization/locale/standard-locale-names">BCP 47 language
```
JeremyKuhne, line 186: We indent xml blocks by a space.

JeremyKuhne (changes requested):
Added some comments. Thanks for the patience here.

DJm00n:
> Added some comments. Thanks for the patience here.

updated PR

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/InputLanguageTests.cs
```diff
@@ -126,19 +126,44 @@ public static IEnumerable<object[]> InputLanguageLayoutId_TestData()
...
+        yield return new object[] { "nqo", "00090C00", "N’Ko" };
+        yield return new object[] { "zgh-Tfng", "0000105F", "Tifinagh (Basic)" };
+    }
+
+    [Theory]
+    [MemberData(nameof(SupplementalInputLanguages_TestData))]
+    public void InputLanguage_FromCulture_SupplementalInputLanguages_Expected(string languageTag, string layoutId, string layoutName)
+    {
+        // Also installs default keyboard layout for this language
+        // https://learn.microsoft.com/windows-hardware/manufacture/desktop/windows-language-pack-default-values
```
DJm00n, line 146: oops. proper link:
https://learn.microsoft.com/windows-hardware/manufacture/desktop/default-input-locales-for-windows-language-packs

DJm00n: fixed

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
DJm00n: done

DJm00n:
@JeremyKuhne @dreddy-work any updates on this?

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
```diff
@@ -180,6 +172,62 @@ internal string LayoutId
         }
     }
 
+    private static readonly int[] s_transientLangIds =
+    {
+        0x2000, // LOCALE_TRANSIENT_KEYBOARD1
```
JeremyKuhne, line 177: These should be added to the NativeMethods.txt file in System.Windows.Forms.Primitives.

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
```diff
@@ -180,6 +172,62 @@ internal string LayoutId
...
+        0x2400, // LOCALE_TRANSIENT_KEYBOARD2
+        0x2800, // LOCALE_TRANSIENT_KEYBOARD3
+        0x2C00  // LOCALE_TRANSIENT_KEYBOARD4
+    };
+
+    private static string s_userProfileRegistryPath => @"Control Panel\International\User Profile";
```
JeremyKuhne, line 183: This should be a constant, not a static.

src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
```diff
@@ -213,13 +261,9 @@ public override bool Equals(object? value)
...
         ArgumentNullException.ThrowIfNull(culture);
 
-        // KeyboardLayoutId is the LCID for built-in cultures, but it is the CU-preferred keyboard language for
-        // custom cultures.
-        int lcid = culture.KeyboardLayoutId;
-
         foreach (InputLanguage? lang in InstalledInputLanguages)
         {
-            if ((unchecked((int)(long)lang!._handle) & 0xFFFF) == lcid)
+            if (lang!.Culture.Equals(culture))
```
JeremyKuhne, line 266: Don't bang (!) this. One option is to flip it: `culture.Equals(lang?.Culture)`

JeremyKuhne (changes requested):
Sorry about the delays in responses, been very busy. Just a few more comments. Please tag me when you've responded.

DJm00n:
@JeremyKuhne please take another look at this. I've made corrections.

JeremyKuhne (dismissed):
Thanks for your patience!

DJm00n:
@JeremyKuhne any updates?

JeremyKuhne (approved):

Olina-Zhang:
Verified it in the Release/8.0-Preview6 branch from this repo, result of running InputLanguage_FromCulture_Roundtrip_Success test is failed as following. Is there some wrong from my setting? @DJm00n 
image

DJm00n:
Hi @Olina-Zhang!
Can you please run `Get-WinUserLanguageList` in PowerShell and provide output on your system?

Olina-Zhang:
@DJm00n Here is Get-WinUserLanguageList PowerShell cmdlet output:
image

DJm00n:
@Olina-Zhang interesting...

- `0xf0402000` is `0x040` which is "Layout Id" under `HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Keyboard Layouts\00090C00` - which looks **okay**
- `0xf03e2000` is `0x03e` layoutId which is "Layout Id" under `HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Keyboard Layouts\00070C00` -  which looks **okay**

`0x2000` (LOCALE_TRANSIENT_KEYBOARD1) in both cases means that these layouts are installed under one input language.. but Get-WinUserLanguageList shows only one of them.

This looks like some kind of configuration issue on your PC.

If it is an RDP connection then it could be related too (since it can install keyboard layouts from the remote system):
https://superuser.com/questions/483086/how-do-i-remove-automatically-added-keyboard-layouts

You can also check these registry paths for this suspicious "00070C00" keyboard layout:
```
HKEY_USERS\.DEFAULT\Control Panel\International\User Profile
HKEY_USERS\.DEFAULT\Keyboard Layout\Preload
HKEY_USERS\.DEFAULT\Keyboard Layout\Substitutes
HKEY_USERS\S-1-5-18\Control Panel\International\User Profile
HKEY_USERS\S-1-5-18\Keyboard Layout\Preload
HKEY_USERS\S-1-5-18\Keyboard Layout\Substitutes
HKEY_CURRENT_USER\Control Panel\International\User Profile
HKEY_CURRENT_USER\Keyboard Layout\Preload
HKEY_CURRENT_USER\Keyboard Layout\Substitutes
```

Olina-Zhang:
@DJm00n  Thanks for your comment! The testing machine I used is connected by remoting. It may be related some machine setting. Today re-tested that test case:  `InputLanguage_FromCulture_Roundtrip_Success`, it passed. And checked machine with more input languages automatically.
Input languages:
image
Case passed:
image
image

PR #8595, opened 2023-02-08
Enable nullability in FeedbackRectangle

gpetrou:
## Proposed changes

- Enable nullability in FeedbackRectangle

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8595)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripPanel.FeedbackRectangle.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripPanel.FeedbackRectangle.cs
@@ -13,7 +13,7 @@ namespace System.Windows.Forms
         // API so the underlying implementation can be replaced if necessary.
         private partial class FeedbackRectangle : IDisposable
         {
-            private FeedbackDropDown _dropDown;
+            private FeedbackDropDown? _dropDown;
 
             public FeedbackRectangle(Rectangle bounds)
             {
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ToolStripPanel.FeedbackRectangle.cs
```diff
@@ -60,7 +58,7 @@ protected void Dispose(bool disposing)
                     {
                         Visible = false;
                         _dropDown.Dispose();
-                        _dropDown = null;
+                        _dropDown = null!;
```
Tanya-Solyanik, line 61: Perhaps declare this field as nullable, as it's tested for null in `Visible` property.

Tanya-Solyanik: You are right, it should not happen. I guess the reason these checks were added in the past was that someone was accessing the Visible property after the disposal and instead of fixing the caller, we changed the framework to be defensive instead of failing fast. Perhaps this was a Watson report. However, we don't know what the scenario was and if it's still relevant. So the compatible thing to do is to not risk an exception in that property.

gpetrou: I added a separate commit that makes `_dropDown` nullable. Please note that this means adding null checks in a couple of places.

Tanya-Solyanik (approved):

PR #8597, opened 2023-02-08
Fix Control COM interface serialization

JeremyKuhne:
`PropertyBagStream` wasn't converting it's values correctly to/from string/object.

Also reinstate `BinaryFormatter` property handling which I removed under the misunderstanding that `BinaryFormatter` would be completely removed.

Add tests.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8597)

### DIFF
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Control.ActiveXImplTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Control.ActiveXImplTests.cs
@@ -46,7 +46,7 @@ public unsafe class Control_ActiveXImplTests : IClassFixture<ThreadExceptionFixt
         using var istream = ComHelpers.GetComScope<IStream>(new GPStream(memoryStream));
         HRESULT hr = persistStream.Save(istream.Value, fClearDirty: BOOL.FALSE);
         Assert.True(hr.Succeeded);
-        control.DataContext = null;
+        control.SerializableValue = default;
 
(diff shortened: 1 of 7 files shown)


src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Control.ActiveXImplTests.cs
```diff
@@ -0,0 +1,77 @@
...
+        // BinaryFormatter code path.
+        SerializableStruct myValue = new() { Value = "HelloThere" };
+        control.SerializableValue = myValue;
+        IPersistStreamInit.Interface persistStream = control;
+
+        using MemoryStream memoryStream = new();
+        using var istream = ComHelpers.GetComScope<IStream>(new GPStream(memoryStream));
+        HRESULT hr = persistStream.Save(istream.Value, fClearDirty: BOOL.FALSE);
+        Assert.True(hr.Succeeded);
+        control.DataContext = null;
```
lonitra, line 49: why do we set this to null?

JeremyKuhne: Whoops, that was supposed to be `.SerializableValue`. I started with `DataContext`, but it isn't really usable for this.

lonitra (approved):
🚀

PR #8618, opened 2023-02-10
Prevent disposing Icons that are not owned by the caller.

### LINKED ISSUE #8513: Automation case about ErrorProvider failed with "Cannot access a disposed object"
### .NET version

8.0.100-alpha.1.23068.16

Yes

Yes, cannot repro in previous versions

### Issue description

image, related causing PR: https://github.com/dotnet/winforms/pull/8486
_System.ObjectDisposedException
  HResult=0x80131622
  Message=Cannot access a disposed object.
  Source=System.Drawing.Common
  StackTrace:
   at System.Drawing.Icon.get_Handle()
   at System.Drawing.Icon.get_Size()
   at System.Windows.Forms.ErrorProvider.IconRegion.get_Size()
   at System.Windows.Forms.ErrorProvider.ErrorWindow.Update(Boolean timerCaused)
   at System.Windows.Forms.ErrorProvider.ErrorWindow.Add(ControlItem item)
   at System.Windows.Forms.ErrorProvider.ControlItem.AddToWindow()
   at System.Windows.Forms.ErrorProvider.ControlItem.set_Error(String value)
   at System.Windows.Forms.ErrorProvider.SetError(Control control, String value)
   at XErrorProvider.set_RightToLeft(TParams p, Boolean rtl) in C:\Users\xxx\Desktop\XErrorProvider_P1\XErrorProvider_P1\XErrorProvider.cs:line 455
   at System.Reflection.MethodInvoker.Invoke(Object obj, IntPtr* args, BindingFlags invokeAttr)_

### Steps to reproduce

It reproduces steadily in our automation case about errorProvider, but we cannot repro manually in TextBox and ErrorProvider, then errorProvider.setError

dreddy-work:
We regressed this issue here: https://github.com/dotnet/winforms/pull/8486.

by accidentally disposing Icons that may not be owned or created by us. In this scenario, It was default Icon that is getting disposed.

Added unit test.

Fixes #8513.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8618)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ErrorProvider.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ErrorProvider.cs
@@ -583,7 +583,6 @@ namespace System.Windows.Forms
             }
             set
             {
-                _icon.Dispose();
                 _icon = value.OrThrowIfNull();
                 DisposeRegion();
                 ErrorWindow[] array = _windows.Values.ToArray();
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ErrorProviderTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ErrorProviderTests.cs
@@ -1274,7 +1274,7 @@ namespace System.Windows.Forms.Tests
 
             Assert.NotNull(provider.Icon);
 
-            // Make sure old icon that is not owned by us, is not disposed
+            // Make sure the old icon that is not owned by us, is not disposed
             Assert.Equal(handle, icon.Handle);
         }
 
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ErrorProviderTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ErrorProviderTests.cs
@@ -1259,6 +1259,20 @@ namespace System.Windows.Forms.Tests
             Assert.Null(exception);
         }
 
+        [WinFormsFact]
+        public void ErrorProvider_Icon_NotDisposed_Unexpectedly()
+        {
+            // Unit test for https://github.com/dotnet/winforms/issues/8513.
+            using var provider = new SubErrorProvider();
+            var Icon = provider.Icon;
+            Assert.NotNull(Icon);
+            nint handle = Icon.Handle;
+
+            provider.Icon = new Icon(typeof(ErrorProvider), "Error");
+            Assert.NotNull(provider.Icon);
+            Assert.Equal(handle, Icon.Handle);
+        }
+
         private class CustomDataSource : IDataErrorInfo
         {
             public string this[string columnName] => string.Empty;
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ErrorProviderTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ErrorProviderTests.cs
@@ -1264,13 +1264,18 @@ namespace System.Windows.Forms.Tests
         {
             // Unit test for https://github.com/dotnet/winforms/issues/8513.
             using var provider = new SubErrorProvider();
-            var Icon = provider.Icon;
-            Assert.NotNull(Icon);
-            nint handle = Icon.Handle;
+            var icon = provider.Icon;
 
-            provider.Icon = new Icon(typeof(ErrorProvider), "Error");
+            Assert.NotNull(icon);
+
+            nint handle = icon.Handle;
+            using Icon newIcon = new (typeof(ErrorProvider), "Error");
+            provider.Icon = newIcon;
+
             Assert.NotNull(provider.Icon);
-            Assert.Equal(handle, Icon.Handle);
+
+            // Make sure old icon that is not owned by us, is not disposed
+            Assert.Equal(handle, icon.Handle);
         }
 
         private class CustomDataSource : IDataErrorInfo
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ErrorProviderTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ErrorProviderTests.cs
@@ -1263,13 +1263,13 @@ namespace System.Windows.Forms.Tests
         public void ErrorProvider_Icon_NotDisposed_Unexpectedly()
         {
             // Unit test for https://github.com/dotnet/winforms/issues/8513.
-            using var provider = new SubErrorProvider();
+            using var provider = new ErrorProvider();
             var icon = provider.Icon;
 
             Assert.NotNull(icon);
 
             nint handle = icon.Handle;
-            using Icon newIcon = new (typeof(ErrorProvider), "Error");
+            using Icon newIcon = new("bitmaps/10x16_one_entry_32bit.ico");
             provider.Icon = newIcon;
 
             Assert.NotNull(provider.Icon);


dreddy-work:
@Olina-Zhang, can you please run some scenarios here apart from the repro app to make sure we are not regressing anything else?

@elachlan as FYI.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ErrorProviderTests.cs
```diff
@@ -1259,6 +1259,20 @@ public void ErrorProvider_CustomDataSource_DoesNotThrowInvalidCastException()
...
+        {
+            // Unit test for https://github.com/dotnet/winforms/issues/8513.
+            using var provider = new SubErrorProvider();
+            var Icon = provider.Icon;
+            Assert.NotNull(Icon);
+            nint handle = Icon.Handle;
+
+            provider.Icon = new Icon(typeof(ErrorProvider), "Error");
+            Assert.NotNull(provider.Icon);
+            Assert.Equal(handle, Icon.Handle);
```
Tanya-Solyanik, line 1273: Had this failed before your fix?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ErrorProviderTests.cs
```diff
@@ -1259,6 +1259,20 @@ public void ErrorProvider_CustomDataSource_DoesNotThrowInvalidCastException()
             Assert.Null(exception);
         }
 
+        [WinFormsFact]
+        public void ErrorProvider_Icon_NotDisposed_Unexpectedly()
+        {
+            // Unit test for https://github.com/dotnet/winforms/issues/8513.
+            using var provider = new SubErrorProvider();
+            var Icon = provider.Icon;
```
JeremyKuhne, line 1267: This is hard to read with the casing wrong.

```suggestion
            var icon = provider.Icon;
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ErrorProviderTests.cs
```diff
@@ -1259,6 +1259,20 @@ public void ErrorProvider_CustomDataSource_DoesNotThrowInvalidCastException()
...
+            Assert.NotNull(Icon);
+            nint handle = Icon.Handle;
+
+            provider.Icon = new Icon(typeof(ErrorProvider), "Error");
```
JeremyKuhne, line 1271: The new Icon should be disposed when you're done with it.

JeremyKuhne (changes requested):
A few minor things in the test.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ErrorProviderTests.cs
lonitra, line 1267: Do we need using statement here? Doesn't seem like the old icon is getting disposed at end of test.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ErrorProviderTests.cs
dreddy-work: Yes. Accessing Icon.Handle throw `ObjectDisposedException`

src/System.Windows.Forms/src/System/Windows/Forms/ErrorProvider.cs
```diff
@@ -583,7 +583,6 @@ public Icon Icon
             }
             set
             {
-                _icon.Dispose();
```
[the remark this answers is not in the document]
dreddy-work: `_icon` is assigned with `DefaultIcon` initially which is owned by WinForms. Subsequent changes to public property  `Icon` should be owned by caller (user).

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ErrorProviderTests.cs
dreddy-work: Not on this one but the one we are creating in this test. This one is default Icon.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ErrorProviderTests.cs
```diff
@@ -1259,6 +1259,20 @@ public void ErrorProvider_CustomDataSource_DoesNotThrowInvalidCastException()
             Assert.Null(exception);
         }
 
+        [WinFormsFact]
+        public void ErrorProvider_Icon_NotDisposed_Unexpectedly()
+        {
+            // Unit test for https://github.com/dotnet/winforms/issues/8513.
+            using var provider = new SubErrorProvider();
```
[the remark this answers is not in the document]
dreddy-work: are you referring to `SubErrorProvider`? I will have to check what all it does to come up with right name but that could be outside of this PR.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ErrorProviderTests.cs
Tanya-Solyanik: So this test is not about equality, but about  the Handle property being set to something.

dreddy-work: Yes. Checking Handle to `null` or `zero` is ideal but Handle accessor on Icon throws exception if it was `null` or `zero`. So, I went this route.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ErrorProviderTests.cs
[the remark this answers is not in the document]
JeremyKuhne: If you don't need something specific from `SubErrorProvider` you should just use `ErrorProvider`. That makes intent and dependencies much clearer.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ErrorProviderTests.cs
```diff
@@ -1259,6 +1259,25 @@ public void ErrorProvider_CustomDataSource_DoesNotThrowInvalidCastException()
...
+            var icon = provider.Icon;
+
+            Assert.NotNull(icon);
+
+            nint handle = icon.Handle;
+            using Icon newIcon = new (typeof(ErrorProvider), "Error");
```
Tanya-Solyanik, line 1272: This constructor is accessing a winforms resource that might be accessed at the same time (this assembly runs tests in parallel) by another test. Is it possible to create an in-memory icon?

Tanya-Solyanik: Sorry, I meant something like this:
```
            using Bitmap bitmap = new(16, 16);
            using Graphics graphics = Graphics.FromImage(bitmap);
            Rectangle rectangle = new(new Point(1, 1), new Size(14, 14));
            using SolidBrush brush = new(Color.Blue);
            graphics.FillRectangle(brush, rectangle);
            Icon icon = Icon.FromHandle(bitmap.GetHicon());
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ErrorProviderTests.cs
```diff
@@ -1259,6 +1259,25 @@ public void ErrorProvider_CustomDataSource_DoesNotThrowInvalidCastException()
...
+
+            Assert.NotNull(icon);
+
+            nint handle = icon.Handle;
+            using Icon newIcon = new("bitmaps/10x16_one_entry_32bit.ico");
+            provider.Icon = newIcon;
+
+            Assert.NotNull(provider.Icon);
+
+            // Make sure old icon that is not owned by us, is not disposed
```
KlausLoeffelmann, line 1277:
```suggestion
            // Make sure the old icon that is not owned by us, is not disposed
```

Olina-Zhang:
@dreddy-work, Tested this PR for GH issue: [8513](https://github.com/dotnet/winforms/issues/8513), it is fixed: the automation case about ErrorProvider passed now, and tested following scenarios manually, no new issue found.
image

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ErrorProviderTests.cs
dreddy-work: That could work but I am using one of the existing Icons in this case. For reading is `Open|Shared` so shouldn't be a problem. Leving as-is.

Tanya-Solyanik: >For reading is Open|Shared so shouldn't be a problem.

What do you mean by open or shared?

dreddy-work: Ico file used in this test is open in those modes for reading. So, even if other test running in parallel and accessing, it shouldn't be an issue.

Tanya-Solyanik: Could you please share the code pointer? Is it really loaded from the file each time the icon is accessed?

dreddy-work: Here is the code that read Icon from file.  It is read only when the Icon is created. https://github.com/dotnet/runtime/blob/52fea7fb9fdc4a80ae5f1646b8461bb6545ab563/src/libraries/System.Drawing.Common/src/System/Drawing/Icon.Windows.cs#L65

Tanya-Solyanik: I think you are using this constructor - https://github.com/dotnet/runtime/blob/52fea7fb9fdc4a80ae5f1646b8461bb6545ab563/src/libraries/System.Drawing.Common/src/System/Drawing/Icon.Windows.cs#L99

We don't ship icons as stand alone files, we embed them into the winforms assembly. I don't see in docs if GetManifestResource supports concurrent read.

dreddy-work: I am confused. Are you referring to latest change? I am using following file to read Icon from.

"bitmaps/10x16_one_entry_32bit.ico"

Tanya-Solyanik: Ah, you are right!

Tanya-Solyanik (approved):

PR #8628, opened 2023-02-12
Enable nullability in PropertyTabCollection

gpetrou:
## Proposed changes

- Enable nullability in PropertyTabCollection

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8628)


dmitrii-drobotov (commented):
Perhaps we can add the ArgumentNullException (via `OrThrowIfNull()`) in this PR?
https://github.com/dotnet/winforms/blob/927e0272797b7afe1d9800834d84059d4c803c85/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGrid.PropertyTabCollection.cs#L19-L21

dreddy-work:
> Perhaps we can add the ArgumentNullException (via `OrThrowIfNull()`) in this PR?

Agree and add "Documentation" label.

dreddy-work (approved):

gpetrou:
> Agree and add "Documentation" label.

This is `internal` constructor, so I assume the documentation will not be affected.

Issue #1107, opened 2019-06-07
Enable nullability across winforms


hughbe:
Once CoreFx and CoreCLR fully support nullability it would be great to do this in winforms

Probably a large project...

merriemcgaw:
Got a few things on the priority list, but we can consider adding this to the roadmap. @OliaG for Roadmap thoughts.

Logerfo:
Implementing NRT to the whole code base would be a massive work, but enabling it only for the API surface would be much easier and yet very helpful for customers who uses both winforms and NRT.

drewnoakes:
Something we did on project system, CPS and other projects is to add `<Nullable>enable</Nullable>` to project files, then use a regular expression to add `#nullable disable` to each and every source file.

This encourages new code to be written using NRT, and provides an annotation to-do list by searching for `#nullable disable`.

For commonly used types where annotations would be useful, but it's too challenging to fix all generated warnings, it's possible to use `#nullable annotations` which allows adding `?` to types without having to address any warnings.

RussKie:
Thank you @drewnoakes, this may be a way forward.

Logerfo:
```c#
#nullable annotations
```
@drewnoakes where is this documented? VS syntax highlight it, but the compiler is bothered:
> CS8637 Expected 'enable', 'disable', or 'restore'

sharwell:
I've done this for a few projects to date, and may be able to help here.

Logerfo:
Now that #2702 and #2756 are merged, I encourage everyone interested to send PRs covering small parts of the API for NRT in SWF.

RussKie:
When annotating Windows Forms types verify that the Runtime types are NRT enabled (refer to https://github.com/dotnet/runtime/issues/2339).

elachlan:
List of files in `System.Windows.Forms` which need annotation, ordered by line count ascending. 
- 157 total files
- 35 files > 1000 LOC
- 92 files < 300 LOC
- 50 files < 100 LOC
```
D:\Development\winforms\src\System.Windows.Forms\src\System\Windows\Forms\IDataGridColumnStyleEditingNotificationService.cs	11
D:\Development\winforms\src\System.Windows.Forms\src\System\Windows\Forms\IFileReaderService.cs	11
D:\Development\winforms\src\System.Windows.Forms\src\System\Drawing\Design\PropertyValueUIHandler.cs	14
[...]

PR #8639, opened 2023-02-13
Null annotate a couple of small line count files - part 2

elachlan:
Related: #1107

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8639)

### DIFF
--- a/src/Common/src/IComparerHelpers.cs
+++ b/src/Common/src/IComparerHelpers.cs
@@ -1,0 +1,40 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Diagnostics.CodeAnalysis;
+
+namespace System
+{
+    internal static class IComparerHelpers
+    {
+        public static bool CompareReturnIfNull<T>([NotNullWhen(false)] T x, [NotNullWhen(false)] T y, [NotNullWhen(true)] out int? compareReturnValue)
+        {
+            compareReturnValue = null;
+            if (x is not null && y is not null)
+            {
+                return false;
+            }
+
+            if (x is null && y is null)
+            {
+                compareReturnValue = 0;
+                return true;
+            }
+
+            if (x is null)
+            {
+                compareReturnValue = -1;
+                return true;
+            }
+
+            if (y is null)
+            {
+                compareReturnValue = 1;
+                return true;
+            }
+
+            return false;
+        }
+    }
+}
--- a/src/System.Windows.Forms/src/System/Windows/Forms/BaseCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/BaseCollection.cs
@@ -17,14 +17,14 @@ namespace System.Windows.Forms
         /// </summary>
         [Browsable(false)]
         [EditorBrowsable(EditorBrowsableState.Advanced)]
-        public virtual int Count => List.Count;
+        public virtual int Count => List!.Count;
 
-        public void CopyTo(Array ar, int index) => List.CopyTo(ar, index);
+        public void CopyTo(Array ar, int index) => List!.CopyTo(ar, index);
 
         /// <summary>
         ///  Gets an IEnumerator for the collection.
         /// </summary>
-        public IEnumerator GetEnumerator() => List.GetEnumerator();
+        public IEnumerator GetEnumerator() => List!.GetEnumerator();
 
         [Browsable(false)]
         [EditorBrowsable(EditorBrowsableState.Advanced)]
@@ -38,6 +38,6 @@ namespace System.Windows.Forms
         [EditorBrowsable(EditorBrowsableState.Advanced)]
         public object SyncRoot => this;
 
-        protected virtual ArrayList List => new();
+        protected virtual ArrayList? List => null;
     }
 }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/BaseCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/BaseCollection.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.Collections;
 using System.ComponentModel;
 
@@ -40,6 +38,6 @@ namespace System.Windows.Forms
         [EditorBrowsable(EditorBrowsableState.Advanced)]
         public object SyncRoot => this;
 
-        protected virtual ArrayList List => null;
+        protected virtual ArrayList List => new();
     }
 }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.ToolTip.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.ToolTip.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 namespace System.Windows.Forms
 {
     public partial class DataGridView
@@ -19,7 +17,7 @@ namespace System.Windows.Forms
 
             public bool Activated { get; private set; }
 
-            public ToolTip ToolTip { get; private set; }
+            public ToolTip? ToolTip { get; private set; }
 
             public void Activate(bool activate)
             {
@@ -43,7 +41,7 @@ namespace System.Windows.Forms
 
                 if (activate)
                 {
-                    ToolTip.Active = true;
+                    ToolTip!.Active = true;
                     ToolTip.Show(_dataGridView.ToolTipPrivate, _dataGridView);
                 }
                 else if (ToolTip is not null)
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.ToolTip.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.ToolTip.cs
@@ -26,10 +26,10 @@ namespace System.Windows.Forms
                     return;
                 }
 
-                // Create the tool tip handle on demand.
-                if (activate && ToolTip is null)
+                if (activate)
                 {
-                    ToolTip = new ToolTip
+                    // Create the tool tip handle on demand.
+                    ToolTip ??= new ToolTip
                     {
                         ShowAlways = true,
                         InitialDelay = 0,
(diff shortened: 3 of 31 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/BaseCollection.cs
```diff
@@ -40,6 +38,6 @@ public class BaseCollection : MarshalByRefObject, ICollection
         [EditorBrowsable(EditorBrowsableState.Advanced)]
         public object SyncRoot => this;
 
-        protected virtual ArrayList List => null;
+        protected virtual ArrayList List => new();
```
dreddy-work, line 41: hmm. Wouldn't this break existing apps? why not make it nullable and override non-nullable as appropriate?

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.ToolTip.cs
```diff
@@ -43,7 +41,7 @@ public void Activate(bool activate)
 
                 if (activate)
                 {
-                    ToolTip.Active = true;
+                    ToolTip!.Active = true;
```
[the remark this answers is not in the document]
elachlan: Yes. Its strange. I am fairly sure it should not be null at that point.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripContainer.ToolStripContainerTypedControlCollection.cs
```diff
@@ -20,7 +20,7 @@ public ToolStripContainerTypedControlCollection(ToolStripContainer c, bool isRea
                 _owner = c;
             }
 
-            public override void Add(Control value)
+            public override void Add(Control? value)
```
gpetrou, line 23: This seems bizarre since we have `ArgumentNullException` below.

src/System.Windows.Forms/src/System/Windows/Forms/TypedControlCollection.cs
```diff
@@ -29,7 +27,7 @@ public TypedControlCollection(Control owner, Type typeOfControl) : base(owner, /
             _ownerControl = owner;
         }
 
-        public override void Add(Control value)
+        public override void Add(Control? value)
```
gpetrou, line 30: Same inconsistency with the `ArgumentNullException` below.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.ToolTip.cs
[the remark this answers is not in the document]
halgab: Wouldn't it be cleaner to change the assignment of `ToolTip` to a null-conditional assignment and merge this `if` clause with the one above? I guess you wouldn't need the `!` anymore?

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripPanel.ToolStripPanelControlCollection.YXComparer.cs
```diff
@@ -17,10 +15,25 @@ public class YXComparer : IComparer<IArrangedElement>
             {
                 public YXComparer() { }
 
-                public int Compare(IArrangedElement first, IArrangedElement second)
+                public int Compare(IArrangedElement? first, IArrangedElement? second)
                 {
-                    Control one = first as Control;
-                    Control two = second as Control;
+                    Control? one = first as Control;
```
dreddy-work, line 20: can you check if this method can be moved to a common place and reuse?

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripContainer.ToolStripContainerTypedControlCollection.cs
elachlan: Yes. I also thought so. but can't think of a way around it because of the inheritance structure. The base class needs the nullability as far as I can tell.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripPanel.ToolStripPanelControlCollection.YXComparer.cs
elachlan: Is it used in multiple places? `XYComparer` and `YXComparer` are slightly different.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripContainer.ToolStripContainerTypedControlCollection.cs
gpetrou: I see that `Control` just checks for null and returns while here and in other places we throw an exception. So, perhaps it is better to change the base method declaration instead.

elachlan: Do you mean `Control.ControlCollection`?
https://github.com/dotnet/winforms/blob/d46ad2e8dc76248739d9ae22b28b399a6a6b299e/src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlCollection.cs#L48-L53

gpetrou: Yes, I should have been more specific. You might want to wait on feedback from a maintainer though if they want this change. These kind of cases are a gray area when it comes to nullability annotations.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.ToolTip.cs
```diff
@@ -41,12 +39,13 @@ public void Activate(bool activate)
...
 
-                if (activate)
+                if (activate && ToolTip is not null)
                 {
                     ToolTip.Active = true;
                     ToolTip.Show(_dataGridView.ToolTipPrivate, _dataGridView);
                 }
-                else if (ToolTip is not null)
+
+                if (ToolTip is not null)
```
halgab, line 48: I was more thinking about deleting the if clause on lines 32 to 39, and replace this one with
```suggestion
                if (activate)
                {
                    // Create the tool tip handle on demand.
                    ToolTip ??= new ToolTip
                    {
                        ShowAlways = true,
                        InitialDelay = 0,
                        UseFading = false,
                        UseAnimation = false,
                        AutoPopDelay = 0
                    };
                    
                    ToolTip.Active = true;
                    ToolTip.Show(_dataGridView.ToolTipPrivate, _dataGridView);
                }
                else if (ToolTip is not null)
```

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripPanel.ToolStripPanelControlCollection.YXComparer.cs
dreddy-work: right.  but this common new added code seems same. I won't insist but this checks around null are common across many compare functions we have in runtime. I won't insist in this PR but keep in mind with goal of moving to a common place and reuse.

elachlan: Something like below? Did you want it in Common so it can be included in both Forms and Design?
```csharp
internal static bool CompareReturnIfNull<T>(T x, T y, out int? CompareReturnValue)
{
    CompareReturnValue = null;
    if (x is not null && y is not null)
    {
        return false;
    }

    if (x is null && y is null)
    {
        CompareReturnValue = 0;
        return true;
    }

    if (x is null)
    {
        CompareReturnValue = -1;
        return true;
    }

    if (y is null)
    {
        CompareReturnValue = 1;
        return true;
    }

    return false;
}
```

dreddy-work: yep.

elachlan: Done.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripContainer.ToolStripContainerTypedControlCollection.cs
elachlan: @dreddy-work can you let me know what the team would prefer? Control.ControlCollection is public so its an API change maybe?

dreddy-work (approved):

src/Common/src/IComparerHelpers.cs
```diff
@@ -0,0 +1,40 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Diagnostics.CodeAnalysis;
+
+namespace System
+{
+    internal static class IComparerHelpers
+    {
+        public static bool CompareReturnIfNull<T>([NotNullWhen(false)] T x, [NotNullWhen(false)] T y, [NotNullWhen(true)] out int? compareReturnValue)
```
halgab, line 11: Don't you think that the fact that `compareReturnValue` is nullable is redundant with the fact that this method returns true? I guess you could either say that the contract of this method is that `compareReturnValue` shouldn't been use if the return value is false, a la `dictionary.TryGetValue`, or return a nullable int directly, right?

elachlan: I am not totally following you. I think either way you have to annotate it so the compiler knows the values aren't null later on.

halgab: Whenever this method returns true, then `compareReturnValue` isn't null. I just think that its doesn't need to be an `int?` and could just be an `int`. Feel free to ignore if you disagree.

elachlan: The issue with that is 0 is a valid value for compare. So that is why I went for a null value. There isn't any way you can get it mixed up.

src/System.Windows.Forms/src/System/Windows/Forms/TypedControlCollection.cs
dreddy-work: Given base is defined as nullable but not derived, taking nullability here is a compile warning. Modifying base is a breaking change.  @elachlan. lets keep track of this and see if there are any other overrides that follow similar suite and make decision on breaking change.

src/Common/src/IComparerHelpers.cs
halgab: That's true, but you could argue the same thing happen when you call `TryGetValue(key, out int value)` on a `Dictionary<T, int>`

halgab: An alternative would be to have the signature of the method be
```suggestion
        public static int? CompareReturnIfNull<T>([NotNullWhen(false)] T x, [NotNullWhen(false)] T y)
```
as you current implementation effectively returns `compareReturnValue is null`.

PR #8645, opened 2023-02-15
Enable nullability in some ToolTip members

gpetrou:
## Proposed changes

- Enable nullability in some ToolTip members.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8645)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolTip.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolTip.cs
@@ -2,9 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.ComponentModel;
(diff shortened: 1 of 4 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ToolTip.cs
```diff
@@ -933,7 +938,7 @@ private void DestroyRegion(Control control)
                 control.RemoveToolTip(this);
             }
         }
-
+#nullable disable
```
dreddy-work, line 941: why this section specific?

gpetrou: No specific reason. This file has lots of lines, so I had to stop somewhere to make it easier to review. We have followed similar approach in other files.

dreddy-work (approved):

Issue #8598, opened 2023-02-09
DragDrop tests are resulting in dotnet.exe process crash


dreddy-work:
### Test

PictureBox_SetData_DoDragDrop_RichTextBox_ReturnsExptected_Async and possibly others in this file.

### Issue description

Although these crashes are not frequent, they happen often enough to impede the CI build process. As these crashes occur in dotnet.exe, our tests are unable to collect dumps or provide detailed logs or call stacks.

I manually setup lab machine and locally debugged to collect dumps. Below are the call stack: 

```CS
 	System.Private.CoreLib.dll!System.ModuleHandle.ResolveMethodHandle(int methodToken, System.RuntimeTypeHandle[] typeInstantiationContext, System.RuntimeTypeHandle[] methodInstantiationContext)	Unknown
 	System.Private.CoreLib.dll!System.Reflection.CustomAttribute.FilterCustomAttributeRecord(System.Reflection.MetadataToken caCtorToken, System.Reflection.MetadataImport scope, System.Reflection.RuntimeModule decoratedModule, System.Reflection.MetadataToken decoratedToken, System.RuntimeType attributeFilterType, bool mustBeInheritable, ref System.RuntimeType.ListBuilder<object> derivedAttributes, out System.RuntimeType attributeType, out System.IRuntimeMethodInfo ctorWithParameters, out bool isVarArg)	Unknown
 	System.Private.CoreLib.dll!System.Reflection.CustomAttribute.AddCustomAttributes(ref System.RuntimeType.ListBuilder<object> attributes, System.Reflection.RuntimeModule decoratedModule, int decoratedMetadataToken, System.RuntimeType attributeFilterType, bool mustBeInheritable, System.RuntimeType.ListBuilder<object> derivedAttributes)	Unknown
 	System.Private.CoreLib.dll!System.Reflection.CustomAttribute.GetCustomAttributes(System.Reflection.RuntimeModule decoratedModule, int decoratedMetadataToken, int pcaCount, System.RuntimeType attributeFilterType)	Unknown
 	System.Private.CoreLib.dll!System.Reflection.CustomAttribute.GetCustomAttributes(System.Reflection.RuntimeMethodInfo method, System.RuntimeType caType, bool inherit)	Unknown
 	System.Private.CoreLib.dll!System.Attribute.GetCustomAttributes(System.Reflection.MemberInfo element, System.Type attributeType, bool inherit)	Unknown
[...]

PR #8649, opened 2023-02-16
Enable nullability in more small line count files

elachlan:
Related: #1107

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8649)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewSelectedColumnCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewSelectedColumnCollection.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.Collections;
 using System.ComponentModel;
 
@@ -14,7 +12,7 @@ namespace System.Windows.Forms
     {
         private readonly List<DataGridViewColumn> _items = new();
 
-        int IList.Add(object value)
+        int IList.Add(object? value)
         {
             throw new NotSupportedException(SR.DataGridView_ReadOnlyCollection);
         }
@@ -24,16 +22,16 @@ namespace System.Windows.Forms
             throw new NotSupportedException(SR.DataGridView_ReadOnlyCollection);
         }
 
-        bool IList.Contains(object value) => ((IList)_items).Contains(value);
+        bool IList.Contains(object? value) => ((IList)_items).Contains(value);
 
-        int IList.IndexOf(object value) => ((IList)_items).IndexOf(value);
+        int IList.IndexOf(object? value) => ((IList)_items).IndexOf(value);
 
-        void IList.Insert(int index, object value)
+        void IList.Insert(int index, object? value)
         {
             throw new NotSupportedException(SR.DataGridView_ReadOnlyCollection);
         }
 
-        void IList.Remove(object value)
+        void IList.Remove(object? value)
         {
             throw new NotSupportedException(SR.DataGridView_ReadOnlyCollection);
         }
@@ -47,7 +45,7 @@ namespace System.Windows.Forms
 
         bool IList.IsReadOnly => true;
 
-        object IList.this[int index]
+        object? IList.this[int index]
         {
             get { return _items[index]; }
             set { throw new NotSupportedException(SR.DataGridView_ReadOnlyCollection); }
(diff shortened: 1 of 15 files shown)


elachlan:
CC: @gpetrou

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.ControlCollection.cs
```diff
@@ -19,11 +17,11 @@ public ControlCollection(TabControl owner) : base(owner)
                 _owner = owner;
             }
 
-            public override void Add(Control value)
+            public override void Add(Control? value)
             {
-                if (!(value is TabPage))
+                if (value is not TabPage)
```
halgab, line 22:
```suggestion
                if (value is not TabPage tabPage)
```
and delete line 27? The same pattern can be found on line 70.

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/MergePropertyDescriptor.MultiMergeCollection.cs
```diff
@@ -12,7 +10,7 @@ internal partial class MergePropertyDescriptor
     {
         private class MultiMergeCollection : ICollection
         {
-            private object[] _items;
+            private object[]? _items;
```
lonitra, line 13: Should this be `object?[]?`

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewSelectedRowCollection.cs
```diff
@@ -28,16 +26,16 @@ void IList.Clear()
...
+        bool IList.Contains(object? value) => ((IList)_items).Contains(value);
 
-        int IList.IndexOf(object value) => ((IList)_items).IndexOf(value);
+        int IList.IndexOf(object? value) => ((IList)_items).IndexOf(value);
 
-        void IList.Insert(int index, object value)
+        void IList.Insert(int index, object? value)
         {
             throw new NotSupportedException(SR.DataGridView_ReadOnlyCollection);
         }
```
lonitra, line 36: We can clean these up a little more by doing the following in the appropriate places
```suggestion
        void IList.Insert(int index, object? value) => throw new NotSupportedException(SR.DataGridView_ReadOnlyCollection);
```

lonitra (commented):
There are some `TabControlControlCollection` tests that are expecting `NullReferenceException` and should be fixed to expect an `ArgumentException` if we are not expecting to be able to add `null`. If we need to make that change this might also be documentation breaking since the exception type has changed it seems.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.ControlCollection.cs
```diff
@@ -19,13 +19,13 @@ public ControlCollection(TabControl owner) : base(owner)
 
             public override void Add(Control? value)
             {
-                if (value is not TabPage)
+                value.OrThrowIfNull(nameof(value));
+
+                if (value is not TabPage tabPage)
                 {
                     throw new ArgumentException(string.Format(SR.TabControlInvalidTabPageType, value?.GetType().Name));
```
halgab, line 26:
```suggestion
                    throw new ArgumentException(string.Format(SR.TabControlInvalidTabPageType, value.GetType().Name));
```
you already checked `value` for `null` on line 22

lonitra (commented):
👍

dreddy-work:
Captured failed error here: https://github.com/dotnet/winforms/issues/8598#issuecomment-1435102622

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.ControlCollection.cs
elachlan: `OrThrowIfNull` does not have the annotations which indicate the input is not null if returned. So replaced with `ArgumentNullException.ThrowIfNull`.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewSelectedColumnCollection.cs
```diff
@@ -69,13 +67,7 @@ internal DataGridViewSelectedColumnCollection()
...
         protected override ArrayList List => ArrayList.Adapter(_items);
 
-        public DataGridViewColumn this[int index]
-        {
-            get
-            {
-                return _items[index];
-            }
-        }
+        public DataGridViewColumn this[int index] => _items[index];
```
dreddy-work, line 70: Q: if `_items[index]` is nullable on line 48, why not here? Build successful suggests this is not nullable?

elachlan: `IList` contract for `this` requires nullability. Where as the `List<DataGridViewColumn>` we have at the top of the class isn't nullable. So in the instance you pointed out, its not nullable.

Issue #8357, opened 2022-12-09
[Tracking] HashTable/ArrayList replacement in Winforms


dreddy-work:
While replacing HashSet and ArrayList we found scenarios where replacements from Generics are not functionally equivalent with non-generics. ex: [HashTable/ArrayLaist replacement in Winforms](https://github.com/dotnet/winforms/pull/8235/files#r1044787174). 

We are going to do a quick review around [here](https://github.com/dotnet/winforms/issues/8140) and [here](https://github.com/dotnet/winforms/issues/8143) and provide further guidance/helpers to make sure replacements are smooth.

related #8143 , #8140  

Community and primarily @elachlan , @Jericho , @hughbe @gpetrou that are helping in this space, can you help us on [these](https://github.com/dotnet/winforms/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22) meanwhile?

elachlan:
@merriemcgaw is this still on the teams radar? I'd like to get it "finished" before .NET8 release. I know its a while away, but time flies.

dreddy-work:
@elachlan , can you prioritize on the linked regression first and see what we can do for it in .NET 8.0 Preview 1?

elachlan:
I believe the error is coming from the designer. But I can't be too sure since I don't have a stack trace.

elachlan:
@dreddy-work I cannot respond in #8207 as I am not the author.

dreddy-work:
> @dreddy-work I cannot respond in #8207 as I am not the author.

no problem. I was mistaken you as owner. I will take closer look.

PR #8651, opened 2023-02-16
Refactor `BindingManagerBase` and related classes to replace `ArrayList`

elachlan:
Related: #8140, #8357

@merriemcgaw I think this is a breaking API change. But if the `ArrayList` contains anything but `PropertyDescriptor` it will cause an exception in `CurrencyManager.GetListName`.

CC: @KlausLoeffelmann because this relates to binding.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8651)

### DIFF
--- a/src/System.Windows.Forms/src/PublicAPI.Shipped.txt
+++ b/src/System.Windows.Forms/src/PublicAPI.Shipped.txt
@@ -375,7 +375,6 @@ override System.Windows.Forms.MonthCalendar.ToString() -> string!
 ~override System.Windows.Forms.PropertyGrid.Text.get -> string
 ~override System.Windows.Forms.PropertyGrid.Text.set -> void
 ~override System.Windows.Forms.PropertyManager.Current.get -> object
-~override System.Windows.Forms.PropertyManager.GetListName(System.Collections.ArrayList listAccessors) -> string
 ~override System.Windows.Forms.PropertyManager.OnCurrentChanged(System.EventArgs ea) -> void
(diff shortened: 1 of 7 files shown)


KlausLoeffelmann:
Only skimmed over this, but this seems to me a public surfacing API change, right?
This would be too much of a breaking change for me, to be honest.

dreddy-work:
@elachlan did we do any breaking changes with respect to public/protected function signatures in this matter?

elachlan:
I believe the changes are `protected internal` which makes this a breaking public API change. It always gets me because I assume AND, not OR with the access modifiers.

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/protected-internal

Otherwise this would have been a nice change.

dreddy-work:
> It always gets me because I assume AND, not OR with the access modifiers.

Agree. Sometimes i double check with documentation :)

elachlan:
@KlausLoeffelmann is it possible to pass anything into it, it seems like `PropertyDescriptor` is what is expected. But I imagine it really depends on the implementation, since `BindingMangerBase` is abstract.

KlausLoeffelmann:
The latter should be decisive, unfortunately. 
And while your ideas, quality and passion set a very high bar for rather taking things: But against backwards compatibility especially with public breaking changes one loses almost always! 🤷‍♂️

PR #8668, opened 2023-02-20
Refactor `BaseContextMenuStrip` to replace `ArrayList` with `List<T>`

elachlan:
Related: #8143, #8357

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8668)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BaseContextMenuStrip.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BaseContextMenuStrip.cs
@@ -4,7 +4,6 @@
 
 #nullable disable
 
-using System.Collections;
 using System.ComponentModel;
 using System.ComponentModel.Design;
 using System.Diagnostics;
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BaseContextMenuStrip.cs
```diff
@@ -80,7 +79,7 @@ private void RefreshSelectionMenuItem()
                 Items.Remove(selectionMenuItem);
             }
 
-            ArrayList parentControls = new ArrayList();
+            List<IComponent> parentControls = new();
```
dmitrii-drobotov, line 82: I think it would be better to use `List<Component>` here and in `AddParentTree` because it is later iterated as a list of `Component`:
https://github.com/dotnet/winforms/blob/4e395cac1a26d9dccf26c3a85ca9ba1de9b64f2c/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BaseContextMenuStrip.cs#L136

I don't think we'd ever add an `IComponent` which wouldn't be `Component` to it, but it seems safer to use `List<Component>`.

dmitrii-drobotov (approved):

lonitra (approved):

PR #8672, opened 2023-02-21
Refactor `DesignerHost` to replace `ArrayList` with `List<T>`

elachlan:
Related: #8143, #8357

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8672)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerHost.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerHost.cs
@@ -53,7 +53,7 @@ namespace System.ComponentModel.Design
         private readonly Dictionary<IComponent, IDesigner> _designers;  // designer -> component mapping
         private readonly EventHandlerList _events; // event list
         private DesignerLoader _loader; // the loader that loads our designers
-        private List<string> _savedSelection; // set of selected components saved across reloads
+        private List<string> _savedSelection; // set of selected components names saved across reloads
         private HostDesigntimeLicenseContext _licenseCtx;
         private IDesignerEventService _designerEventService;
         private static readonly object s_selfLock = new object();
@@ -1184,13 +1184,13 @@ namespace System.ComponentModel.Design
             // If the loader indicated success, but it never created a component, that is an error.
             if (successful && _rootComponent is null)
             {
-                List<object> errorList = new();
-                InvalidOperationException ex = new InvalidOperationException(SR.DesignerHostNoBaseClass)
+                errorCollection = new List<object>()
                 {
-                    HelpLink = SR.DesignerHostNoBaseClass
+                    new InvalidOperationException(SR.DesignerHostNoBaseClass)
+                    {
+                        HelpLink = SR.DesignerHostNoBaseClass
+                    }
                 };
-                errorList.Add(ex);
-                errorCollection = errorList;
                 successful = false;
             }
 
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerHost.cs
```diff
@@ -1181,7 +1184,7 @@ void IDesignerLoaderHost.EndLoad(string rootClassName, bool successful, ICollect
             // If the loader indicated success, but it never created a component, that is an error.
             if (successful && _rootComponent is null)
             {
-                ArrayList errorList = new ArrayList();
+                List<object> errorList = new();
```
lonitra, line 1187: Should this be `List<Exception>`?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerHost.cs
```diff
@@ -53,7 +53,7 @@ internal sealed class DesignerHost : Container, IDesignerLoaderHost2, IDesignerH
         private readonly Dictionary<IComponent, IDesigner> _designers;  // designer -> component mapping
         private readonly EventHandlerList _events; // event list
         private DesignerLoader _loader; // the loader that loads our designers
-        private ICollection _savedSelection; // set of selected components saved across reloads
+        private List<string> _savedSelection; // set of selected components saved across reloads
```
lonitra, line 56: This comment seems a little misleading. Maybe change it to the following
```suggestion
        private List<string> _savedSelection; // set of selected component names saved across reloads
```

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerHost.cs
elachlan: In other places such as `DesignerSerializationManager` we used `List<object>` because there is a possibility exception isn't the only type used.

e.g, `IDesignerSerializationManager.ReportError` contract takes an object and adds it to the errors list.
https://github.com/dotnet/winforms/blob/bba30b55a0ba2a5268709b8c616dc5e2cd2dc6a8/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/DesignerSerializationManager.cs#L876-L889

As far as I can see the code is related. I could be wrong.

elachlan:
I made some more changes designed to avoid problems when an array is passed in and not a list/arraylist. Also fixed a caller to the method which was passing in an arraylist.

elachlan:
@dreddy-work is the test failure a flaky test?

dreddy-work:
Error message

```CS

Stack trace
   at System.Windows.Forms.Tests.ListViewLabelEditAccessibleObjectTests.ListViewLabelEditAccessibleObject_IsPatternSupported_ReturnsExpected() in /_/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewLabelEditAccessibleObjectTests.cs:line 67
   at System.Reflection.MethodInvoker.Invoke(Object obj, IntPtr* args, BindingFlags invokeAttr)

```

dreddy-work:
> @dreddy-work is the test failure a flaky test?

Only on ARM suggests so.  @dmitrii-drobotov can you check why accessible object is disposed in this case?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerHost.cs
lonitra: I see, you are right. Let's keep it as `List<object>` then.

lonitra (approved):

dmitrii-drobotov (approved):

dreddy-work:
#8682

PR #8677, opened 2023-02-23
Refactor `StronglyTypedResourceBuilder` to replace `Hashtable` with `Dictionary`

elachlan:
Related: #8143, #8357

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8677)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Resources/Tools/StronglyTypedResourceBuilder.cs
+++ b/src/System.Windows.Forms.Design/src/System/Resources/Tools/StronglyTypedResourceBuilder.cs
@@ -322,7 +322,7 @@ namespace System.Resources.Tools
 
             // Verify the resource names are valid property names, and they don't conflict.  This includes checking for
             // language-specific keywords, translating spaces to underscores, etc.
-            var cleanedResourceList = VerifyResourceNames(resourceList, codeProvider, errors, out Hashtable reverseFixupTable);
+            var cleanedResourceList = VerifyResourceNames(resourceList, codeProvider, errors, out Dictionary<string, string> reverseFixupTable);
 
             // Verify the class name is legal.
             string className = baseName;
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms.Design/src/System/Resources/Tools/StronglyTypedResourceBuilder.cs
```diff
@@ -898,7 +899,7 @@ private static string VerifyResourceName(string key, CodeDomProvider provider, b
                     }
 
                     // Now see if we've already mapped another key to the same name.
-                    string oldDuplicateKey = (string)reverseFixupTable[newKey];
+                    string oldDuplicateKey = reverseFixupTable[newKey];
```
gpetrou, line 902: When switching `Hashtable` to `Dictionary` it is better to use `TryGetValue` in order to avoid possible exception. The indexers don't behave the same.

elachlan:
@lonitra I think that sorts it.

src/System.Windows.Forms.Design/src/System/Resources/Tools/StronglyTypedResourceBuilder.cs
```diff
@@ -898,8 +899,7 @@ private static string VerifyResourceName(string key, CodeDomProvider provider, b
                     }
 
                     // Now see if we've already mapped another key to the same name.
-                    string oldDuplicateKey = (string)reverseFixupTable[newKey];
-                    if (oldDuplicateKey is not null)
+                    if (reverseFixupTable.TryGetValue(newKey, out string oldDuplicateKey))
```
lonitra, line 902: Is it possible for `reverseFixupTable` to contain a key that maps to a `null` value? If it is, we will either need to keep the null check on `oldDuplicateKey` here and on `fixedUp` below or prevent a null value from being added in the first place if we do not expect dictionary to contain null. `TryGetValue()` returns true if the dictionary contains an element with the specified key, so it is possible to still get `true` even if the key maps to a `null` value.

lonitra (approved):
Thank you 🥳

Issue #8203, opened 2022-11-16
Use interpolated strings in WinForms


JeremyKuhne:
WinForms formatted string usages should be converted to interpolated strings where possible to get the various performance benefits.

https://learn.microsoft.com/dotnet/csharp/language-reference/tokens/interpolated

This is a good community contribution issue. Details:

- Keep PRs to a manageable size
- This includes usages of `+` to concat strings
- `string.Format` is still needed for anything that gets localized strings (`SR.*`)
- In most cases we don't need to specify culture
  - Not needed ever for integers or booleans that aren't special formatted
  - Never really needed in Debug statements

Examples of changes:

``` C#
// No culture needed - non localized string, debug only, boolean
Debug.WriteLineIf(CompModSwitches.ImeMode.Level >= TraceLevel.Info, string.Format(CultureInfo.CurrentCulture, "Inside get_CanEnableIme(), value = {0}, this = {1}", ImeSupported, this));

// Becomes...

Debug.WriteLineIf(CompModSwitches.ImeMode.Level >= TraceLevel.Info, $"Inside get_CanEnableIme(), value = {ImeSupported}, this = {this}");



Debug.WriteLineIf(CompModSwitches.ImeMode.Level >= TraceLevel.Info, "Inside get_CurrentImeContextMode(), this = " + this)

// Becomes...

Debug.WriteLineIf(CompModSwitches.ImeMode.Level >= TraceLevel.Info, $"Inside get_CurrentImeContextMode(), this = {this}")
```

I don't think we have many cases where [raw string literals](https://learn.microsoft.com/dotnet/csharp/language-reference/proposals/csharp-11.0/raw-string-literal) are needed, but that can be included in this work if any come up. (`$"""Count {0}"""`)

We should also identify and collect places in the WinForms code where we're frequently passing formatted strings that may not be used so we can look for ways to leverage skipping formatting in much the same way that `Debug.WriteLineIf`, etc. do.

elachlan:
@Youssef1313 Do you know of an analyzer for converting string.format to interpolation and for string concatenation?

Youssef1313:
Not that I know of.

JeremyKuhne:
There is an auto-fix for converting to interpolated strings that will get you some of the way there.

halgab:
[...]

PR #8681, opened 2023-02-23
Use interpolated strings in Control.cs

halgab:
Contributes to #8203

As the issue mentions that the PR sizes should be manageable, I chose to start with one file that sums up all the cases I encountered. I have plenty more changes ready to submit.

## Test methodology

static code analysis

## Test environment(s)

8.0.100-preview.1.23115.2, but I failed to run the tests locally...

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8681)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
@@ -10726,7 +10726,7 @@ namespace System.Windows.Forms
         protected virtual void SetBoundsCore(int x, int y, int width, int height, BoundsSpecified specified)
         {
             Debug.WriteLineIf(CompModSwitches.SetBounds.TraceInfo,
-                $"{Name}::SetBoundsCore(x={x} y={y} width={width} height={height} specified={specified}");
+                $"{Name}::SetBoundsCore(x={x} y={y} width={width} height={height} specified={specified})");
             // SetWindowPos below sends a WmWindowPositionChanged (not posts) so we immediately
             // end up in WmWindowPositionChanged which may cause the parent to layout.  We need to
             // suspend/resume to defer the parent from laying out until after InitLayout has been called
@@ -11510,7 +11510,6 @@ namespace System.Windows.Forms
                 {
                     Debug.WriteLine($"""
                         WARNING: Bounds changed during OnLocationChanged()
-
                         before={originalBounds} after={Bounds}
                         """);
                 }
@@ -11534,7 +11533,6 @@ namespace System.Windows.Forms
                 {
                     Debug.WriteLine($"""
                         WARNING: Bounds changed during OnSizeChanged()
-
                         before={originalBounds} after={Bounds}
                         """);
                 }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
@@ -75,7 +75,7 @@ namespace System.Windows.Forms
                 string str;
                 try
                 {
-                    str = string.Format(CultureInfo.CurrentCulture, "{0}<{1}>", GetType().Name, Text);
+                    str = $"{GetType().Name}<{Text}>";
                     int maxFrameCount = new StackTrace().FrameCount;
                     if (maxFrameCount > 5)
                     {
@@ -8820,7 +8820,7 @@ namespace System.Windows.Forms
 
                                 break;
                             default:
-                                Debug.Fail("Unknown PaintLayer " + layer);
+                                Debug.Fail($"Unknown PaintLayer {layer}");
                                 break;
                         }
 
@@ -9758,17 +9758,8 @@ namespace System.Windows.Forms
 
                 bool focused = ContainsFocus;
 
-#if DEBUG
-                if (CoreSwitches.PerfTrack.Enabled)
-                {
-                    Debug.Write("RecreateHandle: ");
-                    Debug.Write(GetType().FullName);
-                    Debug.Write(" [Text=");
-                    Debug.Write(Text);
-                    Debug.Write("]");
-                    Debug.WriteLine("");
-                }
-#endif
+                Debug.WriteLineIf(CoreSwitches.PerfTrack.Enabled, $"RecreateHandle: {GetType().FullName} [Text={Text}]");
+
                 bool created = GetState(States.Created);
                 if (GetState(States.TrackingMouseEvent))
                 {
@@ -10017,12 +10008,8 @@ namespace System.Windows.Forms
         /// </summary>
         public void ResumeLayout(bool performLayout)
         {
-#if DEBUG
-            if (CompModSwitches.LayoutSuspendResume.TraceInfo)
-            {
-                Debug.WriteLine($"{GetType().Name}::ResumeLayout( preformLayout = {performLayout}, newCount = {Math.Max(0, LayoutSuspendCount - 1)})");
-            }
-#endif
+            Debug.WriteLineIf(CompModSwitches.LayoutSuspendResume.TraceInfo,
+                $"{GetType().Name}::ResumeLayout( preformLayout = {performLayout}, newCount = {Math.Max(0, LayoutSuspendCount - 1)})");
 
             bool performedLayout = false;
             if (LayoutSuspendCount > 0)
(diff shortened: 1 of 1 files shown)


halgab:
@dotnet-policy-service agree

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -9758,17 +9758,8 @@ internal virtual void RecreateHandleCore()
...
-                {
-                    Debug.Write("RecreateHandle: ");
-                    Debug.Write(GetType().FullName);
-                    Debug.Write(" [Text=");
-                    Debug.Write(Text);
-                    Debug.Write("]");
-                    Debug.WriteLine("");
-                }
-#endif
+                Debug.WriteLineIf(CoreSwitches.PerfTrack.Enabled, $"RecreateHandle: {GetType().FullName} [Text={Text}]");
```
dreddy-work, line 9761: nice!

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -11382,13 +11360,9 @@ public void SuspendLayout()
...
 
-#if DEBUG
             Debug.Assert(LayoutSuspendCount > 0, "SuspendLayout: layoutSuspendCount overflowed.");
-            if (CompModSwitches.LayoutSuspendResume.TraceInfo)
-            {
-                Debug.WriteLine($"{GetType().Name} ::SuspendLayout( newCount = {LayoutSuspendCount}");
-            }
-#endif
+            Debug.WriteLineIf(CompModSwitches.LayoutSuspendResume.TraceInfo,
+                $"{GetType().Name} ::SuspendLayout( newCount = {LayoutSuspendCount}");
```
Tanya-Solyanik, line 11365:
```suggestion
                $"{GetType().Name} ::SuspendLayout(newCount = {LayoutSuspendCount}");
```

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -11508,11 +11482,9 @@ protected void UpdateBounds(int x, int y, int width, int height, int clientWidth
 #if DEBUG
             if (CompModSwitches.SetBounds.TraceVerbose)
             {
-                Debug.WriteLine(string.Format(CultureInfo.CurrentCulture, "{0}::UpdateBounds(", Name));
+                Debug.WriteLine($"{Name}::UpdateBounds(");
```
Tanya-Solyanik, line 11485:
```suggestion
                Debug.WriteLine($"{Name ?? ToString()}::UpdateBounds(");
```

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -10328,13 +10315,8 @@ internal void ScaleControl(SizeF includedFactor, SizeF excludedFactor, Control r
                 {
                     excludedSpecified |= (~RequiredScaling & BoundsSpecified.All);
                 }
-#if DEBUG
-                if (CompModSwitches.RichLayout.TraceInfo)
-                {
-                    Debug.WriteLine(string.Format(CultureInfo.InvariantCulture, "Scaling {0} Included: {1}, Excluded: {2}",
```
dreddy-work, line 10334: I don't see need for Invariant Culture here. We should be okay with interpolation.

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -11536,7 +11508,11 @@ protected void UpdateBounds(int x, int y, int width, int height, int clientWidth
 #if DEBUG
                 if (Bounds != originalBounds && CompModSwitches.SetBounds.TraceWarning)
                 {
-                    Debug.WriteLine(string.Format(CultureInfo.CurrentCulture, "WARNING: Bounds changed during OnLocationChanged()\r\nbefore={0} after={1}", originalBounds, Bounds));
+                    Debug.WriteLine($"""
```
dreddy-work, line 11511: nit: can be converted into `Debug.WriteLinIf`. Same below.

halgab: I thought so at first, but `originalBounds` only exists in debug...

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
halgab: I hadn't notice this one was invariant... I'll check again

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
halgab: While we're trying to fix this (although I had doubts including this change as it's slightly out of scope...), doesn't it look like it's also missing a closing parenthesis?

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
halgab: `Name` appears to be non-nullable though

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
halgab: I think it does have an impact on the output. The original issue mentioned the fact that it's not that relevant for debug statements anyway, but I can replace the code with a version that accounts for the culture if you prefer.

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
Tanya-Solyanik: Good point, add the closing parenthesis as well!

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
halgab: Do you agree with me that the fallback on `ToString()` is not necessary as `Name` cannot be null?

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -11536,7 +11508,11 @@ protected void UpdateBounds(int x, int y, int width, int height, int clientWidth
...
+                        WARNING: Bounds changed during OnLocationChanged()
+
+                        before={originalBounds} after={Bounds}
```
dmitrii-drobotov, line 11514: There was only one line break previously, the empty line is redundant:
```suggestion
                        WARNING: Bounds changed during OnLocationChanged()
                        before={originalBounds} after={Bounds}
```

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -10743,12 +10725,8 @@ public void SetBounds(int x, int y, int width, int height, BoundsSpecified speci
...
         protected virtual void SetBoundsCore(int x, int y, int width, int height, BoundsSpecified specified)
         {
-#if DEBUG
-            if (CompModSwitches.SetBounds.TraceInfo)
-            {
-                Debug.WriteLine(string.Format(CultureInfo.CurrentCulture, "{0}::SetBoundsCore(x={1} y={2} width={3} height={4} specified={5}", Name, x, y, width, height, specified));
-            }
-#endif
+            Debug.WriteLineIf(CompModSwitches.SetBounds.TraceInfo,
+                $"{Name}::SetBoundsCore(x={x} y={y} width={width} height={height} specified={specified}");
```
dmitrii-drobotov, line 10729: Could you add a missing closing parenthesis here?
```suggestion
                $"{Name}::SetBoundsCore(x={x} y={y} width={width} height={height} specified={specified})");
```

dmitrii-drobotov (commented):
Just a couple nits, otherwise looks good!

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
halgab: You're right, thanks. Fixed line 11537 as well

dmitrii-drobotov (approved):

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
dreddy-work: @halgab, Do you have sample of what difference you noticed with `InvariantCulture` for this statement?

dreddy-work (approved):

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
dreddy-work: i see `getter` for name checks for null. We should be good.

Tanya-Solyanik: >Do you agree with me that the fallback on ToString() is not necessary as Name cannot be null?

@halgab - Name can be an empty string, in that case it's better to print out type name via the ToString method.

halgab: I can add a check for empty string in a follow-up PR if you want me to.

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
halgab: I don't, it's just that `includedFactor` and `excludedFactor` are both of type `SizeF`. On first read, I thought this type was culture-dependant, but reading the code again it's not, sorry for the confusion

Issue #8657, opened 2023-02-18
ThreadExceptionFixture Instantiated in Wrong Thread


lonitra:
In all of our test classes we derive from `IClassFixture<ThreadExceptionFixture>` so that, before the first test in the class is ran, we are able to hook into listen for exceptions that are thrown in the thread running the tests and fail a test on an exception instead of allowing the exception dialog to show. However, `ThreadExceptionFixture` is not being instantiated in the same thread as the thread that runs the tests. This means that on an exception, the exception dialog will show and cause the test to hang or, if it manages to bypass the exception dialog, will show that the test passed in some cases.

lonitra:
cc @dreddy-work @JeremyKuhne

PR #8695, opened 2023-02-24
Fix exception dialog handling.

JeremyKuhne:
Test fixtures do not run on the same thread the test method executes on when marked with `[WinFormsFact]` or `[StaFact]`. This change removes the test fixture and adds a new context switch to throw thread exceptions.

Initialization of the switch is done programmatically via a module initializer. The module initializer will also set the starting cursor position to avoid spurious errors in tests due to cursor positioning.

The Cursor tests were the ones modifying the cursor position and causing intermittent errors in other tests. Moved them to their own collection and forced them to run sequentially.

Cleaned up WindowsFormsSynchronizationContext.

Simplified checks in LocalAppContextSwitches. We shouldn't be conditioning on supported OS beyond checking for API availability. We don't support Windows 7, but we won't deliberately cause WinForms apps to crash on Windows by calling later APIs without checking for support. We will not add additional complexity to the code beyond that for unsupported OS platforms.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8695)


JeremyKuhne:
Fixes: #8657

lonitra (approved):
🥳

Tanya-Solyanik:
>We shouldn't be conditioning on supported OS beyond checking for API availability. 

This is pretty much the case of API availability. PMv2 is not available on these OSs, if the user app follows that path, hits the unavailable API, we don't have a way to handle it cleanly.

I would rather  factor this change out of this PR and investigate it more.

JeremyKuhne:
> This is pretty much the case of API availability. PMv2 is not available on these OSs, if the user app follows that path, hits the unavailable API, we don't have a way to handle it cleanly.

It should be done at the consuming site or clearly documented here ("this feature uses APINAME which requires OS blah").

JeremyKuhne:
> I would rather factor this change out of this PR and investigate it more.

I'm out for a week and this is needed for additional test work so I'm not going to pull it out.

RussKie:
Nice!

PR #8698, opened 2023-02-25
Enable nullability in ObjectCollection

gpetrou:
## Proposed changes

- Enable nullability in ObjectCollection

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8698)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ObjectCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ObjectCollection.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.Collections;
 using System.ComponentModel;
 using System.Diagnostics;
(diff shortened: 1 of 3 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ObjectCollection.cs
```diff
@@ -315,7 +313,7 @@ void ICollection.CopyTo(Array destination, int index)
             ///  A SystemException occurs if there is insufficient space available to
             ///  store the new item.
             /// </summary>
-            public void Insert(int index, object item)
+            public void Insert(int index, object? item)
```
lonitra, line 316: We end up throwing if `item` is null so would it be better to remove the `?` to indicate to users that we do not allow null?

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ObjectCollection.cs
```diff
@@ -173,9 +171,9 @@ private int AddInternal(object item)
                 return index;
             }
 
-            int IList.Add(object item)
+            int IList.Add(object? item)
             {
-                return Add(item);
+                return Add(item!);
```
lonitra, line 176: Does this end up as a no-op if `item` is null?

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ObjectCollection.cs
gpetrou: This implements `IList.Insert`, so there is a warning if `?` is missing.

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ObjectCollection.cs
gpetrou: This will result in `ArgumentNullException`, but changing the code here will result in a breaking change and in other PRs we did not want to do this with collections.

dreddy-work (approved):

PR #8702, opened 2023-02-26
Refactor `DesignerSerializationManager` dictionary usage to use `TryGetValue`

elachlan:
Refactor dictionary usage to use `TryGetValue`.

Addresses feedback from:
https://github.com/dotnet/winforms/pull/8353#issuecomment-1443195430

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8702)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/DesignerSerializationManager.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/DesignerSerializationManager.cs
@@ -179,7 +179,7 @@ namespace System.ComponentModel.Design.Serialization
         }
 
         /// <summary>
-        ///  Used to verify that there is an open session.  If there isn't, this method throws.
+        ///  Used to verify that there is an open session. If there isn't, this method throws.
         /// </summary>
         private void CheckSession()
         {
@@ -210,10 +210,7 @@ namespace System.ComponentModel.Design.Serialization
             // If we have been asked to recycle instances, look in our nametable and container first for an object matching this name and type.  If we find it, we will use it.
             if (RecycleInstances && name is not null)
             {
-                if (instancesByName is not null)
-                {
-                    instance = instancesByName[name];
-                }
+                instancesByName?.TryGetValue(name, out instance);
 
                 if (instance is null && addToContainer && Container is not null)
                 {
@@ -420,9 +417,8 @@ namespace System.ComponentModel.Design.Serialization
             {
                 if (serializers is not null)
                 {
-                    // I don't double hash here.  It will be a very rare day where multiple types of serializers will be used in the same scheme. We still handle it, but we just don't cache.
-                    serializer = serializers[objectType];
-                    if (serializer is not null && !serializerType.IsAssignableFrom(serializer.GetType()))
+                    // I don't double hash here. It will be a very rare day where multiple types of serializers will be used in the same scheme. We still handle it, but we just don't cache.
+                    if (serializers.TryGetValue(objectType, out serializer) && !serializerType.IsAssignableFrom(serializer.GetType()))
                     {
                         serializer = null;
                     }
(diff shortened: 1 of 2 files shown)


elachlan:
CC: @KalleOlaviNiemitalo @dreddy-work

elachlan:
> Would this deserve a test too?

I'll look into it when I have more time. I am not sure how to go about it.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/DesignerSerializationManager.cs
```diff
@@ -420,9 +417,8 @@ public IDisposable CreateSession()
             {
                 if (serializers is not null)
                 {
-                    // I don't double hash here.  It will be a very rare day where multiple types of serializers will be used in the same scheme. We still handle it, but we just don't cache.
-                    serializer = serializers[objectType];
-                    if (serializer is not null && !serializerType.IsAssignableFrom(serializer.GetType()))
+                    // I don't double hash here. It will be a very rare day where multiple types of serializers will be used in the same scheme. We still handle it, but we just don't cache.
+                    if (serializers.TryGetValue(objectType, out serializer) && !serializerType.IsAssignableFrom(serializer.GetType()))
```
lonitra, line 421: Is it possible for `objectType` to map to a `null` value in `serializers`? If so, we should check if `serializer` is null before calling `serializer.GetType()`

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/DesignerSerializationManager.cs
```diff
@@ -804,26 +800,22 @@ object IDesignerSerializationManager.CreateInstance(Type type, ICollection? argu
         /// </summary>
         string? IDesignerSerializationManager.GetName(object value)
         {
-            string? name = null;
             ArgumentNullException.ThrowIfNull(value);
 
             CheckSession();
             // Check our local nametable first
-            if (namesByInstance is not null)
+            if (namesByInstance is not null && namesByInstance.TryGetValue(value, out string? name))
```
lonitra, line 807: Ditto here. If `value` can map to a null value in `namesByInstance`, we should check if `name` is not null before returning it.

lonitra:
> > Would this deserve a test too?
> 
> I'll look into it when I have more time. I am not sure how to go about it.

Off the top of my head I am not sure how complicated our unit tests for `DesignerSerializerManager` are, but it is worth revisiting those tests to see if our existing tests that exercise these methods calling to the now dictionaries also exercises it with keys that don't exist to ensure we are not unexpectedly throwing.

dreddy-work:
@elachlan, just checking if this is still on your radar.

elachlan:
Yes. Work has taken over a bit due to some deadlines. I'll see what I can do.

dreddy-work:
@elachlan, did you get chance to respond on couple of open comment above?

elachlan:
I'll look today.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/DesignerSerializationManager.cs
elachlan: namesByInstance/instancesByName do not accept a nullable value.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/DesignerSerializationManager.cs
elachlan: We check `objectType` for null just above on line 416.

elachlan:
@dreddy-work I believe its addressed. But feel free to add to it if I have missed something.

dreddy-work (approved):

PR #8715, opened 2023-02-28
Use interpolated strings in toolstrip-related code

halgab:
Contributes to #8203

## Test methodology

Static code analysis, CI

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8715)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripSplitStackLayout.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripSplitStackLayout.cs
@@ -6,9 +6,6 @@
 
 using System.Diagnostics;
 using System.Drawing;
-#if DEBUG
-using System.Globalization;
-#endif
 using System.Windows.Forms.Layout;
 
 namespace System.Windows.Forms
@@ -89,12 +86,7 @@ namespace System.Windows.Forms
                     // if we have something set to overflow always we need to show an overflow button
                     if (item.Overflow == ToolStripItemOverflow.Always)
                     {
-#if DEBUG
-                        if (DebugLayoutTraceSwitch.TraceVerbose)
-                        {
-                            Debug.WriteLine(string.Format(CultureInfo.CurrentCulture, "OverflowRequired - item set to always overflow: {0} ", item));
-                        }
-#endif
+                        DebugLayoutTraceSwitch.TraceVerbose($"OverflowRequired - item set to always overflow: {item} ");
                         OverflowRequired = true;
                     }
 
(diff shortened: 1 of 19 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ToolStripSplitStackLayout.cs
```diff
@@ -89,12 +86,7 @@ private void CalculatePlacementsHorizontal()
                     // if we have something set to overflow always we need to show an overflow button
                     if (item.Overflow == ToolStripItemOverflow.Always)
                     {
-#if DEBUG
-                        if (DebugLayoutTraceSwitch.TraceVerbose)
-                        {
-                            Debug.WriteLine(string.Format(CultureInfo.CurrentCulture, "OverflowRequired - item set to always overflow: {0} ", item));
-                        }
-#endif
+                        DebugLayoutTraceSwitch.TraceVerbose($"OverflowRequired - item set to always overflow: {item} ");
```
halgab, line 89: This is actually a deoptimisation. I made this change to make the code more consistent, and to put this up for debate because this pattern is used across the whole codebase.
Previously, the message string was created only when tracing was set to verbose. After this change it's going to be created whenever we're in debug, as [TraceVerbose](https://github.com/dotnet/winforms/blob/a8e341cb99955ec3cbdea9815dfa71e487661633/src/Common/src/TraceSwitchExtensions.cs#L12) doesn't handle interpolated strings.
I see 2 potential fixes to this pattern:

1. Introduce a custom interpolated string handler for the `TraceVerbose` extension method
2. inline uses of the extension method as such:
```suggestion
                        Debug.WriteLineIf(DebugLayoutTraceSwitch!.TraceVerbose, $"SendNextItemToOverflow to free space for {item}");
```

gpetrou: The reason we had introduced the extension method was to avoid using the null-forgiving operator. You can check https://github.com/dotnet/winforms/pull/8072#discussion_r1009826425 for more details.

halgab: #8719

dreddy-work: Your proposal, @halgab, is currently under discussion and received positive feedback. Would you like to open the draft for review and potentially utilize it in this PR?

halgab: Sure!

dreddy-work (approved):

PR #8719, opened 2023-02-28
Add interpolated string handler to TraceSwitchExtensions.TraceVerbose

halgab:
Contributes to #8203

## Proposed changes

As stated in #8715, TraceSwitchExtensions.TraceVerbose doesn't handle interpolated strings despite replacing code that does (`Debug.WriteLineIf`). This is a proposed fix. Note that unlike `Debug.WriteIfInterpolatedStringHandler`, which is an obvious inspiration to this PR, we don't pull (yet?) `StringBuilder` instances as the caching mechanism is internal to corelib.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8719)

### DIFF
--- a/src/Common/src/TraceSwitchExtensions.cs
+++ b/src/Common/src/TraceSwitchExtensions.cs
@@ -109,62 +109,16 @@ namespace System.Windows.Forms
             public void AppendFormatted<T>(T value, string? format) => _stringBuilderHandler.AppendFormatted(value, format);
 
             /// <summary>
-            ///   Writes the specified value to the handler.
-            /// </summary>
-            /// <param name="value">The value to write.</param>
-            /// <param name="alignment">Minimum number of characters that should be written for this value.  If the
-            /// value is negative, it indicates left-aligned and the required minimum is the absolute value.</param>
-            /// <typeparam name="T">The type of the value to write.</typeparam>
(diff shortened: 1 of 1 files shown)


JeremyKuhne (approved):
Love it! Thanks!

JeremyKuhne:
We can follow up later on StringBuilder instance caching. For tracing I don't think it is critical.

halgab:
I guess the feedback I got from the pretty similar PR #8736 applies here as well. I'll make the adjustments (at least to the xml docs) shortly

halgab:
I've updated the xml docs. I kept all the overloads to the handler for now. I haven't checked they were all used (I guess that typically alignment isn't used anywhere), but as there are more than 500 call sites, I guess it doesn't hurt keeping the span overload around as it could come up.
I kept the handler in the same file as it's pretty focused, but I can extract it if you want.

Tanya-Solyanik:
Thank you @halgab!

> I kept all the overloads to the handler for now.

I would prefer removing the unused code and readding it when needed.

>I kept the handler in the same file as it's pretty focused, but I can extract it if you want. 

I'm ok with this

halgab:
Okay, that's done. I've kept the currently unused span overload of method AppendFormatted because I'll be using it in #8715. If you prefer I can remove it here and add it back there.

dreddy-work (approved):

PR #8720, opened 2023-02-28
Use interpolated strings in test code

halgab:
Contributes to #8203

## Test methodology

Static code analysis, CI

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8720)

### DIFF
--- a/src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/RichTextBoxes.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/RichTextBoxes.cs
@@ -58,20 +58,20 @@ This is a custom link\v #data#\v0  with hidden text after the link.\par
         private string ReportLinkClickedEventArgs(object sender, LinkClickedEventArgs e)
         {
             var control = (RichTextBox)sender;
-            var prefix = control.Text.AsSpan(0, e.LinkStart);
-            var content = control.Text.AsSpan(e.LinkStart, e.LinkLength);
-            var suffix = control.Text.AsSpan(e.LinkStart + e.LinkLength);
+            var prefix = control.Text.Remove(e.LinkStart);
+            var content = control.Text.Substring(e.LinkStart, e.LinkLength);
+            var suffix = control.Text.Substring(e.LinkStart + e.LinkLength);
 
             var index = prefix.LastIndexOf('\n');
             if (index >= 0)
             {
-                prefix = prefix.Slice(index + 1);
+                prefix = prefix.Substring(index + 1);
             }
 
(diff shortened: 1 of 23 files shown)


src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/RichTextBoxes.cs
```diff
@@ -58,24 +58,32 @@ private unsafe void MakeLink(RichTextBox control, string text)
         private string ReportLinkClickedEventArgs(object sender, LinkClickedEventArgs e)
         {
             var control = (RichTextBox)sender;
-            var prefix = control.Text.Remove(e.LinkStart);
```
dreddy-work, line 61: you may want to do this in a separate PR by enabling analyzers. Posible project by project.

halgab: Okay, no problem. I just thought that as those values were only used in the interpolated string below that was in scope, but I can rollback if you prefer.
I don't know of any analyzer that can flag this though, so I guess this change will probably be lost for now

dreddy-work: Understand. Wanted to group changes to be able to track easily in case any regression issues. I see [doc](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1831) that may help us enable analyzer in follow-up work.

halgab: That's rollbacked. Thanks for the link. Looking at [this file](https://github.com/dotnet/winforms/blob/83f3a8476d682f8b74c7dce5505cc39560298447/eng/CodeAnalysis.test.globalconfig#L325) makes me think this analyzer is already enabled, but I'm not sure I'm looking at the right place...

dmitrii-drobotov: @halgab - yes, that's the correct file, but it seems CA1831 is triggered only with array indexer operators, which is not the case here. There's also [CA1846](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1846), but that one is apparently only for method arguments. I didn't find other rules that would work in this case, so I guess we'll have to search for these cases manually.

dmitrii-drobotov (approved):

PR #8721, opened 2023-02-28
Use interpolated strings in controls

halgab:
Contributes to #8203

This one is pretty big, sorry about that...

## Test methodology 

Static code analysis, CI

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8721)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ThreadExceptionDialog.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ThreadExceptionDialog.cs
@@ -166,7 +166,6 @@ namespace System.Windows.Forms
             }
 
             StringBuilder detailsTextBuilder = new StringBuilder();
-            string newline = "\r\n";
             string separator = SR.ExDlgMsgSeparator;
             string sectionseparator = SR.ExDlgMsgSectionSeparator;
             if (Application.CustomThreadExceptionHandlerAttached)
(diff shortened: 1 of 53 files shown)


dmitrii-drobotov (approved):
Great work!

lonitra (approved):
🥳

src/System.Windows.Forms/src/System/Windows/Forms/ThreadExceptionDialog.cs
```diff
@@ -205,23 +203,22 @@ public ThreadExceptionDialog(Exception t)
...
                 }
 
-                detailsTextBuilder.Append(string.Format(SR.ExDlgMsgLoadedAssembliesEntry, name.Name, name.Version, fileVer, name.EscapedCodeBase));
+                detailsTextBuilder.AppendFormat(SR.ExDlgMsgLoadedAssembliesEntry, name.Name, name.Version, fileVer, name.EscapedCodeBase);
                 detailsTextBuilder.Append(separator);
             }
 #pragma warning restore SYSLIB0044 // Type or member is obsolete
 
-            detailsTextBuilder.Append(string.Format(CultureInfo.CurrentCulture, sectionseparator, SR.ExDlgMsgJITDebuggingSection));
+            detailsTextBuilder.AppendFormat(CultureInfo.CurrentCulture, sectionseparator, SR.ExDlgMsgJITDebuggingSection);
```
halgab, line 211: I'm just noticing, it's been that way since forever, but the format string and arguments look switched... But looking at the resource string, I don't see it using any argument! I'm not sure what the right fix is in that instance.

lonitra: Hm, am I missing something? `sectionseparator` refers to `SR.ExDlgMsgSectionSeparator`, which I am seeing maps to the following
https://github.com/dotnet/winforms/blob/37d7644f28fbbe1f8292d1f4c849984c3d32c572/src/System.Windows.Forms/src/Resources/SR.resx#L2925-L2929

halgab: My bad, I read the code too quickly and was misled by the fact that the `SR.*` part is usually the format string... Sorry for the false alarm!

lonitra: No worries! An easy mistake to make especially after looking at so many of these :)

PR #8743, opened 2023-03-04
Inline CodeDomComponentSerializationService.Trace method

halgab:
Contributes to #8203

## Proposed changes

Method `CodeDomComponentSerializationService.Trace` is only called 5 times, including one time where it performs a useless check. I thought it would be simpler to just inline the method rather than adding a dedicated interpolated string handler.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8743)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs
@@ -178,6 +178,8 @@ namespace System.ComponentModel.Design.Serialization
         {
 #if DEBUG
             private static readonly TraceSwitch s_trace = new TraceSwitch("ComponentSerializationService", "Trace component serialization");
+#else
+            private static readonly TraceSwitch s_trace;
 #endif
             private const string StateKey = "State";
             private const string NameKey = "Names";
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs
```diff
@@ -277,7 +281,7 @@ internal void AddMember(object value, MemberDescriptor member, bool absolute)
                     _objectNames.Add(data._name);
                 }
 
-                Trace("Adding object '{0}' ({1}:{2}) {3}", data._name, data._value.GetType().FullName, member.Name, (absolute ? "NORMAL" : "ABSOLUTE"));
+                Debug.WriteLineIf(s_trace.TraceVerbose, $"ComponentSerialization: Adding object '{data._name}' ({data._value.GetType().FullName}:{member.Name}) {(absolute ? "NORMAL" : "ABSOLUTE")}");
```
dmitrii-drobotov, line 284: It seems it should be this way:
```suggestion
                Debug.WriteLineIf(s_trace.TraceVerbose, $"ComponentSerialization: Adding object '{data._name}' ({data._value.GetType().FullName}:{member.Name}) {(absolute ? "ABSOLUTE" : "NORMAL")}");
```
Could you please fix it here and in `AddObject`?

halgab: good catch

dmitrii-drobotov (approved):

PR #8762, opened 2023-03-07
Enable nullability in ToolStripPanelRow

gpetrou:
## Proposed changes

- Enable nullability in ToolStripPanelRow.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8762)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripPanelRow.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripPanelRow.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.Collections.Specialized;
 using System.ComponentModel;
 using System.Diagnostics;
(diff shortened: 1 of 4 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ToolStripPanelRow.cs
```diff
@@ -530,9 +531,10 @@ private void OnLayoutHorizontalPostFix()
 
         private void OnLayoutVerticalPostFix()
         {
-            ToolStripPanelCell cell = RowManager.GetNextVisibleCell(Cells.Count - 1, /*forward*/false);
+            ToolStripPanelCell? cell = RowManager.GetNextVisibleCell(Cells.Count - 1, forward: false);
+
             // figure out how much space we actually need to free.
-            int spaceToFree = cell.CachedBounds.Bottom - RowManager.DisplayRectangle.Bottom;
+            int spaceToFree = cell!.CachedBounds.Bottom - RowManager.DisplayRectangle.Bottom;
```
dmitrii-drobotov, line 537: Looking at similar methods above, they both have
```
if (cell is null)
{
    ApplyCachedBounds();
    return;
}
```
So I would add the same check here, instead of using `!` operator.

dmitrii-drobotov (approved):
LGTM

PR #8769, opened 2023-03-08
Inline UndoEngine.Trace method

halgab:
Contributes to #8203

## Proposed changes

It's the same story as in #8743. This time I'm inlining a lot more calls though, around 15, so we're paying a higher cost to make the most of interpolated strings.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8769)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.cs
@@ -1160,9 +1160,9 @@ namespace System.ComponentModel.Design
                     _member = e.Member;
 
                     Debug.WriteLineIf(s_traceUndo.TraceVerbose, $"UndoEngine: ---> Creating change undo event for '{_componentName}'");
-                    Debug.WriteLineIf(s_traceUndo.TraceVerbose, $"UndoEngine: ---> Saving before snapshot for change to '{_componentName}'");
                     if (serializeBeforeState)
                     {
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.cs
```diff
@@ -1166,8 +1160,8 @@ public ChangeUndoEvent(UndoEngine engine, ComponentChangingEventArgs e, bool ser
                     _openComponent = e.Component;
                     _member = e.Member;
 
-                    UndoEngine.Trace("---> Creating change undo event for '{0}'", _componentName);
-                    UndoEngine.Trace("---> Saving before snapshot for change to '{0}'", _componentName);
+                    Debug.WriteLineIf(s_traceUndo.TraceVerbose, $"UndoEngine: ---> Creating change undo event for '{_componentName}'");
+                    Debug.WriteLineIf(s_traceUndo.TraceVerbose, $"UndoEngine: ---> Saving before snapshot for change to '{_componentName}'");
```
halgab, line 1164: Don't you think this line should be inside the `if` below?

dmitrii-drobotov: Nice catch, it looks like it should.

halgab: I'll update it then

dmitrii-drobotov (approved):

dreddy-work (approved):

PR #8771, opened 2023-03-09
Fix disabled text label rendering.

JeremyKuhne:
The problem was actually in ControlPaint.DrawStringDisabled that takes an IDeviceContext. In the big refactoring it lost the application of TextFormatFlags to decide whether or not to apply Graphics state. LinkLabel never set those flags historically so it didn't get the clipping it applied to the Graphics object applied to the HDC. Fix is to apply the flags properly.

I've also cleaned up the LinkLabel code and did a thorough pass of the TextExtensions to bring the code up to snuff.

Fixes https://github.com/dotnet/winforms/issues/7341

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8771)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ControlPaint.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ControlPaint.cs
@@ -2040,7 +2040,7 @@ namespace System.Windows.Forms
             // This must come before creating the scope.
             FONT_QUALITY quality = TextRenderer.FontQualityFromTextRenderingHint(dc);
 
-            using var hdc = new DeviceContextHdcScope(dc);
+            using var hdc = new DeviceContextHdcScope(dc, TextRenderer.GetApplyStateFlags(dc, format));
             DrawStringDisabled(hdc, s, font, color, layoutRectangle, format, quality);
         }
 
@@ -2312,7 +2312,7 @@ namespace System.Windows.Forms
         ///  Converts an infinite value to "1".
         /// </summary>
         private static float InfinityToOne(float value)
-            => value == float.NegativeInfinity || value == float.PositiveInfinity ? 1.0f : value;
+            => value is float.NegativeInfinity or float.PositiveInfinity ? 1.0f : value;
 
         /// <summary>
         ///  Creates a new color that is a object of the given color.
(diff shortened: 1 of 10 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ControlPaint.cs
```diff
@@ -2040,7 +2040,7 @@ public static void DrawSizeGrip(Graphics graphics, Color backColor, int x, int y
             // This must come before creating the scope.
             FONT_QUALITY quality = TextRenderer.FontQualityFromTextRenderingHint(dc);
 
-            using var hdc = new DeviceContextHdcScope(dc);
+            using var hdc = new DeviceContextHdcScope(dc, TextRenderer.GetApplyStateFlags(dc, format));
```
JeremyKuhne, line 2043: This is the root of the bug.

src/System.Windows.Forms/src/System/Windows/Forms/LinkLabel.cs
```diff
@@ -1302,13 +1094,12 @@ protected override void OnPaint(PaintEventArgs e)
 
                         PaintLinkBackground(g);
 
-                        // Because the code has been like that since long time, we assume that _textRegion is not null.
-                        g.IntersectClip(_textRegion!);
```
JeremyKuhne, line 1306: This is the only functional change, the rest is cleanup. I moved this inside of UseCompatibleTextRendering as it only really matters there.

JeremyKuhne:
Fixes #7341

elachlan:
> Fixes #7341

This will need to be added to the PR summary.

src/System.Windows.Forms/src/System/Windows/Forms/TextRenderer.cs
```diff
@@ -8,640 +8,632 @@
 using System.Drawing;
 using System.Drawing.Text;
 
-namespace System.Windows.Forms
+namespace System.Windows.Forms;
+
```
dreddy-work, line 12: why this change?

dreddy-work (approved):
RLGTM.

src/System.Windows.Forms/src/System/Windows/Forms/TextRenderer.cs
JeremyKuhne: Reduces nesting, which improves maintainability

lonitra (approved):
🚢

src/System.Windows.Forms/src/System/Windows/Forms/ControlPaint.cs
```diff
@@ -2555,98 +2555,30 @@ private static ColorMatrix RemapBlackAndWhitePreserveTransparentMatrix(Color rep
...
-            }
-            else
-            {
-                result = TextFormatFlags.Left;
-            }
-
-            return result;
-        }
-
-        internal static StringFormat StringFormatForAlignment(ContentAlignment align)
```
Tanya-Solyanik, line 2642: The removed functions were internal, are they not used anywhere?

Tanya-Solyanik (approved):

PR #8774, opened 2023-03-10
Fix: Do not use fixed index in the TreeNode.PrevNode if the TreeView is sorted

### LINKED ISSUE #8768: TreeNodeCollection.AddRange works incorrectly for non-empty collection of the sorted TreeView
### .NET version

6.0

No

No

### Issue description

### Steps to reproduce

```
namespace TreeViewTest
{
  internal static class Program
  {
    [STAThread]
    static void Main()
    {

      treeView.Sort();

      var form = new Form();
      form.Controls.Add(treeView);
        new TreeNode("3"),
        new TreeNode("5"),
        new TreeNode("1"),
        new TreeNode("4")
      });
      Application.Run(form);
    }
  }
}
```

This results in:
Clipboard01

snnz:
Fixes #8768

This corresponds to the following check in `TreeNodeCollection.AddInternal`, that results in the fixed index not being used for node insertion:

```
if (tv is not null && tv.Sorted)
{
    return owner.AddSorted(node);
}
```
## Regression? 

- No

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8774)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TreeNode.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TreeNode.cs
@@ -859,7 +859,11 @@ namespace System.Windows.Forms
 
                 if (fixedInd > 0)
                 {
-                    currentInd = fixedInd;
+                    TreeView tv = TreeView;
+                    if (tv is null || !tv.Sorted)
+                    {
+                        currentInd = fixedInd;
+                    }
                 }
 
                 if (currentInd > 0 && currentInd <= parent.Nodes.Count)


snnz:
@dotnet-policy-service agree

dreddy-work:
@snnz, thank you for sending this. Given this change impact existing application that may have dependency on the current behavior, We follow opt-in mechanism for the changes like this. You can refer to PR: https://github.com/dotnet/winforms/pull/8556 on how to add opt-in switch.

Please also include what kind of tests you performed (attach gifs if possible, in description) and we need to add unit tests around new/old behavior.

PR #8783, opened 2023-03-11
Fix ToolStripItems_FontScaling tests

gpetrou:
## Proposed changes

- Locally, the tests are failing for me and it looks like it is because we are using `DpiHelper.LogicalDpi` which is always `96.0`. I used `Form.DeviceDpi` instead.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8783)

### DIFF
--- a/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Dpi/ToolStripItemTests.Dpi.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Dpi/ToolStripItemTests.Dpi.cs
@@ -50,7 +50,7 @@ namespace System.Windows.Forms.UITests.Dpi
                 form.Show();
 
                 DpiMessageHelper.TriggerDpiMessage(User32.WM.DPICHANGED_BEFOREPARENT, toolStrip, newDpi);
-                var factor = newDpi / DpiHelper.LogicalDpi;
+                var factor = (float)newDpi / form.DeviceDpi;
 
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Dpi/ToolStripItemTests.Dpi.cs
```diff
@@ -50,7 +50,7 @@ public void ToolStripItems_FontScaling(int newDpi)
                 form.Show();
 
                 DpiMessageHelper.TriggerDpiMessage(User32.WM.DPICHANGED_BEFOREPARENT, toolStrip, newDpi);
-                var factor = newDpi / DpiHelper.LogicalDpi;
+                var factor = (float)newDpi / form.DeviceDpi;
```
dreddy-work, line 53: Tests should be running default 100% , which is 96 DPI on build agents. They might be failing locally because of your monitor settings. 

This is succeeding on your local machine only because Fomr doesn't receive DPI change with `DPICHANGED_BEFOREPARENT` message. 

If you plan on making similar changes, Make sure you differentiate messages `DPICHANGED_BEFOREPARENT` and `DPICHANGED` events.

This change is not necessary for build agents but help local. So taking it.

dreddy-work (approved):

PR #8784, opened 2023-03-11
Enable nullability in ComponentCache 

halgab:
Contributes to #1107

I did a bit of cleanup again in the second commit

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8784)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ComponentCache.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ComponentCache.cs
@@ -65,8 +65,6 @@ namespace System.ComponentModel.Design.Serialization
                 // it's a 1:1 relationship so we can go back from entry to  component (if it's not setup yet.. which should not happen, see ComponentCodeDomSerializer.cs::Serialize for more info)
                 if (_cache is not null && component is IComponent)
                 {
-                    ArgumentNullException.ThrowIfNull(value);
-
                     value.Component ??= component;
                     _cache[component] = value;
                 }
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ComponentCache.cs
```diff
@@ -120,39 +107,35 @@ internal bool ContainsLocalName(string name)
...
-        private void OnComponentChanging(object source, ComponentChangingEventArgs ce)
+        private void OnComponentChanging(object? source, ComponentChangingEventArgs ce)
         {
             if (_cache is not null)
             {
                 if (ce.Component is not null)
                 {
                     RemoveEntry(ce.Component);
 
-                    if (!(ce.Component is IComponent) && _serManager is not null)
```
halgab, line 151: `_serManager` is actually never null, so I removed the null checks in my 2nd commit

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ComponentCache.cs
```diff
@@ -258,68 +241,44 @@ internal struct ResourceEntry
         internal sealed class Entry
         {
             private readonly ComponentCache _cache;
```
halgab, line 243: This field is never used. I guess we could remove it, but I don't know if that's kept around for debugging purposes

dreddy-work: Safe to remove it.

dreddy-work (approved):

PR #8792, opened 2023-03-11
Add nullability annotations to DesignerActionList and DesignerActionListCollection

halgab:
Contributes to #1107

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8792)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionList.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionList.cs
@@ -27,7 +27,7 @@ namespace System.ComponentModel.Design
             return Component?.Site?.GetService(serviceType);
         }
 
-        public virtual DesignerActionItemCollection? GetSortedActionItems()
+        public virtual DesignerActionItemCollection GetSortedActionItems()
         {
             var items = new SortedList<string, DesignerActionItem>();
 
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionList.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionList.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.Reflection;
 
 namespace System.ComponentModel.Design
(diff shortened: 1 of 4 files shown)


src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionListCollection.cs
```diff
@@ -21,7 +19,7 @@ public DesignerActionListCollection(DesignerActionList[] value)
 
         public DesignerActionList this[int index]
```
halgab, line 20: This cannot be null as the following code would fail: https://github.com/dotnet/winforms/blob/6da50a7da9c96c217b564974bf63ce39fcec5509/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionUIService.cs#L100

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionList.cs
```diff
@@ -15,21 +13,21 @@ namespace System.ComponentModel.Design
...
+        public IComponent? Component { get; }
 
-        public object GetService(Type serviceType)
+        public object? GetService(Type serviceType)
         {
             return Component?.Site?.GetService(serviceType);
         }
 
-        public virtual DesignerActionItemCollection GetSortedActionItems()
+        public virtual DesignerActionItemCollection? GetSortedActionItems()
```
dreddy-work, line 30: where do you see this method returning null?

dreddy-work:
@halgab, since you have become acquainted with the majority of the codebase, would you be able to prioritize contributing to the issues that have been labeled as "help wanted"?. Team is prioritizing those issues over the nullability when reviewing PRs.

halgab:
Yes, sure.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionList.cs
halgab: I didn't. I figured I would mark this one as nullable as the type is public and the code using it inside the repo always seem to check it against null

dreddy-work: That shouldn't warrant to make this nullable. When we get to work on those callers, we may be able to change those conditions but if we know for sure this method doesn't return null, then better not to do it.

halgab: Fixed

dreddy-work (approved):

PR #8860, opened 2023-03-15
Fix DragDropTests

sharwell:
* Make sure to start a drag/drop operation on DragEnter (DragOver may or may not occur due to test speed)
* Match Mouse Down with Mouse Up
* Remove artificial delays

⚠️ **DO NOT MERGE**: The updated tests are being forced to run 100 times each to demonstrate stability in CI


src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ButtonTests.cs
```diff
@@ -266,7 +263,7 @@ public async Task Button_Mouse_Press_With_Drag_Off_Button_Does_Not_Cause_Button_
                     form,
                     inputSimulator => inputSimulator.Mouse
                         .LeftButtonDown()
-                        .MoveMouseTo(virtualPoint.X, virtualPoint.Y)
```
dreddy-work, line 269: Did you see `MoveMouseTo` was also problematic?

Issue #8802, opened 2023-03-10
[API Proposal]: Add ability to get modern Windows icons to System.Drawing.SystemIcons.


JeremyKuhne:

bartonjs:
We added a `None=0` value to `StockIconOptions`.

Assuming that the shorter overload is passing `StockIconOptions.None`, consider collapsing into one overload with a defaulted parameter.

Consider renaming `StockIcon` to `StockIconId`, just in case there would be a future situation warranting something like `public class StockIcon : Icon`.

```C#
namespace System.Drawing;

public static class SystemIcons
{
    /// <summary>
    ///  Gets the specified stock shell icon.
    /// </summary>
    public static Icon GetStockIcon(StockIcon stockIcon);
    public static Icon GetStockIcon(StockIcon stockIcon, StockIconOptions options);
}

[Flags]
public enum StockIconOptions
{
    None = 0,
    SmallIcon = 0x000000001,
    ShellIconSize = 0x000000004,
    LinkOverlay = 0x000008000,
    Selected = 0x000010000,
}

// Matching https://learn.microsoft.com/windows/win32/api/shellapi/ne-shellapi-shstockiconid with
// Pascal casing and no abbreviations ("Assoc" to "Associations"), will still keep acronyms (CD, not CompactDisc).
public enum StockIcon
{
    DocumentNoAssociation = 0,
    DocumentWithAssociation = 1,
    Application = 2,
    Folder = 3,
    OpenFolder = 4,
    Drive525 = 5,
    Drive35 = 6,
    DriveRemovable = 7,
    DriveFixed = 8,
    DriveNetwork = 9,
    DriveNetworkDisabled = 10,
    DriveCD = 11,
    DriveRAM = 12,
    World = 13,
    Server = 15,
    Printer = 16,
    // ...
}
```

JeremyKuhne:
Discussed further with the API review board, I'll collapse to a single API with a default parameter and rename `StockIconOptions.None` to `StockIconOptions.Default` with a description of what that means.

JeremyKuhne:
> When documenting this, please make it clear whether the caller should eventually Dispose().

@KalleOlaviNiemitalo I will. Given the large number of icon types and options caching them doesn't seem plausible, so disposing them will be clearly called out.

JeremyKuhne:
I've got the basic implementation in place, need to polish it up.

Amy-Li03:
[...]

PR #8878, opened 2023-03-21
Add API to get Windows stock icons.

JeremyKuhne:
Fixes: #8802

I also manually saved the generated Icons from the new API to see that it behaves as expected (`Icon.Save()`).

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8878)

### DIFF
--- a/src/System.Drawing.Common/src/System.Drawing.Common.csproj
+++ b/src/System.Drawing.Common/src/System.Drawing.Common.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFrameworks>$(NetCurrent);$(NetPrevious);$(NetMinimum);netstandard2.0;$(NetFrameworkMinimum)</TargetFrameworks>
     <TargetFramework />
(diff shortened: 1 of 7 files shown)


src/System.Drawing.Common/src/System.Drawing.Common.csproj
```diff
@@ -210,6 +210,8 @@ Since .NET 7, non-Windows platforms are not supported, even with the runtime con
     <Compile Include="System\Drawing\Printing\PrintPreviewGraphics.cs" />
     <Compile Include="System\Drawing\Printing\TriState.cs" />
     <Compile Include="System\Drawing\ScreenDC.cs" />
+    <Compile Include="System\Drawing\StockIconId.cs" />
```
RussKie, line 213: As a possible alternative to `#ifdef`s in the code is to conditionally include the source files in to the project here, e.g.:
```xml
<Compile Include="System\Drawing\StockIconId.cs" Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0'))" />
```

Or even like this:
```xml
  <ItemGroup Condition="$([MSBuild]::IsTargetFrameworkCompatible('$(TargetFramework)', 'net8.0'))">
    <Compile Include="System\Drawing\StockIconId.cs" />
    <Compile Include="System\Drawing\StockIconOptions.cs" />
  </ItemGroup>
```

RussKie (approved):

src/System.Drawing.Common/src/System.Drawing.Common.csproj
JeremyKuhne: @RussKie yeah, I considered it, but went with always including so we can see the code no matter what we're targeting.

dreddy-work (approved):

Shyam-Gupta (approved):

PR #8890, opened 2023-03-23
Add cleaner theory data helpers

JeremyKuhne:
For some common theory data (such as valid and invalid enum values) the current pattern is really unwieldly. I've added specific attributes for the "common" test data to make discovery and test writing easier.

Also moved some types down to the common test utility assembly. Intent is to make functionality available to all test projects, including the newly arrived System.Drawing.

I'll be iterating some more here.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8890)


src/Common/tests/TestUtilities/XUnit/BoolDataAttribute.cs
```diff
@@ -0,0 +1,22 @@
...
+// See the LICENSE file in the project root for more information.
+
+namespace Xunit;
+
+/// <summary>
+///  Generates <see cref="TheoryAttribute"/> data for true/false.
+/// </summary>
+public class BoolDataAttribute : CommonMemberDataAttribute
+{
+    private static readonly TheoryData<bool> _data = new();
```
JeremyKuhne, line 12: This is the start of the new data attributes. They all cache their data currently which should improve perf a bit. Memory pressure shouldn't be too bad, the only one where we might want to switch to a weak ref is the Enum data.

src/Common/tests/TestUtilities/XUnit/NumberDataAttribute.cs
```diff
@@ -0,0 +1,36 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Numerics;
+
+namespace Xunit;
+
+/// <summary>
+///  Generates <see cref="TheoryAttribute"/> data for a representative set of number values.
+/// </summary>
+public class NumberDataAttribute<TNumber> : CommonMemberDataAttribute where TNumber : IBinaryNumber<TNumber>, IMinMaxValue<TNumber>
```
JeremyKuhne, line 12: Using new math generics here so this works for all number types...

Shyam-Gupta (approved):
Looks great to me.

PR #8907, opened 2023-03-27
Add AppContext switch scope functionality

JeremyKuhne:
Adds the `BinaryFormatterScope` ref struct for explicitly setting `BinaryFormatter` enabled state.

Ran through System.Windows.Forms.Tests with the `BinaryFormatter` explicitly disabled to identify tests that were dependent and add the context to said tests.

The intent is to follow this change with additional changes to do the same for the other test assemblies. I'll be using what I find here to open additional issues or add additional context to existing ones. For example, it wasn't clear that CSV text uses the `BinaryFormatter` for the clipboard until I ran through this set of tests.

After all assemblies are scoped I'll be then looking to close test gaps.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8907)

### DIFF
--- a/src/Common/tests/TestUtilities/AppContextSwitchNames.cs
+++ b/src/Common/tests/TestUtilities/AppContextSwitchNames.cs
@@ -1,0 +1,22 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.Serialization.Formatters.Binary;
+
+namespace System;
+
+public static class AppContextSwitchNames
+{
+    /// <summary>
+    ///  The switch that controls whether or not the <see cref="BinaryFormatter"/> is enabled.
+    /// </summary>
+    public static string EnableUnsafeBinaryFormatterSerialization { get; }
+        = "System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization";
+
+    /// <summary>
+    ///  Switch that controls <see cref="AppContext"/> switch caching.
+    /// </summary>
+    public static string LocalAppContext_DisableCaching { get; }
+        = "TestSwitch.LocalAppContext.DisableCaching";
+}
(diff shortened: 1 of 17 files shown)


JeremyKuhne:
cc: @GrabYourPitchforks

src/Common/tests/TestUtilities/AppContextSwitchScope.cs
```diff
@@ -0,0 +1,49 @@
...
+        if (!AppContext.TryGetSwitch(AppContextSwitchNames.LocalAppContext_DisableCaching, out bool isEnabled)
+            || !isEnabled)
+        {
+            // It doesn't make sense to try messing with AppContext switches if they are going to be cached.
+            throw new InvalidOperationException("LocalAppContext switch caching is not disabled.");
+        }
+
+        AppContext.TryGetSwitch(switchName, out _originalState);
+
+        AppContext.SetSwitch(switchName, enable);
```
dreddy-work, line 32: We should you test if switch value is different from `_originalState` before proceeding with setting it.

src/Common/tests/TestUtilities/AppContextSwitchNames.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Runtime.Serialization.Formatters.Binary;
+
+namespace System;
+
+public static class AppContextSwitchNames
+{
+    /// <summary>
+    ///  The switch that controls whether or not the <see cref="BinaryFormatter"/> is enabled.
+    /// </summary>
+    public static string EnableUnsafeBinaryFormatterSerialization { get; }
```
dreddy-work, line 14: Next, should we move all our appcontext switches to this?

JeremyKuhne: Testing them, yes. Putting the string here only makes sense for things that are externally defined.

src/Common/tests/TestUtilities/AppContextSwitchScope.cs
JeremyKuhne: We'd have to check whether or not we actually got the value to begin with. The default is going to be false. I don't think the extra conditional complexity is worth it- the cost of calling this is trivial.

dreddy-work (approved):

PR #8917, opened 2023-03-29
Fix caching of DpiAwarenessContext for Mixed mode DPI hosting scenarios.

dreddy-work:
The `DpiAwarenessContext `is utilized to establish the desired `DpiAwarenessContext ` for a control from the moment it is initialized and is then applied when creating the control's handle. However, the current implementation of the  `DpiAwarenessContext `property evaluates the DpiContext based on the thread's current DPI awareness context. This may not accurately reflect the DPI awareness context that was initially desired for the control, particularly in scenarios where the application is running with mixed-mode DPI hosting and threads DPI awareness context is altered programmatically throughout the application.

Added unit test.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8917)

### DIFF
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/MixedDpiHostingTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/MixedDpiHostingTests.cs
@@ -9,6 +9,31 @@ namespace System.Windows.Forms.Tests
     public class MixedDpiHostingTests
     {
         [WinFormsFact]
+        public void MixedDpiHosting_CacheDpiAwarenessTest()
+        {
+            // Run tests only on Windows 10 versions that support thread dpi awareness.
+            if (!PlatformDetection.IsWindows10Version1803OrGreater)
+            {
+                return;
+            }
+
+            // Set thread awareness context to PerMonitorV2(PMv2). If process/thread is not in PMv2, calling 'EnterDpiAwarenessScope' is a no-op and that is by design.
+            // In this case, we will be setting thread to PMv2 mode and then scope to UNAWARE/SYSTEMAWARE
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/MixedDpiHostingTests.cs
```diff
@@ -8,6 +8,31 @@ namespace System.Windows.Forms.Tests
...
+        {
+            // Run tests only on Windows 10 versions that support thread dpi awareness.
+            if (!PlatformDetection.IsWindows10Version1803OrGreater)
+            {
+                return;
+            }
+
+            // Set thread awareness context to PerMonitorV2(PMv2). If process/thread is not in PMv2, calling 'EnterDpiAwarenessScope' is a no-op and that is by design.
+            // In this case, we will be setting thread to PMv2 mode and then scope to UNAWARE/SYSTEMAWARE
+            PInvoke.SetThreadDpiAwarenessContextInternal(DPI_AWARENESS_CONTEXT.DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2);
```
JeremyKuhne, line 22: Shouldn't we be putting the thread context back the way it was?

JeremyKuhne (changes requested):
This doesn't look like it does anything. Perhaps I'm missing something.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/MixedDpiHostingTests.cs
dreddy-work: > This doesn't look like it does anything. Perhaps I'm missing something.

You are right Jeremy. For some reason, i thought it was happening inside getter. Taking this down.

PR #8921, opened 2023-03-30
InputLanguage: Use RegLoadMUIString instead of SHLoadIndirectString API

DJm00n:
I have extracted part of #8573 into separate PR until decision is made on some aspects of that PR.

## Proposed changes

- Use `RegLoadMUIString` instead of the `SHLoadIndirectString` API as recommended at https://learn.microsoft.com/windows/win32/intl/locating-redirected-strings#load-a-language-neutral-registry-value
- Move InputLanguage `GetKeyboardLayoutNameForHKL` method implementation to the internal `LayoutId` property (to ease debugging)
- Added proper testing of InputLanguage.LayoutName
- Added InputLanguage_Culture_ThrowsArgumentException and InputLanguage_LayoutName_UnknownExpected tests
- Use `PARAM.ToInt` instead of casts in `GetHashCode` implementation

## Customer Impact

- Shouldn't have any visible customer impact

## Regression? 

- No

## Risk

- Little

## Test methodology 

Auto test:
- InputLanguage_InputLanguageLayoutId_Expected (updated)
- InputLanguage_Culture_ThrowsArgumentException (new)
- InputLanguage_LayoutName_UnknownExpected  (new)

Manually:
- Install "Canadian Multilingual Standard"  keyboard layout in Windows.
- Add breakpoint in `InputLanguage_InstalledInputLanguages_Get_ReturnsExpected` test and run it under debugger.
- Check that its `LayoutId` is `00011009` under debugger.
- Check that its `LayoutName` is "Canadian Multilingual Standard".

image

image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8921)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/Windows/Win32/Pinvoke.RegLoadMUIString.cs
+++ b/src/System.Windows.Forms.Primitives/src/Windows/Win32/Pinvoke.RegLoadMUIString.cs
@@ -1,0 +1,47 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.CompilerServices;
+using Microsoft.Win32;
+
+namespace Windows.Win32;
+
+internal static partial class PInvoke
+{
+    /// <inheritdoc cref="RegLoadMUIString(System.Registry.HKEY, string, PWSTR, uint, uint*, uint, string)"/>
+    [SkipLocalsInit]
+    public static unsafe bool RegLoadMUIString(RegistryKey key, string keyName, out string localizedValue)
+    {
+        using BufferScope<char> buffer = new(stackalloc char[128]);
+
+        while (true)
+        {
+            fixed (char* pszOutBuf = buffer)
+            {
+                uint bytes = 0;
+                var errorCode = RegLoadMUIString(
+                    (System.Registry.HKEY)key.Handle.DangerousGetHandle(),
+                    keyName,
+                    pszOutBuf,
+                    (uint)(buffer.Length * sizeof(char)),
+                    &bytes,
+                    0,
+                    null);
+
+                // The buffer is too small. Try again with a larger buffer.
+                if (errorCode == WIN32_ERROR.ERROR_MORE_DATA)
+                {
+                    buffer.EnsureCapacity((int)(bytes / sizeof(char)));
+                    continue;
+                }
+
+                localizedValue = errorCode == WIN32_ERROR.ERROR_SUCCESS
+                    ? buffer[..Math.Max((int)(bytes / sizeof(char)) - 1, 0)].ToString()
+                    : string.Empty;
+
+                return errorCode == WIN32_ERROR.ERROR_SUCCESS;
+            }
+        }
+    }
+}
--- a/src/System.Windows.Forms.Primitives/src/NativeMethods.txt
+++ b/src/System.Windows.Forms.Primitives/src/NativeMethods.txt
@@ -505,6 +505,7 @@ RegisterClass
 RegisterClipboardFormat
 RegisterDragDrop
 RegisterWindowMessage
+RegLoadMUIString
 ReleaseCapture
 ReleaseStgMedium
 RestoreDC
@@ -578,7 +579,6 @@ SHGetSpecialFolderLocation
 ShowCaret
 ShowCursor
 SHGetSpecialFolderLocation
-SHLoadIndirectString
 ShowWindow
 SHParseDisplayName
 SIATTRIBFLAGS
--- a/src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/InputLanguage.cs
@@ -17,7 +17,7 @@ public sealed class InputLanguage
     /// <summary>
     ///  The HKL handle.
     /// </summary>
-    private readonly IntPtr _handle;
+    private readonly nint _handle;
 
     internal InputLanguage(IntPtr handle)
     {
@@ -94,7 +94,7 @@ public sealed class InputLanguage
         }
     }
 
-    private static string KeyboardLayoutsRegistryPath => @"SYSTEM\CurrentControlSet\Control\Keyboard Layouts";
+    private static string s_keyboardLayoutsRegistryPath => @"SYSTEM\CurrentControlSet\Control\Keyboard Layouts";
 
     /// <summary>
     ///  Returns the name of the current keyboard layout as it appears in the Windows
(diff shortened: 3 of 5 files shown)


DJm00n:
@JeremyKuhne any objections on this topic?

src/System.Windows.Forms.Primitives/src/Windows/Win32/Pinvike.RegLoadMUIString.cs
```diff
@@ -0,0 +1,34 @@
...
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Win32;
+
+namespace Windows.Win32
+{
+    internal static partial class PInvoke
+    {
+        /// <inheritdoc cref="RegLoadMUIString(System.Registry.HKEY, string, PWSTR, uint, uint*, uint, string)"/>
+        public static unsafe bool RegLoadMUIString(RegistryKey key, string keyName, out string localizedValue)
```
JeremyKuhne, line 12: This looks like something that we should create an API proposal for in the runtime. `RegistryKey.GetMUIString()` maybe?

src/System.Windows.Forms.Primitives/src/Windows/Win32/Pinvike.RegLoadMUIString.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            localizedValue = string.Empty;
+            if (key.Handle.IsInvalid)
```
JeremyKuhne, line 15: I would expect that this is uncommon, maybe better to just let if fail with the call to RegLoadMUIString().

src/System.Windows.Forms.Primitives/src/Windows/Win32/Pinvike.RegLoadMUIString.cs
```diff
@@ -0,0 +1,34 @@
...
+                return false;
+
+            var hkey = (System.Registry.HKEY)key.Handle.DangerousGetHandle();
+            uint bytes = 0;
+            if (RegLoadMUIString(hkey, keyName, null, 0, &bytes, 0, null) != WIN32_ERROR.ERROR_MORE_DATA)
+                return false;
+
+            Span<char> buffer = stackalloc char[(int)bytes / sizeof(char)];
```
JeremyKuhne, line 23: The data here appears to be unbounded. One presumes that this could be pretty expensive if it follows through to a PE file to load a string resource just to return the length.

The safer/better choice is to stack allocate some reasonable buffer and then rent/return a bigger array from the `ArrayPool` if the size isn't large enough. It may also make sense to just use the `ArrayPool`, as I the benefit of skipping the rental is probably minimal at best. Just using the `ArrayPool` would make the logic simpler, for sure.

JeremyKuhne (changes requested):
Comments are inline.

src/System.Windows.Forms.Primitives/src/Windows/Win32/Pinvike.RegLoadMUIString.cs
JeremyKuhne: To be clear- I don't think we should block here, but I would assume this functionality is generally useful for our developers.

src/System.Windows.Forms.Primitives/src/Windows/Win32/Pinvike.RegLoadMUIString.cs
DJm00n: >One presumes that this could be pretty expensive if it follows through to a PE file to load a string resource just to return the length.

I can bet that RegLoadMUIString is caching MUI strings to avoid this rereading but I'll do the changes.

JeremyKuhne: > I can bet that RegLoadMUIString is caching MUI strings to avoid this rereading but I'll do the changes.

It doesn't look like it. You'll probably get some general IO caching though. The more critical issue is that the buffer size is unbounded, which is a problem for stack allocation.

src/System.Windows.Forms.Primitives/src/Windows/Win32/Pinvike.RegLoadMUIString.cs
DJm00n: @JeremyKuhne I have added `RegistryKeyExtensions.GetMUIString` extension method

src/System.Windows.Forms.Primitives/src/Windows/Win32/Pinvoke.RegLoadMUIString.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    internal static partial class PInvoke
+    {
+        /// <inheritdoc cref="RegLoadMUIString(System.Registry.HKEY, string, PWSTR, uint, uint*, uint, string)"/>
+        public static unsafe bool RegLoadMUIString(RegistryKey key, string keyName, out string localizedValue)
+        {
+            uint bytes;
+            WIN32_ERROR error;
+            var hkey = (System.Registry.HKEY)key.Handle.DangerousGetHandle();
+            char[] buffer = ArrayPool<char>.Shared.Rent(MAX_PATH + 1);
```
DJm00n, line 18: `FolderNameEditor.FolderBrowser.cs` is using `MAX_PATH` for `displayName` buffer too...

src/System.Windows.Forms.Primitives/src/Windows/Win32/Pinvoke.RegLoadMUIString.cs
```diff
@@ -0,0 +1,41 @@
...
+                fixed (char* pszOutBuf = buffer)
+                {
+                    error = RegLoadMUIString(hkey, keyName, pszOutBuf, (uint)(buffer.Length * sizeof(char)), &bytes, 0, null);
+                }
+            }
+
+            localizedValue = new string(buffer, 0, Math.Max((int)(bytes / sizeof(char)) - 1, 0));
+            ArrayPool<char>.Shared.Return(buffer);
+
+            return HRESULT.HRESULT_FROM_WIN32(error).Succeeded;
```
JeremyKuhne, line 38: You should only try to create the string when `ERROR_SUCCESS` is returned and that should be the only thing you check to return success.

JeremyKuhne (changes requested):
Added a few more comments.

DJm00n:
Hi @JeremyKuhne! Do you have any updates on this review?

JeremyKuhne:
@DJm00n I'm in the process of adding a `BufferScope` that makes writing this sort of thing easier. #8999

It would be good to use that to simplify this if possible. You should be able to do something similar to what I did with PInvoke.GetModuleFileNameLongPath.cs in that change.

Hopefully I'll have it checked in this afternoon.

DJm00n:
> GetModuleFileNameLongPath
Nice! This is indeed looks more simple!

Pushed updated version of `PInvoke.RegLoadMUIString`.

DJm00n:
Hi @JeremyKuhne! Any updates on this review?

JeremyKuhne (approved):
Looks great- thanks for your patience!

dreddy-work:
@DJm00n, can you rebase this?

DJm00n:
@dreddy-work done

dreddy-work (approved):

PR #8922, opened 2023-03-30
Remove use of deprecated LoadCursor/LoadIcon Win32 APIs

DJm00n:
[LoadCursor](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-loadcursorw)/[LoadIcon](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-loadiconw) Win32 APIs are superseded by [LoadImage](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-loadimagew). So just use a new function instead of them inplace. 

## Proposed changes

- Remove use of legacy LoadCursor/LoadIcon Win32 APIs and use LoadImage instead

## Customer Impact

- No impact

## Regression? 

- No

## Risk

- Little

## Test methodology

- Run tests

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8922)


DJm00n:
@dreddy-work youre right, as I understand - `superseded by` is same as `replaced by`. Or not?

elachlan:
@DJm00n can you please add an overload for `LoadImage` with `name` and `type` parameters? That would reduce the code needed in the usages of it.

You add it in `System.Windows.Forms.Primitives`.

DJm00n:
@JeremyKuhne alright then. Closing.

PR #8924, opened 2023-03-31
Use CodeAnalysis.test.globalconfig only for overrides

sharwell:
All duplicate items have been removed, leaving only the unique configuration for tests. The specific configuration behavior is not changed.

### DIFF
--- a/eng/CodeAnalysis.test.globalconfig
+++ b/eng/CodeAnalysis.test.globalconfig
@@ -1,271 +1,26 @@
 is_global = true
 
-# AD0001: Analyzer threw an exception
-dotnet_diagnostic.AD0001.severity = suggestion
-
-# CA1000: Do not declare static members on generic types
-dotnet_diagnostic.CA1000.severity = none
-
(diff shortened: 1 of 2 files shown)


gpetrou:
@sharwell we are following what the runtime repository does. I see that https://github.com/dotnet/runtime/blob/e26e33f98a3c24640c48f35a1b183abe6aff37e7/eng/CodeAnalysis.test.globalconfig has not changed. Is the plan to change that as well?

RussKie (commented):
👍 

Is this all manually reviewed and changed, or is there some tool that automates the de-duplication process?

dreddy-work (approved):

sharwell:
> Is this all manually reviewed and changed, or is there some tool that automates the de-duplication process?

I just used Beyond Compare and removed lines that were identical.

sharwell:
> Is the plan to change that as well?

I wasn't planning to (or aware of the similarity), but it could certainly be done.

PR #8925, opened 2023-03-31
Fix nullability of ArgumentValidation class

halgab:
## Proposed changes

The null-checking extension methods defined in ArgumentValidation are lacking an annotation stating that the first argument is checked for null. This PR adds it back.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8925)


halgab:
I noticed this while trying to replace [this method](https://github.com/dotnet/winforms/blob/f5164e780e255e1ce49bb26971450b500063997e/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/NonNullCollection.cs#L136) with calls to `OrThrowIfNull`. If the team is interested, I can push those changes to this PR or in a follow-up.

RussKie (dismissed):

dmitrii-drobotov (approved):

dmitrii-drobotov (approved):

Issue #8535, opened 2023-01-29
DataGridView TopLeftHeaderCell visibility related problems


kirsan31:
### .NET version

All

### Did it work in .NET Framework?

No

### Did it work in any of the earlier releases of .NET Core or .NET 5+?

No

### Issue description

Comes from https://github.com/dotnet/winforms/issues/6926#issuecomment-1236143035 and related discussion. _Sorry had no time to raised it in time and then completely forgot about it_ 😔

- As far as I can tell, **all** `DataGridView` [Flaky tests](https://github.com/dotnet/winforms/issues/7799) have the same root. The `TopLeftHeaderCell` is not visible if we have no columns at all, but `DataGridView` still treat it as visible (still trying to Invalidate it) :(
- Also we have another non consistent behavior here -`DataGridView` without columns (left) look different from `DataGridView` with all invisible columns (right):
![image](https://user-images.githubusercontent.com/17767561/188275713-67683082-9c15-4050-a8b8-892c16004f83.png)
  If we have no columns at all `TopLeftHeaderCell` not visible and if we have columns but they all set as not visible, `TopLeftHeaderCell` is visible 🤔 Also we have third variant [bound `DataGridView` with all invisible columns](https://github.com/dotnet/winforms/issues/4807) 😫

- Another bunch of bugs related to visibility of `TopLeftHeaderCell`. In many places it's checked like `DataGridView.LayoutInfo.TopLeftHeader != Rectangle.Empty`. And `DataGridView.LayoutInfo.TopLeftHeader` updated only in `PerformLayoutPrivate`.  Which is not always called when necessary. For example if we have `DataGridView` with `AutoSize == true` and call `ColumnHeadersVisible = false` then we will get `TopLeftHeaderCell.Visible == false` and `TopLeftHeaderCell.Displayed == true` because `PerformLayoutPrivate` will not be called if `AutoSize == true` 😕

It seems that the option of the absence of columns was not considered in the context of `TopLeftHeaderCell`. I see 2 ways here:
[...]

PR #8574, opened 2023-02-06
Removed extra validation in flaky DataGridView unit tests.

dkazennov:
Fixes #7799 (the main Issue containing a complete list of flaky tests)
Fixes #6597
Fixes #6739
Fixes #6926 (along with #8535)

## Proposed changes
- This fix omits check for `invalidatedCallCount` in unstable unit tests with a commentary reference to this Issue to keep a track record (as [proposed](https://github.com/dotnet/winforms/issues/6926#issuecomment-1246567369) by @RussKie).
- Theory data objects/inline data sets changed as well.

**Tests changed:**

- DataGridView_ColumnHeadersHeightSizeMode_SetNonResizeThenResize_RestoresOldValue
- DataGridView_ColumnHeadersHeightSizeMode_SetWithHandle_GetReturnsExpected
- DataGridView_ColumnHeadersHeight_SetWithHandle_GetReturnsExpected
- DataGridView_ColumnHeadersHeight_SetWithParentWithHandle_GetReturnsExpected
- DataGridView_OnColumnHeadersHeightChanged_InvokeWithHandle_CallsColumnHeadersHeightChanged
- DataGridView_OnColumnHeadersHeightSizeModeChanged_InvokeWithHandle_CallsColumnHeadersHeightSizeModeChanged
- DataGridView_OnRowHeadersWidthChanged_InvokeWithHandle_CallsRowHeadersWidthChanged
- DataGridView_OnRowHeadersWidthSizeModeChanged_InvokeWithHandle_CallsRowHeadersWidthSizeModeChanged
- DataGridView_Parent_SetWithHandle_GetReturnsExpected
- DataGridView_RowHeadersWidthSizeMode_SetNonResizeThenResize_RestoresOldValue
- DataGridView_RowHeadersWidthSizeMode_SetWithHandle_GetReturnsExpected
- DataGridView_RowHeadersWidth_SetWithHandle_GetReturnsExpected
- DataGridView_RowHeadersWidth_SetWithParentWithHandle_GetReturnsExpected
- DataGridView_TopLeftHeaderCell_GetWithHandle_ReturnsExpected
- DataGridView_TopLeftHeaderCell_SetWithHandle_GetReturnsExpected

## Customer Impact

- No impact. This fix affects only unit tests.

## Regression? 

- No

## Risk

- Minimal

## Test methodology 

- Unit tests.

## Test environment(s) 

- dotnet 8.0.100-alpha.1.22607.6

[...]

Issue #7799, opened 2022-09-18
DataGridView Flaky tests due to invalidatedCallCount


kirsan31:
### Test

DataGridView_ColumnHeadersHeightSizeMode_SetNonResizeThenResize_RestoresOldValue
DataGridView_ColumnHeadersHeightSizeMode_SetWithHandle_GetReturnsExpected
DataGridView_ColumnHeadersHeight_SetWithHandle_GetReturnsExpected
DataGridView_ColumnHeadersHeight_SetWithParentWithHandle_GetReturnsExpected
DataGridView_OnColumnHeadersHeightChanged_InvokeWithHandle_CallsColumnHeadersHeightChanged
DataGridView_OnColumnHeadersHeightSizeModeChanged_InvokeWithHandle_CallsColumnHeadersHeightSizeModeChanged
DataGridView_OnRowHeadersWidthChanged_InvokeWithHandle_CallsRowHeadersWidthChanged
DataGridView_OnRowHeadersWidthSizeModeChanged_InvokeWithHandle_CallsRowHeadersWidthSizeModeChanged
DataGridView_Parent_SetWithHandle_GetReturnsExpected
DataGridView_RowHeadersWidthSizeMode_SetNonResizeThenResize_RestoresOldValue
DataGridView_RowHeadersWidthSizeMode_SetWithHandle_GetReturnsExpected
DataGridView_RowHeadersWidth_SetWithHandle_GetReturnsExpected
DataGridView_RowHeadersWidth_SetWithParentWithHandle_GetReturnsExpected
DataGridView_TopLeftHeaderCell_GetWithHandle_ReturnsExpected
DataGridView_TopLeftHeaderCell_SetWithHandle_GetReturnsExpected

### Issue description

All these tests may randomly fail due to extra invalidations. This happens when the mouse pointer is above any cell of DataGridView. We can't properly fix this now because of:
- UTs are run in parallel therefore we can't set the mouse pointer and guarantee that it will not be moved until the test is done.
- DataGridView without columns does not visually display the left top header, but still tries to invalidate it's area. https://github.com/dotnet/winforms/issues/8535

Affected issues:
https://github.com/dotnet/winforms/issues/6597 It's hard to say if they are all, but probably yes.
https://github.com/dotnet/winforms/issues/6739
https://github.com/dotnet/winforms/issues/6926

merriemcgaw:
[...]

PR #7031, opened 2022-04-16
DataGridView flaky tests.

kirsan31:
Attempt to investigate DGV flaky tests. https://github.com/dotnet/winforms/issues/6926 https://github.com/dotnet/winforms/issues/6739

Both tests fail due to extra invalidate. To investigate further we need to understand 2 things:

- [x] [Couse of invalidate - call stack](https://github.com/dotnet/winforms/pull/7031#issuecomment-1101339968).
- [x] Input parameters pattern: `rowHeadersVisible: True` / `columnHeadersVisible: True`

Fails by run №:

11: [Windows_x86-xunit
System.Windows.Forms.Tests.ToolStripItemTests.ToolStripItem_ImageKey_ShouldSerializeValueWithOwnerWithImageList_Success](https://dev.azure.com/dnceng/public/_build/results?buildId=1722813&view=ms.vss-test-web.build-test-results-tab&runId=46748528&resultId=110931&paneView=debug)
27: [Windows_arm64-xunit
System.Windows.Forms.Tests.DataGridViewTests.DataGridView_DefaultCellStyles_Rendering](https://dev.azure.com/dnceng/public/_build/results?buildId=1723543&view=ms.vss-test-web.build-test-results-tab&runId=46761178&resultId=106555&paneView=debug)
**29**: [Windows_x64-xunit
System.Windows.Forms.Tests.DataGridViewTests.DataGridView_RowHeadersWidth_SetWithHandle_GetReturnsExpected](https://dev.azure.com/dnceng/public/_build/results?buildId=1723590&view=ms.vss-test-web.build-test-results-tab&runId=46762248&resultId=104683&paneView=debug). Got it!

The problem is in [random cursor position](https://github.com/dotnet/winforms/pull/7031#issuecomment-1101339968) during tests.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/7031)

PR #8927, opened 2023-03-31
Do not reset cursor position during test discovery

sharwell:
😑

### DIFF
--- a/src/System.Windows.Forms/tests/UnitTests/ModuleInitializer.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/ModuleInitializer.cs
@@ -16,12 +16,6 @@ namespace System.Windows.Forms.Tests
             // handling of a Windows message.
             AppContext.SetSwitch("System.Windows.Forms.DoNotCatchUnhandledExceptions", isEnabled: true);
 
-            // Some controls have behavior that changes when the mouse is over them. Make sure we start with the cursor
-            // at a known position.
-            //
-            // See https://github.com/dotnet/winforms/pull/7031#issuecomment-1101339968 for an example of this.
-            Cursor.Position = default;
-
             Application.EnableVisualStyles();
         }
     }
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewTests.cs
@@ -11,6 +11,15 @@ namespace System.Windows.Forms.Tests
 {
     public partial class DataGridViewTests
     {
+        public DataGridViewTests()
+        {
+            // Some controls have behavior that changes when the mouse is over them. Make sure we start with the cursor
+            // at a known position.
+            //
+            // See https://github.com/dotnet/winforms/pull/7031#issuecomment-1101339968 for an example of this.
+            Cursor.Position = default;
+        }
+
         [WinFormsFact]
         public void DataGridView_Ctor_Default()
         {


src/System.Windows.Forms/tests/UnitTests/ModuleInitializer.cs
```diff
@@ -16,12 +16,6 @@ public static void InitializeModule()
             // handling of a Windows message.
             AppContext.SetSwitch("System.Windows.Forms.DoNotCatchUnhandledExceptions", isEnabled: true);
 
-            // Some controls have behavior that changes when the mouse is over them. Make sure we start with the cursor
```
dreddy-work, line 19: @sharwell, can you add few details on what you observed with this and how it was interfering with tests?

dreddy-work:
@JeremyKuhne as FYI.

src/System.Windows.Forms/tests/UnitTests/ModuleInitializer.cs
sharwell: The mouse randomly moved while editing in winforms solution.

kirsan31:
Currently the change in `DataGridViewTests.cs` is not necessary. I'll try to help a little with what's going on here. 

- Originally cursor position was changed in [tests/TestUtilities/ThreadExceptionFixture.cs](https://github.com/dotnet/winforms/pull/7031/files) for flaky DataGridView tests. But this attempt was no success (because UTs are run in parallel).
- Then https://github.com/dotnet/winforms/issues/7799 was created. _And due to uncontrollable circumstances, I dropped out of this process for a long time :(_
- In https://github.com/dotnet/winforms/pull/8574 these tests were removed but cursor position in `ThreadExceptionFixture.cs` was missed :(
- @JeremyKuhne removed `ThreadExceptionFixture.cs` in https://github.com/dotnet/winforms/pull/8695 and transfer it's logic to `UnitTests/ModuleInitializer.cs` with some modification.

For now we have no cursor dependent tests in `DataGridViewTests.cs` so no need to set cursor position there.

> The Cursor tests were the ones modifying the cursor position and causing intermittent errors in other tests. Moved them to their own collection and forced them to run sequentially.

This is very cool, because similar will be needed if we try to fix https://github.com/dotnet/winforms/issues/8535

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewTests.cs
```diff
@@ -11,6 +11,15 @@ namespace System.Windows.Forms.Tests
 {
     public partial class DataGridViewTests
     {
+        public DataGridViewTests()
+        {
+            // Some controls have behavior that changes when the mouse is over them. Make sure we start with the cursor
+            // at a known position.
+            //
+            // See https://github.com/dotnet/winforms/pull/7031#issuecomment-1101339968 for an example of this.
+            Cursor.Position = default;
```
kirsan31, line 20: I think that this change is not necessary. Pls see my [comment](https://github.com/dotnet/winforms/pull/8927#issuecomment-1492893171) in conversation.

sharwell:
@kirsan31 I'm probably not the best one to comment on the need for this. For me, the important part is just the removal from the module initializer. As long as the mouse stops jumping around randomly while I'm trying to work, the rest can be anything. It's not clear if you want the removal of this call to be a follow-up, or if you want me to combine the fixing of the module initializer with the separate discussion of the need for cursor controls.

Tanya-Solyanik (approved):

kirsan31:
@sharwell 

> I'm probably not the best one to comment on the need for this. For me, the important part is just the removal from the module initializer. As long as the mouse stops jumping around randomly while I'm trying to work, the rest can be anything.

+1
> It's not clear if you want the removal of this call to be a follow-up, or if you want me to combine the fixing of the module initializer with the separate discussion of the need for cursor controls.

I just wanted to clarify how and why original code (in ThreadExceptionFixture.cs) appeared.
And point out that it should have been removed in https://github.com/dotnet/winforms/pull/8574 (this pr was before JeremyKuhne's one).
If this code had remained unchanged until now, I would definitely say that it should 100% be removed.  But @JeremyKuhne  modified it, so now he has to make a decision based on what I wrote above.

KlausLoeffelmann (approved):

JeremyKuhne:
@kirsan31 as there is no guarantee where the mouse will be when starting to run tests I'd love to get this back into a pre-run setup. I don't know how plausible this sort of thing is utilizing xunit functionality. I also don't know if it is plausible to detect that we're in discovery mode in the module initializer.

Personally I like the idea of adding analyzers to the unit tests that flag when we're doing global state things like repositioning the mouse as these sorts of things can be hard to find from intermittently failing tests.

If only there were more time...

sharwell:
@JeremyKuhne My recommendation for that case is to detect the presence of CI execution via an environment variable, and not move the cursor at all for non-CI scenarios.

PR #8934, opened 2023-04-03
Disable BinaryFormatter for most test assemblies

JeremyKuhne:
We want to know when we have additional scenarios that are impacted by the BinaryFormatter, so disabling it by default in tests.

Adds a few formatter scopes around tests that do use BinaryFormatter.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8934)


dreddy-work (approved):

Tanya-Solyanik:
@JeremyKuhne , @dreddy-work  - should we have some tests that still run with binary formatter "on"? Some applications would want to be compatible with downlevel?

JeremyKuhne:
@Tanya-Solyanik the `BinaryFormatterScope` lets you explicitly set the enabled state for the test. With it disabled by default it will flag that a new test actually has a `BinaryFormatter` dependency.

PR #8935, opened 2023-04-03
Handle a few skipped test failures in System.Drawing

JeremyKuhne:
`[KnownIssue]` doesn't actually skip tests in the Test Explorer. This change fixes the tests that were "skipped" and actually failing.

- GDI+ no longer accepts negative pen widths
- Some tests were ported from Mono and skipped as they validated Mono's behavior, not GDI+
- Graphics.CopyFromScreen would call ReleaseHDC even if GetHDC was not called.
- CopyFromScreen assumed -1, -1 was an invalid screen coordinate. Not true for multimonitor.
- SetNativeBrush tests should not be disposing the invalid handle- random badness will occur.
- Prevent using a Graphics object with a null pointer in Font.GetHeight (to avoid AV).

I've also disabled the BinaryFormatter by default for these tests.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8935)

### DIFF
--- a/src/System.Drawing.Common/tests/System.Drawing.Common.Tests.csproj
+++ b/src/System.Drawing.Common/tests/System.Drawing.Common.Tests.csproj
@@ -6,6 +6,7 @@
     <!-- Code warnings that weren't enabled in dotnet/runtime but are raised in winforms.
          TODO: Clean the code up and remove the NoWarns. -->
     <NoWarn>$(NoWarn);CA1825;CA5351;CA1850</NoWarn>
+    <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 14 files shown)


src/System.Drawing.Common/tests/System.Drawing.Common.Tests.csproj
```diff
@@ -6,6 +6,7 @@
     <!-- Code warnings that weren't enabled in dotnet/runtime but are raised in winforms.
          TODO: Clean the code up and remove the NoWarns. -->
     <NoWarn>$(NoWarn);CA1825;CA5351;CA1850</NoWarn>
+    <GenerateRuntimeConfigurationFiles>true</GenerateRuntimeConfigurationFiles>
```
dreddy-work, line 9: Do you need this setting? shouldn't it generate runtimeconfig.json in the output directory by default?

JeremyKuhne: You need it for dlls.

PR #8936, opened 2023-04-04
Disable ServerManager and minimize all open windows on test machines.

dreddy-work:
Infra changes to improve UI tests reliability.

- Stop ServerManager.exe process.
- Minimize all windows on the test machine before launching Integration tests.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8936)

### DIFF
--- a/eng/custom-cibuild.ps1
+++ b/eng/custom-cibuild.ps1
@@ -4,10 +4,10 @@ Param(
     [Parameter(ValueFromRemainingArguments=$true)][String[]]$properties
 )
 
-function _kill($processName) {
+function _endProcess($processName) {
     try {
-       if (Get-Process -Name "ServerManager" -ErrorAction SilentlyContinue) {
-            Write-Host "killing process ${processName}."
+       if (Get-Process -Name ${processName} -ErrorAction SilentlyContinue) {
+            Write-Host "Ending process ${processName}."
             # Redirect stderr to stdout to avoid big red blocks of output in Azure Pipeline logging.
             & cmd /c "taskkill /T /F /IM ${processName}.exe 2>&1"
         }
@@ -15,12 +15,9 @@ function _kill($processName) {
             Write-Host "${processName} process is not running"
         }
     } catch {
-        Write-Host "Failed to kill ${processName}"
+        Write-Host "Failed to end process ${processName}"
     }
 }
-
-# kill server manager process if running on build agents.
-_kill ServerManager
 
 # How long to wait before we consider a build/test run to be unresponsive
 $WaitSeconds = 900 # 15 min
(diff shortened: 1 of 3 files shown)


eng/custom-cibuild.ps1
```diff
@@ -5,18 +5,22 @@ Param(
...
+        }
     } catch {
-        Write-Host "Failed to kill ${processName}."
+        Write-Host "Failed to kill ${processName}"
     }
 }
 
 # kill server manager process if running on build agents.
-_kill severmanager.exe
+_kill ServerManager
```
JeremyKuhne, line 23: We really should change the name to `_endProcess` and avoid using `kill` here and elsewhere. I believe it will get flagged.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/ControlTestBase.cs
```diff
@@ -68,7 +69,14 @@ public virtual Task InitializeAsync()
...
             _joinableTaskCollection = JoinableTaskContext.CreateCollection();
             JoinableTaskFactory = JoinableTaskContext.CreateFactory(_joinableTaskCollection);
-            return Task.CompletedTask;
+
+            ScreenshotService.TakeScreenshot(Path.Combine(DataCollectionService.GetLogDirectory(), "InitialScreenShot.png"));
+
+            // Minimize all windows on test machine.
+            await JoinableTaskFactory.RunAsync(async () =>
+            {
+                await InputSimulator.SendAsync(null, inputSimulator => inputSimulator.Keyboard.ModifiedKeyStroke(VirtualKeyCode.LWIN, VirtualKeyCode.VK_M));
```
JeremyKuhne, line 78: If we want to minimize windows we should not use keyboard input to do so. Keyboard and mouse input should be avoided unless strictly necessary to accomplish tasks.

`IShellDispatch` will do this for you. I'll work with you tomorrow on how that works.

JeremyKuhne (changes requested):
Comments inline

eng/custom-cibuild.ps1
```diff
@@ -67,6 +53,16 @@ if ($CollectDebugInfo) {
...
 }
 
+ if ($ci -and $integrationTest) {
+      # Minimize all windows to avoid interference during integration test runs
+      $shell = New-Object -ComObject "Shell.Application"
+      $shell.MinimizeAll()
+	  Write-Host "Minimized all windows"
+	  
+	  # end server manager process if running on build agents.
+	  _endProcess ServerManager
```
JeremyKuhne, line 63: Is this necessary if we minimize?

eng/custom-cibuild.ps1
```diff
@@ -67,6 +53,16 @@ if ($CollectDebugInfo) {
 
 }
 
+ if ($ci -and $integrationTest) {
+      # Minimize all windows to avoid interference during integration test runs
+      $shell = New-Object -ComObject "Shell.Application"
```
JeremyKuhne, line 58: If we want to do this while running the tests it isn't too hard to do. You can add `IShellDispatch` to the `NativeMethods.txt` and load the COM class. `GlobalInterfaceTable` has an example of using CoCreateInstance.

The class id is: `CLSID_Shell	{13709620-C279-11CE-A49E-444553540000}`

dreddy-work: Thanks Jeremy. I do not see specific need to run it in tests yet. We are conditioning to run against Integration tests and only in CI just before launching tests.

JeremyKuhne (approved):

PR #8943, opened 2023-04-05
String interpolation leftovers

halgab:
This should be the final PR to address #8203.

The first commit should be enough to close the issue.

The second commit goes the extra mile in string creation optimisation, but I can move it to another PR or discard it altogether if the team is more confortable without it.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8943)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/TabOrder.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/TabOrder.cs
@@ -27,6 +27,7 @@ namespace System.Windows.Forms.Design
         private Dictionary<Control, int>? _tabNext;
         private readonly Font _tabFont;
         private readonly StringBuilder _drawString;
+        private readonly Stack<int> _controlIndices;
         private readonly Brush _highlightTextBrush;
         private readonly Pen _highlightPen;
         private readonly int _selSize;
@@ -54,6 +55,7 @@ namespace System.Windows.Forms.Design
 
             // Colors and brushes...
             _drawString = new StringBuilder(12);
+            _controlIndices = new Stack<int>();
             _highlightTextBrush = new SolidBrush(SystemColors.HighlightText);
             _highlightPen = new Pen(SystemColors.Highlight);
 
@@ -231,21 +233,20 @@ namespace System.Windows.Forms.Design
             {
                 Rectangle convertedRectangle = GetConvertedBounds(control);
 
-                _drawString.Length = 0;
+                _drawString.Clear();
+                _controlIndices.Clear();
 
-                Control? parent = GetSitedParent(control);
+                Control? parent = control;
                 Control baseControl = (Control)_host.RootComponent;
 
-                while (parent != baseControl && parent is not null)
+                do
                 {
-                    _drawString.Insert(0, _decimalSep);
-                    _drawString.Insert(0, parent.TabIndex.ToString(CultureInfo.CurrentCulture));
+                    _controlIndices.Push(parent.TabIndex);
                     parent = GetSitedParent(parent);
                 }
+                while (parent != baseControl && parent is not null);
 
-                _drawString.Insert(0, ' ');
-                _drawString.Append(control.TabIndex.ToString(CultureInfo.CurrentCulture));
-                _drawString.Append(' ');
+                _drawString.Append(' ').AppendJoin(_decimalSep, _controlIndices).Append(' ');
 
                 if (_tabProperties[control].IsReadOnly)
                 {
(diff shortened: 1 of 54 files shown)


halgab:
It looks like the test runners all failed... I tried closing and reopening the PR, but it didn't start the tests again.
This is essentially ready, I was just waiting for the tests to come out clean to mark the PR as such

dmitrii-drobotov:
> It looks like the test runners all failed... I tried closing and reopening the PR, but it didn't start the tests again.
> This is essentially ready, I was just waiting for the tests to come out clean to mark the PR as such

@halgab there is a [test failure](https://dev.azure.com/dnceng-public/public/_build/results?buildId=228359&view=ms.vss-test-web.build-test-results-tab) in `System.Windows.Forms.Analyzers.Tests.ApplicationConfigTests+FontDescriptorTests.FontDescriptor_ToString_culture_agnostic`, you can get more details by going on "View more details on Azure Pipelines" link under a failed pipeline. The failure also reproduces when running that test locally.

src/System.Windows.Forms.Analyzers/src/System/Windows/Forms/ApplicationConfig.FontDescriptor.cs
```diff
@@ -33,7 +32,7 @@ public override string ToString()
                     ? "global::System.Windows.Forms.Control.DefaultFont.FontFamily"
                     : $"new global::System.Drawing.FontFamily(\"{name}\")";
 
-                return $"new global::System.Drawing.Font({fontFamily}, {Size.ToString(CultureInfo.InvariantCulture)}f, (global::System.Drawing.FontStyle){(int)Style}, (global::System.Drawing.GraphicsUnit){(int)Unit})";
+                return $"new global::System.Drawing.Font({fontFamily}, {Size}f, (global::System.Drawing.FontStyle){(int)Style}, (global::System.Drawing.GraphicsUnit){(int)Unit})";
```
dmitrii-drobotov, line 35: Looks like removing InvariantCulture causes it to use comma instead of a dot as a decimal separator in certain cultures, which causes this test to fail:
https://github.com/dotnet/winforms/blob/0019303be0fce2d5be96d41221643927f775f786/src/System.Windows.Forms.Analyzers/tests/UnitTests/System/Windows/Forms/Analyzers/ApplicationConfigTests.FontDescriptor.cs#L65-L68

halgab: Okay, I missed that `Size` was a float... Thanks a lot for going through the tests. I'll have another look at potential issues like this one in this PR

dmitrii-drobotov (approved):
Great work!

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/TabOrder.cs
```diff
@@ -231,21 +233,20 @@ private void DrawTabs(List<Control> tabs, Graphics graphics, bool fRegion)
...
-                Control? parent = GetSitedParent(control);
+                Control? parent = control;
                 Control baseControl = (Control)_host.RootComponent;
 
-                while (parent != baseControl && parent is not null)
+                do
                 {
-                    _drawString.Insert(0, _decimalSep);
-                    _drawString.Insert(0, parent.TabIndex.ToString(CultureInfo.CurrentCulture));
+                    _controlIndices.Push(parent.TabIndex);
```
dreddy-work, line 244: if parent is null, this could throw.

dreddy-work (commented):
RLGTM.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/TabOrder.cs
halgab: Here I relied on 2 things. First, this class has nullable annotations, which indicate that `control` (which is the first value of `parent`) is not null. Secondly, the code I'm modifying already calls `control.TabIndex` on line 247 without checking for null, so if my code is broken, it's as broken as the current one.

halgab: As a side-note, in this method we could forego the StringBuilder completely and write to a rented char array as `Graphics.MeasureString` and `Graphics.DrawString` accept a `ReadOnlySpan<char>` input. If there's interest I can look into it

dreddy-work: Right. do-while makes it misleading given `parent` is declared as nullable. And parent cannot be declared as non-nullable :)

dreddy-work (approved):

PR #8947, opened 2023-04-06
Trick the build into enabling IDE1006 but still allowing per-rule severity configuration

sharwell:
The configuration for this repository currently contains several naming rules. The severity is configured for each of these naming rules:

https://github.com/dotnet/winforms/blob/ca8cc15b2bceea5f873ca330e9b2a87397919b4b/.editorconfig#L33
https://github.com/dotnet/winforms/blob/ca8cc15b2bceea5f873ca330e9b2a87397919b4b/.editorconfig#L43
https://github.com/dotnet/winforms/blob/ca8cc15b2bceea5f873ca330e9b2a87397919b4b/.editorconfig#L54

If the build contains a line that sets IDE1006 to warning, it will override _all_ of the above values, and it will not be possible to configure one or more rules as suggestions while others are enforced. Rather than set IDE1006 to warning for all files in the solution, this pull request sets IDE1006 to warning for _exactly one_ file in each project, which is an empty file that contains no violations of IDE1006. This line is sufficient for the compiler to know to enable IDE1006 (it won't run the analyzer at all unless it knows that it may produce a warning), after which point the per-rule configurations listed above control the actual diagnostics emitted in code.

After this change, it will be possible to change any one of the above lines from `suggestion` to `warning` in order to start enforcing that rule. It would also be possible to add other naming rules if desired, and selectively enforce those rules on build.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8947)


JeremyKuhne:
Can you explain this a different way? I'm struggling to understand what this is doing exactly. Perhaps a concrete example of what isn't working?

sharwell:
@JeremyKuhne updated the description

JeremyKuhne (approved):

PR #8949, opened 2023-04-06
Verify key state at the start of each test

sharwell:
This is an attempt to assert that a previous test has not left a keyboard or mouse key in a pressed state going into the test.

### DIFF
--- a/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/ControlTestBase.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/ControlTestBase.cs
@@ -7,6 +7,8 @@ using System.Reflection;
 using System.Runtime.InteropServices;
 using Microsoft.VisualStudio.Threading;
 using Windows.Win32.UI.WindowsAndMessaging;
+using WindowsInput;
+using WindowsInput.Native;
 using Xunit;
 using Xunit.Abstractions;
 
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/ControlTestBase.cs
```diff
@@ -33,6 +34,16 @@ protected ControlTestBase(ITestOutputHelper testOutputHelper)
 
             Application.EnableVisualStyles();
 
+            Span<byte> keyboardState = stackalloc byte[256];
```
dreddy-work, line 37: Can you please add some comments here on what we are ensuring and why? Just by reading the code its not very clear unless you have context.

dreddy-work: And why not reset state instead of assert?

sharwell: oh, yes, sorry

sharwell: > And why not reset state instead of assert?

It's a sign that a previous test terminated in a bad state. I updated the code to also validate at the end of a test to help identify the test which left keys in a bad state.

dreddy-work: Right. But in dispose, it still make sense to fail test and reset state for the following tests to continue? Otherwise, rest of the tests also fail?

sharwell: Only one of the tests will have the call stack showing Dispose, so it should be easy to narrow down. The Screenshots folder also lists items in order of execution.

JeremyKuhne (approved):

PR #8956, opened 2023-04-07
Make sure to invoke GetCursorPos after moving mouse to an absolute position

sharwell:


src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/IMouseSimulatorExtensions.cs
```diff
@@ -28,6 +28,9 @@ public static IMouseSimulator DragMouseTo(this IMouseSimulator simulator, double
         {
             simulator.MoveMouseTo(absoluteX, absoluteY);
 
+            // ⚠ The call to GetCursorPos is required for correct behavior.
+            _ = PInvoke.GetCursorPos(out _);
```
sharwell, line 32: 📝 I don't know how I found out about this originally. The oldest reference I found so far to this behavior is here:
https://github.com/tunnelvisionlabs/MouseFastScroll/commit/306d960c27e656f40020382701a7444a30cbc023

JeremyKuhne: From what I can see this might be a red herring. The Windows code is just a field access (well, with a conversion for DPI if necessary). It appears that this is effectively a very small sleep.

JeremyKuhne:
Fyi, I did a little looking at:

``` C#
                || thread is { ThreadState: ThreadState.Wait, WaitReason: ThreadWaitReason.UserRequest };
```

`UserRequest` aggregates both `WrUserRequest` and `UserRequest`. It looks like `WrUserRequest` might be more specific to "waiting for messages".

This said I wonder why we don't just call `Application.DoEvents` here to make sure the message queue is empty.

Another option would be to call `Control.FromChildHandle` to find the related `Control` and call `Control.Invoke` with a check for `GetQueueStatus`.

sharwell:
> This said I wonder why we don't just call `Application.DoEvents` here to make sure the message queue is empty.

It might not be possible to empty the message queue. Some of the drag/drop tests do not return from COM code until more messages are sent. Also this thread doesn't have a message pump (it's sending messages to other threads).

sharwell:
Marking as draft. The work in this pull request will become obsolete by the follow-up changes from #8959.

JeremyKuhne:
> For now, I think we benefit slightly from the fact that it isn't fully specific to just a single UI strategy. 

I'm not sure how. Can you elaborate?

sharwell:
Superseded by #8970 and #8971

PR #8957, opened 2023-04-07
Resolve more pending test issues in System.Drawing

JeremyKuhne:
Removes suppressions that were for non-Windows platforms and adds tolerance support for ARM related floating point variance.

- Update ValidateBitmapContent helper to hash the raw image byte data instead of the entire serialized image to allow for variance of non-essential image data.
- Refactor comparison code into new `ComparisonHelpers` support class.
- Introduce a `FloatingPointToleranceComparer` to allow comparing float arrays with tolerance (to allow for ARM floating point differences in tests).
- Also add some float `AssertExtensions` for `RectangleF`.
- Adds a check to validate that we successfully got the screen HDC in Graphics.CopyFromScreen.
- Adds `ReadOnlyTheoryData` to ensure that common test data doesn't get modified and reworks test data attributes to be simpler and utilize a new raw `TestData` helper class.
- Rename `NumberData` to `IntegerData`.
- Add a GlobalUsings to TestUtilities test project.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8957)


src/Common/tests/TestUtilities/ComparisonHelpers.cs
```diff
@@ -0,0 +1,119 @@
...
+{
+    public static bool EqualsInteger<T>(T x, T y, T variance)
+        where T : struct, IBinaryInteger<T>
+    {
+        if (variance < T.Zero)
+        {
+            throw new ArgumentOutOfRangeException(nameof(variance));
+        }
+
+        return T.Abs(x > y ? x - y : y - x) <= variance;
```
dreddy-work, line 19: Q: why compare `x>y` when checking on absolute value?

dreddy-work (approved):
RLGTM.

src/System.Drawing.Common/src/System/Drawing/Graphics.cs
```diff
@@ -3097,6 +3097,13 @@ public void CopyFromScreen(int sourceX, int sourceY, int destinationX, int desti
             int destHeight = blockRegionSize.Height;
 
             nint screenDC = Interop.User32.GetDC(0);
+            if (screenDC == 0)
+            {
+                // ERROR_INVALID_HANDLE - if you pass an empty handle to BitBlt you'll get this error.
+                // Checking here to better describe test failures (and avoids taking the Graphics HDC lock).
+                throw new Win32Exception(6);
```
Tanya-Solyanik, line 3104: Why didn't you define `ERROR_INVALID_HANDLE` instead of a hardcoded number?

JeremyKuhne:
Merged as this fixes multiple initialization of test data.

src/System.Drawing.Common/src/System/Drawing/Graphics.cs
sharwell: It's possible CsWin32 can generate the constant if you add it to NativeMethods.txt.

PR #8958, opened 2023-04-07
Support Xunit.Combinatorial in test code

sharwell:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8958)

### DIFF
--- a/src/System.Windows.Forms/tests/UnitTests/CheckedListBoxTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/CheckedListBoxTests.cs
@@ -20,7 +20,7 @@ namespace System.Windows.Forms.Tests
         }
 
         [WinFormsTheory]
-        [BoolData]
+        [CombinatorialData]
         public void CheckedListBox_CheckOnClick(bool expected)
         {
             using var box = new CheckedListBox
@@ -105,7 +105,7 @@ namespace System.Windows.Forms.Tests
         }
 
         [WinFormsTheory]
-        [BoolData]
+        [CombinatorialData]
         public void CheckedListBox_ThreeDCheckBoxes(bool expected)
         {
             using var box = new CheckedListBox
@@ -333,7 +333,7 @@ namespace System.Windows.Forms.Tests
         }
 
         [WinFormsTheory]
-        [EnumData<CheckState>]
+        [CombinatorialData]
         public void CheckedListBox_SetItemCheckState_Invoke_GetReturnsExpected(CheckState value)
         {
             using var control = new CheckedListBox();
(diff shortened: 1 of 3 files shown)


src/System.Windows.Forms/tests/UnitTests/CheckedListBoxTests.cs
```diff
@@ -105,7 +105,7 @@ public void CheckedListBox_SelectionMode_SetInvalidValue_ThrowsInvalidEnumArgume
         }
 
         [WinFormsTheory]
-        [BoolData]
+        [CombinatorialData]
```
JeremyKuhne, line 108: I don't see that this is a better choice than `[BoolData]`. While in this case you can probably guess what it provides, what does it do with other types? It isn't obvious that `int` gets you `0` and `1`.

This library also appears to only intrinsically support `int`, `bool` and `enum`. Additionally, it doesn't appear that there is any caching of the data.

I'd want to see a clear win with it before we take this dependency.

sharwell: For local testing, the library is extremely valuable for running iterations of a test. I can change a test from this:

```cs
[Fact]
public void Test()
{
}
```

to

```cs
[Theory]
[CombinatorialData]
public void Test([CombinatorialRange(0, 100)] int iteration)
{
  _ = iteration;
}
```

to run the test 100 times.

It also can handle cases involving more than one parameter.

sharwell: > While in this case you can probably guess what it provides, what does it do with other types? It isn't obvious that int gets you 0 and 1.

I wasn't aware that you could use it for int directly. I always use either `[CombinatorialValues(A, B, C)]` or `CombinatorialRange` on the parameter for values like int.

It's a pretty common library, so I wouldn't expect it to be problematic in practice for people to use.

JeremyKuhne: > For local testing, the library is extremely valuable for running iterations of a test.

"RunUntilFailure" in the test explorer seems more practical.

>I wasn't aware that you could use it for int directly. I always use either [CombinatorialValues(A, B, C)] or CombinatorialRange on the parameter for values like int.

Do these work if you override them? Could you do a `[BoolValues]`, for example?

> It's a pretty common library, so I wouldn't expect it to be problematic in practice for people to use.

I'd still want to see real compelling usages in a PR before we add this. Ideally, we'd also get our docs updated to provide some more guidance on what test attributes we use and the best practices for them.

sharwell: > "RunUntilFailure" in the test explorer seems more practical.

In theory it could be, but it's multiple orders of magnitude slower in my testing. It also doesn't allow for estimating the percentage of times the test will fail based on number of failures in a large number of total runs.

> Do these work if you override them? Could you do a [BoolValues], for example?

Yes, but the attribute would apply to the parameter and not the method. For `bool` it wouldn't make sense anyway because it already knows to use `true` and `false` by default.

JeremyKuhne:
Discussed offline with the team, we don't feel like this is something we want to take in its current state for the reasons described in the comments.

PR #8959, opened 2023-04-08
Use CsWin32 for SendInput interop

sharwell:
This is the first of four changes to eliminate the dependency or InputSimulatorPlus and allow us to apply MOUSEEVENTF_MOVE_NOCOALESCE to all simulated mouse actions in integration tests.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8959)


JeremyKuhne (approved):
Thanks for this!

PR #8964, opened 2023-04-10
Use SendInput directly instead of InputSimulatorPlus

sharwell:
This is the third of four changes to eliminate the dependency or InputSimulatorPlus and allow us to apply MOUSEEVENTF_MOVE_NOCOALESCE to all simulated mouse actions in integration tests.

Follow-up to #8963

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8964)


src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextProvider.cs
```diff
@@ -129,7 +129,7 @@ public unsafe int SendKeyboardInputVK(VIRTUAL_KEY vk, bool press)
             return SendInput(1, ref keyboardInput, sizeof(INPUT));
         }
 
-        private static bool IsExtendedKey(VIRTUAL_KEY vk)
+        internal static bool IsExtendedKey(VIRTUAL_KEY vk)
```
Tanya-Solyanik, line 132: Would you mind moving this method in one of the helper classes?

sharwell: Not at all. I'm not familiar with the helper layout in this project. Is there one you had in mind?

sharwell: ➡️ Moved to `VirtualKeyUtilities` and updated references

DJm00n:
@sharwell `IsExtendedKey` - is a property of scan code, not VK code so method should take scan code in its param.
Proper scan code is returned from `MapVirtualKey` call with VK code in `MAPVK_VK_TO_VSC_EX` mode.
Extended key scan code have `0xe0` or `0xe1` in its high byte.

Here is list of i8042 scan codes that are mapped from HID usages in Windows: https://learn.microsoft.com/windows/win32/inputdev/about-keyboard-input#scan-codes

Issue #8952, opened 2023-04-07
Saving after changing anything on some forms causes VS to hang.


AhmadKelany:
### Environment

Microsoft Visual Studio Community 2022
Version 17.5.3
VisualStudio.17.Release/17.5.3+33516.290
Microsoft .NET Framework
Version 4.8.04161

Installed Version: Community

Visual C++ 2022   00482-90000-00000-AA004
Microsoft Visual C++ 2022

ASP.NET and Web Tools   17.5.318.41597
ASP.NET and Web Tools

Azure App Service Tools v3.0.0   17.5.318.41597
Azure App Service Tools v3.0.0

Azure Functions and Web Jobs Tools   17.5.318.41597
Azure Functions and Web Jobs Tools

C# Tools   4.5.0-6.23128.14+651d32d8017cfcc60fcbd67383601fa9a3deb145
C# components used in the IDE. Depending on your project type and settings, a different version of the compiler may be used.

Common Azure Tools   1.10
Provides common services for use by Azure Mobile Services and Microsoft Azure Tools.

Extensibility Message Bus   1.4.3 (main@2a4517a)
Provides common messaging-based MEF services for loosely coupled Visual Studio extension components communication and integration.

Microsoft JVM Debugger   1.0
Provides support for connecting the Visual Studio debugger to JDWP compatible Java Virtual Machines

Mono Debugging for Visual Studio   17.5.9 (11975e6)
Support for debugging Mono processes with Visual Studio.

NuGet Package Manager   6.5.0
NuGet Package Manager in Visual Studio. For more information about NuGet, visit https://docs.nuget.org/

Razor (ASP.NET Core)   17.5.2.2316603+9f1b6856460af1e592d387ebef416eadddac453f
Provides languages services for ASP.NET Core Razor.

SQL Server Data Tools   17.2.40119.0
Microsoft SQL Server Data Tools

Test Adapter for Boost.Test   1.0
Enables Visual Studio's testing tools with unit tests written for Boost.Test.  The use terms and Third Party Notices are available in the extension installation directory.

Test Adapter for Google Test   1.0
Enables Visual Studio's testing tools with unit tests written for Google Test.  The use terms and Third Party Notices are available in the extension installation directory.

TypeScript Tools   17.0.20105.2003
TypeScript Tools for Microsoft Visual Studio

Visual Basic Tools   4.5.0-6.23128.14+651d32d8017cfcc60fcbd67383601fa9a3deb145
[...]

PR #8977, opened 2023-04-12
Record processes active at the start of each test

sharwell:
null

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8977)

PR #8982, opened 2023-04-12
Pause NGEN to avoid CPU interference during the run

sharwell:
Logs for #8952 showed copies of **mscorsvw** running in the background. This change pauses the NGEN queue to prevent **mscorsvw** from consuming CPU and potentially interfering with the integration tests.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8982)

### DIFF
--- a/eng/pipelines/build.yml
+++ b/eng/pipelines/build.yml
@@ -64,6 +64,12 @@ jobs:
     - powershell: Remove-Item -Recurse -ErrorAction Ignore "$env:LocalAppData\NuGet\v3-cache"
       displayName: Clear NuGet http cache (if exists)
 
+    - script: C:\Windows\Microsoft.NET\Framework\v4.0.30319\ngen.exe queue pause
+      displayName: Pause NGEN x86
+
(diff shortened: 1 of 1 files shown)


dreddy-work:
Impact is very hard to track here. Actually, is it possible to force ngen to complete before starting tests instead of pause?

sharwell:
> Actually, is it possible to force ngen to complete before starting tests instead of pause?

Yes, but it appeared to add about 10 minutes to the build.

elachlan:
Wow! Good find!

dreddy-work (approved):

KlausLoeffelmann:
> Logs for https://github.com/dotnet/winforms/issues/8952 showed copies of mscorsvw running in the background.

Is that the correct issue reference?

sharwell:
> > Logs for #8952 showed copies of mscorsvw running in the background.
> 
> Is that the correct issue reference?

No, I updated with the correct reference #8977.

PR #8983, opened 2023-04-12
Add ExtractIcon functionality to Icon

### LINKED ISSUE #8929: [API Proposal]: Add Extract methods to Icon
### Background and motivation

`System.Drawing.Icon` has methods for creating icons from managed resources and associated native icons. It doesn't, however, allow you to get all native resource icons, or always specify the size that you require.

The `Icon` constructors that take a path / stream only support `.ico` files. They do allow choosing a particular size, but also try to pick a color depth based on the current display settings. Color depth matching is of limited use and these constructors are costly as they manually load the entire file, parsing it to create the "best-fit" icon. The full stream is kept in memory (this is particularly painful if you're trying to load smaller icons).

`System.Drawing.Icon.ExtractAssociatedIcon()` has some overlap with these new APIs. It will load the first icon from the file it's given or fall back on whatever executable the file is associated with. The API used for this doesn't allow specifying a size. (We could spin our own by calling [`FindExecutable`](https://learn.microsoft.com/windows/win32/api/shellapi/nf-shellapi-findexecutablew) if we find a need in the future.)

### API Proposal

```csharp
namespace System.Drawing;

public class Icon
{
    // The existing Extract method.
    public static Icon ExtractAssociatedIcon(string! filePath);

    // The following work with .ico files as well as PE files (.exe, .dll, etc.). The id is an index when positive or a resource id when negative.

    // Retrieves the specified icon at the current system icon size setting (large by default).
    public static Icon? ExtractIcon(string! filePath, int id, bool smallIcon = false);

    // Allows retrieving the specified icon at a specific size (say 32x32). Icon sizes are always square.
    public static Icon? ExtractIcon(string! filePath, int id, ushort size);

    // Gets the icon count for the specified file.
    public static int GetIconCount(string! filePath);
}

public static class SystemIcons
{
    // Existing.
    public static unsafe Icon GetStockIcon(StockIconId stockIcon, StockIconOptions options = StockIconOptions.Default);

    // New. When asking for an explicit size (as opposed to just large/small) other options aren't available.
    public static unsafe Icon GetStockIcon(StockIconId stockIcon, ushort size);
}
```

### API Usage

```csharp
// Get the first icon in regedit.exe at 32x32 size.
using Icon icon = Icon.ExtractIcon("regedit.exe", id: 0, size: 32);

// Get the icon from devenv.exe with the 1200 resource id, at the default large size
using Icon icon = Icon.ExtractIcon("devenv.exe", id: -1200);
```

### Alternative Designs

Could potentially add additional constructors to `Icon`, but behavior differences between string overloads would probably be confusing. For example, as the entire source file is not copied, resizing via `Copy` is scaled, instead of reparsed.

### Risks

Nothing notable.

### Notes

- This API will scale to the requested size from available sizes. We'll use `SHGetStockIconInfo`- I believe it tries to scale down from larger sizes.
- We will not track the original source or update the copy constructors. When the full data is kept in the other APIs the original data is parsed again to try and find a matching size when using the copy constructors. With `Icon`s loaded through these APIs you'll just get the current backing bitmap scaled if you change sizes with the copy constructors (as you would with `HICON` constructed `Icon`s and `ExtractAssociatedIcon`).

JeremyKuhne:
This adds Icon.Extracticon to allow extracting icons from icon files and native resources from PE files at any specified resolution.

Also adds an overload to SystemIcons.GetStockIcon() to get a specified size.

Fixes #8929

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8983)

### DIFF
--- a/src/System.Drawing.Common/src/Interop/Windows/Shell32/Interop.SHDefExtractIcon.cs
+++ b/src/System.Drawing.Common/src/Interop/Windows/Shell32/Interop.SHDefExtractIcon.cs
@@ -1,0 +1,22 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class Shell32
+    {
+#if NET8_0_OR_GREATER
+        [LibraryImport(Libraries.Shell32, StringMarshalling = StringMarshalling.Utf16, EntryPoint = "SHDefExtractIconW")]
+        internal static unsafe partial HRESULT SHDefExtractIcon(
+            char* pszIconFile,
+            int iIndex,
+            uint uFlags,
+            nint* phiconLarge,
+            nint* phiconSmall,
+            uint nIconSize);
+#endif
+    }
+}
--- a/src/System.Drawing.Common/src/System/Drawing/Icon.cs
+++ b/src/System.Drawing.Common/src/System/Drawing/Icon.cs
@@ -1,4 +1,4 @@
-// Licensed to the .NET Foundation under one or more agreements.
+﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Buffers;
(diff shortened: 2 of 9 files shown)


src/System.Drawing.Common/src/Interop/Windows/Shell32/Interop.SHDefExtractIcon.cs
```diff
@@ -0,0 +1,22 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Runtime.InteropServices;
+
+internal static partial class Interop
+{
+    internal static partial class Shell32
+    {
+#if NET8_0_OR_GREATER
```
Tanya-Solyanik, line 11: Why do we need this restriction, this branch is for .NET8+

src/System.Drawing.Common/src/System/Drawing/Icon.cs
```diff
@@ -12,872 +12,953 @@
...
+
+    internal static class Ole
+    {
+        public const int PICTYPE_ICON = 3;
+    }
 
-            return _isBestImagePng.Value;
+#if NET8_0_OR_GREATER
+    /// <summary>
+    ///  Extracts a specified icon from the given <paramref name="filePath"/>.
```
Tanya-Solyanik, line 885: I think it would be useful to explain that color depth is not modified as opposed to what the constructor does.

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/Buttons.cs
```diff
@@ -111,6 +111,15 @@ protected override void OnLoad(EventArgs e)
                 column: 0,
                 row: 1);
 
+            table.Controls.Add(
+                new Button
+                {
+                    AutoSize = true,
+                    Image = Icon.ExtractIcon("regedit.exe", 0, 256).ToBitmap()
```
Tanya-Solyanik, line 118: What happens if I edit the registry when the test is running?

src/System.Drawing.Common/tests/IconTests.cs
```diff
@@ -863,5 +863,80 @@ public void CorrectColorDepthExtracted()
...
+
+            // Recent builds of Windows have added a few more icons to regedit.
+            Assert.True(count == 7 || count == 5, $"count was {count}, expected 5 or 7");
+        }
+
+        [Fact]
+        public void ExtractIcon_RegeditByResourceId()
+        {
+            Icon? icon = Icon.ExtractIcon("regedit.exe", -100, 256);
+            Assert.NotNull(icon);
```
dreddy-work, line 936: Dispose Icon or `using`?

Tanya-Solyanik (dismissed):
I added some questions

src/System.Drawing.Common/src/Interop/Windows/Shell32/Interop.SHDefExtractIcon.cs
dreddy-work: Roll forward scenarios ?

JeremyKuhne: System.Drawing.Common is built for multiple platforms, not just .NET 8.

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/Buttons.cs
JeremyKuhne: Nothing. It doesn't lock the file to run it.

dreddy-work (approved):

JeremyKuhne:
Note that I've hit multiple UI test failures on this PR.

PR #8986, opened 2023-04-13
Null out FontFamily field on Font when disposing

### LINKED ISSUE #8823: Clarification about System.Drawing.Font and Dispose

JeremyKuhne:
To avoid rooting the FontFamily if the Font itself is still rooted after disposal.

This introduces an ObjectDisposedException if you try to access fields that use the FontFamily after disposing.

Fixes #8823

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8986)

### DIFF
--- a/src/System.Drawing.Common/src/System/Drawing/LocalAppContextSwitches.cs
+++ b/src/System.Drawing.Common/src/System/Drawing/LocalAppContextSwitches.cs
@@ -32,6 +32,6 @@ internal static partial class LocalAppContextSwitches
     public static bool DoNotConsiderDisposedStateInFontEquals
     {
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        get => GetCachedSwitchValue(@"Switch.System.Drawing.DoNotNullOutFontFamilyOnDispose", ref s_fontConsiderDisposedState);
+        get => GetCachedSwitchValue(@"Switch.System.Drawing.DoNotConsiderDisposedStateInFontEquals", ref s_fontConsiderDisposedState);
     }
 }
--- a/src/System.Drawing.Common/src/System/Drawing/LocalAppContextSwitches.cs
+++ b/src/System.Drawing.Common/src/System/Drawing/LocalAppContextSwitches.cs
@@ -1,30 +1,37 @@
-// Licensed to the .NET Foundation under one or more agreements.
+﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Runtime.CompilerServices;
 
-namespace System
-{
-    internal static partial class LocalAppContextSwitches
-    {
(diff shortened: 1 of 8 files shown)


kirsan31:
Thank you @JeremyKuhne
But I think this PR doesn't answer main questions raised in https://github.com/dotnet/winforms/issues/8823 ...

JeremyKuhne:
This is more convoluted than I expected. I think I might have to add some compat switches for this.

It surfaced an issue with caching in the runtime- we were retaining a `Font` cache without a using weak references. Beyond that `Font`s are compared by properties and not instances, which makes it really easy to propagate disposed `Font` objects. I think the right answer here is to never return true for `Equals` when `Font`s are disposed (unless they're ref equal).

JeremyKuhne:
`Switch.System.Drawing.DoNotNullOutFontFamilyOnDispose` will skip the nulling of the FontFamily.
`Switch.System.Drawing.DoNotConsiderDisposedStateInFontEquals` will skip considering disposed state in Font.Equals.

It is really unfortunate that `Font` just considers traits for equality.

src/System.Drawing.Common/src/System/Drawing/LocalAppContextSwitches.cs
```diff
@@ -1,30 +1,37 @@
...
-            get
-            {
-                return GetCachedSwitchValue(@"Switch.System.Drawing.Printing.OptimizePrintPreview", ref s_optimizePrintPreview);
-            }
-        }
+    private static int s_nullOutFontFamily;
+    public static bool DoNotNullOutFontFamilyOnDispose
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        get => GetCachedSwitchValue(@"Switch.System.Drawing.DoNotNullOutFontFamilyOnDispose", ref s_nullOutFontFamily);
```
Tanya-Solyanik, line 28: `DoNotNullOutFontFamilyOnFontDispose`?  - Let's specify which Dispose method we mean.

src/System.Drawing.Common/tests/FontTests.cs
```diff
@@ -341,18 +341,14 @@ public void Ctor_NullFont_ThrowsNullReferenceException()
         }
 
         [Fact]
-        public void Ctor_DisposedFont_Success()
+        public void Ctor_DisposedFont_ThrowsObjectDisposed()
```
Tanya-Solyanik, line 344: Is it possible to test the quirks as well?

JeremyKuhne:
Converting to draft as I think of all of the implications here. I'm going to roll out the other fixes to WinForms code separately.

JeremyKuhne:
Closing this for now until I figure out how to deal with this.

PR #8987, opened 2023-04-13
Implement CachedBitmap

### LINKED ISSUE #8822: Implement CachedBitmap functionality in System.Drawing
### Background

Drawing a `Bitmap` in GDI+ has relatively poor performance, since the image is stored in a device-independent format. The [`CachedBitmap`](https://docs.microsoft.com/en-us/windows/win32/api/gdiplusheaders/nl-gdiplusheaders-cachedbitmap) class stores an image in a format that is optimized for display on a particular device. Thus, rendering an image stored in a `CachedBitmap` is fast, because no processing time is spent converting the image to the format required by the display device. Naturally, such a capability can substantially improve graphics performance in many scenarios. However, for whatever reason, it's not directly usable from the `System.Drawing` APIs.

### API Proposal

This API is derived from [this page of the GDI+ flat API](https://docs.microsoft.com/en-us/windows/win32/gdiplus/-gdiplus-cachedbitmap-flat). (I don't know what `GdipEmfToWmfBits` is doing on that page, but `System.Drawing` doesn't support saving metafiles anyway.)

```diff
namespace System.Drawing.Imaging
{
+   public sealed class CachedBitmap : MarshalByRefObject, IDisposable
+   {
+       public CachedBitmap(Bitmap bitmap, Graphics graphics);
+       public void Dispose();  
+   }
} 

namespace System.Drawing
{
    public sealed class Graphics : MarshalByRefObject, IDisposable, IDeviceContext
    {
+       public void DrawCachedBitmap(CachedBitmap cachedBitmap, int x, int y);	 
    }
}
```

As an aside: currently, libgdiplus doesn't implement this functionality. I imagine it would be fine to just PNSE when not on Windows for now (or, alternatively, making it "lie" about being a CachedBitmap, and simply calling `DrawImage` from `DrawCachedBitmap`).

JeremyKuhne:
Implements GDI+ `CachedBitmap` wrapper which allows caching a device dependent copy of `Bitmap`. Rendering is signficantly faster (5x+), although it has a few caveats:

- It needs to be regenerated if color depth changes
- It cannot be rendered to a Graphics object that has a tranform other than translation (no rotation, scaling)

Fixes #8822

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8987)

### DIFF
--- a/src/System.Drawing.Common/src/System/Drawing/Imaging/CachedBitmap.cs
+++ b/src/System.Drawing.Common/src/System/Drawing/Imaging/CachedBitmap.cs
@@ -1,0 +1,79 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Threading;
+
+#if NET8_0_OR_GREATER
+
+
+using static System.Drawing.SafeNativeMethods;
+
+namespace System.Drawing.Imaging;
+
+/// <summary>
+///  A device dependent copy of a <see cref="Bitmap"/> matching a specified <see cref="Graphics"/> object's current
+///  device (display) settings. Avoids reformatting step when rendering, which can significantly improve performance.
+/// </summary>
+/// <remarks>
+///  <para>
+///   <see cref="CachedBitmap"/> matches the current bit depth of the <see cref="Graphics"/>'s device. If the device bit
+///   depth changes, the <see cref="CachedBitmap"/> will no longer be usable and a new instance will need to be created
+///   that matches. If the <see cref="CachedBitmap"/> was created against <see cref="PixelFormat.Format32bppRgb"/> it
+///   will always work.
+///  </para>
+///  <para>
+///   <see cref="CachedBitmap"/> will not work with any transformations other than translation.
+///  </para>
+///  <para>
+///   <see cref="CachedBitmap"/> cannot be used to draw to a printer or metafile.
+///  </para>
+/// </remarks>
+public sealed class CachedBitmap : IDisposable
+{
+    private nint _handle;
+
+    /// <summary>
+    ///  Create a device dependent copy of the given <paramref name="bitmap"/> for the device settings of the given
+    ///  <paramref name="graphics"/>
+    /// </summary>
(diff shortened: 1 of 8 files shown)


src/System.Drawing.Common/src/System/Drawing/Imaging/CachedBitmap.cs
```diff
@@ -0,0 +1,79 @@
...
+{
+    private nint _handle;
+
+    /// <summary>
+    ///  Create a device dependent copy of the given <paramref name="bitmap"/> for the device settings of the given
+    ///  <paramref name="graphics"/>
+    /// </summary>
+    /// <param name="bitmap">The <see cref="Bitmap"/> to convert.</param>
+    /// <param name="graphics">The <see cref="Graphics"/> object to use to format the cached copy of the <paramref name="bitmap"/>.</param>
+    /// <exception cref="ArgumentNullException"><paramref name="bitmap"/> or <paramref name="graphics"/> is <see langword="null"/>.</exception>
```
dreddy-work, line 41: Why are we not mentioning GDI exceptions thrown via `CheckStatus`

src/System.Drawing.Common/tests/Imaging/BitmapDataTests.cs
```diff
@@ -1,4 +1,4 @@
-// Licensed to the .NET Foundation under one or more agreements.
```
dreddy-work, line 1: No changes in this file.

src/System.Drawing.Common/src/System/Drawing/Imaging/CachedBitmap.cs
JeremyKuhne: Because it is unbounded. Conventionally we only list notable exceptions. We don't, for example, list when the error you get when you use a disposed `Graphics` or `Bitmap`.

src/System.Drawing.Common/tests/Imaging/BitmapDataTests.cs
JeremyKuhne: Encoding changed, letting it flow.

dreddy-work (approved):

PR #8988, opened 2023-04-13
Fix FontConverter InstanceDescriptor conversion

### LINKED ISSUE #8838: FontConverter never generates parameter for GdiVerticalFont value
### Description

This is ported from https://github.com/dotnet/winforms-designer/issues/2350

#### Winforms Designer scenario:
Create a Winforms .Net Core application.
Select Form1 in the property window.
Expanf Font property for the Form.
Change the GdiVerticalFont property value from False to True under the Font node.

Result:
The following constructor is generated:
this.Font = new System.Drawing.Font("Segoe UI", 9F, System.Drawing.FontStyle.Regular, System.Drawing.GraphicsUnit.Point);

Expected:
In .Net Framework, we were generating:
this.Font = new System.Drawing.Font("Microsoft Sans Serif", 8.25F, System.Drawing.FontStyle.Regular, System.Drawing.GraphicsUnit.Point, ((byte)(0)), true);
where the last parameter is GdiVerticalFont

### Configuration
Windows

### Regression?
regression from the .NET Framework

### Other information
This is cause by a change in FontConverter, because it stopppped generating constructor with GdiVertiaclFont parameter.
.NET Core: https://source.dot.net/#System.Drawing.Common/System/Drawing/FontConverter.cs,88
.NET Framework: https://referencesource.microsoft.com/#System.Drawing/commonui/System/Drawing/FontConverter.cs,248

JeremyKuhne:
When porting to Core the `InstanceDescriptor` conversion was simplified to work with only the values that were supported cross platform. This caused data loss on Windows when custom GDI options were present.

Now that `System.Drawing` is only supported on Windows, match the .NET Framework implementation.

Fixes #8838

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8988)

### DIFF
--- a/src/System.Drawing.Common/src/System/Drawing/FontConverter.cs
+++ b/src/System.Drawing.Common/src/System/Drawing/FontConverter.cs
@@ -1,7 +1,8 @@
-// Licensed to the .NET Foundation under one or more agreements.
+﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Collections;
+using System.Collections.Generic;
 using System.ComponentModel;
 using System.ComponentModel.Design.Serialization;
 using System.Diagnostics;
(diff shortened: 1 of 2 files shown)


src/System.Drawing.Common/src/System/Drawing/FontConverter.cs
```diff
@@ -87,14 +88,62 @@ public override bool CanConvertTo(ITypeDescriptorContext? context, [NotNullWhen(
 
                 if (destinationType == typeof(InstanceDescriptor))
                 {
-                    ConstructorInfo? met = typeof(Font).GetConstructor(new Type[] { typeof(string), typeof(float), typeof(FontStyle), typeof(GraphicsUnit) });
-                    object[] args = new object[4];
+                    // Generate the smallest constructor possible.
```
Tanya-Solyanik, line 91: Why is it important to get the smallest constructor possible?

src/System.Drawing.Common/tests/System/Drawing/FontConverterTests.cs
```diff
@@ -112,40 +114,75 @@ public void GetFontPropsSorted()
             Assert.True(propNames.SequenceEqual(expectedPropNames));
         }
 
+        [Theory]
+        [MemberData(nameof(InstanceDescriptorTestData))]
+        public void ConvertToInstanceDescriptor(Font font, int expectedArguments)
```
Tanya-Solyanik, line 119: How is `font` from the theory data disposed?

JeremyKuhne: In the `try .. finally` below

src/System.Drawing.Common/src/System/Drawing/FontConverter.cs
JeremyKuhne: Presumably so that code generated from this is as neat as possible? No matter what the case it is probably best to just put back the .NET Framework behavior unless we're given a really compelling reason not to.

Tanya-Solyanik: makes sence

src/System.Drawing.Common/tests/System/Drawing/FontConverterTests.cs
Tanya-Solyanik: I have no idea how I missed it...

Tanya-Solyanik (approved):

PR #8994, opened 2023-04-14
Implement typed Font.To/FromLogFont

### LINKED ISSUE #8846: [API Proposal]: Trim friendly Font ToLogFont/FromLogFont
### Background and motivation

Right now WinForm use `System.Drawing.Common.Font.ToLogFont/FromLogFont` with their own interop structures. That API produce trimming warning. This API would serve same purpose as existing API, but provide trim friendly alternative.

### API Proposal

```csharp
namespace System.Drawing;

public sealed partial class Font
{
    public void ToLogFont<T>(ref T logFont, Graphics graphics) where T: struct, unmanaged  { throw null; }
    public static Font FromLogFont<T>(in T logFont) where T: struct, unmanaged { throw null; }
    public static Font FromLogFont<T>(in T logFont, IntPtr hdc) where T: struct, unmanaged { throw null; }
    public void ToLogFont<T>(ref T logFont) where T: struct, unmanaged { throw null; }
}
```

### API Usage

```csharp
[StructLayout(LayoutKind.Sequential, CharSet = CharSet.Unicode)]
public unsafe struct MyLOGFONT
{
    public int lfHeight;
    public int lfWidth;
    public int lfEscapement;
    public int lfOrientation;
    public int lfWeight;
    public byte lfItalic;
    public byte lfUnderline;
    public byte lfStrikeOut;
    public byte lfCharSet;
    public byte lfOutPrecision;
    public byte lfClipPrecision;
    public byte lfQuality;
    public byte lfPitchAndFamily;
    private fixed char _lfFaceName[LF_FACESIZE];
}

// And similar 
var f = new Font();
MyLOGFONT interopStruct = default;
f.ToLogFont(ref interopStruct);

```

### Alternative Designs

Ask users of this API, reimplement it itself within their application and make that code trim-friendly.

### Risks

_No response_

Accompanying implementation dotnet/runtime#70224

JeremyKuhne:
This adds the new typed API.

The new namespace conflicts with the Interop class so I had to change a number of files to add a static using for Interop.

Also changes WinForms to consume the new API.

Fixes #8846

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8994)


src/System.Drawing.Common/src/System/Drawing/Interop/LOGFONT.cs
```diff
@@ -0,0 +1,52 @@
...
+    public byte lfPitchAndFamily;
+    private fixed char _lfFaceName[LF_FACESIZE];
+
+#if NET7_0_OR_GREATER
+    [UnscopedRef]
+#endif
+    public Span<char> lfFaceName => MemoryMarshal.CreateSpan(ref _lfFaceName[0], LF_FACESIZE);
+
+    internal string AsString()
+#pragma warning disable format
```
Tanya-Solyanik, line 46: What is this pragma disabling?

JeremyKuhne: The IDE complaint about the code not being formatted "correctly".

Tanya-Solyanik: Why didn't you re-format then? Lines starting with a closing brace look strange.

JeremyKuhne: Because the line is crazy long- this is one of the standard ways to break it up. It's ugly, sure, but you can see it without scrolling.

JeremyKuhne: You have to break it up within the `{}` as that is the only place whitespace doesn't matter.

Tanya-Solyanik (approved):

PR #8995, opened 2023-04-14
Clean up EncoderParameter code

JeremyKuhne:
Was investigating whether there might be logic errors in the marshalling. I've typed the conversions more directly and removed the use of PtrToStructure for performance and avoiding unnecessary reflection.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8995)

### DIFF
--- a/src/System.Drawing.Common/src/System/Drawing/Imaging/EncoderParameter.cs
+++ b/src/System.Drawing.Common/src/System/Drawing/Imaging/EncoderParameter.cs
@@ -1,393 +1,349 @@
-// Licensed to the .NET Foundation under one or more agreements.
+﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Runtime.InteropServices;
 using Gdip = System.Drawing.SafeNativeMethods.Gdip;
 
-namespace System.Drawing.Imaging
+namespace System.Drawing.Imaging;
+
+[StructLayout(LayoutKind.Sequential)]
+public sealed unsafe class EncoderParameter : IDisposable
 {
-    [StructLayout(LayoutKind.Sequential)]
-    public sealed unsafe class EncoderParameter : IDisposable
+    private Guid _parameterGuid;
+    private readonly int _numberOfValues;
+    private readonly EncoderParameterValueType _parameterValueType;
+    private nint _parameterValue;
+
+    internal EncoderParameterPrivate ToNative() => new()
     {
-        private Guid _parameterGuid;                    // GUID of the parameter
-        private readonly int _numberOfValues;                    // Number of the parameter values
-        private readonly EncoderParameterValueType _parameterValueType;   // Value type, like ValueTypeLONG  etc.
-        private IntPtr _parameterValue;                 // A pointer to the parameter values
+        ParameterGuid = _parameterGuid,
+        ParameterValueType = _parameterValueType,
+        NumberOfValues = (uint)_numberOfValues,
+        ParameterValue = (void*)_parameterValue
+    };
 
-        ~EncoderParameter()
+    /// <summary>
+    ///  Gets/Sets the Encoder for the EncoderPameter.
+    /// </summary>
+    public Encoder Encoder
+    {
+        get => new Encoder(_parameterGuid);
+        set => _parameterGuid = value.Guid;
+    }
+
+    /// <summary>
+    ///  Gets the EncoderParameterValueType object from the EncoderParameter.
+    /// </summary>
+    public EncoderParameterValueType Type => _parameterValueType;
+
+    /// <summary>
+    ///  Gets the EncoderParameterValueType object from the EncoderParameter.
+    /// </summary>
+    public EncoderParameterValueType ValueType => _parameterValueType;
+
+    /// <summary>
+    ///  Gets the NumberOfValues from the EncoderParameter.
+    /// </summary>
+    public int NumberOfValues => _numberOfValues;
+
+    public void Dispose()
+    {
+        Dispose(disposing: true);
+        GC.KeepAlive(this);
+        GC.SuppressFinalize(this);
+    }
+
+    private void Dispose(bool disposing)
+    {
+        if (_parameterValue != 0)
         {
-            DisposeInternal();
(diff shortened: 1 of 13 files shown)


Tanya-Solyanik:
Why did you suspect an error in marshaling?

src/System.Drawing.Common/src/System/Drawing/Imaging/EncoderParametersPrivate.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.InteropServices;
+
+namespace System.Drawing.Imaging;
+
+[StructLayout(LayoutKind.Sequential)]
+internal unsafe struct EncoderParametersPrivate
```
Tanya-Solyanik, line 10: You name variables of this type `native` in the code. Should these types be renamed `EncoderParametersNative`?

src/System.Drawing.Common/src/System/Drawing/Imaging/EncoderParameters.cs
```diff
@@ -1,106 +1,66 @@
...
+                (nint)nativeParameters[i].ParameterValue);
         }
 
-        public void Dispose()
+        return parameters;
+    }
+
+    public void Dispose()
+    {
+        foreach (EncoderParameter p in Param)
```
Tanya-Solyanik, line 59: Should we test if `Param` is null before the `foreach`?

JeremyKuhne:
> Why did you suspect an error in marshaling?

Because the repro case doesn't repro. The code was pretty kludgy.

src/System.Drawing.Common/src/System/Drawing/Imaging/EncoderParameter.cs
```diff
@@ -1,393 +1,349 @@
...
+
+[StructLayout(LayoutKind.Sequential)]
+public sealed unsafe class EncoderParameter : IDisposable
 {
-    [StructLayout(LayoutKind.Sequential)]
-    public sealed unsafe class EncoderParameter : IDisposable
+    private Guid _parameterGuid;
+    private readonly int _numberOfValues;
+    private readonly EncoderParameterValueType _parameterValueType;
+    private nint _parameterValue;
```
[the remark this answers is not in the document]
JeremyKuhne: Because I wasn't being that exhaustive in the cleanup. The key thing is I didn't want `IntPtr` to avoid accidental cast issues down-level. This change is more focused on the generation and parsing of the struct data.

src/System.Drawing.Common/src/System/Drawing/Imaging/EncoderParameters.cs
```diff
@@ -1,106 +1,66 @@
...
-        ///
-        /// Also, in 64-bit platforms, 'Count' is aligned in 8 bytes (4 extra padding bytes) so we use IntPtr instead of Int32 to account for
-        /// that.
-        /// </summary>
-        internal unsafe IntPtr ConvertToMemory()
-        {
-            int size = sizeof(EncoderParameterPrivate);
+    internal unsafe nint ConvertToNative()
+    {
+        int length = Param.Length;
```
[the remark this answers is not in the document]
JeremyKuhne: We don't- it doesn't matter here as we end up with a valid struct with zero `Count`. The struct always has the initial `EncoderParameters*` in it so it's pretty safe.

Before this change it would return a technically invalid struct. Would that fall over? I don't know- I haven't gone through the GDI+ code in detail. Better to be safe than sorry. :)

src/System.Drawing.Common/src/System/Drawing/Imaging/EncoderParameters.cs
JeremyKuhne: It's still ok for it to throw here. I'd want some feedback and I don't think the compat risk is worth making throwing on setting `Param` to `null`.

src/System.Drawing.Common/src/System/Drawing/Imaging/EncoderParameter.cs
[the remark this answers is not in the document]
JeremyKuhne: Also, flipping this requires a lot more casting as the `Marshal` alloc methods all return `IntPtr`.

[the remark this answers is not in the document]
Tanya-Solyanik: Makes sense.

src/System.Drawing.Common/src/System/Drawing/Imaging/EncoderParametersNative.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.InteropServices;
+
+namespace System.Drawing.Imaging;
+
+[StructLayout(LayoutKind.Sequential)]
+internal unsafe struct EncoderParametersPrivate
```
Tanya-Solyanik, line 10: Oops, you renamed the file only! The struct is still `Private` :)

src/System.Drawing.Common/src/System/Drawing/Imaging/EncoderParametersNative.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    public uint Count;
```
Tanya-Solyanik, line 12: In the previous code, a comment was saying that the `Count` is an `IntPtr` for the correct alignment on 64 bit OSs. Why is it not relevant anymore?

Tanya-Solyanik: Ah, because the default `Packing` value is `0`, which is the OS behaviour.

Tanya-Solyanik (approved):

PR #8999, opened 2023-04-18
Add BufferScope for handling buffer rentals for interop

JeremyKuhne:
`BufferScope` is a ref struct that handles rental and return from the `ArrayPool` as well as:

- Initial stack allocated buffers
- Reacquisition of buffers (with content copying if necessary)
- Span semantics (converters, range operators, etc.)

It significantly simplifies creating temporary buffers, particularly in cases where you need to make multiple calls (for the buffer not being large enough).

Also add `[SkipLocalsInit]` in a few places (so the stack allocated buffer doesn't get zeroed out).

Did a global removal of unused usings (most of the file changes are just this).

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8999)

### DIFF
--- a/src/System.Windows.Forms.Primitives/tests/UnitTests/System/BufferScopeTests.cs
+++ b/src/System.Windows.Forms.Primitives/tests/UnitTests/System/BufferScopeTests.cs
@@ -1,0 +1,69 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Tests;
+
+public unsafe class BufferScopeTests
+{
+    [Fact]
+    public void Construct_WithStackAlloc()
+    {
+        using BufferScope<char> buffer = new(stackalloc char[10]);
+        Assert.Equal(10, buffer.Length);
+        buffer[0] = 'Y';
+        Assert.Equal("Y", buffer[..1].ToString());
+    }
+
+    [Fact]
+    public void Construct_WithStackAlloc_GrowAndCopy()
+    {
+        using BufferScope<char> buffer = new(stackalloc char[10]);
+        Assert.Equal(10, buffer.Length);
+        buffer[0] = 'Y';
+        buffer.EnsureCapacity(64, copy: true);
+        Assert.True(buffer.Length >= 64);
+        Assert.Equal("Y", buffer[..1].ToString());
+    }
+
+    [Fact]
+    public void Construct_WithStackAlloc_Pin()
+    {
+        using BufferScope<char> buffer = new(stackalloc char[10]);
+        Assert.Equal(10, buffer.Length);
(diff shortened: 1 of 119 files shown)


src/System.Windows.Forms.Primitives/tests/UnitTests/System/BufferScopeTests.cs
```diff
@@ -0,0 +1,69 @@
...
+        Assert.Equal("Z", buffer[..1].ToString());
+    }
+
+    [Fact]
+    public void Construct_GrowAndCopy()
+    {
+        using BufferScope<char> buffer = new(32);
+        Assert.True(buffer.Length >= 32);
+        buffer[0] = 'Y';
+        buffer.EnsureCapacity(64, copy: true);
```
Tanya-Solyanik, line 50: The other test is copying already, perhaps this one can e `copy: false`

src/System.Windows.Forms/src/System/Windows/Forms/ComponentModel/COM2Interop/COM2PropertyDescriptor.cs
```diff
@@ -984,15 +983,13 @@ public override unsafe void SetValue(object? component, object? value)
                     (uint)hr,
                     PInvoke.GetThreadLocale(),
                     b,
-                    255,
+                    (uint)buffer.Length - 2,
```
Tanya-Solyanik, line 986: Could you please clarify the `-2`, I see that it was used before, but don't know why. This size should be in TCHARs.

Tanya-Solyanik (dismissed):
Looks good except for the tests. Also I only skimmed across the "remove unused usings" part of the change assuming it was done by the analyzer.

src/System.Windows.Forms/src/System/Windows/Forms/ComponentModel/COM2Interop/COM2PropertyDescriptor.cs
JeremyKuhne: I can't clarify as I think the code is actually wrong. I think it is the common "I can't figure out the buffer stuff with StringBuilder so I'll pad all over the place". Not super important that we're wasting a few chars here so I'm not planning to spend time validating.

TCHAR is UTF-16 (`char`). It is only 8bit on Win9x.

src/System.Windows.Forms.Primitives/tests/UnitTests/System/BufferScopeTests.cs
JeremyKuhne: `copy: false` is the default and I want to test in both methods as one has a backing `stackalloc` and the other doesn't (and impacts the copy as a result).

Tanya-Solyanik (approved):

PR #9006, opened 2023-04-19
Fix Anchors for Nested containers

### LINKED ISSUE #8494: Updating the bounds of anchored controls leads to an incorrect location of the control
### .NET version

8.0.100-alpha.1.23061.8 + latest winforms dlls

Yes

Yes, it works in .NET Core 3.1 and .NET 5.0/6.0/7.0/8.0(without latest winforms dlls)

### Issue description

The X and Y values in **newRect** should be the same as values in **c.Bounds**.
GetImage

### Steps to reproduce

1. Download the test case project from here: https://microsoft-my.sharepoint.com/:u:/g/personal/v-peqiao_microsoft_com1/ERuveT35qepKkfb15DOVQl0BoIsTvsTQfUc1XyNSeqyYyg?e=FUvmnA
2. Open the **XPropertyGrid** project in VS.
3. Open the **XControl.cs** in **WFCTestLib** project>>**ReflectTools**>>**AutoPME** node.
4. Set the debug point at 3791 line.
5. Star to debug the **XPropertyGrid** project.
6. Check the X and Y values in **newRect** and **c.Bounds** when debug stopped at 3791 line in XControl.cs file.

Setbounds

### LINKED ISSUE #8933: Align and Anchor editors in PropertyGrid do not fill whole width
### .NET version

8.0.0-preview.4.23177.1

Yes

Confirmed that it works in .NET 6 and 7

### Issue description

Anchor and Align editors' UI in PropertyGrid is not filled to the whole width of container.

#### Current Behavior
.NET 8:
Screenshot 2023-04-03 174935
Screenshot 2023-04-03 142235

#### Expected behavior
.NET 7 and .NET Framework:

Screenshot 2023-04-03 175456
Screenshot 2023-04-03 142322

### Steps to reproduce

Issue is visible if PropertyGrid is wide enough (at least 400px) and doesn't seem to be affected by DPI settings. Sample app: 
[AlignAnchorEditorsTest.zip](https://github.com/dotnet/winforms/files/11140356/AlignAnchorEditorsTest.zip)

dreddy-work:
- Controls anchored within the nested containers are never getting chance to compute `AnchorInfo `in V2 layout and thus seeing issues mentioned in #8933. 
- Explicit bounds change on the control either by setting Properties like `Size`, `Location` or by calling public APIs `SetBounds` should recompute `AnchorsInfo` for the control. We documented this but missed in the validation. Issue #8494.
- Added unit tests covering these scenarios.

Fixes #8933 
Fixes #8494 

Anchor UI needs VS servicing to pick these changes.
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9006)


src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -4572,9 +4578,17 @@ internal virtual void AssignParent(Control? value)
                 OnParentChanged(EventArgs.Empty);
             }
 
-            UpdateAnchorsIfRequired();
             SetState(States.CheckedHost, false);
-            ParentInternal?.LayoutEngine.InitLayout(this, BoundsSpecified.All);
+
+            _forceAnchorCalculations = LocalAppContextSwitches.AnchorLayoutV2; // Parent has changed. AnchorsInfo should be recalculated.
```
Tanya-Solyanik, line 4583: Please move comment to a dedicated line.

src/System.Windows.Forms/src/System/Windows/Forms/Layout/DefaultLayout.cs
```diff
@@ -713,18 +714,33 @@ private static bool TryCalculatePreferredSize(IArrangedElement container, bool m
             return CommonProperties.GetAutoSize(container);
         }
 
+        private static void UpdateAnchorsIteratively(Control control)
+        {
+            UpdateAnchorInfoV2(control);
+
+            // If control does not have child controls or control is not yet ready to compute anchors, skip iterating over child controls.
```
Tanya-Solyanik, line 721: Please help me understand this comment, where do we test if this control has child controls?

Tanya-Solyanik: Maybe `_childControlsNeedAnchorLayout ` is `false` because it had not been initialized yet?

src/System.Windows.Forms/src/System/Windows/Forms/Layout/DefaultLayout.cs
```diff
@@ -713,18 +714,33 @@ private static bool TryCalculatePreferredSize(IArrangedElement container, bool m
...
+            if (!control._childControlsNeedAnchorLayout || control.Parent?._childControlsNeedAnchorLayout == true)
+            {
+                return;
+            }
+
+            // Compute anchors if any child controls require it.
+            ControlCollection controls = control.Controls;
+            for (int i = 0; i < controls.Count; i++)
```
Tanya-Solyanik, line 729: Why is this not a `foreach`?

Tanya-Solyanik (commented):
How had you tested this change besides the unit tests? Is it possible to test the actual anchoring info values in the unit test instead of null\not null?

src/System.Windows.Forms/src/System/Windows/Forms/Layout/DefaultLayout.cs
dreddy-work: I explicitly didn't exercise but have seen multiple places commenting that foreach is slower in iterating control collection.

dreddy-work:
> How had you tested this change besides the unit tests? Is it possible to test the actual anchoring info values in the unit test instead of null\not null?

I did integration tests along with unit tests. CTI did some manual testing as well. `AnchorInfo `is internal and not visible in user scenarios but i do have unit tests that actually exercise the expected locations and sizes of the control anchored after the `ResumeLayout`. Which implicitly verify the `AnchorInfo `values.

src/System.Windows.Forms/src/System/Windows/Forms/Layout/DefaultLayout.cs
dreddy-work: That is the case of Control/Container not having any child controls (default). If Control/Container has child controls that were anchored, then this flag is set on its parent.

The flag is set either when the anchor property of the control is set or when the Control is parented/added to the control collection but is not yet ready for anchor calculations due to `Suspendlayout`.

Tanya-Solyanik (dismissed):

PR #9015, opened 2023-04-20
Use native IAccessible handles directly

JeremyKuhne:
When using system IAccessible handles directly we get increased performance and easier to debug code. This also puts us a step closer to trimmability.

This removes the `SystemIAccessbileWrapper` class and instead puts shared `IAccessible` access code in `AccessibleObjectExtensions`.

Also:

- Moves a bit more towards CsWin32
- Change AgileComPointer to allow not tracking disposal
- Only derive AgileComPointer from tracker in debug builds
- Only build tracker in debug builds (don't want it slipping in and taking stack traces)
- Add new API to extract from ComWrappers to ComHelpers

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9015)


src/System.Windows.Forms/src/System/Windows/Forms/AccessibleObject.cs
```diff
@@ -1877,20 +2023,78 @@ internal static void ValidateChildID(ref object childID)
...
+                    int id = variant.data.lVal;
+                    if (id == (int)PInvoke.CHILDID_SELF)
+                    {
+                        return this;
+                    }
+                    else
+                    {
+                        // We never handled this case before. If it comes up should we make the call to get
+                        // the actual child object from the id?
+                        Debug.Fail($"{nameof(TryGetAccessibleObject)} got a child id of {id}");
```
dmitrii-drobotov, line 2040: I can hit this case with Inspect in MSAA mode, and opening ComboBoxes page in WinformsControlsTest app:
Screenshot 2023-04-21 164905
I'm not sure what's the cause of it, and I also see that without these changes when interacting with combo boxes in MSAA mode there were lots of `COMException`-s with `HRESULT.DISP_E_MEMBERNOTFOUND`, which were swallowed in `AccessibleObject.SystemIAccessibleWrapper`. So it might be an issue with ComboBox.

dmitrii-drobotov (commented):
Nice work!

src/System.Windows.Forms/src/System/Windows/Forms/TrackBar.TrackBarAccessibleObject.cs
```diff
@@ -25,16 +25,14 @@ public override Rectangle Bounds
             {
                 get
                 {
-                    if (!_owningTrackBar.IsHandleCreated || GetSystemIAccessibleInternal() is not Accessibility.IAccessible systemIAccessible)
+                    if (!_owningTrackBar.IsHandleCreated)
                     {
                         return Rectangle.Empty;
                     }
 
                     // The "NativeMethods.CHILDID_SELF" constant returns to the id of the trackbar,
```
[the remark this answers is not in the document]
JeremyKuhne: Fixed them

src/System.Windows.Forms/src/System/Windows/Forms/AccessibleObject.cs
JeremyKuhne: Hopefully with this change we can better diagnose and document things

src/System.Windows.Forms/src/System/Windows/Forms/AccessibleObjectExtensions.cs
```diff
@@ -0,0 +1,303 @@
...
+        }
+
+        result = HRESULT.E_NOINTERFACE;
+        return default;
+    }
+
+    public static Rectangle TryGetLocation(this AgileComPointer<IAccessible>? agile, int child)
+        => agile.TryGetLocation((VARIANT)child);
+
+    public static Rectangle TryGetLocation(this AgileComPointer<IAccessible>? agile, VARIANT child)
```
dreddy-work, line 27: Typically, in this code base, the TryGet* functions return a boolean value indicating success or failure, which allows the caller to verify the return value before using the objects. Is there a particular reason why this convention is being changed?

src/System.Windows.Forms/src/System/Windows/Forms/AccessibleObject.EnumVariantObject.cs
```diff
@@ -195,43 +224,15 @@ private unsafe void NextFromChildCollection(uint celt, IntPtr rgVar, uint* pCelt
             }
         }
 
-        /// <summary>
-        ///  Default behavior if there is no custom child collection or
-        ///  system-provided proxy to fall back on. In this case, we return
-        ///  an empty child collection.
-        /// </summary>
-        private unsafe void NextEmpty(uint celt, IntPtr rgvar, uint* pCeltFetched)
+        private ComScope<IEnumVARIANT> TryGetSystemEnum(out HRESULT result)
```
Tanya-Solyanik, line 227: Consider `TryGetSystemEnumVariant`

src/System.Windows.Forms/src/System/Windows/Forms/AccessibleObject.cs
```diff
@@ -179,51 +169,40 @@ private static bool InitializeCanNotifyClients()
...
-        {
-            if (_systemIAccessible.IsIAccessibleCreated)
-            {
-                return _systemIAccessible.get_accValue(NativeMethods.CHILDID_SELF);
-            }
-
-            return string.Empty;
-        }
-        set => _systemIAccessible.set_accValue(NativeMethods.CHILDID_SELF, value);
+        // This might be better to never return null?
```
Tanya-Solyanik, line 203: `null` usually has a special meaning - it tells accessibility tool to use the default. 

Did you mean to keep this comment?

src/System.Windows.Forms/src/System/Windows/Forms/AccessibleObject.cs
```diff
@@ -942,7 +978,39 @@ void IAccessible.accDoDefaultAction(object childID)
...
+    private VARIANT ChildIdToVARIANT(object childId)
+    {
+        if (childId is int integer)
+        {
+            return (VARIANT)integer;
+        }
+
+        if (childId is null)
+        {
+            return default;
```
Tanya-Solyanik, line 999: What is `default` for a VARIANT?

src/System.Windows.Forms/src/System/Windows/Forms/AccessibleObject.cs
```diff
@@ -1509,7 +1628,8 @@ int IAccessible.get_accHelpTopic(out string? pszHelpFile, object childID)
             }
         }
 
-        return _systemIAccessible?.get_accState(childID);
+        // Perhaps would be better to return AccessibleStates.None instead of null here.
```
Tanya-Solyanik, line 1631: Should this comment be converted into an investigation item?

src/System.Windows.Forms/src/System/Windows/Forms/AccessibleObject.cs
```diff
@@ -1786,39 +1912,57 @@ public virtual void Select(AccessibleSelection flags)
...
             objid,
-            ref IID.GetRef<global::Windows.Win32.UI.Accessibility.IAccessible>(),
-            ref acc);
+            IID.Get<UIA.IAccessible>(),
+            (void**)&accessible);
+
+        if (result.Succeeded && accessible is not null)
+        {
+#if DEBUG
+            // AccessibleObject is not set up for deterministic disposal.
```
Tanya-Solyanik, line 1932: Could you please clarify this comment. Accessibility tool holds a proxy to the accessible object. The object itself can be disposed when we are done.

src/System.Windows.Forms/src/System/Windows/Forms/AccessibleObject.cs
```diff
@@ -1875,20 +2019,78 @@ internal static void ValidateChildID(ref object childID)
...
+                if (id == (int)PInvoke.CHILDID_SELF)
+                {
+                    return this;
+                }
+                else
+                {
+                    // We never handled this case before. If it comes up should we make the call to get
+                    // the actual child object from the id?
+                    Debug.Fail($"{nameof(TryGetAccessibleObject)} got a child id of {id}");
+                    return null;
```
Tanya-Solyanik, line 2037: I think we can dispose the VARIANT in this case.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.ListViewItemBaseAccessibleObject.cs
```diff
@@ -285,24 +271,16 @@ public override void Select(AccessibleSelection flags)
...
 
-        internal override void Toggle()
-        {
-            _owningItem.Checked = !_owningItem.Checked;
+            // In Everett, the ListBox accessible children did not have any selection capability.
+            // In Whidbey, they delegate the selection capability to OLEACC.
+            // However, OLEACC does not deal w/ several Selection flags: ExtendSelection, AddSelection, RemoveSelection.
+            // OLEACC instead throws an ArgumentException.
+            // Since Whidbey API's should not throw an exception in places where Everett API's did not, we catch
+            // the ArgumentException and fail silently.
```
Tanya-Solyanik, line 281: I don't see where this exception is caught, could you please double-check?

src/System.Windows.Forms/src/System/Windows/Forms/AccessibleObject.cs
JeremyKuhne: `VT_EMPTY` which is `null` when converted to `object`.

src/System.Windows.Forms/src/System/Windows/Forms/AccessibleObject.cs
JeremyKuhne: I don't think so. I don't think we have bandwidth to track or dig into these unless we run across it during debugging.

src/System.Windows.Forms/src/System/Windows/Forms/AccessibleObject.cs
JeremyKuhne: `AccessibleObject` isn't `IDisposable` and as such these handles always get finalized.

src/System.Windows.Forms/src/System/Windows/Forms/AccessibleObject.cs
JeremyKuhne: Integers don't need disposed- they are embedded in the `VARIANT`.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.ListViewItemBaseAccessibleObject.cs
JeremyKuhne: There is no throwing from `TrySelect()`, we're dealing directly with `HRESULT`. Only a critical exception is possible (no more memory, runtime corruption, etc.).

src/System.Windows.Forms/src/System/Windows/Forms/AccessibleObjectExtensions.cs
JeremyKuhne: Yes, we normally should do the other pattern. In the case of `AccessibleObject`, there is no checking of results, everything just has a default fall-back value. I could rewrite all of the code to out a value and add addition checks to return a `bool` that is never checked, but it didn't seem worth it.

In this case, the `Try` means it won't throw. That's normal. Note that in a lot of the other related interop code we return scopes which require being the return value to be easily used in `using` blocks.

Tanya-Solyanik (approved):

src/System.Windows.Forms/src/System/Windows/Forms/AccessibleObject.cs
JeremyKuhne: I updated the comment. The same comment applies here as in the other case- I don't think we have bandwidth to track this. It is something we can look at when debugging and add the reasoning as to why it is inconsistent or make it consistent with real scenarios.

Tanya-Solyanik: makes sense

PR #9018, opened 2023-04-21
Fix null reference in PermonV2 mode.

dreddy-work:
Null check should be explicit to avoid further access of the Item object. It is not uncovered until we enabled PermonV2 mode for the designer.

Fixes #[5125](https://github.com/microsoft/winforms-designer/issues/5125)
Fixes #[5124](https://github.com/microsoft/winforms-designer/issues/5124)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9018)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripRenderer.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripRenderer.cs
@@ -634,7 +634,7 @@ namespace System.Windows.Forms
                 Point[] arrow = null;
 
                 // We need the Elvis operator here, since at design time at this point Item can be null.
-                if (e.Item?.DeviceDpi != previousDeviceDpi && DpiHelper.IsPerMonitorV2Awareness)
+                if (e.Item is not null && e.Item.DeviceDpi != previousDeviceDpi && DpiHelper.IsPerMonitorV2Awareness)
                 {
                     previousDeviceDpi = e.Item.DeviceDpi;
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ToolStripRenderer.cs
```diff
@@ -634,7 +634,7 @@ protected virtual void OnRenderArrow(ToolStripArrowRenderEventArgs e)
                 Point[] arrow = null;
 
                 // We need the Elvis operator here, since at design time at this point Item can be null.
```
halgab, line 636:
```suggestion
                // We need to check Item for null here, since it can happen at this point at design time.
```

JeremyKuhne: `// At design time Item can be null` is more to the point.

dreddy-work: This is targeting 6.0 and existing comment. FYI - Main had changed and has this fix already. If this comment needs to be changed there, i am not planning on creating a PR just for that against main.

JeremyKuhne (approved):

dreddy-work:
approved over email for servicing.

Tanya-Solyanik (approved):

PR #9039, opened 2023-04-25
Add a scratch project to the solution

JeremyKuhne:
This is to simplify experimenting and looking at repros.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9039)

### DIFF
--- a/src/System.Windows.Forms/tests/IntegrationTests/ScratchProject/Program.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/ScratchProject/Program.cs
@@ -1,0 +1,17 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace ScratchProject;
+
+// This project is meant for temporary testing and experimenting and should be kept as simple as possible.
+
+internal static class Program
+{
+    [STAThread]
+    static void Main()
+    {
+        Application.EnableVisualStyles();
+        Application.SetHighDpiMode(HighDpiMode.PerMonitorV2);
+        Application.Run(new Form1());
+    }
+}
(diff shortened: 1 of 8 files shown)


JeremyKuhne:
Note that we could update the build scripts to ignore the existing files in the scratch project (see the comment in .gitignore). If someone wants to follow up with that...

src/System.Windows.Forms/tests/IntegrationTests/ScratchProject/Program.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace ScratchProject;
+
+// This project is meant for temporary testing and experimenting and should be kept as simple as possible.
+
+internal static class Program
+{
+    [STAThread]
```
dreddy-work, line 10: Any reason you can't use latest template that utilize code generation here?

Also, default is SystemAware.

JeremyKuhne: I deliberately want the non-default here so we look good for the per monitor in our experiments. The code generation stuff doesn't work without significant bending over backwards as we're directly referencing the projects and can't set `UseWindowsForms`. I feel it is better to keep the project as simple as possible to make it easier to understand.

dreddy-work (approved):

RussKie:
I'm curious, why is `WinformsControlsTest` not good?

Tanya-Solyanik:
I prefer minimal projects when debugging or reproing specific bugs, especially when tracing/memory monitoring is on.

PR #9056, opened 2023-04-30
Implement task dialog hyperlink proposal

### LINKED ISSUE #1199: Task Dialog: Improve hyperlink usage
:arrow_double_down::arrow_double_down: [**CLICK HERE TO SCROLL DOWN TO THE FINAL PROPOSAL**](https://github.com/dotnet/winforms/issues/1199#issuecomment-822142698) :arrow_double_down::arrow_double_down:

----

Hi,

as requested by @RussKie in https://github.com/dotnet/winforms/issues/146#issuecomment-502493259, I've opened a new issue for this topic.
**Note:** The listed options don't need to change the currently proposed API in an incompatible way, so this issue can be addressed at a later time.

The native Windows Task Dialog allows to use hyperlinks in certain text elements (text/content, footer, expander) by setting the `TDF_ENABLE_HYPERLINKS` flag. This allows you to specify a link with the form `<a href="mylink">text</a>` (similar to HTML). When the user clicks on the link, the `TDN_HYPERLINK_CLICKED` is sent to the callback with the content of the `href` attribute as parameter.

This is reflected in the currently proposed API in #146: To use links, you set the `TaskDialogPage.HyperlinksEnabled` property to `true`, and then you can set e.g. 
```c#
page.Text = "This text contains <a href=\"link1\">a link</a>.";
```
to show a link in the task dialog text:

grafik

Then, you have to handle the `TaskDialogPage.HyperlinkClicked` event to get notified when the user clicks a link. That event will provide you the text of the `href` attribute, in that case `link1`.

However, there are some problems with this approach (as is simply a mapping of the native API):
* Users need to know the exact syntax to specify a link (which is similar to HTML), and in source code you need to escape the double quotes.
* There is a risk that when you enable hyperlinks and then show strings from external sources, they might get misinterpreted as hyperlink even if you actually just want it to show as plain text. As noted in the documentation for [`TDF_ENABLE_HYPERLINKS`](https://docs.microsoft.com/en-us/windows/desktop/api/commctrl/ns-commctrl-_taskdialogconfig#tdf_enable_hyperlinks), this could cause security vulnerabilities, for example if you did something like `Process.Start(e.Hyperlink)` in the `HyperlinkClicked` event.
(A safer way is not to use links like `https://...` directly in the `href`, but instead use keys like `linkA`, and then in the event, compare the key to open the corresponding link.)
* In order to prevent strings from external sources to be interpreted as links (or when you actually want to display a string like `<a href="..."></a>` as part of the text directly without interpreting as hyperlink), you need to replace `"<a"` e.g. with `"<\u200Ba"` (zero width space), so that the task dialog interprets `<` as literal character.
A minor downside of this is that if you then copy the dialog's text contents with <kbd>Ctrl</kbd>+<kbd>C</kbd>, that invisible character is also copied.
(I didn't find another way for this, as the HTML/XML-like `&lt;` doesn't seem to work; and while replacing `"<a"` with `"<<a></a>a"` (which I thought previously) would fix the copy problem with the invisible space, it cannot be used as it creates empty hyperlinks that are reachable by pressing <kbd>Tab ↹</kbd>.)
* When you enable `HyperlinksEnabled` and there is at least one link (`<a ...>`) present in the text, an `&` character will be interpreted to specify a mnemonic char, instead of displaying it directly (which is the behavior when `HyperlinksEnabled` is not set), although then actually nothing happens when you press <kbd>Alt</kbd> and the character.
Therefore, you would generally need to replace `&` with `&&` in the strings when `HyperlinksEnabled` is set and you have at least one link in the text.
(But if you then copy the task dialog's text contents with <kbd>Ctrl</kbd>+<kbd>C</kbd>, it will copy the escaped form (`&&`) instead of the displayed form (`&`).)

## Option A

As suggested by @GSPP:
Don't change the existing API, but add new methods that can be used to retrieve strings with the correct syntax when `EnableHyperlinks` is set, for example::

```c#
    public static string GetHyperlinkText(string href, string text)
    {
        // Don't allow "</a>" in the link text because we cannot escape that.
        if (text.Contains("</a>", StringComparison.OrdinalIgnoreCase))
            throw new ArgumentException();
        if (href.Contains("\"", StringComparison.OrdinalIgnoreCase))
            throw new ArgumentException();

        return $"<a href=\"{href}\">{text.Replace("&", "&&")}</a>";
    }

    public static string GetNonHyperlinkText(string text)
    {
        return text.Replace("&", "&&").Replace("<a", "<\u200Ba").Replace("<A", "<\u200BA");
    }
```

Then, you could set the text like this:
```c#
    page.Text = TaskDialog.GetHyperlinkText("myLink", "Hello World & others!") +
        TaskDialog.GetNonHyperlinkText(" External text: <a href=\"test\">&</a>");
```
grafik

With the helper methods, you wouldn't need to specify the exact syntax for the link and you wouldn't need to escape special characters like `<` by yourself.

## Option B

As suggested by @jnm2: Also don't change the existing API but add a new method
```c#
    public static string FormatWithHyperlinks(FormattableString str)
```
on the `TaskDialog` which you then could call with an interpolated string, that takes the `href` as argument and the link text as format string (or the other way round):
```c#
    page.Text = TaskDialog.FormatWithHyperlinks(
        $"Some text, then {"https://foo/":a link} & {"https://bar/":another link}; <a href=\"x&y\">no link</a>");

    // Or:
    page.Text = TaskDialog.FormatWithHyperlinks(
        $"Some text, then {"a link":https://foo/} & {"another link":https://bar/}; <a href=\"x&y\">no link</a>");
```

grafik

(However, with this option there will probably need to be another method that allows to escape strings (like `GetNonHyperlinkText()` in Option A) if you then want to embed other strings which shouldn't get interpreted as link.)

Thanks!

jnm2:
Closes #1199

## Proposed changes

- Exposes the Windows Task Dialog API in regard to enabling hyperlinks to be shown (TDF_ENABLE_HYPERLINKS) and hyperlink clicks to be reacted to (TDN_HYPERLINK_CLICKED). This is conceptually 1:1 with the underlying Windows API and does not contain code that interprets or executes the contents of hrefs.

## Customer Impact

- Customers who show task dialogs containing links are unable to move to the new Windows Forms API, away from handwritten implementations or community libraries for TaskDialog which have fewer capabilities in other regards than the Windows Forms API.

## Regression? 

- No

## Risk

- Minimal; this feature is opt-in, and behavior without opting in remains the same as without this feature being implemented.

## Screenshots

TaskDialog demo

## Test methodology

- A new Task Dialog demo is added along with the existing ones, putting all aspects of this feature through their paces. The screenshot for this is above.
- I was not able to identify any existing unit tests for functionality that could be extended to cover hyperlinks.

## Accessibility testing

[Please advise: is this relevant for lighting up the hyperlink feature of the Windows Task Dialog? The winforms codebase has no control over the accessibility of this built-in Windows idiom.]

 

## Test environment(s) 

```
.NET SDK:
 Version:   8.0.100-preview.3.23178.7
 Commit:    e300b0e1e6

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.22621
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\8.0.100-preview.3.23178.7\

.NET workloads installed:
There are no installed workloads to display.

Host:
  Version:      8.0.0-preview.4.23221.1
  Architecture: x64
  Commit:       ab2b80d06d

.NET SDKs installed:
  7.0.203 [C:\Program Files\dotnet\sdk]
  8.0.100-preview.3.23178.7 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 6.0.16 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 7.0.5 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 8.0.0-preview.3.23177.8 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 3.1.32 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 5.0.17 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.12 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.16 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 7.0.5 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 8.0.0-preview.3.23174.8 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 8.0.0-preview.4.23221.1 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 3.1.32 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 5.0.17 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 6.0.6 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 6.0.8 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 6.0.9 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 6.0.16 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 7.0.5 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 8.0.0-preview.3.23178.1 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]

Other architectures found:
  x86   [C:\Program Files (x86)\dotnet]
    registered at [HKLM\SOFTWARE\dotnet\Setup\InstalledVersions\x86\InstallLocation]

Environment variables:
  Not set

global.json file:
  C:\Users\Joseph\Source\Repos\winforms\global.json
```

I tested and created the screenshot on a machine using 200% DPI.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9056)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TaskDialog.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TaskDialog.cs
@@ -271,7 +271,8 @@ public partial class TaskDialog : IWin32Window
         (((GCHandle)lpRefData).Target as TaskDialog)!.HandleTaskDialogCallback(
             hwnd,
             msg,
-            wParam);
+            wParam,
+            lParam);
 
     private static bool IsTaskDialogButtonCommitting(TaskDialogButton? button)
     {
@@ -785,7 +786,8 @@ public partial class TaskDialog : IWin32Window
     private HRESULT HandleTaskDialogCallback(
         HWND hWnd,
         TASKDIALOG_NOTIFICATIONS notification,
-        IntPtr wParam)
+        IntPtr wParam,
+        IntPtr lParam)
     {
         Debug.Assert(_boundPage is not null);
 
@@ -1038,6 +1040,12 @@ public partial class TaskDialog : IWin32Window
 
                 case TASKDIALOG_NOTIFICATIONS.TDN_HELP:
                     _boundPage.OnHelpRequest(EventArgs.Empty);
+                    break;
+
+                case TASKDIALOG_NOTIFICATIONS.TDN_HYPERLINK_CLICKED:
+                    string? linkHref = Marshal.PtrToStringUni(lParam);
+                    Debug.Assert(linkHref is not null);
+                    _boundPage.OnLinkClicked(new TaskDialogLinkClickedEventArgs(linkHref));
                     break;
             }
         }
--- a/src/System.Windows.Forms/src/PublicAPI.Unshipped.txt
+++ b/src/System.Windows.Forms/src/PublicAPI.Unshipped.txt
@@ -1,0 +1,7 @@
+System.Windows.Forms.TaskDialogLinkClickedEventArgs
+System.Windows.Forms.TaskDialogLinkClickedEventArgs.LinkHref.get -> string!
+System.Windows.Forms.TaskDialogLinkClickedEventArgs.TaskDialogLinkClickedEventArgs(string! linkHref) -> void
+System.Windows.Forms.TaskDialogPage.EnableLinks.get -> bool
+System.Windows.Forms.TaskDialogPage.EnableLinks.set -> void
+System.Windows.Forms.TaskDialogPage.LinkClicked -> System.EventHandler<System.Windows.Forms.TaskDialogLinkClickedEventArgs!>?
+System.Windows.Forms.TaskDialogPage.OnLinkClicked(System.Windows.Forms.TaskDialogLinkClickedEventArgs! e) -> void
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TaskDialogLinkClickedEventArgs.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TaskDialogLinkClickedEventArgs.cs
@@ -1,0 +1,25 @@
+﻿namespace System.Windows.Forms;
+
+/// <summary>
+/// Provides data for the <see cref="TaskDialogPage.LinkClicked"/> event.
+/// </summary>
+public class TaskDialogLinkClickedEventArgs : EventArgs
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="TaskDialogLinkClickedEventArgs"/> class.
+    /// </summary>
+    public TaskDialogLinkClickedEventArgs(string linkHref)
+    {
+        LinkHref = linkHref;
+    }
+
+    /// <summary>
+    /// Gets the value of the <c>href</c> attribute of the link that the user clicked.
+    /// </summary>
+    /// <remarks>
+    /// Note: In order to avoid possible security vulnerabilities when showing content
+    /// from unsafe sources in a task dialog, you should always verify the value of this
+    /// property before actually opening the link.
+    /// </remarks>
+    public string LinkHref { get; }
+}
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TaskDialogPage.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TaskDialogPage.cs
@@ -513,6 +513,50 @@ public class TaskDialogPage
     }
 
     /// <summary>
+    /// <para>
+    ///   Gets or sets a value that specifies whether the task dialog should
+    ///   interpret strings in the form <c>&lt;a href="target"&gt;link Text&lt;/a&gt;</c>
+    ///   as hyperlink when specified in the <see cref="Text"/>,
+    ///   <see cref="TaskDialogExpander.Text"/>,
+    ///   or <see cref="TaskDialogFootnote.Text"/> properties.
+    ///   When the user clicks on such a link, the <see cref="LinkClicked"/>
+    ///   event is raised, containing the value of the <c>target</c> attribute.
+    /// </para>
+    /// </summary>
+    /// <value>
+    ///   <see langword="true"/> to enable links; otherwise, <see langword="false"/>.
+    ///   The default value is <see langword="false"/>.
+    /// </value>
+    /// <remarks>
+    /// <para>
+    ///   The Task Dialog will not actually execute any links.
+    ///   Link execution must be handled in the <see cref="LinkClicked"/> event.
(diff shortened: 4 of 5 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/TaskDialog.cs
```diff
@@ -1039,6 +1041,12 @@ internal void UpdateCaption(string? caption)
                 case TASKDIALOG_NOTIFICATIONS.TDN_HELP:
                     _boundPage.OnHelpRequest(EventArgs.Empty);
                     break;
+
+                case TASKDIALOG_NOTIFICATIONS.TDN_HYPERLINK_CLICKED:
+                    string? linkHref = Marshal.PtrToStringUni(lParam);
+                    Debug.Assert(linkHref is not null);
```
jnm2, line 1047: There doesn't seem to be a way to get a null pointer in `lParam`, always a pointer to an empty string, even when there is no `href` attribute, e.g. `<a>Missing href</a>`.

RussKie (approved):

dreddy-work (approved):

PR #9057, opened 2023-04-30
Make task dialog sample menu window readable on high-DPI settings

jnm2:
## Proposed changes

Replace manual coordinate calculation with autosizing with FlowLayoutPanel so that the contributor samples window is readable on systems with high-DPI settings.

## Customer Impact

- Makes no difference to customers, only contributors to the repository.

## Regression? 

- Not that I know of

## Risk

- None, does not affect shipping code

### Before

image

### After

image

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9057)

### DIFF
--- a/src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/TaskDialogSamples.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/TaskDialogSamples.cs
@@ -3,7 +3,6 @@
 // See the LICENSE file in the project root for more information.
 
 using System.ComponentModel;
-using System.Drawing;
 using Timer = System.Windows.Forms.Timer;
 
 namespace WinformsControlsTest;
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/TaskDialogSamples.cs
```diff
@@ -14,21 +13,33 @@ public TaskDialogSamples()
     {
         this.Text = "Task Dialog Demos";
 
-        int currentButtonCount = 0;
-        void AddButtonForAction(string name, Action action)
+        AutoSize = true;
+        AutoSizeMode = AutoSizeMode.GrowAndShrink;
+        MinimizeBox = false;
+        MaximizeBox = false;
```
RussKie, line 19: Why are these disabled?

huynhdev24: What if the font size changes?

jnm2: > Why are these disabled?

To make the window title readable, and because the window is tiny and I thought likely doesn't need them.

dreddy-work:
I am fine taking this sample. In general, there is Font propagation issue here for high DPI device. This is a work around until that is fixed.

dreddy-work (approved):

Tanya-Solyanik (approved):

Olina-Zhang:
Verified this on .NET 8.0 test pass build: 8.0.100-preview.5.23276.4, task dialog has a good display now on HDPI machine:
image

PR #9063, opened 2023-05-02
Scale ContextMenuStrip for high DPI devices

### LINKED ISSUE #9044: ContextMenu items with checkboxes are displayed badly in PMv2 HDPI
### .NET version

.NET 7

Yes

_No response_

### Issue description

In Per-Monitor v2 scenario, the contextmenu items with checkbox causes the contextmenu items having wrong height and the checkbox is not fully visible.
My scenario is 175% primary screen, 100% secondary screen. The sample app looks fine on the primary one, but gets broken on the secondary.

image
image
[DpiTest.zip](https://github.com/dotnet/winforms/files/11331626/DpiTest.zip)

### Steps to reproduce

Set up a primary screen to 175%, secondary to 100%
Run the attached DpiTest project on the secondary screen.
One button opens simple context menu, while the second button opens a contextmenu where the first item is checked.
On the secondary screen, the second contextmenu is not displayed correctly

dreddy-work:
Fixes #9044

There are two problems we're facing when trying to scale `ContextMenuStrip`:

1. `ContextMenuStrip `is a top level window and thus only receive "WM_DPICHANGED" message but not WM_DPICHANGED_BEFOREPARENT/WM_DPICHANGED_AFTERPARENT.  Unfortunately, we do not handle scaling of controls in "WM_DPICHANGED" message. In WinForms, "WM_DPICHANGED" message is intended for Top-level Forms/ContainerControls.              As a result, the `ContextMenuStrip `window doesn't scale itself when moved from one monitor to another on the "PerMonitorV2" process.

2. When `ContextMenuStrip `changes to invisible (with "visible" set to false), owner of the window is changed to a `ParkingWindow `that may possibly be parked on primary monitor. The `GetDpiForWindow()` API on `ContextMenuStrip` thus returns the DPI of the primary monitor.

Because of this inconsistency, we intentionally recreate the handle that triggers scaling according to the new DPI, after setting the "visible" property.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9063)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ContextMenuStrip.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ContextMenuStrip.cs
@@ -120,9 +120,11 @@ public class ContextMenuStrip : ToolStripDropDownMenu
         base.SetVisibleCore(visible);
 
         // There are two problems we're facing when trying to scale ContextMenuStrip:
-        //    1. When the "visible" property is set to false, the control window doesn't receive the "DPI_CHANGED" message.
+        //    1. ContextMenuStrip is a top level window and thus only receive "WM_DPICHANGED" message but not WM_DPICHANGED_BEFOREPARENT/WM_DPICHANGED_AFTERPARENT.
+        //       Unfortunately, we do not handle scaling of controls in "WM_DPICHANGED" message. In WinForms, "WM_DPICHANGED" message is intended for Top-level Forms/ContainerControls.
         //       As a result, the ContextMenuStrip window doesn't scale itself when moved from one monitor to another on the "PerMonitorV2" process.
-        //    2. If a window is invisible(with "visible" set to false), the "GetDpiForWindow()" API returns the DPI of the primary monitor.
+        //    2. When ContextMenuStrip changes to invisible(with "visible" set to false), owner of the window is changed to a ParkingWindow that may possibly parked on primary monitor.
+        //       The "GetDpiForWindow()" API on ContextMenuStrip thus returns the DPI of the primary monitor.
         // Because of this inconsistency, we intentionally recreate the handle that triggers scaling according to the new DPI, after setting the "visible" property.
         if (visible
             && IsHandleCreated
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ContextMenuStrip.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ContextMenuStrip.cs
@@ -118,5 +118,18 @@ public class ContextMenuStrip : ToolStripDropDownMenu
         }
 
         base.SetVisibleCore(visible);
+
+        // There are two problems we're facing when trying to scale ContextMenuStrip:
+        //    1. When the "visible" property is set to false, the control window doesn't receive the "DPI_CHANGED" message.
+        //       As a result, the ContextMenuStrip window doesn't scale itself when moved from one monitor to another on the "PerMonitorV2" process.
+        //    2. If a window is invisible(with "visible" set to false), the "GetDpiForWindow()" API returns the DPI of the primary monitor.
+        // Because of this inconsistency, we intentionally recreate the handle that triggers scaling according to the new DPI, after setting the "visible" property.
+        if (visible
+            && IsHandleCreated
+            && DpiHelper.IsPerMonitorV2Awareness
+            && DeviceDpi != (int)PInvoke.GetDpiForWindow(this))
+        {
+            RecreateHandle();
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ContextMenuStrip.cs
```diff
@@ -118,5 +118,18 @@ protected override void SetVisibleCore(bool visible)
         }
 
         base.SetVisibleCore(visible);
+
+        // There are two problems we're facing when trying to scale ContextMenuStrip:
+        //    1. When the "visible" property is set to false, the control window doesn't receive the "DPI_CHANGED" message.
```
Tanya-Solyanik, line 123: Please spell message name as it's spelled in the docs - WM_DPICHANGED

src/System.Windows.Forms/src/System/Windows/Forms/ContextMenuStrip.cs
```diff
@@ -118,5 +118,18 @@ protected override void SetVisibleCore(bool visible)
...
+        //       As a result, the ContextMenuStrip window doesn't scale itself when moved from one monitor to another on the "PerMonitorV2" process.
+        //    2. If a window is invisible(with "visible" set to false), the "GetDpiForWindow()" API returns the DPI of the primary monitor.
+        // Because of this inconsistency, we intentionally recreate the handle that triggers scaling according to the new DPI, after setting the "visible" property.
+        if (visible
+            && IsHandleCreated
+            && DpiHelper.IsPerMonitorV2Awareness
+            && DeviceDpi != (int)PInvoke.GetDpiForWindow(this))
+        {
+            RecreateHandle();
```
Tanya-Solyanik, line 132: Is this causing any flickering?

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripMenuItem.cs
```diff
@@ -204,16 +204,21 @@ private void Initialize()
...
-            _scaledCheckMarkBitmapSize = DpiHelper.LogicalToDeviceUnits(s_checkMarkBitmapSize);
+            ScaleDpiConstants();
         }
 
         Overflow = ToolStripItemOverflow.Never;
         MouseDownAndUpMustBeInSameItem = false;
         SupportsDisabledHotTracking = true;
     }
 
+    private void ScaleDpiConstants(int newDpi = 0)
```
Tanya-Solyanik, line 215: Suggestion: `ScaleConstantsForDpi`

Tanya-Solyanik (commented):
Looks good. Please get this tested by the CTI. Are there more files that should implementrescaling of constants when DPI changes?

src/System.Windows.Forms/src/System/Windows/Forms/ContextMenuStrip.cs
dreddy-work: No. checked it.

dreddy-work:
> Looks good. Please get this tested by the CTI. Are there more files that should implementrescaling of constants when DPI changes?

This is specific to `ContextMenuStrip`. I will have to check other components in our codebase.

Tanya-Solyanik (approved):

src/System.Windows.Forms/src/System/Windows/Forms/ContextMenuStrip.cs
```diff
@@ -118,5 +118,20 @@ protected override void SetVisibleCore(bool visible)
...
         base.SetVisibleCore(visible);
+
+        // There are two problems we're facing when trying to scale ContextMenuStrip:
+        //    1. ContextMenuStrip is a top level window and thus only receive "WM_DPICHANGED" message but not WM_DPICHANGED_BEFOREPARENT/WM_DPICHANGED_AFTERPARENT.
+        //       Unfortunately, we do not handle scaling of controls in "WM_DPICHANGED" message. In WinForms, "WM_DPICHANGED" message is intended for Top-level Forms/ContainerControls.
+        //       As a result, the ContextMenuStrip window doesn't scale itself when moved from one monitor to another on the "PerMonitorV2" process.
+        //    2. When ContextMenuStrip changes to invisible(with "visible" set to false), owner of the window is changed to a ParkingWindow that may possibly parked on primary monitor.
+        //       The "GetDpiForWindow()" API on ContextMenuStrip thus returns the DPI of the primary monitor.
+        // Because of this inconsistency, we intentionally recreate the handle that triggers scaling according to the new DPI, after setting the "visible" property.
+        if (visible
```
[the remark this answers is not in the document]
dreddy-work: First creation is always scaled well because of "Scaling " done in `OnHandleCreation`(). Issue is only when we try to show it again, but DPI is different on the monitor.

I think recreate handle in this scenario is reasonable given the scenario.

PR #9072, opened 2023-05-05
Enable nullability in ToolStripItemCollection

gpetrou:
## Proposed changes

- Enable nullability in ToolStripItemCollection.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9072)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItemCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItemCollection.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.Collections;
 using System.ComponentModel;
 using System.Drawing;
@@ -16,7 +14,7 @@ namespace System.Windows.Forms;
 [ListBindable(false)]
 public class ToolStripItemCollection : ArrangedElementCollection, IList
 {
-    private readonly ToolStrip _owner;
+    private readonly ToolStrip? _owner;
     private readonly bool _itemsCollection;
     private readonly bool _isReadOnly;
 
@@ -24,12 +22,12 @@ public class ToolStripItemCollection : ArrangedElementCollection, IList
     // references. Note this is not thread safe - but WinForms has to be run in a STA anyways.
     private int _lastAccessedIndex = -1;
 
-    internal ToolStripItemCollection(ToolStrip owner, bool itemsCollection)
+    internal ToolStripItemCollection(ToolStrip? owner, bool itemsCollection)
         : this(owner, itemsCollection, isReadOnly: false)
     {
     }
 
-    internal ToolStripItemCollection(ToolStrip owner, bool itemsCollection, bool isReadOnly)
+    internal ToolStripItemCollection(ToolStrip? owner, bool itemsCollection, bool isReadOnly)
     {
         _owner = owner;
         _itemsCollection = itemsCollection;
@@ -55,7 +53,7 @@ public class ToolStripItemCollection : ArrangedElementCollection, IList
     /// <summary>
     ///  Retrieves the child control with the specified key.
     /// </summary>
-    public virtual ToolStripItem this[string key]
+    public virtual ToolStripItem? this[string? key]
     {
         get
         {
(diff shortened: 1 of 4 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItemCollection.cs
```diff
@@ -378,11 +376,11 @@ private bool IsValidIndex(int index)
     /// <summary>
     ///  Do proper cleanup of ownership, etc.
     /// </summary>
-    private void OnAfterRemove(ToolStripItem item)
+    private void OnAfterRemove(ToolStripItem? item)
```
dreddy-work, line 379: all references of this method should be making sure `item `is not null. Looking at use case in `RemoveAt ` I would throw index out of range instead of propagating `null`.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItemCollection.cs
```diff
@@ -432,7 +430,7 @@ public void RemoveAt(int index)
             throw new NotSupportedException(SR.ToolStripItemCollectionIsReadOnly);
         }
 
-        ToolStripItem item = null;
+        ToolStripItem? item = null;
```
dreddy-work, line 433: nit: remove default assignment.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItemCollection.cs
```diff
@@ -292,14 +290,14 @@ private void FindInternal(string key, bool searchAllChildren, ToolStripItemColle
...
     void IList.Clear() { Clear(); }
     bool IList.IsFixedSize { get { return ((IList)InnerList).IsFixedSize; } }
-    bool IList.Contains(object value) { return InnerList.Contains(value); }
+    bool IList.Contains(object? value) { return InnerList.Contains(value); }
     void IList.RemoveAt(int index) { RemoveAt(index); }
-    void IList.Remove(object value) { Remove(value as ToolStripItem); }
-    int IList.Add(object value) { return Add(value as ToolStripItem); }
-    int IList.IndexOf(object value) { return IndexOf(value as ToolStripItem); }
-    void IList.Insert(int index, object value) { Insert(index, value as ToolStripItem); }
+    void IList.Remove(object? value) { Remove((value as ToolStripItem)!); }
```
dreddy-work, line 295: would change these like direct cast that will throw if null here it self.
```suggestion
    void IList.Remove(object? value) { Remove( (ToolStripItem)(value)); }
```

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItemCollection.cs
```diff
@@ -175,7 +173,7 @@ public virtual void Clear()
             return;
         }
 
-        ToolStripOverflow overflow = null;
+        ToolStripOverflow? overflow = null;
```
dreddy-work, line 176: nit: same, no default assignment.

gpetrou: This will result in a build error.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItemCollection.cs
gpetrou: This will result in a build error.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItemCollection.cs
gpetrou: There is an existing ` if (item is not null)` check in this method. We usually make the parameter nullable in such cases.

dreddy-work: Right @gpetrou. In this case, that check is not needed given `private `use, we know all call sites checking for null. It is safe to change.

gpetrou: @dreddy-work I added a separate commit. Is this what you had in mind?

dreddy-work: yes. thank you.

dreddy-work (approved):

PR #9078, opened 2023-05-06
Use generic list instead of array in TreeNode

gpetrou:
## Proposed changes

- Use generic list instead of array in TreeNode.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9078)


src/System.Windows.Forms/src/System/Windows/Forms/TreeNode.cs
```diff
@@ -1849,15 +1818,15 @@ internal void CheckParentingCycle(TreeNode candidateToAdd)
     /// </summary>
     internal void InsertNodeAt(int index, TreeNode node)
     {
-        EnsureCapacity(1);
+        children.EnsureCapacity(1);
         node.parent = this;
         node.index = index;
         for (int i = childCount; i > index; --i)
         {
             (children[i] = children[i - 1]).index = i;
```
halgab, line 1826: Isn't that line broken now that `children` is a List? I guess now you just have to change the index without the `children[i] = children[i - 1]` part, and this loop can be move after `children.Insert(index, node);` right?

src/System.Windows.Forms/src/System/Windows/Forms/TreeNode.cs
```diff
@@ -2025,12 +1994,7 @@ internal void Remove(bool notify)
         // unlink ourself
         if (notify && parent is not null)
         {
-            for (int i = index; i < parent.childCount - 1; ++i)
-            {
-                (parent.children[i] = parent.children[i + 1]).index = i;
```
halgab, line 2030: Is it okay to no longer update the `index`?

src/System.Windows.Forms/src/System/Windows/Forms/TreeNode.cs
```diff
@@ -1849,15 +1818,15 @@ internal void CheckParentingCycle(TreeNode candidateToAdd)
     /// </summary>
     internal void InsertNodeAt(int index, TreeNode node)
     {
-        EnsureCapacity(1);
+        children.EnsureCapacity(1);
```
halgab, line 1821: Is that still necessary now? Looking at the previous `EnsureCapacity` method, `1` was special-cased but here it will no longer be the case.

src/System.Windows.Forms/src/System/Windows/Forms/TreeNode.cs
```diff
@@ -1377,7 +1375,8 @@ internal void Clear()
                 children[childCount - 1].Remove(true);
             }
 
-            children = null;
+            children.Clear();
+            childCount = 0;
```
halgab, line 1379: Can't `childCount` be replaced with `children.Count` now?

gpetrou: Seems like it can be.

src/System.Windows.Forms/src/System/Windows/Forms/TreeNode.cs
gpetrou: I should not have removed this one.

src/System.Windows.Forms/src/System/Windows/Forms/TreeNodeCollection.cs
```diff
@@ -381,12 +381,11 @@ private int AddInternal(TreeNode node, int delta)
         {
             //if fixedIndex != -1 capacity was ensured by AddRange
             Debug.Assert(delta == 0, "delta should be 0");
-            owner.EnsureCapacity(1);
-            node.index = owner.childCount;
+            owner.children.EnsureCapacity(1);
```
dreddy-work, line 384: please check this case too.

dreddy-work (changes requested):
It would be nice if you could add one or two new tests exercising these changes.

dreddy-work (approved):

PR #9084, opened 2023-05-06
Enable nullability in classes that inherit from DesignerActionList

gpetrou:
## Proposed changes

- Enable nullability in classes that inherit from DesignerActionList.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9084)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.DockingActionList.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.DockingActionList.cs
@@ -21,7 +21,12 @@ public partial class ControlDesigner
 
         private string? GetActionName()
         {
-            PropertyDescriptor? dockProp = TypeDescriptor.GetProperties(Component!)["Dock"];
+            if (Component is null)
+            {
+                return null;
+            }
+
+            PropertyDescriptor? dockProp = TypeDescriptor.GetProperties(Component)["Dock"];
             if (dockProp is not null)
             {
                 DockStyle dockStyle = (DockStyle)dockProp.GetValue(Component)!;
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.DockingActionList.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.DockingActionList.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.ComponentModel;
 using System.ComponentModel.Design;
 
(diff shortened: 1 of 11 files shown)


src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ToolStripActionList.cs
```diff
@@ -38,7 +35,7 @@ private bool CanAddItems
         get
         {
             // Make sure the component is not being inherited -- we can't delete these!
-            InheritanceAttribute ia = (InheritanceAttribute)TypeDescriptor.GetAttributes(_toolStrip)[typeof(InheritanceAttribute)];
+            InheritanceAttribute? ia = (InheritanceAttribute?)TypeDescriptor.GetAttributes(_toolStrip)[typeof(InheritanceAttribute)];
             if (ia is null || ia.InheritanceLevel == InheritanceLevel.NotInherited)
```
halgab, line 39: Maybe we can simplify this down to
```suggestion
            if (!TypeDescriptorHelper.TryGetAttribute(_toolStrip, out InheritanceAttribute? ia) || ia.InheritanceLevel == InheritanceLevel.NotInherited)
```
? Same thing on line 53

gpetrou: Sure.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.DockingActionList.cs
```diff
@@ -13,21 +11,20 @@ public partial class ControlDesigner
...
         {
-            _designer = owner;
             _host = GetService(typeof(IDesignerHost)) as IDesignerHost;
         }
 
-        private string GetActionName()
+        private string? GetActionName()
         {
-            PropertyDescriptor dockProp = TypeDescriptor.GetProperties(Component)["Dock"];
+            PropertyDescriptor? dockProp = TypeDescriptor.GetProperties(Component!)["Dock"];
```
dreddy-work, line 24: i would check `Component  `for null here.  `TypeDescriptor.GetProperties() ` take null, an empty collection is returned.

```CS
if (Component is null)
{
    return null;
}
```

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ListViewActionList.cs
```diff
@@ -36,35 +35,35 @@ public View View
     {
         get
         {
-            return ((ListView)Component).View;
+            return ((ListView)Component!).View;
         }
         set
         {
-            TypeDescriptor.GetProperties(Component)["View"].SetValue(Component, value);
+            TypeDescriptor.GetProperties(Component!)["View"]!.SetValue(Component, value);
```
dreddy-work, line 42: i don't know if there is any better way to avoid `!` all over this file. @JeremyKuhne

dreddy-work: Or Should we throw null exception on `Component ` more aggressively?

JeremyKuhne: You can stash the designer.Component in a field in the constructor and throw there if for some reason it isn't `ListView`. It seems wrong that we're hitting the TypeDescriptor to set the property. What possible purpose does it solve?

JeremyKuhne: Same goes for below.

PR #9086, opened 2023-05-07
Enable nullability in some ToolTip members

gpetrou:
## Proposed changes

- Enable nullability in some ToolTip members.


src/System.Windows.Forms/src/System/Windows/Forms/ToolTip.cs
```diff
@@ -1564,7 +1564,7 @@ internal void ShowKeyboardToolTip(string text, IKeyboardToolTip tool, int durati
             StartTimer(tool.GetOwnerWindow(), duration);
         }
     }
-
+#nullable enable
```
dreddy-work, line 1567: why is this needed?

gpetrou: Because the members in the middle of the file are still not annotated. I will open separate PRs for them.

dreddy-work (approved):

PR #9088, opened 2023-05-08
Create BinaryFormatter replacement for known types

JeremyKuhne:
In a number of cases we know what data we're serializing or deserializing and can avoid using the BinaryFormatter if we weren't concerned about compatibility.

I was considering using a different format on the wire and writing a parser to deal with legacy data. It finally occured to me that if I'm going to parse legacy data I might as well put the data out in the legacy format to maintain compatibility.

This creates a model for the NRBF format that can be used to look at data the BinaryFormatter puts out. The top level of the model is represented by the BinaryFormattedObject class.

The binary records that are part of the model all self serialize and deserialize, which allows manually constructing known models to export data that can be read by the BinaryFormatter. The records never hydrate any complex types other than strings, that must be done manually.

BinaryFormatReader and BinaryFormatWriter have helper methods for creating and reading a set of known types. I've handled primitive types and Hashtable and started on List<T> as well.

Clipboard will no longer use the BinaryFormatter for primitive types. Tests have been updated appropriately.

My next step is to change the ActiveX property bag serialization to use the Hashtable serialization that this change introduces.  I also have a few more tests to port from my prototype repo.

The idea is to then further reduce the need to fall back on the BinaryFormatter by providing support for other BinaryFormatted types in WinForms and other types that WinForms projects would commonly use as public property types. This includes things like PointF and RectangleF, primitive arrays, and other collections of primitive types (List<T> and ArrayList are the priorities there).

Note that performance isn't far off from using the BinaryFormatter. For serializing a really large Hashtable (around 10K records) we are about the same speed and use a slightly larger amount of memory. It's possible it could be significantly better if we use the enumerator to get the data instead of ISerializable- would have to spend some time to figure out the right values for the other private fields without using reflection.

BinaryFormattedObject does not yet support multidimensional or jagged arrays. This doesn't hit any of our scenarios, will implement when needed.

This change also consumes FluentAssertions in the Primitives test assembly. We've discussed using this as a team and will be looking at it in use here before committing to the rest of the test projects.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9088)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/System/IO/BinaryWriterExtensions.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/IO/BinaryWriterExtensions.cs
@@ -1,0 +1,39 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.CompilerServices;
+using System.Text;
+
+namespace System.IO;
+
+internal static class BinaryWriterExtensions
+{
+    /// <summary>
+    ///  Writes a UTF-8 string prefixed by a 7-bit encoded length. Does not include a trailing null.
+    /// </summary>
+    public static void WriteLengthPrefixedString(this BinaryWriter writer, string value)
+    {
+        Encoding encoding = Encoding.UTF8;
+        using BufferScope<byte> buffer = new(stackalloc byte[256], encoding.GetMaxByteCount(value.Length));
+        int length = encoding.GetBytes(value, buffer);
+        writer.Write7BitEncodedInt(length);
+        writer.Write(buffer[..length]);
+    }
+
+    /// <summary>
+    ///  Writes a <see cref="DateTime"/> object to the given <paramref name="writer"/>.
+    /// </summary>
+    public static void Write(this BinaryWriter writer, DateTime value)
+    {
+        // Copied from System.Runtime.Serialization.Formatters.Binary.BinaryFormatterWriter
+
+        // In .NET Framework, BinaryFormatter is able to access DateTime's ToBinaryRaw,
+        // which just returns the value of its sole Int64 dateData field.  Here, we don't
+        // have access to that member (which doesn't even exist anymore, since it was only for
+        // BinaryFormatter, which is now in a separate assembly).  To address that,
+        // we access the sole field directly via an unsafe cast.
+        long dateData = Unsafe.As<DateTime, long>(ref value);
+        writer.Write(dateData);
+    }
+}
--- a/src/System.Windows.Forms.Primitives/src/System/IO/BinaryWriterExtensions.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/IO/BinaryWriterExtensions.cs
@@ -3,24 +3,11 @@
 // See the LICENSE file in the project root for more information.
 
 using System.Runtime.CompilerServices;
-using System.Text;
 
 namespace System.IO;
 
 internal static class BinaryWriterExtensions
 {
-    /// <summary>
-    ///  Writes a UTF-8 string prefixed by a 7-bit encoded length. Does not include a trailing null.
-    /// </summary>
-    public static void WriteLengthPrefixedString(this BinaryWriter writer, string value)
-    {
-        Encoding encoding = Encoding.UTF8;
-        using BufferScope<byte> buffer = new(stackalloc byte[256], encoding.GetMaxByteCount(value.Length));
-        int length = encoding.GetBytes(value, buffer);
-        writer.Write7BitEncodedInt(length);
-        writer.Write(buffer[..length]);
-    }
-
     /// <summary>
     ///  Writes a <see cref="DateTime"/> object to the given <paramref name="writer"/>.
     /// </summary>
--- a/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/ArrayRecord.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/ArrayRecord.cs
@@ -1,0 +1,60 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections;
+
+namespace System.Windows.Forms.BinaryFormat;
+
+/// <summary>
+///  Base class for array records.
+/// </summary>
+/// <devdoc>
+///  <see href="https://learn.microsoft.com/openspecs/windows_protocols/ms-nrbf/f57d41e5-d3c0-4340-add8-fa4449a68d1c">
+///  [MS-NRBF] 2.4</see> describes how item records must follow the array record and how multiple null records
+///  can be coalesced into an <see cref="NullRecord.ObjectNullMultiple"/> or <see cref="NullRecord.ObjectNullMultiple256"/>
+///  record.
+/// </devdoc>
+internal abstract class ArrayRecord : Record, IEnumerable<object>
+{
+    public ArrayInfo ArrayInfo { get; }
+
+    /// <summary>
+    ///  The array items.
+    /// </summary>
+    /// <remarks>
+    ///  <para>
+    ///   Multi-null records are always expanded to individual <see cref="ObjectNull"/> entries when reading.
+    ///  </para>
+    /// </remarks>
+    public IReadOnlyList<object> ArrayObjects { get; }
+
+    /// <summary>
+    ///  Identifier for the array.
+    /// </summary>
+    public Id ObjectId => ArrayInfo.ObjectId;
+
+    /// <summary>
+    ///  Length of the array.
+    /// </summary>
+    public Count Length => ArrayInfo.Length;
+
+    /// <summary>
+    ///  Returns the item at the given index.
+    /// </summary>
+    public object this[int index] => ArrayObjects[index];
+
+    public ArrayRecord(ArrayInfo arrayInfo, IReadOnlyList<object> arrayObjects)
+    {
+        if (arrayInfo.Length != arrayObjects.Count)
+        {
+            throw new ArgumentException($"{nameof(arrayInfo)} doesn't match count of {nameof(arrayObjects)}");
+        }
+
+        ArrayInfo = arrayInfo;
+        ArrayObjects = arrayObjects;
+    }
+
+    IEnumerator<object> IEnumerable<object>.GetEnumerator() => ArrayObjects.GetEnumerator();
+    IEnumerator IEnumerable.GetEnumerator() => ArrayObjects.GetEnumerator();
+}
--- a/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/ArraySinglePrimitive.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/ArraySinglePrimitive.cs
@@ -1,0 +1,52 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Windows.Forms.BinaryFormat;
+
+/// <summary>
+///  Single dimensional array of a primitive type.
+/// </summary>
+/// <remarks>
+///  <para>
+///   <see href="https://learn.microsoft.com/openspecs/windows_protocols/ms-nrbf/3a50a305-5f32-48a1-a42a-c34054db310b">
+///    [MS-NRBF] 2.4.3.3
+///   </see>
+///  </para>
+/// </remarks>
+internal sealed class ArraySinglePrimitive : ArrayRecord, IRecord<ArraySinglePrimitive>
+{
+    public PrimitiveType PrimitiveType { get; }
+
+    public static RecordType RecordType => RecordType.ArraySinglePrimitive;
+
+    public ArraySinglePrimitive(ArrayInfo arrayInfo, PrimitiveType primitiveType, IReadOnlyList<object> arrayObjects)
+        : base(arrayInfo, arrayObjects)
+    {
+        PrimitiveType = primitiveType;
+    }
+
+    static ArraySinglePrimitive IBinaryFormatParseable<ArraySinglePrimitive>.Parse(
+        BinaryReader reader,
+        RecordMap recordMap)
+    {
+        ArrayInfo arrayInfo = ArrayInfo.Parse(reader, out Count length);
+        PrimitiveType primitiveType = (PrimitiveType)reader.ReadByte();
+
+        ArraySinglePrimitive record = new(
+            arrayInfo,
+            primitiveType,
+            ReadPrimitiveTypes(reader, primitiveType, length));
+
+        recordMap[record.ObjectId] = record;
+        return record;
+    }
+
+    public override void Write(BinaryWriter writer)
+    {
+        writer.Write((byte)RecordType);
+        ArrayInfo.Write(writer);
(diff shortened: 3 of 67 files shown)


JeremyKuhne:
There are a little over 500 lines of executable code to implement this. I expect it will total 600 or so when the other items are implemented. I've tried to keep the maintainability index high (currently 84).

I've also leaned heavily on patterns to help avoid parsing mistakes. Usage of BinaryReader/Writer takes a fair amount of validation off of our hands. I've encapsulated counts and identifiers into structs that enforce positive numbers. I've also guarded when reading data from stream provided counts.

JeremyKuhne:
cc: @lonitra, @GrabYourPitchforks

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/ArraySinglePrimitive.cs
```diff
@@ -0,0 +1,52 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Windows.Forms.BinaryFormat;
```
Tanya-Solyanik, line 5: The docs refer to this format as msnrbf. It would be easier to look things up if the names match.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/ArrayRecord.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// <summary>
+    ///  Returns the item at the given index.
+    /// </summary>
+    public object this[int index] => ArrayObjects[index];
+
+    public ArrayRecord(ArrayInfo arrayInfo, IReadOnlyList<object> arrayObjects)
+    {
+        if (arrayInfo.Length != arrayObjects.Count)
+        {
+            throw new ArgumentException($"{nameof(arrayInfo)} doesn't match count of {nameof(arrayObjects)}");
```
Tanya-Solyanik, line 51: Please localize the error message. The same applies to multiple files.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/BinaryFormatReader.cs
```diff
@@ -0,0 +1,145 @@
...
+            || format[2] is not ArraySinglePrimitive array
+            || array.PrimitiveType != Record.GetPrimitiveType(typeof(T)))
+        {
+            throw new SerializationException();
+        }
+
+        int count;
+        try
+        {
+            count = (int)classInfo["_size"];
```
Tanya-Solyanik, line 57: Had the underscore in the field name been there for all versions of .NET?

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/BinaryFormatWriter.cs
```diff
@@ -0,0 +1,425 @@
...
+using System.Text;
+
+namespace System.Windows.Forms.BinaryFormat;
+
+/// <summary>
+///  Writer that writes specific types in binary format without using the BinaryFormatter.
+/// </summary>
+internal static class BinaryFormatWriter
+{
+    private static readonly IReadOnlyList<string> s_hashtableMemberNames = new List<string>()
```
Tanya-Solyanik, line 17: Are these copies over from the runtime repo?

Tanya-Solyanik (commented):
I didn't finish, will get back to it later.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/BinaryFormatReader.cs
JeremyKuhne: Yes, or there would have been a binary breaking change.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/BinaryFormatWriter.cs
JeremyKuhne: They are copies of what is serialized, sometimes that is the field names themselves, other times they are populated through ISerializable. It's all vetted with round trip tests and reviews of the Framework sources.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/ArrayRecord.cs
JeremyKuhne: I don't think we should for this code. .NET does not localize error messages as it makes it difficult to search. If BinaryFormatter doesn't localize, why should we?

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/ArraySinglePrimitive.cs
JeremyKuhne: I don't agree here. Nobody knows this as the "ms-nrbf" spec, that is trivia. It is much easier to associate this with the BinaryFormatter with this name.  (.NET remoting binary format)

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/ArrayRecord.cs
Tanya-Solyanik: Because we are a UI framework where messages might bubble up to the end user if not handled properly.

JeremyKuhne: The same can be said of all of the exceptions coming from the .NET runtime libraries. In this case in particular this will be integrated with the existing BinaryFormatter as a fallback so it would be extra jarring to sometimes get exception messages localized and sometimes not. On top of that some portion of this may be moved to library implementations in the future.

In this particular case the consistency and the value of searching (as described) are more important. If we find a place where errors surface upstream of this that we want a localized message (where we don't already have one), we can always add them as we deem appropriate.

kirsan31: IMHO (but many developers agree on this), localising exceptions is truly evil in any case 🙄

RussKie: > IMHO (but many developers agree on this), localising exceptions is truly evil in any case 🙄

💯 agree.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/BinaryFormatReader.cs
```diff
@@ -0,0 +1,145 @@
...
+        {
+            throw new NotSupportedException($"{nameof(T)} is not primitive.");
+        }
+
+        BinaryFormattedObject format = new(stream, leaveOpen: true);
+        if (format.RecordCount < 4
+            || format[1] is not SystemClassWithMembersAndTypes classInfo
+            || !classInfo.Name.StartsWith($"System.Collections.Generic.List`1[[{typeof(T).FullName}")
+            || format[2] is not ArraySinglePrimitive array
+            || array.PrimitiveType != Record.GetPrimitiveType(typeof(T)))
```
dreddy-work, line 49: why not checking last field as well `format[3]`. Here and other places below.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/BinaryFormatReader.cs
```diff
@@ -0,0 +1,145 @@
...
+                BinaryObjectString keyString => keyString.Value,
+                MemberReference reference => format[reference.IdRef] switch
+                {
+                    BinaryObjectString refString => refString.Value,
+                    _ => throw new SerializationException()
+                },
+                _ => throw new SerializationException()
+            };
+
+            hashtable[key] = values[i] switch
```
dreddy-work, line 103: we may want to ensure we are not overriding when duplicate?

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/BinaryFormattedObject.cs
```diff
@@ -0,0 +1,185 @@
...
+    /// <returns><see langword="true"/> if this represented a primitive type or string.</returns>
+    public bool TryGetPrimitiveType([NotNullWhen(true)] out object? value)
+    {
+        try
+        {
+            return TryGetPrimitiveTypeInternal(out value);
+        }
+        catch (Exception ex) when (!ClientUtils.IsCriticalException(ex))
+        {
+            value = default;
```
dreddy-work, line 110: I was expecting a compile warning here. should this assignment to be moved to `finally`?

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/BinaryFormatWriter.cs
```diff
@@ -0,0 +1,425 @@
...
+    {
+        long originalPosition = stream.Position;
+        try
+        {
+            WritePrimitive(stream, primitive);
+            return true;
+        }
+        catch (Exception ex) when (!ClientUtils.IsCriticalException(ex))
+        {
+            stream.Position = originalPosition;
```
dreddy-work, line 194: Why not reset when it is critical exception?  what if caller like to handle exception ?

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/MemberTypeInfo.cs
```diff
@@ -0,0 +1,108 @@
...
+        {
+            info.Add(((BinaryType)reader.ReadByte(), null));
+        }
+
+        // Check for more clarifying information
+
+        for (int i = 0; i < expectedCount; i++)
+        {
+            BinaryType type = info[i].Type;
+            switch (type)
```
dreddy-work, line 45: why not do this in the first iteration itself?

src/System.Windows.Forms.Primitives/src/System/IO/BinaryWriterExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.Runtime.CompilerServices;
+
+namespace System.IO;
+
+internal static class BinaryWriterExtensions
+{
+    /// <summary>
+    ///  Writes a <see cref="DateTime"/> object to the given <paramref name="writer"/>.
+    /// </summary>
+    public static void Write(this BinaryWriter writer, DateTime value)
```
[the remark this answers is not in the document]
JeremyKuhne: This matches the naming of `BinaryWriter`, it is always `Write` for that class.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/BinaryFormatReader.cs
JeremyKuhne: No need. We don't care if the rest of the non-critical data is well-formed if it has the right type description. As long as the overall stream is generally well-formed (can parse to an end record) that's all that matters. If someone wants to be malicious, they can do so with well-formed data so it doesn't matter that we check for 1-1 `BinaryFormatter` bytes. There is also less risk that I miss some compat case.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/BinaryFormatReader.cs
JeremyKuhne: From a security standpoint it doesn't matter.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/MemberTypeInfo.cs
JeremyKuhne: Because it all follows the first iteration. All types come first, then the clarifying info. It's just the way the spec is.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/BinaryFormattedObject.cs
JeremyKuhne: No. It is only the error case that needs it. The other call is setting the out value.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/BinaryFormatWriter.cs
JeremyKuhne: All bets are off if you have a stack overflow or out-of-memory. These aren't meant to be handled.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/BinaryFormatReader.cs
JeremyKuhne: To be clear, I am checking that the data I need to read is in a good state, but all else is 🤷🏼

src/System.Windows.Forms.Primitives/tests/UnitTests/System/Windows/Forms/BinaryFormat/PrimitiveTypes.cs
```diff
@@ -0,0 +1,95 @@
...
+
+namespace System.Windows.Forms.BinaryFormat.Tests;
+
+public class PrimitiveTypes
+{
+    [Theory]
+    [MemberData(nameof(RoundTrip_Data))]
+    public void RoundTripPrimitiveTypes(byte type, object value)
+    {
+        MemoryStream stream = new();
```
dreddy-work, line 15: using?

dreddy-work (commented):
RLGTM.

src/System.Windows.Forms.Primitives/tests/UnitTests/System/Windows/Forms/BinaryFormat/PrimitiveTypes.cs
JeremyKuhne: The BinaryReader closes it in it's using.

dreddy-work (approved):

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/BinaryFormatWriter.cs
```diff
@@ -0,0 +1,425 @@
...
+    {
+        "flags", "hi", "lo", "mid"
+    };
+
+    private static readonly IReadOnlyList<string> s_dateTimeMemberNames = new List<string>()
+    {
+        "ticks", "dateData"
+    };
+
+    private const string Mscorlib = "mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089";
```
Tanya-Solyanik, line 37: Had you tested copy/pasting between .NET Framework applications and Core applications? At some point it used to work. Perhaps a unit test to test it can be added?

JeremyKuhne: Not explicitly, but the BinaryFormatter uses the mscorlib versions (when it puts out a version).  Everything I've done here is matching what it puts out.

PR #9094, opened 2023-05-09
Enable nullability in ToolStripPanelRowControlCollection

gpetrou:
## Proposed changes

- Enable nullability in ToolStripPanelRowControlCollection.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9094)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripPanelRow.ToolStripPanelRowControlCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripPanelRow.ToolStripPanelRowControlCollection.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.Collections;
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ToolStripPanelRow.ToolStripPanelRowControlCollection.cs
```diff
@@ -149,13 +147,17 @@ public virtual void Clear()
 
         private Control GetControl(int index)
         {
-            Control control = null;
-            ToolStripPanelCell cell = null;
+            Control? control = null;
```
dreddy-work, line 150: This is better rewrite along the line of following and remove locals: 

```CS
 private Control GetControl(int index)
        {
            if (index < Count && index >= 0)
            {
                var cell = (ToolStripPanelCell)(InnerList[index]);
                return cell.Control;
            }

             throw new IndexOutOfRangeException();
        }
```

PR #9104, opened 2023-05-11
Move away from BinaryFormatter for ActiveX purposes

JeremyKuhne:
The PropertyBags now serialize without it where possible. This allows a standard Control to completely serialize without the BinaryFormatter enabled.

For the AxHost case, I've left the BinaryFormatter as a fall-back.

I've also extended the Hashtable serialization to handle all primitive types. Note the two line change to make this happen. :)

This change also moves the "Reader" methods to "BinaryFormattedObjectExtensions". Due to our layering and the need to check multiple states I realized that we need to have all of the reading of objects directly using an existing BinaryFormattedObject instance.

Fixes a few bugs and adds more tests. Some random cleanup related to VARIANT (as I was exploring the scope of things).

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9104)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/Interop/OleAut32/Interop.VARIANT.cs
+++ b/src/System.Windows.Forms.Primitives/src/Interop/OleAut32/Interop.VARIANT.cs
@@ -938,7 +938,7 @@ internal unsafe partial struct VARIANT : IDisposable
     private static T ThrowInvalidCast<T>() => throw new InvalidCastException();
 
     /// <summary>
-    ///  Converts the given object to <see cref="VARIANT"/>. WARNING: Only handles <see cref="string"/>.
+    ///  Converts the given object to <see cref="VARIANT"/>.
     /// </summary>
     public static VARIANT FromObject(object? value)
     {
--- a/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/BinaryFormatReader.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/BinaryFormatReader.cs
@@ -2,7 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System.Collections;
 using System.Runtime.CompilerServices;
 using System.Runtime.Serialization;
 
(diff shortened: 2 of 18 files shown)


src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/BinaryFormatWriter.cs
```diff
@@ -380,46 +385,12 @@ public static void WriteHashtableOfStrings(Stream stream, Hashtable hashtable)
...
+                _ => new MemberPrimitiveTyped(value)
             });
 
         array = new(new(3, values.Length), valueConverter);
         array.Write(writer);
 
         MessageEnd.Instance.Write(writer);
     }
-
-    public static PrimitiveType GetPrimitiveType(object @object)
```
JeremyKuhne, line 392: This was duplicated code.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/BinaryFormatReader.cs
```diff
@@ -13,109 +12,6 @@ namespace System.Windows.Forms.BinaryFormat;
 /// </summary>
 internal static class BinaryFormatReader
 {
-    /// <summary>
```
JeremyKuhne, line 16: Moved to the extensions class

src/System.Windows.Forms.Primitives/src/Interop/OleAut32/Interop.VARIANT.cs
```diff
@@ -938,7 +938,7 @@ private static Span<T> GetSpan<T>(Array array)
     private static T ThrowInvalidCast<T>() => throw new InvalidCastException();
 
     /// <summary>
-    ///  Converts the given object to <see cref="VARIANT"/>. WARNING: Only handles <see cref="string"/>.
+    ///  Converts the given object to <see cref="VARIANT"/>.
     /// </summary>
     public static VARIANT FromObject(object? value)
```
KlausLoeffelmann, line 943: Personal question for my better personal understanding:
Is this for the VARIANT type that has been used in the context of VB6? 
If yes, there once was a `DBNull` subtype also, if I (vaguely) remember correctly (and I can be completely wrong). I am just mentioning it, and maybe it needs to be taken into consideration?

KlausLoeffelmann (approved):
Looks good to me.
Had a question around Variant.

src/System.Windows.Forms.Primitives/src/Interop/OleAut32/Interop.VARIANT.cs
JeremyKuhne: Yes, there is a special null type. It's "handled" for the moment in that we fall back to the .NET interop conversion here. What we have to do is get the rest of this implemented on our side for AOT purposes.

Issue #9111, opened 2023-05-11
[Tracking] DefaultFont is static and could cause issues in PermonitorV2 applications.


dreddy-work:
### .NET version

.NET 8.0

### Did it work in .NET Framework?

No

### Did it work in any of the earlier releases of .NET Core or .NET 5+?

NO

### Issue description

`DefaultFont `in WinForms is static and currently does not consider the DPI of the monitor into consideration. This need to be investigated for all scenarios in PermonitorV2 mode applications.

Currently, `DeviceContext `and Font are defaulting to primary monitor and seems working as expected but would need deeper investigation/understanding

### Steps to reproduce

Any PerMonitor V2 mode WinForms application with dual monitor with different DPI.

dreddy-work:
@Tanya-Solyanik, this is tracking issue to deal with device context and FOnt being constant in PermonitorV2 screens. Its always created for Primary monitor and reused.

PR #9112, opened 2023-05-11
Update/Scale Font on each DPICHNAGED message

### LINKED ISSUE #8932: When a Form is opened and StartPosition is CenterParent, it is opened always on the main monitor when the value of WindowState is Maximized regardless of parent's monitor
### .NET version

SDK DE .NET:
 Version:   7.0.202
 Commit:    6c74320bc3

Entorno de tiempo de ejecución:
 OS Name:     Windows
 OS Version:  10.0.19045
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\7.0.202\

Host:
  Version:      7.0.4
  Architecture: x64
  Commit:       0a396acafe

Yes

_No response_

### Issue description

When a Form is opened and StartPosition is CenterParent, it is opened always on the main monitor when the value of WindowState is Maximized.

If a window is opened on the secondary monitor, instead of being displayed on that monitor it is displayed on the primary.

Because of this, we cannot open maximized forms by default on the same monitor as the parent window.

Here is a test form that demonstrates the observed behavior:

[Form1.vb](https://pastebin.com/v5beGHEh)
[Form1.Designer.vb](https://pastebin.com/XSCBYR4V)

As a workaround, it is possible to set the WindowState in the overridden OnLoad method. (The "Open form that overrides onLoad normal with center parent without parent" and "Open form that overrides onLoad normal with center parent with parent" buttons in Form1). Although if the form contains many controls and takes time to load, you see a rectangle on the wrong monitor before moving to the correct one.

### Steps to reproduce

**Environment**
2 monitor setup (1920x1080 each)
Left monitor assigned as primary monitor

**Reproduction Steps:**

1. Run the test application
2. Move application window to secondary monitor
3. Click the "Open form normal with center parent" button
4. Close the modal window
5. Click the "Open form maximized with center parent" button

**Expected result**
Both modal windows are displayed on the same monitor as their parent (the secondary monitor)

**Actual result**
In the case of the form with the value FormWindowState.Normal, it is shown in the parent's monitor.
However, in the form with value FormWindowState.Maximized it is always displayed on the main monitor.

**Other observations**
It has also been verified that changing the main monitor in the operating system and executing
the steps according to the change (move the application to the secondary monitor before
clicking on the buttons) the behavior is maintained.

### LINKED ISSUE #8946: PerMonitorV2 - Opening a dialog whose owner is on a secondary screen results in broken scaling in .NET 7+
### .NET version

.NET 7.0
.NET 8.0 Preview 2

Yes

Yes, it worked in .NET 6 and earlier.

### Issue description

If a form is opened as a dialog, and the dialog's owner is located on a monitor whose DPI scaling factor doesn't match the primary monitor, the dialog will either be scaled incorrectly on initial scaling, or start scaling incorrectly after it has been moved to a different screen. The issue only occurs in .NET 7 and later. Changing [`ShowInTaskbar`](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.form.showintaskbar)/[`StartPosition`](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.form.startposition) of the dialog seems to produce different (but still incorrect) behavior.

I attached a project that can be used to reproduce the issue, plus 4 videos that demonstrate various side effects. These different effects occur after following *the same replication steps* in different monitor/form configurations. The videos demonstrate correct behavior in .NET 6, and its deterioration as the project is upgraded to .NET 7 and .NET 8 Preview 2.

[WinFormsApp19.zip](https://github.com/dotnet/winforms/files/11172134/WinFormsApp19.zip)
<details>
<summary>Main100_DefaultProperties</summary>

https://user-images.githubusercontent.com/47351802/230427392-03372cc2-2cef-4585-a5ae-e3eef2cdb452.mp4
</details>
<details>
<summary>Main200_DefaultProperties</summary>

https://user-images.githubusercontent.com/47351802/230427985-42082230-3b55-4178-8db7-7b3198d711b2.mp4
</details>
<details>
<summary>Main100_TaskbarAndPositionModified</summary>

https://user-images.githubusercontent.com/47351802/230428166-95fbddfe-355c-47f5-9806-3e74ea5c31e7.mp4
</details>

<details>
<summary>Main200_TaskbarAndPositionModified</summary>

https://user-images.githubusercontent.com/47351802/230428274-631bcbaf-d306-43b3-a424-ed4053121f38.mp4
</details>

### Steps to reproduce

 - Prerequisite: ensure the system has two monitors with different scaling settings. Reboot the system.
1. In Visual Studio 2022, create a new project with the Windows Forms App template. Select .NET 7.
2. In the form designer, add a button to the form. Configure this button to launch another instance of the main form using the `ShowDialog` method: `button1.Click += (sender, e) => { using (var form = new Form1()) form.ShowDialog(); };`.
3. Add `<ApplicationHighDpiMode>PerMonitorV2</ApplicationHighDpiMode>` to the project file.
4. Run the project.
5. Move the main form to the secondary monitor and click on the button.
6. Move the newly opened dialog to the secondary monitor.
7. **The button's bounds are not scaled.**

### LINKED ISSUE #9004: HDPI PMv2 issue in ContainerControl.ScaleContainerForDpi
### .NET version

7.0

Yes

worked in .net 5

### Issue description

The order of functions in ScaleContainerForDpi() is wrong.
Now, it first calls `PInvoke.SetWindowPos` and later it sets Font by calling `GetScaledFont `and raising `OnFontChanged`.

But, in my Per-monitor setup (175% dpi scaling for primary screen and 100% secondary), the `SetWindowPos()` call generates another `OnDpiChanged`, resulting in another `ScaleContainerForDpi `being called (see stacktrace attached).
This inner `ScaleContainerForDpi` call sets the correct font, but the Font is then overwritten by the incorrect value, after the function returns.

So I suppose the correct order should be: First raise OnFontChanged, then call PInvoke.SetWindowPos

[stacktrace.txt](https://github.com/dotnet/winforms/files/11270667/stacktrace.txt)

### Steps to reproduce

I did not found a reliable way how to replicate the issue on smallest scenario possible, it just happens for some of the forms in my application, while some other forms behave correctly

dreddy-work:
fixes #8946 
fixes #9004

By default, control handles are created with the DPI (dots per inch) of the primary monitor. If we assign an owner or parent to a control that is not on the primary monitor and has a different DPI than the primary monitor, it may trigger a DPICHANGED event. This situation leads to issues linked above.

When the DPICHANGED event occurs, we adjust the font and scale the control according to the new DPI. However, if the child control is a top-level window and explicitly or by default placed on the primary monitor, it generates another DPICHANGED message within the first one, causing the scaling process to repeat. This behavior is expected. The problem arises from the current order in which we attempt to scale the font and the control itself. We encounter the second DPICHANGED message before scaling the font, resulting in incorrect font information.

This pull request addresses the issue by fixing the order of operations and also disposing of fonts that are no longer needed.

Created a new issue to further investigate DefaultFont and DPI combination in WInForms. https://github.com/dotnet/winforms/issues/9111

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9112)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ContainerControl.cs
@@ -1452,21 +1452,7 @@ public class ContainerControl : ScrollableControl, IContainerControl
             // WM_DPICHANGED event. Failing to apply SuggestedRectangle will result in a circular WM_DPICHANGED
             // events on the control.
 
-            // Note: SuggestedRectangle supplied  by WM_DPICHANGED event is Dpi (not Font) scaled. if top-level window is
-            // Font scaled, we might see deviations in the expected bounds and may result in adding Scrollbars (horizontal/vertical)
-            if (IsHandleCreated)
-            {
-                PInvoke.SetWindowPos(
-                    this,
-                    HWND.HWND_TOP,
-                    suggestedRectangle.X,
-                    suggestedRectangle.Y,
-                    suggestedRectangle.Width,
-                    suggestedRectangle.Height,
-                    SET_WINDOW_POS_FLAGS.SWP_NOZORDER | SET_WINDOW_POS_FLAGS.SWP_NOACTIVATE);
-            }
-
-            // Bounds are already scaled for the top-level window. We would need to skip scaling of
+            // Bounds are being scaled for the top-level window via SuggestedRectangle. We would need to skip scaling of
             // this control further by the 'OnFontChanged' event.
             _isScaledByDpiChangedEvent = true;
 
@@ -1482,6 +1468,18 @@ public class ContainerControl : ScrollableControl, IContainerControl
                 {
                     OnFontChanged(EventArgs.Empty);
                 }
+            }
+
+            if (IsHandleCreated)
+            {
+                PInvoke.SetWindowPos(
+                    this,
+                    HWND.HWND_TOP,
+                    suggestedRectangle.X,
+                    suggestedRectangle.Y,
+                    suggestedRectangle.Width,
+                    suggestedRectangle.Height,
+                    SET_WINDOW_POS_FLAGS.SWP_NOZORDER | SET_WINDOW_POS_FLAGS.SWP_NOACTIVATE);
             }
         }
         finally
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
@@ -2122,9 +2122,13 @@ public unsafe partial class Control :
     internal Font GetScaledFont(Font font, int newDpi, int oldDpi)
     {
         Debug.Assert(PInvoke.AreDpiAwarenessContextsEqualInternal(DpiAwarenessContext, DPI_AWARENESS_CONTEXT.DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2),
-            $"Fonts need to be cached only for PermonitorV2 mode applications : {DpiHelper.IsPerMonitorV2Awareness} : {DpiAwarenessContext}");
+            $"Fonts need to be cached only for PerMonitorV2 mode applications : {DpiHelper.IsPerMonitorV2Awareness} : {DpiAwarenessContext}");
 
-        _dpiFonts ??= new Dictionary<int, Font>();
+        _dpiFonts ??= new Dictionary<int, Font>
+        {
+            { oldDpi, font.WithSize(font.Size) }
+        };
+
         if (_dpiFonts.TryGetValue(newDpi, out Font? scaledFont))
         {
             return scaledFont!;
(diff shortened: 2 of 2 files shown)


JeremyKuhne (approved):

dreddy-work:
@Olina-Zhang, can your team do a sanity check on this PR?

John-Qiao:
@dreddy-work  we tried the prviate dlls(build by current PR) with issue: [8932](https://github.com/dotnet/winforms/issues/8932), but that issue still repro with private dlls, if that issue cannot repro on your side, please provide your private dlls for us.

dreddy-work:
> @dreddy-work we tried the prviate dlls(build by current PR) with issue: [8932](https://github.com/dotnet/winforms/issues/8932), but that issue still repro with private dlls, if that issue cannot repro on your side, please provide your private dlls for us.

Sorry, it is #8946 but not 8932.

Olina-Zhang:
@dreddy-work Built this PR to verify that issue https://github.com/dotnet/winforms/issues/8946, it was fixed: targeting .Net 8.0 has the same testing result as .Net 6.0.

PR #9113, opened 2023-05-12
Add BinaryArrayType

JeremyKuhne:
Sometimes single dimensional arrays get serialized as BinaryArrayType instead of ArraySingleObject. This change implements reading of BinaryArrayType for this case (writing is still not supported).

This also moves some helper methods up to ClassRecord.

It is also possible for Ids to be negative. They are effectively treated as positive numbers by the BinaryFormatter so Id was changed to do the same (but retain the original value).

Adds a batch of additional tests and moves the BinaryFormat test extensions to the test utilities assembly.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9113)


KlausLoeffelmann:
Ping on the failing tests! 😸

JeremyKuhne:
@KlausLoeffelmann I forgot to include the updated tests. :)

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/Record.cs
```diff
@@ -133,6 +133,7 @@ private protected static unsafe void WritePrimitiveType(BinaryWriter writer, Pri
             // String is handled with a record, never on it's own
             case PrimitiveType.Null:
             case PrimitiveType.String:
+            default:
```
Tanya-Solyanik, line 136: What primitive types are missing from this list? Should the exception say "unexpected type"?

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/Record.cs
```diff
@@ -167,8 +168,8 @@ internal static IRecord ReadBinaryFormatRecord(BinaryReader reader, RecordMap re
             RecordType.SystemClassWithMembersAndTypes => ReadSpecificRecord<SystemClassWithMembersAndTypes>(recordMap),
             RecordType.ClassWithMembersAndTypes => ReadSpecificRecord<ClassWithMembersAndTypes>(recordMap),
             RecordType.BinaryObjectString => ReadSpecificRecord<BinaryObjectString>(recordMap),
-            // The BinaryArray record is used for multidimensional arrays
-            RecordType.BinaryArray => throw new NotImplementedException(),
+            // The BinaryArray record is used all types of arrays, but we currently only support single dimension.
```
Tanya-Solyanik, line 171: A preposition is missing after `used`.

Tanya-Solyanik (dismissed):

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/Record.cs
```diff
@@ -261,5 +262,75 @@ private protected static void WriteRecords(BinaryWriter writer, IReadOnlyList<ob
...
 
+    /// <summary>
+    ///  Reads object member values using <paramref name="memberTypeInfo"/>.
+    /// </summary>
+    private protected static IReadOnlyList<object> ReadValuesFromMemberTypeInfo(
+        BinaryReader reader,
+        RecordMap recordMap,
+        MemberTypeInfo memberTypeInfo)
+    {
+        List<object> memberValues = new();
```
dreddy-work, line 273: why not assign size directly, given we know it upfront? why use `EnsureCapacity `?

dreddy-work (dismissed):

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/Record.cs
JeremyKuhne: There are no others. As a write operation this is probably better as something other than SerializationException. I've changed it.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/Record.cs
JeremyKuhne: Oversight

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/Record.cs
```diff
@@ -182,7 +182,7 @@ internal static IRecord ReadBinaryFormatRecord(BinaryReader reader, RecordMap re
             RecordType.ArraySingleString => ReadSpecificRecord<ArraySingleString>(recordMap),
             RecordType.MethodCall => throw new NotSupportedException(),
             RecordType.MethodReturn => throw new NotSupportedException(),
-            _ => throw new SerializationException("Unexpected record type."),
+            _ => throw new SerializationException("Invalid record type."),
```
Tanya-Solyanik, line 185: Should this be an `ArgumentException` as well?

JeremyKuhne: Input I want to be SerializationException to better match the existing BinaryFormatter experience. On writing the records were created incorrectly, which is a code problem, not a data problem- hence the ArgumentExceptions.

dreddy-work (approved):

PR #9115, opened 2023-05-12
Fix ToolStripButton errors in Accessibility Insights

### LINKED ISSUE #9108: ToolStripButton should not support both Invoke and Toggle patterns
### .NET version

.NET 8

Yes

Yes, the regression was introduced in #8961 and affects only .NET 8 

### Issue description

Accessibility Insights report following error for checked/checkable ToolStripButton:
A screenshot of test result in Accessibility Insights reporting error on ToolStripButton and showing that it should not support both Invoke and Toggle patterns

ToolStripMenuItem also supports both Invoke and Toggle patterns, but AI doesn't have this error because it applies only to "button" control type.

### Steps to reproduce

1. Run WinformsControlTest app from this repo and open "ToolStrips" form, or run sample app: [ToolStripButtonsTest.zip](https://github.com/dotnet/winforms/files/11190660/ToolStripButtonsTest.zip).
2. Open Accessibility Insights and test the form with it

dmitrii-drobotov:
Fixes #9108

## Proposed changes

* Changed ControlType for checkable/checked `ToolStripButton` from `Button` to `CheckBox` to allow having both Toggle and Invoke patterns. We cannot remove Invoke pattern for compatibility reasons, and Toggle pattern is needed to announce checked state. This is done by removing hardcoded `Button` type.
* Extend condition on which accessible object will get `CheckButton` role to include buttons with only `Checked = true` (without `CheckOnClick = true`), so that screen readers can also announce it as "checked";
* Allow setting custom role even if button is checkable/checked. For instance, this enables setting `RadioButton` role for `PropertyGridToolStripButton`, which can have `Checked = true` but shouldn't have `CheckBox` type because it doesn't behave like one.

## Customer Impact

- No errors in Accessibility Insights for ToolStripButton
- Screen readers announce checked/checkable ToolStripButton as "check box" instead of "button". 

## Regression? 

- Yes

## Risk

- Low. Change in screen reader announcement from "button" to "check box" may be somewhat confusing for users.

### Before

A screenshot of Accessibility Insights app showing test pass with 5 errors for a form with different ToolStripButtons

### After

A screenshot of Accessibility Insights app showing test pass without errors for a form with different ToolStripButtons
A screenshot of Accessibility Insights app showing that both Invoke and Toggle patterns are supported

## Test methodology 

- Unit tests
- Manual

## Accessibility testing  

Narrator announcements for ToolStripButtons from StripControls form in Accessibility Core test app:

#### Before

> New, button, Alt+ n, 
> Unchecked_CheckOnClick(F), button, 
> Unchecked_CheckOnClick(T), button, off, 
> Checked_CheckOnClick(T), button, on, 
> Checked_CheckOnClick(F), button, on, 
> Indeterminate_CheckOnClick(T), button, indeterminate, 
> Indeterminate_CheckOnClick(F), button, indeterminate, 

#### After

> New, button, Alt+ n, 
> Unchecked_CheckOnClick(F), button, 
> Unchecked_CheckOnClick(T) check box unchecked, 
> Checked_CheckOnClick(T) check box checked, 
> Checked_CheckOnClick(F) check box checked, 
> Indeterminate_CheckOnClick(T) check box indeterminate, 
> Indeterminate_CheckOnClick(F) check box indeterminate, 

 Tooltip announcements are removed for brevity. Similar result in JAWS and NVDA.

## Test environment(s) 

- .NET 8.0.0-preview.5.23260.3
- Win 11

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9115)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripButton.ToolStripButtonAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripButton.ToolStripButtonAccessibleObject.cs
@@ -70,7 +70,7 @@ public partial class ToolStripButton
 
         internal override void Toggle()
         {
-            if (_ownerItem.CheckOnClick)
+            if (IsPatternSupported(UiaCore.UIA.TogglePatternId))
             {
                 _ownerItem.Checked = !_ownerItem.Checked;
             }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripButton.ToolStripButtonAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripButton.ToolStripButtonAccessibleObject.cs
@@ -20,18 +20,6 @@ public partial class ToolStripButton
             _ownerItem = ownerItem;
         }
 
-        internal override object? GetPropertyValue(UiaCore.UIA propertyID) =>
-            propertyID switch
-            {
-                // If we don't set a default role for the accessible object
-                // it will be retrieved from Windows.
-                // And we don't have a 100% guarantee it will be correct, hence set it ourselves.
-                UiaCore.UIA.ControlTypePropertyId when
-                    _ownerItem.CheckOnClick
-                    => UiaCore.UIA.ButtonControlTypeId,
-                _ => base.GetPropertyValue(propertyID)
-            };
-
         internal override bool IsPatternSupported(UiaCore.UIA patternId) =>
             patternId switch
             {
@@ -43,7 +31,13 @@ public partial class ToolStripButton
         {
             get
             {
-                if (_ownerItem.CheckOnClick)
+                AccessibleRole role = _ownerItem.AccessibleRole;
+                if (role != AccessibleRole.Default)
+                {
+                    return role;
+                }
+
+                if (_ownerItem.CheckOnClick || _ownerItem.Checked)
                 {
                     return AccessibleRole.CheckButton;
                 }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripButton.ToolStripButtonAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripButton.ToolStripButtonAccessibleObject.cs
@@ -23,7 +23,7 @@ public partial class ToolStripButton
         internal override bool IsPatternSupported(UiaCore.UIA patternId) =>
             patternId switch
             {
-                UiaCore.UIA.TogglePatternId => _ownerItem.CheckOnClick || _ownerItem.Checked,
+                UiaCore.UIA.TogglePatternId => Role == AccessibleRole.CheckButton,
                 _ => base.IsPatternSupported(patternId)
             };
 
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripButton.ToolStripButtonAccessibleObjectTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripButton.ToolStripButtonAccessibleObjectTests.cs
@@ -142,6 +142,19 @@ public class ToolStripButton_ToolStripButtonAccessibleObjectTests
         Assert.Equal(expected, actual);
     }
 
+    [WinFormsFact]
+    public void ToolStripButtonAccessibleObject_IsTogglePatternSupported_ReturnsTrue_IfAccessibleRoleIsCheckButton()
+    {
+        using ToolStripButton toolStripButton = new()
+        {
+            AccessibleRole = AccessibleRole.CheckButton
+        };
+
+        object actual = toolStripButton.AccessibilityObject.IsPatternSupported(UiaCore.UIA.TogglePatternId);
+
+        Assert.Equal(true, actual);
+    }
+
     [WinFormsTheory]
     [InlineData(CheckState.Checked, (int)UiaCore.ToggleState.On)]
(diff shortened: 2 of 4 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ToolStripButton.ToolStripButtonAccessibleObject.cs
```diff
@@ -20,18 +20,6 @@ public ToolStripButtonAccessibleObject(ToolStripButton ownerItem) : base(ownerIt
             _ownerItem = ownerItem;
         }
 
-        internal override object? GetPropertyValue(UiaCore.UIA propertyID) =>
-            propertyID switch
-            {
-                // If we don't set a default role for the accessible object
-                // it will be retrieved from Windows.
-                // And we don't have a 100% guarantee it will be correct, hence set it ourselves.
```
dmitrii-drobotov, line 28: This comment doesn't seem to be relevant anymore. In fact, control type will be explicitly set based on Role:
https://github.com/dotnet/winforms/blob/6b0bffb607871ecb37a3184b7a0b3230b97cccd9/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.ToolStripItemAccessibleObject.cs#L110
For instance, for CheckButton role we will get CheckBoxControlTypeId:
https://github.com/dotnet/winforms/blob/6b0bffb607871ecb37a3184b7a0b3230b97cccd9/src/System.Windows.Forms/src/System/Windows/Forms/AccessibleRoleControlTypeMap.cs#L27

Tanya-Solyanik:
>Narrator announcements for ToolStripButtons from StripControls form in Accessibility Core test app:

I assume you are TAB-ing through the toolstrip, right?

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripButton.ToolStripButtonAccessibleObject.cs
```diff
@@ -43,7 +31,13 @@ public override AccessibleRole Role
         {
             get
             {
-                if (_ownerItem.CheckOnClick)
+                AccessibleRole role = _ownerItem.AccessibleRole;
+                if (role != AccessibleRole.Default)
+                {
+                    return role;
```
Tanya-Solyanik, line 37: It is theoretically possible that the owner control sets Role as `CheckButton` but `CheckOnClick` is false. In such a case the  dev can't override `IsPatternSupported` because it's internal. 
I would use `Role` to determine if pattern is supported:
```
                UiaCore.UIA.TogglePatternId => Role == AccessibleRole.CheckButton,
```

dmitrii-drobotov: Good point, I changed it to use Role.

dmitrii-drobotov: @Tanya-Solyanik - in this scenario, do you think it's correct that Toggle action would be no-op, or should we also change this condition to `if (IsPatternSupported(UiaCore.UIA.TogglePatternId))`?
 https://github.com/dotnet/winforms/blob/649b9ccd910f06e9a3d604d1159cbf228ac93295/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripButton.ToolStripButtonAccessibleObject.cs#L77-L83

Tanya-Solyanik: I really like `if (IsPatternSupported(UiaCore.UIA.TogglePatternId))`. THe UIA should not call into Toggle method anyway.

At some point I want to move these tests to the AccessibleObject class. So that interface method implementations are on the AccessibleObject class and these implementations call virtual methods in the specific AccessibleObject override.

```
void UiaCore.IToggleProvider.Toggle()  
{
    if (IsPatternSupported(UiaCore.UIA.TogglePatternId))
    {
        Toggle();
    }
}
```

This gives us an opportunity to open up virtuals for 3rd party developers eventually.

We started moving towards this model when we re-implementted Role/ControlType logic. But this is out of scope right now, as I would prefer that done for all accessible object implementations pattern by pattern.

dmitrii-drobotov: That would be pretty good, I agree! For now I just changed the Toggle method on ToolStripButton.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripButton.ToolStripButtonAccessibleObjectTests.cs
```diff
@@ -195,4 +195,17 @@ public void ToolStripButtonAccessibleObject_Toggle_Invoke()
...
+
+    [WinFormsFact]
+    public void ToolStripButtonAccessibleObject_Toggle_DoesNotChangeChecked_IfTogglePatternNotSupported()
+    {
+        using ToolStripButton toolStripButton = new();
+
+        Assert.False(toolStripButton.AccessibilityObject.IsPatternSupported(UiaCore.UIA.TogglePatternId));
+        Assert.False(toolStripButton.Checked);
+
+        toolStripButton.AccessibilityObject.Toggle();
```
Tanya-Solyanik, line 207: 👍

Tanya-Solyanik (approved):

PR #9116, opened 2023-05-12
Handle WM_DPICHANGED_BEFOREPARENT message for the Form in design scenarios

dreddy-work:
During runtime scenarios in a WinForms application, the main Form is considered a top-level window. Consequently, it will only receive the WM_DPICHANGED message in PermonitorV2 applications. On the other hand, child controls within the Form will receive both WM_DPICHANGED_BEFOREPARENT and WM_DPICHANGED_AFTERPARENT messages.

However, in WinForms designer scenarios, the Form behaves differently. It functions as a child window hosted within Visual Studio. As a result, during design-time, the Form will receive WM_DPICHANGED_BEFOREPARENT and WM_DPICHANGED_AFTERPARENT messages instead of WM_DPICHANGED, just like any other child controls would.

This is applicable only to the WinForms designer scenarios.

Fixes #https://github.com/microsoft/winforms-designer/issues/5192

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9116)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
@@ -2245,7 +2245,11 @@ public unsafe partial class Control :
             //ask the parent if it has the font height
             int localFontHeight = -1;
 
-            if (ParentInternal is not null && ParentInternal.CanAccessProperties)
+            // In design scenarios, Form is hosted within in the editor (VisualStudio). We stop iterating
+            // the parent hierarchy at the Form while getting Font for the WinForms application.
+            if (!(DesignMode && this is Form)
+                && ParentInternal is not null
+                && ParentInternal.CanAccessProperties)
             {
                 localFontHeight = ParentInternal.FontHeight;
             }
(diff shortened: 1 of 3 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -2245,7 +2245,11 @@ protected int FontHeight
             //ask the parent if it has the font height
             int localFontHeight = -1;
 
-            if (ParentInternal is not null && ParentInternal.CanAccessProperties)
+            // In design scenarios, Form is hosted within in the editor (VisualStudio). We stop iterating
```
Tanya-Solyanik, line 2248:
```suggestion
            // In design scenarios, Form is hosted within the editor (VisualStudio). We stop iterating
```

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -2245,7 +2245,11 @@ protected int FontHeight
...
+            // the parent hierarchy at the Form while getting Font for the WinForms application.
+            if (!(DesignMode && this is Form)
```
Tanya-Solyanik, line 2250: Is the same applicable to the MDI child windows?

PR #9120, opened 2023-05-15
Enable nullability in CodeDomComponentSerializationService

halgab:
## Proposed changes

- Enable nullability in `CodeDomComponentSerializationService`
- Change `ArrayList` with `List<T>` (contributes to #8140)
- Change all not serialized `Hashtable`s to `Dictionary<K, V>` and `HashSet<T>` (contributes to #8143)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9120)


src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs
```diff
@@ -179,7 +177,7 @@ private sealed class CodeDomSerializationStore : SerializationStore, ISerializab
         private static readonly TraceSwitch s_trace = new TraceSwitch("ComponentSerializationService", "Trace component serialization");
 #else
 #pragma warning disable CS0649
-        private static readonly TraceSwitch s_trace;
+        private static readonly TraceSwitch? s_trace;
```
dreddy-work, line 180: do you need this in release build?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs
```diff
@@ -289,37 +277,22 @@ internal void AddMember(object value, MemberDescriptor member, bool absolute)
...
         /// <summary>
         ///  The Close method closes this store and prevents any objects  from being serialized into it.  Once closed, the serialization store may be saved.
         /// </summary>
+        [MemberNotNull(nameof(_objectState))]
         public override void Close()
         {
             if (_objectState is null)
             {
-                Hashtable state = new Hashtable(_objects.Count);
+                Hashtable state = new Hashtable(_objects!.Count);
```
dreddy-work, line 295:
```suggestion
                Hashtable state = new(_objects!.Count);
```

dreddy-work: Can this be moved to dictionary as well?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs
```diff
@@ -385,20 +356,14 @@ public override void Close()
         /// <summary>
         ///  Deserializes the saved bits.
         /// </summary>
-        internal ICollection Deserialize(IServiceProvider provider)
+        internal List<object> Deserialize(IServiceProvider provider, IContainer? container = null)
         {
-            return Deserialize(provider, null, false, true, true);
+            List<object> collection = new List<object>();
```
dreddy-work, line 361:
```suggestion
            List<object> collection = new();
```

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs
```diff
@@ -607,19 +563,19 @@ void ISerializable.GetObjectData(SerializationInfo info, StreamingContext contex
...
-            private readonly Hashtable _nameResolveGuard = new Hashtable();
+            private readonly HashSet<string> _nameResolveGuard = new();
 
             public override object Deserialize(IDesignerSerializationManager manager, object state)
             {
                 throw new NotSupportedException();
             }
 
-            private void PopulateCompleteStatements(object data, string name, CodeStatementCollection completeStatements)
+            private void PopulateCompleteStatements(object? data, string name, CodeStatementCollection completeStatements)
```
dreddy-work, line 578: originally callers are ensuring `data `is never null. I don't see a reason to call this method when `data `is null.
Either exit early if `null `or refactor.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs
```diff
@@ -632,15 +588,15 @@ private void PopulateCompleteStatements(object data, string name, CodeStatementC
...
+                    if (!_expressions!.TryGetValue(name, out List<CodeExpression>? exps))
                     {
                         exps = new();
                         _expressions[name] = exps;
                     }
 
                     exps.Add(expression);
                 }
-                else
+                else if (data is not null)
```
dreddy-work, line 599: you  may not need this.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs
```diff
@@ -787,47 +730,46 @@ private static void DeserializeDesignTimeProperties(IDesignerSerializationManage
             ///  This is used to resolve nested component references.  NestedComponents don't exist as sited components within the DesignerHost, they are actually sited within a parent component.  This method takes the FullName defined on INestedSite and returns the component which matches it. outerComponent is the name of the topmost component which does exist in the DesignerHost
             ///  This code also exists in VSCodeDomDesignerLoader -- please keep them in sync.
             /// </summary>
-            private static IComponent ResolveNestedName(IDesignerSerializationManager manager, string name, ref string outerComponent)
+            private static IComponent? ResolveNestedName(IDesignerSerializationManager? manager, string name, [NotNullIfNotNull(nameof(manager))] out string? outerComponent)
             {
-                IComponent curComp = null;
-                if (name is not null && manager is not null)
+                IComponent? curComp = null;
+                if (manager is not null)
```
dreddy-work, line 736: nit: reduce nesting by quitting early.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs
```diff
@@ -713,47 +656,47 @@ internal void Deserialize(IDesignerSerializationManager manager, IDictionary obj
...
-                    _nameResolveGuard.Remove(e.Name);
+                    _nameResolveGuard.Remove(name);
                 }
             }
 
-            private void DeserializeDefaultProperties(IDesignerSerializationManager manager, string name, object state)
+            private void DeserializeDefaultProperties(IDesignerSerializationManager manager, string name, object? state)
             {
                 // Next, default properties, but only if we successfully  resolved.
                 if (state is not null && _applyDefaults)
```
dreddy-work, line 685: nit: reduce nesting with early exit.

dreddy-work (changes requested):

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs
halgab: I followed the blueprint set [here](https://github.com/dotnet/winforms/pull/8072#discussion_r1009826425). That's why I moved `TraceSwitchExtensions` from `System.Windows.Forms` to `System.Windows.Forms.Primitives`. If you prefer another approach I'll change the code accordingly

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs
halgab: `state` is assigned to private field `_objectState`, which is serialized at line 598 (old) / 554 (new). That made me a bit nervous about changing both objects' type, but if you think it's safe and that this code is the only one also deserializing, then I'd be happy to change them as well. I can prepare a commit in that direction that we can scrap if it's incorrect

halgab: Okay, I think you were right, it looks safe. I pushed the change. Thanks!

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs
halgab: Okay, apparently it breaks the unit tests because of [this line](https://github.com/dotnet/winforms/blob/1a1608fe3661728a8c3bc50959f5e0145b62ea48/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs#LL553C11-L553C11), but [this comment](https://github.com/dotnet/winforms/blob/1a1608fe3661728a8c3bc50959f5e0145b62ea48/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs#L547) makes me wonder if we couldn't just update the tests. What do you think?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs
dreddy-work: in such case, i would exit early in the function if it is null.

halgab: I've already made the change following your earlier comments

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs
halgab: Okay, apparently it breaks the unit tests because of [this line](https://github.com/dotnet/winforms/blob/1a1608fe3661728a8c3bc50959f5e0145b62ea48/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs#LL553C11-L553C11), but [this comment](https://github.com/dotnet/winforms/blob/1a1608fe3661728a8c3bc50959f5e0145b62ea48/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs#L547) makes me wonder if we couldn't just update the tests. What do you think? Should I rollback the last commit?

halgab:
Following [this comment](https://github.com/dotnet/winforms/pull/9120#discussion_r1194375880), I tried to modify the state hashtables with dictionaries. It makes the tests break because they control the content of those hashtables, but checking again, only the tests access them directly. They do through the `ISerializable` interface which is implemented by private class [`CodeDomSerializationStore`](https://github.com/dotnet/winforms/blob/1a1608fe3661728a8c3bc50959f5e0145b62ea48/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs#L174)
So I think these objects can be refactored safely. I then tried to push the refactoring further and replace the `object[]` array with a strongly typed, dedicated state object. All tests were updated accordingly.
I understand if you think this is to big of a change, and I can easily rollback the commits that have to do with the state Hashtables

dreddy-work:
This needs a thorough review. Requesting @JeremyKuhne and @Tanya-Solyanik on this.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs
```diff
@@ -125,12 +123,12 @@ public override ICollection Deserialize(SerializationStore store)
...
         ArgumentNullException.ThrowIfNull(store);
 
-        if (!(store is CodeDomSerializationStore cdStore))
+        if (store is not CodeDomSerializationStore cdStore)
         {
             throw new InvalidOperationException(SR.CodeDomComponentSerializationServiceUnknownStore);
         }
 
-        return cdStore.Deserialize(_provider);
+        return cdStore.Deserialize(_provider!);
```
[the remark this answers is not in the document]
halgab: `_provider` is initialised in the constructor (line 31), but there is another public constructor without any arguments that inits this field to null

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs
```diff
@@ -141,12 +139,12 @@ public override ICollection Deserialize(SerializationStore store, IContainer con
...
         ArgumentNullException.ThrowIfNull(container);
 
-        if (!(store is CodeDomSerializationStore cdStore))
+        if (store is not CodeDomSerializationStore cdStore)
         {
             throw new InvalidOperationException(SR.CodeDomComponentSerializationServiceUnknownStore);
         }
 
-        return cdStore.Deserialize(_provider, container);
+        return cdStore.Deserialize(_provider!, container);
```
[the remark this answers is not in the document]
halgab: Same thing here

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs
```diff
@@ -631,7 +583,7 @@ private void PopulateCompleteStatements(object data, string name, CodeStatementC
                 else if (data is CodeExpression expression)
                 {
                     // we handle expressions a little differently since they don't have a LHS or RHS they won't show up correctly in the statement table. We will deserialize them explicitly.
-                    if (!_expressions.TryGetValue(name, out List<CodeExpression> exps))
+                    if (!_expressions!.TryGetValue(name, out List<CodeExpression>? exps))
```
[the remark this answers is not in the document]
halgab: `ComponentListCodeDomSerializer` is used as a singleton. `_expressions` is used only for deserialisation. `_statementsTable` follows the same pattern.

I tried to change the `Deserialize` method to be a static method rather than calling an instance method on `s_instance`, with the hope that I could change `_expressions` to be a local variable that I could pass along methods. I hit a wall because the method `OnResolveName` is called as a callback to an event, and it needs access to `_expressions` (through its call to the `ResolveName` method).

We could always init `_expressions` and `_statementsTable` to a new empty (and arguably useless) dictionary at the creation of any `ComponentListCodeDomSerializer` instance, e.g.

`private Dictionary<string, List<CodeExpression>> _expressions = new();`

Another option would be to write

`private Dictionary<string, List<CodeExpression>> _expressions = null!;`

It's not great, but it's valid given current use cases of the field.

Let me know if/how you want to change this

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs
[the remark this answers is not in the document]
LeafShi1: Sorry for my issue, I mean _provider may be null here, so using ! here may be caused exception.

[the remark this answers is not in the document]
halgab: Okay, sorry. I've just pushed a commit to address this, but it slightly changes the behaviour. One path used to throw an NRE, and that's no longer the case, but I think it's fine. Let me know what you think

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs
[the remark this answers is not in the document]
LeafShi1: Instead of using `private Dictionary<string, List<CodeExpression>> _expressions = null!;` why not add a non-null judgment before using _expressions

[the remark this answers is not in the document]
halgab: you mean something like `if (_expressions is null || !_expressions.TryGetValue(name, out List<CodeExpression>? exps))`?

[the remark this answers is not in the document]
LeafShi1: Yes, like 
 ```
_if (_expressions is not null)
 {
     // we handle expressions a little differently since they don't have a LHS or RHS they won't show up correctly in the statement table. We will deserialize them explicitly.
     if (!_expressions.TryGetValue(name, out List<CodeExpression>? exps))
     {
         exps = new();
         _expressions[name] = exps;
     }

     exps.Add(expression);
 }_
```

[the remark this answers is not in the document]
Comment (halgab) [reply src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs:586]: I didn't add a null check on `_expressions` because here we know for a fact that ir's not null. The method `PopulateCompleteStatements` is only called by the method `Deserialize`, just after `_expressions = new();`
To make that explicit, I'm pushing a change where we pass `_expressions` as an argument to `PopulateCompleteStatements` (which is now static. It's a poor method name btw as it also populates expression, but I can't think of anything good as a replacement)

File: src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs [hunk 11]
```diff
@@ -178,55 +176,45 @@ private sealed class CodeDomSerializationStore : SerializationStore, ISerializab
...
 
         /// <summary>
         ///  Creates a new store.
         /// </summary>
-        internal CodeDomSerializationStore(IServiceProvider provider)
+        internal CodeDomSerializationStore(IServiceProvider? provider)
         {
             _provider = provider;
-            _objects = new Hashtable();
+            _objects = new Dictionary<object, ObjectData>();
```
dreddy-work, line 209: may be we should declare `_objects `read only and avoid assigning to `null`. Instead, just clear dictionary in the close and leave it to GC.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs
```diff
@@ -246,40 +234,34 @@ public override ICollection Errors
...
+        }
+
+        /// <summary>
+        ///  Adds a new member serialization to our list of things to serialize.
+        /// </summary>
+        internal void AddMember(object value, MemberDescriptor member, bool absolute)
+        {
+            ObjectData data = GetOrCreateObjectData(value);
+
+            s_trace.TraceVerbose($"ComponentSerialization: Adding object '{data._name}' ({data._value.GetType().FullName}:{member.Name}) {(absolute ? "ABSOLUTE" : "NORMAL")}");
```
dreddy-work, line 264: why not move this verbose log inside `GetOrCreateObjectData` ?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs
```diff
@@ -606,20 +553,25 @@ void ISerializable.GetObjectData(SerializationInfo info, StreamingContext contex
         /// </summary>
         private class ComponentListCodeDomSerializer : CodeDomSerializer
         {
-            internal static ComponentListCodeDomSerializer s_instance = new();
-            private Hashtable _statementsTable;
-            private Dictionary<string, List<CodeExpression>> _expressions;
-            private Hashtable _objectState; // only used during deserialization
+            internal static readonly ComponentListCodeDomSerializer s_instance = new();
+            private Dictionary<string, OrderedCodeStatementCollection?>? _statementsTable;
+            private Dictionary<string, List<CodeExpression>>? _expressions;
```
dreddy-work, line 558: nit: Would declare this read-only with assignment and clear it before `deserialize()`.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs
```diff
@@ -857,35 +793,35 @@ private bool ResolveName(IDesignerSerializationManager manager, string name, boo
                 // If it doesn't contain an OrderedCodeStatementsCollection this means one of two things:
                 // 1. We already resolved this name and shoved an instance in there.  In this case we just return the instance
                 // 2. There are no statements corresponding to this name, but there might be expressions that have never been deserialized, so we check for that and deserialize those.
-                if (_statementsTable[name] is OrderedCodeStatementCollection statements)
+                _statementsTable!.TryGetValue(name, out OrderedCodeStatementCollection? statements);
+                if (statements is not null)
```
dreddy-work, line 797: nit: return early and reduce nesting?

dreddy-work (dismissed):
adding @JeremyKuhne  and @lonitra  to take a second look.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.cs
halgab: The logs after the 2 callsites of method `GetOrCreateObjectData` are slightly different. I think that putting the tracing inside the method would complicate it with limited benefits, but I can do it anyway if you want me to.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeMethodMap.cs
```diff
@@ -19,14 +19,15 @@ internal class CodeMethodMap
...
+    internal CodeMethodMap(CodeMemberMethod method)
     {
+        Method = method;
+        _targetStatements = Method.Statements;
     }
 
-    internal CodeMethodMap(CodeStatementCollection? targetStatements, CodeMemberMethod method)
+    internal CodeMethodMap(CodeStatementCollection targetStatements)
     {
-        Method = method;
```
lonitra, line 28: Why was this removed from the constructor?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.CodeDomSerializationStore.ComponentListCodeDomSerializer.cs
```diff
@@ -0,0 +1,640 @@
...
+                    {
+                        Type? type = manager.GetType(typeName);
+                        if (type is null)
+                        {
+                            Trace(TraceLevel.Error, $"Type does not exist: {typeName}");
+                            manager.ReportError(new CodeDomSerializerException(string.Format(SR.SerializerTypeNotFound, typeName), manager));
+                        }
+                        else
+                        {
+                            if (statements.Count > 0)
```
lonitra, line 284: Let's use else if to reduce nesting here

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.CodeDomSerializationStore.ComponentListCodeDomSerializer.cs
```diff
@@ -0,0 +1,640 @@
...
+                        {
+                            foreach (CodeExpression exp in exps)
+                            {
+                                object? exValue = DeserializeExpression(manager, name, exp);
+                                if (exValue is not null && !resolved)
+                                {
+                                    if (canInvokeManager && manager.GetInstance(name) is null)
+                                    {
+                                        manager.SetName(exValue, name);
+                                        resolved = true;
```
lonitra, line 361: Is `DeserializeExpression` doing any state change? If not, I believe we maybe be able to `break` after this since once `resolved = true` we'll never go into the if statements again

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.CodeDomSerializationStore.ComponentListCodeDomSerializer.cs
```diff
@@ -0,0 +1,640 @@
...
+                    {
+                        // this is condition 2 of the comment at the start of this method.
+                        if (_expressions.TryGetValue(name, out List<CodeExpression>? exps))
+                        {
+                            foreach (CodeExpression exp in exps)
+                            {
+                                object? exValue = DeserializeExpression(manager, name, exp);
+                                if (exValue is not null && !resolved)
+                                {
+                                    if (canInvokeManager && manager.GetInstance(name) is null)
```
lonitra, line 358: Please combine these

lonitra (commented):
Nice work!

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeMethodMap.cs
halgab: Because it was always called with a null `method`

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.CodeDomSerializationStore.ComponentListCodeDomSerializer.cs
```diff
@@ -0,0 +1,640 @@
...
+            }
+
+            private static void DeserializeEventResets(IDesignerSerializationManager? manager, string name, List<string>? eventNames)
+            {
+                if (eventNames is not null && manager is not null && !string.IsNullOrEmpty(name))
+                {
+                    object? comp = manager.GetInstance(name);
+                    if (comp is not null && manager.GetService(typeof(IEventBindingService)) is IEventBindingService ebs)
+                    {
+                        PropertyDescriptorCollection eventProps = ebs.GetEventProperties(TypeDescriptor.GetEvents(comp));
```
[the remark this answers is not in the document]
halgab: Well, `IEventBindingService` is annotated, and `GetEventProperties` is supposed not to be null. I guess the null check protects us against implementations that don't respect this contract, but I can remove it if you want

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.CodeDomSerializationStore.ComponentListCodeDomSerializer.cs
halgab: I don't think it changes any state but it's such a huge method, I can't say I'm a 100% sure...

lonitra (approved):
LGTM!

halgab:
🥳

PR #9123, opened 2023-05-15
Use array[] and params array[]

JeremyKuhne:
List<T> isn't necessary here. Using params and arrays directly saves some memory and makes writing in the binary format more succinct.

Made a few arrays lazy to save a little memory bloat.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9123)


Shyam-Gupta (approved):
:shipit:

PR #9130, opened 2023-05-16
Enable nullability in remaining ToolTip members

gpetrou:
## Proposed changes

- Enable nullability in remaining ToolTip members.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9130)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
@@ -52,7 +52,7 @@ public partial class ListView : Control
     private ColumnHeaderStyle _headerStyle = ColumnHeaderStyle.Clickable;
     private SortOrder _sorting = SortOrder.None;
     private View _viewStyle = View.LargeIcon;
-    private string _toolTipCaption = string.Empty;
+    private string? _toolTipCaption = string.Empty;
 
     private const int LISTVIEWSTATE_ownerDraw = 0x00000001;
     private const int LISTVIEWSTATE_allowColumnReorder = 0x00000002;
@@ -3578,7 +3578,7 @@ public partial class ListView : Control
         }
     }
 
-    internal override unsafe ComCtl32.ToolInfoWrapper<Control> GetToolInfoWrapper(TOOLTIP_FLAGS flags, string caption, ToolTip tooltip)
+    internal override unsafe ComCtl32.ToolInfoWrapper<Control> GetToolInfoWrapper(TOOLTIP_FLAGS flags, string? caption, ToolTip tooltip)
     {
         // The "ShowItemToolTips" flag is required so that when the user hovers over the ListViewItem,
         // their own tooltip is displayed, not the ListViewItem tooltip.
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
@@ -38,7 +38,7 @@ public partial class TabControl : Control
     private Rectangle _cachedDisplayRect;
     private bool _currentlyScaling;
     private int _selectedIndex = -1;
-    private string _controlTipText = string.Empty;
+    private string? _controlTipText = string.Empty;
     private bool _handleInTable;
 
     // Events
@@ -1924,7 +1924,7 @@ public partial class TabControl : Control
 
         int commandID = (int)ttt->hdr.idFrom;
 
-        string tipText = GetToolTipText(GetTabPage(commandID));
+        string? tipText = GetToolTipText(GetTabPage(commandID));
         if (string.IsNullOrEmpty(tipText))
         {
             tipText = _controlTipText;
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolTip.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolTip.cs
@@ -828,7 +828,7 @@ public partial class ToolTip : Component, IExtenderProvider, IHandle<HWND>
         }
     }
 
-    private void SetToolInfo(Control control, string caption)
+    private void SetToolInfo(Control control, string? caption)
     {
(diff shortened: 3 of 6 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
```diff
@@ -54,7 +54,7 @@ public partial class ListView : Control
     private ColumnHeaderStyle _headerStyle = ColumnHeaderStyle.Clickable;
     private SortOrder _sorting = SortOrder.None;
     private View _viewStyle = View.LargeIcon;
-    private string _toolTipCaption = string.Empty;
+    private string? _toolTipCaption = string.Empty;
```
dreddy-work, line 57: why? i don't see it would ever be null.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
```diff
@@ -38,7 +38,7 @@ public partial class TabControl : Control
     private Rectangle _cachedDisplayRect;
     private bool _currentlyScaling;
     private int _selectedIndex = -1;
-    private string _controlTipText = string.Empty;
+    private string? _controlTipText = string.Empty;
```
dreddy-work, line 41: same here.

src/System.Windows.Forms/src/System/Windows/Forms/TabControl.cs
```diff
@@ -1934,7 +1934,7 @@ private unsafe void WmNeedText(ref Message m)
 
         int commandID = (int)ttt->hdr.idFrom;
 
-        string tipText = GetToolTipText(GetTabPage(commandID));
+        string? tipText = GetToolTipText(GetTabPage(commandID));
```
dreddy-work, line 1937: this shouldn't either.

src/System.Windows.Forms/src/System/Windows/Forms/ToolTip.cs
```diff
@@ -864,7 +864,7 @@ private void SetToolInfo(Control control, string caption)
 
     private void CreateRegion(Control control)
     {
-        string caption = GetToolTip(control);
+        string? caption = GetToolTip(control);
```
dreddy-work, line 867: same.

src/System.Windows.Forms/src/System/Windows/Forms/UpDownBase.cs
```diff
@@ -983,7 +983,7 @@ internal override void SetToolTip(ToolTip toolTip)
             return;
         }
 
-        string caption = toolTip.GetToolTip(this);
+        string? caption = toolTip.GetToolTip(this);
```
dreddy-work, line 986: here too.

dreddy-work (changes requested):

dreddy-work:
waiting on author.

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
gpetrou: I made this and the other items nullable because the methods such as `GetToolTip` return nullable. Should I use a null-forgiving operator instead?

dreddy-work: Current state of `GetToolTip` is not returning null. Can you point me to code where this is true?

gpetrou: In current main, nullability is not enabled in this part of the file.
It is enabled in this PR.
Please check https://github.com/dotnet/winforms/pull/9130/files#diff-f722fa05d83cae80d8fb79d78bcdae58f57170fcf56479a1a8cda42db71e6155R1029
`tipInfo.Caption` is nullable.

src/System.Windows.Forms/src/System/Windows/Forms/ToolTip.cs
```diff
@@ -1024,14 +1024,16 @@ private ToolInfoWrapper<HandleRef<HWND>> GetWinTOOLINFO(IWin32Window window)
...
+    public string? GetToolTip(Control? control)
     {
         if (control is null)
         {
             return string.Empty;
         }
 
-        return _tools.TryGetValue(control, out TipInfo tipInfo) ? tipInfo.Caption : string.Empty;
+        return _tools.TryGetValue(control, out TipInfo? tipInfo)
+            ? tipInfo.Caption
```
dreddy-work, line 1035: @Tanya-Solyanik, I am inclining to make changes on the [TipInfo.Caption](https://github.com/dotnet/winforms/blob/d4388546b5ee57f1bdc6f13bb703d9b1602073c2/src/System.Windows.Forms/src/System/Windows/Forms/ToolTip.TipInfo.cs#L26) as non-nullable and return `string.empty` in those cases and avoid `!` in many places. Do you see any concerns?

gpetrou: The `TipInfo` constructor is called in many places with `caption` possibly null though. What would we do in such cases?

dreddy-work: I wouldn't touch them yet. Only caption property's setter. Let me think it through. Will update this

dreddy-work: `Caption `is passed to native API and native APIs expect null. So, we may not want to change it. Certainly, without complete scenario validations.

dreddy-work (approved):

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
dreddy-work: We can not avoid Caption being null give Native APIs using this allows it to be null.

PR #9168, opened 2023-05-24
[release/7.0] Fix mouse location calculation in BehaviorService mouse message processing

### LINKED ISSUE #9160: System.Windows.Forms.Design.ControlDesigner does not handle mouse messages on adorner window such as resizing
### .NET version

.NET Core 7

Yes

Works fine for .NET 6

### Issue description

We use .NET System.Windows.Forms.Design services for the run-time implementation of the Form designer functionality.
Form designer works fine in .Net core 6 (and .net 4.62), allowing placing and resizing controls on the design surface. However, resizing does not work with .net 7.

### Steps to reproduce

It would be hard to give a reproducible scenario, but by tracing the source code, we've found the following:

.NET 6
BehaviorService.cs
...
private bool ProcessMouseMessage(IntPtr hWnd, Interop.User32.WM msg, int x, int y)
    Point point = new Point(x, y);
    Interop.User32.MapWindowPoints(IntPtr.Zero, adornerWindow.Handle, ref point, 1U);
    Message message = Message.Create(hWnd, msg, (IntPtr)0, Interop.PARAM.FromLowHigh(p.Y, p.X));

.NET 7

BehaviorService.cs
...
private bool ProcessMouseMessage(IntPtr hWnd, Interop.User32.WM msg, int x, int y)
    Point point = new Point(x, y);
    adornerWindow.PointToClient(p); // should be p =adornerWindow.PointToClient(p);
    Message message = Message.Create(hWnd, msg, (IntPtr)0, Interop.PARAM.FromLowHigh(p.Y, p.X));

adornerWindow.PointToClient internally uses MapWindowPoints, but its result is ignored. This is causing sending screen coordinates to the adornerWindow and clearing _hitTestedGlyph field, which results in adorners not behaving as they should.
The same applies to the designer action glyph (displayed at the top right corner of the selected control) - it does nothing on our tests.

dmitrii-drobotov:
Fixes #9160, backport of #8522

## Proposed changes

- Add missing assignment of result of PointToClient call.

## Customer Impact

- Fixes resize operation and action glyphs for runtime designer scenarios

## Regression? 

- Yes, regression in .NET 7

## Risk

- Minimal

### Before

Before

### After

After

## Test methodology 

- Manual

## Test environment(s) 

- .NET 7.0.7

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/Behavior/BehaviorService.AdornerWindow.MouseHook.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/Behavior/BehaviorService.AdornerWindow.MouseHook.cs
@@ -174,8 +174,7 @@ namespace System.Windows.Forms.Design.Behavior
                             try
                             {
                                 _processingMessage = true;
-                                var pt = new Point(x, y);
-                                adornerWindow.PointToClient(pt);
+                                Point pt = adornerWindow.PointToClient(new Point(x, y));
                                 Message m = Message.Create(hWnd, msg, 0, PARAM.FromLowHigh(pt.Y, pt.X));
 
                                 // No one knows why we get an extra click here from VS. As a workaround, we check the TimeStamp and discard it.


dreddy-work (approved):

JeremyKuhne (approved):

Nora-Zhou01:
Verified on .NET 7.0 test pass build: .NET 7.0.108-servicing.23319.8, this issue is fixed. In Winforms repo, the [DemoConsole](https://github.com/dotnet/winforms/tree/main/src/System.Windows.Forms/tests/IntegrationTests/DesignSurface/DemoConsole) test app could resize controls and without exception when moving controls in .Net 7.0.8 Gif as below:

https://github.com/dotnet/winforms/assets/104609169/eeda7e1a-5f06-4a60-8c87-ff28eda52e66

PR #9178, opened 2023-05-25
Hook BinaryFormatter replacement in Resx / Clipboard

JeremyKuhne:
This uses the BinaryFormat support classes to handle common framework types before falling back to the BinaryFormatter.

There is some cleanup in the Resx code and simplification in DataObject by adding internal constant strings for known DataFormats.

Also move MultitargetUtil to the normal CriticalException handler.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9178)

### DIFF
--- a/src/System.Windows.Forms/src/System/Resources/ResXDataNode.cs
+++ b/src/System.Windows.Forms/src/System/Resources/ResXDataNode.cs
@@ -11,6 +11,7 @@ using System.Runtime.Serialization;
 using System.Runtime.Serialization.Formatters.Binary;
 using System.Text;
 using System.Windows.Forms;
+using System.Windows.Forms.BinaryFormat;
 using System.Xml;
 
 namespace System.Resources;
@@ -37,7 +38,8 @@ public sealed class ResXDataNode : ISerializable
     private IFormatter? _binaryFormatter;
 
     // This is going to be used to check if a ResXDataNode is of type ResXFileRef
-    private static readonly ITypeResolutionService s_internalTypeResolver = new AssemblyNamesTypeResolutionService(new AssemblyName[] { new AssemblyName("System.Windows.Forms") });
+    private static readonly ITypeResolutionService s_internalTypeResolver
+        = new AssemblyNamesTypeResolutionService(new AssemblyName[] { new AssemblyName("System.Windows.Forms") });
 
     // Callback function to get type name for multitargeting.
     // No public property to force using constructors for the following reasons:
@@ -89,7 +91,7 @@ public sealed class ResXDataNode : ISerializable
         {
             throw new InvalidOperationException(string.Format(SR.NotSerializableType, name, valueType.FullName));
         }
-#pragma warning restore SYSLIB0050 // Type or member is obsolete
+#pragma warning restore SYSLIB0050
 
         if (value is not null)
         {
@@ -161,12 +163,7 @@ public sealed class ResXDataNode : ISerializable
         }
         set
         {
-            ArgumentNullException.ThrowIfNull(value, nameof(Name));
-            if (value.Length == 0)
-            {
-                throw new ArgumentException(nameof(Name));
-            }
-
+            ArgumentException.ThrowIfNullOrEmpty(value, nameof(Name));
             _name = value;
         }
     }
@@ -201,7 +198,8 @@ public sealed class ResXDataNode : ISerializable
         string raw = Convert.ToBase64String(data);
         if (raw.Length > lineWrap)
         {
-            StringBuilder output = new StringBuilder(raw.Length + (raw.Length / lineWrap) * 3); // word wrap on lineWrap chars, \r\n
+            // Word wrap on lineWrap chars, \r\n
+            StringBuilder output = new StringBuilder(raw.Length + (raw.Length / lineWrap) * 3);
             int current = 0;
             for (; current < raw.Length - lineWrap; current += lineWrap)
             {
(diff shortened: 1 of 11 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/DataObject.BitmapBinder.cs
```diff
@@ -11,66 +11,52 @@ namespace System.Windows.Forms;
...
-        /// <param name="typeName"></param>
         public override void BindToName(Type serializedType, out string? assemblyName, out string? typeName)
         {
-            // null strings will follow the default codepath in BinaryFormatter
+            // Null values will follow the default codepath in BinaryFormatter.
             assemblyName = null;
             typeName = null;
-            if (serializedType is not null && !serializedType.Equals(typeof(string)) && !serializedType.Equals(typeof(Bitmap)))
+
+            // Bitmap and string types are safe type to serialize/deserialize.
```
KlausLoeffelmann, line 58:
```suggestion
            // Bitmap and string types are safe types to serialize/deserialize.
```

src/System.Windows.Forms/src/System/Windows/Forms/DataObject.OleConverter.cs
```diff
@@ -353,55 +351,44 @@ private static unsafe Stream ReadByteStreamFromHandle(nint handle, out bool isSe
         }
 
         /// <summary>
-        ///  Creates a new instance of the Object that has been persisted into the
-        ///  handle.
+        ///  Creates a new instance of the Object that has been persisted into the handle.
```
KlausLoeffelmann, line 354:
```suggestion
        ///  Creates a new instance of the object that has been persisted into the handle.
```

KlausLoeffelmann (dismissed):
Looks good to me - a few typo nits are what I found.

src/System.Windows.Forms/src/System/Resources/ResXDataNode.cs
```diff
@@ -89,7 +91,7 @@ public ResXDataNode(string name, object? value, Func<Type?, string>? typeNameCon
         {
             throw new InvalidOperationException(string.Format(SR.NotSerializableType, name, valueType.FullName));
         }
-#pragma warning restore SYSLIB0050 // Type or member is obsolete
+#pragma warning restore SYSLIB0050
```
Tanya-Solyanik, line 94: I find that comment useful...

src/System.Windows.Forms/src/System/Resources/ResXSerializationBinder.cs
```diff
@@ -35,69 +42,66 @@ internal ResXSerializationBinder(Func<Type?, string>? typeNameConverter)
...
 
-        // Binder couldn't handle it, let the default loader take over.
+        // Hand back what we found or null to let the default loader take over.
         return type;
     }
 
-    // Get the multitarget-aware string representation for the give type.
     public override void BindToName(Type serializedType, out string? assemblyName, out string? typeName)
     {
-        // Normally we don't change typeName when changing the target framework,
```
Tanya-Solyanik, line 73: Why did you remove this comment? This is a test case to try to execute this code.

src/System.Windows.Forms/src/System/Windows/Forms/DataFormats.cs
```diff
@@ -16,143 +16,142 @@ namespace System.Windows.Forms;
 /// </summary>
 public static partial class DataFormats
 {
+    internal const string TextConstant = "Text";
```
Tanya-Solyanik, line 19: In Designer we used pattern `internal const string Text = nameof(Text);` in class named for example `DataFormatConstants`

src/System.Windows.Forms/src/System/Resources/ResXDataNode.cs
```diff
@@ -161,12 +163,7 @@ public string Name
         }
         set
         {
-            ArgumentNullException.ThrowIfNull(value, nameof(Name));
-            if (value.Length == 0)
-            {
-                throw new ArgumentException(nameof(Name));
-            }
-
+            ArgumentException.ThrowIfNullOrEmpty(value, nameof(Name));
```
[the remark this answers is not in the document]
JeremyKuhne: It does the same thing as above. It throws ArgumentNull if null or Argument if empty.

Tanya-Solyanik (approved):

src/System.Windows.Forms/src/System/Windows/Forms/DataFormats.cs
KlausLoeffelmann: I, too, deviate sometimes from that pattern when a name (like `Text`) becomes too generic or may collide (even if only by being a "typical class element name") with a class member.

src/System.Windows.Forms/src/System/Resources/ResXDataNode.cs
JeremyKuhne: I keep it on the matching one above- it's redundant to have the same comment a few lines later.

src/System.Windows.Forms/src/System/Windows/Forms/DataFormats.cs
JeremyKuhne: Collision is a problem here- I wanted discoverability.

src/System.Windows.Forms/src/System/Resources/ResXSerializationBinder.cs
JeremyKuhne: It makes no sense. Outside of that the code here has nothing explicitly to do with a given scenario- anything like this should live with the type name converter or type resolver implementation. But again, it doesn't make sense.

Serialization always happens with the object instance. In reverse, the deserialization is the data not a desired target type. That is just the way it works, so it is super unclear what this is trying to say. There is no "property" involved here or in the elsewhere in the ResX code directly.

Issue #9180, opened 2023-05-26
Failed to Clipboard.GetData() as the specified format


Philip-Wang01:
### .NET version

.NET 8.0.100-preview.5.23274.2

### Did it work in .NET Framework?

Yes

### Did it work in any of the earlier releases of .NET Core or .NET 5+?

Yes, it works in .NET 6.0/7.0/8.0.100-preview5.23253.20. Maybe caused by: https://github.com/dotnet/winforms/pull/9088

### Issue description

Testapp: [Testapp.zip](https://github.com/dotnet/winforms/files/11572062/Testapp.zip)

.NET 8.0
![image](https://github.com/dotnet/winforms/assets/86937911/281bcb3c-3ec9-4f26-8b21-8f113dc4eae4)

.NET 7.0
![image](https://github.com/dotnet/winforms/assets/86937911/eb312cea-aab2-4f02-b255-5f60521144b2)

### Steps to reproduce

1. Create a WinForms .NET Core project.
2. Add a button control to form designer, double click button1 and type below code in function button1_Click.
```
Color c = System.Drawing.Color.Black;
string format = "testabc";
Clipboard.Clear();
Clipboard.SetData(format, c);
Clipboard.ContainsData(format);
object ob = null;
ob = Clipboard.GetData(format);
```
3. Add a breakpoint at the end of the event, build and run the application, click button1.
4. Observe the value of ob.

elachlan:
@JeremyKuhne I guess you are still working through the binary formatter changes?

#6267

JeremyKuhne:
This doesn't reproduce for me with the current bits when the `BinaryFormatter` is enabled, which it should be. When it is disabled, it returns null. @Philip-Wang01 can you debug with first chance exceptions to see if you're getting a NotSupportedException with the binary formatter?

Some notes on the BinaryFormatter being disabled:

Unfortunately, the OLE clipboard window always swallows errors when setting data (in `WM_RENDERFORMAT` it always returns an `LRESULT` of `0`). So setting always succeeds. If we try to get the data out it will fail with `CLIPBRD_E_BAD_DATA`, which we swallow. I've added some debug assistance around that in the #9195 PR.

I'm thinking through what we might be able to do to surface that the `BinaryFormatter` has been disabled and will likely create an issue to track soon.

Philip-Wang01:
[...]

PR #9195, opened 2023-05-30
Clean up OleConverter code

JeremyKuhne:
The OleConverter class had significant readability issues. This brings it closer to standard and cuts over 100 lines of code. Getting the code into better shape is important with BinaryFormatter changes and the need to move this to ComWrappers moving forward.

Recommend reviewing split.

Notable bits:

- The class is never derived, and is now sealed
- Members that do nothing now clearly do nothing (write methods)
- Moved some single call methods inline callees

The name isn't great so I'm going to follow up with a change to ComDataObjectAdapter for increased clarity.

Also add a few tests related to #9180.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9195)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/NativeMethods.txt
+++ b/src/System.Windows.Forms.Primitives/src/NativeMethods.txt
@@ -26,6 +26,7 @@ CloseDesktop
 CloseEnhMetaFile
 CloseHandle
 CloseThemeData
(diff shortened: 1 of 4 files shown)


dreddy-work (approved):
LGTM.

PR #9204, opened 2023-05-31
Fixed extra ListViewItemImageAccessibleObject for ListViewItem without any Image

### LINKED ISSUE #9125: AccessibilityInsight: An on-screen element must not have a null BoundingRectangle property when ListView has DataBinding
### .NET version

.NET 8.0.100-preview.5.23261.16

Yes

Yes, it works in previous .Net 7.0 and 6.0.

### Issue description

AccessibilityInsight detects an error about "An on-screen elememt must not have a null BoundingRectangle property" for ListView with DataBinding, and in Inspect, ListView has extra Image for every ListItem in UIA hierarchy
This issue doesn't repro ListView without DataBinding.

**.Net 8.0 in AccessibilityInsight:**
image

**.Net 8.0 in Inspect:**
image

**.Net 7.0 in AccessibilityInsight:**
image

**.Net 7.0 in Inspect:**
image

### Steps to reproduce

**Repro steps:**
1.	Create a Winforms .Net Core application with ListView added
2.	Create a DataSet with TableData by code
3.	Bind this DataSet table data to ListView
4.	Build and run application 
5.	Use AccessibilityInsight to detect this ListView

or use attached sample application:
[WinFormsApp1.zip](https://github.com/dotnet/winforms/files/11483200/WinFormsApp1.zip)

dkazennov:
Fixes #9125 (this is a regression from ##7672)

## Proposed changes

- changed `HasImage` property check in `ListViewItemBaseAccessibleObject` class.
- Added new regressive unit test `ListViewItemDetailsAccessibleObject_HasImage_ReturnsExpected` (it covers `HasImage` property which is responsible for creation of `ListViewItemImageAccessibleObject` accessible object).
- Fixed faulty unit tests affected by this regression

## Customer Impact

- User can see correct information about ListViewItem without image via accessibility tools

## Regression? 

- Yes

## Risk

- Minimal

### Before

image
image

### After

image
image

## Test methodology 

- Manual
- Unit tests

## Test environment(s) 

- dotnet 8.0.100-preview.3.23178.7

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9204)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.ListViewItemBaseAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.ListViewItemBaseAccessibleObject.cs
@@ -53,7 +53,8 @@ public partial class ListViewItem
         internal override UiaCore.IRawElementProviderFragmentRoot FragmentRoot
             => _owningListView.AccessibilityObject;
 
-        protected bool HasImage => _owningItem.ImageIndex != ImageList.Indexer.DefaultIndex;
+        internal bool HasImage => _owningItem.ImageList is not null && _owningItem.ImageList.Images.Count > 0
+            && _owningItem.ImageIndex != ImageList.Indexer.DefaultIndex;
 
         internal override bool IsItemSelected
             => (State & AccessibleStates.Selected) != 0;
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItem.ListViewItemAccessibleObjectTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItem.ListViewItemAccessibleObjectTests.cs
@@ -1618,6 +1618,7 @@ public class ListViewItem_ListViewItemAccessibleObjectTests
             View = View.Details,
             SmallImageList = imageCollection
         };
+
         listView.Columns.Add(new ColumnHeader());
         var listViewItem = new ListViewItem("Item 1", imageIndex: hasImage ? 0 : -1);
         listView.Items.Add(listViewItem);
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItem.ListViewItemDetailsAccessibleObjectTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItem.ListViewItemDetailsAccessibleObjectTests.cs
@@ -127,4 +127,32 @@ public class ListViewItem_ListViewItemDetailsAccessibleObjectTests
         Assert.Empty(item.AccessibilityObject.TestAccessor().Dynamic._listViewSubItemAccessibleObjects);
         Assert.False(control.IsHandleCreated);
     }
+
+    [WinFormsTheory]
+    [InlineData(true)]
+    [InlineData(false)]
+    public void ListViewItemDetailsAccessibleObject_HasImage_ReturnsExpected(bool addImages)
+    {
+        using ListView listView = new()
+        {
+            View = View.Details
+        };
+
+        listView.Columns.Add(new ColumnHeader());
+
+        using ImageList imageList = new();
+        imageList.Images.Add(Form.DefaultIcon);
+
+        if (addImages)
+        {
+            listView.SmallImageList = imageList;
+        }
+
+        ListViewItem listViewItem = new ListViewItem("1", 0);
+        listView.Items.Add(listViewItem);
+
+        var accessibleObject = (ListViewItemDetailsAccessibleObject)listView.Items[0].AccessibilityObject;
+        Assert.Equal(addImages, accessibleObject.HasImage);
+        Assert.False(listView.IsHandleCreated);
+    }
 }
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItem.ListViewItemWithImageAccessibleObjectTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItem.ListViewItemWithImageAccessibleObjectTests.cs
@@ -35,7 +35,8 @@ public class ListViewItem_ListViewItemWithImageAccessibleObjectTests
         using ListView control = new()
         {
             View = view,
-            SmallImageList = imageCollection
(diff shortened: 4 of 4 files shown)


dkazennov:
> Is it possible to add a regression test?

I'll add such tests.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItem.ListViewItemAccessibleObjectTests.cs
```diff
@@ -1627,6 +1627,34 @@ public void ListViewItemAccessibleObject_GetChildIndex_ReturnsExpected_Image(boo
...
+        if (addImages)
+        {
+            listView.SmallImageList = imageList;
+        }
+
+        ListViewItem listViewItem = new ListViewItem("Item 1", imageIndex: 0);
+        listViewItem.ImageKey = "itemImageKey";
+        listView.Items.Add(listViewItem);
+
+        var accessibleObject = (ListViewItemDetailsAccessibleObject)listView.Items[0].AccessibilityObject;
```
dmitrii-drobotov, line 1653: Perhaps this test would better fit in `ListViewItem_ListViewItemDetailsAccessibleObjectTests`?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItem.ListViewItemAccessibleObjectTests.cs
```diff
@@ -1627,6 +1627,34 @@ public void ListViewItemAccessibleObject_GetChildIndex_ReturnsExpected_Image(boo
...
+        };
+
+        listView.Columns.Add(new ColumnHeader());
+
+        using ImageList imageList = new();
+        imageList.Images.Add("itemImageKey", Form.DefaultIcon);
+        if (addImages)
+        {
+            listView.SmallImageList = imageList;
+        }
```
dmitrii-drobotov, line 1647: Nit:
```suggestion
        if (addImages)
        {
            using ImageList imageList = new();
            imageList.Images.Add("itemImageKey", Form.DefaultIcon);
            listView.SmallImageList = imageList;
        }
```

src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.ListViewItemBaseAccessibleObject.cs
```diff
@@ -53,7 +53,7 @@ internal int CurrentIndex
         internal override UiaCore.IRawElementProviderFragmentRoot FragmentRoot
             => _owningListView.AccessibilityObject;
 
-        protected bool HasImage => _owningItem.ImageIndex != ImageList.Indexer.DefaultIndex;
+        internal bool HasImage => _owningItem.ImageList is not null && _owningItem.ImageList.Images.Count > 0;
```
dmitrii-drobotov, line 56: I think the index condition is still needed:
```suggestion
        internal bool HasImage => _owningItem.ImageList is not null && _owningItem.ImageList.Images.Count > 0 && _owningItem.ImageIndex != ImageList.Indexer.DefaultIndex;
```

If ListView has an ImageList with images, but a specific ListViewItem doesn't have `ImageIndex` set, this item would now have an extra image in accessibility tree.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItem.ListViewItemAccessibleObjectTests.cs
dkazennov: > ListViewItem_ListViewItemDetailsAccessibleObjectTests

Well, this test was initially about different property and was put into a different test class.
I suppose this test does belong to `ListViewItemDetailsAccessibleObjectTests` indeed.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItem.ListViewItemAccessibleObjectTests.cs
dkazennov: This is what I stumbled upon while working on the test. We shouldn't do that, because after the if {} block `ImageList` will be disposed and we will be left with `SmallImageList` being null.

src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.ListViewItemBaseAccessibleObject.cs
dkazennov: > a specific ListViewItem doesn't have `ImageIndex` set

Hm. But in this case `ListViewItem` won't have any image attached to it. Right?
I'm not sure about this extra condition. The fix seems to work as it is now.

dmitrii-drobotov: > But in this case ListViewItem won't have any image attached to it. Right?

Yes, but it will now get an extra accessible object. I've a made a simple repro project, which can demonstrate what I mean:
image
[WinFormsApp7.zip](https://github.com/dotnet/winforms/files/11637199/WinFormsApp7.zip)

This image has non-empty bounds, so it doesn't show up in AI errors, but still seems unnecessary.

dkazennov: @dmitrii-drobotov I've double checked this.
In your application with new WinForms dll's I have no extr accessible object.

image

dkazennov: @dmitrii-drobotov that was false-positive result. Now new dll's with my fix and your proposal indeed work as intended and there is no extra image object in your application as well. Thank you for important feedback!

dkazennov:
Requested CTI test.

Tanya-Solyanik (dismissed):

dmitrii-drobotov (dismissed):
Looks good, I only have a small nit.

Tanya-Solyanik (approved):
Thank you!

PR #9205, opened 2023-05-31
Final round of clipboard cleanup

JeremyKuhne:
Following logic should be significantly easier now.

- Move methods inline in ComDataObjectAdapter
- Make methods static and pass in state
- Move a few methods inlinein DataObject
- Use Span to save filelist
- Use HGLOBAL terms

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9205)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataObject.ComDataObjectAdapter.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataObject.ComDataObjectAdapter.cs
@@ -37,76 +37,9 @@ public partial class DataObject
         public IComDataObject OleDataObject => _innerData;
 
         /// <summary>
-        ///  Uses IStream and retrieves the specified format from the bound IComDataObject.
-        /// </summary>
-        private unsafe object? GetDataFromOleIStream(string format)
-        {
-            STGMEDIUM medium = default;
-
-            FORMATETC formatetc = new()
-            {
-                cfFormat = (short)(ushort)DataFormats.GetFormat(format).Id,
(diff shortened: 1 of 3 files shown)


Tanya-Solyanik:
>Make methods static and pass in state

what is the reason for this change?

src/System.Windows.Forms/src/System/Windows/Forms/DataObject.ComDataObjectAdapter.cs
```diff
@@ -189,9 +215,9 @@ static unsafe string ReadUtf8StringFromHGLOBAL(nint hglobal)
 
             return data;
 
-            // Currently this only supports Bitmap
-            static object? GetDataFromOleOther(string format, IComDataObject dataObject)
+            static object? TryGetBitmapData(IComDataObject dataObject, string format)
             {
+                // Currently this only supports Bitmap.
```
Tanya-Solyanik, line 220: Function name makes this comment redundant.

Tanya-Solyanik (approved):

JeremyKuhne:
> what is the reason for this change?

To make instantly clear what the dependencies are- only the parameters. Otherwise you need to know what state the instance has.

src/System.Windows.Forms/src/System/Windows/Forms/DataObject.ComDataObjectAdapter.cs
JeremyKuhne: I'll remove it in the BF change.

PR #9206, opened 2023-06-01
Handle BinaryFormatter being disabled with the Clipboard

JeremyKuhne:
Windows swallows all errors when getting data for the clipboard. This was resulting in seeming success, but a return of a null when trying to get the data out (as the data would be invalid and we would swallow that error on data retrieval).

There aren't great answers to this. To give some indication of the source of the failure I've implemented binary format support for NotSupportedException and put that on the clipboard if we need the BinaryFormatter.  Getting the value will give this object, with the message that the BinaryFormatter is disabled.

While we could check on settting the data it would involve deserializing to see that it is what is on the clipboard. I don't want to take that performance hit for every (normally working) usage of the clipboard.

cc @GrabYourPitchforks 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9206)


JeremyKuhne:
I also added some unit tests for the SerializationBinder.

dreddy-work (approved):
LGTM.

PR #9245, opened 2023-06-08
Update bug_report.yml

RussKie:
Add "Not Tested/Verified" as an option for "Did it work in .NET Framework" question.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9245)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -25,7 +25,7 @@ body:
       options:
         - 'Yes'
         - 'No'
-        - 'N/A'
+        - 'Not tested/verified'
     validations:
       required: true
   - type: textarea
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -25,6 +25,7 @@ body:
       options:
         - 'Yes'
         - 'No'
+        - 'N/A'
```
dreddy-work, line 28: what is the scenario you choose for NA?

dreddy-work: Is it like, you didn't get chance to verify? or this is completely new to .NET?

dreddy-work: May be we should add something like 

'Not Tested/Verified' and NA is more like it is not applicable to .NETFramework (for example, some of the PerMonitorV2 mode issues)

dreddy-work (approved):

Issue #6590, opened 2022-02-01
Code of the DateTimePicker.Text doesn't match its description in documentation


ArtemTatarinov:
* .NET Core Version: From 3.0 to 7.0
* Have you experienced this same bug with .NET Framework?: Yes, from 2.0 to 4.8

**Problem description:**
`Text` property of the `DateTimePicker` class returns the value of the `Text` property of the base class (i.e. `Control`):
![image](https://user-images.githubusercontent.com/87859299/151927126-aa0f5ddf-a996-47f0-b2e2-1a56836ab998.png)

Yet the documentation says that `Text` is an equivalent of the formatted `Value` property (https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.datetimepicker.text):
![image](https://user-images.githubusercontent.com/87859299/151927841-6770fea0-3355-4ec2-8764-5e34ca5af292.png)

This behavior is the same for all .Net versions: from .Net Framework 2.0 to .Net Core 7.0, and also documentation says the opposite thing for all of these versions. The `Text` property isn't visible at the Visual Studio designer, yet it's visible at the IntelliSense, because it has `public` modifier.

This is how it looks while debugging:
![image](https://user-images.githubusercontent.com/87859299/151927903-3a631906-01f7-4bc0-85ab-c040473482f6.png)

Looks like either the code or the documentation should be changed to be consistent.

ArtemTatarinov:
Hi, @merriemcgaw, @Tanya-Solyanik, @dreddy-work, could you please look at it?

Tanya-Solyanik:
Suggestion:
Change the getter to `return Value;` in .NET 7 and fix documentation for other versions.

Ashley-Li:
Verified with .NET SDK 7.0.100-preview.3.22123.5 build, this issue is fixed.
![image](https://user-images.githubusercontent.com/56664234/155498391-fcecd2d2-19da-4dc7-a970-59000978e720.png)

Cassie-Li01:
Verified on .NET SDK 7.0.100-preview.3.22152.5 build, this issue is fixed.
![image](https://user-images.githubusercontent.com/96758262/156960415-f16fb7e9-2331-4d1f-80f4-991a2064de78.png)

PR #9253, opened 2023-06-08
Update DateTimePicker and Accessibility class

### LINKED ISSUE #9232: A11y_.Net Core_Winforms_MainForm_Common Controls1_Screen reader : Narrator is announcing the Time when the focus lands to the date picker control , but visually time is not present.
### .NET version

.NET-SDK-8.0.100-preview.4.23260.5

No

_No response_

### GitHubTags:
#A11yMAS;#A11yTCS;#BM_.NET Core[Winforms]-Win32-May2023;#.NET Core;#A11ySev3;#WCAG1.3.1;#DesktopApp;#Win11;#Benchmark;#Narrator;#NVDA;

### Actual:
The Narrator screen reader is announcing the Time along with Date when the Focus lands on the Date control in the Date Picker section , But visually there is no Time stamp provided in the screen.

Note : This issue is also reproducing with NVDA screen reader. NVDA is announcing the "Effects" Label but not the "Sample" Label

### Expected:
The Narrator/NVDA screen reader should announce the contents which are present visually. The time stamp should be visible in the UI or Screen reader should not read the time stamp if it is not present visually.

### User Impact:
Low vision and Screen reader user will be confused with this announcement and content not present visually.

### Steps to reproduce

1. Install and Open Winforms App
2. Main Form screen would appear.
3. Tab till "Common Controls_1: Testing the controls under common controls Tab" and Hit enter.
4. Navigate to "Date picker" Section
5. Observe with Screen reader that the Time is announced by Narrator/NVDA but visually it is not present.

### Attachment : 

Narrator Attachment : 
Narrator Time

NVDA Attachment:
NVDA Timing

Epica3055:
Fixes #9232 

## Proposed changes

- 
- 
- 

## Customer Impact

- Narrator is announcing the Time when the focus lands to the date picker control , but visually time is not present.

## Regression? 

- Yes 

### Before

image

## Test methodology 

- Manual

 

## Test environment(s) 

-

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObject.cs
@@ -57,7 +57,18 @@ public partial class DateTimePicker
             get
             {
                 string? baseValue = base.Value;
-                return !string.IsNullOrEmpty(baseValue) ? baseValue : this.GetOwnerText();
+                if (!string.IsNullOrEmpty(baseValue))
+                {
+                    return baseValue;
+                }
+
+                string? baseText = (Owner as DateTimePicker)?.GetBaseText();
+                if (!string.IsNullOrEmpty(baseText))
+                {
+                    return baseText;
+                }
+
+                return "";
             }
         }
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.cs
@@ -819,6 +819,11 @@ public partial class DateTimePicker : Control
 
     internal override bool SupportsUiaProviders => true;
 
+    internal string GetBaseText()
+    {
+        return base.Text;
+    }
+
     /// <summary>
     ///  Overrides Text to allow for setting of the value via a string.  Also, returns
     ///  a formatted Value when getting the text.  The DateTime class will throw


src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObject.cs
```diff
@@ -57,7 +57,18 @@ public override string Value
...
             {
                 string? baseValue = base.Value;
-                return !string.IsNullOrEmpty(baseValue) ? baseValue : this.GetOwnerText();
+                if (!string.IsNullOrEmpty(baseValue))
+                {
+                    return baseValue;
+                }
+
+                string? baseText = (Owner as DateTimePicker)?.GetBaseText();
+                if (!string.IsNullOrEmpty(baseText))
```
Tanya-Solyanik, line 66: This could be simplified as a conditional expression.

src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.cs
```diff
@@ -819,6 +819,11 @@ public bool ShowUpDown
 
     internal override bool SupportsUiaProviders => true;
 
+    internal string GetBaseText()
```
Tanya-Solyanik, line 822: This method is accessed from a nested class, nested classes have access to private members. https://devblogs.microsoft.com/oldnewthing/?p=30273

Tanya-Solyanik (commented):
Looks good!

src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObject.cs
```diff
@@ -57,7 +57,18 @@ public override string Value
             get
             {
                 string? baseValue = base.Value;
-                return !string.IsNullOrEmpty(baseValue) ? baseValue : this.GetOwnerText();
+                if (!string.IsNullOrEmpty(baseValue))
+                {
+                    return baseValue;
+                }
+
+                string? baseText = (Owner as DateTimePicker)?.GetBaseText();
```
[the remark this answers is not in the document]
dmitrii-drobotov: Created https://github.com/dotnet/winforms/issues/9281

src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.cs
```diff
@@ -819,6 +819,11 @@ public bool ShowUpDown
 
     internal override bool SupportsUiaProviders => true;
 
+    private string GetBaseText()
```
dmitrii-drobotov, line 822: @Epica3055 perhaps we should instead change DateTimePicker.Text to be following:
```
get => !string.IsNullOrEmpty(CustomFormat) ? Value.ToString(CustomFormat) : base.Text;
```

In .NET 6 and before, DateTimePicker.Text value matched displayed value:

image

In .NET 7+ (and with this fix as well) it will still be valid only for DateTimePicker with custom format:
image

So I suggest fixing it in DateTimePicker.Text instead of only in AccessibleObject.

Epica3055: At first, I thought this to be a problem too. But while I investigated this [9253](https://github.com/dotnet/winforms/pull/9253) bug, I back traced the code and I found that this bug was caused by a commit which was to fix a bug [6637](https://github.com/dotnet/winforms/pull/6637).

So , according to this page https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.datetimepicker.text?view=windowsdesktop-6.0, I think we should keep it as it is. what do you think?

dmitrii-drobotov: From docs:
>The string returned by this property is equivalent to the Value property with the appropriate formatting or custom formatting applied. 

I interpret it as `Text` value should be adjusted according to both `Format` and `CustomFormat` values, not just `CustomFormat`, like right now.

In #6590 it was assumed that Text is always an empty string, but it's only an empty string when DateTimePicker handle is not created, for example when breakpoint is set right after form load or in form constructor. It executes this code:
https://github.com/dotnet/winforms/blob/6a32b724f3eac5649b942c131ed09a009c87ca55/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs#L3934-L3945

After handle is created, it uses `User32.GetWindowText` which returns correct value for any `Format` or `CustomFormat`.

So we can actually return just `base.Text;` without `Value.ToString(CustomFormat)`, as it was before https://github.com/dotnet/winforms/pull/6637.

In order to fix unit tests from #6637, we need to add `dateTimePicker.CreateControl();` so that handle is created. Also in some tests like this one:
https://github.com/dotnet/winforms/blob/84a94ae7af15c78ca04353e73b1916fc22c5f07d/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DateTimePickerTests.cs#L264-L274
We need to add `dateTimePicker.Format = DateTimePickerFormat.Custom;` to fix it. Setting `CustomFormat` without `DateTimePickerFormat.Custom` doesn't have any effect on an actual control. For example this is a DTP with Long format and `CustomFormat = dateTimeFormat.ShortTimePattern`, the `Text` value doesn't look right:
image

So returning `base.Text` in `DateTimePicker.Text` seems like a correct way, and I think it's acceptable to return empty string until handle is created.

Tanya-Solyanik: Great investigations! 
```
The string returned by this property is equivalent to the Value property with the appropriate formatting or custom formatting applied.

I interpret it as Text value should be adjusted according to both Format and CustomFormat values, not just CustomFormat, like right now.
```

I agree with Dmitrii's assessment. GetWindowText is how we get value for the Text property. 

Do we have a good coverage for DTP controls in the interactive test app? If not, please add missing Formats.

Epica3055: @dmitrii-drobotov @Tanya-Solyanik Thanks for your investigations and explanations.
We just have one format in the interactive test app. I will add the rest formats.

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/DateTimePicker.Designer.cs
```diff
@@ -33,28 +35,77 @@ protected override void Dispose(bool disposing)
...
+        this.dateTimePicker4.TabIndex = 6;
+        // 
+        // dateTimePicker5
+        // 
+        this.dateTimePicker5.Format = DateTimePickerFormat.Custom;
+        this.dateTimePicker5.Location = new Point(31, 221);
+        this.dateTimePicker5.Margin = new Padding(4, 3, 4, 3);
+        this.dateTimePicker5.Name = "dateTimePicker5";
+        this.dateTimePicker5.Size = new Size(233, 23);
+        this.dateTimePicker5.TabIndex = 6;
```
dmitrii-drobotov, line 86: It's a nit, but I noticed tab indices are the same for every DTP, it would be better if first one had index 1, second index 2, etc.

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/DateTimePicker.Designer.cs
```diff
@@ -33,28 +35,77 @@ protected override void Dispose(bool disposing)
...
-        this.Name = "Form1";
+        // 
+        // dateTimePicker2
+        // 
+        this.dateTimePicker2.Format = DateTimePickerFormat.Long;
+        this.dateTimePicker2.Location = new Point(31, 78);
+        this.dateTimePicker2.Margin = new Padding(4, 3, 4, 3);
+        this.dateTimePicker2.Name = "dateTimePicker2";
+        this.dateTimePicker2.Size = new Size(233, 23);
+        this.dateTimePicker2.TabIndex = 6;
```
dmitrii-drobotov, line 59: Let's also set `ShowCheckBox = true;` to one of the DTPs and `ShowUpDown = true;` to another.

dmitrii-drobotov (approved):
Looks good, thank you!

Tanya-Solyanik (approved):

Issue #9256, opened 2023-06-08
[Accessibility] Narrator cannot announce the indeterminate state of DataGridViewCheckBoxCell correctly


v-elnovikova:
### .NET version

.NET SDK 8.0.100-preview.4.23260.5

### Did it work in .NET Framework?

No

### Did it work in any of the earlier releases of .NET Core or .NET 5+?

No

### Issue description

When set DataGridViewCheckBoxCell  value to Indeterminate, Narrator cannot announce that state is Indeterminate, but "Unchecked".

![DataGridViewCeckBoxCell - indeterminate announced as unchecked in 8 0](https://github.com/dotnet/winforms/assets/113603457/bb8e6551-8967-43ab-994e-022ca9b914ad)

![DataGridViewCeckBoxCell - indeterminate is Off in 8 0](https://github.com/dotnet/winforms/assets/113603457/9d124cd1-5a86-49d4-994d-46b933d0ec81)

### Steps to reproduce

1. Create a Winforms .Net Core application with DataGridView.
2. Add a DataGridViewCheckBoxColumn to DataGridView.
3. Build and run application.
4. Launch Narrator to focus on a grid check box cell.

Sample app:  [WinFormsApp1.zip](https://github.com/dotnet/winforms/files/11688232/WinFormsApp1.zip)

Amy-Li03:
Verified this on latest .NET 8.0.100-preview.7.23362.29, issue was fixed, Narrator can announce the indeterminate state of DataGridViewCheckBoxCell.
![image](https://github.com/dotnet/winforms/assets/62929087/de9465db-4c27-4479-a724-5f163658667b)
![image](https://github.com/dotnet/winforms/assets/62929087/aa58b8a5-5c08-48dc-b4de-9b0f8380fa05)

Olina-Zhang:
Verified with .NET SDK 8.0.100-preview.7.23376.3 test pass build, this issue was fixed with above same verification result.

PR #9274, opened 2023-06-12
fix Datagridcheckbox Accessibility checkstate indeterminate(#9256)

Epica3055:
Fixes # https://github.com/dotnet/winforms/issues/9256

## Proposed changes

- The issue is reproduced because we don't have support for indeterminate  state in accessibility in the .NET Core.
- add indeterminate state for DataGridCheckBoxCell
- add some logic for DataGridCheckBoxCell and DataGridCheckBoxCellAccessibilityObject

## Regression? 

- No

### Before

image

### After

image

## Test methodology 

- Manual
 

## Test environment(s) 

.NET 8.0.0-preview6.23308.17

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9274)


src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCheckBoxCell.DataGridViewCheckBoxCellAccessibleObject.cs
```diff
@@ -162,6 +162,11 @@ internal override UiaCore.ToggleState ToggleState
                 switch (Owner.FormattedValue)
                 {
                     case CheckState checkState:
+                        if (checkState == CheckState.Indeterminate)
+                        {
+                            return UiaCore.ToggleState.Indeterminate;
+                        }
+
                         toggledState = checkState == CheckState.Checked;
```
Tanya-Solyanik, line 170: Suggestion: code would be more consistent if you return value from each case. You could use https://learn.microsoft.com/dotnet/csharp/language-reference/statements/selection-statements#case-guards

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCheckBoxCell.cs
```diff
@@ -1015,29 +1015,47 @@ protected override void OnMouseUp(DataGridViewCellMouseEventArgs e)
...
-                isCheckboxChecked = boolValue;
+                checkState = boolValue ? CheckState.Checked : CheckState.Unchecked;
                 break;
         }
 
         if (IsParentAccessibilityObjectCreated)
         {
             var cellName = AccessibilityObject.Name ?? string.Empty;
+            string notificationText = string.Empty;
+            if (checkState == CheckState.Checked)
```
Tanya-Solyanik, line 1044: A switch expression will be suitable here - https://learn.microsoft.com/dotnet/csharp/language-reference/operators/switch-expression

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCheckBoxCell.cs
```diff
@@ -1015,29 +1015,47 @@ protected override void OnMouseUp(DataGridViewCellMouseEventArgs e)
 
     private void NotifyUiaClient()
     {
-        bool isCheckboxChecked = false;
+        CheckState checkState = CheckState.Indeterminate;
         switch (EditingCellFormattedValue)
         {
             case string stringValue:
-                isCheckboxChecked = stringValue == SR.DataGridViewCheckBoxCell_ClipboardChecked;
+                if (stringValue == SR.DataGridViewCheckBoxCell_ClipboardChecked)
```
Tanya-Solyanik, line 1022: Would `when` guard work here?

Tanya-Solyanik: Additionally, both strings in this comparison are localized, but `==` performs ordinary comparison - https://learn.microsoft.com/dotnet/csharp/how-to/compare-strings#default-ordinal-comparisons, please use Compare function and pass in the current UI culture.

Tanya-Solyanik (commented):
Looks good!

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCheckBoxCell.DataGridViewCheckBoxCellAccessibleObject.cs
Epica3055: great suggestion~

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCheckBoxCell.cs
Epica3055: great suggestion~

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCheckBoxCell.cs
Epica3055: great suggestion~

dmitrii-drobotov (commented):
Looks good! Could you also add a `DataGridViewCheckBoxColumn` with `ThreeState = true` to DataGridView test form in WinformsControlsTest app?

Tanya-Solyanik:
@Epica3055  - also please add unit test

Epica3055:
> Looks good! Could you also add a `DataGridViewCheckBoxColumn` with `ThreeState = true` to DataGridView test form in WinformsControlsTest app?

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCheckBoxCell.cs
```diff
@@ -398,6 +398,23 @@ internal bool ThreeStateInternal
         }
     }
 
+    internal CheckState CheckState
+    {
+        get
+        {
+            if ((flags & DATAGRIDVIEWCHECKBOXCELL_checked) != 0x00)
```
Tanya-Solyanik, line 405: We don't usually use HEX numbers in this context, only when positions have a meaning, for example in the HRESULT values.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCheckBoxCell.cs
```diff
@@ -398,6 +398,23 @@ internal bool ThreeStateInternal
...
+            {
+                return CheckState.Checked;
+            }
+            else if ((flags & DATAGRIDVIEWCHECKBOXCELL_indeterminate) != 0x00)
```
Tanya-Solyanik, line 409:
```suggestion
            if ((flags & DATAGRIDVIEWCHECKBOXCELL_indeterminate) != 0)
```

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCheckBoxCell.cs
Tanya-Solyanik: Much better than string comparison! Are `flags` always up-to-date, even before the editing box is committed? I'm wondering why the previous code was looking at the string value

Epica3055: > Much better than string comparison! Are `flags` always up-to-date, even before the editing box is committed? I'm wondering why the previous code was looking at the string value

I think they are always up-to-date. I tested it on my local machine. Even before the editing box is committed, Inspect shows correct value;

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewCheckBoxCellAccessibleObjectTests.cs
```diff
@@ -242,4 +242,26 @@ public void DataGridViewCheckBoxCellAccessibleObject_DoDefaultAction_WorksExpect
         Assert.Equal((UiaCore.ToggleState)expected, accessibleObject.ToggleState);
         Assert.Equal(createControl, control.IsHandleCreated);
     }
+
+    [WinFormsFact]
+    public void DataGridViewCheckBoxCellAccessibleObject_ToggleStateIndeterminate_ReturnsExpected()
+    {
+        using DataGridView control = new();
+        var checkBoxColumn = new DataGridViewCheckBoxColumn();
```
Tanya-Solyanik, line 250: Columns are `IDisposable`, please add `using`.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewCheckBoxCellAccessibleObjectTests.cs
```diff
@@ -242,4 +242,26 @@ public void DataGridViewCheckBoxCellAccessibleObject_DoDefaultAction_WorksExpect
...
+        checkBoxColumn.ThreeState = true;
+        control.Columns.Add(checkBoxColumn);
+        control.Rows.Add(new DataGridViewRow());
```
Tanya-Solyanik, line 253: DataGridViewRow instances are `IDisposable` too.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewCheckBoxCellAccessibleObjectTests.cs
```diff
@@ -242,4 +242,26 @@ public void DataGridViewCheckBoxCellAccessibleObject_DoDefaultAction_WorksExpect
...
+
+        control.CreateControl();
+
+        var checkBoxAccessibleObject = cell.AccessibilityObject;
+
+        Assert.True(checkBoxAccessibleObject.ToggleState == UiaCore.ToggleState.Indeterminate);
+        checkBoxAccessibleObject.DoDefaultAction();
+        Assert.True(checkBoxAccessibleObject.ToggleState == UiaCore.ToggleState.Off);
+        checkBoxAccessibleObject.DoDefaultAction();
+        Assert.True(checkBoxAccessibleObject.ToggleState == UiaCore.ToggleState.On);
```
Tanya-Solyanik, line 264: Consider a second round of default actions to ensure that repeated action yields the same result.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCheckBoxCell.cs
```diff
@@ -398,6 +398,23 @@ internal bool ThreeStateInternal
         }
     }
 
+    internal CheckState CheckState
```
Tanya-Solyanik, line 401:
```suggestion
    private CheckState CheckState
```

Private properties are accessible from the nested classes.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewCheckBoxCellAccessibleObjectTests.cs
```diff
@@ -242,4 +242,26 @@ public void DataGridViewCheckBoxCellAccessibleObject_DoDefaultAction_WorksExpect
...
+    }
+
```
Tanya-Solyanik, line 266: Please remove this blank line.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewCheckBoxCellAccessibleObjectTests.cs
```diff
@@ -242,4 +242,26 @@ public void DataGridViewCheckBoxCellAccessibleObject_DoDefaultAction_WorksExpect
         Assert.Equal((UiaCore.ToggleState)expected, accessibleObject.ToggleState);
         Assert.Equal(createControl, control.IsHandleCreated);
     }
+
+    [WinFormsFact]
+    public void DataGridViewCheckBoxCellAccessibleObject_ToggleStateIndeterminate_ReturnsExpected()
```
Tanya-Solyanik, line 247: Please reference bug number that this test is verifying, for example - https://github.com/dotnet/winforms/blob/6f001b029f4010b7147d8e3dbe5a67d22a4832c3/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/FolderBrowserDialogTests.cs#L16

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCheckBoxCell.cs
```diff
@@ -398,6 +398,20 @@ internal bool ThreeStateInternal
         }
     }
 
+    private CheckState CheckState
+    {
+        get
+        {
+            if ((!EditingCellValueChanged && FormattedValue is CheckState checkState && checkState == CheckState.Indeterminate) ||
```
dmitrii-drobotov, line 405: Could you clarify in which situation `flags` don't have a correct value and there is a need to use `FormattedValue`?

Tanya-Solyanik:
@Epica3055 - please move this PR from the draft state to `ready for review` because it's almost complete.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCheckBoxCell.cs
Epica3055: after the form is rendered before the value is changed

dmitrii-drobotov: I see, in that case there is probably the same problem with `DATAGRIDVIEWCHECKBOXCELL_checked`. Can you look into a way to ensure that `flags` always have correct value, even when value was not edited? Perhaps override `DataGridViewCell.SetValue` and do the same thing as here:
https://github.com/dotnet/winforms/blob/09df23d0e5e6dc3fe15253d229c84144803dcc1a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCheckBoxCell.cs#L79-L109
If it's not possible, we might need to use `FormattedValue` here for all cases.

dmitrii-drobotov (commented):
Could you also test with custom values in `DataGridViewCheckBoxCell.TrueValue`, `DataGridViewCheckBoxCell.FalseValue`, `DataGridViewCheckBoxCell.IndeterminateValue`, and make sure it still sets correct ToggleState?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewCheckBoxCellAccessibleObjectTests.cs
```diff
@@ -242,4 +242,32 @@ public void DataGridViewCheckBoxCellAccessibleObject_DoDefaultAction_WorksExpect
         Assert.Equal((UiaCore.ToggleState)expected, accessibleObject.ToggleState);
         Assert.Equal(createControl, control.IsHandleCreated);
     }
+
+    //unit test for https://github.com/dotnet/winforms/issues/9256
```
Tanya-Solyanik, line 246:
```suggestion
    // Unit test for https://github.com/dotnet/winforms/issues/9256
```

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCheckBoxCell.DataGridViewCheckBoxCellAccessibleObject.cs
```diff
@@ -158,20 +158,12 @@ internal override UiaCore.ToggleState ToggleState
                     throw new InvalidOperationException(SR.DataGridViewCellAccessibleObject_OwnerNotSet);
                 }
 
-                bool toggledState;
-                switch (Owner.FormattedValue)
+                return ((Owner as DataGridViewCheckBoxCell)?.CheckState) switch
```
Tanya-Solyanik, line 161: If the owner is not a DataGridViewCheckBoxCell, then this pattern should not be invoked. Please double check that other cell types do not report this pattern as supported. If this is the case, you can change condition on line 156 to exclude "wrong types" of cells. `if (Owner is not DataGripViewCheckBoxCell { } checkBoxCell) `

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCheckBoxCell.cs
Tanya-Solyanik: @Epica3055  - please add a comment that explains why we must look at the FormattedValue.

Tanya-Solyanik (commented):
Looks good, I added comments though.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCheckBoxCell.DataGridViewCheckBoxCellAccessibleObject.cs
```diff
@@ -153,25 +153,17 @@ internal override UiaCore.ToggleState ToggleState
         {
             get
             {
-                if (Owner is null)
+                if (Owner is not DataGridViewCheckBoxCell)
                 {
-                    throw new InvalidOperationException(SR.DataGridViewCellAccessibleObject_OwnerNotSet);
+                    throw new InvalidOperationException(SR.DataGridViewCellAccessibleObject_OwnerWrongType);
```
Tanya-Solyanik, line 158: Had you double-checked that for other cell types, this pattern is not enabled?

Tanya-Solyanik: Had you verified if `DataGridViewCellAccessibleObject_OwnerNotSet` resource is still used anywhere? I.e. should it be removed?

src/System.Windows.Forms/src/Resources/SR.resx
```diff
@@ -2271,6 +2271,9 @@ To replace this default dialog please handle the DataError event.</value>
   <data name="DataGridViewCellAccessibleObject_OwnerNotSet" xml:space="preserve">
     <value>The owning DataGridViewCell object has not been set.</value>
   </data>
+  <data name="DataGridViewCellAccessibleObject_OwnerWrongType" xml:space="preserve">
+    <value>The owning DataGridViewCell object is a wrong type.</value>
```
Tanya-Solyanik, line 2275: Please provide the actual type of this cell - 
```
DataGridViewCell of type {0} does not provide the Toggle pattern. 
```

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCheckBoxCell.DataGridViewCheckBoxCellAccessibleObject.cs
Epica3055: > Had you double-checked that for other cell types, this pattern is not enabled?

Yes, I searched our code, The only constructor of `DataGridViewCheckBoxCellAccessibleObject` ,which is `public DataGridViewCheckBoxCellAccessibleObject(DataGridViewCell? owner)` , is invoked in several places, where either null or DataGridViewCheckBoxCell is passed . 

> Had you verified if DataGridViewCellAccessibleObject_OwnerNotSet resource is still used anywhere? I.e. should it be removed?

`DataGridViewCellAccessibleObject_OwnerNotSet ` is still used in several places.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataGridViewCheckBoxCellAccessibleObjectTests.cs
```diff
@@ -242,4 +242,32 @@ public void DataGridViewCheckBoxCellAccessibleObject_DoDefaultAction_WorksExpect
...
+    // Unit test for https://github.com/dotnet/winforms/issues/9256
+    [WinFormsFact]
+    public void DataGridViewCheckBoxCellAccessibleObject_ToggleStateIndeterminate_ReturnsExpected()
+    {
+        using DataGridView control = new();
+        using DataGridViewCheckBoxColumn checkBoxColumn = new();
+        checkBoxColumn.ThreeState = true;
+        control.Columns.Add(checkBoxColumn);
+        using DataGridViewRow row = new();
+        control.Rows.Add(row);
```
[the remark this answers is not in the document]
dmitrii-drobotov: Also these cases:
```
checkBoxColumn.TrueValue = "on";
checkBoxColumn.FalseValue = "off";
checkBoxColumn.IndeterminateValue = "not set";

control.Rows.Add("on");
control.Rows.Add("off");
control.Rows.Add("not set");
```

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCheckBoxCell.cs
```diff
@@ -398,6 +398,31 @@ internal bool ThreeStateInternal
...
+    private CheckState CheckState
+    {
+        get
+        {
+            // when the CheckBoxCell has not been focused, the flags is not update to date
+            // so in this case, we use EditingCellValueChanged && FormattedValue to determine CheckState
+            // when users use mouse to change the CheckBoxCell's CheckState but don't commit the value
+            // FormattedValue is not updated, but flag is updated
+            // so in this case, we use flags to determine CheckState.
+            if ((!EditingCellValueChanged && FormattedValue is CheckState checkState && checkState == CheckState.Indeterminate) ||
```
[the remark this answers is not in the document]
Epica3055: yeah, but when EditingCellValueChanged is true, I think I need to write additional code to determine if it is Indeterminate or Checked or Unchecked. So I think maybe we should just keep it.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCheckBoxCell.cs
```diff
@@ -398,6 +398,36 @@ internal bool ThreeStateInternal
...
+    private CheckState CheckState
+    {
+        get
+        {
+            // When the CheckBoxCell has not been focused, the flags variable is not up to date,
+            // so in this case, we use EditingCellValueChanged && FormattedValue to determine CheckState.
+            // When users change the CheckBoxCell's CheckState but don't commit the value,
+            // FormattedValue is not updated, but flags are, so in this case, we use flags to determine CheckState.
+            if ((!EditingCellValueChanged && FormattedValue is CheckState checkState && checkState == CheckState.Indeterminate) ||
+                (flags & DATAGRIDVIEWCHECKBOXCELL_indeterminate) != 0)
```
Tanya-Solyanik, line 410: If flags field is up-to-date only when the cell is focused, why are you not testing for that condition?
something like `owner.Focused && flags & DATADRID....`

Epica3055: Yeah, I have written that in Unit Tests

Tanya-Solyanik: If flags are valid only while the value is not committed, should the condition look like this:
```
if ((!EditingCellValueChanged && FormattedValue is CheckState checkState && checkState == CheckState.Indeterminate) || 
(EditingCellValueChanged &&  (flags & DATAGRIDVIEWCHECKBOXCELL_indeterminate) != 0))
```

Epica3055: According to my investigation, flags are valid after the cell has been focused.

Tanya-Solyanik: How do we know that the cell has been focused at least once when we test the flag value.

Epica3055: I think we don't have to test the flag value, we just return correct CheckState. Or maybe I don't fully understand your words.

According to what I observed: 
when the cell hasn't been focused the flags variable is not up to date, that's why I wrote this code 
`(!EditingCellValueChanged && FormattedValue is CheckState checkState && checkState == CheckState.Indeterminate)` 
When the cell has been focused the flags variable is always up to date , so I think we can use flag to determine CheckState , that's why I wrote this code 
 `flags & DATAGRIDVIEWCHECKBOXCELL_indeterminate) != 0`

The same with `CheckState.Checked`

Tanya-Solyanik: I see, before the cell is clicked, flags field is empty.

Tanya-Solyanik (approved):
Thank you!

Tanya-Solyanik (approved):

PR #9278, opened 2023-06-12
Fully enable ComWrappers for COM TypeDescriptor

JeremyKuhne:
Uses ComHelpers to query for interface pointers to enable all COM scenarios (ComWrappers, Com Interop, our own Com exposed classes).

- Change Com2DataTypeToManagedDataTypeConverter to go between VARIANT and object.
- Remove Activator.CreateInstance call for type converters from Com2PropertyDescriptor and just create the instances directly as needed.
- Add a NullTypeDescriptor to simplify the ComTypeDescriptor.
- Add reverse IUnknown* cast operator to VARIANT.
- Ensure we don't return null from BSTR.ToStringAndFree and only free if it isn't null.
- Change RegisterInterfaceInGlobal to return HRESULT for easier debugging.
- Use centralized IDispatch property gettter/setter helpers and ensure we free any returned exception BSTRs.
- Add IVTable to additional COM structs.
- Add tests to cover both legacy COM and ComWrappers and also validate IProvideMulitpleClassInfo
- Move IDispatchTests to the right namespace.
- Seal classes, revise variable names, split out COM struct files, and other various cleanup.


JeremyKuhne:
ComWrappers are supported on other platforms (Unix/Mac) so there is one more follow up here where I'll try to remove Windows P/Invokes that are strictly unnecessary. Notably the call to get the current thread's LCID.

src/System.Windows.Forms.Primitives/src/Windows/Win32/System/Com/IDispatch.cs
```diff
@@ -70,6 +109,29 @@ public HRESULT SetPropertyValue(int dispatchId, VARIANT value)
...
+                      Description: {pExcepInfo.bstrDescription.ToStringAndFree()}
+                      Source: {pExcepInfo.bstrSource.ToStringAndFree()}
+                      Help file: {pExcepInfo.bstrHelpFile.ToStringAndFree()}
+                    """);
+#else
+                pExcepInfo.bstrDescription.Dispose();
+                pExcepInfo.bstrSource.Dispose();
+                pExcepInfo.bstrHelpFile.Dispose();
+#endif
+        }
```
Shyam-Gupta, line 133: nit: These lines can probably be moved to a private method to avoid code duplication.

JeremyKuhne: I didn't yet as one is getting the error string and the other isn't. I want to do a pass to ensure everything is going through here and that the usages / error reporting is consistent.

src/System.Windows.Forms/src/System/Windows/Forms/ComponentModel/COM2Interop/COM2PropertyBuilderUITypeEditor.cs
```diff
@@ -45,38 +44,34 @@ internal class Com2PropertyBuilderUITypeEditor : Com2ExtendedUITypeEditor
...
 
-            Debug.Assert(obj is IProvidePropertyBuilder.Interface, "object is not IProvidePropertyBuilder");
-            IProvidePropertyBuilder.Interface propBuilder = (IProvidePropertyBuilder.Interface)obj;
+        using var propertyBuilder = ComHelpers.TryGetComScope<IProvidePropertyBuilder>(target, out HRESULT hr);
+        Debug.Assert(hr.Succeeded, $"Failed to get IProvidePropertyBuilder: {hr}");
 
-            // Is it actually necessary to pass the value in?
-            Marshal.GetNativeVariantForObject(value, (nint)(void*)&variantValue);
+        // Is it actually necessary to pass the value in?
+        variantValue = VARIANT.FromObject(value);
```
Shyam-Gupta, line 57: nit: We can declare `variantValue` here itself.

src/System.Windows.Forms/src/System/Windows/Forms/ComponentModel/COM2Interop/ComNativeDescriptor.cs
```diff
@@ -385,7 +323,7 @@ private void OnPropsInfoDisposed(object? sender, EventArgs e)
     }
 
     /// <summary>
-    ///  Looks at at value's type and creates an editor based on that.  We use this to decide which editor to use
+    ///  Looks at at value's type and creates an editor based on that. We use this to decide which editor to use
```
Shyam-Gupta, line 326: nit nit: One 'at' can be removed from comment.

src/System.Windows.Forms/src/System/Windows/Forms/ComponentModel/COM2Interop/COM2ColorConverter.cs
```diff
@@ -3,34 +3,26 @@
...
-
-        return ColorTranslator.FromOle(intVal);
-    }
+            VARENUM.VT_UI4 or VARENUM.VT_UINT => (int)(uint)nativeValue,
+            VARENUM.VT_I4 or VARENUM.VT_INT => (int)nativeValue,
+            _ => 0
+        });
 
-    public override object ConvertManagedToNative(object? managedValue, Com2PropertyDescriptor pd, ref bool cancelSet)
+    public override VARIANT ConvertManagedToNative(object? managedValue, Com2PropertyDescriptor propert, ref bool cancelSet)
```
Shyam-Gupta, line 25: nit: typo

Shyam-Gupta (approved):
Looks great!

Shyam-Gupta (approved):

PR #9297, opened 2023-06-15
Set foreground

sharwell:
Ensure a test window has focus before calling `SendInput`. If the window does not have focus, the test will first try to set focus using the [`SwitchToThisWindow`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-switchtothiswindow), [`SetActiveWindow`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setactivewindow), and [`SetFocus`](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setfocus) APIs, and if that fails it will attempt to click the visible title bar with the mouse. If the test fails to establish focus, it will fail.

In an earlier commit in this PR, the code would attempt to use the console to set focus (a practice that works very well for Visual Studio integration testing). However, this strategy failed in the `AllocConsole` method for unknown reasons when applied to winforms testing.

Merge commit is recommended to preserve history.

### DIFF
--- a/eng/pipelines/build.yml
+++ b/eng/pipelines/build.yml
@@ -118,8 +118,7 @@ jobs:
       env:
         XUNIT_LOGS: $(BUILD.SOURCESDIRECTORY)\artifacts\log\$(_BuildConfig)
       displayName: Run Integration Tests
-      # UI tests are blocked from execution until we find solution for https://github.com/dotnet/winforms/issues/9318.
-      condition: and(eq(variables['System.TeamProject'], 'public'), ne('${{ parameters.targetArchitecture }}', 'arm64'), contains(variables['System.PullRequest.TargetBranch'], 'main'), false)
+      condition: eq(variables['System.TeamProject'], 'public')
       retryCountOnTaskFailure: 3
 
     # Create Nuget package, sign, and publish
--- a/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/SendInput.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/SendInput.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Drawing;
 using System.Runtime.InteropServices;
 using System.Windows.Forms.UITests.Input;
 using Windows.Win32.UI.Input.KeyboardAndMouse;
@@ -69,24 +70,27 @@ public class SendInput
         });
     }
 
-    internal async Task SendAsync(Form window, Action<InputSimulator> actions)
+    internal async Task SendAsync(Form? window, Action<InputSimulator> actions)
     {
         if (actions is null)
         {
             throw new ArgumentNullException(nameof(actions));
         }
 
-        IHandle<HWND> handle = window;
-        if (!TrySetForegroundWindow(handle.Handle))
+        if (window is not null)
         {
-            throw new InvalidOperationException("Failed to set the foreground window.");
+            IHandle<HWND> handle = window;
+            if (!await TrySetForegroundWindowAsync(window, handle.Handle))
+            {
+                throw new InvalidOperationException("Failed to set the foreground window.");
+            }
+
+            // Ensure the window is 'Active' as it may not have been achieved by 'SetForegroundWindow'
+            PInvoke.SetActiveWindow(window);
+
+            // Give the window the keyboard focus as it may not have been achieved by 'SetActiveWindow'
+            PInvoke.SetFocus(window);
         }
-
-        // Ensure the window is 'Active' as it may not have been achieved by 'SetForegroundWindow'
-        PInvoke.SetActiveWindow(window);
-
-        // Give the window the keyboard focus as it may not have been achieved by 'SetActiveWindow'
-        PInvoke.SetFocus(window);
 
         await Task.Run(() => actions(new InputSimulator()));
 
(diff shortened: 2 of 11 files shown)


dreddy-work:
Hello @sharwell, I see you opened an issue. Can you help us understand the changes here and the remaining scope?

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/SendInput.cs
```diff
@@ -98,6 +113,37 @@ private static HWND GetForegroundWindow()
         return foregroundWindow;
     }
 
+    private async Task<bool> TrySetForegroundWindowAsync(Form? form, HWND window)
+    {
+        var activeWindow = PInvoke.GetLastActivePopup(window);
+        activeWindow = PInvoke.IsWindowVisible(activeWindow) ? activeWindow : window;
+        PInvoke.SwitchToThisWindow(activeWindow, true);
```
Tanya-Solyanik, line 120:
```suggestion
        PInvoke.SwitchToThisWindow(activeWindow, fUnknown: true);
```

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/SendInput.cs
```diff
@@ -98,6 +113,37 @@ private static HWND GetForegroundWindow()
...
+
+        if (form is not null && await TrySetForegroundWindowViaMouseAsync(form, activeWindow))
+            return true;
+
+        return false;
+    }
+
+    private async Task<bool> TrySetForegroundWindowViaMouseAsync(Form form, HWND activeWindow)
+    {
+        // If that fails, try setting focus by clicking the title bar of the form
```
Tanya-Solyanik, line 133:
```suggestion
        // If that fails, try setting focus by clicking the title bar of the form.
```

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/SendInput.cs
```diff
@@ -98,6 +113,37 @@ private static HWND GetForegroundWindow()
...
+        var rect = form.DisplayRectangle;
+        var positionOnTitleBar = new Point(GetMiddle(rect.Right, rect.Left), rect.Top + 5);
```
Tanya-Solyanik, line 135:
```suggestion
        Point positionOnTitleBar = new (GetMiddle(rect.Right, rect.Left), rect.Top + 5);
```

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/SendInput.cs
```diff
@@ -98,6 +113,37 @@ private static HWND GetForegroundWindow()
...
+        if (form is not null && await TrySetForegroundWindowViaMouseAsync(form, activeWindow))
+            return true;
+
+        return false;
+    }
+
+    private async Task<bool> TrySetForegroundWindowViaMouseAsync(Form form, HWND activeWindow)
+    {
+        // If that fails, try setting focus by clicking the title bar of the form
+        var rect = form.DisplayRectangle;
```
Tanya-Solyanik, line 134:
```suggestion
        Rectangle rect = form.DisplayRectangle;
```

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/SendInput.cs
```diff
@@ -98,6 +113,37 @@ private static HWND GetForegroundWindow()
...
+        var positionOnTitleBar = new Point(GetMiddle(rect.Right, rect.Left), rect.Top + 5);
+        var positionOnScreen = form.PointToScreen(positionOnTitleBar);
```
Tanya-Solyanik, line 136:
```suggestion
        Point positionOnScreen = form.PointToScreen(positionOnTitleBar);
```

eng/pipelines/build.yml
```diff
@@ -118,8 +118,7 @@ jobs:
       env:
         XUNIT_LOGS: $(BUILD.SOURCESDIRECTORY)\artifacts\log\$(_BuildConfig)
       displayName: Run Integration Tests
-      # UI tests are blocked from execution until we find solution for https://github.com/dotnet/winforms/issues/9318.
-      condition: and(eq(variables['System.TeamProject'], 'public'), ne('${{ parameters.targetArchitecture }}', 'arm64'), contains(variables['System.PullRequest.TargetBranch'], 'main'), false)
+      condition: eq(variables['System.TeamProject'], 'public')
```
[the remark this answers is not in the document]
sharwell: If we do need to disable things, I'm hoping to do so by leaving the pipeline enabled but blocking the failure from being applied to the result. That way we always have running logs over time of teh overall success rate.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/SendInput.cs
sharwell: ➡️ This parameter name didn't seem helpful so I clarified with a comment.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/SendInput.cs
sharwell: I would prefer to leave the rest of the code style suggestions for future cleanup in combination with enabling an analyzer instead of using manual code review as a code style checking tool.

dreddy-work:
Closing the draft PR. Revisit, the strategy when enabling integration tests in 9.0

PR #9309, opened 2023-06-19
Convert more User32 code to CSWin32

JeremyKuhne:
The converts another block of User32 interop to CSWin32.

We had created enums for messages and style flags that were specific to controls. I've decide to not recreate them as it has started to obfuscate that the values overlap with other values.

I'm still giving thought to what we do with WM_. It was convenient for debugging to have an enum, but I think we can solve that in other ways.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9309)


src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextRange.cs
```diff
@@ -577,12 +576,12 @@ private static bool AtWordBoundary(string text, int index)
         return textAttributeIdentifier switch
         {
             TextAttributeIdentifier.BackgroundColorAttributeId => GetBackgroundColor(),
-            TextAttributeIdentifier.CapStyleAttributeId => GetCapStyle(_provider.EditStyle),
+            TextAttributeIdentifier.CapStyleAttributeId => GetCapStyle(_provider.WindowStyle),
```
dreddy-work, line 579: Q: was this a bug earlier? they both are pointing to different values.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextRange.cs
```diff
@@ -751,22 +750,23 @@ void AddLineRectangle(Point startPoint, Point endPoint)
         }
     }
 
-    private static HorizontalTextAlignment GetHorizontalTextAlignment(ES editStyle)
+    private static HorizontalTextAlignment GetHorizontalTextAlignment(WINDOW_STYLE windowStyle)
     {
-        if (editStyle.HasFlag(ES.CENTER))
+        if (((int)windowStyle & PInvoke.ES_CENTER) != 0)
```
dreddy-work, line 755: earlier logical operations were on the same type. May be linking a some description or bug ( if it was) would help understand better.

dreddy-work (commented):
RLGTM.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextRange.cs
JeremyKuhne: Not a bug, one was just casting WindowStyle to EditStyle. As mentioned, these overloads were obfuscating the mechanics.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextRange.cs
JeremyKuhne: It's not an enum now, it's a constant (ES_CENTER). Windows metadata has put the standard styles into an enum for convenience, but they aren't actually defined as such in the headers. They haven't put the other values into enums at this point. There isn't a lot of value for us to write additional code to put the special values into additional enums.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripTextBox.ToolStripTextBoxControl.cs
```diff
@@ -233,7 +233,7 @@ private void WmNCPaint(ref Message m)
 
             // Note that GetWindowDC just calls GetDCEx with DCX_WINDOW | DCX_USESTYLE.
 
-            using var hdc = new User32.GetDcScope(m.HWnd, IntPtr.Zero, User32.DCX.WINDOW | User32.DCX.USESTYLE);
+            using var hdc = new User32.GetDcScope(m.HWND, HRGN.Null, GET_DCX_FLAGS.DCX_WINDOW | (GET_DCX_FLAGS)0x00010000 /* DCX_USESTYLE */);
```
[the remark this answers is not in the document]
JeremyKuhne: It isn't available.

dreddy-work (approved):

PR #9331, opened 2023-06-21
Remove Interop.User32 class

JeremyKuhne:
This change fully removes the Interop.User32 class. This does this by:

- Removing WM
- Moving GetDCScope and RegisteredMessage to Windows.Win32

Using an enum had the advantage of seeing relatively easily what a given message was (presuming it was a standard WM_ message). To help a bit with this I've created a MessageId struct that allows seeing the message description in the debugger.

One nice benefit of this change is that we now have links to the docs for the given message defines, which will help development a lot.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9331)


JeremyKuhne:
Note that the test fix was for an existing issue that I saw here.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/MessageDecoder.cs
```diff
@@ -12,322 +10,23 @@ namespace System.Windows.Forms;
...
-
-    private static string ToString(IntPtr hWnd, User32.WM msg, WPARAM wparam, LPARAM lparam, LRESULT result)
+    public static string ToString(Message message) => ToString(
+        message.HWnd,
+        message.MsgInternal,
+        message.WParamInternal,
+        message.LParamInternal,
+        message.ResultInternal);
+
+    private static string ToString(IntPtr hWnd, MessageId msg, WPARAM wparam, LPARAM lparam, LRESULT result)
```
Tanya-Solyanik, line 20: Consider replacing `msg` with a full word

src/System.Windows.Forms.Primitives/src/Windows/Win32/MessageId.cs
```diff
@@ -0,0 +1,336 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Windows.Win32;
+
+/// <summary>
+///  Simple internal wrapper that enables showing the message identifier string in the debugger.
+/// </summary>
+internal readonly struct MessageId
```
Tanya-Solyanik, line 10: I read your PR description as if you added a `DebuggerDisplay` attribute. Is `ToString` called by default and it's text shown in the locals window?

Tanya-Solyanik (approved):

src/System.Windows.Forms.Primitives/src/Windows/Win32/MessageId.cs
JeremyKuhne: ToString is the default

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/MessageDecoder.cs
JeremyKuhne: I'll iterate on this in the next PR.

PR #9332, opened 2023-06-21
add project reference to System.Drawing.Common to integration tests

Tanya-Solyanik:
This will make debugging more convenient

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9332)

### DIFF
--- a/src/System.Windows.Forms/tests/AccessibilityTests/Accessibility_Core_App.csproj
+++ b/src/System.Windows.Forms/tests/AccessibilityTests/Accessibility_Core_App.csproj
@@ -11,6 +11,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\src\System.Windows.Forms.csproj" />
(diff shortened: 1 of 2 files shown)


dreddy-work:
Isn't this enforces to pick right version of drawing.common? Not just debugging source? As i see it, earlier these test projects were picking drawing.common from SDK and now from the solution?

dreddy-work (approved):

Tanya-Solyanik:
```
Isn't this enforces to pick right version of drawing.common? Not just debugging source? As i see it, earlier these test projects were picking drawing.common from SDK and now from the solution?
```

Sure, but I called out debugging with sources because we had minimal changes in System.Drawing.common, so version does not matter yet, and the changes that we had are not being exercised in these tests

PR #9349, opened 2023-06-23
Add Click Through behavior to `ToolStrip`/`MenuStrip`

### LINKED ISSUE #2544: Fix MenuStrip hover & click behavior in unfocused windows
(split off from #2476)

This issue addresses two differences between native MainMenu and MenuStrip controls when the window is unfocused.

### Issue 1 : Hovering menu items highlights them in an unfocused window
Native control on the right side only highlights hovered items when the window is focused. This also causes confusion with issue 2.

unfocused

### Issue 2 : Clicking a menu item in unfocused window does not open it
To reproduce, open the test project which opens 2 windows as seen on the screenshot above, click the taskbar or another window to unfocus them, and then click **Item 1** in each window. You will see that:

* In the **FormStripSystem** window which uses a MenuStrip, clicking highlights the window but does not open the menu. State after clicking:
focusedstrip

* In the **FormNative** window which uses a native MainMenu, clicking both highlights the window and immediately opens the menu. State after clicking:
focusednative

The behavior of MenuStrip should be changed to match the native control.

As I said in https://github.com/dotnet/winforms/issues/2476#issuecomment-567316912, I tested several applications with both native MainMenu and completely custom menu controls, all of which follow the native Windows behavior, so the non-standard behavior of MenuStrip is likely confusing to Windows users.

### Test Project

[winforms.zip](https://github.com/dotnet/winforms/files/3990977/winforms.zip)

### LINKED ISSUE #9240: ToolStripButton OnClicked handler is not called if be clicked when the app has no focus
### .NET version

.NET6 and .NET7

Yes

_No response_

### Issue description

My project is upgraded from .NET framework 47 to .NET6. `ToolBar/ToolBarButton` has to be replaced with `ToolStrip/ToolStripButton`.

image

I created a sample app to reproduce the problem, there are one `ToolStrip` (three `ToolStripButton` in it) and one `Button`.

If the app is active (has focus), click any button, the event handler for "Clicked" gets called as expected.
If the app is not active, click `ToolStripButton`, the event handler is **not** called, it just active the app, then user has to click again to trigger the event handler. But `Button` still works for this scenario.

In .NET framework, `ToolBarButton` can work if the app is not active. Is it a bug or as design? Is there any way to make `ToolStripButton` keep same behavior as `ToolBarButton` used to do.

### Steps to reproduce

See above

### LINKED ISSUE #9277: Can not raise Click event when mouse click a ToolStripButton in a inactive window
### .NET version

.Net 8.0 Preview 4

Not tested/verified

_No response_

### Issue description

I can not raise a Click event when I first time click a ToolStripButton in a inactive window, have to click twice to raise above event.

### Steps to reproduce

1. Create a WinForms project.
2. Create 2 form classes, and add a ToolStrip to each of above forms.
3. Create a ToolStripButton to each of above ToolStrips.

Issues:
1. Active first form, click ToolStripButton control of another form (inactive), ToolStripButton doesn't raise a Click event.
2. Have to click twice to raise above event.

### LINKED ISSUE #9288: Add `ClickThrough` to `ToolStrip`/`MenuStrip` to bring behavior in-line with other controls.
### Background and motivation

There have recently been a couple of issues asking for this feature. 
- #9240
- #9277

Outline explaining the feature from associated blog post:
> ToolStrip classes do not support "Click Through" between forms. For more information on what click through is, I recommend reading https://daringfireball.net/2003/05/interface_details_itunes_vs_safari . In Windows, pretty much all applications use "click through," whereby if you have one window active and click on to another one, the mouse click activates the window and the click is processed by whatever control you had the mouse over.

https://learn.microsoft.com/en-us/archive/blogs/rickbrew/how-to-enable-click-through-for-net-2-0-toolstrip-and-menustrip

### API Proposal

```csharp
public class ToolStrip
{
   public bool ClickThrough { get; set; }

   protected override void WndProc(ref Message m)
    {
        base.WndProc(ref m);
        if (ClickThrough &&
            m.Msg == NativeConstants.WM_MOUSEACTIVATE &&
            m.Result == (IntPtr)NativeConstants.MA_ACTIVATEANDEAT)
        {
            m.Result = (IntPtr)NativeConstants.MA_ACTIVATE;
        }

    }
}
```

### API Usage

Enable/Disable the `ClickThrough` property to change the behavior.

### Alternative Designs

default value of property set to `false`

### Risks

If default is true, then behavior could change in existing applications.

Yes. Toolstrip will allow click through by default or at least allow click through after the developer enables it.

Unsure on accessibility impact.

Feature is not localizable.

elachlan:
Fixes #9288
Fixes #9240
Fixes #9277
Fixed #2544

Unsure on the attributes for the new property and whether or not an auto implemented property is okay.

I went with 'AllowClickThrough' because I thought it was more self documenting.

My development environment isn't fully setup yet so this was a very quick pr without any testing.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9349)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Controls/ToolStrips/ToolStrip.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Controls/ToolStrips/ToolStrip.cs
@@ -4691,7 +4691,7 @@ public partial class ToolStrip : ScrollableControl, IArrangedElement, ISupportTo
             SnapFocus((HWND)(nint)m.WParamInternal);
         }
 
-        if (m.MsgInternal == PInvoke.WM_MOUSEACTIVATE)
+        if (!AllowClickThrough && m.MsgInternal == PInvoke.WM_MOUSEACTIVATE)
         {
             // We want to prevent taking focus if someone clicks on the toolstrip dropdown itself. The mouse message
             // will still go through, but focus won't be taken. If someone clicks on a child control (combobox,
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Controls/ToolStrips/ToolStrip.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Controls/ToolStrips/ToolStrip.cs
@@ -256,6 +256,9 @@ public partial class ToolStrip : ScrollableControl, IArrangedElement, ISupportTo
         set => base.AutoScrollPosition = value;
     }
 
+    /// <summary>
+    ///  When <see langword="true"/> Allows the control to be interacted when window does not have focus.
+    /// </summary>
     [Browsable(true)]
     [DefaultValue(false)]
     [SRDescription(nameof(SR.ToolStripAllowClickThrough))]
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Controls/ToolStrips/ToolStrip.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Controls/ToolStrips/ToolStrip.cs
@@ -256,6 +256,14 @@ public partial class ToolStrip : ScrollableControl, IArrangedElement, ISupportTo
         set => base.AutoScrollPosition = value;
     }
 
+    [Browsable(true)]
+    [DefaultValue(false)]
+    [SRDescription(nameof(SR.ToolStripAllowClickThrough))]
+    [SRCategory(nameof(SR.CatBehavior))]
+    [EditorBrowsable(EditorBrowsableState.Always)]
+    [DesignerSerializationVisibility(DesignerSerializationVisibility.Visible)]
+    public bool AllowClickThrough { get; set; }
+
     public override bool AllowDrop
     {
         get => base.AllowDrop;
@@ -4732,6 +4740,11 @@ public partial class ToolStrip : ScrollableControl, IArrangedElement, ISupportTo
         }
 
         base.WndProc(ref m);
+
+        if (AllowClickThrough && m.MsgInternal == PInvoke.WM_MOUSEACTIVATE && m.ResultInternal == PInvoke.MA_ACTIVATEANDEAT)
+        {
+            m.ResultInternal = (LRESULT)(nint)PInvoke.MA_ACTIVATE;
+        }
 
         if (m.Msg == (int)PInvoke.WM_NCDESTROY)
         {
--- a/src/System.Windows.Forms/src/PublicAPI.Unshipped.txt
+++ b/src/System.Windows.Forms/src/PublicAPI.Unshipped.txt
@@ -1,3 +1,5 @@
+System.Windows.Forms.ToolStrip.AllowClickThrough.get -> bool
+System.Windows.Forms.ToolStrip.AllowClickThrough.set -> void
 virtual System.Drawing.Design.PropertyValueUIHandler.Invoke(System.ComponentModel.ITypeDescriptorContext! context, System.ComponentModel.PropertyDescriptor! propDesc, System.Collections.ArrayList! valueUIItemList) -> void
 virtual System.Drawing.Design.PropertyValueUIItemInvokeHandler.Invoke(System.ComponentModel.ITypeDescriptorContext! context, System.ComponentModel.PropertyDescriptor! descriptor, System.Drawing.Design.PropertyValueUIItem! invokedItem) -> void
 virtual System.Windows.Forms.Application.MessageLoopCallback.Invoke() -> bool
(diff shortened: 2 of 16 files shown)


elachlan:
@merriemcgaw here is the PR.

elachlan:
My assumption here is that `MenuStrip` inherits from `ToolStrip` and so it should also get the property and behavior.

elachlan:
I have been able to test it and it looks like its working.

elachlan:
@rickbrew here is the ToolStrip ClickThrough changes if you wanted to review.

rickbrew:
Can we have a global/static property for this, or at least to configure the default property value? This seems like a property that will be set to the same value for everything in an app — it would be unusual to have a mix of values throughout an app’s UI, in other words. It would be easy to accidentally not set the property on a new piece of UI, resulting in bug churn. Or maybe some kind of inherited property ala WPF, where we set the property at the top of the UI tree and all the child controls use that for their default value.

elachlan:
Excellent point. I am unsure how to go about that. I'll have to ask the team what they prefer and for an example.

rickbrew:
> ... resulting in bug churn

BTW this is not theoretical -- I found a bug just a few months ago where a top-level `ToolStrip` was not using click-through (with my modified `WndProc`). It had been like that for years and nobody noticed / reported it. It was the only piece of UI that wasn't using it.

elachlan:
> > ... resulting in bug churn
> 
> BTW this is not theoretical -- I found a bug just a few months ago where a top-level `ToolStrip` was not using click-through (with my modified `WndProc`). It had been like that for years and nobody noticed / reported it. It was the only piece of UI that wasn't using it.
> 
> 

Yes, those kinds of bugs suck. We do need a centralised method of switching it.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
```diff
@@ -4884,6 +4892,11 @@ protected override void WndProc(ref Message m)
 
         base.WndProc(ref m);
 
+        if (AllowClickThrough && m.MsgInternal == PInvoke.WM_MOUSEACTIVATE && m.ResultInternal == PInvoke.MA_ACTIVATEANDEAT)
```
dreddy-work, line 4895: Did you get chance to look what `base.WndProc` is doing in `AllowClickThrough` case?

dreddy-work:
Thanks @elachlan. team will review PR this week and share feedback.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
elachlan: I did not. Based on the blog post by Rick, we just leave it alone so everything else processes the same.

lonitra:
This is awaiting API review. Now that we are in .NET 9, we plan on going through all our API suggestions in the beginning of the new year. Thank you for your patience!

ricardobossan:
Needs to solve conflicts. Otherwise LGTM.

elachlan:
@ricardobossan we are waiting on an API review from the team to get this included in .NET9.

merriemcgaw:
@ela1chlan we'll be getting those API reviews early in the new year!

elachlan:
@merriemcgaw Just checking this is still on the radar?

merriemcgaw:
Thanks for the ping @elachlan I do want this in .NET 9 for sure!

@lonitra / @JeremyKuhne can you guys get the API through review? Should we get this in now with a Preview flag while they are pending the official review?

elachlan:
@lonitra can you please review this? API has been approved!!

src/System.Windows.Forms/src/System/Windows/Forms/Controls/ToolStrips/ToolStrip.cs
```diff
@@ -256,6 +256,14 @@ public new Point AutoScrollPosition
         set => base.AutoScrollPosition = value;
     }
 
+    [Browsable(true)]
+    [DefaultValue(false)]
+    [SRDescription(nameof(SR.ToolStripAllowClickThrough))]
+    [SRCategory(nameof(SR.CatBehavior))]
+    [EditorBrowsable(EditorBrowsableState.Always)]
+    [DesignerSerializationVisibility(DesignerSerializationVisibility.Visible)]
+    public bool AllowClickThrough { get; set; }
```
lonitra, line 265: Could we add a summary for this property?

lonitra (commented):
Small comment, but otherwise LGTM. Woohoo! 🥳  Thanks for your patience on this. We are running out of time in .NET 9 to add a global/static property for this. We'll want to look at doing this in .NET 10

lonitra (approved):

PR #9354, opened 2023-06-23
Use `KEYEVENTF_EXTENDEDKEY` flag only extended keys

DJm00n:
Use `KEYEVENTF_EXTENDEDKEY` flag only when calling [SendInput](https://learn.microsoft.com/windows/win32/api/winuser/nf-winuser-sendinput) function with [scan codes](https://learn.microsoft.com/windows/win32/inputdev/about-keyboard-input#scan-codes) that have 0xE0 prefix.

`KEYEVENTF_EXTENDEDKEY` makes sense only when `wScan` value and `KEYEVENTF_SCANCODE` flag are set. It is useless otherwise.

**Update**: Seems `KEYEVENTF_EXTENDEDKEY` may be mapped to `KBDEXT` virtual key flag [in keyboard layout](https://github.com/microsoft/Windows-driver-samples/blob/main/input/layout/kbdus/kbdus.c#L40-L52). The only virtual keys that are having `KBDEXT` flag but don't have `0xE0` prefix in scan code are `VK_RSHIFT`, `VK_NUMLOCK` and `VK_PAUSE` (that is generated on Ctrl+NumLock keystroke). I don't think that this `KEYEVENTF_EXTENDEDKEY` flag is mandatory in this case when we do not provide `KEYEVENTF_SCANCODE` flag.

See:
https://learn.microsoft.com/windows/win32/api/winuser/ns-winuser-keybdinput https://learn.microsoft.com/windows/win32/inputdev/about-keyboard-input#extended-key-flag
https://github.com/microsoft/Windows-driver-samples/blob/main/input/layout/kbdus/kbdus.c#L40-L52

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9354)


src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Input/InputBuilder.cs
```diff
@@ -18,8 +20,8 @@ public static INPUT KeyDown(VIRTUAL_KEY keyCode)
                 ki = new KEYBDINPUT
                 {
                     wVk = keyCode,
-                    wScan = (ushort)(PInvoke.MapVirtualKey((uint)keyCode, MAP_VIRTUAL_KEY_TYPE.MAPVK_VK_TO_VSC) & 0xFFU),
-                    dwFlags = IsExtendedKey(keyCode) ? KEYBD_EVENT_FLAGS.KEYEVENTF_EXTENDEDKEY : 0,
+                    wScan = (ushort)(scanCode & 0xFF),
+                    dwFlags = (scanCode & 0xFF00) == 0xE000 ? KEYBD_EVENT_FLAGS.KEYEVENTF_EXTENDEDKEY : 0,
```
sharwell, line 24: 💭 Do we need to add handling for `0xE100`?

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextProvider.cs
```diff
@@ -112,12 +112,6 @@ public unsafe int SendKeyboardInputVK(VIRTUAL_KEY vk, bool press)
         keyboardInput.Anonymous.ki.wVk = vk;
         keyboardInput.Anonymous.ki.wScan = 0;
         keyboardInput.Anonymous.ki.dwFlags = press ? 0 : KEYBD_EVENT_FLAGS.KEYEVENTF_KEYUP;
-
-        if (VirtualKeyUtilities.IsExtendedKey(vk))
```
sharwell, line 116: ❔ Should we assert that `vk` is only one of the two known values passed by a caller (neither of which is an extended key)?

sharwell, line 116: ❓ Considering this path always used to set `KEYBD_EVENT_FLAGS.KEYEVENTF_EXTENDEDKEY` (the two inputs are VIRTUAL_KEY.VK_RIGHT` and `VIRTUAL_KEY.VK_LEFT`, both of which were considered extended), is there a reason we are no longer doing so?

src/System.Windows.Forms/src/System/Windows/Forms/SendKeys.cs
```diff
@@ -703,12 +703,6 @@ private static unsafe void SendInput(ReadOnlySpan<byte> oldKeyboardState, SKEven
                             currentInput[0].Anonymous.ki.dwFlags |= KEYBD_EVENT_FLAGS.KEYEVENTF_KEYUP;
                         }
 
-                        // Sets KEYEVENTF_EXTENDEDKEY flag if necessary.
-                        if (IsExtendedKey(skEvent))
```
sharwell, line 707: ❔ How do we know `skEvent` isn't going to be extended?

sharwell (changes requested):
Flagging this one for a more detailed internal review.

Reviewer note: Please feel free to reach out to me internally to help move it forward.

JeremyKuhne:
@DJm00n @sharwell do we have a specific scenario that is broken? What is broken that this is fixing?

DJm00n:
@JeremyKuhne there shouldn't be any visible change in behaviour. I just removed a bit of useless code.

DJm00n:
@sharwell do you have any updates on this topic?

sharwell:
I'm hesitant to make a change in this space unless an observable bug is identified, and/or an expert within the Windows team explains why the change would be needed.

DJm00n:
@sharwell maybe you can ask someone from Windows team to review this?

dreddy-work:
Given the concerns and no reported issue, this cleanup isn't prioritized now. Also, some of these constants may be converted to CSwin32.

Closing the PR.

PR #9356, opened 2023-06-23
Include stack trace from log method

sharwell:
Sometimes the first chance exception stack is incomplete, so the additional stack collected at the point where it's written to the log can help narrow down code locations.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9356)

### DIFF
--- a/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/DataCollectionService.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/DataCollectionService.cs
@@ -5,6 +5,7 @@
 using System.Collections.Immutable;
 using System.Runtime.CompilerServices;
 using System.Runtime.ExceptionServices;
+using System.Text;
 using Xunit.Abstractions;
 using Xunit.Sdk;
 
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/DataCollectionService.cs
```diff
@@ -164,7 +165,12 @@ internal static void CaptureFailureState(string testName, Exception ex)
 
             Directory.CreateDirectory(logDir);
 
-            File.WriteAllText(CreateLogFileName(logDir, timestamp, testName, errorId, logId: string.Empty, "log"), ex.ToString());
+            var exceptionDetails = new StringBuilder();
```
Tanya-Solyanik, line 168:
```suggestion
            StringBuilder exceptionDetails = new ();
```

Tanya-Solyanik (approved):

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/DataCollectionService.cs
sharwell: ➡️ We're looking to address this one separately as part of a private code style analyzer.

PR #9358, opened 2023-06-23
Dynamically query interfaces in AxHost

JeremyKuhne:
Get native interfaces by querying when needed. This is more performant than creating a RCW for these interfaces and is a step towards enabling AOT.

This does not change the serialization interfaces, they will come in a follow-up PR. The handling there is a little trickier.

Also:

- Some style cleanup
- Add a interface check helper in ComScope
- Add a simple assert helper in HRESULT
- Fix ref count in COM2PictureConverter test


src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
```diff
@@ -1806,47 +1792,49 @@ protected internal override unsafe bool ProcessMnemonic(char charCode)
...
+        PInvoke.GetCursorPos(out Point point);
+
+        MSG msg = new()
+        {
+            // We don't have a message so we must create one ourselves.
+            // The message we are creating is a WM_SYSKEYDOWN with the right alt key setting.
+            hwnd = (ContainingControl is null) ? HWND.Null : ContainingControl.HWND,
+            message = PInvoke.WM_SYSKEYDOWN,
+            wParam = (WPARAM)char.ToUpper(charCode, CultureInfo.CurrentCulture),
+            lParam = 0x20180001,
```
Tanya-Solyanik, line 1815: What is this number?

src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/HRESULT.cs
```diff
@@ -15,6 +15,9 @@ public static HRESULT HRESULT_FROM_WIN32(WIN32_ERROR error)
         return (HRESULT)ex.HResult;
     }
 
+    [Conditional("DEBUG")]
+    public void AssertSuccess() => Debug.Assert(Succeeded, "Result failed: {hr}");
```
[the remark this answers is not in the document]
JeremyKuhne: That is what `{hr}` does.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
JeremyKuhne: I've clarified it. Why it is always the scan code for the `O` key, I have no idea. Presuming it's arbitrary.

src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/HRESULT.cs
```diff
@@ -15,6 +15,8 @@ public static HRESULT HRESULT_FROM_WIN32(WIN32_ERROR error)
         return (HRESULT)ex.HResult;
     }
 
+    public void AssertSuccess() => Debug.Assert(Succeeded, "Result failed: {hr}");
```
Tanya-Solyanik, line 18:
```suggestion
    public void AssertSuccess() => Debug.Assert(Succeeded, $"Result failed: {hr}");
```

Tanya-Solyanik (approved):

PR #9365, opened 2023-06-25
fix Accessible properties AccessKey and KeyboardShortcut ignore UseMnemonic = false

### LINKED ISSUE #9296: Accessible properties AccessKey and KeyboardShortcut ignore UseMnemonic = false
### .NET version

.NET 8.0.0-preview.6.23307.4

No

No

### Issue description

Button, Label and derived controls have [mnemonics](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.label.usemnemonic) feature, which allows making a shortcut with one of characters in control's Text. For example, a button with `UseMnemonic = true` and `Text = "&Button"` will be displayed as "<ins>B</ins>utton" and can be activated by pressing Alt+B. If `UseMnemonic` is set to `false`, ampersand will be treated as a regular character, and Alt+B shortcut won't work.

Accessible objects set mnemonic shortcut in `AccessKey` and `KeyboardShortcut` properties, which makes screen readers read the shortcut when control is focused. However, when mnemonic is disabled by setting `UseMnemonic = false`, accessible properties ignore it and still set the shortcut. Screen readers will read the shortcut, even though it doesn't work.

#### Actual behavior

https://github.com/dotnet/winforms/assets/102954094/2a100682-d544-4d6a-848e-e25f4ab7a2b7

Narrator reads the form as follows:

> Mnemonic, edit, Alt+ m, 
> No mnemonic, edit, Alt+ n, 
> Mnemonic, button, Alt+ e, 
> No mnemonic, button, Alt+ o, 
> Mnemonic, radio button, selected, Alt+ i, 
> Mnemonic, radio button, non-selected, Alt+ i, 
> No mnemonic, radio button, selected, Alt+ c, 

For controls with "No mnemonic" text, pressing announced shortcut will not do anything

#### Expected behavior

* `AccessKey` and `KeyboardShortcut` should be empty when mnemonics are disabled
* Narrator should not read shortcut when mnemonics are disabled
* Narrator should read ampersand as part of the text when mnemonics are disabled, exactly as displayed

### Steps to reproduce

Run sample app: 
[MnemonicRepro.zip](https://github.com/dotnet/winforms/files/11759886/MnemonicRepro.zip)

Most likely it also reproduces with all controls derived from Button and Label.

Epica3055:
Fixes #9296 

## Proposed changes

- 
- Before return shortcut and name, add if useMnemonic condition
- 

## Regression? 

-  No

### Before

image

### After

image

## Test methodology 

- 
- manually and Unit Tests
- 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9365)


src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlAccessibleObject.cs
```diff
@@ -272,14 +272,36 @@ public IntPtr Handle
...
+                return useMnemonic;
+            }
+
+            if (PreviousLabel is not null)
+            {
+                return PreviousLabel.UseMnemonic;
+            }
+
+            return true;
+        }
```
dreddy-work, line 289:
```suggestion
        private bool GetOwnerUseMnemonic(Control? control)
        {
            if (control is null)
            {
               return true;
            }

            Reflection.PropertyInfo? propertyInfo = control.GetType().GetProperty(nameof(ButtonBase.UseMnemonic));
            if (propertyInfo is not null && propertyInfo.GetValue(control) is bool useMnemonic)
            {
                return useMnemonic;
            }

            if (PreviousLabel is not null)
            {
                return PreviousLabel.UseMnemonic;
            }

            return true;
        }
```

dreddy-work: And change/simplify use cases.

Epica3055: did you mean like this one?
image

src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlAccessibleObject.cs
```diff
@@ -272,14 +272,36 @@ public IntPtr Handle
             }
         }
 
+        private bool GetOwnerUseMnemonic(Control control)
```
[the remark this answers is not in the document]
Epica3055: I have a different point of view. Actually, not every control has a UseMnemonic property(like TextBox), only a few controls have it.
But a control may have a label before it. Label has UseMnemonic Property, I think this label could determine UseMnemonic  of the control. So I think basicly every contol has UseMnemonic  property.
Maybe using reflection is a little bit easier than to implement virtual function in every derived AO class of control accessible object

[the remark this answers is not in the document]
Tanya-Solyanik: Then the default implementation on Control class should return `true`.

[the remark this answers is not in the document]
dmitrii-drobotov: Perhaps we could handle shortcuts from `TextLabel` and from Button/Label's mnemonic separately? If there is a Label with `UseMnemonic = true; Text = "&Text"` and next to it a Button with `UseMnemonic = false; Text = "a&b"`, the button will be accessible by Alt+T shortcut. But with current implementation, it won't have this shortcut in `KeyboardShortcut` property.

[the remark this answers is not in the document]
Tanya-Solyanik: @dmitrii-drobotov  - good point and thank for testing the behavior!  

This comment, however, was about reflection. We are trying to not use it for performance reasons and because reflection prevents making these controls trimmable - https://learn.microsoft.com/dotnet/core/deploying/trimming/trim-self-contained. We are not trimmable as a whole yet, but are working to enable it in the future. So a virtual method is a better route as it avoids reflection.

src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlAccessibleObject.cs
Epica3055: @dreddy-work could you please take a look at this?

src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlAccessibleObject.cs
```diff
@@ -272,14 +272,36 @@ public IntPtr Handle
...
+        {
+            Reflection.PropertyInfo? propertyInfo = control.GetType().GetProperty(nameof(ButtonBase.UseMnemonic));
```
Tanya-Solyanik, line 277: Please get rid of reflection in this code.

Tanya-Solyanik, line 277: UseMnemonic property on button controls can be used to set `ControlStyles.UseTextForAccessibility`, then the existing logic will take it into account

src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.ButtonBaseAccessibleObject.cs
```diff
@@ -26,5 +26,50 @@ public override void DoDefaultAction()
...
+                    return null;
+                }
+
+                return GetKeyboardShortcut(owner, owner.UseMnemonic, PreviousLabel);
+            }
+        }
+
+        public override string? Name
+        {
+            get
```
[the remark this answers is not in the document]
dreddy-work: also , other places in this PR.

src/System.Windows.Forms/src/System/Windows/Forms/Label.LabelAccessibleObject.cs
```diff
@@ -13,5 +13,26 @@ public LabelAccessibleObject(Label owner) : base(owner)
         }
 
         public override AccessibleRole Role => this.GetOwnerAccessibleRole(AccessibleRole.StaticText);
+
+        public override string? KeyboardShortcut
+        {
+            get
+            {
+                if (!this.TryGetOwnerAs(out Label? owner) || !owner.UseMnemonic)
```
dreddy-work, line 21: conditional return with expression body.

dreddy-work (changes requested):

src/System.Windows.Forms/src/System/Windows/Forms/RadioButton.RadioButtonAccessibleObject.cs
```diff
@@ -55,5 +55,25 @@ internal override bool IsPatternSupported(UiaCore.UIA patternId)
                     p == UiaCore.UIA.SelectionItemPatternId => true,
                 _ => base.IsPatternSupported(patternId)
             };
+        public override string? KeyboardShortcut
```
Tanya-Solyanik, line 58: Please add a blank line above this method.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Button.ButtonAccessibleObjectTests.cs
```diff
@@ -124,4 +124,38 @@ public void ButtonAccessibleObject_GetPropertyValue_ControlType_IsExpected_ForCu
...
     }
+
+    // Unit test for https://github.com/dotnet/winforms/issues/9296
+    [WinFormsFact]
+    public void ButtonAccessibleObject_Shortcut_Invoke_ReturnsExpected()
+    {
+        using var button = new Button();
+        button.Text = "&btn";
+        Assert.False(button.IsHandleCreated);
+        var buttonAccessibleObject = new Button.ButtonAccessibleObject(button);
```
Tanya-Solyanik, line 135: Consider introducing an alias for `Button.ButtonAccessibleObject` and other nested accessible objects where you touched the test files in this PR.
```
using ButtonAccessibleObject = Button.ButtonAccessibleObject;
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBox.CheckBoxAccessibleObjectTests.cs
```diff
@@ -286,4 +286,38 @@ public void CheckBoxAccessibleObject_GetPropertyValue_ControlType_IsExpected_For
...
     }
+
+    // Unit test for https://github.com/dotnet/winforms/issues/9296
+    [WinFormsFact]
+    public void CheckBoxAccessibleObject_Shortcut_Invoke_ReturnsExpected()
+    {
+        using var checkBox = new CheckBox();
+        checkBox.Text = "&checkBox";
+        Assert.False(checkBox.IsHandleCreated);
+        var buttonAccessibleObject = new CheckBox.CheckBoxAccessibleObject(checkBox);
```
Tanya-Solyanik, line 297: Consider renaming `buttonAccessibleObject ` -> `accessibleObject` in all changes tests.

Tanya-Solyanik: Alternatively, consider creating a theory test, where you pass in typeof control to create and typeof accessible object to create.  In fact, this theory should be working for almost all controls.  @dmitrii-drobotov  - do we have anything like this already?

This could be done in a follow up PR.

src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.TextBoxBaseAccessibleObject.cs
```diff
@@ -127,5 +128,18 @@ internal override SupportedTextSelection SupportedTextSelectionInternal
...
+        public override string? KeyboardShortcut
+        {
+            get
+            {
+                if (!this.TryGetOwnerAs(out TextBoxBase? owner))
+                {
+                    return null;
+                }
+
+                return ButtonBaseAccessibleObject.GetKeyboardShortcut(owner, false, PreviousLabel);
```
Tanya-Solyanik, line 141: Please add parameter name in front of the literal. This applies to multiple changes in this PR.

Tanya-Solyanik (commented):
Thank you for simplifying the change!

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Button.ButtonAccessibleObjectTests.cs
Epica3055: It looks like I can't use using, because ButtonAccessibleObject doesn't implement IDisposable.

Tanya-Solyanik: `using` as an alias definition - https://learn.microsoft.com/dotnet/csharp/language-reference/keywords/using-directive#using-alias

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBox.CheckBoxAccessibleObjectTests.cs
Epica3055: I don't fully understand.

dmitrii-drobotov: Yes, there is a similar test: https://github.com/dotnet/winforms/blob/918e34631094c184e6b29d659842b540f092ed98/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Control.ControlAccessibleObjectTests.cs#L1251-L1271

@Epica3055 - this PRs adds many very similar tests, it would be cleaner to use a theory test like above with type of control as argument to reduce duplication.

src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.ButtonBaseAccessibleObject.cs
```diff
@@ -26,5 +26,34 @@ public override void DoDefaultAction()
...
+                }
+            }
+
+            return (mnemonic == (char)0) ? null : $"Alt+{mnemonic}";
+        }
+
+        public override string? KeyboardShortcut => this.TryGetOwnerAs(out ButtonBase? owner)
+                                    ? GetKeyboardShortcut(owner, owner.UseMnemonic, PreviousLabel) : null;
+
+        public override string? Name => this.TryGetOwnerAs(out ButtonBase? owner) && owner.UseMnemonic ? base.Name : TextLabel;
```
dmitrii-drobotov, line 57: I'm not sure if it correctly handles custom `AccessibleName` with `UseMnemonic = false`. Could you please test it? We might need to do something like in `base.Name` but return just `TextLabel` without `WindowsFormsUtils.TextWithoutMnemonics`:
https://github.com/dotnet/winforms/blob/918e34631094c184e6b29d659842b540f092ed98/src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlAccessibleObject.cs#L286-L300

src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlAccessibleObject.cs
dreddy-work: Looks `GetOwnerUseMnemonic` method is complete removed. So this comment is not valid anymore.

src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.ButtonBaseAccessibleObject.cs
```diff
@@ -26,5 +26,45 @@ public override void DoDefaultAction()
...
+
+        public override string? KeyboardShortcut => this.TryGetOwnerAs(out ButtonBase? owner)
+                                    ? GetKeyboardShortcut(owner, owner.UseMnemonic, PreviousLabel) : null;
+
+        public override string? Name
+        {
+            get
+            {
+                if (this.TryGetOwnerAs(out ButtonBase? owner) && owner.AccessibleName is { } name)
+                {
```
Tanya-Solyanik, line 62: If the owner is gone, we can short-circuit. If there is no owner, there is no PreviousLabel for TextLabel to read either. See ConstrolAccessibleObject class. And since the base class is repeating logic from this method, I would not call it, but simply repeat the code from the base class here.

```
                if (!this.TryGetOwnerAs(out ButtonBase? owner))
                {
                    return null;
                }
    
                if (owner.AccessibleName is { } name)
                {
                    return name;
                }

                return owner.UseMnemonic ? WindowsFormsUtils.TextWithoutMnemonics(TextLabel) : TextLabel;
```

src/System.Windows.Forms/src/System/Windows/Forms/RadioButton.RadioButtonAccessibleObject.cs
```diff
@@ -55,5 +55,21 @@ internal override bool IsPatternSupported(UiaCore.UIA patternId)
                     p == UiaCore.UIA.SelectionItemPatternId => true,
                 _ => base.IsPatternSupported(patternId)
             };
+
+        public override string? KeyboardShortcut => this.TryGetOwnerAs(out RadioButton? owner) ?
+            ButtonBaseAccessibleObject.GetKeyboardShortcut(control: owner, useMnemonic: owner.UseMnemonic, previousLabel: PreviousLabel) : null;
+
+        public override string? Name
```
Tanya-Solyanik, line 62: This method can be inherited from ButtonBase.

Epica3055: Sorry, I don't fully understand. RadioButtonAccessibleObject doesn't derive from ButtonBaseAccessibleObject

Tanya-Solyanik: Oh, I see, this is unfortunate. This PR is ok as is then.

Tanya-Solyanik (dismissed):

src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.ButtonBaseAccessibleObject.cs
```diff
@@ -26,5 +26,51 @@ public override void DoDefaultAction()
...
+        {
+            char mnemonic = (char)0;
+
+            if (!useMnemonic && previousLabel is not null && previousLabel.UseMnemonic)
+            {
+                string text = previousLabel.Text;
+                if (!string.IsNullOrEmpty(text))
+                {
+                    mnemonic = WindowsFormsUtils.GetMnemonic(text, false);
+                }
```
dreddy-work, line 40:
```suggestion
            mnemonic = WindowsFormsUtils.GetMnemonic(previousLabel.Text, false);
```

You do not need explicit check. `GetMnemonic` handle null or empty strings.

src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.ButtonBaseAccessibleObject.cs
```diff
@@ -26,5 +26,51 @@ public override void DoDefaultAction()
...
+            }
+            else if (useMnemonic)
+            {
+                string text = control.Text;
+                if (!string.IsNullOrEmpty(text))
+                {
+                    mnemonic = WindowsFormsUtils.GetMnemonic(text, false);
+                }
```
dreddy-work, line 48: same here.

src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.ButtonBaseAccessibleObject.cs
```diff
@@ -26,5 +26,51 @@ public override void DoDefaultAction()
...
+            }
```
dmitrii-drobotov, line 49: I'm not sure if it correctly handles case when button has `UseMnemonic = true` and its `Text` doesn't contain a mnemonic shortcut, but previous label does, like this:
image

The button can be focused by Alt+L, but the shortcut will be empty. Could you look into this as well?

src/System.Windows.Forms/src/System/Windows/Forms/Label.LabelAccessibleObject.cs
```diff
@@ -13,5 +13,25 @@ public LabelAccessibleObject(Label owner) : base(owner)
...
 
         public override AccessibleRole Role => this.GetOwnerAccessibleRole(AccessibleRole.StaticText);
+
+        public override string? KeyboardShortcut => !this.TryGetOwnerAs(out Label? owner) || !owner.UseMnemonic ? null : base.KeyboardShortcut;
+
+        public override string? Name
+        {
+            get
+            {
+                if (!this.TryGetOwnerAs(out Label? owner))
```
dreddy-work, line 23: why not move this logic to `ControlAccessibleObject`? and simply call.

```CS
   return owner.UseMnemonic ? base.Name : TextLabel;
```

Also, if natively parented, do we leave accessible name as `null`?

dreddy-work: I see Control does not have `UseMnemonic`  and with offlien discussion with Tanya, it may not worth moving logic to base class here.

src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.ButtonBaseAccessibleObject.cs
```diff
@@ -26,5 +26,51 @@ public override void DoDefaultAction()
                 owner.OnClick(EventArgs.Empty);
             }
         }
+
+        internal static string? GetKeyboardShortcut(Control control, bool useMnemonic, Label? previousLabel)
+        {
+            char mnemonic = (char)0;
```
LeafShi1, line 32: It is more concise and readable to use` char mnemonic = '\0';` here

LeafShi1 (approved):

LeafShi1:
> Fixes #9296
> 
> ## Proposed changes
> * Before return shortcut and name, add if useMnemonic condition
> 
> ## Regression?
> * No
> 
> ## Screenshots
> ### Before
> image
> 
> ### After
> image
> 
> ## Test methodology
> * manually and Unit Tests
> 
> ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9

LeafShi1:
Please add more description on 
image

src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.ButtonBaseAccessibleObject.cs
```diff
@@ -26,5 +26,48 @@ public override void DoDefaultAction()
                 owner.OnClick(EventArgs.Empty);
             }
         }
+
+        private static bool ControlTextContainsAmpersand(Control control)
+        {
+            return control is not null && control.Text is not null && control.Text.Contains('&');
+        }
```
dmitrii-drobotov, line 33: I think we can use `WindowsFormsUtils.ContainsMnemonic` instead of this method

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Control.ControlAccessibleObjectTests.cs
```diff
@@ -1626,6 +1626,84 @@ static Control.ControlAccessibleObject CreateAndDisposeControl(Type type)
         }
     }
 
+    public static IEnumerable<object[]> ControlAccessibleObject_Name_KeyboardShortcut_TestData()
+    {
+        // Controls that have UseMnemonic property.
+        yield return new object[] { typeof(Label) };
+        yield return new object[] { typeof(Button) };
+        yield return new object[] { typeof(RadioButton) };
+        yield return new object[] { typeof(CheckBox) };
```
dmitrii-drobotov, line 1635: There is also the `LinkLabel` with `UseMnemonic`, but its accessible object already extends from `LabelAccessibleObject`. Could you check if it behaves correctly with these changes? And also include it into unit tests.

SimonZhao888 (approved):

src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.TextBoxBaseAccessibleObject.cs
```diff
@@ -127,5 +128,9 @@ internal override SupportedTextSelection SupportedTextSelectionInternal
 
         internal override ITextRangeProvider? GetRangeFromAnnotation(IRawElementProviderSimple annotationElement)
             => _textProvider?.RangeFromAnnotation(annotationElement);
+
+        public override string? KeyboardShortcut => this.TryGetOwnerAs(out TextBoxBase? owner)
+            ? ButtonBaseAccessibleObject.GetKeyboardShortcut(control: owner, useMnemonic: false, previousLabel: PreviousLabel)
```
dmitrii-drobotov, line 133: We usually specify parameter names only for literal values (`false`, `"text"`, `10`, `null`, etc.). For variables or properties we can omit it.
```suggestion
            ? ButtonBaseAccessibleObject.GetKeyboardShortcut(control: owner, useMnemonic: false, PreviousLabel)
```
Same for a few other changes in this PR.

dmitrii-drobotov (commented):
Just some code style comments, the rest looks good to me!

src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.ButtonBaseAccessibleObject.cs
```diff
@@ -26,5 +26,43 @@ public override void DoDefaultAction()
...
+        {
+            char mnemonic = '\0';
+
+            if ((!useMnemonic || !WindowsFormsUtils.ContainsMnemonic(control?.Text)) && previousLabel is not null && previousLabel.UseMnemonic)
+            {
+                mnemonic = WindowsFormsUtils.GetMnemonic(previousLabel.Text, false);
+            }
+            else if (useMnemonic)
+            {
+                mnemonic = WindowsFormsUtils.GetMnemonic(control?.Text, false);
```
dmitrii-drobotov, line 40: Nit: `control` variable is not nullable, so we don't have to check it for null.
```suggestion
            if ((!useMnemonic || !WindowsFormsUtils.ContainsMnemonic(control.Text)) && previousLabel is not null && previousLabel.UseMnemonic)
            {
                mnemonic = WindowsFormsUtils.GetMnemonic(previousLabel.Text, false);
            }
            else if (useMnemonic)
            {
                mnemonic = WindowsFormsUtils.GetMnemonic(control.Text, false);
```

src/System.Windows.Forms/src/System/Windows/Forms/ButtonBase.ButtonBaseAccessibleObject.cs
```diff
@@ -26,5 +26,43 @@ public override void DoDefaultAction()
...
             }
         }
+
+        internal static string? GetKeyboardShortcut(Control control, bool useMnemonic, Label? previousLabel)
+        {
+            char mnemonic = '\0';
+
+            if ((!useMnemonic || !WindowsFormsUtils.ContainsMnemonic(control?.Text)) && previousLabel is not null && previousLabel.UseMnemonic)
+            {
+                mnemonic = WindowsFormsUtils.GetMnemonic(previousLabel.Text, false);
```
dmitrii-drobotov, line 36: Could you also specify parameter name for `GetMnemonic` calls?
```suggestion
                mnemonic = WindowsFormsUtils.GetMnemonic(previousLabel.Text, convertToUpperCase: false);
```

Tanya-Solyanik (approved):
Please let @dmitrii-drobotov and @dreddy-work finish their reviews before merging

dreddy-work (approved):

PR #8588, opened 2023-02-08
Fix ToolStrip/MenuStrip/StatusStrip arrow keys navigation

v-elnovikova:
Fixes #8587

## Proposed changes

- Fix ToolsStrip.TabStop property value handling in calculating the next ToolStripItem to switch to.
- Modify unit tests to ensure GetNextItem method handles TabStop true value correctly.

## Customer Impact

- Arrow keys work correctly when navigation ToolStrip/MenuStrip/StatusStrip items from keyboard.

## Regression? 

- No

## Risk

- Minimal

### Before

Focus goes *right* on both Left and Right keys:

https://user-images.githubusercontent.com/113603457/217533909-e7af42f1-b48c-40e8-b670-bce1797ccac1.mp4

### After

Focus goes *left* on Left key and *right* on the Right key:

https://user-images.githubusercontent.com/113603457/217534004-530e0969-86f0-4470-a9d9-60dd8a06362a.mp4

Tab and Shift+Tab navigation is not affected:

https://user-images.githubusercontent.com/113603457/217534084-6c900fa3-001e-4f39-8c21-6ba12f3df049.mp4

## Test methodology 

- Manual

## Test environment(s) 

- .NET 8.0.0-alpha.1.23057.5

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8588)

Issue #5795, opened 2021-09-17
Accessibility: ToolStrip: TabStop= true, pressing Shift-Tab for first or last item: toolStripComboBox/TextBox in ToolStrips doesn't move to right item right 


Olina-Zhang:
* .NET Core Version: 
.Net SDK 6.0.100-rc.2.21466.17

* Have you experienced this same bug with .NET Framework?: 
Yes, but the result is different from .net 6.0 after the PR: https://github.com/dotnet/winforms/pull/5726

**Minimal repro:**
Scenario A:
1.	Create a Winforms .Net Core application with 2 toolStrips, set their TabStop as true
2.	In the first toolStrip, the last item is not toolStripComboBox/TextBox, and in the second toolStrip, the last item is toolStripComboBox/TextBox
3.	Build and run app
4.	Press Shift-Tab several times
[WinFormsApp14.zip](https://github.com/dotnet/winforms/files/7183233/WinFormsApp14.zip)

Scenario B:
1.	Create a Winforms .Net Core application with 2 toolStrips, set their TabStop as true
2.	In the first toolStrip, the first  item is not toolStripComboBox/TextBox, and in the second toolStrip, the first item is toolStripComboBox/TextBox
3.	Build and run app
4.	Press tab firstly to the first item in the second toolStrip, then press Shift-Tab 
[WinFormsApp15.zip](https://github.com/dotnet/winforms/files/7183239/WinFormsApp15.zip)

**Actual behavior:**
For Scenario A: trap in the last item: toolStripComboBox/TextBox of the second toolStrip
![ShiftTabIssue1](https://user-images.githubusercontent.com/26474449/133738689-412380ae-6179-4d65-8e2b-4ab86a489382.gif)
For Scenario B: it is not in the last item for the first toolStrip, it is in the last item of current toolStrip after pressing Shift-Tab
![ShiftTabIssue2](https://user-images.githubusercontent.com/26474449/133738749-def6a376-f3d7-435e-a5bb-f7d87021a19f.gif)

**Expected behavior:**
For Scenario A: not trap in the last item: toolStripComboBox/TextBox of the second toolStrip
For Scenario B: it is in the last item of the first toolStrip after pressing Shift-Tab

ArtemTatarinov:
[...]

PR #9390, opened 2023-06-29
Fix the next position is not correct when toolstrip using mouse or ke…

### LINKED ISSUE #9181: The next focus position is not correct when selecting on a ToolStripButton using mouse or typing "enter" keyboard and then typing "right or left" keyboard.
### .NET version

.NET SDK 8.0.100-preview.5.23274.2

No

No

For .NET core 3.1 and .NET 5.0, the behavior is same as .NET framework: 
When typing Right keyboard, the next focus position is first item on the left. 
When typing Left keyboard, the next focus position is first item on the Right. 

For .NET 6.0/7.0/8.0, the behavior is: the next focus position is first item on the left, regardless of whether the right or left key is typed. 

### Issue description

When selecting on a ToolStripButton using mouse or typing "enter" keyboard and then typing "right or left" keyboard, the next focus position is first item on the left, regardless of whether the right or left key is typed. 
ReproSteps

### Steps to reproduce

1. Extract and open the test app. 
[TestApp.zip](https://github.com/dotnet/winforms/files/11572425/TestApp.zip)
2. Build and run this project.
3. Select on a ToolStripButton using mouse or typing "enter" keyboard.
4. Type "right or left" keyboard.

SimonZhao888:
Fixes #9181 
Tracing the behavior described in the bug, we find that when calling the GetNextItemHorizontal() method, their input parameters are the same, which caused that, although we clicked on different buttons Left/Right, the behavior of the control is the same.
## Proposed changes

- Modify the assignment of the forward parameter in the GetNextItem(ToolStripItem? start, ArrowDirection direction) method to distinguish Left/Right.

## Customer Impact

- When typing Left/Right keyboard, the next focus position will be the next item, not the first item on the Left/ Right.

## Regression? 

-  No

## Risk

- Minimal

### Before
When selecting on a ToolStripButton using mouse or typing "enter" keyboard and then typing "right or left" keyboard, the next focus position is first item on the left, regardless of whether the right or left key is typed.
image

### After
Keep the same behavior as  .NET framework, .NET 5.0 and .NET core 3.1. When typing Right keyboard, the next focus position is first item on the left, and if typing Left keyboard, the next focus position is first item on the Right.
ToolStrip

## Test methodology 

- Manual

## Test environment(s) 

- 8.0.0-preview.7.23323.5

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9390)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
@@ -45,6 +45,7 @@ public partial class ToolStrip : ScrollableControl, IArrangedElement, ISupportTo
     private bool _showItemToolTips;
     private MouseHoverTimer? _mouseHoverTimer;
     private ToolStripItem? _currentlyActiveTooltipItem;
+    private ToolStripItem? _lastActiveTooltipItem;
     private NativeWindow? _dropDownOwnerWindow;
     private byte _mouseDownID;  // NEVER use this directly from another class, 0 should never be returned to another class.
     private ToolStripRenderer? _renderer;
@@ -1944,6 +1945,8 @@ public partial class ToolStrip : ScrollableControl, IArrangedElement, ISupportTo
                 Invalidate(regionRect, true);
                 Update();
             }
+
+            _lastActiveTooltipItem = null;
         }
         finally
         {
@@ -2159,6 +2162,11 @@ public partial class ToolStrip : ScrollableControl, IArrangedElement, ISupportTo
     /// </summary>
     public virtual ToolStripItem? GetNextItem(ToolStripItem? start, ArrowDirection direction)
     {
+        if (TabStop && start is null && _lastActiveTooltipItem is not null)
+        {
+            start = _lastActiveTooltipItem;
+        }
+
         switch (direction)
         {
             case ArrowDirection.Right:
@@ -4217,6 +4225,14 @@ public partial class ToolStrip : ScrollableControl, IArrangedElement, ISupportTo
         if (KeyboardActive && !Focused && !ContainsFocus)
         {
             KeyboardActive = false;
+        }
+    }
+
+    internal void ResetLastActiveToolTipItem(ToolStripItem? item)
+    {
+        if (item is not null)
+        {
+            _lastActiveTooltipItem = item;
         }
     }
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
@@ -2243,7 +2243,6 @@ public partial class ToolStrip : ScrollableControl, IArrangedElement, ISupportTo
 
     private ToolStripItem GetStartItem(bool forward, bool isDropDown, ArrowDirection arrowDirection = ArrowDirection.Right)
     {
-
         if (forward && arrowDirection == ArrowDirection.Left)
         {
             return DisplayedItems[0];
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
@@ -45,7 +45,6 @@ public partial class ToolStrip : ScrollableControl, IArrangedElement, ISupportTo
     private bool _showItemToolTips;
     private MouseHoverTimer? _mouseHoverTimer;
     private ToolStripItem? _currentlyActiveTooltipItem;
-    private ToolStripItem? _lastActiveTooltipItem;
     private NativeWindow? _dropDownOwnerWindow;
     private byte _mouseDownID;  // NEVER use this directly from another class, 0 should never be returned to another class.
     private ToolStripRenderer? _renderer;
@@ -1945,8 +1944,6 @@ public partial class ToolStrip : ScrollableControl, IArrangedElement, ISupportTo
                 Invalidate(regionRect, true);
                 Update();
             }
-
-            _lastActiveTooltipItem = null;
         }
         finally
         {
@@ -2165,12 +2162,10 @@ public partial class ToolStrip : ScrollableControl, IArrangedElement, ISupportTo
         switch (direction)
         {
             case ArrowDirection.Right:
-                start ??= _lastActiveTooltipItem;
                 return GetNextItemHorizontal(start, forward: true);
             case ArrowDirection.Left:
-                start ??= _lastActiveTooltipItem;
                 bool forward = LastKeyData == Keys.Tab || (TabStop && start is null);
-                return GetNextItemHorizontal(start, forward);
+                return GetNextItemHorizontal(start, forward, ArrowDirection.Left);
             case ArrowDirection.Down:
                 return GetNextItemVertical(start, down: true);
             case ArrowDirection.Up:
@@ -2183,7 +2178,7 @@ public partial class ToolStrip : ScrollableControl, IArrangedElement, ISupportTo
     /// <remarks>
     ///  Helper function for GetNextItem - do not directly call this.
     /// </remarks>
-    private ToolStripItem? GetNextItemHorizontal(ToolStripItem? start, bool forward)
+    private ToolStripItem? GetNextItemHorizontal(ToolStripItem? start, bool forward, ArrowDirection arrowDirection = ArrowDirection.Right)
     {
         if (DisplayedItems.Count <= 0)
         {
@@ -2196,7 +2191,7 @@ public partial class ToolStrip : ScrollableControl, IArrangedElement, ISupportTo
         // backward direction entering the toolstrip, it means that the first
         // toolstrip item should be selected irrespectively TAB or SHIFT+TAB
         // is pressed.
-        start ??= GetStartItem(forward, dropDown is not null);
+        start ??= GetStartItem(forward, dropDown is not null, arrowDirection);
 
         int current = DisplayedItems.IndexOf(start);
         if (current == -1)
@@ -2212,14 +2207,21 @@ public partial class ToolStrip : ScrollableControl, IArrangedElement, ISupportTo
 
         do
         {
-            if (forward)
+            if (forward && arrowDirection == ArrowDirection.Left)
             {
-                current = ++current % count;
+                current = (--current < 0) ? count + current : current;
             }
             else
             {
-                // Provide negative wrap if necessary.
-                current = (--current < 0) ? count + current : current;
+                if (forward)
+                {
+                    current = ++current % count;
+                }
+                else
+                {
+                    // Provide negative wrap if necessary.
+                    current = (--current < 0) ? count + current : current;
+                }
             }
 
             if (dropDown?.OwnerItem is not null && dropDown.OwnerItem.IsInDesignMode)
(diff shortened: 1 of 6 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
```diff
@@ -2615,6 +2615,11 @@ private void HandleMouseDown(MouseEventArgs e)
             OnMouseDown(e);
             RaiseMouseEvent(s_mouseDownEvent, e);
         }
+
+        if (ParentInternal is not null && !ParentInternal.IsDropDown && Enabled)
+        {
+            ParentInternal.ResetLastActiveToolTipItem(this);
```
v-elnovikova, line 2621: I would suggest to move this call to `ToolStripButton` class. `ToolStripItem` is a base class for all ToolStip items, including ToolStripLabel, ToolStripGrip, ToolStripControlHost and others. All of these classes inherit ToolStripItem behavior.

dmitrii-drobotov: Yeah, I'm also not sure we should change it this way. @SimonZhao888 - is it possible to just restore behavior from .NET Framework and .NET Core 3.1 and 5, where pressing Left key would move focus to the last button? Also, since it's a regression, you can try to find a PR that introduced it, and see if there was anything wrong and fix that. My guess would be that it started to reproduce after work on https://github.com/dotnet/winforms/issues/5795.

SimonZhao888: Yeah, thanks @v-elnovikova and @dmitrii-drobotov, I'll think about this a bit more.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
```diff
@@ -2191,7 +2191,7 @@ internal byte GetMouseId()
         // backward direction entering the toolstrip, it means that the first
         // toolstrip item should be selected irrespectively TAB or SHIFT+TAB
         // is pressed.
-        start ??= GetStartItem(forward, dropDown is not null);
+        start ??= GetStartItem(forward, dropDown is not null, arrowDirection);
```
Tanya-Solyanik, line 2194: nitpick: please add a parameter name before `dropDown is not null`

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
```diff
@@ -2165,7 +2165,7 @@ internal byte GetMouseId()
                 return GetNextItemHorizontal(start, forward: true);
             case ArrowDirection.Left:
                 bool forward = LastKeyData == Keys.Tab || (TabStop && start is null);
-                return GetNextItemHorizontal(start, forward);
+                return GetNextItemHorizontal(start, forward, ArrowDirection.Left);
```
dmitrii-drobotov, line 2168: Ideally, we wouldn't need to introduce a new parameter but instead make `forward` as `false` in this case. Did you investigate if we can modify its condition like in https://github.com/dotnet/winforms/pull/8588? Perhaps something like `bool forward = LastKeyData == Keys.Tab || (TabStop && start is null && LastKeyData != Keys.Left);` could work?

Tanya-Solyanik, line 2168: Could you please clarify why the `forward` parameter is not sufficient in the case of the Left arrow only?

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
```diff
@@ -2165,7 +2165,7 @@ internal byte GetMouseId()
                 return GetNextItemHorizontal(start, forward: true);
             case ArrowDirection.Left:
                 bool forward = LastKeyData == Keys.Tab || (TabStop && start is null);
```
Tanya-Solyanik, line 2167: Could you please add a comment that explains scenario that this test covers.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
SimonZhao888: Yes, it works very well, thanks this is definitely the best correction.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
SimonZhao888: Since dmitrii offered a better revision, I restore my commits before, and made a new submission.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
SimonZhao888: In the case of https://github.com/dotnet/winforms/issues/9181, the TabStop = true, and start item is always null, the forward variable will be true, which is the same as ArrowDirection.Right, resulting in the same behavior of Left/Gight events, so we need to update it.

dmitrii-drobotov: Nice! Would it be possible to also add a unit test for it?

SimonZhao888: Done.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
dkazennov: I believe we can add a direct link to the issue. Like we do with tests like `DesignBehaviorsTests_can_DragDrop_ToolboxItem` for example.

v-elnovikova (approved):

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripTests.cs
```diff
@@ -7234,6 +7234,30 @@ public void ToolStripScrollButton_Arrows_Size_ReturnsExpected(string resourceNam
         }
     }
 
+    [WinFormsFact]
+    public void ToolStrip_GetNextItem_ItemsBackwardExpected()
+    {
+        // Regression test for https://github.com/dotnet/winforms/issues/9181, and it verifies that setting TabStop=true,
+        // When typing Right keyboard, the next focus position is first item on the left,
```
Tanya-Solyanik, line 7241: Suggestion: `Right` -> `Right arrow`

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripTests.cs
```diff
@@ -7234,6 +7234,30 @@ public void ToolStripScrollButton_Arrows_Size_ReturnsExpected(string resourceNam
...
+        // When typing Left keyboard, the next focus position is first item on the Right.
+
+        using ToolStrip toolStrip = new() { TabStop = true, Width = 300 };
+        using ToolStripMenuItem toolStripMenuItem1 = new();
+        using ToolStripMenuItem toolStripMenuItem2 = new();
+        using ToolStripMenuItem toolStripMenuItem3 = new();
+        toolStrip.Items.AddRange(new ToolStripItem[] { toolStripMenuItem1, toolStripMenuItem2, toolStripMenuItem3 });
+
+        toolStrip.TestAccessor().Dynamic.LastKeyData = Keys.Left;
+        ToolStripItem previousToolStripItem1 = toolStrip.GetNextItem(null, ArrowDirection.Left);
```
Tanya-Solyanik, line 7251: Please add parameter name before the literal (null)

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripTests.cs
```diff
@@ -7234,6 +7234,30 @@ public void ToolStripScrollButton_Arrows_Size_ReturnsExpected(string resourceNam
...
+        Assert.Equal(toolStrip.Items[2], previousToolStripItem1);
+
+        toolStrip.TestAccessor().Dynamic.LastKeyData = Keys.Right;
```
Tanya-Solyanik, line 7254: Do we need a test where LastKeyData and ArrowDirection don't match?

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
```diff
@@ -2164,7 +2164,7 @@ internal byte GetMouseId()
             case ArrowDirection.Right:
                 return GetNextItemHorizontal(start, forward: true);
             case ArrowDirection.Left:
-                bool forward = LastKeyData == Keys.Tab || (TabStop && start is null);
+                bool forward = LastKeyData == Keys.Tab || (TabStop && start is null && LastKeyData != Keys.Left);
```
Tanya-Solyanik, line 2167: Do we have the same issue with up/down arrows?

Consider adding a vertical toolstrip to the test app if we don't have one already, this could be done in another PR.

Tanya-Solyanik, line 2167: Nice collaboration, @v-elnovikova @dmitrii-drobotov  and @SimonZhao888! this is much more simple!

Tanya-Solyanik (commented):
Looks good!

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
SimonZhao888: There is no issue with up/down arrows.
I'll check the tests list.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripTests.cs
SimonZhao888: I think LastKeyData only affects ArrowDirection.Left, if we want to add a new test case, do we need to consider adding a comprehensive test case that can test the left arrow key？Other related test cases such as [ToolStrip_GetNextItem_CyclesForwardExpected](https://github.com/ArtemTatarinov/winforms/blob/ad7773d3e9f7ad2a29b673d759567c3745b1e0ae/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripTests.cs#L4799C1-L4817C10 )

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripTests.cs
SimonZhao888: Done.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripTests.cs
SimonZhao888: Since LastKeyData  only affects on ArrowDirection.Left, so I removed this line:
`toolStrip.TestAccessor().Dynamic.LastKeyData = Keys.Right;`

SimonZhao888:
Hi @Tanya-Solyanik, Amy-Li03 tested this PR. issue https://github.com/dotnet/winforms/issues/9181 has been fixed.

SimonZhao888:
Here's another question, the behavior of the MenuStrip control is the same from .NET Framework to .NET Core 8.0, but it's different with ToolStrip, do we need to pay attention to this?

Tanya-Solyanik (approved):

PR #9393, opened 2023-06-29
Refactor `LayoutEventArgs` to use a `WeakReference` for the `AffectedComponent` property with a backing field

### LINKED ISSUE #165: Control can't be properly GC-collected after disposing because of it referenced via the System.Windows.Forms.Control.cachedLayoutEventArgs field
The same bug is reproducible with .NET Framework but it seems it has no chance to be ever fixed.
But the .NET Core looks like a good place for introducing some improvements in this area.

### Problem description
Each instance of the `System.Windows.Forms.Control` type contains a private member variable of the  `LayoutEventArgs` type - **cachedLayoutEventArgs** . And, the `LayoutEventArgs` instance typically contains a reference to some specific control. 
Sometimes, the `cachedLayoutEventArgs` field is not cleared when the child control disposing of does not affect the layout process of the parent control due to some reasons.

### Reproduction
Here are the minimal reproduction-steps:
1. Create a form with two buttons.
2. Add the following code for the corresponding `Button.Click` handlers:
```c#
void btnOpenView_Click(object sender, System.EventArgs e) {
    var view = new Panel() { BackColor = Color.Red };
    view.Name = "View";
    view.Bounds = new Rectangle(100, 100, 100, 100);
    view.Parent = this;
}
void btnCloseView_Click(object sender, System.EventArgs e) {
    SuspendLayout();
    var view = this.Controls.Find("View", false)[0];
    if(view != null)
        view.Dispose();
    ResumeLayout(false);
}
```
3. Press the "Open View" button - red panel appears.
4. Press the "Close View" button - red panel disappears.

**Actual behavior:**
The panel is still referencing via the `System.Windows.Forms.Control.cachedLayoutEventArgs` field at the form level. As result, it can't be GC-collected properly and still in memory.

**Expected behavior:**
The panel should not be referenced.

### Proposal for Fix
It looks like we should use the `WeakReference` when caching the `LayoutEventArgs`.
As an alternative solution, we can validate the `cachedLayoutEventArgs` value on child control removing and clear the problematical field.

elachlan:
Attempting to fix #165

Which is a memory leak in `System.Windows.Forms.Control.CachedLayoutEventArgs`.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9393)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/LayoutEventArgs.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/LayoutEventArgs.cs
@@ -8,9 +8,11 @@ namespace System.Windows.Forms;
 
 public sealed class LayoutEventArgs : EventArgs
 {
+    private readonly WeakReference<IComponent>? _affectedComponent;
+
     public LayoutEventArgs(IComponent? affectedComponent, string? affectedProperty)
     {
-        AffectedComponent = affectedComponent;
+        _affectedComponent = affectedComponent is not null ? new(affectedComponent) : null;
         AffectedProperty = affectedProperty;
     }
 
@@ -19,7 +21,15 @@ public sealed class LayoutEventArgs : EventArgs
     {
     }
 
-    public IComponent? AffectedComponent { get; }
+    public IComponent? AffectedComponent
+    {
+        get
+        {
+            IComponent? target = null;
+            _affectedComponent?.TryGetTarget(out target);
+            return target;
+        }
+    }
 
     public Control? AffectedControl => AffectedComponent as Control;
 
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/LayoutEventArgsTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/LayoutEventArgsTests.cs
@@ -55,7 +55,7 @@ public class LayoutEventArgsTests
         static LayoutEventArgs CreateAndDisposeControl()
         {
             // Setup TableLayoutPanel and Panel
-            TableLayoutPanel tableLayoutPanel = new()
+            using TableLayoutPanel tableLayoutPanel = new()
             {
                 ColumnCount = 2,
                 RowCount = 2,
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/LayoutEventArgsTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/LayoutEventArgsTests.cs
@@ -3,6 +3,8 @@
 // See the LICENSE file in the project root for more information.
 
 using System.ComponentModel;
+using System.Reflection;
+using System.Runtime.CompilerServices;
 
 namespace System.Windows.Forms.Tests;
 
@@ -40,6 +42,50 @@ public class LayoutEventArgsTests
         Assert.Equal(affectedControl, e.AffectedComponent);
         Assert.Equal(affectedControl, e.AffectedControl);
         Assert.Equal(affectedProperty, e.AffectedProperty);
+    }
+
+    [WinFormsFact]
+    public void LayoutEventArgs_DoesNotRootControl()
+    {
+        LayoutEventArgs layoutEventArgs = CreateAndDisposeControl();
+        GC.Collect(GC.MaxGeneration, GCCollectionMode.Forced, blocking: true);
+        Assert.Null(layoutEventArgs.AffectedComponent);
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static LayoutEventArgs CreateAndDisposeControl()
+        {
+            // Setup TableLayoutPanel and Panel
+            TableLayoutPanel tableLayoutPanel = new()
+            {
+                ColumnCount = 2,
+                RowCount = 2,
+                AutoScroll = true,
+            };
+
+            tableLayoutPanel.Controls.Add(new Panel()
+            {
+                Name = "Panel",
+                Dock = DockStyle.Fill
+            }, 0, 0); // Add the panel to the first cell
+
+            // Force layout update
+            tableLayoutPanel.PerformLayout();
+
+            var fieldInfo = typeof(Control).GetField("_cachedLayoutEventArgs", BindingFlags.NonPublic | BindingFlags.Instance);
+
+            // Suspend layout
+            tableLayoutPanel.SuspendLayout();
+            using (Panel panel = (Panel)tableLayoutPanel.Controls.Find("Panel", false).First())
+            {
+                tableLayoutPanel.Controls.Remove(panel);
+                panel.Dispose();
+            }
+
+            LayoutEventArgs layoutEventArgs = (LayoutEventArgs)fieldInfo.GetValue(tableLayoutPanel);
+
+            tableLayoutPanel.ResumeLayout();
+            return layoutEventArgs;
+        }
     }
 
     private class SubComponent : IComponent
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/LayoutEventArgsTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/LayoutEventArgsTests.cs
@@ -71,8 +71,6 @@ public class LayoutEventArgsTests
             // Force layout update
             tableLayoutPanel.PerformLayout();
 
-            var fieldInfo = typeof(Control).GetField("_cachedLayoutEventArgs", BindingFlags.NonPublic | BindingFlags.Instance);
-
             // Suspend layout
             tableLayoutPanel.SuspendLayout();
             using (Panel panel = (Panel)tableLayoutPanel.Controls.Find("Panel", false).First())
@@ -81,7 +79,9 @@ public class LayoutEventArgsTests
                 panel.Dispose();
             }
 
-            LayoutEventArgs layoutEventArgs = (LayoutEventArgs)fieldInfo.GetValue(tableLayoutPanel);
+            // Check the cachedLayoutEventArgs
+            ITestAccessor tableLayoutPanelTestAccessor = tableLayoutPanel.TestAccessor();
+            LayoutEventArgs layoutEventArgs = (LayoutEventArgs)tableLayoutPanelTestAccessor.Dynamic._cachedLayoutEventArgs;
 
             tableLayoutPanel.ResumeLayout();
             return layoutEventArgs;
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/LayoutEventArgsTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/LayoutEventArgsTests.cs
@@ -3,7 +3,6 @@
 // See the LICENSE file in the project root for more information.
 
 using System.ComponentModel;
-using System.Reflection;
 using System.Runtime.CompilerServices;
 
 namespace System.Windows.Forms.Tests;


elachlan:
@Olina-Zhang could you team please test this fix based on the reproducible steps in the related issue?

Tanya-Solyanik:
@elachlan  - are you sure that this bug still repros?

elachlan:
No, I am not sure it still repos. I am unsure how to test it.

Amy-Li03:
I am also unsure how to verify this issue, yesterday I use WinDbg check the Panel type before and after disposing in issue: https://github.com/dotnet/winforms/issues/165#issuecomment-1612807436, didn't see any difference, now I don't know how to test this.

elachlan:
Are you saying there is no difference between the tests using this PR and the main branch?

weltkantt suggested that we modify the test to call the following after `dispose()`.
```csharp
GC.Collect(); // clear normal weak references, collect normal objects, mark objects with destructor for finalization
GC.WaitForPendingFinalizers(); // wait until the finalizer callbacks have been processed
GC.Collect(); // clear long weak references, removes finalizeable objects from memory
// now take e.g. a memory snapshot in VS and look at whats still holding the objects alive
```

As far as I understand it. In the main branch this should NOT have an affect, but in this PR you should see the panel be removed from memory.

Amy-Li03:
@elachlan When modifying the test to call the following after dispose(), the tests are different using this PR and the main branch, In the main branch, the Panel still exist in memory, but with this PR, Panel was removed from memory.
```
GC.Collect(); // clear normal weak references, collect normal objects, mark objects with destructor for finalization
GC.WaitForPendingFinalizers(); // wait until the finalizer callbacks have been processed
GC.Collect(); // clear long weak references, removes finalizeable objects from memory
// now take e.g. a memory snapshot in VS and look at whats still holding the objects alive
```
Test result in main branch:
image

Test result with this PR:
image

elachlan:
@Amy-Li03 Thank you so much!!!

Tanya-Solyanik:
@dreddy-work  - it seems to me risky to dispose of a control referenced by LayoutEventArgs. Can thihs field - https://github.com/dotnet/winforms/blob/39c8004591ffe50a4e3bac61520a068229cbd1bc/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs#L313 be used after the affected control is disposed.

elachlan:
> A cleaner way is to clear the cached layout arguments when the control is unparented.

`Control.Dispose` ends up calling `AssignParent(null)` via the ControlCollection. I guess we do it in `AssignParent` when the value is null?

> if the layout is suspended at the time control is unparented, then _cachedLauoutEventArgs should be cleared on resume layout.

I have to read through the code flow a bit, its quite complex. If you have suggests as to where we can make the change that would be fantastic. Also, should we keep the backing field with the weakreference or remove it?

Tanya-Solyanik:
>Also, should we keep the backing field with the weakreference or remove it?

This requires an investigation. `_cachedLayoutEventArgs` field is used in the layout pass. My concern is that the code tests for this field being null, but not the child control (AffectedControl) being null, when applying these args to the layout. After a quick search I found only a single place where this field is used - in TableLayout - https://github.com/dotnet/winforms/blob/70f44451f0eb438000e8212c4c1fe6de51bb331c/src/System.Windows.Forms/src/System/Windows/Forms/Layout/TableLayout.cs#L67. So specific investigation steps are to find when this code is invoked and if it makes sense to invoke it after the control is disposed.

elachlan:
Sorry If I am not understanding correctly. But `ProcessSuspendedLayoutEventArgs` only checks `AffectedProperty` not `AffectedComponent`/`AffectedControl`. So it shouldn't be an issue?

Tanya-Solyanik:
>So it shouldn't be an issue?

I hope so. Can we corrupt the layout if we clear the [cache](https://github.com/dotnet/winforms/blob/70f44451f0eb438000e8212c4c1fe6de51bb331c/src/System.Windows.Forms/src/System/Windows/Forms/Layout/TableLayout.cs#L1396) unnecessarily? It would help to build confidence if we could execute this code. Had you verified that this is the only place that uses it, I only ran a quick search.

elachlan:
> > So it shouldn't be an issue?
> 
> I hope so. Can we corrupt the layout if we clear the [cache](https://github.com/dotnet/winforms/blob/70f44451f0eb438000e8212c4c1fe6de51bb331c/src/System.Windows.Forms/src/System/Windows/Forms/Layout/TableLayout.cs#L1396) unnecessarily? It would help to build confidence if we could execute this code. Had you verified that this is the only place that uses it, I only ran a quick search.

I have no idea how to trigger this. Would it be a table layout on a form with a child control that gets removed inside a SuspendLayout/ResumeLayout?

Tanya-Solyanik:
Define a TableLayoutPanel parent container and place a panel into its cell as a starting point. See what properties cause clearing of the cache.

elachlan:
Ill try and make time this weekend. Apologies for the delay.

elachlan:
Here are the properties that would invalidate the cache. But the LayoutEventArgs are for itself.
https://github.com/dotnet/winforms/blob/e517336a469381c4e5ae926280bf5b347d3846a2/src/System.Windows.Forms/src/System/Windows/Forms/Layout/TableLayout.cs#L46-L60

`ControlCollection.Remove` calls `LayoutTransaction.DoLayout` which triggers `ProcessSuspendedLayoutEventArgs` regardless of the `SuspendLayout`.

I wrote this:
```csharp
[WinFormsFact]
    public void TableLayout_TestLayoutEventArgs_AffectedControl()
    {
        // Setup TableLayoutPanel and Panel
        TableLayoutPanel tableLayoutPanel = new()
        {
            ColumnCount = 2,
            RowCount = 2,
            AutoScroll = true,
        };

        tableLayoutPanel.Controls.Add(new Panel()
        {
            Name= "Panel",
            Dock = DockStyle.Fill
        }, 0, 0); // Add the panel to the first cell

        // Force layout update
        tableLayoutPanel.PerformLayout();

        // Suspend layout
        tableLayoutPanel.SuspendLayout();
        using (Panel panel = (Panel)tableLayoutPanel.Controls.Find("Panel", false).First())
        {
            tableLayoutPanel.Controls.Remove(panel);
            panel.Dispose();
        }

        tableLayoutPanel.ResumeLayout();

        GC.Collect();
        GC.Collect();

        var t = ""; // breakpoint here and check heap.
    }
```

I am fairly sure the `Panel` is kept alive by the `LayoutInfo` in `ContainerInfo`. This is invalidated via `ProcessSuspendedLayoutEventArgs` when it calls `ClearCachedAssignments` because the `AffectedProperty=="Parent"`. If I run the same test on Main without the LayoutEventArgs changes both TableLayoutPanel and Panel stay in memory.

Tanya-Solyanik:
Regarding your unit test - code that performs test steps should be in a helper method, and GC in another helper of in the main unit test for GC to work as expected in Debug and Retail.

I know that your fix ensures that the memory is freed. My concern is that this change introduces regression in the layout processing. Could you please put your code into an app to see if the TLP looks reasonable after the panel is removed.

elachlan:
I don't think it introduces a regression. Because the Panel is held by the tableLayoutPanel containerinfo until after `ResumeLayout` has completed. So all events are executed the same.

Tanya-Solyanik:
Makes sense! Thank you for the investigation! Could you please add your unit test to the change? Please ensure that the TLP under test is disposed and that the test steps are in a helper method, and GC is executed after the helper method completes.

elachlan:
No problem. Do you want it as an integration test or a unit test?

Tanya-Solyanik:
I was thinking unit test like this one - https://github.com/Epica3055/winforms/blob/0bcc6a410c11845329ee0297b788ed64b7e6e9bb/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/Control.ControlAccessibleObjectTests.cs#L1589 but with your TLP sample in place of the `CreateAndDisposeControl` method

elachlan:
Thanks! I'll get on it.

elachlan:
verified the test `LayoutEventArgs_DoesNotRootControl` fails without the `LayoutEventArgs` changes.

Tanya-Solyanik (dismissed):
Thank you!

elachlan:
Can you please check the test. I had to access `_cachedLayoutEventArgs` via reflection. I imagine there is a better way for tests?

elachlan:
@Tanya-Solyanik is this ready to merge or does the test need to change?

Tanya-Solyanik:
> I had to access _cachedLayoutEventArgs via reflection. I imagine there is a better way for tests?

You can use [TextAccessor](https://github.com/dotnet/winforms/blob/43c5f2b3ea81496793f0b14ae5d28877f593bfeb/src/Common/tests/TestUtilities/TestAccessor.cs#L14) in the test.

Tanya-Solyanik:
##[error]src\System.Windows.Forms\tests\UnitTests\System\Windows\Forms\LayoutEventArgsTests.cs(6,1): error IDE0005: (NETCORE_ENGINEERING_TELEMETRY=Build) Using directive is unnecessary. (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/style-rules/ide0005)

Tanya-Solyanik (approved):
Thank you!

Nora-Zhou01:
@Tanya-Solyanik   The result is now the same as the previous test in the master branch, the panel still exists in memory, is this the expected result?
image

elachlan:
@Nora-Zhou01 can you please provide your test? We have a unit test for this issue.

Nora-Zhou01:
@elachlan   I have created a new project result are consistent with the above PR results, please help me to check if my project is correct.  [WinFormsApp4.zip](https://github.com/dotnet/winforms/files/12182070/WinFormsApp4.zip)
test

elachlan:
Try this:
```csharp
private void btnRemoveViewFromTab_Click(object sender, EventArgs e)
{
	var view = tabPage2.Controls.Find("View", false)[0];
	if (view != null)
	{
		tabPage2.Controls.Remove(view);
		view.Dispose();
	}
   
}

private void ClickCollectButton_Click(object sender, EventArgs e)
{
	GC.Collect(GC.MaxGeneration, GCCollectionMode.Forced, blocking: true);
}
```

Nora-Zhou01:
@elachlan I tried the method you gave me, and the results are as follows, but they are different from the test results of the PR at the time, is this expected?

elachlan:
I am not sure. Are you breaking after the GC collect and then running the command?

Nora-Zhou01:
This is my step:

https://github.com/dotnet/winforms/assets/104609169/f01f10cf-b765-432a-a94e-6bb5714f5536

elachlan:
Exception shouldn't be happening.

Nora-Zhou01:
My test results as you said are the same as above, can you please provide me with a project if those steps are still missing?

elachlan:
So your code is this? You click add, remove, collect? there is no exception and the dump has a panel?
```csharp
public partial class Form1 : Form
    {
        public Form1()
        {
            InitializeComponent();
        }

        private void btnAddViewToTab_Click(object sender, EventArgs e)
        {
            var view = new Panel();
            view.Name = "View";
            view.Dock = DockStyle.Fill;
            view.Parent = tabPage2;
        }

		private void btnRemoveViewFromTab_Click(object sender, EventArgs e)
		{
			var view = tabPage2.Controls.Find("View", false)[0];
			if (view != null)
			{
				view.Dispose();
			}
		   
		}

		private void ClickCollectButton_Click(object sender, EventArgs e)
		{
			GC.Collect(GC.MaxGeneration, GCCollectionMode.Forced, blocking: true);
		}
	}
```

Nora-Zhou01:
> This is my step:
> 
>  9393new.mp4

Yes, the code is the same as yours and the steps are the same as here.

elachlan:
I think the steps are wrong. I don't think you should close the app after clicking the "collect" button.

Maybe @Olina-Zhang can help, I might be missing something obvious.

Olina-Zhang:
@elachlan @Nora-Zhou01 I tested it in the latest .Net 8.0 build: 8.0.100-rc.1.23375.11, it was fixed, compared with .Net 7.0 version
**Testing app:** 
[WinFormsApp37.zip](https://github.com/dotnet/winforms/files/12190795/WinFormsApp37.zip)
**.Net 7.0:**
image

**.Net 8.0 build: 8.0.100-rc.1.23375.11:**
image

Issue #9224, opened 2023-06-05
Update AccessibleObject owner references weak references that were missed in PR#9164


LeafShi1:
Update AccessibleObject owner references weak references that were missed in pr#[9164](https://github.com/dotnet/winforms/pull/9164)
After the AccessibleObject  was updated, the corresponding test case "ControlAccessibleObject_DoesNotRootControls_AllPublicControl" added from pr#[9215 ](https://github.com/dotnet/winforms/pull/9215) also need to be updated. 

- [x] CheckedListBox #9324 
- [x] ComboBox #9323 
- [x] DataGridView #9376 
- [x] DataGridViewComboBoxEditingControl #9376 
- [x] DataGridViewTextBoxEditingControl #9376 
- [x] FlowLayoutPanel PR #9246 
- [x] HScrollBar PR #9434 
- [x] Label  PR #9246 
- [x] LinkLabel PR #9246 
- [x] ListBox #9324 
- [x] ListView #9451 
- [x] MaskedTextBox PR #9400 
- [x] MonthCalendar #9454 
- [x] Panel PR #9246 
- [x] PropertyGrid #9452 
- [x] TableLayoutPanel PR #9246 
- [x] TabPage #9453 
- [x] TextBox #9400 
- [x] ToolStripContentPanel PR #9246 
- [x] TreeView #9458 
- [x] VScrollBar PR #9434

elachlan:
Could `ControlAccessibleObject` be refactored to pass in a type parameter? So instead of defining our own property for the given type, we just access the `_ownerControl` property or similar defined in `ControlAccessibleObject`. That would then enforce the weakreference for each usage.

elachlan:
Can anyone explain why are the PRs using `WeakReference` instead of `TryGetOwnerAs` pattern?

LeafShi1:
> Can anyone explain why are the PRs using `WeakReference` instead of `TryGetOwnerAs` pattern?

Some related class cannot invoke function TryGetOwnerAs directly, Like "`[ComboBox.ComboBoxItemAccessibleObjectCollection.cs](https://github.com/dotnet/winforms/blob/main/src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.ComboBoxItemAccessibleObjectCollection.cs)`", so we using `WeakReference`.  but the PRs are still in progress, we will try our best to use the defined methods.

Tanya-Solyanik:
@SimonZhao888  - please resolve this issue

[...]

PR #9400, opened 2023-06-30
Update control owning field of TextBox to using weak reference(#9224)

Epica3055:
Related to #9224 

## Proposed changes

- Update the control owning field of TextBox  to using weak references
- Enable the GC collect test for the TextBox  control in unit test cases

## Customer Impact

## Regression? 

-  No

## Test methodology 

- Unit tests
 

## Test environment(s) 

- dotnet 8.0.0-preview.7.23328.11

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9400)


src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.TextBoxBaseAccessibleObject.cs
```diff
@@ -11,13 +11,11 @@ public abstract partial class TextBoxBase
...
     internal class TextBoxBaseAccessibleObject : ControlAccessibleObject
     {
-        private readonly TextBoxBase _owningTextBoxBase;
         private TextBoxBaseUiaTextProvider? _textProvider;
 
         public TextBoxBaseAccessibleObject(TextBoxBase owner) : base(owner)
         {
-            _owningTextBoxBase = owner;
-            _textProvider = new TextBoxBaseUiaTextProvider(owner);
+            _textProvider = new TextBoxBaseUiaTextProvider(this);
```
Tanya-Solyanik, line 18: Text provider must work with the Control, and is accessing control from the Accessible object for all operations, it's easier to pass in the owner control as a parameter in the constructor.

Epica3055: yeah, but if TextBoxBaseUiaTextProvider has a strong reference to TextBox ,GC will not work.

Tanya-Solyanik: Declare reference to TextBoxBase in the TextBoxBaseUiaTextProvider as  a weak reference as well.

Epica3055: You mean like this ?
image

Epica3055: https://github.com/dotnet/winforms/issues/9224#issuecomment-1610652778

Tanya-Solyanik: Re Lachlan's comment: the reason to pick one pattern over the other is readability. If we are referencing the weakly referenced owner multiple times and we know that it remains valid for the lifetime of the object that references it (the accessible object or the UIA provider), which we know in this particular case, then code is more readable(and more performant) when it uses a field instead of a method call.

Tanya-Solyanik: Yes, your change in the image looks good. Additionally consider renaming this field to `_owner` to improve readability.

Tanya-Solyanik:
@Epica3055  - Do you have permissions to see the build errors in the CI run?

```
src\System.Windows.Forms\src\System\Windows\Forms\TextBoxBase.TextBoxBaseUiaTextProvider.cs(36,13): error CS8631: (NETCORE_ENGINEERING_TELEMETRY=Build) The type 'System.Windows.Forms.Control?' cannot be used as type parameter 'THwnd' in the generic type or method 'PInvoke.SendMessage<THwnd, TWParam, TLParam>(THwnd, MessageId, ref TWParam, ref TLParam)'. Nullability of type argument 'System.Windows.Forms.Control?' doesn't match constraint type 'IHandle<Windows.Win32.Foundation.HWND>'.
```

Epica3055:
@Tanya-Solyanik 

> Do you have permissions to see the build errors in the CI run.

Yes, I have that permission. I usually check the failed tests after I commit PR.

src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.TextBoxBaseUiaTextProvider.cs
```diff
@@ -73,52 +75,53 @@ public TextBoxBaseUiaTextProvider(TextBoxBase owner)
...
-            return new UiaTextRange(_owningTextBoxBase.AccessibilityObject, this, start, start);
+            return new UiaTextRange(Owner.AccessibilityObject, this, start, start);
         }
 
-        public override Rectangle RectangleToScreen(Rectangle rect) => _owningTextBoxBase.RectangleToScreen(rect);
+        public override Rectangle RectangleToScreen(Rectangle rect) => Owner is not null ? Owner.RectangleToScreen(rect) : Rectangle.Empty;
 
-        public override UiaCore.ITextRangeProvider DocumentRange => new UiaTextRange(_owningTextBoxBase.AccessibilityObject, this, start: 0, TextLength);
+        public override UiaCore.ITextRangeProvider? DocumentRange => Owner is not null ?
+                                    new UiaTextRange(Owner.AccessibilityObject, this, start: 0, TextLength) : null;
```
Tanya-Solyanik, line 102: Please use 4 spaces for indentation.

```
        public override UiaCore.ITextRangeProvider? DocumentRange => Owner is not null
            ? new UiaTextRange(Owner.AccessibilityObject, this, start: 0, TextLength) 
            : null;
```

Tanya-Solyanik (commented):
Looks good!

src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.TextBoxBaseAccessibleObject.cs
```diff
@@ -62,19 +63,20 @@ internal override bool IsPatternSupported(UIA patternId)
...
+                    (this.TryGetOwnerAs(out TextBoxBase? owningTextBoxBase) && !owningTextBoxBase.PasswordProtect) ? name : string.Empty;
             }
             set => base.Name = value;
         }
 
-        public override string? Value => !_owningTextBoxBase.PasswordProtect ? ValueInternal : SR.AccessDenied;
+        public override string? Value => this.TryGetOwnerAs(out TextBoxBase? owningTextBoxBase) && !owningTextBoxBase.PasswordProtect ? ValueInternal : SR.AccessDenied;
 
         protected virtual string ValueInternal
             => this.TryGetOwnerAs(out Control? owner) && owner.Text is { } text ? text : string.Empty;
```
Tanya-Solyanik, line 82: This file looks a bit inconsistent because we now use different names for locals in similar patterns. Compare line 79 and 82, both are accessing control that is described by this accessible object, but use different name for the local. Please unify the name to `owner`.

src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.TextBoxBaseUiaTextProvider.cs
```diff
@@ -350,26 +354,37 @@ public override void SetSelection(int start, int end)
...
-            PInvoke.SendMessage(_owningTextBoxBase, PInvoke.EM_SETSEL, (WPARAM)start, (LPARAM)end);
+            PInvoke.SendMessage(Owner, PInvoke.EM_SETSEL, (WPARAM)start, (LPARAM)end);
         }
 
         private RECT GetFormattingRectangle()
         {
-            Debug.Assert(_owningTextBoxBase.IsHandleCreated);
+            if (Owner is null)
+            {
+                return new RECT(0, 0, 0, 0);
```
Tanya-Solyanik, line 364:
```suggestion
                return default;
```

Tanya-Solyanik (commented):
Looks good!

Tanya-Solyanik (approved):
Thank you!

Issue #8231, opened 2022-11-18
Cache is never populated in ComponentDesigner.ShadowPropertyCollection.cs


Jericho:
### .NET version

8.0.0-alpha.1.22559.2

However, the situation I am reporting is not tied to a specific .NET version.

### Did it work in .NET Framework?

No

### Did it work in any of the earlier releases of .NET Core or .NET 5+?

no

### Issue description

The ShadowPropertyCollection class contains a private field called `_descriptors` which, if I understand the original intent correctly, is supposed to be used as a cache of `PropertyDescriptor`. The flow is supposed to be like this:
- the first-time a given property descriptor is requested, the value is calculated in the `private PropertyDescriptor GetShadowedPropertyDescriptor(string propertyName)` private method
- this value is added to the cache
- if the same property descriptor is subsequently requested the cached value is returned

However, what I observe when reading the code of this private method is that the calculated value is NOT added to the cache and therefore it is re-calculated each time the same descriptor is requested.

@elachlan, @dreddy-work and myself had a [discussion](https://github.com/dotnet/winforms/pull/8222#pullrequestreview-1185033949) about this and we concluded that it made more sense to raise a separate issue about this rather than simply fix it as part of #8222.

If the conclusion is that we want to fix the flawed logic, I will be happy to volunteer a PR.

### Steps to reproduce

To be clear, the current logic works without problem and returns the expected result. It's just that we pay a performance penalty if the same descriptor is requested multiple times.

elachlan:
Relevant code where `_descriptors` is initialized and checked, but never added to.
https://github.com/dotnet/winforms/blob/bc73fc2790fdce8ac7399eddf76824161bdd35cc/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.ShadowPropertyCollection.cs#L59-L67

Proposed fix by @Jericho:
```csharp
private PropertyDescriptor GetShadowedPropertyDescriptor(string propertyName)
{
    _descriptors ??= new();

    if (!_descriptors.TryGetValue(propertyName, out PropertyDescriptor property))
    {
        property = TypeDescriptor.GetProperties(_designer.Component.GetType())[propertyName];
        _descriptors.Add(propertyName, property);
    }

    return property;
}
```

elachlan:
@dreddy-work how does the team feel about this change?

dreddy-work:
> @dreddy-work how does the team feel about this change?

[...]

PR #9408, opened 2023-07-02
Enable nullability in ComponentDesigner

halgab:
## Proposed changes

- Enable nullability in `ComponentDesigner`

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9408)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.ShadowPropertyCollection.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.ShadowPropertyCollection.cs
@@ -1,8 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-#nullable disable
-
 namespace System.ComponentModel.Design;
 
 public partial class ComponentDesigner
@@ -13,8 +11,8 @@ public partial class ComponentDesigner
     protected sealed class ShadowPropertyCollection
     {
         private readonly ComponentDesigner _designer;
-        private Dictionary<string, object> _properties;
-        private Dictionary<string, PropertyDescriptor> _descriptors;
+        private Dictionary<string, object?>? _properties;
+        private Dictionary<string, PropertyDescriptor>? _descriptors;
 
         internal ShadowPropertyCollection(ComponentDesigner designer) => _designer = designer;
 
@@ -22,14 +20,14 @@ public partial class ComponentDesigner
         ///  Accesses the given property name. This will throw an exception if the property does not exist on the
         ///  base component.
         /// </summary>
-        public object this[string propertyName]
+        public object? this[string propertyName]
         {
             get
             {
                 ArgumentNullException.ThrowIfNull(propertyName);
 
                 // First, check to see if the name is in the given properties table
-                if (_properties is not null && _properties.TryGetValue(propertyName, out object existing))
+                if (_properties is not null && _properties.TryGetValue(propertyName, out object? existing))
                 {
                     return existing;
                 }
@@ -59,19 +57,19 @@ public partial class ComponentDesigner
         {
             _descriptors ??= new();
 
-            if (_descriptors.TryGetValue(propertyName, out PropertyDescriptor property))
+            if (_descriptors.TryGetValue(propertyName, out PropertyDescriptor? property))
             {
                 return property;
             }
 
-            return TypeDescriptor.GetProperties(_designer.Component.GetType())[propertyName];
+            return TypeDescriptor.GetProperties(_designer.Component.GetType())[propertyName]!;
         }
 
         /// <summary>
         ///  Returns true if the given property name should be serialized, or false if not.
         ///  This is useful in implementing your own ShouldSerialize* methods on shadowed properties.
         /// </summary>
-        internal bool ShouldSerializeValue(string propertyName, object defaultValue)
+        internal bool ShouldSerializeValue(string propertyName, object? defaultValue)
         {
             ArgumentNullException.ThrowIfNull(propertyName);
 
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.ShadowPropertyCollection.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.ShadowPropertyCollection.cs
@@ -34,7 +34,7 @@ public partial class ComponentDesigner
 
                 // Next, check to see if the name is in the descriptors table.  If it isn't, we will search the
                 // underlying component and add it.
-                PropertyDescriptor property = GetShadowedPropertyDescriptor(propertyName);
+                PropertyDescriptor property = GetShadowedPropertyDescriptor(propertyName) ?? throw new ArgumentException("The requested property does not exist", nameof(propertyName));
 
                 return property.GetValue(_designer.Component);
             }
@@ -53,7 +53,7 @@ public partial class ComponentDesigner
         /// <summary>
         ///  Returns the underlying property descriptor for this property on the component
         /// </summary>
-        private PropertyDescriptor GetShadowedPropertyDescriptor(string propertyName)
+        private PropertyDescriptor? GetShadowedPropertyDescriptor(string propertyName)
         {
             _descriptors ??= new();
 
@@ -62,7 +62,7 @@ public partial class ComponentDesigner
                 return property;
             }
 
-            return TypeDescriptor.GetProperties(_designer.Component.GetType())[propertyName]!;
+            return TypeDescriptor.GetProperties(_designer.Component.GetType())[propertyName];
         }
 
         /// <summary>
@@ -73,9 +73,14 @@ public partial class ComponentDesigner
         {
             ArgumentNullException.ThrowIfNull(propertyName);
 
-            return Contains(propertyName)
-                ? !Equals(this[propertyName], defaultValue)
-                : GetShadowedPropertyDescriptor(propertyName).ShouldSerializeValue(_designer.Component);
+            if (Contains(propertyName))
+            {
+                return !Equals(this[propertyName], defaultValue);
+            }
+
+            PropertyDescriptor shadowedPropertyDescriptor = GetShadowedPropertyDescriptor(propertyName) ?? throw new InvalidOperationException("Failed to retrieve the shadowed PropertyDescriptor");
+
+            return shadowedPropertyDescriptor.ShouldSerializeValue(_designer.Component);
         }
     }
 }
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
@@ -547,10 +547,16 @@ public partial class ComponentDesigner : ITreeDesigner, IDesignerFilter, ICompon
     /// <summary>
     ///  Provides a way for a designer to get services from the hosting environment.
     /// </summary>
-    protected virtual object? GetService(Type serviceType) => Component?.Site?.GetService(serviceType);
+    protected virtual object? GetService(Type serviceType) => Component.Site?.GetService(serviceType);
 
     internal T? GetService<T>() where T : class
         => GetService(typeof(T)) as T;
+
+    internal T GetRequiredService<T>() where T : class
+    {
+        ISite componentSite = Component.Site ?? throw new InvalidOperationException("Component should have a site");
+        return componentSite.GetRequiredService<T>();
+    }
 
     internal bool TryGetService<T>([NotNullWhen(true)] out T? service) where T : class
     {
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
@@ -484,29 +484,31 @@ public partial class ComponentDesigner : ITreeDesigner, IDesignerFilter, ICompon
         get
         {
             string? settingsKeyName = (string?)ShadowProperties[SettingsKeyName];
-            if (string.IsNullOrEmpty(settingsKeyName))
+            if (!string.IsNullOrEmpty(settingsKeyName))
             {
-                IComponent? rootComponent = GetService<IDesignerHost>()?.RootComponent;
+                return settingsKeyName;
+            }
 
-                if (Component is IPersistComponentSettings persistableComponent && rootComponent is not null)
+            IComponent? rootComponent = this.GetService<IDesignerHost>()?.RootComponent;
+
+            if (this.Component is IPersistComponentSettings persistableComponent && rootComponent is not null)
+            {
+                if (string.IsNullOrEmpty(persistableComponent.SettingsKey))
                 {
-                    if (string.IsNullOrEmpty(persistableComponent.SettingsKey))
+                    Debug.Assert(this.Component.Site is not null);
+                    if (rootComponent != persistableComponent)
                     {
-                        Debug.Assert(Component.Site is not null);
-                        if (rootComponent != persistableComponent)
-                        {
-                            settingsKeyName = $"{rootComponent.Site!.Name}.{Component.Site.Name}";
-                        }
-                        else
-                        {
-                            settingsKeyName = Component.Site.Name;
-                        }
-
-                        ShadowProperties[SettingsKeyName] = settingsKeyName;
+                        settingsKeyName = $"{rootComponent.Site!.Name}.{this.Component.Site.Name}";
+                    }
+                    else
+                    {
+                        settingsKeyName = this.Component.Site.Name;
                     }
 
-                    persistableComponent.SettingsKey = settingsKeyName;
+                    this.ShadowProperties[SettingsKeyName] = settingsKeyName;
                 }
+
+                persistableComponent.SettingsKey = settingsKeyName;
             }
 
             return settingsKeyName;
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
@@ -441,7 +441,6 @@ public partial class ComponentDesigner : ITreeDesigner, IDesignerFilter, ICompon
     /// <summary>
     ///  Invokes the get inheritance attribute of the specified ComponentDesigner.
     /// </summary>
-    [return: NotNullIfNotNull(nameof(toInvoke))]
     protected InheritanceAttribute? InvokeGetInheritanceAttribute(ComponentDesigner? toInvoke)
         => toInvoke?.InheritanceAttribute;
 
@@ -496,13 +495,14 @@ public partial class ComponentDesigner : ITreeDesigner, IDesignerFilter, ICompon
                 {
                     if (string.IsNullOrEmpty(persistableComponent.SettingsKey))
                     {
+                        Debug.Assert(Component.Site is not null);
                         if (rootComponent != persistableComponent)
                         {
-                            settingsKeyName = $"{rootComponent.Site!.Name}.{Component.Site!.Name}";
+                            settingsKeyName = $"{rootComponent.Site!.Name}.{Component.Site.Name}";
                         }
                         else
                         {
-                            settingsKeyName = Component.Site!.Name;
+                            settingsKeyName = Component.Site.Name;
                         }
 
                         ShadowProperties[SettingsKeyName] = settingsKeyName;
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
@@ -18,6 +18,7 @@ public partial class ComponentDesigner : ITreeDesigner, IDesignerFilter, ICompon
     private DesignerActionListCollection? _actionLists;
     private ShadowPropertyCollection? _shadowProperties;
     private bool _settingsKeyExplicitlySet;
+    private IComponent? _component;
 
     private protected const string SettingsKeyName = "SettingsKey";
 
@@ -34,18 +35,17 @@ public partial class ComponentDesigner : ITreeDesigner, IDesignerFilter, ICompon
 
     private protected virtual void UpdateTextualDefaultProperty()
     {
-        var component = Component;
-        if (component?.Site is { } site)
+        if (_component?.Site is { } site)
         {
-            PropertyDescriptor? defaultProperty = TypeDescriptor.GetDefaultProperty(component);
+            PropertyDescriptor? defaultProperty = TypeDescriptor.GetDefaultProperty(_component);
             if (defaultProperty is null)
             {
                 return;
             }
 
-            if (defaultProperty.TryGetValue(component, out string? currentValue) && string.IsNullOrEmpty(currentValue))
+            if (defaultProperty.TryGetValue(_component, out string? currentValue) && string.IsNullOrEmpty(currentValue))
             {
-                defaultProperty.SetValue(component, site.Name);
+                defaultProperty.SetValue(_component, site.Name);
             }
         }
     }
@@ -169,7 +169,7 @@ public partial class ComponentDesigner : ITreeDesigner, IDesignerFilter, ICompon
     /// <summary>
     ///  Gets or sets the component this designer is designing.
     /// </summary>
-    public IComponent Component { get; private set; } = null!;
+    public IComponent Component => _component ?? throw new InvalidOperationException("Designer is not initialized");
 
     /// <summary>
     ///  Gets the design-time verbs supported by the component associated with the designer.
@@ -356,10 +356,10 @@ public partial class ComponentDesigner : ITreeDesigner, IDesignerFilter, ICompon
         get
         {
             Debug.Assert(
-                Component is not null,
+                _component is not null,
                 "this.component needs to be set before this method is valid.");
 
-            return TryGetService(out IDesignerHost? host) && Component == host.RootComponent;
+            return TryGetService(out IDesignerHost? host) && _component == host.RootComponent;
         }
     }
 
@@ -368,7 +368,7 @@ public partial class ComponentDesigner : ITreeDesigner, IDesignerFilter, ICompon
     /// </summary>
     public virtual void Initialize(IComponent component)
     {
-        Component = component;
+        _component = component;
 
         // For inherited components, save off the current values so we can compute a delta.  We also do this for
         // the root component, but, as it is ALWAYS inherited, the computation of default values favors the
@@ -453,7 +453,7 @@ public partial class ComponentDesigner : ITreeDesigner, IDesignerFilter, ICompon
                 cs.ComponentRename -= new ComponentRenameEventHandler(OnComponentRename);
             }
 
-            Component = null!;
+            _component = null;
             _inheritedProps = null;
         }
     }
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
@@ -407,17 +407,15 @@ public partial class ComponentDesigner : ITreeDesigner, IDesignerFilter, ICompon
             PropertyDescriptor[] values = new PropertyDescriptor[properties.Count];
             properties.CopyTo(values, 0);
 
-            for (int i = 0; i < values.Length; i++)
+            foreach (PropertyDescriptor prop in values)
             {
-                PropertyDescriptor prop = values[i];
-
                 // Skip some properties
                 if (Equals(prop.Attributes[typeof(DesignOnlyAttribute)], DesignOnlyAttribute.Yes))
                 {
                     continue;
                 }
 
-                if (prop.SerializationVisibility == DesignerSerializationVisibility.Hidden && !prop.IsBrowsable)
+                if (prop is { SerializationVisibility: DesignerSerializationVisibility.Hidden, IsBrowsable: false })
                 {
                     continue;
                 }
@@ -628,9 +626,8 @@ public partial class ComponentDesigner : ITreeDesigner, IDesignerFilter, ICompon
         EventDescriptor?[] values = new EventDescriptor[events.Values.Count];
         events.Values.CopyTo(values, 0);
 
-        for (int i = 0; i < values.Length; i++)
+        foreach (EventDescriptor? evt in values)
         {
-            EventDescriptor? evt = values[i];
             if (evt is not null)
             {
                 events[evt.Name] = TypeDescriptor.CreateEvent(evt.ComponentType, evt, ReadOnlyAttribute.Yes);
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
@@ -592,7 +592,7 @@ public partial class ComponentDesigner : ITreeDesigner, IDesignerFilter, ICompon
     ///  Allows a designer to filter the set of member attributes the component it is designing will expose
     ///  through the TypeDescriptor object.
     /// </summary>
-    protected virtual void PostFilterAttributes(IDictionary? attributes)
+    protected virtual void PostFilterAttributes(IDictionary attributes)
     {
         // If this component is being inherited, mark it as such in the class attributes.
         // Also, set our member variable to ensure that what you get by querying through the TypeDescriptor and
@@ -618,7 +618,7 @@ public partial class ComponentDesigner : ITreeDesigner, IDesignerFilter, ICompon
     /// <summary>
     ///  Allows a designer to filter the set of events the component it is designing will expose through the TypeDescriptor object.
     /// </summary>
-    protected virtual void PostFilterEvents(IDictionary? events)
+    protected virtual void PostFilterEvents(IDictionary events)
     {
         // If this component is being privately inherited, we need to filter the events to make them read-only.
         if (events is null)
@@ -648,9 +648,9 @@ public partial class ComponentDesigner : ITreeDesigner, IDesignerFilter, ICompon
     /// <summary>
     ///  Allows a designer to filter the set of properties the component it is designing will expose through the TypeDescriptor object.
     /// </summary>
-    protected virtual void PostFilterProperties(IDictionary? properties)
+    protected virtual void PostFilterProperties(IDictionary properties)
     {
-        if (properties is null || _inheritedProps is null)
+        if (_inheritedProps is null)
         {
             return;
         }
@@ -691,7 +691,7 @@ public partial class ComponentDesigner : ITreeDesigner, IDesignerFilter, ICompon
     ///  Allows a designer to filter the set of member attributes the component it is designing will expose
     ///  through the TypeDescriptor object.
     /// </summary>
-    protected virtual void PreFilterAttributes(IDictionary? attributes)
+    protected virtual void PreFilterAttributes(IDictionary attributes)
     {
     }
 
@@ -699,7 +699,7 @@ public partial class ComponentDesigner : ITreeDesigner, IDesignerFilter, ICompon
     ///  Allows a designer to filter the set of events the component it is designing will expose through the
     ///  TypeDescriptor object.
     /// </summary>
-    protected virtual void PreFilterEvents(IDictionary? events)
+    protected virtual void PreFilterEvents(IDictionary events)
     {
     }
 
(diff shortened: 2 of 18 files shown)


src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
```diff
@@ -140,7 +140,7 @@ public virtual void InitializeNewComponent(IDictionary defaultValues)
     /// <summary>
     ///  Gets or sets the component this designer is designing.
     /// </summary>
-    public IComponent Component { get; private set; }
+    public IComponent Component { get; private set; } = null!;
```
halgab, line 143: This property is not set in the constructor. I chose to annotate it as non-nullable anyway for 2 reasons:
 - this way it matches the annotation of the implicitly implented `IDesigner` interface
 - if I didn't do it, it would have resulted in <ins>a lot</ins> of `!` everywhere in the code
Looking at typical ComponentDesigner  instantation, the `Initialize` method is expected to be called right after the object creation though, so I think it is the best we can do.

elachlan: Maybe use one of these:
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis

It might be more correct if we can track that its null until after initialize is called.

halgab: `MemberNotNull` is the only attribute that could have qualified here, but `Initialize` is called only once, so it would be pretty useless for the majority of the callsites. I can add it to clarify the intent though, even though as I've already set `Component` as not null that wouldn't do much.

halgab: Another approach we could have is to have a nullable backing field to this property, and have the property throw an exception of some sort when it's called too early. But there are some scenarios that try to accomodate for a null `Component`. I have no idea if those are legitimate though, as my understanding is that one shouldn't be using a `ComponentDesigner` in such case (uninitialised or disposed)

elachlan: I guess the team might just clarify within the XML comment and documentation. I'll let them chime in.

Trying to handle it via a new exception would be a breaking api change and I suspect they wouldn't be keen.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
```diff
@@ -56,26 +55,26 @@ protected bool Inherited
...
             return root == Component ? null : root;
         }
     }
 
     /// <summary>
-    ///  Gets or sets the inheritance attribute for this component.
+    ///  Gets the inheritance attribute for this component.
     /// </summary>
-    protected virtual InheritanceAttribute InheritanceAttribute
+    protected virtual InheritanceAttribute? InheritanceAttribute
```
dreddy-work, line 70: I do not see a reason why this would return null. `GetInheritanceAttribute()` is not returning null

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
dreddy-work: I see many checks in this file guarding against `Component ` is not null. Though scenarios are not clear, it indicates that Component can be null. We need to create an issue as a follow-up here. I am okay with no change in this regard in this PR, given it isn't altering any functional behavior with respect to null checks, but add created issue in the comment here.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
```diff
@@ -450,34 +449,37 @@ private void OnComponentRename(object sender, ComponentRenameEventArgs e)
...
-                            ShadowProperties[SettingsKeyName] = Component.Site.Name;
+                            settingsKeyName = Component.Site!.Name;
                         }
+
+                        ShadowProperties[SettingsKeyName] = settingsKeyName;
                     }
 
-                    persistableComponent.SettingsKey = ShadowProperties[SettingsKeyName] as string;
+                    persistableComponent.SettingsKey = settingsKeyName;
                     return persistableComponent.SettingsKey;
```
dreddy-work, line 478: - > return settingsKeyName; for consistency.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
```diff
@@ -626,25 +615,23 @@ protected virtual void PostFilterEvents(IDictionary events)
     /// <summary>
     ///  Allows a designer to filter the set of properties the component it is designing will expose through the TypeDescriptor object.
     /// </summary>
-    protected virtual void PostFilterProperties(IDictionary properties)
+    protected virtual void PostFilterProperties(IDictionary? properties)
```
dreddy-work, line 618: Why do you have to make it nullable here?

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.ControlDesignerAccessibleObject.cs
```diff
@@ -88,7 +86,7 @@ public override AccessibleStates State
 
         private AccessibleObject? GetDesignerAccessibleObject(Control.ControlAccessibleObject cao)
```
dreddy-work, line 87: So, parameter is nullable here.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/SplitContainerDesigner.cs
```diff
@@ -175,18 +174,16 @@ protected override void Dispose(bool disposing)
...
             Rectangle translatedBounds = BehaviorService.ControlRectInAdornerWindow(_splitterPanel1!);
             var panelDesigner = _designerHost?.GetDesigner(_splitterPanel1!) as SplitterPanelDesigner;
             OnSetCursor();
 
-            ControlBodyGlyph? bodyGlyph = null;
             if (panelDesigner is not null)
             {
                 // Create our glyph, and set its cursor appropriately.
-                bodyGlyph = new ControlBodyGlyph(translatedBounds, Cursor.Current, _splitterPanel1, panelDesigner);
+                ControlBodyGlyph bodyGlyph = new ControlBodyGlyph(translatedBounds, Cursor.Current, _splitterPanel1, panelDesigner);
```
dreddy-work, line 186:
```suggestion
                ControlBodyGlyph bodyGlyph = new(translatedBounds, Cursor.Current, _splitterPanel1, panelDesigner);
```

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/SplitContainerDesigner.cs
```diff
@@ -196,7 +193,7 @@ protected override ControlBodyGlyph GetControlGlyph(GlyphSelectionType selection
             if (panelDesigner is not null)
             {
                 // Create our glyph, and set its cursor appropriately.
-                bodyGlyph = new ControlBodyGlyph(translatedBounds, Cursor.Current, _splitterPanel2, panelDesigner);
+                ControlBodyGlyph bodyGlyph = new ControlBodyGlyph(translatedBounds, Cursor.Current, _splitterPanel2, panelDesigner);
```
dreddy-work, line 196:
```suggestion
                ControlBodyGlyph bodyGlyph = new(translatedBounds, Cursor.Current, _splitterPanel2, panelDesigner);
```

src/System.Windows.Forms.Primitives/src/System/ComponentModel/MemberDescriptorExtensions.cs
```diff
@@ -19,4 +19,26 @@ public static bool TryGetAttribute
         <[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicParameterlessConstructor | DynamicallyAccessedMemberTypes.PublicFields)] T>(
         this MemberDescriptor descriptor) where T : Attribute
         => descriptor?.Attributes[typeof(T)] as T;
+
+    public static bool TryGetValue<T>(this PropertyDescriptor descriptor, object? component, out T? value)
+    {
+        if (typeof(T).IsAssignableFrom(descriptor.PropertyType))
+        {
+            value = (T?)descriptor.GetValue(component);
```
dreddy-work, line 27: when true, Ideally, `value ` shoudl be filled with non-null. Can you check all call sites if value being null is okay?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
halgab: I annotated it as nullable because of at least one override expects to return null values ([here](https://github.com/dotnet/winforms/blob/33dcb0586b15cfd4ef144d78f36445777e8eb669/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/SplitterPanelDesigner.cs#L29)). I can update both files if that's more relevant

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
halgab: I did it because of the null check of `properties` below, but that's probably a poor reason

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.ControlDesignerAccessibleObject.cs
halgab: Actually it's never null. I'll update the line below accordingly

src/System.Windows.Forms.Primitives/src/System/ComponentModel/MemberDescriptorExtensions.cs
halgab: Currently it's okay, and looking at all places were we could use this helper it also looks okay

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
halgab: The architecture of this class doesn't seem to fit well with nullable annotations unfortunately... No solution looks satisfying, I just chose the one that looked the least frustrating to me. For the places were we guard against `Component` being null, we could also it in the same way as the classes that guard against some fields set in their constructor and nulled when disposed being null.
Anyway, I can certainly open a follow-up issue. Note that we're going to have the same problem for the (many) child classes, so the decision we take here is going to impact many files.

halgab: #9502

elachlan: For now I suggest `[MemberNotNull(nameof(Component))]` on `Initialize`.

halgab: Unfortunately it won't solve our problem. `Component` still has to be nullable as `Initialize` isn't called in the constructor

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
dreddy-work: Do you know if that override ever returns null? or was mistakenly marked?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
dreddy-work: that's the new check you added in this PR. Unless you have a scenario that proves properties can be null, i wouldn't alter functionality.  Let it throw and warn.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/SplitContainerDesigner.OrientationActionList.cs
```diff
@@ -54,8 +54,8 @@ private void OnOrientationActionClick(object? sender, EventArgs e)
                 orientationProp.SetValue(_ownerComponent, orientation);
             }
 
-            var actionUIService = (DesignerActionUIService)_owner.GetService(typeof(DesignerActionUIService));
-            actionUIService.Refresh(_ownerComponent);
+            DesignerActionUIService? actionUIService = _owner.GetService<DesignerActionUIService>();
+            actionUIService!.Refresh(_ownerComponent);
```
dreddy-work, line 58:
```suggestion
            DesignerActionUIService actionUIService = _owner.GetService<DesignerActionUIService>()!;
            actionUIService.Refresh(_ownerComponent);
```

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
dreddy-work: Addign @Tanya-Solyanik / @JeremyKuhne  to see if they have any alternative suggestions here.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
halgab: Well, it's not really clear to me. I can return `(InheritanceAttribute?)TypeDescriptor.GetAttributes(_splitterPanel.Parent)[typeof(InheritanceAttribute)]`, which theoretically can be null, but I don't know if that can really happen in practice here, and I don't know where to look...

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
halgab: Okay, I'll change this

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.ShadowPropertyCollection.cs
```diff
@@ -60,19 +58,19 @@ private PropertyDescriptor GetShadowedPropertyDescriptor(string propertyName)
         {
             _descriptors ??= new();
 
-            if (_descriptors.TryGetValue(propertyName, out PropertyDescriptor property))
+            if (_descriptors.TryGetValue(propertyName, out PropertyDescriptor? property))
```
sharwell, line 61: 💭 Maybe file a bug to consider avoiding the allocation here?

```cs
if (_descriptors?.TryGetValue(propertyName, out PropertyDescriptor? property) ?? false)
```

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.ShadowPropertyCollection.cs
```diff
@@ -60,19 +58,19 @@ private PropertyDescriptor GetShadowedPropertyDescriptor(string propertyName)
...
             {
                 return property;
             }
 
-            return TypeDescriptor.GetProperties(_designer.Component.GetType())[propertyName];
+            return TypeDescriptor.GetProperties(_designer.Component.GetType())[propertyName]!;
```
sharwell, line 66: 💭 How do we know this will be non-null?

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.ControlDesignerAccessibleObject.cs
```diff
@@ -27,8 +27,7 @@ public ControlDesignerAccessibleObject(ControlDesigner designer, Control control
 
         public override string? Description => _control.AccessibilityObject.Description;
 
-        private IDesignerHost DesignerHost
-            => _host ??= (IDesignerHost)_designer.GetService(typeof(IDesignerHost));
+        private IDesignerHost DesignerHost => _host ??= _designer.GetService<IDesignerHost>()!;
```
sharwell, line 30: 💡 Consider adding an internal method (or extension method) `GetRequiredService<T>`. Applies to all cases where `GetService<T>` is followed by `!`.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.ControlDesignerAccessibleObject.cs
```diff
@@ -38,15 +37,14 @@ private IDesignerHost DesignerHost
 
         public override AccessibleRole Role => _control.AccessibilityObject.Role;
 
-        private ISelectionService SelectionService
-            => _selectionService ??= _designer.GetService<ISelectionService>();
+        private ISelectionService? SelectionService => _selectionService ??= _designer.GetService<ISelectionService>();
```
sharwell, line 40: 💡 Consider filing an issue to make this backed by `Lazy<ISelectionService?>` so it doesn't have to recompute it every time if it's not present.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.ControlDesignerAccessibleObject.cs
```diff
@@ -88,7 +86,7 @@ public override AccessibleStates State
 
         private AccessibleObject? GetDesignerAccessibleObject(Control.ControlAccessibleObject cao)
         {
-            if (cao is null || cao.Owner is not { } owner)
+            if (cao.Owner is not { } owner)
```
sharwell, line 89: ❔ Are we certain the caller will never pass null here? It would be easy to make this `cao?.Owner` and preserve the original behavior.

src/System.Windows.Forms.Design/tests/UnitTests/System/ComponentModel/Design/ComponentDesignerTests.cs
```diff
@@ -170,7 +170,7 @@ public void ComponentDesigner_Children_GetWithInvalidDesignerHost_ReturnsEmpty(I
 
     public static IEnumerable<object[]> InheritanceAttribute_GetValidService_TestData()
     {
-        yield return new object[] { null, 3, 5 };
+        yield return new object[] { null, 2, 4 };
```
sharwell, line 173: 😕 Did this need to change?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
sharwell: The other option here is to make an explicit `IComponent? _component` field, and defined this property as:

```cs
public IComponent Component
{
  get => _component ?? throw new InvalidOperationException(/* message about cannot access before init */);
  set => _component = value;
}
```

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
```diff
@@ -409,7 +407,8 @@ private void InitializeInheritedProperties()
     /// <summary>
     ///  Invokes the get inheritance attribute of the specified ComponentDesigner.
     /// </summary>
-    protected InheritanceAttribute InvokeGetInheritanceAttribute(ComponentDesigner toInvoke)
+    [return: NotNullIfNotNull(nameof(toInvoke))]
+    protected InheritanceAttribute? InvokeGetInheritanceAttribute(ComponentDesigner? toInvoke)
         => toInvoke?.InheritanceAttribute;
```
sharwell, line 412: ❓ Why is `toInvoke.InheritanceAttribute` known to not be null, but above we check `this.InheritanceAttribute` for null?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
```diff
@@ -450,34 +449,36 @@ private void OnComponentRename(object sender, ComponentRenameEventArgs e)
...
             {
-                IComponent rootComponent = GetService<IDesignerHost>()?.RootComponent;
+                IComponent? rootComponent = GetService<IDesignerHost>()?.RootComponent;
 
                 if (Component is IPersistComponentSettings persistableComponent && rootComponent is not null)
                 {
                     if (string.IsNullOrEmpty(persistableComponent.SettingsKey))
                     {
-                        if (rootComponent is not null && rootComponent != persistableComponent)
+                        if (rootComponent != persistableComponent)
```
sharwell, line 465: 💡 Consider a debug assertion to remove the suppressions

```
Debug.Assert(Component.Site is not null);
```

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
halgab: Yes, that's exactly what @elachlan proposed in #9502. I didn't chose it in this PR because I wasn't sure the team would be okay with a new exception being thrown.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.ShadowPropertyCollection.cs
halgab: I could, but this needs revisiting anyway considering #8231. Maybe we could just track this over there.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.ShadowPropertyCollection.cs
halgab: This method is called [here](https://github.com/dotnet/winforms/blob/b3644e50288048afb2ef423a8b89ec3ffedb5c22/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.ShadowPropertyCollection.cs#L39-L38) and [here](https://github.com/dotnet/winforms/blob/b3644e50288048afb2ef423a8b89ec3ffedb5c22/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.ShadowPropertyCollection.cs#L79). If its result were null, the call sites would throw NREs.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.ControlDesignerAccessibleObject.cs
halgab: I can do that. The [other place](https://github.com/dotnet/winforms/blob/f033a225944dfab4f4034a3faa47342432fccd76/src/System.Windows.Forms.Primitives/src/System/ServiceExtensions.cs#L80) that defines such a method throws though. Should we do the same?

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.ControlDesignerAccessibleObject.cs
halgab: Yes. This method is called only at one location that [checks](https://github.com/dotnet/winforms/blob/72c439204e52acc2202db61a14d001a80a34b902/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.ControlDesignerAccessibleObject.cs#L73C14-L73C14) the argument for null before passing it here.

src/System.Windows.Forms.Design/tests/UnitTests/System/ComponentModel/Design/ComponentDesignerTests.cs
halgab: Yes... These are the number of times we expect to call `GetService(typeof(IInheritanceService))`. If you end up caching the value, this test fails.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
halgab: Good catch. I think it's a leftover of a previous version I had locally where I annotated `InvokeGetInheritanceAttribute` as null.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
halgab: Done

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
elachlan: @dreddy-work is the team okay with `InvalidOperationException`? There is precedent for it elsewhere and enforces the pattern. Additionally its a private setter. So its not like consumers check the property and set it in their own logic.

halgab:
FYI I'm waiting on this PR to be merged to go on and annotate the (many!) child classes of ComponentDesigner.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.ControlDesignerAccessibleObject.cs
```diff
@@ -87,7 +85,7 @@ public override AccessibleStates State
 
         private AccessibleObject? GetDesignerAccessibleObject(Control.ControlAccessibleObject cao)
```
Epica3055, line 86: It's recommended not to use abbreviation `cao` and multiple `svc`

src/System.Windows.Forms.Design/tests/UnitTests/System/ComponentModel/Design/ComponentDesignerTests.cs
```diff
@@ -196,7 +196,7 @@ public void ComponentDesigner_Children_GetWithInvalidDesignerHost_ReturnsEmpty(I
...
```
sharwell, line 199: 💡 If it's desirable to cache the possibly-null result of calling `GetService(typeof(IInheritanceService))`, you can use this method:

https://github.com/dotnet/roslyn/blob/0caa880a161057559d1dcaad890e4053f520136e/src/Compilers/Core/Portable/InternalUtilities/InterlockedOperations.cs#L75-L89

src/System.Windows.Forms.Design/tests/UnitTests/System/ComponentModel/Design/ComponentDesignerTests.cs
halgab: This whole test looks a bit strange to me. I'm not sure why we need to test the number of times we get into `GetService`. This looks like a pretty strange way of making sure we're doing what we expect, confusing means with ends, so I'd be inclined to forego this part entirely. Secondly, it checks for a scenario (the first argument being null) that looks like improper use of the API, and expects a consistent result out of it

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
```diff
@@ -484,34 +479,37 @@ private void OnComponentRename(object sender, ComponentRenameEventArgs e)
...
     ///  shadowing the SettingsKey so we can default it to be RootComponent.Name + "." + Control.Name
     /// </summary>
-    private string SettingsKey
+    private string? SettingsKey
     {
         get
         {
-            if (string.IsNullOrEmpty((string)ShadowProperties[SettingsKeyName]))
+            string? settingsKeyName = (string?)ShadowProperties[SettingsKeyName];
+            if (string.IsNullOrEmpty(settingsKeyName))
```
lonitra, line 487: Consider inverting this to reduce the nesting

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
```diff
@@ -484,34 +479,37 @@ private void OnComponentRename(object sender, ComponentRenameEventArgs e)
...
                 if (Component is IPersistComponentSettings persistableComponent && rootComponent is not null)
                 {
                     if (string.IsNullOrEmpty(persistableComponent.SettingsKey))
                     {
-                        if (rootComponent is not null && rootComponent != persistableComponent)
+                        Debug.Assert(Component.Site is not null);
+                        if (rootComponent != persistableComponent)
                         {
-                            ShadowProperties[SettingsKeyName] = $"{rootComponent.Site.Name}.{Component.Site.Name}";
+                            settingsKeyName = $"{rootComponent.Site!.Name}.{Component.Site.Name}";
```
lonitra, line 498: How do we know site is not null here?

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/SplitContainerDesigner.cs
```diff
@@ -267,7 +263,7 @@ protected override void OnPaintAdornments(PaintEventArgs pe)
 
     private void OnSplitContainer(object? sender, MouseEventArgs e)
     {
-        var svc = (ISelectionService)GetService(typeof(ISelectionService));
+        var svc = GetService<ISelectionService>()!;
```
lonitra, line 266: Use `GetRequiredService` here as well

lonitra (commented):
I added a few comments. Thank you for your patience here

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ToolStripPanelDesigner.cs
```diff
@@ -326,16 +326,16 @@ public override void Initialize(IComponent component)
             return;
         }
 
-        _panel = component as ToolStripPanel;
+        _panel = (ToolStripPanel)component;
```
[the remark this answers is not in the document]
halgab: Well, if there were, we would have NRE popping here and there. I don't see any other part of the code that can prove this, but it wouldn't make sense to use a `ToolStripPanelDesigner` to design something that's not a `ToolStripPanel`. Anyway, if you prefer that I make the code more defensive, I can do it

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
halgab: I have no idea... I just maintained the code as it was. As a side note: this (private) getter doesn't seem to be used anywhere...

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ToolStripContainerDesigner.cs
```diff
@@ -32,15 +32,15 @@ internal class ToolStripContainerDesigner : ParentControlDesigner
...
+    private IDesignerHost _designerHost => GetService<IDesignerHost>()!;
 
     /// <summary>
     ///  Shadow the <see cref="ToolStripContainer.TopToolStripPanelVisible"/> property at design-time
     ///  so that we only set the visibility at design time if the user sets it directly.
     /// </summary>
     private bool TopToolStripPanelVisible
     {
-        get => (bool)ShadowProperties[nameof(TopToolStripPanelVisible)];
+        get => (bool)ShadowProperties[nameof(TopToolStripPanelVisible)]!;
```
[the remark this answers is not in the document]
halgab: The thing is I don't know... What happened here is that Many designers were annotated after I opened this PR, and I had to deal with warnings coming from classes that didn't wait for the base class to be annotated. So I tried to solve those problems without spending too much time with them, but that explains why so many files are touched here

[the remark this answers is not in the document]
lonitra: Thanks for the explaination here. I am seeing that these shadow properties are set in `Initialize()` so I think it would be sufficient to add a comment that these are set to non-null value on initialization. Same for the rest of them.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ToolStripPanelDesigner.cs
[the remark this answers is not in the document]
lonitra: Good point. I think this is fine to leave as is.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.cs
Tanya-Solyanik: We don't work with settings file right now, we might have to put it in eventually. I would rather keep it, as the recent pattern was to pull code from the framework in.

halgab:
[This change](https://github.com/dotnet/winforms/pull/9408#discussion_r1271117226) breaks *a lot* of tests (33 on the last count). It looks like tests create designers without initializing them. We either need to update the tests to account for this change, or change the designers code to add targeted checks to a newly created `IsInitialized` protected field.

lonitra:
> [This change](https://github.com/dotnet/winforms/pull/9408#discussion_r1271117226) breaks **a lot** of tests (33 on the last count). It looks like tests create designers without initializing them. We either need to update the tests to account for this change, or change the designers code to add targeted checks to a newly created `IsInitialized` protected field.

Could you please update the tests? I don't think we should be creating designers and using them without first initializing them. It would also be good to add a test that showcases that it is expected to get an `InvalidOperationException` if trying to use designer without first initializing it

halgab:
Should we throw a NRE when passing null to the `Initialize` method? It's legal today, but doesn't make much sense now...

elachlan:
If we do. It should be documented. Unsure if it's a true breaking change, would it end up throwing later on anyway if you did pass null?

halgab:
@lonitra tests are fixed. I also had to modify `ControlDesigner` so that it's `Dispose` method wouldn't throw. It used a null `Component` to prevent double disposal. I had to use another field instead. It's the last commit.

lonitra:
> Should we throw a NRE when passing null to the `Initialize` method? It's legal today, but doesn't make much sense now...

Good call. We should throw `ArgumentNullException` when null is passed into `initialize` method. This would be considered a breaking change since it does not throw today. I'll get that documented once this is merged.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.cs
```diff
@@ -464,7 +464,7 @@ protected override void Dispose(bool disposing)
     {
         if (disposing)
         {
-            if (Control is not null)
+            if (_statusCommandUI is not null)
```
lonitra, line 467: Why did we choose to use `_statusCommandUI`? This doesn't seem very inuitive perhaps consider making a local helper method to avoid throws in dispose when checking `Control` for null. Something like:
```c#
bool IsControlNull()
{
   try 
   {
      _ = Control;
      return false;
   }
   catch (InvalidOperationException) 
   {
       return true;
   }
}
```

halgab: I used `_statusCommandUI` because it's the only field that it set in method `Initialize` besides `_component`, which we don't have access to here. I figured it was the most simple way to replicate current behaviour. If that's not okay, I can imagine the following changes:

- exit early of the `Dispose` method when `_statusCommandUI` is null. That might make things a bit clearer 
- introduce a `isDisposed` field

I'm not a fan of using a try catch for an exception we ourselves introduced. It looks a bit pythonic to me. If you insist I'll change the code that way.

lonitra: I see, could we introduce isDisposed field? This is more explicit and there seem to be other places we have this sort of pattern.

halgab: I looked a bit more, and while this pattern is used elsewhere, it's never in an actual `Dispose` implementation... I pushed another change which introduce a private protected property in ComponentDesigner so that designers beyond ControlDesigner also have an escape hatch for this pattern. Tell me what you think

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.ShadowPropertyCollection.cs
lonitra: Took another look, looks like it's valid that return may be null and we should annotate it as so. Let's update the callsite to throw an exception if return ends up being null. [this callsite](https://github.com/dotnet/winforms/blob/b3644e50288048afb2ef423a8b89ec3ffedb5c22/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.ShadowPropertyCollection.cs#L39-L38) seems to specifically call out it will throw an exception, if if the property does not exist on the base component, so an `ArgumentException` should be thrown here. The other callsite `InvalidOperationException` may be appropriate.

lonitra (commented):
Couple more comments, otherwise LGTM! 👍 Appreciate your diligence here.

lonitra (approved):
LGTM. Appreciate your awesome work here! 🥳

halgab:
Thanks! It's a huge step forward for the annotations of this project 🎊

lonitra:
breaking change doc: https://github.com/dotnet/docs/issues/38799

PR #9414, opened 2023-07-03
DateTimePicker has incorrect AccessKey and KeyboardShortcut accessibility properties when Text contains '&' character

### LINKED ISSUE #9281: DateTimePicker has incorrect AccessKey and KeyboardShortcut accessibility properties when Text contains '&' character
### .NET version

.NET 8

No

No

### Issue description

Follow-up on https://github.com/dotnet/winforms/pull/9253/files#r1227133642. 

When DateTimePicker has text with '&' character before another character, DateTimePickerAccessibleObject [interprets](https://github.com/dotnet/winforms/blob/d1c6829dfa2c26757efab792fd0bab012537c629/src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObject.cs#L38-L45) it as a mnemonic character — as if it's a shortcut to select it by Alt+mnemonic. Some controls like Label support it by adding an underscore for a mnemonic character and making it a shortcut. But DateTimePicker doesn't seem to support it, '&' is interpreted as a regular character without any special meaning, it doesn't create a shortcut. I didn't find a way to turn '&' into a mnemonic by escaping it with `\` or similar, and there is no documentation on Win32 DateTimePicker regarding mnemonic characters.

As a result, Narrator announces it as a shortcut, although it doesn't work, unlike a mnemonic shortcut from Label with `"&Date"` text:
datetimepicker-shortcut

AccessKey and KeyboardShortcut accessibility properties should be empty instead of "Alt+t":
image

### Steps to reproduce

1. Add a DateTimePicker to form
2. Set `Format` as `Custom`
3. Add `&` before any character in `CustomFormat` property, for example set `CustomFormat` as `'Date&Time' hh:mm dd/MM`

Sample app:
[WinFormsApp8.zip](https://github.com/dotnet/winforms/files/11734357/WinFormsApp8.zip)

Epica3055:
Fixes #9281 

## Proposed changes

- 
- fix DateTimePicker's KeyboardShortcut property
- add unit tests
- 

## Regression? 

- No

### Before

image

### After

image

## Test methodology 

- 
- Manually and Unit Test 
- 

 

## Test environment(s) 

8.0.0-preview.7.23330.6

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9414)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObject.cs
@@ -32,7 +32,7 @@ public partial class DateTimePicker
                     }
                 }
 
-                // WIn32 DTP does not interpret ampersand in its Text as an escape character for a mnemonic.
+                // Win32 DTP does not interpret ampersand in its Text as an escape character for a mnemonic.
                 return null;
             }
         }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObject.cs
@@ -32,6 +32,7 @@ public partial class DateTimePicker
                     }
                 }
 
+                // WIn32 DTP does not interpret ampersand in its Text as an escape character for a mnemonic.
                 return null;
             }
         }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObject.cs
@@ -32,18 +32,7 @@ public partial class DateTimePicker
                     }
                 }
 
-                string? baseShortcut = base.KeyboardShortcut;
-
-                if (baseShortcut is null || baseShortcut.Length == 0)
-                {
-                    char ownerTextMnemonic = WindowsFormsUtils.GetMnemonic(this.GetOwnerText(), convertToUpperCase: false);
-                    if (ownerTextMnemonic != '\0')
-                    {
-                        return $"Alt+{ownerTextMnemonic}";
-                    }
-                }
-
-                return baseShortcut;
+                return null;
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObject.cs
```diff
@@ -32,18 +32,7 @@ public DateTimePickerAccessibleObject(DateTimePicker owner) : base(owner)
...
-                {
-                    char ownerTextMnemonic = WindowsFormsUtils.GetMnemonic(this.GetOwnerText(), convertToUpperCase: false);
-                    if (ownerTextMnemonic != '\0')
-                    {
-                        return $"Alt+{ownerTextMnemonic}";
-                    }
-                }
-
-                return baseShortcut;
+                return null;
```
Tanya-Solyanik, line 35: Please add a comment that WIn32 DTP does not interpret ampersand in its Text as an escape character for a mnemonic.

Tanya-Solyanik (commented):
Looks good

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObjectTests.cs
```diff
@@ -342,4 +342,39 @@ public void DateTimePickerAccessibleObject_DoDefaultAction_IfHandleIsCreated_Ret
...
+        dateTimePicker1.Name = "dateTimePicker1";
+        dateTimePicker1.TabIndex = 0;
+
+        label1.AutoSize = true;
+        label1.Name = "label1";
+        label1.TabIndex = 1;
+        label1.Text = "&Date";
+
+        dateTimePicker2.Name = "dateTimePicker2";
+        dateTimePicker2.TabIndex = 2;
```
dmitrii-drobotov, line 366: Nit: some of these assignments can be removed because they don't affect keyboard shortcuts behavior.
```suggestion
        dateTimePicker1.CustomFormat = "'Date&Time' hh:mm dd/MM";
        dateTimePicker1.Format = DateTimePickerFormat.Custom;

        label1.Text = "&Date";
```

Epica3055: TabIndex  is related to PreviousLabel , so I think it can not be remove.

src/System.Windows.Forms/src/System/Windows/Forms/DateTimePicker.DateTimePickerAccessibleObject.cs
```diff
@@ -32,6 +32,7 @@ public DateTimePickerAccessibleObject(DateTimePicker owner) : base(owner)
                     }
                 }
 
+                // WIn32 DTP does not interpret ampersand in its Text as an escape character for a mnemonic.
```
Tanya-Solyanik, line 35: Oops, wrong case - Win32

dreddy-work (approved):

PR #9418, opened 2023-07-04
Enable nullability in some ToolStripManager members

gpetrou:
## Proposed changes

- Enable nullability in some ToolStripManager members.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9418)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripManager.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripManager.cs
@@ -63,24 +63,22 @@ public static partial class ToolStripManager
                 {
                     // Default to menu font
                     sysFont = SystemInformation.GetMenuFontForDpi(dpi);
-                    if (sysFont is not null)
-                    {
-                        // Ensure font is in pixels so it displays properly in the property grid at design time.
-                        if (sysFont.Unit != GraphicsUnit.Point)
-                        {
-                            retFont = ControlPaint.FontInPoints(sysFont);
-                            sysFont.Dispose();
-                        }
-                        else
-                        {
-                            retFont = sysFont;
-                        }
 
-                        s_defaultFontCache[dpi] = retFont;
+                    // Ensure font is in pixels so it displays properly in the property grid at design time.
+                    if (sysFont.Unit != GraphicsUnit.Point)
+                    {
+                        retFont = ControlPaint.FontInPoints(sysFont);
+                        sysFont.Dispose();
                     }
+                    else
+                    {
+                        retFont = sysFont;
+                    }
+
+                    s_defaultFontCache[dpi] = retFont;
                 }
 
-                return retFont!;
+                return retFont;
             }
             else
             {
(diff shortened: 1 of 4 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ToolStripManager.cs
```diff
@@ -54,14 +52,14 @@ internal static Font DefaultFont
...
+            Font? sysFont = null;
 
             // We need to cache the default fonts for the different DPIs.
             if (DpiHelper.IsPerMonitorV2Awareness)
             {
                 int dpi = CurrentDpi;
 
-                Font retFont = null;
+                Font? retFont = null;
                 if (s_defaultFontCache.TryGetValue(dpi, out retFont) == false || retFont is null)
```
halgab, line 63: if `s_defaultFontCache` can't hold null `Font`s, this can simply be
``
```suggestion
                if (!s_defaultFontCache.TryGetValue(dpi, out Font? retFont))
```

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripManager.cs
```diff
@@ -83,12 +81,12 @@ internal static Font DefaultFont
                     }
                 }
 
-                return retFont;
+                return retFont!;
```
halgab, line 84: This looks odd. I guess it's to accomodate for the fact that we check `sysFont` for null even though it can't be null?

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripManager.cs
```diff
@@ -120,7 +118,7 @@ internal static Font DefaultFont
                                 }
                             }
 
-                            return retFont;
+                            return retFont!;
```
halgab, line 121: same thing here

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripManager.cs
```diff
@@ -15,22 +13,22 @@ public static partial class ToolStripManager
     // WARNING: ThreadStatic initialization happens only on the first thread at class CTOR time.
     // use InitializeThread mechanism to initialize ThreadStatic members
     [ThreadStatic]
-    private static WeakRefCollection t_toolStripWeakArrayList;
+    private static WeakRefCollection? t_toolStripWeakArrayList;
```
halgab, line 16: Sorry if it's a dumb question, but couldn't `WeakRefCollection` be generic? I see it's an internal class used only here.

gpetrou: Sounds like you are talking about a different issue/PR? I wouldn't change that in this PR.

halgab: I just happened to notice that while looking at your PR. Of course it doesn't have to be changed here. I just took the opportunity to bring this forward as it would remove a few casts

dreddy-work:
@gpetrou, did you get chance to look at other comments from @halgab ?

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripManager.cs
gpetrou: Yes, this is to avoid a build warning.

gpetrou:
@dreddy-work I added an extra commit for the `if` statement.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripManager.cs
halgab: I understand that it's more impactful, but since the output of `SystemInformation.GetMenuFontForDpi` can't be null, isn't it cleaner to remove the null check of `sysFont` one line 69/66? That way you wouldn't have to change this line.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripManager.cs
```diff
@@ -83,12 +80,12 @@ internal static Font DefaultFont
...
                 }
 
-                return retFont;
+                return retFont!;
             }
             else
             {
                 // Threadsafe local reference
-                Font retFont = s_defaultFont;
+                Font? retFont = s_defaultFont;
```
dreddy-work, line 88: on line 104 below, null check on `sysFont ` is redundant. It cannot be null. If we remove that check, we could avoid `!` on line 120.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripManager.cs
dreddy-work: I was typing same and noticed the comment here :)

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripManager.cs
gpetrou: @dreddy-work I added a separate commit per your suggestion.

dreddy-work (dismissed):

dreddy-work:
@gpetrou can you merge this with main?

gpetrou:
> can you merge this with main?

@dreddy-work done.

dreddy-work (approved):

PR #9423, opened 2023-07-04
Small refactoring in CoreSwitches class

Lehonti:
## Proposed changes

- In the original `CoreSwitches` class there is a field, `s_perfTrack`, that is initialized the first time that the `PerfTrack` property is accessed. The problem is that this is not thread-safe. Moving the initialization of the property to the static constructor ensures that its initialization is made thread-safe at the runtime level, and that it's initialized only once.

## Customer Impact

- No direct customer impact

## Regression? 

- No

## Risk

- Low

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9423)


JeremyKuhne (changes requested):
If this class gets more switches, we would be initializing all of them even if we don't use them, which would be undesirable from a memory perspective. This one in particular is only used in debug builds.

The worst that could happen here is that we'd end up with an extra instance that would get collected eventually. It is extremely unlikely that we'd see that happen. In general, we try to avoid protecting ourselves from exceedingly rare cases that only result in extra allocations if it has other performance implications.

This can, however, be simplified to the `=>` syntax.

Lehonti:
Fair enough. I have made it so that only the syntax is changed

dreddy-work (approved):

PR #9432, opened 2023-07-05
Change `ContainerSelectorGlyph` to use Icon internally

### LINKED ISSUE #9421: Moving icon of GroupBox/Panel is not clear in DemoConsole application
### .NET version

.NET 8.0.100-preview.7.23330.16

Yes

Yes, it doesn't reproduce in .Net 7.0

### Issue description

In [DemoConsole](https://github.com/dotnet/winforms/tree/main/src/System.Windows.Forms/tests/IntegrationTests/DesignSurface/DemoConsole) test app of Winforms repo, the moving icon of GroupBox/Panel is not clear now.

**The result of Main branch for .Net 8.0:**
image

**The result of release/7.0 branch for .Net 7.0:**
image

### Steps to reproduce

elachlan:
Fixes #9421
Caused by #8358

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9432)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/Behavior/ContainerSelectorGlyph.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/Behavior/ContainerSelectorGlyph.cs
@@ -14,7 +14,7 @@ namespace System.Windows.Forms.Design.Behavior;
 internal sealed class ContainerSelectorGlyph : Glyph
 {
     private readonly Rectangle _glyphBounds;
-    private Icon? _glyph;
+    private readonly Icon _glyph;
     private readonly ContainerSelectorBehavior? _relatedBehavior;
 
     /// <summary>
@@ -25,6 +25,7 @@ internal sealed class ContainerSelectorGlyph : Glyph
     {
         _relatedBehavior = behavior;
         _glyphBounds = new Rectangle(containerBounds.X + glyphOffset, containerBounds.Y - (int)(glyphSize * .5), glyphSize, glyphSize);
+        _glyph = new Icon(typeof(ContainerSelectorGlyph), "MoverGlyph");
     }
 
     /// <summary>
@@ -40,17 +41,8 @@ internal sealed class ContainerSelectorGlyph : Glyph
     public override Cursor? GetHitTest(Point p)
         => _glyphBounds.Contains(p) || _relatedBehavior?.OkToMove == true ? Cursors.SizeAll : null;
 
-    private Icon MoveGlyph
-    {
-        get
-        {
-            _glyph ??= new Icon(typeof(ContainerSelectorGlyph), "MoverGlyph");
-            return _glyph;
-        }
-    }
-
     /// <summary>
     ///  Very simple paint logic.
     /// </summary>
-    public override void Paint(PaintEventArgs pe) => pe.Graphics.DrawIcon(MoveGlyph, _glyphBounds);
+    public override void Paint(PaintEventArgs pe) => pe.Graphics.DrawIcon(_glyph, _glyphBounds);
 }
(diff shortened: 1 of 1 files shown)


elachlan:
@dreddy-work here is the PR.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/Behavior/ContainerSelectorGlyph.cs
```diff
@@ -44,20 +44,14 @@ public override Rectangle Bounds
     /// </summary>
     public override Cursor? GetHitTest(Point p)
     {
-        if (_glyphBounds.Contains(p) || _relatedBehavior?.OkToMove == true)
-        {
-            return Cursors.SizeAll;
-        }
-
-        return null;
+        return _glyphBounds.Contains(p) || _relatedBehavior?.OkToMove == true ? Cursors.SizeAll : null;
```
dreddy-work, line 47: nit: make these expressional body statements.

elachlan:
Should I move the glyph initialization to the constructor? it would simplify the class and avoid creating the icon when paint is called?

dreddy-work:
> Should I move the glyph initialization to the constructor? it would simplify the class and avoid creating the icon when paint is called?

I would move initialization of it to the Paint override instead. that is the only place i see it is being used. If the Container glyph is not selected, It would avoid initialization of Icon.

dreddy-work (commented):
LGTM. thank you.

dreddy-work (approved):

dreddy-work:
@elachlan, in the designer, we made this transparent with following code. Can you capture how it looks with this chnage?
```CS

          static Bitmap LoadBitmap()
          {
              var glyph = new Bitmap(typeof(ContainerSelectorGlyph), "MoverGlyph.bmp");
              glyph.MakeTransparent();

              return glyph;
          }
```

elachlan:
> ```cs
> glyph.MakeTransparent();
> ```

I think I would have to change it back to a bitmap. Load it to an icon and convert it to bitmap. Which might make it not scale well. I'll give it a go.

elachlan:
image
It works.

elachlan:
The change works better with the conversion to bitmap with higher DPIs. I tested 200% on my desktop. The lines are clearer.

dreddy-work:
> The change works better with the conversion to bitmap with higher DPIs. I tested 200% on my desktop. The lines are clearer.
> Bitmap:

We could make Icon load right one with out converting to Bitmap. Did you try transparency?

dreddy-work:
Nevermind, i see code change. Is the screenshots after transparency?

elachlan:
> Nevermind, i see code change. Is the screenshots after transparency?

Yes. Bitmap is after transparency.

dreddy-work:
> Yes. Bitmap is after transparency.

Strange, i do not see difference between transparent and non-transparent here. WinForms designer is using `Cursor.SizeAll` from runtime(windows) one. I do not think we could change that in this. Let me run it by team before merge it.

elachlan:
Could you add waiting on team tag so we don't get lost?

dreddy-work:
With Internal offline discussion, we are taking this as-is for the runtime.

dreddy-work:
@elachlan, before i merge, can you capture a screenshot without `paint` override method in this? Wanted to check if it sets cursor from Windows in those cases ( just like designer)

elachlan:
There is no image on the left hand corner. But the mouse cursor icon changes to the system "move" icon (I am unable to get a screenshot of it).
image

It makes me wonder if we can use a win32 API call to get that icon instead.

Do you still want the `_glyph.MakeTransparent();` call?

elachlan:
Would it be a bad idea to use this?
https://learn.microsoft.com/en-us/windows/win32/menurc/about-cursors
image

dreddy-work:
> Do you still want the `_glyph.MakeTransparent();` call?

Given no difference with transparent, we can remove it.

>Would it be a bad idea to use this?

For this, we may want to try it as a separate PR and validate all the possible scenarios that goes through this call.

dreddy-work (approved):

PR #9439, opened 2023-07-07
Fix issue 9438: when setting GripStyle = hidden, the next focus position is not correct.

### LINKED ISSUE #9438: The next focus position is not correct when setting GripStyle = hidden and selecting on a ToolStripButton using mouse or typing "enter" keyboard and then typing "right or left" keyboard.
### .NET version

.NET SDK 8.0.100-preview.5.23274.2

No

No

For .NET core 3.1 and .NET 5.0, the behavior is same as .NET framework, setting GripStyle = hidden:
When typing Right keyboard, the next focus position is first item on the left.
When typing Left keyboard, the next focus position is first item on the Right.

For .NET 6.0/7.0/8.0, the behavior is: the next focus position is first item on the left, regardless of whether the right or left key is typed.

### Issue description

When selecting on a ToolStripButton using mouse or typing "enter" keyboard and then typing "right or left" keyboard, the next focus position is first item on the left, regardless of whether the right or left key is typed.
ToolStrip

### Steps to reproduce

1. Extract and open the test app. [TestApp.zip](https://github.com/dotnet/winforms/files/11973778/TestApp.zip)
2. Build and run this project.
3. Select on a ToolStripButton using mouse or typing "enter" keyboard.
4. Type "right or left" keyboard.

SimonZhao888:
Fixes #9438 

When the GripStyle = Visible, the ToolStrip control can correctly find out the first item of can be selected by Keyboard.
but when we set the GripStyle = hidden, the ToolStrip control will find out the second item of can be selected by Keyboard, not the first. It may be caused this issue.

## Proposed changes

- Modify the method of finding the starting item - GetStartItem(bool forward, bool isDropDown).
- Refine test cases to cover more scenarios.

## Customer Impact

- Whatever the value of GripStyle is, when typing Left/Right arrow keyboard, ToolStrip  can find the correct starting item.

## Regression? 

- Yes

## Risk

- Minimal

### Before
Setting GripStyle = hidden, when selecting on a ToolStripButton using mouse or typing "enter" keyboard and then typing "right or left" keyboard, the next focus position is first item on the left, regardless of whether the right or left key is typed.
image

### After
The behavior is the same as when GripStyle = Visible, when typing Right keyboard, the next focus position is first item on the left, and if typing Left keyboard, the next focus position is first item on the Right.
ToolStrip1

## Test methodology 

- Manual

## Test environment(s) 

- 8.0.0-preview.7.23323.5

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9439)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
@@ -2191,7 +2191,7 @@ public partial class ToolStrip : ScrollableControl, IArrangedElement, ISupportTo
         // backward direction entering the toolstrip, it means that the first
         // toolstrip item should be selected irrespectively TAB or SHIFT+TAB
         // is pressed.
-        start ??= GetStartItem(forward);
+        start ??= forward ? DisplayedItems[DisplayedItems.Count - 1] : DisplayedItems[0];
 
         int current = DisplayedItems.IndexOf(start);
         if (current == -1)
@@ -2233,8 +2233,6 @@ public partial class ToolStrip : ScrollableControl, IArrangedElement, ISupportTo
 
         return null;
     }
-
-    private ToolStripItem GetStartItem(bool forward) => forward ? DisplayedItems[DisplayedItems.Count - 1] : DisplayedItems[0];
 
     /// <remarks>
     ///  Helper function for GetNextItem - do not directly call this.
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
@@ -2245,7 +2245,7 @@ public partial class ToolStrip : ScrollableControl, IArrangedElement, ISupportTo
         {
             // For the drop-down up-directed loop should be preserved.
             // So if the current item is topmost, then the bottom item should be selected on up-key press.
-            return DisplayedItems[DisplayedItems.Count > 1 ? 1 : 0];
+            return (DisplayedItems.Count > 1 && DisplayedItems[0].CanKeyboardSelect) ? DisplayedItems[0] : DisplayedItems[DisplayedItems.Count > 1 ? 1 : 0];
         }
 
         return DisplayedItems[0];
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
@@ -2191,7 +2191,7 @@ public partial class ToolStrip : ScrollableControl, IArrangedElement, ISupportTo
         // backward direction entering the toolstrip, it means that the first
         // toolstrip item should be selected irrespectively TAB or SHIFT+TAB
         // is pressed.
-        start ??= GetStartItem(forward, dropDown is not null);
+        start ??= GetStartItem(forward);
 
         int current = DisplayedItems.IndexOf(start);
         if (current == -1)
@@ -2234,22 +2234,7 @@ public partial class ToolStrip : ScrollableControl, IArrangedElement, ISupportTo
         return null;
     }
 
-    private ToolStripItem GetStartItem(bool forward, bool isDropDown)
-    {
-        if (forward)
-        {
-            return DisplayedItems[DisplayedItems.Count - 1];
-        }
-
-        if (!isDropDown)
-        {
-            // For the drop-down up-directed loop should be preserved.
-            // So if the current item is topmost, then the bottom item should be selected on up-key press.
-            return (DisplayedItems.Count > 1 && DisplayedItems[0].CanKeyboardSelect) ? DisplayedItems[0] : DisplayedItems[DisplayedItems.Count > 1 ? 1 : 0];
-        }
-
-        return DisplayedItems[0];
-    }
+    private ToolStripItem GetStartItem(bool forward) => forward ? DisplayedItems[DisplayedItems.Count - 1] : DisplayedItems[0];
 
     /// <remarks>
     ///  Helper function for GetNextItem - do not directly call this.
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
```diff
@@ -2245,7 +2245,7 @@ private ToolStripItem GetStartItem(bool forward, bool isDropDown)
         {
             // For the drop-down up-directed loop should be preserved.
             // So if the current item is topmost, then the bottom item should be selected on up-key press.
-            return DisplayedItems[DisplayedItems.Count > 1 ? 1 : 0];
+            return (DisplayedItems.Count > 1 && DisplayedItems[0].CanKeyboardSelect) ? DisplayedItems[0] : DisplayedItems[DisplayedItems.Count > 1 ? 1 : 0];
```
dmitrii-drobotov, line 2248: Looks good, although I don't really understand why we needed to return `DisplayedItems[1]`. If there is a grip, it will be skipped here (it has `CanKeyboardSelect = false`): https://github.com/dotnet/winforms/blob/a719980fd85ca545cc8c9b382ce2f3298788c193/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs#L2225-L2230

So it seems this method could look like this: 
```
private ToolStripItem GetStartItem(bool forward) => forward ? DisplayedItems[DisplayedItems.Count - 1] : DisplayedItems[0];
```
@SimonZhao888 - can you try this implementation and see if it has any issues? I would suggest to simplify it this way, if it works correctly.

SimonZhao888: Very good suggestions, it works correctly.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
```diff
@@ -2191,7 +2191,7 @@ internal byte GetMouseId()
         // backward direction entering the toolstrip, it means that the first
         // toolstrip item should be selected irrespectively TAB or SHIFT+TAB
         // is pressed.
-        start ??= GetStartItem(forward, dropDown is not null);
+        start ??= GetStartItem(forward);
```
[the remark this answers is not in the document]
SimonZhao888: Done.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
```diff
@@ -2234,23 +2234,6 @@ internal byte GetMouseId()
...
-    private ToolStripItem GetStartItem(bool forward, bool isDropDown)
-    {
-        if (forward)
-        {
-            return DisplayedItems[DisplayedItems.Count - 1];
-        }
-
-        if (!isDropDown)
-        {
-            // For the drop-down up-directed loop should be preserved.
```
dreddy-work, line 2246: Is this scenario is still good with new change?

SimonZhao888: Yes, it works well.
dropdown

dmitrii-drobotov: For toolstrip dropdown nothing is changed actually, we're still returning `DisplayedItems[0]`.

dmitrii-drobotov (approved):
LGTM

dmitrii-drobotov:
@Olina-Zhang - could your team test this change? It would be good to cover navigation by left/right and up/down arrows within a ToolStrip, and also by Tab/Shift-Tab beetween multiple ToolStrips. Please test with different values of `ToolStrip.LayoutStyle`, `ToolStrip.GripStyle`, `ToolStrip.TabStop` and `ToolStrip.RightToLeft`.

dreddy-work (approved):

Tanya-Solyanik:
@SimonZhao888  - feel free to merge once test team signs off.

Olina-Zhang:
> @Olina-Zhang - could your team test this change?

Hi @dmitrii-drobotov, @SimonZhao888  During our testing, we found the GH issue: https://github.com/dotnet/winforms/issues/9438 is not a real issue, when GripStyle is hidden, .Net framework, .Net 5.0 and .Net 8.0 have same behavior: the next focus position keeps in the first item on the left, regardless of whether the right or left key is typed as the steps we did.  So we suggest this PR is not needed.

Note: the behaviors for GripStyle: hidden and GripStyle: visible are different. When GripStyle is visible, the result is the "After" described in https://github.com/dotnet/winforms/pull/9390#issue-1780286642

Olina-Zhang:
@Tanya-Solyanik @dmitrii-drobotov another interesting thing: for .Net framework different versions, the results are different: In .Net framework 4.7.2, set GripStyle = hidden, when typing Right keyboard, the next focus position is first item on the left, and if typing Left keyboard, the next focus position is first item on the Right. But in .Net framework 4.8 and 4.8.1, it is changed: set the GripStyle = hidden, the next focus position keeps in the first item on the left, regardless of whether the right or left key is typed. I am not clear it is based on some background or a regression. Regression is more likely since we should keep the same behavior for GripStyle = hidden and GripStyle = visible cases. What is your opinion?

Tanya-Solyanik:
@Olina-Zhang - thank you for the investigation. I agree that whether GripStyle is visible or not should not matter for this behavior. Since 4.7.2 is more common, I would go for compatibility with it. So we are keeping this PR. This is not a common scenario. Usually a click (enter) on a toolstrip button would move focus somewhere else in the real UI, so I wouldn't worry about the regression in 4.8.1 unless a customer request comes in.

@SimonZhao888  - you have permission to rerun CI tests in the pipeline now - expand details link and click the rerun button. Alternatively, you can type in the comment `/azp run` command. Also you can just `git rebase` your branch on the later main and push the updated branch out to trigger the tests.

Tanya-Solyanik (approved):

PR #9444, opened 2023-07-08
Enable nullability in RelatedPropertyManager

gpetrou:
## Proposed changes

- Enable nullability in RelatedPropertyManager.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9444)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/RelatedPropertyManager.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/RelatedPropertyManager.cs
@@ -24,11 +24,8 @@ internal class RelatedPropertyManager : PropertyManager
     {
         Debug.Assert(parentManager is not null, "How could this be a null parentManager.");
         _parentManager = parentManager;
-        _fieldInfo = parentManager.GetItemProperties().Find(dataField, true)!;
-        if (_fieldInfo is null)
-        {
+        _fieldInfo = parentManager.GetItemProperties().Find(dataField, true) ??
             throw new ArgumentException(string.Format(SR.RelatedListManagerChild, dataField));
-        }
 
         parentManager.CurrentItemChanged += new EventHandler(ParentManager_CurrentItemChanged);
         Refresh();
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/RelatedPropertyManager.cs
```diff
@@ -2,31 +2,30 @@
...
     private void Bind(BindingManagerBase parentManager, string dataField)
     {
         Debug.Assert(parentManager is not null, "How could this be a null parentManager.");
-        this.parentManager = parentManager;
-        this.dataField = dataField;
-        fieldInfo = parentManager.GetItemProperties().Find(dataField, true);
-        if (fieldInfo is null)
+        _parentManager = parentManager;
+        _fieldInfo = parentManager.GetItemProperties().Find(dataField, true)!;
+        if (_fieldInfo is null)
```
dreddy-work, line 28: May be changing to throw right away looks cleaner.
```suggestion
        _fieldInfo = parentManager.GetItemProperties().Find(dataField, true) ?? throw new ArgumentException(string.Format(SR.RelatedListManagerChild, dataField));
```

src/System.Windows.Forms/src/System/Windows/Forms/RelatedPropertyManager.cs
```diff
@@ -46,13 +45,13 @@ internal override string GetListName()
         return base.GetListName();
     }
 
-    protected internal override string GetListName(ArrayList listAccessors)
+    protected internal override string GetListName(ArrayList? listAccessors)
```
dreddy-work, line 48: can you check call sites if any one expecting null reference exception?  or returning string.empty is an option when input is null?

dreddy-work (changes requested):

src/System.Windows.Forms/src/System/Windows/Forms/RelatedPropertyManager.cs
gpetrou: The only call site I see is the `GetListName()` above. So, I added a null check.

dreddy-work: returning empty make sense to me.

dreddy-work (approved):
LGTM.

PR #9454, opened 2023-07-10
Update control owning field of MonthCalendar to using weak reference

SimonZhao888:
Related to https://github.com/dotnet/winforms/issues/9224

## Proposed changes

- Update the control owning field of MonthCalendar to using weak references
- Enable the GC collect test for the MonthCalendar control in unit test cases

## Customer Impact

## Regression? 

- No

## Test methodology 

- Unit tests

## Test environment(s) 

- dotnet 8.0.0-preview.7.23358.3

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9454)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/MonthCalendar.MonthCalendarAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/MonthCalendar.MonthCalendarAccessibleObject.cs
@@ -25,10 +25,7 @@ public partial class MonthCalendar
 
         public MonthCalendarAccessibleObject(MonthCalendar owner) : base(owner)
         {
-            if(this.TryGetOwnerAs(out MonthCalendar? owningMonthCalendar))
-            {
-                owningMonthCalendar.DisplayRangeChanged += OnMonthCalendarStateChanged;
-            }
+            owner.DisplayRangeChanged += OnMonthCalendarStateChanged;
         }
 
         // Use a LinkedList instead a List for the following reasons:
@@ -47,7 +44,7 @@ public partial class MonthCalendar
         {
             get
             {
-                if (!this.TryGetOwnerAs(out MonthCalendar? owningMonthCalendar) || !owningMonthCalendar.IsHandleCreated)
+                if (!this.TryGetOwnerAs(out MonthCalendar? owner) || !owner.IsHandleCreated)
                 {
                     return null;
                 }
@@ -127,13 +124,13 @@ public partial class MonthCalendar
                 _ => DayOfWeek.Sunday
             };
 
-        internal MONTH_CALDENDAR_MESSAGES_VIEW CalendarView => this.TryGetOwnerAs(out MonthCalendar? owningMonthCalendar) ? owningMonthCalendar._mcCurView : MONTH_CALDENDAR_MESSAGES_VIEW.MCMV_MONTH;
+        internal MONTH_CALDENDAR_MESSAGES_VIEW CalendarView => this.TryGetOwnerAs(out MonthCalendar? owner) ? owner._mcCurView : MONTH_CALDENDAR_MESSAGES_VIEW.MCMV_MONTH;
 
         internal override int ColumnCount
         {
             get
             {
-                if (!this.TryGetOwnerAs(out MonthCalendar? owningMonthCalendar) || !owningMonthCalendar.IsHandleCreated || CalendarsAccessibleObjects is null)
+                if (!this.TryGetOwnerAs(out MonthCalendar? owner) || !owner.IsHandleCreated || CalendarsAccessibleObjects is null)
                 {
                     return -1;
                 }
@@ -160,7 +157,7 @@ public partial class MonthCalendar
             int innerX = (int)x;
             int innerY = (int)y;
 
-            if (!this.TryGetOwnerAs(out MonthCalendar? owningMonthCalendar) || !owningMonthCalendar.IsHandleCreated)
+            if (!this.TryGetOwnerAs(out MonthCalendar? owner) || !owner.IsHandleCreated)
             {
                 return base.ElementProviderFromPoint(x, y);
             }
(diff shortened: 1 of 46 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/MonthCalendar.MonthCalendarAccessibleObject.cs
```diff
@@ -201,13 +201,14 @@ internal override int ColumnCount
             };
         }
 
-        internal DayOfWeek FirstDayOfWeek => CastDayToDayOfWeek(_owningMonthCalendar.FirstDayOfWeek);
+        internal DayOfWeek FirstDayOfWeek => this.TryGetOwnerAs(out MonthCalendar? owner) ? CastDayToDayOfWeek(owner.FirstDayOfWeek) : CastDayToDayOfWeek(Day.Default);
 
-        internal bool Focused => _owningMonthCalendar.Focused;
+        internal bool Focused => this.TryGetOwnerAs(out MonthCalendar? owner) ? owner.Focused : false;
```
Tanya-Solyanik, line 206: Please unify local variable names. Either use `owner` or `owningMonthCalendar` everywhere

src/System.Windows.Forms/src/System/Windows/Forms/MonthCalendar.MonthCalendarAccessibleObject.cs
```diff
@@ -26,9 +25,10 @@ internal class MonthCalendarAccessibleObject : ControlAccessibleObject
 
         public MonthCalendarAccessibleObject(MonthCalendar owner) : base(owner)
         {
-            _owningMonthCalendar = owner;
-
-            _owningMonthCalendar.DisplayRangeChanged += OnMonthCalendarStateChanged;
+            if(this.TryGetOwnerAs(out MonthCalendar? owningMonthCalendar))
+            {
+                owningMonthCalendar.DisplayRangeChanged += OnMonthCalendarStateChanged;
+            }
```
dmitrii-drobotov, line 31: Can we do it this way?
```suggestion
            owner.DisplayRangeChanged += OnMonthCalendarStateChanged;
```

src/System.Windows.Forms/src/System/Windows/Forms/MonthCalendar.MonthCalendarAccessibleObject.cs
```diff
@@ -201,13 +201,14 @@ internal override int ColumnCount
...
 
-        internal bool Focused => _owningMonthCalendar.Focused;
+        internal bool Focused => this.TryGetOwnerAs(out MonthCalendar? owningMonthCalendar) ? owningMonthCalendar.Focused : false;
 
         internal CalendarCellAccessibleObject? FocusedCell
             => UiaCore.UiaClientsAreListening()
-                ? _focusedCellAccessibleObject ??= GetCellByDate(_owningMonthCalendar._focusedDate)
+                ? _focusedCellAccessibleObject ??= this.TryGetOwnerAs(out MonthCalendar? owningMonthCalendar) ? GetCellByDate(owningMonthCalendar._focusedDate)
+                : null
                 : null;
```
dmitrii-drobotov, line 212: It's not related to this PR, but could you also remove `UiaCore.UiaClientsAreListening()` check here? It's might have been needed for `FocusedCell?.RaiseAutomationEvent(...)` usages, but `RaiseAutomationEvent` already has this check:
https://github.com/dotnet/winforms/blob/b5695a8d2321deebc8885610b6fb8ced4d7933e3/src/System.Windows.Forms/src/System/Windows/Forms/AccessibleObject.cs#L2350-L2359

We also don't use this check in any other AO, so this seems redundant.

src/System.Windows.Forms/src/System/Windows/Forms/MonthCalendar.MonthCalendarAccessibleObject.cs
```diff
@@ -569,20 +568,37 @@ internal CalendarTodayLinkAccessibleObject TodayLinkAccessibleObject
...
-                        return $"{_owningMonthCalendar.SelectionStart:y}";
+                        if(this.TryGetOwnerAs(out owningMonthCalendar))
+                        {
+                            return $"{owningMonthCalendar.SelectionStart:y}";
+                        }
+                        else
+                        {
+                            return string.Empty;
+                        }
+
```
dmitrii-drobotov, line 591: Nit: `else` is not needed here and in similar places below.
```suggestion
                        if (this.TryGetOwnerAs(out owningMonthCalendar))
                        {
                            return $"{owningMonthCalendar.SelectionStart:y}";
                        }
                        
                        return string.Empty;
```

src/System.Windows.Forms/src/System/Windows/Forms/MonthCalendar.MonthCalendarAccessibleObject.cs
```diff
@@ -569,20 +568,37 @@ internal CalendarTodayLinkAccessibleObject TodayLinkAccessibleObject
...
             get
             {
+                MonthCalendar? owningMonthCalendar;
                 SelectionRange? range;
 
                 switch (CalendarView)
                 {
                     case MONTH_CALDENDAR_MESSAGES_VIEW.MCMV_MONTH:
-                        range = _owningMonthCalendar.SelectionRange;
+                        range = this.TryGetOwnerAs(out owningMonthCalendar) ? owningMonthCalendar.SelectionRange : new SelectionRange();
```
dmitrii-drobotov, line 577: Perhaps we could return an empty string if owner is null? Now we return a range from `DateTime.MinValue` to `DateTime.MaxValue` (default values with parameter-less constructor) - not sure if it's a good value.

Tanya-Solyanik (approved):
Thank you!

PR #9419, opened 2023-07-04
Changes to fix issues in `DemoConsole`

elachlan:
Fixes #9412
Fixes #9416
Fixes #9420
Fixes #9389

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9419)

PR #9457, opened 2023-07-10
Enable nullability in ResourceCodeDomSerializer.SerializationResourceManager

halgab:
As dicussed [here](https://github.com/dotnet/winforms/pull/9419/files#r1253754529)

## Proposed changes

- Enable nullability in ResourceCodeDomSerializer.SerializationResourceManager
- Fix what looks to be a bug introduced in #8332

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9457)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ComponentCache.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ComponentCache.cs
@@ -8,7 +8,7 @@ namespace System.ComponentModel.Design.Serialization;
 /// <summary>
 ///  This class is used to cache serialized properties and events of components to speed-up Design to Code view switches
 /// </summary>
-internal sealed class ComponentCache : IDisposable
+internal sealed partial class ComponentCache : IDisposable
 {
     private Dictionary<object, Entry>? _cache;
     private readonly IDesignerSerializationManager _serManager;
@@ -217,24 +217,6 @@ internal sealed class ComponentCache : IDisposable
                 }
             }
         }
-    }
-
-    internal readonly struct ResourceEntry(
-        string name,
-        object? value,
-        bool forceInvariant,
-        bool shouldSerializeValue,
-        bool ensureInvariant,
-        PropertyDescriptor propertyDescriptor,
-        ExpressionContext expressionContext)
-    {
-        public readonly bool ForceInvariant = forceInvariant;
-        public readonly bool EnsureInvariant = ensureInvariant;
-        public readonly bool ShouldSerializeValue = shouldSerializeValue;
-        public readonly string Name = name;
-        public readonly object? Value = value;
-        public readonly PropertyDescriptor PropertyDescriptor = propertyDescriptor;
-        public readonly ExpressionContext ExpressionContext = expressionContext;
     }
 
     // A single cache entry
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ComponentCache.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ComponentCache.cs
@@ -55,9 +55,9 @@ internal sealed class ComponentCache : IDisposable
         }
         set
         {
-            if (_cache is null && Enabled)
+            if (Enabled)
             {
-                _cache = new Dictionary<object, Entry>();
+                _cache ??= new Dictionary<object, Entry>();
             }
 
             // it's a 1:1 relationship so we can go back from entry to  component (if it's not setup yet.. which should not happen, see ComponentCodeDomSerializer.cs::Serialize for more info)
@@ -219,15 +219,22 @@ internal sealed class ComponentCache : IDisposable
         }
     }
 
-    internal struct ResourceEntry
+    internal readonly struct ResourceEntry(
+        string name,
+        object? value,
+        bool forceInvariant,
+        bool shouldSerializeValue,
+        bool ensureInvariant,
+        PropertyDescriptor propertyDescriptor,
+        ExpressionContext expressionContext)
     {
-        public bool ForceInvariant;
-        public bool EnsureInvariant;
-        public bool ShouldSerializeValue;
-        public string Name;
-        public object Value;
-        public PropertyDescriptor PropertyDescriptor;
-        public ExpressionContext ExpressionContext;
+        public readonly bool ForceInvariant = forceInvariant;
+        public readonly bool EnsureInvariant = ensureInvariant;
+        public readonly bool ShouldSerializeValue = shouldSerializeValue;
+        public readonly string Name = name;
+        public readonly object? Value = value;
+        public readonly PropertyDescriptor PropertyDescriptor = propertyDescriptor;
+        public readonly ExpressionContext ExpressionContext = expressionContext;
     }
 
     // A single cache entry
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ResourceCodeDomSerializer.SerializationResourceManager.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ResourceCodeDomSerializer.SerializationResourceManager.cs
@@ -1,8 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-#nullable disable
-
 using System.CodeDom;
 using System.Collections;
 using System.Globalization;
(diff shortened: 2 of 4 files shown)


src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ResourceCodeDomSerializer.SerializationResourceManager.cs
```diff
@@ -414,29 +373,20 @@ public object GetObject(string resourceName, bool forceInvariant)
                     }
                     else
                     {
-                        // Provide a sentinel so we don't repeatedly ask for the same resource.  If this is the invariant culture, always provide one.
+                        // Provide null as a sentinel so we don't repeatedly ask for the same resource.  If this is the invariant culture, always provide one.
                         ResourceTable[culture] = culture.Equals(CultureInfo.InvariantCulture)
-                            ? new Dictionary<string, object>() : s_resourceSetSentinel;
+                            ? new Dictionary<string, object?>() : null;
```
halgab, line 378: Before #8332 the sentinel value used to be an `object`. As `ResourceTable` was a `Hashtable`, it was fine putting both dictionaries and simple objects in it. But changing it to a strongly typed dictionary, the type of the sentinel was also changed to a dictionary. The problem with that approach is that now we're actually updating the sentinel dictionary. Replacing the sentinel object by null should solve it

halgab: @elachlan

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ResourceCodeDomSerializer.SerializationResourceManager.cs
```diff
@@ -221,14 +186,14 @@ private CompareValue CompareWithParentValue(string name, object value)
...
         {
             Debug.Assert(culture.Parent != culture, "should have returned when culture = InvariantCulture");
             CultureInfo parent = culture.Parent;
-            Dictionary<string, object> resourceSet = GetResourceSet(culture);
-            if (resourceSet is not null && resourceSet.TryGetValue(name, out object parentValue))
+            Dictionary<string, object?>? resourceSet = GetResourceSet(culture);
+            if (resourceSet is not null && resourceSet.TryGetValue(name, out object? parentValue))
             {
-                return !parentValue.Equals(value) || parentValue is null ? CompareValue.Different : CompareValue.Same;
+                return parentValue is null || !parentValue.Equals(value) ? CompareValue.Different : CompareValue.Same;
```
halgab, line 196: I figured those were flipped as a mistake...

elachlan: If value is nullable then it might not be. Most likely you are correct though.

elachlan:
Thank you for taking this on.

halgab:
Note that if we merge this in .Net 9 timeframe, we are going to keep a bug that was not here in .Net 7. Is this deemed less risky than taking a fix so late? @dreddy-work

halgab:
@dreddy-work do you want me to send over a PR just with the bugfix for .Net 8?

dreddy-work:
> @dreddy-work do you want me to send over a PR just with the bugfix for .Net 8?

Yes please. We are trying to mitigate risk by limiting what we ship in .NET 8.0.

elachlan:
> @dreddy-work do you want me to send over a PR just with the bugfix for .Net 8?

I'd just submit the PR and let them decide.

halgab:
Now that #9642 was merged, I rebased this one, removing the second commit that was about the bugfix.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ResourceCodeDomSerializer.SerializationResourceManager.cs
```diff
@@ -154,20 +120,17 @@ private IResourceWriter Writer
...
-                {
-                    Name = name,
-                    Value = value,
-                    ForceInvariant = forceInvariant,
-                    ShouldSerializeValue = shouldSerializeValue,
-                    EnsureInvariant = ensureInvariant,
-                    PropertyDescriptor = (PropertyDescriptor)manager.Context[typeof(PropertyDescriptor)],
-                    ExpressionContext = (ExpressionContext)manager.Context[typeof(ExpressionContext)]
-                };
+                ComponentCache.ResourceEntry re = new ComponentCache.ResourceEntry(name,
```
lonitra, line 127:
```suggestion
                ComponentCache.ResourceEntry re = new(
                name,
```

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ResourceCodeDomSerializer.SerializationResourceManager.cs
```diff
@@ -154,20 +120,17 @@ private IResourceWriter Writer
...
+                    value,
+                    forceInvariant,
+                    shouldSerializeValue,
+                    ensureInvariant,
+                    manager.GetContext<PropertyDescriptor>()!,
+                    manager.GetContext<ExpressionContext>()!);
```
lonitra, line 133: Please add comment of how do we know these are not null

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ResourceCodeDomSerializer.SerializationResourceManager.cs
```diff
@@ -546,15 +502,11 @@ public void SetMetadata(IDesignerSerializationManager manager, string resourceNa
...
-                {
-                    invariantWriter = service.GetResourceWriter(CultureInfo.InvariantCulture);
-                }
+                IResourceService? service = manager.GetService<IResourceService>();
+                IResourceWriter? invariantWriter = service?.GetResourceWriter(CultureInfo.InvariantCulture);
 
-                Dictionary<string, object> invariant = GetResourceSet(CultureInfo.InvariantCulture);
-                Dictionary<string, object> t;
+                Dictionary<string, object?>? invariant = GetResourceSet(CultureInfo.InvariantCulture);
+                Dictionary<string, object?>? t;
```
lonitra, line 509: Let's rename this as this is not very descriptive. Perhaps `metaData` ?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ResourceCodeDomSerializer.SerializationResourceManager.cs
```diff
@@ -565,7 +517,7 @@ public void SetMetadata(IDesignerSerializationManager manager, string resourceNa
                     }
 
                     // Note that when we read metadata, for backwards compatibility, we also merge in regular data from the invariant resource. We need to clear that data here, since we are going to write out metadata separately.
-                    invariant.Remove(resourceName);
+                    invariant!.Remove(resourceName);
```
lonitra, line 520: How do we know this is not nullable?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ComponentCache.cs
```diff
@@ -219,15 +219,22 @@ internal void RemoveEntry(object component)
         }
     }
 
-    internal struct ResourceEntry
+    internal readonly struct ResourceEntry(
```
lonitra, line 222: Could you break this out to a separate file?

lonitra (commented):
I added a few comments. Appreciate your patience here

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ResourceCodeDomSerializer.SerializationResourceManager.cs
halgab: `GetResourceSet` returns null only in an error scenario. Which I know is not satisfying... As often in internal classes, checks are pretty inconsistent

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ResourceCodeDomSerializer.SerializationResourceManager.cs
halgab: Well, I don't know the actual reason... I just know the call sites of those properties always expect them not to be null as they are used as parameters in `manager.Context.Push` calls. That's another case of inconsistent checks I think

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ResourceCodeDomSerializer.SerializationResourceManager.cs
lonitra: Generally we are moving to add comments whenever `!` is used, but given that the codebase is so old, it can be hard to determine when exactly things are null in which case we should throw a specific exception instead of getting an NRE. For this case I think the right answer would be to throw an `InvalidOperationException`, but of course if there was more context, we should throw exception that would be most helpful.

```C#
if (invariant is null) {
}

invariant.Remove(resourceName);
```

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ResourceCodeDomSerializer.SerializationResourceManager.cs
lonitra: I see, it looks like these are used as arguments to `manager.Context.Push` only when `isMetadata = false`. So it seems that it is OK for `ResourceEntry.PropertyDescription` and `ResourceEntry.ExpressionContext` to be nullable and should be annotated as so. If these are null when `isMetadata = false`, we should actually throw `ArgumentException` here instead of later getting an `ArgumentNullException` from `manager.Context.Push` which might be confusing.

halgab: I've pushed something to address this. I've tried to stick to what you indicated. Let me know if this needs some more tweaking

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ResourceCodeDomSerializer.SerializationResourceManager.cs
halgab: I was too quick in my first reply. Turns out that we can perfectly explain why `invariant` can't be null. I've added a comment near the null forgiving operator (which move earlier in the code).

lonitra (approved):
Awesome :shipit:

PR #9458, opened 2023-07-11
Update control owning field of TreeView to using weak reference

Epica3055:
Related to #9224 

## Proposed changes

- Update the control owning field of TreeView to using weak references
- Enable the GC collect test for the TreeView control in unit test cases

## Customer Impact

## Regression? 

-  No

## Test methodology 

- Unit tests
 

## Test environment(s) 

- dotnet 8.0.0-preview.7.23358.3

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9458)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TreeView.TreeViewAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TreeView.TreeViewAccessibleObject.cs
@@ -94,7 +94,7 @@ public partial class TreeView
 
                 if (!this.TryGetOwnerAs(out TreeView? owningTreeView))
                 {
-                    return state;
+                    return AccessibleStates.None;
                 }
 
                 if (owningTreeView.Focused)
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/TreeView.TreeViewAccessibleObject.cs
```diff
@@ -47,55 +43,66 @@ internal override int GetChildIndex(AccessibleObject? child)
...
         public override AccessibleStates State
         {
             get
             {
                 AccessibleStates state = AccessibleStates.Focusable;
 
-                if (_owningTreeView.Focused)
+                if (!this.TryGetOwnerAs(out TreeView? owningTreeView))
+                {
+                    return state;
```
Tanya-Solyanik, line 97: If this accessible object is disconnected from the control, then it shouldn't be focusable

dmitrii-drobotov (approved):

Issue #721, opened 2019-04-09
Add unit tests to System.Windows.Forms.Design.ControlDesigner


JuditRose:
Add unit tests (if not exists) for

1) Methods:
- [x] M:System.Windows.Forms.Design.ControlDesigner.#ctor
- [x] M:System.Windows.Forms.Design.ControlDesigner.get_SelectionRules

2) Classes:
- [x] System/ComponentModel/Design/DesignerActionListsChangedEventArgs.cs
- [ ] System/ComponentModel/Design/DesignerActionPanel.cs
- [ ] System/ComponentModel/Design/DesignerActionUIService.cs
- [ ] System/Windows/Forms/Design/BaseContextMenuStrip.cs
- [ ] System/Windows/Forms/Design/Behavior/ContainerSelectorBehavior.cs
- [ ] System/Windows/Forms/Design/Behavior/DragAssistanceManager.cs
- [ ] System/Windows/Forms/Design/Behavior/DropSourceBehavior.cs
- [ ] System/Windows/Forms/Design/Behavior/ResizeBehavior.cs
- [x] System/Windows/Forms/Design/ControlDesigner.cs
- [ ] System/Windows/Forms/Design/DesignerActionPanel.cs
- [ ] System/Windows/Forms/Design/DesignerActionService.cs
- [ ] System/Windows/Forms/Design/DesignerActionUI.cs
- [ ] System/Windows/Forms/Design/DesignerFrame.cs
- [ ] System/Windows/Forms/Design/DesignerUtils.cs
- [ ] System/Windows/Forms/Design/FormDocumentDesigner.cs
- [ ] System/Windows/Forms/Design/ParentControlDesigner.cs
- [ ] System/Windows/Forms/Design/SelectionManager.cs
- [ ] System/Windows/Forms/Design/ToolStripAdornerWindowService.cs
- [ ] System/Windows/Forms/Design/ToolStripDesigner.cs
- [ ] System/Windows/Forms/Design/ToolStripDesignerUtils.cs
- [ ] System/Windows/Forms/Design/ToolStripDropDownDesigner.cs
- [ ] System/Windows/Forms/Design/ToolStripDropDownItemDesigner.cs
- [ ] System/Windows/Forms/Design/ToolStripEditorManager.cs
- [ ] System/Windows/Forms/Design/ToolStripInSituService.cs
- [ ] System/Windows/Forms/Design/ToolStripItemBehavior.cs
[...]

PR #9462, opened 2023-07-11
Add unit test for System.Windows.Forms.Design.ControlDesigner.#ctor

SimonZhao888:
Related to https://github.com/dotnet/winforms/issues/721

## Proposed changes

- Add unit test for System.Windows.Forms.Design.ControlDesigner.#ctor
- The test cases for SelectionRules already exist, here is the link:  [SelectionRulesProperty ](https://github.com/SimonZhao888/v-weidzh-winforms/blob/4b56c489bbe0858806bee1a0365824b945c06949/src/System.Windows.Forms.Design/tests/UnitTests/ControlDesignerTests.cs#L55C21-L55C21)

## Customer Impact

- The new test case can run normally through.

## Regression? 

- No

## Test methodology 

- Unit test

## Test environment(s) 

- 8.0.100-preview.7.23324.2

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9462)

### DIFF
--- a/src/System.Windows.Forms.Design/tests/UnitTests/ControlDesignerTests.cs
+++ b/src/System.Windows.Forms.Design/tests/UnitTests/ControlDesignerTests.cs
@@ -11,12 +11,28 @@ public class ControlDesignerTests
     [WinFormsFact]
     public void ControlDesigner_Ctor_Default()
     {
-        using var designer = new ControlDesigner();
-        Assert.Null(designer.Control);
-        Assert.Null(designer.Component);
-        Assert.Equal(SelectionRules.Visible, designer.SelectionRules);
-        Assert.True(designer.ParticipatesWithSnapLines);
-        Assert.False(designer.AutoResizeHandles);
+        using TestControlDesigner controlDesigner = new TestControlDesigner();
+        Assert.False(controlDesigner.AutoResizeHandles);
+        Assert.Null(controlDesigner.Control);
+        Assert.True(controlDesigner.ControlSupportsSnaplines);
+        Assert.Null(controlDesigner.Component);
+        Assert.True(controlDesigner.ForceVisible);
+        Assert.Null(controlDesigner.GetParentComponentProperty());
+        Assert.False(controlDesigner.SerializePerformLayout);
+    }
+
+    [WinFormsFact]
+    public void ControlDesigner_PropertiesTest()
+    {
+        using TestControlDesigner controlDesigner = new TestControlDesigner();
+        using Button button = new Button();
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms.Design/tests/UnitTests/ControlDesignerTests.cs
```diff
@@ -8,6 +8,17 @@ namespace System.Windows.Forms.Design.Tests;
...
 {
+    [WinFormsFact]
+    public void ControlDesigner_Ctor_Default()
+    {
+        using var designer = new ControlDesigner();
+        Assert.Null(designer.Control);
+        Assert.Null(designer.Component);
+        Assert.Equal(SelectionRules.Visible, designer.SelectionRules);
+        Assert.True(designer.ParticipatesWithSnapLines);
+        Assert.False(designer.AutoResizeHandles);
```
dmitrii-drobotov, line 19: I see there are more non-private properties on ControlDesigner, could you test them here as well? You can see a full list in Class View panel in VS:
List of properties for ControlDesigner

In addition to those, I think it would be good to test properties from base class - see `ComponentDesigner_Ctor_Default`. Also, if a property is not available due to the protection layer, you can add it to `TestControlDesigner` similar to `SubComponentDesigner`.

SimonZhao888: Thank you for your guidance.

I had updated the test case for  ComponentDesigner_Ctor, I added a new test case.
And now almost of the non-private properties are already in test cases. except the property of MoveBehavior.

src/System.Windows.Forms.Design/tests/UnitTests/TestControlDesigner.cs
```diff
@@ -10,6 +10,18 @@ namespace System.Windows.Forms.Design.Tests;
 
 internal class TestControlDesigner : ControlDesigner
 {
+    internal new bool ControlSupportsSnaplines => base.ControlSupportsSnaplines;
```
dreddy-work, line 13: what are we trying to achieve with this `new `properties?

SimonZhao888: It's my mistake, all these properties are internal, not protected, they are directly accessible.

Tanya-Solyanik (approved):

dmitrii-drobotov (approved):

v-elnovikova (approved):

PR #9463, opened 2023-07-11
Add the logical judgment of scaledHeightPadding in the function Detai…

### LINKED ISSUE #9449: [Multi-Monitor] Buttons on ThreadExceptionDialog are truncated when DPI is differ
### .NET version

.Net 8.0.100-preview.7.23356.13

No

No

### Issue description

Buttons on ThreadExceptionDialog are truncated when expanding Details button in threadExpectionDialog on 300%DPI primary screen, then move to 100% secondary screen, and collapse Details button. --- This issue was found after PR: https://github.com/dotnet/winforms/pull/9399 merged
**The screenshot:** 
100%Secondary

**The testing result from 300%DPI primary screen to 100%DPI secondary screen:**

https://github.com/dotnet/winforms/assets/26474449/5b45f518-ebb7-410d-8b1c-8532dd14bf85

**The testing result from 100%DPI primary screen to 300%DPI secondary screen:**

https://github.com/dotnet/winforms/assets/26474449/5a70c21c-b20e-4d00-9086-ddda1df08f28

### Steps to reproduce

1. Open following testing app on 300% primary screen:
[WinFormsApp6.zip](https://github.com/dotnet/winforms/files/11996946/WinFormsApp6.zip)
2. Build and run it, click **Open ThreadExceptionDialog** button to open ThreadExceptionDialog
3. Expand **Details** button in ThreadExpectionDialog
4. Move this ThreadExpectionDialog to 100% secondary screen
5. Collapse **Details** button, and observe the buttons on ThreadExceptionDialog

LeafShi1:
Fixes #9449 

## Proposed changes

- Add the logical judgment of scaledHeightPadding in the function DetailsClick

- Expand the details in threadExpectionDialog at 300%DPI, the height of the expanded details bar is detailHeight (height is 454) + scaledHeightPadding (which is a fixed value of 8), move to 100% secondary screen and collapse the details, collapsed height is detailHeight (because auto scaling is added, detailHeight will automatically scale to 154) + scaledHeightPadding (this value is hard-coded, so it has not changed due to movement, so my fix idea is to adjust the height according to the scaling `scaledHeightPadding = (int)(_scaledHeightPadding * (currentAutoScaleDimensions. Height / savedAutoScaleDimensions. Height));`)

## Customer Impact

-  Buttons on ThreadExceptionDialog are truncated when DPI is differ

## Regression? 

-  No

## Risk

- Minimal

### Before
Buttons on ThreadExceptionDialog are truncated when expanding Details button in threadExpectionDialog on 300%DPI primary screen, then move to 100% secondary screen, and collapse Details button.
BeforeScaledExceptionDialog

### After
Buttons on ThreadExceptionDialog can be fully displayed when expanding Details button in threadExpectionDialog on 300%DPI primary screen, then move to 100% secondary screen, and collapse Details button.
ScaledExceptionDialog

## Test methodology 

- Manually

## Test environment(s) 

- .net 8.0.100-preview.7.23357.8

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9463)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ThreadExceptionDialog.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ThreadExceptionDialog.cs
@@ -69,8 +69,7 @@ public class ThreadExceptionDialog : Form
     private Bitmap? _expandImage;
     private Bitmap? _collapseImage;
     private bool _detailsVisible;
-    private SizeF savedAutoScaleDimensions;
-    private int scaledHeightPadding = HEIGHTPADDING;
+    private SizeF _savedAutoScaleDimensions;
 
     /// <summary>
     ///  Initializes a new instance of the <see cref="ThreadExceptionDialog"/> class.
@@ -358,7 +357,7 @@ public class ThreadExceptionDialog : Form
 
         _detailsButton.Image = _detailsVisible ? _collapseImage : _expandImage;
 
-        savedAutoScaleDimensions = AutoScaleDimensions;
+        _savedAutoScaleDimensions = AutoScaleDimensions;
     }
 
     /// <summary>
@@ -386,10 +385,11 @@ public class ThreadExceptionDialog : Form
     /// </summary>
     private void DetailsClick(object? sender, EventArgs eventargs)
     {
-        if (savedAutoScaleDimensions.Height > 0)
+        int scaledHeightPadding = _scaledHeightPadding;
+        if (_savedAutoScaleDimensions.Height > 0)
         {
             SizeF currentAutoScaleDimensions = CurrentAutoScaleDimensions;
-            scaledHeightPadding = (int)(_scaledHeightPadding * (currentAutoScaleDimensions.Height / savedAutoScaleDimensions.Height));
+            scaledHeightPadding = (int)Math.Round((_scaledHeightPadding * (currentAutoScaleDimensions.Height / _savedAutoScaleDimensions.Height)));
         }
 
         int delta = _details.Height + scaledHeightPadding;
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ThreadExceptionDialog.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ThreadExceptionDialog.cs
@@ -48,7 +48,6 @@ public class ThreadExceptionDialog : Form
     private readonly int _scaledButtonTopPadding = BUTTONTOPPADDING;
     private readonly int _scaledButtonDetailsLeftPadding = BUTTONDETAILS_LEFTPADDING;
     private readonly int _scaledMessageTopPadding = MESSAGE_TOPPADDING;
-    private readonly int _scaledHeightPadding = HEIGHTPADDING;
     private readonly int _scaledButtonWidth = BUTTONWIDTH;
     private readonly int _scaledButtonHeight = BUTTONHEIGHT;
     private readonly int _scaledButtonAlignmentWidth = BUTTONALIGNMENTWIDTH;
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ThreadExceptionDialog.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ThreadExceptionDialog.cs
@@ -1,4 +1,4 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
+// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
@@ -69,7 +69,7 @@ public class ThreadExceptionDialog : Form
     private Bitmap? _expandImage;
     private Bitmap? _collapseImage;
     private bool _detailsVisible;
-    private SizeF _savedAutoScaleDimensions;
+    private int _scaledHeightPadding = HEIGHTPADDING;
 
     /// <summary>
     ///  Initializes a new instance of the <see cref="ThreadExceptionDialog"/> class.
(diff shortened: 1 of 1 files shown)


dreddy-work:
@LeafShi1, can you add details description here on what was happening and how your fix is helping?  Also, did you try setting AutoScaleMode as `DPI` on this dialog?  I see it has `InHerit ` as default.

src/System.Windows.Forms/src/System/Windows/Forms/ThreadExceptionDialog.cs
```diff
@@ -69,6 +69,8 @@ public class ThreadExceptionDialog : Form
     private Bitmap? _expandImage;
     private Bitmap? _collapseImage;
     private bool _detailsVisible;
+    private SizeF savedAutoScaleDimensions;
```
Tanya-Solyanik, line 72: Please prefix field names with an underscore.

src/System.Windows.Forms/src/System/Windows/Forms/ThreadExceptionDialog.cs
```diff
@@ -382,7 +386,13 @@ public override bool AutoSize
     /// </summary>
     private void DetailsClick(object? sender, EventArgs eventargs)
     {
-        int delta = _details.Height + _scaledHeightPadding;
+        if (savedAutoScaleDimensions.Height > 0)
+        {
+            SizeF currentAutoScaleDimensions = CurrentAutoScaleDimensions;
+            scaledHeightPadding = (int)(_scaledHeightPadding * (currentAutoScaleDimensions.Height / savedAutoScaleDimensions.Height));
```
Tanya-Solyanik, line 392: Does this work when moving this dialog from screen to screen multiple times and opening on each screen? I'm asking because the description was not specific enough.

Tanya-Solyanik, line 392: `scaledHeightPadding` - is not used as a private field, rather as a local variable.

Tanya-Solyanik (commented):
Nice fix!

src/System.Windows.Forms/src/System/Windows/Forms/ThreadExceptionDialog.cs
LeafShi1: > Does this work when moving this dialog from screen to screen multiple times and opening on each screen? I'm asking because the description was not specific enough.

I added detailed descrption on PR's Proposed changes.

But because the zoom ratio of the two screens (96/288=0.34027) is a float value, the code is rounded, which also results in the height of each collapse on the 100% screen and the height on the 300% screen The expanded height cannot be precisely scaled, so there will still be breakage after multiple scalings, but this will happen after more than five such operations. Combined with the user's usage habits, the current repair should be able to satisfy needs, what do you think?

LeafShi1:
> @LeafShi1, can you add details description here on what was happening and how your fix is helping? Also, did you try setting AutoScaleMode as `DPI` on this dialog? I see it has `InHerit ` as default.

Yes, I added [AutoScaleMode = AutoScaleMode.Dpi;](https://github.com/dotnet/winforms/blob/7b621e585cc53860755ea5968ae817d4b6079dbe/src/System.Windows.Forms/src/System/Windows/Forms/ThreadExceptionDialog.cs#L338C33-L338C33)  before.

src/System.Windows.Forms/src/System/Windows/Forms/ThreadExceptionDialog.cs
Tanya-Solyanik: I agree that this is acceptable in this case. However, this is a bad pattern, and we don't want it copied over to other dialogs. A better way is on each click to scale the same constant starting value instead of the previous padding height(_scaledHeightPadding). Scale HEIGHTPADDING using LogicalToDeviceUnits

LeafShi1: The logic of the original code is like this, but the expand and collapse are not performed under the same screen, so the scale the same constant starting value caused this issue, I'm not sure if I misunderstood what you meant, you meant to change to following?
image

src/System.Windows.Forms/src/System/Windows/Forms/ThreadExceptionDialog.cs
```diff
@@ -382,7 +385,14 @@ public override bool AutoSize
     /// </summary>
     private void DetailsClick(object? sender, EventArgs eventargs)
     {
-        int delta = _details.Height + _scaledHeightPadding;
+        int scaledHeightPadding = _scaledHeightPadding;
```
dreddy-work, line 388: why can't we adjust  `_scaledHeightPadding` inside `ThreadExceptionDialog_DpiChanged` according to the new DPI  and rest work as expected?

src/System.Windows.Forms/src/System/Windows/Forms/ThreadExceptionDialog.cs
Tanya-Solyanik: I would expect that Control.DeviceDpi value matches the current screen when the details button is clicked. Is it not the case?

LeafShi1: It should be like this, but here, the height of the collapsed should depend on how much the height is expanded, because  currently the current logic for autoscaling is currentAutoScaleDimensions/ savedAutoScaleDimensions
image

dreddy-work (approved):

PR #9475, opened 2023-07-12
Fix ListBox rendering issues on horizontal scroll

### LINKED ISSUE #9392: The GridView's columns display incorrectly when setting PropertyGrid's selectedObject to GridView and slide the scrollbar in DataGridViewColumn Collection Editor dialog
### .NET version

.Net 8.0 SDK: 8.0.100-preview.7.23328.8

Not tested/verified

Yes, cannot repro in .Net 6.0 and Net 7.0.

### Issue description

When setting PropertyGrid's selectedObject to GridView with Columns, open DataGridViewColumn Collection Editor dialog in propertyGrid, use the mouse to slide the scrollbar left and right to observe the Columns, the GridView's columns display incorrectly.
The issue also can repro in ListView control.

**.Net 8.0 result:**
image

**.Net 7.0/.Net 6.0**
When using the mouse to slide the scrollbar left and right, the GridView's columns display correctly.
image

### Steps to reproduce

Use following sample application:
[WinFormsApp9.zip](https://github.com/dotnet/winforms/files/11903662/WinFormsApp9.zip)

https://github.com/dotnet/winforms/assets/56664234/a9324606-f2c2-4d4a-aa9d-44cc98f82ca9

dmitrii-drobotov:
Fixes #9392

Regression was introduced in #9309 and specifically in ListBox change:
```diff
   using var e = new DrawItemEventArgs(
-      dis->hDC,
+      dis->hDC.CreateGraphics(),
       Font,
       bounds,
-      dis->itemID,
-      dis->itemState,
+      (int)dis->itemID,
+      (DrawItemState)(int)dis->itemState,
       ForeColor,
       BackColor);
```            

`DrawItemEventArgs.Graphics` object started to be created by `hDC.CreateGraphics()`, which only calls `Graphics.FromHdcInternal(hdc)`, but previously it was a more complex initialization in `DrawingEventArgs`:
https://github.com/dotnet/winforms/blob/ea36ea9229ef001ffe21629d1c4e2fb3047a04f9/src/System.Windows.Forms/src/System/Windows/Forms/Internal/DrawingEventArgs.cs#L94-L118

There are different constructors for `DrawItemEventArgs`, so I restored it to use one accepting `HDC` object, which will initialize `Graphics` with above snippet.

## Proposed changes

- Use compatible constructor for creating `DrawItemEventArgs` in `ListBox`

## Customer Impact

- No rendering issues on ListBox horizontal scroll

## Regression? 

- Yes

## Risk

- Low

### Before

ListBox with rendering issues on horizontal scroll

### After

ListBox without rendering issues on horizontal scroll

## Test methodology 

- Manual

## Accessibility testing  

No differences found.
 

## Test environment(s) 

- .NET  8.0.100-preview.7.23360.13
- Win 11 22H2

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9475)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListBox.cs
@@ -2383,11 +2383,11 @@ public partial class ListBox : ListControl
         }
 
         using var e = new DrawItemEventArgs(
-            dis->hDC.CreateGraphics(),
+            dis->hDC,
             Font,
             bounds,
-            (int)dis->itemID,
-            (DrawItemState)(int)dis->itemState,
+            dis->itemID,
+            dis->itemState,
             ForeColor,
             BackColor);
 


Tanya-Solyanik (approved):
Thank you!

JeremyKuhne (approved):
Thanks! This might be because the `Graphics.PageUnit` is explicitly being set to Pixel when lazily created.

JeremyKuhne:
`ComboBox` should also be fixed

Tanya-Solyanik:
@LeafShi1 @SimonZhao888 @Epica3055  - could one of you please try to repro the same with combobox and then apply the same fix to it?

SimonZhao888:
> 

Hi @Tanya-Solyanik, @JeremyKuhne we had verified the ComBox control, and it seems that this issue does not exist, following is our recorded screen:
image

SimonZhao888:
> @SimonZhao888 - the problem is not in the collection editor but in the ComboBox control - 

Hi @Tanya-Solyanik, we don't repro this issue with ComboBox.
ComboBox
Whether we need to fix it?

dmitrii-drobotov:
@SimonZhao888 - I think it reproduces only with `DrawMode = DrawMode.OwnerDraw*`, otherwise `WmReflectDrawItem` method is not called. I can see some rendering issues with owner draw mode:
ComboBox rendering issues
Here's my sample app, please confirm that your fix resolves these problems:
[WinFormsApp9.zip](https://github.com/dotnet/winforms/files/12038497/WinFormsApp9.zip)

dmitrii-drobotov:
> @SimonZhao888 - I think it reproduces only with `DrawMode = DrawMode.OwnerDraw*`, otherwise `WmReflectDrawItem` method is not called. I can see some rendering issues with owner draw mode: ComboBox rendering issues Here's my sample app, please confirm that your fix resolves these problems: [WinFormsApp9.zip](https://github.com/dotnet/winforms/files/12038497/WinFormsApp9.zip)

I looked deeper into my example, and it turned out to be not related to #9309, I get the same result in .NET 7 and .NET Framework, and also with graphics fix in .NET 8. It's probably due to the way I implemented DrawItem method or my setup.

In ListBox it reproduces only with horizontal scroll, and does not with vertical scroll. In ComboBox I don't think there is a way to make items scrollable horizontally, so it probably doesn't have the exact same issue, but it might get issues in some other DrawItem implementation.

JeremyKuhne:
@SimonZhao888 it does need fixed as it also is a potential performance regression. We try to avoid creating the Graphics object unless we really need it. If all we do is use GDI calls we won't create the Graphics object when we pass the HDC.

PR #9488, opened 2023-07-13
Enable nullability in HtmlShim

gpetrou:
## Proposed changes

- Enable nullability in HtmlShim.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9488)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/HtmlShim.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/HtmlShim.cs
@@ -89,11 +89,8 @@ internal abstract class HtmlShim : IDisposable
         if (disposing)
         {
             DisconnectFromEvents();
-            if (_events is not null)
-            {
-                _events.Dispose();
-                _events = null;
-            }
+            _events?.Dispose();
+            _events = null;
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/HtmlShim.cs
```diff
@@ -31,9 +29,9 @@ private EventHandlerList Events
     {
         get
         {
-            events ??= new EventHandlerList();
+            _events ??= new EventHandlerList();
```
dreddy-work, line 32: nit: expressional body?

src/System.Windows.Forms/src/System/Windows/Forms/HtmlShim.cs
```diff
@@ -98,17 +96,17 @@ protected virtual void Dispose(bool disposing)
         if (disposing)
         {
             DisconnectFromEvents();
-            if (events is not null)
+            if (_events is not null)
```
dreddy-work, line 99: _events?.Dispose();
_events = null;

gpetrou: In that case we would always set to null even if it is already null, right?

dreddy-work: yes. that should be fine.

dreddy-work (approved):

Issue #9470, opened 2023-07-12
'System.ArgumentException: Value does not fall within the expected range' when using Windows Forms user control within VB6 form via COM interop


oryhanen:
### .NET version

.NET 8.0 preview 5

### Did it work in .NET Framework?

Yes

### Did it work in any of the earlier releases of .NET Core or .NET 5+?

.NET 7

### Issue description

Using a Windows Forms user control embedded into a VB6 form via COM interop fails, with an exception:

System.ArgumentException: Value does not fall within the expected range.
   at System.Runtime.InteropServices.Marshal.ThrowExceptionForHR(Int32 errorCode, IntPtr errorInfo)
   at Windows.Win32.System.Com.IGlobalInterfaceTable.RegisterInterfaceInGlobal(IUnknown* pUnk, Guid* riid, UInt32* pdwCookie)
   at Windows.Win32.Foundation.GlobalInterfaceTable.RegisterInterface[TInterface](TInterface* interface)
   at Windows.Win32.Foundation.AgileComPointer`1..ctor(TInterface* interface, Boolean takeOwnership)
   at System.Windows.Forms.Control.ActiveXImpl.InPlaceActivate(OLEIVERB verb)
   at System.Windows.Forms.Control.ActiveXImpl.OnFocus(Boolean focus)
   at System.Windows.Forms.Control.ActiveXOnFocus(Boolean focus)
   at System.Windows.Forms.Control.ChildGotFocus(Control child)
   at System.Windows.Forms.Control.OnGotFocus(EventArgs e)
   at System.Windows.Forms.TextBox.OnGotFocus(EventArgs e)
   at System.Windows.Forms.Control.WmSetFocus(Message& m)
   at System.Windows.Forms.Control.WndProc(Message& m)
   at System.Windows.Forms.TextBoxBase.WndProc(Message& m)
   at System.Windows.Forms.TextBox.WndProc(Message& m)
   at System.Windows.Forms.NativeWindow.Callback(HWND hWnd, WM msg, WPARAM wparam, LPARAM lparam)

### Steps to reproduce

I built a minimal repro application based on the code found in: https://github.com/dotnet/samples/tree/main/core/extensions/COMServerDemo

[ComInteropTest.zip](https://github.com/dotnet/winforms/files/12026351/ComInteropTest.zip)

[...]

PR #9489, opened 2023-07-13
Fix ActiveXImpl.InPlaceActivate

JeremyKuhne:
The `UIWindow` can be null, we need to check for that before trying to wrap it in an `AgileComPointer`. This was an oversight- we check to see if the object is null in the rest of the method. Also checking `pFrame` as it is possible we would fall out before trying to use it.

Also assert that the method succeeded for our own debugging purposes. Historically the result wasn't checked.

Lastly add a missing `AddRef()` and use `ThrowOnFailure()` instead of calling `Marshal` directly (for consistency).

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9489)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
@@ -866,8 +866,8 @@ public partial class Control
                     ThrowHr(hr);
                 }
 
-                var posRect = default(RECT);
-                var clipRect = default(RECT);
+                RECT posRect = default;
+                RECT clipRect = default;
 
                 _inPlaceUiWindow?.Dispose();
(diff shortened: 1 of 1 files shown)


JeremyKuhne:
This is meant to address #9470, but I'm struggling to validate the fix with the repro so I'm not marking it as a fix until I can manually validate. (I'm having issues setting the repro up with the fix- it requires using a shared install, and I'm struggling with redirection.)

The fix is obvious, however, and needs to be made regardless. I want to get this in so it will flow.

I'm giving some thought to how we might be able to add some tests for hosting WinForms controls as ActiveX objects. I do have a simple host in my thirtytwo project which I might use.

dreddy-work (approved):
LGTM

Issue #8842, opened 2022-03-23
SystemIcons outdated?


Rand-Random:

JeremyKuhne:
Resolved with #8802 & #8878

PR #9492, opened 2023-07-14
Utilize StockIcons in a couple interfaces

toehead2001:
## Proposed changes

- Use the new `SystemIcons.GetStockIcon()` method to display up-to-date icons from the OS

## Customer Impact

- Negligible
- Customers will no longer think these User Interface elements look like they're from Windows Vista/7

## Regression? 

- No

## Risk

-

### Before

Exception Dialog (this screenshot was taken at 200% scaling)
exBefore

Paint exception in Winforms Designer
designerBefore

### After

Exception Dialog (this screenshot was taken at 200% scaling)
exAfter

Paint exception in Winforms Designer
(I didn't know how to get the Winforms Designer to use the DLL I built, so testing was simulated by copying code from `ControlDesigner.cs`  into a test control. Which is why it doesn't show the same call stack in this screenshot)
designerAfter

## Test methodology 

- Manual
- tested the Exception Dialog at 100% and 200% scaling
- tested at 100% in the Winforms Designer

## Test environment(s) 

- 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9492)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.cs
@@ -2300,7 +2300,7 @@ public partial class ControlDesigner : ComponentDesigner
                 e.Graphics.DrawRectangle(pen, borderRectangle);
             }
 
-            Icon err = SystemIcons.Error;
+            using Icon err = SystemIcons.GetStockIcon(StockIconId.Error);
             e.Graphics.FillRectangle(Brushes.White, imageRect);
             e.Graphics.DrawIcon(err, imageRect.X, imageRect.Y);
             textRect.X++;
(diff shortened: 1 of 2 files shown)


toehead2001:
@dotnet-policy-service agree

elachlan:
`SystemIcons` is cached and doesn't have a lifetime. `GetStockIcon` isn't cached and the icon needs to be disposed of.

The team will be better able to assess this. Maybe `SystemIcons` can be themed?

elachlan:
Related #8802 #8842

elachlan:
CC: @JeremyKuhne

JeremyKuhne:
> The team will be better able to assess this. Maybe SystemIcons can be themed?

We can't really as some people will be expecting these to never change (which they really haven't). It's part of the contract for the new API that it's whatever the OS gives us. (Ok, so is the other one, but Windows has never changed them.)

As far as having a static for these, in this case I don't think it is that useful.

JeremyKuhne (approved):
Thanks!

JeremyKuhne:
We need to look at updating the default form icon as well, but that is a bit more tricky as these icons are just icon handles, not whole icon files. WinForms uses multiple sizes from the `Icon` for the `Form`.

I suspect in the immediate we can just manually call the same StockIcon Windows APIs directly and get the file path and load that.

@merriemcgaw I had forgotten about this- it's probably worth doing a quick pass to leverage these before we get too much closer to RC.

elachlan:
@JeremyKuhne want me to go over all references to SystemIcons and replace them with `SystemIcons.GetStockIcon`?

Amy-Li03:
Verified this on .NET latest build: 8.0.100 - rc.1.23375.11, issue was fixed. The new icon displays on Exception Dialog at 100% and 200% scaling.

Exception Dialog on 200% scalling:
Exception

Nora-Zhou01:
Verified with .NET SDK 8.0.100-preview.7.23376.3 test pass build, this issue was fixed, test result same as above.

PR #9503, opened 2023-07-16
`PropertyStore` refactor for generic wrappers

### LINKED ISSUE #8990: Disposing of an MDI parent form twice throws an exception
### .NET version

.NET 8 Preview 3

Yes

Yes, it works in .NET 7 and earlier

### Issue description

Calling `this.Dispose()` twice inside a form that has a child MDI form crashes the application with `NullReferenceException`.

Inspecting platform code code reveals an inconsistency inside the [PropertyStore.TryGetObject](https://github.com/dotnet/winforms/blob/f3a6aa1d4313472705cf3f00515a496c44552b10/src/System.Windows.Forms/src/System/Windows/Forms/PropertyStore.cs#L190-L195) method:

image

### Steps to reproduce

1. In Visual Studio 2022, create a new project with the Windows Forms App template. Select .NET 8 Preview 3.
2. In the form designer, add a button to the form.
3. Configure the form to be an MDI container, add an MDI child form, and add a handler to `Button.Click`:
```cs
public Form1() {
    InitializeComponent();
    IsMdiContainer = true;
    var form = new Form();
    form.Text = "TEST";
    form.MdiParent = this;
    form.Show();

    button1.Click += (sender, e) =>
    {
        this.Dispose();
        this.Dispose();
    };
}
```
4. Run the project and click on the button.
5. **The application crashes.**

elachlan:
This PR refactors the `PropertyStore` wrappers to generics so they can be used on other struct types. It also adds separate functions for structs with `SetValue<T>`, `GetValue<T>`, and `TryGetValue<T>`.

Fixes #8990

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9503)


elachlan:
CC: @JeremyKuhne

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/FormTests.cs
```diff
@@ -2625,6 +2625,17 @@ public void Form_ParentThenSetShowInTaskbarToFalse()
         form.Show();
     }
 
+    [WinFormsFact]
+    public void Form_MdiContainer_WithChild_DoubleDispose()
+    {
+        using Form parent = new() { IsMdiContainer = true };
```
JeremyKuhne, line 2631: You should put the URL in to the issue.

JeremyKuhne (commented):
Thanks @elachlan!

This class bothers me in general. I suspect that we'd be far better off with a wrapper for `Dictionary<int, object>`. I don't think the boxing would amount to more than what would be gained by simplifying the logic. I would expect that we would be much faster, but even if it was a wash we'd be far better off as the complexity would be vastly reduced.

One has to remember that we were originally developing for a baseline of Windows 95/98 25 years ago. Single cores, megabytes of total memory. :)

Is it possible to do a super constrained fix here and decide on the right long-term solution for .NET 9? You change is great, but I'm wondering if we should just go all the way with this. Running some benchmarks on the existing and potential replacement would be great.

If there are any properties in here that are always stored when the control is created, they should probably be private fields instead.

elachlan:
I also thought about private fields. It is a strange setup and I'm not sure why it exists.

elachlan:
To fix the bug I'll move to a private field, is that okay?

elachlan:
Closing in favor of #9509 and refactoring work continuing in #9508

PR #9504, opened 2023-07-17
Enable nullability in remaining RichTextBox members

gpetrou:
## Proposed changes

- Enable nullability in remaining RichTextBox members.

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/RichTextBox.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/RichTextBox.cs
@@ -425,7 +425,7 @@ public partial class RichTextBox : TextBoxBase
 
             try
             {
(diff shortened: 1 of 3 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/RichTextBox.cs
```diff
@@ -1274,6 +1275,7 @@ public override string SelectedText
         set
         {
             ForceHandleCreate();
+            value ??= string.Empty;
```
dreddy-work, line 1278: what is this change for?

gpetrou: We use `AllowNull` in `TextBoxBase` and inherited classes, so I used the same approach here.

PR #9509, opened 2023-07-17
Fix `Form.Dispose` `NullReferenceException`

### LINKED ISSUE #8990: Disposing of an MDI parent form twice throws an exception

elachlan:
Fixes #8990

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9509)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/PropertyStore.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/PropertyStore.cs
@@ -184,11 +184,22 @@ internal partial class PropertyStore
     /// </summary>
     public object? GetObject(int key) => GetObject(key, out _);
 
+    /// <summary>
+    /// Retrieves an object value from our property list.
+    /// This will set value to null and return false if the
+    /// list does not contain the given key.
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    /// <param name="key"></param>
+    /// <param name="value"></param>
+    /// <remarks><para>If a null value is set for a given key
+    /// it will return true and a null value.</para></remarks>
+    /// <returns></returns>
     public bool TryGetObject<T>(int key, out T? value)
     {
         object? entry = GetObject(key, out bool found);
         Debug.Assert(!found || entry is null || entry is T, $"Entry is not of type {typeof(T)}, but of type {entry?.GetType()}");
-        if (typeof(T).IsValueType && !typeof(T).IsPrimitive && !typeof(T).IsEnum)
+        if (typeof(T).IsValueType || typeof(T).IsEnum || typeof(T).IsPrimitive)
         {
             value = found && entry is not null ? (T?)entry : default;
             return found;
--- a/src/System.Windows.Forms/src/System/Windows/Forms/PropertyStore.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/PropertyStore.cs
@@ -197,7 +197,6 @@ internal partial class PropertyStore
     ///  it will return true and a null value.</para></remarks>
     /// <returns>True if an object (including null) is found for the given key; otherwise, false.</returns>
     public bool TryGetObject<T>(int key, out T? value)
-
     {
         object? entry = GetObject(key, out bool found);
         Debug.Assert(!found || entry is null || entry is T, $"Entry is not of type {typeof(T)}, but of type {entry?.GetType()}");
--- a/src/System.Windows.Forms/src/System/Windows/Forms/PropertyStore.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/PropertyStore.cs
@@ -184,7 +184,7 @@ internal partial class PropertyStore
     /// </summary>
     public object? GetObject(int key) => GetObject(key, out _);
 
-    public bool TryGetObject<T>(int key, [NotNullWhen(true)] out T? value)
+    public bool TryGetObject<T>(int key, out T? value)
     {
         object? entry = GetObject(key, out bool found);
         Debug.Assert(!found || entry is null || entry is T, $"Entry is not of type {typeof(T)}, but of type {entry?.GetType()}");
@@ -195,7 +195,7 @@ internal partial class PropertyStore
         }
 
         value = found ? (T?)entry : default;
-        return found && entry is not null;
+        return found;
     }
 
     public bool ContainsObjectThatIsNotNull(int key)
--- a/src/System.Windows.Forms/src/System/Windows/Forms/PropertyStore.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/PropertyStore.cs
@@ -187,8 +187,8 @@ internal partial class PropertyStore
     public bool TryGetObject<T>(int key, [NotNullWhen(true)] out T? value)
     {
         object? entry = GetObject(key, out bool found);
-        value = !found ? default : (T?)entry;
-        return found;
+        value = !found || entry is null ? default : (T?)entry;
+        return found && entry is not null;
     }
 
     public bool ContainsObjectThatIsNotNull(int key)
--- a/src/System.Windows.Forms/src/System/Windows/Forms/PropertyStore.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/PropertyStore.cs
@@ -195,7 +195,7 @@ internal partial class PropertyStore
         }
 
         value = found ? (T?)entry : default;
-        return found;
+        return found && entry is not null;
     }
 
     public bool ContainsObjectThatIsNotNull(int key)
--- a/src/System.Windows.Forms/src/System/Windows/Forms/PropertyStore.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/PropertyStore.cs
@@ -185,17 +185,19 @@ internal partial class PropertyStore
     public object? GetObject(int key) => GetObject(key, out _);
 
     /// <summary>
-    /// Retrieves an object value from our property list.
-    /// This will set value to null and return false if the
-    /// list does not contain the given key.
+    ///  Retrieves an object value from our property list.
+    ///  This will set value to null and return false if the
+    ///  list does not contain the given key.
     /// </summary>
-    /// <typeparam name="T"></typeparam>
-    /// <param name="key"></param>
-    /// <param name="value"></param>
+    /// <typeparam name="T">The type of object to retrieve.</typeparam>
+    /// <param name="key">The key corresponding to the object in the property list.</param>
+    /// <param name="value">Output parameter where the object will be set if found.
+    ///  Will be set to null if the key is not present.</param>
     /// <remarks><para>If a null value is set for a given key
-    /// it will return true and a null value.</para></remarks>
-    /// <returns></returns>
+    ///  it will return true and a null value.</para></remarks>
+    /// <returns>True if an object (including null) is found for the given key; otherwise, false.</returns>
     public bool TryGetObject<T>(int key, out T? value)
+
     {
         object? entry = GetObject(key, out bool found);
         Debug.Assert(!found || entry is null || entry is T, $"Entry is not of type {typeof(T)}, but of type {entry?.GetType()}");
--- a/src/System.Windows.Forms/src/System/Windows/Forms/PropertyStore.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/PropertyStore.cs
@@ -187,8 +187,15 @@ internal partial class PropertyStore
     public bool TryGetObject<T>(int key, [NotNullWhen(true)] out T? value)
     {
         object? entry = GetObject(key, out bool found);
-        value = !found || entry is null ? default : (T?)entry;
-        return found && entry is not null;
+        Debug.Assert(!found || entry is null || entry is T, $"Entry is not of type {typeof(T)}, but of type {entry?.GetType()}");
+        if (typeof(T).IsValueType && !typeof(T).IsPrimitive && !typeof(T).IsEnum)
+        {
+            value = found && entry is not null ? (T?)entry : default;
+            return found;
+        }
+
+        value = found ? (T?)entry : default;
+        return found;
     }
 
     public bool ContainsObjectThatIsNotNull(int key)
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
@@ -1758,7 +1758,7 @@ public partial class ToolStrip : ScrollableControl, IArrangedElement, ISupportTo
     {
         get
         {
-            if (!Properties.TryGetObject(ToolStrip.s_propToolTip, out ToolTip? toolTip))
+            if (!Properties.TryGetObject(ToolStrip.s_propToolTip, out ToolTip? toolTip) || toolTip is null)
             {
                 toolTip = new ToolTip();
                 Properties.SetObject(ToolStrip.s_propToolTip, toolTip);
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyStoreTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyStoreTests.cs
@@ -50,10 +50,10 @@ public class PropertyStoreTests
     [MemberData(nameof(PropertyStore_TryGetValue_Exists_TestData))]
     public void PropertyStore_TryGetValue_Exists(int key, object? value)
     {
-        PropertyStore properties = new PropertyStore();
-        properties.SetObject(key, value);
-        Assert.True(properties.ContainsObject(key), "PropertyStore does not contain key.");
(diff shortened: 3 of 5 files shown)


elachlan:
@JeremyKuhne here is the more simple fix.

JeremyKuhne:
@elachlan wouldn't it be better to guard against the cast in the overload I added? Would something like this work?

``` C#
    public bool TryGetObject<T>(int key, [NotNullWhen(true)] out T? value)
    {
        object? entry = GetObject(key, out bool found);
        if (!found || entry is null)
        {
            value = default;
            return false;
        }

        value = (T?)entry!;
        return true;
    }
```

The idea with the `PropertyStore` is to only use memory for a property when it's needed. I think there is still some value there perhaps. There are thirty-two entries for the Control itself.

That said, we've sort of accrued debt here. For example (@dreddy-work):

``` C#
private Dictionary<int, Font>? _dpiFonts;
```

Why are we centrally caching this for every single control? Why isn't there a central cache? I presume there are typically only a few fonts in use in a typical app. Looks like this could be vastly improved?

@KlausLoeffelmann can you give me an idea of a typical large app control count? It would be good to know the scales we're talking about here. 100 total controls? 1000? 10,000?

elachlan:
@JeremyKuhne one of the apps I maintain has logic to avoid handle exhaustion (it was possible to hit 10k handles pretty easily because of the amount of controls). Basically we have a form which runs through a workflow of user designed screens. The screens can have any number of controls. We added caching so that the controls weren't disposed when switching back and forth to avoid the multisecond delay of loading the controls. 

The app allows a user to open multiple workflow forms at once, hence the exhaustion. But one form was able to use upwards of 2-3k handles. We now only cache the top N number of workflow forms based on some logic.

> The idea with the PropertyStore is to only use memory for a property when it's needed. I think there is still some value there perhaps. There are thirty-two entries for the Control itself.

Can we not just have a nullable backing field? or does a private field take up memory no matter what?

elachlan:
> wouldn't it be better to guard against the cast in the overload I added? Would something like this work?

@JeremyKuhne, I started doing this but then remembered its why I went down the rabbit whole of the refactor with Value vs Object. Because some object types are nullable and should show as being "Found" but return null. I'll push through a fix and see if that causes any failures.

dreddy-work:
> That said, we've sort of accrued debt here. For example . Why are we centrally caching this for every single control? Why isn't there a central cache? 

`Control `has public `Font` property and may differ from other controls in the application. This dictionary was intended to cache Font for each Dpi in PerMonitorV2 application (instead of recreating Font on each monitor move).  Typically this is storing the various Sizes of the Font and only impact PermonitorV2 applications.

JeremyKuhne:
> Control has public Font property and may differ from other controls in the application. This dictionary was intended to cache Font for each Dpi in PerMonitorV2 application (instead of recreating Font on each monitor move). Typically this is storing the various Sizes of the Font and only impact PermonitorV2 applications.

@dreddy-work it's taking what is likely one or two fonts and creating multiple copies for every single control. Not only is this bloating the memory footprint, but it's also blowing out our HFONT cache (killing font rendering performance), and probably other things. If I can get some time, I'll look at a fix this week.

> Can we not just have a nullable backing field? or does a private field take up memory no matter what?

@elachlan the size of the class will be bigger by the size of `Nullable<T>` for the field, which is twice the size of whatever `T` is if it is a value type or the size of two pointers for a reference type.

src/System.Windows.Forms/src/System/Windows/Forms/PropertyStore.cs
```diff
@@ -188,7 +188,14 @@ public int GetInteger(int key, out bool found)
     public bool TryGetObject<T>(int key, [NotNullWhen(true)] out T? value)
     {
         object? entry = GetObject(key, out bool found);
-        value = !found ? default : (T?)entry;
+        Debug.Assert(!found || entry is null || entry is T, $"Entry is not of type {typeof(T)}, but of type {entry?.GetType()}");
+        if (typeof(T).IsValueType && !typeof(T).IsPrimitive && !typeof(T).IsEnum)
```
JeremyKuhne, line 192: Why exclude Primitive and Enum?

elachlan: Because we are handling structs only?

src/System.Windows.Forms/src/System/Windows/Forms/PropertyStore.cs
```diff
@@ -188,7 +188,14 @@ public int GetInteger(int key, out bool found)
...
+        {
+            value = found && entry is not null ? (T?)entry : default;
+            return found;
+        }
+
+        value = found ? (T?)entry : default;
```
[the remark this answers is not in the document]
elachlan: As far as I can tell there are usages that are broken. Here is an example where the test is expecting a null value.

https://github.com/dotnet/winforms/blob/eb926001a1d4592b43b7995c4a388ca3b4433ae0/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs#L1900-L1913

https://github.com/dotnet/winforms/blob/eb926001a1d4592b43b7995c4a388ca3b4433ae0/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripTests.cs#L4227-L4244

elachlan:
@Tanya-Solyanik could you please give this a review?

src/System.Windows.Forms/src/System/Windows/Forms/PropertyStore.cs
```diff
@@ -185,10 +185,17 @@ public int GetInteger(int key, out bool found)
     /// </summary>
     public object? GetObject(int key) => GetObject(key, out _);
 
-    public bool TryGetObject<T>(int key, [NotNullWhen(true)] out T? value)
+    public bool TryGetObject<T>(int key, out T? value)
```
Tanya-Solyanik, line 188: Hm, this is now `TryGetNullableObject`...

src/System.Windows.Forms/src/System/Windows/Forms/PropertyStore.cs
```diff
@@ -185,10 +185,17 @@ public int GetInteger(int key, out bool found)
...
     {
         object? entry = GetObject(key, out bool found);
-        value = !found ? default : (T?)entry;
+        Debug.Assert(!found || entry is null || entry is T, $"Entry is not of type {typeof(T)}, but of type {entry?.GetType()}");
+        if (typeof(T).IsValueType && !typeof(T).IsPrimitive && !typeof(T).IsEnum)
+        {
+            value = found && entry is not null ? (T?)entry : default;
```
Tanya-Solyanik, line 194: Would this work - 
```suggestion
            value = found && entry is not null ? (T)entry : default;
```

src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs
```diff
@@ -1759,7 +1759,7 @@ internal ToolTip ToolTip
     {
         get
         {
-            if (!Properties.TryGetObject(ToolStrip.s_propToolTip, out ToolTip? toolTip))
+            if (!Properties.TryGetObject(ToolStrip.s_propToolTip, out ToolTip? toolTip) || toolTip is null)
```
Tanya-Solyanik, line 1762: Is this method called after or during the disposal?

src/System.Windows.Forms/src/System/Windows/Forms/PropertyStore.cs
elachlan: Yes, apparently its expected that the output can be null. Which is what this whole issue has been over. ToolStripItem.Text is nullable and returns `String.Empty` when it doesn't find a value. but if the user sets null explicitly then it will return null.

https://github.com/dotnet/winforms/blob/eb926001a1d4592b43b7995c4a388ca3b4433ae0/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs#L1900-L1913

src/System.Windows.Forms/src/System/Windows/Forms/PropertyStore.cs
elachlan: I believe this is because the output value is `T?` and you need to match it.

src/System.Windows.Forms/src/System/Windows/Forms/PropertyStore.cs
Tanya-Solyanik: Yes, it would be nice to reflect the fact that this method returns a null, if that is what the user stored, in the method name.

Tanya-Solyanik, line 188: Had you reviewed the places where this method is invoked? Are they affected by the nullability change?

elachlan: `ToolStripItem` is the only warning I received from the removal of the `[NotNullWhen(true)]` attribute.

Tanya-Solyanik: ..or add a `///<remarks>` comment.

Tanya-Solyanik: Also, would you be able to look at adding new PropertyStore tests to cover nulls as nullable structs and objects?

src/System.Windows.Forms/src/System/Windows/Forms/PropertyStore.cs
Tanya-Solyanik: Why are you excluding primitives, they will throw in the cast if user sets a null value in the store. TryGetObject<int>.

src/System.Windows.Forms/src/System/Windows/Forms/PropertyStore.cs
elachlan: Done.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyStoreTests.cs
```diff
@@ -0,0 +1,204 @@
...
+        yield return new object[] { s_object, new object() };
+        yield return new object[] { s_color, Color.Red };
+        yield return new object[] { s_formWindowState, FormWindowState.Maximized };
+    }
+
+    [WinFormsTheory]
+    [MemberData(nameof(PropertyStore_TryGetValue_Exists_TestData))]
+    public void PropertyStore_TryGetValue_Exists(int key, object? value)
+    {
+        PropertyStore properties = new PropertyStore();
```
Tanya-Solyanik, line 53: Consider the following change across the file: 
```suggestion
        PropertyStore store = new();
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyStoreTests.cs
```diff
@@ -0,0 +1,204 @@
...
+        Assert.False(properties.ContainsObject(s_color), "PropertyStore contains key.");
+        Assert.False(properties.TryGetObject(s_color, out Color outValue), "PropertyStore contains key.");
+        Color color = default;
+        Assert.Equal(color, outValue);
+    }
+
+    [WinFormsFact]
+    public void PropertyStore_TryGetValue_Int_Unset_IsDefault()
```
Tanya-Solyanik, line 104:
```suggestion
    public void PropertyStore_TryGetValue_Primitive_Unset_IsDefault()
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyStoreTests.cs
```diff
@@ -0,0 +1,204 @@
...
+        PropertyStore properties = new PropertyStore();
+        properties.SetObject(s_color, null);
+        Assert.True(properties.ContainsObject(s_color), "PropertyStore does not contain key.");
+        Assert.True(properties.TryGetObject(s_color, out Color outValue), "PropertyStore does not contain key.");
+        Color color = default;
+        Assert.Equal(color, outValue);
+    }
+
+    [WinFormsFact]
+    public void PropertyStore_TryGetValue_Int_Null()
```
Tanya-Solyanik, line 136:
```suggestion
    public void PropertyStore_TryGetValue_Primitive_Null()
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyStoreTests.cs
```diff
@@ -0,0 +1,204 @@
...
+    {
+        PropertyStore properties = new PropertyStore();
+        properties.SetObject(s_color, null);
+        Assert.True(properties.ContainsObject(s_color), "PropertyStore does not contain key.");
+        Assert.True(properties.TryGetObject(s_color, out Color? outValue), "PropertyStore does not contain key.");
+        Assert.Null(outValue);
+    }
+
+    [WinFormsFact]
+    public void PropertyStore_TryGetValue_Int_Nullable()
```
Tanya-Solyanik, line 167:
```suggestion
    public void PropertyStore_TryGetValue_Primitive_Nullable()
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyStoreTests.cs
```diff
@@ -0,0 +1,204 @@
...
+    public void PropertyStore_TryGetValue_Struct_Unset_Nullable()
+    {
+        PropertyStore properties = new PropertyStore();
+        Assert.False(properties.ContainsObject(s_color), "PropertyStore contains key.");
+        Assert.False(properties.TryGetObject(s_color, out Color? outValue), "PropertyStore contains key.");
+        Assert.Null(outValue);
+    }
+
+    [WinFormsFact]
+    public void PropertyStore_TryGetValue_Int_Unset_Nullable()
```
Tanya-Solyanik, line 195:
```suggestion
    public void PropertyStore_TryGetValue_Primitive_Unset_Nullable()
```

src/System.Windows.Forms/src/System/Windows/Forms/PropertyStore.cs
```diff
@@ -184,10 +184,28 @@ public int GetInteger(int key, out bool found)
     /// </summary>
     public object? GetObject(int key) => GetObject(key, out _);
 
-    public bool TryGetObject<T>(int key, [NotNullWhen(true)] out T? value)
+    /// <summary>
+    /// Retrieves an object value from our property list.
```
Tanya-Solyanik, line 188: Please indent content of doc comments by 2 spaces

Tanya-Solyanik (commented):
Looks good, I added some minor styling comments

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyStoreTests.cs
```diff
@@ -0,0 +1,204 @@
...
+        PropertyStore properties = new PropertyStore();
+        properties.SetObject(key, value);
+        Assert.True(properties.ContainsObject(key), "PropertyStore does not contain key.");
+        Assert.True(properties.TryGetObject(key, out object? outValue));
+        Assert.Equal(value, outValue);
+    }
+
+    [WinFormsTheory]
+    [MemberData(nameof(PropertyStore_TryGetValue_Exists_TestData))]
+    public void PropertyStore_TryGetValue_Exists_Null(int key, object? value)
```
Tanya-Solyanik, line 62: Is it possible to use a discard instead of the `value` argument?

elachlan: No, its not possible. If I remove the value parameter we get a runtime error around the member data and changing it to underscore and not using it causes a warning of unused parameter.

JeremyKuhne (approved):

PR #9527, opened 2023-07-19
Cleanup for utilization of target new

JeremyKuhne:
Tweaks the editor config settings to encourage behavior that matches our coding style:

- We prefer `Foo foo = new()` over `var foo =`.
- `var` for built in types is an error
- Expression bodied members are always ok

A number of the rules got added by just using the editor. I'll continue to iterate.

Changed a big chunk of the code to use target typed new using fixers and regex's for the majority of the changes.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9527)


.editorconfig
```diff
@@ -70,6 +70,21 @@ dotnet_style_collection_initializer = true:suggestion
...
+dotnet_style_prefer_conditional_expression_over_return = true:silent
+dotnet_style_prefer_inferred_tuple_names = true:suggestion
+dotnet_style_prefer_inferred_anonymous_type_member_names = true:suggestion
+dotnet_style_prefer_compound_assignment = true:suggestion
+dotnet_style_prefer_simplified_interpolation = true:suggestion
+dotnet_style_namespace_match_folder = true:suggestion
+dotnet_style_operator_placement_when_wrapping = beginning_of_line
+tab_width = 4
+end_of_line = crlf
+dotnet_style_allow_multiple_blank_lines_experimental = true:silent
```
Tanya-Solyanik, line 86: We have another rule that does not allow multiple blank lines. This could be `false:error`

Tanya-Solyanik (approved):
I assume that this was done by the analyzer?

.editorconfig
JeremyKuhne: Thanks- I'll try to flip that one next.

PR #9540, opened 2023-07-21
Reduced nesting levels by inverting a few top-level `if`s

Lehonti:
## Proposed changes

The logic of certain `if` blocks has been inverted. Namely:
- Some setters and some methods have `if` blocks that span more than a screen full of text, which makes it hard to see where the block ends. By inverting the logic, the `if` block becomes an `if` guard that is just one more element in the sequence of instructions.
- Also, some `if` blocks in certain setters could be better described and understood as guards, which is not very clear if the block spans all or at least most of the body of the setter. 

## Customer Impact

- No direct customer impact. The code is functionally identical

## Regression? 

- No

## Risk

- Low. Only safe, localized refactorings have been applied, aided by Visual Studio

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9540)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
@@ -368,7 +368,10 @@ public partial class Form : ContainerControl
         }
         set
         {
-            if (value == (_formState[FormStateAllowTransparency] != 0)) return;
+            if (value == (_formState[FormStateAllowTransparency] != 0))
+            {
+                return;
+            }
 
             _formState[FormStateAllowTransparency] = (value ? 1 : 0);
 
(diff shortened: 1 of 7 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
```diff
@@ -368,29 +368,27 @@ public bool AllowTransparency
         }
         set
         {
-            if (value != (_formState[FormStateAllowTransparency] != 0))
-            {
-                _formState[FormStateAllowTransparency] = (value ? 1 : 0);
+            if (value == (_formState[FormStateAllowTransparency] != 0)) return;
```
Tanya-Solyanik, line 371: Could you please move `return` statements to a dedicated line? That style is much more common in this repo.

PR #9512, opened 2023-07-18
Enable nullability in DataGridViewCellLinkedList

gpetrou:
## Proposed changes

- Enable nullability in DataGridViewCellLinkedList.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9512)

PR #9541, opened 2023-07-22
Enable nullability in DataGridViewIntLinkedList

gpetrou:
## Proposed changes

- Enable nullability in DataGridViewIntLinkedList.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9541)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridView/DataGridViewIntLinkedList.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridView/DataGridViewIntLinkedList.cs
@@ -161,14 +161,14 @@ internal class DataGridViewIntLinkedList : IEnumerable
 
         if (tmp2 is not null && tmp2.Int == integer)
         {
-            DataGridViewIntLinkedListElement? tmp3 = tmp2.Next;
+            DataGridViewIntLinkedListElement? removeTargetNext = tmp2.Next;
             if (removeTargetPrevious is null)
             {
-                _headElement = tmp3;
+                _headElement = removeTargetNext;
             }
             else
             {
-                removeTargetPrevious.Next = tmp3;
+                removeTargetPrevious.Next = removeTargetNext;
             }
 
             Count--;
@@ -191,14 +191,14 @@ internal class DataGridViewIntLinkedList : IEnumerable
             index--;
         }
 
-        DataGridViewIntLinkedListElement? tmp3 = tmp2!.Next;
+        DataGridViewIntLinkedListElement? removeTargetNext = tmp2!.Next;
         if (removeTargetPrevious is null)
         {
-            _headElement = tmp3;
+            _headElement = removeTargetNext;
         }
         else
         {
-            removeTargetPrevious.Next = tmp3;
+            removeTargetPrevious.Next = removeTargetNext;
         }
 
         Count--;
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridView/DataGridViewIntLinkedList.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridView/DataGridViewIntLinkedList.cs
@@ -38,10 +38,11 @@ internal class DataGridViewIntLinkedList : IEnumerable
     {
         get
         {
-            Debug.Assert(index >= 0);
-            Debug.Assert(index < Count);
+            ArgumentOutOfRangeException.ThrowIfLessThan(index, 0);
+            ArgumentOutOfRangeException.ThrowIfGreaterThanOrEqual(index, Count);
             if (_lastAccessedIndex == -1 || index < _lastAccessedIndex)
             {
+                // Since we have checked the index value, tmp should not be null here.
                 DataGridViewIntLinkedListElement tmp = _headElement!;
                 int tmpIndex = index;
                 while (tmpIndex > 0)
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridView/DataGridViewIntLinkedList.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridView/DataGridViewIntLinkedList.cs
@@ -146,7 +146,7 @@ internal class DataGridViewIntLinkedList : IEnumerable
 
     public bool Remove(int integer)
     {
-        DataGridViewIntLinkedListElement? tmp1 = null;
+        DataGridViewIntLinkedListElement? removeTargetPrevious = null;
         DataGridViewIntLinkedListElement? tmp2 = _headElement;
         while (tmp2 is not null)
         {
@@ -155,20 +155,20 @@ internal class DataGridViewIntLinkedList : IEnumerable
                 break;
             }
 
-            tmp1 = tmp2;
+            removeTargetPrevious = tmp2;
             tmp2 = tmp2.Next;
         }
 
         if (tmp2 is not null && tmp2.Int == integer)
         {
             DataGridViewIntLinkedListElement? tmp3 = tmp2.Next;
-            if (tmp1 is null)
+            if (removeTargetPrevious is null)
             {
                 _headElement = tmp3;
             }
             else
             {
-                tmp1.Next = tmp3;
+                removeTargetPrevious.Next = tmp3;
             }
 
             Count--;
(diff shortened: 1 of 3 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewIntLinkedList.cs
```diff
@@ -45,27 +43,27 @@ public DataGridViewIntLinkedList(DataGridViewIntLinkedList source)
...
                 {
-                    tmp = tmp.Next;
+                    tmp = tmp!.Next;
                     tmpIndex--;
                 }
 
                 _lastAccessedElement = tmp;
                 _lastAccessedIndex = index;
-                return tmp.Int;
+                return tmp!.Int;
```
dreddy-work, line 56: i feel `tmp` and `_lastAccessedElement` elements can be  reorder to avoid using `!`.  If tmp cannot be null at this line,  why it should be declared as nullable?

gpetrou: I did similar changes to https://github.com/dotnet/winforms/pull/9512. What do you mean by reorder?

dreddy-work: Right. I remember that and make similar comment there.  In this method, we are declaring `tmp `as nullable but contradicting while returning it. Can you see a scenario in this method where `tmp ` could be null but become non-null by the time it reached line 56?

gpetrou: Since we are using the index, I would expect that this is not null and it will not throw. Do you want me to use `!` differently? `Next` is also nullable, so we will still have a couple of null forgiving operators anyway.

gpetrou: @dreddy-work I added a separate commit with the proposed change.

gpetrou:
@dreddy-work any more feedback on this one?

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewIntLinkedList.cs
```diff
@@ -160,9 +159,9 @@ public bool Remove(int integer)
             tmp2 = tmp2.Next;
         }
 
-        if (tmp2.Int == integer)
+        if (tmp2!.Int == integer)
```
dreddy-work, line 162: These methods are not written correctly.  reading code it could be null on line 162 but it seems in real-time, it won't. 

@gpetrou / @elachlan, do you have cycle to look into restructuring this code to avoid obvious `!`  and we can take it to 9.0?

gpetrou: Should we try to use LinkedList<T> instead?

elachlan: I just tried it and there are a couple of issues.

- Can't index into a LinkedList
- There is no Add method - we would have to use addFirst/AddLast or write an extension method

We could wrap it, but that doesn't reduce the complexity.

elachlan: I am fairly sure a LinkedList is best and it shouldn't be too bad to refactor it. I'll give it a quick stab and @gpetrou can look over it to see if I am crazy or missed something stupid.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewIntLinkedList.cs
```diff
@@ -183,15 +182,16 @@ public bool Remove(int integer)
 
     public void RemoveAt(int index)
     {
-        DataGridViewIntLinkedListElement tmp1 = null, tmp2 = _headElement;
+        DataGridViewIntLinkedListElement? tmp1 = null;
+        DataGridViewIntLinkedListElement? tmp2 = _headElement;
```
lonitra, line 186:
```suggestion
        DataGridViewIntLinkedListElement? removeTarget = _headElement;
```

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewIntLinkedList.cs
```diff
@@ -183,15 +182,16 @@ public bool Remove(int integer)
 
     public void RemoveAt(int index)
     {
-        DataGridViewIntLinkedListElement tmp1 = null, tmp2 = _headElement;
+        DataGridViewIntLinkedListElement? tmp1 = null;
```
lonitra, line 185:
```suggestion
        DataGridViewIntLinkedListElement? removeTargetPrevious = null;
```

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewIntLinkedList.cs
```diff
@@ -183,15 +182,16 @@ public bool Remove(int integer)
...
         while (index > 0)
         {
             tmp1 = tmp2;
-            tmp2 = tmp2.Next;
+            tmp2 = tmp2!.Next;
             index--;
         }
 
-        DataGridViewIntLinkedListElement tmp3 = tmp2.Next;
+        DataGridViewIntLinkedListElement? tmp3 = tmp2!.Next;
```
lonitra, line 194:
```suggestion
        DataGridViewIntLinkedListElement? removeTargetNext = tmp2!.Next;
```

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewIntLinkedList.cs
```diff
@@ -148,7 +146,8 @@ public int IndexOf(int integer)
 
     public bool Remove(int integer)
     {
-        DataGridViewIntLinkedListElement tmp1 = null, tmp2 = _headElement;
+        DataGridViewIntLinkedListElement? tmp1 = null;
+        DataGridViewIntLinkedListElement? tmp2 = _headElement;
```
lonitra, line 150:
```suggestion
        DataGridViewIntLinkedListElement? removeTarget = _headElement;
```

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewIntLinkedList.cs
```diff
@@ -148,7 +146,8 @@ public int IndexOf(int integer)
 
     public bool Remove(int integer)
     {
-        DataGridViewIntLinkedListElement tmp1 = null, tmp2 = _headElement;
+        DataGridViewIntLinkedListElement? tmp1 = null;
```
lonitra, line 149:
```suggestion
        DataGridViewIntLinkedListElement? removeTargetPrevious = null;
```

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewIntLinkedList.cs
```diff
@@ -45,11 +43,11 @@ public DataGridViewIntLinkedList(DataGridViewIntLinkedList source)
             Debug.Assert(index < Count);
             if (_lastAccessedIndex == -1 || index < _lastAccessedIndex)
             {
-                DataGridViewIntLinkedListElement tmp = _headElement;
+                DataGridViewIntLinkedListElement tmp = _headElement!;
```
lonitra, line 46:
```suggestion
                DataGridViewIntLinkedListElement targetElement = _headElement!;
```

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCellLinkedList.cs
```diff
@@ -31,17 +31,17 @@ public DataGridViewCellLinkedList()
             Debug.Assert(index < _count);
             if (_lastAccessedIndex == -1 || index < _lastAccessedIndex)
```
lonitra, line 32: Please do the same here as I suggested in `DataGridViewIntLinkedList` below.

lonitra (commented):
Apologies for this longstanding PR. I've added some comments here that would help with readability and understanding of why things cannot be null at certain points if you have cycle to make these changes. Appreciate your patience 🙏

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView/DataGridViewIntLinkedList.cs
```diff
@@ -182,22 +182,26 @@ public bool Remove(int integer)
 
     public void RemoveAt(int index)
     {
-        DataGridViewIntLinkedListElement tmp1 = null, tmp2 = _headElement;
+        ArgumentOutOfRangeException.ThrowIfLessThan(index, 0);
+        ArgumentOutOfRangeException.ThrowIfGreaterThanOrEqual(index, Count);
+
+        DataGridViewIntLinkedListElement? removeTargetPrevious = null;
+        DataGridViewIntLinkedListElement? removeTarget = _headElement;
```
lonitra, line 189:
```suggestion
        // removeTarget will not null since the index is within the range of the list
        DataGridViewIntLinkedListElement removeTarget = _headElement!;
```

lonitra (commented):
couple small comments, otherwise LGTM!

lonitra (approved):
Thank you 🚀

PR #9543, opened 2023-07-22
Enable nullability in remaining AxHost members

gpetrou:
## Proposed changes

- Enable nullability in remaining AxHost members.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9543)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
@@ -2639,7 +2639,7 @@ public abstract unsafe partial class AxHost : Control, ISupportInitialize, ICust
                     continue;
                 }
 
-                if (!_properties.ContainsKey(propName))
+                if (!_properties.TryGetValue(propName, out PropertyDescriptor? propDesc))
                 {
                     if (propInfo is not null)
                     {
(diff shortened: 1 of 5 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
```diff
@@ -3677,7 +3678,7 @@ private AxContainer GetParentContainer()
                 _axContainer.AddControl(this);
             }
 
-            return _axContainer;
+            return _axContainer!;
```
halgab, line 3681: Since `_axContainer` and `_newParent` are always set in these lines, maybe you could change the null check to `_newParent` and remove the `!` on the last line?

gpetrou: Sorry, not sure I understand the suggestion.

halgab: Sorry I wasn't clear. `_axContainer` and `_newParent` are always set at the same time. So the null check on `_newParent` on line 3674 is actually checking if both fields were set. It is equivalent to check `_axContainer` for null instead, with the bonus that then you no longer need the `!` on line 3681

gpetrou:
@dreddy-work any thoughts on this one?

dreddy-work:
> @dreddy-work any thoughts on this one?

Haven't got a chance to go over it. Team will look into it in my absence.

gpetrou:
@lonitra, @JeremyKuhne any thoughts on this one?

JeremyKuhne:
@gpetrou sorry about the slow responses, we're pretty focused on closing down .NET 8 RC2.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
JeremyKuhne: Or maybe check both for null to satisfy the validation `(_newParent is null || _axContainer is null)`.

gpetrou: @JeremyKuhne I modified the check.

dreddy-work (approved):

PR #9548, opened 2023-07-22
Reduced nesting levels by inverting a few top-level `if`s

Lehonti:
In #9540 I was asked to rebase my changes on `Net9.0-Test`, so I closed that pull request and opened this one.

## Proposed changes

The logic of certain `if` blocks has been inverted. Namely:
- Some setters and some methods have `if` blocks that span more than a screen full of text, which makes it hard to see where the block ends. By inverting the logic, the `if` block becomes an `if` guard that is just one more element in the sequence of instructions.
- Also, some `if` blocks in certain setters could be better described and understood as guards, which is not very clear if the block spans all or at least most of the body of the setter. 

## Customer Impact

- No direct customer impact. The code is functionally identical

## Regression? 

- No

## Risk

- Low. Only safe, localized refactorings have been applied, aided by Visual Studio

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9548)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -34,6 +34,7 @@ pr:
       - internal/release/*
       - internal/experimental/*
       - feature/win32
+      - Net9.0-Test
   paths:
     include:
       - '*'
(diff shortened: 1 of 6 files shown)


dreddy-work:
FYI, you do not need to close the PR to rebase.

PR #9554, opened 2023-07-24
Enable nullability in BindToObject

gpetrou:
## Proposed changes

- Enable nullability in BindToObject.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9554)


src/System.Windows.Forms/src/System/Windows/Forms/Binding.BindToObject.cs
```diff
@@ -103,7 +104,7 @@ internal void SetBindingManagerBase(BindingManagerBase lManager)
         ///  Returns any data error info on the data source for the bound data field
         ///  in the current row
         /// </summary>
-        private string GetErrorText(object value)
+        private string GetErrorText(object? value)
```
dreddy-work, line 107: can `text` local nullable and avoid duplicate assignment?

gpetrou: Sorry, what do you mean?

src/System.Windows.Forms/src/System/Windows/Forms/Binding.BindToObject.cs
```diff
@@ -0,0 +1,251 @@
...
+
+            // The wait is over: DataSource is initialized.
+            _waitingOnDataSource = false;
+            _dataSourceInitialized = true;
+
+            // Rebind.
+            CheckBinding();
+        }
+
+        internal void SetBindingManagerBase(BindingManagerBase lManager)
```
halgab, line 80: Are you sure `lManager` can't be null? I was under the impression the `BindingManagerBase` property whose setter calls this method can be null.

src/System.Windows.Forms/src/System/Windows/Forms/Binding.BindToObject.cs
```diff
@@ -0,0 +1,251 @@
...
+                    }
+
+                    return type;
+                }
+
+                return FieldInfo?.PropertyType;
+            }
+        }
+
+        internal void SetValue(object value)
```
halgab, line 170: Same here, I thought the argument could be null on the call site.

src/System.Windows.Forms/src/System/Windows/Forms/Binding.BindToObject.cs
dreddy-work: local on line 109 can be declared as `string? text` ?

gpetrou: Actually, I see that text can never be null. Maybe we should remove `?? string.Empty;` instead from line 130?

dreddy-work: final change could alter functionality. Though Interface says default is empty string, I see implementations of this interface returning null either for `errorInfo[FieldInfo.Name]` or `errorInfo.Error`.  So, i would go with declaration as `string? text ` and return `text??striong.Empty`;

gpetrou: I kept the removal of `text` and added some null checks instead.

dreddy-work (approved):

PR #9558, opened 2023-07-24
Merge main into .Net9.0-Test branch

dreddy-work:
NA

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9558)


Tanya-Solyanik:
@LeafShi1 @SimonZhao888 @Epica3055  - please start merging Main to .Net9.0-Test daily, like Devendar did in this PR. Use "Merge Commit" option to make merging back to Main easier when Main is re-branded for NET9

JeremyKuhne:
@dreddy-work @Tanya-Solyanik I don't think we should have separate branches for test- there should only be 1 .NET 9 branch, otherwise we're going to have a multiplicative effect on FI/RIs.

JeremyKuhne (changes requested):
We should only have one .NET 9 branch.

Tanya-Solyanik:
@JeremyKuhne - this is not a Test branch, this is  The .NET 9.0 branch for development.

JeremyKuhne:
> This is .NET 9.0 branch and do not want to name it as just NET9.0 as it may be conflicting with some of the keyword information that infra team might put into YAMLS when they are close to make main as 9.0. This is temporary ( until we snap RC , which is almost a month away)

That seems extremely unlikely. Following our naming patterns `feature\9.0` makes the most sense.

We'll discuss this at the team meeting to settle on a name for this branch. Calling it "Test" is confusing and misleading. On a related note, we should not have any other personal developer branches in the runtime repo. It's extra noise for our external contributors. `feature\*` is the pattern to use.

dreddy-work:
Renaming branch to feature/6.0 and this pull request is not needed anymore as i am taking snap from todays bits.

PR #9577, opened 2023-07-27
Enable nullability in ToolStripSplitStackLayout

gpetrou:
## Proposed changes

- Enable nullability in ToolStripSplitStackLayout.


src/System.Windows.Forms/src/System/Windows/Forms/ToolStripSplitStackLayout.cs
```diff
@@ -364,7 +362,7 @@ private bool LayoutHorizontal()
             }
             else
             {
-                item.ParentInternal = (item.Placement == ToolStripItemPlacement.Overflow) ? toolStrip.OverflowButton.DropDown : null;
+                item!.ParentInternal = (item.Placement == ToolStripItemPlacement.Overflow) ? toolStrip.OverflowButton.DropDown : null;
```
dreddy-work, line 365: `else `is reached when `item `is null.   This change is making sure we are not altering functionally but just enable nullability. However, this doesn't seem right. Marking for next release to see if we can refactor code to better align.

gpetrou: The null check seems redundant. I could remove that if you want?

dreddy-work: yes please. if it is not required.

gpetrou: @dreddy-work I added a separate commit.

dreddy-work (approved):

PR #9588, opened 2023-07-27
Guard against re-entry in ActiveXImpl / AgileComPointer.

### LINKED ISSUE #9551: COleControlSite re-enter destructor
### .NET version

.NET8 preview6, and daily build, and build with latest code

Yes

It works fine before .NET8

### Issue description

See following call stack, `COleControlSite::~COleControlSite` is called again, it makes app crash.

image

### Steps to reproduce

I don't know OLE, it's not easy for me to make a test app, but I compare .NET8 and .NET7 source code. It may be helpful for you to understand the problem.

In .NET8, the second destructor is called from this line:
https://github.com/dotnet/winforms/blob/e9a35997645b440b605f0d975305e313a6808883/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs#L1705

Before .NET8, this line doesn't call destructor again:
https://github.com/dotnet/winforms/blob/32a4f45583a44acc811da8c452d4a5f673342893/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs#L1918 

So I guess it's probably related to @JeremyKuhne 's https://github.com/dotnet/winforms/pull/8420

JeremyKuhne:
Releasing a COM interface can lead to a call back. We need to clear fields before releasing so we don't attempt to release twice.

Fixes #9551 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9588)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
@@ -2261,8 +2261,12 @@ public partial class Control
             _inPlaceFrame = null;
             _inPlaceUiWindow?.Dispose();
             _inPlaceUiWindow = null;
-            _clientSite?.Dispose();
+
+            // Disposing the client site handle can get us called back with SetClientSite(null). We need to
+            // make sure that we clear the field before disposing it.
+            var clientSite = _clientSite;
             _clientSite = null;
+            clientSite?.Dispose();
         }
     }
 }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
@@ -865,11 +865,8 @@ public partial class Control
                 RECT posRect = default;
                 RECT clipRect = default;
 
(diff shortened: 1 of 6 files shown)


lonitra (approved):

src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
```diff
@@ -2261,8 +2261,12 @@ public void Dispose()
             _inPlaceFrame = null;
             _inPlaceUiWindow?.Dispose();
             _inPlaceUiWindow = null;
-            _clientSite?.Dispose();
+
+            // Disposing the client site handle can get us called back with SetClientSite(null). We need to
+            // make sure that we clear the field before disposing it.
+            var clientSite = _clientSite;
```
dreddy-work, line 2267: can't this be identified with `disposing` flag?

JeremyKuhne: We could add a flag, but then we would have to check everywhere we access the field for that flag. It's better to prevent the need for it entirely.

lonitra (approved):
Awesome 😎

PR #9589, opened 2023-07-27
Remove `RequiresPreviewFeatures` attributes from Command Binding APIs

KlausLoeffelmann:
We introduce Command Binding APIs under Preview in .NET 7 and aimed to remove this in the .NET 8 timeframe, when it did not cause any breaking changes. 

https://github.com/dotnet/winforms/pull/7143

We got some feedback, but mainly asking to invest further into data biding, like simplifying Converters or allow automatic marshalling from other threads to the UI thread when updating controls from the data sources. We want to address that in the .NET 9 release.

We found some issues in the Designer though, but they were due to the fact that the Out-Of-Proc Designer couldn't handle preview-assemblies per se and had nothing to do with the new APIs themselves. In so far, having those new APIs attributed with `RequiresPreviewFeatures` helped us to make the Designer more robust and able to deal with binding to properties in the runtime, which are attributed to be under preview.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9589)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
@@ -13,7 +13,6 @@ using IComDataObject = System.Runtime.InteropServices.ComTypes.IDataObject;
 
 namespace System.Windows.Forms;
 
-#pragma warning disable CA2252 // Suppress 'Opt in to preview features' (https://aka.ms/dotnet-warnings/preview-features)
 [DesignTimeVisible(false)]
 [Designer($"System.Windows.Forms.Design.ToolStripItemDesigner, {AssemblyRef.SystemDesign}")]
 [DefaultEvent(nameof(Click))]
@@ -3728,5 +3727,3 @@ public abstract partial class ToolStripItem : BindableComponent,
         return local is not null;
     }
 }
-#pragma warning restore CA2252
-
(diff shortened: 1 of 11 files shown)


dreddy-work:
What about removing the from `proj `file?

elachlan:
Test failures are from `Using directive is unnecessary.` warnings.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
```diff
@@ -475,11 +470,9 @@ public bool Available
     [SRDescription(nameof(SR.CommandComponentCommandParameterDescr))]
     public object? CommandParameter
     {
-        [RequiresPreviewFeatures]
         get => _commandParameter;
 
         // We need to opt into previre features here, because we calling a preview feature from the setter.
```
dreddy-work, line 475: remove comment.

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripItem.cs
```diff
@@ -2750,7 +2742,6 @@ private void OnAnimationFrameChanged(object? o, EventArgs e)
     ///  Raises the <see cref="ToolStripItem.CommandChanged"/> event.
     /// </summary>
     /// <param name="e">An empty <see cref="EventArgs"/> instance.</param>
-    [RequiresPreviewFeatures]
```
dreddy-work, line 2753: I see some comments still left out . for ex : line 2705.

KlausLoeffelmann:
I scanned the original PR again. 
I should have caught every comment now. 
Also, answered @Tanya-Solyanik remaining comments from the old PR.

Tanya-Solyanik (approved):
Looks good to me. But let Devendar finish his review too.

dreddy-work (approved):

PR #9603, opened 2023-07-28
Add ref to IOleClientSite in SetClientSite.

### LINKED ISSUE #9551: COleControlSite re-enter destructor

JeremyKuhne:
We were not adding a ref to the IOleClientSite pointer passed to SetClientSite, which was a mistake made when we moved this to directly utilizing COM pointers.

Fixes #9551

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9603)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
@@ -1697,11 +1697,22 @@ public partial class Control
         internal void SetClientSite(IOleClientSite* value)
         {
             DisposeHelper.NullAndDispose(ref _clientSite);
-            _clientSite = value is null ? null : new(value, takeOwnership: true);
 
-            _control.Site = _clientSite is not null
-                ? new AxSourcingSite(_control, _clientSite, "ControlAxSourcingSite")
-                : (ISite?)null;
+            if (value is not null)
+            {
+                // Callers don't increment the ref count when they pass IOleClientSite, it is up to us to do so as we're
+                // maintaining a reference to the pointer. Validated this behavior with the ATL/MFC sources.
+                //
+                // https://learn.microsoft.com/windows/win32/api/oleidl/nf-oleidl-ioleobject-setclientsite#notes-to-implementers
+
+                value->AddRef();
+                _clientSite = new(value, takeOwnership: true);
+                _control.Site = new AxSourcingSite(_control, _clientSite, "ControlAxSourcingSite");
+            }
+            else
+            {
+                _control.Site = null;
+            }
 
             // Get the ambient properties that effect us.
             using VARIANT property = GetAmbientProperty(PInvoke.DISPID_AMBIENT_UIDEAD);


src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
```diff
@@ -1697,11 +1697,22 @@ internal HRESULT SetAdvise(DVASPECT aspects, ADVF advf, IAdviseSink* pAdvSink)
...
-                : (ISite?)null;
+            if (value is not null)
+            {
+                // Callers don't increment the ref count when they pass IOleClientSite, it is up to us to do so as we're
+                // maintaining a reference to the pointer. Validated this behavior with the ATL/MFC sources.
+                //
+                // https://learn.microsoft.com/windows/win32/api/oleidl/nf-oleidl-ioleobject-setclientsite#notes-to-implementers
+
+                value->AddRef();
+                _clientSite = new(value, takeOwnership: true);
```
dreddy-work, line 1709: Q: it seems constructor is adding and removing `Ref` count (by design). How is this helping? Is the ref count added here is removed after this `SetClientSite` method returned?

Tanya-Solyanik: It is released [here](https://github.com/dotnet/winforms/blob/f1c767e3020148ed985dfaa903a15b525a2e9015/src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/AgileComPointer.cs#L48) because _clientSite is an AgileCOMPointer wrapper,

Tanya-Solyanik (approved):

src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
JeremyKuhne: The global interface table adds a ref count. By removing one after that we let the `AgileComPointer` "own" it. While I could technically pass `takeOwnership: false`, it would obscure the intent.

JeremyKuhne: Another way to look at this is that the code passing in the pointer should always "own" a reference already. That's the conceptual model. Taking advantage of the internal details confuses the API usage.

I'll continue to reflect on the API and will iterate if I can come up with cleaner paradigms.

PR #9607, opened 2023-07-29
Enable nullability in UndoEngine

halgab:
## Proposed changes

- Enable nullability in `UndoEngine` and add some helper methods to simplify the code

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9607)


src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.cs
```diff
@@ -351,24 +327,24 @@ private void OnComponentChanged(object sender, ComponentChangedEventArgs e)
...
+    private void OnComponentChanging(object? sender, ComponentChangingEventArgs e)
     {
         // Open a new unit unless there is already one open or we are currently executing a unit. If we need to create a unit, we will have to fabricate a good name.
-        if (_enabled && _executingUnit is null && _unitStack.Count == 0)
+        if (Enabled && _executingUnit is null && _unitStack.Count == 0)
         {
             string name;
 
-            if (e.Member is not null && e.Component is not null)
+            if (e.Component is null)
```
dreddy-work, line 337: I wouldn't change these blocks. it is more readable. What is the benefit here?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.cs
```diff
@@ -437,11 +411,11 @@ private void OnComponentRemoving(object sender, ComponentEventArgs e)
                 PropertyDescriptorCollection props = TypeDescriptor.GetProperties(comp);
                 foreach (PropertyDescriptor prop in props)
                 {
-                    if (prop.PropertyType.IsAssignableFrom(e.Component.GetType()) &&
+                    if (prop.PropertyType.IsInstanceOfType(e.Component) &&
```
dreddy-work, line 414: Why this change?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.cs
halgab: That way we don't have to check `e.Component` for null twice, so it's a little bit lighter, and as much readable. I'll rollback if you prefer the previous code

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.cs
halgab: It felt simpler, but I guess it's unnecessary. I'll rollback

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.cs
dreddy-work: Lets keep it. Readability is subjective but i do not see it is complex here.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.cs
dreddy-work: Yah. if we do not have full context on why one is better than other,  we should rollback.

halgab: Okay, I had another look. If I rollback, I have to handle the potential nullability of `e.Component`:

```suggestion
                    if (prop.PropertyType.IsAssignableFrom(e.Component!.GetType() &&
```
But `e.Component` is checked for null a few lines before (line 388), so the new code here is more defensive. Here is the code of `IsInstanceOfType` : 

`public virtual bool IsInstanceOfType([NotNullWhen(true)] object? o) => o == null ? false : IsAssignableFrom(o.GetType());`

So for me the options are:

1. Keep it that way
2. Rollback
3. Rollback and add a manual null check for `e.Component`

What do you think?

dreddy-work: in this case, `IsInstanceOfType` seems safer to take. Lets keep it.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.cs
```diff
@@ -549,15 +523,15 @@ public ReferencingComponent(IComponent component, MemberDescriptor member)
     /// </summary>
     protected class UndoUnit
     {
-        private List<UndoEvent> _events; // the list of events we've captured
-        private List<ChangeUndoEvent> _changeEvents; // the list of change events we're currently capturing.  Only valid until Commit is called.
-        private List<AddRemoveUndoEvent> _removeEvents; // the list of remove events we're currently capturing.  Only valid until a matching Removed is encountered.
-        private List<IComponent> _ignoreAddingList; // the list of objects that are currently being added.  We ignore change events between adding and added.
-        private List<IComponent> _ignoreAddedList; // the list of objects that are added. We do not serialize before state for change events that happen in the same transaction
+        private List<UndoEvent>? _events; // the list of events we've captured
```
dreddy-work, line 526: you can move this to a seperate file.

halgab: done

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.UndoUnit.cs
```diff
@@ -0,0 +1,793 @@
+﻿using System.Collections;
```
dreddy-work, line 1: Missing license file.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.UndoUnit.cs
```diff
@@ -0,0 +1,793 @@
...
+        private List<ChangeUndoEvent>? _changeEvents; // the list of change events we're currently capturing.  Only valid until Commit is called.
+        private List<AddRemoveUndoEvent>? _removeEvents; // the list of remove events we're currently capturing.  Only valid until a matching Removed is encountered.
+        private List<IComponent>? _ignoreAddingList; // the list of objects that are currently being added.  We ignore change events between adding and added.
+        private List<IComponent>? _ignoreAddedList; // the list of objects that are added. We do not serialize before state for change events that happen in the same transaction
+        private bool _reverse; // if true, we walk the events list from the bottom up
+        private readonly Dictionary<string, IContainer>? _lastSelection; // the selection as it was before we gathered undo info
+
+        public UndoUnit(UndoEngine engine, string? name)
+        {
+            name ??= string.Empty;
```
dreddy-work, line 25:
```suggestion
            Name = name ?? string.Empty;
```

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.UndoUnit.cs
```diff
@@ -0,0 +1,793 @@
...
+
+            Debug.WriteLineIf(s_traceUndo.TraceVerbose, $"UndoEngine: Creating undo unit '{name}'");
+
+            Name = name;
```
dreddy-work, line 29:
```suggestion
            Name = name ?? string.Empty;

            Debug.WriteLineIf(s_traceUndo.TraceVerbose, $"UndoEngine: Creating undo unit '{Name}'");
```

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.UndoUnit.cs
```diff
@@ -0,0 +1,793 @@
...
+                AddEvent(evt);
+                _removeEvents.Add(evt);
+            }
+            catch (TargetInvocationException) { }
+        }
+
+        /// <summary>
+        ///  The undo engine will cal this on the active undo unit in response to a component rename event.
+        /// </summary>
+        public virtual void ComponentRename(ComponentRenameEventArgs e)
```
dreddy-work, line 342: convert to expressional body?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.UndoUnit.cs
```diff
@@ -0,0 +1,793 @@
...
+        /// </summary>
+        public void Undo()
+        {
+            Debug.WriteLineIf(s_traceUndo.TraceVerbose, $"UndoEngine: Performing undo '{Name}'");
+            UndoUnit? savedUnit = UndoEngine._executingUnit;
+            UndoEngine._executingUnit = this;
+            DesignerTransaction? transaction = null;
+            try
+            {
+                if (savedUnit is null)
```
dreddy-work, line 374: when does this `null`? wouldn't it throw on line 370 if it is?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.UndoUnit.cs
```diff
@@ -0,0 +1,793 @@
...
+                }
+            }
+
+            _reverse = !_reverse;
+        }
+
+        /// <summary>
+        ///  This undo event handles addition and removal of components.
+        /// </summary>
+        private sealed class AddRemoveUndoEvent : UndoEvent
```
dreddy-work, line 503: Move to seperate file?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.UndoUnit.cs
```diff
@@ -0,0 +1,793 @@
...
+            /// <summary>
+            ///  Actually performs the undo action.
+            /// </summary>
+            public override void Undo(UndoEngine engine)
+            {
+                if (NextUndoAdds)
+                {
+                    Debug.WriteLineIf(s_traceUndo.TraceVerbose, $"UndoEngine: ---> Adding '{_componentName}'");
+                    // We need to add this component.  To add it, we deserialize it and then we add it to the designer host's container.
+                    if (engine.GetRequiredService(typeof(IDesignerHost)) is IDesignerHost host)
```
dreddy-work, line 563: should this be ? or even move out of the `if-else` block?
```suggestion
                    IDesignerHost host = engine.GetRequiredService<IDesignerHost>();
```

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.UndoUnit.cs
```diff
@@ -0,0 +1,793 @@
...
+                    {
+                        host.DestroyComponent(component);
+                    }
+                }
+
+                NextUndoAdds = !NextUndoAdds;
+            }
+        }
+
+        private sealed class ChangeUndoEvent : UndoEvent
```
dreddy-work, line 587: move to seperate file.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.UndoUnit.cs
```diff
@@ -0,0 +1,793 @@
...
+
+                Debug.WriteLineIf(s_traceUndo.TraceVerbose, $"UndoEngine: ---> Creating change undo event for '{_componentName}'");
+                if (serializeBeforeState)
+                {
+                    Debug.WriteLineIf(s_traceUndo.TraceVerbose, $"UndoEngine: ---> Saving before snapshot for change to '{_componentName}'");
+                    _before = Serialize(engine, OpenComponent!, _member);
+                }
+            }
+
+            public ComponentChangingEventArgs ComponentChangingEventArgs
```
dreddy-work, line 614: nit: expressionalbody whereever we can.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.UndoUnit.cs
```diff
@@ -0,0 +1,793 @@
...
+                    {
+                        engine._serializationService.DeserializeTo(_before, host.Container);
+                    }
+                }
+
+                (_after, _before) = (_before, _after);
+            }
+        }
+
+        private sealed class RenameUndoEvent : UndoEvent
```
dreddy-work, line 739: same here and below. Moving classes to separate file.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.UndoUnit.cs
halgab: `_executingUnit` starts out as `null`, so `savedUnit` can actually be null. Not sure why it would throw though

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.UndoUnit.cs
halgab: Sure, but the UndoEvent classes are 4 sub-sub-classes. I'm not sure there is precedent of spinning off classes this deep. Should I proceed anyway? Other options would be

- putting all those 4 classes in a single file
- make those sub-sub classes only subclasses (at the same level as `UndoUnit`

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.UndoUnit.cs
halgab: I didn't make this change because the generic version of `GetRequiredService` is slighly different than the non-generic one:

- `GetRequiredService<T>` throws if we don't find a service of type T
- `GetRequiredService` throws only if `GetService` returns null. It's fine with it returning an object of a different type than the one requested. I have no idea if it's actually possible though (it's certainly not the contract of `GetService`)

I figured it would be a behaviour change here to use the generic version (even though I'd very much like to!), so I chose to live this as is

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.UndoUnit.cs
halgab: see https://github.com/dotnet/winforms/pull/9607#discussion_r1284162467

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.UndoUnit.cs
dreddy-work: We do not want to change the scope of those classes. Moving them into separate file  ( and file name syntax follows) should be fine.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.UndoUnit.cs
dreddy-work: Though the difference, i do not expect it to be the case.  Happy to go with Net9.0 as well but simplifying the code is an improvement here.

halgab: Okay, I made the change

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.UndoUnit.cs
halgab: that's done

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/UndoEngine.UndoUnit.cs
dreddy-work: line numbers changed. I mean this `UndoEngine._executingUnit = this;`. Wasn't it warning if `UndoEngine._executingUnit` can be null?

halgab: Setting a nullable field with a non-null value is a valid scenario, so I'm not sure why this should warn. And as a matter of fact it doesn't. Still this field should remain null for the reasons stated above. Does that answer your question?

PR #9608, opened 2023-07-29
Enable nullability in CursorEditor

halgab:
## Proposed changes

- Enable nullability in `CursorEditor`

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9608)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Drawing/Design/CursorEditor.CursorUI.cs
+++ b/src/System.Windows.Forms.Design/src/System/Drawing/Design/CursorEditor.CursorUI.cs
@@ -2,8 +2,6 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-#nullable disable
-
 using System.ComponentModel;
 using System.Windows.Forms;
 using System.Windows.Forms.Design;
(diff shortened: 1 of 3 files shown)


src/System.Windows.Forms.Design/src/System/Drawing/Design/CursorEditor.CursorUI.cs
```diff
@@ -36,26 +33,26 @@ public CursorUI()
...
                 }
             }
         }
 
-        public object Value => _value;
+        public object? Value { get; private set; }
 
         public void End()
         {
             _editorService = null;
```
dreddy-work, line 47: i would change this assignment as `_editorService = null!` and make `_editorService` not nullable.

halgab: `_editorService` is not set in the constructor though, so I would need to change line 18 to:

> `private IWindowsFormsEditorService _editorService = null!;`

Should I make the change anyway?

dreddy-work: hmm. its assigned in `Start`. That doesn't help. lets keep as-in now.

dreddy-work (approved):

dreddy-work:
Please resolve conflict.

halgab:
Should be good now

dreddy-work (approved):

PR #9613, opened 2023-07-30
Checking the dictionary once is enough

gero3:
## Proposed changes

 - simplified source code

## Customer Impact

- None

## Regression? 

- No

## Risk

- None

## Test methodology 

- I ran the tests

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9613)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
@@ -30,7 +30,7 @@ public class KeysConverter : TypeConverter, IComparer
 
     private void AddLocalizedKeyNames(CultureInfo cultureInfo)
     {
-        if (CultureToDisplayOrder.ContainsKey(cultureInfo))
+        if (CultureToDisplayOrder.ContainsKey(cultureInfo) && CultureToKeyName.ContainsKey(cultureInfo))
         {
             return;
         }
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/KeysConverter.cs
```diff
@@ -30,7 +30,7 @@ private void Initialize()
 
     private void AddLocalizedKeyNames(CultureInfo cultureInfo)
     {
-        if (CultureToDisplayOrder.ContainsKey(cultureInfo) && CultureToDisplayOrder.ContainsKey(cultureInfo))
+        if (CultureToDisplayOrder.ContainsKey(cultureInfo))
```
dmitrii-drobotov, line 33: Thanks for bringing up attention to this, but I believe it was intended to do the following:
```suggestion
        if (CultureToDisplayOrder.ContainsKey(cultureInfo) && CultureToKeyName.ContainsKey(cultureInfo))
```

This method inserts values to both of these dictionaries, so it makes sense to check them both.

gero3: That would make more sense indeed

dmitrii-drobotov (approved):
Thank you!

PR #3644, opened 2020-07-27
Add caching for Pen/SoldBrush

JeremyKuhne:
Adds base classes and caches for System.Drawing pens and brushes. Signficantly reduces allocations for Graphics drawing.

- Remove other instances of Brush/Pen caching
- Write helpers to retain the orginal Color when finding nearest color (allowing existing caches to be used)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/winforms/pull/3644)

PR #9621, opened 2023-08-01
Refactor `DataGridView` to use `SinglyLinkedList` instead of `DataGridViewIntLinkedList`

elachlan:
Based on discussion at https://github.com/dotnet/winforms/pull/9541#discussion_r1280115417

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9621)


elachlan:
CC: @gpetrou @dreddy-work

Most of the logic was easy to convert and probably could get refactored into a helper method.

The main method to look at is `ProcessDeleteKey` as I am unsure if I got the logic totally right there. I tried to unwrap it a bit as well to reduce nesting.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.Methods.cs
```diff
@@ -12907,7 +12911,7 @@ private void OnCellMouseDown(HitTestInfo hti, bool isShiftDown, bool isControlDo
                                 Debug.Assert(_selectedBandIndexes.Count <= 1);
                                 if (_selectedBandIndexes.Count > 0)
```
gpetrou, line 12912: I guess all these checks should now be `_selectedBandIndexes.First is not null` to avoid null forgiving operators when this huge file is annotated?

elachlan: We could maybe write a `TryGetFirst` extension method to make it a bit nicer or just use pattern matching?

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.Methods.cs
```diff
@@ -20563,98 +20572,101 @@ protected bool ProcessDeleteKey(Keys keyData)
             {
                 case DataGridViewSelectionMode.FullRowSelect:
                 case DataGridViewSelectionMode.RowHeaderSelect:
-                    int remainingSelectedRows = 0;
```
elachlan, line 20566: @JeremyKuhne it would be great if you could review this function. I am not totally confident on the unwrap and making sure the loop runs correctly.

halgab:
Did you also evaluate [this class](https://github.com/dotnet/winforms/blob/main/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/SinglyLinkedList.cs) as a replacement for those custom linked lists? It looks closer to what is already used.

halgab:
I have no idea... I just thought that:

1. It was added for a reason, apparently after `LinkedList<T>` appeared in the BCL (more specifically in #3644)
2. It's singly linked while `LinkedList<T>` is doubly linked, which is apparently overkill here
3. Since it's internal, we can more easily add methods to fit our needs here

On the other hand, `LinkedList<T>` is of course the more used class so it's a safe bet.

elachlan:
I wonder why runtime never added `SinglyLinkedList<T>`? I'll refactor this to use it since (i think) we are always working in the forward direction.

elachlan:
@JeremyKuhne In `SinglyLinkedList<T>` was `Enumerable<T>` avoided on purpose? It seems super simple and I don't want to mess with it in case of perf changes.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/SinglyLinkedList.cs
```diff
@@ -56,13 +68,116 @@ public Node AddLast(T value)
...
+    {
+        Node? node = First;
+        EqualityComparer<T> c = EqualityComparer<T>.Default;
+        if (node is not null)
+        {
+            if (value is not null)
+            {
+                while (node is not null)
+                {
+                    if (c.Equals(node!.Value, value))
```
halgab, line 101:
```suggestion
                    if (c.Equals(node.Value, value))
```

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/SinglyLinkedList.cs
```diff
@@ -56,13 +68,116 @@ public Node AddLast(T value)
...
+                    }
+
+                    node = node.Next;
+                }
+            }
+            else
+            {
+                while (node is not null)
+                {
+                    if (node!.Value is null)
```
halgab, line 113:
```suggestion
                    if (node.Value is null)
```

Also, in this loop and the one above, you could put this `if` into the `while`, and return `node` at the end of the method, but that's more of a style preference

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/SinglyLinkedList.cs
```diff
@@ -56,13 +68,116 @@ public Node AddLast(T value)
...
+    internal void InternalRemoveNode(Node node)
+    {
+        Node? tmp1 = null, tmp2 = First;
+        while (tmp2 is not null)
+        {
+            if (tmp2 == node)
+            {
+                break;
+            }
+
```
halgab, line 152:
```suggestion
        while (tmp2 is not null && tmp2 != node)
        {
```

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.Methods.cs
```diff
@@ -3649,17 +3649,15 @@ protected void ClearSelection(int columnIndexException, int rowIndexException, b
                 case DataGridViewSelectionMode.FullRowSelect:
                 case DataGridViewSelectionMode.RowHeaderSelect:
                     {
-                        int bandIndex = 0;
-                        while (bandIndex < _selectedBandIndexes.Count)
+                        var enumerator = _selectedBandIndexes.GetEnumerator();
```
halgab, line 3652: maybe replace `var` with the actual type?

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.Methods.cs
```diff
@@ -3649,17 +3649,15 @@ protected void ClearSelection(int columnIndexException, int rowIndexException, b
...
-                                SetSelectedRowCore(_selectedBandIndexes[bandIndex], false);
-                            }
-                            else
+                            SinglyLinkedList<int>.Node currentNode = enumerator.Current;
+
+                            if (currentNode.Value != rowIndexException)
                             {
-                                bandIndex++;
+                                // deselect currently selected column
+                                SetSelectedColumnCore(currentNode.Value, false);
```
halgab, line 3660: if I understand correctly, `SetSelectedColumnCore` is going to modify `_selectedBandIndexes` by removing `currentNode.Value`. The previous version of the code handled that by not increasing the loop index in that situation. With the new code,  are we confortable with iterating over a collection we are modifying?

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.Methods.cs
```diff
@@ -3674,17 +3672,15 @@ protected void ClearSelection(int columnIndexException, int rowIndexException, b
...
+
+                            if (currentNode.Value != columnIndexException)
                             {
                                 // deselect currently selected column
-                                SetSelectedColumnCore(_selectedBandIndexes[bandIndex], false);
-                            }
-                            else
-                            {
-                                bandIndex++;
+                                SetSelectedColumnCore(currentNode.Value, false);
```
halgab, line 3683: same question as above

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.Methods.cs
```diff
@@ -12644,16 +12639,15 @@ private void OnCellMouseDown(HitTestInfo hti, bool isShiftDown, bool isControlDo
...
+
+                                        if (currentNode.Value != hti._col)
                                         {
                                             // deselect currently selected column
-                                            SetSelectedColumnCore(_selectedBandIndexes[bandIndex], false);
-                                        }
-                                        else
-                                        {
-                                            bandIndex++;
+                                            SetSelectedColumnCore(currentNode.Value, false);
```
halgab, line 12650: same question

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.Methods.cs
```diff
@@ -12833,16 +12826,15 @@ private void OnCellMouseDown(HitTestInfo hti, bool isShiftDown, bool isControlDo
...
+
+                                        if (currentNode.Value != hti._row)
                                         {
                                             // deselect currently selected row
-                                            SetSelectedRowCore(_selectedBandIndexes[bandIndex], false);
-                                        }
-                                        else
-                                        {
-                                            bandIndex++;
+                                            SetSelectedRowCore(currentNode.Value, false);
```
halgab, line 12837: same question

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.Methods.cs
```diff
@@ -14323,16 +14314,15 @@ private void OnColumnHeaderMouseDown(HitTestInfo hti, bool isShiftDown, bool isC
...
+
+                                        if (currentNode.Value != hti._col)
                                         {
                                             // deselect currently selected column
-                                            SetSelectedColumnCore(_selectedBandIndexes[bandIndex], false);
-                                        }
-                                        else
-                                        {
-                                            bandIndex++;
+                                            SetSelectedColumnCore(currentNode.Value, false);
```
halgab, line 14325: same question

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.Methods.cs
```diff
@@ -17736,25 +17714,22 @@ internal void OnRemovingColumn(DataGridViewColumn dataGridViewColumn, out Point
...
                 {
-                    int columnIndexSelected = _selectedBandIndexes[columnEntry];
-                    if (columnIndex == columnIndexSelected)
+                    SinglyLinkedList<int>.Node currentNode = enumerator.Current;
+
+                    if (columnIndex == currentNode.Value)
                     {
-                        _selectedBandIndexes.RemoveAt(columnEntry);
-                        columnEntries--;
+                        _selectedBandIndexes.Remove(currentNode);
```
halgab, line 17724: same question (it's even more obvious here)

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.Methods.cs
```diff
@@ -17855,48 +17830,42 @@ internal void OnRemovingRow(int rowIndexDeleted, out Point newCurrentCell, bool
...
                     {
-                        int rowIndex = _selectedBandSnapshotIndexes[rowEntry];
-                        if (rowIndexDeleted == rowIndex)
+                        SinglyLinkedList<int>.Node currentNode = enumerator.Current;
+
+                        if (rowIndexDeleted == currentNode.Value)
                         {
-                            _selectedBandSnapshotIndexes.RemoveAt(rowEntry);
-                            rowEntries--;
+                            _selectedBandSnapshotIndexes.Remove(currentNode);
```
halgab, line 17861: same question

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.Methods.cs
```diff
@@ -18627,16 +18595,14 @@ private void OnRowHeaderMouseDown(HitTestInfo hti, bool isShiftDown, bool isCont
...
+                                        SinglyLinkedList<int>.Node currentNode = enumerator.Current;
+                                        if (currentNode.Value != hti._row)
                                         {
                                             // deselect currently selected row
-                                            SetSelectedRowCore(_selectedBandIndexes[bandIndex], false);
-                                        }
-                                        else
-                                        {
-                                            bandIndex++;
+                                            SetSelectedRowCore(currentNode.Value, false);
```
halgab, line 18605: same question. I'll stop flagging the occurences now

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/SinglyLinkedList.cs
```diff
@@ -56,13 +68,116 @@ public Node AddLast(T value)
...
+        return Find(value) is not null;
+    }
+
+    public Node? Find(T value)
+    {
+        Node? node = First;
+        EqualityComparer<T> c = EqualityComparer<T>.Default;
+        if (node is not null)
+        {
+            if (value is not null)
```
dreddy-work, line 97: why can't you just check null `Value` with  null of `Node.Value`? why do you need to go to `else `clause and do that?

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/SinglyLinkedList.cs
```diff
@@ -56,13 +68,116 @@ public Node AddLast(T value)
...
+
+        return null;
+    }
+
+    public bool Remove(T value)
+    {
+        Node? node = Find(value);
+        if (node is not null)
+        {
+            InternalRemoveNode(node);
```
dreddy-work, line 131: isn't this iterating twice?

dreddy-work:
Thanks @elachlan for making time for this. It may take some time to get this ironed out due to #9652. I marked this for 9.0 but please continue the discussion.

JeremyKuhne:
>  In SinglyLinkedList<T> was Enumerable<T> avoided on purpose? It seems super simple and I don't want to mess with it in case of perf changes.

Yes, it's perf. You're allocating and boxing and it wasn't necessary.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/SinglyLinkedList.cs
```diff
@@ -5,9 +5,21 @@
 
 namespace System.Windows.Forms;
 
-internal class SinglyLinkedList<T>
+internal sealed class SinglyLinkedList<T>
```
JeremyKuhne, line 8: There should be new unit tests for anything added here.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/SinglyLinkedList.cs
```diff
@@ -56,13 +68,116 @@ public Node AddLast(T value)
...
+    }
+
+    public sealed class Node
     {
         public Node(T value) => Value = value;
         public Node? Next { get; set; }
         public T Value { get; set; }
 
         public static implicit operator T(Node? node) => node is null ? default! : node.Value;
+        public void Invalidate() => Next = null;
```
JeremyKuhne, line 180: I don't see why you're adding this.

JeremyKuhne:
I'm a little concerned about this change. What is the win we're getting in DataGridView? What is the impact going to be to the current performance characteristics?

I did very carefully make the collection to support our caching (with significant perf testing). This collection should remain fine-tuned for performance. I would expect that we have adequate unit tests and that we've benchmarked the new methods to make sure they're efficient.

elachlan:
I think the ultimate goal here is to remove the special purpose classes `DataGridViewIntLinkedList` and `DataGridViewCellLinkedList` to avoid repeating code. We also use `LinkedList<T>` in a few `AccessibleObject` classes where double linking isn't required. Ideally runtime should provide a `SinglyLinkedList<T>`. Do you think this is something runtime would be willing to look at in .NET9?

I'd rather leave your class alone if its specifically tuned for the caching scenario.

elachlan:
@JeremyKuhne Do you think runtime would add a `SinglyLinkedList<T>` or is it too specialized?

JeremyKuhne:
@elachlan Sorry about the delay, scrambling around trying to make sure we're closing down .NET 8 successfully

> Do you think this is something runtime would be willing to look at in .NET9?

I don't know what the appetite for one would be. I'd look to see if someone has already suggested, then consider suggesting it if you're interested in seeing this happen.

One other possibility is to just create one here first and use that to judge the usefulness.

> I'd rather leave your class alone if its specifically tuned for the caching scenario.

I think that's wise

dreddy-work:
@elachlan, please rebase this to main. Feature branch is being closed.

dreddy-work:
Sorry, it is automatically closed when branch is deleted. You may need to rebase and reopen it.

lonitra:
@elachlan it seems like we wanted to take this PR before merging https://github.com/dotnet/winforms/pull/9541 , but this was closed as a result of deleting old feature/9.0 branch. Do you have cycle to rebase to main and reopen?

PR #9622, opened 2023-08-01
use generic overload of Enum.IsDefined to avoid boxing

halgab:
## Proposed changes

- Replace usage of `Enum.IsDefined` by `Enum.IsDefined<T>` across codebase

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9622)


elachlan:
@halgab how did you identify these? Is there an analyzer we need to enable?

halgab:
I found one occurrence, which led me to do a systematic search and replace. I don't know about any analyser that provides this service unfortunately.

dreddy-work (approved):

PR #9624, opened 2023-08-01
fix [Accessibility] SubItems displayed incorrectly in Inspect or AccessibilityInsights after a column is removed from the ListView. #7492

### LINKED ISSUE #7492: [Accessibility] SubItems displayed incorrectly in Inspect or AccessibilityInsights after a column is removed from the ListView.
### .NET version

.NET SDK 7.0.100-preview.7.22370.3

Yes

1. This issue doesn't repro on .NET Core 3.1.
2. The subItems doesn't display in Inspect in .NET 5.0.
3. This issue all can reproduced on .NET 6.0 & 7.0.

### Issue description

The SubItems displayed incorrectly in Inspect and Accessibility Insight tool after remove a column of ListView.
.NET Core:
Repro

.NET Framework:
FX

### Steps to reproduce

1. Extract and open attached app: 
[TestApp.zip](https://github.com/dotnet/winforms/files/9180066/TestApp.zip)
2. Build and run this app.
3. Launch Inspect or Accessibility Insight tool.
4. Focus on the subItems --- these subItems displayed correctly in Inspect.
5. Click "Remove Column" button.
6. Refresh Inspect tool.
7. Focus on the subItems --- these subItems displayed incorrectly in Inspect.

ReproSteps

Epica3055:
Fixes #7492

## Proposed changes

- 
- Change the way to get childAO of a ListViewItem
- 

## Regression? 

- Yes 

## Test methodology 

- 
- Manual and Automation 
- 

## Test environment(s) 

8.0.0-rc.1.23376.5

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9624)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
@@ -3881,6 +3881,19 @@ public partial class ListView : Control
         }
     }
 
+    internal void UpdateColumnHeaderCorrespondingListViewSubItemIndex()
+    {
+        if (_columnHeaders is null)
+        {
+            return;
+        }
+
+        for (int i = 0; i < _columnHeaders.Length; i++)
+        {
+            _columnHeaders[i]._correspondingListViewSubItemIndex = i;
+        }
+    }
+
     /// <summary>
     ///  Inserts a new Column into the ListView
     /// </summary>
@@ -3978,6 +3991,8 @@ public partial class ListView : Control
         {
             RealizeAllSubItems();
         }
+
+        UpdateColumnHeaderCorrespondingListViewSubItemIndex();
 
         return ch;
     }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.ListViewItemDetailsAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.ListViewItemDetailsAccessibleObject.cs
@@ -30,7 +30,8 @@ public partial class ListViewItem
         /// </summary>
         /// <param name="accessibleChildIndex">The index of the child <see cref="AccessibleObject"/>.</param>
         /// <returns>The index of an owning <see cref="ListViewSubItem"/>'s object.</returns>
-        private int AccessibleChildToSubItemIndex(int accessibleChildIndex) => HasImage ? _owningListView.Columns[accessibleChildIndex - 1]._correspondingListViewSubItemIndex
+        private int AccessibleChildToSubItemIndex(int accessibleChildIndex) => HasImage
+            ? _owningListView.Columns[accessibleChildIndex - 1]._correspondingListViewSubItemIndex
             : _owningListView.Columns[accessibleChildIndex]._correspondingListViewSubItemIndex;
 
         internal override UiaCore.IRawElementProviderFragment? FragmentNavigate(UiaCore.NavigateDirection direction)
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.ListViewItemDetailsAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.ListViewItemDetailsAccessibleObject.cs
@@ -30,7 +30,8 @@ public partial class ListViewItem
         /// </summary>
         /// <param name="accessibleChildIndex">The index of the child <see cref="AccessibleObject"/>.</param>
         /// <returns>The index of an owning <see cref="ListViewSubItem"/>'s object.</returns>
-        private int AccessibleChildToSubItemIndex(int accessibleChildIndex) => HasImage ? accessibleChildIndex - 1 : accessibleChildIndex;
+        private int AccessibleChildToSubItemIndex(int accessibleChildIndex) => HasImage ? _owningListView.Columns[accessibleChildIndex - 1]._correspondingListViewSubItemIndex
+            : _owningListView.Columns[accessibleChildIndex]._correspondingListViewSubItemIndex;
 
         internal override UiaCore.IRawElementProviderFragment? FragmentNavigate(UiaCore.NavigateDirection direction)
         {
@@ -107,7 +108,16 @@ public partial class ListViewItem
             }
 
             int subItemIndex = _owningItem.SubItems.IndexOf(subItemAccessibleObject.OwningSubItem);
-            int accessibleChildIndex = SubItemToAccessibleChildIndex(subItemIndex);
+            int accessibleChildIndex = InvalidIndex;
+            for (int i = 0; i < _owningListView.Columns.Count; i++)
+            {
+                if (_owningListView.Columns[i]._correspondingListViewSubItemIndex == subItemIndex)
+                {
+                    accessibleChildIndex = i + (HasImage ? 1 : 0);
+                    break;
+                }
+            }
+
             return accessibleChildIndex > LastChildIndex ? InvalidIndex : accessibleChildIndex;
         }
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.ListViewItemDetailsAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.ListViewItemDetailsAccessibleObject.cs
@@ -126,7 +126,7 @@ public partial class ListViewItem
         // when there is no ListViewSubItem, but the cell for it is displayed and the user can interact with it.
         internal AccessibleObject? GetDetailsSubItemOrFake(int subItemIndex)
         {
-            int accessibleChildIndex = SubItemToAccessibleChildIndex(subItemIndex);
+            int accessibleChildIndex = HasImage ? subItemIndex + 1 : subItemIndex;
             return GetDetailsSubItemOrFakeInternal(accessibleChildIndex);
         }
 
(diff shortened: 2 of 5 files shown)


LeafShi1 (commented):
Please add a before and after comparison animation and description

SimonZhao888 (approved):

src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.ListViewItemDetailsAccessibleObject.cs
```diff
@@ -107,7 +108,16 @@ internal override int GetChildIndex(AccessibleObject? child)
             }
 
             int subItemIndex = _owningItem.SubItems.IndexOf(subItemAccessibleObject.OwningSubItem);
-            int accessibleChildIndex = SubItemToAccessibleChildIndex(subItemIndex);
```
Tanya-Solyanik, line 110: Is `SubItemToAccessibleChildIndex` used elsewhere in the code? Or should this method be removed?

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
```diff
@@ -3881,6 +3881,19 @@ private unsafe void InvalidateColumnHeaders()
...
+    internal void UpdateColumnHeaderCorrespondingListViewSubItemIndex()
+    {
+        if (_columnHeaders is null)
+        {
+            return;
+        }
+
+        for (int i = 0; i < _columnHeaders.Length; i++)
+        {
+            _columnHeaders[i]._correspondingListViewSubItemIndex = i;
```
Tanya-Solyanik, line 3893: It looks like `_correspondingListViewSubItemIndex ` is always equal to the column header index in the array. Do we really need this field?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItem.ListViewItemDetailsAccessibleObjectTests.cs
```diff
@@ -155,4 +156,69 @@ public void ListViewItemDetailsAccessibleObject_HasImage_ReturnsExpected(bool ad
         Assert.Equal(addImages, accessibleObject.HasImage);
         Assert.False(listView.IsHandleCreated);
     }
+
+    private ListView GetListView()
+    {
+        ListView listView1 = new();
+        ColumnHeader columnHeader1 = new();
+        ColumnHeader columnHeader2 = new();
+        ColumnHeader columnHeader3 = new();
```
Tanya-Solyanik, line 165: Please make sure that column headers are disposed

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewItem.ListViewItemDetailsAccessibleObjectTests.cs
```diff
@@ -155,4 +156,69 @@ public void ListViewItemDetailsAccessibleObject_HasImage_ReturnsExpected(bool ad
...
+        using ImageList imageList1 = new();
+        using Image image = new Bitmap(1, 1);
+        imageList1.Images.Add(image);
+        listView1.LargeImageList = imageList1;
+        listView1.SmallImageList = imageList1;
+
+        form.Controls.Add(listView1);
+        form.Show();
+
+        AccessibleObject listItemAO = listView1.AccessibilityObject.GetChild(1);
```
Tanya-Solyanik, line 197: Please avoid abbreviations in variable names.

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
```diff
@@ -3979,6 +3992,8 @@ internal ColumnHeader InsertColumn(int index, ColumnHeader ch, bool refreshSubIt
             RealizeAllSubItems();
         }
 
+        UpdateColumnHeaderCorrespondingListViewSubItemIndex();
```
Tanya-Solyanik, line 3995: Why isn't this method called when columns are removed?

Tanya-Solyanik (commented):
I added some questions

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
Epica3055: No, `_correspondingListViewSubItemIndex ` is not always equal to the column header index in the array. 

The cause of This ([7492](https://github.com/dotnet/winforms/issues/7492))  bug is 
say, there is a ListView with 3 columns,two rows,each row has 3 items.
image

When col1 column is removed, ListView is re-rendered , but SubItemList didn't remove the data(Row1-1, Row2-1). 
image


then I want to access this AccessibleObject. This is our old code logic: `listview.Items[0].SubItems[0]`, it is actually Row1-1. what I expected is Row1-2.
image

How I fix this:

I add `_correspondingListViewSubItemIndex ` to column. When ListView is rendered first , `_correspondingListViewSubItemIndex `  is  equal to the column header index in the array.

Then col1 column is removed. This is our new code logic: `listview.Items[0].SubItems[ listview.Columns[0]._correspondingListViewSubItemIndex   ]`, this result should be what we expect.

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
Epica3055: Please see the comments above

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
Tanya-Solyanik: I think each subitem should maintain its DisplayIndex, like the ColumnHeaders do

src/System.Windows.Forms/src/System/Windows/Forms/ListViewItem.ListViewItemDetailsAccessibleObject.cs
```diff
@@ -30,7 +30,8 @@ public ListViewItemDetailsAccessibleObject(ListViewItem owningItem) : base(ownin
         /// </summary>
         /// <param name="accessibleChildIndex">The index of the child <see cref="AccessibleObject"/>.</param>
         /// <returns>The index of an owning <see cref="ListViewSubItem"/>'s object.</returns>
-        private int AccessibleChildToSubItemIndex(int accessibleChildIndex) => HasImage ? accessibleChildIndex - 1 : accessibleChildIndex;
+        private int AccessibleChildToSubItemIndex(int accessibleChildIndex) => HasImage ? _owningListView.Columns[accessibleChildIndex - 1]._correspondingListViewSubItemIndex
+            : _owningListView.Columns[accessibleChildIndex]._correspondingListViewSubItemIndex;
```
Tanya-Solyanik, line 34:
```suggestion
        private int AccessibleChildToSubItemIndex(int accessibleChildIndex) => HasImage 
            ? _owningListView.Columns[accessibleChildIndex - 1]._correspondingListViewSubItemIndex
            : _owningListView.Columns[accessibleChildIndex]._correspondingListViewSubItemIndex;
```

Tanya-Solyanik:
Please test this change with all possible ListView Views

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
```diff
@@ -3881,6 +3881,19 @@ private unsafe void InvalidateColumnHeaders()
         }
     }
 
+    internal void UpdateColumnHeaderCorrespondingListViewSubItemIndex()
```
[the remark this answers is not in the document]
Epica3055: No, It is called from `ListView.InsertColumn()`

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
Epica3055: >What happens if you remove column, and then add a different column? On Add indices will be recalculated and the subitem whose column had been removed, will be assigned to a new column.

Yeah, Adding a different column would cause recalculating  `_correspondingListViewSubItemIndex` 

>I think each subitem should maintain its DisplayIndex, like the ColumnHeaders do

You were suggesting another solution right?
I consider it before. What am thinking is: what if it is a very huge ListView. Would adding an extra field (DisplayIndex) cost a lot of Memory? Maybe I am thinking too much, maybe this additional memory means nothing to modern device.

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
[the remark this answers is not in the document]
Tanya-Solyanik: So `AddRange` will break your fix?

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
Tanya-Solyanik: Yes, we need a different solution, because it seems that this one is not working when adding and removing columns. Memory is still a consideration, so it would be nice to put subitem info on the column header, but column headers can be removed so we must take it into account when recalculating indices.
Suggestion - can we have an additional array, of the same length as the number of subitems, that will maintain `DisplayIndices` for subitems. Instead of storing display indices on each subitem, since they are the same .

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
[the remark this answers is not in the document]
Epica3055: > So AddRange will break your fix? Please add Update... to the AddRange path.

I think that probably won't be necessary because `AddRange` will call `Add`, `Add` will call `InsertColumn`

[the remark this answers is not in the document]
Tanya-Solyanik: If AddRange eventually calls InsertColumn, then we are good here

Epica3055:
@Tanya-Solyanik  Yes, I have tested the scenarios that I can think of.

Tanya-Solyanik:
@Epica3055  - does your fix work correctly when deleting and adding columns?

Epica3055:
> @Epica3055 - does your fix work correctly when deleting and adding columns?

Yes, I tested it, it worked correctly

Tanya-Solyanik (approved):

PR #9632, opened 2023-08-01
Use AdviseHolder in ActiveXImpl

JeremyKuhne:
This change is the first of a series that is auditing and documenting our ActiveX control usages.

Our handling of Advise/Unadvise wasn't properly ref counting. MFC just uses the Windows provided IOleAdviseHolder, this change does the same.

This change also adds a number of inheritdoc tags to pick up the documentation generated by CSWin32, Note that the interface names don't always match as the documentation is only on the base interface.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9632)


src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
```diff
@@ -2096,20 +2081,16 @@ internal HRESULT UIDeactivate()
...
-        ///  Implements IOleObject::Unadvise
-        /// </summary>
+        /// <inheritdoc cref="IOleObject.Unadvise(uint)"/>
         internal HRESULT Unadvise(uint dwConnection)
         {
-            if (dwConnection > _adviseList.Count || _adviseList[(int)dwConnection - 1] is null)
+            if (_adviseHolder is null)
             {
-                return HRESULT.OLE_E_NOCONNECTION;
+                return HRESULT.E_FAIL;
```
Tanya-Solyanik, line 2089: Why isn't OLE_E_NOCONNECTION appropriate here? Advise had not been called, so connection had not been established.. https://learn.microsoft.com/en-us/windows/win32/api/oleidl/nf-oleidl-ioleobject-unadvise

JeremyKuhne: Because this is exactly what MFC does.

Tanya-Solyanik (approved):

lonitra (approved):
Nice!

PR #9641, opened 2023-08-02
Implement: Mark some properties and methods as `readonly`

### LINKED ISSUE #9618: Mark some properties and methods as `readonly`
### Background and motivation

Some properties and methods should be marked `readonly`. This way they can be used on `readonly` references without causing defensive copies.

This is something I ran into recently with `Padding`, so I went and found (hopefully) all of the types which could be improved by marking them / some of their members as `readonly`. I have an analyser which detected a defensive copy for my use of `Padding`, and I have currently worked around it by using `Unsafe.AsRef` (obviously I could have also accepted the defensive copy) - but I would like to remove this workaround as it should be unnecessary.

Note that `Point`, `PointF`, etc. have already had this done to them in https://github.com/dotnet/corefx/pull/40407.

I would like to make a PR for this when approved.

### API Proposal

```diff
namespace System.Drawing
{
	//`: IEquatable<CharacterRange>` is currently not in ref-only dll, but is in source dll:
-	public struct CharacterRange
+	public struct CharacterRange : IEquatable<CharacterRange>
	{
-		public int First { get; set; }
+		public int First { readonly get; set; }
-		public int Length { get; set; }
+		public int Length { readonly get; set; }
-		public override bool Equals(object? obj);
+		public override readonly bool Equals(object? obj);
		//`bool Equals(CharacterRange other)` is currently not in ref-only dll, but is in source dll:
+		public readonly bool Equals(CharacterRange other);
-		public override int GetHashCode();
+		public override readonly int GetHashCode();
		//...
	}
}

namespace System.Windows.Forms
{
-	public struct BindingMemberInfo
+	public readonly struct BindingMemberInfo
	{
		public string BindingPath { get; }
		public string BindingField { get; }
		public string BindingMember { get; }
		public override bool Equals(object? otherObject);
		public bool Equals(BindingMemberInfo other);
		public override int GetHashCode();
		//...
	}
	//or we could make each member readonly instead? (applicable members listed above if this is what we do)

-	public struct ImeModeConversion
+	public readonly struct ImeModeConversion
	{
		//...
	}

	public struct LinkArea
	{
-		public int Start { get; set; }
+		public int Start { readonly get; set; }
-		public int Length { get; set; }
+		public int Length { readonly get; set; }
-		public bool IsEmpty { get; }
+		public bool IsEmpty { readonly get; }
-		public override bool Equals(object? o);
+		public override readonly bool Equals(object? o);
-		public bool Equals(LinkArea other);
+		public readonly bool Equals(LinkArea other);
-		public override string ToString();
+		public override readonly string ToString();
-		public override int GetHashCode();
+		public override readonly int GetHashCode();
		//...
	}

	public struct Message
	{
-		public int Msg { get; set; }
+		public int Msg { readonly get; set; }
-		public IntPtr WParam { get; set; }
+		public IntPtr WParam { readonly get; set; }
-		public IntPtr LParam { get; set; }
+		public IntPtr LParam { readonly get; set; }
-		public IntPtr Result { get; set; }
+		public IntPtr Result { readonly get; set; }
-		public object? GetLParam(Type cls);
+		public readonly object? GetLParam(Type cls);
-		public override bool Equals(object? o);
+		public override readonly bool Equals(object? o);
-		public bool Equals(Message other);
+		public readonly bool Equals(Message other);
-		public override int GetHashCode();
+		public override readonly int GetHashCode();
-		public override string ToString();
+		public override readonly string ToString();
		//...
	}

	public struct Padding
	{
-		public int All { get; set; }
+		public int All { readonly get; set; }
-		public int Bottom { get; set; }
+		public int Bottom { readonly get; set; }
-		public int Left { get; set; }
+		public int Left { readonly get; set; }
-		public int Right { get; set; }
+		public int Right { readonly get; set; }
-		public int Top { get; set; }
+		public int Top { readonly get; set; }
-		public int Horizontal { get; }
+		public int Horizontal { readonly get; }
-		public int Vertical { get; }
+		public int Vertical { readonly get; }
-		public Size Size { get; }
+		public Size Size { readonly get; }
-		public override bool Equals(object? other);
+		public override readonly bool Equals(object? other);
-		public bool Equals(Padding other);
+		public readonly bool Equals(Padding other);
-		public override int GetHashCode();
+		public override readonly int GetHashCode();
-		public override string ToString();
+		public override readonly string ToString();
		//...
	}

	public struct TableLayoutPanelCellPosition
	{
-		public override bool Equals(object? other);
+		public override readonly bool Equals(object? other);
-		public bool Equals(TableLayoutPanelCellPosition other);
+		public readonly bool Equals(TableLayoutPanelCellPosition other);
-		public override string ToString();
+		public override readonly string ToString();
-		public override int GetHashCode();
+		public override readonly int GetHashCode();
		//...
	}
}
```

Pre-existing attributes omitted for clarity.

Edit: Removed `TableLayoutPanelCellPosition.Row`, `TableLayoutPanelCellPosition.Column`, `Message.HWnd`, and all of the properties on `TextMetrics`, as these were already marked as `readonly` due to being auto properties.

Edit 2: Add implementation of `IEquatable<CharacterRange>` to `CharacterRange` + its method, since it's declared in source, but not in the ref project.

Edit 3: Add missing `Padding.GetHashCode()` and `Padding.ToString()` members.

Edit 4 (question): Should we fix the `NotSupported.cs`'s definition of `CharacterRange` to have the correct fields. It doesn't really matter, but would better match what C# would generate for a ref assembly.

### API Usage

The applicable APIs can now be used anywhere that these types are declared as readonly without causing a defensive copy, thus improving performance.

e.g. (a simplified version of my scenario)
```csharp
struct SomeLargeStruct
{
	public Padding Padding;
	//...
}

int SomeMethod(in SomeLargeStruct referenceToABunchOfFields, int index)
{
	//some condition
	if (condition) return referenceToABunchOfFields.Padding.Top;
}
```

### Alternative Designs

For `BindingMemberInfo`, we could do:
```diff
namespace System.Windows.Forms
{
	public struct BindingMemberInfo
	{
-		public string BindingPath { get; }
+		public readonly string BindingPath { get; }
-		public string BindingField { get; }
+		public readonly string BindingField { get; }
-		public string BindingMember { get; }
+		public readonly string BindingMember { get; }
-		public override bool Equals(object? otherObject);
+		public override readonly bool Equals(object? otherObject);
-		public bool Equals(BindingMemberInfo other);
+		public bool readonly Equals(BindingMemberInfo other);
-		public override int GetHashCode();
+		public override readonly int GetHashCode();
	}
}
```
instead of
```diff
namespace System.Windows.Forms
{
-	public struct BindingMemberInfo
+	public readonly struct BindingMemberInfo
	{
		public string BindingPath { get; }
		public string BindingField { get; }
		public string BindingMember { get; }
		public override bool Equals(object? otherObject);
		public bool Equals(BindingMemberInfo other);
		public override int GetHashCode();
	}
}
```

### Risks

This should be pretty uncontroversial.

Maybe I have missed an API somewhere, I'm pretty sure I got all the public structs, but I may have missed a nested one or something like that.

I would hope not.

hamarb123:
Fixes #9618

## Proposed changes

<details>
<summary>
API proposal
</summary>

```diff
namespace System.Drawing
{
	public struct CharacterRange
	{
-		public int First { get; set; }
+		public int First { readonly get; set; }
-		public int Length { get; set; }
+		public int Length { readonly get; set; }
-		public override bool Equals(object? obj);
+		public override readonly bool Equals(object? obj);
-		public bool Equals(CharacterRange other); //net7.0+
+		public readonly bool Equals(CharacterRange other); //net7.0+
-		public override int GetHashCode();
+		public override readonly int GetHashCode();
	}
}

namespace System.Windows.Forms
{
-	public struct BindingMemberInfo
+	public readonly struct BindingMemberInfo
	{
	}

-	public struct ImeModeConversion
+	public readonly struct ImeModeConversion
	{
	}

	public struct LinkArea
	{
-		public int Start { get; set; }
+		public int Start { readonly get; set; }
-		public int Length { get; set; }
+		public int Length { readonly get; set; }
-		public bool IsEmpty { get; }
+		public bool IsEmpty { readonly get; }
-		public override bool Equals(object? o);
+		public override readonly bool Equals(object? o);
-		public bool Equals(LinkArea other);
+		public readonly bool Equals(LinkArea other);
-		public override string ToString();
+		public override readonly string ToString();
-		public override int GetHashCode();
+		public override readonly int GetHashCode();
	}

	public struct Message
	{
-		public int Msg { get; set; }
+		public int Msg { readonly get; set; }
-		public IntPtr WParam { get; set; }
+		public IntPtr WParam { readonly get; set; }
-		public IntPtr LParam { get; set; }
+		public IntPtr LParam { readonly get; set; }
-		public IntPtr Result { get; set; }
+		public IntPtr Result { readonly get; set; }
-		public object? GetLParam(Type cls);
+		public readonly object? GetLParam(Type cls);
-		public override bool Equals(object? o);
+		public override readonly bool Equals(object? o);
-		public bool Equals(Message other);
+		public readonly bool Equals(Message other);
-		public override int GetHashCode();
+		public override readonly int GetHashCode();
-		public override string ToString();
+		public override readonly string ToString();
	}

	public struct Padding
	{
-		public int All { get; set; }
+		public int All { readonly get; set; }
-		public int Bottom { get; set; }
+		public int Bottom { readonly get; set; }
-		public int Left { get; set; }
+		public int Left { readonly get; set; }
-		public int Right { get; set; }
+		public int Right { readonly get; set; }
-		public int Top { get; set; }
+		public int Top { readonly get; set; }
-		public int Horizontal { get; }
+		public int Horizontal { readonly get; }
-		public int Vertical { get; }
+		public int Vertical { readonly get; }
-		public Size Size { get; }
+		public Size Size { readonly get; }
-		public override bool Equals(object? other);
+		public override readonly bool Equals(object? other);
-		public bool Equals(Padding other);
+		public readonly bool Equals(Padding other);
-		public override int GetHashCode();
+		public override readonly int GetHashCode();
-		public override string ToString();
+		public override readonly string ToString();
	}

	public struct TableLayoutPanelCellPosition
	{
-		public override bool Equals(object? other);
+		public override readonly bool Equals(object? other);
-		public bool Equals(TableLayoutPanelCellPosition other);
+		public readonly bool Equals(TableLayoutPanelCellPosition other);
-		public override string ToString();
+		public override readonly string ToString();
-		public override int GetHashCode();
+		public override readonly int GetHashCode();
	}
}
```

</details>

## Customer Impact

- Allows avoiding defensive copies

## Regression? 

- No

## Risk

- No

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9641)

### DIFF
--- a/src/System.Drawing.Common/src/NotSupported.cs
+++ b/src/System.Drawing.Common/src/NotSupported.cs
@@ -230,7 +230,8 @@ namespace System.Drawing
     }
     public partial struct CharacterRange : System.IEquatable<CharacterRange>
     {
-        private int _dummyPrimitive;
+        private int _first;
+        private int _length;
         public CharacterRange(int First, int Length) { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  }
         public int First { readonly get { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  } set { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  } }
         public int Length { readonly get { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  } set { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  } }
--- a/src/System.Drawing.Common/src/NotSupported.cs
+++ b/src/System.Drawing.Common/src/NotSupported.cs
@@ -230,8 +230,7 @@ namespace System.Drawing
     }
     public partial struct CharacterRange
     {
-        private int _first;
-        private int _length;
+        private int _dummyPrimitive;
         public CharacterRange(int First, int Length) { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  }
         public int First { readonly get { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  } set { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  } }
         public int Length { readonly get { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  } set { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  } }
--- a/src/System.Drawing.Common/src/NotSupported.cs
+++ b/src/System.Drawing.Common/src/NotSupported.cs
@@ -228,14 +228,15 @@ namespace System.Drawing
     {
         public static System.Drawing.BufferedGraphicsContext Current { get { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  } }
     }
-    public partial struct CharacterRange
+    public partial struct CharacterRange : System.IEquatable<CharacterRange>
     {
         private int _dummyPrimitive;
         public CharacterRange(int First, int Length) { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  }
-        public int First { get { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  } set { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  } }
-        public int Length { get { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  } set { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  } }
-        public override bool Equals([System.Diagnostics.CodeAnalysis.NotNullWhenAttribute(true)] object? obj) { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  }
-        public override int GetHashCode() { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  }
+        public int First { readonly get { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  } set { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  } }
+        public int Length { readonly get { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  } set { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  } }
+        public override readonly bool Equals([System.Diagnostics.CodeAnalysis.NotNullWhenAttribute(true)] object? obj) { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  }
+        public readonly bool Equals(CharacterRange other) { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported); }
+        public override readonly int GetHashCode() { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  }
         public static bool operator ==(System.Drawing.CharacterRange cr1, System.Drawing.CharacterRange cr2) { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  }
         public static bool operator !=(System.Drawing.CharacterRange cr1, System.Drawing.CharacterRange cr2) { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  }
     }
--- a/src/System.Drawing.Common/src/NotSupported.cs
+++ b/src/System.Drawing.Common/src/NotSupported.cs
@@ -228,7 +228,7 @@ namespace System.Drawing
     {
         public static System.Drawing.BufferedGraphicsContext Current { get { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  } }
     }
-    public partial struct CharacterRange : System.IEquatable<CharacterRange>
+    public partial struct CharacterRange
     {
         private int _first;
         private int _length;
@@ -236,7 +236,6 @@ namespace System.Drawing
         public int First { readonly get { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  } set { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  } }
         public int Length { readonly get { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  } set { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  } }
         public override readonly bool Equals([System.Diagnostics.CodeAnalysis.NotNullWhenAttribute(true)] object? obj) { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  }
-        public readonly bool Equals(CharacterRange other) { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported); }
         public override readonly int GetHashCode() { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  }
         public static bool operator ==(System.Drawing.CharacterRange cr1, System.Drawing.CharacterRange cr2) { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  }
         public static bool operator !=(System.Drawing.CharacterRange cr1, System.Drawing.CharacterRange cr2) { throw new System.PlatformNotSupportedException(System.SR.SystemDrawingCommon_PlatformNotSupported);  }
--- a/src/System.Drawing.Common/src/System/Drawing/Font.cs
+++ b/src/System.Drawing.Common/src/System/Drawing/Font.cs
@@ -3,6 +3,7 @@
 
 using System.ComponentModel;
 using System.Globalization;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Runtime.Serialization;
 using Gdip = System.Drawing.SafeNativeMethods.Gdip;
@@ -596,7 +597,7 @@ internal
             throw new ArgumentException(SR.Format(SR.GdiplusNotTrueTypeFont, logFont.AsString()));
         }
 
-        bool gdiVerticalFont = logFont.lfFaceName[0] == '@';
+        bool gdiVerticalFont = logFont._lfFaceName[0] == '@';
         return new Font(font, logFont.lfCharSet, gdiVerticalFont);
     }
 
--- a/src/System.Drawing.Common/src/System/Drawing/Font.cs
+++ b/src/System.Drawing.Common/src/System/Drawing/Font.cs
@@ -3,7 +3,6 @@
 
 using System.ComponentModel;
 using System.Globalization;
-using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Runtime.Serialization;
 using Gdip = System.Drawing.SafeNativeMethods.Gdip;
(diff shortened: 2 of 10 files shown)


elachlan:
@hamarb123 There are some build/test failures. It might be because of the build targets for .net framework. 
```
API compatibility errors between 'lib/netstandard2.0/System.Drawing.Common.dll' (left) and 'lib/net462/System.Drawing.Common.dll'
Type 'System.Drawing.CharacterRange' does not implement interface 'System.IEquatable<System.Drawing.CharacterRange>' on lib/net462/System.Drawing.Common.dll but it does on lib/netstandard2.0/System.Drawing.Common.dll
API breaking changes found. If those are intentional, the APICompat suppression file can be updated by rebuilding with '/p:ApiCompatGenerateSuppressionFile=true'
API compatibility errors between 'lib/netstandard2.0/System.Drawing.Common.dll' (left) and 'lib/net462/System.Drawing.Common.dll' (right):
```

hamarb123:
It seems like it got removed in whenever https://github.com/dotnet/winforms/blob/9cf5b5ac66b48fa77fe9af28e2957ff73f5bbadf/src/System.Drawing.Common/ref/System.Drawing.Common.netcoreapp.cs got removed. Would that mean that there were other APIs removed on accident?

hamarb123:
Nevermind, I must have tested it on `net6.0` last time.

hamarb123:
I ran my analyser on the projects and discovered some defensive copies on internal types. I have made them no longer cause a defensive copy.
There is 1 more in `ToolStripManager.ModalMenuFilter.cs` that I didn't fix (`IsChildOrSameWindow`), it's on the parameters. Let me know if I should fix it too - I didn't fix it, because there's not an obvious solution other than just using `Unsafe.AsRef`.

src/System.Drawing.Common/src/NotSupported.cs
```diff
@@ -230,12 +230,13 @@ public static partial class BufferedGraphicsManager
     }
     public partial struct CharacterRange
     {
-        private int _dummyPrimitive;
+        private int _first;
```
JeremyKuhne, line 233: This should stay `_dummyPrimitive`.

src/System.Windows.Forms/src/System/Windows/Forms/Control.Ime.cs
```diff
@@ -1311,12 +1311,17 @@ public static void SetOpenStatus(bool open, IntPtr handle)
...
 ///  ALPHANUMERIC or HIRAGANA and depending on its value the IME enables/disables the IME conversion window appropriately.
 /// </summary>
-public struct ImeModeConversion
+public readonly struct ImeModeConversion
 {
     private static Dictionary<ImeMode, ImeModeConversion>? imeModeConversionBits;
 
-    internal IME_CONVERSION_MODE setBits;
-    internal IME_CONVERSION_MODE clearBits;
+    internal readonly IME_CONVERSION_MODE setBits;
```
JeremyKuhne, line 1318: Better to change these to autoprops `internal IME_CONVERSION_MODE SetBits { get; init; }`

Then you can do the following:

``` C#
s_imeModeConversionBits.Add(
    ImeMode.Hiragana,
    new()
    {
        SetBits = IME_CONVERSION_MODE.IME_CMODE_FULLSHAPE | IME_CONVERSION_MODE.IME_CMODE_NATIVE,
        ClearBits = IME_CONVERSION_MODE.IME_CMODE_KATAKANA
    });
```

src/System.Windows.Forms/src/System/Windows/Forms/Control.Ime.cs
```diff
@@ -1311,12 +1311,17 @@ public static void SetOpenStatus(bool open, IntPtr handle)
 ///  Helper class that provides information about IME conversion mode.  Conversion mode refers to how IME interprets input like
 ///  ALPHANUMERIC or HIRAGANA and depending on its value the IME enables/disables the IME conversion window appropriately.
 /// </summary>
-public struct ImeModeConversion
+public readonly struct ImeModeConversion
 {
     private static Dictionary<ImeMode, ImeModeConversion>? imeModeConversionBits;
```
JeremyKuhne, line 1316: Since we're touching this, can you please update this to `s_imeModeConversionBits`

src/System.Drawing.Common/src/System/Drawing/Interop/LOGFONT.cs
```diff
@@ -29,18 +30,18 @@ unsafe struct LOGFONT
     public byte lfClipPrecision;
     public byte lfQuality;
     public byte lfPitchAndFamily;
-    private fixed char _lfFaceName[LF_FACESIZE];
+    internal fixed char _lfFaceName[LF_FACESIZE];
```
JeremyKuhne, line 33: This shouldn't be necessary.

JeremyKuhne (changes requested):
Added a few comments

src/System.Drawing.Common/src/System/Drawing/Font.cs
```diff
@@ -596,7 +596,7 @@ static unsafe Font FromLogFont(in LOGFONT logFont, IntPtr hdc)
             throw new ArgumentException(SR.Format(SR.GdiplusNotTrueTypeFont, logFont.AsString()));
         }
 
-        bool gdiVerticalFont = logFont.lfFaceName[0] == '@';
+        bool gdiVerticalFont = logFont._lfFaceName[0] == '@';
```
hamarb123, line 599: @JeremyKuhne `_lfFaceName` is made `internal` so it can be used here. `lfFaceName` causes a defensive copy as it's a `Span`, not a `ReadOnlySpan`, and therefore cannot be exposed as a readonly member - this means it would copy the whole struct just to check if one `char` is `@`.

src/System.Windows.Forms/src/System/Windows/Forms/Control.Ime.cs
```diff
@@ -1459,62 +1459,93 @@ internal static ImeMode[] InputLanguageTable
     {
         get
         {
-            if (imeModeConversionBits is null)
+            if (s_imeModeConversionBits is null)
```
hamarb123, line 1462: Also, is this code even multi-threading safe? It doesn't seem like it is to me. If it was built up in a local and then stored using a volatile write it would be multi-threading safe, I can implement that if you want.

src/System.Drawing.Common/src/System/Drawing/Interop/LOGFONT.cs
hamarb123: Replied to this [here](https://github.com/dotnet/winforms/pull/9641#discussion_r1283761197).

src/System.Drawing.Common/src/System/Drawing/Font.cs
JeremyKuhne: You're right, thought it would be smart enough to know it didn't have to. We might be exposed elsewhere with this pattern.

hamarb123: I fixed all of the violations my analyser picked up on except the one I mentioned in a comment. It's probably not written perfectly, so maybe it missed something, but I think it can find most things.

src/System.Windows.Forms/src/System/Windows/Forms/Control.Ime.cs
JeremyKuhne: Yeah, it's probably better to build this in a local. Using `=> s_imeModeConversionBits ??= new(7) {{...}};` should do the right thing I think.

src/System.Drawing.Common/src/System/Drawing/Font.cs
elachlan: Maybe we ping the Roslyn team and see if they can improve the IL around this pattern?

hamarb123: It's not something that can be solved currently, it has to make a defensive copy to be correct. @jaredpar and some other people including me were talking about a new language feature which could achieve something like this yesterday on the C# discord, but it gets into the realm of generics over readonly to solve this particular case (without just adding a ROS property also obviously).

hamarb123: The reason it's an invalid operation to not make a defensive copy is because if I marked `lfFaceName` as readonly, then you could do `logFont.lfFaceName[0] = '@';` on a `ref readonly`. The C# compiler doesn't have special knowledge of `Span`, nor should it (for all it knows, `this[int]` could do a bunch of mutation on its ref - this would not break the span's `readonly` contract as nothing in the span itself changes). This would be something which should be solved in a more general way imo, which is difficult without things like `[MatchedMutabilityRef]` (what we were discussing yesterday) and generics over mutability.

elachlan: Instead of exposing _lfFaceName would it make sense to add an internal `readonly bool IsGdiVerticalFont =>_lfFaceName[0] == '@';` to `LOGFONT`? or am I way off?

JeremyKuhne (approved):
Looks good!

Issue #966, opened 2019-05-10
Add Tests related to Code Dom Serializers


zsd4yr:
We have a testing gap for types in System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization

We would like to add tests in this area.

/cc @DustinCampbell

elachlan:
@dreddy-work This is a duplicate of #719

PR #9642, opened 2023-08-03
Fix bugs in ResourceCodeDomSerializer.SerializationResourceManager

halgab:
As discussed in #9457.

#8332 introduced bugs in `ResourceCodeDomSerializer.SerializationResourceManager`, in method `GetResourceSet`. When we fail to get a resource reader, the original code used to return a resource set for the invariant culture, and null otherwise. After #8332,,

- we're returning `null` the first time we fill the `ResourceTable` cache
- for the non-invariant culture case, we started filling the `ResourceTable` cache with a static dictionary,  which looks like a pretty big issue

These bugs were introduced in .Net 8 and we still have a chance to fix them before the release

@dreddy-work, @elachlan 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9642)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ResourceCodeDomSerializer.SerializationResourceManager.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ResourceCodeDomSerializer.SerializationResourceManager.cs
@@ -227,7 +227,7 @@ internal partial class ResourceCodeDomSerializer
             Dictionary<string, object> resourceSet = GetResourceSet(culture);
             if (resourceSet is not null && resourceSet.TryGetValue(name, out object parentValue))
             {
-                return !parentValue.Equals(value) || parentValue is null ? CompareValue.Different : CompareValue.Same;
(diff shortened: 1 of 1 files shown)


dreddy-work:
> * we started filling the `ResourceTable` cache with a static dictionary,  which looks like a pretty big issue

@halgab, i see original code before @elachlan's change was also doing this. Can you point me to the original code where it wasn't?

halgab:
Before #8332, `s_resourceSetSentinel` was not a dictionary or a hashtable but an object (see line 21). Then on line 505, we checked that the content of `ResourceTable` was actually a `Hashtable`, which proved to be false when we were returning `s_resourceSetSentinel`

Tanya-Solyanik:
@halgab - is it possible to add a unit test for this scenario?

halgab:
I agree that would be ideal, but right now code coverage of serializers is so scarce I wouldn't really know where to start...

dreddy-work:
We can track test debt under separate issue. @Tanya-Solyanik, you good with changes here?

elachlan:
Related: #719, #966

lonitra (approved):
LGTM

PR #9643, opened 2023-08-03
Enable nullability in ToolboxItem and derived classes

halgab:
## Proposed changes

- Enable nullability in `ToolboxItem` and derived classes

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9643)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Drawing/Design/ToolboxItem.cs
+++ b/src/System.Windows.Forms.Design/src/System/Drawing/Design/ToolboxItem.cs
@@ -136,13 +136,7 @@ public class ToolboxItem : ISerializable
     /// <summary>
     ///  The Component Type is ".Net Component" -- unless otherwise specified by a derived toolboxitem
     /// </summary>
-    public virtual string ComponentType
-    {
-        get
-        {
-            return SR.DotNET_ComponentType;
-        }
-    }
+    public virtual string ComponentType => SR.DotNET_ComponentType;
 
     /// <summary>
     ///  Description is a free-form, multiline capable text description that will be displayed in the tooltip
@@ -214,13 +208,7 @@ public class ToolboxItem : ISerializable
     ///  Determines if this toolbox item is locked.  Once locked, a toolbox item will
     ///  not accept any changes to its properties.
     /// </summary>
-    public virtual bool Locked
-    {
-        get
-        {
-            return _locked;
-        }
-    }
+    public virtual bool Locked => _locked;
 
     /// <summary>
     ///  The properties dictionary is a set of name/value pairs.  The keys are property
@@ -229,15 +217,7 @@ public class ToolboxItem : ISerializable
     ///  Values in the properties dictionary are validated through ValidateProperty
     ///  and default values are obtained from GetDefaultProperty.
     /// </summary>
-    public IDictionary Properties
-    {
-        get
-        {
-            _properties ??= new LockableDictionary(this, 8 /* # of properties we have */);
-
-            return _properties;
-        }
-    }
+    public IDictionary Properties => _properties ??= new LockableDictionary(this, 8 /* # of properties we have */);
 
     /// <summary>
     ///  Gets or sets the fully qualified name of the type this toolbox item will create.
@@ -260,10 +240,7 @@ public class ToolboxItem : ISerializable
     ///  overridden in a derived toolboxitem. This can be overridden to
     ///  return an empty string to suppress its display in the toolbox tooltip.
     /// </summary>
-    public virtual string Version
-    {
-        get => AssemblyName?.Version?.ToString() ?? string.Empty;
-    }
+    public virtual string Version => AssemblyName?.Version?.ToString() ?? string.Empty;
 
     /// <summary>
     ///  Occurs when components are created.
@@ -299,10 +276,7 @@ public class ToolboxItem : ISerializable
     ///  Creates objects from the type contained in this toolbox item.
     /// </summary>
     /// <returns></returns>
-    public IComponent[]? CreateComponents()
-    {
-        return CreateComponents(null);
-    }
+    public IComponent[]? CreateComponents() => CreateComponents(null);
 
     /// <summary>
     ///  Creates objects from the type contained in this toolbox item.  If designerHost is non-null
@@ -462,7 +436,7 @@ public class ToolboxItem : ISerializable
             return false;
         }
 
-        if (!(obj.GetType() == GetType()))
+        if (obj.GetType() != GetType())
         {
             return false;
         }
@@ -523,10 +497,7 @@ public class ToolboxItem : ISerializable
     ///  The designer host is used to access an implementation of ITypeResolutionService.
     ///  However, the loaded type is not added to the list of references in the designer host.
     /// </summary>
-    public Type? GetType(IDesignerHost? host)
-    {
-        return GetType(host, AssemblyName, TypeName, false);
-    }
+    public Type? GetType(IDesignerHost? host) => GetType(host, AssemblyName, TypeName, false);
 
     /// <summary>
     ///  This utility function can be used to load a type given a name.  AssemblyName and
@@ -537,17 +508,11 @@ public class ToolboxItem : ISerializable
     [UnconditionalSuppressMessage("SingleFile", "IL3002", Justification = "Single-file case is handled")]
     protected virtual Type? GetType(IDesignerHost? host, AssemblyName? assemblyName, string typeName, bool reference)
     {
-        ITypeResolutionService? ts = null;
         Type? type = null;
 
         ArgumentNullException.ThrowIfNull(typeName);
 
-        if (host is not null)
-        {
-            ts = host.GetService(typeof(ITypeResolutionService)) as ITypeResolutionService;
-        }
-
-        if (ts is not null)
+        if (host.TryGetService(out ITypeResolutionService? ts))
         {
             if (reference)
             {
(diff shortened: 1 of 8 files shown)


src/System.Windows.Forms.Design/src/System/Drawing/Design/ToolboxItem.cs
```diff
@@ -165,11 +157,12 @@ public string Description
     /// <summary>
     ///  Gets or sets the display name for this <see cref="ToolboxItem"/>.
     /// </summary>
+    [AllowNull]
     public string DisplayName
     {
         get
         {
-            return (string)Properties["DisplayName"];
+            return (string)Properties["DisplayName"]!;
```
halgab, line 165: "DisplayName" is explicitely handled in the getter of `LockableDictionary`. Same with "Filter", "IsTransient", "TypeName". That's why those cannot be null. See the method `FilterPropertyValue`

src/System.Windows.Forms.Design/src/System/Drawing/Design/ToolboxItem.cs
```diff
@@ -886,43 +840,13 @@ protected virtual object ValidatePropertyValue(string propertyName, object value
...
-                        {
-                            filter[filterCount++] = tfa;
-                        }
-                    }
-                }
-
-                value = filter;
-                break;
+                ICollection? col = (ICollection?)value;
+                return col?.OfType<ToolboxItemFilterAttribute>().ToArray() ?? Array.Empty<ToolboxItemFilterAttribute>();
```
halgab, line 849: `OfType` method includes a null check of the elements

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/OleDragDropHandler.CfCodeToolboxItem.cs
```diff
@@ -72,10 +72,9 @@ protected override void Deserialize(SerializationInfo info, StreamingContext con
             }
         }
 
-        protected override IComponent[]? CreateComponentsCore(IDesignerHost host, IDictionary? defaultValues)
+        protected override IComponent[]? CreateComponentsCore(IDesignerHost? host, IDictionary? defaultValues)
         {
-            IDesignerSerializationService? ds = host.GetService<IDesignerSerializationService>();
-            if (ds is null || _serializationData is null)
+            if (!host.TryGetService(out IDesignerSerializationService? ds) || _serializationData is null)
```
halgab, line 77: Note that this adds a null check on `host` that wasn't there. Looking at just the first commit makes that more explicit.

src/System.Windows.Forms.Primitives/src/System/ComponentModel/TypeDescriptorHelper.cs
```diff
@@ -14,6 +14,15 @@ public static bool TryGetAttribute
         return attribute is not null;
     }
 
+    public static bool TryGetAttribute
```
halgab, line 17: Tests fail if I don't add this new overload

src/System.Windows.Forms.Design/src/System/Drawing/Design/ToolboxItem.cs
```diff
@@ -680,25 +641,20 @@ public virtual void Initialize(Type type)
...
-                    {
-                        Company = company.Company;
-                    }
+                    Company = company.Company;
                 }
 
                 //set the description based off the description attribute of the given type.
-                DescriptionAttribute descattr = (DescriptionAttribute)TypeDescriptor.GetAttributes(type)[typeof(DescriptionAttribute)];
-                if (descattr is not null)
+                if (TypeDescriptorHelper.TryGetAttribute(type, out DescriptionAttribute? descattr))
```
halgab, line 650: It's because of this change

src/System.Windows.Forms.Primitives/src/System/ComponentModel/TypeDescriptorHelper.cs
halgab: This overload is then automatically used [here](https://github.com/dotnet/winforms/blob/51001036223cb4211810c0f6a2abfef8ce6fb8d9/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGrid.cs#L2423). The switch to the `TryGetAttribute` helper was done in #5931, and now I'm worried it is not correct without this new ovleroad being introduced.
@JeremyKuhne

halgab:
I'm not sure why this PR keeps having the 'Draft' label 3 weeks after I marked it as ready for review...
@dreddy-work, @lonitra for visibility (sorry for the ping, I'm just worried this erroneous label makes this PR fall off your radar)

dreddy-work:
> I'm not sure why this PR keeps having the 'Draft' label 3 weeks after I marked it as ready for review.

Thanks for bringing this to our notice. Bot action to remove `Draft `label somehow didn't work on this PR. I will keep an eye if this is a consistent issue and get it addressed by service team.

src/System.Windows.Forms.Design/src/System/Drawing/Design/ToolboxItem.cs
```diff
@@ -680,25 +641,20 @@ public virtual void Initialize(Type type)
             if (!type.Assembly.ReflectionOnly)
             {
                 object[] companyattrs = type.Assembly.GetCustomAttributes(typeof(AssemblyCompanyAttribute), true);
-                if (companyattrs is not null && companyattrs.Length > 0)
+                if (companyattrs is [AssemblyCompanyAttribute { Company: not null } company, ..])
```
lonitra, line 644: Does this include the check that `company.Company` isn't null?

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/OleDragDropHandler.CfCodeToolboxItem.cs
```diff
@@ -85,27 +84,24 @@ protected override void Deserialize(SerializationInfo info, StreamingContext con
...
 
             // Parent and locate each Control
             defaultValues ??= new Dictionary<string, object>();
-            Control? parentControl = defaultValues["Parent"] as Control;
-            if (parentControl is not null)
+            if (defaultValues["Parent"] is Control parentControl)
             {
-                ParentControlDesigner? parentControlDesigner = host.GetDesigner(parentControl) as ParentControlDesigner;
-                if (parentControlDesigner is not null)
+                if (host.GetDesigner(parentControl) is ParentControlDesigner parentControlDesigner)
```
lonitra, line 97: Can you combine these if statements to reduce the nesting here

src/System.Windows.Forms.Design/src/System/Drawing/Design/ToolboxItem.cs
```diff
@@ -58,11 +56,11 @@ public ToolboxItem(Type toolType) : this()
     ///  The assembly name for this toolbox item. The assembly name describes the assembly
     ///  information needed to load the toolbox item's type.
     /// </summary>
-    public AssemblyName AssemblyName
+    public AssemblyName? AssemblyName
     {
         get
         {
-            return (AssemblyName)Properties["AssemblyName"];
+            return (AssemblyName?)Properties["AssemblyName"];
```
lonitra, line 63: Could you use body expression for this and some of the other properties that you are touching below

src/System.Windows.Forms.Primitives/src/System/ComponentModel/TypeDescriptorHelper.cs
lonitra: 😮 how were the tests failing?

halgab: I opened #9980 to showcase the failure. It's the same as this branch but without the 2nd commit

lonitra (approved):
Thanks for providing the PR that showed the failure. The [docs](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.typedescriptor.getattributes?view=net-7.0#system-componentmodel-typedescriptor-getattributes(system-type)) seem to indicate that we need to be calling the `GetAttribute(Type)` method that the overload that you added calls when we do not have an instance, which in this case we did not so this LGTM!

halgab:
No problem. Merging this PR will then also [fix potential problems in `PropertyGrid`](https://github.com/dotnet/winforms/pull/9643#discussion_r1311559821). Should I put the second commit in a dedicated PR?

lonitra:
> Should I put the second commit in a dedicated PR?

I think leaving it in this PR is fine as we can clearly see the usage of the overload in this PR

elachlan:
Do we need to service the Overload to .NET 8?

halgab:
Knowing the bug has been around since .NET 7

dreddy-work:
> Do we need to service the Overload to .NET 8?

I would wait for a real customer report here.

lonitra:
I did some more investigation around here and I actually do not think that we will have problems with this in PropertyGrid (perhaps by sheer luck..? 😥 ). 
In the PR that was showcasing the failure, the test was expecting to get the attribute type `DescriptionAttribute` that was non-empty, but it was actually getting `DescriptionAttribute.Default`, which is empty. So `TryGetAttribute` will return `true` since technically the attribute is not null, but tests were failing due to that `DescriptionAttribute` being empty. I looked through the docs and the code for `GetAttribute` as well as the indexer on `AttributeCollection` and found the only time that it will ever return `null` is when the attribute does not exist in collection _**and there is no static field `Default` on whatever attribute type you are trying to get.**_ ([AttributeCollection.Indexer docs](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.attributecollection.item?view=net-7.0#system-componentmodel-attributecollection-item(system-type)) ,[AttributeCollection.Indexer code](https://source.dot.net/#System.ComponentModel.TypeConverter/System/ComponentModel/AttributeCollection.cs,127)) In `PropertyGrid`, we utilize `TryGetAttribute` to try to get `PropertyTabAttribute`. Note that it does not have a static field `Default`. So when we index into what we get from `GetAttribute` in `TryGetAttribute` we will never get an empty attribute like we did in the failure, we will get `null`, which will cause `TryGetAttribute` to return `false`.
That said, I think may be worth adding a summary comment to the `TypeDescriptorHelper` methods to emphasize that it is possible to get empty attribute type to encourage checking for that before proceeding if its necessary.

PR #9649, opened 2023-08-03
Enable nullability in DockEditor, ContentAlignmentEditor and SelectionPanelBase

halgab:
## Proposed changes

- Enable nullability in `SelectionPanelBase`
- Enable nullability in `DockEditor`
- Enable nullability in `ContentAlignmentEditor`

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9649)


src/System.Windows.Forms.Design/src/System/Drawing/Design/SelectionPanelBase.cs
```diff
@@ -12,12 +10,12 @@ namespace System.Drawing.Design;
...
     private bool _allowExit = true;
-    private RadioButton _checkedControl;
-    private IWindowsFormsEditorService _editorService;
+    private RadioButton? _checkedControl;
+    private IWindowsFormsEditorService? _editorService;
 
     protected RadioButton CheckedControl
     {
-        get => _checkedControl;
+        get => _checkedControl!;
```
dreddy-work, line 18: I do not have a real scenario here but forgiving null only propagate error somewhere else in this case and setter doesn't ensure it  isn't null. Can you check if you can remove backing field here? Is it possible to enforce null check in this?

halgab: The setter prevents us from removing the backing field. It ensures the value passed to it isn't null because it calls `FocusCheckedControl` which crashes with a NRE if `CheckedControl` is null. So my first version of the annotation of this property was `RadioButton?` with `[DisallowNull]`. But I realised that almost all the callsites then need a null forgiving operator, so I thought it was simpler to just do it the way it is here

dreddy-work (approved):

dreddy-work:
Please resolve conflict.

dreddy-work (approved):

PR #9654, opened 2023-08-03
Handle `ref` to `in` change

JeremyKuhne:
A number of Libarary APIs have had their parameters changed from `ref` to `in` (or `readonly ref`).

This change modifies code that we control and disables the "passing ref to in" warning for the interop assembly until we can get https://github.com/microsoft/CsWin32/issues/1014 resolved.

This change requires a more current Roslyn build and as such I've updated it. Note that VS Intellisense does not support this new feature yet. You can filter out intellisense errors by selecting "Build Only" in the error pane in the meantime (this will not remove red squiggles unfortunately).

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9654)

### DIFF
--- a/src/System.Drawing.Common/src/System/Drawing/DrawingCom.ComWrappers.cs
+++ b/src/System.Drawing.Common/src/System/Drawing/DrawingCom.ComWrappers.cs
@@ -10,10 +10,13 @@ using static Interop;
 namespace System.Drawing;
 
 /// <summary>
-/// The ComWrappers implementation for System.Drawing.Common's COM interop usages.
-///
-/// Supports IStream and IPicture COM interfaces.
+///  The ComWrappers implementation for System.Drawing.Common's COM interop usages.
 /// </summary>
+/// <remarks>
+///  <para>
+///   Supports IStream and IPicture COM interfaces.
+///  </para>
+/// </remarks>
 internal unsafe partial class DrawingCom : ComWrappers
 {
     private const int S_OK = (int)HRESULT.S_OK;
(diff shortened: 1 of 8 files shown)


JeremyKuhne:
cc: @tannergooding

src/System.Drawing.Common/src/System/Drawing/DrawingCom.ComWrappers.cs
```diff
@@ -52,7 +55,11 @@ protected override object CreateObject(IntPtr externalComObject, CreateObjectFla
         Debug.Assert(flags == CreateObjectFlags.UniqueInstance);
 
         Guid pictureIID = IPicture.IID;
+#if NET8_0_OR_GREATER
```
dreddy-work, line 58: Can you add a comment on why we are conditioning it for NET8.0. Also, why not in other files?

dreddy-work (approved):
LGTM.

KlausLoeffelmann (approved):

src/System.Drawing.Common/src/System/Drawing/DrawingCom.ComWrappers.cs
```diff
@@ -52,7 +55,11 @@ protected override object CreateObject(IntPtr externalComObject, CreateObjectFla
...
+        int hr = Marshal.QueryInterface(externalComObject, in pictureIID, out IntPtr comObject);
+#else
         int hr = Marshal.QueryInterface(externalComObject, ref pictureIID, out IntPtr comObject);
```
tannergooding, line 61: As a note, you could alternatively `#pragma warning disable CS....` here for the multi-target case.

`ref` still works, you just get a warning recommending you use `in` for clarity.

tannergooding (approved):

Issue #4908, opened 2021-05-14
Support designer capabilities at runtime


RussKie:
# Context

The Windows Forms SDK consist of two distinct parts – the **runtime** (the code that executes, open sourced in this repo) and the **designer** (consists of the "_general purpose_" and "_Visual Studio-specific_"  components, close sourced). In .NET Framework these two parts lived together, and that allowed customers to invoke and use VS-specific functionality. Skip forward to .NET Core/.NET and these two components are now split and evolve (for the most part) independently of each other.
 
The more and more customers are starting to migrate their apps to .NET the more we see the demand to the missing designer capabilities to be ported over to .NET (e.g. #1395, #2573, #4456, #4886, #4887, [AB#1314797](https://devdiv.visualstudio.com/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_workitems/edit/1314797)).

# In scope:

However we need to port the general purpose designer infrastructure – i.e. API that enable building "_a designer that the user can embed in their application_". For example, a user might have a business application that contains a report designer feature, which can use our "general designer" framework.

* The following designers are currently available (though may have restricted or incomplete functionality):
  - [x] System.ComponentModel.Design.ComponentDesigner
  - [x] System.Windows.Forms.Design.DocumentDesigner (#122)
  - [x] System.Windows.Forms.Design.FormDocumentDesigner (#681)
  - [x] System.Windows.Forms.Design.MaskedTextBoxDesigner (#9709)
  - [x] System.Windows.Forms.Design.ParentControlDesigner (#122)
  - [x] System.Windows.Forms.Design.ScrollableControlDesigner (#122)
  - [x] System.Windows.Forms.Design.TextBoxBaseDesigner
  - [x] System.Windows.Forms.Design.ToolStripDesigner
  - [x] System.Windows.Forms.Design.ToolStripDropDownDesigner
  - [x] System.Windows.Forms.Design.ToolStripItemDesigner
  - [x] System.Windows.Forms.Design.ToolStripMenuItemDesigner
 
* The following designers are missing and will be ported in scope of #4860:
  - [x] System.Windows.Forms.Design.ButtonBaseDesigner
  - [x] System.Windows.Forms.Design.ControlDesigner
  - [x] System.Windows.Forms.Design.DocumentDesigner - functionality restored
  - [x] System.Windows.Forms.Design.LabelDesigner
  - [x] System.Windows.Forms.Design.ParentControlDesigner - functionality restored
[...]

PR #9677, opened 2023-08-07
Port DataGridviewDesigner to runtime

Epica3055:
Related to #4908 

## Proposed changes

- Port DataGridviewDesigner to runtime

## Customer Impact

- DataGridviewDesigner can be designed in runtime

## Regression? 

-  No

## Risk

- Minimal

### Before

image

## Test methodology 

- Manually (added DataGridviewDesigner to DemoConsole test app)

## Test environment(s) 

- .Net 8.0.0-rc.1.23401.22

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9677)


src/System.Windows.Forms.Design/src/System.Windows.Forms.Design.csproj
```diff
@@ -84,6 +84,15 @@
     <EmbeddedResource Include="Resources\System\WinForms\Design\*.ico">
       <LogicalName>System.Windows.Forms.Design.%(FileName)</LogicalName>
     </EmbeddedResource>
+    <EmbeddedResource Include="Resources\System\WinForms\Design\DataGridViewColumnsDialog.selectedColumns.bmp">
```
Tanya-Solyanik, line 87: Please add a single EmbeddedResource element with a wildcard, like we do for icons in line 84.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BindingSourceDesigner.cs
```diff
@@ -0,0 +1,119 @@
...
+    /// </summary>
+    public override void Initialize(IComponent component)
+    {
+        base.Initialize(component);
+
+        IComponentChangeService componentChangeSvc = (IComponentChangeService)GetService(typeof(IComponentChangeService));
+        if (componentChangeSvc is not null)
+        {
+            componentChangeSvc.ComponentChanged += new ComponentChangedEventHandler(OnComponentChanged);
+            componentChangeSvc.ComponentRemoving += new ComponentEventHandler(OnComponentRemoving);
```
Tanya-Solyanik, line 50: Please remove unnecessary new for event handlers in other places in this PR as well
```suggestion
            componentChangeSvc.ComponentChanged += OnComponentChanged;
            componentChangeSvc.ComponentRemoving += OnComponentRemoving;
```

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BindingSourceDesigner.cs
```diff
@@ -0,0 +1,119 @@
...
+    /// </summary>
+    private void OnComponentChanged(object? sender, ComponentChangedEventArgs e)
+    {
+        if (_bindingUpdatedByUser &&
+            e.Component == this.Component &&
+            e.Member is not null && (e.Member.Name == "DataSource" || e.Member.Name == "DataMember"))
+        {
+            _bindingUpdatedByUser = false;
+
+            DataSourceProviderService dspSvc = (DataSourceProviderService)GetService(typeof(DataSourceProviderService));
```
Tanya-Solyanik, line 84: Please don't use abbreviations in variable names.
```suggestion
            var dataSourceProviderService = (DataSourceProviderService)GetService(typeof(DataSourceProviderService));
```

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BindingSourceDesigner.cs
```diff
@@ -0,0 +1,119 @@
...
+            dspSvc?.NotifyDataSourceComponentAdded(Component);
+        }
+    }
+
+    private void OnComponentRemoving(object? sender, ComponentEventArgs e)
+    {
+        BindingSource? b = this.Component as BindingSource;
```
Tanya-Solyanik, line 91: Please rename b->bindingSource.

Comment about not using abbreviations applied in multiple places in this PR. But feel free to do this clean up in the next PR or in a dedicated commit.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewAddColumnDialog.cs
```diff
@@ -0,0 +1,1012 @@
...
+
+using System.ComponentModel.Design.Serialization;
+using System.ComponentModel.Design;
+using System.ComponentModel;
+using System.Globalization;
+using System.Drawing;
+
+namespace System.Windows.Forms.Design;
+
+internal class DataGridViewAddColumnDialog :Form
```
Tanya-Solyanik, line 12:
```suggestion
internal class DataGridViewAddColumnDialog : Form
```

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewAddColumnDialog.cs
```diff
@@ -0,0 +1,1012 @@
...
+            _insertAtPosition++;
+        }
+
+        // if we persist changes directly to the designer then:
+        // 1. set the column property values
+        // 2. Add the new column
+        // 3. site the column
+
+        // 1. set the property values in the column
+        //
```
Tanya-Solyanik, line 117: Please remove the empty comment line

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewAddColumnDialog.cs
```diff
@@ -0,0 +1,1012 @@
...
+
+        while (!ValidName(columnName,
+                            _dataGridViewColumns!,
+                            container,
+                            null /*nameCreationService*/,
+                            _liveDataGridView!.Columns,
+                            !_persistChangesToDesigner))
+        {
+            colId++;
+            // columnName = (SR.DataGridView_ColumnName, colId.ToString());
```
Tanya-Solyanik, line 183: If this line is not needed, remove it

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewAddColumnDialog.cs
```diff
@@ -0,0 +1,1012 @@
...
+        else
+        {
+            _dataBoundColumnRadioButton!.Enabled = false;
+            _unboundColumnRadioButton!.Checked = true;
+            _nameTextBox!.Focus();
+            _headerTextBox!.Text = _nameTextBox.Text = AssignName();
+        }
+    }
+
+    //
```
Tanya-Solyanik, line 217: Use `<devdoc>`  tag for such comments

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewAddColumnDialog.cs
```diff
@@ -0,0 +1,1012 @@
...
+        // AddColumn takes all the information from this dialog, makes a DataGridViewColumn out of it
+        // and inserts it at the right index in the data grid view column collection
+        AddColumn();
+    }
+
+    private void cancelButton_Click(object? sender, EventArgs e)
+    {
+        Close();
+        /*
+        if (this.cancelButtonIsCloseButton)
```
Tanya-Solyanik, line 784: Please remove the commented out code.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewAddColumnDialog.cs
```diff
@@ -0,0 +1,1012 @@
...
+                    if (ValidName(_nameTextBox!.Text,
+                                  _dataGridViewColumns!,
+                                  container,
+                                  nameCreationService,
+                                  _liveDataGridView!.Columns,
+                                  !_persistChangesToDesigner,
+                                  out errorString))
+                    {
+                        this.AddColumn();
+                        this.Close();
```
Tanya-Solyanik, line 821: Please remove the `this` modifier though out this PR. This could be done in  a separate commit or in a follow up PR
```suggestion
                        AddColumn();
                        Close();
```

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewAddColumnDialog.cs
```diff
@@ -0,0 +1,1012 @@
...
+
+        return -1;
+    }
+
+    public static bool ValidName(string name,
+                                 DataGridViewColumnCollection columns,
+                                 IContainer? container,
+                                 INameCreationService? nameCreationService,
+                                 DataGridViewColumnCollection liveColumns,
+                                 bool allowDuplicateNameInLiveColumnCollection)
```
Tanya-Solyanik, line 892:
```suggestion
    public static bool ValidName(
        string name,
        DataGridViewColumnCollection columns,
        IContainer? container,
        INameCreationService? nameCreationService,
        DataGridViewColumnCollection liveColumns,
        bool allowDuplicateNameInLiveColumnCollection)
```

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewAddColumnDialog.cs
```diff
@@ -0,0 +1,1012 @@
...
+            {
+                return false;
+            }
+        }
+
+        return true;
+    }
+
+    /// <devdoc>
+    ///     A column name is valid if it does not cause any name conflicts in the DataGridViewColumnCollection
```
Tanya-Solyanik, line 919: Please use 2 spaces to indent content in xml-doc comments. This applies to the whole PR.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewAddColumnDialog.cs
```diff
@@ -0,0 +1,1012 @@
...
+                errorString = string.Format(SR.CodeDomDesignerLoaderInvalidIdentifier, name);
+                return false;
+            }
+        }
+
+        errorString = string.Empty;
+        return true;
+    }
+
+    private class ListBoxItem
```
Tanya-Solyanik, line 958: Please move this class to a dedicated file.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewAddColumnDialog.cs
```diff
@@ -0,0 +1,1012 @@
...
+    {
+        private Type propertyType;
```
Tanya-Solyanik, line 960: Use the underscore in filed names per convention.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.cs
```diff
@@ -2012,6 +2013,7 @@ public string DataMember
         remove => Events.RemoveHandler(s_dataMemberChangedEvent, value);
     }
 
+    [Editor($"System.Windows.Forms.Design.DataSourceListEditor, {AssemblyRef.SystemDesign}", typeof(UITypeEditor))]
```
Tanya-Solyanik, line 2016: Had you tested this change in the OOP designer?

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/ClientUtils.cs
```diff
@@ -47,4 +47,14 @@ public static int GetWordBoundaryStart(string text, int endIndex)
 
         return index + 1;
     }
+
```
Tanya-Solyanik, line 50: This one is already defined  - https://github.com/dotnet/winforms/blob/6268411a43735e67119df9afe2c45b68a29bd4c9/src/System.Windows.Forms.Primitives/src/System/ExceptionExtensions.cs#L12

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.cs
```diff
@@ -291,14 +294,14 @@ public static Size LogicalToDeviceUnits(Size logicalSize, int deviceDpi = 0)
     /// <param name="logicalImage">The image to scale from logical units to device units</param>
     /// <param name="targetImageSize">The size to scale image to</param>
     [return: NotNullIfNotNull("logicalImage")]
-    public static Bitmap? CreateResizedBitmap(Bitmap? logicalImage, Size targetImageSize)
+    public static Bitmap? CreateResizedBitmap(Bitmap? logicalImage, Size targetImageSize, bool doNotOffset = false)
```
Tanya-Solyanik, line 297: Double negation is more difficult than no negation. I.e. `bool offset=true`.

But more importantly, I don't see callers that set this parameter to a non-default value.

Tanya-Solyanik: I think that changes outside the `.Design` assembly are not required.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.cs
Epica3055: To be honest I don't fully understand the code here.

```  c#
// Licensed to the .NET Foundation under one or more agreements.

using System.Collections;

namespace System.ComponentModel
{
    [MergableProperty(false)]
    [Editor("System.Windows.Forms.Design.DataSourceListEditor, System.Design, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a",
            "System.Drawing.Design.UITypeEditor, System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a")]
    [TypeConverterAttribute("System.Windows.Forms.Design.DataSourceConverter, System.Design, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a")]
    public interface IListSource
    {
        bool ContainsListCollection { get; }

        IList GetList();
    }
}
```
But if don't add this `[Editor($"System.Windows.Forms.Design.DataSourceListEditor, {AssemblyRef.SystemDesign}", typeof(UITypeEditor))]` ,the editor will not show.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewAddColumnDialog.cs
Epica3055: This is a small nested class, there are quite a few of this class in this PR. 
Do really need to extract every one of them into a separate file?
Or did you mean: just extract every one of them out of the class but it still remain in the file? like this:
image

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.cs
Epica3055: I am sorry. I don't fully understand.

> I think that changes outside the .Design assembly are not required.
We should not change this file?

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewAddColumnDialog.cs
Tanya-Solyanik: If these classes were nested before, they should stay nested. I meant extracting each small class into a small file.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.cs
Tanya-Solyanik: I see, we are probably using TypeDescriptor to shadow this property. Let DataSource property stay readonly in the sample app for now and open an issue to add support for editor later. To find out how it works, please debug the InProc designer.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BindingSourceDesigner.cs
```diff
@@ -0,0 +1,119 @@
...
+    ///  Dispose this designer.
+    /// </summary>
+    protected override void Dispose(bool disposing)
+    {
+        if (disposing)
+        {
+            IComponentChangeService componentChangeService = (IComponentChangeService)GetService(typeof(IComponentChangeService));
+            if (componentChangeService is not null)
+            {
+                componentChangeService.ComponentChanged -= (OnComponentChanged);
```
Tanya-Solyanik, line 64: Please remove the extra parens.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewAddColumnDialog.cs
Epica3055: Yes , there are nested before.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.cs
Epica3055: > Let DataSource property stay readonly in the sample app for now

you mean we should halt this PR? open a new issue to Port DataGridviewDesigner to runtime right?

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.cs
Epica3055: image

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BindingSourceDesigner.cs
```diff
@@ -0,0 +1,119 @@
...
+        {
+            IComponentChangeService ComponentChangeService = (IComponentChangeService)GetService(typeof(IComponentChangeService));
+            string previousDataMember = bindingSource.DataMember;
+
+            PropertyDescriptorCollection props = TypeDescriptor.GetProperties(bindingSource);
+            PropertyDescriptor? propertyDescriptor = props?["DataMember"];
+
+            if (ComponentChangeService is not null)
+            {
+                if (propertyDescriptor is not null)
```
SimonZhao888, line 102: Conditions could be combined into a single one to improve readability of this code.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BindingSourceDesigner.cs
```diff
@@ -0,0 +1,113 @@
...
+    ///  that the BindingSource is now bound to, so it can generate any table adapters or fill
+    ///  statements necessary to properly set up that data source.
+    ///
+    ///  This serves the scenario of an advanced user configuring a BindingSource manually. In
+    ///  normal user scenarios, BindingSources get auto-generated as a result of various drag/drop
+    ///  operations, and in those scenarios the above notification is sent automatically.
+    ///
+    ///  VSWhidbey#256272
+    /// </summary>
+    public bool BindingUpdatedByUser
```
lonitra, line 31: Can this just be made into a property?

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BindingSourceDesigner.cs
```diff
@@ -0,0 +1,113 @@
...
+
+    /// <summary>
+    ///  Dispose this designer.
+    /// </summary>
+    protected override void Dispose(bool disposing)
+    {
+        if (disposing)
+        {
+            IComponentChangeService componentChangeService = (IComponentChangeService)GetService(typeof(IComponentChangeService));
+            if (componentChangeService is not null)
```
lonitra, line 62: Let's combine these if statements
```suggestion
        if (disposing && GetService<IComponentChangeService>() is { } componentChangeService)
        {
```

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ChangeToolStripParentVerb.cs
```diff
@@ -28,7 +28,7 @@ internal ChangeToolStripParentVerb(string text, ToolStripDesigner designer)
         _designer = designer;
         _provider = designer.Component.Site;
         _host = (IDesignerHost)_provider.GetService(typeof(IDesignerHost));
```
lonitra, line 30: I believe this API is available to make this simplification:
```suggestion
        _host = _provider.GetService<IDesignerHost>();
```
The same change can be applied below.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/CollectionEditVerbManager.cs
```diff
@@ -55,9 +55,9 @@ private IComponentChangeService ChangeService
     {
         get
         {
-            _componentChangeSvc ??= (IComponentChangeService)((IServiceProvider)this).GetService(typeof(IComponentChangeService));
+            _componentChangeService ??= (IComponentChangeService)((IServiceProvider)this).GetService(typeof(IComponentChangeService));
```
lonitra, line 58:
```suggestion
            _componentChangeService ??= ((IServiceProvider)this).GetService<IComponentChangeService>();
```

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewAddColumnDialog.cs
```diff
@@ -0,0 +1,1005 @@
...
+
+        while (!ValidName(columnName,
+                            _dataGridViewColumns!,
+                            container,
+                            null /*nameCreationService*/,
+                            _liveDataGridView!.Columns,
+                            !_persistChangesToDesigner))
+        {
+            colId++;
+            columnName = "Column" + colId.ToString(CultureInfo.InvariantCulture);
```
lonitra, line 183: Let's use string interpolation:
```suggestion
            columnName = $"Column{colId.ToString()}";
```

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewAddColumnDialog.cs
```diff
@@ -0,0 +1,1005 @@
...
+        return columnName;
+    }
+
+    /// <summary>
+    ///  Clean up any resources being used.
+    /// </summary>
+    protected override void Dispose(bool disposing)
+    {
+        base.Dispose(disposing);
+    }
```
lonitra, line 195: nit:
```suggestion
    protected override void Dispose(bool disposing) => base.Dispose(disposing);
```

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewAddColumnDialog.cs
```diff
@@ -0,0 +1,1005 @@
...
+        {
+            Type? designerType = null;
+            if (typeResolutionService is not null)
+            {
+                designerType = typeResolutionService.GetType(designerAttribute.DesignerTypeName);
+            }
+            else
+            {
+                designerType = Type.GetType(designerAttribute.DesignerTypeName);
+            }
```
lonitra, line 251: This can be simplified.
```suggestion
            Type? designerType = typeResolutionService is null 
            ? Type.GetType(designerAttribute.DesignerTypeName) 
            : typeResolutionService.GetType(designerAttribute.DesignerTypeName);
```

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewColumnCollectionDialog.cs
```diff
@@ -0,0 +1,1505 @@
...
+        finally
+        {
+            _columnCollectionChanging = false;
+        }
+    }
+
+    [SuppressMessage("Microsoft.Performance", "CA1808:AvoidCallsThatBoxValueTypes")]
+    private void CommitChanges()
+    {
+        if (_formIsDirty)
```
lonitra, line 190: Please invert this if statement.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewColumnCollectionDialog.cs
```diff
@@ -0,0 +1,1505 @@
...
+        PropertyDescriptor? propertyDescriptor = TypeDescriptor.GetProperties(col)["UserAddedColumn"];
+        object? val = propertyDescriptor?.GetValue(col);
+        if (propertyDescriptor is not null && val is not null)
+        {
+            return (bool)val;
+        }
+        else
+        {
+            return false;
+        }
```
lonitra, line 655:
```suggestion
        return (bool)val ?? false;
```

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewColumnCollectionDialog.cs
```diff
@@ -0,0 +1,1505 @@
...
+    {
+        int insertIndex;
+        if (_selectedColumns!.SelectedIndex == -1)
+        {
+            insertIndex = _selectedColumns.Items.Count;
+        }
+        else
+        {
+            insertIndex = _selectedColumns.SelectedIndex + 1;
+        }
```
lonitra, line 857:
```suggestion
        int insertIndex = _selectedColumns!.SelectedIndex == -1 
        ? _selectedColumns.Items.Count
        : _selectedColumns.SelectedIndex + 1;
```

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewColumnCollectionDialog.cs
```diff
@@ -0,0 +1,1505 @@
...
+
+        if (_selectedColumns.Items.Count == 0)
+        {
+            _propertyGridLabel!.Text = string.Format(SR.DataGridViewProperties);
+        }
+    }
+
+    private void propertyGrid1_PropertyValueChanged(object? sender, PropertyValueChangedEventArgs e)
+    {
+        if (!_columnCollectionChanging)
```
lonitra, line 906: Please invert this if statement

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewColumnCollectionDialog.cs
```diff
@@ -0,0 +1,1505 @@
...
+    ///  internal because the DataGridViewColumnDataPropertyNameEditor needs to get at the ListBoxItem
+    ///  IComponent because some editors for some dataGridViewColumn properties - DataGridViewComboBox::DataSource editor -
+    ///  need the site
+    /// </summary>
+    internal class ListBoxItem : ICustomTypeDescriptor, IComponent
+    {
+        private DataGridViewColumn _column;
+        private DataGridViewColumnCollectionDialog _owner;
+        private ComponentDesigner? _compDesigner;
+        private Image? _toolboxBitmap;
```
lonitra, line 1222: I think we can just convert these into properties

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewColumnCollectionDialog.cs
```diff
@@ -0,0 +1,1505 @@
...
+        public override string ToString()
+        {
+            return _column.HeaderText;
+        }
+
+        // ICustomTypeDescriptor implementation
+        AttributeCollection ICustomTypeDescriptor.GetAttributes()
+        {
+            return TypeDescriptor.GetAttributes(_column);
+        }
```
lonitra, line 1295:
```suggestion
        AttributeCollection ICustomTypeDescriptor.GetAttributes() => TypeDescriptor.GetAttributes(_column);
```
Ditto below as well.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewColumnCollectionDialog.cs
```diff
@@ -0,0 +1,1505 @@
...
+
+        ISite? IComponent.Site
+        {
+            get
+            {
+                return _owner._liveDataGridView?.Site;
+            }
+            set
+            {
+            }
```
lonitra, line 1403:
```suggestion
            get => _owner._liveDataGridView?.Site;
            set;
```

lonitra (commented):
Mostly styling/simplification comments

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewColumnCollectionDialog.cs
Epica3055: How about this case: propertyDescriptor is null , val is not null

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewColumnCollectionDialog.cs
Epica3055: why?

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewColumnCollectionDialog.cs
lonitra: Oops, my mistake here 🤦‍♀️ please ignore my suggestion

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewColumnCollectionDialog.cs
lonitra: For more cleanliness. It seems like we do not do anything special with these fields besides initially set them in the constructor and then get their value in various places, so we can clean this further by removing these backing fields and just defining these as read-only properties.

e.g. we can remove `_column` field and update `DataGridViewColumn` property to look like the following and it would be equivalent behavior to what we have now:
```c#
public DataGridViewColumn DataGridViewColumn { get; }
```

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.cs
Tanya-Solyanik: >you mean we should halt this PR? 

No, the PR is fine, except for the change in SYstem.Windows.Forms.dll. It seems that only the editor is not working.  The new issue would be about the DataSourceListEditor being disabled and `DataSource` entry in the property grid being gray (meaning readonly). Use the gif you pasted above in the new issue.

Tanya-Solyanik: Normally PropertyGrid control uses [TypeDescriptor ](https://learn.microsoft.com/dotnet/api/system.componentmodel.typedescriptor.geteditor?view=net-7.0)to discover Editor, first TypeDescriptor searches for an editor attribute on the specific property (DataSource in this case) then it searches for editor attribute on the type of that property (`object` in this case, so there is no appropriate editor found). I guess that the InProc designer provides a custom PropertyTypeDescriptor for the DataSource property to override the GetEditor method and provide the editor.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewColumnCollectionDialog.cs
Epica3055: With all due respect, I don't see any cleanliness.

lonitra: > If you did mean public, could you please tell me why.

`DataGridViewColumn` was `public` before so I assume it is being accessed outside of the class in some places. If we were to make it `private`, we'd no longer be able to access it outside of the class. Normally in a property, a `set` will also be specified, but since we did not specify that, it means that `DataGridViewColumn` is a read-only property, so it can be accessed outside the class, but cannot be set anywhere except the constructor of the class it is in.

> I am wondering, in your opinion, in what cases should we use fields.

If we make the change I mentioned, the compiler creates the fields for us behind the scenes, so we are still using fields, we just do not need to explicitly declare the fields. I think there are lots of instances when we need to explicitly declare the fields. Some that come to mind are when we need to keep track of values that we are not exposing at all or if we are needing to do any logic in the property `get`/`set` aside from a simple get/set such as the following:
https://github.com/dotnet/winforms/blob/aaf7e95fe4a4d8ea057f06763d1f7f3eb4157544/src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroup.cs#L130-L143
In this case, we are not doing any special logic like above, so imho there is no reason to explicitly declare the field when we can just use a read-only property and have the compiler declare the fields for us, but I am open to other opinions :)

Epica3055: Many thanks for your detailed explanation. 👍

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewDesigner.cs
```diff
@@ -0,0 +1,1196 @@
...
+            {
+                return ((DataGridView)_owner.Component).AllowUserToAddRows;
+            }
+
+            [
+                SuppressMessage("Microsoft.Performance", "CA1811:AvoidUncalledPrivateCode") // DAP calls this method thru Reflection.
+            ]
+            set
+            {
+                if (value != AllowUserToAddRows)
```
lonitra, line 1011: Let's invert this if statement. Some of the properties below also have this same sort of if check in `set`, could you invert those as well?

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewAddColumnDialog.cs
```diff
@@ -0,0 +1,996 @@
...
+        if (nameCreationService is not null && !nameCreationService.IsValidName(name))
+        {
+            if (!allowDuplicateNameInLiveColumnCollection || liveColumns is null || !liveColumns.Contains(name))
+            {
+                return false;
+            }
+        }
+
+        return true;
+    }
```
lonitra, line 900: This method and the method below are almost exactly the same. The only difference seems to be that the one below also has an `out string errorString` in its signature. 
It is hard to tell from looking on github, but could you see if it is necessary to keep both around? Could we just remove this method and update any calls to this to use the method below? 
If there are a lot of callers to this one then alternatively let's update this method to look like the following:

```suggestion
    public static bool ValidName(string name,
        DataGridViewColumnCollection columns,
        IContainer? container,
        INameCreationService? nameCreationService,
        DataGridViewColumnCollection liveColumns,
        bool allowDuplicateNameInLiveColumnCollection)
            => ValidName(name, columns, container, nameCreationService, liveColumns, allowDuplicateNameInLiveColumnCollection, out _);
```

lonitra (commented):
Looks like there are conflicts as well, but otherwise it LGTM!

lonitra (approved):
👍

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/DpiHelper.cs
```diff
@@ -139,7 +139,7 @@ private static InterpolationMode InterpolationMode
         }
     }
 
-    private static Bitmap ScaleBitmapToSize(Bitmap logicalImage, Size deviceImageSize)
+    internal static Bitmap ScaleBitmapToSize(Bitmap logicalImage, Size deviceImageSize, bool offset = true)
```
Tanya-Solyanik, line 142: The change in this file is not needed, because out of 6 callers into this API, none is setting `offset` to false.

image

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewAddColumnDialog.cs
Tanya-Solyanik: Could you please move these classes into dedicated files. You should name these new files like this - DataGridViewAddColumnDialog.ListBoxItem.cs. You will have to declare DataGridViewAddColumnDialog as partial class.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewAddColumnDialog.cs
```diff
@@ -0,0 +1,973 @@
...
+            }
+        }
+
+        errorString = string.Empty;
+        return true;
+    }
+
+    private class ListBoxItem
+    {
+        private Type _propertyType;
```
Tanya-Solyanik, line 921: both these fields can be declared `readonly` to improve nullability reasoning.

Tanya-Solyanik (approved):
Thank you!

John-Qiao:
Verified it in the main branch of Winforms repo for .Net 9.0, the DataGridviewDesigner can be designed in DemoConsole app runtime now.

9677-testresult

PR #9679, opened 2023-08-07
Don't serialize Hashtable with custom comparer

JeremyKuhne:
We don't support `Hashtable` manual serialization with custom comparers or hash code generators. Check for this case and allow it to fall through to the `BinaryFormatter`.  Add a few tests to validate.

Also move `BinaryFormatReader` code to `BinaryReaderExtensions`.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9679)


JeremyKuhne:
At some point in the future we might want to offer a "safe" mode where we ignore troublesome fields.

dreddy-work (approved):

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/BinaryFormattedObjectExtensions.cs
```diff
@@ -392,7 +392,9 @@ static bool Get(BinaryFormattedObject format, [NotNullWhen(true)] out object? ha
         {
             hashtable = null;
 
-            if (format.RecordCount < 5
+            // Note that hashtables with custom comparers and/or hash code providers will have that information before
+            // the value pair arrays.
+            if (format.RecordCount != 5
```
lonitra, line 397: Sorry, I don't think I follow the change to `!= 5`?

JeremyKuhne: Just being more restrictive.

PR #9681, opened 2023-08-08
Fix `Help.ShowHelp` to use `FindExecutable` to enable passing full local URL to default browser

### LINKED ISSUE #9599: Support HelpProvider HelpKeyword for local HTML files
### Background and motivation

Where `HelpNamespace` is set to a local HTML file, e.g. a HTML file that is bundled with the application (surely a common use case?), a HTML anchor that is specified using `SetHelpKeyword` will not be shown when pressing F1.

In `Help::ShowHTMLFile`, the URI is constructed correctly to include the keyword, `file = new Uri($"{file}#{stringParam}");`, but the URI is then presented using:

```
PInvoke.ShellExecute(handle.Handle, lpOperation: null, file.ToString(), lpParameters: null, lpDirectory: null, SHOW_WINDOW_CMD.SW_NORMAL);
```

For a local HTML file, this will throw away the anchor. However, at this point we already know that the URI is a HTML file, therefore this command could be updated to use the default browser to open the URI.

### API Proposal

In `Help`, first determine the default browser, e.g. by reading the registry key:

```
Registry.CurrentUser.OpenSubKey("Software\\Microsoft\\Windows\\CurrentVersion\\Explorer\\FileExts\\.htm\\UserChoice", false);
```

Then in `Help::ShowHTMLFile`, update the `ShellExecute` command to use the default browser to open the URI:

```
PInvoke.ShellExecute(handle.Handle, lpOperation: null, defaultBrowser, lpParameters: file.ToString(), lpDirectory: null, SHOW_WINDOW_CMD.SW_NORMAL);
```

### API Usage

No change

### Alternative Designs

_No response_

### Risks

_No response_

No

elachlan:
Fixes #9599

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9681)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Help.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Help.cs
@@ -238,20 +238,17 @@ public static class Help
         HandleRef<HWND> handle = parent is not null ? new(parent) : Control.GetHandleRef(PInvoke.GetActiveWindow());
         s_windowsFormsHelpTrace.TraceVerbose($"\tExecuting '{file}'");
         string fileName = file.ToString();
+        string? executable = null;
         if (file.IsFile)
         {
-            string? executable = FindExecutable(file.LocalPath.ToString());
-            PInvoke.ShellExecute(handle.Handle, lpOperation: null, executable ?? fileName, executable is not null ? fileName : null, lpDirectory: null, SHOW_WINDOW_CMD.SW_NORMAL);
-        }
-        else
-        {
-            PInvoke.ShellExecute(handle.Handle, lpOperation: null, fileName, lpParameters: null, lpDirectory: null, SHOW_WINDOW_CMD.SW_NORMAL);
+            executable = FindExecutableInternal(file.LocalPath.ToString());
         }
 
+        PInvoke.ShellExecute(handle.Handle, lpOperation: null, executable ?? fileName, executable is not null ? fileName : null, lpDirectory: null, SHOW_WINDOW_CMD.SW_NORMAL);
         GC.KeepAlive(handle.Wrapper);
     }
 
-    private static unsafe string? FindExecutable(string uri)
+    private static unsafe string? FindExecutableInternal(string uri)
     {
         HINSTANCE result;
         Span<char> buffer = stackalloc char[PInvoke.MAX_PATH + 1];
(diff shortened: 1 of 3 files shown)


elachlan:
CC: @decobeirne

src/System.Windows.Forms/src/System/Windows/Forms/Help.cs
```diff
@@ -240,24 +235,40 @@ private static void ShowHTMLFile(Control? parent, string? url, HelpNavigator com
...
+        if (file.IsFile)
         {
-            handle = new(parent);
+            string? executable = FindExecutable(file.LocalPath.ToString());
+            PInvoke.ShellExecute(handle.Handle, lpOperation: null, executable ?? fileName, executable is not null ? fileName : null, lpDirectory: null, SHOW_WINDOW_CMD.SW_NORMAL);
         }
         else
         {
-            handle = Control.GetHandleRef(PInvoke.GetActiveWindow());
+            PInvoke.ShellExecute(handle.Handle, lpOperation: null, fileName, lpParameters: null, lpDirectory: null, SHOW_WINDOW_CMD.SW_NORMAL);
```
dreddy-work, line 248: Seems this can be simplified as: 
```CS
 string? executable;
 if (file.IsFile)
        {
            executable = FindExecutable(file.LocalPath.ToString());
        }
      
  PInvoke.ShellExecute(handle.Handle, lpOperation: null, executable ?? fileName, executable is not null ? fileName : null, lpDirectory: null, SHOW_WINDOW_CMD.SW_NORMAL);
```

src/System.Windows.Forms/src/System/Windows/Forms/Help.cs
```diff
@@ -240,24 +235,40 @@ private static void ShowHTMLFile(Control? parent, string? url, HelpNavigator com
...
         }
 
-        WindowsFormsHelpTrace.TraceVerbose($"\tExecuting '{file}'");
-        PInvoke.ShellExecute(handle.Handle, lpOperation: null, file.ToString(), lpParameters: null, lpDirectory: null, SHOW_WINDOW_CMD.SW_NORMAL);
         GC.KeepAlive(handle.Wrapper);
     }
 
+    private static unsafe string? FindExecutable(string uri)
```
dreddy-work, line 254: we try to suffix the function names with `local` ( i.e: `FindExecutableLocal`) for the windows APIs, to avoid assumptions while reading code.

src/System.Windows.Forms/src/System/Windows/Forms/Help.cs
elachlan: Done.

elachlan:
@dreddy-work another flaky test

PR #9691, opened 2023-08-09
Enable nullability in BinaryEditor

halgab:
## Proposed changes

- Enable nullability in `BinaryEditor`

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9691)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/BinaryEditor.BinaryUI.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/BinaryEditor.BinaryUI.cs
@@ -94,8 +94,8 @@ public sealed partial class BinaryEditor
 
         private void ButtonOK_click(object? source, EventArgs e)
         {
-            object? localValue = _value;
-            ConvertToValue(_byteViewer.GetBytes(), ref localValue!);
+            object localValue = _value!;
+            ConvertToValue(_byteViewer.GetBytes(), ref localValue);
             _value = localValue;
         }
 
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/BinaryEditor.BinaryUI.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/BinaryEditor.BinaryUI.cs
@@ -95,7 +95,7 @@ public sealed partial class BinaryEditor
         private void ButtonOK_click(object? source, EventArgs e)
         {
             object? localValue = _value;
-            ConvertToValue(_byteViewer.GetBytes(), ref localValue);
+            ConvertToValue(_byteViewer.GetBytes(), ref localValue!);
             _value = localValue;
         }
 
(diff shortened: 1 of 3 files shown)


src/System.Windows.Forms.Design/src/System/ComponentModel/Design/BinaryEditor.BinaryUI.cs
```diff
@@ -95,7 +95,7 @@ private void RadioUnicode_checkedChanged(object? source, EventArgs e)
         private void ButtonOK_click(object? source, EventArgs e)
         {
             object? localValue = _value;
-            ConvertToValue(_byteViewer.GetBytes(), ref localValue);
+            ConvertToValue(_byteViewer.GetBytes(), ref localValue!);
```
dreddy-work, line 98: i would keep `!` on line 97 as `_value!`, if it can't be null here.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/BinaryEditor.cs
```diff
@@ -55,7 +59,8 @@ internal static byte[] ConvertToBytes(object value)
         {
             int size = stringValue.Length * 2;
             byte[] buffer = new byte[size];
-            Encoding.Unicode.GetBytes(stringValue.ToCharArray(), 0, size / 2, buffer, 0);
+            Encoding.Unicode.GetBytes(stringValue.AsSpan(), buffer.AsSpan());
```
dreddy-work, line 62: Do we have any tests covering this change?

halgab: I can't find any... Should I rollback?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/BinaryEditor.cs
```diff
@@ -55,7 +59,8 @@ internal static byte[] ConvertToBytes(object value)
...
+            buffer = Encoding.Unicode.GetBytes(stringValue);
```
lonitra, line 63: Curious why we are doing `GetBytes` twice here?

halgab: ... I didn't want to add this line, sorry about that. Reading the code, I didn't understand why we didn't do just that and had to manually allocate a buffer. But I figured the person who wrote this knew better, and decided against replacing line 62 with line 63.
I'll delete this right away

lonitra: 🤔 Good point! Honestly looking at the [source code](https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Text/Encoding.cs,624) for `GetBytes(string)`, it looks like we are essentially doing the same thing as that implementation except it seems we are just arbitrarly allocating a byte array that doubles the length of the string so I would be in favor of using `GetBytes(string)` instead of doing that allocation ourselves.

halgab: Yes, the current version avoids computing the array length. I don't see how it's a good enough optimisation to justify the added complexity

lonitra (approved):

PR #9707, opened 2023-08-10
Enable nullability in MaskDescriptor

halgab:
## Proposed changes

- Enable nullability in `MaskDescriptor`

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9707)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/MaskDescriptorComparer.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/MaskDescriptorComparer.cs
@@ -34,7 +34,8 @@ internal class MaskDescriptorComparer : IComparer<MaskDescriptor>
             return 0;
         }
 
-        string? textA, textB;
+        string? textA;
+        string? textB;
 
         switch (_sortType)
         {
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/MaskDescriptorComparer.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/MaskDescriptorComparer.cs
@@ -1,8 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-#nullable disable
-
 namespace System.Windows.Forms.Design;
 
 /// <summary>
@@ -27,7 +25,7 @@ internal class MaskDescriptorComparer : IComparer<MaskDescriptor>
         _sortOrder = sortOrder;
     }
 
-    public int Compare(MaskDescriptor maskDescriptorA, MaskDescriptor maskDescriptorB)
+    public int Compare(MaskDescriptor? maskDescriptorA, MaskDescriptor? maskDescriptorB)
     {
         if (maskDescriptorA is null || maskDescriptorB is null)
         {
@@ -36,7 +34,7 @@ internal class MaskDescriptorComparer : IComparer<MaskDescriptor>
             return 0;
         }
 
-        string textA, textB;
+        string? textA, textB;
 
         switch (_sortType)
         {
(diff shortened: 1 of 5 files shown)


src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/MaskDescriptorComparer.cs
```diff
@@ -36,7 +34,7 @@ public int Compare(MaskDescriptor maskDescriptorA, MaskDescriptor maskDescriptor
             return 0;
         }
 
-        string textA, textB;
+        string? textA, textB;
```
Tanya-Solyanik, line 37: Please move `textB` to a separate line.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/MaskDescriptorTemplate.cs
```diff
@@ -12,45 +10,39 @@ namespace System.Windows.Forms.Design;
...
 internal class MaskDescriptorTemplate : MaskDescriptor
 {
-    private readonly string _mask;
-    private readonly string _name;
-    private readonly string _sample;
-    private readonly Type _type;
-    private readonly CultureInfo _culture;
-
-    public MaskDescriptorTemplate(string mask, string name, string sample, Type validatingType, CultureInfo culture) : this(mask, name, sample, validatingType, culture, false)
+    public MaskDescriptorTemplate(string? mask, string name, string? sample, Type? validatingType, CultureInfo culture) : this(mask, name, sample, validatingType, culture, false)
```
Tanya-Solyanik, line 13: Please add argument name before `false`

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/MaskDesignerDialog.cs
```diff
@@ -31,7 +31,7 @@ internal class MaskDesignerDialog : Form
     private ErrorProvider _errorProvider;
 
     private readonly List<MaskDescriptor> _maskDescriptors = new();
-    private MaskDescriptor _customMaskDescriptor;
+    private MaskDescriptorTemplate _customMaskDescriptor;
```
Tanya-Solyanik, line 34: Why did you change field type?

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/MaskDesignerDialog.cs
```diff
@@ -346,7 +346,7 @@ private bool ContainsMaskDescriptor(MaskDescriptor maskDescriptor)
         {
             Debug.Assert(descriptor is not null, "Null mask descriptor in the collection.");
 
-            if (maskDescriptor.Equals(descriptor) || maskDescriptor.Name.Trim() == descriptor.Name.Trim())
+            if (maskDescriptor.Equals(descriptor) || maskDescriptor.Name!.Trim() == descriptor.Name!.Trim())
```
Tanya-Solyanik, line 349: Consider trimming and testing for null outside the loop, and return false if you got a null.
```
if (maskDescroptor is null || maskdescriptor.Name is null)
{
    return false;
}
```
and inside the loop - 
```
if (descriptor is null || descriptok.Name is null)
{
    return false;
}
```

halgab: Not sure I'm following here. All those can't be null, so why add the extra check? I'm moving the trimming outside the loop though

halgab: Sure. Should I explain it here or add a comment directly in the code?

halgab: We only add elements to the `_maskDescriptors` list

- at line 555, where we check `MaskDescriptor` for validity, which includes `Name` not being null
- at line 522, where we add `_customMaskDescriptor` whose `Name` we know is not null (because we created it)

halgab:
It should be good now

lonitra (dismissed):

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/MaskDesignerDialog.cs
Tanya-Solyanik: Thank you. Could you please add a comment in the code.

halgab: I added an assert for `maskdescriptor.Name` and a comment for `descriptor.Name`

lonitra (approved):

PR #9713, opened 2023-08-13
Use throw helpers for ArgumentOutOfRangeException

halgab:
I applied the CA1512 fixer across the codebase, and just deleted some blank lines manually.

There are places that the analyzer did not fix. I'm doing a manual pass locally and can follow-up on that.

Also, in dotnet/runtime#79460, custom error messages were replaced by the standard one. If that's something that we are willing to consider here, I can follow-up with another commit/PR

## Proposed changes

- Apply CA1512 fixer: Use ArgumentOutOfRangeException throw helper

## Risk

- low: changes are from anaylzers

## Test methodology 

- CI

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9713)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/NullRecord.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/NullRecord.cs
@@ -15,10 +15,7 @@ internal abstract partial class NullRecord
         get => _count;
         private protected set
         {
-            if (value == 0)
-            {
-                throw new ArgumentOutOfRangeException(nameof(value));
-            }
+            ArgumentOutOfRangeException.ThrowIfZero(value);
 
             _count = value;
         }
--- a/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/NullRecord.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/NullRecord.cs
@@ -15,7 +15,10 @@ internal abstract partial class NullRecord
         get => _count;
         private protected set
         {
-            ArgumentOutOfRangeException.ThrowIfZero(value);
+            if (value == 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(value));
+            }
 
             _count = value;
         }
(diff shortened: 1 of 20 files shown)


src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/NullRecord.cs
```diff
@@ -15,10 +15,7 @@ public virtual Count NullCount
         get => _count;
         private protected set
         {
-            if (value == 0)
-            {
-                throw new ArgumentOutOfRangeException(nameof(value));
-            }
+            ArgumentOutOfRangeException.ThrowIfZero(value);
```
halgab: This one doesn't work. The analyzer is broken for this specific scenario apparently

lonitra (approved):
Awesome! Thanks for doing this

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/NullRecord.cs
lonitra: 😮 Curious what error you were seeing when you did have this change?

lonitra:
> Also, in https://github.com/dotnet/runtime/pull/79460, custom error messages were replaced by the standard one. If that's something that we are willing to consider here, I can follow-up with another commit/PR

I think it would be good to replace custom error messages with standard ones especially if they are user facing, but I will let others give their opinion on this as well.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/NullRecord.cs
halgab: The problem is that `value` is of type `System.Windows.Forms.Count`, which doesn't implement the `INumberBase` interface required for the throw helper, even though it's implicitely convertible to `int`

Tanya-Solyanik:
@halgab  - could you please open a work item for this - 
`Also, in https://github.com/dotnet/runtime/pull/79460, custom error messages were replaced by the standard one. If that's something that we are willing to consider here, I can follow-up with another commit/PR`

halgab:
Not sure I used the best wording, but here it is: #9726 9726

PR #9728, opened 2023-08-15
Use throw helpers for ArgumentOutOfRangeException: manual cases

halgab:
This is the first follow-up to #9713. These are all the cases not caught by CA1512 but are legitimate uses of the new `ArgumentOutOfRangeException` throw helpers, because the error message is not overridden.

## Proposed changes

- Use ArgumentOutOfRangeException throw helpers

## Risk

- low, but higher than the pure fixer modifications. I still used the IDE as much as possible (e.g. to break if clauses into 2) so that CA1512 could apply

## Test methodology

- CI

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9728)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRow.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRow.cs
@@ -1217,10 +1217,8 @@ public partial class DataGridViewRow : DataGridViewBand
                 throw new InvalidOperationException(SR.DataGridView_InvalidOperationOnSharedRow);
             }
 
-            if (rowIndex < 0 || rowIndex >= DataGridView.Rows.Count)
-            {
-                throw new ArgumentOutOfRangeException(nameof(rowIndex));
-            }
+            ArgumentOutOfRangeException.ThrowIfNegative(rowIndex);
+            ArgumentOutOfRangeException.ThrowIfGreaterThanOrEqual(rowIndex, DataGridView.Rows.Count);
 
             if (DataGridView.VirtualMode || DataGridView.DataSource is not null)
             {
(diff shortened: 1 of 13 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRowHeaderCell.cs
```diff
@@ -641,9 +640,10 @@ protected override object GetValue(int rowIndex)
     {
         // We allow multiple rows to share the same row header value. The row header cell's cloning does this.
         // So here we need to allow rowIndex == -1.
-        if (DataGridView is not null && (rowIndex < -1 || rowIndex >= DataGridView.Rows.Count))
+        if (DataGridView is not null)
         {
-            throw new ArgumentOutOfRangeException(nameof(rowIndex));
+            ArgumentOutOfRangeException.ThrowIfLessThan(rowIndex, -1);
+            ArgumentOutOfRangeException.ThrowIfGreaterThanOrEqual(rowIndex, DataGridView.Rows.Count);
```
lonitra, line 646: Here as well.

src/System.Windows.Forms/src/System/Windows/Forms/LinkClickedEventArgs.cs
```diff
@@ -28,8 +28,9 @@ public class LinkClickedEventArgs : EventArgs
     public LinkClickedEventArgs(string? linkText, int linkStart, int linkLength)
     {
         ArgumentOutOfRangeException.ThrowIfNegative(linkStart);
+        ArgumentOutOfRangeException.ThrowIfNegative(linkLength);
 
-        if (linkLength < 0 || linkStart + linkLength < 0)
+        if (linkStart + linkLength < 0)
             throw new ArgumentOutOfRangeException(nameof(linkLength));
```
lonitra, line 34: It seems like this can be removed completely. We already check that neither `linkStart` or `linkLength` is negative so I don't see how `linkStart + linkLength` can be negative.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRow.cs
```diff
@@ -1378,9 +1372,10 @@ internal bool GetSelected(int rowIndex)
     [EditorBrowsable(EditorBrowsableState.Advanced)]
     public virtual DataGridViewElementStates GetState(int rowIndex)
     {
-        if (!(DataGridView is null || (rowIndex >= 0 && rowIndex < DataGridView.Rows.Count)))
+        if (DataGridView is not null)
         {
-            throw new ArgumentOutOfRangeException(nameof(rowIndex));
+            ArgumentOutOfRangeException.ThrowIfNegative(rowIndex);
+            ArgumentOutOfRangeException.ThrowIfGreaterThanOrEqual(rowIndex, DataGridView.Rows.Count);
```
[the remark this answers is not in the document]
lonitra: Nevermind on this. I am seeing now it was just me 😄

lonitra (approved):
👍

PR #9732, opened 2023-08-16
Fix IRawElementProviderSimple.GetPropertyValue

JeremyKuhne:
Our interop handle structs (such as HWND) cannot be marshalled directly and will fail "silently" on callbacks (they will throw but the marshaller will convert that to an HRESULT that we won't see unless first-chance exceptions are on when we're debugging).

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9732)


JeremyKuhne:
We'll want to pull this into RC2 at the very least.

Tanya-Solyanik (approved):

PR #9745, opened 2023-08-17
Ensure Form Doesn't Prevent Shutdown

### LINKED ISSUE #9210: .NET 8.0 Preview 3 Winforms application doesn't quit on shutdown or logoff
### .NET version

.NET 8.0 Preview 3

Yes

Yes, the code works well in .NET 7.0 and previous.

### Issue description

I created a simple Winforms application written in C# with VS Professional 2022 (64-bit) version 17.6.2.4 and .NET 8.0 Preview 3
The application contains an empty form that does nothing.
If I try to shutdown Windows the system stucks and tells me: "This app is preventing shutdown".
The Windows system is Windows 10 Pro (v. 10.0.18362) but the behaviour is the same in Windows 11 too.

### Steps to reproduce

Create an empty Winforms application targeting .NET 8.0 Preview 3.
Compile in release, execute, and shutdown Windows.

lonitra:
Fixes #9210 

In #7561 the result for `WM_QUERYENDSESSION` was accidentally flipped causing the return for `WM_QUERYENDSESSION` to be `false`, which would prevent shutdown/log off. This change fixes this and also adds a test that ensures that `Form` will return `true` for a `WM_QUERYENDSESSION` message.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9745)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Form.cs
@@ -6238,7 +6238,7 @@ public partial class Form : ContainerControl
 
             if (m.MsgInternal == PInvoke.WM_QUERYENDSESSION)
             {
-                m.ResultInternal = (LRESULT)(nint)(BOOL)e.Cancel;
+                m.ResultInternal = (LRESULT)(BOOL)!e.Cancel;
             }
             else if (e.Cancel && (MdiParent is not null))
             {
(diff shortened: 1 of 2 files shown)


JeremyKuhne (approved):
:shipit:

Issue #9726, opened 2023-08-15
Review replacing ArgumentOutOfRangeException with custom error messages with default ones using its throw helpers 


halgab:
In dotnet/runtime#79460, a choice was made to replace several `throw new ArgumentOutOfRangeException` with custom error messages with `ArgumentOutOfRangeException.ThrowIf` methods that use a standard message. As a follow up to #9713, we could do the same in the winforms repo

elachlan:
@merriemcgaw is the team okay with this?

merriemcgaw:
Yes, I think it looks like something we can do for .NET 9

halgab:
#9755 is a draft PR that showcases some examples of the kind of modification I think would be profitable.

PR #9755, opened 2023-08-18
Replacing some ArgumentOutOfRangeException with custom messages with throw helpers

halgab:
Contributes to #9726

## Proposed changes

This is a follow-up to #9728, only this time we are dealing with less standard `ArgumentOutOfRangeException`s, so the message will change a bit. Here, I made changes only to cases where I figured the messages added little value to the default ones:

- SR.InvalidArgument
- SR.InvalidLowBoundArgumentEx
- SR.InvalidLowBoundArgument
- SR.InvalidHighBoundArgumentEx
- SR.InvalidHighBoundArgument
- SR.InvalidBoundArgument
- SR.IndexOutOfRange

## Risk

- low, but higher than the pure fixer modifications. I still used the IDE as much as possible (e.g. to break if clauses into 2) so that CA1512 could apply. To do so, I manually modified the exception to remove the custom message after verfying that we do not lose any information by doing so.

## Test methodology

- CI

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9755)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/HtmlElementCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/HtmlElementCollection.cs
@@ -48,10 +48,8 @@ public sealed unsafe class HtmlElementCollection : ICollection
         get
         {
             //do some bounds checking here...
-            if (index < 0 || index >= Count)
-            {
-                throw new ArgumentOutOfRangeException(nameof(index), index, string.Format(SR.InvalidBoundArgument, nameof(index), index, 0, Count - 1));
-            }
+            ArgumentOutOfRangeException.ThrowIfNegative(index);
+            ArgumentOutOfRangeException.ThrowIfGreaterThanOrEqual(index, Count);
 
             if (NativeHtmlElementCollection is not null)
             {
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListView.ColumnHeaderCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListView.ColumnHeaderCollection.cs
@@ -26,9 +26,11 @@ public partial class ListView
         {
             get
             {
-                if (_owner._columnHeaders is null || index < 0 || index >= _owner._columnHeaders.Length)
+                ArgumentOutOfRangeException.ThrowIfNegative(index);
+
+                if (_owner._columnHeaders is null || index >= _owner._columnHeaders.Length)
                 {
-                    throw new ArgumentOutOfRangeException(nameof(index), index, string.Format(SR.InvalidArgument, nameof(index), index));
+                    throw new ArgumentOutOfRangeException(nameof(index));
                 }
 
                 return _owner._columnHeaders[index];
@@ -410,10 +412,8 @@ public partial class ListView
 
         public void Insert(int index, ColumnHeader value)
         {
-            if (index < 0 || index > Count)
-            {
-                throw new ArgumentOutOfRangeException(nameof(index), index, string.Format(SR.InvalidArgument, nameof(index), index));
-            }
+            ArgumentOutOfRangeException.ThrowIfNegative(index);
+            ArgumentOutOfRangeException.ThrowIfGreaterThan(index, Count);
 
             _owner.InsertColumn(index, value);
         }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListView.ColumnHeaderCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListView.ColumnHeaderCollection.cs
@@ -22,20 +22,7 @@ public partial class ListView
         ///  Given a Zero based index, returns the ColumnHeader object
         ///  for the column at that index
         /// </summary>
-        public virtual ColumnHeader this[int index]
-        {
-            get
-            {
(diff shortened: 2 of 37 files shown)


elachlan:
The number of lines difference is awesome. Great work.

Tanya-Solyanik:
@halgab  - is this still a draft?

halgab:
I initially put it as a draft to showcase the related issue, but reading again the reaction from the team, I figured I'd put it ready for review. And once again the draft label wasn't automatically removed (and I don't think I can change it manually)

src/System.Windows.Forms/src/System/Windows/Forms/HtmlElementCollection.cs
```diff
@@ -42,10 +42,8 @@ internal HtmlElementCollection(HtmlShimManager shimManager, HtmlElement[] array)
         get
         {
             //do some bounds checking here...
-            if (index < 0 || index >= Count)
-            {
-                throw new ArgumentOutOfRangeException(nameof(index), index, string.Format(SR.InvalidBoundArgument, nameof(index), index, 0, Count - 1));
```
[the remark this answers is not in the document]
lonitra: I think it is ok in this scenario

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
```diff
@@ -5066,30 +5058,12 @@ protected void RealizeProperties()
...
-            }
-
-            if (endIndex < 0 || endIndex >= Items.Count)
-            {
-                throw new ArgumentOutOfRangeException(nameof(endIndex), endIndex, string.Format(SR.InvalidArgument, nameof(endIndex), endIndex));
-            }
-        }
+        int maxSize = VirtualMode ? VirtualListSize : Items.Count;
+        ArgumentOutOfRangeException.ThrowIfGreaterThanOrEqual(startIndex, maxSize);
+        ArgumentOutOfRangeException.ThrowIfGreaterThanOrEqual(endIndex, maxSize);
```
lonitra, line 5066: Nice!

src/System.Windows.Forms/src/System/Windows/Forms/ListView.ColumnHeaderCollection.cs
```diff
@@ -26,9 +26,11 @@ public ColumnHeaderCollection(ListView owner)
         {
             get
             {
-                if (_owner._columnHeaders is null || index < 0 || index >= _owner._columnHeaders.Length)
+                ArgumentOutOfRangeException.ThrowIfNegative(index);
+
+                if (_owner._columnHeaders is null || index >= _owner._columnHeaders.Length)
```
lonitra, line 31: Any reason we are not doing `ArgumentOutOfRangeException(index, _owner._columnHeaders?.Length ?? 0)` like we did in ComboBox.cs here?

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
```diff
@@ -5352,9 +5326,9 @@ internal unsafe void SetColumnInfo(LVCOLUMNW_MASK mask, ColumnHeader ch)
     /// </summary>
     internal void SetColumnWidth(int columnIndex, ColumnHeaderAutoResizeStyle headerAutoResize)
     {
-        if ((columnIndex < 0) ||
-            (columnIndex >= 0 && _columnHeaders is null) ||
-            (columnIndex >= _columnHeaders!.Length))
+        ArgumentOutOfRangeException.ThrowIfNegative(columnIndex);
+
+        if (_columnHeaders is null || columnIndex >= _columnHeaders.Length)
```
lonitra, line 5331: Same here with using `ArgumentOutOfRangeException.ThrowIfGreaterOrEqual(columnIndex, _columnHeaders?.Length ?? 0)`

src/System.Windows.Forms/src/System/Windows/Forms/ListView.ColumnHeaderCollection.cs
halgab: `ArgumentOutOfRangeException.ThrowIfGreaterOrEqual` has an error message that will mention 0, which can be misleading for the null case. I'd be tempted to change what I did in ComboBox.cs rather than use that pattern everywhere. What do you think?

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
halgab: https://github.com/dotnet/winforms/pull/9755/files#r1379358840

lonitra: I think it would be helpful here as well as it would indicate user is trying to set column width on a column of a certain index in an empty collection.

src/System.Windows.Forms/src/System/Windows/Forms/ListView.ColumnHeaderCollection.cs
halgab: This raised some problems with NRT annotations, so I spun out this code to a helper function, that I also use at the other place you wanted to see changed

lonitra (approved):

PR #9769, opened 2023-08-22
Additional COM fixes

JeremyKuhne:
COM code audit fixes:

- AxContainer.IOleInPlaceFrame.SetActiveObject was not ref counting correctly or checking for null
- OleInterfaces.IOleControlSite.GetExtendedControl was unnecessarily using Marshal
- OleIntefaces.IOleInPlaceSite.GetWindowContext brought in line with MFC behavior with null pointers
- Use static char array in ActiveXImpl (minor perf fix)
- ActiveXImpl.InPlaceActivate was incorrectly ref counting
- ActiveXImpl Save/Load were releasing IPropertyBag when it was passed directly from native code, move the using local to the internal code
- ActiveXImpl.QuickActivate should not fail if we cannot get IConnectionPointContainer
- Add DisposeHelper.NullAndRelease helper for safer release of pointer fields
- Harden AgileComPointer a bit more by using Interlocked

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9769)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/AgileComPointer.cs
+++ b/src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/AgileComPointer.cs
@@ -102,7 +102,7 @@ internal unsafe class AgileComPointer<TInterface> :
         // Clear the cookie before revoking the interface to guard against re-entry.
 
         uint cookie = Interlocked.Exchange(ref _cookie, 0);
-        if (_cookie == 0)
+        if (cookie == 0)
         {
             return;
         }
--- a/src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/AgileComPointer.cs
+++ b/src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/AgileComPointer.cs
@@ -99,14 +99,13 @@ internal unsafe class AgileComPointer<TInterface> :
 
     protected virtual void Dispose(bool disposing)
     {
+        // Clear the cookie before revoking the interface to guard against re-entry.
+
+        uint cookie = Interlocked.Exchange(ref _cookie, 0);
         if (_cookie == 0)
         {
             return;
         }
-
-        // Clear the cookie before revoking the interface to guard against re-entry.
-        uint cookie = _cookie;
-        _cookie = 0;
 
         HRESULT hr = GlobalInterfaceTable.RevokeInterface(cookie);
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/AxHost.AxContainer.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/AxHost.AxContainer.cs
@@ -684,7 +684,12 @@ public abstract partial class AxHost
                 // Release the field before disposing to avoid accessing it during disposal on callbacks.
                 activeHost._iOleInPlaceActiveObjectExternal = null;
                 existing.Dispose();
-                activeHost._iOleInPlaceActiveObjectExternal = new(pActiveObject, takeOwnership: true);
+
+                if (pActiveObject is not null)
+                {
+                    pActiveObject->AddRef();
+                    activeHost._iOleInPlaceActiveObjectExternal = new(pActiveObject, takeOwnership: true);
+                }
             }
 
             if (pActiveObject is null)
@@ -741,7 +746,7 @@ public abstract partial class AxHost
             return HRESULT.S_OK;
         }
 
-        unsafe HRESULT IOleInPlaceFrame.Interface.InsertMenus(HMENU hmenuShared, OLEMENUGROUPWIDTHS* lpMenuWidths)
+        HRESULT IOleInPlaceFrame.Interface.InsertMenus(HMENU hmenuShared, OLEMENUGROUPWIDTHS* lpMenuWidths)
         {
             s_axHTraceSwitch.TraceVerbose("in InsertMenus");
             return HRESULT.S_OK;
--- a/src/System.Windows.Forms/src/System/Windows/Forms/AxHost.OleInterfaces.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/AxHost.OleInterfaces.cs
@@ -429,7 +429,6 @@ public abstract partial class AxHost
                 return HRESULT.E_POINTER;
             }
 
-            *ppDoc = null;
             *ppFrame = ComHelpers.GetComPointer<IOleInPlaceFrame>(_host.GetParentContainer());
             *lprcPosRect = _host.Bounds;
             *lprcClipRect = WebBrowserHelper.GetClipRect();
--- a/src/System.Windows.Forms/src/System/Windows/Forms/AxHost.OleInterfaces.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/AxHost.OleInterfaces.cs
@@ -160,13 +160,13 @@ public abstract partial class AxHost
                 return HRESULT.E_POINTER;
             }
 
-            object? proxy = _host.GetParentContainer().GetExtenderProxyForControl(_host);
+            AxContainer.ExtenderProxy? proxy = _host.GetParentContainer().GetExtenderProxyForControl(_host);
             if (proxy is null)
             {
                 return HRESULT.E_NOTIMPL;
             }
 
-            *ppDisp = (IDispatch*)Marshal.GetIDispatchForObject(proxy);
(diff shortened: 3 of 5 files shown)


src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/AgileComPointer.cs
```diff
@@ -99,15 +99,14 @@ public void Dispose()
 
     protected virtual void Dispose(bool disposing)
     {
+        // Clear the cookie before revoking the interface to guard against re-entry.
+
+        uint cookie = Interlocked.Exchange(ref _cookie, 0);
```
lonitra, line 104: Should this be after the if statement? Or check for `cookie == 0` instead?

src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
```diff
@@ -116,6 +116,7 @@ private sealed unsafe partial class ActiveXImpl : MarshalByRefObject, IWindowTar
         private static readonly int s_uiActive = BitVector32.CreateMask(s_inPlaceVisible);
         private static readonly int s_uiDead = BitVector32.CreateMask(s_uiActive);
         private static readonly int s_adjustingRect = BitVector32.CreateMask(s_uiDead);
+        private static readonly char[] s_whitespace = new char[] { ' ', '\r', '\n' };
```
hughbe, line 119: Use SearchValues? https://learn.microsoft.com/en-us/dotnet/api/system.buffers.searchvalues?view=net-8.0

https://github.com/search?q=repo%3Adotnet%2Fruntime%20SearchValues&type=code

src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
```diff
@@ -529,7 +530,7 @@ internal void Close(OLECLOSE dwSaveOption)
         /// </summary>
         private static byte[] FromBase64WrappedString(string text)
         {
-            if (text.IndexOfAny(new char[] { ' ', '\r', '\n' }) != -1)
+            if (text.IndexOfAny(s_whitespace) != -1)
```
halgab, line 533: Or simply 
```suggestion
            if (text.AsSpan().ContainsAny(' ', '\r', '\n'))
```
though we could use the computed index to slightly optimise the loop below

src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/AgileComPointer.cs
JeremyKuhne: Oops, good catch.

src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
JeremyKuhne: Thanks, hadn't seen this one yet.

src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
JeremyKuhne: I'm going to go the `SearchValues` route for now. Happy to entertain any other optimization PRs.

lonitra (approved):
:shipit:

src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
halgab: Wouldn't it be better in that case to replace the other manual references to `' ', '\r', '\n'` by a `s_whitespace.Contains` around line 540?

halgab: Alternatively I could do that in a follow-up PR that would consolidate the version of `FromBase64WrappedString` and [the one in `ResXDataNode`](https://github.com/dotnet/winforms/blob/061d0faf1241dff04a876dd47c2c98cbb44bf765/src/System.Windows.Forms/src/System/Resources/ResXDataNode.cs#L637)

src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
```diff
@@ -116,6 +117,7 @@ private sealed unsafe partial class ActiveXImpl : MarshalByRefObject, IWindowTar
         private static readonly int s_uiActive = BitVector32.CreateMask(s_inPlaceVisible);
         private static readonly int s_uiDead = BitVector32.CreateMask(s_uiActive);
         private static readonly int s_adjustingRect = BitVector32.CreateMask(s_uiDead);
+        private static readonly SearchValues<char> s_whitespace = SearchValues.Create(new char[] { ' ', '\r', '\n' });
```
halgab, line 120:
```suggestion
        private static readonly SearchValues<char> s_whitespace = SearchValues.Create(stackalloc char[] { ' ', '\r', '\n' });
```

Shyam-Gupta (dismissed):

src/System.Windows.Forms/src/System/Windows/Forms/Control.ActiveXImpl.cs
JeremyKuhne: Interesting, didn't know you could do this. That said, in the runtime they're using string literals, so that's what I'm going to do here.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.OleInterfaces.cs
```diff
@@ -412,16 +412,24 @@ HRESULT IOleInPlaceSite.Interface.OnUIActivate()
             RECT* lprcClipRect,
             OLEINPLACEFRAMEINFO* lpFrameInfo)
         {
-            ppDoc = null;
-
-            if (ppFrame is null || lprcPosRect is null || lprcClipRect is null)
+            if (ppDoc is null || ppFrame is null)
             {
                 return HRESULT.E_POINTER;
```
Tanya-Solyanik, line 417: Docs say that in the case of an error, all parameters should be set to NULL, we are not doing it.

Tanya-Solyanik (dismissed):

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.AxContainer.cs
```diff
@@ -684,7 +684,12 @@ HRESULT IOleInPlaceFrame.Interface.SetActiveObject(IOleInPlaceActiveObject* pAct
                 // Release the field before disposing to avoid accessing it during disposal on callbacks.
                 activeHost._iOleInPlaceActiveObjectExternal = null;
                 existing.Dispose();
-                activeHost._iOleInPlaceActiveObjectExternal = new(pActiveObject, takeOwnership: true);
+
+                if (pActiveObject is not null)
+                {
+                    pActiveObject->AddRef();
+                    activeHost._iOleInPlaceActiveObjectExternal = new(pActiveObject, takeOwnership: true);
```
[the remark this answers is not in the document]
JeremyKuhne: It would throw an exception if it was null as the `AgileComPointer` does not allow a null.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.OleInterfaces.cs
JeremyKuhne: Good catch. I was looking at the wrong implementation in MFC, updated to match the host class.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.OleInterfaces.cs
```diff
@@ -412,16 +412,28 @@ HRESULT IOleInPlaceSite.Interface.OnUIActivate()
...
+            {
+                *ppDoc = null;
+            }
+
+            if (ppDoc is null || ppFrame is null || lprcPosRect is null || lprcClipRect is null)
             {
                 return HRESULT.E_POINTER;
             }
 
+            *ppDoc = null;
```
Tanya-Solyanik, line 432: This arguments is already nulled-out on line424

Tanya-Solyanik (approved):

lonitra (approved):

PR #9776, opened 2023-08-24
More COM audit fixes

JeremyKuhne:
Additional audit fixes:

- IGetOleObject should be returning IOleObject, not IUnknown. Confirmed against VB6 code.
- ClassPropertyDispatchAdapter was allowing conflicts with explicit [DispId]
- Clarify code a bit in ClassPropertyDispatchAdapter.Invoke, use VARIANT marshalling, and allow no result in getter
- Tweak ITypeLib import to throw for clarity in StandardDispatch
- Validate DISPPARAMS parameter in StandardDispatch Invoke methods
- Remove unused method in StandardDispatch
- Add a few unit tests for ClassPropertyDispatchAdapterTests
- Change namespace for IDispatchTests
- Minor source cleaning in COM type descriptor code
- Throw with OleCreatePropertyFrame as we used to.
- Use ComHelpers for one missed IPerPropertyBrowsing usage in Com2IPerPropertyBrowsingEnum
- Don't create in VARIANT for ExecuteBuilder call in COM2PropertyBuilderUITypeEditor (validated with all implementations in VS and Windows)
- Com2PropertyDescriptor.SetValue should store the object not the VARIANT in last value and should free the VARIANT
- Remove IErrorInfo wrapper and use the API directly
- Use VARIANT method to get object in ProcessTypeInfoEnum
- Add a simple test to validate interop behavior for int/uint and add comment to VARIANT

Thanks to @lonitra for going through this with me

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9776)


JeremyKuhne:
Added another round of changes as this hasn't gotten reviewed yet.

src/System.Windows.Forms/tests/InteropTests/PropertyGridTests.cs
```diff
@@ -27,7 +27,7 @@ public void ISupportErrorInfo_Supported_ButNoIErrorInfoGiven()
                 try
                 {
                     encodingEntry.SetPropertyTextValue("333");
-                    Assert.False(true, "Invalid set values should produce ExternalException which will be presenttted to the user.");
+                    Assert.False(true, "Invalid set values should produce ExternalException which will be presented to the user.");
```
KlausLoeffelmann, line 30: Took me a while to 'decipher' this Assert.
```suggestion
                    // We should never arrive here. If we do:
                    Assert.False(true, "Passing an invalid set of values should produce 'ExternalException' which will be presented to the user.");
```

src/System.Windows.Forms.Primitives/src/Windows/Win32/System/Ole/ClassPropertyDispatchAdapter.cs
```diff
@@ -175,28 +184,21 @@ public bool TryGetMemberName(int dispId, [NotNullWhen(true)] out string? name)
...
             if (parameters->cArgs != 1)
             {
                 return HRESULT.DISP_E_BADPARAMCOUNT;
             }
 
             try
             {
-                object? value = InteropMarshal.GetObjectForNativeVariant((nint)parameters->rgvarg);
+                VARIANT* variantValue = parameters->rgvarg;
+                object? value = variantValue is null ? null : variantValue->ToObject();
```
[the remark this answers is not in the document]
JeremyKuhne: It doesn't work with pointers :(

src/System.Windows.Forms/tests/InteropTests/PropertyGridTests.cs
JeremyKuhne: Yeah, I don't know why I didn't do this as Assert.Fail

Shyam-Gupta (approved):
LGTM

src/System.Windows.Forms.Primitives/src/Interop/OleAut32/Interop.VARIANT.cs
```diff
@@ -889,6 +889,7 @@ private static Span<T> GetSpan<T>(Array array)
     public static explicit operator VARIANT(int value)
         => new()
         {
+            // Legacy marshalling uses VT_I4, not VT_INT
```
Tanya-Solyanik, line 892: Just curious, what is legacy marshalling? Is it possible to add a pointer to the source code?

Tanya-Solyanik (approved):

lonitra (approved):

src/System.Windows.Forms.Primitives/src/Interop/OleAut32/Interop.VARIANT.cs
JeremyKuhne: Built-in COM marshalling as opposed to ComWrappers.

src/System.Windows.Forms/tests/InteropTests/PropertyGridTests.cs
JeremyKuhne: I'll turn this into an Assert.Fail in another PR.

PR #9797, opened 2023-08-28
More COM audit fixes (#9776)

JeremyKuhne:
## Port of #9776

Additional audit fixes:

- IGetOleObject should be returning IOleObject, not IUnknown. Confirmed against VB6 code.
- ClassPropertyDispatchAdapter was allowing conflicts with explicit [DispId]
- Clarify code a bit in ClassPropertyDispatchAdapter.Invoke, use VARIANT marshalling, and allow no result in getter
- Tweak ITypeLib import to throw for clarity in StandardDispatch
- Validate DISPPARAMS parameter in StandardDispatch Invoke methods
- Remove unused method in StandardDispatch
- Add a few unit tests for ClassPropertyDispatchAdapterTests
- Change namespace for IDispatchTests
- Minor source cleaning in COM type descriptor code
- Throw with OleCreatePropertyFrame as we used to.
- Use ComHelpers for one missed IPerPropertyBrowsing usage in Com2IPerPropertyBrowsingEnum
- Don't create in VARIANT for ExecuteBuilder call in COM2PropertyBuilderUITypeEditor (validated with all implementations in VS and Windows)
- Com2PropertyDescriptor.SetValue should store the object not the VARIANT in last value and should free the VARIANT
- Remove IErrorInfo wrapper and use the API directly
- Use VARIANT method to get object in ProcessTypeInfoEnum
- Add a simple test to validate interop behavior for int/uint and add comment to VARIANT

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9797)


Shyam-Gupta (approved):

src/System.Windows.Forms.Primitives/src/Interop/OleAut32/Interop.VARIANT.cs
```diff
@@ -889,6 +889,7 @@ private static Span<T> GetSpan<T>(Array array)
     public static explicit operator VARIANT(int value)
         => new()
         {
+            // Legacy marshalling uses VT_I4, not VT_INT
```
Tanya-Solyanik, line 892: Is this legacy marshalling implemented in a public repo we could add a link to?

PR #9834, opened 2023-09-01
Remove Tracing from AxHost

lonitra:
We have trace statements everywhere in AxHost and most of our trace statements do not seem to add much value as it just highlights what methods were entered. If needed, we can use debug actions in VS to output messages capturing points of interest when debugging instead of having traces throughout our code, so removing them.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9834)


Tanya-Solyanik (approved):
Looks good. 
Difference between these traces and VS actions is that these do not require a debugger attached and can be turned on/off as needed by the app user, at least in the .NET Framework. I don't think this works in Core fully, so I don't mind removing them

PR #9733, opened 2023-08-16
Make sure the Narrator won't announce the password text

SimonZhao888:
Fixes #7852

## Proposed changes

- Implemented HasKeyboardFocusPropertyId for TextBoxBaseAccessibleObject.
- Update the Text property in the TextBoxBaseUiaTextProvider to ensure that passwords are not displayed in plaintext in Insights.

## Customer Impact

- When enable UseSystemPasswordChar or PasswordChar properties in TextBox/MaskedTextBox, Narrator won't announce the text.
- Passwords are not displayed as plaintext in Insights.

## Regression? 

- No

## Risk

- Low

### Before

When typing text in TextBox/MaskedTextBox with UseSystemPasswordChar or PasswordChar, Narrator announce text to "Hidden", but when focus on TextBox/MaskedTextBox or specific text, it will announce as specific text.
image

### After

When enable UseSystemPasswordChar or PasswordChar properties in TextBox/MaskedTextBox, Narrator won't announce the text, and passwords are not displayed as plaintext in Insights.
PassWord

image

## Test methodology 

- Manual

## Test environment(s) 

-  8.0.0-rc.1.23381.3
- Windows 11 22H2

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9733)

PR #9838, opened 2023-09-05
Add Password tests to WinformsControlsTest

SimonZhao888:
Related Comments #https://github.com/dotnet/winforms/pull/9733#discussion_r1313501924

## Proposed changes

- Add Password tests to WinformsControlsTest, so that we can use it to test the Password functionality of textBoxes and the like.

## Customer Impact

- We can use WinformsControlsTest App to test the Password functionality of textBoxes and the like.

## Regression? 

- No

## Risk

- Minimal

### Before
No Password test form in WinformsControlsTest App.
image

### After
We can use WinformsControlsTest App to test the Password functionality of textBoxes and the like.
image

## Test methodology 

- Manual (WinformsControlsTest)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9838)


src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/PassWord.Designer.cs
```diff
@@ -0,0 +1,282 @@
...
+        this.label5 = new System.Windows.Forms.Label();
+        this.label6 = new System.Windows.Forms.Label();
+        this.label7 = new System.Windows.Forms.Label();
+        this.label8 = new System.Windows.Forms.Label();
+        this.label9 = new System.Windows.Forms.Label();
+        this.SuspendLayout();
+        // 
+        // textBox1
+        // 
+        this.textBox1.AccessibleName = "Test_Box1";
```
Tanya-Solyanik, line 56: Good job setting up AccessibleNames! However in this case, when textboxes are preceded by labels, when AccessibleName is not set up explicitly, AccessibleName comes from the label's Text. In this case, if you remove the accessible name, AND change the tab order:
textBox1.TabIndex = 1;
label1.TabIndex=0;   

Accessible name on the textBox1 will be "UseSystemPasswordChar = true", which is appropriate because it matches what the sighted user reads, and is descriptive.

You can change tab order in the VS designer , go to View menu->Tab Order item.
image

Tanya-Solyanik (commented):
Looks good, I have a comment about accessible names, and additionally, please spell Password with a low case W in all file names and variable names, and text labels, this is a common spelling , see UseSystemPasswordChar property as an example.

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/Password.Designer.cs
```diff
@@ -0,0 +1,288 @@
...
+
+using System.Drawing;
+
+namespace WinformsControlsTest;
+
+partial class Password
+{
+    /// <summary>
+    /// Required designer variable.
+    /// </summary>
```
Epica3055, line 12:
```suggestion
    /// <summary>
    ///  Required designer variable.
    /// </summary>
```
And the below

Epica3055:
LGTM 👍

Epica3055 (approved):

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/WinformsControlsTest.csproj
```diff
@@ -136,6 +136,9 @@
 
   <ItemGroup>
     <Compile Update="MediaPlayer.cs" />
+    <Compile Update="Password.cs">
+      <SubType>Form</SubType>
+    </Compile>
```
LeafShi1, line 141: It seems that no other form files have been added here. Is this necessary?

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/Password.Designer.cs
```diff
@@ -0,0 +1,288 @@
...
+        label7 = new Label();
+        label8 = new Label();
+        label9 = new Label();
+        label10 = new Label();
+        label11 = new Label();
+        SuspendLayout();
+        // 
+        // textBox1
+        // 
+        textBox1.Location = new Point(26, 113);
```
Tanya-Solyanik, line 60: Thank you for removing `this.` modifiers!

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/WinformsControlsTest.csproj
Tanya-Solyanik: This is a perf optimization to open this file in designer in the VS. Normally we would keep it, however in this project we can't open Forms in designer view because we reference WInform project as project-to-project reference instead of using a `UseWinforms` property in the project file, so OOP designer does not recognize winforms assembly. Please remove it.

Tanya-Solyanik (approved):
Looks good, please address Leaf's comment and merge.

Tanya-Solyanik (approved):

Eudora-Li01:
Verified it in the main branch of Winforms repo for .Net 9.0, password test has been added to WinformsControlTest successfully.
image

PR #9845, opened 2023-09-05
Create IUnknownStrategy for GlobalInterfaceTable

JeremyKuhne:
Create a Global Interface Table IUnknownStrategy so we can use the same machinery the runtime uses to expose COM interfaces as objects with AgileComPointer-like support.

Change all of the direct Marshal calls to use a ComHelpers wrapper method.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9845)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Application.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Application.cs
@@ -41,10 +41,6 @@ public sealed partial class Application
     private static readonly object s_internalSyncObject = new();
     private static bool s_useWaitCursor;
 
-    private static bool s_useEverettThreadAffinity;
-    private static bool s_checkedThreadAffinity;
-    private const string EverettThreadAffinityValue = "EnableSystemEventsThreadAffinityCompatibility";
-
     /// <summary>
     ///  Events the user can hook into
     /// </summary>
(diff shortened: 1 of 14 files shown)


JeremyKuhne:
@jkoritzinsky here we're trying to leverage the stuff we talked about

src/System.Windows.Forms/src/System/Windows/Forms/Application.cs
```diff
@@ -143,43 +139,6 @@ public static RegistryKey CommonAppDataRegistry
     internal static string CommonAppDataRegistryKeyName
         => $"Software\\{CompanyName}\\{ProductName}\\{ProductVersion}";
 
-    internal static bool UseEverettThreadAffinity
```
lonitra, line 146: Why was this removed?

lonitra (approved):
Just a question, otherwise LGTM

src/System.Windows.Forms/src/System/Windows/Forms/Application.cs
JeremyKuhne: It is never called. :) Forgot to mention it.

jkoritzinsky (approved):
Love to see new usages of StrategyBasedComWrappers!

PR #9847, opened 2023-09-05
Convert AxHost to use Pointers

lonitra:
- Add `StorageType` enumto track what interface `_instance` is
- Remove `T.Interface` fields in `AxHost` and query interface on `_instance` based on`StorageType`
- Use `StorageType.PropertyBag` where appropriate. We never explicitly defined this before, but were essentially following the same pattern as the other storage cases
- Fix PropertyBag storage cases. This was likely never working due to this case always throwing in `AxHost.DepersistControl`
- Update methods in `AxHost.State` to use pointers instead of `T.Interface`


src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
```diff
@@ -2812,35 +2826,36 @@ private void DepersistControl()
             // Must init new:
             if (_instance is IPersistStreamInit.Interface init)
```
JeremyKuhne, line 2827: All of these `is` checks need to go away. You need to query for the desired interface pointer.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
```diff
@@ -137,10 +137,10 @@ public abstract unsafe partial class AxHost : Control, ISupportInitialize, ICust
 
     private object? _instance;
     private AgileComPointer<IOleInPlaceActiveObject>? _iOleInPlaceActiveObjectExternal;
-    private IPersistPropertyBag.Interface? _iPersistPropBag;
-    private IPersistStream.Interface? _iPersistStream;
-    private IPersistStreamInit.Interface? _iPersistStreamInit;
-    private IPersistStorage.Interface? _iPersistStorage;
+    private AgileComPointer<IPersistPropertyBag>? _iPersistPropBag;
```
JeremyKuhne, line 140: It would be better to get rid of the fields completely and just use the STG_ flags to identify which one we should re-query for when we need it.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
```diff
@@ -2768,27 +2784,24 @@ private void ActivateAxControl()
...
     {
-        using var pPropBag = ComHelpers.TryGetComScope<IPropertyBag>(propBag, out HRESULT hr);
-        hr.AssertSuccess();
-        _iPersistPropBag.Load(pPropBag, pErrorLog: null).ThrowOnFailure();
+        using var persistPropBag = _iPersistPropBag.GetInterface();
+        persistPropBag.Value->Load(propBag, pErrorLog: null).ThrowOnFailure();
     }
 
-    private void DepersistFromIStream(IStream.Interface istream)
+    private void DepersistFromIStream(IStream* istream)
```
JeremyKuhne, line 2793: These methods should probably be moved inline.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
```diff
@@ -2812,35 +2826,36 @@ private void DepersistControl()
...
+                _iPersistStorage = new(ComHelpers.GetComPointer<IPersistStorage>(persistStorage), takeOwnership: true);
                 using var storage = _ocxState.GetStorage();
-                _iPersistStorage.InitNew(storage).AssertSuccess();
-
+                using var persistStoragePtr = _iPersistStorage.GetInterface();
+                persistStoragePtr.Value->InitNew(storage).AssertSuccess();
                 return;
             }
 
             if (_instance is IPersistPropertyBag.Interface persistPropertyBag)
```
JeremyKuhne, line 2854: It still strikes me as super odd that this case will throw (that it doesn't return). Also that we don't have a `STG_PROPERTYBAG`.

JeremyKuhne (changes requested):
Added a few comments. We probably should convert `STG_*` into a `StorageType` enum and add something for `PropertyBag`.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
lonitra: Same here 🤔 had to re-check Framework to make sure it didn't accidentally get removed, but seems like it was always like this..

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.StorageType.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace System.Windows.Forms;
+
+public abstract partial class AxHost
+{
+    [Flags]
```
JeremyKuhne, line 8: Don't use the `[Flags]` if multiple won't be set at the same time. If these are actually flags you'll want to make sure the values don't overlap.

In either case, `Unknown` should be `0`.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
```diff
@@ -2768,79 +2766,43 @@ private void ActivateAxControl()
...
-
     private void DepersistControl()
     {
         FreezeEvents(true);
 
         if (_ocxState is null)
         {
             // Must init new:
-            if (_instance is IPersistStreamInit.Interface init)
+            if (ComHelpers.SupportsInterface<IPersistStreamInit>(_instance))
```
JeremyKuhne, line 2776: It is better to query here and not use `SupportsInterface` as otherwise you're querying twice. Probably worth adding a comment to the helper method to remind on this case.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
```diff
@@ -2768,79 +2766,43 @@ private void ActivateAxControl()
...
+            if (ComHelpers.SupportsInterface<IPersistStream>(_instance))
             {
-                _storageType = STG_STREAM;
-                _iPersistStream = persistStream;
+                _storageType = StorageType.Stream;
                 return;
             }
 
-            if (_instance is IPersistStorage.Interface persistStorage)
+            if (ComHelpers.SupportsInterface<IPersistStorage>(_instance))
```
JeremyKuhne, line 2790: Same comment as above, try to get the ComScope directly

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
```diff
@@ -2768,79 +2766,43 @@ private void ActivateAxControl()
...
                 using var storage = _ocxState.GetStorage();
-                _iPersistStorage.InitNew(storage).AssertSuccess();
-
+                using var persistStoragePtr = ComHelpers.GetComScope<IPersistStorage>(_instance);
+                persistStoragePtr.Value->InitNew(storage).AssertSuccess();
                 return;
             }
 
-            if (_instance is IPersistPropertyBag.Interface persistPropertyBag)
+            if (ComHelpers.SupportsInterface<IPersistPropertyBag>(_instance))
```
JeremyKuhne, line 2800: Same comment as above, try to get the ComScope directly

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
```diff
@@ -2850,67 +2812,67 @@ private void DepersistControl()
...
+                    _storageType = StorageType.Stream;
+                    using var persistStream = ComHelpers.GetComScope<IPersistStream>(_instance);
+                    persistStream.Value->Load(stream).AssertSuccess();
                 }
 
                 break;
-            case STG_STREAMINIT:
-                if (_instance is IPersistStreamInit.Interface persistStreamInit)
+            case StorageType.StreamInit:
+                if (ComHelpers.SupportsInterface<IPersistStreamInit>(_instance))
```
JeremyKuhne, line 2825: Same comment as above, try to get the ComScope directly

JeremyKuhne (changes requested):
A few more comments

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.StorageType.cs
lonitra: We have a public constructor in AxHost.State that takes an int for storage type https://github.com/dotnet/winforms/blob/01d50d1f4479d150234190718abfdda885e4293f/src/System.Windows.Forms/src/System/Windows/Forms/AxHost.State.cs#L63
Originally `Unknown = -1` so is it fine to make the change that `Unknown = 0` even though we have this public constructor?

JeremyKuhne: Ugh. It's safer to have `default` (i.e. `0`) be unknown. I'd just translate here by `+1`.

src/System.Windows.Forms/src/System/Windows/Forms/AxHost.cs
JeremyKuhne: This one still needs updated.

JeremyKuhne (approved):
Looks good!

PR #9864, opened 2023-09-07
reverted pinned .NET versions

### LINKED ISSUE #9851: Remove hardcoded .NET versions when Arcade updates SDK to 9.0
Until Arcade updates to a 9.0 SDK, you want to set these properties in your root Directory.Build.props file with a TODO to remove it when a 9.0 SDK is used: 

```
<NetMinimum>net8.0</NetMinimum>
<NetPrevious>net8.0</NetPrevious>
<NetCurrent>net8.0</NetCurrent>
```

This will drop the net6.0 and net7.0 assets from System.Drawing.Common (which is intentional) and doesn't yet use the net9.0 TFM (which would require some msbuild work). I think winforms in the past continued building the current TFM until an SDK is used that supports the newer TFM.

_Originally posted by @ViktorHofer in https://github.com/dotnet/winforms/issues/9836#issuecomment-1708860563_

Once Arcade is upgraded, we should revert [that change](https://github.com/dotnet/winforms/pull/9836/commits/3be1f719e3944b0a7b51a487f5c002f1e469e79a) and re-evaluate if we still need conditional compilation in System.Drawing.Common

Tanya-Solyanik:
Fixes #9851 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9864)


Tanya-Solyanik:
\azp run

dreddy-work:
you need to address references of those definitions.

Tanya-Solyanik:
>you need to address references of those definitions.

No, this is not on us, we are waiting for the arcade update.

Tanya-Solyanik:
\azp run

PR #9866, opened 2023-09-07
Fix folder browser code

### LINKED ISSUE #9859: An System.AccessViolationException appears when changing InitialDirectory property of OpenFileDialog control
### .NET version

.NET SDK 8.0.100-rc.2.23456.6

Not tested/verified

Yes, it cannot repro in .Net 7.0 SDK/.NET 6.0 SDK build.

### Issue description

After setting PropertyGrid's selectedObject to OpenFileDialog control and run the app, enter an absolute path to the D-disk for the InitialDirectory peoperty of the OpenFileDialog control,  an System.AccessViolationException pops up.

The issue also repro in FolderBrowserDialog/SaveFileDialog controls.

**.Net 8.0 result:**
[Exception Details.txt](https://github.com/dotnet/winforms/files/12546662/Exception.Details.txt)
image

repro_video

**.Net 7.0/.Net 6.0 result:**
There is no error dialog pop up and the InitialDirectory property of the OpenFileDialog control can be changed successfully.
image

Expected_video

### Steps to reproduce

Use following sample application:
[WinFormsApp16.zip](https://github.com/dotnet/winforms/files/12546745/WinFormsApp16.zip)

JeremyKuhne:
The old style folder browser was broken a few different ways:

- FolderBrowserDialog would mutate the managed result string, usually corrupting memory in terrible ways
- FolderNameEditor would AV
- All were leaking shell handles

This adds a centralized helper to avoid these problems. This also:

- Moves the exception string to primitives
- Uses CSIDL defines and clarifies usage
- Unblocks other CSIDL values
- Uses the updated API to get the path from the PIDL

We can't use the shared dialog in FolderNameEditor as we expose dialog styles there that are not exposed in FolderBrowserDialog.

Fixes #9859

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/FolderNameEditor.FolderBrowser.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/FolderNameEditor.FolderBrowser.cs
@@ -1,9 +1,7 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-using System.Buffers;
 using System.ComponentModel;
-using Windows.Win32.UI.Shell.Common;
 
 namespace System.Windows.Forms.Design;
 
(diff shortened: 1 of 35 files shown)


JeremyKuhne:
Manually validated

JeremyKuhne:
I'm going to cherry pick this back to main as there are conflicts.

lonitra (approved):
👍 LGTM

Tanya-Solyanik:
Where was the AV?

JeremyKuhne:
> Where was the AV?

@Tanya-Solyanik Yes, there. We were also pinning `string.Empty` and writing to it in the other dialog. 😲

Tanya-Solyanik (approved):

PR #9867, opened 2023-09-08
Fix issue #9807 Review AFTER PR ##11655

### LINKED ISSUE #9807: Refactor VB's Application Framework Network Class
The class [`Microsoft.VisualBasic.Devices.Network`](https://github.com/dotnet/winforms/blob/e47bebd2e1c8ce466f16ab0d40d0d7d60d5dcffa/src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/Devices/Network.vb#L37), which is part of the Visual Basic Application Framework, is based on the obsoleted [`WebClient`](https://learn.microsoft.com/en-us/dotnet/api/system.net.webclient?view=net-7.0) class.

We need to modernize this part of the Application Framework to avoid breaking changes, should the `WebClient` class no longer be included in future versions of .NET.

This should be done in the .NET 9-time frame.

This also addresses https://github.com/dotnet/winforms/issues/3936.

Here is a potential general approach, which uses `HttpClient` instead of `WebClient`.
(!not tested, and my last Web-dev-experiences is quite a while, so, please double-check me here!)

```VB
Imports System.IO
Imports System.Net
Imports System.Net.Http

Public Class Network

    Public Async Function DownloadFileWithCredentialsAndProgressAsAStartingPointAsync(
        url As String,
        destinationPath As String,
        credentials As NetworkCredential,
        progress As IProgress(Of Double)) As Task

        Dim handler As New HttpClientHandler() With
        {
            .Credentials = credentials
        }

        Using httpClient As New HttpClient(handler)

            Dim response = Await httpClient.GetAsync(url, HttpCompletionOption.ResponseHeadersRead)
            Dim contentLength = response.Content.Headers.ContentLength

            Using responseStream As Stream = Await response.Content.ReadAsStreamAsync()
                Using fileStream As New FileStream(destinationPath, FileMode.Create, FileAccess.Write, FileShare.None)

                    Dim buffer(8191) As Byte
                    Dim totalBytesRead As Long = 0
                    Dim bytesRead As Integer

                    While (Await responseStream.ReadAsync(buffer, 0, buffer.Length)) > 0
                        Await fileStream.WriteAsync(buffer, 0, bytesRead)
                        totalBytesRead += bytesRead

                        If contentLength.HasValue Then
                            Dim percentage As Double = (CDbl(totalBytesRead) / CDbl(contentLength.Value)) * 100
                            progress.Report(percentage)
                        End If

                    End While
                End Using
            End Using
        End Using
    End Function
End Class
```

paul1956:
Fixes #9807

## Proposed changes

- Add Remove dependance on WebClient for VB FileDownload by using HttpClient
- Add a Visual Basic Scratch Project so that Microsoft.VisualBasic.Forms dependent code can be tested from a Visual Basic project (even an empty WinForms project written in VB does not work)
-  Upload still uses WebClient that also needs to be done as a separate PR

## Customer Impact
- Adds Async file download support to VB, and if WebClient goes away this will allow existing Visual Basic programs using synchronous File Download to still work.

## Regression? 

- Yes when WebClient if and when goes away, no today 

## Risk

- The code as is completely untested as I can't test without a working Scratch Project.
- The existing code uses ICredentials in some of the paths, it is unclear how compatible that is between WebClient and HttpClient 

## Test methodology
-  EXISTING CODE IS UNTESTED, and don't know if any tests currently exist for FileDownload.
- I will write tests if none exist.

## Test environment(s) 
- TBD

- Visual Basic code to replace deprecated WebClient, before it is removed from .Net

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9867)


src/System.Windows.Forms/tests/IntegrationTests/ScratchProjectVB/My Project/Application.Designer.vb
```diff
@@ -0,0 +1,37 @@
...
+            Me.IsSingleInstance = False
+            Me.EnableVisualStyles = True
+            Me.SaveMySettingsOnExit = True
+            Me.ShutDownStyle = Global.Microsoft.VisualBasic.ApplicationServices.ShutdownMode.AfterMainFormCloses
+        End Sub
+
+        <Global.System.Diagnostics.DebuggerStepThroughAttribute()>
+        Protected Overrides Sub OnCreateMainForm()
+            Me.MainForm = Form1
+        End Sub
```
elachlan, line 35: This is causing the failure.

@KlausLoeffelmann

paul1956: @elachlan that comes with new WinForms project, I have nothing to with the Application Designer.vb file it is autogenerated and the way VB Programs start.

elachlan: That is what is causing the build failure. Unsure on how to fix it.

paul1956:
@elachlan I have removed the Scratch Project and move all the tests into a new Network tests file to unit/Microsoft.VisualBasic.Forms.Tests/VisualBasic.Tests. under Forms it turns out there are no tests for upload or download just a comment.

I commented out all the test code because Test Projects don't have access to My.Computer.Network. All the tests would pass except for the one using the UI and the errors returned are the same as the current functions. The UI tests fail in a real VB project. I also made DownloadFileAsync Private so this adds no new API services. There is nothing else I can do without at this time. To make DownloadFileAsync Public the ProgressBar would need to be addressed

src/Microsoft.VisualBasic.Forms/tests/UnitTests/System/Windows/Forms/NetworkTexts.vb
```diff
@@ -0,0 +1,147 @@
...
+        '        Do While Directory.Exists(folder) OrElse File.Exists(folder)
+        '            folder = Path.Combine(_baseTempPath, $"Test{i}")
+        '            i += 1
+        '        Loop
+        '    End SyncLock
+        '    Return folder
+        'End Function
+
+        '<WinFormsFact>
+        'Public Sub FileDownloadWithJustUriAndDestinationFileName_method()
```
elachlan, line 50: Instead of commenting these out, just mark them with skip. Winforms team can probably advise if there is something better.

```csharp
[WinFormsFact(Skip = "Application class not available in CI, see: https://github.com/dotnet/winforms/issues/9807")]
```

KlausLoeffelmann:
> I commented out all the test code because Test Projects don't have access to My.Computer.Network...

Why is that?
I am just on my phone right now, but I'd say, there should be a way to check-in the additional generated classes which get created on writing the application.myapp file?

KlausLoeffelmann:
That sounds to me that the tests don't reflect the settings for the Application Framework _enabled_.

> Sorry that is not enough information for me to know how to fix this issue...

Have you guys read this (especially the paragraphed I tagged:)?

https://devblogs.microsoft.com/dotnet/update-to-winforms-vb-appframework/#a-look-behind-the-scenes-of-the-winforms-vb-application-framework

It's quite probable that it'll help in the context.
I don't have the time to do it right now, but we should find a way to include the Application Framework generated files in the unit tests. So, again, without really having dig into the whole topic, maybe checking in the code generated part helps here.

src/Microsoft.VisualBasic.Forms/tests/UnitTests/System/Windows/Forms/NetworkTexts.vb
paul1956: No tests now fail on desktop and CI. The one that uses ReportDialog is skipped using the line above.

paul1956:
@KlausLoeffelmann Figured something out, adding the 3 lines below make the tests compile.
```
<MyType>WindowsFormsWithCustomSubMain</MyType>
<UseApplicationFramework>True</UseApplicationFramework>
<OutputType>Library</OutputType>
```
All Tests pass in local computer and all pass on CI with 1 skipped.
Exceptions are identical to original code.

KlausLoeffelmann:
Nice!

It'll take a bit, until I am able to dig deeper into it, but I have it on my radar.
We won't take this for .NET 8, since it's to late, but I think it's also not that urgent, so .NET 9 is fine. Please remember, we also need the sync version for backwards compat and the except same overloads.

And since we're introducing async versions, we need a list of the new APIs and a rationale. The rationale needs to point out, that the primary motivation is not to introduce async overloads, but replacing an obsolete usage of an internal API.

And please don't rush it, it's taking some time to get this in, anyway!

paul1956:
@KlausLoeffelmann The Async Overloads are private. I added no new public functions (there is 1 new SR message because I didn't know how to access the existing one). Making them public would require more work as ProgressDialog is also not Public. The implementation of the existing API's is Async they call the private DownloadFileAsync functions, there is just an Await at the end.

elachlan:
Some tests are failing. Very close though.

paul1956:
> Some tests are failing. Very close though.

They were not so something changed. I will look but this really should wait until VB-Code-Cleanup PR to avoid duplicate work.

paul1956:
@elachlan I have changed nothing except merging with main and now the tests passed.

elachlan:
Probably a transient test issue. It happens.

Tanya-Solyanik:
Assert.Throws() Failure: No exception was thrown\r\nExpected: typeof(System.Net.WebException)



Stack trace
   at Microsoft.VisualBasic.Forms.Tests.Microsoft.VisualBasic.Forms.Tests.NetworkTexts.SimpleFileDownloadWithExpectedTimeOut_method() in /_/src/Microsoft.VisualBasic.Forms/tests/UnitTests/System/Windows/Forms/NetworkTexts.vb:line 100
   at System.Reflection.MethodBaseInvoker.InterpretedInvoke_Method(Object obj, IntPtr* args)
   at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)

.editorconfig
```diff
@@ -99,6 +95,12 @@ dotnet_style_prefer_collection_expression = true:suggestion
 dotnet_diagnostic.RS0041.severity = none
 dotnet_diagnostic.IDE0005.severity = error
 
+# Expression-level preferences
```
Tanya-Solyanik, line 98: Is this line preserved when saving in the VS editor?

.editorconfig
```diff
@@ -194,6 +196,16 @@ dotnet_diagnostic.SYSLIB1054.severity = none
...
 
+# IDE0079: Remove unnecessary suppression
+dotnet_diagnostic.IDE0079.severity = none
+
+# Expression-level preferences
+dotnet_style_coalesce_expression = false:suggestion
+dotnet_style_prefer_simplified_boolean_expressions = true:suggestion
+dotnet_style_coalesce_expression = false:suggestion
+dotnet_style_prefer_conditional_expression_over_assignment = false:silent
+dotnet_style_prefer_conditional_expression_over_return = false:silent
```
Tanya-Solyanik, line 207: Why did you repeat 4 rules that are already listed under C#+VB section?

src/Microsoft.VisualBasic.Forms/tests/UnitTests/System/Windows/Forms/NetworkTexts.vb
```diff
@@ -0,0 +1,156 @@
...
+' The .NET Foundation licenses this file to you under the MIT license.
+
+Imports System.IO
+Imports System.Net
+
+Imports Xunit
+
+Namespace Microsoft.VisualBasic.Forms.Tests
+
+    Partial Public Class NetworkTexts
```
Tanya-Solyanik, line 11: Did you mean `NetworkTests` as the class and file name?

.editorconfig
paul1956: They should be.

src/Microsoft.VisualBasic.Forms/tests/UnitTests/System/Windows/Forms/NetworkTexts.vb
paul1956: Fixed

.editorconfig
paul1956: Because C# and VB need different options, so I deleted them from common section and preserved C# values and changed VB values. To do this you need to duplicate the sections under C# and #VB so these 4 times appear twice with different options. Lines 204-207 are the VB Versions

paul1956:
Ping, is there an ETA when someone might review this?

KlausLoeffelmann:
No, not immediately. I just finalized a big VB change for CodeGen in the Designer, so VB won't get another higher prio tasks this or next week.

paul1956:
@KlausLoeffelmann old comment. I wrote tests for all the non interactive code, there are no tests specific to the interactive code though I have tested the interactive code, in a separate project on my machine. To test the interactive code inside this solution I would need a VB scratch project like exists for the C# code and creating one is not obvious.

paul1956:
@KlausLoeffelmann an update on testing, functions that require passwords were not tested by me or any of the tests I added. I would need a server that supports username and password that I could download the test files from and a way to do it without exposing the server to attack. Also, interactive functions were tested using separate project. I don't know how to write interactive tests. None of this if different than what was here before the file Network Tests is new. I have not exposed any new API's, DownloadFileAsync is private and could be made public but that requires API review and exposing ProgressDialog and a decision on how many of the options to make public.

elachlan:
@Olina-Zhang and team might be able to help test those.

I'd say we should land it as private and do an API Review afterwards to move it to public.

Olina-Zhang:
@paul1956 I still see you have some updates in this PR, please let us know if the PR is ready for testing. It would be great if you could help us provide some testing direction about your changing.

paul1956:
@Olina-Zhang I have been adding tests to improve coverage. VB's overall coverage is very low. There are two tests I just added that run locally and then had them skipped (<WinFormsFact(Skip:=) that improve coverage. From what I understand the skipped tests will not run on server (I don't know if this is true or how to fix if it is). If you could get them to run on server, I would be happy to add more tests to PR and do separate PR(s) to improve coverage to VB in general. At this point the only thing I am doing with my 2 PR's is keeping them updated with latest "Main". I also don't understand why there are test failures all tests pass locally,

src/Microsoft.VisualBasic.Forms/tests/UnitTests/System/Windows/Forms/NetworkTests.vb
```diff
@@ -0,0 +1,286 @@
...
+                                                                                      False,
+                                                                                      connectionTimeout:=100000,
+                                                                                      overwrite:=False)
+                                              Catch ex As Exception
+                                                  Throw
+                                              Finally
+                                                  CleanupTempDirectory(tmpFilePath)
+                                              End Try
+                                          End Sub
+                                             )
```
KlausLoeffelmann, line 237: I know I said, let's see what Roslyn does, but I don't think this is feasible.
Let's rather go for the C# rules, and aim for one level indentation, where it makes sense. Where it doesn't make sense, let's discuss, what we should do in VB alternatively.

src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/ApplicationServices/SingleInstanceHelpers.vb
```diff
@@ -66,7 +65,7 @@ Namespace Microsoft.VisualBasic.ApplicationServices
                     If bytesRead = 0 Then
                         Exit While
                     End If
-                    stream.Write(buffer, 0, bytesRead)
+                    Await stream.WriteAsync(buffer.AsMemory(0, bytesRead), cancellationToken).ConfigureAwait(False)
```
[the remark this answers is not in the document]
paul1956: @KlausLoeffelmann 
Because there is a code rule that recommends that change, I made it. The originally code was very old.

KlausLoeffelmann:
Oh, and one more thing:
As I said, we should have different PRs for addressing the formatting stuff and the .editorConfig, and then the download issue.

It's maybe good anyway to consolidate almost 100 commits in a meaningful way. For the future, when we have big changes like this, let's aim for rebasing the branch regularly, rather than merging them. Should not be a problem for the VB part, since we're pretty much the only people doing changes here, anyway! 😉

src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/ApplicationServices/WindowsFormsApplicationBase.vb
```diff
@@ -499,12 +504,13 @@ Namespace Microsoft.VisualBasic.ApplicationServices
...
             '    the ApplyDefaults event.
-            Dim applicationDefaultsEventArgs = New ApplyApplicationDefaultsEventArgs(
-                MinimumSplashScreenDisplayTime,
-                HighDpiMode)
 
             ' Overriding MinimumSplashScreenDisplayTime needs still to keep working!
-            applicationDefaultsEventArgs.MinimumSplashScreenDisplayTime = MinimumSplashScreenDisplayTime
+            Dim applicationDefaultsEventArgs = New ApplyApplicationDefaultsEventArgs(
+                    MinimumSplashScreenDisplayTime,
+                    HighDpiMode) With {
```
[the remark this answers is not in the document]
paul1956: @KlausLoeffelmann  The changes were again based on rules. If you want to see all the rule based changes it is better to review my CodeCleanup PR, it has no logic changes just formatting to get rid of all coding issues. A few appear in this PR. My recommendations was to review and merge that PR first to avoid a lot of duplicate effore.

src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/Devices/Audio.vb
```diff
@@ -3,7 +3,9 @@
 
 Option Explicit On
 Option Strict On
+
 Imports System.IO
+
```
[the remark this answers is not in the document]
paul1956: The LF's are added automatically between import groups System vs Microsoft.

paul1956:
> Hey Paul,
> 
> I spent some time with that, and I still have a series of issues with this change, especially around the sync over async part, and also, that we still don't seem to have a comprehensive test app?
> 
> So, one general issue: We need to separate the formatting issues and the actual Download changes. The amount of changes we need for both are just too many, so we cannot mix them.
> 
> It's up to you, what we want to address in what order, also, again, since I will be somewhat slow in following up due to current workload (spending part of my holiday on this, currently).
> 
> I very much appreciate your work, so, please don't be demotivated by my responses - we need to ensure some consistent standard for VB and have not really figured it out - and my somewhat slow reacting time. I still very much like to see this happening in .NET 9!

@KlausLoeffelmann 
I have a separate PR for just the formatting changes "VB-Code-Cleanup" it should be reviewed and merged first..

src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/Devices/Network.vb
```diff
@@ -129,7 +130,7 @@ Namespace Microsoft.VisualBasic.Devices
             ' We're safe from Ping(Nothing, ...) due to overload failure (Ping(String,...) vs. Ping(Uri,...)).
             ' However, it is good practice to verify address before calling address.Host.
             If address Is Nothing Then
-                Throw ExceptionUtils.GetArgumentNullException("address")
+                Throw ExUtils.GetArgumentNullException("address")
```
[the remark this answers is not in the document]
paul1956: It is an Alias to shorten the lines, it is used a lot.

Imports ExUtils = Microsoft.VisualBasic.CompilerServices.ExceptionUtils

elachlan:
See: #10117

src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/Devices/Network.vb
```diff
@@ -165,18 +163,280 @@ Namespace Microsoft.VisualBasic.Devices
...
         ''' <param name="address">Address to the remote file, http, ftp etc...</param>
         ''' <param name="destinationFileName">Name and path of file where download is saved</param>
         Public Sub DownloadFile(address As String, destinationFileName As String)
-            DownloadFile(address, destinationFileName, DEFAULT_USERNAME, DEFAULT_PASSWORD, False, DEFAULT_TIMEOUT, False)
+            Try
+                DownloadFileAsync(address, destinationFileName, DEFAULT_USERNAME,
+                              DEFAULT_PASSWORD, Nothing,
+                              DEFAULT_TIMEOUT,
+                              False
+                             ).Wait()
```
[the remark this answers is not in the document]
paul1956: @KlausLoeffelmann 
I use the pattern below when the code has to run Async. Not sure why you think the other could deadlock. If it throws it is caught and correct response is raised.
```
                Dim t As Task = DownloadFileAsync(address,
                                                  destinationFileName,
                                                  userName,
                                                  password,
                                                  dialog, connectionTimeout, overwrite,
                                              UICancelOption.ThrowException
                                             )
                If t.IsFaulted Then
                    Throw t.Exception
                Else
                    dialog?.ShowProgressDialog()
                    t.Wait()
                End If
            Catch ex As Exception
                If ex.InnerException IsNot Nothing Then
                    Throw ex.InnerException
                End If
                Throw
```
VB does not like the VB equivalent of the C# code you wrote.
image

src/Microsoft.VisualBasic.Forms/tests/UnitTests/System/Windows/Forms/NetworkTests.vb
paul1956: @KlausLoeffelmann C# handles formatting with bracket rules, VB has no enforced rules. I think Sam was interested in doing something for VB but I think it got dropped. Better to handle this in #10117 first than adopt here.

src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/Devices/Network.vb
```diff
@@ -218,17 +506,47 @@ Namespace Microsoft.VisualBasic.Devices
...
+                                                  password,
+                                                  dialog, connectionTimeout, overwrite,
+                                              UICancelOption.ThrowException
+                                             )
+                If t.IsFaulted Then
+                    Throw t.Exception
+                Else
+                    dialog?.ShowProgressDialog()
+                    t.Wait()
+                End If
```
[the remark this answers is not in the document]
paul1956: @KlausLoeffelmann 
I believe this emulates the original code; I am not trying to make the I/O Sync that is a different but useful PR.
I agree I need to handle t.IsFaulted TWICE. The first IsFaulted test will fail when any of the parameters are bad,  you can't create the ProgressDialog if the true Async code does not start it will deadlock.   
```  
If t.IsFaulted Then ' This will be true if any parameters are bad
     Throw t.Exception
Else
      dialog?.ShowProgressDialog()
      t.Wait()
      If t.IsFaulted Then
          Throw t.Exception
     End If
End If
```

paul1956:
> Oh, and one more thing: As I said, we should have different PRs for addressing the formatting stuff and the .editorConfig, and then the download issue.
> 
> It's maybe good anyway to consolidate almost 100 commits in a meaningful way. For the future, when we have big changes like this, let's aim for rebasing the branch regularly, rather than merging them. Should not be a problem for the VB part, since we're pretty much the only people doing changes here, anyway! 😉

That would be a lot less work, I have never rebased so will need help there.

paul1956:
@KlausLoeffelmann what is "that we still don't seem to have a comprehensive test app". VB Test coverage is very low. I added a Network Test (winforms\src\Microsoft.VisualBasic.Forms\tests\UnitTests\System\Windows\Forms\NetworkTests.vb) which has reasonable coverage if you include the functions that don't run on server (I need help with getting them working), I don't believe the 2 functions using ProgressDialog they will compile on the server.

src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/Devices/Network.vb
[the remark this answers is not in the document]
KlausLoeffelmann: What you wrote is not the equivalent.
From the line...

```
Task.Run(async () => await SomeAsyncMethod()).GetAwaiter().GetResult();
```

...which by the way could be even further shortened to...

```
Task.Run(SomeAsyncMethod).GetAwaiter().GetResult();
```

...this is what the compiler makes out of that:

```
public void YourSyncMethod()
    {
        Task.Run(new Func<Task>(SomeAsyncMethod)).GetAwaiter().GetResult();
    }
```

Which means this would be the equivalent:

image

Concerning `Task.Wait`: We do not allow this in WinForms Apps. (And actually, I would not use it anywhere.) Here is the reason:

WinForms applications have a synchronization context that ensures all UI operations are executed on the main thread. When you use `Task.Wait` (or `Task.Result`), it blocks the calling thread until the task is completed. If this task is waiting for an operation that needs to execute on the main thread (like updating the UI), a deadlock occurs because the task is waiting for the main thread to be free, while the main thread is blocked waiting for the task to complete. So, to avoid that we run in likely problems here, we are not allowing `Task.Wait`. This needs to be refactored.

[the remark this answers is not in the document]
paul1956: @KlausLoeffelmann The above does not work
image

Change to Async Sub is better but still an issue, no idea how to address the issue
image

The VB recommendations (https://stackoverflow.com/questions/42009868/how-should-task-run-call-an-async-method-in-vb-net) seem to say don't use GetResult

paul1956:
@KlausLoeffelmann I moved the test to #10117 and ran them against the old code. It was useful to make sure that all errors match (they didn't).  More importantly if you run the original ProjectDialog tests they are run serially, and you can't actually cancel a long download. The new code (without Task.Run) works correctly. When I tried to convert using your example, most of the tests actually fail.

Olina-Zhang:
> @Olina-Zhang I have been adding tests to improve coverage. VB's overall coverage is very low. There are two tests I just added that run locally and then had them skipped (<WinFormsFact(Skip:=) that improve coverage. From what I understand the skipped tests will not run on server (I don't know if this is true or how to fix if it is). If you could get them to run on server, I would be happy to add more tests to PR and do separate PR(s) to improve coverage to VB in general. At this point the only thing I am doing with my 2 PR's is keeping them updated with latest "Main". I also don't understand why there are test failures all tests pass locally, but I just changed failing test to use very larger download file from Web.

Sorry, I'm not sure about all this, haven't done any research on it yet. But leave me a message for anything manual testing or verification related!

merriemcgaw:
Hi @paul1956 thank you so much for working on this! I think our best course of action at the moment is to focus on #10117 and get that in first. Then we can pick this back up after rebasing your branch. At this point it makes much more sense to focus on the cleanup and not worry about this one as it is today.

paul1956:
@merriemcgaw I completely agree. I have added the tests from this PR to that one to improve coverage. There were 0 networking tests and adding them was useful to get a baseline the existing network code is synchronize and does block UI. I could remove them and do a separate PR it's just 1 file. Up to you.

merriemcgaw:
> . There were 0 networking tests and adding them was useful to get a baseline the existing network code is synchronize and does block UI. I could remove them and do a separate PR it's just 1 file. Up to you.

I'll let @KlausLoeffelmann and @JeremyKuhne make the call on that one.

paul1956:
@KlausLoeffelmann @JeremyKuhne I did a new PR #10684, it does not change any existing code. It just uses NetworkTest from other this PR and Adds Tests for other device subsystems as I get around to them. So far, I have added "Clock" as a test to get 100% code coverage.

paul1956:
@KlausLoeffelmann I have looked at calling the internal DownLoadFileAsync using Task.Run instead of Await and in every attempt the tests against Framework fail. I believe the issue is that DownloadFile in Framework does block the UI.

I have implemented the full set of DownloadFileAsync except 1 function (currently commented out) and with an API review they could be exposed, and new tests written that would expect Async operation that don't block the UI. I have also simplified the file by removing all the code that hasn't changed (Upload, Ping...) to separate files and moved all of what was Network,vb to a Sub directory using "rtial Public Class Network" for all the parts so no API changes.

paul1956:
> I still see you have some updates in this PR, please let us know if the PR is ready for testing. It would be great if you could help us provide some testing direction about your changing.
@Olina-Zhang I have written all the tests for code directly changed in this PR as a separate Pr, providing 100% coverage that should be reviewed and Merged first, as it had no code changes just tests for the original code. I would like to write additional tests for the VB code (for stuff not directly changed) but I am waiting for someone to enable VisibleToTest attribute on VB code, my attempts have all resulted in ambiguous class definition for "Utils" but I don't see any so I guess I am doing the attribute wrong.

There are several sections of code in ExceptionUtilities that can't be reached nested 3 levels deep and are overly complex to do a trivial function have removed all of it in the PR but want test to prove 100% coverage after the change, I don't think it's possible with current code as most of it is unreachable.

elachlan:
@paul1956 can you please update the title? also is it possible to land the testing changes first in a PR and then merge the changes? or are they interconnected too much?

paul1956:
> @paul1956 can you please update the title? also is it possible to land the testing changes first in a PR and then merge the changes? or are they interconnected too much?

@elachlan Testing is part of the VB CodeCleanup that is how I found the issues. I added a lot of tests but the review got tripped up with formatting issues so PR #11215 is cleanup with tests. The SR entries aren’t accessible in that PR so they are hardcoded strings but the PR #9867 has SR entries I added (probably incorrectly).

paul1956:
@elachlan can I update title without breaking the connection between Visual Studio and GitHub?

src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/ApplicationServices/ApplyApplicationDefaultsEventArgs.vb
```diff
@@ -13,7 +9,7 @@ Imports System.Windows.Forms
 Namespace Microsoft.VisualBasic.ApplicationServices
 
     ''' <summary>
-    ''' Provides context for the ApplyApplicationDefaults event.
+    '''  Provides context for the ApplyApplicationDefaults event.
```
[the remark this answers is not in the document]
lonitra: Ah sorry, I realize that they are already in that PR

src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/MyServices/Internal/HttpClientCopy.vb
```diff
@@ -0,0 +1,201 @@
...
+
+                        If m_ProgressDialog IsNot Nothing Then
+                            RemoveHandler m_ProgressDialog.UserHitCancel, AddressOf m_ProgressDialog_UserHitCancel
+                        End If
+                    End If
+
+                Case HttpStatusCode.Unauthorized
+                    Throw New WebException(SR.net_webstatus_Unauthorized, WebExceptionStatus.ProtocolError)
+                Case Else
+                    Throw New WebException()
```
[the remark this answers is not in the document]
paul1956: @lonitra Not sure what the alternative is, existing code only looks for 2 exceptions timeout and Unauthorized.  The else clause could also throw Unreachable but existing code just passes through the WebException.
```
        Private Sub m_WebClient_DownloadFileCompleted(sender As Object, e As System.ComponentModel.AsyncCompletedEventArgs) Handles m_WebClient.DownloadFileCompleted
            Try
                ' If the download was interrupted by an exception, keep track of the exception, which we'll throw from the main thread
                If e.Error IsNot Nothing Then
                    _exceptionEncounteredDuringFileTransfer = e.Error
                End If

                If Not e.Cancelled AndAlso e.Error Is Nothing Then
                    InvokeIncrement(100)
                End If
            Finally
                'We don't close the dialog until we receive the WebClient.DownloadFileCompleted event
                CloseProgressDialog()
            End Try
        End Sub
```

paul1956:
> @paul1956 can you please update the title? also is it possible to land the testing changes first in a PR and then merge the changes? or are they interconnected too much?

I updated title to point to correct dependent PR and updated title of PR #11215 to
"Code cleanup only and Add VB Tests #11215"

src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/ApplicationServices/ApplyApplicationDefaultsEventArgs.vb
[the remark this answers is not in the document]
paul1956: @lonitra I did some merging with #11215 before it because too difficult as files are deleted and added and moved around with this PR. Since I am not great with GitHub my plan was to manually merge #11215 after it is merged into Main. Reviewing this before #11215 will be a lot of wasted effort.

src/System.Windows.Forms/tests/IntegrationTests/ScratchProjectVB/My Project/Application.Designer.vb
paul1956: Issue fixed

paul1956:
Replaces #11867, all PR Feedback is addressed and now targets feature/10.0

PR #9872, opened 2023-09-08
Enable nullability in some Binding members

gpetrou:
## Proposed changes

- Enable nullability in some Binding members.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9872)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/BindingContext.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/BindingContext.cs
@@ -348,7 +348,7 @@ public partial class BindingContext : ICollection
     {
         ArgumentNullException.ThrowIfNull(binding);
 
-        BindingManagerBase oldManager = binding.BindingManagerBase;
+        BindingManagerBase? oldManager = binding.BindingManagerBase;
         oldManager?.Bindings.Remove(binding);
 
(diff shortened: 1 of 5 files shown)


SimonZhao888 (approved):

src/System.Windows.Forms/src/System/Windows/Forms/BindingContext.cs
```diff
@@ -360,7 +360,7 @@ public static void UpdateBinding(BindingContext? newBindingContext, Binding bind
                 CheckPropertyBindingCycles(newBindingContext, binding);
             }
 
-            BindingManagerBase newManager = newBindingContext.EnsureListManager(binding.DataSource, binding.BindingMemberInfo.BindingPath);
+            BindingManagerBase newManager = newBindingContext.EnsureListManager(binding.DataSource!, binding.BindingMemberInfo.BindingPath);
```
LeafShi1, line 363: How to confirm that binding.DataSource is not null?

gpetrou: If this is null, it will eventually throw in `HashKey`. So, we keep the current behavior.

PR #9875, opened 2023-09-09
Migrating `ComCtl32` interop to CsWin32

elachlan:
Related to: #7445

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9875)


elachlan:
@lonitra I might need a hand with some of the code. I think I have made a pretty good start.

elachlan:
I think I fixed the issues I had. The final compile issue is waiting on:
https://github.com/microsoft/win32metadata/pull/1703

src/System.Windows.Forms/src/System/Windows/Forms/TreeNode.cs
```diff
@@ -26,8 +26,8 @@ namespace System.Windows.Forms;
 public partial class TreeNode : MarshalByRefObject, ICloneable, ISerializable
```
lonitra, line 26: I believe we may want to implement `IHandle<HTREEITEM>` to this similar to our implementation of `IHandle<HWND>` in Control.cs to follow existing patterns.
https://github.com/dotnet/winforms/blob/main/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs#L13591-L13593

Perhaps also add a `HTREEITEMInternal` property for when we just want a handle without forcing creation and then using that instead of casting everywhere.

lonitra (commented):
Nice work on this! Things LRGTM, just a few comments :)

src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/PCWSTR.cs
```diff
@@ -0,0 +1,46 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Runtime.InteropServices;
+
+namespace Windows.Win32.Foundation;
+
+internal readonly unsafe partial struct PCWSTR
+{
+    public string? ToStringAndCoTaskMemFree()
```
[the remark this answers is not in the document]
elachlan: I needed `IsNull`, I figured the rest would be useful. But I don't think I got around to using it.

src/System.Windows.Forms/src/System/Windows/Forms/ListView.cs
```diff
@@ -2904,7 +2906,7 @@ private unsafe void CustomDraw(ref Message m)
                             int mask = 0xFF0000;
                             do
                             {
-                                int C = nmcd->clrText & mask;
+                                int C = (int)(nmcd->clrText & mask);
```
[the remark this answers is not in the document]
elachlan: renamed to `color`

src/System.Windows.Forms/src/System/Windows/Forms/TreeNode.cs
```diff
@@ -2195,13 +2166,16 @@ internal unsafe void UpdateImage()
...
-            hItem = Handle,
+            hItem = HTREEITEMInternal,
             iImage = Math.Max(0, ((ImageIndexer.ActualIndex >= tv.ImageList.Images.Count) ? tv.ImageList.Images.Count - 1 : ImageIndexer.ActualIndex))
         };
 
         PInvoke.SendMessage(tv, PInvoke.TVM_SETITEMW, 0, ref item);
     }
 
+    HTREEITEM IHandle<HTREEITEM>.Handle => HTREEITEMInternal;
+    internal HTREEITEM HTREEITEMInternal => (HTREEITEM)_handle;
```
lonitra, line 2177: Digging through the code again, I don't see anywhere we are doing P/Invoke calls that explicitly need HTREEITEM, so I am not sure that implementing `IHandle<HTREEITEM>` is totally necessary. Let's remove the implementation for now and we can easily add it if we see it is needed later, my apologies for that. 🙏 
We should still add some internal properties for getting the HTREEITEM though, so this is close to what we want, but as it is right now, this would be changing the behavior since we are no longer calling `Handle`, which would force handle creation if it hasn't already been created. I think we should have something like the following:
```suggestion
    internal HTREEITEM HTREEITEM => (HTREEITEM)Handle
    internal HTREEITEM HTREEITEMInternal => (HTREEITEM)_handle;
```

then anywhere that was originally calling `Handle` can be replaced with `HTREEITEM` (I believe I also saw a few in TreeNodeTest.cs) while anywhere that was originally calling `_handle` can be replaced with `HTREEITEMInternal`. You may be able to actually get rid of `_handle` entirely and just have `HTREEITEMInternal` serve as an auto prop for it, but I am not entirely sure by just looking 🤔

src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/PCWSTR.cs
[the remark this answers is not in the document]
lonitra: Gotcha, let's remove the methods that aren't in use and we can add it later as needed

elachlan:
I ran the test locally 1000 times and didn't have a failure. So someone might want to look at it?

lonitra:
We are getting the following exception. Didn't have time to dig further on what was going on
```
System.ArgumentOutOfRangeException : Specified argument was out of the range of valid values.

Stack trace
   at Windows.Win32.UI.Controls.LVITEMW.UpdateText(ReadOnlySpan`1 text) in /_/src/System.Windows.Forms.Primitives/src/Windows/Win32/UI/Controls/LVITEMW.cs:line 25
   at System.Windows.Forms.Tests.InteropTests.LVITEMWTests.UpdateText_should_limit_input_text_to_cchTextMax_less_1_text_longer(String originalText, Int32 maxLength, String newText, String expected) in /_/src/System.Windows.Forms.Primitives/tests/UnitTests/Interop/ComCtl32/LVITEMWTests.cs:line 34
   at InvokeStub_LVITEMWTests.UpdateText_should_limit_input_text_to_cchTextMax_less_1_text_longer(Object, Span`1)
   at System.Reflection.MethodBaseInvoker.InvokeWithFewArgs(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   ```

File: src/System.Windows.Forms.Primitives/src/Windows/Win32/UI/Controls/LVITEMW.cs [hunk 5]
```diff
@@ -0,0 +1,29 @@
...
+        {
+            text = text[..(cchTextMax - 1)];
+        }
+        else
+        {
+            cchTextMax = text.Length + 1;
+        }
+
+        // Create a span from the pszText and copy the input text to it
+        Span<char> targetSpan = pszText.AsSpan()[..cchTextMax];
```
lonitra, line 25: Okay so I'm not sure why the CI points to the failing test as being `LVITEMWTests.UpdateText_should_limit_input_text_to_cchTextMax_less_1_text_longer` , but I found that the real test that is failing/aborting is [this one](https://github.com/dotnet/winforms/blob/f59965e707d1a68bf381f59a721a97ff0c6fbcd2/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewTests.cs#L4374). I think we will need to revert this back to what this originally was to resolve this. I am seeing that in the test `pszText` points to an empty string so this change results in `OutOfRangeException`.
```suggestion
        Span<char> targetSpan = new(pszText, cchTextMax);
```

elachlan:
@lonitra CI is passing!

lonitra (approved):
Awesome! Thank you for your work here! :)

src/System.Windows.Forms/src/System/Windows/Forms/TreeNode.cs
```diff
@@ -37,7 +37,8 @@ public partial class TreeNode : MarshalByRefObject, ICloneable, ISerializable
     // ie. nodes which don't use fancy fonts or colors (ie. that use the TreeView settings for these)
     //     will take up less memory than those that do.
     internal OwnerDrawPropertyBag propBag;
-    internal IntPtr _handle;
+    internal HTREEITEM HTREEITEMInternal;
```
JeremyKuhne, line 40: This is a little confusing. Better choice is to put this next to the `Handle` property and make them properties I think.

``` C#
public IntPtr Handle => (nint)HTREEITEM;
internal HTREEITEM HTREEITEMInternal { get; private set; }
internal HTREEITEM HTREEITEM
{
    get
    {
        if (HTREEITEMInternal // ...
    }
}
```

src/System.Windows.Forms.Primitives/src/Windows/Win32/UI/Controls/LVITEMW.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Windows.Win32.UI.Controls;
+
+internal partial struct LVITEMW
+{
+    /// <summary>
+    /// Set the new text. The text length is limited by <see cref="cchTextMax"/>.
```
JeremyKuhne, line 9: Can you please indent XML tag content by 1 space?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewInsertionMarkTests.cs
```diff
@@ -522,9 +521,9 @@ public unsafe void ListViewInsertionMark_Index_GetInsertMarkWithColor_Success(in
             // Set same.
             Assert.NotEqual(IntPtr.Zero, control.Handle);
             control.InsertionMark.Index = 0;
-            var insertMark = new ComCtl32.LVINSERTMARK
+            var insertMark = new LVINSERTMARK
```
JeremyKuhne, line 524: When you change code it would be good to flip to target type new to match our desired pattern

```suggestion
            LVINSERTMARK insertMark = new()
```

JeremyKuhne (changes requested):
Great work! I only have a few minor comments.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ListViewInsertionMarkTests.cs
JeremyKuhne: Don't sweat changing everything in the file, just what you touch.

elachlan: It was a small file, so I did everything.

JeremyKuhne (approved):
Great work! Thanks!

PR #9887, opened 2023-09-12
Enable nullability in ElementProxy

gpetrou:
## Proposed changes

- Enable nullability in ElementProxy.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9887)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Layout/CommonProperties.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Layout/CommonProperties.cs
@@ -485,9 +485,9 @@ internal partial class CommonProperties
         if (GetAutoSize(element))
         {
             // check for legacy layout engine
-            if (element.Container is Control)
+            if (element.Container is Control control)
             {
-                if (((Control)element.Container).LayoutEngine is DefaultLayout)
+                if (control.LayoutEngine is DefaultLayout)
                 {
                     return GetSelfAutoSizeInDefaultLayout(element);
                 }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Layout/CommonProperties.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Layout/CommonProperties.cs
@@ -485,12 +485,9 @@ internal partial class CommonProperties
         if (GetAutoSize(element))
         {
             // check for legacy layout engine
-            if (element.Container is Control control)
+            if (element.Container is Control { LayoutEngine: DefaultLayout })
             {
-                if (control.LayoutEngine is DefaultLayout)
-                {
-                    return GetSelfAutoSizeInDefaultLayout(element);
-                }
+                return GetSelfAutoSizeInDefaultLayout(element);
             }
 
             // else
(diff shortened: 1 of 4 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Layout/FlowLayout.ElementProxy.cs
```diff
@@ -16,13 +14,13 @@ internal partial class FlowLayout
...
     {
-        private IArrangedElement _element;
+        private IArrangedElement? _element;
 
         public virtual AnchorStyles AnchorStyles
         {
             get
             {
-                AnchorStyles anchorStyles = LayoutUtils.GetUnifiedAnchor(Element);
+                AnchorStyles anchorStyles = LayoutUtils.GetUnifiedAnchor(Element!);
```
LeafShi1, line 23: Element may be null here, so if use !, it will cause exception. 
same issue in below code.

LeafShi1 (changes requested):

src/System.Windows.Forms/src/System/Windows/Forms/Layout/CommonProperties.cs
```diff
@@ -485,9 +485,9 @@ internal static bool ShouldSelfSize(IArrangedElement element)
         if (GetAutoSize(element))
         {
             // check for legacy layout engine
-            if (element.Container is Control)
+            if (element.Container is Control control)
             {
-                if (((Control)element.Container).LayoutEngine is DefaultLayout)
+                if (control.LayoutEngine is DefaultLayout)
```
halgab, line 490: nit: You could have a single if clause without introducing a new variable:

`if (element.Container is Control { LayoutEngine: DefaultLayout })`

src/System.Windows.Forms/src/System/Windows/Forms/Layout/FlowLayout.ElementProxy.cs
gpetrou: The usage of this class is a bit weird. We set `Element` before using it as seen in https://github.com/dotnet/winforms/blob/b3ce60f0b2af034ce8326b7e3da81c2bef88d2b9/src/System.Windows.Forms/src/System/Windows/Forms/Layout/FlowLayout.cs#L204 So, not sure if it is worth adding lots of null checks in this file.

src/System.Windows.Forms/src/System/Windows/Forms/Layout/CommonProperties.cs
gpetrou: As mentioned before, I am not a big fan of this syntax. But if the maintainers want, I could do the change.

src/System.Windows.Forms/src/System/Windows/Forms/Layout/FlowLayout.ElementProxy.cs
```diff
@@ -85,7 +83,7 @@ public bool Stretches
 
         public virtual Size GetPreferredSize(Size proposedSize)
         {
-            return _element.GetPreferredSize(proposedSize);
+            return _element!.GetPreferredSize(proposedSize);
```
dreddy-work, line 86: nit: expressional body

dreddy-work (changes requested):

dreddy-work (approved):

PR #9888, opened 2023-09-12

gpetrou:
## Proposed changes

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9888)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyGridView.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyGridView.cs
@@ -1329,7 +1329,7 @@ internal sealed partial class PropertyGridView :
             {
                 object[] values = gridEntry.GetPropertyValueList();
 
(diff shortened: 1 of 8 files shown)


elachlan:
@gpetrou there are a couple of errors.

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyGridView.cs
```diff
@@ -1338,7 +1338,7 @@ public void DoubleClickRow(int row, bool toggleExpand, int type)
                     index++;
                 }
 
-                CommitValue(values[index]);
+                CommitValue(values![index]);
```
lonitra, line 1341: It doesn't seem like the return from `GetPropertyValueList` can be null. Is the `!` required because of the null check on `values` above?

lonitra (approved):
🚀

PR #9897, opened 2023-09-12
Enable nullability in RelatedCurrencyManager

gpetrou:
## Proposed changes

- Enable nullability in RelatedCurrencyManager.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9897)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/RelatedCurrencyManager.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/RelatedCurrencyManager.cs
@@ -14,9 +14,9 @@ namespace System.Windows.Forms;
 /// </summary>
 internal class RelatedCurrencyManager : CurrencyManager
 {
-    private BindingManagerBase parentManager;
-    private string dataField;
-    private PropertyDescriptor fieldInfo;
+    private BindingManagerBase _parentManager;
+    private string _dataField;
+    private PropertyDescriptor _fieldInfo;
     private static readonly List<BindingManagerBase> IgnoreItemChangedTable = new();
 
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/RelatedCurrencyManager.cs
```diff
@@ -173,25 +172,25 @@ private void ParentManager_CurrentItemChanged(object sender, EventArgs e)
...
                 {
-                    IgnoreItemChangedTable.Add(curManager);
-                    curManager.CancelCurrentEdit();
+                    IgnoreItemChangedTable.Add(currencyManager);
+                    currencyManager.CancelCurrentEdit();
                 }
                 finally
                 {
-                    if (IgnoreItemChangedTable.Contains(curManager))
+                    if (IgnoreItemChangedTable.Contains(currencyManager))
```
halgab, line 183: I guess this line is redundant?

gpetrou: Based on the long comment above, I don't think so.

halgab: I'm not sure I get how it addresses the need to call Contains before Remove, but it does indicate this code is tricky, so I guess staying on the safe side and not changing anything is not a bad idea...

lonitra (approved):
LGTM

PR #9902, opened 2023-09-13
Enable nullability in DataGridViewCellStyle

gpetrou:
## Proposed changes

- Enable nullability in DataGridViewCellStyle.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9902)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewCellStyleBuilder.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewCellStyleBuilder.cs
@@ -75,7 +75,7 @@ internal class DataGridViewCellStyleBuilder : Form
         {
             _cellStyle = new DataGridViewCellStyle(value);
             _cellStyleProperties.SelectedObject = _cellStyle;
-            ListenerDataGridViewDefaultCellStyleChanged(null, EventArgs.Empty);
+            ListenerDataGridViewDefaultCellStyleChanged(sender: null, EventArgs.Empty);
             _listenerDataGridView.DefaultCellStyle = _cellStyle;
             _listenerDataGridView.DefaultCellStyleChanged += ListenerDataGridViewDefaultCellStyleChanged;
         }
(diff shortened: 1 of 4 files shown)


src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewCellStyleBuilder.cs
```diff
@@ -88,7 +88,7 @@ public ITypeDescriptorContext Context
 
     private void ListenerDataGridViewDefaultCellStyleChanged(object? sender, EventArgs e)
     {
-        DataGridViewCellStyle cellStyleTmp = new DataGridViewCellStyle(_cellStyle);
+        DataGridViewCellStyle cellStyleTmp = new DataGridViewCellStyle(_cellStyle!);
```
SimonZhao888, line 91: Since this variable _cellStyle is nullable, how do we ensure that it must be non-null?
And throwing null exceptions in the constructor of class DataGridViewCellStyle, should we also need throw exceptions for null values before it?

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCellStyle.cs
```diff
@@ -395,33 +387,18 @@ internal Padding PaddingInternal
...
-    internal PropertyStore Properties
-    {
-        get
-        {
-            return propertyStore;
-        }
-    }
+    internal PropertyStore Properties => _propertyStore;
 
     internal DataGridViewCellStyleScopes Scope
```
halgab, line 392: Could it be changed to an auto-property?

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCellStyle.cs
```diff
@@ -395,33 +387,18 @@ internal Padding PaddingInternal
...
     }
 
-    internal PropertyStore Properties
-    {
-        get
-        {
-            return propertyStore;
-        }
-    }
+    internal PropertyStore Properties => _propertyStore;
```
halgab, line 390: Could it be changed to an auto-property?

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewCellStyleBuilder.cs
gpetrou: This is only called after we initialize `_cellStyle` in `CellStyle` property. So, I don't think we need to check it.

lonitra (approved):
LGTM

SimonZhao888 (approved):

PR #9909, opened 2023-09-14
Migrate `Shell32` enums to `Cswin32`

elachlan:
Migrates shell32 custom implemented enums to use ones generated from cswin32.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9909)


elachlan:
CC: @lonitra

src/System.Windows.Forms.Primitives/src/Interop/Shell32/Interop.Shell_NotifyIconW.cs
```diff
@@ -8,6 +8,6 @@ internal static partial class Interop
     internal static partial class Shell32
     {
         [DllImport(Libraries.Shell32, CharSet = CharSet.Unicode, ExactSpelling = true)]
```
JeremyKuhne, line 10: Wouldn't it be better to change this PInvoke over? The defines should come with that, I believe.

elachlan: `NOTIFYICONDATAW` has different versions for different architectures (x86/x64). I wasn't sure how to support that so I left it alone.

elachlan:
@JeremyKuhne anything else I need to address?

src/System.Windows.Forms.Primitives/src/Interop/Shell32/Interop.Shell_NotifyIconW.cs
JeremyKuhne: @elachlan Sorry that I missed your response. The right thing to do is move the struct to the same namespace that it would live under CsWin32 and put a reference to the active CsWin32 issue. https://github.com/microsoft/CsWin32/issues/882

JeremyKuhne: Same for the PInvoke itself

elachlan: Done.

JeremyKuhne (approved):
Looks good, thanks!

Issue #9823, opened 2023-08-31
RichEdit control needs moved to CsWin32


JeremyKuhne:
This is particularly important as it has a few COM interfaces that need to be moved to native pointers to support AOT/Trimming.

elachlan:
Did you want me to look into pulling all the enums from CsWin32 to begin with?

A lot of the structs are separated by X86 vs X64/Arm64 in the metadata. So I don't think we can do anything about those.

PR #9912, opened 2023-09-14
Convert interop RichEdit enums to use `CsWin32`

elachlan:
There are a few enums that were removed and replaced with constants, which I don't love. `SF` is a good example of where it isn't super great.

PFA/PFM enums are waiting on changes from https://github.com/microsoft/win32metadata/pull/1707 and so I left them alone.

Related: #9823

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9912)


elachlan:
CC: @lonitra

lonitra (approved):
LGTM

PR #8675, opened 2023-02-22
Enable nullability and refactor `LocalizationCodeDomSerializer` to replace `Hashtable` with `HashSet`

elachlan:
Related: #8143, #8357

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8675)

PR #9914, opened 2023-09-14
Enable nullability in CodeDomLocalizationProvider and clean it up

halgab:
## Proposed changes

- Enable nullability in `CodeDomLocalizationProvider`
- Replace `Hashtable`s with generic dictionaries (contributes to #8143)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9914)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomLocalizationProvider.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomLocalizationProvider.cs
@@ -1,8 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-#nullable disable
-
 using System.Collections;
 using System.Globalization;
 using System.Resources;
@@ -20,10 +18,9 @@ public sealed partial class CodeDomLocalizationProvider : IDisposable, IDesigner
 {
     private IExtenderProviderService _providerService;
     private readonly CodeDomLocalizationModel _model;
-    private readonly CultureInfo[] _supportedCultures;
     private LanguageExtenders _extender;
-    private Hashtable _memberSerializers;
-    private Hashtable _nopMemberSerializers;
+    private Hashtable? _memberSerializers;
+    private Hashtable? _nopMemberSerializers;
 
     /// <summary>
     ///  Creates a new adapter and attaches it to the serialization manager.  This
@@ -37,7 +34,7 @@ public sealed partial class CodeDomLocalizationProvider : IDisposable, IDesigner
         ArgumentNullException.ThrowIfNull(provider);
 
         _model = model;
-        Initialize(provider);
+        Initialize(provider, null);
     }
 
     /// <summary>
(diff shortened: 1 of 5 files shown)


src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomLocalizationProvider.cs
```diff
@@ -218,15 +188,15 @@ private object GetMemberCodeDomSerializer(IDesignerSerializationManager manager,
...
     {
         if (serializerType == typeof(CodeDomSerializer))
         {
-            return GetCodeDomSerializer(manager, currentSerializer, objectType, serializerType);
+            return GetCodeDomSerializer(manager, currentSerializer, objectType);
         }
         else if (serializerType == typeof(MemberCodeDomSerializer))
         {
-            return GetMemberCodeDomSerializer(manager, currentSerializer, objectType, serializerType);
+            return GetMemberCodeDomSerializer(manager, (MemberCodeDomSerializer?)currentSerializer, objectType);
```
halgab, line 199: this cast used to happen inside `GetMemberCodeDomSerializer`

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomLocalizationProvider.cs
```diff
@@ -218,15 +188,15 @@ private object GetMemberCodeDomSerializer(IDesignerSerializationManager manager,
     /// <summary>
     ///  Returns an appropriate serializer for the object.
     /// </summary>
-    object IDesignerSerializationProvider.GetSerializer(IDesignerSerializationManager manager, object currentSerializer, Type objectType, Type serializerType)
+    object? IDesignerSerializationProvider.GetSerializer(IDesignerSerializationManager manager, object? currentSerializer, Type? objectType, Type serializerType)
     {
         if (serializerType == typeof(CodeDomSerializer))
         {
-            return GetCodeDomSerializer(manager, currentSerializer, objectType, serializerType);
+            return GetCodeDomSerializer(manager, currentSerializer, objectType);
```
halgab, line 195: I suspect we could cast `currentSerializer` to `CodeDomSerializer` but I'm not 100% sure. The upside of this would be to make `_currentSerializer` in `LocalizationCodeDomSerializer` non-nullable. A null-check was added in #8675 but it is recent and historically the code would rely on it being non-null

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomLocalizationProvider.cs
```diff
@@ -47,14 +43,13 @@ public CodeDomLocalizationProvider(IServiceProvider provider, CodeDomLocalizatio
     ///  and provide the Language and Localizable properties.  The latter piece is only
     ///  supplied if CodeDomLocalizationModel is not �none�.
     /// </summary>
-    public CodeDomLocalizationProvider(IServiceProvider provider, CodeDomLocalizationModel model, CultureInfo[] supportedCultures)
+    public CodeDomLocalizationProvider(IServiceProvider provider, CodeDomLocalizationModel model, CultureInfo?[] supportedCultures)
```
halgab, line 46: Making this constructor accept a null `supportedCultures` which would make it behave the other overload would also make sense, though it would be a behaviour change

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomLocalizationProvider.cs
lonitra: I looked up the chain where `LocalizationCodeDomSerializer` is created and the only way `_currentSerializer` could be null is if the passed in type for `LocalizationCodeDomSerializer` constructor isn't of type `CodeDomSerializer`. Looking at the code, I am not confident that passed in type will always be `CodeDomSerializer` so I think we should leave `_currentSerializer` as nullable.

lonitra (commented):
Some small comments otherwise LGTM

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomLocalizationProvider.cs
halgab: I've just found out that [the documentation for `IDesignerSerializationProvider.GetSerializer`
](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.design.serialization.idesignerserializationprovider.getserializer?view=net-7.0#system-componentmodel-design-serialization-idesignerserializationprovider-getserializer(system-componentmodel-design-serialization-idesignerserializationmanager-system-object-system-type-system-type)) says that we expect `currentSerializer` to be of type `serializerType`. As you point out, I have no idea if all implementations repsect that though

lonitra (approved):
LGTM

PR #9928, opened 2023-09-15
Enable nullability in ControlBindingsCollection

gpetrou:
## Proposed changes

- Enable nullability in ControlBindingsCollection.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9928)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ControlBindingsCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ControlBindingsCollection.cs
@@ -14,14 +14,14 @@ namespace System.Windows.Forms;
 [TypeConverter($"System.Windows.Forms.Design.ControlBindingsConverter, {AssemblyRef.SystemDesign}")]
 public class ControlBindingsCollection : BindingsCollection
 {
-    private readonly IBindableComponent? _control;
+    private readonly IBindableComponent _control;
 
-    public ControlBindingsCollection(IBindableComponent? control)
+    public ControlBindingsCollection(IBindableComponent control)
     {
         _control = control;
     }
 
-    public IBindableComponent? BindableComponent => _control;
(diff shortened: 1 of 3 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ControlBindingsCollection.cs
```diff
@@ -16,18 +14,18 @@ namespace System.Windows.Forms;
 [TypeConverter($"System.Windows.Forms.Design.ControlBindingsConverter, {AssemblyRef.SystemDesign}")]
 public class ControlBindingsCollection : BindingsCollection
 {
-    private readonly IBindableComponent _control;
+    private readonly IBindableComponent? _control;
 
-    public ControlBindingsCollection(IBindableComponent control)
+    public ControlBindingsCollection(IBindableComponent? control)
```
dreddy-work, line 19: All references tell me that this shouldn't be nullable.

gpetrou: This is a public class and in the past if the constructor did not throw on null parameter, we used to just set the parameter as nullable. Do you still want me to change this one?

dreddy-work: Nowhere in the documentation we say it could be null. All the examples in the documentation lead me to believe this is not expected to be null. Lets enforce and make breaking change and throw exception.  We add `breakingchnage ` label and document it.

@JeremyKuhne  / @lonitra as FYI.

dreddy-work (approved):

gpetrou:
@dreddy-work I fixed the conflicts on this one.

PR #9938, opened 2023-09-18
Enable nullability in OleDragDropHandler and SelectionUIHandler

halgab:
Another step towards the annotation of `ComponentTray`

## Proposed changes

- Enable nullability in `OleDragDropHandler`
- Enable nullability in `SelectionUIHandler`
- Remove unused fields and fix field names

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9938)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/OleDragDropHandler.ComponentDataObject.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/OleDragDropHandler.ComponentDataObject.cs
@@ -13,58 +13,51 @@ internal partial class OleDragDropHandler
 {
     protected class ComponentDataObject : IDataObject
     {
-        private readonly IServiceProvider serviceProvider;
-        private object[]? components;
+        private readonly IServiceProvider _serviceProvider;
+        private object[]? _components;
 
-        private Stream? serializationStream;
-        private object? serializationData;
-        private readonly int initialX;
-        private readonly int initialY;
-        private readonly IOleDragClient dragClient;
-        private CfCodeToolboxItem? toolboxitemdata;
+        private MemoryStream? _serializationStream;
+        private object? _serializationData;
+        private CfCodeToolboxItem? _toolboxItemData;
 
-        public ComponentDataObject(IOleDragClient dragClient, IServiceProvider sp, object[] comps, int x, int y)
+        public ComponentDataObject(IServiceProvider serviceProvider, object[] components)
         {
-            serviceProvider = sp;
-            components = GetComponentList(comps);
-            initialX = x;
-            initialY = y;
-            this.dragClient = dragClient;
+            _serviceProvider = serviceProvider;
+            _components = GetComponentList(components);
         }
 
-        public ComponentDataObject(IOleDragClient dragClient, IServiceProvider sp, object serializationData)
+        public ComponentDataObject(IServiceProvider serviceProvider, object serializationData)
         {
-            serviceProvider = sp;
-            this.serializationData = serializationData;
-            this.dragClient = dragClient;
+            _serviceProvider = serviceProvider;
+            _serializationData = serializationData;
         }
 
-        private Stream? SerializationStream
+        private MemoryStream? SerializationStream
         {
             get
             {
-                if (serializationStream is null && components is not null)
+                if (_serializationStream is null && _components is not null)
                 {
-                    IDesignerSerializationService? ds = serviceProvider.GetService<IDesignerSerializationService>();
+                    IDesignerSerializationService? ds = _serviceProvider.GetService<IDesignerSerializationService>();
                     if (ds is not null)
                     {
-                        IComponent[] comps = new IComponent[components.Length];
-                        for (int i = 0; i < components.Length; i++)
+                        IComponent[] components = new IComponent[_components.Length];
+                        for (int i = 0; i < _components.Length; i++)
                         {
-                            Debug.Assert(components[i] is IComponent, $"Item {components[i].GetType().Name} is not an IComponent");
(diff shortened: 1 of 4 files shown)


src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/OleDragDropHandler.ComponentDataObject.cs
```diff
@@ -228,71 +203,61 @@ public string[] GetFormats()
...
+                objects.CopyTo(_components, 0);
 
                 // only do top-level components here,
                 // because other are already parented.
                 // otherwise, when we process these
                 // components it's too hard to know what we
                 // should be reparenting.
-                ArrayList topComps = new ArrayList();
-                for (int i = 0; i < components.Length; i++)
+                List<object> topComponents = new(_components.Length);
```
dreddy-work, line 244: Do we need to assign size here? are you sure the size of the `topComponents ` is always same as `_components` ? but not less.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/OleDragDropHandler.cs
```diff
@@ -24,134 +22,96 @@ internal partial class OleDragDropHandler
...
 
                 ICollection objects = ds.Deserialize(serializationData);
                 if (objects.Count > 0)
                 {
-                    bool dropOk = true;
-
                     foreach (object o in objects)
                     {
-                        if (!(o is IComponent))
+                        if (o is IComponent component && !Destination.IsDropOk(component))
```
dreddy-work, line 108: nice!.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/OleDragDropHandler.ComponentDataObject.cs
halgab: We could very well write `List<object> topComponents = new();`. It's just that we know `topComponents` cannot be bigger than `_components.Length`, so setting that as the capacity to avoid list resizes seemed like a good idea to me.
That being said, I find this code a bit odd. I don't understand why we need to copy `objects` into `_components` before filtering it and why we can't do both at the same time...

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/OleDragDropHandler.ComponentDataObject.cs
```diff
@@ -15,119 +13,97 @@ internal partial class OleDragDropHandler
...
                     Deserialize(null, false);
-                    if (components is null)
+                    if (_components is null)
                     {
                         return Array.Empty<object>();
                     }
                 }
 
-                return (object[])components.Clone();
+                return (object[])_components!.Clone();
```
halgab, line 77: I don't like that I had to write `_components!`. Either some null checks are redundant on line 68 or the check on line 71 should be move to the outer scope I think

dreddy-work:
@halgab, can you get this to a review state by fixing buil issues?

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/OleDragDropHandler.ComponentDataObject.cs
dreddy-work: `Deserialize(null, false);` on line 70 ensure `_components` is not null.  So check on line 71 and the block of code is unnecessary. 

I think it was intended to be moved out  and make line 77  as
 
```CS

return _components is null ? Array.Empty<object>() : (object[])_components!.Clone()

```

This change will be breaking.  @halgab, can you see if we have unit tests that are running through this code path?

halgab:
Sorry, I forgot to check the CI after my last rebase...

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/OleDragDropHandler.ComponentDataObject.cs
halgab: I don't see any test coverage, but I've looked manually. I don't have any tool that allows me to measure code coverage, so I can't be sure. Should I make the change anyway? Or maybe in a separate PR?

dreddy-work: @halgab , i lost context here around `breaking change` comment. Did you revert or rebase code here?

halgab: So far I haven't done anything. I checked the code wasn't covered (not 100% sure though), and I was asking if I should go ahead and make the (breaking) change anyway considering the current code looks broken

dreddy-work: sure. Go ahead. If you still have to end up using `!`, i would go with debug assert.

dreddy-work (changes requested):

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/OleDragDropHandler.ComponentDataObject.cs
halgab: done. Sorry for the delay, this fell off my radar

lonitra (approved):
LGTM

PR #9947, opened 2023-09-19
Enable nullability in DataGridViewImageColumn

gpetrou:
## Proposed changes

- Enable nullability in DataGridViewImageColumn.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9947)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewImageColumn.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewImageColumn.cs
@@ -1,8 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-#nullable disable
-
 using System.ComponentModel;
 using System.Drawing;
 
@@ -13,10 +11,11 @@ namespace System.Windows.Forms;
 public class DataGridViewImageColumn : DataGridViewColumn
 {
     private static readonly Type s_columnType = typeof(DataGridViewImageColumn);
-    private Image _image;
-    private Icon _icon;
+    private Image? _image;
+    private Icon? _icon;
 
-    public DataGridViewImageColumn() : this(false /*valuesAreIcons*/)
+    public DataGridViewImageColumn()
+        : this(valuesAreIcons: false)
     {
     }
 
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewImageColumn.cs
```diff
@@ -121,26 +115,17 @@ public Icon Icon
...
     }
 
-    private DataGridViewImageCell ImageCellTemplate
-    {
-        get
-        {
-            return (DataGridViewImageCell)CellTemplate;
-        }
-    }
+    private DataGridViewImageCell? ImageCellTemplate => (DataGridViewImageCell?)CellTemplate;
```
halgab, line 128: Every call site of this property needs an extra `!`, except [here](https://github.com/dotnet/winforms/blob/1aa66e4505a8d38518ea7439b3867a653d4376fb/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewImageColumn.cs#L180C43-L180C43) where we check it for null. Can't we imagine adding this exception throwing null check to this property itself?

dreddy-work: @gpetrou, Everywhere else, it would anyway throw null reference (if null).

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewImageColumn.cs
```diff
@@ -247,24 +217,19 @@ public override object Clone()
         }
         else
         {
-            //
-
-            dataGridViewColumn = (DataGridViewImageColumn)System.Activator.CreateInstance(thisType);
+            dataGridViewColumn = (DataGridViewImageColumn)Activator.CreateInstance(thisType)!;
```
dreddy-work, line 220: so, it can't be null here?

gpetrou: I think so. We already have similar usages in other classes where we create instance of a known type.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewImageColumn.cs
```diff
@@ -166,7 +144,7 @@ public DataGridViewImageCellLayout ImageLayout
         {
             if (ImageLayout != value)
             {
-                ImageCellTemplate.ImageLayout = value;
+                ImageCellTemplate!.ImageLayout = value;
```
dreddy-work, line 147: why do you need `!` here?

dreddy-work (approved):

lonitra (approved):
👍

PR #9949, opened 2023-09-19
[release/8.0] Make System.Drawing.Common stable for rtm

dkurepa:
System.Drawing.Common has been ported from runtime during the .NET 8 previews. In it's current form, it will have an unstable version when we switch to rtm. This PR fixes that

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9949)

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -9,6 +9,9 @@
     <PreReleaseVersionIteration>2</PreReleaseVersionIteration>
     <VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
     <UsingToolMicrosoftNetCompilers>true</UsingToolMicrosoftNetCompilers>
+    <!-- StabilizePackageVersion should be switched to true for rtm -->
+    <StabilizePackageVersion Condition="'$(StabilizePackageVersion)' == ''">false</StabilizePackageVersion>
+    <DotNetFinalVersionKind Condition="'$(StabilizePackageVersion)' == 'true'">release</DotNetFinalVersionKind>
   </PropertyGroup>
   <!-- Below have corresponding entries in Versions.Details.XML because they are updated via Maestro -->
   <PropertyGroup>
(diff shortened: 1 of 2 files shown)


dreddy-work:
@ViktorHofer, can you help review this?

eng/Versions.props
```diff
@@ -9,6 +9,9 @@
     <PreReleaseVersionIteration>2</PreReleaseVersionIteration>
     <VersionPrefix>$(MajorVersion).$(MinorVersion).$(PatchVersion)</VersionPrefix>
     <UsingToolMicrosoftNetCompilers>true</UsingToolMicrosoftNetCompilers>
+    <!-- StabilizePackageVersion should be switched to true for rtm -->
```
Tanya-Solyanik, line 12: release/8.0 is already branded for rtm, can we set it to true now?

dkurepa: No, not yet. We will do that once we start working on the 8.0.0 release, after rc2

src/System.Drawing.Common/src/System.Drawing.Common.csproj
```diff
@@ -10,6 +10,8 @@
     <CLSCompliant>true</CLSCompliant>
     <ProduceReferenceAssembly>true</ProduceReferenceAssembly>
     <IsPackable>true</IsPackable>
+    <IsShipping>true</IsShipping>
```
Tanya-Solyanik, line 13: Do we need this change in main as well?

Tanya-Solyanik (approved):
The change looks reasonable, but @ViktorHofer is the expert in this area and should review this change.

src/System.Drawing.Common/src/System.Drawing.Common.csproj
Tanya-Solyanik: never mind, I see it's there already.

ViktorHofer (approved):

dkurepa:
@Tanya-Solyanik @dreddy-work can you merge the PR please? I don't have the authorization to do so

PR #9952, opened 2023-09-19
Enable nullability in DataGridViewLinkCell

gpetrou:
## Proposed changes

- Enable nullability in DataGridViewLinkCell.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9952)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewLinkCell.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewLinkCell.cs
@@ -271,10 +271,7 @@ public partial class DataGridViewLinkCell : DataGridViewCell
         }
     }
 
-    private bool ShouldSerializeLinkVisited()
-    {
-        return _linkVisitedSet = true;
-    }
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewLinkCell.cs
```diff
@@ -721,7 +752,7 @@ protected override Size GetPreferredSize(Graphics graphics, DataGridViewCellStyl
         return preferredSize;
     }
 
-    protected override object GetValue(int rowIndex)
+    protected override object? GetValue(int rowIndex)
```
dreddy-work, line 755: is this nullability coming from base? Did you check any case where this could be null?

gpetrou: If you check the base method, you will notice a case where we have `return null;`.

dreddy-work (approved):

PR #9955, opened 2023-09-20
Fix Accessibility: Can't select TreeView items using UIAutomation SelectionItem.Select

### LINKED ISSUE #9822: Accessibility: Can't select TreeView items using UIAutomation SelectionItem.Select
### .NET version

8.0 Preview 7.

Yes

.NET 6 was working

### Issue description

Using UIAutomation API to select a TreeView item fails - the root item gets selected but not the desired item.

### Steps to reproduce

Create a basic WinForms app and add a TreeView. Edit the Nodes collection and add a root and 5 children. Run the app.
Using Inspect.exe (included in Visual Studio), expand the app and the tree and the root node. Click on one of the nodes then in the menu select Action=>SelectionItem.Select. Note that the node does not get selected. This is a regression from .NET 6 (didn't try 7).

Epica3055:
Fix #9822

## Proposed changes

- 
- TreeViewItem add `SelectItem` method.
- 

## Regression? 

- Yes 

### Before

9955_before

### After

9955_after

## Test methodology 

- 
- unit test
- 

## Test environment(s) 
8.0.0-preview.7.23375.6
- 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9955)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TreeNode.TreeNodeAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TreeNode.TreeNodeAccessibleObject.cs
@@ -98,6 +98,11 @@ public partial class TreeNode
                 _ => base.FragmentNavigate(direction),
             };
 
+        internal override void SelectItem()
+        {
+            _owningTreeView.SelectedNode = _owningTreeNode;
+        }
+
         internal override object? GetPropertyValue(UiaCore.UIA propertyID)
             => propertyID switch
             {
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TreeNode.TreeNodeAccessibleObjectTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TreeNode.TreeNodeAccessibleObjectTests.cs
@@ -465,6 +465,27 @@ public class TreeNodeAccessibleObjectTests
     }
 
     [WinFormsFact]
+    public void TreeNodeAccessibleObject_SelectItem_WorksExpected()
+    {
+        using TreeView control = new();
+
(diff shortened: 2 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/TreeNode.TreeNodeAccessibleObject.cs
```diff
@@ -98,6 +98,11 @@ internal override UiaCore.IRawElementProviderFragmentRoot FragmentRoot
                 _ => base.FragmentNavigate(direction),
             };
 
+        internal override void SelectItem()
```
Tanya-Solyanik, line 101: To fully implement the pattern, please implement AddToSelection and RemoveFromSelection too - https://github.com/dotnet/winforms/blob/6560d1902eb0016c652557cb2f5d4920e824b14f/src/System.Windows.Forms/src/System/Windows/Forms/AccessibleObject.cs#L698-L704

Tanya-Solyanik: And also investigate if we should add any events associated with this pattern. See https://learn.microsoft.com/dotnet/api/system.windows.automation.selectionitempattern?view=windowsdesktop-7.0

Epica3055: I think these two methods might not suit our situation. because it seems TreeView doesn't support mutilSelection.

by the way, other methods that `ISelectionItemProvider`  requires have been implemented.

Tanya-Solyanik (approved):

PR #9960, opened 2023-09-20
Convert structs to primitive types when passing them to the interop methods that expect VARIANTs

Tanya-Solyanik:
This problem manifests as a first chance NotSupportedException    at System.StubHelpers.ObjectMarshaler.ConvertToNative(Object objSrc, IntPtr pDstVariant)
```
System.NotSupportedException
  HResult=0x80131515
  Message=Type 'Windows.Win32.Foundation.COLORREF' cannot be marshalled to a Variant. Type library is not registered.
  Source=<Cannot evaluate the exception source>
  StackTrace:
<Cannot evaluate the exception stack trace>
```
This exception is thrown when the user is selecting text in a text box (or any text control that has a UIA text provider) under a Narrator. UIA is querying the foreground and background text colors and the bounding rectangle of the control. Colors and bounding rectangle are c# structs, and our UIA implementations cast these structs to an object. However, the interop layer can't convert such objects to VARIANT types expected by the UIA.   

The fix is a copy of #9732, i.e. explicitly convert struct to a primitive type that has a conversion to a VARIANT. Specific types are documented with the UIA property or attribute definitions - https://learn.microsoft.com/en-us/windows/win32/winauto/uiauto-textattribute-ids and https://learn.microsoft.com/windows/win32/api/uiautomationcore/nf-uiautomationcore-irawelementproviderfragment-get_boundingrectangle.

I had reviewed all overrides of GetPropertyValue and GetAttributeValue methods, and did not find any other non-primitive values that require additional conversion.
 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9960)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextProvider.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextProvider.cs
@@ -95,6 +95,15 @@ internal abstract class UiaTextProvider : ITextProvider
         return doubles;
     }
 
+    /// <summary>
+    ///  Bounding rectangles are represented by a VT_ARRAY of doubles in a native VARIANT
+    ///  in accessibility interfaces. This method does the conversion. Accessibility will then convert it to an UiaRect.
+    ///  https://learn.microsoft.com/windows/win32/api/uiautomationcore/nf-uiautomationcore-irawelementproviderfragment-get_boundingrectangle
+    ///  https://learn.microsoft.com/windows/win32/api/uiautomationcore/ns-uiautomationcore-uiarect
+    /// </summary>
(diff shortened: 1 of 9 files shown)


JeremyKuhne (approved):
Nice!

dreddy-work:
First commit in this PR shouldn't be in this. I am surprised why you are not seeing a conflict here.

Tanya-Solyanik:
>First commit in this PR shouldn't be in this. I am surprised why you are not seeing a conflict here.

This is because I submitted this PR before you merged the nullability fix :) - I had to run some errands, and wanted to get Jeremy's review ASAP. Now I'm back and will clean up the branch

Issue #9824, opened 2023-08-31
Move Mshtml to Cswin32


lonitra:
Interop.Mshtml has quite a few COM interfaces that need to be moved to use Cswin32 to support trimming/AOT

elachlan:
@lonitra is there an example of something similar we have done previously in winforms? I'd like to understand the pattern around this so I can help implement it.

lonitra:
> is there an example of something similar we have done previously in winforms?

Sorry for the delay in response! I've just put up a PR that illustrates the pattern here, so this would be a good starting point to understanding the pattern around moving COM interfaces over to Cswin32. I'm not sure if there is another singular PR to point to that demonstrates the pattern -- AFAIK it is kind of scattered throughout PRs as we were incrementally designing it.

merriemcgaw:
Does #10611 address this?

elachlan:
@merriemcgaw no, that is separate.

PR #9972, opened 2023-09-21
Convert Mshtml COM interfaces to Cswin32

lonitra:
Related: #9824 
This PR moves a chunk of our Mshtml COM interfaces over to utilizing Cswin32 definition. This PR also:
- Adds implicit conversations to/from `VARIANT_BOOL` and `bool`
- Adds helper methods for getting `FDEX_PROP_FLAGS` for properties/methods/fields into `IDispatch`. These were determined using the same value that runtime uses
-  Add `SafeArrayScope` + tests to ensure that `SAFEARRAY`s get destroyed correctly
- Remove `IReflect` off of `HtmlToClrEventProxy` and implement `StandardDispatch` manually give responses for the properties/methods there + add tests for it
- Update tests which include utilizing pointers instead of interfaces as a workaround due to bug involving late binding-only interfaces

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9972)


src/System.Windows.Forms/src/System/Windows/Forms/HtmlElement.cs
```diff
@@ -677,14 +818,23 @@ public void SetAttribute(string attributeName, string value)
...
 
     public static bool operator !=(HtmlElement left, HtmlElement right) => !(left == right);
 
-    public override int GetHashCode() => _htmlElement?.GetHashCode() ?? 0;
+    public override int GetHashCode()
+    {
+        using var htmlElement = _htmlElement.GetInterface();
+        using var unknown = htmlElement.TryQuery<IUnknown>(out HRESULT hr);
+        return hr.Succeeded ? (int)unknown.Value : 0;
+    }
```
lonitra, line 837: Originally I had this code as 
```c#
public override int GetHashCode()
{
    using var htmlElement = _htmlElement.GetInterface();
    return htmlElement.Value->GetHashCode();
}
```
but found out via tests that even with different `HtmlElement` objects, the hash code was the same, which is unexpected. I also alternatively tried `return unknown.Value->GetHashCode()` but again was getting the same hash code for different `HtmlElement`. The code now seems to be the right behavior as `HtmlElement` objects that are the same will now return the same hash code while differing `HtmlElement` objects will return different hash codes. We also do something similar for determining if they are equal above.

src/System.Windows.Forms/src/System/Windows/Forms/HtmlDocument.HtmlDocumentShim.cs
```diff
@@ -103,11 +128,7 @@ protected override void Dispose(bool disposing)
             base.Dispose(disposing);
             if (disposing)
             {
-                if (_htmlDocument is not null)
-                {
-                    Marshal.FinalReleaseComObject(_htmlDocument.NativeHtmlDocument2);
-                }
-
+                _htmlDocument?.NativeHtmlDocument2.Dispose();
```
lonitra, line 131: Ideally I wanted to use `DisposeHelper.NullAndDispose()` to also null out the `AgileComPointer` field that `_htmlDocument.NativeHtmlDocument2` is associated with, but wasn't sure how to without changing visibility of the field to `internal`, so is it fine that the field isn't nulled out?

src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/VARIANT_BOOL.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Windows.Win32.Foundation;
+
+internal readonly partial struct VARIANT_BOOL
+{
+    public static implicit operator bool(VARIANT_BOOL value) => value != VARIANT_FALSE;
+
+    public static implicit operator VARIANT_BOOL(bool value) => value ? VARIANT_TRUE : VARIANT_FALSE;
```
elachlan, line 10: @AArnott would it make sense to add these operators in Cswin32?

AArnott: Yes! Wanna file a PR, or merely an issue?

elachlan: I'll file an issue. It's been a while since I worked on CsWin32 so I'd be pretty rusty.

src/System.Windows.Forms.Primitives/src/Windows/Win32/System/Com/IDispatch.cs
```diff
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using Windows.Win32.System.Variant;
+using FDEX_PROP_FLAGS = Windows.Win32.System.Ole.FDEX_PROP_FLAGS;
```
JeremyKuhne, line 5: Is there a reason you don't include the namespace? If you want to make the methods a little cleaner you could also add a static using. Then you'd be able to just use `fdex...`.

```suggestion
using Windows.Win32.System.Ole;
using static Windows.Win32.System.Ole.FDEX_PROP_FLAGS;
```

src/System.Windows.Forms.Primitives/src/Windows/Win32/System/Com/SafeArrayScope.cs
```diff
@@ -0,0 +1,37 @@
...
+///  <para>
+///   Use in a <see langword="using" /> statement to ensure the <see cref="SAFEARRAY"/> gets disposed.
+///   Use <see cref="Value"/> to pass the <see cref="SAFEARRAY"/> to APIs that populate it.
+///  </para>
+/// </remarks>
+internal unsafe ref struct SafeArrayScope
+{
+    public SAFEARRAY* Value { get; private set; }
+
+    public SafeArrayScope(VARENUM vt, uint cDims, SAFEARRAYBOUND bound)
```
JeremyKuhne, line 22: Consider simplifying this. I don't think there is ever a case we'll want more than a single dimensional 0 based array.

src/System.Windows.Forms/src/System/Windows/Forms/HtmlDocument.HtmlDocumentShim.cs
```diff
@@ -39,9 +38,28 @@ internal HtmlDocumentShim(HtmlDocument htmlDocument)
...
+            get
+            {
+                if (_associatedWindow is null)
+                {
+                    return null;
+                }
 
-        public IHTMLDocument2 NativeHtmlDocument2 => _htmlDocument.NativeHtmlDocument2;
+                using var window = _associatedWindow.GetInterface();
+                return (IHTMLWindow2.Interface?)ComHelpers.GetObjectForIUnknown(window.AsUnknown);
```
JeremyKuhne, line 51: We're doing this a lot. Might be useful to add a helper on `AgileComPointer`. Something like "object GetManagedObject()". That would also let us tweak the shared code if necessary.

src/System.Windows.Forms/src/System/Windows/Forms/HtmlDocument.HtmlDocumentShim.cs
JeremyKuhne: It's probably ok

src/System.Windows.Forms/src/System/Windows/Forms/HtmlDocument.cs
```diff
@@ -144,7 +200,10 @@ public Color ForeColor
         {
             try
             {
-                return ColorFromObject(NativeHtmlDocument2.GetFgColor());
+                using var htmlDoc2 = NativeHtmlDocument2.GetInterface();
+                VARIANT color = default;
+                htmlDoc2.Value->get_fgColor(&color).ThrowOnFailure();
+                return ColorFromObject(color.ToObject());
```
JeremyKuhne, line 206: I'd change `ColorFromObject` to take `VARIANT`.

src/System.Windows.Forms/src/System/Windows/Forms/HtmlDocument.cs
```diff
@@ -144,7 +200,10 @@ public Color ForeColor
         {
             try
             {
-                return ColorFromObject(NativeHtmlDocument2.GetFgColor());
+                using var htmlDoc2 = NativeHtmlDocument2.GetInterface();
+                VARIANT color = default;
```
JeremyKuhne, line 204: These returned `VARIANT`s need a using.

src/System.Windows.Forms/src/System/Windows/Forms/HtmlDocument.cs
```diff
@@ -273,70 +413,114 @@ public HtmlElementCollection Forms
...
+        };
+
+        using SafeArrayScope scope = new(VARENUM.VT_BSTR, 1, saBound);
+        if (scope.IsNull)
+        {
+            return;
+        }
+
+        Span<int> indices = stackalloc int[] { 0 };
+        using BSTR bstrText = new(text);
```
JeremyKuhne, line 438: This would probably be good to move into the scope. Add a `string` indexer to get and set strings.

It wasn't clear to me whether or not the `SafeArrayPutElement` would copy the string, had to look at the source code to confirm it does.

src/System.Windows.Forms/src/System/Windows/Forms/HtmlElement.cs
```diff
@@ -508,25 +640,34 @@ public unsafe object InvokeMember(string methodName, params object[] parameter)
 
     public void RemoveFocus()
     {
-        ((IHTMLElement2)NativeHtmlElement).Blur();
+        using var htmlElement2 = GetHtmlElement<IHTMLElement2>();
+        htmlElement2.Value->blur();
```
JeremyKuhne, line 644: ThrowOnFailure?

src/System.Windows.Forms/src/System/Windows/Forms/HtmlElement.cs
JeremyKuhne: Did you try just keeping the original code? I'm not sure if there is value in trying to find the `IUnknown`. Particularly given the GIT. Built-in COM does unify, so maybe we could expose a hash of the original pointer from `AgileComPointer` by putting it through `HashCode<nint>.Combine()`?

src/System.Windows.Forms/src/System/Windows/Forms/HtmlToClrEventProxy.cs
```diff
@@ -13,136 +14,161 @@ namespace System.Windows.Forms;
...
+                if (result is not null)
+                {
+                    *result = VARIANT.Empty;
+                }
+
+                return HRESULT.S_OK;
+            case EventNamePropertyDispId:
+                if (flags.HasFlag(DISPATCH_FLAGS.DISPATCH_PROPERTYGET))
+                {
+                    if (result is not null)
```
JeremyKuhne, line 154: Maybe better to return `DISP_E_PARAMNOTOPTIONAL` for null in these cases? I have mixed feelings. Not too concerned in either case.

src/System.Windows.Forms/src/System/Windows/Forms/WebBrowser.cs
```diff
@@ -334,20 +348,18 @@ public string DocumentTitle
...
-                {
-                    documentTitle = htmlDocument2.GetTitle();
-                }
-                catch (COMException)
-                {
-                    documentTitle = string.Empty;
-                }
+                using var htmlDocument2 = ComHelpers.GetComScope<IHTMLDocument2>(htmlDocument.DomDocument);
+                Debug.Assert(!htmlDocument2.IsNull, "The HtmlDocument object must implement IHTMLDocument2.");
+                BSTR title = default;
```
JeremyKuhne, line 360: Probably better to be consistent and do a using with the BSTR even though it probably isn't necessary here (that is we're unlikely to throw out before the `ToStringAndFree()`.

src/System.Windows.Forms/src/System/Windows/Forms/WebBrowser.cs
```diff
@@ -1073,16 +1130,30 @@ protected override void Dispose(bool disposing)
...
     protected override void AttachInterfaces(object nativeActiveXObject)
     {
-        _axIWebBrowser2 = (Mshtml.IWebBrowser2)nativeActiveXObject;
+        if (nativeActiveXObject is null)
+        {
+            return;
+        }
+
+        IWebBrowser2* webBrowser2 = ComHelpers.GetComPointer<IWebBrowser2>(nativeActiveXObject);
+        if (_axIWebBrowser2 is not null)
```
JeremyKuhne, line 1139: Not sure it's useful to check here? Probably better to just NullAndDispose and create a new one even if it happens to be the same object?

src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/AgileComPointer.cs
```diff
@@ -61,6 +53,19 @@ public AgileComPointer(TInterface* @interface, bool takeOwnership)
...
+    ///  <see cref="AgileComPointer{TInterface}"/> was created from.
+    /// </summary>
+    /// <remarks>
+    ///  <para>
+    ///   This is useful in avoiding recreating a new <see cref="AgileComPointer{TInterface}"/> for the same
+    ///   object.
+    ///  </para>
+    /// </remarks>
+    public bool MatchesOriginalPointer(TInterface* @interface)
+        => _cookie != 0 && @interface == _originalPointer;
```
JeremyKuhne, line 67: You should query IUnknown here for comparison.

src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/AgileComPointer.cs
```diff
@@ -28,20 +28,7 @@ internal unsafe class AgileComPointer<TInterface> :
...
-    /// </summary>
-    /// <remarks>
-    ///  <para>
-    ///   This is useful in avoiding recreating a new <see cref="AgileComPointer{TInterface}"/> for the same
-    ///   object.
-    ///  </para>
-    /// </remarks>
-    public bool MatchesOriginalPointer(TInterface* @interface)
-        => _cookie != 0 && @interface == _originalPointer;
+    private readonly IUnknown* _originalPointer;
```
JeremyKuhne, line 31: _originalObject might be clearer

src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/AgileComPointer.cs
```diff
@@ -98,6 +103,14 @@ public ComScope<TAsInterface> TryGetInterface<TAsInterface>(out HRESULT hr)
         return scope;
     }
 
+    public object GetManagedObject()
```
JeremyKuhne, line 106: Comment

src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/AgileComPointer.cs
```diff
@@ -61,6 +53,19 @@ public AgileComPointer(TInterface* @interface, bool takeOwnership)
...
+    ///  Returns <see langword="true"/> if the given <paramref name="interface"/> is the same pointer this
+    ///  <see cref="AgileComPointer{TInterface}"/> was created from.
+    /// </summary>
+    /// <remarks>
+    ///  <para>
+    ///   This is useful in avoiding recreating a new <see cref="AgileComPointer{TInterface}"/> for the same
+    ///   object.
+    ///  </para>
+    /// </remarks>
+    public bool MatchesOriginalPointer(TInterface* @interface)
```
JeremyKuhne, line 66: MatchesOriginalObject might be clearer

src/System.Windows.Forms.Primitives/src/Windows/Win32/System/Com/SafeArrayScope.cs
```diff
@@ -0,0 +1,115 @@
...
+///  </para>
+/// </remarks>
+internal unsafe ref struct SafeArrayScope<T>
+{
+    public SAFEARRAY* Value { get; private set; }
+
+    public SafeArrayScope(uint size)
+    {
+        VARENUM vt = VARENUM.VT_EMPTY;
+        Type type = typeof(T);
```
JeremyKuhne, line 25: I'm not sure what the matching rules are for the JIT to elide the branches. I've usually seen this as `typeof(T) ==` repeatedly. I'd play it safe and not stick it in a local.

src/System.Windows.Forms.Primitives/src/Windows/Win32/System/Com/SafeArrayScope.cs
```diff
@@ -0,0 +1,115 @@
...
+        Value = PInvoke.SafeArrayCreate(vt, 1, &saBound);
+    }
+
+    public readonly T? this[int i]
+    {
+        get
+        {
+            Span<int> indices = stackalloc int[] { i };
+
+            Type type = typeof(T);
```
JeremyKuhne, line 57: Again, probably not stick it in a local. You could validate by using sharplab.io.

src/System.Windows.Forms/src/System/Windows/Forms/HtmlElement.cs
```diff
@@ -419,33 +534,43 @@ public void Focus()
...
     public string GetAttribute(string attributeName)
     {
-        object oAttributeValue = NativeHtmlElement.GetAttribute(attributeName, 0);
-        return oAttributeValue is null ? "" : oAttributeValue.ToString();
+        using var htmlElement = NativeHtmlElement.GetInterface();
+        using BSTR name = new(attributeName);
+        using VARIANT attributeValue = default;
+        htmlElement.Value->getAttribute(name, 0, &attributeValue).ThrowOnFailure();
+        return attributeValue.Type == VARENUM.VT_NULL || attributeValue.ToObject() is not string validString
+            ? ""
```
JeremyKuhne, line 542: nit: `string.Empty` (although I think this may do the right thing now, still a good habit to be in)

src/System.Windows.Forms/src/System/Windows/Forms/HtmlWindow.cs
```diff
@@ -385,14 +483,18 @@ public void ScrollTo(Point point)
         }
 
         // Neither are null. Get the IUnknowns and compare them.
-        using var leftUnknown = ComHelpers.GetComScope<IUnknown>(left.NativeHtmlWindow);
-        using var rightUnknown = ComHelpers.GetComScope<IUnknown>(right.NativeHtmlWindow);
+        using var leftHtmlWindow = left.NativeHtmlWindow.GetInterface();
+        using var rightHtmlWindow = right.NativeHtmlWindow.GetInterface();
+        using var leftUnknown = leftHtmlWindow.TryQuery<IUnknown>(out HRESULT hr);
```
JeremyKuhne, line 488: This would be better done by comparing the two AgileComPointers. As we don't want to expose the original IUnknown I'd add an overload to the Are that takes the T* the `MatchesOriginalPointer`.  Maybe `IsSameNativeObject()` would be a better name?

src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/AgileComPointer.cs
```diff
@@ -61,6 +53,33 @@ public AgileComPointer(TInterface* @interface, bool takeOwnership)
...
+        return _cookie != 0 && unknownScope.Value == _originalObject;
+    }
+
+    /// <summary>
+    ///  Returns <see langword="true"/> if <paramref name="other"/> has the same pointer this
+    ///  <see cref="AgileComPointer{TInterface}"/> was created from.
+    /// </summary>
+    public bool IsSameNativeObject(AgileComPointer<TInterface> other)
+    {
+        using var otherInterface = other.GetInterface();
```
JeremyKuhne, line 79: This should just be comparing `_originalObject == other._originalObject`. `GetInterface()` is coming from the GIT, there is a good chance you'll get a proxy pointer as the `IUnknown`.

JeremyKuhne (approved):
Nice work!!

PR #9979, opened 2023-09-22
Fix TabPage cannot add to TablControl in DemoConsole

### LINKED ISSUE #9968: Failed to "Add Tab" for TabControl in DemoConsole application
### .NET version

Yes

Yes, it works well in .Net 8.0 from Release/8.0 branch of Winforms repo

### Issue description

In [DemoConsole](https://github.com/dotnet/winforms/tree/main/src/System.Windows.Forms/tests/IntegrationTests/DesignSurface/DemoConsole) test app of Winforms repo, failed to "Add Tab" for TabControl with popping up an error dialog. And "Remove Tab" works well.

Actual result in DemoConsole application when **testing main branch**:
image
AddTab_Failed

Expected result in DemoConsole application when **testing release/8.0 branch**:
AddTab_success

### Steps to reproduce

SimonZhao888:
Fixes #9968 

When we click the control of 'Add Tab', an exception message will be thrown in the CreateName method:
image
DataGridViewComboBoxColumn cannot be converted to component, 
image
image

So we need to modify the creation method of the DataGridViewComboBoxColumn object.

## Proposed changes

- Modify the creation way of DataGridViewComboBoxColumn object.

## Customer Impact

- TabPage can add to TabControl in DemoConsole.

## Regression? 

- No

## Risk

-Minimal

### Before

image

### After

TabControl_Image

## Test methodology 

- Manual(DemoConsole)

## Test environment(s) 

- 8.0.100-preview.7.23376.3

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9979)

### DIFF
--- a/src/System.Windows.Forms/tests/IntegrationTests/DesignSurface/DemoConsole/MainForm.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/DesignSurface/DemoConsole/MainForm.cs
@@ -259,7 +259,7 @@ public partial class MainForm : Form
                         bindingSource.DataSource = new List<string> { "a1", "b2", "c3", "d4", "e5", "f6" };
                         listBox.DataSource = bindingSource;
                         DataGridView dataGridView = surface.CreateControl<DataGridView>(new Size(200, 150), new Point(470, 220));
-                        DataGridViewComboBoxColumn comboBoxColumn = new DataGridViewComboBoxColumn();
+                        DataGridViewComboBoxColumn comboBoxColumn = surface.CreateComponent<DataGridViewComboBoxColumn>();
                         comboBoxColumn.HeaderText = "Column1";
                         dataGridView.Columns.AddRange([comboBoxColumn]);
                     }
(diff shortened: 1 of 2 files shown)


dreddy-work:
What regressed this?

lonitra:
This issue was only occurring in DemoConsole? Adding TabPage works properly in other WinForms apps outside of DemoConsole?

Tanya-Solyanik:
>This issue was only occurring in DemoConsole? Adding TabPage works properly in other WinForms apps outside of DemoConsole?

Yes.

Tanya-Solyanik:
@SimonZhao888  - creation of a new tab should not depend on other controls present on the form. Ehy is the datagridview column name re-created?

SimonZhao888:
> @SimonZhao888 - creation of a new tab should not depend on other controls present on the form. Why is the datagridview column name re-created? I would expect that creation of this page should always fail, not only when clicking in the property browser....

Actually the DataGridView column name is not re-created. When we click the Add Tab button in DemoConsole,  the method of CreateName is always accessed, which will accessed all the Components in the current container and is used to supply a name to the control just created.  But the  DataGridViewComboBoxColumn cannot be converted to a Componet, so we will get a null value at this line, and it will throw an Exception when we use the comp.GetType() method.
image

And you also can see the call stack from the following picture:
image

Tanya-Solyanik (approved):
Thank you!

PR #9982, opened 2023-09-23
Enable nullability in remaining DataGridView members

gpetrou:
## Proposed changes

- Enable nullability in remaining DataGridView members.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9982)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.cs
@@ -4160,8 +4160,8 @@ public partial class DataGridView : Control, ISupportInitialize
 
     [Browsable(false)]
     [EditorBrowsable(EditorBrowsableState.Advanced)]
-    public Cursor UserSetCursor => _dataGridViewState1[State1_CustomCursorSet]
-        ? _oldCursor!
+    public Cursor? UserSetCursor => _dataGridViewState1[State1_CustomCursorSet]
+        ? _oldCursor
         : Cursor;
 
     internal int VerticalOffset
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.cs
@@ -1154,7 +1154,11 @@ public partial class DataGridView : Control, ISupportInitialize
                     _borderStyle = value;
                     if (!AutoSize)
                     {
-                        PerformLayoutPrivate(false /*useRowShortcut*/, false /*computeVisibleRows*/, true /*invalidInAdjustFillingColumns*/, true /*repositionEditingControl*/);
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.cs
```diff
@@ -4291,7 +4131,7 @@ public DataGridViewHeaderCell TopLeftHeaderCell
                 TopLeftHeaderCell = new DataGridViewTopLeftHeaderCell();
             }
 
-            return _topLeftHeaderCell;
+            return _topLeftHeaderCell!;
```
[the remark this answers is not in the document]
gpetrou: We are setting this in the setter below, so I don't think we need any checks.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.cs
```diff
@@ -3118,11 +2981,11 @@ internal bool IsSharedCellVisible(DataGridViewCell dataGridViewCell, int rowInde
             && dataGridViewCell.OwningColumn.Visible;
     }
 
-    internal ToolTip KeyboardToolTip
+    internal ToolTip? KeyboardToolTip
```
dreddy-work, line 2984: When does this return null?

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.cs
```diff
@@ -3516,7 +3380,7 @@ public DataGridViewCellStyle RowHeadersDefaultCellStyle
             _rowHeadersDefaultCellStyle = value;
             if (value is not null)
             {
-                _rowHeadersDefaultCellStyle.AddScope(this, DataGridViewCellStyleScopes.RowHeaders);
+                _rowHeadersDefaultCellStyle!.AddScope(this, DataGridViewCellStyleScopes.RowHeaders);
```
dreddy-work, line 3383: I would change this to `_rowHeadersDefaultCellStyle?.AddScope(this, DataGridViewCellStyleScopes.RowHeaders);`

dreddy-work (changes requested):

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.cs
gpetrou: `TryGetObject` will return nullable value. Without `?` there is a warning.

dreddy-work: hmm. return value in those cases should be `false`.  However, Its unfortunate that its not honoring that convention in `GetObject` call made by this API.

dreddy-work (approved):

PR #9992, opened 2023-09-26
Enable nullability in MenuCommandService

halgab:
## Proposed changes

- Enable nullability in `MenuCommandService`

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9992)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/MenuCommandService.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/MenuCommandService.cs
@@ -1,8 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-#nullable disable
-
 using System.Collections;
 using System.Globalization;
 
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms.Design/src/System/ComponentModel/Design/MenuCommandService.cs
```diff
@@ -294,23 +288,24 @@ protected void EnsureVerbs()
...
-                    verbsOrder.Add(localVerbs[i]);
-                    buildVerbs[key] = verbsOrder.Count - 1;
+                    DesignerVerb localVerb = localVerbs[i]!;
+                    verbsOrder.Add(localVerb);
+                    buildVerbs[localVerb.Text] = verbsOrder.Count - 1;
                 }
             }
 
             // look for duplicate, prepare the result table
+            //TODO: isn't 'result' too big if there are duplicates?
```
halgab, line 308: I'm worried we're missing an `Array.Resize(ref result, j);` after the loop below...

dreddy-work: Based on logic, `buildVerbs ` and `result` both will have same size.  `verbsOrder` size may be bigger (dues to duplicates).

dreddy-work (dismissed):
LGTM

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/MenuCommandService.cs
halgab: Okay, I get it now. Thanks! I removed the useless comment

dreddy-work (approved):

PR #9994, opened 2023-09-26
Don't invalidate Cursors cursors

### LINKED ISSUE #9985: Exception in .Net Core(5.0/6.0/7.0) when Corsor of Form is Disposed
### .NET version

7.0

Yes

.Net Core 3.1 works, 
.Net Core 5.0, 6.0, 7.0 doesn't work
.Net Core 8.0 seems to work

### Issue description

When the Cursor of a Form gets Disposed, then there are endless exceptions and the App may even crash.
This didn't happen in the .Net Framework, and doesn't seems to happen in .Net Core 3.1 or 8.0 Preview.
This Exception seems to happen in .Net Core 5.0, 6.0 & 7.0

As far as I tracked this back, the Problem seems that in the FW when the cursor is set, then it uses the internal Value of 'System.Windows.Forms.handle', and .Net Core 7.0 seems to use 'System.Windows.Forms.Handle'.
'System.Windows.Forms.handle' throws a exception when System.Windows.Forms.handle is IntPtr.Zero.

This is even more problematic when (acidently) any one of teh Cursors returned by the 'System.Windows.Forms.Cursors'-Properties gets Disposed, since they are used internal and there is no way to replace the Disposed Cursors with Not Disposed ones, prevent the Dispose or react to the Dispose (No Event)

### Steps to reproduce

1.) Create a New WinForms Project.
2.) Add a Button
3.) Add this Click-Event Code:
`
    Private Sub Button1_Click(sender As Object, e As EventArgs) Handles Button1.Click
        Dim cur As Cursor
        cur = System.Windows.Forms.Cursors.Default
        Me.Cursor = cur
        Me.Cursor.Dispose()
    End Sub
`
4.) Click the Button and move the Mousecursor around the Form

JeremyKuhne:
Our static Cursors should never be disposed. This was a regression for the ones wrapped around system cursors. Our own resource cursors were never properly guarded, this fixes them.

Fixes #9985

## Proposed changes

- Don't dispose Cursor instances we shouldn't dispose

## Customer Impact

- Without this any code can dispose of the Cursors statics, making them unusable
- There is no workaround without using reflection

## Regression? 

- Yes

## Risk

- Minimal


dreddy-work:
@JeremyKuhne, test is failing. Please send email to Tactics.

dreddy-work (approved):

PR #9998, opened 2023-09-27
Enable nullability in ByteViewer and remove allocations when drawing

halgab:
## Proposed changes

- Enable nullability in `ByteViewer`
- Cleanup code and remove most allocations when drawing

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9998)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ByteViewer.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ByteViewer.cs
@@ -427,7 +427,7 @@ public class ByteViewer : TableLayoutPanel
     /// </summary>
     private void InitUnicode()
     {
-        _edit.Text = string.Create(_dataBuf.Length / 2 + 1, _dataBuf, static (text, dataBuf) =>
+        _edit.Text = string.Create(_dataBuf.Length / sizeof(char) + 1, _dataBuf, static (text, dataBuf) =>
         {
             Encoding.Unicode.GetChars(dataBuf.AsSpan(), text);
             text.Replace('\0', '\v');
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ByteViewer.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ByteViewer.cs
@@ -43,7 +43,7 @@ public class ByteViewer : TableLayoutPanel
     private TextBox _edit;
     private readonly int _columnCount = DEFAULT_COLUMN_COUNT;
     private int _rowCount = DEFAULT_ROW_COUNT;
-    private byte[]? _dataBuf;
+    private byte[] _dataBuf = Array.Empty<byte>();
     private int _startLine;
     private int _displayLinesCount;
     private int _linesCount;
@@ -236,8 +236,7 @@ public class ByteViewer : TableLayoutPanel
         }
 
         int maxLength = _columnCount * 3 + 1;
-        using BufferScope<char> charsBufferScope = new(stackalloc char[256], maxLength);
-        Span<char> charsBuffer = charsBufferScope;
+        using BufferScope<char> charsBuffer = new(stackalloc char[256], maxLength);
         for (int i = 0; i < linesCount; i++)
         {
             ReadOnlySpan<byte> lineBuffer = GetLineBytes(startLine + i);
@@ -250,7 +249,7 @@ public class ByteViewer : TableLayoutPanel
         {
             int offset = (startLine + line) * _columnCount;
             int length;
-            if (offset + _columnCount > _dataBuf!.Length)
+            if (offset + _columnCount > _dataBuf.Length)
             {
                 length = _dataBuf.Length % _columnCount;
             }
@@ -280,7 +279,7 @@ public class ByteViewer : TableLayoutPanel
         int unicodeCount = 0;
         int size;
 
-        if ((_dataBuf is null) || (_dataBuf.Length >= 0 && (_dataBuf.Length < 8)))
+        if (_dataBuf.Length is >= 0 and < 8)
         {
             return DisplayMode.Hexdump;
         }
@@ -364,7 +363,7 @@ public class ByteViewer : TableLayoutPanel
     /// <summary>
     ///  Gets the bytes in the buffer.
     /// </summary>
-    public virtual byte[]? GetBytes() => _dataBuf;
+    public virtual byte[] GetBytes() => _dataBuf;
 
     /// <summary>
     ///  Gets the display mode for the control.
@@ -415,7 +414,7 @@ public class ByteViewer : TableLayoutPanel
         int bufferSize;
         fixed (byte* pDataBuff = _dataBuf)
         {
-            bufferSize = PInvoke.MultiByteToWideChar(PInvoke.CP_ACP, 0, (PCSTR)pDataBuff, _dataBuf!.Length, null, 0);
+            bufferSize = PInvoke.MultiByteToWideChar(PInvoke.CP_ACP, 0, (PCSTR)pDataBuff, _dataBuf.Length, null, 0);
             charsBuffer.EnsureCapacity(bufferSize + 1);
             fixed (char* pText = charsBuffer)
             {
@@ -436,7 +435,7 @@ public class ByteViewer : TableLayoutPanel
     /// </summary>
     private void InitUnicode()
     {
-        _edit.Text = string.Create(_dataBuf!.Length / 2 + 1, _dataBuf, static (text, dataBuf) =>
+        _edit.Text = string.Create(_dataBuf.Length / 2 + 1, _dataBuf, static (text, dataBuf) =>
         {
             Encoding.Unicode.GetChars(dataBuf.AsSpan(), text);
             text.Replace('\0', '\v');
@@ -487,7 +486,7 @@ public class ByteViewer : TableLayoutPanel
     private void InitState()
     {
         // calculate number of lines required (being careful to count 1 for the last partial line)
-        _linesCount = (_dataBuf!.Length + _columnCount - 1) / _columnCount;
+        _linesCount = (_dataBuf.Length + _columnCount - 1) / _columnCount;
 
         _startLine = 0;
         if (_linesCount > _rowCount)
@@ -600,11 +599,6 @@ public class ByteViewer : TableLayoutPanel
     /// </summary>
     public virtual void SaveToFile(string path)
     {
-        if (_dataBuf is null)
-        {
-            return;
-        }
-
         using FileStream currentFile = new(path, FileMode.OpenOrCreate, FileAccess.ReadWrite, FileShare.None);
         currentFile.Write(_dataBuf.AsSpan());
     }
@@ -625,11 +619,6 @@ public class ByteViewer : TableLayoutPanel
     public virtual void SetBytes(byte[] bytes)
     {
         ArgumentNullException.ThrowIfNull(bytes);
-
-        if (_dataBuf is not null)
-        {
-            _dataBuf = null;
-        }
 
         _dataBuf = bytes;
         InitState();
@@ -698,7 +687,7 @@ public class ByteViewer : TableLayoutPanel
     /// </summary>
     public virtual void SetStartLine(int line)
     {
-        if (line < 0 || line >= _linesCount || line > _dataBuf?.Length / _columnCount)
+        if (line < 0 || line >= _linesCount || line > _dataBuf.Length / _columnCount)
         {
             _startLine = 0;
         }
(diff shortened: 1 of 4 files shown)


src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ByteViewer.cs
```diff
@@ -258,12 +257,20 @@ private void DrawHex(Graphics g, byte[] lineBuffer, int line)
     /// </summary>
     private void DrawLines(Graphics g, int startLine, int linesCount)
     {
+        if (linesCount == 0)
+        {
+            return;
+        }
+
+        int maxLength = _columnCount * 3 + 1;
+        using BufferScope<char> charsBufferScope = new(maxLength);
```
JeremyKuhne, line 266: Another improvement you can do here is to provide a stack allocated space. Something like this provides a nice safe alternative:

```suggestion
        using BufferScope<char> charsBuffer = new(stackalloc char[256], maxLength);
```

You also don't need to assign it to a local. The pattern is to just use it and let the implicit cast operator get out the span.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ByteViewer.cs
```diff
@@ -720,28 +698,19 @@ public virtual void SetDisplayMode(DisplayMode mode)
...
-            currentFile.Close();
-        }
-        catch
-        {
-            currentFile.Close();
-            throw;
-        }
+        using FileStream currentFile = new FileStream(path, FileMode.Open, FileAccess.Read, FileShare.None);
+        int length = (int)currentFile.Length;
+        byte[] buf = new byte[length + 1];
```
JeremyKuhne, line 703: This should all be collapsed to `SetBytes(File.ReadAllBytes(path));`

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ByteViewer.cs
```diff
@@ -619,23 +613,14 @@ public virtual void SaveToFile(string path)
...
-        {
-            currentFile.Write(_dataBuf, 0, _dataBuf.Length);
-            currentFile.Close();
-        }
-        catch
-        {
-            currentFile.Close();
-            throw;
-        }
+        using FileStream currentFile = new FileStream(path, FileMode.OpenOrCreate, FileAccess.ReadWrite, FileShare.None);
```
JeremyKuhne, line 616: We prefer to use target typed new:

```suggestion
        using FileStream currentFile = new(path, FileMode.OpenOrCreate, FileAccess.ReadWrite, FileShare.None);
```

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ByteViewer.cs
```diff
@@ -450,19 +444,19 @@ private unsafe void InitAnsi()
...
-        for (int i = 0; i < text.Length; i++)
-            if (text[i] == '\0')
-                text[i] = (char)0x0B;
-
-        text[text.Length - 1] = '\0';
-        _edit.Text = new string(text);
+        _edit.Text = string.Create(_dataBuf!.Length / 2 + 1, _dataBuf, static (text, dataBuf) =>
+        {
+            Encoding.Unicode.GetChars(dataBuf.AsSpan(), text);
+            text.Replace('\0', (char)0x0B);
```
JeremyKuhne, line 450:
```suggestion
            text.Replace('\0', '\v');
```

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ByteViewer.cs
```diff
@@ -433,14 +434,7 @@ private unsafe void InitAnsi()
...
         text[size] = '\0';
-
-        for (int i = 0; i < size; i++)
-        {
-            if (text[i] == '\0')
-            {
-                text[i] = (char)0x0B;
-            }
-        }
+        text.AsSpan(0, size).Replace('\0', (char)0x0B);
```
JeremyKuhne, line 437:
```suggestion
        text.AsSpan(0, size).Replace('\0', `\v`);
```

JeremyKuhne, line 437: Consider replacing `text` with `BufferScope<char>`.

JeremyKuhne (changes requested):
Nice! I've added a few comments.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ByteViewer.cs
halgab: I assigned it to a local because I wanted to avoid having the implicit cast happening over and over inside the loop, but I can remove it if you think that's overkill

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ByteViewer.cs
halgab: Will do. Just 2 things on this:

1. I don't get why `byte[]` wasn't set to `new byte[length]`
2. We are going to fix a potential bug making the change, as `currentFile.Read` can very well read less than `length`, and we never check for that scenario. I've seen this same problem at several other places in the repo btw

halgab: This change makes tests fail:

 ```
    System.ComponentModel.Design.Tests.ByteViewerTests.ByteViewer_SetFile_InvokeWithBytes_Success [FAIL]
      Assert.Equal() Failure
                           (pos 3)
      Expected: [1, 2, 3, 0]
      Actual:   [1, 2, 3]
      Stack Trace:
        /_/src/System.Windows.Forms.Design/tests/UnitTests/System/ComponentModel/Design/ByteViewerTests.cs(896,0): at System.ComponentModel.Design.Tests.ByteViewerTests.ByteViewer_SetFile_InvokeWithBytes_Success()
           at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
    System.ComponentModel.Design.Tests.ByteViewerTests.ByteViewer_SetFile_InvokeNoBytes_Success [FAIL]
      Assert.Equal() Failure
                           (pos 3)
      Expected: [1, 2, 3, 0]
      Actual:   [1, 2, 3]
      Stack Trace:
        /_/src/System.Windows.Forms.Design/tests/UnitTests/System/ComponentModel/Design/ByteViewerTests.cs(885,0): at System.ComponentModel.Design.Tests.ByteViewerTests.ByteViewer_SetFile_InvokeNoBytes_Success()
           at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
```

I guess my first point above was relevant. Should I roll this back @JeremyKuhne?

Comment (JeremyKuhne) [reply src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ByteViewer.cs]: @halgab Sorry, didn't notice that it was putting a trailing `0` into the buffer. That seems wrong, but there is a lot of wrong here. Opening and saving the same file name is going to add a byte that doesn't exist. That said, saving to the file doesn't reset the length of the file, which doesn't seem sensible (that you could end up with data after the buffer).

I'm inclined to fix this and change the test. It makes the file usage symmetrical, and it won't change the behavior of `BinaryEditor`.

File: src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ByteViewer.cs (hunk 1 above)
Comment (JeremyKuhne) [reply src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ByteViewer.cs]: Yeah, it isn't going to measure in the scope of the cost of rendering here, so I'd go with simpler.

File: src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ByteViewer.cs [hunk 6]
```diff
@@ -45,7 +43,7 @@ public class ByteViewer : TableLayoutPanel
     private TextBox _edit;
     private readonly int _columnCount = DEFAULT_COLUMN_COUNT;
     private int _rowCount = DEFAULT_ROW_COUNT;
-    private byte[] _dataBuf;
+    private byte[]? _dataBuf;
```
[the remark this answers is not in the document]
JeremyKuhne: This one I think is pretty important.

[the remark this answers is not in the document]
halgab: Yes, I was preparing it for after my first batch of changes. But if I make the change, the behaviour of `SaveToFile` will change slightly. Today it's a no-op when we haven't called `SetBytes`. Now it's going to dump an empty file. It's minor, but that's still a change

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ByteViewer.cs
```diff
@@ -258,12 +230,35 @@ private void DrawHex(Graphics g, byte[] lineBuffer, int line)
...
             DrawAddress(g, startLine, i);
-            DrawHex(g, lineBuffer, i);
-            DrawDump(g, lineBuffer, i);
+            DrawHex(g, lineBuffer, i, charsBuffer);
+            DrawDump(g, lineBuffer, i, charsBuffer);
+        }
+
+        ReadOnlySpan<byte> GetLineBytes(int line)
+        {
+            int offset = (startLine + line) * _columnCount;
```
JeremyKuhne, line 250: `startLine` shouldn't be used here

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ByteViewer.cs
```diff
@@ -258,12 +230,35 @@ private void DrawHex(Graphics g, byte[] lineBuffer, int line)
...
+            int length;
+            if (offset + _columnCount > _dataBuf.Length)
```
JeremyKuhne, line 252: Consider using the ternary operator.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ByteViewer.cs
halgab: Oops... Fixed

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ByteViewer.cs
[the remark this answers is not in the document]
halgab: The change stated above makes the tests fail. Should I try to preserve existing behaviour this time, or should I go ahead and fix the tests?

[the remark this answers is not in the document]
halgab: @JeremyKuhne the current behaviour of method `GetBytes` is to return `null` if `SetBytes` has never been called. Would you like me to maintain this behaviour or should the method return an empty array from now on?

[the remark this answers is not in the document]
halgab: I went ahead and change the test, assuming we're happy we the behaviour change.

halgab:
This one is still waiting on team @lonitra, @JeremyKuhne

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ByteViewer.cs
```diff
@@ -450,19 +427,19 @@ private unsafe void InitAnsi()
...
     {
-        char[] text = new char[_dataBuf.Length / 2 + 1];
-        Encoding.Unicode.GetChars(_dataBuf, 0, _dataBuf.Length, text, 0);
-        for (int i = 0; i < text.Length; i++)
-            if (text[i] == '\0')
-                text[i] = (char)0x0B;
-
-        text[text.Length - 1] = '\0';
-        _edit.Text = new string(text);
+        _edit.Text = string.Create(_dataBuf.Length / 2 + 1, _dataBuf, static (text, dataBuf) =>
```
JeremyKuhne, line 430:
```suggestion
        _edit.Text = string.Create(_dataBuf.Length / sizeof(char) + 1, _dataBuf, static (text, dataBuf) =>
```

JeremyKuhne: @halgab just this one nit left

halgab: Thanks! So you're fine with https://github.com/dotnet/winforms/pull/9998#discussion_r1355711840?

JeremyKuhne: Yes. Sorry about the delay on this, I was out for most of October.

JeremyKuhne (approved):

PR #10005, opened 2023-09-27
Simplify Cursor code

JeremyKuhne:
Construction of Cursors statics had gotten a bit over complicated. ToString() was also very expensive when creating Cursors from handles.

The core of the change is to pass the property name when constructing a Cursor from Cursors for consistency and performance.

These changes will also allow making the CursorConverter reflection free in the future.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10005)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Cursor.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Cursor.cs
@@ -459,8 +459,6 @@ public sealed class Cursor : IDisposable, ISerializable, IHandle<HICON>, IHandle
         return (byte[])_cursorData.Clone();
     }
 
-    internal bool IsValid() => _handle != IntPtr.Zero;
-
     /// <summary>
     ///  Displays the cursor. For every call to Cursor.show() there must have been
     ///  a previous call to Cursor.hide().
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Cursor.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Cursor.cs
@@ -26,6 +26,9 @@ public sealed class Cursor : IDisposable, ISerializable, IHandle<HICON>, IHandle
     private HCURSOR _handle;
     private readonly bool _freeHandle;
 
+    /// <summary>
+    ///  If created by the <see cref="Cursors"/> class, this is the property name that created it.
+    /// </summary>
     internal string? CursorsProperty { get; }
 
     internal unsafe Cursor(PCWSTR nResourceId, string cursorsProperty)
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Cursor.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Cursor.cs
@@ -24,22 +24,14 @@ public sealed class Cursor : IDisposable, ISerializable, IHandle<HICON>, IHandle
 
     private readonly byte[]? _cursorData;
     private HCURSOR _handle;
-    private readonly Flags _flags;
-    private readonly PCWSTR _resourceId;
+    private readonly bool _freeHandle;
 
-    private bool OwnHandle => (_flags & Flags.OwnHandle) != 0;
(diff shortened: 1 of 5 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Cursor.cs
```diff
@@ -24,41 +24,40 @@ public sealed class Cursor : IDisposable, ISerializable, IHandle<HICON>, IHandle
 
     private readonly byte[]? _cursorData;
     private HCURSOR _handle;
-    private readonly Flags _flags;
-    private readonly PCWSTR _resourceId;
+    private readonly bool _freeHandle;
 
-    private bool OwnHandle => (_flags & Flags.OwnHandle) != 0;
-    private bool IsWellKnown => (_flags & Flags.InternalCursor) != 0;
+    internal string? CursorsProperty { get; }
```
Tanya-Solyanik, line 29: Please add a comment to this property - what does it contain?

src/System.Windows.Forms/src/System/Windows/Forms/Cursor.cs
```diff
@@ -77,14 +76,8 @@ public Cursor(string fileName)
     ///  Initializes a new instance of the <see cref="Cursor"/> class from the specified <paramref name="resource"/>.
     /// </summary>
     public Cursor(Type type, string resource)
-        : this((type.OrThrowIfNull()).Module.Assembly.GetManifestResourceStream(type, resource)!)
+        : this(type.OrThrowIfNull().Module.Assembly.GetManifestResourceStream(type, resource)!)
     {
-        _flags |= Flags.OwnHandle;
```
Tanya-Solyanik, line 82: Not related to this change, but I don't understand why this case was marked as "OwnHandle" in the past and is marked as _freeHandle== true now. This could be a system assembly or the WinForms assembly. and multiple instances of this cursor could be used in the same exe....

lonitra (approved):
:shipit:

src/System.Windows.Forms/src/System/Windows/Forms/Cursor.cs
JeremyKuhne: `_freeHandle` is just way clearer. `_ownHandle` meant that you should free the handle, that is all it meant.

src/System.Windows.Forms/src/System/Windows/Forms/Cursor.cs
```diff
@@ -94,16 +90,17 @@ public Cursor(Type type, string resource)
     public Cursor(Stream stream)
     {
         ArgumentNullException.ThrowIfNull(stream);
-        MemoryStream memoryStream = new();
-        // reset stream position to start, there are no gaurantees it is at the start.
+        using MemoryStream memoryStream = new();
+
+        // Reset stream position to start, there are no gaurantees it is at the start.
```
dreddy-work, line 95: Does it make sense to restore stream position to what it was before , after copying to local?

JeremyKuhne: I wouldn't as some users might be depending on it to be in a specific position.

dreddy-work (approved):

src/System.Windows.Forms/src/System/Windows/Forms/Cursor.cs
Tanya-Solyanik: That was not my question. What is the difference between line 45 and line 80, both are loading manifestresourcestream. One from an arbitrary assemby, another from the winforms assembly. Why are we freeing cursor loaded from an arbitrary assembly when that assembly could be the winforms assembly?

JeremyKuhne: Because it only matters that we don't unload the ones from our own static `Cursors` class.

Olina-Zhang:
Tested controls with different cursor value setting based on main branch of Winforms repo, it works well.

PR #10009, opened 2023-09-28

gpetrou:
## Proposed changes

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10009)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyGridView.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyGridView.cs
@@ -498,7 +498,7 @@ internal sealed partial class PropertyGridView :
                 }
             }
 
-            GridEntry equivalentEntry = FindEquivalentGridEntry(new GridEntryCollection(new[] { value }, disposeItems: false));
+            GridEntry? equivalentEntry = FindEquivalentGridEntry(new GridEntryCollection(new[] { value }, disposeItems: false));
 
             if (equivalentEntry is not null)
             {
@@ -591,7 +591,7 @@ internal sealed partial class PropertyGridView :
     /// </summary>
     internal GridEntryCollection? TopLevelGridEntries { get; private set; }
 
-    internal GridEntryCollection AccessibilityGetGridEntries() => GetAllGridEntries();
+    internal GridEntryCollection? AccessibilityGetGridEntries() => GetAllGridEntries();
 
     internal Rectangle AccessibilityGetGridEntryBounds(GridEntry gridEntry)
     {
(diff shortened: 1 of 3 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyGridView.cs
```diff
@@ -2297,26 +2297,26 @@ protected override void OnHandleDestroyed(EventArgs e)
...
     }
 
-    private void OnListMouseUp(object sender, MouseEventArgs me)
+    private void OnListMouseUp(object? sender, MouseEventArgs me)
     {
         OnListClick(sender, me);
     }
 
-    private void OnListClick(object sender, EventArgs e)
+    private void OnListClick(object? sender, EventArgs e)
```
dreddy-work, line 2319: why not `private void OnListClick(object? sender, EventArgs? e)` and avoid `!` on caller?  Its privat emethod.

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyGridView.cs
```diff
@@ -2405,11 +2405,11 @@ protected override void OnLostFocus(EventArgs e)
 
         if (FocusInside)
         {
-            base.OnLostFocus(e);
+            base.OnLostFocus(e!);
```
dreddy-work, line 2408: I don't see `e` being null here? why `!`?

dreddy-work: And also condition on line 2401.

gpetrou: There are warnings without these. I guess due to `if (e is not null)` above. We had to use `!` in some similar cases in the past.

gpetrou: Check `OnMouseWheel` in `ScrollBar` class for example.

dreddy-work: We need to revisit those cases with a practical scenario here.

If `e` can be null, `OnLostFocus` should be annotated so. Reading this code as is doesn't sound right. If base isn't expecting null, but override method expects null and doesn't annotate.

Can you test the scenario and see if there is ever a case where `e` could be null here?

gpetrou: Check https://github.com/dotnet/winforms/pull/7491/files#r928648176 for a related discussion. I could add the same comment here.

dreddy-work: Thank you for pointing me to this. Given it was already discussed and finalized, lets go with comment (easy to revisit if we decided to make breaking change).

dreddy-work (approved):

PR #10012, opened 2023-09-29
Enable nullability in HtmlWindow

gpetrou:
## Proposed changes

- Enable nullability in HtmlWindow.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10012)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/HtmlWindow.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/HtmlWindow.cs
@@ -448,20 +448,18 @@ public sealed unsafe partial class HtmlWindow
 
     public static unsafe bool operator ==(HtmlWindow? left, HtmlWindow? right)
     {
-        // Not equal if only one's null.
-        if (left is null != right is null)
+        if (left is null)
+        {
+            return right is null;
+        }
+
+        if (right is null)
         {
             return false;
         }
 
-        // Equal if both are null.
-        if (left is null)
-        {
-            return true;
-        }
-
         // Neither are null. Compare their native pointers.
-        return left.NativeHtmlWindow.IsSameNativeObject(right!.NativeHtmlWindow);
+        return left.NativeHtmlWindow.IsSameNativeObject(right.NativeHtmlWindow);
     }
 
     public static bool operator !=(HtmlWindow? left, HtmlWindow? right) => !(left == right);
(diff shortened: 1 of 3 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/HtmlWindow.cs
```diff
@@ -53,23 +51,18 @@ private HtmlWindowShim WindowShim
...
-                HtmlWindowShim shim = ShimManager.GetWindowShim(this);
-                if (shim is null)
-                {
-                    _shimManager.AddWindowShim(this);
-                    shim = ShimManager.GetWindowShim(this);
-                }
-
-                return shim;
+                _shimManager.AddWindowShim(this);
+                shim = ShimManager.GetWindowShim(this);
```
dreddy-work, line 58: So you don't get warning here but on line 61, by analyzer, to use `!` ?

src/System.Windows.Forms/src/System/Windows/Forms/HtmlWindow.cs
```diff
@@ -482,12 +461,12 @@ public void ScrollTo(Point point)
         }
 
         // Neither are null. Compare their native pointers.
-        return left.NativeHtmlWindow.IsSameNativeObject(right.NativeHtmlWindow);
+        return left.NativeHtmlWindow.IsSameNativeObject(right!.NativeHtmlWindow);
```
dreddy-work, line 464: Seems this `==` isn't right according to comment on line 484.  `right `could be `null `here.

ex:  `left `is not `null `and `right` is `null`.

src/System.Windows.Forms/src/System/Windows/Forms/HtmlWindow.cs
gpetrou: Yes, because `GetWindowShim` returns a nullable.
But we have used `AddWindowShim` in the previous line, so we can use the `!`.

src/System.Windows.Forms/src/System/Windows/Forms/HtmlWindow.cs
gpetrou: I added a separate commit with an alternative implementation to avoid the `!`.

dreddy-work:
Same here. We have to keep an eye if equality operator chnage break any existing applications.

dreddy-work (dismissed):

dreddy-work:
@gpetrou, can you rebase to unblock merge?

gpetrou:
> can you rebase to unblock merge?

@dreddy-work done.

dreddy-work (approved):

PR #10016, opened 2023-09-29
Enable nullability in DataGridViewLinkColumn

gpetrou:
## Proposed changes

- Enable nullability in DataGridViewLinkColumn.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10016)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewImageColumn.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewImageColumn.cs
@@ -80,20 +80,22 @@ public class DataGridViewImageColumn : DataGridViewColumn
 
         set
         {
-            if (Description != value)
+            if (Description == value)
             {
-                ImageCellTemplate.Description = value;
-                if (DataGridView is not null)
+                return;
+            }
+
+            ImageCellTemplate.Description = value;
+            if (DataGridView is not null)
+            {
+                DataGridViewRowCollection dataGridViewRows = DataGridView.Rows;
+                int rowCount = dataGridViewRows.Count;
+                for (int rowIndex = 0; rowIndex < rowCount; rowIndex++)
                 {
-                    DataGridViewRowCollection dataGridViewRows = DataGridView.Rows;
-                    int rowCount = dataGridViewRows.Count;
-                    for (int rowIndex = 0; rowIndex < rowCount; rowIndex++)
+                    DataGridViewRow dataGridViewRow = dataGridViewRows.SharedRow(rowIndex);
+                    if (dataGridViewRow.Cells[Index] is DataGridViewImageCell dataGridViewCell)
                     {
-                        DataGridViewRow dataGridViewRow = dataGridViewRows.SharedRow(rowIndex);
-                        if (dataGridViewRow.Cells[Index] is DataGridViewImageCell dataGridViewCell)
-                        {
-                            dataGridViewCell.Description = value;
-                        }
+                        dataGridViewCell.Description = value;
                     }
                 }
             }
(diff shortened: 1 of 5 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewLinkColumn.cs
```diff
@@ -101,7 +100,7 @@ public LinkBehavior LinkBehavior
         {
             if (!LinkBehavior.Equals(value))
             {
-                ((DataGridViewLinkCell)CellTemplate).LinkBehavior = value;
+                ((DataGridViewLinkCell)CellTemplate!).LinkBehavior = value;
```
dreddy-work, line 103: can you add Debug assert on `CellTemplate` and see if we can avoid `!`?

gpetrou: Not sure if it makes much sense to do that here. We already throw in the getter. We have used null-forgiving operator in other cases.

dreddy-work: If Debug.Assert help avoid using `!`, that should be done. Also, it helps while debugging by asserting right away rather than failing down the line.

gpetrou: What I mean is that in line 101 we already have `if (!LinkBehavior.Equals(value))`. This will call line 92 `if (CellTemplate is null)` and will throw an exception if that is the case. So, not sure I understand why we need an assert. You want to place one before the if statement? Sounds like something we have not done before.

lonitra: Perhaps we should add `MemberNotNullAttribute` to `LinkBehavior` and any other properties that check for `CellTemplate` being null and throws if so. That way we can get rid of the `!`. We have some similar instances where we do this e.g.: https://github.com/dotnet/winforms/blob/e6d1ad98adb7825970935942de483273912c82f1/src/System.Windows.Forms/src/System/Windows/Forms/RelatedCurrencyManager.cs#L25-L39

gpetrou: I was actually thinking about that before. I updated this and some other already annotated classes for consistency.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewLinkColumn.cs
```diff
@@ -343,22 +347,15 @@ public override object Clone()
         }
         else
         {
-            //
-
-            dataGridViewColumn = (DataGridViewLinkColumn)System.Activator.CreateInstance(thisType);
+            dataGridViewColumn = (DataGridViewLinkColumn)Activator.CreateInstance(thisType)!;
         }
 
-        if (dataGridViewColumn is not null)
```
lonitra, line 351: Is it possible that we get `null` for `dataGridViewColumn` in the else statement above?

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewImageColumn.cs
```diff
@@ -65,22 +65,35 @@ public override DataGridViewCellStyle DefaultCellStyle
...
+            }
+
+            return ImageCellTemplate.Description;
+        }
+
         set
         {
-            ImageCellTemplate.Description = value;
-            if (DataGridView is not null)
+            if (Description != value)
```
lonitra, line 83: Consider inverting this

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewImageColumn.cs
```diff
@@ -112,26 +125,21 @@ public string Description
...
-        get
-        {
-            if (CellTemplate is null)
-            {
-                throw new InvalidOperationException(SR.DataGridViewColumn_CellTemplateRequired);
-            }
-
-            return (DataGridViewImageCell)CellTemplate;
-        }
-    }
```
lonitra, line 126: Sorry, I did not see this before, but I actually think it would be cleaner if we keep it like this and do the same sort of pattern for `ButtonCellTemplate` and `LinkCellTemplate` in the other files. This way I don't think we need to add the `MemberNotNullAttribute` everywhere and most property getters would now be one liners.

gpetrou: This means that we check for null twice, correct?

lonitra: We would only check for null once in `ImageCellTemplate`, `ButtonCellTemplate`, and `LinkCellTemplate`, throwing if `CellTemplate` is null. There is no need to check for null again in the other properties that are using either `ImageCellTemplate`, `ButtonCellTemplate` or `LinkCellTemplate`

gpetrou: But the setter calls the getter.

lonitra: Ah gotcha I missed that, nevermind then please ignore this suggestion :)

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewLinkColumn.cs
gpetrou: I don't think so. We have done similar changes in other files.

Issue #9795, opened 2023-08-28
Make AccessibleObject trimmable


JeremyKuhne:
In order for AOT/Trimming to be a viable option we need to change AccessibleObject to fully utilize native VARIANTs and pointers and use ComWrappers for our CCW projections.

There are a number of steps to this, including:

- Implement the CsWin32 IAccessible.Interface on AccessibleObject and have the existing public IAccessible interface projection just wrap this.
- Convert the UIA interfaces to use CsWin32.
- Ensure that we're using IManagedWrapper on all of our accessible objects.

`AccessibleObject` is a little difficult to make efficient due to it being public with numerous virtual members.

elachlan:
Related: #9784

PR #10022, opened 2023-09-29
Convert ITextRangeProvider/ITextProvider to Cswin32

lonitra:
- Move references of `ITextProvider`, `ITextProvider2`, `ITextRangeProvider` to Cswin32
- Add Improvements to `SafeArrayScope` to be able to receive `SAFEARRAY*` (similar to `ComScope` pattern)
- Add `ComSafeArrayScope` for better usability when we need a `SAFEARRAY` that contains COM interfaces (`SafeArrayScope<nint>`)
- Add explicit operators for converting to/from `VARIANT` and `double`
- Add implicit operator to convert to/from `UiaPoint` and `Point`
- Update tests

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10022)


src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextProvider.cs
```diff
@@ -101,8 +119,34 @@ public static double[] RectListToDoubleArray(List<Rectangle> rectArray)
     ///  https://learn.microsoft.com/windows/win32/api/uiautomationcore/nf-uiautomationcore-irawelementproviderfragment-get_boundingrectangle
     ///  https://learn.microsoft.com/windows/win32/api/uiautomationcore/ns-uiautomationcore-uiarect
     /// </summary>
-    internal static double[] BoundingRectangleAsArray(Rectangle bounds)
-        => new double[] { bounds.X, bounds.Y, bounds.Width, bounds.Height };
+    internal static VARIANT BoundingRectangleAsArray(Rectangle bounds)
+    {
+        SAFEARRAYBOUND saBound = new()
```
lonitra, line 124: Thinking about using `SafeArrayScope` as it would leaving us with the `SAFEARRAY` creation as simple as:
```c#
SafeArrayScope<double> array = new(size: 4);
array[0] = bounds.X;
array[1] = bounds.Y;
array[2] = bounds.Width;
array[3] = bounds.Height;
// ...
```
there are other instances in this PR where we can make it more simple like this, but I am a bit hesitant since our pattern has been to add a `using` statement with scopes, but in this case we should not have a using statement since we are returning this..

src/System.Windows.Forms.Primitives/src/Interop/OleAut32/Interop.VARIANT.cs
```diff
@@ -951,6 +951,18 @@ private static Span<T> GetSpan<T>(Array array)
     public static explicit operator IUnknown*(VARIANT value)
         => value.vt == VT_UNKNOWN ? value.data.punkVal : throw new InvalidCastException();
 
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static explicit operator double(VARIANT value)
+        => value.data.dblVal;
```
JeremyKuhne, line 956: This should have the same type check as the other cast operators.

src/System.Windows.Forms.Primitives/src/Interop/OleAut32/Interop.VARIANT.cs
```diff
@@ -984,6 +996,10 @@ public static VARIANT FromObject(object? value)
         {
             return (VARIANT)uintValue;
         }
+        else if (value is double dblVal)
```
JeremyKuhne, line 999: nit: `doubleValue`

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextRange.cs
```diff
@@ -3,21 +3,22 @@
...
 using Windows.Win32.System.Variant;
+using Windows.Win32.UI.Accessibility;
 #endif
 using Windows.Win32.UI.Input.KeyboardAndMouse;
-using static Interop.UiaCore;
 
 namespace System.Windows.Forms.Automation;
 
-internal class UiaTextRange : ITextRangeProvider
+internal unsafe class UiaTextRange : ITextRangeProvider.Interface
```
lonitra, line 15: We have an `AgileComPointer` that will not get disposed here. Does this need to implement `IDisposable`? No where that uses this implements `IDisposable` either

src/System.Windows.Forms.Primitives/tests/UnitTests/Windows/Win32/System/Com/SafeArrayScopeTests.cs
```diff
@@ -53,6 +53,6 @@ public void SafeArrayScope_Dispose_Success()
         Assert.False(scope.IsNull);
 
         scope.Dispose();
-        Assert.True(scope.IsNull);
+        Assert.False(scope.IsNull);
```
JeremyKuhne, line 56: This should still be true.

src/System.Windows.Forms.Primitives/src/Windows/Win32/System/Com/SafeArrayScope.cs
```diff
@@ -44,70 +63,117 @@ public SafeArrayScope(uint size)
...
+        {
+            return;
+        }
+
+        SAFEARRAY* safeArray = (SAFEARRAY*)_value;
+
+        // Really want _value to be null after disposal to avoid double destroy, but we also want
+        // to maintain the readonly state of the struct to allow passing as `in` without creating implicit
+        // copies (which would break the T** and void** operators).
+        *(void**)_value = null;
```
JeremyKuhne, line 163: This needs to be the `this` pointer.

```suggestion
        *(void**)this = null;
```

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextRange.cs
JeremyKuhne, line 15: If classes aren't derived, they should be sealed. Additionally, this should be able to get `IManagedWrapper` now I think.

src/System.Windows.Forms.Primitives/src/Windows/Win32/ComHelpers.cs
```diff
@@ -135,76 +135,70 @@ internal static unsafe partial class ComHelpers
...
-                Debug.WriteLine($"{nameof(TryGetManagedInterface)}: Found a manual CCW, but couldn't unwrap to {typeof(TWrapper).Name}");
+                Debug.WriteLine($"{nameof(TryGetObjectForIUnknown)}: Found a manual CCW, but couldn't unwrap to {typeof(TWrapper).Name}");
             }
         }
 
         @interface = default;
         return false;
     }
 
+#pragma warning disable CS1574 // XML comment has cref attribute 'TryGetObjectForIUnknown{TObject}(IUnknown*, bool, out TObject?)' that could not be resolved
```
JeremyKuhne, line 161: Does Intellisense work for the method or not? If it doesn't then we should figure out how to fix this or manually add the comments.

lonitra: Intellisense can find the method, but for some reason it is not resolving the cref. I can look into more how this can be resolved or just add the comments manually if nothing else.

JeremyKuhne (approved):
Good work!

PR #10023, opened 2023-09-30
Enable nullability in some CurrencyManager members

gpetrou:
## Proposed changes

- Enable nullability in some CurrencyManager members.


src/System.Windows.Forms/src/System/Windows/Forms/CurrencyManager.cs
```diff
@@ -515,20 +517,20 @@ private void FindGoodRow()
     /// </summary>
     internal void SetSort(PropertyDescriptor property, ListSortDirection sortDirection)
     {
-        if (_list is IBindingList && ((IBindingList)_list).SupportsSorting)
+        if (_list is IBindingList bindingList && bindingList.SupportsSorting)
```
lonitra, line 520: We want to start using this syntax:
```suggestion
        if (_list is IBindingList { SupportsSorting: true } bindingList)
```
Ditto in a few places below

src/System.Windows.Forms/src/System/Windows/Forms/CurrencyManager.cs
```diff
@@ -577,46 +579,39 @@ internal int Find(PropertyDescriptor property, object key, bool keepIndex)
...
+            : finalType!.Name;
 
     /// <summary>
     ///  Gets the name of the specified list.
     /// </summary>
-    protected internal override string GetListName(ArrayList listAccessors)
+    protected internal override string GetListName(ArrayList? listAccessors)
     {
-        if (_list is ITypedList)
+        if (listAccessors is null)
```
lonitra, line 592: Thanks for fixing this bug!

src/System.Windows.Forms/src/System/Windows/Forms/CurrencyManager.cs
```diff
@@ -317,10 +317,15 @@ public override void CancelCurrentEdit()
...
+
+    private void ChangeRecordState(
+        int newPosition,
+        bool validating,
+        bool endCurrentEdit,
+        bool firePositionChange,
+        bool pullData)
     {
-        if (newPosition == -1 && _list.Count == 0)
+        if (newPosition == -1 && _list!.Count == 0)
```
lonitra, line 328: I'm not sure that `_list` is meant to be nullable in this class. In `SetDataSource`, which is called from constructor, if we cannot set `_list` to something nonnull, we end up throwing.

gpetrou: I made this non-nullable, but one way or the other we have to use `_list = null!` in a couple of places.

PR #10031, opened 2023-10-03
Enable nullability in some Binding members

gpetrou:
## Proposed changes

- Enable nullability in some Binding members.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10031)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Binding.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Binding.cs
@@ -573,30 +573,30 @@ public partial class Binding
 
     internal bool ControlAtDesignTime()
     {
-        if (!(BindableComponent is IComponent comp))
+        if (BindableComponent is not IComponent comp)
         {
             return false;
         }
 
         return comp.Site?.DesignMode ?? false;
     }
-#nullable disable
-    private object GetDataSourceNullValue(Type type)
+
+    private object? GetDataSourceNullValue(Type? type)
     {
         return _dsNullValueSet ? _dsNullValue : Formatter.GetDefaultDataSourceNullValue(type);
     }
 
-    private object GetPropValue()
+    private object? GetPropValue()
     {
-        if (_propIsNullInfo is not null && (bool)_propIsNullInfo.GetValue(BindableComponent))
+        if (_propIsNullInfo is not null && (bool?)_propIsNullInfo.GetValue(BindableComponent) == true)
         {
             return DataSourceNullValue;
         }
 
-        return _propInfo.GetValue(BindableComponent) ?? DataSourceNullValue;
+        return _propInfo!.GetValue(BindableComponent) ?? DataSourceNullValue;
     }
 
-    private BindingCompleteEventArgs CreateBindingCompleteEventArgs(BindingCompleteContext context, Exception ex)
+    private BindingCompleteEventArgs CreateBindingCompleteEventArgs(BindingCompleteContext context, Exception? ex)
     {
         bool cancel = false;
         string errorText;
@@ -612,7 +612,7 @@ public partial class Binding
         else
         {
             // If data error info on data source for this binding, report that
-            errorText = _bindToObject.DataErrorText;
+            errorText = _bindToObject!.DataErrorText;
 
             // We should not cancel with an IDataErrorInfo error - we didn't in Everett
             if (!string.IsNullOrEmpty(errorText))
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Binding.cs
```diff
@@ -614,7 +614,7 @@ private BindingCompleteEventArgs CreateBindingCompleteEventArgs(BindingCompleteC
         else
         {
             // If data error info on data source for this binding, report that
-            errorText = _bindToObject.DataErrorText;
+            errorText = _bindToObject!.DataErrorText;
```
lonitra, line 617: Is possible for `_bindToObject` to be null in this class? It looks like we set it in the constructor. The only place I see it being null is in private constructor, but I don't see any references to that.

src/System.Windows.Forms/src/System/Windows/Forms/Binding.cs
```diff
@@ -575,30 +575,30 @@ private void CheckBinding()
...
+    private object? GetPropValue()
     {
-        if (_propIsNullInfo is not null && (bool)_propIsNullInfo.GetValue(BindableComponent))
+        if (_propIsNullInfo is not null && (bool?)_propIsNullInfo.GetValue(BindableComponent) == true)
         {
             return DataSourceNullValue;
         }
 
-        return _propInfo.GetValue(BindableComponent) ?? DataSourceNullValue;
+        return _propInfo!.GetValue(BindableComponent) ?? DataSourceNullValue;
```
lonitra, line 598: How do we know `_propInfo` is nonnull here?

src/System.Windows.Forms/src/System/Windows/Forms/Binding.cs
gpetrou: I used `_bindToObject = null!` in the private constructor instead to avoid the warning.

src/System.Windows.Forms/src/System/Windows/Forms/Binding.cs
gpetrou: I replaced this with null propagation.

src/System.Windows.Forms/src/System/Windows/Forms/Binding.cs
```diff
@@ -573,30 +574,30 @@ private void CheckBinding()
...
+    private object? GetDataSourceNullValue(Type? type)
     {
         return _dsNullValueSet ? _dsNullValue : Formatter.GetDefaultDataSourceNullValue(type);
     }
 
-    private object GetPropValue()
+    private object? GetPropValue()
     {
-        if (_propIsNullInfo is not null && (bool)_propIsNullInfo.GetValue(BindableComponent))
+        if (_propIsNullInfo is not null && (bool?)_propIsNullInfo.GetValue(BindableComponent) == true)
```
halgab, line 592: Why do you cast to `bool`? Shouldn't this be `(bool)_propIsNullInfo.GetValue(BindableComponent)!`?

src/System.Windows.Forms/src/System/Windows/Forms/Binding.cs
```diff
@@ -677,16 +678,16 @@ protected virtual void OnFormat(ConvertEventArgs cevent)
         }
     }
 
-    private object ParseObject(object value)
+    private object? ParseObject(object? value)
     {
-        Type type = _bindToObject.BindToType;
+        Type? type = _bindToObject.BindToType;
```
halgab, line 683: `type` always seems to be used with a `!`. Do we want to add an assert here?

src/System.Windows.Forms/src/System/Windows/Forms/Binding.cs
```diff
@@ -162,6 +162,7 @@ public Binding(string propertyName, object? dataSource, string? dataMember)
     /// </summary>
     private Binding()
```
lonitra, line 163: If this is not used anywhere, I think it is safe to remove.

src/System.Windows.Forms/src/System/Windows/Forms/Binding.cs
gpetrou: What is the difference?

halgab: What bothers me is that `bool?` is not the same type as `bool`. Perhaps the code is eventually lowered to the same thing though.

lonitra (approved):

PR #9422, opened 2023-07-04
Designer action panel line refactoring

halgab:
Depends on #9367. Only the last commit is not in the quoted PR and should be reviewed here.

As dicussed [here](https://github.com/dotnet/winforms/pull/9367#discussion_r1251266055), we're attempting a refactoring to limit the use of the `!` operator when annotating `DesignerActionPanel.Line` objects.

The idea is to stop creating `Line` objects early without initialising them and start creating then only when we are actually going to use them. To do so we change the `LineInfo` type to be a factory type.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9422)

PR #10035, opened 2023-10-03
Refactor DesignerActionPanel

halgab:
As discussed in #9422 @dreddy-work 

## Proposed changes

- Refactor `DesignerActionPanel` `Line` and `LineInfo` objects to increase the number of immutable fields in `Line` objects. The goal is to have a design that works better with NRT annotations

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10035)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionPanel.CheckBoxPropertyLine.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionPanel.CheckBoxPropertyLine.cs
@@ -12,14 +12,10 @@ internal sealed partial class DesignerActionPanel
 {
     private sealed class CheckBoxPropertyLine : PropertyLine
     {
-        private CheckBox _checkBox;
+        private readonly CheckBox _checkBox;
 
-        public CheckBoxPropertyLine(IServiceProvider serviceProvider, DesignerActionPanel actionPanel)
+        private CheckBoxPropertyLine(IServiceProvider serviceProvider, DesignerActionPanel actionPanel)
             : base(serviceProvider, actionPanel)
-        {
-        }
-
-        protected override void AddControls(List<Control> controls)
         {
             _checkBox = new CheckBox
             {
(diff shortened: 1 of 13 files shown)


src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionPanel.CheckBoxPropertyLine.cs
```diff
@@ -66,5 +62,15 @@ protected override void OnValueChanged()
         {
             _checkBox.Checked = (bool)Value;
         }
+
+        public sealed class Info(DesignerActionList list, DesignerActionPropertyItem item) : PropertyLineInfo(list, item)
```
Tanya-Solyanik, line 66: Please move nested class into a dedicated file

halgab: I don't think that's a good idea here. I've designed the refactoring in such a way that each `Line` object has its own nested LineInfo class, and putting those classes in separate files will make us introduce several new files while degrading the readability of the code.
Of course, if you insist we should do it, I'll make the change. I'll just wait for you to confirm

Tanya-Solyanik: The reason I suggested it was that the abstract base class, LineInfo, is in a dedicated file already and that seemed inconsistent to me. @dreddy-work  - what is your vote?

Tanya-Solyanik: @halgab - ok, let's keep your change as is.

Tanya-Solyanik (approved):

lonitra (approved):

PR #10038, opened 2023-10-03

gpetrou:
## Proposed changes

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10038)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyGridView.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyGridView.cs
@@ -3255,8 +3255,9 @@ internal sealed partial class PropertyGridView :
             // Equivalent to large change scrolls
             if (fullNotches != 0)
             {
+                Debug.Assert(_scrollBar is not null);
                 int originalOffset = initialOffset;
-                int large = fullNotches * _scrollBar!.LargeChange;
+                int large = fullNotches * _scrollBar.LargeChange;
                 int newOffset = Math.Max(0, initialOffset - large);
                 newOffset = Math.Min(newOffset, TotalProperties - _visibleRows + 1);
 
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyGridView.cs
```diff
@@ -3256,7 +3256,7 @@ protected override void OnMouseWheel(MouseEventArgs e)
             if (fullNotches != 0)
             {
                 int originalOffset = initialOffset;
-                int large = fullNotches * _scrollBar.LargeChange;
+                int large = fullNotches * _scrollBar!.LargeChange;
```
dreddy-work, line 3259: can you add debug assert  on  `_scrollBar` and see if we can avoid `!`?

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyGridView.cs
```diff
@@ -3463,9 +3463,9 @@ protected override void OnPaint(PaintEventArgs e)
         }
     }
 
-    private void OnGridEntryLabelDoubleClick(object s, EventArgs e)
+    private void OnGridEntryLabelDoubleClick(object? s, EventArgs e)
```
dreddy-work, line 3466: why define nullable? and mark non null immediately.  This is private method.

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyGridView.cs
```diff
@@ -3477,9 +3477,9 @@ private void OnGridEntryLabelDoubleClick(object s, EventArgs e)
         DoubleClickRow(row, gridEntry.Expandable, RowLabel);
     }
 
-    private void OnGridEntryValueDoubleClick(object s, EventArgs e)
+    private void OnGridEntryValueDoubleClick(object? s, EventArgs e)
```
dreddy-work, line 3480: same here.

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyGridView.cs
gpetrou: This is used in the event handler.

dreddy-work (approved):

PR #10057, opened 2023-10-06
Enable nullability in some DataGridViewColumnCollection members

gpetrou:
## Proposed changes

- Enable nullability in some DataGridViewColumnCollection members.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10057)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.Methods.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.Methods.cs
@@ -1621,7 +1621,7 @@ public partial class DataGridView
         }
 
         // Sort the array.
-        Array.Sort(finalClonedColumns, System.Windows.Forms.DataGridViewColumnCollection.ColumnCollectionOrderComparer);
+        Array.Sort(finalClonedColumns, DataGridViewColumnCollection.ColumnCollectionOrderComparer);
 
         // 4. Add new columns for the Fields which were not data bound previously ( ie, for fields which do not have a clone ).
         if (boundColumns is not null)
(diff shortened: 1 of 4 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewColumnCollection.ColumnOrderComparer.cs
```diff
@@ -5,7 +5,7 @@ namespace System.Windows.Forms;
 
 public partial class DataGridViewColumnCollection
 {
-    private class ColumnOrderComparer : IComparer<DataGridViewColumn>
+    private class ColumnOrderComparer : IComparer<DataGridViewColumn?>
```
lonitra, line 8: Why was this changed to be nullable?

gpetrou: There is a warning in https://github.com/dotnet/winforms/pull/10057/files#diff-9c795a0e7795add33c2eefe872513e1adb161e4a5f9a557f57bc5041439a7c14R1624 about it.

lonitra: Ah I see what is going on now, thank you!

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.Methods.cs
```diff
@@ -1621,7 +1621,7 @@ private void AutoGenerateDataBoundColumns(List<DataGridViewColumn> boundColumns)
         }
 
         // Sort the array.
-        Array.Sort(finalClonedColumns, System.Windows.Forms.DataGridViewColumnCollection.ColumnCollectionOrderComparer);
+        Array.Sort(finalClonedColumns, DataGridViewColumnCollection.ColumnCollectionOrderComparer);
```
lonitra, line 1624: I'm not sure `finalClonedColumns` should be marked as nullable here. It is copying from `clonedColumns` which is ultimately cloning from `Columns`. I believe `DataGridViewColumnCollection` ensures that we cannot add a null to the collection, so `finalClonedColumns` shouldn't have any nulls. We could might be able to now also remove the nullable added to `ColumnOrderComparer` once that is changed

gpetrou: We have https://github.com/dotnet/winforms/blob/5013e65b12f7a276f43e44376853c6e5e8cfb0db/src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.Methods.cs#L1641 and https://github.com/dotnet/winforms/blob/5013e65b12f7a276f43e44376853c6e5e8cfb0db/src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.Methods.cs#L1651 though.
What should we do there?

lonitra (approved):

PR #10079, opened 2023-10-09
Enable nullability in CommandSet

halgab:
Just the one internal class, but it's pretty huge...

## Proposed changes

- Enable nullability in `CommandSet`

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10079)


sharwell:
Thank you for separating the code style changes from the nullability changes. Please make sure to _not_ squash merge this pull request after the review is complete.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/CommandSet.cs
```diff
@@ -62,12 +60,12 @@ public CommandSet(ISite site)
     {
         this.site = site;
 
-        _eventService = (IEventHandlerService)site.GetService(typeof(IEventHandlerService));
+        _eventService = site.GetService<IEventHandlerService>()!;
         Debug.Assert(_eventService is not null, "Command set must have the event service.  Is command set being initialized too early?");
```
sharwell, line 64: 💡 Consider creating a `GetRequiredService` extension method for this pattern.

https://github.com/dotnet/roslyn/blob/0caa880a161057559d1dcaad890e4053f520136e/src/Workspaces/Core/Portable/Workspace/Host/HostWorkspaceServices.cs#L43-L57

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/CommandSet.cs
```diff
@@ -439,7 +414,7 @@ protected void EndDragManager()
     }
 
     // Returns true if the action is successful, false otherwise
-    internal static bool ExecuteSafely(Action action, bool throwOnException)
+    internal static bool ExecuteSafely(Action? action, bool throwOnException)
```
sharwell, line 417: 😕 It's strange for `action` to be null here, _especially_ considering the false return value when it's null.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/CommandSet.cs
```diff
@@ -464,7 +434,7 @@ internal static bool ExecuteSafely(Action action, bool throwOnException)
 
     // This function will return true if call to func is successful, false otherwise
     // Output of call to func is available in result out parameter
-    private static bool ExecuteSafely<T>(Func<T> func, bool throwOnException, out T result)
+    private static bool ExecuteSafely<T>(Func<T> func, bool throwOnException, out T? result)
```
sharwell, line 437: Final parameter should be:

```
[MaybeNullWhen(false)] out T result
```

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/CommandSet.cs
```diff
@@ -529,12 +494,12 @@ private IComponent[] FilterSelection(IComponent[] components, SelectionRules sel
     /// </summary>
     protected virtual ICollection GetCopySelection()
     {
-        ICollection selectedComponents = SelectionService.GetSelectedComponents();
+        ICollection selectedComponents = SelectionService!.GetSelectedComponents();
```
sharwell, line 497: Can we not throw a more specific exception (e.g. `InvalidOperationException`) when `SelectionService` is unavailable?

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/CommandSet.cs
```diff
@@ -750,264 +726,253 @@ protected virtual bool OnKeyCancel(object sender)
...
-    protected virtual void OnKeyMove(object sender, EventArgs e)
+    protected virtual void OnKeyMove(object? sender, EventArgs e)
     {
         // Arrow keys.  Begin a drag if the selection isn't locked.
         //
-        ISelectionService selSvc = SelectionService;
+        ISelectionService? selSvc = SelectionService;
 
-        if (selSvc is not null)
+        if (selSvc is not null && selSvc.PrimarySelection is IComponent comp)
```
sharwell, line 754: ❔ Can this not also use the following?

```
if (SelectionService?.PrimarySelection is IComponent comp)
```

I was under the impression that this syntax would end up treating _both_ `SelectionService` and `comp` as non-null for the following block.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/CommandSet.cs
```diff
@@ -750,264 +726,253 @@ protected virtual bool OnKeyCancel(object sender)
...
+                    {
+                        Debug.Fail($"Unknown command mapped to OnKeyMove: {cmd}");
+                    }
 
-                        DesignerTransaction trans = selSvc.SelectionCount > 1
-                            ? host.CreateTransaction(string.Format(SR.DragDropMoveComponents, selSvc.SelectionCount))
-                            : host.CreateTransaction(string.Format(SR.DragDropMoveComponent, comp.Site.Name));
+                    DesignerTransaction trans = selSvc.SelectionCount > 1
+                        ? host.CreateTransaction(string.Format(SR.DragDropMoveComponents, selSvc.SelectionCount))
+                        : host.CreateTransaction(string.Format(SR.DragDropMoveComponent, comp.Site!.Name));
```
sharwell, line 808: Any reason not to use `comp.Site?.Name` here?

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/CommandSet.cs
```diff
@@ -750,264 +726,253 @@ protected virtual bool OnKeyCancel(object sender)
...
 
-                                    if (snapOn && !snapSize.IsEmpty)
+                                    if (primaryControl is not null)
                                     {
-                                        moveOffsetX *= snapSize.Width;
-                                        moveOffsetY *= snapSize.Height;
-
-                                        if (primaryControl is not null)
+                                        //ask the parent to adjust our wanna-be snapped position
+                                        if (host.GetDesigner(primaryControl.Parent!) is ParentControlDesigner parentDesigner)
```
sharwell, line 884: 💡 Consider adding a `GetRequiredParent()` extension method, similar to what we did in Roslyn

https://github.com/dotnet/roslyn/blob/0caa880a161057559d1dcaad890e4053f520136e/src/Workspaces/SharedUtilitiesAndExtensions/Compiler/Core/Extensions/SyntaxNodeExtensions.cs#L23-L24

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/CommandSet.cs
halgab: I'll wait for the team to state their intent for this one

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/CommandSet.cs
```diff
@@ -204,16 +202,16 @@ public CommandSet(ISite site)
...
+        SelectionService = site.GetService<ISelectionService>();
         Debug.Assert(SelectionService is not null, "CommandSet relies on the selection service, which is unavailable.");
         if (SelectionService is not null)
         {
             SelectionService.SelectionChanged += new EventHandler(OnSelectionChanged);
         }
 
-        _menuService = (IMenuCommandService)site.GetService(typeof(IMenuCommandService));
+        _menuService = site.GetService<IMenuCommandService>();
         if (_menuService is not null)
```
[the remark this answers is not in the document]
halgab: I cached `MenuService` to a local variable to avoid checking for null inside the loop or using the null forgiving operator

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/CommandSet.cs
halgab: It raises the question of throwing new exceptions, so I'll let the team decide what they want to do with this one

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/CommandSet.cs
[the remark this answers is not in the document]
sharwell: Note that the compiler flow analysis will remember that `MenuService` is not null and allow additional references to the property without null checks or suppressions. I was commenting because the assignment to `_menuService` here is duplicating the exact code of `MenuService` property.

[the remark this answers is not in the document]
halgab: Understood. It's just that the property does not work under the hood (it's not an auto-property). My change avoids the null check inside `MenuService`. I could also have used the underlying field inside the loop, but it's harder to read

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/CommandSet.cs
```diff
@@ -4092,13 +3561,13 @@ public int Compare(object p, object q)
     /// </summary>
     private class ComponentTopCompare : IComparer
     {
-        public int Compare(object p, object q)
+        public int Compare(object? p, object? q)
```
lonitra, line 3564: Is this method ever called with null parameters?

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/CommandSet.cs
lonitra: imo I would much rather have it throw an exception like this rather than a null reference, but I will check with team on this among some of the other questions about raising a more specific exception

lonitra:
> Thank you for separating the code style changes from the nullability changes. Please make sure to _not_ squash merge this pull request after the review is complete.

We typically squash merge commits if the commit messages do not add significant value in understanding what is going on in the branch. In this case, I'm not sure the messages add value here and think the PR commit from squash merge is sufficient, but am open to other thoughts.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/CommandSet.cs
halgab: I guess not, but our hands are tied bu the annotations of the interface. I could change it to IComparer<IComponent> though

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/CommandSet.cs
```diff
@@ -349,9 +311,11 @@ private bool CheckComponentEditor(object obj, bool launchEditor)
                     }
                 }
 
+                bool success;
```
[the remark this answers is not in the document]
halgab: It is set on lines 329 and 333, and used on line 336

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/CommandSet.cs
Tanya-Solyanik: +1 for a specific exception

halgab: Done. I used the already existing `GetRequiredService` and deleted the assert

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/CommandSet.cs
halgab: `SelectionService` null check looks to be a mess. There is an assert checking for its presence. Sometimes we check it for null, which I guess is needed in release, and sometimes we don't, like here, and also on line 486. This would need to be cleaned up, either here or in a follow-up PR. What do you think @lonitra?

lonitra: Let's do that clean up as a separate PR

sharwell:
> > Thank you for separating the code style changes from the nullability changes. Please make sure to _not_ squash merge this pull request after the review is complete.
> 
> We typically squash merge commits if the commit messages do not add significant value in understanding what is going on in the branch. In this case, I'm not sure the messages add value here and think the PR commit from squash merge is sufficient, but am open to other thoughts.

For this pull request, a squash merge will invalidate git blame by changing many lines from 'CommandSet: fix name violations and use auto-props' (correct) to 'Enable nullability in CommandSet' (incorrect).

lonitra (approved):
RLGTM

halgab:
> > > Thank you for separating the code style changes from the nullability changes. Please make sure to _not_ squash merge this pull request after the review is complete.
> > 
> > We typically squash merge commits if the commit messages do not add significant value in understanding what is going on in the branch. In this case, I'm not sure the messages add value here and think the PR commit from squash merge is sufficient, but am open to other thoughts.
> 
> For this pull request, a squash merge will invalidate git blame by changing many lines from 'CommandSet: fix name violations and use auto-props' (correct) to 'Enable nullability in CommandSet' (incorrect). If squash merge is desired, it would be best to extract the first commit to its own PR, wait for it to merge, and then rebase this change onto the new branch.

I could also try to group some commits and rename them

lonitra:
> I could also try to group some commits and rename them

If you could regroup commits to separate nullability change and more stylistic changes then we can go ahead and do a merge commit here.

halgab:
> > I could also try to group some commits and rename them
> 
> If you could regroup commits to separate nullability change and more stylistic changes that would be great. Then we can go ahead and do a merge commit here.

I'll have a look, but it shouldn't be a problem.

halgab:
That's done. I'm ready to do always keep my PRs with clean commit history from now on, but I think that merging PR review feedback inside existing commits make it harder for reviewers. Perhaps maintainers could say when commits are ready to be squashed?

lonitra:
> That's done. I'm ready to do always keep my PRs with clean commit history from now on, but I think that merging PR review feedback inside existing commits make it harder for reviewers. Perhaps maintainers could say when commits are ready to be squashed?

Yes please keep review feedback in a separate commit so that it is easier to review. Once review is complete and there is value to keep the history, we'll let you know to clean commit history so that a merge-commit can be done

lonitra (approved):
🥳 Thank you for cleaning the commit history!

PR #10081, opened 2023-10-09
Convert UIA Enum to Cswin32

lonitra:
Removes `UIA` enum in preparation for more changes relating to converting COM interfaces on `AccessibleObject`.


src/System.Windows.Forms.Primitives/src/Windows/Win32/UI/Accessibility/UIA_CONTROLTYPE_ID.cs
```diff
@@ -0,0 +1,49 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Windows.Win32.UI.Accessibility;
+
+internal enum UIA_CONTROLTYPE_ID : int
```
lonitra, line 6: Although Cswin32 has a defined `UIA_CONTROLTYPE_ID`, it is defined as `uint`, which causes Narrator to give incorrect announcements. This may be a consequence of having partially converted Cswin32 definitions. `IRawElementProviderSimple` is the primary interface that relies on the UIA enum and this hasn't been converted. For now, we will add this version of `UIA_CONTROLTYPE_ID` definition and revisit whether or not `UIA_CONTROLTYPE_ID` being defined as a `uint` enum is a bug when we convert `IRawElementProviderSimple` to Cswin32, which is the next step.

KlausLoeffelmann:
Looks like a pretty mechanical change. Did you use find/replace or regex to swap the consts out?
I did spot checks, and everything looks plausible.

lonitra:
> Looks like a pretty mechanical change. Did you use find/replace or regex to swap the consts out? I did spot checks, and everything looks plausible.

I used find/replace whenever there were multiple of the same UIA type (`UIA_PROPERTY_TYPE`, `UIA_PATTERN_TYPE`, etc.) that needed to be changed in a file, otherwise I manually converted it if there was only one UIA type to convert.

src/System.Windows.Forms.Primitives/src/Windows/Win32/UI/Accessibility/UIA_CONTROLTYPE_ID.cs
Tanya-Solyanik: All of these enums should be signed ints per docs and the idl file. This one is the easiest to test. We have to return tha documented type from GetPropertyValue methods becasue this is what the accessibility tools are expecting. We can't expect accessibility tools coded to be resilient and try to cast uint to int

lonitra: Added `UIA_PATTERN_ID` and `UIA_PROPERTY_ID` and made them int enums. I will follow up filing an issue with metadata to make all the enums signed ints.

lonitra: https://github.com/microsoft/win32metadata/issues/1723

Tanya-Solyanik (commented):
Looks good, but some test data can be simplified a little by removing casts to an int. Look for (int)UIA_ pattern

Tanya-Solyanik (approved):
Feel free to merge as is and follow up on the comments in another PR

lonitra:
Will follow up on these comments in the next PR

PR #10084, opened 2023-10-10
Add nullability to BaseContextmenuStrip

ricardobossan:
Related #1107 

## Proposed changes

- Enabled nullability for BaseContextmenuStrip
 

## Regression? 

- No

## Risk

- Minimal

## Test methodology 

- 

 

## Test environment(s) 
- .net 9.0.100-alpha.1.23464.17

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10084)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BaseContextMenuStrip.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BaseContextMenuStrip.cs
@@ -1,8 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-#nullable disable
-
 using System.ComponentModel;
 using System.ComponentModel.Design;
 using System.Drawing;
@@ -16,7 +14,7 @@ internal class BaseContextMenuStrip : GroupedContextMenuStrip
 {
     private readonly IServiceProvider serviceProvider;
     private readonly Component component;
-    private ToolStripMenuItem selectionMenuItem;
+    private ToolStripMenuItem? selectionMenuItem;
 
     public BaseContextMenuStrip(IServiceProvider provider, Component component) : base()
     {
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BaseContextMenuStrip.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BaseContextMenuStrip.cs
@@ -150,9 +150,7 @@ internal class BaseContextMenuStrip : GroupedContextMenuStrip
     private void AddVerbMenuItem()
     {
         //Add Designer Verbs..
-        IMenuCommandService? menuCommandService = serviceProvider.GetService<IMenuCommandService>();
-
-        if (menuCommandService is not null)
+        if (serviceProvider.TryGetService(out IMenuCommandService? menuCommandService))
         {
             DesignerVerbCollection verbCollection = menuCommandService.Verbs;
             foreach (DesignerVerb verb in verbCollection)
@@ -197,7 +195,7 @@ internal class BaseContextMenuStrip : GroupedContextMenuStrip
         //this.Opening += new CancelEventHandler(OnContextMenuOpening);
         Name = "designerContextMenuStrip";
 
-        if (serviceProvider.GetService(typeof(IUIService)) is IUIService uis)
+        if (serviceProvider.TryGetService(out IUIService? uis))
         {
             Renderer = (ToolStripProfessionalRenderer)uis.Styles["VsRenderer"]!;
 
@@ -224,7 +222,7 @@ internal class BaseContextMenuStrip : GroupedContextMenuStrip
     /// </summary>
     public override void RefreshItems()
     {
-        if (serviceProvider.GetService(typeof(IUIService)) is IUIService uis)
+        if (serviceProvider.TryGetService(out IUIService? uis))
         {
             Font = (Font)uis.Styles["DialogFont"]!;
         }
@@ -258,19 +256,15 @@ internal class BaseContextMenuStrip : GroupedContextMenuStrip
             _serviceProvider = provider;
             // Get NestedSiteName...
             string? compName = null;
-            if (_comp is not null)
+            if (_comp is not null && _comp.Site is { } site)
             {
-                ISite? site = _comp.Site;
-                if (site is not null)
+                if (site is INestedSite nestedSite && !string.IsNullOrEmpty(nestedSite.FullName))
                 {
-                    if (site is INestedSite nestedSite && !string.IsNullOrEmpty(nestedSite.FullName))
-                    {
-                        compName = nestedSite.FullName;
-                    }
-                    else if (!string.IsNullOrEmpty(site.Name))
-                    {
-                        compName = site.Name;
-                    }
+                    compName = nestedSite.FullName;
+                }
+                else if (!string.IsNullOrEmpty(site.Name))
+                {
+                    compName = site.Name;
                 }
             }
 
@@ -289,11 +283,11 @@ internal class BaseContextMenuStrip : GroupedContextMenuStrip
                     // else attempt to get the resource from a known place in the manifest. if and only if the namespace of the type is System.Windows.Forms. else attempt to get the resource from a known place in the manifest
                     if (_itemType.Namespace == s_systemWindowsFormsNamespace)
                     {
-                        _image = ToolboxBitmapAttribute.GetImageFromResource(_itemType, null, false);
+                        _image = ToolboxBitmapAttribute.GetImageFromResource(_itemType, imageName: null, large: false);
                     }
 
                     // if all else fails, throw up a default image.
-                    _image ??= ToolboxBitmapAttribute.GetImageFromResource(_comp!.GetType(), null, false);
+                    _image ??= ToolboxBitmapAttribute.GetImageFromResource(_comp!.GetType(), imageName: null, large: false);
                 }
 
                 return _image;
@@ -310,7 +304,7 @@ internal class BaseContextMenuStrip : GroupedContextMenuStrip
         /// </summary>
         protected override void OnClick(EventArgs e)
         {
-            if (_comp is not null && _serviceProvider.GetService(typeof(ISelectionService)) is ISelectionService selectionService)
+            if (_comp is not null && _serviceProvider.TryGetService(out ISelectionService? selectionService))
             {
                 selectionService.SetSelectedComponents(new object[] { _comp }, SelectionTypes.Replace);
             }
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BaseContextMenuStrip.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BaseContextMenuStrip.cs
@@ -118,7 +118,7 @@ internal class BaseContextMenuStrip : GroupedContextMenuStrip
 
             if (serviceProvider.GetService(typeof(IUIService)) is IUIService uis)
             {
-                var rendererStyle = uis.Styles["VsRenderer"];
+                object? rendererStyle = uis.Styles["VsRenderer"];
                 if (rendererStyle is ToolStripProfessionalRenderer renderer)
                 {
                     selectionMenuItem.DropDown.Renderer = renderer;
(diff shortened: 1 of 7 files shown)


ricardobossan:
Drafted the pull request because there's still warnings to correct regarding the changes I'm making.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BaseContextMenuStrip.cs
```diff
@@ -121,9 +118,15 @@ private void RefreshSelectionMenuItem()
 
             if (serviceProvider.GetService(typeof(IUIService)) is IUIService uis)
             {
-                selectionMenuItem.DropDown.Renderer = (ToolStripProfessionalRenderer)uis.Styles["VsRenderer"];
-                //Set the right Font
-                selectionMenuItem.DropDown.Font = (Font)uis.Styles["DialogFont"];
+                var rendererStyle = uis.Styles["VsRenderer"];
```
elachlan, line 121: try to use the type directly instead of var.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BaseContextMenuStrip.cs
```diff
@@ -152,7 +156,7 @@ private void RefreshSelectionMenuItem()
     private void AddVerbMenuItem()
     {
         //Add Designer Verbs..
-        IMenuCommandService menuCommandService = (IMenuCommandService)serviceProvider.GetService(typeof(IMenuCommandService));
+        IMenuCommandService? menuCommandService = serviceProvider.GetService(typeof(IMenuCommandService)) as IMenuCommandService;
```
elachlan, line 159: Is there a TryGetService?

LeafShi1: Do you mean `serviceProvider.TryGetService(out IMenuCommandService? menuCommandService);`，I think there is.

@dreddy-work Which method is better to call here? It seems that using TryGetService is simpler, but we still use GetService in many places in current project.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BaseContextMenuStrip.cs
```diff
@@ -104,9 +102,8 @@ private void RefreshSelectionMenuItem()
                     parentControl = parentControl.Parent;
                 }
             }
-            else if (selectionService.PrimarySelection is ToolStripItem)
+            else if (selectionService.PrimarySelection is ToolStripItem selectedItem)
             {
-                ToolStripItem selectedItem = selectionService.PrimarySelection as ToolStripItem;
                 if (host.GetDesigner(selectedItem) is ToolStripItemDesigner itemDesigner)
                 {
                     parentControls = itemDesigner.AddParentTree();
```
halgab, line 109: Looking at `AddParentTree` which is not annotated yet, I'm not 100% sure it returns only non-null items

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BaseContextMenuStrip.cs
```diff
@@ -244,22 +245,22 @@ public override void RefreshItems()
     /// </summary>
     private class SelectToolStripMenuItem : ToolStripMenuItem
     {
-        private readonly Component _comp;
+        private readonly Component? _comp;
```
halgab, line 248: if itemDesigner.AddParentTree only return non-null elements (see [here](url)), then I don't get why this needs to be null. But if it can be null, then this is not consistent with this class constructor signature

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BaseContextMenuStrip.cs
```diff
@@ -292,7 +293,10 @@ public override Image Image
                     }
 
                     // if all else fails, throw up a default image.
-                    _image ??= ToolboxBitmapAttribute.GetImageFromResource(_comp.GetType(), null, false);
+                    if (_comp is not null)
```
halgab, line 296: To me, the null checks you needed to add to `_comp` are a hint it might not be possible for it to be null after all

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ToolStripContentPanelDesigner.cs
```diff
@@ -11,26 +11,32 @@ internal class ToolStripContentPanelDesigner : PanelDesigner
...
     {
         get
         {
-            if (_contextMenu is null)
+            ISite? site = Component.Site;
+            var component = Component as Component;
+
+            if (site is not null && component is not null)
             {
-                _contextMenu = new BaseContextMenuStrip(Component.Site, Component as Component);
```
halgab, line 20: Actually, the second argument in `BaseContextMenuStrip`'s constructor is only used to initialise a field that isn't used anywhere. Perhaps we could remove it entirely, and stop caring here about `Component` being a `Component`?

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ToolStripPanelDesigner.cs
```diff
@@ -58,23 +58,29 @@ private Pen BorderPen
...
-                    StandardGroups.Custom,
-                    StandardGroups.Selection,
-                    StandardGroups.Edit,
-                    StandardGroups.Properties
-                ]);
-                _contextMenu.Text = "CustomContextMenu";
+                ISite? site = Component.Site;
+                var component = Component as Component;
+
+                if (site is not null && component is not null)
```
halgab, line 70: Same as above

halgab (commented):
Note that base class `GroupedContextMenuStrip` isn't annotated yet

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BaseContextMenuStrip.cs
ricardobossan: If I did not make _comp nullable, the constructor would complain:

image

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BaseContextMenuStrip.cs
ricardobossan: As I [had to make `_comp` nullable to remove warnings from the constructor](https://github.com/dotnet/winforms/pull/10084#discussion_r1354761914), it triggered a warning in this section of code:
image

To address that, I added the null check in line 296.

image

However, maybe a better approach would be this:

image

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BaseContextMenuStrip.cs
halgab: I think it's because of the null check at the bottom of your screenshot. I'm always uncomfortable in those situations were we try to check for something that should not happen... What if you used `c` instead of `_comp` after setting `_comp`?

ricardobossan: As the code was originally like that, generated no warnings, and _comp is a property of the class, I would rather not change it to avoid unforeseen effects.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BaseContextMenuStrip.cs
```diff
@@ -152,7 +150,8 @@ private void RefreshSelectionMenuItem()
     private void AddVerbMenuItem()
     {
         //Add Designer Verbs..
-        IMenuCommandService menuCommandService = (IMenuCommandService)serviceProvider.GetService(typeof(IMenuCommandService));
+        IMenuCommandService? menuCommandService = serviceProvider.GetService<IMenuCommandService>();
+
         if (menuCommandService is not null)
```
lonitra, line 155: I believe we have a TryGetService method that will check for null so this can be simplified to
```suggestion
        if (serviceProvider.TryGetService(out IMenuCommandService? menuCommandService))
```

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BaseContextMenuStrip.cs
```diff
@@ -197,12 +196,14 @@ private void InitializeContextMenu()
     {
         //this.Opening += new CancelEventHandler(OnContextMenuOpening);
         Name = "designerContextMenuStrip";
+
         if (serviceProvider.GetService(typeof(IUIService)) is IUIService uis)
```
lonitra, line 200:
```suggestion
        if (serviceProvider.TryGetService(out IUIService? uis)
```
Ditto is a couple places below as well

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BaseContextMenuStrip.cs
```diff
@@ -244,22 +245,22 @@ public override void RefreshItems()
...
             _comp = c;
             _serviceProvider = provider;
             // Get NestedSiteName...
-            string compName = null;
+            string? compName = null;
             if (_comp is not null)
             {
-                ISite site = _comp.Site;
+                ISite? site = _comp.Site;
                 if (site is not null)
```
lonitra, line 264: Let's combine these if statements to reduce nesting:
```c#
if (_comp is not null && _comp.Site is { } site)
```

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BaseContextMenuStrip.cs
```diff
@@ -292,7 +293,7 @@ public override Image Image
                     }
 
                     // if all else fails, throw up a default image.
-                    _image ??= ToolboxBitmapAttribute.GetImageFromResource(_comp.GetType(), null, false);
+                    _image ??= ToolboxBitmapAttribute.GetImageFromResource(_comp!.GetType(), null, false);
```
lonitra, line 296: nit: could you add the parameter names for `null` and `false`

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ToolStripContentPanelDesigner.cs
```diff
@@ -11,26 +11,31 @@ internal class ToolStripContentPanelDesigner : PanelDesigner
...
     {
         get
         {
-            if (_contextMenu is null)
+            ISite? site = Component.Site;
+
+            if (site is not null)
             {
-                _contextMenu = new BaseContextMenuStrip(Component.Site, Component as Component);
+                if (_contextMenu is null)
```
lonitra, line 22: Please combine these if statements

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ToolStripPanelDesigner.cs
```diff
@@ -58,23 +58,28 @@ private Pen BorderPen
...
-                    StandardGroups.Verbs,
-                    StandardGroups.Custom,
-                    StandardGroups.Selection,
-                    StandardGroups.Edit,
-                    StandardGroups.Properties
-                ]);
-                _contextMenu.Text = "CustomContextMenu";
+                ISite? site = Component.Site;
+
+                if (site is not null )
```
lonitra, line 69: Ditto here with combining if statement

lonitra (commented):
Nice work! A few comments but otherwise LGTM!

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/BaseContextMenuStrip.cs
dreddy-work: For consistency sake, If we had to check if returned service is not null in the source, `TryGetService` is right choice. 

If explicitly throwing an exception when `ServiceProvider` is null, then we should be using `serviceProvider.GetService<IMenuCommandService>()`

lonitra (approved):

elachlan:
@ricardobossan it hasn't been merged yet. Are you sure you want to close it?

PR #9367, opened 2023-06-25
Enable nullability in DesignerActionService and DesignerActionUIService

halgab:
## Proposed changes

- Enable nullability in `DesignerActionService`, `DesignerActionUIService`, and all the internal code these classes use: `DesignerActionUI`, `DesignerActionPanel` and `SelectionManager`

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9367)

PR #10088, opened 2023-10-10
Enable nullability in DesignerActionPanel

halgab:
Now that #10035 is merged, my plan is to annotate just `DesignerActionPanel`, and then rebase #9367 to remove the annotation changes I made on it there, but keep all the other changes with their review.

My changes here include the modification of the annotation of a public interface. The two internal implementers of method `IWindowsFormsEditorService.DropDownControl` cannot handle a null `Control` as argument, and nothing in the docs said the implementers should prepare for that scenario

## Proposed changes

- Enable nullability in `DesignerActionPanel`
- Update nullability annotations of interface `IWindowsFormsEditorService`

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10088)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionPanel.EditorPropertyLine.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionPanel.EditorPropertyLine.cs
@@ -15,7 +15,8 @@ internal sealed partial class DesignerActionPanel
     {
         private readonly EditorButton _button;
         private UITypeEditor? _editor;
-        private bool _hasSwatch;
+        [MemberNotNullWhen(true, nameof(_editor))]
+        private bool HasSwatch { get; set; }
         private Image? _swatch;
         private FlyoutDialog? _dropDownHolder;
         private bool _ignoreNextSelectChange;
@@ -59,7 +60,7 @@ internal sealed partial class DesignerActionPanel
                 TypeConverter.StandardValuesCollection? standardValues = GetStandardValues();
                 if (standardValues is not null)
                 {
-                    foreach (object standardValue in standardValues)
+                    foreach (object? standardValue in standardValues)
                     {
                         string newItem = PropertyDescriptor.Converter.ConvertToString(TypeDescriptorContext, CultureInfo.CurrentCulture, standardValue)!;
                         listBox.Items.Add(newItem);
@@ -135,7 +136,7 @@ internal sealed partial class DesignerActionPanel
 
         protected override int GetTextBoxLeftPadding(int textBoxHeight)
         {
-            if (_hasSwatch)
+            if (HasSwatch)
             {
                 return base.GetTextBoxLeftPadding(textBoxHeight) + textBoxHeight + 2 * EditorLineSwatchPadding;
             }
@@ -230,14 +231,14 @@ internal sealed partial class DesignerActionPanel
             if (_editor is not null)
             {
                 _button.Ellipsis = (_editor.GetEditStyle(TypeDescriptorContext) == UITypeEditorEditStyle.Modal);
-                _hasSwatch = _editor.GetPaintValueSupported(TypeDescriptorContext);
+                HasSwatch = _editor.GetPaintValueSupported(TypeDescriptorContext);
             }
             else
             {
                 _button.Ellipsis = false;
             }
 
-            EditControl!.AccessibleRole = _button.Ellipsis
+            EditControl.AccessibleRole = _button.Ellipsis
                 ? IsReadOnly() ? AccessibleRole.StaticText : AccessibleRole.Text
                 : IsReadOnly() ? AccessibleRole.DropList : AccessibleRole.ComboBox;
 
@@ -272,7 +273,7 @@ internal sealed partial class DesignerActionPanel
             base.OnValueChanged();
 
             _swatch = null;
-            if (_hasSwatch)
+            if (HasSwatch)
             {
                 ActionPanel.Invalidate(new Rectangle(EditRegionLocation, EditRegionSize), false);
             }
@@ -282,7 +283,7 @@ internal sealed partial class DesignerActionPanel
         {
             base.PaintLine(g, lineWidth, lineHeight);
 
-            if (_hasSwatch)
+            if (HasSwatch)
             {
                 if (_swatch is null)
                 {
@@ -292,7 +293,7 @@ internal sealed partial class DesignerActionPanel
                     Rectangle rect = new Rectangle(1, 1, width - 2, height - 2);
                     using (Graphics swatchGraphics = Graphics.FromImage(_swatch))
                     {
-                        _editor!.PaintValue(Value, swatchGraphics, rect);
+                        _editor.PaintValue(Value, swatchGraphics, rect);
                         swatchGraphics.DrawRectangle(SystemPens.ControlDark, new Rectangle(0, 0, width - 1, height - 1));
                     }
                 }
@@ -408,7 +409,10 @@ internal sealed partial class DesignerActionPanel
 
         void IWindowsFormsEditorService.DropDownControl(Control control)
         {
-            ShowDropDown(control, ActionPanel.BorderColor);
+            if (control is not null)
+            {
+                ShowDropDown(control, ActionPanel.BorderColor);
+            }
         }
 
         DialogResult IWindowsFormsEditorService.ShowDialog(Form dialog)
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionPanel.EditorPropertyLine.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionPanel.EditorPropertyLine.cs
@@ -1,8 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-#nullable disable
-
 using System.Drawing;
 using System.Drawing.Design;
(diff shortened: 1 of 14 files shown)


src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionPanel.MethodLine.cs
```diff
@@ -52,27 +50,27 @@ public override Size LayoutControls(int top, int width, bool measureOnly)
...
 
-        private void OnLinkLabelLinkClicked(object sender, LinkLabelLinkClickedEventArgs e)
+        private void OnLinkLabelLinkClicked(object? sender, LinkLabelLinkClickedEventArgs e)
         {
             Debug.Assert(!ActionPanel.InMethodInvoke, "Nested method invocation");
             ActionPanel.InMethodInvoke = true;
             try
             {
-                _methodItem.Invoke();
+                _methodItem!.Invoke();
```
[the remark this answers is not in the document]
halgab: Because `_methodItem` cannot be null here. I attempted to refactor the code even further to make the field entirely non-nullable, but it made the initialisation code of `Line`s pretty hard to read

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionPanel.EditorPropertyLine.cs
```diff
@@ -299,7 +297,7 @@ public override void PaintLine(Graphics g, int lineWidth, int lineHeight)
                     Rectangle rect = new Rectangle(1, 1, width - 2, height - 2);
                     using (Graphics swatchGraphics = Graphics.FromImage(_swatch))
                     {
-                        _editor.PaintValue(Value, swatchGraphics, rect);
+                        _editor!.PaintValue(Value, swatchGraphics, rect);
```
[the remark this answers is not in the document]
halgab: It adds a null check that is unnecessary. It's pretty much the same situation as [here](https://github.com/dotnet/winforms/pull/10088#discussion_r1354393166)

Tanya-Solyanik:
>My changes here include the modification of the annotation of a public interface. The two internal implementers of method IWindowsFormsEditorService.DropDownControl cannot handle a null Control as argument, and nothing in the docs said the implementers should prepare for that scenario

If a public API does not handle null input gracefully, I'd rather fix the implementation to do nothing in this case. We don't want to crash if users ignored a warning.

Implementation in the property grid handles null gracefully, except got the Tracing , which can be fixed. - https://github.com/dotnet/winforms/blob/ac594bf200c2e34e7943c3abed1bac3b69114359/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyGridView.cs#L1461

Implementation in the designer action panel - https://github.com/dotnet/winforms/blob/ac594bf200c2e34e7943c3abed1bac3b69114359/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionPanel.EditorPropertyLine.cs#L416 can exit and do nothing is control is null

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionPanel.EditorPropertyLine.cs
```diff
@@ -58,15 +56,15 @@ private unsafe void ActivateDropDown()
                 listBox.SelectedIndexChanged += new EventHandler(OnListBoxSelectedIndexChanged);
                 listBox.KeyDown += new KeyEventHandler(OnListBoxKeyDown);
 
-                TypeConverter.StandardValuesCollection standardValues = GetStandardValues();
+                TypeConverter.StandardValuesCollection? standardValues = GetStandardValues();
                 if (standardValues is not null)
                 {
-                    foreach (object o in standardValues)
+                    foreach (object standardValue in standardValues)
```
Tanya-Solyanik, line 62: `standardValue` can be null, this collection does not guarantee that it's not a null. And you test for null on line 67. Why did you declare it as not-nullable?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionPanel.MethodLine.cs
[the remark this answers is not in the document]
Tanya-Solyanik: Please add a comment explaining why it can't be null.  The rule of thumb for comments is that if code raised questions in the PR once, it would raise question again later.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionPanel.EditorPropertyLine.cs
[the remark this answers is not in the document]
Tanya-Solyanik: Please clarify why _editor is initialized here, but we are not guaranteed that PropertyDescriptor.GetEditor found anything in the "Update" method. It seems that internal EditorPropertyLine class is created only once we made sure that editor is available, and yet the "update" method is coded defensively.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionPanel.MethodLine.cs
```diff
@@ -52,27 +50,27 @@ public override Size LayoutControls(int top, int width, bool measureOnly)
...
             }
             catch (Exception ex)
             {
                 if (ex is TargetInvocationException)
                 {
-                    ex = ex.InnerException;
+                    ex = ex.InnerException!;
```
Tanya-Solyanik, line 65: InnerException might be null - https://learn.microsoft.com/dotnet/api/system.exception.innerexception?view=net-8.0#system-exception-innerexception

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionPanel.PropertyLine.cs
```diff
@@ -58,7 +56,7 @@ protected void SetValue(object newValue)
                             // If we can't convert it, show an error
                             if (!PropertyDescriptor.Converter.CanConvertFrom(_typeDescriptorContext, valueType))
                             {
-                                ActionPanel.ShowError(string.Format(SR.DesignerActionPanel_CouldNotConvertValue, newValue, _propDesc.PropertyType));
+                                ActionPanel.ShowError(string.Format(SR.DesignerActionPanel_CouldNotConvertValue, newValue, _propDesc!.PropertyType));
```
Tanya-Solyanik, line 59: Please use PropertyDescriptor instead of the backing field.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionPanel.PropertyLine.cs
```diff
@@ -81,7 +79,7 @@ protected void SetValue(object newValue)
             {
                 if (e is TargetInvocationException)
                 {
-                    e = e.InnerException;
+                    e = e.InnerException!;
```
Tanya-Solyanik, line 82: We need a null-check, we don't know who implemented this property descriptor.
```
                  if (e.InnerException is not null)
                  {
                      e = e.InnerException;
                  }
```

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionPanel.TextBoxPropertyLine.cs
```diff
@@ -355,7 +349,7 @@ public EditorLabelAccessibleObject(EditorLabel owner)
                 {
                 }
 
-                public override string Value => Owner.Text;
+                public override string Value => Owner!.Text;
```
Tanya-Solyanik, line 352: We would want to think that accessible objects go out of scope when control is destroyed, but this is not the case unfortunately. Please add a null-check.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionPanel.EditorPropertyLine.cs
halgab: Good catch. This didn't raise any warning, so I missed it. I wish Implicit casts in `foreach`s would raise more warnings...

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionPanel.MethodLine.cs
halgab: This pattern is used a lot across the codebase. If we are to check for null here, there are many other places that need updating. This perhaps could be its own PR?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionPanel.PropertyLine.cs
halgab: See [here](https://github.com/dotnet/winforms/pull/10088#discussion_r1359371804)

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerActionPanel.MethodLine.cs
Tanya-Solyanik: ok

Tanya-Solyanik (approved):
Thank you!

lonitra:
https://github.com/dotnet/docs/issues/37547

PR #9381, opened 2023-06-27
Enable nullability in DesignerUtils and replace ArrayList with List

halgab:
## Proposed changes

- Enable nullability in internal helper `DesignerUtils`.
- Replace a few `ArrayList`s with` List<T>`s (contributes to #8140). This change should make it easier to annotate all the files touched by this refactoring

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9381)

PR #10092, opened 2023-10-11
DesignerUtils refactorings

halgab:
Also extracted from #9381

## Proposed changes

- Modernisation of `DesignerUtils`' code
- Change signature of methods `GenerateSnapShotWithBitBlt` and `GenerateSnapShotWithWM_PRINT` to better reflect their action

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10092)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DesignerUtils.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DesignerUtils.cs
@@ -130,10 +130,7 @@ internal static class DesignerUtils
     /// <summary>
     ///  Used by Designer action glyphs to render a 'mouse hover' state.
     /// </summary>
-    public static Brush HoverBrush
-    {
-        get => s_hoverBrush;
-    }
+    public static Brush HoverBrush => s_hoverBrush;
 
     /// <summary>
     ///  Demand created size used to determine how far the user needs to drag the mouse before a drag operation starts.
@@ -228,24 +225,17 @@ internal static class DesignerUtils
     /// </summary>
     public static void DrawFrame(Graphics g, Region resizeBorder, FrameStyle style, Color backColor)
     {
-        Brush brush;
         Color color = SystemColors.ControlDarkDark;
         if (backColor != Color.Empty && backColor.GetBrightness() < .5)
         {
             color = SystemColors.ControlLight;
         }
 
-        switch (style)
+        Brush brush = style switch
         {
-            case FrameStyle.Dashed:
-                brush = new HatchBrush(HatchStyle.Percent50, color, Color.Transparent);
-                break;
-            case FrameStyle.Thick:
-            default:
-                brush = new SolidBrush(color);
-                break;
-        }
-
+            FrameStyle.Dashed => new HatchBrush(HatchStyle.Percent50, color, Color.Transparent),
+            _ => new SolidBrush(color),
+        };
         g.FillRegion(brush, resizeBorder);
         brush.Dispose();
     }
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DesignerUtils.cs
```diff
@@ -363,9 +353,9 @@ public static Size GetAdornmentDimensions(AdornmentType adornmentType)
 
     public static bool UseSnapLines(IServiceProvider provider)
     {
-        bool useSnapLines = true;
         object optionValue = null;
-        if (provider.GetService(typeof(DesignerOptionService)) is DesignerOptionService options)
+        DesignerOptionService options = provider.GetService<DesignerOptionService>();
+        if (options is not null)
```
lonitra, line 358: Consider using TryGetService here

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DesignerUtils.cs
```diff
@@ -374,43 +364,41 @@ public static bool UseSnapLines(IServiceProvider provider)
...
-        return optionValue;
+        DesignerOptionService designerOptionService = provider.GetService<DesignerOptionService>();
+        if (designerOptionService is not null)
+        {
+            PropertyDescriptor prop = designerOptionService.Options.Properties[name];
+            return prop?.GetValue(null);
+        }
+
+        IDesignerOptionService optionService = provider.GetService<IDesignerOptionService>();
+        if (optionService is not null)
```
lonitra, line 390: Ditto here with TryGetService

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DesignerUtils.cs
```diff
@@ -489,27 +477,15 @@ public static bool GenerateSnapShotWithWM_PRINT(Control control, ref Image image
...
-            case SelectionBorderGlyphType.Right:
-                bounds = new Rectangle(originalBounds.Right, originalBounds.Top - borderSize, borderSize, originalBounds.Height + 2 * borderSize);
-                break;
-            case SelectionBorderGlyphType.Body:
-                bounds = originalBounds;
-                break;
-        }
-
-        return bounds;
+        return type switch
```
lonitra, line 480: nit: body expression

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DesignerUtils.cs
```diff
@@ -666,7 +642,7 @@ public static string GetUniqueSiteName(IDesignerHost host, string name)
         }
 
         // Get the name creation service from the designer host
-        INameCreationService nameCreationService = (INameCreationService)host.GetService(typeof(INameCreationService));
+        INameCreationService nameCreationService = host.GetService<INameCreationService>();
         if (nameCreationService is null)
```
lonitra, line 646: Here as well with TryGetService

lonitra (commented):
Some small comments but otherwise LGTM!

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DesignerUtils.cs
halgab: I deliberately didn't use it here because it would be a behaviour change. Today with throw a NRE if `provider` is null. If I make the change to `TryGetService`, the method returns `true` for a null `provider`

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DesignerUtils.cs
halgab: `provider` would be checked twice for null. If we don't care about that, then I'll make the change

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DesignerUtils.cs
halgab: Here again, it's a behaviour change if `host` is null (NRE vs silent null return)

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DesignerUtils.cs
Tanya-Solyanik: Could you please throw a better exception if provider is null, and then use the TryGet method?

Tanya-Solyanik (approved):
Loos good!

lonitra (approved):
🚀

PR #10105, opened 2023-10-12

gpetrou:
## Proposed changes

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10105)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.Methods.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.Methods.cs
@@ -10372,7 +10372,7 @@ public partial class DataGridView
             top = rowDisplayRect.Top;
         }
 
-        rowDisplayRect = GetRowDisplayRectangle(hi, true /*cutOverflow*/);
+        rowDisplayRect = GetRowDisplayRectangle(hi, cutOverflow: true);
 
         if (rowDisplayRect.IsEmpty)
         {
@@ -10857,12 +10857,13 @@ public partial class DataGridView
                 return DataGridViewAutoSizeRowMode.RowHeader;
         }
     }
-#nullable disable
+
     private void MoveColumnHeadersOrRowResize(MouseEventArgs e)
     {
         _lastRowSplitBar = _currentRowSplitBar;
         _currentRowSplitBar = e.Y;
         Rectangle lastSplitBarRect = CalcRowResizeFeedbackRect(_lastRowSplitBar);
+        Debug.Assert(_editingPanel is not null);
         if (EditingControl is not null &&
             !_dataGridViewState1[State1_EditingControlHidden] &&
             _editingPanel.Bounds.IntersectsWith(lastSplitBarRect))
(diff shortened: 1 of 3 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewAutoSizeColumnsModeEventArgs.cs
```diff
@@ -5,10 +5,11 @@ namespace System.Windows.Forms;
 
 public class DataGridViewAutoSizeColumnsModeEventArgs : EventArgs
 {
-    public DataGridViewAutoSizeColumnsModeEventArgs(DataGridViewAutoSizeColumnMode[]? previousModes)
+    public DataGridViewAutoSizeColumnsModeEventArgs(DataGridViewAutoSizeColumnMode[] previousModes)
```
LeafShi1, line 8: It seems that this change caused the test case `Ctor_DataGridViewAutoSizeColumnModeArray` to fail verification. Maybe this file does not need to be changed.

LeafShi1:
Just one comment, other changes are OK in my opinion.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.Methods.cs
```diff
@@ -11692,9 +11724,10 @@ protected virtual void OnAutoSizeColumnsModeChanged(DataGridViewAutoSizeColumnsM
                             dataGridViewColumn.CachedThickness = dataGridViewColumn.Thickness;
                         }
 
-                        AutoResizeColumnInternal(dataGridViewColumn.Index,
-                                                 (DataGridViewAutoSizeColumnCriteriaInternal)autoSizeColumnMode,
-                                                 (((DataGridViewAutoSizeRowsModeInternal)_autoSizeRowsMode) & DataGridViewAutoSizeRowsModeInternal.AllColumns) == 0 /*fixedHeight*/);
+                        AutoResizeColumnInternal(
+                            dataGridViewColumn.Index,
+                            (DataGridViewAutoSizeColumnCriteriaInternal)autoSizeColumnMode,
+                            (((DataGridViewAutoSizeRowsModeInternal)_autoSizeRowsMode) & DataGridViewAutoSizeRowsModeInternal.AllColumns) == 0 /*fixedHeight*/);
```
lonitra, line 11730: Looks like this was missed
```suggestion
                        AutoResizeColumnInternal(
                            dataGridViewColumn.Index,
                            (DataGridViewAutoSizeColumnCriteriaInternal)autoSizeColumnMode,
                            fixedHeight: (((DataGridViewAutoSizeRowsModeInternal)_autoSizeRowsMode) & DataGridViewAutoSizeRowsModeInternal.AllColumns) == 0);
```
Could we actually use `HasFlags()` for input for `fixedHeight`?

lonitra: I changing to use `HasFlags()` here was missed, but I won't block on this. Could you address this in your next change?

lonitra (approved):

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.Methods.cs
gpetrou: We have many usages of the bitwise operators instead of `HasFlag` in the codebase. If you want to replace them, perhaps open a separate issue? As far as I remember though, this is not guaranteed to be optimized as per https://github.com/dotnet/runtime/issues/55455#issuecomment-877639565, so not sure if we should keep doing that.

PR #10110, opened 2023-10-13

gpetrou:
## Proposed changes

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10110)

### DIFF
--- a/src/System.Windows.Forms/src/PublicAPI.Shipped.txt
+++ b/src/System.Windows.Forms/src/PublicAPI.Shipped.txt
@@ -241,10 +241,10 @@ override System.Windows.Forms.PropertyGrid.BackgroundImage.set -> void
 ~override System.Windows.Forms.PropertyGrid.OnResize(System.EventArgs e) -> void
 ~override System.Windows.Forms.PropertyGrid.OnSystemColorsChanged(System.EventArgs e) -> void
 ~override System.Windows.Forms.PropertyGrid.OnVisibleChanged(System.EventArgs e) -> void
-~override System.Windows.Forms.PropertyGrid.Site.get -> System.ComponentModel.ISite
-~override System.Windows.Forms.PropertyGrid.Site.set -> void
-~override System.Windows.Forms.PropertyGrid.Text.get -> string
-~override System.Windows.Forms.PropertyGrid.Text.set -> void
+override System.Windows.Forms.PropertyGrid.Site.get -> System.ComponentModel.ISite?
+override System.Windows.Forms.PropertyGrid.Site.set -> void
+override System.Windows.Forms.PropertyGrid.Text.get -> string!
+override System.Windows.Forms.PropertyGrid.Text.set -> void
 override System.Windows.Forms.TreeNode.ToString() -> string!
 override System.Windows.Forms.TreeView.BackgroundImage.get -> System.Drawing.Image?
 override System.Windows.Forms.TreeView.BackgroundImage.set -> void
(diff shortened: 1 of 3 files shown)


Epica3055 (approved):

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGrid.cs
```diff
@@ -847,18 +847,18 @@ public PropertySort PropertySort
 
             if ((value & PropertySort.Categorized) != 0)
             {
-                newButton = _viewSortButtons[CategorySortButtonIndex];
+                newButton = _viewSortButtons![CategorySortButtonIndex];
```
halgab, line 850: I think `_viewSortButtons` can be declared as non-null. Method `void SetupToolbar(bool fullRebuild)` can be marked as `[MemberNotNull(nameof(_viewSortButtons))]`, method `void SetupToolbar()` as well, and it's called in the class constructor

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGrid.cs
```diff
@@ -1098,9 +1099,9 @@ public object[] SelectedObjects
             // If you select an events tab for your designer and double click to go to code, it should
             // be the events tab when you get back to the designer. Check for that state here and
             // make sure we select and refresh that tab when we load.
-            if (_selectedObjects.Length > 0 && GetFlag(Flags.ReInitTab))
+            if (_selectedObjects!.Length > 0 && GetFlag(Flags.ReInitTab))
```
halgab, line 1102: I'm wondering if we can't make `_selectedObjects` non-nullable by initiating it as `Array.Empty<object>()`. Looking at its usages, I think it's functionally equivalent

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGrid.cs
```diff
@@ -1098,9 +1099,9 @@ public object[] SelectedObjects
...
             {
-                object designerKey = ActiveDesigner;
+                object? designerKey = ActiveDesigner;
```
halgab, line 1104: Not sure we want to do anything about it here, but just to raise awareness: `designerKey` doesn't look to be used anywhere.

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGrid.cs
```diff
@@ -1330,15 +1333,15 @@ private int AddImage(Bitmap image)
             image = DpiHelper.CreateResizedBitmap(image, s_normalButtonSize);
         }
 
-        int result = _normalButtonImages.Images.Count;
+        int result = _normalButtonImages!.Images.Count;
```
halgab, line 1336: `_normalButtonImages` also seems to be set in `void SetupToolbar()`, in which case it can be declared as non-null

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGrid.cs
Tanya-Solyanik: Let's make this change in a dedicated PR if at all.

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGrid.cs
Tanya-Solyanik: We could use a discard for the desigenrKey, but we should invoke ActiveDesigner anyway to initialize _designerHost.

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGrid.cs
```diff
@@ -1546,13 +1553,13 @@ internal void AddTab(Type tabType, PropertyTabScope scope, object @object = null
         {
             try
             {
-                object[] tabComponents = tab.Components;
+                object[]? tabComponents = tab.Components;
                 int oldArraySize = tabComponents is null ? 0 : tabComponents.Length;
 
                 object[] newComponents = new object[oldArraySize + 1];
                 if (oldArraySize > 0)
```
Tanya-Solyanik, line 1560: test tabComponents for null explicitly to avoid the bang operator.

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGrid.cs
gpetrou: This method checks parameters such as `_tabsDirty` and `fullRebuild`. We don't usually use `MemberNotNull` in such cases.

halgab: Sorry, I don't know why I missed the beginning of the method. Do we want to introduce a `ForceSetupToolbar` method consisting of everything that's called after this inital check, annotate it with `MemberNotNull` and call it directly in the constructor? Or is that too much effort to remove a few `!`?

lonitra:
@Tanya-Solyanik gentle ping to keep this on your radar to take to completion

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGrid.cs
Tanya-Solyanik: @gpetrou  - had you tried @halgab's suggestion?

> Do we want to introduce a ForceSetupToolbar method consisting of everything that's called after this inital check, annotate it with MemberNotNull and call it directly in the constructor?

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGrid.cs
Tanya-Solyanik: @gpetrou  - another alternative is to factor out the !isSame condition into a helper function and decorate that function with a MemberNotNull attribute for _selectedObjects.

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGrid.cs
Tanya-Solyanik: @gpetrou  - had you tried adding MemberNotNull attribute for _normalButtonImages to the SetupToolbar method?

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGrid.cs
gpetrou: No, I didn't. But do we really need such a change?

Tanya-Solyanik: It's an alternative to multiple `!`, and it did not seem to be a big change, just split a function in 2. and edit the call site in the constructor.

gpetrou: In which multiple `!` are you referring to? The ones in `PropertySort`? I don't think that these would be affected by such a change, right?

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGrid.cs
gpetrou: This is only called inside `SetupToolbar`, right? So, I don't think this would make any difference.

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGrid.cs
Tanya-Solyanik: I meant all fields that are allocated in SetupToolbar. _viewSortButtons collection and content, _normalButtonImages, _alphaBitmap, _categoryBitmap, _propertyPageBitmap. We could make them non-nullable. That would include _viewSortButtons in PropertySort getter. We never set fields that are allocated in SetupToolbar to null unless we are disposing them, we only allocate or re-allocate these fields.
If you feel that this is too big of a change, it can be done in the next PR.

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGrid.cs
Tanya-Solyanik: You are right!

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGrid.cs
Tanya-Solyanik: We can follow up on refactoring later - https://github.com/dotnet/winforms/issues/10686

Tanya-Solyanik (approved):
Thank you!

PR #9970, opened 2023-09-21
Fix Narrator cannot announce correctly and getting incorrect rectanges

SimonZhao888:
Fixes #9828

## Proposed changes

- Base on [PR 9960](https://github.com/dotnet/winforms/pull/9960), we need update the ITextRangeProvider.GetBoundingRectangles() method to make sure the rectangeles for the selection text in TextBox can be correctly drawn.
- Update the method of ProcessCmdKey(ref Message m, Keys keyData) in TextBox class.

## Customer Impact

- When using keyboard: Ctrl+A or mouse to choose all texts in TextBox/ComboBox/DomainUpDown/NumericUpDown, Narrator cannot focus entire texts and announce "Document selected" to user

## Regression? 

- Yes 

## Risk

- Minimal

### Before

image

### After

Keep the same behavior with .Net 7.0, narrator can focus entire texts and announce "Document selected" to user.
image

## Test methodology 

- Manual (Narrator)

## Test environment(s) 

- 8.0.100-preview.7.23376.3

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/9970)

PR #10128, opened 2023-10-16
Fix Narrator cannot select ComboBox's text area with Ctrl+a

### LINKED ISSUE #10007: Accessibility: Narrator cannot focus entire texts and announce "Document selected" when pressing Ctrl+A to choose all texts in ComboBox
### .NET version

.Net 8.0: 8.0.100-rtm.23477.34

Yes

No, this issue can repro in .Net 6.0 and .Net 7.0.

### Issue description

When using keyboard: Ctrl+A or mouse to choose all texts in ComboBox, Narrator cannot focus entire texts and announce "Document selected" to user.

The original issue: https://github.com/dotnet/winforms/issues/9828

The result in .Net 8.0 RC2 latest build:
image
test80trm

The result in .Net framework:
image
test481framework

### Steps to reproduce

1. Create a winforms .Net Core application.
2. Add a ComboBox control to form designer with text.
3. Build and run.
4. Use Narrator to focus on the individual text, then presse Ctrl+A to select all texts.

ricardobossan:
Fixes #10007

## Proposed changes

- Base on PR 9970 https://github.com/dotnet/winforms/pull/9970, we need update the ComboBox.Select() method to make sure the rectangeles for the selection text in TextBox can be correctly drawn.
- Update the method of ProcessCmdKey(ref Message m, Keys keyData) in TextBox class.

## Customer Impact

- When using keyboard: Ctrl+A or mouse to choose all texts in ComboBox, Narrator cannot focus entire texts and announce `<say-as interpret-as="spell">a</say-as>` to user

## Regression?

- Yes

## Risk

- Minimal

### Before

issue_10007

### After

issue_10007_after

## Test methodology

- Manual (Narrator)

## Test environment(s)

- 9.0.100-alpha.1.23511.2

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10128)


src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.cs
```diff
@@ -3078,6 +3078,12 @@ protected override bool ProcessCmdKey(ref Message msg, Keys keyData)
     {
         bool returnedValue = base.ProcessCmdKey(ref msg, keyData);
 
+        if (DropDownStyle == ComboBoxStyle.DropDown && keyData == (Keys.Control | Keys.A))
+        {
+            Select(0, Text.Length);
+            SelectedText = Text;
```
Tanya-Solyanik, line 3084: It's not clear to me why we are not exiting after processing this command

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.cs
```diff
@@ -3390,7 +3396,7 @@ public void Select(int start, int length)
             throw new ArgumentOutOfRangeException(nameof(length), length, string.Format(SR.InvalidArgument, nameof(length), length));
         }
 
-        PInvoke.SendMessage(this, PInvoke.CB_SETEDITSEL, (WPARAM)0, LPARAM.MAKELPARAM(start, end));
+        PInvoke.SendMessage(this, PInvoke.CB_SETEDITSEL, (WPARAM)start, (LPARAM)end);
```
Tanya-Solyanik, line 3399: This does not look like a correct change, as wParam is not used for this message - https://learn.microsoft.com/windows/win32/Controls/cb-seteditsel

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.cs
```diff
@@ -3078,6 +3078,12 @@ protected override bool ProcessCmdKey(ref Message msg, Keys keyData)
     {
         bool returnedValue = base.ProcessCmdKey(ref msg, keyData);
 
+        if (DropDownStyle == ComboBoxStyle.DropDown && keyData == (Keys.Control | Keys.A))
```
Tanya-Solyanik, line 3081: If the base class was able to process this key, we can skip our own processing.

Tanya-Solyanik, line 3081:
```suggestion
        if (!returnedValue && DropDownStyle == ComboBoxStyle.DropDown && keyData == (Keys.Control | Keys.A))
```

src/System.Windows.Forms/src/System/Windows/Forms/ComboBox.cs
```diff
@@ -3078,6 +3078,18 @@ protected override bool ProcessCmdKey(ref Message msg, Keys keyData)
     {
         bool returnedValue = base.ProcessCmdKey(ref msg, keyData);
 
+        if (!returnedValue && keyData == (Keys.Control | Keys.A))
+        {
+            Select(0, Text.Length);
+            SelectedText = Text;
+            int boundaryStart = ClientUtils.GetWordBoundaryStart(Text, SelectionStart);
```
Tanya-Solyanik, line 3085: In the Ctrl+A case we don't have to select complete words, we are selecting everything as is, so there is no need to search for the word boundary.

ricardobossan: I added that code because otherwise (code commented out bellow, to demonstrate), when pressing `ctrl+a`, it would not select the full text and just focus at the end of the text and then narrator buddy would say "new line" at the end, and, as seen bellow :

image

withoutcode

However, adding that code (uncommented bellow), the whole text would be properly selected on pressing `ctrl+a`:

image

withcode

Tanya-Solyanik: >However, adding that code (uncommented bellow), the whole text would be properly selected on pressing ctrl+a:

Does this work when content of the combobox has some non-word characters for example, punctuation and whitespaces.

ricardobossan: No. It did not work once I insert spaces and special characters.

So now I just set SelectionStart to 0 and SelectionLength to the length of the whole text and it works:

image

issue1007fix

Tanya-Solyanik (approved):
Thank you!

PR #10131, opened 2023-10-17
Fix issue 9999: Update IconEditor to offset the icon within the property grid entry

### LINKED ISSUE #9999: The icon property of the NotifyIcon control cannot be modified in DemoConsole application
### .NET version

Not tested/verified

No, it doesn't work in .Net 8.0 from Release/8.0 branch of Winforms repo

### Issue description

In [DemoConsole](https://github.com/dotnet/winforms/tree/main/src/System.Windows.Forms/tests/IntegrationTests/DesignSurface/DemoConsole) test app of Winforms repo, when changing the icon property of the NotifyIcon control, cannot be modified.

**Actual result** in [DemoConsole](https://github.com/dotnet/winforms/tree/main/src/System.Windows.Forms/tests/IntegrationTests/DesignSurface/DemoConsole) test app:

https://github.com/dotnet/winforms/assets/56664234/dab536e8-e545-4370-a641-0008ba2bc5a8

**Expected result:**
When changing the icon property of the NotifyIcon control, can be modified.

https://github.com/dotnet/winforms/assets/56664234/df710638-e2e2-477a-85da-5dd0f85c5f1f

### Steps to reproduce

LeafShi1:
Fixes #9999

## Proposed changes

-  Update function Icon.Draw to use targetRect to calculate offset

## Customer Impact

- The icon property of the NotifyIcon control can be modified

## Regression? 

-  No

## Risk

- Minimal

### Before

The icon property of the NotifyIcon control cannot be modified

https://github.com/dotnet/winforms/assets/56664234/dab536e8-e545-4370-a641-0008ba2bc5a8

### After
The icon property of the NotifyIcon control can be modified

AfterChange

## Test methodology 

- Manually (Verified in Demo project)

## Test environment(s) 

- .net 9.0.0-alpha.1.23515.3

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10131)

### DIFF
--- a/src/System.Drawing.Common/src/System/Drawing/Icon.cs
+++ b/src/System.Drawing.Common/src/System/Drawing/Icon.cs
@@ -426,7 +426,9 @@ public sealed partial class Icon : MarshalByRefObject, ICloneable, IDisposable, 
         Rectangle copy = targetRect;
 
         using Matrix transform = graphics.Transform;
-        PointF offset = transform.Offset;
+        PointF point = new PointF(targetRect.X, targetRect.Y);
+        transform.TransformPoints(new PointF[] { point });
+        PointF offset = point;
         copy.X += (int)offset.X;
         copy.Y += (int)offset.Y;
 
--- a/src/System.Drawing.Common/src/System/Drawing/Icon.cs
+++ b/src/System.Drawing.Common/src/System/Drawing/Icon.cs
@@ -426,9 +426,7 @@ public sealed partial class Icon : MarshalByRefObject, ICloneable, IDisposable, 
         Rectangle copy = targetRect;
 
         using Matrix transform = graphics.Transform;
-        PointF point = new PointF(targetRect.X, targetRect.Y);
-        transform.TransformPoints(new PointF[] { point });
-        PointF offset = point;
+        PointF offset = transform.Offset;
         copy.X += (int)offset.X;
         copy.Y += (int)offset.Y;
 
--- a/src/System.Windows.Forms.Design/src/System/Drawing/Design/IconEditor.cs
+++ b/src/System.Windows.Forms.Design/src/System/Drawing/Design/IconEditor.cs
@@ -116,6 +116,10 @@ public class IconEditor : UITypeEditor
 
         // If icon is smaller than rectangle, just center it unscaled in the rectangle.
         Rectangle rectangle = e.Bounds;
+        Graphics g = e.Graphics;
+        using Matrix transform = g.Transform;
+        rectangle.Offset(new Point(-(int)transform.OffsetX, -(int)transform.OffsetY));
+
         if (icon.Width < rectangle.Width)
         {
             rectangle.X = (rectangle.Width - icon.Width) / 2;
--- a/src/System.Windows.Forms.Design/src/System/Drawing/Design/IconEditor.cs
+++ b/src/System.Windows.Forms.Design/src/System/Drawing/Design/IconEditor.cs
@@ -117,8 +117,9 @@ public class IconEditor : UITypeEditor
         // If icon is smaller than rectangle, just center it unscaled in the rectangle.
         Rectangle rectangle = e.Bounds;
         Graphics g = e.Graphics;
-        using Matrix transform = g.Transform;
-        rectangle.Offset(new Point(-(int)transform.OffsetX, -(int)transform.OffsetY));
+        using var transform = g.Transform;
+        rectangle.X -= (int)transform.OffsetX;
+        rectangle.Y -= (int)transform.OffsetY;
 
         if (icon.Width < rectangle.Width)
         {


Epica3055 (commented):
looks good 👍

src/System.Drawing.Common/src/System/Drawing/Icon.cs
```diff
@@ -426,7 +426,9 @@ internal void Draw(Graphics graphics, Rectangle targetRect)
         Rectangle copy = targetRect;
 
         using Matrix transform = graphics.Transform;
-        PointF offset = transform.Offset;
+        PointF point = new PointF(targetRect.X, targetRect.Y);
+        transform.TransformPoints(new PointF[] { point });
```
Tanya-Solyanik, line 430: This function is used in multiple applications besides the property grid, this change will regress applications that depend on that behavior. Is it possible to fix this but in the sample application instead?

LeafShi1: > This function is used in multiple applications besides the property grid, this change will regress applications that depend on that behavior. Is it possible to fix this but in the sample application instead?

I'm trying to do this, but I'm running into some problems, some of the methods I'm trying to override have classes that are Sealed and can't be inherited, for example `NotifyIcon.cs`，so I still need more time to investigate this issue.

Tanya-Solyanik: No problem, please change this PR to draft then.

Tanya-Solyanik: @LeafShi1  - one thing to do is to investigate when this transform is set on the graphics object.  Set selected object in property grid to a very simple class that has only a single icon property, then the propertygrid will have a single `GridEntry` line and it will be easier to set breakpoints in the drawing methods. And then look up history of the file where this transform is added
 
```
public class A
{
    public A(){};
    public Icon AIcon {get;set;}
}
```

lonitra:
This PR has not had any activity in a while. Should this still be kept active?

LeafShi1:
> This PR has not had any activity in a while. Should this still be kept active?

Please keep active.

src/System.Drawing.Common/src/System/Drawing/Icon.cs
LeafShi1: The Graphics value is changes from https://github.com/dotnet/winforms/blob/558619be6e585825c6f7eb768f449583c1dd720f/src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGridInternal/PropertyGridView.cs#L1178

After this, The location of the graphics.ClipBounds will be changed.

Tanya-Solyanik: I would offset the rectangle in the IconEditor - https://github.com/dotnet/winforms/blob/4052920b2695a15c81a83ea7dd6099b3e8b748e5/src/System.Windows.Forms.Design/src/System/Drawing/Design/IconEditor.cs#L110
```
    public override void PaintValue(PaintValueEventArgs e)
    {
        if (e?.Value is not Icon icon)
        {
            return;
        }

        // If icon is smaller than rectangle, just center it unscaled in the rectangle.
        Rectangle rectangle = e.Bounds;
        Graphics g = e.Graphics;
        using Matrix transform = g.Transform;
        rectangle.Offset(new Point(-(int)transform.OffsetX, -(int)transform.OffsetY));
```
Offset is applied by Icon.Draw, Icon.cs  line434, unnecessarily in this case, as you pointed out in the draft PR, because graphics is already clipped to the PropertyGridView entry, value part only (the right hand side of the row). I think that then WindowGraphics is created from that clipped graphics and the transform is applied again. 
However "undoing" transform on the Icon class level will potentially break other application, this is why I suggest fixing this issue on the Edit level, because we know that editor is invoked in the propertyGridView context.

Tanya-Solyanik: @LeafShi1  - FYI, to debug drawing code, it's convenient to disable double buffering in the control you are looking at. https://github.com/dotnet/winforms/blob/4052920b2695a15c81a83ea7dd6099b3e8b748e5/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs#L1968 Then you will see each drawing command on the screen right away. For example - this is how icon is drawn right after `e.Graphics.DrawIcon(icon, rectangle);`  call, before the background is filled in.

src/System.Windows.Forms.Design/src/System/Drawing/Design/IconEditor.cs
```diff
@@ -116,6 +116,10 @@ public override void PaintValue(PaintValueEventArgs e)
 
         // If icon is smaller than rectangle, just center it unscaled in the rectangle.
         Rectangle rectangle = e.Bounds;
+        Graphics g = e.Graphics;
+        using Matrix transform = g.Transform;
+        rectangle.Offset(new Point(-(int)transform.OffsetX, -(int)transform.OffsetY));
```
Tanya-Solyanik, line 121: Something like this would avoid a `new`
        rectangle.X -= (int)transform.OffsetX;
        rectangle.Y -= (int)transform.OffsetY;

src/System.Windows.Forms.Design/src/System/Drawing/Design/IconEditor.cs
```diff
@@ -116,6 +116,10 @@ public override void PaintValue(PaintValueEventArgs e)
 
         // If icon is smaller than rectangle, just center it unscaled in the rectangle.
         Rectangle rectangle = e.Bounds;
+        Graphics g = e.Graphics;
+        using Matrix transform = g.Transform;
```
Tanya-Solyanik, line 120:
```suggestion
        using var transform = g.Transform;
```

Tanya-Solyanik (approved):
Thank you!

PR #10145, opened 2023-10-18
Enable nullability in EditorServiceContext

halgab:
## Proposed changes

- Enable nullability in `EditorServiceContext`

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10145)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/EditorServiceContext.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/EditorServiceContext.cs
@@ -20,17 +20,17 @@ internal class EditorServiceContext : IWindowsFormsEditorService, ITypeDescripto
     internal EditorServiceContext(ComponentDesigner designer, PropertyDescriptor? prop)
     {
         _designer = designer;
-        _targetProperty = prop!;
         if (prop is null)
         {
             prop = TypeDescriptor.GetDefaultProperty(designer.Component);
-            if (prop is not null && typeof(ICollection).IsAssignableFrom(prop.PropertyType))
+            if (!typeof(ICollection).IsAssignableFrom(prop?.PropertyType))
             {
-                _targetProperty = prop;
+                prop = null;
+                Debug.Fail("Need PropertyDescriptor for ICollection property to associate collection editor with.");
             }
         }
 
-        Debug.Assert(_targetProperty is not null, "Need PropertyDescriptor for ICollection property to associate collection editor with.");
+        _targetProperty = prop;
     }
 
     internal EditorServiceContext(ComponentDesigner designer, PropertyDescriptor? prop, string newVerbText) : this(designer, prop)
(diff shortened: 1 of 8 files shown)


src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/EditorServiceContext.cs
```diff
@@ -16,18 +14,13 @@ namespace System.Windows.Forms.Design;
...
-    {
-        _designer = designer;
-    }
-
-    internal EditorServiceContext(ComponentDesigner designer, PropertyDescriptor prop)
+    internal EditorServiceContext(ComponentDesigner designer, PropertyDescriptor? prop)
     {
         _designer = designer;
-        _targetProperty = prop;
+        _targetProperty = prop!;
```
Epica3055, line 23: This constructor `EditorServiceContext` permits the parameter  `PropertyDescriptor`to be null. But inside the constructor , it requires `PropertyDescriptor` not to be null.

So , maybe it's better to mark parameter  `PropertyDescriptor` not nullable
`internal EditorServiceContext(ComponentDesigner designer, PropertyDescriptor prop)`

halgab: I changed the code to make it clearer why `prop` can be null and `_targetProperty` cannot

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/EditorServiceContext.cs
```diff
@@ -16,48 +14,43 @@ namespace System.Windows.Forms.Design;
 internal class EditorServiceContext : IWindowsFormsEditorService, ITypeDescriptorContext
 {
     private readonly ComponentDesigner _designer;
-    private IComponentChangeService _componentChangeService;
+    private IComponentChangeService? _componentChangeService;
     private readonly PropertyDescriptor _targetProperty;
 
-    internal EditorServiceContext(ComponentDesigner designer)
```
halgab, line 22: This wasn't used anywhere and made the annotation of `_targetProperty` problematic

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/EditorServiceContext.cs
```diff
@@ -16,48 +14,43 @@ namespace System.Windows.Forms.Design;
...
+            if (!typeof(ICollection).IsAssignableFrom(prop?.PropertyType))
             {
-                _targetProperty = prop;
+                prop = null;
+                Debug.Fail("Need PropertyDescriptor for ICollection property to associate collection editor with.");
             }
         }
 
-        Debug.Assert(_targetProperty is not null, "Need PropertyDescriptor for ICollection property to associate collection editor with.");
+        _targetProperty = prop;
```
lonitra, line 33: It still seems like `_targetProperty` could be set to null here. Should it be annotated as nullable?

halgab: It cannot be nullable in debug. But I guess that's not enough. I'll rollback my last change, and make it nullable.

lonitra: In this case, I think it is better to make it nullable if we are not sure it won't be null in release

lonitra (approved):
🚢

Issue #10150, opened 2023-10-19
Remove OS Condition in InputLanguageTests.InputLanguage_FromCulture_SupplementalInputLanguages_Expected


lonitra:
`InputLanguageTests.InputLanguage_FromCulture_SupplementalInputLanguages_Expected', specifically N'ko input, has started failing after CI machine updated to Windows 11 version 21H2. This failure does not repro on Windows 11 version 22H2. A condition has been placed on the test to skip N'ko input if OS is not on version Windows 11, 22H2. This is tracking issue to remove condition when CI machines update to Windows 11 version 22H2

PR #10151, opened 2023-10-19
[release/8.0] Disable N'Ko in InputLanguageTests

lonitra:
This test has started failing after CI machine update. This is blocking dependency code flow among other PRs. I've created https://github.com/dotnet/winforms/issues/10150 to track investigating and re-enabling this test.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10151)


dreddy-work:
Don't disable it in main. Just do it in release to unblock runtime flow. Main can wait for runtime update. This is the result of OS update and we need to check if this was an intentional change.

lonitra:
I will retarget this to release then

Tanya-Solyanik (approved):

PR #10154, opened 2023-10-19
Add OS Condition for InputLanguageTests.InputLanguage_FromCulture_SupplementalInputLanguages_Expected

lonitra:
Related: https://github.com/dotnet/winforms/issues/10150

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10154)


dreddy-work (changes requested):
I don't see reason why you should. You could always condition to run on higher version OS (especially when we know it doesn't repo on latest) Also link issue as comment for future reference to remove this condition when close issue.

dreddy-work (approved):
LGTM.

PR #10158, opened 2023-10-20
Enable nullability in DataGridViewHeaderCell

gpetrou:
## Proposed changes

- Enable nullability in DataGridViewHeaderCell.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10158)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewHeaderCell.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewHeaderCell.cs
@@ -1,8 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-#nullable disable
-
 using System.ComponentModel;
 using System.Drawing;
 
@@ -96,11 +94,11 @@ public partial class DataGridViewHeaderCell : DataGridViewCell
         }
     }
 
-    internal Bitmap FlipXPThemesBitmap
+    internal Bitmap? FlipXPThemesBitmap
     {
         get
         {
-            return (Bitmap)Properties.GetObject(s_propFlipXPThemesBitmap);
+            return (Bitmap?)Properties.GetObject(s_propFlipXPThemesBitmap);
         }
         set
         {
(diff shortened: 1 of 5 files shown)


gpetrou:
@lonitra any thoughts on this one?

lonitra (dismissed):
LGTM

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewHeaderCell.cs
```diff
@@ -487,23 +471,27 @@ protected override object GetValue(int rowIndex)
...
 
     protected override bool MouseLeaveUnsharesRow(int rowIndex)
     {
+        Debug.Assert(DataGridView is not null);
         return ButtonState != ButtonState.Normal && DataGridView.ApplyVisualStylesToHeaderCells;
     }
 
     protected override bool MouseUpUnsharesRow(DataGridViewCellMouseEventArgs e)
     {
+        Debug.Assert(DataGridView is not null);
```
lonitra, line 494: Looks like these `Debug.Assert` is causing test failure in the debug test runs since we will throw a difference exception. Perhaps we should update the related tests to assert that it throws `InvalidOperationException` instead of NRE if we are running tests on debug

gpetrou: I just used null-forgiving operator instead to make the Debug tests to work as before. I guess another approach could be to introduce breaking changes by checking `if DataGridView is null` and return false.

lonitra: Yeah actually I would prefer we check if `DataGridView` is null and return false. I don't think anyone is expecting this to throw NRE based on [docs](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.datagridviewheadercell.mouseupunsharesrow?view=windowsdesktop-7.0)

gpetrou: I added a separate commit for this. Is this what you had in mind?

lonitra: Yes, thank you

lonitra (approved):

gpetrou:
@lonitra perhaps we could merge this?

lonitra:
breaking change doc request created https://github.com/dotnet/docs/issues/38283

PR #10172, opened 2023-10-22
Review usage of string.Split across codebase

halgab:
There is no issue tied to this, but I encountered several times pieces of code that could use more recent overloads of `string.Split` to reduce allocations, and I thought I'd let the team decide if it's worth the effort or not.

I split the work in four distinct commits. Each one can be spun out in a separate PR or simply discarded

## Proposed changes

- Use `StringSplitOptions.TrimEntries` where relevant. This includes a bit of cleanup in BindingSource that helps readability quite a bit
- Remove useless usage of string.Split. Sometimes it's not even more readable anyway
- Introduce and use TypeConverterHelper. This factorise a pattern often found in `TypeConverter`s in a way that reduces allocations
- Remove allocs in LinkUtilities.GetIELinkBehavior

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10172)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/System/ComponentModel/TypeConverterHelper.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/ComponentModel/TypeConverterHelper.cs
@@ -1,0 +1,31 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Globalization;
+
+namespace System.ComponentModel;
+
+internal static class TypeConverterHelper
+{
+    public static bool TryParseAsSpan<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)] T>(ITypeDescriptorContext? context, CultureInfo? culture, ReadOnlySpan<char> text, Span<T> output)
+    {
+        culture ??= CultureInfo.CurrentCulture;
+
+        Span<Range> tokens = stackalloc Range[output.Length + 1];
+        int tokensCount = text.Split(tokens, culture.TextInfo.ListSeparator[0]);
+
+        if (tokensCount != output.Length)
+        {
+            return false;
+        }
+
+        TypeConverter converter = TypeDescriptor.GetConverter(typeof(T));
+        for (int i = 0; i < output.Length; i++)
+        {
+            // Note: ConvertFromString will raise exception if value cannot be converted.
+            output[i] = (T)converter.ConvertFromString(context, culture, text[tokens[i]].ToString())!;
+        }
+
+        return true;
+    }
+}
--- a/src/System.Windows.Forms.Primitives/src/System/ComponentModel/TypeConverterHelper.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/ComponentModel/TypeConverterHelper.cs
@@ -7,11 +7,20 @@ namespace System.ComponentModel;
 
 internal static class TypeConverterHelper
 {
+    /// <summary>
+    /// Converts the given text to list of objects, using the specified context and culture information.
+    /// </summary>
+    /// <param name="context">An ITypeDescriptorContext that provides a format context.</param>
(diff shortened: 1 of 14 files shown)


src/System.Windows.Forms.Primitives/src/System/ComponentModel/TypeConverterHelper.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace System.ComponentModel;
+
+internal static class TypeConverterHelper
+{
+    public static bool TryParseAsSpan<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)] T>(ITypeDescriptorContext? context, CultureInfo? culture, ReadOnlySpan<char> text, Span<T> output)
+    {
+        culture ??= CultureInfo.CurrentCulture;
+
+        Span<Range> tokens = stackalloc Range[output.Length + 1];
```
elachlan, line 14: Is the `output.length` always < 1024?

src/System.Windows.Forms/src/System/Windows/Forms/LinkUtilities.cs
```diff
@@ -36,15 +36,17 @@ private static Color GetIEColor(string name)
 
             if (s is not null)
             {
-                string[] rgbs = s.Split(new char[] { ',' });
-                int[] rgb = new int[3];
+                Span<Range> rgbs = stackalloc Range[4];
```
lonitra, line 39: Curious why we are doing 4?

src/System.Windows.Forms.Primitives/src/System/ComponentModel/TypeConverterHelper.cs
lonitra: Yes, please add an assert here :)

src/System.Windows.Forms.Primitives/src/System/ComponentModel/TypeConverterHelper.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Globalization;
+
+namespace System.ComponentModel;
+
+internal static class TypeConverterHelper
+{
+    public static bool TryParseAsSpan<[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)] T>(ITypeDescriptorContext? context, CultureInfo? culture, ReadOnlySpan<char> text, Span<T> output)
```
lonitra, line 10: Could you add docs for this?

lonitra (commented):
Looks good! 🥳 Just some minor asks

src/System.Windows.Forms/src/System/Windows/Forms/LinkUtilities.cs
halgab: Because if we do 3, we change behaviour. Before this PR, it's a simple split, which doesn't make any assumption on the number of parts. If we split in 3 parts when there are actually more than 3, the last one will end up having extra `,` in it. The 4 avoids that

src/System.Windows.Forms.Primitives/src/System/ComponentModel/TypeConverterHelper.cs
halgab: Reading this again, I thought it would be more robust to use `BufferScope`. I can rollback to the previous version and add an assert if you prefer

src/System.Windows.Forms/src/System/Windows/Forms/LinkUtilities.cs
lonitra: I see, thank you for the explanation!

lonitra (approved):

PR #10177, opened 2023-10-23
Convert IRawElementProviderSimple to Cswin32

lonitra:

This change converts `IRawElementProviderSimple` to use Cswin32 definition. This interface includes definition of `GetPropertyValue`, which now returns `VARIANT` instead of `object`, so the bulk of this change involves casting what was returned from `GetPropertyValue` to type `VARIANT` in XAccessibleObject. Note we cannot directly cast enums/structs to `VARIANT` so these need to be first casted to what they really represent e.g. `int` or `nint` before casting to `VARIANT`. 
Other changes include:
- Convert methods: `UiaDisconnectProvider` , `UiaHostProviderFromHwnd` , `UiaRaiseAutomationEvent`, `UiaRaiseAutomationPropertyChangedEvent`, `UiaRaiseNotificationEvent` , `UiaRaiseStructureChangedEvent` , `UiaReturnRawElementProvider` to Cswin32
- Add `IsWindows10_1709OrGreater()` API since `UiaRaiseNotificationEvent` requires it. We didn't have this check before, so I also kept old definition of `UiaRaiseNotificationEvent` in cases where the OS version requirement is not met
- Add new constructor to `SafeArrayScope` that takes an array and add test
- Update XAccessibleObjectTests since `GetPropertyValue` no longer returns `object`

I tried to separate the test changes from the actual changes to make this somewhat easier to digest for review.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10177)


src/System.Windows.Forms.Primitives/src/Interop/OleAut32/Interop.VARIANT.cs
```diff
@@ -963,6 +963,18 @@ private static Span<T> GetSpan<T>(Array array)
...
 
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static explicit operator nint(VARIANT value)
+        => value.vt == VT_INT ? value.data.intVal : ThrowInvalidCast<nint>();
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static explicit operator VARIANT(nint value)
+        => new()
+        {
+            vt = VT_INT,
```
Tanya-Solyanik, line 974: Why is this not a VT_INT_PTR? https://learn.microsoft.com/windows/win32/api/wtypes/ne-wtypes-varenum

```
VT_INT_PTR
Value: 37
A signed machine register size width.
```

lonitra: We use `Marshal.GetNativeVariantForObject` as a fallback in `VARIANT.FromObject()` when we don't have a conversion defined for the object https://github.com/dotnet/winforms/blob/b2685139fddc05467bf3ab21e24fc8828e9e1b8c/src/System.Windows.Forms.Primitives/src/Interop/OleAut32/Interop.VARIANT.cs#L972-L1010

I followed the variant definition given back by `Marshal.GetNativeVariantForObject` when it is called with a `nint`. In the existing tests, as well as the new test I wrote in `VARIANTTests.cs`, `Marshal.GetNativeVariantForObject` gives back a VARIANT with `VT_INT` for the vt when passing a `nint`.

src/System.Windows.Forms/src/System/Windows/Forms/AccessibleObject.cs
```diff
@@ -2202,12 +2194,11 @@ MemberInfo[] IReflect.GetMembers(BindingFlags bindingAttr)
...
+            HRESULT result = PInvoke.UiaRaiseNotificationEvent(
                 this,
                 notificationKind,
                 notificationProcessing,
-                notificationText,
-                string.Empty);
+                notificationText);
             return result == HRESULT.S_OK;
         }
         catch (EntryPointNotFoundException)
```
Tanya-Solyanik, line 2204: This catch and a static that it sets is another way to ensure that this API is invoked only on the OSs that implement the notification API. Feel free to replace `s_notificationEventAvailable` with IsWindows10_1709OrGreater, but we don't need both.

Tanya-Solyanik: I didn't mean to replace inplace, your original approach was cleaner because OS version check was outside the "business logic" and avoided a build warning, keeping version check in Interop and removing exception handling here is fine.

src/System.Windows.Forms.Primitives/src/Interop/OleAut32/Interop.VARIANT.cs
Tanya-Solyanik: What does `GetNativeVariantForObject` return for a C# IntPtr object. I didn't find docs about VT_INT, besides the one that said `integer`. Also, I don't know if VT produced by `GetNativeVariantForObject` depends on the bitness of the app, do you know?

src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlAccessibleObject.cs
```diff
@@ -483,41 +483,42 @@ internal override bool IsPatternSupported(UIA_PATTERN_ID patternId)
...
                 UIA_PROPERTY_ID.UIA_IsKeyboardFocusablePropertyId when
                     Owner?.SupportsUiaProviders ?? false
-                    => Owner.CanSelect,
+                    => (VARIANT)Owner.CanSelect,
                 UIA_PROPERTY_ID.UIA_LiveSettingPropertyId => Owner is IAutomationLiveRegion owner
-                    ? owner.LiveSetting
+                    ? (VARIANT)(int)owner.LiveSetting
                     : base.GetPropertyValue(propertyID),
-                UIA_PROPERTY_ID.UIA_NativeWindowHandlePropertyId => (nint)(Owner?.InternalHandle ?? HWND.Null),
+                UIA_PROPERTY_ID.UIA_NativeWindowHandlePropertyId => new() { vt = VARENUM.VT_I4, data = new() { intVal = (int)(nint)(Owner?.InternalHandle ?? HWND.Null) } },
```
Tanya-Solyanik, line 501: Consider adding a helper method that builds up an appropriate VARIANT for UIA_NativeWindowHandlePropertyId

src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlAccessibleObject.cs
```diff
@@ -483,41 +483,42 @@ internal override bool IsPatternSupported(UIA_PATTERN_ID patternId)
         internal override bool IsIAccessibleExSupported()
             => Owner is IAutomationLiveRegion ? true : base.IsIAccessibleExSupported();
 
-        internal override object? GetPropertyValue(UIA_PROPERTY_ID propertyID) =>
+        /// <inheritdoc/>
```
Tanya-Solyanik, line 486: Is this tag really required? The doc comment is supposed to be inherited by default.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjects/ButtonBase.ButtonBaseAccessibleObjectTests.cs
```diff
@@ -196,10 +197,10 @@ public void ButtonBaseBoxAccessibleObject_ControlType_IsPane_IfAccessibleRoleIsD
         }
 
         AccessibleObject accessibleObject = buttonBase.AccessibilityObject;
-        object actual = accessibleObject.GetPropertyValue(UIA_PROPERTY_ID.UIA_ControlTypePropertyId);
+        VARIANT actual = accessibleObject.GetPropertyValue(UIA_PROPERTY_ID.UIA_ControlTypePropertyId);
```
Tanya-Solyanik, line 200: Theoretically, if this property is implemented correctly, these VARIANTs don't leak because they are integral types. So up to you if you want to clear them or not.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjects/ButtonBase.ButtonBaseAccessibleObjectTests.cs
```diff
@@ -237,9 +238,9 @@ public void ButtonBaseAccessibleObject_GetPropertyValue_AutomationId_ReturnsExpe
     {
         using SubButtonBase ownerControl = new() { Name = "test name" };
         string expected = ownerControl.Name;
-        object actual = ownerControl.AccessibilityObject.GetPropertyValue(UIA_PROPERTY_ID.UIA_AutomationIdPropertyId);
+        VARIANT actual = ownerControl.AccessibilityObject.GetPropertyValue(UIA_PROPERTY_ID.UIA_AutomationIdPropertyId);
```
Tanya-Solyanik, line 241: Please add using, it's a BSTR

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjects/Control.ControlAccessibleObjectTests.cs
```diff
@@ -1060,7 +1067,7 @@ public void ControlAccessibleObject_RaiseAutomationPropertyChangedEvent_IsHandle
             Assert.NotEqual(IntPtr.Zero, ownerControl.Handle);
         }
 
-        Assert.Equal(isHandleCreated, accessibleObject.RaiseAutomationPropertyChangedEvent(UIA_PROPERTY_ID.UIA_NamePropertyId, ownerControl.Name, ownerControl.Name));
+        Assert.Equal(isHandleCreated, accessibleObject.RaiseAutomationPropertyChangedEvent(UIA_PROPERTY_ID.UIA_NamePropertyId, (VARIANT)ownerControl.Name, (VARIANT)ownerControl.Name));
```
Tanya-Solyanik, line 1070: This variant(s) is not cleared

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjects/ListViewItem.ListViewSubItem.ListViewSubItemAccessibleObjectTests.cs
```diff
@@ -927,9 +927,9 @@ public void ListViewSubItemAccessibleObject_GetPropertyValue_BoundingRectangle_R
         listView.Items.Add(listViewItem);
         ListViewItem.ListViewSubItem listViewSubItem = new(listViewItem, "Test subItem");
         ListViewSubItemAccessibleObject listViewSubItemAccessibleObject = new(listViewSubItem, listViewItem);
-        object actual = listViewSubItemAccessibleObject.GetPropertyValue(UIA_PROPERTY_ID.UIA_BoundingRectanglePropertyId);
+        using VARIANT actual = listViewSubItemAccessibleObject.GetPropertyValue(UIA_PROPERTY_ID.UIA_BoundingRectanglePropertyId);
         using SafeArrayScope<double> rectArray = UiaTextProvider.BoundingRectangleAsArray(listViewSubItem.AccessibilityObject.BoundingRectangle);
-        Assert.Equal(((VARIANT)rectArray).ToObject(), actual);
+        Assert.Equal(((VARIANT)rectArray).ToObject(), actual.ToObject());
```
Tanya-Solyanik, line 932: (VARIANT)rectArray is not Cleared. Is it possible to simplify this test and compare rectangles instead?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjects/ListViewLabelEditAccessibleObjectTests.cs
```diff
@@ -14,30 +14,31 @@ namespace System.Windows.Forms.Tests.AccessibleObjects;
...
         Assert.Equal(listView.Enabled, (bool)accessibilityObject.GetPropertyValue(UIA_PROPERTY_ID.UIA_IsEnabledPropertyId));
-        Assert.Equal("1", accessibilityObject.GetPropertyValue(UIA_PROPERTY_ID.UIA_AutomationIdPropertyId));
-        Assert.Empty((string)accessibilityObject.GetPropertyValue(UIA_PROPERTY_ID.UIA_HelpTextPropertyId));
+        Assert.Equal("1", ((BSTR)accessibilityObject.GetPropertyValue(UIA_PROPERTY_ID.UIA_AutomationIdPropertyId)).ToStringAndFree());
+        Assert.Empty(((BSTR)accessibilityObject.GetPropertyValue(UIA_PROPERTY_ID.UIA_HelpTextPropertyId)).ToStringAndFree());
         Assert.True((bool)accessibilityObject.GetPropertyValue(UIA_PROPERTY_ID.UIA_IsContentElementPropertyId));
         Assert.False((bool)accessibilityObject.GetPropertyValue(UIA_PROPERTY_ID.UIA_IsPasswordPropertyId));
-        Assert.Equal(labelEdit.Handle, accessibilityObject.GetPropertyValue(UIA_PROPERTY_ID.UIA_NativeWindowHandlePropertyId));
+        Assert.Equal((int)labelEdit.Handle, accessibilityObject.GetPropertyValue(UIA_PROPERTY_ID.UIA_NativeWindowHandlePropertyId).ToObject());
```
Tanya-Solyanik, line 41: Why is this ToObject needed?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjects/TabControl.TabControlAccessibleObjectTests.cs
```diff
@@ -672,7 +679,7 @@ public void TabControlAccessibleObject_GetPropertyValue_NativeWindowHandleProper
 
         TabControlAccessibleObject accessibleObject = Assert.IsType<TabControlAccessibleObject>(tabControl.AccessibilityObject);
 
-        Assert.Equal((nint)tabControl.InternalHandle, (nint)accessibleObject.GetPropertyValue(UIA_PROPERTY_ID.UIA_NativeWindowHandlePropertyId));
+        Assert.Equal((int)(nint)tabControl.InternalHandle, accessibleObject.GetPropertyValue(UIA_PROPERTY_ID.UIA_NativeWindowHandlePropertyId).ToObject());
```
Tanya-Solyanik, line 682: Why doesn't a cast to int work instead of ToObject?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjects/TextBoxAccessibleObjectTests.cs
```diff
@@ -139,7 +139,7 @@ public void TextBoxAccessibleObject_IsPassword_IsExpected_WithPasswordChar(char
         using TextBox textBox = new TextBox();
         textBox.PasswordChar = passwordChar;
 
-        object actual = textBox.AccessibilityObject.GetPropertyValue(UIA_PROPERTY_ID.UIA_IsPasswordPropertyId);
+        var actual = (bool)textBox.AccessibilityObject.GetPropertyValue(UIA_PROPERTY_ID.UIA_IsPasswordPropertyId);
```
Tanya-Solyanik, line 142: Usually we don't use `var` for primitive types.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjects/ToolStripItemAccessibleObjectTests.cs
```diff
@@ -94,14 +95,17 @@ public void ToolStripItemAccessibleObject_GetPropertyValue_Custom_Name_ReturnsEx
 
         // By default Name has string.Empty value, because if AccessibleName is not defined
         // then control uses the value of "Text" property from owner Item (by default an empty string)
-        Assert.Equal(string.Empty, toolStripItemAccessibleObject.GetPropertyValue(UIA_PROPERTY_ID.UIA_NamePropertyId));
+        VARIANT result = toolStripItemAccessibleObject.GetPropertyValue(UIA_PROPERTY_ID.UIA_NamePropertyId);
```
Tanya-Solyanik, line 98: BBTW, `using` would clear it if line 99 throws.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjects/ToolStripItemAccessibleObjectTests.cs
```diff
@@ -94,14 +95,17 @@ public void ToolStripItemAccessibleObject_GetPropertyValue_Custom_Name_ReturnsEx
...
+        Assert.Equal(string.Empty, ((BSTR)result).ToStringAndFree());
 
         item.Name = "Name1";
         item.AccessibleName = "Test Name";
 
         var accessibleName = toolStripItemAccessibleObject.GetPropertyValue(UIA_PROPERTY_ID.UIA_NamePropertyId);
 
-        Assert.Equal("Test Name", accessibleName);
+        Assert.Equal(VARENUM.VT_BSTR, accessibleName.vt);
+        Assert.Equal("Test Name", ((BSTR)accessibleName).ToString());
```
Tanya-Solyanik, line 108: This needs freeing.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjects/UpDownBase.UpDownButtons.UpDownButtonsAccessibleObjectTests.cs
```diff
@@ -112,9 +112,9 @@ public void UpDownButtonsAccessibleObject_GetPropertyValue_BoundingRectangle_Ret
     {
         using SubUpDownBase upDownBase = new();
         UpDownButtons upDownButtons = upDownBase.UpDownButtonsInternal;
-        object actual = upDownButtons.AccessibilityObject.GetPropertyValue(UIA_PROPERTY_ID.UIA_BoundingRectanglePropertyId);
+        using VARIANT actual = upDownButtons.AccessibilityObject.GetPropertyValue(UIA_PROPERTY_ID.UIA_BoundingRectanglePropertyId);
         using SafeArrayScope<double> rectArray = UiaTextProvider.BoundingRectangleAsArray(upDownButtons.AccessibilityObject.BoundingRectangle);
-        Assert.Equal(((VARIANT)rectArray).ToObject(), actual);
+        Assert.Equal(((VARIANT)rectArray).ToObject(), actual.ToObject());
```
Tanya-Solyanik, line 117: Either Clear this `(VARIANT)rectArray` or compare rectangles.

Tanya-Solyanik (commented):
Looks great!

src/System.Windows.Forms.Primitives/src/Windows/Win32/UI/Accessibility/UIA_PATTERN_ID.cs
```diff
@@ -3,6 +3,8 @@
 
 namespace Windows.Win32.UI.Accessibility;
 
+// Remove this definition and use Cswin32 definition instead once
+// https://github.com/microsoft/win32metadata/issues/1723 is resolved
```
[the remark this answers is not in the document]
lonitra: Cswin32 still needs to push out a version that includes the fix so we can update our version. I will update comment to call this out.

src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.TextBoxBaseUiaTextProvider.cs
```diff
@@ -164,8 +164,8 @@ public override HRESULT GetCaretRange(BOOL* isActive, ITextRangeProvider** pRetV
                 return HRESULT.S_OK;
             }
 
-            var hasKeyboardFocus = Owner.AccessibilityObject.GetPropertyValue(UIA_PROPERTY_ID.UIA_HasKeyboardFocusPropertyId);
-            *isActive = hasKeyboardFocus is true;
+            VARIANT hasKeyboardFocus = Owner.AccessibilityObject.GetPropertyValue(UIA_PROPERTY_ID.UIA_HasKeyboardFocusPropertyId);
```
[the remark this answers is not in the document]
lonitra: No need, this `VARIANT` is either empty or holding a `bool` which doesn't require disposal.

src/System.Windows.Forms/tests/InteropTests/AccessibleObjectTests.cs
```diff
@@ -98,9 +103,17 @@ public static IEnumerable<object[]> GetPropertyValue_TestData()
     [MemberData(nameof(GetPropertyValue_TestData))]
     public void AccessibleObject_IRawElementProviderSimpleGetPropertyValue_Invoke_ReturnsExpected(object propertyId, object expected)
     {
-        var o = new AccessibleObject();
-        AssertSuccess(Test_IRawElementProviderSimpleGetPropertyValue(o, (UIA_PROPERTY_ID)propertyId, out VARIANT variant));
-        Assert.Equal(expected, variant.ToObject());
+        using var elementProvider = ComHelpers.GetComScope<IRawElementProviderSimple>(new AccessibleObject());
+        VARIANT variant = default;
```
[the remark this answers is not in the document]
lonitra: We also don't need to do a `using` here  because this variant will either be empty or have `bool` written in it which doesn't require disposal.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.DataGridViewEditingPanelAccessibleObject.cs
```diff
@@ -65,24 +66,24 @@ internal override void SetFocus()
...
                     : base.GetPropertyValue(propertyId),
                 UIA_PROPERTY_ID.UIA_HasKeyboardFocusPropertyId
-                    => _ownerDataGridView.TryGetTarget(out var owner) && owner.CurrentCell is not null,
-                UIA_PROPERTY_ID.UIA_IsContentElementPropertyId => true,
-                UIA_PROPERTY_ID.UIA_IsControlElementPropertyId => true,
-                UIA_PROPERTY_ID.UIA_IsEnabledPropertyId => _ownerDataGridView.TryGetTarget(out var owner) && owner.Enabled,
-                UIA_PROPERTY_ID.UIA_IsKeyboardFocusablePropertyId => true,
-                UIA_PROPERTY_ID.UIA_ProviderDescriptionPropertyId => SR.DataGridViewEditingPanelUiaProviderDescription,
+                    => (VARIANT)(_ownerDataGridView.TryGetTarget(out var owner) && owner.CurrentCell is not null),
+                UIA_PROPERTY_ID.UIA_IsContentElementPropertyId => (VARIANT)true,
```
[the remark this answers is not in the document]
lonitra: Good call!

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjects/ButtonBase.ButtonBaseAccessibleObjectTests.cs
lonitra: We are clearing below with `ToStringAndFree()` but I will change to do using on variant

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjects/Control.ControlAccessibleObjectTests.cs
lonitra: Thanks for the catch!

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjects/TabControl.TabControlAccessibleObjectTests.cs
lonitra: We can do cast to int, I will update to use that instead

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjects/UpDownBase.UpDownButtons.UpDownButtonsAccessibleObjectTests.cs
lonitra: This is being cleared already with the using on the `SafeArrayScope<double>` above

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjects/ListViewItem.ListViewSubItem.ListViewSubItemAccessibleObjectTests.cs
lonitra: `rectArray` is being cleared with the using statement above on `SafeArrayScope<double>`. Could you clarify with comparing the rectangles instead? Both `GetPropertyValue(UIA_PROPERTY_ID.UIA_BoundingRectanglePropertyId)` and `UiaTextProvider.BoundingRectangleAsArray(listViewSubItem.AccessibilityObject.BoundingRectangle)` are returning the rectangle but in the form of an array. Do you want me to actually create the rectangle to compare here?

src/System.Windows.Forms/src/System/Windows/Forms/ListBox.AccessibleObject.cs
```diff
@@ -120,27 +120,27 @@ private protected virtual ListBoxItemAccessibleObject CreateItemAccessibleObject
...
         /// <param name="propertyID">The accessible property ID.</param>
         /// <returns>The accessible property value.</returns>
-        internal override object? GetPropertyValue(UIA_PROPERTY_ID propertyID)
+        internal override VARIANT GetPropertyValue(UIA_PROPERTY_ID propertyID)
         {
             switch (propertyID)
             {
                 case UIA_PROPERTY_ID.UIA_BoundingRectanglePropertyId:
-                    return ((VARIANT)UiaTextProvider.BoundingRectangleAsArray(BoundingRectangle)).ToObject();
+                    return (VARIANT)UiaTextProvider.BoundingRectangleAsArray(BoundingRectangle);
```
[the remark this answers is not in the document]
lonitra: I'm not sure we can do this change. `BoundingRectangleAsArray` is used [here](https://github.com/dotnet/winforms/blob/3766951fd7f8630261287ad1cfdc682480699b50/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextRange.cs#L362), which is not expecting a `VARIANT`

src/System.Windows.Forms/src/System/Windows/Forms/Button.ButtonAccessibleObject.cs
```diff
@@ -13,22 +14,22 @@ public ButtonAccessibleObject(Button owner) : base(owner)
...
+        internal override VARIANT GetPropertyValue(UIA_PROPERTY_ID propertyID)
             => propertyID switch
             {
                 UIA_PROPERTY_ID.UIA_ControlTypePropertyId
                     // If we don't set a default role for Button and ButtonBase accessible objects
                     // it will be retrieved from Windows.
                     // And we don't have a 100% guarantee it will be correct, hence set it ourselves.
                     => this.TryGetOwnerAs(out Control? owner) && owner.AccessibleRole == AccessibleRole.Default
-                       ? UIA_CONTROLTYPE_ID.UIA_ButtonControlTypeId
+                       ? (VARIANT)(int)UIA_CONTROLTYPE_ID.UIA_ButtonControlTypeId
```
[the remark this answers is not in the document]
lonitra: Hm, that would be nice, but I'm not sure I want to add an operator in `VARIANT.cs` as it is not for general purpose and once we move to using Cswin32 definition of UIA_ enum, I don't think there is a way for us to add the operator.

src/System.Windows.Forms/src/System/Windows/Forms/ListBox.AccessibleObject.cs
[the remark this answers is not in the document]
Tanya-Solyanik: Good point, we can keep the old `BoundingRectangleAsArray` in that spot and replace it everywhere else. What bothers we with the code as is, is that we have a helper method that takes us "half way" and requiring a cast, while we could have a helper method (or an extension method) to do the complete conversion.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjects/ListViewItem.ListViewSubItem.ListViewSubItemAccessibleObjectTests.cs
Tanya-Solyanik: Yes, compare listViewSubItem.AccessibilityObject.BoundingRectangle to `actual` converted to a rectangle, because the code will be simpler(less allocations) and closer to the business logic (as is we are invoking the same method (`BoundingRectangleToArray` twice, once directly and once in the `GetPropertyValue` method. Theoretically if we had a bug in that method, we wouldn't have detected it.), but this is not a big deal, up to you.

src/System.Windows.Forms/src/System/Windows/Forms/Button.ButtonAccessibleObject.cs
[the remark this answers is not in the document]
Tanya-Solyanik: ok

src/System.Windows.Forms.Primitives/src/Interop/OleAut32/Interop.VARIANT.cs
```diff
@@ -15,6 +15,10 @@ internal unsafe partial struct VARIANT : IDisposable
 {
     public static VARIANT Empty { get; }
 
+    public static VARIANT True => new() { vt = VT_BOOL, data = new() { boolVal = VARIANT_BOOL.VARIANT_TRUE } };
```
Tanya-Solyanik, line 18: Why do we allocate the `boolVal`?

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Automation/UiaTextProvider.cs
```diff
@@ -112,6 +112,10 @@ internal static SafeArrayScope<double> BoundingRectangleAsArray(Rectangle bounds
         return result;
     }
 
+    /// <inheritdoc cref="BoundingRectangleAsArray(Rectangle)"/>
```
Tanya-Solyanik, line 115: Yes, this is what this tag is for.

src/System.Windows.Forms.Primitives/src/Interop/OleAut32/Interop.VARIANT.cs
lonitra: When vt is `VT_BOOL` data is read from `boolVal`

Tanya-Solyanik: But this is not a pointer data, it's a -1 or a `0`, why are we not assigning it directly?
 data.boolVal = VARIANT_BOOL.VARIANT_TRUE;

lonitra: Sorry I'm not sure I follow. Where are we giving pointer data? `boolVal` is a `VARIANT_BOOL`

Tanya-Solyanik: I was thinking something like this:

```
    public static VARIANT True { get; } = InitTrue();

    private static VARIANT InitTrue()
    {
        VARIANT a = new() { vt = VT_BOOL };
        a.data.boolVal = VARIANT_BOOL.VARIANT_TRUE;
        return a;
    }
```

src/System.Windows.Forms.Primitives/src/Windows/Win32/UI/Accessibility/UIAHelper.cs
```diff
@@ -9,9 +9,21 @@ internal static class UIAHelper
...
     /// </summary>
     /// <param name="handle">The handle to the window</param>
     /// <returns>The <see cref="VARIANT"/> version of the window handle.</returns>
+    /// <remarks>
+    ///  <para>
+    ///   Although <see cref="global::System.Runtime.InteropServices.Marshal.GetNativeVariantForObject"/>
+    ///   with the nint/IntPtr passed in returns a VARIANT with <see cref="VARENUM.VT_INT"/>,
+    ///   <see cref="IRawElementProviderSimple.GetPropertyValue(IRawElementProviderSimple*, UIA_PROPERTY_ID, VARIANT*)"/>
+    ///   with <see cref="UIA_PROPERTY_ID.UIA_NativeWindowHandlePropertyId"/> is expecting a VARIANT with <see cref="VARENUM.VT_I4"/>.
+    ///   See <see href="https://learn.microsoft.com/en-us/windows/win32/winauto/uiauto-automation-element-propids"/>
```
Tanya-Solyanik, line 23: Please remove locale from the URL

Tanya-Solyanik (approved):

Tanya-Solyanik (approved):

PR #9704, opened 2023-08-10
Enable nullability in DataGridViewDataConnection

gpetrou:
## Proposed changes

- Enable nullability in DataGridViewDataConnection.

PR #10186, opened 2023-10-25
Enable nullability in some DataGridViewComboBoxCell members

gpetrou:
## Proposed changes

- Enable nullability in some DataGridViewComboBoxCell members.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10186)


src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewComboBoxCell.cs
```diff
@@ -1409,10 +1451,12 @@ private void InitializeDisplayMemberPropertyDescriptor(string displayMember)
             {
                 BindingMemberInfo displayBindingMember = new BindingMemberInfo(displayMember);
                 // make the DataManager point to the sublist inside this.DataSource
-                DataManager = DataGridView.BindingContext[DataSource, displayBindingMember.BindingPath] as CurrencyManager;
+                Debug.Assert(DataGridView?.BindingContext is not null);
+                Debug.Assert(DataSource is not null);
```
Epica3055, line 1455: Why `Debug.Assert` here, Why not `!` like other places

line 1483 as well

Tanya-Solyanik: Are we actually guaranteed that both BindingContext and DataSource are not null? Please add a comment with an explanation.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewComboBoxCell.cs
```diff
@@ -1437,10 +1481,12 @@ private void InitializeValueMemberPropertyDescriptor(string valueMember)
             {
                 BindingMemberInfo valueBindingMember = new BindingMemberInfo(valueMember);
                 // make the DataManager point to the sublist inside this.DataSource
-                DataManager = DataGridView.BindingContext[DataSource, valueBindingMember.BindingPath] as CurrencyManager;
+                Debug.Assert(DataGridView?.BindingContext is not null);
+                Debug.Assert(DataSource is not null);
```
Tanya-Solyanik, line 1485: Please add a comment with an explanation, why these are not null. Debug.Assert is present only in the debug bits, it doesn't ensure that BindingCntext and DataSource are not null, so essentially it is equivalent to a `!`

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewComboBoxCell.cs
gpetrou: Using `!` is my preference as well, but recently I was getting suggestions to use `Debug.Assert`. Check https://github.com/dotnet/winforms/pull/9704#discussion_r1350728551 for example.
As for the values being null, looking at `GetDataManager` didn't seem like  a possibility.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewComboBoxCell.cs
```diff
@@ -1587,18 +1634,18 @@ public override bool KeyEntersEditMode(KeyEventArgs e)
     ///  ComboBox datasource. We then use DisplayMember to get the
     ///  text to display.
     /// </summary>
-    private bool LookupDisplayValue(int rowIndex, object value, out object displayValue)
+    private bool LookupDisplayValue(int rowIndex, object value, out object? displayValue)
```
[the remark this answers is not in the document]
gpetrou: Hmm, this can still be null if we return `true` according to `GetItemDisplayValue`.

lonitra (approved):
LGTM

Issue #10193, opened 2023-10-26
An exception pops up when testing HtmlWindow's Opener and WindowFrameElement properties


Olina-Zhang:
### .NET version

.NET 9.0.100-alpha.1.23523.12

### Did it work in .NET Framework?

Yes

### Did it work in any of the earlier releases of .NET Core or .NET 5+?

Yes

### Issue description

Automation case about XHtmlWindow met an error "Object reference not set to an instance of an object" when testing HtmlWindow's Opener and WindowFrameElement properties. 

We created a test sample with that exception in .Net 9, got following result:
**HtmlWindow's Opener** 
![image](https://github.com/dotnet/winforms/assets/26474449/d6844085-3826-4d13-9b6f-8440d2c537dc)
_System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=System.Windows.Forms.Primitives
  StackTrace:
   at Windows.Win32.System.Com.IDispatch.QueryInterface(Guid* riid, Void** ppvObject)
   at System.Windows.Forms.HtmlWindow.get_Opener()
   at WinFormsApp20.Form1.button1_Click(Object sender, EventArgs e) in C:\Users\xxx\source\repos\WinFormsApp20\WinFormsApp20\Form1.cs:line 12_

**HtmlWindow's WindowFrameElement** 
![image](https://github.com/dotnet/winforms/assets/26474449/235d2233-3862-471e-b81f-5e437ee0983b)
_System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=System.Windows.Forms.Primitives
  StackTrace:
   at Windows.Win32.Web.MsHtml.IHTMLFrameBase.QueryInterface(Guid* riid, Void** ppvObject)
   at System.Windows.Forms.HtmlWindow.get_WindowFrameElement()
   at WinFormsApp20.Form1.button2_Click(Object sender, EventArgs e) in C:\Users\xxx\source\repos\WinFormsApp20\WinFormsApp20\Form1.cs:line 17_

### Steps to reproduce

1. Download and extract project: [HtmlWindowSample.zip](https://github.com/dotnet/winforms/files/13172458/HtmlWindowSample.zip)
2. Build and run
3. Click buttons: HtmlWindow.Opener and HtmlWindow.WindowFrameElement

MelonWang1:
[...]

PR #10198, opened 2023-10-26
Harden Against Null in HTMLX files

lonitra:
Fixes https://github.com/dotnet/winforms/issues/10193

- Add null check to `HtmlWindow.Opener` and `HtmlWindow.WindowFrameElement` and add regression tests
- Also went ahead and looked through HTMLX files and added additional null checks where it is valid to receive an empty pointer and using it would result in null reference.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10198)

### DIFF
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/HtmlWindowTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/HtmlWindowTests.cs
@@ -1,0 +1,60 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Text;
+
+namespace System.Windows.Forms.Tests;
+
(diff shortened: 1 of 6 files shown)


src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/HtmlWindowTests.cs
```diff
@@ -0,0 +1,60 @@
...
+        using var parent = new Control();
+        using var control = new WebBrowser
+        {
+            Parent = parent
+        };
+
+        const string Html = "<html><body>test</body></html>";
+        HtmlDocument document = await GetDocument(control, Html);
+        HtmlWindow window = document.Window;
+        Assert.NotSame(window, document.Window);
```
Tanya-Solyanik, line 23: Could you please explain what is happeninghere?

Tanya-Solyanik (approved):

PR #10199, opened 2023-10-27
Removed a noisy Debug.Assert

Tanya-Solyanik:
Steps:
Run a debug build of WinformsControlsTest app,
open the DataGridView page and resize columns or rows

Result: an Assert.

We assumed that lifetime of editing control is not longer than that of its container panel, and tested only the editing control for null. However their lifetimes are not tied explicitly, it's safer to have an explicit null-check. It's unlikely that anyone was processing this NRE other than by ignoring it, as it happens on mouse move.

Another change is a removal of null check for EditingPanel get property as this property allocated the panel on the first access.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10199)


lonitra (approved):
:shipit:

Olina-Zhang:
Verified in main branch of Winforms repo for a debug build of WinformsControlsTest app, no debug.Assert pops up when resizing columns or rows in DataGridView form.

PR #10200, opened 2023-10-27
Start migrating UiaCore to CsWin32

elachlan:
Related: #7445, #9795

This is just a start, focusing on the enums. I mostly used find/replace and corrected any compile errors.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10200)


elachlan:
It looks like @lonitra has already started this at #10177

elachlan:
I might be wrong actually.

lonitra:
Changes LGTM, thanks for doing this! 🥳 The PR mentioned above only converts enums that are related to `IRawElementProviderSimple`, so definitely not the amount of enum conversion this PR covers. Did you have the chance to test changes by going through `WinformsControlsTest` with Narrator?

lonitra:
We recently ran into an issue while converting UIA enums that metadata had defined their enums incorrectly. In that case metadata had defined enum as `uint` instead of `int` which caused Narrator to announce things incorrectly as it was expecting `int`. I think we are ok here, but it is good to make sure. I will pull down changes and test with Narrator some time today.

elachlan:
I generally check ilspy to confirm all values are in the enums.

I'll make a note to check the data type as well.

lonitra (approved):
Just tested. Narrator seems to be announcing things correctly 🚀

PR #10204, opened 2023-10-27
Improve nullable annotations in ListBindingHelper

halgab:
I noticed that `ListBindingHelper.GetListItemType` annotations could be improved. I updated other annotations accordingly.

Fixes #

## Proposed changes

- Change NRT annotations of `ListBindingHelper.GetListItemType`

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/CurrencyManager.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/CurrencyManager.cs
@@ -134,7 +134,7 @@ public class CurrencyManager : BindingManagerBase
     /// </summary>
     public override object? Current => this[Position];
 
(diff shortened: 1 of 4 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/CurrencyManager.cs
```diff
@@ -134,7 +134,7 @@ internal bool AllowRemove
     /// </summary>
     public override object? Current => this[Position];
 
-    internal override Type? BindType => ListBindingHelper.GetListItemType(List);
+    internal override Type BindType => ListBindingHelper.GetListItemType(List);
```
lonitra, line 137: Should this be kept as nullable? `List` may be null here

lonitra (approved):

PR #10205, opened 2023-10-28

gpetrou:
## Proposed changes

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10205)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.Methods.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.Methods.cs
@@ -3426,7 +3426,7 @@ public partial class DataGridView
     }
 
     // determines if a data bound cell can be validated or not
-    private bool CanValidateDataBoundDataGridViewCell(DataGridViewCell dataGridViewCurrentCell)
+    private bool CanValidateDataBoundDataGridViewCell(DataGridViewCell? dataGridViewCurrentCell)
     {
         if (dataGridViewCurrentCell is null)
         {
@@ -3859,7 +3859,7 @@ public partial class DataGridView
     {
         if (IsCurrentCellInEditMode)
         {
-            DataGridViewCell dataGridViewCurrentCell = CurrentCellInternal;
+            DataGridViewCell? dataGridViewCurrentCell = CurrentCellInternal;
             DataGridViewDataErrorEventArgs? dgvdee = CommitEdit(
                 ref dataGridViewCurrentCell,
                 context,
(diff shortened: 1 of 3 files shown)


Epica3055 (commented):
looks good

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.Methods.cs
```diff
@@ -11540,39 +11540,49 @@ protected virtual void OnAllowUserToAddRowsChanged(EventArgs e)
...
+            !IsDisposed)
         {
             eh(this, e);
         }
     }
 
     protected virtual void OnAllowUserToDeleteRowsChanged(EventArgs e)
     {
-        if (Events[s_allowUserToDeleteRowsChangedEvent] is EventHandler eh && !_dataGridViewOper[OperationInDispose] && !IsDisposed)
+        if (Events[s_allowUserToDeleteRowsChangedEvent] is EventHandler eh &&
```
Epica3055, line 11553: I noticed that you wrapped a lot of lines, like this.
I was wondering why did you do this for readability or just because the line is too long, if the line is wrapped , so there is no need to use the horizontal scroll bar.

This is what I observed from my environment, 1920*1080 monitor.

I think modern monitor is big enough to show at least 120 chars.


image

@Tanya-Solyanik what do you think?

gpetrou: I did that for readability.

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView.Methods.cs
```diff
@@ -6368,7 +6368,7 @@ public bool EndEdit(DataGridViewDataErrorContexts context)
         {
             int curRowIndex = _ptCurrentCell.Y;
             int curColIndex = _ptCurrentCell.X;
-            DataGridViewCell dataGridViewCurrentCell = CurrentCellInternal;
+            DataGridViewCell? dataGridViewCurrentCell = CurrentCellInternal;
```
lonitra, line 6371: Is `CurrentCellInternal` annotated as nullable? If it is, I'm not so sure that this should be annotated as nullable since we are ultimately returning a `DataGridViewCell` attained from a `DataGridViewCellCollection`. I believe we guard against adding a null to the collection in `DataGridViewCellCollection`

gpetrou: No, it is not. But we pass `dataGridViewCurrentCell` by ref in some methods which can set it to null. Check inside `CommitEdit` and the methods that it calls.

lonitra: Ah I see, thanks! I was concerned about the `!` on `dataGridViewCurrentCell` below, but digging through the code, it is hard to tell whether or not there is a chance is could be null and what exactly that means.. There is definitely room for clean up here in general that would help make it easier to understand what is going on, so I think it is ok to do `!` for now and revisit later when clean up happens.

lonitra (approved):

Issue #10218, opened 2023-11-02
Inconsistent DomWindow values


MelonWang1:
### .NET version

.NET 9.0.100-alpha.1.23551.16

### Did it work in .NET Framework?

Yes

### Did it work in any of the earlier releases of .NET Core or .NET 5+?

Yes, it cannot repro in .NET 3.1/5.0/6.0/7.0/8.0.

### Issue description

Automation case about XHtmlWindow met an error "FAIL: equality between windows failed" when testing HtmlWindow's op_Equality, get_Frames methods.

We created a test sample in .NET 9, got following result:
**The DomWindow has inconsistent values. In .NET 9.0:**
![CORE9 0](https://github.com/dotnet/winforms/assets/94418985/3781aac1-2060-4710-8659-143dc35a3f2f)

**In .NET 3.1~8.0 and Framework app:**
![FRAMEWORK](https://github.com/dotnet/winforms/assets/94418985/06943c25-1861-451e-9c31-2b7beff344c4)

### Steps to reproduce

1. Download and extract project: [DomWindow.zip](https://github.com/dotnet/winforms/files/13233689/DomWindow.zip)
2. Build and run
3. Click button1 and observe DomWindow value

elachlan:
Is the issue that the `DomWindow` in main branch doesn't have the same properties as the olded .net versions?

> Gets the unmanaged interface wrapped by this class.

https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.htmlwindow.domwindow?view=windowsdesktop-7.0

It's typed as an object:
```csharp
public object DomWindow { get; }
```

lonitra:
Yes, we had changed the inner type to an `AgileComPointer` and this is now no longer giving back the COM interface, which is a mistake.

MelonWang1:
Verified this on .NET 9.0.100-alpha.1.23558.10, issue was fixed. But the automation case about XHtmlWindow still has error, filed a new issue [10261](https://github.com/dotnet/winforms/issues/10261).
![domwindow](https://github.com/dotnet/winforms/assets/94418985/231fb12d-9033-40f8-a904-876d87f01dd0)

PR #10220, opened 2023-11-02
Return Interface for HtmlWindow.DomWindow

lonitra:
Fixes https://github.com/dotnet/winforms/issues/10218

We previously changed internal `NativeHtmlWindow` to type `AgileComPointer` after removing old `IHTMLWindow2` Interface definition and did not update `DomWindow` to return the interface like it always had been. This was missed due to `DomWindow` being of type `object`. I've updated code to return the interface as well as added regression test for this.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10220)

### DIFF
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/HtmlWindowTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/HtmlWindowTests.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Text;
+using Windows.Win32.Web.MsHtml;
 
 namespace System.Windows.Forms.Tests;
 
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/HtmlWindowTests.cs
```diff
@@ -40,6 +41,27 @@ public async Task HtmlWindow_WindowFrameElement_NoneReturnsNull()
...
+        const string Html = "<html><body>test</body></html>";
+        HtmlDocument document = await GetDocument(control, Html);
+        HtmlWindow window = document.Window;
+        object domWindow = window.DomWindow;
+
+        Assert.Same(domWindow, window.DomWindow);
+        Assert.True(domWindow.GetType().IsCOMObject);
+        Assert.True(domWindow is IHTMLWindow2.Interface);
+        Assert.True(domWindow is IHTMLWindow3.Interface);
+        Assert.True(domWindow is IHTMLWindow4.Interface);
```
gpetrou, line 62: Since we can use FluentAssertions now, perhaps you could use `BeOfType` and `BeAssignableTo` to get better error messages?

lonitra: Good call!

Tanya-Solyanik (approved):

PR #10223, opened 2023-11-03
Convert `System.Drawing.Printing.PreviewPrintController` `_list` from `ArrayList` to `List<T>`

elachlan:
Related: #8140

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10223)

### DIFF
--- a/src/System.Drawing.Common/src/System/Drawing/Printing/PreviewPrintController.cs
+++ b/src/System.Drawing.Common/src/System/Drawing/Printing/PreviewPrintController.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Collections;
+using System.Collections.Generic;
 using System.Drawing.Drawing2D;
 using System.Drawing.Imaging;
 using System.Drawing.Internal;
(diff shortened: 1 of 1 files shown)


lonitra (approved):

src/System.Drawing.Common/src/System/Drawing/Printing/PreviewPrintController.cs
```diff
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Collections;
```
halgab, line 4: I know I'm late to the party, but this line should have been removed as part of this PR. Apparently this does not raise an error in this project.

elachlan: Good Catch. I think `system.drawing` has different analyzer settings.

elachlan: https://github.com/dotnet/winforms/blob/3687127d555f75914afd958d19bfededa1e6254f/src/System.Drawing.Common/Directory.Build.props#L8-L10

PR #10230, opened 2023-11-06
When the cursor is not over any control and not in a row/column, chan…

LeafShi1:
Porting fix from the Out-Of-Process designer

## Proposed changes
When the cursor is not over any control and not in a row/column, change insertIndex to the last one.

## Root cause of this bug
In the original code, when the cursor is not on any control and not in a row/column, only the IBar is erased without processing the InsertionIndex of the control, so the top control cannot move to the bottom.

## Before
The top control cannot be moved to the bottom, it always backs to the original position.
BeforeChange

## After
The top control can be moved to the bottom
AfterChange

## Test methodology 

- Manually

## Test environment(s) 
- .Net 9.0.0-alpha.1.23519.1

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10230)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/FlowLayoutPanelDesigner .cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/FlowLayoutPanelDesigner .cs
@@ -898,6 +898,7 @@ internal partial class FlowLayoutPanelDesigner : FlowPanelDesigner
         {
             // Here, we're at the 'end' of the FlowLayoutPanel - not over
             // any controls and not in a row/column.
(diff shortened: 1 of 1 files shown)


ricardobossan:
@LeafShi1 LGTM!

Tanya-Solyanik (approved):

John-Qiao:
Verified this PR on release/9.0 branch of winforms repo, the top control can be moved to the bottom now.


https://github.com/dotnet/winforms/assets/45864985/eeca637b-78f5-4364-b155-a9e80267df95

PR #10231, opened 2023-11-06
Introduce a throw helper in DataGridView.Methods

halgab:
While working on #9726, I encountered this factorisation opportunity. This cleans up the code and optimises it by getting the throw expressions out of the way.

As a follow-up, if the team is willing to modify the error messages, we could use `ArgumentOutOfRangeException` throw helpers inside the new method, but this change is less valuable

## Proposed changes

- Factorise EventArgs index bound checks in `DataGridView` in a dedicated helper method

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10231)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCellEventArgs.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCellEventArgs.cs
@@ -3,7 +3,7 @@
 
 namespace System.Windows.Forms;
 
-public class DataGridViewCellEventArgs : EventArgs
+public class DataGridViewCellEventArgs : EventArgs, IDataGridViewCellEventArgs
 {
     internal DataGridViewCellEventArgs(DataGridViewCell dataGridViewCell)
         : this(dataGridViewCell.ColumnIndex, dataGridViewCell.RowIndex)
@@ -22,4 +22,10 @@ public class DataGridViewCellEventArgs : EventArgs
     public int ColumnIndex { get; }
 
     public int RowIndex { get; }
+}
+
+internal interface IDataGridViewCellEventArgs
+{
+    int ColumnIndex { get; }
+    int RowIndex { get; }
 }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCellEventArgs.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCellEventArgs.cs
@@ -23,9 +23,3 @@ public class DataGridViewCellEventArgs : EventArgs, IDataGridViewCellEventArgs
 
     public int RowIndex { get; }
 }
-
-internal interface IDataGridViewCellEventArgs
-{
-    int ColumnIndex { get; }
-    int RowIndex { get; }
-}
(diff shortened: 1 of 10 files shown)


lonitra (approved):
Awesome 🥳

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCellEventArgs.cs
```diff
@@ -23,3 +23,9 @@ public DataGridViewCellEventArgs(int columnIndex, int rowIndex)
 
     public int RowIndex { get; }
 }
+
+internal interface IDataGridViewCellEventArgs
```
gpetrou, line 27: Maybe move this to a separate file?

halgab: Good point. DataGridViewCellEventArgs.cs is already taken though. Any idea?

gpetrou: I guess `IDataGridViewCellEventArgs.cs`?

halgab: All `DataGridView` interfaces drop the prefix I in their name files. But we're lacking alternatives here, so I guess I'll end up doing that

halgab: done

gpetrou: > All `DataGridView` interfaces drop the prefix I in their name files.

This sounds like a mistake to me. I guess we could fix it in the future.

lonitra: This does seem like a mistake.. Which interfaces are you seeing that are dropping the prefix?

lonitra (approved):

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCellEventArgs.cs
halgab: Sorry I wasn't clear. The interfaces are named just fine. It's just the file names that are slightly inconsistent. For instance, `IDataGridViewEditingCell` sits in DataGridViewEditingCell.cs, not IDataGridViewEditingCell.cs. I don't know if it's expected. If you want me to change those occurences, I can take care of it.

lonitra: Ah gotcha, I was just misunderstanding -- that is good to hear :) yes this seems like a mistake and would be safe to rename. Appreciate it if you could follow up on renaming.

halgab: On that topic, would it make sense to put all the `DataGridView` files in a dedicated subfolder (after all open PRs are merged to avoid conflicts)?

halgab: #10270

PR #10243, opened 2023-11-07
Allows DataGridView to start row/column on index 1 on Narrator

### LINKED ISSUE #7154: [Accessibility] Change DataGridView's Row/Column counting with starting at 1, not 0
### .NET version

**.Net 7.0 SDK build:** 7.0.100-preview.5.22256.2

No

No, this is not a regression issue. And @merriemcgaw wants a runtime issue to ask to change the counting with starting at 1.

### Issue description

We need to change DataGridView's Row/Column counting with starting at 1, not 0

**Accessibility Insight:**
image

**Inspect:**
image

**Narrator:**
image

### Steps to reproduce

**Repro steps:**
1. Open attached application
[DataGridView_Sample.zip](https://github.com/dotnet/winforms/files/8649203/DataGridView_Sample.zip)
2. Build and run
3. Use Accessibility tools: AccessiblityInsight, Narrator, Inspect focus on the first DataGridView cell, to observe the rows/columns starting count

ricardobossan:
Fixes #7154 

## Proposed changes

- Creates a new switch that allows developers to configure their application to make the starting index of rows/cells in DataGridView to be read by Narrator from 1, instead of 0;
- Adds checks on rows and cells of DataGridView to see if the switch is true. If true, starting index of rows/cells will be 1, instead of 0.

## Customer Impact

- Developers will be able to configure their WinForms DataGridView rows/cells to be read by Narrator from index 1, instead of 0.

## Regression? 

- No

## Risk

- Minimal

### Before

image

### After

image
image

## Test methodology 

- Manual (Narrator)

## Test environment(s) 

- 9.0.100-alpha.1.23511.2

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
@@ -20,12 +20,14 @@ internal static partial class LocalAppContextSwitches
     internal const string ServicePointManagerCheckCrlSwitchName = "System.Windows.Forms.ServicePointManagerCheckCrl";
     internal const string TrackBarModernRenderingSwitchName = "System.Windows.Forms.TrackBarModernRendering";
     private const string DoNotCatchUnhandledExceptionsSwitchName = "System.Windows.Forms.DoNotCatchUnhandledExceptions";
+    internal const string DataGridViewStartRowColumnIndexSwitchName = "System.Windows.Forms.DataGridViewRowColumnStartIndex";
 
     private static int s_scaleTopLevelFormMinMaxSizeForDpi;
     private static int s_anchorLayoutV2;
     private static int s_servicePointManagerCheckCrl;
     private static int s_trackBarModernRendering;
     private static int s_doNotCatchUnhandledExceptions;
+    private static int s_dataGridViewStartRowColumnIndex;
 
     private static FrameworkName? s_targetFrameworkName;
 
@@ -155,5 +157,11 @@ internal static partial class LocalAppContextSwitches
     {
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         get => GetCachedSwitchValue(ServicePointManagerCheckCrlSwitchName, ref s_servicePointManagerCheckCrl);
+    }
+
+    public static bool DataGridViewStartRowColumnIndex
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        get => GetCachedSwitchValue(DataGridViewStartRowColumnIndexSwitchName, ref s_dataGridViewStartRowColumnIndex);
     }
 }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCell.DataGridViewCellAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCell.DataGridViewCellAccessibleObject.cs
@@ -3,6 +3,7 @@
 
 using System.ComponentModel;
 using System.Drawing;
+using System.Windows.Forms.Primitives;
 using Windows.Win32.System.Com;
 using Windows.Win32.System.Variant;
 using Windows.Win32.UI.Accessibility;
@@ -58,7 +59,7 @@ public abstract partial class DataGridViewCell
 
                 int rowIndex = _owner.DataGridView is null
                     ? -1
-                    : _owner.DataGridView.Rows.GetVisibleIndex(_owner.OwningRow);
+                    : _owner.DataGridView.Rows.GetVisibleIndex(_owner.OwningRow) + RowColumnStartIndex;
 
                 string name = string.Format(SR.DataGridView_AccDataGridViewCellName, _owner.OwningColumn.HeaderText, rowIndex);
 
@@ -115,6 +116,19 @@ public abstract partial class DataGridViewCell
         }
 
         public override AccessibleRole Role => AccessibleRole.Cell;
+
+        public int RowColumnStartIndex
+        {
+            get
+            {
+                if (LocalAppContextSwitches.DataGridViewStartRowColumnIndex)
+                {
+                    return 1;
+                }
+
+                return 0;
+            }
+        }
 
         public override AccessibleStates State
         {
@@ -725,12 +739,12 @@ public abstract partial class DataGridViewCell
 
         internal override int Row
             => _owner?.OwningRow?.Visible is true && _owner.DataGridView is not null
-                ? _owner.DataGridView.Rows.GetVisibleIndex(_owner.OwningRow)
+                ? _owner.DataGridView.Rows.GetVisibleIndex(_owner.OwningRow) 
                 : -1;
 
         internal override int Column
             => _owner?.OwningColumn?.Visible is true && _owner.DataGridView is not null
-                ? _owner.DataGridView.Columns.GetVisibleIndex(_owner.OwningColumn)
+                ? _owner.DataGridView.Columns.GetVisibleIndex(_owner.OwningColumn) 
                 : -1;
 
         internal override IRawElementProviderSimple.Interface? ContainingGrid => _owner?.DataGridView?.AccessibilityObject;
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRow.DataGridViewRowAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRow.DataGridViewRowAccessibleObject.cs
@@ -4,6 +4,7 @@
 using System.Drawing;
 using System.Globalization;
 using System.Text;
+using System.Windows.Forms.Primitives;
 using Windows.Win32.System.Variant;
 using Windows.Win32.UI.Accessibility;
 using static Interop;
@@ -35,7 +36,7 @@ public partial class DataGridViewRow
                 ? dataGridView.ColumnHeadersVisible
                     ? dataGridView.Rows.GetVisibleIndex(_owningDataGridViewRow) + 1
                     : dataGridView.Rows.GetVisibleIndex(_owningDataGridViewRow)
-                : -1;
+            : -1;
 
         public override Rectangle Bounds
         {
@@ -96,7 +97,7 @@ public partial class DataGridViewRow
                 }
 
                 int index = _owningDataGridViewRow is { Visible: true, DataGridView: { } }
-                        ? _owningDataGridViewRow.DataGridView.Rows.GetVisibleIndex(_owningDataGridViewRow)
+                        ? _owningDataGridViewRow.DataGridView.Rows.GetVisibleIndex(_owningDataGridViewRow) + RowColumnStartIndex
                         : -1;
 
                 return string.Format(SR.DataGridView_AccRowName, index.ToString(CultureInfo.CurrentCulture));
@@ -133,6 +134,19 @@ public partial class DataGridViewRow
         }
 
         public override AccessibleRole Role => AccessibleRole.Row;
+
+        public int RowColumnStartIndex
+        {
+            get
+            {
+                if (LocalAppContextSwitches.DataGridViewStartRowColumnIndex)
+                {
+                    return 1;
+                }
+
+                return 0;
+            }
+        }
 
         internal override int[] RuntimeId
             => _runtimeId ??= new int[]


ricardobossan:
Drafted the pull request because I could not yet configure properly the ScratchProject config files to produce the desired result on build. Had to manually add the switch on `ScratchProject.runtimeconfig.json` to test the changes:
image

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewRow.DataGridViewRowAccessibleObject.cs
```diff
@@ -35,7 +36,7 @@ private int VisibleIndex
                 ? dataGridView.ColumnHeadersVisible
                     ? dataGridView.Rows.GetVisibleIndex(_owningDataGridViewRow) + 1
                     : dataGridView.Rows.GetVisibleIndex(_owningDataGridViewRow)
-                : -1;
+            : -1;
```
Epica3055, line 39:
```suggestion
                : -1;
```
I guess this `:` should be aligned with `?` above not the `=>`

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCell.DataGridViewCellAccessibleObject.cs
```diff
@@ -116,6 +117,19 @@ public override string DefaultAction
...
+            get
+            {
+                if (LocalAppContextSwitches.DataGridViewStartRowColumnIndex)
+                {
+                    return 1;
+                }
+
+                return 0;
+            }
+        }
```
Epica3055, line 131: this could be simplified 
```suggestion
        public int RowColumnStartIndex => LocalAppContextSwitches.DataGridViewStartRowColumnIndex ? 1 : 0;
        
```

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -156,4 +158,10 @@ public static bool ServicePointManagerCheckCrl
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         get => GetCachedSwitchValue(ServicePointManagerCheckCrlSwitchName, ref s_servicePointManagerCheckCrl);
     }
+
+    public static bool DataGridViewStartRowColumnIndex
```
Epica3055, line 162: I think this property name kind of doesn't match the type `bool`.
In my understanding, I think the type of this property  should be `int` like
`public static int DataGridViewStartRowColumnIndex`
Or `public static bool DataGridViewStartRowColumnIndexStartWithOne`

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCell.DataGridViewCellAccessibleObject.cs
```diff
@@ -116,6 +117,19 @@ public override string DefaultAction
 
         public override AccessibleRole Role => AccessibleRole.Cell;
 
+        public int RowColumnStartIndex
```
Epica3055, line 120: Do we need to expose this property to users?

Epica3055:
- Is it possible to programmatically configure this option?
- Is it possible to write UnitTests?

ricardobossan:
> Drafted the pull request because I could not yet configure properly the ScratchProject config files to produce the desired result on build. Had to manually add the switch on `ScratchProject.runtimeconfig.json` to test the changes: image

Managed to apply the configuration in my tests, so that the option to start counting the rows on the DataGridView at 1 is applied.

a runtimeconfig.template.json file should be created at the root folder of the project, enabling that option, as it is said in the docs: https://learn.microsoft.com/en-us/dotnet/core/runtime-config/#runtimeconfigjson

ricardobossan:
@Epica3055 

> * Is it possible to programmatically configure this option?
> * Is it possible to write UnitTests?

- Yes. You can programmatically configure it, by adding a `runtimeconfig.template.json` at the root folder of the test project:

```json
{
    "configProperties": {
      "System.Windows.Forms.DataGridViewRowStartsAtOne": true
    }
}
```

source: https://learn.microsoft.com/en-us/dotnet/core/runtime-config/#runtimeconfigjson

- Added a unit test

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -156,4 +158,12 @@ public static bool ServicePointManagerCheckCrl
...
         get => GetCachedSwitchValue(ServicePointManagerCheckCrlSwitchName, ref s_servicePointManagerCheckCrl);
     }
+
+    public static bool DataGridViewRowStartsAtOne
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        get => GetCachedSwitchValue(DataGridViewRowStartsAtOneSwitchName, ref s_dataGridViewRowStartsAtOne);
+    }
+
+    internal static void Test_SetDataGridViewRowStartsAtOne(bool value) => s_dataGridViewRowStartsAtOne = value ? 1 : 0;
```
Epica3055, line 168: Is this `Test_SetDataGridViewRowStartsAtOne` a deliberate method name? If so, then why added a prefix `Test_`

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjects/DataGridViewRowAccessibleObjectTests.cs
```diff
@@ -2379,6 +2380,26 @@ public void DataGridViewRowAccessibleObject_GetPropertyValue_ValueValuePropertyI
         Assert.False(dataGridView.IsHandleCreated);
     }
 
+    [WinFormsFact]
+    public void DataGridView_SwitchConfigured_AdjustsRowAndColumnStartIndices()
+    {
+        var originalSwitchValue = LocalAppContextSwitches.DataGridViewRowStartsAtOne;
+        if (!originalSwitchValue)
+        { LocalAppContextSwitches.Test_SetDataGridViewRowStartsAtOne(true); }
```
Epica3055, line 2388: just for readability 
```suggestion
        {
            LocalAppContextSwitches.Test_SetDataGridViewRowStartsAtOne(true); 
        }
```

Epica3055:
I am not sure if we need this feature. 
But what I meant by _programmatically_ is : how to configure this option at run time.
Simply put, if there are two DataViews, I need one's startIndex to be 0 and the other to be 1. How to implement it?

@Tanya-Solyanik could you please take a look at this

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjects/DataGridViewCellAccessibleObjectTests.cs
```diff
@@ -1442,6 +1443,25 @@ private DataGridView CreateDataGridView(int columnCount, bool createControl = tr
...
+    {
+        Assert.False(LocalAppContextSwitches.DataGridViewRowStartsAtOne);
+
+        LocalAppContextSwitches.SetDataGridViewRowStartsAtOne(true);
+
+        using DataGridView dataGridView = new();
+        dataGridView.Columns.Add(new DataGridViewTextBoxColumn());
+        dataGridView.Rows.Add(new DataGridViewRow());
+
+        Assert.Equal(string.Format(SR.DataGridView_AccRowName, 1) + ", Not sorted.", dataGridView.Rows[0].Cells[0].AccessibilityObject.Name);
```
Tanya-Solyanik, line 1458: Please use string interpolation - https://learn.microsoft.com/dotnet/csharp/language-reference/tokens/interpolated . It generates less IL code. Even if this does not difference in tests, it's a better practice.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -20,12 +20,14 @@ internal static partial class LocalAppContextSwitches
     internal const string ServicePointManagerCheckCrlSwitchName = "System.Windows.Forms.ServicePointManagerCheckCrl";
     internal const string TrackBarModernRenderingSwitchName = "System.Windows.Forms.TrackBarModernRendering";
     private const string DoNotCatchUnhandledExceptionsSwitchName = "System.Windows.Forms.DoNotCatchUnhandledExceptions";
+    internal const string DataGridViewRowStartsAtOneSwitchName = "System.Windows.Forms.DataGridViewRowStartsAtOne";
```
Tanya-Solyanik, line 23: The bug was about both rows and columns, both can be implemented under the same switch - "System.Windows.Forms.DataGridViewRowColumnStartsAtOneInNarration"
Also I think it's important for the switch name to reflect that this feature applies only to accessibility announcements

src/System.Windows.Forms/src/System/Windows/Forms/DataGridViewCell.DataGridViewCellAccessibleObject.cs
```diff
@@ -116,6 +117,8 @@ public override string DefaultAction
 
         public override AccessibleRole Role => AccessibleRole.Cell;
 
+        private static int RowStartIndex => LocalAppContextSwitches.DataGridViewRowStartsAtOne ? 1 : 0;
```
Tanya-Solyanik, line 120: I think name of this property should reflect that this is applicable only to accessibility announcements, for example `RowStartIndexForAnnouncements`

Tanya-Solyanik (commented):
Looks good, but the bug was about both rows and columns, please expand the same implementation to columns under the same switch.

Tanya-Solyanik:
> if there are two DataViews, I need one's startIndex to be 0 and the other to be 1. How to implement it?

@Epica3055  - Accessibility tools should be consistent in how they announce columns or rows, any application business logic should be conveyed by appropriate names.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
Tanya-Solyanik: @merriemcgaw and I had discussed this issue again, we think it will not be too breaking, i.e. only UIA tests will be affected while the users of AT will benefit from the change. So we want 1-indexing to be "on" by default and we will provide an opt-out switch to support the broken automation tests. New suggested switch name "System.Windows.Forms.DataGridViewUIAStartCountAtZero", it will be `false` by default, i.e. your change will be "on" by default.

ricardobossan:
> Looks good, but the bug was about both rows and columns, please expand the same implementation to columns under the same switch.

@Tanya-Solyanik , I had made changes to the names of accessible objects, in order to change only how narrator announces row and column indexes, with no impact to existing logic.

However, columns are only announced in narrator by the values set to their header, **if there's one**, as seen below (default behavior, unchanged by this PR):

### If a value is set for column header:

image
- Red: The value "C" is announced for the column;
- Purple: Column's header value "C" is announced;
- Green: Narrator already announces column index considering 1 as starting index, when you change from one column to the other.

### Else if no value is set to column header:

image
- Red: There's no value announced for column
- Green: Narrator already announces column index considering 1 as starting index, when you change from one column to the other.

So existing behavior **does not** announce column indexes in Narrator.

Also, while there are `DataGridCellAccessibleObject` and `DataGridRowAccessibleObject` in code, where I could alter their `Name` fields, there's no `DataGridColumnAccessibleObject`  for me to do the same. I imagine that is because the column index is not announced in Narrator, as seen above.

So it seems to me that just changing the name of row is enough, as Narrator does not announce column indexes.

But if that's not enough, we can change Narrator's existing behavior for columns:

We could change how Narrator announces columns. If no value is set, it could announce the column index. That could be achieved by changing `DataGridViewAccessibleObject`, making it consider the column's index, in this section:

image

Tanya-Solyanik:
@ricardobossan  - thank you for the explanation, makes sense to change only the row-related announcements. Then we can name the switch "System.Windows.Forms.DataGridViewUIAStartRowCountAtZero"

Tanya-Solyanik:
@ricardobossan - when you fix these tests, please add a comment that whether indexing is 1-based or 0-based is controlled by a switch, this will help with debugging tests later. Alternatively, you can set the switch value explicitly in each test.

Tanya-Solyanik:
@ricardobossan  - the next step with this fix is to create a [breaking change issue with detailed description](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fdotnet%2Fdocs%2Fissues%2Fnew%3Fassignees%3Dgewarren%26labels%3Dbreaking-change%252CPri1%252Cdoc-idea%26projects%3D%26template%3Dbreaking-change.yml%26title%3D%255BBreaking%2Bchange%255D%253A%2B&data=05%7C01%7CTatiana.Solyanik%40microsoft.com%7Cc2432fff5dcd420d9f0b08db7748d5b7%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C638234926932469199%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=bYVvtAiSgbhdcxTnzdzOaYBjkpH0LhT5EFFtITYmgFs%3D&reserved=0). Then pink me to review it. Thanks!

ricardobossan:
> @ricardobossan - the next step with this fix is to create a [breaking change issue with detailed description](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fdotnet%2Fdocs%2Fissues%2Fnew%3Fassignees%3Dgewarren%26labels%3Dbreaking-change%252CPri1%252Cdoc-idea%26projects%3D%26template%3Dbreaking-change.yml%26title%3D%255BBreaking%2Bchange%255D%253A%2B&data=05%7C01%7CTatiana.Solyanik%40microsoft.com%7Cc2432fff5dcd420d9f0b08db7748d5b7%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C638234926932469199%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=bYVvtAiSgbhdcxTnzdzOaYBjkpH0LhT5EFFtITYmgFs%3D&reserved=0). Feel free to add a sample json file to opt out from this change. Then ping me to review it. Thanks!

@Tanya-Solyanik , how's this? https://github.com/dotnet/docs/issues/38527

merriemcgaw:
@ricardobossan your docs Issue looks great.
Do we have this enabled by default? Do all of the accessibility tools (JAWs NVDA etc) recognize the new count? I know some accessibility champs that are going to be _very_ happy!! 🎉

PR #10247, opened 2023-11-08
Implement the ITypeDiscoveryService service

### LINKED ISSUE #9870: Column types cannot be displayed in the DataGridView's Add Column window
### Issue description
The column type cannot be show in the Add Column window of the DataGridView
Related to #4908 

### Steps to reproduce

1. Clone current repo
2. Set "DemoConsole" as startup project and debug the project
3. Switch tab to "TabControl and TableLayoutPanel"
4. Try to add an Unbound column to DataGridView 

Actual:
The column type is empty
image

Expected: 
The column type show up
image

LeafShi1:
Fixes #9870

## Proposed changes

- Add ITypeDiscoveryService in default service container `ServiceContainer `of IDesignerHost
- Implement the ITypeDiscoveryService service

## Root cause of this issue:
- In the `PopulateColumnTypesCombo ` function, the code attempts to obtain an `ITypeDiscoveryService `instance through the GetService method of `IDesignerHost `in order to discover all subtypes of `DataGridViewColumn`.
However, because `ITypeDiscoveryService` is not added in the default service container `ServiceContainer` of `IDesignerHost`, `GetService `returns null.

https://github.com/dotnet/winforms/blob/1d604b5425cbbaf54f2b5be181096d1cf6b0271d/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DataGridViewAddColumnDialog.cs#L657-L673

## Customer Impact

- The DataGridViewColumn can be added normally in DemoConsole project

## Risk

- Minimal

### Before
The column type is empty
image

### After
The DataGridView column can be added normally from the Add column page

AfterChange

## Test methodology 

- Manually (Add test to Dome project)

## Test environment(s) 

- .net 9.0.0-alpha.1.23519.1

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/TypeDiscoveryService.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/TypeDiscoveryService.cs
@@ -1,0 +1,56 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Collections;
+using System.Collections.Concurrent;
+using System.Collections.Immutable;
+using System.Reflection;
+
+namespace System.ComponentModel.Design;
+
+/// <summary>
+///  This service is requested by TypeDescriptor when asking for type information for a component.
+/// </summary>
+internal sealed class TypeDiscoveryService : ITypeDiscoveryService
+{
+    internal TypeDiscoveryService()
+    {
+    }
+
+    private readonly ConcurrentDictionary<Type, ImmutableArray<Type>> _discoveredTypesCache = new();
+
+    public ICollection GetTypes(Type? baseType, bool excludeGlobalTypes)
+    {
+        return baseType is null
+            ? throw new ArgumentNullException(nameof(baseType))
+            : (ICollection)_discoveredTypesCache.GetOrAdd(baseType, type => FindTypes(type, AppDomain.CurrentDomain.GetAssemblies()));
+
+        static ImmutableArray<Type> FindTypes(Type baseType, Assembly[] assemblies)
+        {
+            var builder = ImmutableArray.CreateBuilder<Type>();
+
+            foreach (var assembly in assemblies)
+            {
+                Type[] types;
+                try
+                {
+                    types = assembly.GetTypes();
+                }
+                catch (ReflectionTypeLoadException exception)
+                {
+                    types = exception.Types!;
+                }
+
+                foreach (var type in types)
+                {
+                    if (baseType.IsAssignableFrom(type))
+                    {
+                        builder.Add(type);
+                    }
+                }
+            }
+
+            return builder.ToImmutable();
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms.Design/src/System/ComponentModel/Design/TypeDiscoveryService.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+    }
+
+    private readonly ConcurrentDictionary<Type, ImmutableArray<Type>> _discoveredTypesCache = new();
+
+    public ICollection GetTypes(Type? baseType, bool excludeGlobalTypes)
+    {
+        return baseType is null
+            ? throw new ArgumentNullException(nameof(baseType))
+            : (ICollection)_discoveredTypesCache.GetOrAdd(baseType, type => FindTypes(type, AppDomain.CurrentDomain.GetAssemblies()));
```
Tanya-Solyanik, line 26: Implementation of this service depends on the application that consumes it. For example, application might want to see types that are not loaded yet. Could you please move this type to demo application instead? It might require addition of the service providers.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/TypeDiscoveryService.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Reflection;
+
+namespace System.ComponentModel.Design;
+
+/// <summary>
+///  This service is requested by TypeDescriptor when asking for type information for a component.
+/// </summary>
+internal sealed class TypeDiscoveryService : ITypeDiscoveryService
+{
+    internal TypeDiscoveryService()
```
Tanya-Solyanik, line 16: Is the default constructor required?

LeafShi1: > Is the default constructor required?

No, it isn't, I just referred to the implementation of other services.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/TypeDiscoveryService.cs
LeafShi1: > Implementation of this service depends on the application that consumes it. For example, application might want to see types that are not loaded yet. Could you please move this type to demo application instead? It might require addition of the service providers.

I directly add the `GetTypes ` method in `DataGridViewAddColumnDialog.cs` instead of using service. Do you think this is okay?

> Implementation of this service depends on the application that consumes it. For example, application might want to see types that are not loaded yet. Could you please move this type to demo application instead? It might require addition of the service providers.

src/System.Windows.Forms/tests/IntegrationTests/DesignSurface/DesignSurfaceExt/TypeDiscoveryService.cs
```diff
@@ -0,0 +1,54 @@
...
+
+using System.Collections;
+using System.Collections.Concurrent;
+using System.Collections.Immutable;
+using System.Reflection;
+
+namespace System.ComponentModel.Design;
+
+/// <summary>
+///  This service is requested by TypeDescriptor when asking for type information for a component.
```
Tanya-Solyanik, line 12: Please use xml-doc tags in comments - https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/recommended-tags#see

Tanya-Solyanik, line 12:
```suggestion
///  This service is requested by <see cref="TypeDescriptor" /> when asking for type information for a component.
///  This is a sample implementation suitable for this sample application.
```

Tanya-Solyanik (commented):
Looks good, I added one minor comment.

PR #10253, opened 2023-11-08
Enable nullability in remaining BindingSource members

gpetrou:
## Proposed changes

- Enable nullability in remaining BindingSource members.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10253)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/BindingSource.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/BindingSource.cs
@@ -78,7 +78,8 @@ public class BindingSource : Component,
     {
     }
 
-    public BindingSource(object? dataSource, string dataMember) : base()
+    public BindingSource(object? dataSource, string dataMember)
+        : base()
     {
         // Set data source and data member
         _dataSource = dataSource;
@@ -563,28 +564,28 @@ public class BindingSource : Component,
         return instancedObject;
     }
 
-    private void CurrencyManager_PositionChanged(object sender, EventArgs e)
+    private void CurrencyManager_PositionChanged(object? sender, EventArgs e)
     {
         Debug.Assert(sender == _currencyManager, "only receive notifications from the currency manager");
         OnPositionChanged(e);
     }
 
-    private void CurrencyManager_CurrentChanged(object sender, EventArgs e)
+    private void CurrencyManager_CurrentChanged(object? sender, EventArgs e)
     {
         OnCurrentChanged(EventArgs.Empty);
     }
 
-    private void CurrencyManager_CurrentItemChanged(object sender, EventArgs e)
+    private void CurrencyManager_CurrentItemChanged(object? sender, EventArgs e)
     {
         OnCurrentItemChanged(EventArgs.Empty);
     }
 
-    private void CurrencyManager_BindingComplete(object sender, BindingCompleteEventArgs e)
+    private void CurrencyManager_BindingComplete(object? sender, BindingCompleteEventArgs e)
     {
         OnBindingComplete(e);
     }
 
-    private void CurrencyManager_DataError(object sender, BindingManagerDataErrorEventArgs e)
+    private void CurrencyManager_DataError(object? sender, BindingManagerDataErrorEventArgs e)
     {
         OnDataError(e);
     }
@@ -736,7 +737,7 @@ public class BindingSource : Component,
         return false;
     }
 
-    private void InnerList_ListChanged(object sender, ListChangedEventArgs e)
+    private void InnerList_ListChanged(object? sender, ListChangedEventArgs e)
     {
         // Set recursive flag
         // Basically, we can have computed columns that cause our parent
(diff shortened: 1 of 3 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/BindingSource.cs
```diff
@@ -1785,13 +1787,13 @@ public virtual void ApplySort(ListSortDescriptionCollection sorts)
     [EditorBrowsable(EditorBrowsableState.Never)]
     public virtual ListSortDescriptionCollection SortDescriptions
     {
-        get => List is IBindingListView iblv ? iblv.SortDescriptions : null;
+        get => List is IBindingListView iblv ? iblv.SortDescriptions : null!;
```
lonitra, line 1790: Callers of this are checking if this is null so it seems like we should be marking the return as nullable instead of doing a `!`

src/System.Windows.Forms/src/System/Windows/Forms/BindingSource.cs
```diff
@@ -1549,14 +1551,14 @@ public virtual void RemoveAt(int index)
     [Browsable(false)]
     public virtual bool IsReadOnly => List.IsReadOnly;
 
-    public virtual string GetListName(PropertyDescriptor[] listAccessors)
+    public virtual string GetListName(PropertyDescriptor[]? listAccessors)
     {
         return ListBindingHelper.GetListName(List, listAccessors);
     }
```
lonitra, line 1557: nit: body expression

src/System.Windows.Forms/src/System/Windows/Forms/BindingSource.cs
gpetrou: The problem here is that this implements `IBindingListView.SortDescriptions` which is non-nullable. https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.ibindinglistview.sortdescriptions does not seem to have it as nullable. Perhaps I should make it here and just disable the warning?

gpetrou: I added a separate commit with the last proposed change.

lonitra: Ah I see, I will get team's thoughts on this. This would be a breaking change since we've already documented that we will return `null` (https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.bindingsource.sortdescriptions?view=windowsdesktop-7.0) but this is contradicting annotations on `IBindingListView.SortDescriptions`

JeremyKuhne: Personally, I think we should return an empty collection and update the docs. I'd wait for @KlausLoeffelmann's opinion as well (it may be a few days).

KlausLoeffelmann:
FWIW, this can be non-nullable. There doesn't seem to be a spot, where a concrete implementation of this does return `null`. Also, please keep in mind, that guarding against `null` _always_ is a must despite using Nullables. These are compile-time agreements, not runtime checks! And VB doesn't understand those, so, a VB developer wouldn't even get a warning that they've passed an argument which has a "different signature": VB simply does not have nullable aware contexts, nor nullable reference types, nor the null static analysis warnings, and also not the null-forgiving operator.

KlausLoeffelmann:
Or in other words: What @JeremyKuhne said. 😸

KlausLoeffelmann:
That said...

> Also, please keep in mind, that guarding against null always is a must despite using Nullables...

we cannot retrofit guarding and then behave differently than before (or even throw). 
That would be a breaking change, I would not at all support.

src/System.Windows.Forms/src/System/Windows/Forms/BindingSource.cs
gpetrou: I updated the last commit to fix the failing tests.

src/System.Windows.Forms/src/System/Windows/Forms/BindingSource.cs
```diff
@@ -1479,12 +1481,12 @@ public virtual int Add(object value)
         }
 
         // Throw if user tries to add items to list that don't match the current item type
-        if (value is not null && !_itemType.IsAssignableFrom(value.GetType()))
+        if (value is not null && !_itemType!.IsAssignableFrom(value.GetType()))
```
lonitra, line 1484: Can you add comment that `_itemType` will not be null here because the above code will have set it either with the call to `List` or `SetList`. This will save some time for someone looking at the code later down the road.

gpetrou: OK. I added a separate commit.

lonitra (approved):
🚀

lonitra:
https://github.com/dotnet/docs/issues/38363

PR #10254, opened 2023-11-08
Use ITypeInfo directly in StandardDispatch

JeremyKuhne:
CreateStdDispatch doesn't do anything other than pass through to ITypeInfo so we'll do the same directly and save some of the overhead and make the code clearer.

To avoid loading and unloading the ITypeLib repeatedly we'll simply keep it alive (via the ITypeInfo) once we've loaded it once. We can expect that in an app that uses these we'll continue to need them throughout the lifetime of the app.

This allows removing the IWrapperInitialize interface, which further simplifies the way things work. I'll follow up with a change that simplifies our ComWrappers a bit based on this.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10254)


JeremyKuhne:
@AaronRobinsonMSFT, @jkoritzinsky this is a revision of the pattern we're using to replace IReflect IDispatch support.

@jacdavis fyi.

src/System.Windows.Forms.Primitives/src/Windows/Win32/System/Com/StandardDispatch.cs
```diff
@@ -11,95 +10,84 @@ namespace Windows.Win32.System.Com;
...
 {
-    private readonly Guid _typeLibrary;
-    private readonly ushort _majorVersion;
-    private readonly ushort _minorVersion;
-    private readonly Guid _interfaceId;
-    private AgileComPointer<IDispatch>? _standardDispatch;
-
-    // StdOle32.tlb
-    private static readonly Guid s_stdole = new("00020430-0000-0000-C000-000000000046");
+    private ITypeInfo* _typeInfo;
```
lonitra, line 22: Should this be in an AgileComPointer?

JeremyKuhne: It doesn't need to be proxied, so I don't want to put it in the global interface table. We could add a finalizer to release it, but we're never actually releasing the ITypeInfo so I didn't want to add the overhead.

PR #10268, opened 2023-11-09
Enable nullability in some DataGridViewCell members

gpetrou:
## Proposed changes

- Enable nullability in some DataGridViewCell members.


gpetrou:
@lonitra  any feedback on this one?

JeremyKuhne (changes requested):
I believe we should not be making the `ToolTipText` nullable and instead should fall back to `string.Empty` when necessary.

gpetrou:
> I believe we should not be making the `ToolTipText` nullable and instead should fall back to `string.Empty` when necessary.

Hmm, do you see a specific problem here? Going though the existing code, we seem to be having null checks when needed, so I assume that this was meant to be nullable.

JeremyKuhne:
> Hmm, do you see a specific problem here? Going though the existing code, we seem to be having null checks when needed, so I assume that this was meant to be nullable.

@gpetrou We should get consistent for all `ToolTipText`. Better that we always prevent null then have some that do and some that don't. See `ListViewItem.ToolTipText` for the ideal pattern.

gpetrou:
@JeremyKuhne @lonitra any more feedback on this one?

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView/DataGridViewCell.cs
```diff
@@ -2571,64 +2542,55 @@ public virtual void InitializeEditingControl(int rowIndex, object initialFormatt
 
         if (AccessibleRestructuringNeeded)
         {
-            dgv.EditingControlAccessibleObject.SetParent(AccessibilityObject);
+            dgv.EditingControlAccessibleObject!.SetParent(AccessibilityObject);
```
lonitra, line 2545: Please add comment that this cannot be null since we have already checked `EditingControl` is not null

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView/DataGridViewCell.cs
```diff
@@ -2571,64 +2542,55 @@ public virtual void InitializeEditingControl(int rowIndex, object initialFormatt
...
+        string? text,
+        Font font,
+        int maxWidth,
+        TextFormatFlags flags) => MeasureTextHeight(
+            graphics,
+            text,
+            font,
+            maxWidth,
+            flags,
+            out bool widthTruncated);
```
lonitra, line 2584: It looks like this is unused
```suggestion
            out _);
```

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView/DataGridViewCell.cs
```diff
@@ -2514,12 +2488,12 @@ protected virtual object GetValue(int rowIndex)
...
         else if (OwningColumn is not null && OwningColumn.IsDataBound)
         {
-            DataGridView.DataGridViewDataConnection dataConnection = dataGridView.DataConnection;
+            DataGridView.DataGridViewDataConnection? dataConnection = dataGridView.DataConnection;
             if (dataConnection is null)
             {
                 return null;
             }
-            else if (dataConnection.CurrencyManager.Count <= rowIndex)
+            else if (dataConnection.CurrencyManager!.Count <= rowIndex)
```
lonitra, line 2496: How do we know this is not null here?

lonitra (approved):

gpetrou:
@lonitra could we merge this one?

lonitra (approved):

PR #10270, opened 2023-11-10
Rename files containg DataGridView interfaces

halgab:
See https://github.com/dotnet/winforms/pull/10231#discussion_r1388617034

The downside to this change is that DataGridView related files will no longer be next to each other in the explorer. Perhaps we could put all of them in a dedicated subfolder?

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10270)


JeremyKuhne:
> Perhaps we could put all of them in a dedicated subfolder?

I'm sorely tempted to put some of these files in subdirectories. In general we want to match the namespace, but it is getting unwieldly in the primary namespace. I'll discuss it at our next internal runtime sync.

PR #10274, opened 2023-11-10
Add a project for testing trimming

JeremyKuhne:
Adds a simple sample project for testing trimming. It has a TextBox, SplitContainer, Button, and ComboBox. It accepts text drops into the TextBox and uses the ClipBoard class to copy text out of the ComboBox.

This is meant to be a starting point to assess and fix the most important warnings / issues with trimming.

@vitek-karas, @LakshanF, @sbomer, @agocke 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10274)

### DIFF
--- a/src/System.Windows.Forms/tests/IntegrationTests/TrimTest/TrimTest.csproj
+++ b/src/System.Windows.Forms/tests/IntegrationTests/TrimTest/TrimTest.csproj
@@ -1,0 +1,25 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>WinExe</OutputType>
+    <Nullable>enable</Nullable>
+    <OutputType>WinExe</OutputType>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <!-- These are needed to suppress the localization step picked up from Arcade targets -->
+    <EnableXlfLocalization>false</EnableXlfLocalization>
+    <UpdateXlfOnBuild>false</UpdateXlfOnBuild>
+  </PropertyGroup>
+
+  <!-- These normally come from $(UseWindowsForms) when $(ImplicitUsings) is enabled -->
+  <ItemGroup>
+    <Using Include="System.Drawing" />
+    <Using Include="System.Windows.Forms" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\..\..\System.Drawing.Common\src\System.Drawing.Common.csproj" />
+    <ProjectReference Include="..\..\..\src\System.Windows.Forms.csproj" />
+  </ItemGroup>
+
+</Project>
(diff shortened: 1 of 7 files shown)


src/System.Windows.Forms/tests/IntegrationTests/TrimTest/TrimTest.csproj
```diff
@@ -0,0 +1,25 @@
...
+  <PropertyGroup>
+    <OutputType>WinExe</OutputType>
+    <Nullable>enable</Nullable>
+    <OutputType>WinExe</OutputType>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <!-- These are needed to suppress the localization step picked up from Arcade targets -->
+    <EnableXlfLocalization>false</EnableXlfLocalization>
+    <UpdateXlfOnBuild>false</UpdateXlfOnBuild>
+  </PropertyGroup>
```
Tanya-Solyanik, line 12: Is `<PublishTrimmed>true</PublishTrimmed>`  coming from somewhere else?

vitek-karas: I think we should add it here - adding it on a command line won't help here, it will trigger an error because the property would be applied to dependent projects - in this case source generators.

Or we might as well use `PublishAot=true` instead, since that's the final goal anyway.

Currently doing so also triggers another error:
```dotnetcli
Unable to
find package Microsoft.WindowsDesktop.App.Runtime.win-x64 with version (= 9.0.0-alpha.1.23560.1)
```

I assume this is because the project gets TFM from the repo, but it would also need a live built WindowsDesktop runtime pack.

LakshanF: I haven't tried this on the WinForms repo but am able to see the trim warnings with `PublishTrimmed ` set to true with these files as an app. I also needed to bypass the default publish error by setting the property `_SuppressWinFormsTrimError` to true (and also needed to the set the property `SuppressTrimAnalysisWarnings `to false to see the warnings after that).

JeremyKuhne: > Is `<PublishTrimmed>true</PublishTrimmed>` coming from somewhere else?

@Tanya-Solyanik I deliberately didn't set it, this was just to make a sample app that we want to start trying to trim.

@LakshanF My hope is that you can help figure out how to get past the error that @vitek-karas mentioned. I hit the same thing.

Tanya-Solyanik: Ok, once the workaround to build this project as trimmable is known, please add it to the readme file.

Tanya-Solyanik (approved):

Amy-Li03:
Tested the TrimTest project on winforms repo from .NET 9.0 branch, they work well, we can drag drop text to TextBox and copy text out of the ComboBox.

PR #8673, opened 2023-02-22
Refactor `DropSourceBehaviour` and related code to replace `ArrayList` with `List<T>`

elachlan:
Related: #8143, #8357

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/8673)

PR #10278, opened 2023-11-11
Use List<T> instead of ICollection in BehaviorDataObject

halgab:
Finish work started in #8673 
Related: #8140

Now that `DropSourceBehavior` is passed a `List<IComponent>` in its constructor, we can update `BehaviorDataObject` to do the same instead of using `ICollection`

## Proposed changes

- Refactor `BehaviorDataObject.DragComponents` to be a `List<IComponent>` and update code accordingly
- Another small `BehaviorDataObject` refacto while we're touching this code, but in a separate commit

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10278)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ParentControlDesigner.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ParentControlDesigner.cs
@@ -1478,8 +1478,7 @@ public partial class ParentControlDesigner : ControlDesigner, IOleDragClient
         object[] dragComps;
         if (behDataObject is not null && behDataObject.DragComponents is not null)
         {
-            dragComps = new object[behDataObject.DragComponents.Count];
-            behDataObject.DragComponents.CopyTo(dragComps, 0);
+            dragComps = behDataObject.DragComponents.ToArray();
         }
         else
         {
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/Behavior/DropSourceBehavior.BehaviorDataObject.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/Behavior/DropSourceBehavior.BehaviorDataObject.cs
@@ -33,6 +33,6 @@ internal sealed partial class DropSourceBehavior
 
         internal void CleanupDrag() => _sourceBehavior.CleanupDrag();
 
-        internal List<IComponent> GetSortedDragControls(ref int primaryControlIndex) => _sourceBehavior.GetSortedDragControls(ref primaryControlIndex);
+        internal List<IComponent> GetSortedDragControls(out int primaryControlIndex) => _sourceBehavior.GetSortedDragControls(out primaryControlIndex);
     }
 }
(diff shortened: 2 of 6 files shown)


src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ParentControlDesigner.cs
```diff
@@ -1478,8 +1478,7 @@ protected override void OnDragEnter(DragEventArgs de)
         object[] dragComps;
```
elachlan, line 1478: `dragComps` can not be `IComponent[]`

elachlan:
As apart of this change maybe refactor this?
https://github.com/dotnet/winforms/blob/cc1a96870bfb794efc2005b843f3908e422865fe/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/TableLayoutPanelDesigner.cs#L967-L977

Something like:
```csharp
private Control ExtractControlFromDragEvent(DragEventArgs de) =>
    (de.Data as DropSourceBehavior.BehaviorDataObject)?.DragComponents.FirstOrDefault() as Control;
```

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ParentControlDesigner.cs
halgab: It doesn't look that simple... `OleDragDropHandler.GetDraggingObjects` return an `object[]`. I think it can be refactored to return an `IComponent[]`, but that's more involved. I'm not sure I should add it to this PR

halgab:
> As apart of this change maybe refactor this?
> 
> https://github.com/dotnet/winforms/blob/cc1a96870bfb794efc2005b843f3908e422865fe/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/TableLayoutPanelDesigner.cs#L967-L977
> 
> Something like:
> 
> ```cs
> private Control ExtractControlFromDragEvent(DragEventArgs de) =>
>     (de.Data as DropSourceBehavior.BehaviorDataObject)?.DragComponents.FirstOrDefault() as Control;
> ```

Nice catch, I hadn't noticed this one. `dragComps` is a field though, so we have to keep it around. We could definitely change it to be a `List<IComponent>` after `DragComponents` itself becomes generic. As it's a change to a field in another type, perhaps is it more suited to a follow-up PR?

elachlan:
I agree, follow up PR to make this change quicker to review.

JeremyKuhne (approved):
Thanks!

halgab:
> I agree, follow up PR to make this change quicker to review.

#10287

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ParentControlDesigner.cs
halgab: Looking at it again, it might be possible, but I'm not comfortable with making such a change. That would lead to assumptions that I'm not sure would hold all the time.

elachlan: Yeah, it depends if the team ever decided to implement another Component interface.

PR #10279, opened 2023-11-12
Refactor `SnapLines` to use `IList` directly instead of casting to `ArrayList`

elachlan:
Kind of related: #8140

Mostly trying to remove `ArrayList` usage through the code base. This code is using a typesafe cast to `ArrayList`, where it can just use IList directly.

We cannot convert to `List<T>` because we have a public API that uses `ArrayList` ParentControlDesigner.AddPaddingSnapLines:
https://github.com/dotnet/winforms/blob/cc1a96870bfb794efc2005b843f3908e422865fe/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ParentControlDesigner.cs#L295-L319

Ultimately it would be great if the public API could be changed to `List<SnapLine>` to be more explicit.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10279)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/AdapterHelpers.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/AdapterHelpers.cs
@@ -1,0 +1,32 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Collections;
+
+namespace System.Windows.Forms.Design;
+
+/// <summary>
+/// Helpers for adapting between generic and non-generic lists.
+/// </summary>
+internal static class AdapterHelpers
+{
+    /// <summary>
+    ///  Provides an extension method to safely extract the underlying non-generic <see cref="IList"/>
+    ///  from a wrapped <see cref="IList&lt;T&gt;"/> when available.
+    /// </summary>
+    /// <typeparam name="T">The type of elements in the <see cref="IList&lt;T&gt;"/>.</typeparam>
+    /// <param name="list">The IList&lt;T&gt; to unwrap.</param>
+    /// <returns>The underlying non-generic <see cref="IList"/> if available; otherwise, the original <see cref="IList&lt;T&gt;"/>.</returns>
+    internal static IList Unwrap<T>(this IList<T> list) => list is IWrapper<IList> wrapper ? wrapper.Unwrap() : (IList)list;
+
+    /// <summary>
+    ///  Provides an extension method to adapt a non-generic IList to a generic <see cref="IList&lt;T&gt;"/> by creating
+    ///  a <see cref="ListAdapter&lt;T&gt;"/> when needed.
+    /// </summary>
+    /// <typeparam name="T">The desired type of elements in the resulting <see cref="IList&lt;T&gt;"/>.</typeparam>
+    /// <param name="list">The non-generic <see cref="IList"/> to adapt.</param>
+    /// <returns>
+    /// A generic <see cref="IList&lt;T&gt;"/> if the input IList can be cast to <see cref="IList&lt;T&gt;"/> otherwise, a new <see cref="ListAdapter&lt;T&gt;"/> wrapping the input IList.
+    /// </returns>
+    internal static IList<T> Adapt<T>(this IList list) => list is IList<T> iList ? iList : new ListAdapter<T>(list);
+}
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/AdapterHelpers.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/AdapterHelpers.cs
@@ -6,7 +6,7 @@ using System.Collections;
 namespace System.Windows.Forms.Design;
 
 /// <summary>
-/// Helpers for adapting between generic and non-generic lists.
+///  Helpers for adapting between generic and non-generic lists.
 /// </summary>
 internal static class AdapterHelpers
 {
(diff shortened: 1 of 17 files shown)


src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/TextBoxBaseDesigner.cs
halgab, line 1: Changes in this file seem only stylistic. I root for them, but don't look like they belong to this PR

elachlan:
I'll set aside some time and get the last changes sorted.

elachlan:
@JeremyKuhne I am not really following on how the `ListAdapter` helps here? `IList<T>` can't be implicitly converted to `IList`.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/FlowPanelDesigner.cs
```diff
@@ -1,8 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-#nullable disable
```
halgab, line 4: I'm not sure annotating `FlowPanelDesigner` for NRT should be part of this PR.

JeremyKuhne:
> I am not really following on how the `ListAdapter` helps here? `IList<T>` can't be implicitly converted to `IList`.

@elachlan `ListAdapter` is for our own "casting" to `IList<T>`. We know we want only `T`, not `T?`, and not `object?`. If the data structure is `IList<T>` it just casts to it, otherwise it creates an adapter for the `IList` from the property. That way we can write consuming code as if it is always `IList<T>`.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/FlowPanelDesigner.cs
elachlan: Are you sure? Its a very small change.

elachlan:
@JeremyKuhne I think it clicked. Let me know if this makes sense.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/FlowPanelDesigner.cs
halgab: I though the parent class wasn't annotated, but apparently it is, so in that light it's a small change (though the parent's parent isn't, which exposes us to incorrect assumptions)

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/LabelDesigner.cs
```diff
@@ -93,7 +93,7 @@ public override IList SnapLines
                 }
             }
 
-            return snapLines;
+            return (IList)snapLines;
```
JeremyKuhne, line 96: This won't work if the adapter is in play. One could add `IList` to the adapter, but then everything would get the wrapper at that point. If they try to do the same thing again.

You can just not expose the internal and do the `ListAdapter` call in each usage and return the original `IList`. Another option is to create some internal interface that we could use as a general pattern to handle this sort of wrapping. Something like:

``` C#
internal interface IWrapper<T>
{
    T Unwrap();
}

internal ListAdapter<T> : IList<T>, IWrapper<IList>
{
    // ...
}

internal static AdapterHelpers
{
    internal IList Unwrap(this IList list) => list is IWrapper<IList> wrapper ? wrapper.Unwrap() : (IList)list;
    internal IList<T> Adapt(this IList list) => list is IList<T> ilist ? ilist : new ListWrapper(list);
}

// Example usage:

IList<SnapLine> snapLines = SnapLines.Adapt<SnapLine>();
// ...
return snapLines.Unwrap();
```

JeremyKuhne (changes requested):
Added some comments. Sorry about the delay, a bit busy currently.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/LabelDesigner.cs
elachlan: I think this results in having to use `List<SnapLine>` and not `IList<SnapLine>` so as to avoid `((IList)snapLines).Unwrap();`

```csharp
// Example usage:

List<SnapLine> snapLines = base.SnapLines.Adapt<SnapLine>();
// ...
return snapLines.Unwrap();
```

Or should I just do the explicit cast?

JeremyKuhne:
@elachlan this looks good outside of adding comments to the new types and making sure they are in individual files per guidelines.

JeremyKuhne (changes requested):
Looks good, just get comments on the new types and move them into their own files.

elachlan:
@JeremyKuhne how is that?

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/AdapterHelpers.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Collections;
+
+namespace System.Windows.Forms.Design;
+
+/// <summary>
+/// Helpers for adapting between generic and non-generic lists.
```
JeremyKuhne, line 9: nit: indent

JeremyKuhne (changes requested):
Just one more nit

JeremyKuhne (approved):

elachlan:
Weird test failure I think is unrelated.

PR #10281, opened 2023-11-12
Refactor `WeakRefCollection` to replace `ArrayList` with `List<T>`

elachlan:
Related: #8140

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10281)

### DIFF
--- a/src/System.Drawing.Common/src/System/Drawing/ClientUtils.cs
+++ b/src/System.Drawing.Common/src/System/Drawing/ClientUtils.cs
@@ -1,7 +1,8 @@
-// Licensed to the .NET Foundation under one or more agreements.
+﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Collections;
+using System.Collections.Generic;
 using System.Security;
 
 namespace System.Drawing;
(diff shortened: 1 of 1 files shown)


elachlan:
@halgab could you please give this a quick review?

halgab (approved):
Happy to help!
I just see a minor potential improvement.

I don;t know if you've noticed, but this type is also defined [here](https://github.com/dotnet/winforms/blob/ad0a844cca5e4fc02217dd585823531588356ff1/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/WeakRefCollection.cs#L6). I don't know if code consolidation between the winforms projects and System.Drawing.Common is something that should be considered, knowing that the latter targetting netfx makes it less profitable.

Other open questions I have about this type (but that are outside the scope of this PR):

- should this be made generic?
- if that's the case, should we be using `WeakReference<T>` under the hood? And more generally should we try and replace `WeakReference` with its generic counterpart across codebase?

JeremyKuhne:
@elachlan I don't think there is any real difference between the two `WeakRefCollection` classes. Presuming they are the same, it would be worth merging to `Common` and linking into both projects.

It would also be worth looking at whether these can be typed. Can this be `WeakRefCollection<T>`? And more correctly: `WeakReferenceList<T> : IList<T>`??

elachlan:
@JeremyKuhne what namespace do you want me to use if we merge them?

elachlan:
Closing in favor of #10291

Issue #10259, opened 2023-11-09
[Accessibility] Narrator cannot focus on and announce the list items when pressing up/down keyboard for ListView/ListBox/CheckListBox


Amy-Li03:
### .NET version

.NET 9.0.100-alpha.1.23558.10

### Did it work in .NET Framework?

Yes

### Did it work in any of the earlier releases of .NET Core or .NET 5+?

It's a regression issue, doesn't repro on .NET 8.0/7.0/6.0.

### Issue description

Narrator cannot focus on and announce the list items when pressing up/down keyboard for ListView/ListBox/CheckListBox.

**The behavior for .NET 9.0:**
![9 0](https://github.com/dotnet/winforms/assets/62929087/4d70623b-6503-48bb-91d6-e2f6ff479540)
![image](https://github.com/dotnet/winforms/assets/62929087/9432a837-3e89-4405-9c1d-670baeb188af)

**The behavior for .NET 8.0:**
![8 0](https://github.com/dotnet/winforms/assets/62929087/03f0f719-4c25-480f-8c4d-5722ec12d07f)

### Steps to reproduce
[WinFormsApp134.zip](https://github.com/dotnet/winforms/files/13304574/WinFormsApp134.zip)

1. Create a WinForms .NET Core application with ListView control.
2. Add some columns and Items for it.
3. Set ListView's View as Details.
4. Build and run the application.
5. Launch the Narrator tool, try to focus on the first item of ListView.
6. Press up/down keyboard to focus on the next item.

Eudora-Li01:
verified this issue on .NET 9.0+latest dll files, it has been fixed. narrator can focus on and announce the list items when pressing up/down keyboard for ListView/ListBox/CheckListBox controls.

https://github.com/dotnet/winforms/assets/130345015/c04c185f-e0d0-4aba-8c4f-2c8b0d8c237a

Issue #10258, opened 2023-11-09
[Accessibility] Accessibility Insights tool keeps loading when detecting listBox control with scrollbar


Philip-Wang01:
### .NET version

9.0.100-alpha.1.23558.10

### Did it work in .NET Framework?

Yes

### Did it work in any of the earlier releases of .NET Core or .NET 5+?

It's a regression issue. Not repro on .NET 8.0/7.0/6.0.

### Issue description

Accessibility Insights tool keeps loading and does not display test results.

https://github.com/dotnet/winforms/assets/86937911/e54014b2-8d7e-45b7-a50e-66abc009a750

![image](https://github.com/dotnet/winforms/assets/86937911/bb9349e7-cfa2-4be5-b929-0bb0aa7e3c26)

### Steps to reproduce

1. Create a WinForms .NET Core application with ListBox control.
2. Add multiple Items to the ListBox and make sure it has scrollbars.
3. Build and run the application.
4. Open the Accessibility Insights tool, Select the ListBox and click Test Element.

**More Info:**
1. This issue is also reproduced in the CheckListBox control.
2. This issue is not reproduced in the ComboBox and TreeView controls.
3. **This issue is only reproduced when the control has a scrollbar.**

Amy-Li03:
This issue also reproes for ListView control, no matter it has scrollBar or not.

Tanya-Solyanik:
Seems to be related to https://github.com/dotnet/winforms/commit/383225d62ba125f3dfc0183d2f24dfd03da04a3d

Zheng-Li01:
Verified the issue on 9.0.0-alpha.1.23558.4 with dlls build form main branch, the issue has been fixed that the Accessibility Insights tool working well when detecting listBox/CheckListBox /LisView controls with scrollbar or not.
![10258](https://github.com/dotnet/winforms/assets/38325459/befe6909-8e79-4c74-b749-aadd0a6ea56e)

PR #10284, opened 2023-11-13
Return 0 on UIA_PROPERTY_ID.UIA_NativeWindowHandlePropertyId for Objects with No Handle

lonitra:
Fixes https://github.com/dotnet/winforms/issues/10259
Fixes https://github.com/dotnet/winforms/issues/10258

These AccessibleObjects were added in early .NET Core and did not exist in Framework. Since these objects do not have windows handles, we should be returning `HWND.Null` (0) on `GetPropertyValue(UIA_PROPERTY_ID.UIA_NativeWindowHandlePropertyId)`. We had recently changed to stop doing legacy marshaling for `GetPropertyValue`, which brought this to light. Did manual testing to verify issue no longer repros.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10284)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ListBox.ItemAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ListBox.ItemAccessibleObject.cs
@@ -171,7 +171,7 @@ public partial class ListBox
                  UIA_PROPERTY_ID.UIA_HasKeyboardFocusPropertyId => (VARIANT)(_owningListBox.Focused && _owningListBox.FocusedIndex == CurrentIndex),
                  UIA_PROPERTY_ID.UIA_IsEnabledPropertyId => (VARIANT)_owningListBox.Enabled,
                  UIA_PROPERTY_ID.UIA_IsKeyboardFocusablePropertyId => (VARIANT)State.HasFlag(AccessibleStates.Focusable),
-                 UIA_PROPERTY_ID.UIA_NativeWindowHandlePropertyId => UIAHelper.WindowHandleToVariant(_owningListBox.IsHandleCreated ? _owningListBox.Handle : 0),
+                 UIA_PROPERTY_ID.UIA_NativeWindowHandlePropertyId => UIAHelper.WindowHandleToVariant(HWND.Null),
                  _ => base.GetPropertyValue(propertyID)
              };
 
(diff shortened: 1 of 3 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ListBox.ItemAccessibleObject.cs
```diff
@@ -171,7 +171,7 @@ internal override VARIANT GetPropertyValue(UIA_PROPERTY_ID propertyID)
                  UIA_PROPERTY_ID.UIA_HasKeyboardFocusPropertyId => (VARIANT)(_owningListBox.Focused && _owningListBox.FocusedIndex == CurrentIndex),
                  UIA_PROPERTY_ID.UIA_IsEnabledPropertyId => (VARIANT)_owningListBox.Enabled,
                  UIA_PROPERTY_ID.UIA_IsKeyboardFocusablePropertyId => (VARIANT)State.HasFlag(AccessibleStates.Focusable),
-                 UIA_PROPERTY_ID.UIA_NativeWindowHandlePropertyId => UIAHelper.WindowHandleToVariant(_owningListBox.IsHandleCreated ? _owningListBox.Handle : 0),
+                 UIA_PROPERTY_ID.UIA_NativeWindowHandlePropertyId => UIAHelper.WindowHandleToVariant(HWND.Null),
```
JeremyKuhne, line 174: Do we need to respond to this property at all? That is, does an empty VARIANT work and if so, is that more correct?

Tanya-Solyanik: This is close to the [docs ](https://learn.microsoft.com/en-us/windows/win32/winauto/uiauto-automation-element-propids)-
 _Identifies the NativeWindowHandle property, which is an integer that represents the handle (HWND) of the automation element window, if it exists; otherwise, this property is 0._

Tanya-Solyanik: I did not notice any difference between not handling this property and returning a null in narrator, inspect or AI4Win, but UIA Core usually tries different heuristics when the first access fails, so when we return an actual value, UIACore performs less work

Tanya-Solyanik (approved):

src/System.Windows.Forms/src/System/Windows/Forms/ListViewGroup.ListViewGroupAccessibleObject.cs
```diff
@@ -188,7 +188,7 @@ internal override VARIANT GetPropertyValue(UIA_PROPERTY_ID propertyID)
                 UIA_PROPERTY_ID.UIA_HasKeyboardFocusPropertyId => (VARIANT)(_owningListView.Focused && Focused),
                 UIA_PROPERTY_ID.UIA_IsEnabledPropertyId => (VARIANT)_owningListView.Enabled,
                 UIA_PROPERTY_ID.UIA_IsKeyboardFocusablePropertyId => (VARIANT)State.HasFlag(AccessibleStates.Focusable),
-                UIA_PROPERTY_ID.UIA_NativeWindowHandlePropertyId => UIAHelper.WindowHandleToVariant(_owningListView.IsHandleCreated ? _owningListView.Handle : 0),
+                UIA_PROPERTY_ID.UIA_NativeWindowHandlePropertyId => UIAHelper.WindowHandleToVariant(HWND.Null),
```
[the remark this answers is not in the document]
lonitra: Yes, we would get a hang on when testing `ListView` control with `ListViewGroup` in Accessibility Insights

PR #10290, opened 2023-11-13
Remove a couple of RichEdit interface definitions

JeremyKuhne:
Move ITextDocument and ITextRange to CsWin32 defs and away from built-in COM.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10290)


src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.cs
```diff
@@ -1681,75 +1680,55 @@ public unsafe void ScrollToCaret()
...
-                {
-                    if (iTextDocument != IntPtr.Zero)
-                    {
-                        Marshal.Release(iTextDocument);
-                    }
-                }
-            }
-        }
-        finally
+        if (hr.Succeeded)
```
lonitra, line 1693: Consider inverting this to avoid needing to do `return;` inside the if

lonitra (dismissed):
👍

src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.cs
JeremyKuhne: There are two returns and we always have to fall back to the send message. I didn't want to have a bunch of these- I did add one as you can see above.

lonitra (approved):

PR #10291, opened 2023-11-14
Refactor `WeakRefCollection` to Generics

elachlan:
* Unify `WeakRefCollection`s into single implementation in Common and share between projects.
* Refactor to use Generics

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10291)


elachlan:
CC: @JeremyKuhne @halgab

JeremyKuhne (changes requested):
I started adding comments and realized how messy this is. A much better solution would be to replace it with the following:

``` C#
/// <summary>
///  Weak reference collection of <typeparamref name="T"/>.
/// </summary>
internal class WeakRefCollection<T>() where T : class
{
    private readonly List<WeakReference<T>> _list = new();
    private bool _scavenge;

    public T? this[int index]
    {
        get
        {
            _scavenge = !_list[index].TryGetTarget(out T? target);
            return target;
        }
        set => _list[index] = new(value.OrThrowIfNull());
    }

    public void ScavengeReferences()
    {
        for (int i = Count; i >= 0; i--)
        {
            if (!_list[i].TryGetTarget(out T? target))
            {
                _list.RemoveAt(i);
            }
        }

        _scavenge = false;
    }

    public void RemoveAt(int index) => _list.RemoveAt(index);

    public void Remove(T value)
    {
        int index = IndexOf(value);
        if (index != -1)
        {
            _list.RemoveAt(index);
        }
    }

    public int IndexOf(T value)
    {
        for (int i = 0; i < _list.Count; i++)
        {
            if (!_list[i].TryGetTarget(out T? target))
            {
                _scavenge = true;
            }
            else if (target == value)
            {
                return i;
            }
        }

        return -1;
    }

    public void Add(T value)
    {
        if (_scavenge)
        {
            ScavengeReferences();
        }

        _list.Add(new(value));
    }

    public int Count => _list.Count;
}
```

Depending on the amount of iterators it might be useful to write an enumerator for it as well that already filters out all of the collected values.

The System.Drawing usage can ultimately go away. It is only used to get call backs on the DeviceContext disposals. If someone is feeling up for a challenge they can kill the DeviceContext class in System.Drawing. I did the same for WinForms some time ago. https://github.com/dotnet/winforms/pull/3553 was the last of that change.

elachlan:
> The System.Drawing usage can ultimately go away. It is only used to get call backs on the DeviceContext disposals. If someone is feeling up for a challenge they can kill the DeviceContext class in System.Drawing. I did the same for WinForms some time ago. https://github.com/dotnet/winforms/pull/3553 was the last of that change.

Is the gist that we switch `DeviceContext` with `IDeviceContext`, then use the extension methods on that?

elachlan:
I think the old WeakRefCollection never `ScavengeReferences` unless `RefCheckThreshold` was hit. The `RefCheckThreshold` was never set, so the error case was never hit, because we never cleaned up the collected references.

If I step through the tests slowly in debug, they all pass. If step through fast, some fail.

If I change remove to this they all pass, because `ScavengeReferences` isn't being triggered by the `IndexOf` searching for the value.
```csharp
public void Remove(T value)
{
    for (int i = 0; i < _list.Count; i++)
    {
        if (_list[i].TryGetTarget(out T? target) && target == value)
        {
            _list.RemoveAt(i);
        }
    }
}
```

JeremyKuhne:
> If I step through the tests slowly in debug, they all pass. If step through fast, some fail.

@elachlan all you're getting is this assert, right?

``` C#
        Debug.Assert(t_cellCount <= ToolStripManager.ToolStrips.Count, "who is allocating an extra toolstrippanel cell?");
```

I don't think the assert is correct and should be removed.

Also, you should always iterate backwards [when removing items from a list](https://www.techiedelight.com/remove-elements-from-list-while-iterating-csharp).

src/Common/src/WeakRefCollection.cs
```diff
@@ -0,0 +1,97 @@
...
+    public int Count => _list.Count;
+
+    public IEnumerator<T> GetEnumerator()
+    {
+        foreach (WeakReference<T>? weakRef in _list)
+        {
+            if (weakRef is not null)
+            {
+                // Safely try to get the target. If it's still alive, yield return it.
+                if (weakRef.TryGetTarget(out T? target) && target is not null)
```
halgab, line 90:
```suggestion
                if (weakRef.TryGetTarget(out T? target))
```

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripManager.cs
```diff
@@ -318,21 +318,7 @@ internal static void NotifyMenuModeChange(bool invalidateText, bool activationCh
...
-
-        for (int i = t_toolStripWeakArrayList.Count - 1; i >= 0; i--)
-        {
-            if (t_toolStripWeakArrayList[i] is null)
-            {
-                t_toolStripWeakArrayList.RemoveAt(i);
-            }
-        }
-    }
+    internal static void PruneToolStripList() => t_toolStripWeakArrayList?.ScavengeReferences();
```
JeremyKuhne, line 321: You can probably get rid of the helper method and just do this directly.

src/Common/src/WeakRefCollection.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            ScavengeReferences();
+        }
+
+        _list.Add(new(value));
+    }
+
+    public int Count => _list.Count;
+
+    public IEnumerator<T> GetEnumerator()
```
JeremyKuhne, line 83: If you're adding this you should implement `IEnumerable<T>` and actually use it.

JeremyKuhne (changes requested):
Added a few more comments.

src/Common/src/WeakRefCollection.cs
JeremyKuhne: That is, do `for .. each` on the collection where we're already filtering these out.

elachlan: There might be issues with using the for each, since `ToolStripManager` was triggering `ScavengeReferences` when it found deadrefs whilst looping.

To combat this, I changed the enumerator to set `_scavenge = true;` when it found a deadref, so `ScavengeReferences` is run at the next opportunity.

src/Common/src/WeakRefCollection.cs
```diff
@@ -0,0 +1,99 @@
...
+    {
+        foreach (WeakReference<T> weakRef in _list)
+        {
+            // Safely try to get the target. If it's still alive, yield return it.
+            if (weakRef.TryGetTarget(out T? target))
+            {
+                yield return target;
+                continue;
```
JeremyKuhne, line 91: nit: better to `if ... else` here rather than `continue`

src/System.Windows.Forms/src/System/Windows/Forms/ToolStripManager.cs
```diff
@@ -1481,11 +1406,6 @@ public static bool RevertMerge(ToolStrip targetToolStrip, ToolStrip sourceToolSt
     public static bool RevertMerge(string targetName)
     {
         ToolStrip? target = FindToolStrip(targetName);
-        if (target is null)
-        {
-            return false;
-        }
-
-        return RevertMerge(target);
+        return target is not null && RevertMerge(target);
```
JeremyKuhne, line 1409: nit: `=>`

src/Common/src/WeakRefCollection.cs
JeremyKuhne: It would be good to document where it does that as it isn't obvious looking at the code.

JeremyKuhne (changes requested):
Minor nits

elachlan:
@JeremyKuhne I think I have covered everything here, can you take another quick look?

src/Common/src/WeakRefCollection.cs
```diff
@@ -0,0 +1,107 @@
...
+
+namespace System.Collections.Generic;
+
+/// <summary>
+///  Weak reference collection of <typeparamref name="T"/>.
+/// </summary>
+internal sealed class WeakRefCollection<T>() : IEnumerable<T> where T : class
+{
+    private readonly List<WeakReference<T>> _list = [];
+    /// <summary>
```
JeremyKuhne, line 12: Nit: needs blank line before comment.

src/Common/src/WeakRefCollection.cs
```diff
@@ -0,0 +1,107 @@
...
+        }
+        set
+        {
+            ArgumentNullException.ThrowIfNull(value, nameof(value));
+            _list[index] = new(value);
+        }
+    }
+
+    /// <summary>
+    /// Cleans up dead weak references.
```
JeremyKuhne, line 33: nit: Indent the comment

JeremyKuhne (changes requested):
Just a couple nits. Looks good otherwise!

elachlan:
@JeremyKuhne rebased to fix conflict. No other changes.

JeremyKuhne (approved):

PR #10306, opened 2023-11-14
Issue 5761 normalize runtime id to base

### LINKED ISSUE #5761: Move the implementation of properties "RuntimeId" to the base "ControlAccessibleObject" class for all inheritors
Many accessibility objects for controls have different implementations for `RuntimeId` properties. Although in fact it is one and the same property and its implementation should not have logic specific to which control. Therefore, I suggest to use the implementation of `RuntimeId` properties from the base `ControlAccessibilityObject` class for all accessibility objects. For the base class, I would suggest using one of two implementations:

**Current base implementation:**
```
if (_runtimeId is null)
{
    _runtimeId = new int[] { 0x2a, (int)(long)HandleInternal };
}

return _runtimeId;
```

**Implementation from some accessibility objects:**
```
internal override int[] RuntimeId
{
    get
    {
        var runtimeId = new int[3];
        runtimeId[0] = RuntimeIDFirstItem;
        runtimeId[1] = (int)(long)Owner.InternalHandle;
        runtimeId[2] = Owner.GetHashCode();

        return runtimeId;
    }
}
```

Probably we can even use the current base implementation. Because if the `Handle` is not created, then the external tool will not have access to the control. If the `Handle` is created, then it will be unique for each `Control`.

ricardobossan:
Fixes #5761 

## Proposed changes

- Unifies the identical implementation of the `RuntimeId` from various controls to base.RuntimeId of `ControlAccessibleObject`.

## Customer Impact

- None

## Regression? 

- No

## Risk

- Minimal

### Before

All controls already had `RuntimeId`.

#### ListBox.ItemAccessibleObject.cs:

image

#### UpDownBase.UpDownButtons.UpDownButtonsAccessibleObject.cs:

image

#### LinkLabel.LinkLabelAccessibleObject.cs:

image

#### PropertyGridInternal\PropertyGridView.GridViewTextBox.GridViewTextBoxAccessibleObject.cs

image

#### 

## Test methodology 

- Manual tests
- Inspect

## Test environment(s) 

- 9.0.100-alpha.1.23511.2

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10306)


src/System.Windows.Forms/src/System/Windows/Forms/TreeView.TreeViewAccessibleObject.cs
```diff
@@ -69,23 +68,7 @@ internal override VARIANT GetPropertyValue(UIA_PROPERTY_ID propertyID)
...
-                        RuntimeIDFirstItem,
-                        PARAM.ToInt(owningTreeView.InternalHandle),
-                        owningTreeView.GetHashCode()
-                    };
-                }
-
-                return base.RuntimeId;
-            }
-        }
+        internal override int[] RuntimeId => base.RuntimeId;
```
Tanya-Solyanik, line 71: Since this override returns the base method directly, we don't need it.

src/System.Windows.Forms/src/System/Windows/Forms/Control.ControlAccessibleObject.cs
```diff
@@ -201,7 +201,12 @@ internal override bool GetSysChild(AccessibleNavigation navdir, out AccessibleOb
 
         // This is used only if control supports IAccessibleEx. We need to provide a unique ID. Others are implementing this in the same manner.
         // First item is static - 0x2a (RuntimeIDFirstItem). Second item can be anything, but it's good to supply HWND.
-        internal override int[] RuntimeId => new int[] { RuntimeIDFirstItem, PARAM.ToInt(HandleInternal) };
+        internal override int[] RuntimeId => new int[]
```
Tanya-Solyanik, line 204: BTW, the reason that we are not caching this value in a local field on the base class is that control might be re-create and its handle would change, and we are not updating the ID after the handle change. We only cache it for controls that we don't expect that handles are re-created.

ricardobossan: @Tanya-Solyanik I don't understand very well. Could you please clarify?

- I don't see how caching relates to this issue;
- Also the point of this PR is to move the implementation of `RuntimeId` to the `ControlAccessibleObject` base class that all those controls inherit from, [as suggested on the issue](https://github.com/dotnet/winforms/issues/5761#issue-995825628). However, the code suggestion you made assigns an empty int vector to the `RuntimeId` property in the base class:
image
Should we not move the specific implementations of `RuntimeId` from the controls into the base class [as suggested](https://github.com/dotnet/winforms/issues/5761#issue-995825628) and as I did in my code?

Tanya-Solyanik (approved):

PR #10308, opened 2023-11-15
Enable nullability in remaining PropertyDescriptorGridEntry members

gpetrou:
## Proposed changes

- Enable nullability in remaining PropertyDescriptorGridEntry members.


JeremyKuhne (dismissed):
There are a lot of `!` being added here with no explanation. Please comment as to why they are legit in the code or change things to make sure we're satisfying null analysis.

gpetrou:
> There are a lot of `!` being added here with no explanation. Please comment as to why they are legit in the code or change things to make sure we're satisfying null analysis.

Is there a particular one that you worry about? For example, some of them are in `try/catch` statements. I assume it is OK to use `!` there. We have done that in previous PRs without adding a comment.

JeremyKuhne:
> Is there a particular one that you worry about? For example, some of them are in try/catch statements. I assume it is OK to use ! there. We have done that in previous PRs without adding a comment.

We shouldn't ever do them without a comment. It would also be better to turn these into deliberate throws. I'll add a comment for one of them.

src/System.Windows.Forms/src/System/Windows/Forms/PropertyGridInternal/PropertyDescriptorGridEntry.cs
```diff
@@ -791,22 +800,22 @@ protected bool ViewEvent(object owner, string newHandler, EventDescriptor eventD
...
-        DesignerTransaction transaction = null;
+        IDesignerHost? host = DesignerHost;
+        DesignerTransaction? transaction = null;
 
         try
         {
             // This check can cause exceptions if the event has unreferenced dependencies, which we want to cath.
             // This must be done before the transaction is started or the commit/cancel will also throw.
-            if (eventDescriptor.EventType is null)
+            if (eventDescriptor!.EventType is null)
```
JeremyKuhne, line 810: A better way to do this is:

``` C#
if (eventDiscriptor is null)
{
     // I wouldn't bother adding descriptive text in these cases as this shouldn't be happening in normal circumstances.
     // (Strings have a cost to localize and they bloat the size of the assembly. We can always add helpful strings if feedback warrants.)
     //
     // Additionally there wasn't helpful context in the `NullReferenceException` as it was. In general I'm fine with changing
     // the exception type in these cases. Ideally we would throw `ArgumentException` if we have enough context, but if we don't,
     // `InvalidOperationException` is the right one I believe.
}

if (eventDescriptor.EventType is null)
// ...
```

Fyi: @lonitra

gpetrou:
@JeremyKuhne @lonitra any more feedback on this one?

src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGridInternal/PropertyDescriptorGridEntry.cs
```diff
@@ -471,10 +471,12 @@ public override void OnComponentChanged()
...
         NotifyParentsOfChanges(this);
     }
-#nullable disable
+
     public override bool OnMouseClick(int x, int y, int count, MouseButtons button)
     {
-        if (_propertyValueUIItems is not null && count == 2 && ((button & MouseButtons.Left) == MouseButtons.Left))
+        if (_propertyValueUIItems is not null
+            && _uiItemRects is not null
+            && count == 2 && ((button & MouseButtons.Left) == MouseButtons.Left))
```
lonitra, line 479: nit:
```suggestion
            && count == 2 
            && ((button & MouseButtons.Left) == MouseButtons.Left))
```

src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGridInternal/PropertyDescriptorGridEntry.cs
```diff
@@ -610,12 +621,11 @@ private object SetPropertyValue(object owner, object value, bool reset, string u
...
                     object eventObj = owner;
 
-                    if (PropertyDescriptor is MergePropertyDescriptor && owner is Array)
+                    if (PropertyDescriptor is MergePropertyDescriptor && owner is Array objArray)
                     {
-                        var objArray = owner as Array;
                         if (objArray.Length > 0)
                         {
-                            eventObj = objArray.GetValue(0);
+                            eventObj = objArray.GetValue(0)!;
```
lonitra, line 628: Is it possible `objArray` contains nulls?

gpetrou: I don't see how this can happen.

lonitra: We typically want comments along with usages of `!`. Could you add a comment explaining why it doesn't contain null. Alternatively, If it's not clear why, it looks like even if it does contain null in the array, no exception will be thrown. We could make `eventObj` nullable and describe above line `eventDescriptor = TypeDescriptor.GetEvents(eventObj)[PropertyDescriptor.Name];` that if `eventObj = null` then `TypeDescriptor.GetEvents(eventObj)` returns empty collection while indexing via string will result in `null`

lonitra: Since it does not seem that we would get a NRE if `eventObj = null` before, this change looks like it would change behavior as we would no longer go into the if/else branch in the if statement below. The remarks section of [TypeDescriptor.GetEvents](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.typedescriptor.getevents?view=net-7.0#system-componentmodel-typedescriptor-getevents(system-object)) mentions that if `component` is null, an empty collection will be returned. It doesn't seem that this method is null annotated so we could pass `eventObj!` to this method and note that it will not cause exception and return empty collection in comments.

lonitra (approved):
🚀

Issue #5092, opened 2021-06-14
Painting bug in IconEditor


cadilhac:
* .NET Core Version: 
Any

* Have you experienced this same bug with .NET Framework?: 
Yes

**Problem description:**
The problem happens in System.Drawing.Design.IconEditor.PaintValue.

```
public override void PaintValue(PaintValueEventArgs e)
{
    if (!(e?.Value is Icon icon))
    {
        return;
    }

    // If icon is smaller than rectangle, just center it unscaled in the rectangle
    Rectangle rectangle = e.Bounds;
    if (icon.Width < rectangle.Width)
    {
        rectangle.X = (rectangle.Width - icon.Width) / 2;
        rectangle.Width = icon.Width;
    }

    if (icon.Height < rectangle.Height)
    {
        rectangle.X = (rectangle.Height - icon.Height) / 2;
        rectangle.Height = icon.Height;
    }

    e.Graphics.DrawIcon(icon, rectangle);
}

```
When the icon size is smaller than the provided bounds, the X position is reset as if it was expected to be initially 0. Actually, I guess this code is only called by the MS PropertyGrid and it works well in that case, surely because the provided bounds has X==0.
In my code where I manually ask this editor to paint its value on some bounds inside a window, this makes the drawing appear on the left side of the window instead of in the provided bounds. The solution would be to call rectangle.X += ..... instead of X = ..... This would not break the existing code (for instance the MS PropertyGrid that calls PaintValue) and it would allow custom codes like mine to also behave correctly.

**Expected behavior:**
The painting should occur in the provided bounds.

**Minimal repro:**
Not easy to provide some repro code.

RussKie:
Are you in a position to submit a fix?



> Not easy to provide some repro code.

This makes it hard to triage and fix

cadilhac:
> Are you in a position to submit a fix?

Yes. As I said, replacing "X = " by "X += " fixes it.

weltkante:
looking at the code you copy-pasted it seems the vertical adjustment code is setting `X` instead of `Y` - which is likely to be another bug

> Not easy to provide some repro code.

Why is that so? Can't you just obtain an editor from the TypeDescriptor infrastructure and invoke it to present your scenario, say, in the paint handler of a panel or something?

cadilhac:
> Why is that so? Can't you just obtain an editor from the TypeDescriptor infrastructure and invoke it to present your scenario, say, in the paint handler of a panel or something?

[...]

PR #10309, opened 2023-11-15
fix bug Painting bug in IconEditor

Epica3055:
Fix #[5092](https://github.com/dotnet/winforms/issues/5092)

## Regression? 

-  No

## Risk

- Minimal

### Before

The icon property of the NotifyIcon control cannot be modified

image

### After
The icon property of the NotifyIcon control can be modified

image

## Test methodology 

- Manually (Verified in Scratch project)

## Test environment(s) 

- .net 9.0.0-alpha.1.23510.12

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10131)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Drawing/Design/IconEditor.cs
+++ b/src/System.Windows.Forms.Design/src/System/Drawing/Design/IconEditor.cs
@@ -119,12 +119,14 @@ public class IconEditor : UITypeEditor
         if (icon.Width < rectangle.Width)
         {
             rectangle.X = (rectangle.Width - icon.Width) / 2;
+            rectangle.X += (rectangle.Width - icon.Width) / 2;
             rectangle.Width = icon.Width;
         }
 
         if (icon.Height < rectangle.Height)
         {
             rectangle.X = (rectangle.Height - icon.Height) / 2;
+            rectangle.Y += (rectangle.Height - icon.Height) / 2;
             rectangle.Height = icon.Height;
         }
 
--- a/src/System.Windows.Forms.Design/src/System/Drawing/Design/IconEditor.cs
+++ b/src/System.Windows.Forms.Design/src/System/Drawing/Design/IconEditor.cs
@@ -118,14 +118,12 @@ public class IconEditor : UITypeEditor
         Rectangle rectangle = e.Bounds;
         if (icon.Width < rectangle.Width)
         {
-            rectangle.X = (rectangle.Width - icon.Width) / 2;
             rectangle.X += (rectangle.Width - icon.Width) / 2;
             rectangle.Width = icon.Width;
         }
 
         if (icon.Height < rectangle.Height)
         {
-            rectangle.X = (rectangle.Height - icon.Height) / 2;
             rectangle.Y += (rectangle.Height - icon.Height) / 2;
             rectangle.Height = icon.Height;
         }


src/System.Windows.Forms.Design/src/System/Drawing/Design/IconEditor.cs
```diff
@@ -119,12 +119,14 @@ public override void PaintValue(PaintValueEventArgs e)
         if (icon.Width < rectangle.Width)
         {
             rectangle.X = (rectangle.Width - icon.Width) / 2;
+            rectangle.X += (rectangle.Width - icon.Width) / 2;
```
hughbe, line 122: Surely both these can now be consolidated to `rectangle.X = rectangle.Width - icon.Width;`

Epica3055: I am sorry, what did you mean?

hughbe: We've assigned `X` to be `(rectangle.Width - icon.Width) / 2`, then we add `(rectangle.Width - icon.Width) / 2` to it.

So we can just assign `X` to be `(rectangle.Width - icon.Width)`

Epica3055: Sorry for the confusing code.  I must not have enough coffee the other day.  😴 
``` c#
    public override void PaintValue(PaintValueEventArgs e)
    {
        if (e?.Value is not Icon icon)
        {
            return;
        }
        // If icon is smaller than rectangle, just center it unscaled in the rectangle.
        Rectangle rectangle = e.Bounds;
        if (icon.Width < rectangle.Width)
        {
            rectangle.X = (rectangle.Width - icon.Width) / 2;
            rectangle.Width = icon.Width;
        }

        if (icon.Height < rectangle.Height)
        {
            rectangle.X = (rectangle.Height - icon.Height) / 2;
            rectangle.Height = icon.Height;
        }

        e.Graphics.DrawIcon(icon, rectangle);
    }
```
As we can see, this method is to re-calculate the icon position.

The logic here is not right
``` c#
        if (icon.Width < rectangle.Width)
        {
            rectangle.X = (rectangle.Width - icon.Width) / 2;
            rectangle.Width = icon.Width;
        }
```
As we can see:  `(rectangle.Width - icon.Width) / 2` is the offsetX. I think it is not right to assign offsetX to `rectangle.X`, which I think is the cause of this issue.
To address this error is simple, just `rectangle.X = rectangle.X + (rectangle.Height - icon.Height) / 2`, which is `rectangle.X += (rectangle.Width - icon.Width) / 2;`.
But for some reason, I forgot to remove the old code.

Tanya-Solyanik (commented):
Looks good, but please double check that runtime property grid scenario had not regressed, in the case of icons smaller than 16x16. Also consider adding an Icon property to https://github.com/dotnet/winforms/blob/573d21d5ca575b940186f5a1a29512a28a682185/src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/UserControls/UserControlWithObjectCollectionEditor.cs#L11 , and adding a smaller icon resource to the project for easy experiments

Tanya-Solyanik:
Never mind, I noticed that icon drawing in property grid is blocked by an unexpected transform - https://github.com/dotnet/winforms/pull/10131

Tanya-Solyanik (approved):

PR #10317, opened 2023-11-16
Replace `Gdip` `HashTable` with `Dictionary<object, object>`

elachlan:
Related: #8143

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10317)

### DIFF
--- a/src/System.Drawing.Common/src/System/Drawing/Brushes.cs
+++ b/src/System.Drawing.Common/src/System/Drawing/Brushes.cs
@@ -312,13 +312,13 @@ public static class Brushes
 
     private static Brush GetBrush(object key, Color color)
     {
-        Brush? brush = (Brush?)Gdip.ThreadData[key];
-        if (brush is null)
+        if (Gdip.ThreadData.TryGetValue(key, out object? objectBrush) && objectBrush is Brush brush)
         {
-            brush = new SolidBrush(color);
-            Gdip.ThreadData[key] = brush;
+            return brush;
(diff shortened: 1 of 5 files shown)


elachlan:
@lonitra I think the test failure is unrelated?

lonitra (approved):
LGTM, just a thought - if we find that we are converting to Dictionary<object, object> often as we go through more in System.Drawing.Common, it may be useful to create a helper/extension method that would take care of checking if we get a null/a specific type e.g.
```c#
        public bool TryGetValue<T>(Dictionary<object, object> dictionary, object key, out T? result) where T : class
        {
            dictionary.TryGetValue(key, out object? temp);
            result = temp is T typed ? typed : null;
            return result != null;
        }
```

elachlan:
I think the nullability should be handled by annotations. Its mostly just pattern matching to create a variable to use further down below.

lonitra:
Yes sorry that is was what I meant. This is just something to consider to avoid having to do the following pattern everywhere if do end up having a lot of Dictionary<object, object> - not so much about whether it is null
```c#
if (Dictionary<object, object>.TryGetValue(key, out object? result) && result is SomeType typedResult)
```

elachlan:
Ahh, so the output type ends up being the desired type. Sorry, that is a good suggestion and would reduce the code a tiny bit.

PR #10318, opened 2023-11-16
Always Query for IUnknown when Comparing AgileComPointer for Identity

### LINKED ISSUE #10261: The parent window of a frame should same with the original window of the web browser
### .NET version

.NET 9.0.100-alpha.1.23558.10

Yes

Yes, it cannot repro in .NET /6.0/7.0/8.0.

### Issue description

Automation case about XHtmlWindow met an error "FAIL: equality between windows failed" when testing HtmlWindow's op_Equality, get_Frames methods.
We created a test sample in .NET 9, got following result:
**Actual result:**
Shows "Not expected!" message. The "htmlWindow.Frames[0].Parent.Document.Window" not consistent with "webBrowser1.Document.Window".

https://github.com/dotnet/winforms/assets/94418985/8783ddb8-56ab-4cae-9978-07bb6153309a

**Expected result:**
Shows "Expected!" message. The "htmlWindow.Frames[0].Parent.Document.Window" should be consistent with "webBrowser1.Document.Window".

https://github.com/dotnet/winforms/assets/94418985/4fe74015-b350-4846-99a8-d9ea909d018e

### Steps to reproduce

1. Download and extract project: [XHtmlWindow failed.zip](https://github.com/dotnet/winforms/files/13306587/XHtmlWindow.failed.zip)
2. Build and run
3. Click button1

lonitra:
Fixes #10261

When obtaining a COM pointer to an object that was created on a different thread, we get a proxy so we cannot just directly compare for identity like we had been doing with maintaining `_originalObject` in `AgileComPointer`. If we want to know identity, we need to register the COM pointers in the GIT, query for the IUnknowns on the _same_ thread, and then compare the values.

As a consequence of this, updated `GetHashCode()` on HTMLX. We can no longer get the hashcode the same manner we were doing so before (getting the interface and using the value) because it would change depending on the thread if we had an AgileComPointer created from a proxy. So, we will now get the initial value of the `AgileComPointer` and use that for the hash code to make it constant. 

Added a regression test for this specific scenario that issue outlines, but also added a general test to categorize this a bit more in `AgileComPointerTest` - this test was failing before changes.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10318)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/AgileComPointer.cs
+++ b/src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/AgileComPointer.cs
@@ -28,7 +28,6 @@ internal unsafe class AgileComPointer<TInterface> :
     where TInterface : unmanaged, IComIID
 {
     private uint _cookie;
-    private readonly IUnknown* _originalObject;
 
 #if DEBUG
     public AgileComPointer(TInterface* @interface, bool takeOwnership, bool trackDisposal = true)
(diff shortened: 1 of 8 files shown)


lonitra:
cc: @AaronRobinsonMSFT

JeremyKuhne (approved):
Yay!

PR #10327, opened 2023-11-16
Start putting System.Windows.Forms files in subfolders

JeremyKuhne:
There are well over 1000 files in the root namespace, which makes navigating exceptionally difficult. While we prefer to keep things in folders matching namespaces, we've decided as a team that subfoldering beneath that for clarity is desired.

This is the start of this. I need to finish going through the rest. Not everything will get put in a subfolder.

All renames other than deleting the unreferenced internal Triangle.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10327)


lonitra (approved):
🥳

PR #10334, opened 2023-11-17
Convert Rest of IRawElementProviderX to Cswin32

lonitra:
Converts `IRawElementProviderFragmentRoot` and `IRawElementProviderHwndOverride` to Cswin32. Each conversion is separated into its own commit.  Note that we do not currently return anything from `IRawElementProviderHwndOverride.GetOverrideProviderForHwnd` aside from null so I've added a commit to remove the corresponding virtual method in separate commit.

Tested manually with Narrator and AccessibilityInsights.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10334)


src/System.Windows.Forms/tests/InteropTests/AccessibleObjectTests.cs
```diff
@@ -842,16 +844,6 @@ private class StandardAccessibleObject : AccessibleObject
         [MarshalAs(UnmanagedType.IUnknown)] object pUnk,
         double value, double expected);
 
-    [DllImport(NativeTests, ExactSpelling = true, CharSet = CharSet.Unicode)]
```
Tanya-Solyanik, line 845: We don't have to remove these definitions, even if we don't use them yet. This is the way to expose surface area from the native test. What am I missing?

Tanya-Solyanik (commented):
Looks good. Minor comment about styling changes in formatting ternary conditional, my preference is to indent the 'true' and 'false' forks. And I thought it was the more common pattern in this repo. Let's vote in the meeting?

src/System.Windows.Forms/tests/InteropTests/AccessibleObjectTests.cs
lonitra: My understanding is that we don't need these anymore as these are methods that are exposed through Cswin32 now and that is how we are testing them above.

Tanya-Solyanik: I see, I missed that test above. This test does not validate that marshaling between managed and native code is done correctly because it's 100% managed, the previous test was doing actual marshaling to 
https://github.com/dotnet/winforms/blob/7a6f7349b6c57cc74f346498df279c2078818075/src/System.Windows.Forms/tests/InteropTests/NativeTests/AccessibleObjectTests.cpp#L408 and invoked our managed `ElementProviderFromPoint` implementation from CPP.  I would prefer keeping the old test.

lonitra: Ah I see. I have reverted all the tests including changes I had made in previous PRs. I needed to make some adjustments to `AccessibleObjectTests.cpp` given that we are not tolerating passing in null for some methods now.

src/System.Windows.Forms/tests/InteropTests/NativeTests/AccessibleObjectTests.cpp
```diff
@@ -351,15 +351,13 @@ TEST const WCHAR* WINAPI Test_IRawElementProviderFragmentGetRuntimeId(IUnknown*
...
+        SAFEARRAY *result;
         hr = pRawElementProviderFragment->GetRuntimeId(&result);
-        assertEqualHr(COR_E_NOTSUPPORTED, hr);
-        SafeArrayDestroy(result);
+        assertEqualHr(S_OK, hr);
 
         // Negative tests.
         hr = pRawElementProviderFragment->GetRuntimeId(NULL);
-        assertEqualHr(COR_E_NOTSUPPORTED, hr);
+        assertEqualHr(E_POINTER, hr);
```
lonitra, line 360: @JeremyKuhne it seems that in old code we were returning `HRESULT.COR_E_NOTSUPPORTED` in 2 cases:
1. Passing in a `SAFEARRAY**` where it was already holding some valid value to a `SAFEARRAY*`
2. Passing in null for `SAFEARRAY**`

For 1. I don't see in documentation that explicitly calls out that we should be checking the passed in.. Should we continue this behavior?

For 2. Our pattern is we had been returning `E_POINTER` if null is passed in for a `SOMETYPE**`, I think we should maintain our pattern for returning `E_POINTER` for this case, but what are your thoughts?

src/System.Windows.Forms/tests/InteropTests/AccessibleObjectTests.cs
```diff
@@ -807,6 +799,28 @@ private class StandardAccessibleObject : AccessibleObject
     private static extern string Test_IServiceProviderQueryService(
         [MarshalAs(UnmanagedType.IUnknown)] object pUnk);
 
+    [DllImport(NativeTests, ExactSpelling = true, CharSet = CharSet.Unicode)]
+    private static extern string Test_IRawElementProviderSimpleHostRawElementProvider(
+    [MarshalAs(UnmanagedType.IUnknown)] object pUnk,
```
JeremyKuhne, line 804: nit: lost an indent

src/System.Windows.Forms/tests/InteropTests/NativeTests/AccessibleObjectTests.cpp
JeremyKuhne: `COR_E_NOT_SUPPORTED` is `NotSupportedException`. `E_POINTER` is much more clear.

JeremyKuhne (changes requested):
Just an indent miss

JeremyKuhne (approved):

PR #10347, opened 2023-11-20
Remove unneeded catch

davkean:
This catch is unneeded, its setting a value that never escapes in the case of an exception. This ruins debugging crashes caused by this path and I hit this trying to look at crashes tracked by 1906584 internally.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10347)


Tanya-Solyanik (approved):
Thank you!

PR #10353, opened 2023-11-21

gpetrou:
## Proposed changes

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10353)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataGridView/DataGridView.Methods.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataGridView/DataGridView.Methods.cs
@@ -11232,7 +11232,6 @@ public partial class DataGridView
                         dataGridViewCellNew.OwningRow = dataGridViewRow;
                         dataGridViewCellNew.OwningColumn = dataGridViewColumn;
 
-                        Debug.Assert(KeyboardToolTip is not null);
                         KeyboardToolTipStateMachine.Instance.Hook(dataGridViewCellNew, KeyboardToolTip);
                     }
                 }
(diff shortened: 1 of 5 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/DataGridView/DataGridView.Methods.cs
```diff
@@ -17482,6 +17494,7 @@ internal void OnRemovingColumn(DataGridViewColumn dataGridViewColumn, out Point
             DataGridViewRow dataGridViewRow = Rows.SharedRow(rowIndex);
             if (dataGridViewRow.Cells.Count > newColumnCount)
             {
+                Debug.Assert(KeyboardToolTip is not null);
```
lonitra, line 17497: Taking a look at the implementation `KeyboardToolTip`, I don't see how this could ever be null. Please change its return to be nonnullable then the `Debug.Asserts` for this can be removed

src/System.Windows.Forms/src/System/Windows/Forms/DataGridView/DataGridView.Methods.cs
```diff
@@ -18074,7 +18107,7 @@ private void OnRowEnter(ref DataGridViewCell dataGridViewCell, int columnIndex,
                 && DataConnection.InterestedInRowEvents
                 && !DataConnection.PositionChangingOutsideDataGridView
                 && !DataConnection.ListWasReset
-                && (!calledAddNewOnTheDataConnection || DataConnection.List.Count > 0))
+                && (!calledAddNewOnTheDataConnection || DataConnection.List!.Count > 0))
```
lonitra, line 18110: How do we know `DataConnection.List` is not null here? Please add a comment or consider doing the following to avoid NRE
```c#
(DataConnection.List?.Count ?? 0) > 0
```

lonitra (commented):
Just a few comments, but otherwise looks good! 👍

gpetrou:
@lonitra any more feedback on this one?

lonitra (approved):
Thanks for the ping, changes look good thank you!

PR #10295, opened 2023-11-14
Add ValuePattern.Value changed event to TextBox

SimonZhao888:
Fixes #10257 

## Proposed changes

- Add raise ValuePattern.Value property changed event to TextBoxBase.
- Override Name property of TextBoxBaseAccessibleObject

## Customer Impact

- ValuePattern.Value property changed event will be sent from TextBox control.
- When we don't set the accessible name of the textbox or the previous label, the accessible name of the textbox will return "" not the value of textbox.

## Regression? 

- Yes 

## Risk

- Minimal

### Before
ValuePattern.Value property changed event is not sent from TextBox.
image

### After
When we make changes to the text in the textbox, it will send the ValuePattern.Value property changed event.
TextBox_ValuePattern1

## Test methodology 

- Manual (Insights)

## Test environment(s) 

- 9.0.100-alpha.1.23504.14

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10295)

PR #10369, opened 2023-11-23
[release/8.0] Add ValuePattern.Value changed event to TextBox

SimonZhao888:
Backport of https://github.com/dotnet/winforms/pull/10295 to release/8.0

## Proposed changes

- Add raise ValuePattern.Value property changed event to TextBoxBase.

## Customer Impact

- ValuePattern.Value property changed event will be sent from TextBox control. 
- When we don't set the accessible name of the textbox or the previous label, the accessible name of the textbox will return "" not the value of textbox.

## Regression? 

- Yes

## Risk

- Minimal

### Before

ValuePattern.Value property changed event is not sent from TextBox.
image

### After
When we make changes to the text in the textbox, it will send the ValuePattern.Value property changed event.
TextBox_ValuePattern1

## Test methodology 

- Manual (Insights)

## Test environment(s) 

- 8.0.0-beta.23564.4

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10369)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.TextBoxBaseAccessibleObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TextBoxBase.TextBoxBaseAccessibleObject.cs
@@ -70,9 +70,9 @@ public abstract partial class TextBoxBase
             get
             {
                 var name = base.Name;
-                return name is not null ||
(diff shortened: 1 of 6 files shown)


Tanya-Solyanik (approved):

lonitra:
Merging as February servicing is open

PR #10392, opened 2023-11-29

gpetrou:
## Proposed changes

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10392)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Controls/DataGridView/DataGridView.Methods.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Controls/DataGridView/DataGridView.Methods.cs
@@ -5887,13 +5887,13 @@ public partial class DataGridView
         GC.KeepAlive(this);
     }
 
-    private void EditingControls_CommonMouseEventHandler(object sender, MouseEventArgs e, DataGridViewMouseEvent dgvme)
+    private void EditingControls_CommonMouseEventHandler(object? sender, MouseEventArgs e, DataGridViewMouseEvent dgvme)
     {
         Debug.Assert(_editingPanel is not null);
         Debug.Assert(_ptCurrentCell.X != -1);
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Controls/DataGridView/DataGridView.Methods.cs
```diff
@@ -29400,41 +29417,62 @@ private void UpdateMouseEnteredCell(HitTestInfo hti, MouseEventArgs e)
             &&  oldEdgeRowIndex == newEdgeRowIndex)
         {
             // h1
-            DataGridViewColumn dataGridViewColumn = Columns.GetNextColumn(Columns[oldEdgeColumnIndex], DataGridViewElementStates.Visible, DataGridViewElementStates.None);
+            DataGridViewColumn? dataGridViewColumn = Columns.GetNextColumn(Columns[oldEdgeColumnIndex], DataGridViewElementStates.Visible, DataGridViewElementStates.None);
             Debug.Assert(dataGridViewColumn is not null);
```
lonitra, line 29421: I see that sometimes we are doing this `Debug.Assert` and sometimes checking `dataGridViewColumn` with an if statement to see if it is null. It seems that we are not expecting `dataGridViewColumn` to be null across this method where it is defined, so can you change to throw an `InvalidOperationException` instead? 

If we have a lot of instances where we are expecting non-null `DataGridViewColumn` on calls to `GetNextColumn/GetPreviousColumn` perhaps consider adding an internal helper method on `DataGridViewColumnCollection` like `GetNextColumnRequired/GetPreviousColumnRequired`. Otherwise a small local helper method here would be ok.

lonitra (approved):
🚀

PR #10400, opened 2023-11-30
Small refactor of `ToolStripDesignerUtils` to remove `ArrayList` usage

elachlan:
Related: #8140

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10400)


src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ToolStripDesignerUtils.cs
```diff
@@ -303,10 +303,8 @@ public static Type[] GetCustomItemTypes(IComponent component, ITypeDiscoveryServ
 
                 if (creatableTypes.Count > 0)
```
halgab, line 304: I don't think this if clause is necessary. I may be wrong, but I think the collection expression here will be equivalent to `Array.Empty` in the empty case

PR #10406, opened 2023-11-30
Enable nullability in remaining DataGridViewCell members

gpetrou:
## Proposed changes

- Enable nullability in remaining DataGridViewCell members.


src/System.Windows.Forms/src/System/Windows/Forms/Controls/DataGridView/DataGridViewCellPaintingEventArgs.cs
```diff
@@ -149,20 +155,23 @@ public void PaintContent(Rectangle clipBounds)
...
-                CellBounds,
-                RowIndex,
-                State,
-                Value,
-                FormattedValue,
-                ErrorText,
-                CellStyle,
-                AdvancedBorderStyle,
-                DataGridViewPaintParts.ContentBackground | DataGridViewPaintParts.ContentForeground | DataGridViewPaintParts.ErrorIcon);
+        if (Graphics is not null && CellStyle is not null && AdvancedBorderStyle is not null)
```
gpetrou, line 158: Some tests such as `DataGridViewCellPaintingEventArgs_PaintContent_NullAdvancedBorderStyle_ThrowsArgumentNullException` fail after this change. Should I remove it and use the null-forgiving operator instead?

gpetrou: I used null-forgiving operator and added a comment.

PR #10407, opened 2023-11-30
Convert Multiple AccessibleObject Interfaces to Cswin32

lonitra:

Converts various interfaces on AccessibleObject to Cswin32 and updates correspondings tests. The conversions are separated by commits.

Also introduce `IComInterface<T>` to bound COM structs to their COM interfaces. Currently this is used for the purposes of new method `ComSafeArrayScope.CreateFromInterfaceArray` in order to generically go from COM interface [] -> SafeArray of COM pointers, but may be useful for other purposes as we continue making WinForms trimmable. Tests are also added around this.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10407)


Tanya-Solyanik (approved):
Looks good

PR #10432, opened 2023-12-05
Replace ArrayList with List in several designers

halgab:
Contributes to #8140

## Proposed changes

- Replace `ArrayList` with `List<IComponent>` in `ControlDesigner`
- Replace `ArrayList` with `List<Control>` in `FlowLayoutPanelDesigner`
- ~~Replace `ArrayList` with `List<SnapLine>` in `FlowPanelDesigner`~~ see https://github.com/dotnet/winforms/pull/10432#issuecomment-1841717995
- Replace `ArrayList` with `List<IComponent>` in `TabControlDesigner`

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10432)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/FlowLayoutPanelDesigner .cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/FlowLayoutPanelDesigner .cs
@@ -637,7 +637,7 @@ internal partial class FlowLayoutPanelDesigner : FlowPanelDesigner
             else
             {
                 // We are doing a copy, so let's copy the controls.
-                ArrayList tempList = new ArrayList();
+                List<Control> tempList = new();
                 tempList.AddRange(_dragControls);
 
                 DesignerUtils.CopyDragObjects(tempList, Component.Site);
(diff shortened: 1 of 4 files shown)


src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/FlowLayoutPanelDesigner .cs
```diff
@@ -637,7 +637,7 @@ private void ReorderControls(DragEventArgs de)
             else
             {
                 // We are doing a copy, so let's copy the controls.
-                ArrayList tempList = new ArrayList();
+                List<Control> tempList = new();
                 tempList.AddRange(_dragControls);
 
                 DesignerUtils.CopyDragObjects(tempList, Component.Site);
```
halgab, line 643: Shouldn't this be 
```suggestion
                tempList = DesignerUtils.CopyDragObjects(tempList, Component.Site);
```
?

elachlan:
Might be overlap with #10279

halgab:
@lonitra not sure what the CLA check has been up to for a week here...

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/FlowLayoutPanelDesigner .cs
lonitra: Hmm.. This does look a bit suspicious, but I am concerned about whether this was intentional and what effects this would have if we do make this change. It's not immediately clear from looking here. Let's open a separate issue on this to investigate.

lonitra (approved):
👍

PR #10440, opened 2023-12-07
Fix `ErrorProvider.HasErrors` so maintains correct value

### LINKED ISSUE #10424: ErrorProvider HasErrors returns incorrect value after clearing an error
### .NET version

7.0

Not tested/verified

_No response_

### Issue description

The implementation of #5912 does not take into account clearing an individual error. The documentation for ErrorProvider.SetError states for the value "The error description string, or null or Empty to remove the error."

So HasErrors returns an incorrect value if SetError has ever been used, even if there is now no error.

### Steps to reproduce

Basic application logic is
  SetError(control, "Some error");
  _HasErrors should be true at this point_
  SetError(control, null);
  _HasErrors should be **false** at this point_
  
Given that GetError also creates an entry in _items with a null error the following will also cause incorrect results:
  _ = GetError(control);
  _HasErrors should be **false** at this point_

elachlan:
Fixes #10424

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10440)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ErrorProvider/ErrorProvider.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ErrorProvider/ErrorProvider.cs
@@ -54,6 +54,7 @@ public partial class ErrorProvider : Component, IExtenderProvider, ISupportIniti
     private EventHandler? _onRightToLeftChanged;
 
     private bool _rightToLeft;
+    private int _errorCount;
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


eriawan (approved):
LGTM all. Thanks for fixing this! ✔️ 👍

lonitra (approved):
🚀

PR #10441, opened 2023-12-07
Add coding style documentation

JeremyKuhne:
Add a more detailed document on our team coding style.

To demonstrate code to the style standard I've updated the code in the Buttons subfolder. Also updated a few files from Primitives. Informational messages in the updated files are now greatly reduced.

Some of the things done:

- Fix spelling
- Use ternary operators
- Use target-typed new
- Fix some internal parameter names
- Use expression bodies
- Use pattern matching
- Add argument names for bools and nulls
- Remove regions

Thread static naming and shipped public API naming issues aren't easily addressable without suppressions. Added a global suppression file to suppress these particular issues to make the code less difficult to read. We shouldn't put other types of suppressions in this file.

Note that for some reason RadioButton.PerformAutoUpdates was getting the TypeDescriptor to clear sibling RadioButtons. I've set the property directly as this is significantly more performant and friendly to trimming.

PaddingConverter was unnecessarily complicated in creating it's strings. `int` to string never is impacted by culture.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10441)

### DIFF
--- a/docs/coding-style.md
+++ b/docs/coding-style.md
@@ -4,20 +4,20 @@ Our coding style is informed by the [.NET Runtime coding style](https://github.c
 
 When we establish guidelines we consider how they impact the ability to comprehend the code. The easier the code is to understand, the easier it is to maintain. There are a few core guiding principles:
 
-- Code should be succinct (and as such we lean into new C# language features)
-- Code should be clear and unambiguous
-- Guidelines are mutable where they inhibit our guiding principles
+- Code should be clear and unambiguous.
+- Code should be efficient and concise, using the minimum amount of code necessary to achieve the desired functionality without sacrificing readability or maintainability (and as such we lean into new C# language features).
+- Guidelines are mutable where they inhibit our guiding principles.
 
 There are a few key things we strive for to attain our clarity goals:
 
-- Minimal code block nesting
-- Minimal scrolling for code blocks (almost no horizontal and limited vertical)
-- Minimal code visibility (via access modifiers, type nesting, and local functions)
-- All context captured in source (via code or comments)
+- Minimal code block nesting.
+- Minimal scrolling for code blocks (almost no horizontal and limited vertical).
+- Minimal code visibility (via access modifiers, type nesting, and local functions).
+- All context captured in source (via code or comments).
 
-Windows Forms dates back to the mid 1990s (originally Windows Foundation Classes). It has seen over 25 years of language and coding evolution. We believe that updating the code to modern style makes the code easier to maintain and update in the longer term. While we do not update everything as the style guidelines evolve, we do make continual investments in cleaning the codebase as an engineering best practice.
+Windows Forms dates back to the mid 1990s (originally [Windows Foundation Classes](https://web.archive.org/web/20050915003238/http://msdn.microsoft.com/library/en-us/vjcore98/html/vjconintroductiontowfcprogramming.asp)). It has seen over 25 years of language and coding evolution. We believe that updating the code to modern style makes the code easier to maintain and update in the longer term. While we do not update everything as the style guidelines evolve, we do make continual investments in cleaning the codebase as an engineering best practice.
 
-A number of the following detailed guidelines are captured in our .editorconfig and as such you'll get the appropriate feedback in Visual Studio. We're iterating to enable more automated analyzers in the future.
+A number of the following detailed guidelines are captured in our `.editorconfig` and as such you'll get the appropriate feedback in Visual Studio. We're iterating to enable more automated analyzers in the future.
 
 ### Spacing
 
@@ -36,6 +36,42 @@ Enabling viewing white space will make conforming to these rules easier. In Visu
 1. When breaking logical statements, all logical operators (`&&`, `||`, etc.) at a given parenthetical scope should be indented on individual lines at one indent in.
 1. When breaking ternary operators (`?:`), they should be broken into two singly indented lines, starting with `?` and `:`.
 
+<details>
+<summary>Code Samples</summary>
+
+More complicated line breaks:
+
+``` C#
+if (_trackColumn != previousColumnIndex
+    && !(previousColumnIndex == -1 && hti._col == _trackColumn)
+    && (dataGridViewColumnNext is null || _trackColumn != dataGridViewColumnNext.Index))
+```
+
+``` C#
+if (Focused
+    && !IsCurrentCellInEditMode
+    && (EditMode == DataGridViewEditMode.EditOnEnter
+        || (EditMode != DataGridViewEditMode.EditProgrammatically && CurrentCellInternal.EditType is null)))
+```
+
+Why `=>` isn't at the beginning of a line:
+
+``` C#
+// Further breaks logically align
+internal bool SingleVerticalBorderAdded =>
+    !_layout.RowHeadersVisible
+    && (AdvancedCellBorderStyle.All == DataGridViewAdvancedCellBorderStyle.Single
+        || CellBorderStyle == DataGridViewCellBorderStyle.SingleVertical);
+
+// Same example, broken alignment
+internal bool SingleVerticalBorderAdded
+    => !_layout.RowHeadersVisible
+    && (AdvancedCellBorderStyle.All == DataGridViewAdvancedCellBorderStyle.Single
+        || CellBorderStyle == DataGridViewCellBorderStyle.SingleVertical);
+
+```
+</details>
+
 ### Code Blocks
 
 1. Use [Allman style](http://en.wikipedia.org/wiki/Indent_style#Allman_style) braces, where each brace begins on a new line.
@@ -48,8 +84,10 @@ Enabling viewing white space will make conforming to these rules easier. In Visu
 
 1. Code comments should be on their own line and precede the code they refer to. In condition blocks, comments should be inside the blocks (e.g., never before the `else`). Switch expression conditions have no blocks and should have the comment before the relevant condition line.
 1. Avoid comments labeling the end of a block (never `if {} else {} // else`).
-1. Methods and properties should have XML comments. `public` API comments should align with the documentation on https://learn.microsoft.com.
+1. Methods and properties should have XML comments, not `\\` comments. `public` API comments should align with the documentation on https://learn.microsoft.com.
+1. XML comment blocks that provide no further clarification are not required, with the exception of public API. (You don't have to document the return value if it is obvious, for example.)
 1. Do not replicate comments from base methods or interfaces. Use `\\\ <inheritdoc cref='interfacemethod'\>` when docs aren't automatically inherited or you need to modify only a part of the documentation.
+1. Avoid `\* *\` comments.
 
 ### General Naming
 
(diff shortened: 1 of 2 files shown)


docs/coding-style.md
```diff
@@ -0,0 +1,129 @@
...
+
+- Minimal code block nesting
+- Minimal scrolling for code blocks (almost no horizontal and limited vertical)
+- Minimal code visibility (via access modifiers, type nesting, and local functions)
+- All context captured in source (via code or comments)
+
+Windows Forms dates back to the mid 1990s (originally Windows Foundation Classes). It has seen over 25 years of language and coding evolution. We believe that updating the code to modern style makes the code easier to maintain and update in the longer term. While we do not update everything as the style guidelines evolve, we do make continual investments in cleaning the codebase as an engineering best practice.
+
+A number of the following detailed guidelines are captured in our .editorconfig and as such you'll get the appropriate feedback in Visual Studio. We're iterating to enable more automated analyzers in the future.
+
```
KlausLoeffelmann, line 21: **Note:** The following sections are for C# only. [Visual Basic's _Pretty Lister_](https://learn.microsoft.com/en-us/dotnet/visual-basic/programming-guide/program-structure/coding-conventions#layout-conventions) is turned on in most projects and never switched of by a development team which means indenting and formatting is automatically done. We will address VB's special requirements in upcoming updates.

docs/coding-style.md
```diff
@@ -0,0 +1,129 @@
...
+1. Avoid more than one empty line at any time. For example, do not have two blank lines between members of a type.
+1. Closing brackets (`}`) on a line by themself should get a blank line after them unless followed by another line with just a closing bracket.
+
+### Line Breaks
+
+1. Lines should be less than 120 characters long to limit horizontal scrolling in development tools. If it improves clarity, they can be slightly over, but they never should be longer than 150. Statements should not be broken into multiple lines when they fall under the 120 character limit.
+1. When using expression body definitions (`=>`) they should be on the same line if they fit, otherwise they should be broken after the `=>` and indented once on the next line.
+1. When breaking arguments to a method, all arguments should be indented on individual lines at one indent in.
+1. When breaking logical statements, all logical operators (`&&`, `||`, etc.) at a given parenthetical scope should be indented on individual lines at one indent in.
+1. When breaking ternary operators (`?:`), they should be broken into two singly indented lines, starting with `?` and `:`.
```
KlausLoeffelmann, line 37: Can we break those always for better readability, even if the complete line wouldn't reach the char limit?

docs/coding-style.md
```diff
@@ -0,0 +1,129 @@
...
+
+1. Use four spaces of indentation (never use tabs). XML blocks should get a single space indent (this includes XML comments).
+1. Lines should not have trailing white space or more than one space between code elements (`=` can be aligned when there is notable value in doing so, such as bit flag values).
+1. Avoid more than one empty line at any time. For example, do not have two blank lines between members of a type.
+1. Closing brackets (`}`) on a line by themself should get a blank line after them unless followed by another line with just a closing bracket.
+
+### Line Breaks
+
+1. Lines should be less than 120 characters long to limit horizontal scrolling in development tools. If it improves clarity, they can be slightly over, but they never should be longer than 150. Statements should not be broken into multiple lines when they fall under the 120 character limit.
+1. When using expression body definitions (`=>`) they should be on the same line if they fit, otherwise they should be broken after the `=>` and indented once on the next line.
```
KlausLoeffelmann, line 34: Why is this in the same line, while the Boolean operators start a line? Isn't that sort of inconsistent?

docs/coding-style.md
```diff
@@ -0,0 +1,129 @@
...
+1. `fixed` statements must use code blocks. When there are multiple `using` statements in a row, they should share a code block to reduce nesting.
+1. `unsafe` should generally be applied to either whole methods or classes (when many methods are unsafe). When unsafe blocks are strictly necessary (say in an async or yield method) they must use code blocks.
+
+### Comments
+
+1. Code comments should be on their own line and precede the code they refer to. In condition blocks, comments should be inside the blocks (e.g., never before the `else`). Switch expression conditions have no blocks and should have the comment before the relevant condition line.
+1. Avoid comments labeling the end of a block (never `if {} else {} // else`).
+1. Methods and properties should have XML comments. `public` API comments should align with the documentation on https://learn.microsoft.com.
```
KlausLoeffelmann, line 51: Should private methods/properties also have XML comments?

docs/coding-style.md
```diff
@@ -0,0 +1,129 @@
...
+1. Do not replicate comments from base methods or interfaces. Use `\\\ <inheritdoc cref='interfacemethod'\>` when docs aren't automatically inherited or you need to modify only a part of the documentation.
+
+### General Naming
+
+1. Avoid abbreviations and type names for variable or property names (e.g. `Point firstPoint` not `pt1`).
```
KlausLoeffelmann, line 56: Should we have an exception list?
(`jtf`, `e`, `g`, `i`, `T`, `hwnd`, `dc`, etc?)

docs/coding-style.md
```diff
@@ -0,0 +1,129 @@
...
+1. Always specify the visibility, even if it's the default (e.g. `private string _foo` not `string _foo`). Visibility should be the first modifier (e.g. `public abstract` not `abstract public`).
+
+### Namespace Usings
+
+1. Namespace `using`s should be specified at the top of the file, before `namespace` declarations, and should be sorted alphabetically, with the exception of `System.*` namespaces, which are to be placed on top of all others.
+1. Global usings go should in a file called `globalusings.cs`.
+
+### Namespaces
+
+1. Use file-scoped namespaces.
```
KlausLoeffelmann, line 71:
```suggestion
1. Use [file-scoped namespaces](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/namespace).
```

docs/coding-style.md
```diff
@@ -0,0 +1,129 @@
...
+1. All types should be in their own files. This includes nested types.
+1. All internal and private classes should be sealed when they are not derived from.
+1. Use PascalCasing to name all types. The only exception is for interop types, which should match the native casing.
+1. Consider [primary constructors](https://learn.microsoft.com/dotnet/csharp/programming-guide/classes-and-structs/instance-constructors#primary-constructors) for "simple" structs and classes where there is only one constructor with no logic other than field assignment.
+
+### Fields
+
+1. Type constants, statics, and fields should be specified at the top within the type declaration.
+1. Use `_camelCase` for internal and private fields and use `readonly` where possible. Prefix instance fields with `_`, static fields with `s_` and thread static fields with `t_`.
+1. Use PascalCasing to name all constants. The only exception is for interop code where the constant value should match the naming in the native code.
```
KlausLoeffelmann, line 85:
```suggestion
1. Use `PascalCasing` to name all constants. The only exception is for interop code where the constant value should match the naming in the native code.
```

docs/coding-style.md
```diff
@@ -0,0 +1,129 @@
...
+1. `public` or `internal` fields should not be used (use properties).
```
KlausLoeffelmann, line 86: What about protected fields as a performance exception?

docs/coding-style.md
```diff
@@ -0,0 +1,129 @@
...
+1. Use [expression body definitions](https://learn.microsoft.com/dotnet/csharp/programming-guide/statements-expressions-operators/expression-bodied-members) (`=>`) for single line methods.
+1. Avoid methods with a single calling method. Use [local functions](https://learn.microsoft.com/dotnet/csharp/programming-guide/classes-and-structs/local-functions). Don't use a method at all in these cases if the logic is simple and not replicated.
+1. Use static methods when instance fields are not accessed, and consider passing fields in to make methods static to clarify what a method modifies.
+1. Prefer methods where top level blocks fit in a single editor screen (around 25 lines of code). Local functions can be used to break the logic into digestible chunks.
+1. Keep block nesting to a minimum (preferably no more than 3 indents, 5 at the maximum). Use early outs and inverted conditionals to help manage nesting.
+1. Exit (return) early from methods to reduce nesting.
+
+### Variables
+
+1. We never use `var` for intrinsic or base types, e.g. `var count = 2`. `var` should only be used when the type name inhibits code readability.
```
KlausLoeffelmann, line 108:
```suggestion
1. We never use `var` for intrinsic/primitive/predefined types, e.g. don't use `var count = 2`, instead use `int count = 2`. `var` should only be used when the type name inhibits code readability.
```

docs/coding-style.md
```diff
@@ -0,0 +1,129 @@
...
+1. [Target-typed `new()`](https://learn.microsoft.com/dotnet/csharp/language-reference/operators/new-operator#target-typed-new) should be used to remove redundancy, not var. For example: `FileStream stream = new(...);`, not `var stream = new FileStream(...);`. Using target-typed `new()` for succinctness is allowed where the type is reasonably obvious, e.g. `Point[] points = [new(1, 2), new(5, 6)];`
+1. [Collection expressions](https://learn.microsoft.com/dotnet/csharp/language-reference/operators/collection-expressions) are preferred, e.g. `Point[] points = [new(1, 2)];` over `Point[] points = new Point[] { new(1, 2) }`;
+
+### General
+
+1. Use language keywords instead of BCL types (e.g. `int, string, float` instead of `Int32, String, Single`, etc.) for both type references as well as method calls (e.g. `int.Parse` instead of `Int32.Parse`).
```
KlausLoeffelmann, line 114: Note that in contrast to C#, `Date` is a primitive type in VB. Remember the differences, when working on PRs where both C# and VB are involved (so `DateTime` in C# but `Date` in VB or `string` in C# over `String`).

KlausLoeffelmann (approved):
I am wondering - would it make more sense to separate the doc update from the .editorConfig update (with the related code changes)?

Looks good to me otherwise, I had a few suggestions for the docs.

docs/coding-style.md
JeremyKuhne: I find that worse as I end up with random line breaks depending on the size of my editor pane.

docs/coding-style.md
JeremyKuhne: Because further formatting (e.g. line breaks) don't line up. I'm adding a code samples block (as discussed) to further clarify.

``` C#
// Further breaks logically align
internal bool SingleVerticalBorderAdded =>
    !_layout.RowHeadersVisible
    && (AdvancedCellBorderStyle.All == DataGridViewAdvancedCellBorderStyle.Single
        || CellBorderStyle == DataGridViewCellBorderStyle.SingleVertical);

// Same example, broken alignment
internal bool SingleVerticalBorderAdded
    => !_layout.RowHeadersVisible
    && (AdvancedCellBorderStyle.All == DataGridViewAdvancedCellBorderStyle.Single
        || CellBorderStyle == DataGridViewCellBorderStyle.SingleVertical);

```

docs/coding-style.md
JeremyKuhne: You almost never really have a perf situation where this would show up. I did add a bullet for `protected` fields.

docs/coding-style.md
JeremyKuhne: I clarified this to "built-in types" and added a doc link.

KlausLoeffelmann (approved):
Cool, I like it!

KlausLoeffelmann:
> This rule does sound a bit inconsistent with the principle of having

Honestly, I would like this better, too.

src/System.Windows.Forms/src/GlobalSuppressions.cs
```diff
@@ -0,0 +1,25 @@
...
+[assembly: SuppressMessage("Naming", "CA1725:Parameter names should match base declaration", Justification = "Public API", Scope = "member", Target = "~M:System.Windows.Forms.ButtonBase.OnMouseMove(System.Windows.Forms.MouseEventArgs)")]
+[assembly: SuppressMessage("Naming", "CA1725:Parameter names should match base declaration", Justification = "Public API", Scope = "member", Target = "~M:System.Windows.Forms.ButtonBase.OnMouseUp(System.Windows.Forms.MouseEventArgs)")]
+[assembly: SuppressMessage("Naming", "CA1725:Parameter names should match base declaration", Justification = "Public API", Scope = "member", Target = "~M:System.Windows.Forms.ButtonBase.OnPaint(System.Windows.Forms.PaintEventArgs)")]
+[assembly: SuppressMessage("Naming", "CA1725:Parameter names should match base declaration", Justification = "Public API", Scope = "member", Target = "~M:System.Windows.Forms.Button.OnMouseEnter(System.EventArgs)")]
+[assembly: SuppressMessage("Naming", "CA1725:Parameter names should match base declaration", Justification = "Public API", Scope = "member", Target = "~M:System.Windows.Forms.Button.OnMouseLeave(System.EventArgs)")]
+
+// Thread locals
```
Tanya-Solyanik, line 16: Consider repeating the comment form the PR description that there is not editorconfig option to special-case thread locals in here.

JeremyKuhne:
> This rule does sound a bit inconsistent with the principle of having the important operators on the left. It will also result in code churn especially in the Designer code. Can we make it optional?
This could be aligned like this:

``` C#
internal bool SingleVerticalBorderAdded
    => !_layout.RowHeadersVisible
        && (AdvancedCellBorderStyle.All == DataGridViewAdvancedCellBorderStyle.Single
        || CellBorderStyle == DataGridViewCellBorderStyle.SingleVertical);
```

@Tanya-Solyanik 

The `!` doesn't align with the operators that are at the same level of grouping. The `||` is contained with the `()`, which makes it a sub clause of the `&&` and that is lost.

Full indenting would be something like this:

``` C#
internal bool SingleVerticalBorderAdded =>
    !_layout.RowHeadersVisible
    &&
    (
         AdvancedCellBorderStyle.All == DataGridViewAdvancedCellBorderStyle.Single
         || CellBorderStyle == DataGridViewCellBorderStyle.SingleVertical
    );
```

In practice that is a bit unwieldly. Collapsing the parenthesis "blocks", while not perfect, is a reasonable balance I've found.

``` C#
internal bool SingleVerticalBorderAdded =>
    !_layout.RowHeadersVisible
    && (AdvancedCellBorderStyle.All == DataGridViewAdvancedCellBorderStyle.Single
        || CellBorderStyle == DataGridViewCellBorderStyle.SingleVertical);
```

As to making the `=>` placement optional I believe that it doesn't provide any real value in readability over being consistent.

We can always revisit with enough evidence to the contrary.

JeremyKuhne:
The coding style doc is a living document. We'll continue to add to it and evolve the rules based on real-world experience.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Padding.cs
```diff
@@ -11,13 +11,15 @@ namespace System.Windows.Forms;
 [Runtime.CompilerServices.TypeForwardedFrom("System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089")]
 public struct Padding : IEquatable<Padding>
 {
-    private bool _all; // Do NOT rename (binary serialization).
-    private int _top; // Do NOT rename (binary serialization).
-    private int _left; // Do NOT rename (binary serialization).
-    private int _right; // Do NOT rename (binary serialization).
-    private int _bottom; // Do NOT rename (binary serialization).
+    private bool _all;      // Do NOT rename (binary serialization).
```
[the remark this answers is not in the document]
Tanya-Solyanik: >Which you would be responsible for fixing up if you run it. 

This looks like an unnecessary complexity. Are we really achieving any readability objectives by adding these spaces?

KlausLoeffelmann:
Still with Tanya about this one. 
For me, this is WAY better readable than any other version, although not perfectly aligned. I can't even tell way - coming from typesetting computer books, all alarm bells should go off seeing this, but curiously enough, it is just _really_ nice to read. I have no clue, why.

image

KlausLoeffelmann:
Said all that, it wouldn't be important enough, to have a schnapps-duel over this with you.
(Although, on second thought...)

Tanya-Solyanik:
Yeah, the example with misalignment is not what I had in mind when I suggested making this rule optional. I was thinking about single line methods
```
private int SomeLongName()
    => ASingleLineButLongMethodCall();
```

PR #10447, opened 2023-12-07
Update ScaleHelper file name

JeremyKuhne:
Update the file names to match and pull two methods back into the main file.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10447)


lonitra (approved):
Does ScaleHelper still need to be a partial class after this change?

PR #10459, opened 2023-12-11
Update the logic of the function GetIconResourceAsBitmap in class Sca…

### LINKED ISSUE #10444: The smart tag icon in the DemoConsole application becomes very large
### .NET version

Yes

Yes, this is a regression issue.

### Issue description

In [DemoConsole](https://github.com/dotnet/winforms/tree/main/src/System.Windows.Forms/tests/IntegrationTests/DesignSurface/DemoConsole) test app of Winforms repo, Smart tag icons getting very large.

image

https://github.com/dotnet/winforms/assets/86937911/59da3f2e-a1df-4d4b-bb9d-39cb5ac38c6d

### Steps to reproduce

LeafShi1:
Fixes #10444

## Cause of this issue:
- Since the icon` Close_left` and` Open_left` are not standard logical sizes (16X16), after calling `GetIconResourceAsBitmap(Type type, string resource, int dpi)`, thay are scaled from the standard logical size 16x16 to 200% size - 32X32

## Proposed changes

- Specify custom icon size in code for the non-standard icons (glyphs in the designer)
- Using  `GetIconResourceAsBitmap(Type type, string resource, Size size)` instead of "GetIconResourceAsBitmap(Type type, string resource, int dpi)" when initial Image "GlyphImageClosed" and "GlyphImageOpened"

## Customer Impact

- The smart tag icon can be show normally

## Regression? 

- Yes (From PR#[10434](https://github.com/dotnet/winforms/pull/10434))

## Risk

- Minimal

### Before
Smart tag icons getting very large

image

### After
The size of the Smart tag icon is normal

image

## Test methodology 

- Manually

## Test environment(s) 

- .net 9.0.0-alpha.1.23609.5

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10459)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/Behavior/DesignerActionGlyph.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/Behavior/DesignerActionGlyph.cs
@@ -15,9 +15,8 @@ internal sealed class DesignerActionGlyph : Glyph
 {
     internal const int CONTROLOVERLAP_X = 5;                    // number of pixels the anchor should be offset to the left of the control's upper-right
     internal const int CONTROLOVERLAP_Y = 2;                    // number of pixels the anchor overlaps the control in the y-direction
-    private const byte IconsWidth = 10;
-    private const byte IconsHeight = 10;
 
+    private const byte IconSize = 10;
     private Rectangle _alternativeBounds = Rectangle.Empty;     // if !empty, this represents the bounds of the tray control this glyph is related to
     private Rectangle _bounds;                                  // the bounds of our glyph
     private readonly Adorner? _adorner;                         // A ptr back to our adorner - so when we decide to change state, we can invalidate
@@ -27,6 +26,8 @@ internal sealed class DesignerActionGlyph : Glyph
     private DockStyle _dockStyle;
     private Bitmap? _glyphImageClosed;
     private Bitmap? _glyphImageOpened;
+    private readonly byte _iconsWidth = IconSize;
+    private readonly byte _iconsHeight = IconSize;
 
     /// <summary>
     ///  Constructor that passes empty alternative bounds and parents.
@@ -49,6 +50,7 @@ internal sealed class DesignerActionGlyph : Glyph
     private DesignerActionGlyph(DesignerActionBehavior? behavior, Adorner? adorner, Rectangle alternativeBounds, Control? alternativeParent)
         : base(behavior)
     {
+        _iconsWidth = _iconsHeight = (byte)ScaleHelper.ScaleToInitialSystemDpi(IconSize);
         _adorner = adorner;
         _alternativeBounds = alternativeBounds;
         _alternativeParent = alternativeParent;
@@ -100,12 +102,12 @@ internal sealed class DesignerActionGlyph : Glyph
     private Image GlyphImageClosed => _glyphImageClosed ??= ScaleHelper.GetIconResourceAsBitmap(
         typeof(DesignerActionGlyph),
         "Close_left",
-        new Size(IconsWidth, IconsHeight));
+        new Size(_iconsWidth, _iconsHeight));
 
     private Image GlyphImageOpened => _glyphImageOpened ??= ScaleHelper.GetIconResourceAsBitmap(
         typeof(DesignerActionGlyph),
         "Open_left",
-        new Size(IconsWidth, IconsHeight));
+        new Size(_iconsWidth, _iconsHeight));
 
     internal void InvalidateOwnerLocation()
     {
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/Behavior/DesignerActionGlyph.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/Behavior/DesignerActionGlyph.cs
@@ -15,6 +15,8 @@ internal sealed class DesignerActionGlyph : Glyph
 {
     internal const int CONTROLOVERLAP_X = 5;                    // number of pixels the anchor should be offset to the left of the control's upper-right
     internal const int CONTROLOVERLAP_Y = 2;                    // number of pixels the anchor overlaps the control in the y-direction
+    private const byte IconsWidth = 10;
+    private const byte IconsHeight = 10;
 
     private Rectangle _alternativeBounds = Rectangle.Empty;     // if !empty, this represents the bounds of the tray control this glyph is related to
     private Rectangle _bounds;                                  // the bounds of our glyph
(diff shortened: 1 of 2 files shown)


LeafShi1:
I tried use the function `GetIconResourceAsBitmap(Type type, string resource, int dpi)` in other control, the icon is also displayed incorrectly. 
For example: Replace [ ScaleHelper.GetIconResourceAsBitmap(typeof(DataGridViewHeaderCell), iconName, new Size(s_iconsWidth, s_iconsHeight));](https://github.com/dotnet/winforms/blob/001b753d61752d9366c50e73e150b323ad12b095/src/System.Windows.Forms/src/System/Windows/Forms/Controls/DataGridView/DataGridViewRowHeaderCell.cs#L87C114-L87C114) with `ScaleHelper.GetIconResourceAsBitmap(typeof(DataGridViewHeaderCell), iconName, ScaleHelper.InitialSystemDpi)`, then the Datagridview show as below
image

Excepted value:
image

Tanya-Solyanik:
@LeafShi1  - thank you for the investigation! This function is used in too many places, we should restore the old logic instead of scaling all bitmaps to the initial system DPI. In some cases this will be incorrect. @JeremyKuhne is actively working in this area, it's best to let him fix this issue.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/ScaleHelper.cs
```diff
@@ -332,7 +332,7 @@ internal static Bitmap GetIconResourceAsDefaultSizeBitmap(Type type, string reso
     ///  Gets the given icon resource as a <see cref="Bitmap"/> scaled to the specified dpi.
     /// </summary>
     internal static Bitmap GetIconResourceAsBitmap(Type type, string resource, int dpi)
-        => GetIconResourceAsBitmap(type, resource, ScaleToDpi(SystemIconSize, dpi));
+        => GetIconResourceAsBestMatchBitmap(type, resource, ScaleToDpi(SystemIconSize, dpi));
```
Tanya-Solyanik, line 335: In some cases `SystemIconSize` is not the right size. Some icon resources are supposed to be custom-sized, for example the glyphs are much smaller. In these cases the resource size should be read from the bitmap and scaled according to DPI.

Tanya-Solyanik: For ThreadExceptionDialog, and GridErrorDialog up/down arrows, and for ToolStripTemplateNode, `SystemIconSize` is the right size because the ico files show icons 16x16 pix if you open them in the VS.

But for DesignerActioGlyph and InheritanceUI we need a new function that uses the resource size as the logical size.
https://github.com/dotnet/winforms/blob/001b753d61752d9366c50e73e150b323ad12b095/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/Behavior/DesignerActionGlyph.cs#L98-L106
https://github.com/dotnet/winforms/blob/001b753d61752d9366c50e73e150b323ad12b095/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/InheritanceUI.cs#L23-L26

Tanya-Solyanik: I would add hardcoded glyph sizes to the callsites(becasue we know that these icons are in our control) and call ScaleHelper.ScaleToInitialSystemDpi on them. This would allow all icons follow the same optimized code path. @JeremyKuhne  - does this look reasonable.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/Behavior/DesignerActionGlyph.cs
```diff
@@ -98,12 +100,12 @@ public bool IsInComponentTray
     private Image GlyphImageClosed => _glyphImageClosed ??= ScaleHelper.GetIconResourceAsBitmap(
         typeof(DesignerActionGlyph),
         "Close_left",
-        ScaleHelper.InitialSystemDpi);
+        new Size(IconsWidth, IconsHeight));
```
Tanya-Solyanik, line 103: Please scale these sizes with ScaleHelper.ScaleToInitialSystemDpi

Tanya-Solyanik: And the same applies to InheritanceUI.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/Behavior/DesignerActionGlyph.cs
```diff
@@ -15,6 +15,8 @@ internal sealed class DesignerActionGlyph : Glyph
 {
     internal const int CONTROLOVERLAP_X = 5;                    // number of pixels the anchor should be offset to the left of the control's upper-right
     internal const int CONTROLOVERLAP_Y = 2;                    // number of pixels the anchor overlaps the control in the y-direction
+    private const byte IconsWidth = 10;
```
Tanya-Solyanik, line 18: I would define a single value `private const int IconSize`

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/Behavior/DesignerActionGlyph.cs
LeafShi1: > And the same applies to InheritanceUI.

In InheritanceUI, the size of the [InheritedGlyph icon ](https://raw.githubusercontent.com/dotnet/winforms/001b753d61752d9366c50e73e150b323ad12b095/src/System.Windows.Forms.Design/src/Resources/System/WinForms/Design/InheritedGlyph.ico)is right logic size 16X16, so this one don't need to be changed.

https://github.com/dotnet/winforms/blob/001b753d61752d9366c50e73e150b323ad12b095/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/InheritanceUI.cs#L23-L26

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/Behavior/DesignerActionGlyph.cs
```diff
@@ -47,6 +50,7 @@ public DesignerActionGlyph(DesignerActionBehavior? behavior, Rectangle alternati
     private DesignerActionGlyph(DesignerActionBehavior? behavior, Adorner? adorner, Rectangle alternativeBounds, Control? alternativeParent)
         : base(behavior)
     {
+        _iconsWidth = _iconsHeight = (byte)ScaleHelper.ScaleToInitialSystemDpi(IconSize);
```
Tanya-Solyanik, line 53: We only need this size before we display the icon. There is no reason to scale it in the constructor. To save some space and work, we can make the loaded icons static, both the property and the backing field, as this code does not support the PerMonitorV2 mode and is always scaled according to the primary screen DPI as available at the start up time.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/Behavior/DesignerActionGlyph.cs
```diff
@@ -98,12 +102,12 @@ public bool IsInComponentTray
     private Image GlyphImageClosed => _glyphImageClosed ??= ScaleHelper.GetIconResourceAsBitmap(
         typeof(DesignerActionGlyph),
         "Close_left",
-        ScaleHelper.InitialSystemDpi);
+        new Size(_iconsWidth, _iconsHeight));
```
Tanya-Solyanik, line 105: ScaleHelper.ScaleToInitialSystemDpi(IconSize)
```suggestion
        ScaleToDpi(new Size(IconSize, IconSize),  ScaleHelper.InitialSystemDpi));
```

Tanya-Solyanik (commented):
Looks good!

Tanya-Solyanik (approved):
Thank you! Please select squash-merge, a single commit will be the cleanest option here.

lonitra (approved):

PR #10460, opened 2023-12-11
Suppress finalizeation of ThreadMethodEntry class. Fix #10361.

### LINKED ISSUE #10361: Suppress finalizeation of ThreadMethodEntry class
### Background and motivation

[ThreadMethodEntry](https://github.com/dotnet/winforms/blob/d20f9c71a189693c6c3475253397665bce0a1354/src/System.Windows.Forms/src/System/Windows/Forms/Control.ThreadMethodEntry.cs#L11) class have a Finalizer but not implement an `IDisposable` interface. So, after **every** call to `BeginInvoke`/`Invoke` instance of `ThreadMethodEntry` will end up in Freachable queue:
Snipaste_2023-11-22_14-23-10

Finalizer here to free WH:
https://github.com/dotnet/winforms/blob/d20f9c71a189693c6c3475253397665bce0a1354/src/System.Windows.Forms/src/System/Windows/Forms/Control.ThreadMethodEntry.cs#L51-L54
We have 3 cases here (for all of them `ThreadMethodEntry` will end up in Freachable queue any way) :
- We not allocate WH.
- We allocate WH and use it internally (`Invoke`) - we not free it. This already violate `IAsyncResult.AsyncWaitHandle` [using rules](https://learn.microsoft.com/en-us/dotnet/api/system.iasyncresult.asyncwaithandle?view=net-8.0#notes-to-callers).
- User allocate WH through `IAsyncResult.AsyncWaitHandle` (`BeginInvoke`). And even if he properly Close this WH, `ThreadMethodEntry` will end up in Freachable queue :(

### Proposal

Suggestions:
- implement `IDisposable`?
- register for finalization only when WH is allocated.
- SuppressFinalize after WH disposed.

If you don't mind in principle, you can assignee this to me. I have some ideas on this matter, and will try to implement them as soon as find time.

### Risks

Must be minimal...

No

kirsan31:
Fixes #10361
After every call to `BeginInvoke`/`Invoke` instance of `ThreadMethodEntry` will end up in Freachable queue.

## Proposed changes
~~- Register for finalization only when WH is allocated.~~
~~- SuppressFinalize after WH disposed.~~
- Remove `ThreadMethodEntry` finalizer.
- Dispose `WaitHandle`  during  `Invoke` call.

## Customer Impact
- `ThreadMethodEntry` will not end in Freachable queue - GC will be more faster - overall improve performance.

## Regression? 

- No

## Risk

Minimal.

## Test methodology 

~~New UT added.
I would like to add tests for the triggering of the finalizer, but unfortunately we cannot correctly handle exceptions in it :( That is, the only option is to add some kind of debug only test logic - is this necessary?~~

--------------------------

 P.S. I have many test failure (not related to this PR) on my pc:
testError

--------------------------

P.P.S. I have build error with VS 17.8.3:
```
21>MSBuild version 17.8.3+195e7f5a3 for .NET Framework
21>C:\Program Files\Microsoft Visual Studio\2022\Community\MSBuild\Microsoft\VC\v170\Microsoft.CppCommon.targets(254,5): error MSB8066: Custom build for 'D:\Projects\winforms\artifacts\obj\NativeTests\x86\Debug\CMakeFiles\6366f68e1e32f772e531aff683e51b2d\Contract_i.c.rule' exited with code 9009. [D:\Projects\winforms\artifacts\obj\NativeTests\x86\Debug\NativeTests.vcxproj]
21>C:\Users\_\.nuget\packages\microsoft.dotnet.cmake.sdk\9.0.0-beta.23607.2\build\Microsoft.DotNet.CMake.Sdk.targets(190,5): error MSB3073: The command "
21>C:\Users\_\.nuget\packages\microsoft.dotnet.cmake.sdk\9.0.0-beta.23607.2\build\Microsoft.DotNet.CMake.Sdk.targets(190,5): error MSB3073:       call "D:\Projects\winforms\eng\init-vs-env.cmd" 
21>C:\Users\_\.nuget\packages\microsoft.dotnet.cmake.sdk\9.0.0-beta.23607.2\build\Microsoft.DotNet.CMake.Sdk.targets(190,5): error MSB3073:       
21>C:\Users\_\.nuget\packages\microsoft.dotnet.cmake.sdk\9.0.0-beta.23607.2\build\Microsoft.DotNet.CMake.Sdk.targets(190,5): error MSB3073:       cmake --build "D:\Projects\winforms\artifacts\obj\NativeTests\x86\Debug" --target install  --config Debug -- -verbosity:quiet /p:PlatformToolset=%__PlatformToolset%
21>C:\Users\_\.nuget\packages\microsoft.dotnet.cmake.sdk\9.0.0-beta.23607.2\build\Microsoft.DotNet.CMake.Sdk.targets(190,5): error MSB3073:       " exited with code 1.
21>Done building project "NativeTests.proj" -- FAILED.
```

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10460)


lonitra (commented):
Thank you for your patience! Added some small comments

lonitra (approved):
LGTM!

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ControlTests.Methods.cs
```diff
@@ -4177,6 +4177,115 @@ public void Control_Invoke_InvokeInvalidParameters_ThrowsTargetParameterCountExc
         Assert.Throws<TargetParameterCountException>(() => control.Invoke(method, Array.Empty<object>()));
     }
 
+#if DEBUG
+    // Here we will test for ThreadMethodEntry destructor will not be called.
+    // If we use AsyncWaitHandle and not close it, then we will crash to xUnit runner:
+    // The active test run was aborted. Reason: Test host process crashed : Unhandled exception. DebugAssertException: Method Debug.Fail failed with 'ThreadMethodEntry finalization hit!'
+    [WinFormsFact]
+    public void Control_Invoke_SameThread_ThreadMethodEntry_Finalizeation_Success()
```
JeremyKuhne, line 4185:
```suggestion
    public void Control_Invoke_SameThread_ThreadMethodEntry_Finalization_Success()
```

Same with the others.

JeremyKuhne (changes requested):
In general this looks good, just a few questions / comments.

kirsan31:
@JeremyKuhne 
- Suggestions implemented.
- Fixed some tests (they never failed). _P.s. it's hard to have no way to check for failures :(_ For example this code:  
  ```cs
  using (Control control = new())
  {
      control.CreateControl();
      Action method = () => { };
      IAsyncResult res = control.BeginInvoke(method);
      WaitHandle waitHandle = res.AsyncWaitHandle;
      while (!waitHandle.WaitOne(1))
      {
          Application.DoEvents();
      }
  }
  
  GC.Collect();
  GC.Collect();
  ```
  will not trigger `ThreadMethodEntry` finalization for some reason  🤷‍♂️ And I had to move the code to a local function...
- [Question](https://github.com/dotnet/winforms/pull/10460#discussion_r1453862120) about `#if DEBUG` is still actual.


Also I have noticed some _interesting_ thing. In `ManualResetEventPrivate` I use:
```cs
protected override void Dispose(bool explicitDisposing)
{
    base.Dispose(explicitDisposing);
    if (explicitDisposing)
    {
        // The owner (ThreadMethodEntry) need to free only this waitHandle.
        // But ThreadMethodEntry can't be Disposed directly because it exposed though IAsyncResult interface.
        // So after freeing this waitHandle we mark owner ThreadMethodEntry to suppress finalization.
        GC.SuppressFinalize(_owner);
    }
}
```
and `explicitDisposing` here is always true (even when called from finalizer): 
```cs
~ThreadMethodEntry()
{
    _resetEvent?.Close();
    Debug.Fail($"{nameof(ThreadMethodEntry)} finalization hit!");
}
```

Due to `WaitHandle` code:
```cs
public virtual void Close() => Dispose();

protected virtual void Dispose(bool explicitDisposing)
{
    _waitHandle?.Close();
}

public void Dispose()
{
    Dispose(true);
    GC.SuppressFinalize(this);
}
```

could this be a problem?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ControlTests.Methods.cs
```diff
@@ -4177,6 +4177,115 @@ public void Control_Invoke_InvokeInvalidParameters_ThrowsTargetParameterCountExc
         Assert.Throws<TargetParameterCountException>(() => control.Invoke(method, Array.Empty<object>()));
     }
 
+#if DEBUG
```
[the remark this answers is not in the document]
JeremyKuhne: @kirsan31 There is probably at least some value in running the tests in release on the slim chance that they fail. If we're concerned about test performance I'd rather see these conditionalized in code.

JeremyKuhne:
@kirsan31

> will not trigger ThreadMethodEntry finalization for some reason 🤷‍♂️ And I had to move the code to a local function...

Presumably it is still rooted in a local? At least historically I remember the GC keeping all locals rooted to the end of a stack frame in debug.

src/System.Windows.Forms/src/System/Windows/Forms/Control.ThreadMethodEntry.cs
```diff
@@ -40,17 +40,18 @@ private class ThreadMethodEntry : IAsyncResult
...
             _executionContext = executionContext;
+            // We need finalization only if we will create _resetEvent instance.
+            GC.SuppressFinalize(this);
         }
 
         ~ThreadMethodEntry()
         {
+            // If something new is added here, it is necessary to reconsider the conditions of finalization.
+            // Look GC.SuppressFinalize(this); in constructor and GC.ReRegisterForFinalize(this); in AsyncWaitHandle.
             _resetEvent?.Close();
```
JeremyKuhne, line 53: @kirsan31 I hadn't really internalized that we're touching a class in the finalizer. Why do we even have a finalizer? The event has its own `SafeHandle` that should take care of things.

kirsan31: Ha, I think you're right - why did I decide to overcomplicate everything? 🤦‍♂️ Indeed, removing the destructor and disposing `WaitHandle` after `Invoke` should solve all the problems...
I'll correct...

JeremyKuhne (approved):

PR #10475, opened 2023-12-14
Fix RichTextBox drag and drop

### LINKED ISSUE #10470: Memory corruption/crash when using drag'n'drop over a WinForms RichTextBox
### .NET version

.NET 8.0

Yes

yes

### Issue description

When you enable drag'n'drop for RichTextBox and drag something over/into it a few times the application eventually crashes due to memory corruption

### Steps to reproduce

Place a RichTextBox on a Form and set `EnableAutoDragDrop` to true. Run the application and drag something into the RichTextBox a few times, for example texts and images from the browser.

JeremyKuhne:
When receiving dropped data the COM IDataObject was not adding a ref count as appropriately. (The RCW adds a ref, but we were also releasing it.)

This causes random memory corruption, and crashes .NET.

Fixes #10470

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10475)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.cs
@@ -66,7 +66,7 @@ public unsafe partial class DataObject :
         // Get the RCW for the pointer and continue.
         bool success = ComHelpers.TryGetObjectForIUnknown(
             (Com.IUnknown*)data,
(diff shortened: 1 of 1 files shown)


JeremyKuhne:
Need to service .NET 8 with this.

lonitra (approved):

PR #10478, opened 2023-12-14
Hide the RichTextBox drag image and drop description if the Effect is set to None

### LINKED ISSUE #8903: A drag image remnant remains when dragging between different application instances
### .NET version

.NET 7

No

This feature is new as of .NET 7.

### Issue description

When dragging between different application instances, and when a drag image is specified, a remnant of the drag image is incorrectly left displayed due to what I believe is a possible oversight in the DropSource implementation. Note that this issue does not occur when dragging within the same application instance.

A video demonstrating the drag image remnant when dragging between different application instances.

https://user-images.githubusercontent.com/5017479/227811248-964e91b9-bff8-491d-a045-1b0940a2e35e.mp4

A video demonstrating the drag image behavior when dragging within the same application instance.

https://user-images.githubusercontent.com/5017479/227811261-fdb8457b-64a2-4e6d-8ad8-eb9acfa275c9.mp4

### Steps to reproduce

WinformsControlTest project -> DragDrop

See the video demonstrating the issue when dragging between different application instances.

willibrandon:
Fixes #8903 

## Proposed changes

- The RichTextBox behaves differently in the cross-process scenario and the drag image manager doesn't automatically synthesize a call to DragLeave on drop when the Effect is set to None, like it does in the in-process scenario, leaving a drag image artefact.
- Hide the RichTextBox drag image and drop description if the application sets the Effect to None.

## Customer Impact

- Fixes a drag image artefact that occurs when dropping onto the RichTextBox cross-process with an Effect of None.

## Regression? 

- No

## Risk

-

### Before

Cross-process behavior on drop when the Effect is set to None.

https://github.com/dotnet/winforms/assets/5017479/784fa0a4-bf5e-44c4-807c-a2ecc1152b7b

In-process behavior on drop when the Effect is set to None. Note how the drag image manager automatically synthesizes a drag leave on drop. I'm very interested in learning why there is a difference since obviously this is the preferred behavior.

https://github.com/dotnet/winforms/assets/5017479/2ace6804-8994-4e41-96d8-4268b265d763

### After

This after video also includes the change made in #10475 for a complete demonstration.

https://github.com/dotnet/winforms/assets/5017479/bd47839c-179f-4631-b215-91661c09a735

One more video demonstrating a workaround for this minor annoyance by setting the RichTextBox Effect to something other than None in DragEnter in order to display the drag image and drop description, and more importantly, not leave an artefact.

https://github.com/dotnet/winforms/assets/5017479/cad6f4bb-1c00-4e65-990f-7cc1c890ff97

## Test methodology 

- Manual

## Test environment(s) 

- 8.0.0

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10478)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Controls/RichTextBox/RichTextBox.OleCallback.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Controls/RichTextBox/RichTextBox.OleCallback.cs
@@ -180,7 +180,7 @@ public partial class RichTextBox
                 e.Effect = keyState.HasFlag(MODIFIERKEYS_FLAGS.MK_CONTROL) ? DragDropEffects.Copy : DragDropEffects.Move;
                 _owner.OnDragEnter(e);
 
-                if ((e.Effect != DragDropEffects.None && e.DropImageType > DropImageType.Invalid) && _owner.IsHandleCreated)
+                if (CanShowImage(e))
                 {
                     UpdateDropDescription(e);
                     DragDropHelper.DragEnter(_owner.Handle, e);
@@ -295,7 +295,7 @@ public partial class RichTextBox
                     _owner.OnDragOver(e);
                     _lastEffect = e.Effect;
 
-                    if (e.Effect != DragDropEffects.None && e.DropImageType > DropImageType.Invalid)
+                    if (CanShowImage(e))
                     {
                         UpdateDropDescription(e);
                         DragDropHelper.DragOver(e);
@@ -324,6 +324,9 @@ public partial class RichTextBox
 
             return HRESULT.S_OK;
         }
+
+        private bool CanShowImage(DragEventArgs e)
+            => e.Effect != DragDropEffects.None && e.DropImageType > DropImageType.Invalid && _owner.IsHandleCreated;
 
         private void UpdateDropDescription(DragEventArgs e)
         {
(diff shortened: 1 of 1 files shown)


willibrandon:
Perfecto!

willibrandon:
One thing that should be said is I really do need a better way to test drag and drop, and not have to worry so much about it breaking in CI. It has been a big source of frustration for me and I can't imagine what some of these missteps must look like to some of you. I desperately need to find a way to reliably automate all of the drag and drop use cases.

src/System.Windows.Forms/src/System/Windows/Forms/Controls/RichTextBox/RichTextBox.OleCallback.cs
```diff
@@ -295,7 +295,7 @@ public unsafe HRESULT GetDragDropEffect(BOOL fDrag, MODIFIERKEYS_FLAGS grfKeySta
                     _owner.OnDragOver(e);
                     _lastEffect = e.Effect;
 
-                    if (e.DropImageType > DropImageType.Invalid)
+                    if (e.Effect != DragDropEffects.None && e.DropImageType > DropImageType.Invalid)
```
Tanya-Solyanik, line 298: Would it make sense to introduce a local function that encapsulated this conditions? `CanShowImage` for example.

willibrandon: Agreed, it reads better that way. I also wonder since we have the same/similar check in `QueryAcceptData`, if both should be combined into a private `CanShowImage` method.

Tanya-Solyanik (approved):
Thank you!

PR #10483, opened 2023-12-14
Fix RichTextBox drag and drop (#10475)

JeremyKuhne:
Port of #10475 to release.

When receiving dropped data the COM IDataObject was not adding a ref count as appropriately. (The RCW adds a ref, but we were also releasing it.)

This causes random memory corruption, and crashes .NET.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10483)


Tanya-Solyanik (approved):

lonitra:
February servicing branch is not open yet, this will need to wait until then for merge.

PR #10493, opened 2023-12-18
Do not use fixed index in the `TreeNode.PrevNode` if the `TreeView` is sorted 

### LINKED ISSUE #8768: TreeNodeCollection.AddRange works incorrectly for non-empty collection of the sorted TreeView

elachlan:
Fixes #8768

This is a re-roll of the PR in #8774, but with the switch added.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10493)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
@@ -21,6 +21,7 @@ internal static partial class LocalAppContextSwitches
     internal const string TrackBarModernRenderingSwitchName = "System.Windows.Forms.TrackBarModernRendering";
     private const string DoNotCatchUnhandledExceptionsSwitchName = "System.Windows.Forms.DoNotCatchUnhandledExceptions";
     internal const string DataGridViewUIAStartRowCountAtZeroSwitchName = "System.Windows.Forms.DataGridViewUIAStartRowCountAtZero";
+    internal const string TreeNodePrevNodeDoNotUseFixedIndexWithSortedTreeViewSwitchName = "System.Windows.Forms.TreeNodeDoNotUseFixedIndexWithSortedTreeView";
 
     private static int s_scaleTopLevelFormMinMaxSizeForDpi;
     private static int s_anchorLayoutV2;
@@ -28,6 +29,7 @@ internal static partial class LocalAppContextSwitches
     private static int s_trackBarModernRendering;
     private static int s_doNotCatchUnhandledExceptions;
     private static int s_dataGridViewUIAStartRowCountAtZero;
+    private static int s_treeNodePrevNodeDoNotUseFixedIndexWithSortedTreeView;
 
     private static FrameworkName? s_targetFrameworkName;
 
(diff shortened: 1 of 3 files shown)


elachlan:
CC: @snnz

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -165,5 +167,11 @@ public static bool DataGridViewUIAStartRowCountAtZero
         get => GetCachedSwitchValue(DataGridViewUIAStartRowCountAtZeroSwitchName, ref s_dataGridViewUIAStartRowCountAtZero);
     }
 
+    public static bool TreeNodeDoNotUseFixedIndexWithSortedTreeView
```
lonitra, line 170: Please add comments for this switch

elachlan:
@lonitra I added the xml comment and renamed the switch to be more specific to its behavior.

snnz:
But this re-roll, it is totally incorrect. The purpose was to make sure that fixed index is NOT used if the tree is sorted, not vice versa. See the original PR: there should be `!TreeView.Sorted` in the condition.

lonitra:
> @lonitra I added the xml comment and renamed the switch to be more specific to its behavior.

Thank you! The change looks good. Could you add tests to capture the behavior of the switch when it is on?

elachlan:
Sorry! I'll try and get this sorted soon.

elachlan:
@lonitra I am unsure how we can unit test this. As the behavior of affects the underlying win32 tree view control.

elachlan:
I did some manual testing and was getting different results than what was specified in the #8768.

I then looked into the differences, we changed the behavior of the insert in #9078 by @gpetrou 

Did we want to change the behavior back, or keep it as is? It used to add in "almost" reverse order, but now it adds it in the order provided (almost).

**Before:**
223826227-6cf6d1aa-3605-4e39-98a8-65116d5db21a

**After:**
image

Method of adding:
```csharp
  treeView1.Nodes.AddRange([
  new TreeNode("3"),
  new TreeNode("5"),
  new TreeNode("1"),
  new TreeNode("4")
]);
```

elachlan:
Test is showing some weird behavior and I am not sure why. It doesn't line up with what I observed when manually testing.

Expected: `2,5,3,1,4`
Actual:  `2,3,4,5,1`

lonitra:
> Did we want to change the behavior back, or keep it as is? It used to add in "almost" reverse order, but now it adds it in the order provided (almost).

Had you investigated what exactly caused the behavior change? From glancing at the PR it doesn't seem like the change was intentional.

> Test is showing some weird behavior and I am not sure why. It doesn't line up with what I observed when manually testing.

Hm.. That is concerning. It seems like the state between our control and the underlying one is mismatched and the results you are getting are actually being sorted when it comes to the underlying control.. This looks a bit suspicious to me
https://github.com/dotnet/winforms/blob/226a5b5d3782f40d8c84cdacfa3c49b3f7cc042e/src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNodeCollection.cs#L315-L318
this is where it looks like we eventually get to from `AddRange`. Could you maybe debug through this and see if somewhere along the line we are adding the tree node to a unexpected index for the underlying control? -- I cannot look into this too deepy right now, apologies!

elachlan:
Sorry I haven't sorted this yet. Hopefully soon.

src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNode.cs
```diff
@@ -812,17 +813,20 @@ public TreeNodeCollection Nodes
...
+            int currentIndex = _index;
+            int fixedIndex = _parent.Nodes.FixedIndex;
 
-            if (fixedInd > 0)
+            if (fixedIndex > 0)
             {
-                currentInd = fixedInd;
+                if (!LocalAppContextSwitches.TreeNodePrevNodeDoNotUseFixedIndexWithSortedTreeView || TreeView is null || !TreeView.Sorted)
+                {
+                    currentIndex = fixedIndex;
```
Tanya-Solyanik, line 823: The idea of this fix is that we want to skip the fixed index perf optimization when we are sorting the treeview.

elachlan: Yes, based on this comment from the issue reporter.
https://github.com/dotnet/winforms/pull/10493#issuecomment-1873621647

Tanya-Solyanik: Then we don't have to set FIxedIndex to the currently displayed number of child nodes, I prefer fix in https://github.com/dotnet/winforms/pull/11300

elachlan: No problem. Ill close this then. The tests still seem weird. So worth investigating.

Issue #10443, opened 2023-12-07
DataGridView and its elements accessible objects are leaking in memory


MelonWang1:
### .NET version

.NET 9.0.100-alpha.1.23606.1 + private dlls(build from main branch of WinForms repo)

### Did it work in .NET Framework?

Not tested/verified

### Did it work in any of the earlier releases of .NET Core or .NET 5+?

Yes, this issue cannot repro in .NET 9.0.100-alpha.1.23606.1, it may be caused by PR: [10407](https://github.com/dotnet/winforms/pull/10407).

### Issue description

The original issue is [7347](https://github.com/dotnet/winforms/issues/7347).
DataGridView and its elements accessible objects are left in memory after the DataGridView control instance with its rows was inspected with accessibility tool (like Accessibility Insights/Inspect/Narrtor) and the form with DataGridView is closed, garbage is collected, but accessibility tool is still opened. 

Report from WinDbg:
![datagridview-memory leak](https://github.com/dotnet/winforms/assets/94418985/f8b3804d-b339-4dde-aff2-a36bba165a3f)

### Steps to reproduce

Tested sample application:
[DataGridView Sample.zip](https://github.com/dotnet/winforms/files/13595985/DataGridView.Sample.zip)

1.Create a parent form with two buttons: one will be opening a child form with the DataGridView control and the other will collect the garbage.
2. Start project without debugging(Ctrl+F5).
3. Attaching project to WinDbg and adding symbols folder.
4. Start the Accessibility Insights tool.
5. Press the second button to open the child form with DatGridView.
6. Automatic checks for the DataGridView control using Accessibility Insight.
7. Close the child form.
8. Clear memory with the first button of the parent form.
9. Don't close the Accessibility Insights tool.
10. Look for DataGridView accessible objects left in memory (using WinDbg).

Tanya-Solyanik:
@MelonWang1  - thank you for incorporating the leak detection tests in your regular testing!

Tanya-Solyanik:
I see DGV column arrays rooted to an array before and after PR #10407

Tanya-Solyanik:
Though I see crashes on application shutdowns related to UIA providers that were not disposed - 
 	KernelBase.dll!RaiseException() + 108 bytes	Unknown
 	ucrtbase.dll!_CxxThrowException() + 154 bytes	Unknown
[...]

PR #10501, opened 2023-12-18
Release Pointer when Index Setting ComSafeArrayScope

lonitra:
Fixes https://github.com/dotnet/winforms/issues/10443

If the object being placed into a SafeArray is a IDispatch or IUnknown, SafeArray does an `AddRef` on the object (https://learn.microsoft.com/en-us/windows/win32/api/oleauto/nf-oleauto-safearrayputelement#remarks). This means to maintain the correct ref count, we should be releasing the object that we put into the SafeArray. This was missed in `ComSafeArrayScope.CreateFromInterfaceArray`

Before:
image

After: No AccessibleObject leaks
image

Note:
This should also fix other AccessibleObject memory leaks that have been noted as .NET 9 regressions.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10501)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/Windows/Win32/System/Com/ComSafeArrayScope.cs
+++ b/src/System.Windows.Forms.Primitives/src/Windows/Win32/System/Com/ComSafeArrayScope.cs
@@ -61,7 +61,8 @@ internal readonly unsafe ref struct ComSafeArrayScope<T> where T : unmanaged, IC
         ComSafeArrayScope<T> scope = new(length);
         for (int i = 0; i < length; i++)
         {
-            scope[i] = ComHelpers.GetComPointer<T>(array[i]);
+            using var pointer = ComHelpers.GetComScope<T>(array[i]);
+            scope[i] = pointer;
         }
 
         return scope;
--- a/src/System.Windows.Forms.Primitives/src/Windows/Win32/System/Com/ComSafeArrayScope.cs
+++ b/src/System.Windows.Forms.Primitives/src/Windows/Win32/System/Com/ComSafeArrayScope.cs
@@ -61,6 +61,8 @@ internal readonly unsafe ref struct ComSafeArrayScope<T> where T : unmanaged, IC
         ComSafeArrayScope<T> scope = new(length);
         for (int i = 0; i < length; i++)
         {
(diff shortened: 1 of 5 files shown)


Tanya-Solyanik (approved):

src/System.Windows.Forms.Primitives/src/Windows/Win32/System/Com/ComSafeArrayScope.cs
```diff
@@ -61,7 +61,8 @@ public static ComSafeArrayScope<T> CreateFromInterfaceArray<TComInterface>(TComI
         ComSafeArrayScope<T> scope = new(length);
         for (int i = 0; i < length; i++)
         {
-            scope[i] = ComHelpers.GetComPointer<T>(array[i]);
+            using var pointer = ComHelpers.GetComScope<T>(array[i]);
```
JeremyKuhne, line 64: You should add a comment here about the ref counts.

JeremyKuhne (changes requested):
I'd add comments for extra clarity to hopefully avoid regressions here.

src/System.Windows.Forms/src/System/Windows/Forms/Controls/TextBox/TextBoxBase.TextBoxBaseUiaTextProvider.cs
```diff
@@ -67,7 +68,8 @@ public override HRESULT GetVisibleRanges(SAFEARRAY** pRetVal)
             GetVisibleRangePoints(out int start, out int end);
 
             ComSafeArrayScope<ITextRangeProvider> result = new(1);
-            result[0] = ComHelpers.GetComPointer<ITextRangeProvider>(new UiaTextRange(Owner.AccessibilityObject, this, start, end));
+            using var ranges = ComHelpers.GetComScope<ITextRangeProvider>(new UiaTextRange(Owner.AccessibilityObject, this, start, end));
```
JeremyKuhne, line 71: `// Adding to the SAFEARRAY adds a reference` on each of these would help I think

JeremyKuhne (approved):

PR #10502, opened 2023-12-19
Enable `IDE0008` / `IDE0090` analyzers

### LINKED ISSUE #10500: Enable IDE0008 and IDE0090 in System.Windows.Forms.Tests.csproj
Currently IDE0008 and IDE0090 are disabled in System.Windows.Forms.Tests.csproj:
https://github.com/dotnet/winforms/blob/main/src/System.Windows.Forms/tests/UnitTests/System.Windows.Forms.Tests.csproj#L13

This should be enabled to help enforce style guidelines in the project. Existing build errors will need to be fixed upon enabling.

elachlan:
Fixes #10500

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10502)


.editorconfig
```diff
@@ -166,7 +166,7 @@ csharp_space_between_parentheses = false
 csharp_space_between_square_brackets = false
 
 # IDE0008: Use explicit type
-dotnet_diagnostic.IDE0008.severity = silent
+dotnet_diagnostic.IDE0008.severity = warning
```
elachlan, line 169: I think we should probably just remove it from `.editorconfig`

elachlan:
@lonitra any suggestions on the build failures?

```
src\System.Windows.Forms\tests\UnitTests\System\Windows\Forms\ListViewGroupTests.cs(1427,9): error CS0436: (NETCORE_ENGINEERING_TELEMETRY=Build) The type 'TestAccessors' in 'D:\a\_work\1\s\src\System.Windows.Forms\tests\UnitTests\System\Windows\Forms\PropertyGridInternal\TestAccessors.PropertyGridTestAccessor.cs' conflicts with the imported type 'TestAccessors' in 'System.Windows.Forms.Common.TestUtilities, Version=9.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35'. Using the type defined in 'D:\a\_work\1\s\src\System.Windows.Forms\tests\UnitTests\System\Windows\Forms\PropertyGridInternal\TestAccessors.PropertyGridTestAccessor.cs'.
```

https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs0436

src/System.Windows.Forms/tests/UnitTests/System/Drawing/Design/UITypeEditorTests.cs
```diff
@@ -44,7 +44,7 @@ public static IEnumerable<object[]> EditValue_IServiceProvider_Object_TestData()
     [MemberData(nameof(EditValue_IServiceProvider_Object_TestData))]
     public void UITypeEditor_EditValue_Invoke_CallsVirtualEditValue(IServiceProvider provider, object value)
     {
-        var result = new object();
+        object result = new object();
```
Tanya-Solyanik: Use target-typed `new` , this applies in multiple places in this PR.

.editorconfig
Tanya-Solyanik: The code style insists on specifying explicit types for built-in types - https://github.com/dotnet/winforms/blob/main/docs/coding-style.md#variables, not in cases like this one - 

>using var stream = ComHelpers.GetComScope<IStream>(new Interop.Ole32.GPStream(memoryStream));

That was the reason for `silent` I assume

lonitra: Recent arcade update has started causing IDE0008 build errors. I had gone through a pass to fix the errors to get arcade PR merged. All have been fixed except for ‘System.Windows.Forms.Test’ project. I was mistaken and thought that errors from arcade PR meant we needed to enforce IDE0008 but after some more reading it now seems like our editorconfig is not being respected.. 
@elachlan could you revert enabling IDE0008 for now? I’ll need to investigate why build errors are occurring for this. We do want IDE0090 enabled though.

elachlan: Thanks! Ill get this sorted early in the new year. I'll be afk until then.

lonitra: Sure thing, enjoy the holidays! (:

sharwell, line 169: I'm going to tentatively request changes here since the preferred strategy is using dotnet/roslyn#69843. I need to review the status of that work.

sharwell (changes requested):

.editorconfig
sharwell: It looks like this functionality was completed in dotnet/roslyn#69844. I'll test to see if it's working in the latest branches of this repository.

sharwell: 🎉 Verified that if you remove _all_ references to IDE0008 from the solution, it will use the three lines above for the severity. This allows the three different cases to each have their own severity instead of requiring an all-or-nothing approach. Based on this, I believe this pull request will end up dramatically simplified (items falling under `csharp_style_var_elsewhere = false:silent` no longer need to change), so it might be best to just start over on it.

elachlan: @sharwell thank you for doing the leg work here.

src/System.Windows.Forms/tests/UnitTests/DataStreamFromComStreamTests.cs
```diff
@@ -14,7 +14,7 @@ public class DataStreamFromComStreamTests
     public unsafe void Write_ThrowsInvalidCount(int bufferSize, int index, int count)
     {
         using MemoryStream memoryStream = new();
-        using var stream = ComHelpers.GetComScope<IStream>(new Interop.Ole32.GPStream(memoryStream));
+        using ComScope<IStream> stream = ComHelpers.GetComScope<IStream>(new Interop.Ole32.GPStream(memoryStream));
```
sharwell, line 17: ❓ Was this change applied before or after the configuration was updated to make this location not be a warning? I would have expected this to stay as `var`.

elachlan: I started from scratch and changed the `.editorconfig` before applying the codefix.

sharwell: Interesting. Sounds like the Fix All infrastructure isn't filtering diagnostics based on severity.

lonitra: We don't want changes from `var` -> `ComScope`. Explicit type should only be used for built-in types, or if it is necessary to know the type to understand what the code is doing and type is not immediately apparent from reading the code. 

@sharwell what would be your recommendation to avoid this? Should we have `csharp_style_var_elsewhere = true:silent` instead?

sharwell: @lonitra The only way to avoid it currently is to manually review the changes and only commit the ones you want. Roslyn needs to fix the way it runs a Fix All operation with mixed severities.

Note that reverting the var&rarr;ComScope change will not introduce warnings.

elachlan: I'll get this sorted today.

elachlan: @sharwell is there an issue tracking the problem with the Fix All?

src/System.Windows.Forms/tests/UnitTests/System/Resources/ResXResourceWriterTests.cs
```diff
@@ -136,7 +136,7 @@ public void ResXResourceWriter_AddMetadata_ByteValue_AddsMetadataToResourceFile(
 
         // Assert
         using ResXResourceReader reader = new(_resxFileName);
-        var metadataEnumerator = reader.GetMetadataEnumerator();
+        IDictionaryEnumerator metadataEnumerator = reader.GetMetadataEnumerator();
```
lonitra, line 139: Please revert some of these other ones back to `var` as well. We should only use explicit type for built in types or if the code cannot be understood easily without the explicit type. For most of these, there doesn't seem to be a need to know the type expicitly so we can revert back to `var`. If you feel strongly that a line should use the explicit type then please let me know, happy to discuss it.

elachlan: I'll revert the changes and then got through and individually fix each of the 150 errors. That way only the ones we want to be fixed will be fixed.

elachlan: @lonitra how is that?

lonitra (approved):
Awesome! LGTM thanks for going through that.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/InternalTestAccessors.KeyboardToolTipStateMachineTestAccessor.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace System;
 
-public static partial class TestAccessors
+internal static partial class InternalTestAccessors
```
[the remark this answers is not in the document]
elachlan: So most of it was "undone" already. Just the rename was needing to be undone.

Basically if we used an explicit type, it would cause the error.

src/System.Windows.Forms/tests/UnitTests/DataStreamFromComStreamTests.cs
sharwell: Not currently.I 've been discussing it with @mavasani

sharwell: https://github.com/dotnet/roslyn/pull/71486

lonitra (approved):

PR #10521, opened 2023-12-21
Fix descriptors cache in `ShadowPropertyCollection`

### LINKED ISSUE #8231: Cache is never populated in ComponentDesigner.ShadowPropertyCollection.cs

elachlan:
Fixes #8231

Winforms Team still would like to do a bit of investigation on the VS side of things so we can make sure that we don't run into something unexpected.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10521)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.ShadowPropertyCollection.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.ShadowPropertyCollection.cs
@@ -40,7 +40,7 @@ public partial class ComponentDesigner
             }
             set
             {
-                _properties ??= new();
+                _properties ??= [];
                 _properties[propertyName] = value;
             }
         }
@@ -55,14 +55,18 @@ public partial class ComponentDesigner
         /// </summary>
         private PropertyDescriptor? GetShadowedPropertyDescriptor(string propertyName)
         {
-            _descriptors ??= new();
+            _descriptors ??= [];
 
-            if (_descriptors.TryGetValue(propertyName, out PropertyDescriptor? property))
+            if (!_descriptors.TryGetValue(propertyName, out PropertyDescriptor? descriptor))
             {
-                return property;
+                descriptor = TypeDescriptor.GetProperties(_designer.Component.GetType())[propertyName];
+                if (descriptor is not null)
+                {
+                    _descriptors.Add(propertyName, descriptor);
+                }
             }
 
-            return TypeDescriptor.GetProperties(_designer.Component.GetType())[propertyName];
+            return descriptor;
         }
 
         /// <summary>


SimonZhao888 (dismissed):
LGTM!

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.ShadowPropertyCollection.cs
```diff
@@ -55,14 +55,18 @@ protected sealed class ShadowPropertyCollection
         /// </summary>
         private PropertyDescriptor? GetShadowedPropertyDescriptor(string propertyName)
         {
-            _descriptors ??= new();
+            _descriptors ??= [];
```
Tanya-Solyanik, line 58: FYI - I would have moved this into the constructor and declared this field readonly. If ShadowPropertyCollection is instantiated, the "shadowed" descriptors are populated. It's useless otherwise. But this is outside of scope for this PR.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.ShadowPropertyCollection.cs
```diff
@@ -55,14 +55,18 @@ protected sealed class ShadowPropertyCollection
...
 
-            if (_descriptors.TryGetValue(propertyName, out PropertyDescriptor? property))
+            if (!_descriptors.TryGetValue(propertyName, out PropertyDescriptor? property))
```
Tanya-Solyanik, line 60: Suggestion: rename `property` to `descriptor` because we later add it to a `_descriptors` collection

Tanya-Solyanik (dismissed):
I added a minor comment, take it or leave it.
This code is not executed in the VS designer. But you should hit a breakpoint here in the demo console.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/ComponentDesigner.ShadowPropertyCollection.cs
elachlan: I'll get all these changes sorted next week. Thank you for the review!

Tanya-Solyanik (approved):

Issue #10524, opened 2023-12-22
XML comment crefs unusable with shared/duplicate CsWin32


JeremyKuhne:
XML comment `cref`s don't resolve the same way code does. They have visibility to everything, which prevents us from using them when we have duplicate type names in dependent assemblies or types that cross assemblies. We either have to turn off [CS1574](https://learn.microsoft.com/dotnet/csharp/misc/cs1574) and [CS1580](https://learn.microsoft.com/dotnet/csharp/misc/cs1574), or remove lots of existing `cref`s.

JeremyKuhne:
@jaredpar, can give guidance on where we should create an issue for this?

@AArnott, fyi.

jaredpar:
Tracking the issue here https://github.com/dotnet/roslyn/issues/71442

elachlan:
@JeremyKuhne was this resolved by using a common base library?

JeremyKuhne:
This issue is just a tracking issue for the linked C# feature.

PR #10320, opened 2023-11-16
Start using `CsWin32` in `System.Drawing.Common`

elachlan:
Related: #9879

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10320)

PR #10525, opened 2023-12-22
Prototype shared CsWin32 with System.Drawing.Common

JeremyKuhne:
This is a rough prototype of a shared interop with System.Drawing.Common. It introduces a new base assembly for shared CsWin32 and other low-level support code.

Converted HGLOBAL, DEVMODE and related stuff to the new shared pattern. Also used the CsWin32 generated GdiplusStartup.

I have almost no idea if packaging is correct here. Assuming it isn't.

`crefs` don't work right when there is overlap in internals. #10524 tracks this, and this change disables the relevant warnings.

I've only smoke tested this with the WinFormsControlsTest.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10525)


JeremyKuhne:
Note that I'm also hitting some weird file locking during builds.

This is not necessarily the way we'll go. There are definitely benefits in simpler sharing of low-level code, but we still need to discuss options.

Mostly out for the holidays so unlikely to respond to anything until I'm back.

eng/packageContent.targets
```diff
@@ -56,8 +56,8 @@
       This means we build a real assembly that has no associated official intellisense docs.
       Contact the intellisense team for guidance.
     -->
-    <Error  Condition="!Exists('$(IntellisenseXmlFileSource)') and '$(IsFacadeAssembly)' != 'true' and '$(IsAnalyzerProject)' != 'true'" 
-            Text="$(AssemblyName).xml was not found in $(_DotnetApiDocsPackageName) package. Contact the intellisense team about adding the docs for this assembly."
+    <Error Condition="!Exists('$(IntellisenseXmlFileSource)') and '$(IsFacadeAssembly)' != 'true' and '$(IsAnalyzerProject)' != 'true' and '$(HasNoPublicIntellisense)' != 'true'"
```
JeremyKuhne, line 59: @ericstj Tried to work around this error here, but there are others and I'm unsure if this is the right thing to do for this one.

src/System.Windows.Primitives/src/System.Windows.Primitives.csproj
```diff
@@ -0,0 +1,41 @@
...
+      See https://github.com/dotnet/winforms/issues/4649
+    <NoWarn>$(NoWarn);IL2026;IL2050</NoWarn>
+    -->
+    <!--
+      Libraries code has changed a number of APIs from `ref` to `in`. CSWin32 generates code that passes by `ref`
+      to some of these. Disabling for now until we can get https://github.com/microsoft/CsWin32/issues/1014 resolved.
+    <NoWarn>$(NoWarn);CS9195</NoWarn>
+    -->
+    <Deterministic>true</Deterministic>
+    <ProduceReferenceAssembly>true</ProduceReferenceAssembly>
```
ericstj, line 30: I think you want this to be false.  That might fix your build issue.  You also don't want this thing to be included in your ref-pack if it has no public API.

eng/packageContent.targets
ericstj: Build is failing because this is still included: 
https://github.com/dotnet/winforms/blob/bec1f5a39e01cf3a5ada1db09d481fe004087dad/eng/packageContent.targets#L67-L69C44

You might want to add the `HasNoPublicIntellisense` to that line or set `ProduceReferenceAssembly` to false and remove the `HasNoPublicIntellisense` if that makes more sense.

Directory.Build.props
```diff
@@ -19,6 +19,12 @@
     <IsAnalyzerProject Condition="!$(IsTestProject) and
                                   ($(MSBuildProjectName.EndsWith('.Analyzers')) or $(MSBuildProjectName.EndsWith('.Analyzers.CSharp')))"
                                   >true</IsAnalyzerProject>
+
+    <!--
+     XML cref tags don't resolve the same way code does. Until we can get this resolved we need to turn this off in
+     order to use CsWin32 in both System.Drawing and System.Windows.Forms.
+    -->
+    <NoWarn>$(NoWarn);CS1574;CS1580</NoWarn>
```
sharwell, line 27: ❓ Are you able to work around this issue using the approach described here?
https://github.com/dotnet/winforms/pull/10320#issuecomment-1874591789

JeremyKuhne: Can't specify the root namespace for CsWin32 currently.

sharwell: I believe you may have misunderstood my comment. My proposed solution is only relevant for cases where you cannot specify the root namespace.

JeremyKuhne: > My proposed solution is only relevant for cases where you cannot specify the root namespace.

@sharwell that is every case afaik

src/System.Drawing.Common/src/System/Drawing/Printing/PageSettings.cs
```diff
@@ -203,14 +202,12 @@ public PrinterResolution PrinterResolution
                 return _printerResolution;
             }
 
-            IntPtr modeHandle = _printerSettings.GetHdevmode();
-            IntPtr modePointer = Kernel32.GlobalLock(new HandleRef(this, modeHandle));
-            Gdi32.DEVMODE mode = Marshal.PtrToStructure<Gdi32.DEVMODE>(modePointer)!;
+            HGLOBAL modeHandle = (HGLOBAL)_printerSettings.GetHdevmode();
```
lonitra, line 205: seems like we are freeing this a lot. May want to consider adding a scope in the future 🤔

lonitra (approved):
👍

src/System.Drawing.Common/src/System/Drawing/Printing/PageSettings.cs
JeremyKuhne: Definitely intend to :)

src/System.Windows.Forms/tests/IntegrationTests/ScratchProject/Form1.cs
```diff
@@ -9,8 +9,18 @@ namespace ScratchProject;
 [DesignerCategory("Default")]
 public partial class Form1 : Form
 {
+    private RichTextBox _textBox;
+
     public Form1()
     {
         InitializeComponent();
+
+        _textBox = new RichTextBox
```
Tanya-Solyanik, line 18: @JeremyKuhne  - please remove the content, we all use this project for random stuff, it's purpose was to stay empty for easy testing.

JeremyKuhne: Sorry, missed it

PR #10529, opened 2023-12-25
Scale icon to Dpi when initializing icon in ThreadExceptionDialog

### LINKED ISSUE #10512: The expand/collapse arrow icon is double-scaled in the ThreadExceptionDialog
### .NET version

9.0.100-alpha.1.23620.7

Not tested/verified

Yes, it works well on .NET 8.0.

### Issue description

The down arrow icon align with Details is not tangible in Thread exception dialog 

## Repro steps:
1. Git clone  https://github.com/dotnet/winforms
2. Open CMD from clone folder and run "start-vs.cmd" command open project
3. Select WinformsControlTest app, right click it to set as startup project
4. Build and run, click  "Dialog" button -> "Thread exception dialog" button

## Actual result:  
The down arrow icon align with Details is not tangible.

image

## Expect result: 
The down arrow icon align with Details is tangible.
image

## More Info:
1. This issue exists in DPI 100%.

LeafShi1:
Fixes #10512 

## Proposed changes

-   Adding methods `GetLargeIconResourceAsBitmap `and `GetSmallIconResourceAsBitmap `to ScaleHelper.cs to 100% scaling the icon resource.

## The cause of this issue
-  In the original logic, the resource was scaled once when calling `GetSystemMetrics(SM_CXICON)` (this is not expected), and was scaled again when calling `ScaleToDpi`. The two scalings resulted in abnormal icon display.

## Customer Impact

- The down arrow icon align with Details in Thread exception dialog can be show normally

## Regression? 

- Yes 

## Risk

- Minimal

### Before

In 200% screen:
image

### After
image

## Test methodology 

- Manually
 

## Test environment(s) 

- .net 9.0.0-alpha.1.23623.1

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10529)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/InheritanceUI.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/InheritanceUI.cs
@@ -16,6 +16,7 @@ internal class InheritanceUI
     private static Bitmap s_inheritanceGlyph;
     private static Rectangle s_inheritanceGlyphRect;
     private ToolTip _tooltip;
+    private const int IconSize = 16;
 
     /// <summary>
     ///  The bitmap we use to show inheritance.
@@ -23,7 +24,7 @@ internal class InheritanceUI
     public static Bitmap InheritanceGlyph => s_inheritanceGlyph ??= ScaleHelper.GetIconResourceAsBitmap(
         typeof(InheritanceUI),
         "InheritedGlyph",
-        ScaleHelper.InitialSystemDpi);
+        ScaleHelper.ScaleToDpi(new Size(IconSize, IconSize), ScaleHelper.InitialSystemDpi));
 
     /// <summary>
     ///  The rectangle surrounding the glyph.
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/InheritanceUI.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/InheritanceUI.cs
@@ -16,15 +16,14 @@ internal class InheritanceUI
     private static Bitmap s_inheritanceGlyph;
     private static Rectangle s_inheritanceGlyphRect;
     private ToolTip _tooltip;
-    private const int IconSize = 16;
 
     /// <summary>
     ///  The bitmap we use to show inheritance.
     /// </summary>
-    public static Bitmap InheritanceGlyph => s_inheritanceGlyph ??= ScaleHelper.GetIconResourceAsBitmap(
+    public static Bitmap InheritanceGlyph => s_inheritanceGlyph ??= ScaleHelper.GetSmallIconResourceAsBitmap(
         typeof(InheritanceUI),
         "InheritedGlyph",
-        ScaleHelper.ScaleToDpi(new Size(IconSize, IconSize), ScaleHelper.InitialSystemDpi));
+        ScaleHelper.InitialSystemDpi);
 
     /// <summary>
     ///  The rectangle surrounding the glyph.
--- a/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/ScaleHelper.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/ScaleHelper.cs
@@ -26,6 +26,7 @@ internal static partial class ScaleHelper
 
     // Backing field, indicating that we will need to send a PerMonitorV2 query in due course.
     private static bool s_processPerMonitorAware;
+    private static Size? _logicalSystemIconSize;
 
     /// <summary>
     ///  The initial primary monitor DPI (logical pixels per inch) for the process.
@@ -322,20 +323,28 @@ internal static partial class ScaleHelper
         PInvoke.GetSystemMetrics(SYSTEM_METRICS_INDEX.SM_CXICON),
         PInvoke.GetSystemMetrics(SYSTEM_METRICS_INDEX.SM_CYICON));
 
-    internal static Size LogicalSystemIconSize
+    [NotNull]
+    internal static Size? LogicalSystemIconSize
     {
         get
         {
-            Size logicalSystemIconSize = new(16, 16);
+            if (_logicalSystemIconSize is not null)
+            {
+                return _logicalSystemIconSize;
+            }
 
             if (OsVersion.IsWindows10_1607OrGreater())
             {
-                logicalSystemIconSize = new(
+                _logicalSystemIconSize = new(
                     PInvoke.GetSystemMetricsForDpi(SYSTEM_METRICS_INDEX.SM_CXICON, OneHundredPercentLogicalDpi),
                     PInvoke.GetSystemMetricsForDpi(SYSTEM_METRICS_INDEX.SM_CYICON, OneHundredPercentLogicalDpi));
             }
+            else
+            {
+                _logicalSystemIconSize = new Size(16, 16);
+            }
 
-            return logicalSystemIconSize;
+            return _logicalSystemIconSize;
         }
     }
 
@@ -349,7 +358,7 @@ internal static partial class ScaleHelper
     ///  Gets the given icon resource as a <see cref="Bitmap"/> scaled to the specified dpi.
     /// </summary>
     internal static Bitmap GetIconResourceAsBitmap(Type type, string resource, int dpi)
-        => GetIconResourceAsBitmap(type, resource, ScaleToDpi(LogicalSystemIconSize, dpi));
+        => GetIconResourceAsBitmap(type, resource, ScaleToDpi((Size)LogicalSystemIconSize, dpi));
 
     /// <summary>
     ///  Gets the given icon resource as a <see cref="Bitmap"/> of the given size.
--- a/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/ScaleHelper.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/ScaleHelper.cs
@@ -324,35 +324,23 @@ internal static partial class ScaleHelper
         PInvoke.GetSystemMetrics(SYSTEM_METRICS_INDEX.SM_CXICON),
         PInvoke.GetSystemMetrics(SYSTEM_METRICS_INDEX.SM_CYICON));
 
-    internal static Size LogicalLargeSystemIconSize => s_logicalLargeSystemIconSize ??= OsVersion.IsWindows10_1607OrGreater()
-      ? new(
-          PInvoke.GetSystemMetricsForDpi(SYSTEM_METRICS_INDEX.SM_CXICON, OneHundredPercentLogicalDpi),
-          PInvoke.GetSystemMetricsForDpi(SYSTEM_METRICS_INDEX.SM_CYICON, OneHundredPercentLogicalDpi))
-      : new(32, 32);
-
     internal static Size LogicalSmallSystemIconSize => s_logicalSmallSystemIconSize ??= OsVersion.IsWindows10_1607OrGreater()
-     ? new(
-         PInvoke.GetSystemMetricsForDpi(SYSTEM_METRICS_INDEX.SM_CXSMICON, OneHundredPercentLogicalDpi),
-         PInvoke.GetSystemMetricsForDpi(SYSTEM_METRICS_INDEX.SM_CXSMICON, OneHundredPercentLogicalDpi))
-     : new(16, 16);
+        ? new(
+            PInvoke.GetSystemMetricsForDpi(SYSTEM_METRICS_INDEX.SM_CXSMICON, OneHundredPercentLogicalDpi),
+            PInvoke.GetSystemMetricsForDpi(SYSTEM_METRICS_INDEX.SM_CXSMICON, OneHundredPercentLogicalDpi))
+        : new(16, 16);
 
     /// <summary>
     ///  Gets the given icon resource as a <see cref="Bitmap"/> at the default icon size.
     /// </summary>
-    internal static Bitmap GetIconResourceAsDefaultSizeBitmap(Type type, string resource)
-        => GetIconResourceAsBestMatchBitmap(type, resource, Size.Empty);
-
-    /// <summary>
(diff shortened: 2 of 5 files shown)


src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/InheritanceUI.cs
```diff
@@ -16,14 +16,15 @@ internal class InheritanceUI
     private static Bitmap s_inheritanceGlyph;
     private static Rectangle s_inheritanceGlyphRect;
     private ToolTip _tooltip;
+    private const int IconSize = 16;
```
Tanya-Solyanik, line 19: But this is the standard size (SystemIconSize), it should work correctly, had you seen problems with this code?

src/System.Windows.Forms/src/System/Windows/Forms/Dialogs/ThreadExceptionDialog.cs
```diff
@@ -278,20 +279,26 @@ public ThreadExceptionDialog(Exception t)
...
-            b.SetBounds(_scaledButtonDetailsLeftPadding, buttonTop, _scaledButtonWidth, _scaledButtonHeight);
-            b.Image = _expandImage;
-            b.ImageAlign = ContentAlignment.MiddleLeft;
-            Controls.Add(b);
+            button = _detailsButton;
+
+            _expandImage = ScaleHelper.GetIconResourceAsBitmap(
+                GetType(),
+                DownBitmapName,
+                ScaleHelper.ScaleToDpi(new Size(IconSize, IconSize), ScaleHelper.InitialSystemDpi));
```
Tanya-Solyanik, line 292: What is `PInvoke.GetSystemMetrics(SYSTEM_METRICS_INDEX.SM_CXICON)` returning in this case?

LeafShi1: In 200% screen, the return value is 64, after scaling, the size becomes 192*192
image

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/InheritanceUI.cs
LeafShi1: I couldn't find the UI interface corresponding to this icon to verify this icon, but according to the problem of ThreadExceptionDialog, the code here has the same logic as that there, so I modified it here.

LeafShi1: And the standard size (SystemIconSize) only for 100% screen, if in 200% screen, even if it is standard size, there will still be problems with display errors, as mentioned in another comment.

src/System.Windows.Forms/src/System/Windows/Forms/Dialogs/ThreadExceptionDialog.cs
Tanya-Solyanik: Wow, I didn't expect that GetSystemMetrics(SM_CXICON) does any scaling... There is another API that is supposed to scale sizes https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getsystemmetricsfordpi. But actually GetSystemMetrics does some scaling - https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getsystemmetrics#remarks
>System metrics can vary from display to display.

Great find!

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/InheritanceUI.cs
Tanya-Solyanik: That icon is a design time glyph, you probably need to make changes to our interactive test to see it
1 create a form with a button - Form1
2. build the app
3. create a form that inherits from Form1

image

src/System.Windows.Forms/src/System/Windows/Forms/Dialogs/ThreadExceptionDialog.cs
Tanya-Solyanik: This means that all standard icons are double-scaled. To fix the rootcause we should fix the ScaleHelper API

Tanya-Solyanik: We need a logical system icon size property - 

    internal static Size LogicalSystemIconSize => new(
        PInvoke.GetSystemMetricsForDpi(SYSTEM_METRICS_INDEX.SM_CXICON, OneHundredPercentLogicalDpi),
        PInvoke.GetSystemMetricsForDpi(SYSTEM_METRICS_INDEX.SM_CYICON, OneHundredPercentLogicalDpi));

And it should be used wherever we scale SystemIconSize to DPI, for example `ScaleToDpi(LogicalSystemIconSize, dpi) `

Tanya-Solyanik: But when we load resources or creating an icon, we should use the "old" , scaled for the primary screen `SystemIconSize`

Tanya-Solyanik: >In 200% screen, the return value is 64,

This is strange, I would have expected 32 on 200% screen as the initial value

LeafShi1: The app is System Aware but 200% screen is my secondary screen.

LeafShi1: > internal static Size LogicalSystemIconSize => new(
>     PInvoke.GetSystemMetricsForDpi(SYSTEM_METRICS_INDEX.SM_CXICON, OneHundredPercentLogicalDpi),
>     PInvoke.GetSystemMetricsForDpi(SYSTEM_METRICS_INDEX.SM_CYICON, OneHundredPercentLogicalDpi));

Yes, we can add it, but the "PInvoke.GetSystemMetricsForDpi" is only supported on 'Windows 10.0.14393 and later' 
image

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/InheritanceUI.cs
LeafShi1: What is the ”interactive test“? I don't know how to do this, can you give me more guidance

src/System.Windows.Forms/src/System/Windows/Forms/Dialogs/ThreadExceptionDialog.cs
Tanya-Solyanik: Then this property can be implemented conditional on the OSVersion. If version is below 14393, then return 16x16

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/InheritanceUI.cs
Tanya-Solyanik: I meant the DemoConsole  :), but don't spend too much time on it if too much infrastructure is missing. Just try adding UserControl and the inherit another UserControl from it, and add the second control to the DemoConsole tab

src/System.Windows.Forms/src/System/Windows/Forms/Dialogs/ThreadExceptionDialog.cs
LeafShi1: The return value of `PInvoke.GetSystemMetricsForDpi(SYSTEM_METRICS_INDEX.SM_CXICON, OneHundredPercentLogicalDpi) `is 32. 
image

After executing `ScaleToDpi(LogicalSystemIconSize, dpi)`, the Size becomes 64, and the icon is still twice as big as expected.
image

LeafShi1: And I found another question, I tried update the `ApplicationHighDpiMode `of the `WinformsControlsTest `to `PerMonitorV2`, then move ThreaException dialog from 200% screen to 100% screen, the icon scaling is also incorrect

BeforeChange

After added LogicalSystemIconSize and changed `ScaleToDpi(SystemIconSize, dpi)` to `ScaleToDpi(LogicalSystemIconSize, dpi) `of function `GetIconResourceAsBitmap(Type type, string resource, int dpi)`
AfterChange

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/ScaleHelper.cs
```diff
@@ -322,6 +322,23 @@ internal static int ScaleToDpi(int logicalValue, int dpi)
         PInvoke.GetSystemMetrics(SYSTEM_METRICS_INDEX.SM_CXICON),
         PInvoke.GetSystemMetrics(SYSTEM_METRICS_INDEX.SM_CYICON));
 
+    internal static Size LogicalSystemIconSize
+    {
+        get
+        {
+            Size logicalSystemIconSize = new(16, 16);
```
Tanya-Solyanik, line 329: It is safe to cache the backing field in this case because we always are querying the same value. Also this allocation might not be needed , we can do it only for OSs below the 16070

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/InheritanceUI.cs
LeafShi1: In fact, the existing button "I'm the second button" in Tab1 is a custom inherited control, but the icon "InheritedGlyph" is not displayed.
image

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/ScaleHelper.cs
LeafShi1: > Also this allocation might not be needed , we can do it only for OSs below the 16070

I don't quite understand this. Do you mean that there is no need to make judgments about OSs below the 16070?

src/System.Windows.Forms/src/System/Windows/Forms/Dialogs/ThreadExceptionDialog.cs
Tanya-Solyanik: I would prefer first fixing the regression in System Aware applications, and then moving on to PerMonitor. Did pPerMonitor work in NET8?

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/ScaleHelper.cs
```diff
@@ -332,7 +358,7 @@ internal static Bitmap GetIconResourceAsDefaultSizeBitmap(Type type, string reso
     ///  Gets the given icon resource as a <see cref="Bitmap"/> scaled to the specified dpi.
     /// </summary>
     internal static Bitmap GetIconResourceAsBitmap(Type type, string resource, int dpi)
```
Tanya-Solyanik, line 360: And this method should be changed into 2, one for Large system icons and another one for small system icons. Lost call sites are expecting the small icon, but they should be analyzed anyway

Tanya-Solyanik (commented):
Great findings! Thank you!

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/InheritanceUI.cs
LeafShi1: > Hm, this means that we have missing design time functionality. Please open a bug for this issue, but we'll have to triage it first, do not jump on it, we are implementing design time functionality that was requested by customers, and we didn't get a request for this one yet.

Track this issue by #10607 10607

src/System.Windows.Forms/src/System/Windows/Forms/Dialogs/ThreadExceptionDialog.cs
LeafShi1: In NET8, PerMonitorV2 work fine. Because the logic of getting Icon is different from main
image

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/Internals/ScaleHelper.cs
Tanya-Solyanik: I meant that `new` can be moved towards the end of the function.

src/System.Windows.Forms/src/System/Windows/Forms/Dialogs/ThreadExceptionDialog.cs
```diff
@@ -329,8 +335,8 @@ public ThreadExceptionDialog(Exception t)
     private void ThreadExceptionDialog_DpiChanged(object? sender, DpiChangedEventArgs e)
     {
         _expandImage?.Dispose();
-        _expandImage = ScaleHelper.GetIconResourceAsBitmap(GetType(), DownBitmapName, DeviceDpi);
-        _collapseImage = ScaleHelper.GetIconResourceAsBitmap(GetType(), UpBitmapName, DeviceDpi);
+        _expandImage = ScaleHelper.GetSmallIconResourceAsBitmap(GetType(), DownBitmapName, DeviceDpi);
```
Tanya-Solyanik, line 338: I see, this dialog is supporting PerMonitor DPI mode.

elachlan:
You probably have already address this, but I ran into this issue when I entered an incorrect value into a propertygrid:
image

Tanya-Solyanik:
@elachlan  - yes, see GridErrorDialog

LeafShi1:
> You probably have already address this, but I ran into this issue when I entered an incorrect value into a propertygrid: image

This has returned to normal
image

Tanya-Solyanik (approved):

Tanya-Solyanik:
@LeafShi1  - please squash-merge this PR

Issue #10462, opened 2023-12-11
TabControl ImageList adds a white background in .NET8


Monotek18:
### .NET version

.NET8

### Did it work in .NET Framework?

Not tested/verified

### Did it work in any of the earlier releases of .NET Core or .NET 5+?

Yes, .NET7

### Issue description

When upgrading my Winforms project from .NET7 to .NET8 (while changing nothing to the code) I noticed that the images dynamically added to the tab through an `ImageList` add a blank background.

.NET7:

![net7](https://github.com/dotnet/winforms/assets/20025273/64130185-7654-4a47-9c8e-d64d4edeecae)

.NET8:
![net8](https://github.com/dotnet/winforms/assets/20025273/02aa8bc1-196d-4236-b5b4-aa6a252275bf)

### Steps to reproduce

The following code is used for both of the screenshots above:

```csharp
ImageList iconsList = new()
{
    TransparentColor = Color.Transparent,
    ColorDepth = ColorDepth.Depth32Bit,
    ImageSize = new Size(16, 16)
};
iconsList.Images.Add(Resources.GambitTabControl_0);
iconsList.Images.Add(Resources.GambitTabControl_1);
iconsList.Images.Add(Resources.GambitTabControl_2);
iconsList.Images.Add(Resources.GambitTabControl_3);
iconsList.Images.Add(Resources.GambitTabControl_4);
iconsList.Images.Add(Resources.GambitTabControl_5);
TBC_Gambit.ImageList = iconsList;
```

(Note: I am not sure if I need to update the code/UI or if this an issue with Winforms since .NET8, but I did not see anything regarding changes to `TabControl`)

elachlan:
We don't have test coverage in the `WinformsControlsTest` test project. All the uses of `ImageList` use a white background.

I managed to reproduce with the treelist test by setting the backcolor to lightgray:
![image](https://github.com/dotnet/winforms/assets/2433737/22b3f5db-7805-4dd9-b793-47dca258a726)

@Monotek18 what image type are you using?

Monotek18:
Hello @elachlan,

I use PNG images. (with transparent background if that matter)

JeremyKuhne:
@Monotek18 would you mind sharing an image you're actually using so we can make sure we correctly address this?

Monotek18:
@JeremyKuhne here are the images I used above:
[GambitTabControl.zip](https://github.com/dotnet/winforms/files/13671500/GambitTabControl.zip)

[...]

PR #10536, opened 2023-12-27
dispose graphics in TabControl.WmReflectDrawItem()

### LINKED ISSUE #10534: Memory leak in TabControlWmReflectDrawItem(), dispose not called on graphics
### .NET version

8.0

Not tested/verified

unknown

### Issue description

See code below.  The `Graphics` object created by `dis->hDC.CreateGraphics()` is not disposed (neither via `DrawItemEventArgs.Dispose()` nor `WmReflectDrawItem()`).  This results in the object having its finalizer run.

https://github.com/dotnet/winforms/blob/9e2fc95fca4e6d18444c819b35bb1338ac6ca63d/src/System.Windows.Forms/src/System/Windows/Forms/Controls/TabControl/TabControl.cs#L1935

```
    private unsafe void WmReflectDrawItem(ref Message m)
    {
        DRAWITEMSTRUCT* dis = (DRAWITEMSTRUCT*)(nint)m.LParamInternal;

        using DrawItemEventArgs e = new(
            dis->hDC.CreateGraphics(),
            Font,
            dis->rcItem,
            (int)dis->itemID,
            (DrawItemState)(int)dis->itemState);

        OnDrawItem(e);

        m.ResultInternal = (LRESULT)1;
    }
```

### Steps to reproduce

n/a

jgilbert2017:
fixes #10534

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10536)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Controls/TabControl/TabControl.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Controls/TabControl/TabControl.cs
@@ -1933,11 +1933,13 @@ public partial class TabControl : Control
         DRAWITEMSTRUCT* dis = (DRAWITEMSTRUCT*)(nint)m.LParamInternal;
 
         using DrawItemEventArgs e = new(
-            dis->hDC.CreateGraphics(),
+            dis->hDC,
             Font,
             dis->rcItem,
-            (int)dis->itemID,
-            (DrawItemState)(int)dis->itemState);
+            dis->itemID,
+            dis->itemState,
+            ForeColor,
+            BackColor);
 
         OnDrawItem(e);
 


Tanya-Solyanik (approved):

jgilbert2017:
@dotnet-policy-service agree

Tanya-Solyanik (approved):
There ae more instances of CreateGraphicsInternal that could be fixed. Would you be able to take care of them as well?

jgilbert2017:
@Tanya-Solyanik sure.  can you point them out?  i checked ListControl and ComboBox that @kirsan31 mentioned and the bad code doesn't exist on the main branch.

Tanya-Solyanik:
List box is correct in a sense that it has a try/finally, but a `using` would simplify the code.

ComponentEditorForm, ThreadExceptionDialog, SplitterContainer do not have try/finally.

Tanya-Solyanik:
@jgilbert2017  - what version of .NET do you need this fixed in? Does this affect your production code?

jgilbert2017:
@Tanya-Solyanik a net8 patch would be ideal :-).  i've gotten sporadic production OOM exceptions due to the tab control lack of Dispose() call.

jgilbert2017:
@Tanya-Solyanik lmk know what you think i should do.  i'd be happy to create another PR to do a more thorough cleanup.

Tanya-Solyanik:
@jgilbert2017  - 
Would you be able to confirm that this change fixes your sporadic OOM exceptions? Run your app on a private version built off release\8.0 in test environment perhaps? I would have expected your application to freeze for the time it takes the finalizer to run, due to this bug.

I would prefer:
1. a very targeted PR with the fix (i.e. your first version) and a unit test against Main. 
2. PR with the same change against release/8.0
3. a follow up PR against main that enforces the `using` pattern for other instances of CreateGraphics* methods

kirsan31:
> i've gotten sporadic production OOM exceptions due to the tab control lack of Dispose() call.

As I mentioned [here](https://github.com/dotnet/winforms/issues/10534#issuecomment-1870148689) you should not have OOM (or other leaks problems) due to `Graphics.Dispose()` not being called. Because `Graphics` have a finalizer. Unnecessary finalization must be only a performance problem. Unless of course we have an error in the garbage collector logic 🙄

jgilbert2017:
@kirsan31 graphics handles are limited resources and finalization may be arbitrarily delayed causing a bunch of objects to be pending finalization and sometimes i like to switch tabs very quickly ;-).

jgilbert2017:
@kirsan31 just to be clear, the OOME is not due to being out of memory, it's due to failure to allocate a new handle.

kirsan31:
> @kirsan31 just to be clear, the OOME is not due to being out of memory, it's due to failure to allocate a new handle.

So you are reaching GDI/USER objects limit - make sense... To be sure that we caught the root - this problem can (mast) be solved with manual triggering GC?

kirsan31:
Another thing that worries me is that in the [original version](https://github.com/dotnet/winforms/commit/e9fc1dac4f9792856f0bc8932486303dd8a8fe02#diff-edd2fb85653c19f4c380ec1ae98b35d6723ca3ecb80ef29a4be708f83fa30463L1949) we pass HDC - not Graphics. For `ComboBox` and `ListBox` this [was fixed](https://github.com/dotnet/winforms/issues/10534#issuecomment-1870591253) later. Bat not for `TabControl`.
Passing HDC will result:
https://github.com/dotnet/winforms/blob/d4caca7a596df9e6d2c4854fb6f14f284ec6ad0c/src/System.Windows.Forms/src/System/Windows/Forms/Rendering/DrawingEventArgs.cs#L99-L113
And `CreateGraphics`:
https://github.com/dotnet/winforms/blob/d4caca7a596df9e6d2c4854fb6f14f284ec6ad0c/src/System.Drawing.Common/src/System/Drawing/Graphics.cs#L172-L176

I think the code is a little different 🤔 Could this change lead to undesirable consequences, for example #10462?
In the end, I would [agree with @elachlan](https://github.com/dotnet/winforms/issues/10534#issuecomment-1870644909) and return the original version with HDC...

Tanya-Solyanik:
@kirsan31 @elachlan - Original version was using graphics - https://referencesource.microsoft.com/#System.Windows.Forms/winforms/Managed/System/WinForms/TabControl.cs,5780e62ffb386499
[PR9309](https://github.com/dotnet/winforms/commit/e9fc1dac4f9792856f0bc8932486303dd8a8fe02#diff-edd2fb85653c19f4c380ec1ae98b35d6723ca3ecb80ef29a4be708f83fa30463) was undoing what [PR3553](https://github.com/dotnet/winforms/commit/a6a308fa452770caafca55ee4d8ac54d1ee1015d#diff-edd2fb85653c19f4c380ec1ae98b35d6723ca3ecb80ef29a4be708f83fa30463) had introduced. Keeping graphics seems correct.

kirsan31:
@Tanya-Solyanik 
> [PR9309](https://github.com/dotnet/winforms/commit/e9fc1dac4f9792856f0bc8932486303dd8a8fe02#diff-edd2fb85653c19f4c380ec1ae98b35d6723ca3ecb80ef29a4be708f83fa30463) was undoing what [PR3553](https://github.com/dotnet/winforms/commit/a6a308fa452770caafca55ee4d8ac54d1ee1015d#diff-edd2fb85653c19f4c380ec1ae98b35d6723ca3ecb80ef29a4be708f83fa30463) had introduced. Keeping graphics seems correct.

Then we need to restore `CreateGraphic` for `ComboBox` and `ListBox`. Before PR3553 all of them (`ComboBox`, `ListBox`, `TabControl`) were using `CreateGraphic`.

jgilbert2017:
The comment below is on the ctor that takes the HDC.  The graphics object is ultimately passed to `OnDrawItem` which can be implemented by the user and subject to breaking the assumption below.  This would seem to imply the ctor that takes Graphics is the one to use.

```
    /// <summary>
    ///  For internal use to improve performance. DO NOT use this method if you modify the Graphics Clip or Transform.
    /// </summary>
    internal Graphics GetOrCreateGraphicsInternal(Action<Graphics>? creationAction = null)
```

elachlan:
@JeremyKuhne I think can comment on if we should be passing the device context or graphics here.

Jeremy did the work refactoring away from graphics to device context (HDC) to avoid creating graphics unnecessarily to improve performance. This is to do with GDI being more performant than GDI+

kirsan31:
> This would seem to imply the ctor that takes Graphics is the one to use.

Yea we discussed this with @Tanya-Solyanik above - originally there were `CreateGraphic`. Then I suggested revert `ComboBox` and `ListBox` to `CreateGraphic` too.

Tanya-Solyanik:
@kirsan31  

>Then I suggested revert ComboBox and ListBox to CreateGraphic too.

There is an issue of pallet to be double-checked. 

Original code was setting up a halftone pallet on the DC before creating the Graphics object.
EventArg constructor that takes the DC is still doing that - [here](https://github.com/dotnet/winforms/blob/58ffd1a410d00f7f5b6e609ddf88114117c0e1c0/src/System.Windows.Forms/src/System/Windows/Forms/Rendering/DrawingEventArgs.cs#L93). 
CreateGraphics route is not setting up this pallet. So the question is to @JeremyKuhne  - why did he move back to CreateGraphics here - [PR9309](https://github.com/dotnet/winforms/commit/e9fc1dac4f9792856f0bc8932486303dd8a8fe02#diff-edd2fb85653c19f4c380ec1ae98b35d6723ca3ecb80ef29a4be708f83fa30463)

JeremyKuhne:
Sorry, I've been out.

The simple answer is that I make mistakes. :) Intellisense gave me significant grief when I was converting this so I'm not completely surprised that I messed this up.

Creating the `DrawingEventArgs` via `HDC` is _always_ preferable if you haven't created the `Graphics` object first. `IGraphicsHdcProvider` is then able to unwrap the `HDC` when drawing via GDI internally without creating the `Graphics` object.

src/System.Windows.Forms/src/System/Windows/Forms/Controls/ComboBox/ComboBox.cs
```diff
@@ -3566,7 +3566,7 @@ private void UpdateItemHeight()
         else if (DrawMode == DrawMode.OwnerDrawVariable)
         {
             PInvoke.SendMessage(this, PInvoke.CB_SETITEMHEIGHT, (WPARAM)(-1), (LPARAM)ItemHeight);
-            Graphics graphics = CreateGraphicsInternal();
+            using Graphics graphics = CreateGraphicsInternal();
```
JeremyKuhne, line 3569: I don't know how often consuming code never uses the `Graphics`. Possible performance optimization in the future to make this lazy in `MeasureItemEventArgs` and make the args disposable.

src/System.Windows.Forms/src/System/Windows/Forms/Controls/TabControl/TabControl.cs
```diff
@@ -1936,8 +1936,9 @@ private unsafe void WmReflectDrawItem(ref Message m)
     {
         DRAWITEMSTRUCT* dis = (DRAWITEMSTRUCT*)(nint)m.LParamInternal;
 
+        using var graphics = dis->hDC.CreateGraphics();
         using DrawItemEventArgs e = new(
-            dis->hDC.CreateGraphics(),
+            graphics,
```
JeremyKuhne, line 1941: This really should use the `dis->hDC` in the overload that takes `HDC`.

elachlan:
@jgilbert2017 Did you need anymore guidance on this issue?

jgilbert2017:
@Tanya-Solyanik to recap... i can resubmit a patch
- which branch do you want it against?
- i will use the dis->hDC ctor() per @JeremyKuhne 's comment

elachlan:
@jgilbert2017 against main branch please.

Tanya-Solyanik:
>i will use the dis->hDC ctor() per @JeremyKuhne 's comment

yes, but there will be several places that regressed in [9309](https://github.com/dotnet/winforms/pull/9309)

>@jgilbert2017 against main branch please.

yes

jgilbert2017:
@Tanya-Solyanik can you point them out? I'm only aware of the TabControl issue. i checked ListControl and ComboBox that @kirsan31 mentioned and the bad code doesn't exist on the main branch.

kirsan31:
@jgilbert2017
If I understand right (we need hDC everywhere) so only TabControl need to fixed.

Tanya-Solyanik:
>@Tanya-Solyanik can you point them out? 

I only looked at the PR, yes, combobox and listbox are fixed in main

Tanya-Solyanik (approved):

PR #10538, opened 2023-12-27
Avoid finalization in Cursor.cs if freeing handle not needed

jgilbert2017:
call GC.SuppressFinalize(this) for constructors where we don't own the cursor handle so we can avoid being queued for finalization.

Fixes https://github.com/dotnet/winforms/issues/10539

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10538)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Input/Cursor.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Input/Cursor.cs
@@ -56,7 +56,7 @@ public sealed class Cursor : IDisposable, ISerializable, IHandle<HICON>, IHandle
     /// </summary>
     public Cursor(IntPtr handle)
     {
-        GC.SuppressFinalize(this);    
+        GC.SuppressFinalize(this);
         if (handle == 0)
         {
             throw new ArgumentException(string.Format(SR.InvalidGDIHandle, (typeof(Cursor)).Name), nameof(handle));
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Input/Cursor.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Input/Cursor.cs
@@ -46,7 +46,7 @@ public sealed class Cursor : IDisposable, ISerializable, IHandle<HICON>, IHandle
     internal Cursor(string resource, string cursorsProperty)
         : this(typeof(Cursors).Assembly.GetManifestResourceStream(typeof(Cursor), resource).OrThrowIfNull())
     {
-        GC.SuppressFinalize(this);    
+        GC.SuppressFinalize(this);
         CursorsProperty = cursorsProperty;
         _freeHandle = false;
     }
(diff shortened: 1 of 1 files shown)


Tanya-Solyanik:
@jgilbert2017  - please create a new issue with description of the impact of this issue.

src/System.Windows.Forms/src/System/Windows/Forms/Input/Cursor.cs
```diff
@@ -33,6 +33,7 @@ public sealed class Cursor : IDisposable, ISerializable, IHandle<HICON>, IHandle
 
     internal unsafe Cursor(PCWSTR nResourceId, string cursorsProperty)
     {
+        GC.SuppressFinalize(this);
```
Tanya-Solyanik, line 36:
```suggestion
        GC.SuppressFinalize(this);
```

src/System.Windows.Forms/src/System/Windows/Forms/Input/Cursor.cs
```diff
@@ -45,6 +46,7 @@ internal unsafe Cursor(PCWSTR nResourceId, string cursorsProperty)
     internal Cursor(string resource, string cursorsProperty)
         : this(typeof(Cursors).Assembly.GetManifestResourceStream(typeof(Cursor), resource).OrThrowIfNull())
     {
+        GC.SuppressFinalize(this);
```
Tanya-Solyanik, line 49:
```suggestion
        GC.SuppressFinalize(this);
```

src/System.Windows.Forms/src/System/Windows/Forms/Input/Cursor.cs
```diff
@@ -54,6 +56,7 @@ internal Cursor(string resource, string cursorsProperty)
     /// </summary>
     public Cursor(IntPtr handle)
     {
+        GC.SuppressFinalize(this);
```
Tanya-Solyanik, line 59:
```suggestion
        GC.SuppressFinalize(this);
```

Tanya-Solyanik:
Looks good, please add a unit test

jgilbert2017:
> Looks good, please add a unit test

What what you like the unit test to do?  The change prevents a call to ~Cursor().  I'm not sure how I would test for this without things getting very ugly.

kirsan31:
> > Looks good, please add a unit test
> 
> What what you like the unit test to do? The change prevents a call to ~Cursor(). I'm not sure how I would test for this without things getting very ugly.

We can test it from one side (ensure that destructor will not be called when we not need it) and not make _ugly_ things :) Something like [this](https://github.com/dotnet/winforms/pull/10460/files). From other side - yes things become more complicated :(

Tanya-Solyanik (approved):

lonitra (approved):

PR #10552, opened 2023-12-29
Implement screen recording for integration tests

sharwell:
Record the screen during integration tests, and save an animated PNG (*.apng) if the test fails. Derived from a similar implementation in dotnet/roslyn.

Integration tests will be re-enabled for CI as a follow-up to this.

The APNG code is derived from code originally written for Roslyn in #70983, with significant modifications to make it operate on WinForms types instead of WPF types.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10552)

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -116,6 +116,10 @@ Note: if the Uri is a new place, you will need to add a subscription from that p
       <Uri>https://github.com/dotnet/runtime</Uri>
       <Sha>9f6debd3d9a4e8ccfd53845c6d268daab1ab7a2f</Sha>
     </Dependency>
+    <Dependency Name="System.IO.Hashing" Version="9.0.0-preview.2.24081.10">
+      <Uri>https://github.com/dotnet/runtime</Uri>
+      <Sha>9f6debd3d9a4e8ccfd53845c6d268daab1ab7a2f</Sha>
+    </Dependency>
     <Dependency Name="System.IO.Packaging" Version="9.0.0-preview.2.24081.10">
       <Uri>https://github.com/dotnet/runtime</Uri>
       <Sha>9f6debd3d9a4e8ccfd53845c6d268daab1ab7a2f</Sha>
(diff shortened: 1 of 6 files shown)


eng/Version.Details.xml
```diff
@@ -116,6 +116,10 @@ Note: if the Uri is a new place, you will need to add a subscription from that p
       <Uri>https://github.com/dotnet/runtime</Uri>
       <Sha>b4ba5da5a0b8e0c7e3027a695f2acb2d9d19137b</Sha>
     </Dependency>
+    <Dependency Name="System.IO.Hashing" Version="9.0.0-alpha.1.23628.3">
```
sharwell, line 119: 📝 I manually edited this file to add this. I'm not sure if anything else needs to be changed to ensure it gets updated along with the other packages.

JeremyKuhne (changes requested):
@sharwell spoke with @lonitra offline. We're a little concerned about maintaining our own screen recording code. We'd be more comfortable using existing functionality. One example (from a cursory look) is https://github.com/sskodje/ScreenRecorderLib.

Can you take a look at options here? Thanks!

sharwell:
@JeremyKuhne I believe the only thing that would need maintenance here is the screenshot timing code, and this would need to be maintained regardless of the solution you go with:
https://github.com/dotnet/winforms/blob/1b508b1bea081bb8ed669b81fa2002e53cf5c410/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/ScreenRecordService.cs#L560-L610

In theory the compression strategy could be further improved, but in practice I'm seeing saved image sizes that are acceptably small even on high resolution displays.

JeremyKuhne:
> I believe the only thing that would need maintenance here is the screenshot timing code

All of the code needs to be maintained. The manual PNG manipulation code is a concern.

> but in practice I'm seeing saved image sizes that are acceptably small even on high resolution displays.

What is "acceptably small"?

sharwell:
> > but in practice I'm seeing saved image sizes that are acceptably small even on high resolution displays.
> 
> What is "acceptably small"?

Some of the longer Roslyn tests produced 300+ images @ 4K resolution in a test, and the saved files were 5-7MB. The shorter duration and lower resolution files for CI were <1MB. WinForms seemed to come in a bit smaller despite having a higher image capture rate because the tests are particularly short.

JeremyKuhne (dismissed):
Approved after further internal discussion with @lonitra

lonitra:
@sharwell seems that there are conflicts. Could you rebase this so we can get this merged?

sharwell:
This is now updated. Note that if another automated dependency merge occurs before merging this, it will need to be rebased again due to the introduction of the System.IO.Hashing dependency.

PR #10555, opened 2023-12-29
Stabilize ImageList_FinalizerReleasesNativeHandle_ReturnsExpected

### LINKED ISSUE #6635: Re-enable skipped test: ImageList_FinalizerReleasesNativeHandle_ReturnsExpected
* .NET Core Version: 7.0

**Problem description:**
- `ImageList_FinalizerReleasesNativeHandle_ReturnsExpected` unit test fails sometime in x86 Debug CI build for successful PRs, that blocking deployment of fixes into sdk.

**Stack trace** ([System.Windows.Forms.UI.IntegrationTests_net7.0_x86.log](https://dev.azure.com/dnceng/public/_build/results?buildId=1596505&view=artifacts&pathAsName=false&type=publishedArtifacts))
```
System.Windows.Forms.UITests.ImageListTests.ImageList_FinalizerReleasesNativeHandle_ReturnsExpected [FAIL]
      Assert.Equal() Failure
      Expected: 32
      Actual:   31
      Stack Trace:
        /_/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ImageListTests.cs(47,0): at System.Windows.Forms.UITests.ImageListTests.ImageList_FinalizerReleasesNativeHandle_ReturnsExpected()
      Output:
        GDI handles before: 32
        GDI handles after: 31
```
**Expected behavior:**
- All correct tests pass every time.

**Minimal repro:**
- Run unit tests on x86 environment (or create a PR with changes, CI build will do it)

sharwell:
Closes #6635

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10555)

### DIFF
--- a/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ImageListTests.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ImageListTests.cs
@@ -3,6 +3,7 @@
 
 using System.ComponentModel;
 using System.Drawing;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using Windows.Win32.System.Threading;
 using Xunit.Abstractions;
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ImageListTests.cs
```diff
@@ -20,48 +21,70 @@ public ImageListTests(ITestOutputHelper testOutputHelper)
...
         Assert.Equal(startGdiHandleCount, endGdiHandleCount);
 
+        [MethodImpl(MethodImplOptions.NoInlining)]
         static uint GetGdiHandles()
         {
-            GC.Collect();
-            GC.WaitForPendingFinalizers();
-            GC.Collect(0);
+            uint result = GetGdiHandlesOnce();
+            while (true)
```
Tanya-Solyanik, line 49: I would prefer to have an additional exit condition based on the number of attempts, i.e. while (i++ < 10000) and report the timeout error with a descriptive message.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ImageListTests.cs
```diff
@@ -20,48 +21,70 @@ public ImageListTests(ITestOutputHelper testOutputHelper)
...
+                }
+
+                result = updatedResult;
+            }
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static uint GetGdiHandlesOnce()
+        {
+            GC.GetTotalMemory(true);
```
Tanya-Solyanik, line 65: Is this the preferred way to force GC.Collect? How many generations does it collect?

sharwell: > Is this the preferred way to force GC.Collect?

As best I can tell, yes.

> How many generations does it collect?

It doesn't have a fixed limit. It runs collections until the available memory stabilizes within a threshold.
https://github.com/dotnet/runtime/blob/b4ba5da5a0b8e0c7e3027a695f2acb2d9d19137b/src/coreclr/System.Private.CoreLib/src/System/GC.CoreCLR.cs#L337-L363

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/ImageListTests.cs
sharwell: The maximum theoretical number of iterations allowed in the current implementation is 65536 (GDI handle limit for the system), but in practice I'm seeing limits around 30-40. Every iteration is required to make progress relative to the previous iteration, so it will never iterate more than the total number of GDI handles in use.

sharwell: Ah, I didn't account for the possibility that something creates a GDI handle during the execution of this method (how would that even happen?). If that seems to be a real concern, I could update it to throw an exception if the number ever increases.

sharwell: Note that running through the test runner should automatically be providing `--blame-hang` as a timeout prevention measure. I'm not sure we need to do anything more than that within the context of a specific test.

Tanya-Solyanik: >something creates a GDI handle during the execution of this method (how would that even happen?)

I was thinking about somebody copy/pasting this code elsewhere or running these tests in parallel. 
Inside the test we can provide a more descriptive error than a standard timeout hung reporting. Or you can add a comment that we assume that no GDI handles are creates and the count only goes down ( to prevent copy/pasting this code)

Tanya-Solyanik: >I could update it to throw an exception if the number ever increases.

This is the best option.

sharwell: ➡️ Updated

Tanya-Solyanik (approved):

PR #10575, opened 2024-01-03
Fix PropertyGrid.RemoveTabType method

### LINKED ISSUE #10574: PropertyGrid.RemoveTabType does not effectively remove tab
### .NET version

8.0.100

Yes

Yes, it works in 5.0

### Issue description

The PropertyGrid.RemoveTabType method does not effectively remove the PropertyTab. Calling this method with a previously added tab does not suppress the tab. It looks like the check on type is wrong in the method RemoveTabType  since the commit https://github.com/dotnet/winforms/pull/6108

The type check is not consistent between AddTabType and RemoveTabType method:
In AddTabType: https://github.com/dotnet/winforms/blob/39c1cfba04f0682ae79e6629188eb291bc0d203e/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGrid.cs#L1480

In RemoveTabType:
https://github.com/dotnet/winforms/blob/39c1cfba04f0682ae79e6629188eb291bc0d203e/src/System.Windows.Forms/src/System/Windows/Forms/PropertyGrid.cs#L3565

### Steps to reproduce

1. Create a new form with a PropertyGrid
2. Create a class MyPropertyTab which inherits from PropertyTab
3. Properly implement MyPropertyTab 
4. Call propertyGrid.AddTabType(typeof(MyPropertyTab))
5. Call propertyGrid.RemoveTabType(typeof(MyPropertyTab))

In .net50, MyPropertyTab is removed
in .net80, MyPropertyTab is not removed

thomasnaletsg:
Tab removal wasn't matching the tab type so it wouldn't remove tabs.

Fixes #10574

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGrid.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGrid.cs
@@ -3564,7 +3564,7 @@ public partial class PropertyGrid : ContainerControl, IComPropertyBrowser, IProp
         int tabIndex = -1;
         for (int i = 0; i < _tabs.Count; i++)
         {
-            if (tabType == _tabs[i].GetType())
+            if (tabType == _tabs[i].Tab.GetType())
             {
                 tabIndex = i;
                 break;
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyGridTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PropertyGridTests.cs
@@ -11,6 +11,7 @@ using System.Windows.Forms.TestUtilities;
 using System.Runtime.CompilerServices;
 using Point = System.Drawing.Point;
 using Size = System.Drawing.Size;
+using System.Windows.Forms.Design;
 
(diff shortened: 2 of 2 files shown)


thomasnaletsg:
Related to issue https://github.com/dotnet/winforms/issues/10574

Tanya-Solyanik:
@thomasnaletsg the fix looks good, could you please add a unit test that removes tabs?

thomasnaletsg:
> @thomasnaletsg the fix looks good, could you please add a unit test that removes tabs?

@JeremyKuhne as an author of the initial refactoring, could you please add a UT covering this use case? Thanks in advance.

lonitra:
@thomasnaletsg I've gone ahead and added a unit test. Please read CLA above.

Tanya-Solyanik (approved):

kirsan31:
@lonitra 
> /backport to release/8.0

And what will be with 6.0 and 7.0?

lonitra:
> And what will be with 6.0 and 7.0?

We think that servicing this to 8.0 will be enough, but if there is a customer request we can bring this to shiproom.

lonitra:
@thomasnaletsg gentle ping. Could you please accept CLA so this can be merged?

thomasnaletsg:
@dotnet-policy-service agree

PR #10577, opened 2024-01-04
Replace `ArrayList` usage in `FlowLayoutPanelDesigner` with List

elachlan:
Related:
- #8140
- #10432

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10577)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/FlowLayoutPanelDesigner .cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/FlowLayoutPanelDesigner .cs
@@ -617,8 +617,7 @@ internal partial class FlowLayoutPanelDesigner : FlowPanelDesigner
             }
 
             // We use this list when doing a Drag-Copy, so that we can correctly restore state when we are done.
-            // List<Control> originalControls = new();
-            ArrayList originalControls = new();
+            List<Control> originalControls = [];
 
             // Remove the controls in the drag collection - don't need to do this if we are copying.
             if (!performCopy)
@@ -637,8 +636,7 @@ internal partial class FlowLayoutPanelDesigner : FlowPanelDesigner
             else
             {
                 // We are doing a copy, so let's copy the controls.
-                List<Control> tempList = new();
-                tempList.AddRange(_dragControls);
+                List<Control> tempList = [.. _dragControls];
 
                 DesignerUtils.CopyDragObjects(tempList, Component.Site);
 
@@ -656,10 +654,10 @@ internal partial class FlowLayoutPanelDesigner : FlowPanelDesigner
                     // Remember to set the new primary control.
                     if (_primaryDragControl.Equals(_dragControls[j]))
                     {
-                        _primaryDragControl = tempList[j] as Control;
+                        _primaryDragControl = tempList[j];
                     }
 
-                    _dragControls[j] = tempList[j] as Control;
+                    _dragControls[j] = tempList[j];
                 }
             }
 
@@ -705,7 +703,7 @@ internal partial class FlowLayoutPanelDesigner : FlowPanelDesigner
             {
                 for (int i = 0; i < originalControls.Count; i++)
                 {
-                    _dragControls[i] = (Control)originalControls[i];
+                    _dragControls[i] = originalControls[i];
                 }
             }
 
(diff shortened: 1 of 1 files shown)


elachlan:
@halgab it looks like we missed one in #8140.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/FlowLayoutPanelDesigner .cs
```diff
@@ -637,8 +636,7 @@ private void ReorderControls(DragEventArgs de)
             else
             {
                 // We are doing a copy, so let's copy the controls.
-                List<Control> tempList = new();
-                tempList.AddRange(_dragControls);
+                List<Control> tempList = [.. _dragControls];
 
                 DesignerUtils.CopyDragObjects(tempList, Component.Site);
```
elachlan, line 641: @lonitra this line is still a bit strange. Unsure if it was ever followed up:
https://github.com/dotnet/winforms/pull/10432#discussion_r1426233245

lonitra: Thanks for bringing this to my attention again! Did some quick investigation here and in Framework, this was in fact `tempList = DesignerUtils.CopyDragObjects(tempList, Component.Site);`. 
Looking at the history here, `FlowLayoutPanelDesigner` was recently ported to core as of .NET 9, which is likely why we have not received any issues revolving around this if it did cause any regressions. It seems like it may have been a mistake that `tempList =` was removed. @LeafShi1 can you confirm this? We can go ahead and fix this as part of this PR if so

elachlan: Done.

LeafShi1: Thank you for spotting this issue, it does look like a bug.

elachlan: It was all @halgab , I was just following up.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/FlowLayoutPanelDesigner .cs
```diff
@@ -637,10 +636,9 @@ private void ReorderControls(DragEventArgs de)
             else
             {
                 // We are doing a copy, so let's copy the controls.
-                List<Control> tempList = new();
-                tempList.AddRange(_dragControls);
+                List<IComponent> tempList = [.. _dragControls];
 
-                DesignerUtils.CopyDragObjects(tempList, Component.Site);
+                tempList = DesignerUtils.CopyDragObjects(tempList, Component.Site);
```
halgab, line 641:
```suggestion
                List<IComponent> tempList = DesignerUtils.CopyDragObjects(_dragControls, Component.Site);
```
and remove the line above?

elachlan: I am unsure what is correct. Other usages have a similar pattern of creating a temp list and passing it in. But I think they could be wrong as well.

elachlan: Just tested it in DemoConsole and it looks like it works fine. So we will go with that.

elachlan: @halgab did you want to investigate the rest of the usages of `DesignerUtils.CopyDragObjects` to make sure they aren't creating a list unnecessarily?

lonitra (approved):
🚀

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/FlowLayoutPanelDesigner .cs
halgab: I've just checked, and it was the only such occurence. `DesignerUtils.CopyDragObjects` always returns a new list by the way. What I have my eye on is changing this method's signature to accept something more strongly typed than `ICollection`. Moving `ToolStripItemDataObject.DragComponents` off `ArrayList` is a prerequisite though.

halgab: Scratched my own itch: #10626

elachlan: I had a quick look, it seems possible to narrow everything to `IEnumerable<IComponent>`, but `CopyDragObjects ` would need a little bit of refactoring to avoid calling `Count()` multiple times.

`ToolStripItemDataObject.DragComponents` might be an issue depending on serialization, it will need to be tested. I suggest that change goes into its own PR gets thoroughly tested.

PR #10585, opened 2024-01-04
Fix calls to Component in ComponentDesigner.Dispose overrides

halgab:
In #9408, we forced the `ComponentDesigner.Component` property to be non-null. We noticed this had implications in the method `ControlDesigner.Dispose`. This PR addresses all other such cases that were broken by this PR but 2 instances that will be addressed in followup PRs.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10585)


lonitra (approved):
Thank you!

PR #10625, opened 2024-01-09
Disable Removing Milestones from Closed Issues

lonitra:
We want to stop the bot from removing milestones off of closed issues to better track what work has been done in different previews.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10625)

### DIFF
--- a/.github/fabricbot.json
+++ b/.github/fabricbot.json
@@ -919,7 +919,8 @@
             "parameters": {}
           }
         ]
-      }
+      },
+      "disabled": true
     },
     {
       "taskType": "trigger",


sharwell:
I would agree with disabling that automation. Ideally, when an issue is automatically closed as a result of merging a PR, the PR and issue should both be given a milestone of when that change will ship.

It would also be good to make sure that stale/duplicate issues get closed as "not planned". FabricBot likes to close issues as completed even when that is not correct. GitHub Actions is more capable in this regard, and has the additional benefit of being publicly usable.

Tanya-Solyanik:
Just curious, why disabling and not removing the whole thing?

Tanya-Solyanik (approved):

lonitra:
>Ideally, when an issue is automatically closed as a result of merging a PR, the PR and issue should both be given a milestone of when that change will ship.

Yep, this is exactly what we want to do. Unfortunately this doesn't seem so easily achieved automatically with fabric bot, so the plan is to cross check to make sure issue/PR milestones match.

> It would also be good to make sure that stale/duplicate issues get closed as "not planned". FabricBot likes to close issues as completed even when that is not correct. GitHub Actions is more capable in this regard, and has the additional benefit of being publicly usable.

Good point! There are plans to migrate off of fabric bot in the works right now. I have not had a chance to explore GitHub Actions very much but it would be good to take advantage of its capability here and maybe it would have some ability to do the above as well. 🤔

lonitra:
> Just curious, why disabling and not removing the whole thing?

I was hesitant to remove because I saw that we had other actions that were just disabled instead of removed. If preferred, I can follow up and remove instead.

PR #10627, opened 2024-01-09
added hash algorithm value to the assembly table in metadata for Accessibility.dll

Tanya-Solyanik:
Other .NET and .NET Framework assemblies have this value set to SHA1. And Accessibility.dll in .NET Framework has the same value. This issue is unlikely to cause any problems, it only looks inconsistent. Most likely we had overlooked this attribute when editing the IL when moving to Core.
Per @marklio this value is used for:
```
•	calculating the hashes for File table metadata. This is used in multi-file and multi-module assembly scenarios, which are rarer (multi-module basically only happens in System.EnterpriseServices and no where else, multi-file is more common, but isn't in play for Accessibility.dll)
•	Calculating the "strong name signature" data used in validating strong names. The content of the file is hashed with that algorithm and then signed with the private key and stored in these bytes, and can be validated using the public key that exists in the Assembly table. Without the algorithm, you can't validate the signature. I haven't validated whether the strong name signature is present and is a SHA1, but that is what I suspect (the strong name signature bytes are non-zero, which suggests that it is appropriately signed).
```

.NET9 before:
image

.NET Framework
image

.NET 9 after
image


marklio (approved):

Philip-Wang01:
Verified this on .NET 9.0.100-alpha.1.24062.11, issue was fixed. Successfully added hash algorithm value to the assembly table in metadata for Accessibility.dll.
image

PR #10628, opened 2024-01-10
Convert Region and Graphics to use CsWin32

JeremyKuhne:
Converts Region and Graphics to mostly use CsWin32 (modulo enumerating metafiles).

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10628)


JeremyKuhne:
VS hangs trying to do the diff on Graphics. 😆 Note that I've done some GC.KeepAlive handling in error helper methods to reduce the number of places I have to type it (notably for the `this` pointer). Most methods are only called once so I'm generally not going to bother creating a second wrapper to keep objects from being collected.

src/System.Drawing.Common/src/misc/DpiHelper.cs
```diff
@@ -25,7 +22,7 @@ internal static class DpiHelper
 
     private static double s_logicalToDeviceUnitsScalingFactorX;
     private static double s_logicalToDeviceUnitsScalingFactorY;
-    private static InterpolationMode s_interpolationMode = InterpolationMode.Invalid;
+    private static Drawing2D.InterpolationMode s_interpolationMode = Drawing2D.InterpolationMode.Invalid;
```
Tanya-Solyanik, line 25: Why didn't you add using System.Drawing.Drawing2D?

src/System.Drawing.Common/src/Interop/Windows/Gdi32/Interop.OffsetViewportOrgEx.cs
```diff
@@ -19,9 +18,9 @@ internal static partial class Gdi32
             IntPtr hdc,
             int x,
             int y,
-            ref Point lppt);
+            ref System.Drawing.Point lppt);
```
Tanya-Solyanik, line 21: What was the reason for this change?

Tanya-Solyanik (commented):
I only took a quick look.

src/System.Drawing.Common/src/misc/DpiHelper.cs
JeremyKuhne: This is resolving conflict. It is included, there are two `InterpolationMode` definitions. It just so happens that these are the exact same names Gdi+ uses.

src/System.Drawing.Common/src/Interop/Windows/Gdi32/Interop.OffsetViewportOrgEx.cs
JeremyKuhne: Conflict. The native GDI+ type is `Point` and `PointF`. (`Rect` and `RectF` are the rectangles)

src/System.Drawing.Common/src/System/Drawing/Graphics.cs
```diff
@@ -359,52 +364,38 @@ public float DpiY
...
 
-            // GDI+ interprets the value of InterpolationMode and sets a value accordingly.
-            // Libgdiplus does not, so do this manually here.
-            switch (value)
-            {
-                case InterpolationMode.Default:
-                case InterpolationMode.Low:
-                    value = InterpolationMode.Bilinear;
-                    break;
-                case InterpolationMode.High:
```
Tanya-Solyanik, line 390: This switch is being replaced by a cast, is this correct?

JeremyKuhne: It was a hack for Unix

src/System.Drawing.Common/src/System/Drawing/Graphics.cs
```diff
@@ -897,748 +782,710 @@ public void DrawArc(Pen pen, int x, int y, int width, int height, int startAngle
...
-            CheckErrorStatus(Gdip.GdipFillPolygonI(
-                new HandleRef(this, NativeGraphics),
-                new HandleRef(brush, brush.NativeBrush),
-                p, points.Length,
-                fillMode));
+            CheckErrorStatus(PInvoke.GdipFillPolygonI(
+                NativeGraphics,
+                (GpBrush*)brush.NativeBrush,
+                (GdiPlus.Point*)p, points.Length,
+                (GdiPlus.FillMode)fillMode));
```
Tanya-Solyanik, line 1318: GC.KeepAlive(brush)?

src/System.Drawing.Common/src/System/Drawing/Graphics.cs
```diff
@@ -897,748 +782,710 @@ public void DrawArc(Pen pen, int x, int y, int width, int height, int startAngle
...
         {
-            CheckErrorStatus(Gdip.GdipFillClosedCurve(
-                new HandleRef(this, NativeGraphics),
-                new HandleRef(brush, brush.NativeBrush),
-                p, points.Length));
+            CheckErrorStatus(PInvoke.GdipFillClosedCurve(
+                NativeGraphics,
+                (GpBrush*)brush.NativeBrush,
+                (GdiPlus.PointF*)p, points.Length));
         }
```
Tanya-Solyanik, line 1426: GC.KeepAlive(brush)?

src/System.Drawing.Common/src/System/Drawing/Graphics.cs
JeremyKuhne: Good catch, I'll add it in my next change

src/System.Drawing.Common/src/System/Drawing/Graphics.cs
JeremyKuhne: Yes, I'll add it in the next change

src/System.Drawing.Common/src/System/Drawing/Region.cs
```diff
@@ -95,241 +105,212 @@ public void Dispose()
     private void Dispose(bool disposing)
     {
 #if FINALIZATION_WATCH
-        Debug.WriteLineIf(!disposing && NativeRegion != IntPtr.Zero, $"""
+        Debug.WriteLineIf(!disposing && NativeRegion is null, $"""
```
lonitra, line 108:
```suggestion
        Debug.WriteLineIf(!disposing && NativeRegion is not null, $"""
```

lonitra (approved):
Some small comments, but these can be addressed in next change

src/System.Drawing.Common/src/System/Drawing/Graphics.cs
```diff
@@ -897,748 +782,710 @@ public void DrawArc(Pen pen, int x, int y, int width, int height, int startAngle
...
         fixed (Point* p = points)
         {
-            CheckErrorStatus(Gdip.GdipFillClosedCurveI(
-                new HandleRef(this, NativeGraphics),
-                new HandleRef(brush, brush.NativeBrush),
-                p, points.Length));
+            CheckErrorStatus(PInvoke.GdipFillClosedCurveI(
+                NativeGraphics,
+                (GpBrush*)brush.NativeBrush,
+                (GdiPlus.Point*)p, points.Length));
```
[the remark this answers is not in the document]
JeremyKuhne: I'll double check the KeepAlives in the next PR.

src/System.Drawing.Common/src/System/Drawing/Region.cs
JeremyKuhne: Thanks, I'll get it in the next. I have a feeling FINALIZAION_WATCH has other breaks. I'd like to make it more general too. Maybe see if we can add a base class in DEBUG only.

Tanya-Solyanik (approved):

PR #10639, opened 2024-01-10
Ignore System.Private.Windows.Core in FileClassification

lonitra:
This helps with failures in https://github.com/dotnet/windowsdesktop/pull/4063. We want System.Private.Windows.Core to be in the runtime pack of the shared Framework, so this does not need to be omitted from FrameworkListFile 

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10639)

### DIFF
--- a/pkg/Microsoft.Private.Winforms/sdk/dotnet-windowsdesktop/System.Windows.Forms.FileClassification.props
+++ b/pkg/Microsoft.Private.Winforms/sdk/dotnet-windowsdesktop/System.Windows.Forms.FileClassification.props
@@ -17,6 +17,9 @@
     <FrameworkListFileClass Include="System.Windows.Forms.dll" Profile="WindowsForms" />
     <FrameworkListFileClass Include="System.Windows.Forms.Primitives.dll" Profile="WindowsForms" />
   </ItemGroup>
+  <!--
+    Private assemblies should not be referenced.
+  -->
   <ItemGroup>
     <IgnoredReference Include="System.Private.Windows.Core" />
   </ItemGroup>
--- a/pkg/Microsoft.Private.Winforms/sdk/dotnet-windowsdesktop/System.Windows.Forms.FileClassification.props
+++ b/pkg/Microsoft.Private.Winforms/sdk/dotnet-windowsdesktop/System.Windows.Forms.FileClassification.props
@@ -16,6 +16,8 @@
     <FrameworkListFileClass Include="System.Windows.Forms.Design.Editors.dll" Profile="WindowsForms" />
     <FrameworkListFileClass Include="System.Windows.Forms.dll" Profile="WindowsForms" />
     <FrameworkListFileClass Include="System.Windows.Forms.Primitives.dll" Profile="WindowsForms" />
-    <FrameworkListFileClass Include="System.Private.Windows.Core.dll" Profile="WindowsForms" />
+  </ItemGroup>
+  <ItemGroup>
+    <IgnoredReference Include="System.Private.Windows.Core" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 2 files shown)


JeremyKuhne (dismissed):
🎉

ericstj:
How does this exclude `System.Private.Windows.Core` from the references?

lonitra:
> How does this exclude `System.Private.Windows.Core` from the references?

Adding `<IgnoredReference Include="System.Private.Windows.Core" />` to the props file causes `System.Private.Windows.Core` to be ignored when going through `VerifyClosure` target in [sharedfx.targets](https://github.com/dotnet/arcade/blob/3717ff2d2bb08cf54315f96ffc62fbdf89defb58/src/Microsoft.DotNet.SharedFramework.Sdk/targets/sharedfx.targets#L484-L487). Otherwise without this, when this flows up to windowsdesktop repo, we will get an error about our public assemblies missing a dependency to `System.Private.Windows.Core` from this [line](https://github.com/dotnet/arcade/blob/3717ff2d2bb08cf54315f96ffc62fbdf89defb58/src/Microsoft.DotNet.PackageTesting/VerifyClosure.cs#L212)

JeremyKuhne (approved):

pkg/Microsoft.Private.Winforms/sdk/dotnet-windowsdesktop/UpdateFileClassification.ps1
```diff
@@ -74,6 +75,12 @@ else {
             $output += "    <FrameworkListFileClass Include=`"$assembly`" Profile=`"WindowsForms`" />`r`n"
         }
     $output += "  </ItemGroup>
+  <!--
+    Private assemblies should not be referenced.
+  -->
+  <ItemGroup>
+    <IgnoredReference Include=`"System.Private.Windows.Core`" />
+  </ItemGroup>
```
ericstj, line 83: Oh my -- I hadn't realized that this was what was failing.  I see.  To avoid tight coupling you could place that `IgnoredReference` in a different props file that isn't tested in this way.  So long as it's included in the ref project it should be fine.

pkg/Microsoft.Private.Winforms/sdk/dotnet-windowsdesktop/System.Windows.Forms.FileClassification.props
```diff
@@ -16,6 +16,11 @@
     <FrameworkListFileClass Include="System.Windows.Forms.Design.Editors.dll" Profile="WindowsForms" />
     <FrameworkListFileClass Include="System.Windows.Forms.dll" Profile="WindowsForms" />
     <FrameworkListFileClass Include="System.Windows.Forms.Primitives.dll" Profile="WindowsForms" />
-    <FrameworkListFileClass Include="System.Private.Windows.Core.dll" Profile="WindowsForms" />
+  </ItemGroup>
+  <!--
+    Private assemblies should not be referenced.
+  -->
+  <ItemGroup>
+    <IgnoredReference Include="System.Private.Windows.Core" />
```
ericstj, line 24: I wonder if this should also have a PackageTargetRuntime condition?  You only want to apply this for the ref pack.

pkg/Microsoft.Private.Winforms/sdk/dotnet-windowsdesktop/UpdateFileClassification.ps1
lonitra: Agree here on the coupling. Once this change flows to windowsdesktop and depdency flow is unblocked, I will make necessary changes here with your suggestions -- make a separate props file that includes `IgnoredReference` and make updates to ref project to include the props file.

PR #10640, opened 2024-01-11
Remove *I overloads and move scopes

JeremyKuhne:
Int overloads just cast to float, removing non-array ones to cut back on interop code.

- Fix some missing KeepAlives
- Move error handling for GdiPlus to Core
- Move some scopes down and convert RegionScope to use IPointer

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10640)


JeremyKuhne:
This looks bigger than it is as it adds loc files to the Core assembly.

JeremyKuhne:
There is some intermittent packaging issue with the resources that I'm exploring, but otherwise this is good for review.

src/System.Private.Windows.Core/src/Windows/Win32/Graphics/Gdi/RegionScope.cs
```diff
@@ -15,7 +16,7 @@ namespace Windows.Win32.Graphics.Gdi;
 ///  </para>
 /// </remarks>
 #if DEBUG
-internal class RegionScope : DisposalTracking.Tracker, IDisposable
+internal unsafe class RegionScope : DisposalTracking.Tracker, IDisposable
 #else
 internal ref struct RegionScope
```
lonitra, line 21: Think we may also need unsafe here
```suggestion
internal unsafe ref struct RegionScope
```

JeremyKuhne:
The resources I copied were from `System.Drawing.Common`, which aren't currently localized.

lonitra (approved):
🚀

PR #10647, opened 2024-01-12
Update initial function and add function GetStandardValuesByCulture i…

### LINKED ISSUE #10610: KeysConverter.GetStandardValues() throws an exception
### .NET version

8.0.100

Yes

Used to work in .Net 6 and 7.

### Issue description

Simply calling

`var test = new KeysConverter().GetStandardValues();`

throws the exception:

> System.Collections.Generic.KeyNotFoundException: 'The given key 'en-US' was not present in the dictionary.'

Looking at the KeysConverter source code, it's easy to see that the CultureToKeyName property is initialized only with the invariant culture at construction time. Other cultures are added only if ConvertFrom() or ConvertTo() are called first. This is a serious breaking change.

If this was intended, is it documented? What is a workaround?

### Steps to reproduce

Create a console app targetting .net 8. Simply add the code:

`var test = new KeysConverter().GetStandardValues();`

### Possible fix

In GetStandardValue(), this guard should be added:

```
if (!CultureToKeyName.ContainsKey(CultureInfo.CurrentCulture))
	AddLocalizedKeyNames(CultureInfo.CurrentCulture);
```

Note that I think CultureInfo.CurrentCulture was changed to CurrentUICulture in .Net 9 preview if I remember well a post that I saw previously.

On a deeper level, prior to .Net 8, ConvertTo, ConvertFrom and GetStandardValues were not using culture for localization. With .Net 8, it seems it changed but partially only. If it was working correctly, it seems that it is always expected that GetStandardValues() returns results for the current culture, although ConvertTo and ConvertFrom can adapt to any culture... ?

LeafShi1:
Fixes #10610

## Proposed changes

- Update initial function and add function GetStandardValuesByCulture

## Customer Impact

- The KeysConverter.GetStandardValues() can be used normally

## Regression? 

- Yes 

## Risk

- Minimal

### Before

Simply calling

`var test = new KeysConverter().GetStandardValues();`

throws the exception:

`System.Collections.Generic.KeyNotFoundException: 'The given key 'en-US' was not present in the dictionary.'`

### After

Call `var test = new KeysConverter().GetStandardValues();` to return standard key values
Call `var test = new KeysConverter().GetStandardValues();` to return key values matching current culture

## Test methodology 

- Manually

## Test environment(s) 

- .net 9.0.0-alpha.1.24058.11

### DIFF
--- a/src/System.Windows.Forms/src/PublicAPI.Unshipped.txt
+++ b/src/System.Windows.Forms/src/PublicAPI.Unshipped.txt
@@ -20,6 +20,7 @@ System.Windows.Forms.FolderBrowserDialog.Multiselect.get -> bool
 System.Windows.Forms.FolderBrowserDialog.Multiselect.set -> void
 System.Windows.Forms.FolderBrowserDialog.SelectedPaths.get -> string![]!
 System.Windows.Forms.ImageList.ImageCollection.AddRange(params System.Drawing.Image![]! images) -> void
+System.Windows.Forms.KeysConverter.GetStandardValuesByCulture() -> System.ComponentModel.TypeConverter.StandardValuesCollection!
 System.Windows.Forms.ListBox.IntegerCollection.AddRange(params int[]! items) -> void
 System.Windows.Forms.ListBox.ObjectCollection.AddRange(params object![]! items) -> void
 System.Windows.Forms.ListView.ListViewItemCollection.AddRange(params System.Windows.Forms.ListViewItem![]! items) -> void
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Input/KeysConverter.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Input/KeysConverter.cs
@@ -24,7 +24,15 @@ public class KeysConverter : TypeConverter, IComparer
     {
         _cultureToDisplayOrder = new();
         _cultureToKeyName = new();
-        AddLocalizedKeyNames(CultureInfo.InvariantCulture);
+
+        if (CultureToKeyName.ContainsKey(CultureInfo.CurrentCulture))
+        {
+            AddLocalizedKeyNames(CultureInfo.CurrentCulture);
+        }
+        else
+        {
+            AddLocalizedKeyNames(CultureInfo.InvariantCulture);
+        }
     }
 
     private void AddLocalizedKeyNames(CultureInfo cultureInfo)
@@ -363,6 +371,21 @@ public class KeysConverter : TypeConverter, IComparer
     {
         if (_values is null)
         {
+            Keys[] list = CultureToKeyName[CultureInfo.InvariantCulture].Values.ToArray();
+            Array.Sort(list, this);
+            _values = new StandardValuesCollection(list);
+        }
+
+        return _values;
+    }
+
+    public StandardValuesCollection GetStandardValuesByCulture()
+    {
+        if (_values is null)
+        {
+            if (!CultureToKeyName.ContainsKey(CultureInfo.CurrentCulture))
+                AddLocalizedKeyNames(CultureInfo.CurrentCulture);
+
             Keys[] list = CultureToKeyName[CultureInfo.CurrentCulture].Values.ToArray();
             Array.Sort(list, this);
             _values = new StandardValuesCollection(list);


src/System.Windows.Forms/src/System/Windows/Forms/Input/KeysConverter.cs
```diff
@@ -24,7 +24,15 @@ private void Initialize()
     {
         _cultureToDisplayOrder = new();
         _cultureToKeyName = new();
-        AddLocalizedKeyNames(CultureInfo.InvariantCulture);
+
+        if (CultureToKeyName.ContainsKey(CultureInfo.CurrentCulture))
```
Tanya-Solyanik, line 28: Why did you select CurrentCulture vs CurrentUICulture?

Tanya-Solyanik: This code could be simplified by using a ternary conditional expression

src/System.Windows.Forms/src/System/Windows/Forms/Input/KeysConverter.cs
```diff
@@ -363,6 +371,21 @@ public override StandardValuesCollection GetStandardValues(ITypeDescriptorContex
...
+
+        return _values;
+    }
+
+    public StandardValuesCollection GetStandardValuesByCulture()
+    {
+        if (_values is null)
+        {
+            if (!CultureToKeyName.ContainsKey(CultureInfo.CurrentCulture))
+                AddLocalizedKeyNames(CultureInfo.CurrentCulture);
```
Tanya-Solyanik, line 387: We still can return null from this method if the current culture is no in the list of our resources.

src/System.Windows.Forms/src/System/Windows/Forms/Input/KeysConverter.cs
```diff
@@ -363,6 +371,21 @@ public override StandardValuesCollection GetStandardValues(ITypeDescriptorContex
     {
         if (_values is null)
         {
+            Keys[] list = CultureToKeyName[CultureInfo.InvariantCulture].Values.ToArray();
```
Tanya-Solyanik, line 374: Why is this method getting only invariant culture?

Tanya-Solyanik: Suggestion: rename `list` to `values` because it's unintuitive to call an array a list.

src/System.Windows.Forms/src/PublicAPI.Unshipped.txt
```diff
@@ -20,6 +20,7 @@ System.Windows.Forms.FolderBrowserDialog.Multiselect.get -> bool
 System.Windows.Forms.FolderBrowserDialog.Multiselect.set -> void
 System.Windows.Forms.FolderBrowserDialog.SelectedPaths.get -> string![]!
 System.Windows.Forms.ImageList.ImageCollection.AddRange(params System.Drawing.Image![]! images) -> void
+System.Windows.Forms.KeysConverter.GetStandardValuesByCulture() -> System.ComponentModel.TypeConverter.StandardValuesCollection!
```
Tanya-Solyanik, line 23: I think we still can return a null here....

Tanya-Solyanik (commented):
Please add unit tests!

src/System.Windows.Forms/src/System/Windows/Forms/Input/KeysConverter.cs
LeafShi1: > Why did you select CurrentCulture vs CurrentUICulture?

This is based on the logic added in line 382 

https://github.com/dotnet/winforms/blob/9b5acd9f99cec277d7d451367cd3e3cd11e9712e/src/System.Windows.Forms/src/System/Windows/Forms/Input/KeysConverter.cs#L387

src/System.Windows.Forms/src/System/Windows/Forms/Input/KeysConverter.cs
LeafShi1: > Why is this method getting only invariant culture?

In the original code logic, the `CultureToKeyName` property is initialized only with the invariant culture at construction time. 

https://github.com/dotnet/winforms/blob/62612eea70eef656e0b2f19442b126e296e4ffee/src/System.Windows.Forms/src/System/Windows/Forms/Input/KeysConverter.cs#L100-L109


https://github.com/dotnet/winforms/blob/62612eea70eef656e0b2f19442b126e296e4ffee/src/System.Windows.Forms/src/System/Windows/Forms/Input/KeysConverter.cs#L23-L28


This problem is caused by [PR 8401](https://github.com/dotnet/winforms/pull/8401/files#diff-af538a9c7d8e8fabf308658199124823509b02c260ab6b60df907b0d3518c4c8). 
In the logic before the PR, `GetStandardValues` is a fixed return value, which does not change due to culture.
But the code logic after the PR is to get the default value based on the current culture, However, no culture was added to `CultureToKeyName`, so the exception "The given key 'en-US' was not present in the dictionary" was caused. so I added another method `GetStandardValuesByCulture `below.

src/System.Windows.Forms/src/PublicAPI.Unshipped.txt
LeafShi1: Remove this function GetStandardValuesByCulture

src/System.Windows.Forms/src/System/Windows/Forms/Input/KeysConverter.cs
LeafShi1: Removing the function "GetStandardValuesByCulture"

src/System.Windows.Forms/src/System/Windows/Forms/Input/KeysConverter.cs
Tanya-Solyanik: This function should be doing what the [docs ](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.typeconverter.getstandardvalues?view=net-8.0&viewFallbackFrom=windowsdesktop-8.0#system-componentmodel-typeconverter-getstandardvalues)specify - 
```
Returns a collection of standard values from the default context for the data type this type converter is designed for.
```

"default context" sounds like the invariant culture indeed. Could you please verify what the .NET Framework KeyConverter was returning in this method? The goal of this fix is to simplify migration from .NET Framework to .NET9, so ideally the behavior would be the same.

LeafShi1: In [.net Framework](https://vstfdevdiv.corp.microsoft.com/DevDiv2/DevDiv/_versionControl?path=%24/DevDiv/FX/Rel/Net481Rel1Last_C/NDP/fx/src/WinForms/Managed/System/WinForms/KeysConverter.cs&version=T&line=339&lineEnd=354&lineStartColumn=1&lineEndColumn=10&lineStyle=plain&_a=contents) , this function also returns a fixed value, which is the same as the current fix

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/KeysConverterTests.cs
```diff
@@ -84,4 +84,20 @@ public void ConvertToEnumArray_ShouldConvertKeys(Keys keys, Enum[] expectedResul
...
+    public void GetStandardValues()
+    {
+        Keys[] expectedValues =
+        [
+            Keys.None, Keys.D0, Keys.D1, Keys.D2, Keys.D3, Keys.D4, Keys.D5, Keys.D6, Keys.D7, Keys.D8, Keys.D9, Keys.Alt, Keys.Back, Keys.Control,
+            Keys.Delete, Keys.End, Keys.Enter, Keys.F1, Keys.F10, Keys.F11, Keys.F12, Keys.F2, Keys.F3, Keys.F4, Keys.F5, Keys.F6, Keys.F7, Keys.F8,
+            Keys.F9, Keys.Home, Keys.Insert, Keys.Next, Keys.PageUp, Keys.Shift
+        ];
+
+        KeysConverter converter = new();
```
Tanya-Solyanik, line 98: Is it possible to add test that runs on a non-default locale, for example de-De. See https://github.com/dotnet/winforms/blob/e6d1830fee32ec9d1c81f669d344ba0b99134cb3/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripMenuItemTests.cs#L148-L161 for an example.

Tanya-Solyanik (dismissed):

Tanya-Solyanik:
@LeafShi1  - had you tested this change in the designer? TO do that:
1 install the latest SDK https://github.com/dotnet/installer#installers-and-binaries
2. build privates
3. copy over the following dlls
    System.Windows.Forms.dll
    System.Windows.Forms.Primitives.dll
    System.Drawing.Common.dll
    System.Private.Windows.Core.dll
    
from `artifacts\bin\System.Windows.Forms\Debug\net9.0` to your shared SDK location, for example `C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App\9.0.0-alpha?????`
4 create a form targeting NET9 indesigner with a toolstrip, and edit shortcut keys in the property browser

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/KeysConverterTests.cs
```diff
@@ -84,4 +85,49 @@ public void ConvertToEnumArray_ShouldConvertKeys(Keys keys, Enum[] expectedResul
...
+    [InlineData("zh-TW")]
+    [InlineData("cs-CZ")]
+    [InlineData("ja-JP")]
+    public void GetStandardValues(string cultureName)
+    {
+        // Record original culture
+        CultureInfo originalCulture = Thread.CurrentThread.CurrentCulture;
+
+        // Update CurrentCulture
+        Thread.CurrentThread.CurrentCulture = cultureName is not null ? CultureInfo.GetCultureInfo(cultureName) : originalCulture;
```
Tanya-Solyanik, line 106: Please also set the CurrentUICulture to ensure that result does not depend on it.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/KeysConverterTests.cs
```diff
@@ -84,4 +85,49 @@ public void ConvertToEnumArray_ShouldConvertKeys(Keys keys, Enum[] expectedResul
...
+        try
+        {
+            var standardValuesCollection = new KeysConverter().GetStandardValues();
+            object[] actualValues = new ArrayList(standardValuesCollection).ToArray();
+
+            Assert.Equal(expectedValues.Length, standardValuesCollection.Count);
+
+            foreach (object key in expectedValues)
+            {
+                Assert.Contains(key, actualValues);
```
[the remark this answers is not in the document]
LeafShi1: Yes, when executing the following code, the order of keys will be rearranged differently based on the CurrentCulture https://github.com/dotnet/winforms/blob/cb4d1d1dfe69d4fc747c608fdbe85d0777bae386/src/System.Windows.Forms/src/System/Windows/Forms/Input/KeysConverter.cs#L367

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/KeysConverterTests.cs
LeafShi1: Okay, I'll add that.
I debugged the code and it gets the default values in a fixed order and then reorders them via CurrentCulture, which doesn't seem to be related to CurrentUICulture, so I ignore CurrentUICulture.

https://github.com/dotnet/winforms/blob/cb4d1d1dfe69d4fc747c608fdbe85d0777bae386/src/System.Windows.Forms/src/System/Windows/Forms/Input/KeysConverter.cs#L362-L368

LeafShi1:
> @LeafShi1 - had you tested this change in the designer? TO do that: 1 install the latest SDK https://github.com/dotnet/installer#installers-and-binaries 2. build privates 3. copy over the following dlls System.Windows.Forms.dll System.Windows.Forms.Primitives.dll System.Drawing.Common.dll System.Private.Windows.Core.dll
> 
> from `artifacts\bin\System.Windows.Forms\Debug\net9.0` to your shared SDK location, for example `C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App\9.0.0-alpha?????` 4 create a form targeting NET9 indesigner with a toolstrip, and edit shortcut keys in the property browser

OK， I will do this test.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/KeysConverterTests.cs
Tanya-Solyanik: I was thinking about something like this:
```
public void GetStandardValues(string cultureName, string uiCultureName)
    {
        // Record original culture
        CultureInfo originalCulture = Thread.CurrentThread.CurrentCulture;
        CultureInfo originalUICulture = Thread.CurrentThread.CurrentUICulture;

        // Update CurrentCulture
        Thread.CurrentThread.CurrentCulture = cultureName is not null ? CultureInfo.GetCultureInfo(cultureName) : originalCulture;
        Thread.CurrentThread.CurrentUICulture = uiCultureName is not null ? CultureInfo.GetCultureInfo(uiCultureName) : originalUICulture;
```

to show that standard values do not depend on either current culture of current UI culture

LeafShi1:
> @LeafShi1 - had you tested this change in the designer? TO do that: 1 install the latest SDK https://github.com/dotnet/installer#installers-and-binaries 2. build privates 3. copy over the following dlls System.Windows.Forms.dll System.Windows.Forms.Primitives.dll System.Drawing.Common.dll System.Private.Windows.Core.dll
> 
> from `artifacts\bin\System.Windows.Forms\Debug\net9.0` to your shared SDK location, for example `C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App\9.0.0-alpha?????` 4 create a form targeting NET9 indesigner with a toolstrip, and edit shortcut keys in the property browser

When CurrentCulture and CurrentUICulture match or don't match, the shortcut default value displays fine and the value can be set and reset successfully.
VerifyInDesigner

Tanya-Solyanik (approved):

LeafShi1:
backport to release/8.0

Issue #10615, opened 2024-01-09
[9.0.100-alpha.1.24055.6] HtmlDocument.Write() throw System.Runtime.InteropServices.COMException (0x80020008): Bad variable type.


Junjun-zhao:
### .NET version

```
.NET SDK:
 Version:           9.0.100-alpha.1.24055.6
 Commit:            ca1439c8ae
 Workload version:  9.0.100-manifests.f3b52893

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.19044
 OS Platform: Windows
 RID:         win-x64
 Base Path:   C:\Program Files\dotnet\sdk\9.0.100-alpha.1.24055.6\

.NET workloads installed:
 There are no installed workloads to display


Host:
  Version:      9.0.0-alpha.1.24054.18
  Architecture: x64
  Commit:       9325d3836d

.NET SDKs installed:
    9.0.100-alpha.1.24055.6 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 9.0.0-alpha.1.24055.3 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 9.0.0-alpha.1.24054.18 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 9.0.0-alpha.1.24055.1 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
```

### Did it work in .NET Framework?

Yes

### Did it work in any of the earlier releases of .NET Core or .NET 5+?

Yes
1. Windows10 21h2 x64 + dotnet-sdk-7.0.405: Pass
2. Windows10 21h2 x64 + dotnet-sdk-8.0.101： Pass
3. Windows10 21h2 x64 + dotnet-sdk-9.0.100-alpha.1.24055.6: Fail

### Issue description

When calling ```HtmlDocument.Write()``` to write a HTML page on dotnet-sdk-9.0.100-alpha.1.24055.6, it throws ```"System.Runtime.InteropServices.COMException (0x80020008): Bad variable type"``` exception.

**Application Name**: Character-Builder-5
**OS**： Windows 10 21H2
**CPU**: X64
**.NET Build Number**: dotnet-sdk-9.0.100-alpha.1.24055.6
**App or App source checking at:**  https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1936945

[...]

PR #10674, opened 2024-01-16
Fix SafeArray type in HtmlDocument.Write

lonitra:
Fixes https://github.com/dotnet/winforms/issues/10615 , regression from https://github.com/dotnet/winforms/pull/9972
- Avoid throwing on failure in `IHtmlDocument2.Write` as original code did not throw on failure 
- While making changes in the regression PR, I had accidentally commented out some tests covering this scenario 🤦‍♀️ , I re-enabled those tests which covers scenario outlined by issue
- The SafeArray type passed to `IHtmlDocument2.Write` needs to be `VT_VARIANT` so I've added to `SafeArrayScope` to be able to handle this.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10674)

### DIFF
--- a/src/System.Private.Windows.Core/src/Windows/Win32/System/Com/SafeArrayScope.cs
+++ b/src/System.Private.Windows.Core/src/Windows/Win32/System/Com/SafeArrayScope.cs
@@ -66,6 +66,13 @@ internal readonly unsafe ref struct SafeArrayScope<T>
         {
             throw new ArgumentException("Use ComSafeArrayScope instead");
         }
+        else if (typeof(T) == typeof(object))
+        {
+            if (value->VarType is not VARENUM.VT_VARIANT)
+            {
+                throw new ArgumentException($"Wanted SafeArrayScope<{typeof(T)}> but got SAFEARRAY with VarType={value->VarType}");
+            }
+        }
         else
         {
             // The type has not been accounted for yet in the SafeArrayScope
@@ -95,6 +102,10 @@ internal readonly unsafe ref struct SafeArrayScope<T>
         else if (typeof(T) == typeof(nint) || typeof(T).IsAssignableTo(typeof(IComIID)))
         {
             throw new ArgumentException("Use ComSafeArrayScope instead");
+        }
+        else if (typeof(T) == typeof(object))
+        {
+            vt = VARENUM.VT_VARIANT;
         }
         else
         {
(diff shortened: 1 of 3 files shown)


src/System.Windows.Forms.Primitives/src/Windows/Win32/System/Com/SafeArrayScope.cs
```diff
@@ -153,14 +164,26 @@ public SafeArrayScope(T[] array) : this((uint)array.Length)
...
             // thing as we run into new types.
             return default;
         }
         set
         {
-            if (value is string s)
+            if (Value->VarType == VARENUM.VT_VARIANT)
+            {
+                using VARIANT variant = VARIANT.FromObject(value);
+                PutElement(i, &variant);
```
lonitra, line 184: I worry about cases where we need a SafeArray of type `VT_VARIANT`, but the objects we are doing `VARIANT.FromObject`/`VARIANT.ToObject` on in order to get/set in SafeArray are not covered in FromObject/ToObject method and perhaps should not be covered in this methods. But I have not seen such a case yet so perhaps we should rethink this when we have a scenario?

Tanya-Solyanik (dismissed):

src/System.Windows.Forms.Primitives/src/Windows/Win32/System/Com/SafeArrayScope.cs
```diff
@@ -153,14 +164,26 @@ public SafeArrayScope(T[] array) : this((uint)array.Length)
             {
                 return (T)(object)GetElement<nint>(i);
             }
+            else if (typeof(T) == typeof(object))
+            {
+                // This may need to change if we have a case where
```
JeremyKuhne, line 169: Where wouldn't we want to dispose it? We effectively always dispose of the other patterns above, although we don't technically need to do it (and don't) for the intrinsic types as they're contained in the VARIANT struct.

lonitra: I was thinking about cases where we'd want a SafeArray of VARIANTs that hold pointers. I don't think we'd want to dispose that, but again I haven't seen any cases yet.

lonitra: Ah you are right, SafeArray actually gives out copies even VARIANT pointers it will make a copy and add ref appropriately so it is always safe to dispose, I will update outlining that detail

Tanya-Solyanik (approved):

Issue #8302, opened 2022-11-29
Add Test Coverage For MFC ActiveX Controls


lonitra:
Related: #8294 

This is another area we are lacking as we do not have any [MFC ActiveX](https://learn.microsoft.com/en-us/cpp/mfc/mfc-activex-controls?view=msvc-170) test coverage. Any help here is appreciated!

As a first step an MFC ActiveX control will need to be created, registered, and follow a pattern similar to #8253 to get it into the winforms solution. The gaps for test coverage can then be expanded from there :)

weltkante:
duplicate of #8294 ? either way I'll probably contribute some as part of #10583 to get COM source generator based ActiveX bindings working

Issue #8294, opened 2022-11-28
Add test coverage for AxHost related classes


JeremyKuhne:
We have some tests, but there are significant gaps. Any additional coverage here will help keep this code in good shape for the next few decades. :)

Help here is appreciated. Live on the bleeding edge of the mid-nineties. :)

Supporting this scenario is important for Windows Forms customers with legacy dependencies. This code is how Visual Basic 6 (VB6) controls are supported, and we plan to add specific tests for that scenario as well. See #8253.

lonitra:
One form of test coverage that we are lacking is test coverage against existing ActiveX controls such as Microsoft Web Browser, System Monitor Control, etc. Any additional coverage here would be immensely helpful and greatly appreciated. We have a core project that targets Framework called `AxHosts` so that dlls for these controls can be grabbed more easily to start writing tests. The process to do this would be as follows:
1. In WinForms solution, find and open `AxHosts` form. Right click on the toolbox and click "Choose Items" then "COM Components"
2. There will be a list of COM controls. Check which control(s) you would like to add and click 'OK'
3. After clicking 'OK', the control(s) should appear in the ToolBox. Add the control(s) to the form and save
4. In `AxHosts.csproj`, COM reference(s) should've been added. Change the `EmbedInteropTypes` tag to `false` then save and build Debug configuration
5. A successful build should've generated dlls in `winforms\artifacts\bin\AxHosts\Debug\net472` folder. 
6. Paste the following into the .csproj of the project you will be writing tests in:
```.csproj
<ItemGroup>
    <Reference Include="<dll_name_without_.dll_extension>">
      <HintPath>$(ArtifactsBinDir)\AxHosts\$(Configuration)\net472\<dll_name_with_.dll_extension></HintPath>
    </Reference>
</ItemGroup>
```
The dll name can be found in the folder path in step 4. If there were multipe dlls generated, add another `Reference` tag under the same `ItemGroup` tag.
7. Start writing tests! :)

We have some tests written for Windows Media Player that may be helpful as examples. See #8216 , #8251
Note - the following code snipped may be necessary to add to the beginning of your tests for the ActiveX control you've added to behave as expected.
```c#
using Form form = new();
using <Your_Added_ActiveX_Control> control= new();
((ISupportInitialize)control).BeginInit();
form.Controls.Add(control);
((ISupportInitialize)control).EndInit();
```

weltkante:
[...]

PR #10675, opened 2024-01-17
Adds tests for AxSystemMonitor

ricardobossan:
Related #8302 

## Proposed changes

- Instead of consolidating multiple controls into a single pull request, opted for separate PRs for each control for better organization.
- Followed [Lonitra's instructions](https://github.com/dotnet/winforms/issues/8294#issuecomment-1331026574) on related issue #[8294](https://github.com/dotnet/winforms/issues/8294) by adding Ax control references to both the AxHosts project and the WinForms test project. Additionally, added corresponding tests.

## Customer Impact

- None

## Regression? 

- No

## Risk

- Minimal

## Test methodology 

- Unit tests

## Test environment(s) 

- 9.0.100-alpha.1.23618.3

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10675)

### DIFF
--- a/src/System.Windows.Forms/tests/UnitTests/System.Windows.Forms.Tests.csproj
+++ b/src/System.Windows.Forms/tests/UnitTests/System.Windows.Forms.Tests.csproj
@@ -42,6 +42,15 @@
   </ItemGroup>
 
   <ItemGroup>
+    <Reference Include="AxInterop.SystemMonitor">
+      <HintPath>$(ArtifactsBinDir)\AxHosts\$(Configuration)\net472\AxInterop.SystemMonitor.dll</HintPath>
+    </Reference>
+    <Reference Include="Interop.SystemMonitor">
+      <HintPath>$(ArtifactsBinDir)\AxHosts\$(Configuration)\net472\Interop.SystemMonitor.dll</HintPath>
+    </Reference>
+  </ItemGroup>
+
+  <ItemGroup>
     <Content Include="TestResources\Files\*">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
@@ -67,14 +76,4 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
   </ItemGroup>
-
-  <ItemGroup>
-    <Reference Include="AxInterop.WMPLib">
-      <HintPath>$(ArtifactsBinDir)\AxHosts\$(Configuration)\net472\AxInterop.WMPLib.dll</HintPath>
-    </Reference>
-    <Reference Include="Interop.WMPLib">
-      <HintPath>$(ArtifactsBinDir)\AxHosts\$(Configuration)\net472\Interop.WMPLib.dll</HintPath>
-    </Reference>
-  </ItemGroup>
-
 </Project>
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AxSystemMonitorTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AxSystemMonitorTests.cs
@@ -31,6 +31,7 @@ public class AxSystemMonitorTests : IDisposable
 
     public void Dispose()
     {
+        // This line was added due to https://github.com/dotnet/winforms/issues/10692
         using NoAssertContext context = new();
         _control.Dispose();
         _form.Dispose();
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AxSystemMonitorTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AxSystemMonitorTests.cs
@@ -1,4 +1,6 @@
 ﻿using System.ComponentModel;
+using System.Reflection;
+using Castle.Core.Internal;
 
 namespace System.Windows.Forms.Tests;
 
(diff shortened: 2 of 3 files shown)


ricardobossan:
Drafted because I may more tests.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AxSystemMonitorTests.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System.ComponentModel;
+
+namespace System.Windows.Forms.Tests;
+
+public class AxSystemMonitorTests
+{
+    private readonly Form form;
+    private readonly AxSystemMonitor.AxSystemMonitor control;
```
lonitra, line 8: Fields should have `_` prefix
```suggestion
    private readonly Form _form;
    private readonly AxSystemMonitor.AxSystemMonitor _control;
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AxSystemMonitorTests.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System.ComponentModel;
+
+namespace System.Windows.Forms.Tests;
+
+public class AxSystemMonitorTests
```
lonitra, line 5: Could you implement `IDisposable` so that the fields can be properly disposed similar to how [ErrorProviderAccessibleObjectTests](https://github.com/dotnet/winforms/blob/0956586af08d74cf4e87f1e9e837e39af2174587/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjects/ErrorProviderAccessibleObjectTests.cs#L9) does so

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AxSystemMonitorTests.cs
```diff
@@ -0,0 +1,40 @@
...
+    public void AxSystemMonitor_WhenInitialized_IsEnabled()
+    {
+        Assert.True(control.Enabled);
+    }
+
+    [WinFormsFact]
+    public void AxSystemMonitor_Counter_ReturnsDefaultCounter()
+    {
+        Assert.Equal(0, control.Counters.Count);
+    }
```
lonitra, line 39: Consider condensing these tests into one if we are just checking the state of the control when it has been initialized.

lonitra (commented):
Were you able to add `AxSystemMonitor` in the `AxHosts` form designer and see it during runtime?

src/System.Windows.Forms/tests/UnitTests/System.Windows.Forms.Tests.csproj
```diff
@@ -67,14 +76,4 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
   </ItemGroup>
-
-  <ItemGroup>
-    <Reference Include="AxInterop.WMPLib">
```
Tanya-Solyanik, line 72: Did you delete reference to the media player intentionally?

ricardobossan: Yes. It was duplicating lines from 35 to 42.

image

ricardobossan:
> Were you able to add `AxSystemMonitor` in the `AxHosts` form designer and see it during runtime?

@lonitra, I successfully added `AxSystemMonitor` to the designer. However, upon attempting to run the AxHosts project, I encountered the following error:

AxHosts Runtime Error

I haven't identified the root cause of this issue yet and would appreciate any insights or suggestions on how to address it.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AxSystemMonitorTests.cs
lonitra: @JeremyKuhne was able to investigate https://github.com/dotnet/winforms/issues/10692 and looks like the error you were running into while implementing `IDisposable` is an issue with only the `AxSystemMonitor` control. In order to implement `IDisposable` successfully, could you also add the following to the beginning of your implementation of `Dispose()`:
```c#
using NoAssertContext context = new();
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AxSystemMonitorTests.cs
```diff
@@ -0,0 +1,31 @@
...
+        _control = new AxSystemMonitor.AxSystemMonitor();
+        ((ISupportInitialize)_control).BeginInit();
+        _form.Controls.Add(_control);
+        ((ISupportInitialize)_control).EndInit();
+    }
+
+    [WinFormsFact]
+    public void AxSystemMonitor_WhenInitialized_ExpectsProperties()
+    {
+        var properties = TypeDescriptor.GetProperties(_control);
```
lonitra, line 22: Do we know exactly what properties/events are here? It might be good to test that the properties/events that are here are what we expect.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AxSystemMonitorTests.cs
ricardobossan: Worked out fine now
image

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AxSystemMonitorTests.cs
ricardobossan: I just added a test to verify this.

lonitra (approved):
LGTM 👍

PR #10693, opened 2024-01-18
Add code coverage for buttons

ricardobossan:
## Proposed changes

- Adds code coverage for buttons

## Customer Impact

- None

## Regression? 

- No

## Risk

- Minimal

## Test methodology 

- Unit tests

## Test environment(s) 

- 9.0.100-alpha.1.23618.3


ricardobossan:
Drafted because there's more code to add coverage to.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ButtonTests.cs
```diff
@@ -3649,6 +3649,63 @@ public void Button_WndProc_InvokeReflectCommandWithHandle_Success(FlatStyle flat
         Assert.Equal(0, createdCallCount);
     }
 
+    [WinFormsFact]
+    public unsafe void Button_Flat_ValidBorder()
```
SimonZhao888, line 3653: Please rename this test method.
image

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ButtonTests.cs
```diff
@@ -3649,6 +3649,52 @@ public void Button_WndProc_InvokeReflectCommandWithHandle_Success(FlatStyle flat
...
+            button.FlatAppearance.BorderSize = borderSize;
+            Assert.Equal(borderSize, button.FlatAppearance.BorderSize);
+        }
+    }
+
+    [WinFormsTheory]
+    [InlineData(255, 0, 0)]
+    [InlineData(0, 255, 0)]
+    [InlineData(0, 0, 255)]
+    public unsafe void Button_Flat_ProperColor(int red, int green, int blue)
```
sharwell, line 3680: 💭 The properties tested in this method seem to be inherited from ButtonBase. It may be helpful to create an AbstractButtonBaseTests class that contains tests that apply to all subtypes of ButtonBase, and then create make ButtonTests extend AbstractButtonBaseTests so it can run those inherited tests against `Button`.

ricardobossan: I understand you mean we should unify common test logic for all buttons under an abstract class, and then implement specificities in each inheriting test class.

However, we already have around 12,000 lines of existing tests, including a ButtonBaseTests class with 9,379 lines unrelated to derived controls.

Existing Tests (including the ones I already wrote on the PR draft):

- ButtonTests.cs: 3,837 lines
- CheckBoxTests.cs: 632 lines
- ButtonRenderingTests.cs: 248 lines
- ButtonBaseTests.cs: 9,379 lines (unrelated to those other three test classes)

Your idea of an abstract class for common button tests makes sense, but with our existing structure, I'm unsure about the practicality, especially given the challenges of integrating with the large `ButtonBaseTests` class..

@Tanya-Solyanik @merriemcgaw @LeafShi1 @SimonZhao888 @Epica3055  How would you guys think we should approach this?

SimonZhao888: Considering that the FlatButtonAppearance class is called in the ButtonBase class, 
https://github.com/dotnet/winforms/blob/63f2d85bdc754063f935729029b1287f5b22d285/src/System.Windows.Forms/src/System/Windows/Forms/Controls/Buttons/ButtonBase.cs#L354
the best option would be to put the relevant test case method in ButtonBaseTest.
https://github.com/dotnet/winforms/blob/63f2d85bdc754063f935729029b1287f5b22d285/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ButtonBaseTests.cs#L1028

sharwell: @ricardobossan I'm not necessarily suggesting that `ButtonTests` derive from `ButtonBaseTests`. I'm suggesting that `ButtonTests` derives from `AbstractButtonBaseTests` (a new type that does not currently exist), and tests that cover the `ButtonBase` behavior _as it applies to a derived implementation_ would go in that location. `ButtonTests` would then contain tests specific to `Button`, and also inherit tests for its `ButtonBase` behavior.

ricardobossan: @sharwell Interesting idea. How would we go about this? Should we consider creating new tests in AbstractButtonBaseTests for common behaviors, or do you suggest moving existing tests from ButtonBaseTests?

@merriemcgaw @Tanya-Solyanik what are your thoughts on this approach?

Tanya-Solyanik: Makes sense to reduce amoun of duplicated code. Inheritance hierarchy:
                 AbstractButtonBaseTests
                    /                  |             \
     ButtonBaseTests     ButtonTests    OtherButtonTests

AbstractButtonBaseTests contains helper methods, and Specific tests classes call those helper methods for specific data.

>creating new tests in AbstractButtonBaseTests for common behaviors 

This sounds reasonable to me

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ButtonBaseTests.cs
```diff
@@ -9244,6 +9244,52 @@ public void ButtonBase_WndProc_InvokeSetStateWithHandle_Success(FlatStyle flatSt
         Assert.Equal(0, createdCallCount);
     }
 
+    [WinFormsTheory]
+    [InlineData(-1)]
+    [InlineData(0)]
+    [InlineData(1)]
+    public unsafe void Button_Flat_ValidBorder(int borderSize)
+    {
+        using SubButton button = new()
```
Tanya-Solyanik, line 9253: In the case of this test, I would put a helper method in the base class ButtonTestBase(name can be improved on)- 

```
protected void FlatAppearance_BorderSize_Valid(ButtonBase control, int borderSize)
{
        Control.FlatStyle = FlatStyle.Flat,

        Assert.Throws<NotSupportedException>(() => control.FlatAppearance.BorderColor = Color.Transparent);

        if (borderSize < 0)
        {
            Assert.Throws<ArgumentOutOfRangeException>(() => control.FlatAppearance.BorderSize = borderSize);
        }
        else
        {
            control.FlatAppearance.BorderSize = borderSize;
            Assert.Equal(borderSize, control.FlatAppearance.BorderSize);
        }
}
```

And then derive ButtonBaseTests from that class

>public class ButtonTests : ButtonTestBase 

and implement test like this:
    [WinFormsTheory]
    [InlineData(-1)]
    [InlineData(0)]
    [InlineData(1)]
    public unsafe void Button_FlatAppearance_BorderSize_Valid(int borderSize)
    {
        using SubButton button = new();
       FlatAppearance_BorderSize_Valid(button, borderSize);
    }

And the same could be done for other button-like controls, if implementations differer.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ButtonBaseTests.cs
```diff
@@ -9244,6 +9244,52 @@ public void ButtonBase_WndProc_InvokeSetStateWithHandle_Success(FlatStyle flatSt
...
+
+        button.FlatAppearance.CheckedBackColor = expectedColor;
+        button.FlatAppearance.BorderColor = expectedColor;
+
+        Assert.Equal(expectedColor, button.BackColor);
+        Assert.Equal(expectedColor, button.FlatAppearance.BorderColor);
+        Assert.Equal(expectedColor, button.FlatAppearance.CheckedBackColor);
+    }
+
     private class SubButton : Button
```
Tanya-Solyanik, line 9293: Hm, Button control should not be present in the ButtonBase tests...

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ButtonTests.cs
Tanya-Solyanik: > Considering that the FlatButtonAppearance class is called in the ButtonBase class,
> 
> https://github.com/dotnet/winforms/blob/63f2d85bdc754063f935729029b1287f5b22d285/src/System.Windows.Forms/src/System/Windows/Forms/Controls/Buttons/ButtonBase.cs#L354
> 
> 
> the better option would be to put the relevant test case method in ButtonBaseTest.
> https://github.com/dotnet/winforms/blob/63f2d85bdc754063f935729029b1287f5b22d285/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ButtonBaseTests.cs#L1028

@SimonZhao888  - do we need this test as applied to Button class once we have the same already applied to ButtonBase? - Do code coverage numbers go up when we add test against Button?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBoxRenderingTests.cs
```diff
@@ -0,0 +1,171 @@
...
+public class CheckBoxRenderingTests
+{
+    [WinFormsTheory]
+    [InlineData(CheckBoxState.CheckedNormal)]
+    [InlineData(CheckBoxState.MixedNormal)]
+    public void CheckBoxRenderer_DrawCheckBox(CheckBoxState state)
+    {
+        using Bitmap bitmap = new(100, 100);
+        using Graphics graphics = Graphics.FromImage(bitmap);
+        Point rectangle = new Point(10, 20);
```
Tanya-Solyanik, line 19:
```suggestion
        Point point= new Point(10, 20);
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBoxRenderingTests.cs
```diff
@@ -0,0 +1,171 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Drawing;
+using System.Windows.Forms.Metafiles;
+using System.Windows.Forms.VisualStyles;
+
+namespace System.Windows.Forms.Tests;
+
+public class CheckBoxRenderingTests
```
Tanya-Solyanik, line 10: Class name should follow pattern `<name of the type being tested>Tests`, and file should be named the same.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBoxRenderingTests.cs
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    [WinFormsTheory]
+    [InlineData(CheckBoxState.CheckedNormal)]
+    [InlineData(CheckBoxState.MixedNormal)]
+    public void CheckBoxRenderer_DrawCheckBox_OverloadWithSizeAndText(CheckBoxState state)
+    {
+        using Bitmap bitmap = new(100, 100);
+        using Graphics graphics = Graphics.FromImage(bitmap);
+        Point point = new Point(10, 20);
```
Tanya-Solyanik, line 31:
```suggestion
        Point point = new(10, 20);
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ButtonBaseTests.cs
```diff
@@ -9244,6 +9244,52 @@ public void ButtonBase_WndProc_InvokeSetStateWithHandle_Success(FlatStyle flatSt
...
+        Assert.Throws<NotSupportedException>(() => button.FlatAppearance.BorderColor = Color.Transparent);
+
+        if (borderSize < 0)
+        {
+            Assert.Throws<ArgumentOutOfRangeException>(() => button.FlatAppearance.BorderSize = borderSize);
+        }
+        else
+        {
+            button.FlatAppearance.BorderSize = borderSize;
+            Assert.Equal(borderSize, button.FlatAppearance.BorderSize);
```
Tanya-Solyanik, line 9267: We are moving towards https://fluentassertions.com/ in the newly added tests. We already have some examples in the codebase to follow.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ButtonBaseTests.cs
```diff
@@ -9244,6 +9244,52 @@ public void ButtonBase_WndProc_InvokeSetStateWithHandle_Success(FlatStyle flatSt
...
+        }
+    }
+
+    [WinFormsTheory]
+    [InlineData(255, 0, 0)]
+    [InlineData(0, 255, 0)]
+    [InlineData(0, 0, 255)]
+    public unsafe void Button_Flat_ProperColor(int red, int green, int blue)
```
Tanya-Solyanik, line 9275: test names should follow a pattern - 

`<TypeBeing tested>_<property or method name>_<result or characteristic of the input>`

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBoxRenderingTests.cs
```diff
@@ -0,0 +1,171 @@
...
+        Rectangle bounds = new Rectangle(10, 20, 30, 40);
+
+        CheckBoxRenderer.DrawCheckBox(graphics, point, bounds, "Text", SystemFonts.DefaultFont, false, state);
```
Tanya-Solyanik, line 34: Is it possible to add validation to this test case?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RadioButtonRenderingTests.cs
```diff
@@ -0,0 +1,170 @@
...
+public class RadioButtonRenderingTests
+{
+    [WinFormsTheory]
+    [InlineData(RadioButtonState.CheckedNormal)]
+    [InlineData(RadioButtonState.CheckedPressed)]
+    public void RadioButtonRenderer_DrawRadioButton(RadioButtonState state)
+    {
+        using Bitmap bitmap = new(100, 100);
+        using Graphics graphics = Graphics.FromImage(bitmap);
+        Point rectangle = new Point(10, 20);
```
Tanya-Solyanik, line 19:
```suggestion
        Point point= new(10, 20);
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RadioButtonRenderingTests.cs
```diff
@@ -0,0 +1,170 @@
...
+            types.Add(record.Type);
+            details.Add(record.ToString());
+            return true;
+        });
+    }
+
+    [WinFormsFact]
+    public unsafe void Button_VisualStyles_off_Default_LineDrawing()
+    {
+        if (Application.RenderWithVisualStyles)
```
Tanya-Solyanik, line 72: Do we run tests with RenderWithVisualStyles == false?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RadioButtonTests.cs
```diff
@@ -1625,4 +1625,52 @@ internal override bool RaiseAutomationPropertyChangedEvent(UIA_PROPERTY_ID prope
...
+    [InlineData(Appearance.Button, FlatStyle.Flat)]
+    [InlineData(Appearance.Button, FlatStyle.Popup)]
+    [InlineData(Appearance.Button, FlatStyle.System)]
+    [InlineData(Appearance.Normal, FlatStyle.Standard)]
+    [InlineData(Appearance.Normal, FlatStyle.Flat)]
+    [InlineData(Appearance.Normal, FlatStyle.Popup)]
+    [InlineData(Appearance.Normal, FlatStyle.System)]
+    public void RadioButton_OverChangeRectangle_Get(Appearance appearance, FlatStyle flatStyle)
+    {
+        SubRadioButton control = new()
```
Tanya-Solyanik, line 1640: This test could be factored out into the base class.

Tanya-Solyanik (commented):
Great new tests! I added some comments that apply to multiple spots, but I had not repeated them everywhere, please review your changes with these comments in mind.

It will be outside of scope for this PR to move all tests to a new model where test is defined in the base class, but the Fact is implemented in a derived class. But you could implement new tests so that they follow this pattern.

Test naming is clarified here - https://github.com/dotnet/winforms/blob/main/docs/testing.md#unit-test-best-practices

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ButtonTests.cs
SimonZhao888: If we have the same unit test to ButtonBase, I think we don't need apply this test to Button class, and the coverage numbers won't go up.

Actually, although we do test for FlatAppearance in ButtonBaseTest class, but it's not perfect.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AbstractButtonBaseTests.cs
```diff
@@ -0,0 +1,95 @@
...
+public abstract class AbstractButtonBaseTests
+{
+    protected abstract ButtonBase CreateButton();
+
+    public virtual void ButtonBase_Click_RaisesClickEvent()
+    {
+        using var button = (Button)CreateButton();
+        bool clickEventRaised = false;
+        button.Click += (sender, e) => clickEventRaised = true;
+        button.PerformClick();
```
Tanya-Solyanik, line 14: Since PerformClick is available only in the Button class, and we can test this scenario only for Button, this test should be moved to ButtonTests.cs

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ButtonBaseTests.cs
```diff
@@ -9244,6 +9247,11 @@ public void ButtonBase_WndProc_InvokeSetStateWithHandle_Success(FlatStyle flatSt
         Assert.Equal(0, createdCallCount);
     }
 
+    protected override ButtonBase CreateButton()
+    {
+        return new SubButton();
```
Tanya-Solyanik, line 9252: Use expression-bodied syntax

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBoxTests.cs
```diff
@@ -527,4 +530,129 @@ internal override bool RaiseAutomationPropertyChangedEvent(UIA_PROPERTY_ID prope
...
+        checkBox.CheckedChanged += (sender, args) => eventFired = true;
+        checkBox.Checked = !checkBox.Checked;
+
+        Assert.True(eventFired);
+    }
+
+    [WinFormsFact]
+    public void CheckBox_CheckStateChangedEvent_Fires()
+    {
+        CheckBox checkBox = new CheckBox();
```
Tanya-Solyanik, line 549: Please add `using`

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AbstractButtonBaseTests.cs
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        control.Appearance = appearance;
+        control.FlatStyle = flatStyle;
+
+        Rectangle overChangeRectangle;
+
+        // ButtonBase.Adapter prohibits this
+        if (appearance == Appearance.Normal && (flatStyle != FlatStyle.Standard && flatStyle != FlatStyle.Popup &&
+                                                flatStyle != FlatStyle.Flat))
```
Tanya-Solyanik, line 69:
```suggestion
        if (appearance == Appearance.Normal 
            && (flatStyle != FlatStyle.Standard 
            && flatStyle != FlatStyle.Popup 
            && flatStyle != FlatStyle.Flat))
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBoxTests.cs
```diff
@@ -527,4 +530,129 @@ internal override bool RaiseAutomationPropertyChangedEvent(UIA_PROPERTY_ID prope
             return base.RaiseAutomationPropertyChangedEvent(propertyId, oldValue, newValue);
         }
     }
+
+    [WinFormsFact]
+    public void CheckBox_CheckedChangedEvent_Fires()
+    {
+        CheckBox checkBox = new CheckBox();
```
Tanya-Solyanik, line 537: Please add `using`

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AbstractButtonBaseTests.cs
```diff
@@ -0,0 +1,95 @@
...
+        control.FlatAppearance.BorderColor = expectedColor;
+
+        Assert.Equal(expectedColor, control.BackColor);
+        Assert.Equal(expectedColor, control.FlatAppearance.BorderColor);
+        Assert.Equal(expectedColor, control.FlatAppearance.CheckedBackColor);
+    }
+
+    public virtual void ButtonBase_OverChangeRectangle_Get(Type controlType, Appearance appearance, FlatStyle flatStyle)
+    {
+        dynamic control = Activator.CreateInstance(controlType);
```
Tanya-Solyanik, line 55: Why not use the `using var control = (Control)CreateButton();` pattern?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBoxRendererTests.cs
```diff
@@ -0,0 +1,171 @@
...
+    public unsafe void CaptureButtonOnForm()
+    {
+        using Form form = new();
+        using CheckBox checkBox = new();
+        form.Controls.Add(checkBox);
+
+        using EmfScope emf = new();
+        form.PrintToMetafile(emf);
+
+        string details = emf.RecordsToString();
```
Tanya-Solyanik, line 169: Is it possible to add validation here?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBoxTests.cs
```diff
@@ -527,4 +530,129 @@ internal override bool RaiseAutomationPropertyChangedEvent(UIA_PROPERTY_ID prope
...
+    [InlineData(Appearance.Button, FlatStyle.Flat)]
+    [InlineData(Appearance.Button, FlatStyle.Popup)]
+    [InlineData(Appearance.Button, FlatStyle.System)]
+    [InlineData(Appearance.Normal, FlatStyle.Standard)]
+    [InlineData(Appearance.Normal, FlatStyle.Flat)]
+    [InlineData(Appearance.Normal, FlatStyle.Popup)]
+    [InlineData(Appearance.Normal, FlatStyle.System)]
+    public void CheckBox_DownChangeRectangle_ReturnsExpectedRectangle(Appearance appearance, FlatStyle flatStyle)
+    {
+        SubCheckBox checkBox = new SubCheckBox()
```
Tanya-Solyanik, line 580: Please add `using`.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBoxTests.cs
```diff
@@ -527,4 +530,129 @@ internal override bool RaiseAutomationPropertyChangedEvent(UIA_PROPERTY_ID prope
...
+        }
+    }
+
+    [WinFormsTheory]
+    [InlineData(true)]
+    [InlineData(false)]
+    public void CheckBox_LeftClick_MouseUpCounts(bool capture)
+    {
+        Form form = new();
+        SubCheckBox control = new() { Capture = capture };
```
Tanya-Solyanik, line 604:
```suggestion
        using Form form = new();
        using SubCheckBox control = new() { Capture = capture };
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBoxTests.cs
```diff
@@ -527,4 +530,129 @@ internal override bool RaiseAutomationPropertyChangedEvent(UIA_PROPERTY_ID prope
...
+    }
+
+    [WinFormsTheory]
+    [InlineData(true, '&', "&MnemonicText")]
+    [InlineData(true, 'N', "NonMnemonicText")]
+    [InlineData(true, 'M', "&MnemonicText")]
+    [InlineData(false, 'M', "&MnemonicText")]
+    public void CheckBox_ProcessMnemonic_ValidCases(bool useMnemonic, char charCode, string buttonText)
+    {
+        Form form = new();
```
Tanya-Solyanik, line 631:
```suggestion
        using Form form = new();
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBoxTests.cs
```diff
@@ -527,4 +530,129 @@ internal override bool RaiseAutomationPropertyChangedEvent(UIA_PROPERTY_ID prope
...
+        }
+        else
+        {
+            Assert.False(result);
+        }
+    }
+
+    protected override ButtonBase CreateButton()
+    {
+        return new SubCheckBox();
```
Tanya-Solyanik, line 656: use expression-bodied syntax.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBoxTests.cs
```diff
@@ -527,4 +530,129 @@ internal override bool RaiseAutomationPropertyChangedEvent(UIA_PROPERTY_ID prope
...
+    [InlineData(false)]
+    public void CheckBox_LeftClick_MouseUpCounts(bool capture)
+    {
+        Form form = new();
+        SubCheckBox control = new() { Capture = capture };
+        form.Controls.Add(control);
+        control.TabIndex = 9999;
+        form.Show();
+
+        MouseEventArgs eventArgs = new MouseEventArgs(MouseButtons.Left, 1, new Point(0, 0), 0);
```
Tanya-Solyanik, line 609:
```suggestion
        MouseEventArgs eventArgs = new(MouseButtons.Left, 1, new Point(0, 0), 0);
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RadioButtonRendererTests.cs
```diff
@@ -0,0 +1,172 @@
...
+    [WinFormsTheory]
+    [InlineData(RadioButtonState.CheckedNormal)]
+    [InlineData(RadioButtonState.CheckedPressed)]
+    public void RadioButtonRenderer_DrawRadioButton(RadioButtonState state)
+    {
+        using Bitmap bitmap = new(100, 100);
+        using Graphics graphics = Graphics.FromImage(bitmap);
+        Point point = new Point(10, 20);
+
+        RadioButtonRenderer.DrawRadioButton(graphics, point, state);
```
Tanya-Solyanik, line 21: Is it possible to add validation to all tests in this file?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RadioButtonTests.cs
```diff
@@ -1625,4 +1625,20 @@ internal override bool RaiseAutomationPropertyChangedEvent(UIA_PROPERTY_ID prope
...
+    [InlineData(typeof(SubRadioButton), Appearance.Button, FlatStyle.System)]
+    [InlineData(typeof(SubRadioButton), Appearance.Normal, FlatStyle.Standard)]
+    [InlineData(typeof(SubRadioButton), Appearance.Normal, FlatStyle.Flat)]
+    [InlineData(typeof(SubRadioButton), Appearance.Normal, FlatStyle.Popup)]
+    [InlineData(typeof(SubRadioButton), Appearance.Normal, FlatStyle.System)]
+    public void RadioButton_OverChangeRectangle_Get(Type controlType, Appearance appearance, FlatStyle flatStyle) => base.ButtonBase_OverChangeRectangle_Get(controlType, appearance, flatStyle);
+
+    protected override ButtonBase CreateButton()
+    {
+        return new SubRadioButton();
```
Tanya-Solyanik, line 1642: Use expression-bodied syntax.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AbstractButtonBaseTests.cs
```diff
@@ -1,24 +1,18 @@
...
-    public virtual void ButtonBase_Click_RaisesClickEvent()
-    {
-        using var button = (Button)CreateButton();
-        bool clickEventRaised = false;
-        button.Click += (sender, e) => clickEventRaised = true;
-        button.PerformClick();
-
-        Assert.True(clickEventRaised);
-    }
+    protected abstract ButtonBase CreateButton { get; }
```
Tanya-Solyanik, line 10: `CreateButton` sounds like a method name because it starts with an action verb. It's a better name for a method. Property name would be `ButtonUnderTest`, but that name does not convey if the button should be disposed by the caller. So `CreateButton()` method is preferable in this case.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AbstractButtonBaseTests.cs
```diff
@@ -65,8 +59,10 @@ public virtual void ButtonBase_OverChangeRectangle_Get(Type controlType, Appeara
         Rectangle overChangeRectangle;
 
         // ButtonBase.Adapter prohibits this
-        if (appearance == Appearance.Normal && (flatStyle != FlatStyle.Standard && flatStyle != FlatStyle.Popup &&
-                                                flatStyle != FlatStyle.Flat))
+        if (appearance == Appearance.Normal
+            && (flatStyle != FlatStyle.Standard
+            && flatStyle != FlatStyle.Popup
+            && flatStyle != FlatStyle.Flat))
```
Tanya-Solyanik, line 65:
```suggestion
        if (appearance == Appearance.Normal
            && (flatStyle != FlatStyle.Standard
                && flatStyle != FlatStyle.Popup
                && flatStyle != FlatStyle.Flat))
```

Indentation should reflect the level of the statement.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ButtonTests.cs
```diff
@@ -3665,6 +3671,7 @@ protected override ButtonBase CreateButton()
     [InlineData(0, 255, 0)]
     [InlineData(0, 0, 255)]
     public void Button_Flat_ProperColor(int red, int green, int blue) => base.ButtonBase_FlatStyle_ProperFlatButtonColor(red, green, blue);
+    protected override ButtonBase CreateButton => new SubButton();
```
Tanya-Solyanik, line 3674: Please separate methods by a blank line.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBoxRendererTests.cs
```diff
@@ -119,7 +119,8 @@ public unsafe void Button_VisualStyles_off_Default_WithText_LineDrawing()
             return;
         }
 
-        using CheckBox checkBox = new() { Text = "Hello" };
+        using CheckBox checkBox = (CheckBox)CreateButton  ;
```
Tanya-Solyanik, line 122:
```suggestion
        using CheckBox checkBox = (CheckBox)CreateButton ;
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RadioButtonRendererTests.cs
```diff
@@ -161,12 +162,14 @@ public unsafe void Button_VisualStyles_off_Default_WithText_LineDrawing()
     public unsafe void CaptureButtonOnForm()
     {
         using Form form = new();
-        using RadioButton RadioButton = new();
-        form.Controls.Add(RadioButton);
+        using RadioButton radioButton = (Forms.RadioButton)CreateButton;
```
Tanya-Solyanik, line 165:
```suggestion
        using RadioButton radioButton = (RadioButton)CreateButton;
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AbstractButtonBaseTests.cs
```diff
@@ -0,0 +1,91 @@
...
+namespace System.Windows.Forms.Tests;
+
+public abstract class AbstractButtonBaseTests
+{
+    protected abstract ButtonBase CreateButton { get; }
+
+    public virtual void ButtonBase_FlatStyle_ValidFlatButtonBorder(int borderSize)
+    {
+        using var control = CreateButton;
+        ;
```
Tanya-Solyanik, line 15:
```suggestion
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AbstractButtonBaseTests.cs
```diff
@@ -0,0 +1,91 @@
...
+
+using System.Drawing;
+
+namespace System.Windows.Forms.Tests;
+
+public abstract class AbstractButtonBaseTests
+{
+    protected abstract ButtonBase CreateButton { get; }
+
+    public virtual void ButtonBase_FlatStyle_ValidFlatButtonBorder(int borderSize)
```
Tanya-Solyanik, line 12: Are we overriding this method? It doesn't have to be virtual.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBoxRendererTests.cs
```diff
@@ -0,0 +1,174 @@
...
+    [InlineData(TextFormatFlags.PreserveGraphicsTranslateTransform, CheckBoxState.CheckedPressed)]
+    [InlineData(TextFormatFlags.TextBoxControl, CheckBoxState.UncheckedNormal)]
+    public void CheckBoxRenderer_DrawCheckBox_OverloadWithTextFormat(TextFormatFlags textFormat, CheckBoxState state)
+    {
+        using Bitmap bitmap = new(100, 100);
+        using Graphics graphics = Graphics.FromImage(bitmap);
+        Point point = new Point(10, 20);
+        Rectangle bounds = new Rectangle(10, 20, 30, 40);
+
+        CheckBoxRenderer.DrawCheckBox(graphics, point, bounds, "Text", SystemFonts.DefaultFont, textFormat, false, state);
```
Tanya-Solyanik, line 50: Please add argument name before the literal (`false`).

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBoxRendererTests.cs
```diff
@@ -0,0 +1,174 @@
...
+    [InlineData(TextFormatFlags.Default, CheckBoxState.CheckedNormal)]
+    [InlineData(TextFormatFlags.Default, CheckBoxState.MixedNormal)]
+    [InlineData(TextFormatFlags.GlyphOverhangPadding, CheckBoxState.MixedHot)]
+    [InlineData(TextFormatFlags.PreserveGraphicsTranslateTransform, CheckBoxState.CheckedPressed)]
+    [InlineData(TextFormatFlags.TextBoxControl, CheckBoxState.UncheckedNormal)]
+    public void CheckBoxRenderer_DrawCheckBox_OverloadWithTextFormat(TextFormatFlags textFormat, CheckBoxState state)
+    {
+        using Bitmap bitmap = new(100, 100);
+        using Graphics graphics = Graphics.FromImage(bitmap);
+        Point point = new Point(10, 20);
```
Tanya-Solyanik, line 47:
```suggestion
        Point point = new(10, 20);
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBoxRendererTests.cs
```diff
@@ -0,0 +1,174 @@
...
+        Rectangle bounds = new Rectangle(10, 20, 30, 40);
```
Tanya-Solyanik, line 48:
```suggestion
        Rectangle bounds = new(10, 20, 30, 40);
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBoxTests.cs
```diff
@@ -527,4 +530,126 @@ internal override bool RaiseAutomationPropertyChangedEvent(UIA_PROPERTY_ID prope
             return base.RaiseAutomationPropertyChangedEvent(propertyId, oldValue, newValue);
         }
     }
+
+    [WinFormsFact]
+    public void CheckBox_CheckedChangedEvent_Fires()
```
Tanya-Solyanik, line 535:
```suggestion
    public void CheckBox_CheckedChangedEvent_Raised()
```

And the same applies to the variable name and the next test name

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RadioButtonRendererTests.cs
```diff
@@ -0,0 +1,175 @@
...
+public class RadioButtonRendererTests : AbstractButtonBaseTests
+{
+    [WinFormsTheory]
+    [InlineData(RadioButtonState.CheckedNormal)]
+    [InlineData(RadioButtonState.CheckedPressed)]
+    public void RadioButtonRenderer_DrawRadioButton(RadioButtonState state)
+    {
+        using Bitmap bitmap = new(100, 100);
+        using Graphics graphics = Graphics.FromImage(bitmap);
+        Point point = new Point(10, 20);
```
Tanya-Solyanik, line 19:
```suggestion
        Point point = new(10, 20);
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RadioButtonRendererTests.cs
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    [WinFormsTheory]
+    [InlineData(RadioButtonState.CheckedNormal)]
+    [InlineData(RadioButtonState.CheckedPressed)]
+    public void RadioButtonRenderer_DrawRadioButton_OverloadWithSizeAndText(RadioButtonState state)
+    {
+        using Bitmap bitmap = new(100, 100);
+        using Graphics graphics = Graphics.FromImage(bitmap);
+        Point point = new Point(10, 20);
```
Tanya-Solyanik, line 31:
```suggestion
        Point point = new(10, 20);
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RadioButtonRendererTests.cs
```diff
@@ -0,0 +1,175 @@
...
+        Rectangle bounds = new Rectangle(10, 20, 30, 40);
```
Tanya-Solyanik, line 32:
```suggestion
        Rectangle bounds = new(10, 20, 30, 40);
```

This comment applies to multiple places in this file

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RadioButtonRendererTests.cs
```diff
@@ -0,0 +1,175 @@
...
+
+        RadioButtonRenderer.DrawRadioButton(graphics, point, bounds, "Text", SystemFonts.DefaultFont, false, state);
```
Tanya-Solyanik, line 34: Please add parameter name before `false`. This comment applies to multiple places in this file.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AbstractButtonBaseTests.cs
ricardobossan: In response to the [previous comment suggesting the application of expression-bodied syntax to inheriting methods][expression-bodied], I transformed the `CreateButton` method into a property. This decision was prompted by an attempt to apply expression-bodied syntax to override the `CreateButton()` method in inheriting classes, which revealed a limitation. The inheriting methods had their override keyword highlighted in red:

screenshot

I discovered that [abstract methods cannot declare a body using expression-bodied syntax](https://www.informit.com/articles/article.aspx?p=2414582). It appears that applying expression-bodied syntax to override abstract methods is not possible.

The motivation behind changing the method to an abstract property was to address this highlighted issue.

**Before the change**:

- AbstractControlBaseTests:

  image

- RadioButtonTests:

  image

**After the change**:

- AbstractControlBaseTests:

  image

- RadioButtonTests:

  image

In conclusion, due to the limitation in applying expression-bodied syntax to override abstract methods, I chose to change the `CreateButton` method to an abstract property. This addressed the issue where the override keyword was highlighted in red, as illustrated above.

It seems a choice between using a property named `ButtonUnderTest` or an abstract method called `CreateButton` that is overriden in inheriting classes with normal function body syntax.

For the moment, I'm switching back to abstract method, and overriding it in inheriting classes with regular method body, wich better conveys that the button should be disposed later.

[expression-bodied]: https://github.com/dotnet/winforms/pull/10693#discussion_r1485676758

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AbstractButtonBaseTests.cs
Tanya-Solyanik: I don't think your change went through.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBoxRendererTests.cs
```diff
@@ -119,7 +119,7 @@ public unsafe void Button_VisualStyles_off_Default_WithText_LineDrawing()
             return;
         }
 
-        using CheckBox checkBox = (CheckBox)CreateButton  ;
+        using CheckBox checkBox = (CheckBox)CreateButton()  ;
```
Tanya-Solyanik:
```suggestion
        using CheckBox checkBox = (CheckBox)CreateButton();
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AbstractButtonBaseTests.cs
```diff
@@ -0,0 +1,90 @@
...
+        control.FlatAppearance.BorderColor = expectedColor;
+
+        Assert.Equal(expectedColor, control.BackColor);
+        Assert.Equal(expectedColor, control.FlatAppearance.BorderColor);
+        Assert.Equal(expectedColor, control.FlatAppearance.CheckedBackColor);
+    }
+
+    public virtual void ButtonBase_OverChangeRectangle_Get(Type controlType, Appearance appearance, FlatStyle flatStyle)
+    {
+        using dynamic control = Activator.CreateInstance(controlType);
```
Tanya-Solyanik, line 48: Why doesn't CreateButton method work here?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBoxRendererTests.cs
```diff
@@ -0,0 +1,174 @@
...
+    [InlineData(TextFormatFlags.Default, CheckBoxState.CheckedNormal)]
+    [InlineData(TextFormatFlags.Default, CheckBoxState.MixedNormal)]
+    [InlineData(TextFormatFlags.GlyphOverhangPadding, CheckBoxState.MixedHot)]
+    [InlineData(TextFormatFlags.PreserveGraphicsTranslateTransform, CheckBoxState.CheckedPressed)]
+    [InlineData(TextFormatFlags.TextBoxControl, CheckBoxState.UncheckedNormal)]
+    public void CheckBoxRenderer_DrawCheckBox_OverloadWithTextFormat(TextFormatFlags textFormat, CheckBoxState state)
+    {
+        using Bitmap bitmap = new(100, 100);
+        using Graphics graphics = Graphics.FromImage(bitmap);
+        Point point = new (10, 20);
```
Tanya-Solyanik, line 47: Please remove the whitespace between the `new` and the opening parenthesis.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBoxTests.cs
```diff
@@ -527,4 +530,125 @@ internal override bool RaiseAutomationPropertyChangedEvent(UIA_PROPERTY_ID prope
...
+        using SubCheckBox checkBox = new()
+        {
+            UseMnemonic = useMnemonic,
+            Text = buttonText,
+        };
+        form.Controls.Add(checkBox);
+        form.Show();
+
+        bool result = checkBox.ProcessMnemonic(charCode);
+
```
Tanya-Solyanik, line 640: You could compare result to the expected value - 
Assert.Equal(useMnemonic, result);
Assert.Equal(useMnemonic && charCode != `&`, checkBox.Focused);

and also test that the checkbox is going into the checked state when mnemonic is typed in.

Assert.Equal(useMnemonic && charCode != `&`);

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RadioButtonRendererTests.cs
```diff
@@ -0,0 +1,175 @@
...
+    public unsafe void CaptureButtonOnForm()
+    {
+        using Form form = new();
+        using RadioButton radioButton = (RadioButton)CreateButton();
+        form.Controls.Add(radioButton);
+
+        using EmfScope emf = new();
+        form.PrintToMetafile(emf);
+
+        string details = emf.RecordsToString();
```
Tanya-Solyanik, line 171: Is it possible to add validation to `details`?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AbstractButtonBaseTests.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        control.FlatAppearance.CheckedBackColor = expectedColor;
+        control.FlatAppearance.BorderColor = expectedColor;
+
+        Assert.Equal(expectedColor, control.BackColor);
+        Assert.Equal(expectedColor, control.FlatAppearance.BorderColor);
+        Assert.Equal(expectedColor, control.FlatAppearance.CheckedBackColor);
+    }
+
+    public virtual void ButtonBase_OverChangeRectangle_Get(Type controlType, Appearance appearance, FlatStyle flatStyle)
```
Tanya-Solyanik, line 46:
```suggestion
    public virtual void ButtonBase_OverChangeRectangle_Get(Appearance appearance, FlatStyle flatStyle)
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBoxRenderingTests.cs
ricardobossan: I included those test cases initially to verify that the method could draw without errors, similar to the approach taken in [ControlPaintTests](https://github.com/dotnet/winforms/blob/main/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ControlPaintTests.cs). At the time, I didn't have a specific validation plan in mind. I'm open to suggestions on how we could enhance validation for this method. If you have any ideas or specific criteria in mind, I'd be happy to work on incorporating them into the test.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBoxRendererTests.cs
ricardobossan: This was modeled after an existing test case that was successfully merged into the repository ([ButtonRenderingTests.CaptureButtonOnForm()](https://github.com/dotnet/winforms/blob/389c8da733c43105275cfc9f9bc482b144073880/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ButtonRenderingTests.cs#L210)). Drawing from this precedent, I've reproduced the test to align with established practices. However, I'm unsure on how to implement further validation for it. Any suggestions would be greatly appreciated.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RadioButtonRendererTests.cs
ricardobossan: This was modeled after an existing test case that was successfully merged into the repository ([ButtonRenderingTests.CaptureButtonOnForm()](https://github.com/dotnet/winforms/blob/389c8da733c43105275cfc9f9bc482b144073880/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ButtonRenderingTests.cs#L210)). Drawing from this precedent, I've reproduced the test to align with established practices. However, I'm unsure on how to implement further validation for it. Any suggestions would be greatly appreciated.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBoxTests.cs
ricardobossan: I refactored it to make compare the result to the expected value in a clearer way, with fluent assertion syntax.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RadioButtonRendererTests.cs
ricardobossan: This was modeled after an existing test case that was successfully merged into the repository ([ButtonRenderingTests.CaptureButtonOnForm()](https://github.com/dotnet/winforms/blob/389c8da733c43105275cfc9f9bc482b144073880/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ButtonRenderingTests.cs#L210)). Drawing from this precedent, I've reproduced the test to align with established practices. However, I'm unsure on how to implement further validation for it. Any suggestions would be greatly appreciated.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBoxRenderingTests.cs
Tanya-Solyanik: We have metafile-based validation for [rendering tests](https://github.com/dotnet/winforms/blob/main/docs/testing.md#rendering-tests).

However in this case checkbox could be rendered differently based on the OS, so metafile validation will be unreliable. But please review other rendering test with metafile validation in mind.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RadioButtonRendererTests.cs
Tanya-Solyanik: Look up `emf.Validate` calls in this file.

Tanya-Solyanik (dismissed):

Tanya-Solyanik:
Please rebase on the latest.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ButtonRenderingTests.cs
```diff
@@ -222,11 +224,9 @@ public unsafe void CaptureButtonOnForm()
     [WinFormsFact]
     public unsafe void Button_FlatStyle_WithText_Rectangle()
     {
-        using Button button = new()
```
Tanya-Solyanik, line 225: The previous version is the preferred initialization style now because it works better with compiler optimizations.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBoxRendererTests.cs
```diff
@@ -0,0 +1,155 @@
...
+namespace System.Windows.Forms.Tests;
+
+public class CheckBoxRendererTests : AbstractButtonBaseTests
+{
+    [WinFormsTheory]
+    [InlineData(CheckBoxState.CheckedNormal)]
+    [InlineData(CheckBoxState.MixedNormal)]
+    public void CheckBoxRenderer_DrawCheckBox(CheckBoxState cBState)
+    {
+        using Form form = new Form();
```
Tanya-Solyanik, line 17: `using Form form = new();`

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBoxRendererTests.cs
```diff
@@ -0,0 +1,155 @@
...
+            );
+        }
+    }
+
+    [WinFormsTheory]
+    [InlineData(CheckBoxState.CheckedNormal)]
+    [InlineData(CheckBoxState.MixedNormal)]
+    public void CheckBoxRenderer_DrawCheckBox_OverloadWithSizeAndText(CheckBoxState cBState)
+    {
+        using Form form = new Form();
```
Tanya-Solyanik, line 48: use target-typed `new`

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ButtonRenderingTests.cs
Tanya-Solyanik: Never mind, you are now using a CreateButton method rather than a constructor.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RadioButtonRendererTests.cs
```diff
@@ -0,0 +1,140 @@
...
+        form.Handle.Should().NotBe(IntPtr.Zero);
+
+        using EmfScope emf = new();
+        DeviceContextState state = new(emf);
+        using Graphics graphics = Graphics.FromHdc((IntPtr)emf.HDC);
+        Point point = new(control.Location.X, control.Location.Y);
+        Rectangle bounds = control.Bounds;
+        control.Text = "Text";
+
+        RadioButtonRenderer.DrawRadioButton(graphics, point, bounds, control.Text, SystemFonts.DefaultFont, TextFormatFlags.Default, focus, RadioButtonState.CheckedNormal, HWND.Null);
```
Tanya-Solyanik, line 118: This line is too long, please break it up.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RadioButtonTests.cs
```diff
@@ -1625,4 +1625,17 @@ internal override bool RaiseAutomationPropertyChangedEvent(UIA_PROPERTY_ID prope
             return base.RaiseAutomationPropertyChangedEvent(propertyId, oldValue, newValue);
         }
     }
+
+    [WinFormsTheory]
+    [InlineData(Appearance.Button, FlatStyle.Standard)]
```
Tanya-Solyanik, line 1630: I would suggest creating a theory data with these values, as you use it in multiple places.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBoxRendererTests.cs
```diff
@@ -0,0 +1,152 @@
...
+
+        Point point = new(control.Location.X, control.Location.Y);
+        Rectangle bounds = control.Bounds;
+        control.Text = "Text";
+
+        CheckBoxRenderer.DrawCheckBox(graphics, point, bounds, control.Text, SystemFonts.DefaultFont, textFormat, false, cBState);
+
+        emf.Validate(
+            state,
+           Application.RenderWithVisualStyles
```
Tanya-Solyanik, line 100: This line should be indented by 1 space in

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckBoxRendererTests.cs
```diff
@@ -0,0 +1,152 @@
...
+        emf.Validate(
+            state,
+            Application.RenderWithVisualStyles
+                ? Validate.SkipType(ENHANCED_METAFILE_RECORD_TYPE.EMR_ALPHABLEND)
+                : Validate.Repeat(Validate.SkipType(ENHANCED_METAFILE_RECORD_TYPE.EMR_STRETCHDIBITS), 1),
+            Validate.TextOut(
+                control.Text,
+                    bounds: new Rectangle(3, 0, 20, 12),
+                    State.FontFace(SystemFonts.DefaultFont.Name)
+            ),
```
Tanya-Solyanik, line 141: lines 139 and 140 should be aligned with line 138

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/RadioButtonRendererTests.cs
```diff
@@ -0,0 +1,150 @@
...
+        emf.Validate(
+            state,
+            Application.RenderWithVisualStyles
+                ? Validate.SkipType(ENHANCED_METAFILE_RECORD_TYPE.EMR_ALPHABLEND)
+                : Validate.Repeat(Validate.SkipType(ENHANCED_METAFILE_RECORD_TYPE.EMR_STRETCHDIBITS), 1),
+            Validate.TextOut(
+                control.Text,
+                    bounds: new Rectangle(3, 0, 20, 12),
+                    State.FontFace(SystemFonts.DefaultFont.Name)
+            ),
```
Tanya-Solyanik, line 139: lines 137 and 138 should be aligned with line 136

Tanya-Solyanik (approved):

PR #10698, opened 2024-01-19

gpetrou:
## Proposed changes

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10698)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGrid.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGrid.cs
@@ -167,7 +167,7 @@ public partial class PropertyGrid : ContainerControl, IComPropertyBrowser, IProp
 
         try
         {
-            _gridView = CreateGridView(null);
+            _gridView = CreateGridView(serviceProvider: null);
             _gridView.TabStop = true;
             _gridView.MouseMove += OnChildMouseMove;
             _gridView.MouseDown += OnChildMouseDown;
@@ -955,7 +955,7 @@ public partial class PropertyGrid : ContainerControl, IComPropertyBrowser, IProp
                         continue;
                     }
 
-                    Type oldType = GetUnwrappedObject(i).GetType();
+                    Type oldType = GetUnwrappedObject(i)!.GetType();
 
                     object newObject = value[i];
 
@@ -1569,7 +1569,7 @@ public partial class PropertyGrid : ContainerControl, IComPropertyBrowser, IProp
             ShowEventsButton(false);
         }
     }
-#nullable disable
+
     /// <summary> Collapses all the nodes in the PropertyGrid</summary>
     public void CollapseAllGridItems()
         => _gridView.RecursivelyExpand(_rootEntry, initialize: false, expand: false, maxExpands: -1);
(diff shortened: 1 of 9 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGrid.cs
```diff
@@ -1987,8 +1987,13 @@ private static void DrawXorBar(Control targetControl, Rectangle rcFrame)
...
     }
 
-    private bool EnablePropPageButton(object obj)
+    private bool EnablePropPageButton(object? obj)
     {
+        if (_viewPropertyPagesButton is null)
+        {
+            throw new InvalidOperationException();
+        }
+
```
Epica3055, line 1996: why don't you add `!` below

src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGrid.cs
```diff
@@ -955,7 +955,7 @@ public object[] SelectedObjects
                         continue;
                     }
 
-                    Type oldType = GetUnwrappedObject(i).GetType();
+                    Type oldType = GetUnwrappedObject(i)!.GetType();
```
lonitra, line 958: Please add a comment as to why `GetUnwrappedObject` will not be null

src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGrid.cs
```diff
@@ -1684,7 +1684,7 @@ private PropertyTab CreateTab(Type tabType, IDesignerHost host)
                 else
                 {
                     // Just call the default constructor.
-                    tab = (PropertyTab)Activator.CreateInstance(tabType);
+                    tab = (PropertyTab)Activator.CreateInstance(tabType)!;
```
lonitra, line 1687: Would this be better since `tab` is type `PropertyTab?`
```suggestion
                    tab = (PropertyTab?)Activator.CreateInstance(tabType);
```

src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGrid.cs
lonitra: This is desired. If we cannot gauruntee that the variable is non null by reasoning with the current code or refactor the code so that the variable is non null at the point of use, we should throw either `InvalidOperationException` or some argument exception to avoid a null reference exception.

src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGrid.cs
```diff
@@ -2105,7 +2115,7 @@ private void AddLargeImage(Bitmap originalBitmap)
         try
         {
             Bitmap largeBitmap = ScaleHelper.CopyAndScaleToSize(originalBitmap, s_largeButtonSize);
-            _largeButtonImages.Images.Add(largeBitmap);
+            _largeButtonImages!.Images.Add(largeBitmap);
```
lonitra, line 2118: Please add a comment for why `_largeButtonImages` is non null here

src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGrid.cs
gpetrou: We have used null-forgiving operator with `Activator.CreateInstance` in most of our codebase.

lonitra (approved):
LGTM, thank you!

PR #10699, opened 2024-01-19
Add inheritdoc to PInvoke overloads

### LINKED ISSUE #10695: Add <inheritdoc> to all PInvoke overloads
Part of the value of using CsWin32 is that it has links to the docs. We lose that when we add "overloads" for `PInvoke` defintions and don't use `<inheritdoc cref=""/>` from the original we're calling.

We should make sure that we've added this for all helpers we've created.

willibrandon:
Fixes #10695

## Proposed changes

- Add `<ineritdoc>` to PInvoke overloads to get links to the docs from CsWin32.

## Customer Impact

- Improve the developer experience.

## Regression? 

- No

## Risk

- None.

## Test methodology 

- Manual.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10699)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.ChildWindowFromPointEx.cs
+++ b/src/System.Windows.Forms.Primitives/src/Windows/Win32/PInvoke.ChildWindowFromPointEx.cs
@@ -7,6 +7,7 @@ namespace Windows.Win32;
 
 internal static partial class PInvoke
 {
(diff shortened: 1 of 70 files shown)


JeremyKuhne (approved):
Awesome, thank you!

PR #10701, opened 2024-01-20

gpetrou:
## Proposed changes

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10701)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGridInternal/GridEntry.EventEntry.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGridInternal/GridEntry.EventEntry.cs
@@ -7,11 +7,11 @@ internal abstract partial class GridEntry
 {
     private sealed class EventEntry
     {
-        public Delegate Handler { get; set; }
+        public Delegate? Handler { get; set; }
         public object Key { get; }
-        public EventEntry Next { get; set; }
+        public EventEntry? Next { get; set; }
 
-        internal EventEntry(EventEntry next, object key, Delegate handler)
+        internal EventEntry(EventEntry? next, object key, Delegate? handler)
         {
             Next = next;
             Key = key;
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGridInternal/GridEntry.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGridInternal/GridEntry.cs
@@ -1441,7 +1441,7 @@ internal abstract partial class GridEntry : GridItem, ITypeDescriptorContext
     /// </summary>
     public virtual object GetService(Type serviceType)
         => serviceType == typeof(GridItem) ? this : (_parent?.GetService(serviceType));
-
+#nullable enable
     /// <summary>
     ///  Paints the label portion of this <see cref="GridEntry"/> into the given <see cref="Graphics"/> object.
     /// </summary>
@@ -1615,6 +1615,11 @@ internal abstract partial class GridEntry : GridItem, ITypeDescriptorContext
                 g.FillRectangle(brush, outline);
             }
 
+            if (g is null)
+            {
+                throw new InvalidOperationException();
+            }
+
             if (ColorInversionNeededInHighContrast || !expanded)
             {
                 VisualStyleElement element = expanded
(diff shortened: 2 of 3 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGridInternal/GridEntry.cs
```diff
@@ -2002,17 +2012,17 @@ internal bool ShouldSerializePropertyValue()
...
     {
-        PropertyDescriptor[] newProperties = null;
+        PropertyDescriptor[]? newProperties = null;
         int newPosition = 0;
 
         // First scan the list and move any parenthesized properties to the front.
         for (int i = 0; i < props.Length; i++)
         {
-            if (props[i].GetAttribute<ParenthesizePropertyNameAttribute>().NeedParenthesis)
+            if (props[i].GetAttribute<ParenthesizePropertyNameAttribute>()!.NeedParenthesis)
```
lonitra, line 2021: Please add a comment for why this is nonnull

src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGridInternal/GridEntry.cs
```diff
@@ -2023,11 +2033,11 @@ private static PropertyDescriptor[] SortParenProperties(PropertyDescriptor[] pro
             {
                 if (props[i] is not null)
                 {
-                    newProperties[newPosition++] = props[i];
+                    newProperties![newPosition++] = props[i];
                 }
             }
 
-            props = newProperties;
+            props = newProperties!;
```
lonitra, line 2040: Consider changing the if statement condition `newPosition > 0` to `newProperties is not null` to avoid these `!`. This should be ok since if `newProperties` is nonnull, `newPosition` should also be > 0

src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGridInternal/GridEntry.cs
```diff
@@ -2158,7 +2168,7 @@ private void ResetState()
     /// <summary>
     ///  Sets the value of this <see cref="GridEntry"/> from text.
     /// </summary>
-    public bool SetPropertyTextValue(string text)
+    public bool SetPropertyTextValue(string? text)
     {
```
ricardobossan, line 2172: If text is null here, shouldn't it return false?

If so, maybe add

```cs
if (text is null)
{
    return false;// or  throw new ArgumentNullException();
}
```

src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGridInternal/GridEntry.EventEntry.cs
```diff
@@ -7,11 +7,11 @@ internal abstract partial class GridEntry
 {
     private sealed class EventEntry
```
ricardobossan, line 8: Why not convert those to full properties?

src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGridInternal/GridEntry.cs
gpetrou: We use `text` to set `PropertyValue`.

src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGridInternal/GridEntry.EventEntry.cs
gpetrou: Sorry, I don't understand what you mean.

ricardobossan: Thanks for your patience. I've reviewed the coding style documentation and now see the recommendation for using auto-implemented properties when possible. My previous suggestion to consider full properties was based on a misunderstanding, and I apologize for any confusion.

lonitra (approved):

PR #10707, opened 2024-01-21
7027 : fix treeview sorter issue when treeviewnodesorter is null set …

### LINKED ISSUE #7027: TreeView.Sorted Property has Attribute [EditorBrowsable(EditorBrowsableState.Never)]
### .NET version

6.0 / 4.8

No

### Issue description

If you have a TreeView control, and you set TreeView.TreeViewNodeSorter = new TreeNodeComparer().
Then .net will set TreeView.Sorted=true.

https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.treeview.treeviewnodesorter

But if you set  TreeView.TreeViewNodeSorter = null, Then .net will not set TreeView.Sorted=false. This means .net **will still** sort all nodes.

So if you have set  TreeView.TreeViewNodeSorter = new TreeNodeComparer(), you may need to call two steps:
1. Set  TreeView.TreeViewNodeSorter = null
2. Set TreeView.Sorted=false.

Then you can cancel the automatic sorting function of TreeView.

But TreeView.Sorted Property has Attribute [EditorBrowsable(EditorBrowsableState.Never)], **So you may not find** this property in a text editor.

https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.treeview.sorted

### Steps to reproduce

Full Test .net code: .net 6.0 with Visual studio 2022

```
using System.Collections;

namespace TreeViewNodeSorterTest
{
    internal static class Program
    {
        [STAThread]
        static void Main()
        {
            Test();
        }

        private class TreeNodeComparer : IComparer
        {
            public int Compare(object x, object y)
            {
                throw new NotImplementedException();
            }
        }

        private static void Test()
        {
            TreeView view = new TreeView();

            MessageBox.Show("Sorted: " + view.Sorted);

            view.TreeViewNodeSorter = new TreeNodeComparer();

            MessageBox.Show("Sorted: " + view.Sorted);

            view.TreeViewNodeSorter = null;

            MessageBox.Show("Sorted: " + view.Sorted);
        }
    }
}
```

RutulPatel8:
Fixes #7027
If you have a TreeView control, and you set TreeView.TreeViewNodeSorter = new TreeNodeComparer().
Then .net will set TreeView.Sorted=true.

## Proposed changes
- when we set TreeViewNodeSorter null system will update Sorted = false

## Customer Impact

- No
- 

## Regression? 

- Yes

## Risk

- No

### Before

- When we set TreeViewNodeSorter to null after previously assigning a value to it, it still appears to be sorted as true even after updating TreeViewNodeSorter with null.

### After

- When we set TreeViewNodeSorter to null after previously assigning a value to it, it is set sorted as false.

## Test methodology 

- Added testcase 

## Test environment(s) 

- tested with .net core 8.0

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10707)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeView.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeView.cs
@@ -1256,6 +1256,11 @@ public partial class TreeView : Control
                 {
                     Sort();
                 }
+                else
+                {
+                    Sorted = false;
+                    RefreshNodes();
+                }
             }
         }
     }
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TreeViewTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TreeViewTests.cs
@@ -4723,6 +4723,19 @@ public class TreeViewTests
     }
 
     [WinFormsFact]
+    public void TreeViewNodeSorter_Set_SortedFalseIfNull()
+    {
+        using var treeView = new TreeView
+        {
+            TreeViewNodeSorter = StringComparer.CurrentCulture
+        };
+        Assert.True(treeView.Sorted);
+
+        treeView.TreeViewNodeSorter = null;
+        Assert.False(treeView.Sorted);
+    }
+
+    [WinFormsFact]
     public void AddExistingNodeAsChild_ThrowsArgumentException()
     {
         using TreeView treeView = new();
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TreeViewTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TreeViewTests.cs
@@ -4711,7 +4711,7 @@ public class TreeViewTests
     [MemberData(nameof(TreeViewNodeSorter_TestData))]
     public void TreeViewNodeSorter_Set_GetReturnsExpected(IComparer value)
     {
-        using TreeView treeView = new()
+        using TreeView treeView = new ()
         {
             TreeViewNodeSorter = value
         };
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TreeViewTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TreeViewTests.cs
@@ -4725,7 +4725,7 @@ public class TreeViewTests
     [WinFormsFact]
     public void TreeViewNodeSorter_Set_SortedFalseIfNull()
     {
-        using var treeView = new ()
+        using TreeView treeView = new ()
         {
             TreeViewNodeSorter = StringComparer.CurrentCulture
         };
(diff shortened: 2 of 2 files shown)


RutulPatel8:
> @dotnet-policy-service agree

RutulPatel8:
> @dotnet-policy-service agree

@dotnet-policy-service agree

src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeView.cs
```diff
@@ -1256,6 +1256,11 @@ public bool Sorted
                 {
                     Sort();
                 }
+                else
+                {
+                    Sorted = false;
+                    RefreshNodes();
```
lonitra, line 1262: Is `RefreshNodes()` needed here? When we set `Sorted = false`, `RefreshNodes()` should've been called already I believe.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TreeViewTests.cs
```diff
@@ -4722,6 +4722,19 @@ public void TreeViewNodeSorter_Set_GetReturnsExpected(IComparer value)
         Assert.Same(value, treeView.TreeViewNodeSorter);
     }
 
+    [WinFormsFact]
+    public void TreeViewNodeSorter_Set_SortedFalseIfNull()
+    {
+        using var treeView = new TreeView
```
lonitra, line 4728: Prefer target type new:
```suggestion
        using TreeView treeView = new()
```

src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeView.cs
RutulPatel8: Yes, 
```RefreshNodes()```  is required here because when we set ```Sorted = false``` , ```RefreshNode()``` is not called inside the property due to Sorted being set to false. Therefore, it is necessary to include it here.

MS-WB-Forms-1

Comment (lonitra) [reply src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeView.cs:1262]: Whoops! You’re right. Thanks for checking that!

File: src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TreeViewTests.cs [hunk 3]
```diff
@@ -4711,7 +4711,7 @@ public static IEnumerable<object[]> TreeViewNodeSorter_TestData()
     [MemberData(nameof(TreeViewNodeSorter_TestData))]
     public void TreeViewNodeSorter_Set_GetReturnsExpected(IComparer value)
     {
-        using TreeView treeView = new()
+        using TreeView treeView = new ()
```
lonitra, line 4714: Please remove the space between ‘new’ and ‘()’. We typically do not have a space between those when doing target type new in this repo.
```suggestion
        using TreeView treeView = new()
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TreeViewTests.cs
```diff
@@ -4722,6 +4722,19 @@ public void TreeViewNodeSorter_Set_GetReturnsExpected(IComparer value)
         Assert.Same(value, treeView.TreeViewNodeSorter);
     }
 
+    [WinFormsFact]
+    public void TreeViewNodeSorter_Set_SortedFalseIfNull()
+    {
+        using TreeView treeView = new ()
```
lonitra, line 4728: Ditto here as well

RutulPatel8:
@lonitra have you review my PR again?

PR #10710, opened 2024-01-22
Add test if running in DesignMode to the MessageBox.Show method

### LINKED ISSUE #10677: VS is not responding after adding a user control that with events handling to the Form designer
### .NET version

9.0.100-alpha.1.24066.10

Not tested/verified

_No response_

### Issue description

VS is not responding after adding a user control that with events handling to the Form designer.
84885709-fee9c300-b0c5-11ea-986c-1c4ae0352f57

### Steps to reproduce

1. Create a Winforms core project.
2. Add a User Control to the same project.
3. Generate the UserControl1_Load event with following code: MessageBox.Show("Not Responding");
4. Build and reopen the Form designer.
5. Drag the user control to the Form designer.

Or

1. Open the attached project. [NotResponding.zip](https://github.com/dotnet/winforms-designer/files/4791793/NotResponding.zip)
2. Build and Drag the user control to the Form designer.

SimonZhao888:
Fixes #10677

## Proposed changes

- Add checking CanNotifyClients to the MessageBox.Show method, to ensure that the MessageBox is not popped up at DesignMode.

## Customer Impact

- In DesignMode, the MessageBox will not pop up, and at runtime, the MessageBox will pop up normally.

## Regression? 

- Yes

## Risk

- Minimal

### Before
VS is not responding after adding a user control that with events handling to the Form designer.

297284676-e19ca3bd-f364-40ff-bc57-ec83e49adcec

### After
No MessageBox pop up when adding a user control with events handing to the Form designer.
DesignMode

Runtime Mode
MessageBox will pop up when we run the Form designer.
RuntimeMode

## Test methodology 

- Manually

## Test environment(s) 

- .net 9.0.0-alpha.1.24058.11

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10710)


src/System.Windows.Forms/src/System/Windows/Forms/Dialogs/MessageBox.cs
```diff
@@ -446,6 +446,11 @@ public static DialogResult Show(IWin32Window? owner, string? text)
         MessageBoxOptions options,
         bool showHelp)
     {
+        if (LicenseManager.UsageMode == LicenseUsageMode.Designtime)
```
Tanya-Solyanik, line 449: Had you tested this code?
FYI: Leaf had recently tested her fix in the designer context - https://github.com/dotnet/winforms/pull/10647#issuecomment-1896441688.

SimonZhao888: > Had you tested this code? FYI: Leaf had recently tested her fix in the designer context - [#10647 (comment)](https://github.com/dotnet/winforms/pull/10647#issuecomment-1896441688).

Yes, but it didn't pass on our local machine, without CurrentContext the value of UsageMode will be Runtime.
https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.licensemanager.usagemode?view=net-8.0#remarks

Tanya-Solyanik: Yeah, I don't think LicenseManager is actually working

SimonZhao888: > Yeah, I don't think LicenseManager is actually working

I don't really understand this method, it seems like he just handles AccessibleObject, so why would it work for MessageBox as well?
https://github.com/dotnet/winforms/blob/9ba3ad31fdf187f8c82985ba6ff5a9703922a747/src/System.Windows.Forms/src/System/Windows/Forms/Dialogs/MessageBox.cs#L505

src/System.Windows.Forms/src/System/Windows/Forms/Dialogs/MessageBox.cs
```diff
@@ -491,4 +491,22 @@ public static DialogResult Show(IWin32Window? owner, string? text)
             PInvoke.SendMessage(handle, PInvoke.WM_SETFOCUS);
         }
     }
+
+    private static bool s_canNotifyClients => InitializeCanNotifyClients();
```
Tanya-Solyanik, line 495: This code should be factored out into a common helper class. Please search if we have an appropriate class to use.

src/System.Windows.Forms/src/System/Windows/Forms/Dialogs/MessageBox.cs
```diff
@@ -486,4 +491,22 @@ public static DialogResult Show(IWin32Window? owner, string? text)
...
+
+    private static bool InitializeCanNotifyClients()
+    {
+        // While handling accessibility events, accessibility clients (JAWS, Inspect),
+        // can access AccessibleObject associated with the event. In the designer scenario, controls are not
+        // receiving messages directly and might not respond to messages while in the notification call.
+        // This will make the server process unresponsive and will cause VisualStudio to become unresponsive.
+        //
+        // The following compat switch is set in the designer server process to prevent controls from sending notification.
+        if (AppContext.TryGetSwitch("Switch.System.Windows.Forms.AccessibleObject.NoClientNotifications", out bool isEnabled))
```
lonitra, line 505: Where is this switch defined?

Tanya-Solyanik: It's set in the DesignToolsServer.exe on startup

JeremyKuhne:
`AppContext` switches should be added to [LocalAppContextSwitches.cs](https://github.com/dotnet/winforms/blob/daf3f4db5de296ac78b51d279e86a38c62255e0b/src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs)

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -21,13 +21,15 @@ internal static partial class LocalAppContextSwitches
...
     internal const string DataGridViewUIAStartRowCountAtZeroSwitchName = "System.Windows.Forms.DataGridViewUIAStartRowCountAtZero";
+    private const string WindowsFormsAccessibleObjectNoClientNotificationsSwitchName = "Switch.System.Windows.Forms.AccessibleObject.NoClientNotifications";
 
     private static int s_scaleTopLevelFormMinMaxSizeForDpi;
     private static int s_anchorLayoutV2;
     private static int s_servicePointManagerCheckCrl;
     private static int s_trackBarModernRendering;
     private static int s_doNotCatchUnhandledExceptions;
     private static int s_dataGridViewUIAStartRowCountAtZero;
+    private static int s_canNotifyClients;
```
Tanya-Solyanik, line 32: Please reverse the existing logic to match the pattern that other switches are following.
Switch name - NoClientNotifications
Backing static field - s_noClientNotifications
Internal property - NoClientNotifications

This will require logic reversal at all call sites. But will simplify code maintenance and initialization function.

src/System.Windows.Forms/src/System/Windows/Forms/Accessibility/AccessibleObject.cs
```diff
@@ -143,23 +143,7 @@ private AccessibleObject(AgileComPointer<UIA.IAccessible> accessible)
...
-        //
-        // The following compat switch is set in the designer server process to prevent controls from sending notification.
-        if (AppContext.TryGetSwitch("Switch.System.Windows.Forms.AccessibleObject.NoClientNotifications", out bool isEnabled))
-        {
-            return !isEnabled;
-        }
-
-        return true;
-    }
+    internal static bool CanNotifyClients => LocalAppContextSwitches.CanNotifyClients;
```
Tanya-Solyanik, line 146: code can directly the property from LocalAppContextSwitches class, it is visible to the same callers.

Tanya-Solyanik:
@SimonZhao888  - please test design time scenarios, install NET9 private binaries, create a .NET9 winforms application , open designer, and attach to the DesignToolsServer to verifuy the switch value.

SimonZhao888:
> @SimonZhao888 - please test design time scenarios, install NET9 private binaries, create a .NET9 winforms application , open designer, and attach to the DesignToolsServer to verifuy the switch value.

Hi Tanya, I test it in the design mode, and found that the value we set for the switch will always be true whether we set it to true or false.
And in runtime scenarios, the results are same with Olina's team.

Tanya-Solyanik:
>the value we set for the switch will always be true whether we set it to true or false.

This is expected

Tanya-Solyanik (commented):
Had you investigated how to add a unit test for MessageBox.Show?

SimonZhao888:
> Had you investigated how to add a unit test for MessageBox.Show?

We have only added a test case for one scenario(Switch.System.Windows.Forms.AccessibleObject.NoClientNotifications=true), in all other scenarios, there will be message boxes popping up that will prevent our test case from running.

Do we still need to add test cases？

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/MessageBoxTests.cs
```diff
@@ -135,6 +136,14 @@ public void MessageBox_MessageBoxDefaultButton_Valid(MessageBoxDefaultButton val
                 | (MESSAGEBOX_STYLE)MessageBoxOptions.DefaultDesktopOnly);
     }
 
+    [WinFormsTheory]
```
Tanya-Solyanik, line 139: Change this test to a fact

Tanya-Solyanik (commented):
Looks good, added a minor comment

SimonZhao888:
> We don't need more tests, this is the important case to test

I'm now using Mock to test the MessageBox without the UI popping up, and the test scenarios are more refined.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/MessageBoxTests.cs
SimonZhao888: Done, and move this unit test to UI test project.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -160,5 +167,28 @@ public static bool DataGridViewUIAStartRowCountAtZero
...
+            s_noClientNotifications = value ? 1 : 0;
+        }
+
+        if (switchName == DataGridViewUIAStartRowCountAtZeroSwitchName)
+        {
+            s_dataGridViewUIAStartRowCountAtZero = value ? 1 : 0;
+        }
+    }
+
+    internal static bool GetCachedSwitchValue(string switchName)
```
Tanya-Solyanik, line 189: Where is this method used?

SimonZhao888: This is used to get the default value of the switch so that it can be reset after the unit tests have run.
https://github.com/dotnet/winforms/blob/89a2460850825e96bbf8449d357c5a7ab46cf585/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/MessageBoxTests.cs#L13

Tanya-Solyanik (approved):

PR #10716, opened 2024-01-22
Clarify ImageConverter

JeremyKuhne:
ImageConverter has some conversion code with somewhat magic numbers in it. This change documents the format it's looking for and validates it more throroughly.

It also adds a new high-performance Span reader class for scenarios like this that allow bounds checked iterating through data. It makes code written for this much easier to visually validate as opposed to slicing through spans manually.

Also:

- Use SpanReader in Icon parsing code
- Fix ordering of class keywords in a number of files
- Add readonly modifier to some fields
- Minor comment cleanup
- Move ICONDIR and ICONDIRENTRY to interop folder
- Delete ENHMETAHEADER
- Move SpanHelpers
- Move POINTS to Core
- Delete unused  code in ARGB
- Move method from ARGB to ArgbBuffer

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10716)


src/System.Private.Windows.Core/src/Windows/Win32/UI/WindowsAndMessaging/ICONDIR.cs
```diff
@@ -0,0 +1,20 @@
...
+public struct ICONDIR
+{
+    // Must be 0
+    public ushort idReserved;
+    // Must be 1
+    public ushort idType;
+    // Count of entries
+    public ushort idCount;
+    // First entry (anysize array)
+    // public ICONDIRENTRY idEntries;
```
lonitra, line 19: Is this supposed to be commented out?

JeremyKuhne: Yes, so the reader only reads the other data.

lonitra (approved):
:shipit:

PR #10720, opened 2024-01-23
fix bug: No text for new tabpage item that added by TabPages property in the properties window in the DemoConsole application

### LINKED ISSUE #10545: No text for new tabpage item that added by TabPages property in the properties window in the DemoConsole application
### .NET version

9.0.0-alpha.1.23621.3

Not tested/verified

The issue also can reproduce .NET 8.0, which start support the TabControl.

### Issue description

No text for new tabpage item added by TabPages property in the properties window in the DemoConsole application

### Steps to reproduce

1. Test sample: [DemoConsole](https://github.com/dotnet/winforms/tree/main/src/System.Windows.Forms/tests/IntegrationTests/DesignSurface/DemoConsole) in Winforms repo
2. Add new tabpage item by TabPages property in the properties window 

**Actually Results:**
No text for new tabpage item added by TabPages property in the properties window
image
TabControl

**Expected Results:**
Have text for new tabpage item added by TabPages property in the properties window

**More info:**
The issue can't reproduce when added new tabpage item by using Command/SmartTag

Epica3055:
Fixes #10545

## Root cause.

- 
-  `PropertyGrid` requires `DesignerHost` to perform some functions . in our case it is to CreateNewObject
- 

https://github.com/dotnet/winforms/blob/a05dd925fe03e64ee026ad2fda65638b75d08998/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/CollectionEditor.cs#L91-L114

please see line 95. if the `host` here is not null, then our existing code can create target object **and set `Text` property(line 102)**

if the `host` is null, some properties would not be initialized inclining `Text`, which is the cause of this issue.

## Risk

- low

### Before

TabControl

### After

https://github.com/dotnet/winforms/assets/135201996/c6e1f5bf-d614-41e2-ba07-ca82902a53e1

## Test methodology 

- 
- manual
- 

## Test environment(s) 

9.0.0-alpha.1.24066.33

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10720)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/CollectionEditor.CollectionEditorCollectionForm.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/CollectionEditor.CollectionEditorCollectionForm.cs
@@ -301,6 +301,12 @@ public partial class CollectionEditor
                 {
                     AddItems(multipleInstance);
                 }
+
+                if (instance is IComponent component &&
+                    (Context?.Instance as Control)?.Site?.GetService<IDesignerHost>()?.GetDesigner(component) is IComponentInitializer initializer)
+                {
+                    initializer.InitializeNewComponent(defaultValues: null);
+                }
             }
             catch (Exception e)
             {
@@ -374,10 +380,10 @@ public partial class CollectionEditor
 
         private void HookEvents()
         {
-            _listbox.KeyDown += Listbox_keyDown;
-            _listbox.DrawItem += Listbox_drawItem;
-            _listbox.SelectedIndexChanged += Listbox_SelectedIndexChanged;
-            _listbox.HandleCreated += Listbox_HandleCreated;
+            _listbox.KeyDown += ListBox_keyDown;
+            _listbox.DrawItem += ListBox_drawItem;
+            _listbox.SelectedIndexChanged += ListBox_SelectedIndexChanged;
+            _listbox.HandleCreated += ListBox_HandleCreated;
             _upButton.Click += UpButton_Click;
             _downButton.Click += DownButton_click;
             _propertyGrid.PropertyValueChanged += PropertyGrid_propertyValueChanged;
@@ -552,9 +558,9 @@ public partial class CollectionEditor
         }
 
         /// <summary>
-        ///  This draws a row of the listbox.
+        ///  This draws a row of the listBox.
         /// </summary>
-        private void Listbox_drawItem(object? sender, DrawItemEventArgs e)
+        private void ListBox_drawItem(object? sender, DrawItemEventArgs e)
         {
             if (e.Index != -1)
             {
@@ -621,7 +627,7 @@ public partial class CollectionEditor
                         e.Bounds with { X = e.Bounds.X + offset, Width = e.Bounds.Width - offset });
                 }
 
-                // Check to see if we need to change the horizontal extent of the listbox
+                // Check to see if we need to change the horizontal extent of the listBox
                 int width = offset + (int)g.MeasureString(itemText, Font).Width;
                 if (width > e.Bounds.Width && _listbox.HorizontalExtent < width)
                 {
@@ -633,9 +639,9 @@ public partial class CollectionEditor
         /// <summary>
         ///  Handles keypress events for the list box.
         /// </summary>
-        private void Listbox_keyDown(object? sender, KeyEventArgs kevent)
+        private void ListBox_keyDown(object? sender, KeyEventArgs keyEvent)
         {
-            switch (kevent.KeyData)
+            switch (keyEvent.KeyData)
             {
                 case Keys.Delete:
                     PerformRemove();
@@ -649,7 +655,7 @@ public partial class CollectionEditor
         /// <summary>
         ///  Event that fires when the selected list box index changes.
         /// </summary>
-        private void Listbox_SelectedIndexChanged(object? sender, EventArgs e)
+        private void ListBox_SelectedIndexChanged(object? sender, EventArgs e)
         {
             UpdateEnabled();
         }
@@ -657,7 +663,7 @@ public partial class CollectionEditor
         /// <summary>
         ///  Event that fires when the list box's window handle is created.
         /// </summary>
-        private void Listbox_HandleCreated(object? sender, EventArgs e)
+        private void ListBox_HandleCreated(object? sender, EventArgs e)
         {
             UpdateItemWidths(null);
         }
@@ -845,7 +851,7 @@ public partial class CollectionEditor
         {
             _dirty = true;
 
-            // Refresh selected listbox item so that it picks up any name change
+            // Refresh selected listBox item so that it picks up any name change
             SuspendEnabledUpdates();
             try
             {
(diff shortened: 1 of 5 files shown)


src/System.Windows.Forms.Design/src/System/ComponentModel/Design/CollectionEditor.CollectionEditorCollectionForm.cs
```diff
@@ -633,9 +639,9 @@ private void Listbox_drawItem(object? sender, DrawItemEventArgs e)
         /// <summary>
         ///  Handles keypress events for the list box.
         /// </summary>
-        private void Listbox_keyDown(object? sender, KeyEventArgs kevent)
+        private void ListBox_keyDown(object? sender, KeyEventArgs keyEvent)
```
Tanya-Solyanik, line 642: we should match parameter names used in the public API - https://learn.microsoft.com/dotnet/api/system.windows.forms.keyeventhandler?view=windowsdesktop-8.0

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/CollectionEditor.CollectionEditorCollectionForm.cs
```diff
@@ -374,10 +380,10 @@ private static string GetDisplayText(ListItem? item)
 
         private void HookEvents()
         {
-            _listbox.KeyDown += Listbox_keyDown;
-            _listbox.DrawItem += Listbox_drawItem;
-            _listbox.SelectedIndexChanged += Listbox_SelectedIndexChanged;
-            _listbox.HandleCreated += Listbox_HandleCreated;
+            _listbox.KeyDown += ListBox_keyDown;
```
Tanya-Solyanik, line 383: Consider renaming `_listbox` to `_listBox`

src/System.Windows.Forms/tests/IntegrationTests/DesignSurface/DemoConsole/MainForm.cs
```diff
@@ -315,7 +315,10 @@ private void SelectRootComponent()
         // - find out the DesignSurfaceExt control hosted by the TabPage
         IDesignSurfaceExt isurf = _listOfDesignSurface[tabControl1.SelectedIndex];
         if (isurf is not null)
+        {
             propertyGrid.SelectedObject = isurf.GetIDesignerHost().RootComponent;
+            propertyGrid.Site = isurf.GetIDesignerHost().RootComponent.Site;
```
Tanya-Solyanik, line 320: I don't think that PropertyBrowser in the InProc designer has its Site set. Had you looked it up? I would expect that Designer Host should come from the DesignSurfaceExt.
What is the type of `Context` property in the CollectionEditor.CreateInstance?

Tanya-Solyanik: What is the value of PropertyGrid.ActiveDesigner when you run Demo console?

Epica3055: > I don't think that PropertyBrowser in the InProc designer has its Site set. Had you looked it up? I would expect that Designer Host should come from the DesignSurfaceExt.

I didn't check InProc PropertyBrowser Site ,I may check it latter.

I just want to set DesignerHost to PropertyGrid , but there is no way to set it directly because it is private field or internal method.

I observed that setting `site` could set it's own DesignerHost.

https://github.com/dotnet/winforms/blob/17fd2a33ecb8f56c03a8798b5025f115bdea3d58/src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGrid.cs#L1154-L1169

https://github.com/dotnet/winforms/blob/17fd2a33ecb8f56c03a8798b5025f115bdea3d58/src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGrid.cs#L248-L308

> What is the value of PropertyGrid.ActiveDesigner when you run Demo console?

Null.

Tanya-Solyanik: >I didn't check InProc PropertyBrowser Site ,I may check it latter.

This is important because components that have Site behave differently than the ones that don't have Site. Components that have Site behave as components on the design surface and they might have some functionality disabled or altered.

Tanya-Solyanik: Is there other way for PropertyGrid.GetService<IDesignerHost> to get the DesignerHost?

Epica3055: Somehow I found it that in InProc it is `PropertyGridHolder`, which descend from `PropertyGrid`. And it has a `Site` in our case.

> Is there other way for PropertyGrid.GetService to get the DesignerHost?

I don't think we can set `DesignerHost` directly, because `PropertyGrid` doesn't expose any public mehtod to set `DesignerHost`.

or we can use `InternalsVisibleTo` to set its `ActiveDesigner`,
https://github.com/dotnet/winforms/blob/17fd2a33ecb8f56c03a8798b5025f115bdea3d58/src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGrid.cs#L248-L308
 but I don't know if it is acceptable.

Epica3055: Another thing is:
I can do something like InProc designer.

``` c#
namespace DesignSurfaceExt;
public class PropertyGridExt: PropertyGrid
{
    public IDesignerHost DesignerHost { get; set; }

    protected override object GetService(Type serviceType)
    {
        object service = DesignerHost?.GetService(serviceType);
        return service ?? base.GetService(serviceType);
    }
}
```

and we use `PropertyGridExt` instead of `PropertyGrid`. 
But that just doesn't work. because we have several designers in DemoConsole.

https://github.com/dotnet/winforms/blob/00b11a73177daf5687673640b3629e51e7a9c4d3/src/System.Windows.Forms/tests/IntegrationTests/DesignSurface/DemoConsole/MainForm.cs#L21-L52

Epica3055: But I can bring up another solution.

We can use `PropertyGridExt` mentioned above. And make the `PropertyGridExt` multiple-instance. i.e. one DesignSurface for one `PropertyGridExt`
image

Tanya-Solyanik: >make the PropertyGridExt multiple-instance. i.e. one DesignSurface for one PropertyGridExt

Good idea! Is it possible to **re-initialize** PropertyGridExt when we click the tab to switch to a different DesignSurafaceEx instead of **re-allocating** it? For example, ` propetyGridExt.DesignSurface = <the new design surface we switched to>`;

```
namespace DesignSurfaceExt;
public class PropertyGridExt: PropertyGrid
{
    public DesignerSurfaceEx DesignSurface;

    protected override object GetService(Type serviceType)
    {
        object service = DesignSurface.GetIDesignerHost()?.GetService(serviceType);
        return service ?? base.GetService(serviceType);
    }
}
```

Epica3055: > For example,  `propetyGridExt.DesignSurface = <the new design surface we switched to>;`

this didn't work. I think it is because `PropetyGrid._designerHost` never get a chance to get updated.

PR #10728, opened 2024-01-24
Add MiscValue to UserControlWithObjectCollectionEditor to enable manual testing of "misc" label on PropertyGrid

elachlan:
Helps with testing the issue in https://github.com/dotnet/runtime/issues/94365

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10728)

### DIFF
--- a/src/System.Windows.Forms/src/SRCategoryAttribute.cs
+++ b/src/System.Windows.Forms/src/SRCategoryAttribute.cs
@@ -8,7 +8,6 @@ namespace System.Windows.Forms;
 [AttributeUsage(AttributeTargets.All)]
 internal sealed class SRCategoryAttribute : CategoryAttribute
 {
-    private static volatile SRCategoryAttribute? s_defaultAttribute;
     public SRCategoryAttribute(string category) : base(category)
     {
     }
@@ -45,9 +44,4 @@ internal sealed class SRCategoryAttribute : CategoryAttribute
         "WindowStyle" => SR.CatWindowStyle,
         _ => category
     };
-
-    public static new SRCategoryAttribute Default
-    {
-        get => s_defaultAttribute ??= new SRCategoryAttribute("PropertyCategoryMisc");
-    }
 }
--- a/src/System.Windows.Forms/src/SRCategoryAttribute.cs
+++ b/src/System.Windows.Forms/src/SRCategoryAttribute.cs
@@ -8,6 +8,7 @@ namespace System.Windows.Forms;
 [AttributeUsage(AttributeTargets.All)]
 internal sealed class SRCategoryAttribute : CategoryAttribute
 {
+    private static volatile SRCategoryAttribute? s_defaultAttribute;
     public SRCategoryAttribute(string category) : base(category)
     {
     }
@@ -15,5 +16,38 @@ internal sealed class SRCategoryAttribute : CategoryAttribute
     protected override string? GetLocalizedString(string value)
     {
         return SR.GetResourceString(value);
+    }
+
+    public static string GetLocalizedCategory(string category) => category switch
+    {
+        "Accessibility" => SR.CatAccessibility,
+        "Action" => SR.CatAction,
+        "Appearance" => SR.CatAppearance,
+        "Asynchronous" => SR.CatAsynchronous,
+        "Behavior" => SR.PropertyCategoryBehavior,
+        //"Config" => SR.PropertyCategoryConfig,
+        "Data" => SR.PropertyCategoryData,
+        "DDE" => SR.PropertyCategoryDDE,
+        "Misc" => SR.PropertyCategoryMisc,
+        "Default" => SR.PropertyCategoryMisc,
+        //"Design" => SR.PropertyCategoryDesign,
+        "DragDrop" => SR.CatDragDrop,
+        "Focus" => SR.CatFocus,
+        "Font" => SR.PropertyCategoryFont,
+        //"Format" => SR.PropertyCategoryFormat,
+        "Key" => SR.CatKey,
+        "Layout" => SR.CatLayout,
+        "List" => SR.PropertyCategoryList,
+        "Mouse" => SR.CatMouse,
+        "Position" => SR.PropertyCategoryPosition,
+        "Scale" => SR.PropertyCategoryScale,
+        "Text" => SR.PropertyCategoryText,
+        "WindowStyle" => SR.CatWindowStyle,
+        _ => category
+    };
+
+    public static new SRCategoryAttribute Default
+    {
+        get => s_defaultAttribute ??= new SRCategoryAttribute("PropertyCategoryMisc");
     }
 }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGridInternal/PropertyDescriptorGridEntry.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGridInternal/PropertyDescriptorGridEntry.cs
@@ -154,7 +154,7 @@ internal partial class PropertyDescriptorGridEntry : GridEntry
                 category = base.PropertyCategory;
             }
 
-            return category;
+            return SRCategoryAttribute.GetLocalizedCategory(category);
         }
     }
 
--- a/src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/PropertyGridLocalized.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/PropertyGridLocalized.cs
@@ -1,0 +1,47 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.ComponentModel;
+using System.Globalization;
+
+namespace WinformsControlsTest;
+
+public partial class PropertyGridLocalized : Form
+{
+    // https://github.com/dotnet/runtime/issues/94365
+    public PropertyGridLocalized()
+    {
+        InitializeComponent();
+        propertyGrid1.SelectedObject = new MyObjectClass();
+        cultureComboBox.DisplayMember = "EnglishName";
+        cultureComboBox.ValueMember = "Name";
+        LoadCultures();
+        cultureComboBox.SelectedItem = CultureInfo.CurrentUICulture;
+        cultureComboBox.SelectedIndexChanged += CultureComboBox_SelectedIndexChanged;
+    }
+
+    private void CultureComboBox_SelectedIndexChanged(object sender, EventArgs e)
+    {
+        if (cultureComboBox.SelectedItem is CultureInfo selectedCulture && selectedCulture != CultureInfo.CurrentUICulture)
+        {
+            CultureInfo.CurrentUICulture = selectedCulture;
+            CultureInfo.CurrentCulture = selectedCulture;
+            propertyGrid1.SelectedObject = new MyObjectClass();
+        }
+    }
+
+    private void LoadCultures()
+    {
+        foreach (InputLanguage lang in InputLanguage.InstalledInputLanguages)
+        {
+            cultureComboBox.Items.Add(lang.Culture);
+        }
+    }
+
+    private class MyObjectClass : UserControl
+    {
+        [Browsable(true)]
+        [EditorBrowsable(EditorBrowsableState.Always)]
+        public string MiscValue { get; set; }
+    }
+}
--- a/src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/PropertyGridLocalized.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/PropertyGridLocalized.cs
@@ -8,10 +8,13 @@ namespace WinformsControlsTest;
 
 public partial class PropertyGridLocalized : Form
 {
+    private CultureInfo _originalCulture;
+
     // https://github.com/dotnet/runtime/issues/94365
     public PropertyGridLocalized()
     {
         InitializeComponent();
+        _originalCulture = CultureInfo.CurrentUICulture;
         propertyGrid1.SelectedObject = new MyObjectClass();
         cultureComboBox.DisplayMember = "EnglishName";
         cultureComboBox.ValueMember = "Name";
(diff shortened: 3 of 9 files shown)


elachlan:
CC: @Tanya-Solyanik

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/UserControls/UserControlWithObjectCollectionEditor.cs
```diff
@@ -25,6 +25,10 @@ public IList<int> AAAAAFirstCollection
         get { return new List<int>(new int[] { 1, 2, 3 }); }
         set { }
     }
+
+    [Browsable(true)]
+    [EditorBrowsable(EditorBrowsableState.Always)]
+    public string MiscValue { get; set; }
```
Tanya-Solyanik, line 31: This is a runtime(libraries) issue though, fix on the winforms runtime level will be hacky, i.e. involve string comparison. TypeDescriptor always adds default attribute value when querying for an attribute, and we are always querying for CategoryAttribute, not SRCategoryAttribute... Consider adding a link to the runtime issue.

But if you want to add a test case, it's much easier to have a class with a single public read/write property, rather than a whole control.

elachlan: Are any translations included in runtime? 

It uses `SR.PropertyCategoryDefault`. 
https://github.com/dotnet/runtime/blob/76b5b76ceb08baa587178f8654df1a964547930b/src/libraries/System.ComponentModel.Primitives/src/System/ComponentModel/CategoryAttribute.cs#L211

Tanya-Solyanik: I don't think runtime assemblies are translated at all. They could add hardcoded translated values into the source code for this particular default value because we (users) don't have a way to fix it. ResourceManager in the TypeDescriptor code marks this value "localized" but it's a lie, they supply a hardcoded ENU default string, go with that default string and claim that the value is localized.

Tanya-Solyanik:
@elachlan this test can be improved by adding a dedicated page to the main form, we can remove it after the issue is fixed. This will allow you to set a different UI culture and to define a small custom class with a single property to use as a SelectedObject in the PropertyGrid.

elachlan:
Great Idea! It will take a little longer but would be a great way to test this issue without having to constantly re-log switching languages.

Tanya-Solyanik:
@elachlan  - should we close this PR for now?

elachlan:
Hi @Tanya-Solyanik, I'll put something together today. Sorry I hadn't got it done sooner. Keeping this open is a reminder for me.

elachlan:
@Tanya-Solyanik It kind of works. But property grid is caching the category titles between selected objects I think.

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/PropertyGridLocalized.cs
```diff
@@ -0,0 +1,47 @@
...
+        cultureComboBox.SelectedItem = CultureInfo.CurrentUICulture;
+        cultureComboBox.SelectedIndexChanged += CultureComboBox_SelectedIndexChanged;
+    }
+
+    private void CultureComboBox_SelectedIndexChanged(object sender, EventArgs e)
+    {
+        if (cultureComboBox.SelectedItem is CultureInfo selectedCulture && selectedCulture != CultureInfo.CurrentUICulture)
+        {
+            CultureInfo.CurrentUICulture = selectedCulture;
+            CultureInfo.CurrentCulture = selectedCulture;
```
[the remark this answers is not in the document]
Tanya-Solyanik: Also, does this update category names? I would expect that you must refresh the cached categories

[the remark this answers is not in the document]
elachlan: I do not know how to refresh the cached categories? Any hints?

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/PropertyGridLocalized.cs
```diff
@@ -0,0 +1,47 @@
...
+
+    private void LoadCultures()
+    {
+        foreach (InputLanguage lang in InputLanguage.InstalledInputLanguages)
+        {
+            cultureComboBox.Items.Add(lang.Culture);
+        }
+    }
+
+    private class MyObjectClass : UserControl
```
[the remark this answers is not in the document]
elachlan: I used `UserControl` so I could see that the culture had been applied to other categories.

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/UserControls/UserControlWithObjectCollectionEditor.cs
elachlan: Could we use an interceptor for `TypeDescriptor`?

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/PropertyGridLocalized.cs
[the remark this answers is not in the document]
Tanya-Solyanik: Never mind, these are not cached, the default categories are static readonly properties. So there is no reason to switch languages dynamically as these categories are not updatable.
 We could select the locale on the main page, before propertygrid is created, then the existing proeprty grid test page is good enough

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/UserControls/UserControlWithObjectCollectionEditor.cs
Tanya-Solyanik: I'm not familiar with interceptors, would they work with static properties?

elachlan:
@Tanya-Solyanik I kind of got it working?

image

src/System.Windows.Forms/src/SRCategoryAttribute.cs
```diff
@@ -16,4 +17,37 @@ public SRCategoryAttribute(string category) : base(category)
     {
         return SR.GetResourceString(value);
     }
+
+    public static string GetLocalizedCategory(string category) => category switch
+    {
+        "Accessibility" => SR.CatAccessibility,
```
Tanya-Solyanik, line 23: Does this work? TYpedescriptor is doing reflection against CategoryAttribute type, not the actual derived type...

src/System.Windows.Forms/src/SRCategoryAttribute.cs
```diff
@@ -16,4 +17,37 @@ public SRCategoryAttribute(string category) : base(category)
...
+        "Position" => SR.PropertyCategoryPosition,
+        "Scale" => SR.PropertyCategoryScale,
+        "Text" => SR.PropertyCategoryText,
+        "WindowStyle" => SR.CatWindowStyle,
+        _ => category
+    };
+
+    public static new SRCategoryAttribute Default
+    {
+        get => s_defaultAttribute ??= new SRCategoryAttribute("PropertyCategoryMisc");
```
elachlan, line 51: Having s_defaultAttribute might cause the switching of cultures to not work properly. Say, if you started with japanese and switched to english. I might remove this and just use the default.

src/System.Windows.Forms/src/SRCategoryAttribute.cs
elachlan: Screenshot of the results is in a comment above. I think it looks correct?

src/System.Windows.Forms/src/SRCategoryAttribute.cs
```diff
@@ -16,4 +17,37 @@ public SRCategoryAttribute(string category) : base(category)
...
+        "Action" => SR.CatAction,
+        "Appearance" => SR.CatAppearance,
+        "Asynchronous" => SR.CatAsynchronous,
+        "Behavior" => SR.PropertyCategoryBehavior,
+        //"Config" => SR.PropertyCategoryConfig,
+        "Data" => SR.PropertyCategoryData,
+        "DDE" => SR.PropertyCategoryDDE,
+        "Misc" => SR.PropertyCategoryMisc,
+        "Default" => SR.PropertyCategoryMisc,
+        //"Design" => SR.PropertyCategoryDesign,
```
Tanya-Solyanik, line 33: `Design` is added dynamically at design time

src/System.Windows.Forms/src/SRCategoryAttribute.cs
```diff
@@ -16,4 +16,32 @@ public SRCategoryAttribute(string category) : base(category)
...
+        "DragDrop" => SR.CatDragDrop,
+        "Focus" => SR.CatFocus,
+        "Font" => SR.PropertyCategoryFont,
+        //"Format" => SR.PropertyCategoryFormat,
+        "Key" => SR.CatKey,
```
Tanya-Solyanik, line 37: Why do you need all translations, everything else, besides Misc works correctly? Is this for culture switching only?

src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGridInternal/PropertyDescriptorGridEntry.cs
```diff
@@ -154,7 +154,7 @@ public override string PropertyCategory
                 category = base.PropertyCategory;
             }
 
-            return category;
+            return SRCategoryAttribute.GetLocalizedCategory(category);
```
Tanya-Solyanik, line 157: I see, you are ignoring values defined in the ComponentModel...

Tanya-Solyanik: But you are not guaranteed to get an English string from base.PropertyCategory....

Tanya-Solyanik: Theoretically we need this re-localization for Misc only, but even that could break someone who overrides PropertyDescriptor.Category.

src/System.Windows.Forms/src/SRCategoryAttribute.cs
elachlan: Yes, culture switching will not work without it. If I limit it to Misc:
image

src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGridInternal/PropertyDescriptorGridEntry.cs
elachlan: If we limit it to "Misc" then its basically:

```csharp
return category == "Misc" ? SR.PropertyCategoryMisc : category;
```

elachlan: We can restructure it so the Misc check happens before the null check?
```csharp
public override string PropertyCategory
{
    get
    {
        string? category = PropertyDescriptor.Category;
        if (category is not null and "Misc")
        {
            category = SR.PropertyCategoryMisc;
        }

        if (category is null || category.Length == 0)
        {
            category = base.PropertyCategory;
        }

        return category;
    }
}
```

elachlan: The other option I looked at was implementing a default PropertyDescriptor, but that seemed "heavy".

Tanya-Solyanik: This code snippet breaks someone who implements custom PropertyDescriptor and wants to see category named "Misc". We can say that this is an unlikely scenario, but I would prefer to hardcode all standard category names in ComponentModel. The drawback is language switching, which is a nice to have...
Had switching languages after initialization worked in the .NET Framework?

elachlan: The hackiest solution yet:
```csharp
public override string PropertyCategory
{
    get
    {
        string? category = PropertyDescriptor.Category;
        Type type = PropertyDescriptor.GetType();
        if (type.Name == "ReflectPropertyDescriptor" && category is not null and "Misc")
        {
            category = SR.PropertyCategoryMisc;
        }

        if (category is null || category.Length == 0)
        {
            category = base.PropertyCategory;
        }

        return category;
    }
}
```

Tanya-Solyanik: Yeah, hardcoded translations to all supported languages in ComponentModel is a less hacky approach. Had you tried NET Framework?

elachlan: I have not tested NET Framework. Does `ComponentModel` currently have any of the translations? From memory when I investigated this originally they didn't have any of it setup like `winforms`.

Tanya-Solyanik: I think runtime is not localized, SR properties are resolved from the single resx file. We would have to add localized strings for each culture into the .cs file.

Tanya-Solyanik:
@elachlan  - let's restrict this PR to a test case.

elachlan:
@Tanya-Solyanik Did you just want a form with PropertyGrid and a class with a single misc property? The culture switching won't work without the translation changes.

Tanya-Solyanik:
> @Tanya-Solyanik Did you just want a form with PropertyGrid and a class with a single misc property? The culture switching won't work without the translation changes.

I would like to have some language switching added before your new form is shown, it would be sufficient to simply Set CurrentUICulture and CurrentCulture, run ShowDialog on the new form, and set the culture back to the original one.

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/UserControls/UserControlWithObjectCollectionEditor.cs
KlausLoeffelmann: We don't know yet how interceptors will finally work, or do we? 
You're talking about C#, right?

KlausLoeffelmann:
I don't think that addressing this in does make sense this way.

First of all, not assigning a category leads to assigning "Misc" as a fall-back mechanisam and should not be happening to begin with. I know, it unfortunately happened in a few cases in .NET Framework 2.0, but I am not really sure if that was intended. For example, it would be perfectly fine to assign the `AutoComplete...` properties of the `ComboBox` to the `Behavior` category. And if the `MaskEditBox` has `PasswordChar` and the `TextMaskFormat` under `Behavior`, I think the `ComboBox` `Format...` properties or the `TextBox` `PlaceholderText` should, too. 

I just would dare this to be breaking change and assign the unassigned Categories now.

Also, `PropertyGrid` is just one component, which surfaces the concept of `CategoryAttribute`. So, if, at all, it would be the `PropertyGrid` we would need to change. And it should be the `PropertyGrid` in this case to provide the localisation. But as I said, I would rather have an Analyzer which points out unassigned Categories in a Control or Component.

So, my strongly suggestion is: 
let's address this in another way.

Thanks,

Klaus

Tanya-Solyanik:
@elachlan  - should we close this PR?

elachlan:
I think so. I am really unsure where to take it or what to do.

Tanya-Solyanik:
You could add a private analyzer if interested.

PR #10729, opened 2024-01-24
Enable CA1815 - Override equals and operator equals on value types

### LINKED ISSUE #8284: Enable CA1815 Override equals and operator equals on value types
### Is your feature request related to a problem? Please describe

https://learn.microsoft.com/en-au/dotnet/fundamentals/code-analysis/quality-rules/ca1815
> For [non-blittable](https://learn.microsoft.com/en-au/dotnet/framework/interop/blittable-and-non-blittable-types) value types, the inherited implementation of [Equals](https://learn.microsoft.com/en-us/dotnet/api/system.object.equals) uses the [System.Reflection](https://learn.microsoft.com/en-us/dotnet/api/system.reflection) library to compare the contents of all fields. Reflection is computationally expensive, and comparing every field for equality might be unnecessary. If you expect users to compare or sort instances, or use them as hash table keys, your value type should implement [Equals](https://learn.microsoft.com/en-us/dotnet/api/system.object.equals).

`ImeModeConversion`
https://github.com/dotnet/winforms/blob/5a28fa9c294581314d5675188316fbf09882c114/src/System.Windows.Forms/src/System/Windows/Forms/Control.Ime.cs#L1316

`ThemedScrollbarWindow`
https://github.com/dotnet/winforms/blob/5a28fa9c294581314d5675188316fbf09882c114/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ThemedScrollbarWindow.cs#L1-L12

### Describe the solution you'd like and alternatives you've considered

Add Equals/GetHashCode to the structs.

This involves an API change to the public types.

Unknown

elachlan:
Fixes #8284

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10729)

### DIFF
--- a/src/System.Drawing.Common/src/GlobalSuppressions.cs
+++ b/src/System.Drawing.Common/src/GlobalSuppressions.cs
@@ -1,6 +1,5 @@
-﻿// This file is used by Code Analysis to maintain SuppressMessage
-// attributes that are applied to this project.
-// Project-level suppressions either have no target or are given
-// a specific target and scoped to a namespace, type, member, etc.
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
 
-[assembly: SuppressMessage("Performance", "CA1815:Override equals and operator equals on value types", Justification = "<Pending>", Scope = "type", Target = "~T:System.Drawing.Interop.LOGFONT")]
+// This file is used by Code Analysis to maintain SuppressMessage
+[assembly: SuppressMessage("Performance", "CA1815:Override equals and operator equals on value types", Justification = "Public API", Scope = "type", Target = "~T:System.Drawing.Interop.LOGFONT")]
--- a/src/System.Drawing.Common/src/GlobalSuppressions.cs
+++ b/src/System.Drawing.Common/src/GlobalSuppressions.cs
@@ -1,0 +1,6 @@
+﻿// This file is used by Code Analysis to maintain SuppressMessage
+// attributes that are applied to this project.
+// Project-level suppressions either have no target or are given
+// a specific target and scoped to a namespace, type, member, etc.
+
+[assembly: SuppressMessage("Performance", "CA1815:Override equals and operator equals on value types", Justification = "<Pending>", Scope = "type", Target = "~T:System.Drawing.Interop.LOGFONT")]
--- a/src/System.Windows.Forms.Primitives/src/GlobalSuppressions.cs
+++ b/src/System.Windows.Forms.Primitives/src/GlobalSuppressions.cs
@@ -1,6 +1,5 @@
-﻿// This file is used by Code Analysis to maintain SuppressMessage
-// attributes that are applied to this project.
-// Project-level suppressions either have no target or are given
-// a specific target and scoped to a namespace, type, member, etc.
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
 
-[assembly: SuppressMessage("Performance", "CA1815:Override equals and operator equals on value types", Justification = "<Pending>", Scope = "type", Target = "~T:System.Windows.Forms.VisualStyles.TextMetrics")]
+// This file is used by Code Analysis to maintain SuppressMessage
+[assembly: SuppressMessage("Performance", "CA1815:Override equals and operator equals on value types", Justification = "Public API", Scope = "type", Target = "~T:System.Windows.Forms.VisualStyles.TextMetrics")]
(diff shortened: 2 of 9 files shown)


elachlan:
CC: @JeremyKuhne

JeremyKuhne (dismissed):

elachlan:
I've added more suppressions. Unsure if we want to implement the equals override or not. Since they are public types.

lonitra:
Looks like there are a few more that need to be suppressed

elachlan:
I am unsure on the value of the analyzer if we are suppressing it. Should I just close this and we leave it alone?

src/System.Private.Windows.Core/src/Windows/Win32/UI/Controls/Dialogs/DEVNAMES.cs
```diff
@@ -21,7 +21,9 @@ namespace Windows.Win32.UI.Controls.Dialogs;
 ///
 ///  https://github.com/microsoft/CsWin32/issues/882
 /// </devdoc>
+#pragma warning disable CA1815 // Override equals and operator equals on value types
```
JeremyKuhne, line 24: The idea is to add these to a global suppressions file for two primary reasons:

1. We already shipped some public API that doesn't conform
2. We're extending something that CsWin32 generates

Generally speaking, I don't want the `#pragma` suppressions in the code as they inhibit readability. The biggest caveat is we should continue to use `#pragma` for anything that is related to using depreciated APIs (such as `BinaryFormatter` 
 related stuff).

I think this one is moot though. The type was accidentally public, I don't know that it will get flagged for this once you rebase.

elachlan:
@lonitra I think there is an unrelated test failure here.

src/System.Drawing.Common/src/GlobalSuppressions.cs
```diff
@@ -0,0 +1,6 @@
+﻿// This file is used by Code Analysis to maintain SuppressMessage
```
lonitra, line 1: Please add licensing for these new files
```suggestion
// Licensed to the .NET Foundation under one or more agreements.
// The .NET Foundation licenses this file to you under the MIT license.

﻿// This file is used by Code Analysis to maintain SuppressMessage
```

src/System.Drawing.Common/src/GlobalSuppressions.cs
```diff
@@ -0,0 +1,6 @@
...
+// attributes that are applied to this project.
+// Project-level suppressions either have no target or are given
+// a specific target and scoped to a namespace, type, member, etc.
+
+[assembly: SuppressMessage("Performance", "CA1815:Override equals and operator equals on value types", Justification = "<Pending>", Scope = "type", Target = "~T:System.Drawing.Interop.LOGFONT")]
```
lonitra, line 6: I believe `LOGFONT` is a public API so I think we can just write "Public API" as the justification.

src/System.Windows.Forms.Primitives/src/GlobalSuppressions.cs
```diff
@@ -0,0 +1,6 @@
+﻿// This file is used by Code Analysis to maintain SuppressMessage
+// attributes that are applied to this project.
+// Project-level suppressions either have no target or are given
+// a specific target and scoped to a namespace, type, member, etc.
+
+[assembly: SuppressMessage("Performance", "CA1815:Override equals and operator equals on value types", Justification = "<Pending>", Scope = "type", Target = "~T:System.Windows.Forms.VisualStyles.TextMetrics")]
```
lonitra, line 6: Ditto here with public API justification

src/System.Windows.Forms/src/GlobalSuppressions.cs
```diff
@@ -30,3 +31,5 @@
 [assembly: SuppressMessage("Style", "IDE1006:Naming Styles", Justification = "Thread local", Scope = "member", Target = "~F:System.Windows.Forms.VisualStyles.VisualStyleRenderer.t_themeHandles")]
 [assembly: SuppressMessage("Style", "IDE1006:Naming Styles", Justification = "Thread local", Scope = "member", Target = "~F:System.Windows.Forms.VisualStyles.VisualStyleRenderer.t_threadCacheVersion")]
 [assembly: SuppressMessage("Style", "IDE1006:Naming Styles", Justification = "Thread local", Scope = "member", Target = "~F:System.Windows.Forms.RadioButtonRenderer.t_visualStyleRenderer")]
+
+[assembly: SuppressMessage("Performance", "CA1815:Override equals and operator equals on value types", Justification = "Public API", Scope = "type", Target = "~T:System.Windows.Forms.ImeModeConversion")]
```
lonitra, line 35: Consider moving this up by the other suppressions for public APIs

lonitra (approved):
LGTM!

PR #10730, opened 2024-01-24
cleaned up implementations of RuntimeId in AccessibleObjects

Tanya-Solyanik:
1  fixed style
2. replaced PARAM.ToInt with a cast (int) because InpPtr is now implemented as a nint
3. fixed xml-doc comments

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10730)


lonitra (approved):
👍

PR #10732, opened 2024-01-24
Use CurrentUICulture in KeysConverter.cs when converting to name in d…

### LINKED ISSUE #10687: KeysConverter should use CurrentUICulture when converting to name in default context.
KeysConverter in NET8 started converting keys to strings based on the CurrentCulture - https://github.com/dotnet/winforms/blob/f7d6850caa53f62816d977e737d6cb2a537f0d9a/src/System.Windows.Forms/src/System/Windows/Forms/Input/KeysConverter.cs#L332-L354

However, in .NET Framework, we were looking up names based on the CurrentUICulture - https://referencesource.microsoft.com/#System.Windows.Forms/winforms/Managed/System/WinForms/KeysConverter.cs,64 from the resources like this:

AddKey(SR.**GetString**(SR.toStringPageDown), Keys.Next);

this code goes to [ResourceManager.GetString Method (System.Resources) | Microsoft Learn](https://learn.microsoft.com/dotnet/api/system.resources.resourcemanager.getstring?view=netframework-4.8.1#system-resources-resourcemanager-getstring(system-string))
and returns

> The value of the resource localized for the caller's current UI culture, or null if name cannot be found in a resource set.

_Originally posted by @cadilhac in https://github.com/dotnet/winforms/issues/10610#issuecomment-1897509534_

LeafShi1:
Fixes #10687 

## Proposed changes

- Use CurrentUICulture instead of CurrentCulture in function ConvertTo and ConvertFrom 

## Customer Impact

- Key values and order will be displayed according to CurrentUICulture

## Regression? 

- Yes 

## Risk

- Minimal

### Before

- Key values and order are displayed according to CurrentCulture

### After

 - Key values and order are displayed according to CurrentUICulture

## Test methodology 

- Unit test

## Test environment(s) 

- .net 9.0.0-alpha.1.24066.33

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10732)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Input/KeysConverter.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Input/KeysConverter.cs
@@ -331,7 +331,8 @@ public class KeysConverter : TypeConverter, IComparer
 
     private IList<string> GetDisplayOrder(CultureInfo? culture)
     {
-        // Use CurrentUICulture as default to match other TypeConverters.
+        // Use CurrentUICulture as default because we assume that this converter is primarily
+        // used in user-facing applications (I.e. what key to press on the keyboard).
         culture ??= CultureInfo.CurrentUICulture;
         if (!CultureToDisplayOrder.ContainsKey(culture))
         {
@@ -343,7 +344,8 @@ public class KeysConverter : TypeConverter, IComparer
 
     private IDictionary<string, Keys> GetKeyNames(CultureInfo? culture)
     {
-        // Use CurrentUICulture as default to match other TypeConverters.
+        // Use CurrentUICulture as default because we assume that this converter is primarily
+        // used in user-facing applications (I.e. what key to press on the keyboard).
         culture ??= CultureInfo.CurrentUICulture;
         if (!CultureToKeyName.ContainsKey(culture))
         {
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Input/KeysConverter.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Input/KeysConverter.cs
@@ -331,8 +331,8 @@ public class KeysConverter : TypeConverter, IComparer
 
     private IList<string> GetDisplayOrder(CultureInfo? culture)
     {
-        // Use CurrentCulture as default to match other TypeConverters.
-        culture ??= CultureInfo.CurrentCulture;
+        // Use CurrentUICulture as default to match other TypeConverters.
+        culture ??= CultureInfo.CurrentUICulture;
         if (!CultureToDisplayOrder.ContainsKey(culture))
         {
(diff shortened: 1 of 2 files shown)


LeafShi1:
The unit tests  and manual testing are in progress

Tanya-Solyanik:
>The unit tests and manual testing are in progress

Additionally please compare the new behavior to the .NET Framework, the goal is to have the same results as in the .NET Framework.

src/System.Windows.Forms/src/System/Windows/Forms/Input/KeysConverter.cs
```diff
@@ -331,8 +331,8 @@ string GetTermsString(Keys key)
 
     private IList<string> GetDisplayOrder(CultureInfo? culture)
     {
-        // Use CurrentCulture as default to match other TypeConverters.
-        culture ??= CultureInfo.CurrentCulture;
+        // Use CurrentUICulture as default to match other TypeConverters.
```
Tanya-Solyanik, line 334:
```suggestion
        // Use CurrentUICulture as default because we assume that this converter is primarily used in user-facing applications (I.e. what key to press on the keyboard).
```

src/System.Windows.Forms/src/System/Windows/Forms/Input/KeysConverter.cs
```diff
@@ -343,8 +343,8 @@ private IList<string> GetDisplayOrder(CultureInfo? culture)
 
     private IDictionary<string, Keys> GetKeyNames(CultureInfo? culture)
     {
-        // Use CurrentCulture as default to match other TypeConverters.
-        culture ??= CultureInfo.CurrentCulture;
+        // Use CurrentUICulture as default to match other TypeConverters.
```
Tanya-Solyanik, line 346:
```suggestion
        // Use CurrentUICulture as default because we assume that this converter is primarily used in user-facing applications (I.e. what key to press on the keyboard).
```

LeafShi1:
> > The unit tests and manual testing are in progress
> 
> Additionally please compare the new behavior to the .NET Framework, the goal is to have the same results as in the .NET Framework.

Culture: Chinese, UICulture: French
image

The order of the key is always the same no matter which UICulture or Culture is set, and the results of netframework and net projects are consistent.

SimonZhao888 (approved):
Looks good!

Tanya-Solyanik:
@LeafShi1  - thank you for additional verification. Interesting to know that we have different translations for `Shift` key, but this is fine.

Tanya-Solyanik (approved):
Looks good, please prepare a PR for release/8.0 branch and send this fix to @Olina-Zhang's team to test

PR #10736, opened 2024-01-24
Clean up ImageListStreamer

JeremyKuhne:
Clean up the code in ImageListStreamer. It should be more efficient now as it uses vector optimized reading and writing from the .NET libraries (through the SpanReader/Writer) .

- Factor out RLE encoding into RunLengthEncoder
- Add a SpanWriter to go with SpanReader
- Tweak SpanReader to have Position property and add IsNext, TryAdvancePast, and AdvancePast
- Add inherit doc to ImageList P/Invoke overloads

Also add a test to catch public types in System.Private.Windows.Core and fix the prior misses.

### DIFF
--- a/src/System.Private.Windows.Core/src/Windows/Win32/System/Com/Interop.GPStream.cs
+++ b/src/System.Private.Windows.Core/src/Windows/Win32/System/Com/Interop.GPStream.cs
@@ -5,7 +5,7 @@ using System.Runtime.InteropServices;
 
 namespace Windows.Win32.System.Com;
 
-public sealed unsafe class ComManagedStream : IStream.Interface, IManagedWrapper<IStream, ISequentialStream>
+internal sealed unsafe class ComManagedStream : IStream.Interface, IManagedWrapper<IStream, ISequentialStream>
 {
     private readonly Stream _dataStream;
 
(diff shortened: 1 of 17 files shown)


src/System.Private.Windows.Core/src/Windows/Win32/System/Com/Interop.GPStream.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace Windows.Win32.System.Com;
 
-public sealed unsafe class ComManagedStream : IStream.Interface, IManagedWrapper<IStream, ISequentialStream>
```
Tanya-Solyanik, line 8: Dumb question - why didn't this API show up in the public APIs text file?

lonitra (approved):
👍

src/System.Private.Windows.Core/src/Windows/Win32/System/Com/Interop.GPStream.cs
Tanya-Solyanik: Because we don't have this file for this assembly...

lonitra: I'm not 100% how that analyzer that flags that we don't have public APIs in the PublicAPI.Unshipped.txt works, but we don't have those txt files for System.Private.Windows.Core since this assembly should never have any public APIs in it.

Issue #8835, opened 2020-07-23
Add Effects to System.Drawing


reflectronic:
### Background

A major addition in GDI+1.1 was the addition of effects. GDI+ comes with eleven effect types which perform various adjustments and transformations on images, like blurs, color balancing, brightness/contrast adjustments, etc. This proposal is one of many to add missing GDI+ 1.1 functionality to `System.Drawing`.

### API Proposal

Updated proposal: https://github.com/dotnet/winforms/issues/8835#issuecomment-1912901179

See the documentation for [the gdipluseffects.h header](https://docs.microsoft.com/en-us/windows/win32/api/gdipluseffects/) and the corresponding [GDI+ flat APIs](https://docs.microsoft.com/en-us/windows/win32/gdiplus/-gdiplus-bitmap-flat).

```diff
+namespace System.Drawing.Effects
+{
+   public abstract class Effect : IDisposable
+   {
+       public void Dispose();
+
+       public bool UseAuxiliaryData { get; set; }
+
!       I know this is a bit of a mouthful - see below for details.
+       public ColorLookupTable GetColorLookupTableFromAuxiliaryData();
+   }

!   These APIs deviate slightly from the native definitions. See below.
+   public sealed class BlurEffect : Effect
+   {
+       public BlurEffect();
+
+       public float Radius { get; set; }
+       public bool ExpandEdge { get; set; }
+   }

+   public sealed class BrightnessContrastEffect : Effect
+   {
+       public BrightnessContrastEffect();
+
+       public int BrightnessLevel { get; set; }
+       public bool ContrastLevel { get; set; }
+   }

+   public sealed class ColorBalanceEffect : Effect
+   {
+       public ColorBalanceEffect();

+       public int CyanRed { get; set; }
+       public int MagentaGreen { get; set; }
+       public int YellowBlue { get; set; }
+   }

+   public sealed class ColorCurveEffect : Effect
+   {
+       public ColorCurveEffect();

+       public CurveAdjustments Adjustment { get; set; }
+       public CurveChannel Channel { get; set; }
+       public int AdjustmentValue { get; set; }
+   }

+   public enum CurveAdjustments
+   {
+       Exposure,
+       Density,
+       Contrast,
+       Highlight,
+       Shadow,
+       Midtone,
+       WhiteSaturation,
+       BlackSaturation
+   }

+   public enum CurveChannel
+   {
+       All,
+       Red,
+       Green,
+       Blue
+   }

[...]

Issue #4338, opened 2020-12-08
API Proposal: Add IDeviceContext overloads to ControlPaint


JeremyKuhne:
### Summary

Painting using `System.Drawing` (GDI+) has significant performance overhead over utilizing GDI. Internally WinForms has moved much of it's rendering to GDI (starting in .NET 2.0) and has provided some entry points with GDI via `IDeviceContext`. Adding overloads to `ControlPaint` methods will allow more performant rendering of custom controls. It allows direct use of paint event arguments in `OnPaint`  as they now expose `IDeviceContext` (as opposed to `e.Graphics`, which has huge overhead).

### API

``` C#
    public sealed class ControlPaint
    {
        // Existing APIs that take Graphics
        public static void DrawBorder(Graphics graphics, Rectangle bounds, Color color, ButtonBorderStyle style);
        public static void DrawBorder(Graphics graphics, Rectangle bounds, Color leftColor, int leftWidth, ButtonBorderStyle leftStyle, Color topColor, int topWidth, ButtonBorderStyle topStyle, Color rightColor, int rightWidth, ButtonBorderStyle rightStyle, Color bottomColor, int bottomWidth, ButtonBorderStyle bottomStyle);
        public static void DrawBorder3D(Graphics graphics, int x, int y, int width, int height);
        public static void DrawBorder3D(Graphics graphics, int x, int y, int width, int height, Border3DStyle style);
        public static void DrawBorder3D(Graphics graphics, int x, int y, int width, int height, Border3DStyle style, Border3DSide sides);
        public static void DrawBorder3D(Graphics graphics, Rectangle rectangle);
        public static void DrawBorder3D(Graphics graphics, Rectangle rectangle, Border3DStyle style);
        public static void DrawBorder3D(Graphics graphics, Rectangle rectangle, Border3DStyle style, Border3DSide sides);
        public static void DrawButton(Graphics graphics, int x, int y, int width, int height, ButtonState state);
        public static void DrawButton(Graphics graphics, Rectangle rectangle, ButtonState state);
        public static void DrawCaptionButton(Graphics graphics, int x, int y, int width, int height, CaptionButton button, ButtonState state);
        public static void DrawCaptionButton(Graphics graphics, Rectangle rectangle, CaptionButton button, ButtonState state);
        public static void DrawCheckBox(Graphics graphics, int x, int y, int width, int height, ButtonState state);
        public static void DrawCheckBox(Graphics graphics, Rectangle rectangle, ButtonState state);
[...]

PR #10750, opened 2024-01-27
Initial implementation of GDI+ effects

JeremyKuhne:
This exposes most of the GDI+ effects under the [RequiresPreviewFeatures] while we take it to .NET API review.

Effects are added under the `Imaging\Effects` namespace. The new effects are:

- BlackSaturation
- Blur
- BrightnessContrast
- ColorBalance
- ColorLookupTable
- ColorMatrix
- Contrast
- Density
- Exposure
- Highlight
- Levels
- Midtone
- Shadow
- Sharpen
- Tint
- WhiteSaturation

This also adds:

- New, efficient constructor to ColorMatrix that takes a span of floats
- Overload for Graphics.DrawImage that allows you to draw with an effect (and fix the perf of the indexer)
- An ApplyEffect method on Bitmap that lets you apply an effect to an existing bitmap
- Adds the uses preview features flag to the scratch projects

Effects are applied to bitmaps, not associated with them. The image doesn't retain a handle, it just modifies the bits using the effect. As such I have not exposed a way to mutate the effect classes after creation. The cost of creating an effects class is not high, but they should be disposed to avoid having them clutter the finalizer queue.

I also did not include the red eye removal effect as it is sort of a one-off from general image effects. GDI+ won't be adding any new effects (these were added over 20 years ago).

Tests are pending.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10750)


JeremyKuhne:
Related to #4338

KlausLoeffelmann (commented):
Didn't look at each individual line, but it looks very reasonable to me.

One question though:
Don't we need a) modified public-API text files, and b) a list of concrete API changes in the PR/Issue description for the API review?

JeremyKuhne:
@KlausLoeffelmann #8835 has the API proposal and we don't have api.txt files for System.Drawing. We build for multiple frameworks so I'm not even sure it is possible; I'll look into it when I get a chance.

src/System.Drawing.Common/src/NativeMethods.txt
```diff
@@ -546,4 +546,28 @@ SHDefExtractIcon
 SHGSI_FLAGS
 SHSTOCKICONID
 StartDoc
-StartPage
\ No newline at end of file
+StartPage
+GdipSetEffectParameters
```
lonitra, line 550: These are not in alphabetical order

src/System.Windows.Forms/tests/IntegrationTests/ScratchProject/ScratchProject.csproj
```diff
@@ -22,4 +23,10 @@
     <ProjectReference Include="..\..\..\src\System.Windows.Forms.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <None Update="Resources\bill-gates.jpg">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+  </ItemGroup>
```
lonitra, line 30: Was this meant to be added?

src/System.Drawing.Common/src/System/Drawing/Imaging/ColorMatrix.cs
```diff
@@ -334,58 +329,60 @@ internal void SetMatrix(float[][] newColorMatrix)
         _matrix44 = newColorMatrix[4][4];
     }
 
-    internal float[][] GetMatrix()
```
lonitra, line 337: Do we still need to retain this code in case `NET9_0_OR_GREATER` is false?

lonitra (commented):
Few minor comments but overall looks good! This is exciting 🥳

src/System.Drawing.Common/src/System/Drawing/Imaging/ColorMatrix.cs
JeremyKuhne: No- it wouldn't compile against .NET 8 without it.

lonitra (approved):
:shipit:

PR #10759, opened 2024-01-29
Always Wrap Incoming Clipboard Data

### LINKED ISSUE #4555: Broken custom IDataObject implementations
* .NET Core Version: 5.0

* Have you experienced this same bug with .NET Framework?:  No

**Problem description:**

PR #3388 removed the `ComVisible` attribute from `System.Windows.Forms.IDataObject`. That largely breaks the marshaling custom implementation of `IDataObject` through clipboard *within the same application*. The concern about marshaling of individual calls is not applicable in that case because the runtime could resolve the `IDataObject` to the original .NET object implementation.

It likely broke this case in `Clipboard.GetDataObject` and now all returned objects are wrapped in the default `DataObject` which doesn't support all the custom types:
https://github.com/dotnet/winforms/blob/2fd2a5840b345ef83e0d15a91747325af27441f1/src/System.Windows.Forms/src/System/Windows/Forms/Clipboard.cs#L170-L173

**Expected behavior:**

**Minimal repro:**

TBD

lonitra:
Fixes #4555 

**Background:**
WinForms original clipboard behavior is that:
- The original data coming in must implement `System.Runtime.InteropServices.ComTypes.IDataObject` to be taken as is. If it does not, we will wrap the data in a `DataObject`.
- The original data going out must implement `System.Windows.Forms.IDataObject` to be passed out as is. If it does not, we will wrap the data in a `DataObject` before passing it out

Native call to `OleGetClipboard` will always return a proxy that will forward all IDataObject method calls to the real data object, without giving out the real data object. The only time that the proxy ever gives back the real object is when we query for an interface it does not know about. WinForms `DataObject` class also implements `IComCallableWrapper` which is an internal interface meant to help indicate to us that this object is one of our CCWs. The proxy does not know about `IComCallableWrapper` meaning that when we query for it on the proxy, the proxy will give us back the real data object. This is how we have been able to give the user back the original object they placed on the clipboard (if the original object meets the 2 requirements outlined above) when they ask for it.

**Issue:**
When users create a custom data object that implements both `System.Runtime.InteropServices.ComTypes.IDataObject` and `System.Windows.Forms.IDataObject`, the expectation is that the user will be able to set their custom data object onto the clipboard and get the data object back out as is i.e. not wrapped. 
This is not the case with today's code since:
1. The custom data object coming in will not be wrapped due to it implementing 
    `System.Runtime.InteropServices.ComTypes.IDataObject`
2. `OleGetClipboard` will continue to return a proxy. Since the user's custom data object does not implement an interface that WinForms knows about and the proxy does not, e.g. `IComCallableWrapper`, WinForms cannot retrieve the real data object from the proxy and will just return the proxy wrapped in a `DataObject`

**Proposed Solution:**
WinForms new clipboard behavior:
- The original data coming will always be wrapped in a `DataObject`.
- The original data must implement `System.Windows.Forms.IDataObject` to be passed out as is. If it does not, we will leave the data wrapped in a `DataObject` when passing it out

How it works:
1. Always wrap the incoming data to the clipboard in a `DataObject` and flag that we've wrapped it for clipboard purposes using new property `IsWrappedForClipboard`
2. When passing the data out from the clipboard, try to query for `IComCallableWrapper` per usual. If we were able to query for `IComCallableWrapper` successfully, this means that we a pointer to the data object we placed onto the clipboard, not the proxy
3. Get the managed object from the pointer received from querying for `IComCallableWrapper` successfully and check if the managed object is a `DataObject` that has `IsWrappedForClipboard = true`. If so, we know we have the `DataObject` that we created in step 1 that wraps the real data object. We can now utilize new internal method `DataObject.UnwrapInnerIDataObject` to retrieve the real data object we wrapped if the real data object had implemented `System.Windows.Forms.IDataObject` and return it. 
Implementation for this new method is the opposite of the wrapping behavior we have in place for `DataObject` construction where we only track the real data object as is via `_innerData` if it implements `System.Windows.Forms.IDataObject`:
https://github.com/dotnet/winforms/blob/47131cd723c596d567cd2d80480f6f64272f41aa/src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.cs#L104-L120
Note: We must only unwrap the real data object from the `DataObject` if the real data object implements `System.Windows.Forms.IDataObject` to maintain our original heuristic. If the real data object does not implement `System.Windows.Forms.IDataObject`, we return `DataObject`, which is still inline with our original heuristic as the data is wrapped when passing it out.
4. If we were not able to successfully query for `IComCallableWrapper` in step 2 or the `DataObject` on the clipboard has `IsWrappedForClipboard = false`, we did not place that data on the clipboard so we should not unwrap it. Fallback to old behavior.

Other thoughts:
- Looking through windows sources, native DragDrop operation does not have this same odd behavior of returning a proxy so we do not need do something similar for that scenario
- Getting data via WinForms clipboard and setting clipboard data by other means should still work as expected since we fall back to old behavior if WinForms did not set the data on the clipboard
- Setting the data via WinForms clipboard and getting the clipboard data by other means should also still work as expected since native call `OleGetClipboard` will continue to return a proxy that forwards `IDataObject` method calls to the underlying data, which would now be our `DataObject` and our `DataObject` forwards calls to the real data object we wrap.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10759)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/OLE/Clipboard.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/OLE/Clipboard.cs
@@ -129,9 +129,7 @@ public static class Clipboard
             // There is a DataObject on the clipboard that we placed there. If the real data object
             // implements IDataObject, we want to unwrap it and return it. Otherwise return
             // the DataObject as is.
-            return wrappedData.UnwrapInnerIDataObject() is { } innerData
-                ? innerData
-                : wrappedData;
+            return wrappedData.TryUnwrapInnerIDataObject();
         }
 
         // We did not place the data on the clipboard. Fall back to old behavior.
--- a/src/System.Windows.Forms/src/System/Windows/Forms/OLE/Clipboard.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/OLE/Clipboard.cs
@@ -41,7 +41,11 @@ public static class Clipboard
         ArgumentOutOfRangeException.ThrowIfNegative(retryTimes);
         ArgumentOutOfRangeException.ThrowIfNegative(retryDelay);
 
-        using var dataObject = ComHelpers.GetComScope<Com.IDataObject>(data is IComDataObject ? data : new DataObject(data));
+        // Always wrap the data in our DataObject since we know how to retrieve our DataObject from the proxy OleGetClipboard returns.
+        DataObject wrappedData = data is DataObject { IsWrappedForClipboard: true } alreadyWrapped
+            ? alreadyWrapped
+            : new DataObject(data) { IsWrappedForClipboard = true };
+        using var dataObject = ComHelpers.GetComScope<Com.IDataObject>(wrappedData);
 
         HRESULT hr;
         int retry = retryTimes;
@@ -120,6 +124,17 @@ public static class Clipboard
             return null;
         }
 
+        if (dataObject is DataObject { IsWrappedForClipboard: true } wrappedData)
+        {
+            // There is a DataObject on the clipboard that we placed there. If the real data object
+            // implements IDataObject, we want to unwrap it and return it. Otherwise return
+            // the DataObject as is.
+            return wrappedData.UnwrapInnerIDataObject() is { } innerData
+                ? innerData
+                : wrappedData;
+        }
+
+        // We did not place the data on the clipboard. Fall back to old behavior.
         return dataObject is IDataObject ido && !Marshal.IsComObject(dataObject)
                 ? ido
                 : new DataObject(dataObject);
--- a/src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.cs
@@ -130,7 +130,7 @@ public unsafe partial class DataObject :
     /// <summary>
     ///  Flags that the original data was wrapped for clipboard purposes.
     /// </summary>
-    internal bool IsWrappedForClipboard { get; set; }
+    internal bool IsWrappedForClipboard { get; init; }
 
     /// <summary>
     ///  Returns the inner data that the <see cref="DataObject"/> was created with if the original data implemented
--- a/src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.cs
@@ -134,18 +134,15 @@ public unsafe partial class DataObject :
 
     /// <summary>
     ///  Returns the inner data that the <see cref="DataObject"/> was created with if the original data implemented
-    ///  <see cref="IDataObject"/>. Otherwise, returns null.
+    ///  <see cref="IDataObject"/>. Otherwise, returns this.
     ///  This method should only be used if the <see cref="DataObject"/> was created for clipboard purposes.
     /// </summary>
-    internal IDataObject? UnwrapInnerIDataObject()
+    internal IDataObject TryUnwrapInnerIDataObject()
     {
-        if (!IsWrappedForClipboard)
-        {
-            throw new InvalidOperationException("This method should only be used for clipboard purposes.");
-        }
+        Debug.Assert(IsWrappedForClipboard, "This method should only be used for clipboard purposes.");
 
         return _innerData is DataStore or ComDataObjectAdapter
-            ? null
+            ? this
             : _innerData;
     }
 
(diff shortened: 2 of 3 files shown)


lonitra:
cc: @filipnavara , @weltkante , @AaronRobinsonMSFT , @jacdavis for comments, suggestions, and/or other scenarios to consider that I might've missed

src/System.Windows.Forms/src/System/Windows/Forms/OLE/Clipboard.cs
```diff
@@ -120,6 +124,17 @@ public static unsafe void SetDataObject(object data, bool copy, int retryTimes,
             return null;
         }
 
+        if (dataObject is DataObject { IsWrappedForClipboard: true } wrappedData)
+        {
+            // There is a DataObject on the clipboard that we placed there. If the real data object
+            // implements IDataObject, we want to unwrap it and return it. Otherwise return
```
AaronRobinsonMSFT, line 130: So this says `is { }`, which matches on non-null. I assume that `UnwrapInnerIDataObject()` returns `null` as an indication of `false`, is that the intent here?

src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.cs
```diff
@@ -127,6 +127,28 @@ public DataObject(object data)
 
     internal DataObject(string format, bool autoConvert, object data) : this() => SetData(format, autoConvert, data);
 
+    /// <summary>
+    ///  Flags that the original data was wrapped for clipboard purposes.
+    /// </summary>
+    internal bool IsWrappedForClipboard { get; set; }
```
AaronRobinsonMSFT, line 133:
```suggestion
    internal bool IsWrappedForClipboard { get; init; }
```

I assume this should be an immutable value.

src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.cs
```diff
@@ -127,6 +127,28 @@ public DataObject(object data)
...
+    /// <summary>
+    ///  Returns the inner data that the <see cref="DataObject"/> was created with if the original data implemented
+    ///  <see cref="IDataObject"/>. Otherwise, returns null.
+    ///  This method should only be used if the <see cref="DataObject"/> was created for clipboard purposes.
+    /// </summary>
+    internal IDataObject? UnwrapInnerIDataObject()
+    {
+        if (!IsWrappedForClipboard)
+        {
+            throw new InvalidOperationException("This method should only be used for clipboard purposes.");
```
AaronRobinsonMSFT, line 144: Should this be an assert instead? If not, does the message need to localized?

lonitra: This was an attempt to ensure that this method does not get used outside of this clipboard scenario. It is an internal method and it's usage outside this scenario should be caught before it is anywhere public so I don't think this needs to be localized.

src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.cs
lonitra: Good catch!

weltkante:
This may need consideration and testing for cross-process scenarios to make sure you don't fail with exceptions or otherwise go down the wrong code paths, because you'll likely be able to see "your own" COM interface yet probably not be able to unwrap the object since its in another process? Not sure how the WinForms COM logic works here, I'm not used to the new style COM interop, so maybe this already has been considered previously, since I don't see a lot of changes in that regard in the PR.

src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.cs
AaronRobinsonMSFT: Sounds good. Does that mean it can be converted into an assert?

src/System.Windows.Forms/src/System/Windows/Forms/OLE/Clipboard.cs
AaronRobinsonMSFT: > I suppose I could update this method to just return itself in this case to be more clear as well

That sounds like a great idea. +1

AaronRobinsonMSFT:
> because you'll likely be able to see "your own" COM interface yet probably not be able to unwrap the object since its in another process?

That is quite possible. The out-of-proc aspect of this is likely to make things complicated. It is a very common scenario in Visual Studio. Testing should be simple though.

JeremyKuhne (approved):
Nice!

AaronRobinsonMSFT (approved):

PR #10770, opened 2024-01-31

gpetrou:
## Proposed changes

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10770)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGrid.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGrid.cs
@@ -3050,7 +3050,7 @@ public partial class PropertyGrid : ContainerControl, IComPropertyBrowser, IProp
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGrid.cs
```diff
@@ -3660,13 +3669,15 @@ internal void ReplaceSelectedObject(object oldObject, object newObject)
 
     private void SaveSelectedTabIndex()
     {
-        if (_designerHost is not null)
+        if (_designerHost is not null && _selectedTab is not null)
```
lonitra, line 3672: This may be a change in behavior. Previously this code would set `_designerSelections[_designerHost.GetHashCode()] = -1` if `_selectedTab` is null. In this case we can leave the condition as it originally was and add `!` for `IndexOf` since docs say that the parameter can be null. A comment should be left as well.

lonitra (approved):
Thank you

PR #10793, opened 2024-02-01
Add Support for Bitmap BinaryFormat

lonitra:
Add support for binary formatting `Bitmap` without using `BinaryFormatter`
- Move Bitmap serialization code to Core assembly for centralized use
- Adds tests:
    - Deserialize from a object that was binary formatted via `BinaryFormatter` using our Binary Format reader
    - Deserialize from a object that was binary formatted via our Binary Format writer using `BinaryFormatter`
    - Round trip serialization/deserialization using our Binary Format reader/writer

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10793)


src/System.Private.Windows.Core/src/System/Drawing/ImageExtensions.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        using var iStream = stream.ToIStream();
+        PInvokeCore.GdipSaveImageToStream(image.Pointer, iStream, &encoder, encoderParameters).ThrowIfFailed();
+    }
+
+    internal static void Save(this IImage image, MemoryStream stream)
+    {
+        Guid format = default;
+        PInvokeCore.GdipGetImageRawFormat(image.Pointer, &format).ThrowIfFailed();
+        GC.KeepAlive(image);
```
JeremyKuhne, line 33: The `KeepAlive` isn't necessary as we reference it at the end of the method.

src/System.Windows.Forms/src/System/Windows/Forms/BinaryFormat/WinFormsBinaryFormatWriter.cs
```diff
@@ -12,6 +14,23 @@ internal static class WinFormsBinaryFormatWriter
 
     private static readonly string s_currentWinFormsFullName = typeof(WinFormsBinaryFormatWriter).Assembly.FullName!;
 
+    public static unsafe void WriteBitmap(Stream stream, Bitmap bitmap)
+    {
+        using MemoryStream memoryStream = new();
+        bitmap.Save(memoryStream);
+        byte[] data = memoryStream.GetBuffer();
```
JeremyKuhne, line 21: You need to slice this into an ArraySegment<byte> based on the MemoryStream length.

src/System.Windows.Forms/src/System/Windows/Forms/BinaryFormat/WinFormsBinaryFormatWriter.cs
```diff
@@ -12,6 +14,24 @@ internal static class WinFormsBinaryFormatWriter
 
     private static readonly string s_currentWinFormsFullName = typeof(WinFormsBinaryFormatWriter).Assembly.FullName!;
 
+    public static unsafe void WriteBitmap(Stream stream, Bitmap bitmap)
+    {
+        using MemoryStream memoryStream = new();
+        bitmap.Save(memoryStream);
+
+        Debug.Assert(memoryStream.TryGetBuffer(out ArraySegment<byte> data));
```
JeremyKuhne, line 22: You can't do this inside an Assert as it will not be in release builds

JeremyKuhne (approved):
Looks good!

PR #10801, opened 2024-02-02
Enable nullability in remaining PropertyGrid members

gpetrou:
## Proposed changes

- Enable nullability in remaining PropertyGrid members.


src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGrid.cs
```diff
@@ -4147,9 +4152,9 @@ internal void UpdateSelection()
...
         }
 
-        string tabName = $"{_selectedTab.Tab.TabName}{_propertySortValue}";
+        string tabName = $"{_selectedTab?.Tab.TabName}{_propertySortValue}";
 
-        if (_viewTabProperties is not null && _viewTabProperties.TryGetValue(tabName, out GridEntry value))
+        if (_viewTabProperties is not null && _viewTabProperties.TryGetValue(tabName, out GridEntry? value))
         {
             _rootEntry = value;
             _rootEntry?.Refresh();
```
lonitra, line 4160: Perhaps I missed something, but I don't think we ever place null as a value in `_viewTabProperties` so this `?` is not needed here.

src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGrid.cs
```diff
@@ -4188,8 +4193,8 @@ internal void UpdateSelection()
         }
 
         // Get entries.
-        _currentEntries = _rootEntry.Children;
-        _defaultEntry = _rootEntry.DefaultChild;
+        _currentEntries = _rootEntry?.Children;
+        _defaultEntry = _rootEntry?.DefaultChild;
```
lonitra, line 4197: Related here -- if we do not place `null` as a value in `_viewTabProperties`, I don't think we will have a null `_rootEntry` at this point.

PR #10808, opened 2024-02-02
Split out effects and unseal classes

JeremyKuhne:
I realized part of the fun is writing your own sets of effect settings so I've unsealed all of the classes and broken out the statics for Grayscale, Sepia, Invert, and Vivid. Improves discoverability and allows users to build their own "suite" of effects. (Rather than second guess which ones might be interesting I've unsealed them all.)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10808)

### DIFF
--- a/src/System.Drawing.Common/src/System/Drawing/Imaging/Effects/ColorMatrixEffect.cs
+++ b/src/System.Drawing.Common/src/System/Drawing/Imaging/Effects/ColorMatrixEffect.cs
@@ -17,7 +17,7 @@ namespace System.Drawing.Imaging.Effects;
 ///  </para>
 /// </remarks>
 [RequiresPreviewFeatures]
-public sealed unsafe class ColorMatrixEffect : Effect
+public unsafe class ColorMatrixEffect : Effect
 {
     private readonly ColorMatrix _matrix;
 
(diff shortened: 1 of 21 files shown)


src/System.Drawing.Common/src/System/Drawing/Imaging/Effects/ColorMatrixEffect.cs
```diff
@@ -35,76 +35,5 @@ public ColorMatrixEffect(ColorMatrix matrix) : base(PInvoke.ColorMatrixEffectGui
 
         _matrix = matrix;
     }
-
-    /// <summary>
-    ///  An effect that converts an image to grayscale.
-    /// </summary>
-    public static ColorMatrixEffect GrayScaleEffect()
```
Tanya-Solyanik, line 42: Should these proeprties stay and be conditionally compiled for the standalone package?

JeremyKuhne: I'm not sure what you're asking. They were moved to classes.

lonitra (approved):

src/System.Drawing.Common/src/System/Drawing/Imaging/Effects/ColorMatrixEffect.cs
Tanya-Solyanik: The new classes are only included in NET9+. Are we still building a standalone nuget package from these sources?

Issue #10185, opened 2023-10-25
Cursor icon does not display correctly on form designer with .NET 9.0 targeted


Philip-Wang01:
### .NET version

9.0.100-alpha.1.23524.13

### Did it work in .NET Framework?

Yes

### Did it work in any of the earlier releases of .NET Core or .NET 5+?

Yes, this is a regression issue that will not be reproduced in .NET 6.0/7.0/8.0.

### Issue description

Cursor icon does not display correctly on form designer with .NET 9.0 targeted
![image](https://github.com/dotnet/winforms/assets/86937911/e43cd095-8d95-46c7-8de8-49dbf241b703)
**The cursor icon changes are not particularly stable.**
![test90cursor](https://github.com/dotnet/winforms/assets/86937911/1684c1ed-3c65-4baa-811d-41b7f35f244a)

### Steps to reproduce

1. Create a WinForms .NET 8.0 project (VS does not currently support the creation of 9.0 projects).
2. Change the `TargetFramework` from `net8.0-windows` to `net9.0-windows` in the csproj file.
3. Clean and build the solution, reload form designer.
4. Add some controls to the form designer and observe the cursor icon.

lonitra:
The cause of this issue is due to the removal of code in `CursorConverter.ConvertTo` method in https://github.com/dotnet/winforms/pull/10005. For some reason, we are unable to get the property name for one of our own cursors..

elachlan:
@Philip-Wang01 does this issue still reproduce?

Philip-Wang01:
@elachlan This issue still repro on the latest .NET 9.0: 9.0.100-alpha.1.24059.8.

elachlan:
@lonitra I changed the existing test for convert to string and tested each Cursor value. It passed. 

I also haven't noticed anything strange in the demo console. Should it reproduce there or only in the designer?

lonitra:
> @lonitra I changed the existing test for convert to string and tested each Cursor value. It passed.
> 
> I also haven't noticed anything strange in the demo console. Should it reproduce there or only in the vs designer?

Interesting 🤔 I hadn't tried to repro it on demo console before. I had only been able to repro it in vs designer by using dlls built from main. I have not had a chance to spend more time with this, but my original suspicion was something needed to be updated on designer side.

MelonWang1:
[...]

PR #10814, opened 2024-02-03
Add Fallback to CursorConverter.ConvertTo

lonitra:
Fixes https://github.com/dotnet/winforms/issues/10185

We had previously removed this code in https://github.com/dotnet/winforms/pull/10005 in an attempt to avoid using reflection when converting Cursor->string. We had tried to avoid reflection by adding property `CursorsProperty` to `Cursor` to track the "string version" of the Cursor. This will work as expected when we are setting the cursor, however there are times when other processes set the cursor, in which we only have the handle available to us i.e. we cannot utilize `CursorsProperty` to properly convert Cursor->string since we will not have that information. To handle this case, we should fall back to our old code to convert the cursor properly.

Manually testing results:
cursor

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/winforms/pull/10814)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Input/CursorConverter.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Input/CursorConverter.cs
@@ -91,6 +91,32 @@ public class CursorConverter : TypeConverter
                     return propertyName;
                 }
 
+                PropertyInfo[] props = GetProperties();
+                int bestMatch = -1;
+
+                for (int i = 0; i < props.Length; i++)
+                {
+                    PropertyInfo prop = props[i];
+                    object[]? tempIndex = null;
+                    Cursor? c = (Cursor?)prop.GetValue(null, tempIndex);
+                    if (c == cursor)
+                    {
+                        if (ReferenceEquals(c, value))
+                        {
+                            return prop.Name;
+                        }
+                        else
+                        {
+                            bestMatch = i;
+                        }
+                    }
+                }
+
+                if (bestMatch != -1)
+                {
+                    return props[bestMatch].Name;
+                }
+
                 // We throw here because we cannot meaningfully convert a custom
                 // cursor into a string. In fact, the ResXResourceWriter will use
                 // this exception to indicate to itself that this object should
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Input/CursorConverter.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Input/CursorConverter.cs
@@ -97,10 +97,10 @@ public class CursorConverter : TypeConverter
                 for (int i = 0; i < props.Length; i++)
                 {
                     PropertyInfo prop = props[i];
-                    Cursor? c = (Cursor?)prop.GetValue(obj: null, index: null);
-                    if (c == cursor)
+                    Cursor? knownCursor = (Cursor?)prop.GetValue(obj: null, index: null);
+                    if (knownCursor == cursor)
                     {
-                        if (ReferenceEquals(c, value))
+                        if (ReferenceEquals(knownCursor, value))
                         {
                             return prop.Name;
                         }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Input/CursorConverter.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Input/CursorConverter.cs
@@ -97,8 +97,7 @@ public class CursorConverter : TypeConverter
                 for (int i = 0; i < props.Length; i++)
                 {
                     PropertyInfo prop = props[i];
-                    object[]? tempIndex = null;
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Input/CursorConverter.cs
```diff
@@ -91,6 +91,31 @@ public override bool CanConvertTo(ITypeDescriptorContext? context, Type? destina
                     return propertyName;
                 }
 
+                PropertyInfo[] props = GetProperties();
+                int bestMatch = -1;
+
+                for (int i = 0; i < props.Length; i++)
+                {
+                    PropertyInfo prop = props[i];
+                    Cursor? c = (Cursor?)prop.GetValue(obj: null, index: null);
```
Tanya-Solyanik, line 100: Please use full words for local variable names

Tanya-Solyanik (dismissed):
Great fix!

src/System.Windows.Forms/src/System/Windows/Forms/Input/CursorConverter.cs
lonitra: Whoops, thank you!

Tanya-Solyanik (approved):

src/System.Windows.Forms/src/System/Windows/Forms/Input/CursorConverter.cs
```diff
@@ -91,6 +91,31 @@ public override bool CanConvertTo(ITypeDescriptorContext? context, Type? destina
                     return propertyName;
                 }
 
+                PropertyInfo[] props = GetProperties();
+                int bestMatch = -1;
+
+                for (int i = 0; i < props.Length; i++)
```
JeremyKuhne, line 97: As we discussed, this is somewhat obtuse. It would be good to add a comment that explains what this is doing.

``` C#
// Try to find an exact instance match to a Cursors property, falling back on HCURSOR equality (==) with one
```

src/System.Windows.Forms/src/System/Windows/Forms/Input/CursorConverter.cs
```diff
@@ -91,6 +91,31 @@ public override bool CanConvertTo(ITypeDescriptorContext? context, Type? destina
...
+                {
+                    PropertyInfo prop = props[i];
+                    Cursor? knownCursor = (Cursor?)prop.GetValue(obj: null, index: null);
+                    if (knownCursor == cursor)
+                    {
+                        if (ReferenceEquals(knownCursor, value))
```
JeremyKuhne, line 103: In theory we shouldn't have to do this anymore as it would always be caught by the optimized path above. You wouldn't need `bestMatch` either, just the first `==`.  I think I'm seeing this right?

JeremyKuhne (changes requested):

src/System.Windows.Forms/src/System/Windows/Forms/Input/CursorConverter.cs
JeremyKuhne: The wording here is contingent on my other comment. Explaining the scenario would probably be good too.

src/System.Windows.Forms/src/System/Windows/Forms/Input/CursorConverter.cs
lonitra: We can remove this, but seems like `bestMatch` was actually there because Cursors.Arrow and Cursors.Default share the same handle, so we had always ended up returning "Default" for Cursors.Arrow. I still removed this, but I needed to also add this case. Decided to put it before the loop to avoid doing the check every time.

JeremyKuhne (dismissed):
Looks good! One could walk backwards to replicate the old behavior, but I think it is better to be explicit about the Default case (which may have been accidental, honestly).

src/System.Windows.Forms/src/System/Windows/Forms/Input/CursorConverter.cs
```diff
@@ -90,6 +90,25 @@ public override bool CanConvertTo(ITypeDescriptorContext? context, Type? destina
...
                     return propertyName;
                 }
+                else if (cursor.Handle == Cursors.Arrow.Handle)
+                {
+                    // Arrow and Default cursors share the same HCURSOR.
+                    // Always return "Default" in this case.
+                    return nameof(Cursors.Default);
+                }
+
+                // We have a cursor that only has handle information. This can happen if other processes set the cursor.
```
Tanya-Solyanik, line 100: It doesn't matter what process sets the cursor, the same process can be setting the cursor and then reading the cursor with Cursor.Current.

Tanya-Solyanik (approved):

JeremyKuhne (approved):

PR #10832, opened 2024-02-05

gpetrou:
## Proposed changes

###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/10832)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGridInternal/GridEntry.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGridInternal/GridEntry.cs
@@ -764,7 +764,7 @@ internal abstract partial class GridEntry : GridItem, ITypeDescriptorContext
     }
 
     // Note: we don't do set because of the value class semantics, etc.
-
+#nullable enable
     public sealed override object Value => PropertyValue;
 
     internal Point ValueToolTipLocation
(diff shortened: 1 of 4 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGridInternal/GridEntry.cs
```diff
@@ -849,11 +849,11 @@ internal void ClearCachedValues(bool clearChildren = true)
     /// <summary>
     ///  Converts the given string of text to a value.
     /// </summary>
-    public object ConvertTextToValue(string text)
+    public object? ConvertTextToValue(string? text)
     {
         if (TypeConverter.CanConvertFrom(this, typeof(string)))
         {
-            return TypeConverter.ConvertFromString(this, text);
+            return TypeConverter.ConvertFromString(this, text!);
```
lonitra, line 856: Do we get `NotSupportedException` thrown when `text = null`? If so, I don't think we should be indicating that null is an acceptable parameter for this method

gpetrou: This seems to return an empty string when text is null.

lonitra: Thanks for checking that! Could you add a comment that we will get empty string when text = null for others looking at this next time?

lonitra (approved):
🚀

Issue #10827, opened 2024-02-05
Refactor classes with multiple private boolean fields into `BitVector32`


elachlan:
I performed a search for "private bool _" and filtered out classes with 2 or less fields, the collated and ordered results are in the table below.

I believe we can convert these to use `BitVector32` to save runtime memory usage.

| File | Count |
|--------|--------|
|D:\Development\winforms\src\System.Windows.Forms\src\System\Windows\Forms\Controls\ComboBox\ComboBox.cs|	16|
|D:\Development\winforms\src\System.Windows.Forms\src\System\Windows\Forms\Layout\Containers\SplitContainer.cs|	13|
|D:\Development\winforms\src\System.Windows.Forms\src\System\Windows\Forms\Controls\ListBoxes\ListBox.cs|	12|
|D:\Development\winforms\src\System.Windows.Forms\src\System\Windows\Forms\DataBinding\BindingSource.cs|	12|
|D:\Development\winforms\src\System.Windows.Forms\src\System\Windows\Forms\ToolTip\ToolTip.cs|	9|
|D:\Development\winforms\src\System.Windows.Forms\src\System\Windows\Forms\Controls\PropertyGrid\PropertyGrid.cs|	6|
|D:\Development\winforms\src\System.Windows.Forms\src\System\Windows\Forms\Controls\ToolStrips\ToolStripManager.ModalMenuFilter.cs|	6|
|D:\Development\winforms\src\System.Windows.Forms\src\System\Windows\Forms\DataBinding\Binding.cs|	6|
|D:\Development\winforms\src\System.Windows.Forms\src\System\Windows\Forms\Dialogs\TaskDialog\TaskDialog.cs|	6|
|D:\Development\winforms\src\System.Windows.Forms\src\System\Windows\Forms\Controls\MonthCalendar\MonthCalendar.cs|	5|
|D:\Development\winforms\src\System.Windows.Forms\src\System\Windows\Forms\DataBinding\CurrencyManager.cs|	5|
|D:\Development\winforms\src\System.Windows.Forms\src\System\Windows\Forms\ErrorProvider\ErrorProvider.cs|	5|
|D:\Development\winforms\src\System.Windows.Forms\src\System\Windows\Forms\Printing\PrintPreviewControl.cs|	5|
|D:\Development\winforms\src\System.Windows.Forms\src\System\Windows\Forms\Controls\ListControl\ListControl.cs|	4|
[...]

PR #10843, opened 2024-02-06
Refactor `BindingSource` and `CurrencyManager` to use flags enum field instead of individual boolean fields

elachlan:
Related: #10827
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/10843)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataBinding/BindingSource.BindingSourceStates.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataBinding/BindingSource.BindingSourceStates.cs
@@ -21,5 +21,4 @@ public partial class BindingSource
         InnerListChanging           = 0b0000_0100_0000_0000,
         EndingEdit                  = 0b0000_1000_0000_0000,
     }
-
 }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/DataBinding/BindingSource.BindingSourceStates.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/DataBinding/BindingSource.BindingSourceStates.cs
@@ -6,7 +6,7 @@ namespace System.Windows.Forms;
 public partial class BindingSource
 {
     [Flags]
-    private enum BindingSourceStates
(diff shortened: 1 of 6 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/DataBinding/BindingSource.BindingSourceStates.cs
```diff
@@ -0,0 +1,24 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace System.Windows.Forms;
+
+public partial class BindingSource
+{
+    [Flags]
+    private enum BindingSourceStates : short
+    {
+        ParentsCurrentItemChanging = 1 << 0,
```
JeremyKuhne, line 11: Ideally we'd have all of our bitflags using the new binary constant feature:

``` C#
= 0b00000000_00000001,
```

They should also be aligned on the equals so you can see the bit positioning clearly.

src/System.Windows.Forms/src/System/Windows/Forms/DataBinding/BindingSource.cs
```diff
@@ -33,20 +33,23 @@ public class BindingSource : Component,
...
+
     // Public property values
     private object? _dataSource;
     private string _dataMember = string.Empty;
     private string? _sort;
     private string? _filter;
     private readonly CurrencyManager _currencyManager;
-    private bool _parentsCurrentItemChanging;
-    private bool _disposedOrFinalized;
+    private bool _parentsCurrentItemChanging { get => _state.HasFlag(BindingSourceStates.ParentsCurrentItemChanging); set => _state.ChangeFlags(BindingSourceStates.ParentsCurrentItemChanging, value); }
```
JeremyKuhne, line 45: It is better to do this inline everywhere now. We don't need wrapper properties (it will likely optimize better as well).

lonitra (approved):
LGTM

PR #10849, opened 2024-02-08
Fix empty `OpenFileDialog.FileNames` when `AutoUpgradeEnabled=false` and `Multiselect = true`

### LINKED ISSUE #10846: [BUG] System::Windows::Forms::OpenFileDialog::AutoUpgradeEnabled does not work correctly with NetCore 8.0
When setting AutoUpgradeEnabled to false in an System::Windows::Forms::OpenFileDialog

- everything goes right in a .Net 4.8 project
- no filenames are returned when exiting the dialog in a .Net 8.0 project (netcore8.0-windows)

It seems related to Multiselect is true

To be able to use System::Windows::Forms::OpenFileDialog in a NetCore 8.0 project, `<UseWindowsForms>true</UseWindowsForms>` must be used.

I don't know if it's relevant, but I have also referenced Microsoft.Windows.Compatibility in my project.

Is there any workaround ?

I need this feature to use customized dialogs.

My reproducer :

```
System::Windows::Forms::OpenFileDialog^ openFileDialog = gcnew System::Windows::Forms::OpenFileDialog();
openFileDialog->AutoUpgradeEnabled = false;
openFileDialog->Title = "Load...";
openFileDialog->AddExtension = true;
openFileDialog->Multiselect = true;
openFileDialog->DefaultExt = "cfg";
System::String^ filterString = "";
filterString += System::String::Format("{0}{1}",
  System::String::IsNullOrEmpty(filterString) ? "" : "|",
  "Cfg files" + " (*.cfg)|*.cfg");
filterString += System::String::Format("{0}{1}",
  System::String::IsNullOrEmpty(filterString) ? "" : "|",
  "All files" + " (*.*)|*.*");
openFileDialog->Filter = filterString;
System::Windows::Forms::DialogResult code = openFileDialog->ShowDialog(this);
//select one or a few files
System::Console::WriteLine("selected count = {0}", openFileDialog->FileNames->Length);//returns 0
```

elachlan:
Fixes #10846
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/10849)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Dialogs/CommonDialogs/FileDialog.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Dialogs/CommonDialogs/FileDialog.cs
@@ -457,6 +457,9 @@ public abstract partial class FileDialog : CommonDialog
     {
         var directory = fileBuffer.SliceAtFirstNull();
         var fileNames = fileBuffer[(directory.Length + 1)..];
+
+        // When a single file is returned, the directory is not null delimited.
+        // So we check here to see if the filename starts with a null.
         if (fileNames[0] == '\0')
         {
             return [directory.ToString()];
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Dialogs/CommonDialogs/FileDialog.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Dialogs/CommonDialogs/FileDialog.cs
@@ -456,7 +456,7 @@ public abstract partial class FileDialog : CommonDialog
     private static string[] GetMultiselectFiles(ReadOnlySpan<char> fileBuffer)
     {
         var directory = fileBuffer.SliceAtFirstNull();
-        var fileNames = fileBuffer[directory.Length..];
+        var fileNames = fileBuffer[(directory.Length + 1)..];
         if (fileNames.Length == 0)
         {
             return [directory.ToString()];
@@ -470,7 +470,7 @@ public abstract partial class FileDialog : CommonDialog
                 ? fileName.ToString()
                 : Path.Join(directory, fileName));
 
-            fileNames = fileNames[fileName.Length..];
+            fileNames = fileNames[(fileName.Length + 1)..];
             fileName = fileNames.SliceAtFirstNull();
         }
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Dialogs/CommonDialogs/FileDialog.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Dialogs/CommonDialogs/FileDialog.cs
@@ -457,12 +457,12 @@ public abstract partial class FileDialog : CommonDialog
     {
         var directory = fileBuffer.SliceAtFirstNull();
         var fileNames = fileBuffer[(directory.Length + 1)..];
-        if (fileNames.Length == 0)
+        if (fileNames[0] == '\0')
         {
             return [directory.ToString()];
         }
 
-        List<string> names = new();
+        List<string> names = [];
         var fileName = fileNames.SliceAtFirstNull();
         while (fileName.Length > 0)
         {
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Dialogs/CommonDialogs/FileDialog.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Dialogs/CommonDialogs/FileDialog.cs
@@ -460,7 +460,7 @@ public abstract partial class FileDialog : CommonDialog
 
         // When a single file is returned, the directory is not null delimited.
         // So we check here to see if the filename starts with a null.
-        if (fileNames[0] == '\0')
+        if (fileNames.Length == 0 || fileNames[0] == '\0')
         {
             return [directory.ToString()];
         }
--- a/src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/Dialogs.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/Dialogs.cs
@@ -34,7 +34,7 @@ public partial class Dialogs : Form
             if (propertyGrid1.SelectedObject is OpenFileDialog openFileDialog)
             {
                 openFileDialog.ShowDialog(this);
-                MessageBox.Show(string.Join(',', openFileDialog.FileNames),"File Names");
+                MessageBox.Show(string.Join(',', openFileDialog.FileNames), "File Names");
                 return;
             }
 
--- a/src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/Dialogs.cs
+++ b/src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/Dialogs.cs
@@ -31,6 +31,13 @@ public partial class Dialogs : Form
 
         _btnOpen.Click += (s, e) =>
         {
+            if (propertyGrid1.SelectedObject is OpenFileDialog openFileDialog)
+            {
+                openFileDialog.ShowDialog(this);
+                MessageBox.Show(string.Join(',', openFileDialog.FileNames),"File Names");
+                return;
+            }
+
             if (propertyGrid1.SelectedObject is CommonDialog dialog)
             {
                 dialog.ShowDialog(this);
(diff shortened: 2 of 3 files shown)


elachlan:
I am unsure on a regression test.

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/Dialogs.cs
```diff
@@ -31,6 +31,13 @@ public Dialogs()
 
         _btnOpen.Click += (s, e) =>
         {
+            if (propertyGrid1.SelectedObject is OpenFileDialog openFileDialog)
+            {
+                openFileDialog.ShowDialog(this);
+                MessageBox.Show(string.Join(',', openFileDialog.FileNames),"File Names");
```
RussKie, line 37: [nit]
```suggestion
                MessageBox.Show(string.Join(',', openFileDialog.FileNames), "File Names");
```

lonitra:
Thanks for jumping on this! Changes look good. We have [TestAccessors](https://github.com/dotnet/winforms/blob/226a5b5d3782f40d8c84cdacfa3c49b3f7cc042e/src/Common/tests/TestUtilities/TestAccessors.cs#L44) which can dynamically access private methods -- though this is a method coming from an abstract class so I'm not too sure if this would still work, but could you try writing a test that calls `GetMultiselectFiles` using `TestAccessors`, passing in file buffer that would represent 2 items selected to see if it gets parsed correctly?

elachlan:
@lonitra we run into 
`CS1978	Cannot use an expression of type 'ReadOnlySpan<char>' as an argument to a dynamically dispatched operation.`

```csharp
[WinFormsFact]
public void FileDialog_GetMultiselectFiles_ReturnsExpected()
{
    // Test with directory
    string buffer = "C:\\test\0testfile.txt\0testfile2.txt\0";
    string[] expected = ["C:\\test\\testfile.txt", "C:\\test\\testfile2.txt"];
    string[] result = typeof(FileDialog).TestAccessor().Dynamic.GetMultiselectFiles(buffer.AsSpan());
    Assert.Equal(expected, result);

    // Test without directory
    buffer = "C:\\\0testfile.txt\0testfile2.txt\0";
    expected = ["C:\\testfile.txt", "C:\\testfile2.txt"];
    result = typeof(FileDialog).TestAccessor().Dynamic.GetMultiselectFiles(buffer);
    Assert.Equal(expected, result);
}
```

If I change `GetMultiselectFiles` to internal we can test it quite easily since its static.
```csharp
string buffer = "C:\\test\0testfile.txt\0testfile2.txt\0";
string[] expected = ["C:\\test\\testfile.txt", "C:\\test\\testfile2.txt"];
string[] result = FileDialog.GetMultiselectFiles(buffer);
Assert.Equal(expected, result);
```

lonitra:
Hm, we are trying to avoid changing visibility to `internal` for solely test purposes which is why `TestAccessors` was added to begin with. @JeremyKuhne any ideas if there are other ways we can do this without changing visibility of method?

src/System.Windows.Forms/src/System/Windows/Forms/Dialogs/CommonDialogs/FileDialog.cs
```diff
@@ -456,7 +456,7 @@ private protected static bool FileExists(string? fileName)
     private static string[] GetMultiselectFiles(ReadOnlySpan<char> fileBuffer)
     {
         var directory = fileBuffer.SliceAtFirstNull();
-        var fileNames = fileBuffer[directory.Length..];
+        var fileNames = fileBuffer[(directory.Length + 1)..];
```
Dave-Lowndes, line 459: When debugging this code from the example repro (using .NET 8), the input (fileBuffer) looks like this:
"C:\\Test\\AFile.ext\0\0\0\0..."
I don't see how this change is going to fix the problem.
Can you explain?

src/System.Windows.Forms/src/System/Windows/Forms/Dialogs/CommonDialogs/FileDialog.cs
```diff
@@ -456,7 +456,7 @@ private protected static bool FileExists(string? fileName)
...
         if (fileNames.Length == 0)
```
Dave-Lowndes, line 460: To resolve the single item issue, this line probably needs to be:
if ( fileNames[0] == '\0')

Some comments in the code to explain the intention wouldn't go amiss either.

elachlan:
@lonitra we have already implemented tests previously for `OpenFileDialog`. I have been investigating if we can use something similar.

https://github.com/dotnet/winforms/blob/7dda32bf3dd18f6c20b4299dec09f80509f2267c/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/OpenFileDialogTests.cs#L34-L54

This is what I thought might work, but the buffer gets overwritten with the full path.
```csharp
    [WinFormsFact]
    public void OpenFileDialog_ResultWithMultiselect_AutoUpgradeDisabled()
    {
        using var tempFile = TempFile.Create(0);
        using var tempFile2 = TempFile.Create(0);
        string initialDirectory = Path.GetDirectoryName(tempFile.Path)!;
        string[] fileNames = [Path.GetFileName(tempFile.Path), Path.GetFileName(tempFile2.Path)];
        using AcceptDialogForm dialogOwnerForm = new();
        using OpenFileDialog dialog = new()
        {
            AutoUpgradeEnabled = false,
            Multiselect = true,
            CheckFileExists = false,
            InitialDirectory = initialDirectory,
            FileName = string.Join("\0", fileNames)
        };

        Assert.Equal(DialogResult.OK, dialog.ShowDialog(dialogOwnerForm));
        Assert.Equal(fileNames, dialog.FileNames);
    }
```

Doing this with `AutoUpgradeEnabled=true` works, but fails with `AutoUpgradeEnabled=false`:
```csharp
openFileDialog.FileName = "\"test.txt\" \"test2.txt\"";
```

lonitra:
@elachlan Does it work if full path is given instead of just file names?
I meant to comment this but I guess I did not actually comment here. We can need to define a delegate if we want to use `TestAccessor` to make a test. An example is outlined here:
https://github.com/dotnet/winforms/blob/2b00f7d2bf2ce788056218c5c1f5ea0afeaf4015/src/Common/tests/TestUtilities/ITestAccessor.cs#L40-L57

It would be good to add a test for one file selected as well.

JeremyKuhne:
> Hm, we are trying to avoid changing visibility to `internal` for solely test purposes which is why `TestAccessors` was added to begin with. @JeremyKuhne any ideas if there are other ways we can do this without changing visibility of method?

See `ITestAccessor.CreateDelegate` for details on how this is done.

https://github.com/dotnet/winforms/blob/e13647da8f159e329100697189992282640561a4/src/Common/tests/TestUtilities/ITestAccessor.cs#L29-L60

src/System.Windows.Forms/src/System/Windows/Forms/Dialogs/CommonDialogs/FileDialog.cs
elachlan: Thank you!

elachlan:
@lonitra Thank you for that! I have pushed through the tests and they are passing locally.

src/System.Windows.Forms/src/System/Windows/Forms/Dialogs/CommonDialogs/FileDialog.cs
Dave-Lowndes: You're welcome; though to be more robust, it should probably be: if ((fileNames.Length == 0) || (fileNames[0]=='\0')) - with a comment to explain why!

elachlan: I've added a comment already. 

Can you explain why you think the parenthesis are needed?

src/System.Windows.Forms/src/System/Windows/Forms/Dialogs/CommonDialogs/FileDialog.cs
elachlan: In the single file scenario, the directory holds the full path to the file.

lonitra (approved):
Awesome 🚀

src/System.Windows.Forms/src/System/Windows/Forms/Dialogs/CommonDialogs/FileDialog.cs
Dave-Lowndes: It's how I write code - don't rely on my memory for any vagaries of operator precedence :)

PR #10854, opened 2024-02-08
Use Native DataObject Pointer in ComDataObjectAdapter

### LINKED ISSUE #10853: Drag/drop Not Working with External Items When Trimming On
Drag/drop external items does not work when trimming is turned on. We are expecting the managed object we get on [DropTarget.PrepareOutgoingDropData](https://github.com/dotnet/winforms/blob/f7d1be6c8af5445fe9ad041ad2f8f62ddc4a8ece/src/System.Windows.Forms/src/System/Windows/Forms/OLE/DropTarget.cs#L56) to be either a `System.Windows.Forms.IDataObject` or `System.Runtime.InteropServices.ComTypes.IDataObject`, but the object we get back is neither of these. This is because the object is not a ComWrappers created object and built in com interop is not supported, so we are not getting that automatic wrapping that built in com provides. We need to be wrapping the COM object ourselves.

Repro:
1. Navigate to winforms TrimTest project
2. Edit the .csproj to turn on trimming:
```.csproj
<Project Sdk="Microsoft.NET.Sdk">
 <PropertyGroup>
    <OutputType>WinExe</OutputType>
    <TargetFramework>net9.0-windows</TargetFramework>
    <UseWindowsForms>true</UseWindowsForms>
    <PublishTrimmed>true</PublishTrimmed>
    <ImplicitUsings>true</ImplicitUsings>
    <_SuppressWinFormsTrimError>true</_SuppressWinFormsTrimError>
  </PropertyGroup>
</Project>
```
3. Run 'dotnet publish' in command line where TrimTest lives
4. Run published TrimTest and drag a string from somewhere external onto the TextBox
5. Observe being unable to drop

lonitra:
Fixes #10853 

When dealing with COM pointers that are not ComWrappers-created unmanaged wrappers and built-in COM interop is turned on, `Marshal.GetObjectForIUnknown` generates an RCW under the hood for us to use, which is why we are able to cast the COM data object  we receive back seamlessly to `System.Runtime.InteropServices.ComTypes`. However when built-in COM interop is turned off and the COM pointer we receive is not a ComWrappers-created unmanaged wrapper, we need to be able to work off of the native pointer

Trimmed drag/drop
Before:
dragb4

After:
drag

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/10854)


lonitra:
cc: @LakshanF

src/System.Windows.Forms/src/System/Windows/Forms/OLE/DropTarget.cs
```diff
@@ -50,19 +50,29 @@ private void ClearDropDescription()
 
     /// <summary>
     ///  Prepares the drag data to be passed out.
-    ///  <paramref name="dataObj"/> should implement <see cref="IDataObject"/> to be passed out as is.
+    ///  <paramref name="managedDataObject"/> should implement <see cref="IDataObject"/> to be passed out as is.
     ///  Otherwise, the data will be wrapped in a <see cref="DataObject"/>.
     /// </summary>
-    private static IDataObject? PrepareOutgoingDropData(object dataObj)
+    private static IDataObject? PrepareOutgoingDropData(object managedDataObject)
```
lonitra, line 56: It would be nice if we could directly pass `Com.IDataObject*` here to avoid ptr->object->ptr in the last case, but I couldn't make that work in tests when we invoke this method dynamically.

src/System.Windows.Forms.Primitives/src/Windows/Win32/System/Com/FORMATETC.cs
```diff
@@ -5,15 +5,25 @@
...
-    public static unsafe implicit operator FORMATETC(ComType.FORMATETC formatetc) =>
+    public static implicit operator FORMATETC(ComType.FORMATETC formatEtc) =>
         new()
         {
-            cfFormat = (ushort)formatetc.cfFormat,
-            ptd = (DVTARGETDEVICE*)formatetc.ptd,
-            dwAspect = (uint)formatetc.dwAspect,
-            lindex = formatetc.lindex,
-            tymed = (uint)formatetc.tymed
+            cfFormat = (ushort)formatEtc.cfFormat,
```
JeremyKuhne, line 13: As these are bit-equivalent you can use `Unsafe.As<T,T>` to convert them or simply cast pointer types when you are dealing with pointers (such as passing FORMATETC* in).

JeremyKuhne (changes requested):
Triple check ref counts.

src/System.Windows.Forms.Primitives/src/Windows/Win32/System/Com/STATDATA.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace Windows.Win32.System.Com;
+
+internal unsafe partial struct STATDATA
+{
+    public static implicit operator STATDATA(ComType.STATDATA statData) =>
+        new()
+        {
+            formatetc = Unsafe.As<ComType.FORMATETC, FORMATETC>(ref statData.formatetc),
+            advf = (uint)statData.advf,
+            pAdvSink = ComHelpers.GetComPointer<IAdviseSink>(statData.advSink),
```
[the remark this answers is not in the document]
JeremyKuhne: I'm still unclear on how this IAdviseSink add ref gets released.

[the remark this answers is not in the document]
lonitra: We have one usage of this cast currently, which is here https://github.com/dotnet/winforms/pull/10854/files#diff-23196ea13d7ab5a81c9001467c40bd4ada54be0f7a591a7e7e6845c987ecd709R43 and I am releasing before end of that method. Maybe I could implement `IDisposable` additionally?

src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.ComDataObjectAdapter.cs
```diff
@@ -472,46 +460,37 @@ public string[] GetFormats(bool autoConvert)
...
-            HashSet<string> distinctFormats = new();
+            HashSet<string> distinctFormats = [];
 
-            enumFORMATETC.Reset();
+            enumFORMATETC.Value->Reset();
 
-            FORMATETC[] formatetc = [default];
-            int[] retrieved = [1];
+            Com.FORMATETC[] formatEtc = [default];
+            uint[] retrieved = [1];
```
JeremyKuhne, line 477: Better to just use `uint` and not an array. Then you won't allocate an array or have to pin.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataObjectTests.cs
```diff
@@ -2104,10 +2105,10 @@ public void DataObject_EnumFormatEtc_InvokeNotGet_ThrowsExternalException(DATADI
 
     public static IEnumerable<object[]> EnumFormatEtc_CustomComDataObject_TestData()
     {
-        yield return new object[] { DATADIR.DATADIR_GET, null };
+/*        yield return new object[] { DATADIR.DATADIR_GET, null };
```
JeremyKuhne, line 2108: Uncomment

lonitra: This should actually be removed since the test below is mocking `ComTypes.IDataObject.EnumFormatEtc` to return null, but this shouldn't return null if native method succeeded.

src/System.Windows.Forms.Primitives/src/Windows/Win32/System/Com/STATDATA.cs
```diff
@@ -0,0 +1,43 @@
...
+            advSink = ComHelpers.TryGetObjectForIUnknown(
+                pAdvSink.Query<IUnknown>(),
+                takeOwnership: true,
+                out ComType.IAdviseSink? adviseSink)
+                    ? adviseSink
+                    : new AdviseSinkWrapper(pAdvSink),
+            connection = (int)statData.dwConnection
+        };
+
+        if (adviseSink is not null)
```
JeremyKuhne, line 36: I'm pretty sure we always need to release `pAdvSink` if it was passed to us, presumably by just putting a `using` up above.

lonitra: We will always release it. If `adviseSink` is null, it will be released in `AdviseSinkWrapper` via the `AgileComPointer` we create there.

src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.EnumStatDataWrapper.cs
```diff
@@ -0,0 +1,83 @@
...
+        {
+            using var enumStatData = _enumStatData.GetInterface();
+            Com.IEnumSTATDATA* result;
+            enumStatData.Value->Clone(&result).ThrowOnFailure();
+            newEnum = new EnumStatDataWrapper(result);
+        }
+
+        unsafe int IEnumSTATDATA.Next(int celt, STATDATA[] rgelt, int[] pceltFetched)
+        {
+            if ((pceltFetched is not null && pceltFetched.Length == 0)
```
JeremyKuhne, line 38: if (pceltFetched is null || pceltFetched.Length == 0) return E_POINTER

src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.EnumStatDataWrapper.cs
```diff
@@ -0,0 +1,83 @@
...
+            using var enumStataData = _enumStatData.GetInterface();
+            Com.STATDATA[] nativeStatData = new Com.STATDATA[rgelt.Length];
+            HRESULT result;
+            fixed (int* ppceltFetched = pceltFetched)
+            fixed (Com.STATDATA* pNativeStatData = nativeStatData)
+            {
+                result = enumStataData.Value->Next((uint)celt, pNativeStatData, (uint*)ppceltFetched);
+            }
+
+            for (int i = 0; i < nativeStatData.Length; i++)
```
JeremyKuhne, line 63: This has to be ppceltFetched, not the full length.

src/System.Windows.Forms.Primitives/src/Windows/Win32/System/Com/STATDATA.cs
JeremyKuhne: Ah. It would be clearer if we did not takeOwnership and just always release.

src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.EnumStatDataWrapper.cs
JeremyKuhne: Both are effectively "null" for what would be the int*

JeremyKuhne (approved):

PR #10855, opened 2024-02-09
Fix the automation for merged/closed PRs

jeffhandley:
This is follow-up to #10811. Automation related to closed/merged PRs was not working because of an issue in the migration where there wasn't a condition for checking if a PR was merged. The 'merged' action was carried over as 'Null' in the previous PR.


lonitra (approved):
Thank you!

PR #10857, opened 2024-02-09
Add missing Stream overloads in DataStreamFromComStream

halgab:
## Proposed changes

- Add optimised overloads to methods `DataStreamFromComStream.ReadByte` and `WriteByte`

## Customer Impact

- remove some small arrays allocations

## Risk

- low

## Test methodology 

- CI

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/10857)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ActiveX/DataStreamFromComStream.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ActiveX/DataStreamFromComStream.cs
@@ -82,6 +82,13 @@ internal unsafe class DataStreamFromComStream : Stream
         return (int)bytesRead;
     }
 
+    public override int ReadByte()
+    {
+        Span<byte> oneByteArray = stackalloc byte[1];
+        int r = Read(oneByteArray);
+        return r == 0 ? -1 : oneByteArray[0];
+    }
+
     public override void SetLength(long value)
     {
         _comStream->SetSize((ulong)value);
@@ -139,6 +146,8 @@ internal unsafe class DataStreamFromComStream : Stream
             throw new IOException(SR.DataStreamWrite);
         }
     }
+
+    public override void WriteByte(byte value) => Write([value]);
 
     protected override void Dispose(bool disposing)
     {
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ActiveX/DataStreamFromComStream.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ActiveX/DataStreamFromComStream.cs
@@ -84,9 +84,9 @@ internal unsafe class DataStreamFromComStream : Stream
 
     public override int ReadByte()
     {
-        Span<byte> oneByteArray = stackalloc byte[1];
-        int r = Read(oneByteArray);
-        return r == 0 ? -1 : oneByteArray[0];
+        byte data = default;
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ActiveX/DataStreamFromComStream.cs
```diff
@@ -82,6 +82,13 @@ public override int Read(Span<byte> buffer)
         return (int)bytesRead;
     }
 
+    public override int ReadByte()
+    {
+        Span<byte> oneByteArray = stackalloc byte[1];
```
JeremyKuhne, line 87: You should be able to use the new span constructor, which would be the prefered method:

``` C#
byte data;
int r = Read(new Span<byte>(ref data));
// ...
```

JeremyKuhne (changes requested):
Thanks for this! Just a minor comment.

src/System.Windows.Forms/src/System/Windows/Forms/ActiveX/DataStreamFromComStream.cs
halgab: Done. I hesitated between the two versions and initially chose this one because it didn't require me to initialise the data field.

lonitra (approved):

src/System.Windows.Forms/src/System/Windows/Forms/ActiveX/DataStreamFromComStream.cs
JeremyKuhne: @halgab you could mark this method as `[SkipLocalsInit]` if you see that it makes it more efficient (look at sharplap.io).

JeremyKuhne (approved):
This is good, if you want to further optimize see my comment on [SkipLocalsInit]

src/System.Windows.Forms/src/System/Windows/Forms/ActiveX/DataStreamFromComStream.cs
halgab: I tried looking at the il and the assembly in sharplab [(here)](https://sharplab.io/#v2:C4LglgNgNAJiDUAfAAgJgIwFgBQyAMABMugHQCSA8gNw75GkBKArgHbBgC2ApiQMID2HAA6QuAJwDK4gG5gAxlwDONXAGYiqArwIBvHAQMEJwMVwCGHAopPnLAXgIsuAdwIBZLh35iAnsdMWABQAlCqGROpgbO4huvrhhgBGPsBcBDBmwGYEDjBcAGZmTBDAYQkGUcAEYjlWNhYkDOYwgU6uEkJmLAA8yakAfIGm+emZZsGh8eXIAOzVOQ6EAPwEALToBCCjWWUGAL5TBIcA2hIA1mBCADL8cmYQimQsYMAAuofIkdFu55exethygY+mkMllanlCsVSodwpV5g5rAEOI1mq0XEZOj0QYNhttxpNAUDZgjFgQVutNvjdgQDtg9kA=). I can't really interpret the results. I just notice that the asm looks bigger with the attribute. Also, we still have to set `data` to something despite the attribute.

I guess all those concerns are minor anyway compared to the gains of just defining this method overload properly

PR #10864, opened 2024-02-12
Update System.Private.Windows.Core.csproj

ViktorHofer:
https://github.com/dotnet/winforms/pull/10525 introduced System.Private.Windows.Core.csproj which multi-targets. Unfortunately that project was declared incorrectly as the `<TargetFramework>` property needs to be unset as it's defined in the root of the repo.

This caused overbuilds of System.Private.Windows.Core.csproj which ultimately manifested in build race conditions.

The binlog shows that the `Csc` task is invoked twice for the same TFM and hence the race condition:

### Global properties for CSC 1 project evaluation
image

### Global properties for CSC 2 project evaluation
image

Notice that the only difference between those two is the `TargetFramework` global property. The first evaluation also builds for the same TFM as that property is defined in the root repo.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/10864)

### DIFF
--- a/eng/SourceBuild.props
+++ b/eng/SourceBuild.props
@@ -4,6 +4,5 @@
   <PropertyGroup>
     <GitHubRepositoryName>winforms</GitHubRepositoryName>
     <SourceBuildManagedOnly>true</SourceBuildManagedOnly>
-    <InnerBuildArgs>$(InnerBuildArgs) -warnNotAsError:MSB3026</InnerBuildArgs>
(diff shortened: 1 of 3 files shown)


akoeplinger (approved):

JeremyKuhne:
Thanks @ViktorHofer and @akoeplinger!

Issue #10813, opened 2024-02-02
Add a new feature switch for WinForms trimming


LakshanF:
Trimming WinForms will require a [feature switch](https://github.com/dotnet/designs/blob/main/accepted/2020/feature-switch.md) to be used similar to [other feature switches](https://github.com/dotnet/runtime/blob/main/docs/workflow/trimming/feature-switches.md) when trimming WinForms applications.

Exploring the trim warnings from the [TrimTest project](https://github.com/dotnet/winforms/blob/main/src/System.Windows.Forms/tests/IntegrationTests/TrimTest/TrimTest.csproj) shows the following where having a feature switch could help:
 - The vast majority of the trim warnings are actually not hit by simple WinForms applications like the TrimTest project. Code hitting trim warnings that is not currently tracked by #4649 could be refactored such that setting the feature switch would remove these code paths.
 - Its very likely the type extension mechanism offered by WinForms need to be curtailed with a feature switch for trimming. For example, [ICustomTypeDescriptor ](https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.icustomtypedescriptor) likely need to be restricted and code paths that use this interface will need to be removed via the feature switch. 
[Default_Trimmed_Warnings_wpf_attributes_excluded_ComponentModel_ICustomTypeDescriptor.txt](https://github.com/dotnet/winforms/files/14145862/Default_Trimmed_Warnings_wpf_attributes_excluded_ComponentModel_ICustomTypeDescriptor.txt)
 - Remove the current WinForms trimming when this feature switch

Issue #10741, opened 2024-01-25
[Umbrella-Item] Make WinForms DataBinding Apps trimmable


KlausLoeffelmann:
This is an investigation and prototype issue at this point (with some good prospects of success, though).

- [ ] Investigate options to improve Designer-side code generation to support Trimming for WinForms App which use Designer-hooked-up Data Binding (we will *explicitly* exclude non-type-safe Data Binding, like Runtime-Binding to DataTables/DataViews which use Property Descriptors to create "Types" at runtime.)
- [ ] Investigate runtime API-Improvements around
  - [ ] [Binding](https://source.dot.net/#System.Windows.Forms/System/Windows/Forms/DataBinding/Binding.cs,bed1e5beea29de5c)
  - [ ] [BindingSource](https://source.dot.net/#q=BindingSource)
  - [ ] [Formatter](https://source.dot.net/#System.Windows.Forms/System/Windows/Forms/Internal/Formatter.cs,e12c083610518727)
  - [ ] [ListBindingHelper](https://source.dot.net/#System.Windows.Forms/System/Windows/Forms/DataBinding/ListBindingHelper.cs,cdf058a848f1cd26)
  - [ ] [ListBindingConverter](https://source.dot.net/#System.Windows.Forms/System/Windows/Forms/DataBinding/ListBindingConverter.cs,b89feeecce85a487)
  - [ ] [BindingList<T>](https://source.dot.net/#System.ComponentModel.TypeConverter/System/ComponentModel/BindingList.cs,55afabf0a74c12cc)

merriemcgaw:
I'm super excited to see how this goes!

PR #10866, opened 2024-02-12
Explore WinForms feature switch coverage in System.Windows.Forms project

LakshanF:
This is a draft PR to solicit feedback and not meant to be merged.

Remove `nowarn` for [IL2026](https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/trim-warnings/il2026) and explore the impact on trimming. The RUC warnings are currently at leaf level and making WinForms assemblies [trim compatible](https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/fixing-warnings) require blubbling these warning up to public API level or suppressing the warnings. Bubbling up the warnings from the leaf level to public API root level will encompass a fairly large graph of API nodes, and likely make WinForms unusable with trimming. Suppressing warnings, especially at the leaf level can reduce the size of this graph but can be potentially very dangerous if not done correctly and require careful analysis to ensure that the trimmed WinForms surface area work correctly. The current plan is to use a feature switch to guard against trim problematic features in a trimmed WinForms application.

This PR explores the space of suppression so as to understand the WinForms feature switch coverage and get feedback from domain experts on problematic areas for trimming to further expand the feature switch coverage. Current identified areas where a [feature switch](https://github.com/dotnet/winforms/issues/10813) will restrict trimmed WinForms are:
- Use of [ComponentModel](https://github.com/dotnet/runtime/issues/97841) area
- Use of ICustomTypeDescriptor
- [DefaultValueAttribute](https://github.com/dotnet/runtime/issues/97842)
- [Binding ](https://github.com/dotnet/winforms/issues/10741)to move to a strongly typed model
- Designer

The commits in this PR show the impacted APIs from the leaf level bubbling up the call graph. It doesn't show the full graph since the goal is to identify the suppression points closer to the leaf level APIs. These commits also show suppressions applied at leaf level (those no longer need to be bubbled up). Current suppressions are not done as we envision how we do this in the final product. For example, the suppression is applied at method level instead of the exact location callsite. This approach was adopted to expedite the feedback.

These commits show base classes and interfaces that could cause trimming concerns since derived classes call RUC methods. This can be seen via the [IL2046](https://learn.microsoft.com/en-us/dotnet/core/deploying/trimming/trim-warnings/il2046) warning. This might mean that the feature guard could restrict usage on these types for a trimmed WinForms application. Some of these impacted types are;
 - ITypeResolutionService, ISerializable, IPropertyNotifySink, Control.BackColor
 - Binding helper classes (BindingSource, BindingManagerBase)
 - Com2ExtendedBrowsingHandler<T>
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/10866)


src/System.Windows.Forms/src/System/Windows/Forms/ActiveX/AxHost.cs
```diff
@@ -2398,6 +2399,7 @@ private void CreateInstance()
         return null;
     }
 
+    [RequiresUnreferencedCode("Calls System.ComponentModel.TypeDescriptor.GetProperties(Object)")]
```
JeremyKuhne, line 2402: @LakshanF 

This only comes into play when running in the designer (when you actually have a Site for the Component). I'm not sure how we would deal with this in a way that would give meaningful feedback. It isn't an expected runtime scenario.

I think this is one big bucket of things- anything that touches the Site.

src/System.Windows.Forms/src/System/Windows/Forms/ComponentModel/COM2Interop/COM2IManagedPerPropertyBrowsingHandler.cs
```diff
@@ -30,6 +30,7 @@ public override void RegisterEvents(Com2PropertyDescriptor[]? properties)
     ///  Here is where we handle IVsPerPropertyBrowsing.GetLocalizedPropertyInfo and IVsPerPropertyBrowsing.
     ///  Hide properties such as IPerPropertyBrowsing, IProvidePropertyBuilder, etc.
     /// </summary>
+    [RequiresUnreferencedCode("Calls System.Windows.Forms.ComponentModel.Com2Interop.Com2IManagedPerPropertyBrowsingHandler.GetComponentAttributes(IVSMDPerPropertyBrowsing*, Int32)")]
```
JeremyKuhne, line 33: This class is highly unlikely to come up. It is a VS specific support interface.

JeremyKuhne:
There is a big swath of design-time only issues here.

MichalStrehovsky:
Would it make sense to approach this from a different angle - build a blank WinForms app (with WPF removed) and start with addressing the warnings there until we get into a clean state for the blank app?

This diff is rather overwhelming and there's a lot of advanced scenarios such as ActiveX that we maybe don't need to solve _right now_. The diff will only get bigger. (Also the unnecessary whitespace diffs in enum formatting are not helping.)

Maybe this is not the case for the other reviewers, but I don't have much here to comment on because it touches so many things and I don't see myself drilling into the implementation and static callgraph of every method this touches to understand how it's used.

I just have the generic guideline of "_please please_ don't suppress warnings". There's a good chance the suppression will be wrong even with all the expertise we have Cc'd on the review. I'm filing an issue on a wrong suppression (#10868) I found in existing code while clicking around this PR.

src/System.Windows.Forms/src/System/Windows/Forms/ActiveX/AxHost.cs
agocke: I would actually expect annotating is the right answer here. Basically, we need to wall off the designer code from the app code, right?

So a mechanical way to do that is to annotate everything that uses the designer as RUC. If that code is successfully removed from all trimming code paths, the RUC warning won't fire because the code won't be called.

But this approach also helps catch bugs because it ensures that if a reference is ever added in the future, it will warn.

agocke:
Agreed with Michal on the "start with an app" approach. I do think the eventual goal should be to annotate everything, so we can get static validation that it's all being properly removed, but there's just a lot of code in WinForms and starting with the template app makes sense

PR #10872, opened 2024-02-13

gpetrou:
## Proposed changes

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/10872)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGridInternal/GridEntry.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGridInternal/GridEntry.cs
@@ -346,7 +346,7 @@ internal abstract partial class GridEntry : GridItem, ITypeDescriptorContext
             _flags |= Flags.Checked;
 
             TypeConverter converter = TypeConverter;
-            UITypeEditor editor = UITypeEditor;
+            UITypeEditor? editor = UITypeEditor;
             object value = Instance;
             bool forceReadOnly = ForceReadOnly;
 
@@ -724,11 +724,11 @@ internal abstract partial class GridEntry : GridItem, ITypeDescriptorContext
     ///  Returns the Type of the value of this <see cref="GridEntry"/>, or null if the value is null.
     /// </summary>
     public virtual Type PropertyType => PropertyValue?.GetType();
-
+#nullable enable
     /// <summary>
     ///  Gets or sets the value for the property that is represented by this <see cref="GridEntry"/>.
     /// </summary>
-    public virtual object PropertyValue
+    public virtual object? PropertyValue
     {
         get => _cacheItems?.LastValue;
         set { }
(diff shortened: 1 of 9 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGridInternal/ImmutablePropertyDescriptorGridEntry.cs
```diff
@@ -36,42 +36,45 @@ internal sealed class ImmutablePropertyDescriptorGridEntry : PropertyDescriptorG
...
             // Create a new instance of the value and set it into the parent grid entry.
-            object owner = GetValueOwner();
+            object? owner = GetValueOwner();
             object? newObject = null;
             GridEntry parentEntry = InstanceParentGridEntry;
             TypeConverter parentConverter = parentEntry.TypeConverter;
 
-            PropertyDescriptorCollection? properties = parentConverter.GetProperties(parentEntry, owner);
-            if (properties is not null)
+            if (owner is not null)
```
lonitra, line 44: Can we invert this to avoid further nesting

gpetrou: There is another if after this one.

src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGridInternal/SingleSelectRootGridEntry.cs
```diff
@@ -172,15 +172,18 @@ internal override PropertyGridView OwnerGridView
...
+    public override object? PropertyValue
     {
         get => Target;
         set
         {
             object old = Target;
-            Target = value;
-            _valueClassName = TypeDescriptor.GetClassName(Target);
-            OwnerGrid.ReplaceSelectedObject(old, value);
+            if (value is not null)
```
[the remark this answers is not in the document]
gpetrou: PropertyValue is nullable in other cases. I used AllowNull. Is this better?

src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGridInternal/ImmutablePropertyDescriptorGridEntry.cs
lonitra: I think we can still invert in this case since we will only go into the if statement after this one if `newObject is not null` , which will never be true if `owner = null`

src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGridInternal/GridEntry.cs
```diff
@@ -1218,7 +1221,7 @@ public int GetValueTextWidth(string text, Graphics graphics, Font font)
     ///  This internal override allows <see cref="CategoryGridEntry"/> to skip to its parent <see cref="PropertyValue"/>
     ///  and <see cref="MultiPropertyDescriptorGridEntry"/> to return it's set of owners.
     /// </devdoc>
-    internal virtual object GetValueOwnerInternal() => PropertyValue;
+    internal virtual object? GetValueOwnerInternal() => PropertyValue;
```
lonitra, line 1224: Now that we changed nullability of `PropertyValue`, this shouldn't be nullable either now and I believe this will be the same case for some other methods that were changed in this PR.

gpetrou: This still gives a warning if we make it non-nullable.

lonitra: Whoops, you are right. I was mistaken and thought we had changed the virtual `PropertyValue` but that is not the case.

lonitra (approved):
LGTM 🚀

PR #10879, opened 2024-02-14
Mimic Marshal Behavior on Clipboard when Built-in Com is Off

lonitra:
This PR fixes some issues found when running our test suite with `BuiltInComSupport = false` (mimic trimming).

When there is data on the clipboard that is from another process or when the clipboard has been flushed ([OleFlushClipboard](https://learn.microsoft.com/en-us/windows/win32/api/ole2/nf-ole2-oleflushclipboard) was called), we are no longer able to retrieve the real `IDataObject` pointer from the proxy that is returned by `OleGetClipboard` if we had placed something on the clipboard. In this case, when built in COM is turned on, marshal will simply return the proxy in a wrapper that implements `System.Runtime.InteropServices.ComTypes.IDataObject` via `Marshal.GetObjectForIUnknown`. When built in COM is turned off, we need to mimic this behavior ourselves.

Also added `IManagedWrapper` onto our `RichTextBox.OleCallback` as part of fixing failing tests when ` BuiltInComSupport = false`
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/10879)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/OLE/Clipboard.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/OLE/Clipboard.cs
@@ -100,9 +100,9 @@ public static class Clipboard
         }
 
         // OleGetClipboard always returns a proxy. The proxy forwards all IDataObject method calls to the real data object,
-        // without giving out the real data object. If the data placed on the clipboard is not one of our CCWs or the data
+        // without giving out the real data object. If the data placed on the clipboard is not one of our CCWs or the clipboard
         // has been flushed, marshal will create a wrapper around the proxy for us to use. However, if the data placed on
-        // the clipboard is one of our own and the data has not been flushed, we need to retrieve the real data object
+        // the clipboard is one of our own and the clipboard has not been flushed, we need to retrieve the real data object
         // pointer in order to retrieve the original managed object via ComWrappers. To do this, we must query for an
         // interface that is not known to the proxy e.g. IComCallableWrapper. If we are able to query for IComCallableWrapper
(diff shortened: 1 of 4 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/OLE/Clipboard.cs
```diff
@@ -118,12 +119,21 @@ public static unsafe void SetDataObject(object data, bool copy, int retryTimes,
...
             return null;
         }
 
+        if (managedDataObject is not IComDataObject dataObject)
+        {
+            // If we do not have a IComDataObject, built-in com support is turned off and
+            // we have a proxy where there is no way to retrieve the original data object
+            // pointer from it likely because either the clipboard was flushed or the data on the
+            // clipboard is from another process. We need to mimic built-in com behavior and wrap the proxy ourselves.
+            return new DataObject(proxyDataObject, managedDataObject);
```
JeremyKuhne, line 134: It looks like you haven't flipped the Interfaces on DataObject yet. We should never see the runtime interface coming back through.

lonitra: I was planning to make the flip as part of the refactoring we discussed, but I can flip it here if that is preferred.

JeremyKuhne: Just make sure to come back through and clear everything up.

JeremyKuhne (approved):

PR #10880, opened 2024-02-14
replace a null-forgiving operator with a null-check in TreeNodeAccessibleObject 

### LINKED ISSUE #10876: .NET Core 8.0 WinForms TreeView null reference exception 'owningTreeView'
### .NET version

.NET 8 Core

Yes

This code has been working since .NET Framework 3.5.
It's been migrated to Framework 4.x
Then .NET Core 3.1
And lastly .NET Core 6.0
It's now in .NET Core 8.0 that there is a problem.

### Issue description

I have code that has been migrated from .NET Framework over to .NET Core 3.1 then 6.0 and now I’m migrating to 8.0. It’s worked fine in all versions up to 8.0 where I’m now getting a null reference exception internal to the Core code.

Setting breakpoints will cause the error to become less likely to happen which leads me to suspect that this is some weird Windows messaging related issue where a slight change in the code execution time allows messages to process slightly different. I’m posting now in the hope that this is a known issue or with a known solution. I will be continuing to work on debugging further to narrow down what’s going on.

The exception:
System.ArgumentNullException
HResult=0x80004003
Message=Value cannot be null. (Parameter ‘owningTreeView’)
Source=System.Private.CoreLib
StackTrace:
at System.ArgumentNullException.Throw(String paramName) in /_/src/libraries/System.Private.CoreLib/src/System/ArgumentNullException.cs:line 87

Top call stack entries:
System.Private.CoreLib.dll! System.ArgumentNullException.Throw(string paramName) Line 87 C#
System.Windows.Forms.Primitives.dll! System.Windows.Forms.ArgumentValidation.OrThrowIfNull<System.__Canon>(System.__Canon argument, string paramName) Line 12 C#
System.Windows.Forms.dll! System.Windows.Forms.TreeNode.TreeNodeAccessibleObject.TreeNodeAccessibleObject(System.Windows.Forms.TreeNode owningTreeNode, System.Windows.Forms.TreeView owningTreeView) Line 19 C#
System.Windows.Forms.dll! System.Windows.Forms.TreeNode.AccessibilityObject.get() Line 1172 C#
System.Windows.Forms.dll! System.Windows.Forms.TreeView.OnAfterSelect(System.Windows.Forms.TreeViewEventArgs e) Line 2330 C#
System.Windows.Forms.dll! System.Windows.Forms.TreeView.TvnSelected(Interop.ComCtl32.NMTREEVIEW* nmtv) Line 2663 C#
System.Windows.Forms.dll! System.Windows.Forms.TreeView.WmNotify(ref System.Windows.Forms.Message m) Line 3112 C#
System.Windows.Forms.dll! System.Windows.Forms.TreeView.WndProc(ref System.Windows.Forms.Message m) Line 3343 C#
System.Windows.Forms.dll! System.Windows.Forms.NativeWindow.Callback(Windows.Win32.Foundation.HWND hWnd, Windows.Win32.MessageId msg, Windows.Win32.Foundation.WPARAM wparam, Windows.Win32.Foundation.LPARAM lparam) Line 372 C#

### Steps to reproduce

This only seems to happen when clicking on TreeView items that have sub items.  
I'm continuing to investigate.

Tanya-Solyanik:
Fixes #10876

## Proposed changes
TreeNodeAccessibleObject can't be created when the parent TreeView is not found. This could happen when a node is being added or deleted from the Nodes collection. If AccessibleObject is not created on this attempt, it will be created on the next access. If TreeView is not available, we skip object creation.

## Regression? 

- Yes  from NET7

## Risk

Low - adding a null-check

## Testing
Fix is based on the exception callstack, I couldn't reproduce this crash

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/10880)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeView.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeView.cs
@@ -10,6 +10,7 @@ using System.Windows.Forms.Layout;
 using System.Windows.Forms.VisualStyles;
 using Windows.Win32.System.Variant;
 using Windows.Win32.UI.Accessibility;
+using static System.Windows.Forms.TreeNode;
 
 namespace System.Windows.Forms;
 
@@ -1830,12 +1831,10 @@ public partial class TreeView : Control
     }
 
     /// <summary>
-    ///  Fires the DrawNode event.
+    ///  Raises the DrawNode event.
     /// </summary>
-    protected virtual void OnDrawNode(DrawTreeNodeEventArgs e)
-    {
+    protected virtual void OnDrawNode(DrawTreeNodeEventArgs e) =>
         _onDrawNode?.Invoke(this, e);
-    }
 
     protected override void OnHandleCreated(EventArgs e)
     {
@@ -2080,15 +2079,13 @@ public partial class TreeView : Control
     }
 
     /// <summary>
-    ///  Fires the beforeLabelEdit event.
+    ///  Raises the beforeLabelEdit event.
     /// </summary>
-    protected virtual void OnBeforeLabelEdit(NodeLabelEditEventArgs e)
-    {
+    protected virtual void OnBeforeLabelEdit(NodeLabelEditEventArgs e) =>
         _onBeforeLabelEdit?.Invoke(this, e);
-    }
 
     /// <summary>
-    ///  Fires the afterLabelEdit event.
+    ///  Raises the afterLabelEdit event.
     /// </summary>
     protected virtual void OnAfterLabelEdit(NodeLabelEditEventArgs e)
     {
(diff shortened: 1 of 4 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeView.cs
```diff
@@ -2217,7 +2211,12 @@ protected virtual void OnAfterSelect(TreeViewEventArgs e)
         // Raise an event to highlight & announce the selected node.
         if (IsAccessibilityObjectCreated)
         {
-            AccessibleObject nodeAccessibleObject = e.Node!.AccessibilityObject;
+            TreeNodeAccessibleObject? nodeAccessibleObject = e.Node!.AccessibilityObject;
```
lonitra, line 2214: Should `TreeViewEventArgs.Node` be made non nullable? We seem to have a `!` everytime it is used.

lonitra (approved):
LGTM, just a question on nullability of TreeViewArgs.Node -- but doesn't need to be addressed as part of this PR

src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeView.cs
Tanya-Solyanik: TreeViewEventArgs could be created during node insertion according to this comment-  https://github.com/dotnet/winforms/blob/1f2e5f1702a30bd334ad58b2f4d57e3bca342e95/src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeView.cs#L1823-#L1826

And we were raising events in the past(.NET Framework) in these cases. So probably a better approach is to handle e.Node == null gracefully in more cases.

lonitra (approved):

PR #10881, opened 2024-02-14
Make MsoComponent code trim friendly

JeremyKuhne:
This change converts the interfaces to match the same patterns we're using with other Com interfaces. Also yanks the trace statements for these.

I'm probably going to try and factor this stuff out into an opt-in feature.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/10881)

### DIFF
--- a/src/System.Private.Windows.Core/src/Windows/Win32/System/Com/WinFormsComWrappers.cs
+++ b/src/System.Private.Windows.Core/src/Windows/Win32/System/Com/WinFormsComWrappers.cs
@@ -71,15 +71,28 @@ internal unsafe partial class WinFormsComWrappers : ComWrappers
         }
     }
 
+    internal static TReturnType UnwrapAndInvoke<TThis, TInterface, TReturnType>(TThis* @this, Func<TInterface, TReturnType> func)
+        where TThis : unmanaged
+        where TInterface : class
+        where TReturnType : unmanaged
+    {
+        try
+        {
+            TInterface? @object = ComInterfaceDispatch.GetInstance<TInterface>((ComInterfaceDispatch*)@this);
+            return @object is null ? default : func(@object);
+        }
+        catch (Exception ex)
+        {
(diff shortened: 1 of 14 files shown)


src/System.Private.Windows.Core/src/Windows/Win32/System/Com/WinFormsComWrappers.cs
```diff
@@ -71,15 +71,28 @@ protected override void ReleaseObjects(IEnumerable objects)
         }
     }
 
+    internal static TReturnType UnwrapAndInvoke<TThis, TInterface, TReturnType>(TThis* @this, Func<TInterface, TReturnType> func)
```
lonitra, line 74: nit: docs

src/System.Windows.Forms/src/System/Windows/Forms/Application.ComponentManager.cs
```diff
@@ -21,23 +20,23 @@ public sealed partial class Application
     ///   This class is not used when running inside the Visual Studio shell.
     ///  </para>
     /// </remarks>
-    private unsafe class ComponentManager : IMsoComponentManager
+    private unsafe class ComponentManager : IMsoComponentManager.Interface
```
lonitra, line 23: Do we need `IManagedWrapper` on this?

JeremyKuhne: No, we never expose it out. We just stub it in if we don't have one to attach to.

src/System.Private.Windows.Core/src/Windows/Win32/System/Com/WinFormsComWrappers.cs
JeremyKuhne: I'll catch it in my next PR.

lonitra (approved):

PR #10893, opened 2024-02-15
Make IMsoComponent support opt-in

JeremyKuhne:
There is quite a bit of overhead on message handling and main thread set-up. Additionally it is non-trivial to follow what is going on when we process Windows messages. This PR creates two `ThreadContext` classes, one that supports `IMsoComponentManager` and one that does not.

Starting in .NET 9, you'll need to set the "Switch.System.Windows.Forms.EnableMsoComponentManager" switch to get integration with process registered `IMsoComponentManager`s.

Reviewing this should be done by looking at the individual commits.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/10893)


src/System.Windows.Forms/src/System/Windows/Forms/Application.ComponentThreadContext.cs
```diff
@@ -0,0 +1,489 @@
...
+                    if (_componentManager is not null)
+                    {
+                        _externalComponentManager = true;
+                    }
+                    else
+                    {
+                        _componentManager = new ComponentManager();
+                    }
+
+                    if (_componentManager is not null)
```
lonitra, line 78: I think this if statement can be removed

src/System.Windows.Forms/src/System/Windows/Forms/Application.ComponentThreadContext.cs
```diff
@@ -0,0 +1,489 @@
...
+
+        public override void EnsureReadyForIdle() =>
+            // Ensure the component manager is created.
+            _ = ComponentManager;
+
+        /// <summary>
+        ///  Retrieves the component manager for this process.  If there is no component manager
+        ///  currently installed, we install our own.
+        /// </summary>
+        internal unsafe IMsoComponentManager.Interface? ComponentManager
```
lonitra, line 46:
```suggestion
        internal IMsoComponentManager.Interface? ComponentManager
```

src/System.Windows.Forms/src/System/Windows/Forms/Application.ComponentThreadContext.cs
```diff
@@ -0,0 +1,489 @@
...
+                // Reset the flag since we are exiting out of a ModalMessageLoop.
+                _ourModalLoop = wasOurLoop;
+            }
+        }
+
+        internal override void FormActivated(bool activate)
+        {
+            if (activate)
+            {
+                if (ComponentManager is { } manager && manager is not Application.ComponentManager)
```
lonitra, line 214: This can be combined. Same with method below as well

src/System.Windows.Forms/src/System/Windows/Forms/Application.ComponentThreadContext.cs
```diff
@@ -0,0 +1,489 @@
...
+
+        void IMsoComponent.Interface.OnAppActivate(BOOL fActive, uint dwOtherThreadID)
+        {
+        }
+
+        void IMsoComponent.Interface.OnLoseActivation()
+        {
+        }
+
+        unsafe void IMsoComponent.Interface.OnActivationChange(
```
lonitra, line 401:
```suggestion
        void IMsoComponent.Interface.OnActivationChange(
```

lonitra (commented):
Changes look good, just some small comments

JeremyKuhne:
@lonitra I'll catch your comments in a follow up PR.

lonitra (approved):

lonitra:
https://github.com/dotnet/docs/issues/42195

PR #10905, opened 2024-02-19
Fix `null` values in `AutoCompleteCustomSource` 

### LINKED ISSUE #10904: Adding Nothing to TextBox.AutoCompleteCustomSource stops app without any info
### .NET version

.NET 9.0.0-preview.2.24079.1

No, `The program '[23732] FrameworkApp.exe' has exited with code 3221225477 (0xc0000005) 'Access violation'.`

Yes, this is not a regression issue, can reproduce from 3.1 ~ 9.0.

### Issue description

The app doesn’t respond for some seconds and will end after input the text for the textBox1. However there is no error message, exception or such a thing. And the user expect an exception message, like says: Adding null/Nothing to the AutoCompleteCustomSource is invalid.

### Steps to reproduce

1. Open attached the reproduce application, then build and run
[FeedBack1949667.zip](https://github.com/dotnet/winforms/files/14320889/FeedBack1949667.zip)
2. Click the button1
3. Entering the textbox and typing any letter

**Actual Results:**
 The app doesn’t respond for some seconds and will end after input the text for the textBox1. However there is no error message, exception or such a thing for user.
FeedBack1949667

**Expected Results:**
Expect an exception message, that says: Adding null/Nothing to the AutoCompleteCustomSource is invalid.

elachlan:
- Refactoring to add `ArgumentNullException`/`InvalidOperationException` to `AutoCompleteCustomSource`
- add tests to verify
- Cleanup/Refactor of `StringSource`

I think this is an API change, so it might need to be documented. Since the APIs now throw exceptions.

#Fixes #10904
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/10905)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/AutoCompleteStringCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/AutoCompleteStringCollection.cs
@@ -59,7 +59,11 @@ public class AutoCompleteStringCollection : IList
     /// </summary>
     public int Add(string value)
     {
-        ArgumentNullException.ThrowIfNull(value);
+        if (value is null)
+        {
+            return -1;
+        }
+
         int index = ((IList)_data).Add(value);
         OnCollectionChanged(new CollectionChangeEventArgs(CollectionChangeAction.Add, value));
         return index;
@@ -72,15 +76,16 @@ public class AutoCompleteStringCollection : IList
     {
         ArgumentNullException.ThrowIfNull(value);
 
+        List<string> nonNullItems = [];
         foreach (string item in value)
         {
-            if (item is null)
+            if (item is not null)
             {
-                throw new InvalidOperationException(SR.InvalidNullItemInCollection);
+                nonNullItems.Add(item);
             }
         }
 
-        _data.AddRange(value);
+        _data.AddRange(nonNullItems);
         OnCollectionChanged(new CollectionChangeEventArgs(CollectionChangeAction.Refresh, null));
     }
 
@@ -121,9 +126,12 @@ public class AutoCompleteStringCollection : IList
     {
         ArgumentOutOfRangeException.ThrowIfNegative(index);
         ArgumentOutOfRangeException.ThrowIfGreaterThan(index, _data.Count);
-        ArgumentNullException.ThrowIfNull(value);
-        _data.Insert(index, value);
-        OnCollectionChanged(new CollectionChangeEventArgs(CollectionChangeAction.Add, value));
+
+        if (value is not null)
+        {
+            _data.Insert(index, value);
+            OnCollectionChanged(new CollectionChangeEventArgs(CollectionChangeAction.Add, value));
+        }
     }
 
     /// <summary>
--- a/src/System.Windows.Forms/src/System/Windows/Forms/AutoCompleteStringCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/AutoCompleteStringCollection.cs
@@ -12,7 +12,7 @@ namespace System.Windows.Forms;
 public class AutoCompleteStringCollection : IList
 {
     private CollectionChangeEventHandler? _onCollectionChanged;
-    private readonly List<string> _data = new();
+    private readonly List<string> _data = [];
 
     public AutoCompleteStringCollection()
     {
@@ -59,6 +59,7 @@ public class AutoCompleteStringCollection : IList
     /// </summary>
     public int Add(string value)
     {
+        ArgumentNullException.ThrowIfNull(value);
         int index = ((IList)_data).Add(value);
         OnCollectionChanged(new CollectionChangeEventArgs(CollectionChangeAction.Add, value));
         return index;
@@ -70,6 +71,14 @@ public class AutoCompleteStringCollection : IList
     public void AddRange(params string[] value)
     {
         ArgumentNullException.ThrowIfNull(value);
+
+        foreach (string item in value)
+        {
+            if (item is null)
+            {
+                throw new InvalidOperationException(SR.InvalidNullItemInCollection);
+            }
+        }
 
         _data.AddRange(value);
         OnCollectionChanged(new CollectionChangeEventArgs(CollectionChangeAction.Refresh, null));
@@ -110,6 +119,7 @@ public class AutoCompleteStringCollection : IList
     /// </summary>
     public void Insert(int index, string value)
     {
+        ArgumentNullException.ThrowIfNull(value);
         _data.Insert(index, value);
         OnCollectionChanged(new CollectionChangeEventArgs(CollectionChangeAction.Add, value));
     }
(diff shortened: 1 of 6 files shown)


elachlan:
CC: @Zheng-Li01

KlausLoeffelmann (commented):
Looks good to me, but I want either Jeremy or Loni to have a look.
Also, do you think the additional Range-checks could be a breaking change, or are the exceptions only "brought forward in time", so to speak? (I'd assume the latter, but rather ask.)

Thanks for the PR, I really appreciate, especially, since someone from the German community reported this issue, and it was rejected at first, and that guy was (rightfully so!) somewhat disappointed, but now it has been addressed so quickly! 

This is so nice and helpful - I am happy!

elachlan:
@KlausLoeffelmann you are correct. It was to bring the checks forward. They are also done in the insert call to the underlying list. But it doesn't do a null check.

src/System.Windows.Forms/src/System/Windows/Forms/AutoCompleteStringCollection.cs
```diff
@@ -59,6 +59,7 @@ public AutoCompleteStringCollection()
     /// </summary>
     public int Add(string value)
     {
+        ArgumentNullException.ThrowIfNull(value);
```
Tanya-Solyanik, line 62: We could also do nothing in case of null here and skip null values in `AddRange` method.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AutoCompleteStringCollectionTests.cs
```diff
@@ -36,6 +36,20 @@ public void AutoCompleteStringCollection_AddRange_NullValue_ThrowsArgumentNullEx
     }
 
     [WinFormsFact]
+    public void AutoCompleteStringCollection_AddRange_NullValues_ThrowsInvalidOperationException()
+    {
+        AutoCompleteStringCollection collection = new();
+        Assert.Throws<InvalidOperationException>(() => collection.AddRange([null]));
```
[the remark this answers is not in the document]
elachlan: That would show a warning on the null values and break the build wouldn't it?

src/System.Windows.Forms/src/System/Windows/Forms/AutoCompleteStringCollection.cs
elachlan: I think it would be inconsistent with the `add` and `insert`.

src/System.Windows.Forms/src/System/Windows/Forms/StringSource.cs
```diff
@@ -13,21 +13,18 @@ namespace System.Windows.Forms;
...
-    private int size;
+    private string[] _strings;
+    private int _current;
+    private int _size;
     private IAutoComplete2* _autoComplete2;
 
     /// <summary>
     ///  Constructor.
     /// </summary>
     public StringSource(string[] strings)
```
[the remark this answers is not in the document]
elachlan: I think because the class is internal, we should keep the null-ability the same and make sure we aren't passing in null in the consumers. We already do this, we just pass an empty string collection to `StringSource` when `AutoCompleteStringCollection` is empty.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AutoCompleteStringCollectionTests.cs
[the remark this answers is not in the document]
Tanya-Solyanik: Well, you would add null forgiving operators too.

src/System.Windows.Forms/src/System/Windows/Forms/AutoCompleteStringCollection.cs
Tanya-Solyanik: Let me re-phrase: Add, Insert and AddRange would handle individual strings being null silently. Add and Insert would exit if value is null, and AddRange would throw if the array is null, but ignore null values in that array.

elachlan: Therefore no API change, I get it. Ill make the change.

Tanya-Solyanik: My reasoning was, if we have a way to handle a null gracefully, we should do it. In this case, the developer who enables nullability in the app, will have a hint as to what is going wrong, but the existing application would just work with potentially some missing autocomplete values.

src/System.Windows.Forms/src/System/Windows/Forms/AutoCompleteStringCollection.cs
```diff
@@ -59,6 +59,11 @@ public AutoCompleteStringCollection()
     /// </summary>
     public int Add(string value)
     {
+        if (value is null)
+        {
+            return -1;
```
Tanya-Solyanik, line 64: We might have to add this to the docs

src/System.Windows.Forms/src/System/Windows/Forms/AutoCompleteStringCollection.cs
```diff
@@ -71,7 +76,16 @@ public void AddRange(params string[] value)
...
+        List<string> nonNullItems = [];
+        foreach (string item in value)
+        {
+            if (item is not null)
+            {
+                nonNullItems.Add(item);
+            }
+        }
+
+        _data.AddRange(nonNullItems);
```
Tanya-Solyanik, line 88: Make sure that the nonNullItems list is not empty before adding because we want to skip the Changed event

src/System.Windows.Forms/src/System/Windows/Forms/AutoCompleteStringCollection.cs
elachlan: Added to xml comment.

src/System.Windows.Forms/src/System/Windows/Forms/AutoCompleteStringCollection.cs
elachlan: Done.

src/System.Windows.Forms/src/System/Windows/Forms/Controls/ComboBox/ComboBox.cs
```diff
@@ -3326,15 +3316,15 @@ private void SetAutoComplete(bool reset, bool recreate)
 
             if (_stringSource is null)
             {
-                _stringSource = new StringSource(GetStringsForAutoComplete(Items));
```
Tanya-Solyanik, line 3329: The call to `Items` was previously allocating this collection. This change does not have an effect on the output of GetStringsForAutoComplete method, but might cause side effects elsewhere, for example in the SelectedItem setter. I think GetStringsForAutoComplete should read the proeprty rather than the backing field

elachlan: Done!

lonitra (approved):
LGTM

Issue #5391, opened 2021-08-04
Menus and *Strips and SplitContainer controls do not respect application font


RussKie:
As reported in https://github.com/dotnet/winforms-designer/issues/3611

**Problem description:**

The font for these controls: ToolStrip/menuStrip/StatusStrip, MonthCalendar, SplitContainer is not rendered after setting an application font, is not same as button's
![image](https://user-images.githubusercontent.com/26474449/126601047-51e49925-e4f8-4b89-b4be-6a7df4ee8b65.png)

<details>
<summary>.NET 6.0 RC1</summary>

* Runtime
![image](https://user-images.githubusercontent.com/4403806/126939007-93cba08c-343a-4d67-9dcc-82c58da8d05a.png)
* Designer
![image](https://user-images.githubusercontent.com/4403806/126939122-ddabf7e8-da7a-4b04-80e1-3a544d0ef8f0.png)

</details>

<details>
<summary>.NET 5.0</summary>

* Runtime
![image](https://user-images.githubusercontent.com/4403806/126946460-15191dd4-24ed-4876-bac7-2bdc8d12fb13.png)

* Designer:
![image](https://user-images.githubusercontent.com/4403806/126946426-2aad4a32-5a4c-4cc6-8efe-f3d7efc7364d.png)

</details>

**Expected behavior:**

The font for these controls: ToolStrip/menuStrip/StatusStrip, MonthCalendar, SplitContainer is rendered as that application font

**Repro steps:**

1. Create a Winforms .Net application
2. Set a custom application font in .csproj file as following:
```xml
          <Project Sdk="Microsoft.NET.Sdk">
            <PropertyGroup>
              <OutputType>WinExe</OutputType>
              <TargetFramework>net6.0-windows</TargetFramework>
              <Nullable>enable</Nullable>
              <UseWindowsForms>true</UseWindowsForms>
	          <ApplicationFont>Arial, 12pt, style=Bold, Italic</ApplicationFont>
            </PropertyGroup>
          </Project>
```
3. Add some controls: Button, ToolStrip/menuStrip/StatusStrip, MonthCalendar/SplitContainer to form designer, then to observe

A test solution is also available at https://github.com/RussKie/DesignerTestbed

dreddy-work:
Can we repro this by just setting Font on the Form and no explicit Default font?

dreddy-work:
@Olina-Zhang , Can you verify if any other winforms controls, other than what is specified here,  experiencing same issue?

[...]

PR #10906, opened 2024-02-19
Add AppContext switch "System.Windows.Forms.ApplyParentFontToMenus"

### LINKED ISSUE #10774: Korean localization font issue
### Description

In 200 dpi mode, I set default font to old font that framework used, `Application.SetDefaultFont(new Font(new FontFamily("Microsoft Sans Serif"), 8.25F, FontStyle.Regular, GraphicsUnit.Point, 0))`, the font of toolstrip and datagridview is different in Korean. If set default font to "Segoe UI", the font of toolstrip and datagridview is same in Korean.
Sample code: [TestKoreanWinform.zip](https://github.com/dotnet/runtime/files/14108547/TestKoreanWinform.zip)

BTW, "Microsoft Sans Serif" seems ok in 100 dpi mode.
image
image

### Reproduction Steps

In 200 dpi mode, build TestKoreanWinform project and run TestKoreanWinform.exe.

### Expected behavior

The font of toolstrip and datagridview is same in Korean.

### Actual behavior

The font of toolstrip and datagridview is different in Korean.

### Regression?

_No response_

### Known Workarounds

set ToolStrip.Font to the desired font for each toolstrip (or define a base class for all your toolstrips that sets the desired font).

### Configuration

_No response_

### Other information

_No response_

LeafShi1:
Fixes #10774 

## Proposed changes

- Add a new AppContext switch "System.Windows.Forms.ApplyParentFontToMenus"
- Add a condition in ToolStrip.cs to return the font of the form when LocalAppContextSwitches.ApplyParentFontToMenus is true.

## Customer Impact

- The ToolStrip control font can be consistent with the Form by setting the property "`System.Windows.Forms.ApplyParentFontToMenus: true`" in the configuration file

## Regression? 

- No

## Risk

- Minimal

### Before

In 200 dpi mode, using `Application.SetDefaultFont(new Font(new FontFamily("Microsoft Sans Serif"), 8.25F, FontStyle.Regular, GraphicsUnit.Point, 0))`, the font of toolstrip and datagridview is different in Korean.

image

### After
In 200 dpi mode, the fonts of toolstrip and datagridview are the same regardless of the font used by the Form.

image

## Test methodology 

- Manually

## Test environment(s) 

- .net 9.0.0-preview.2.24105.1

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/10906)


ricardobossan (commented):
Maybe add a test for switches on/off, like [this](https://github.com/dotnet/winforms/blob/852f7ef241402c2336e085c82dca6f70b7400c21/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjects/DataGridViewRowAccessibleObjectTests.cs#L2387-L2402) for `DataGridViewUIAStartRowCountAtZeroSwitchName`, but adapting it to this PR's switch:

```cs

    // Unit test for https://github.com/dotnet/winforms/issues/7154
    [WinFormsFact]
    public void DataGridView_SwitchConfigured_AdjustsRowStartIndices()
    {
        LocalAppContextSwitches.SetLocalAppContextSwitchValue(LocalAppContextSwitches.DataGridViewUIAStartRowCountAtZeroSwitchName, true); 

        using DataGridView dataGridView = new();
        dataGridView.Columns.Add(new DataGridViewTextBoxColumn());
        dataGridView.Rows.Add(new DataGridViewRow());

        Assert.Equal(string.Format(SR.DataGridView_AccRowName, 0), dataGridView.Rows[0].AccessibilityObject.Name);

        LocalAppContextSwitches.SetLocalAppContextSwitchValue(LocalAppContextSwitches.DataGridViewUIAStartRowCountAtZeroSwitchName, false);

        Assert.Equal(string.Format(SR.DataGridView_AccRowName, 1), dataGridView.Rows[0].AccessibilityObject.Name);
    }
```

All the changes already made LGTM.

src/System.Windows.Forms/src/System/Windows/Forms/Controls/ToolStrips/ToolStrip.cs
```diff
@@ -550,6 +551,11 @@ public override Font Font
             // thread local storage for it.
             _defaultFont ??= ToolStripManager.DefaultFont;
 
+            if (_defaultFont != base.Font && LocalAppContextSwitches.ApplyApplicationFontToMenus is true)
```
Tanya-Solyanik, line 554: Is it possible to extend unit tests for Application font to test menu font as well?

src/System.Windows.Forms/src/System/Windows/Forms/Controls/ToolStrips/ToolStrip.cs
```diff
@@ -550,6 +551,11 @@ public override Font Font
...
+            {
+                return base.Font;
```
Tanya-Solyanik, line 556: Had you tested https://github.com/dotnet/winforms/issues/5391 too?

src/System.Windows.Forms/src/System/Windows/Forms/Controls/ToolStrips/ToolStrip.cs
```diff
@@ -550,6 +551,11 @@ public override Font Font
             // thread local storage for it.
             _defaultFont ??= ToolStripManager.DefaultFont;
```
Tanya-Solyanik, line 552: I would short-circuit ToolStripManager.DefaultFont by testing for ApplyApplicationFontToMenus  switch there.

Tanya-Solyanik (commented):
Looks good!

src/System.Windows.Forms/src/System/Windows/Forms/Controls/ToolStrips/ToolStrip.cs
Tanya-Solyanik: I'm not sure what is going on, maybe try showing the form and expanding menus to parent controls properly?

src/System.Windows.Forms/src/System/Windows/Forms/Controls/ToolStrips/ToolStrip.cs
LeafShi1: Yes, the controls _ToolStrip/menuStrip/StatusStrip, SplitContainer_ all work well after setting "`System.Windows.Forms.ApplyApplicationFontToMenus: true`"
image

src/System.Windows.Forms/src/System/Windows/Forms/Controls/ToolStrips/ToolStrip.cs
LeafShi1: Sorry, I forgot to add judgment in function `SetLocalAppContextSwitchValue`

src/System.Windows.Forms/src/System/Windows/Forms/Controls/ToolStrips/ToolStrip.cs
LeafShi1: > I would short-circuit ToolStripManager.DefaultFont by testing for ApplyApplicationFontToMenus switch there.

I added a test, please review it.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripTests.cs
```diff
@@ -1618,6 +1619,31 @@ public void ToolStrip_Font_SetWithItemsWithFontWithHandler_CallsFontChanged()
...
+        toolStrip1.Items.Add(item2);
+        form.Controls.Add(toolStrip1);
+        form.Font = font;
+
+        Assert.Same(form.Font, toolStrip1.Font);
+        Assert.Same(form.Font, item1.Font);
+        Assert.Same(form.Font, item2.Font);
+
+        LocalAppContextSwitches.SetLocalAppContextSwitchValue(LocalAppContextSwitches.ApplyApplicationFontToMenusSwitchName, false);
+        Assert.Same(ToolStripManager.DefaultFont, item2.Font);
```
Tanya-Solyanik, line 1642: Add a blank line after line 1641

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripTests.cs
```diff
@@ -1618,6 +1619,32 @@ public void ToolStrip_Font_SetWithItemsWithFontWithHandler_CallsFontChanged()
         Assert.Equal(3, callCount);
     }
 
+    [WinFormsFact]
+    public void ToolStrip_Font_ApplyApplicationFontToMenus_GetReturnsExpected()
+    {
+        LocalAppContextSwitches.SetLocalAppContextSwitchValue(LocalAppContextSwitches.ApplyApplicationFontToMenusSwitchName, true);
```
lonitra, line 1625: It might be good to place this in a try/catch. If one of the asserts fail or we get an exception somewhere, we will not set the switch back to original value, which could inadvertently affect other tests.
e.g.

```c#
LocalAppContextSwitches.SetLocalAppContextSwitchValue(LocalAppContextSwitches.ApplyApplicationFontToMenusSwitchName, true);
try
{
    // tests....
}
finally
{
LocalAppContextSwitches.SetLocalAppContextSwitchValue(LocalAppContextSwitches.ApplyApplicationFontToMenusSwitchName, false);
}
```

src/System.Windows.Forms/src/System/Windows/Forms/Controls/ToolStrips/ToolStrip.cs
lonitra, line 554:
```suggestion
            if (_defaultFont != base.Font && LocalAppContextSwitches.ApplyApplicationFontToMenus)
```

Also, consider switching to check the switch value first. If there is work being done in `base.Font` that we do not need unless the switch is `true` then we are doing extra work to no benefit.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -140,6 +142,12 @@ public static bool AnchorLayoutV2
         get => GetCachedSwitchValue(AnchorLayoutV2SwitchName, ref s_anchorLayoutV2);
     }
 
+    public static bool ApplyApplicationFontToMenus
```
Tanya-Solyanik, line 145: Please add detailed description in `/// <summary> ` comments to this property

src/System.Windows.Forms/src/System/Windows/Forms/Controls/ToolStrips/ToolStrip.cs
Tanya-Solyanik: I meant to add a test for ApplyApplicationFontToMenus to https://github.com/dotnet/winforms/blob/dd5ae276be5f5e8c2203fd5f07bdc725a9891945/src/System.Windows.Forms/src/System/Windows/Forms/Controls/ToolStrips/ToolStripManager.cs#L50

LeafShi1: Only adding the judgment at `DefaultFont ` in `ToolStripManager.cs`, when the default value of Application is set, the font of ToolStrip will change according to the font of App.

But if the App's default value remains "Segoe UI", and then sets the Form's font to "Microsoft Sans Serif", the ToolStrip's font will not change according to the Form's font. To change according to the font of the Form, this judgment needs to be added in ToolStrip.cs. 

The test results are as follows after deleting the judgment here, 

Test1 will **pass**:
```
LocalAppContextSwitches.SetLocalAppContextSwitchValue(LocalAppContextSwitches.ApplyApplicationFontToMenusSwitchName, true);
using Font font1 = new("Arial", 8.25f);
Application.SetDefaultFont(SystemFonts.CaptionFont);
using ToolStrip toolStrip = new();
Assert.Equal(font1, toolStrip.Font);
```


Test 2 will **fail**:
```
LocalAppContextSwitches.SetLocalAppContextSwitchValue(LocalAppContextSwitches.ApplyApplicationFontToMenusSwitchName, true);
using Font font = new("Arial", 8.25f);
using Form form = new();
using ToolStrip toolStrip1 = new();
form.Controls.Add(toolStrip1);
form.Font = font;
Assert.Equal(form.Font, toolStrip1.Font);
```

So if this switch only works for the DefaultFont of Application, I think the judgment here is redundant.

Tanya-Solyanik: I see, it was not my intention to allow Form font to propagate to the menus. But this makes sense. Then we should rename the app compat switch. How about `ApplyParentFontToMenus`?

src/System.Windows.Forms/src/System/Windows/Forms/Controls/ToolStrips/ToolStrip.cs
```diff
@@ -540,7 +541,7 @@ public override Font Font
     {
         get
         {
-            if (IsFontSet())
+            if (IsFontSet() || LocalAppContextSwitches.ApplyApplicationFontToMenus)
```
Tanya-Solyanik, line 544:
```suggestion
            if (LocalAppContextSwitches.ApplyApplicationFontToMenus || IsFontSet())
```

src/System.Windows.Forms/src/System/Windows/Forms/Controls/ToolStrips/ToolStrip.cs
LeafShi1: I think the name is good.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripTests.cs
```diff
@@ -1618,6 +1619,77 @@ public void ToolStrip_Font_SetWithItemsWithFontWithHandler_CallsFontChanged()
...
+        try
+        {
+            toolStrip1.Items.Add(item1);
+            toolStrip1.Items.Add(item2);
+            form.Controls.Add(toolStrip1);
+            form.Shown += (s, e) => { form.Close(); };
+            form.ShowDialog();
+
+            Assert.True(LocalAppContextSwitches.ApplyParentFontToMenus);
+            Assert.Equal(font, toolStrip1.Font);
```
LeafShi1, line 1649: It's strange that this assertion passes when running on the local machine, but it always fails in PR run. @Tanya-Solyanik Do you have any suggestions?

LeafShi1: Remove this test from here, I will submit a separate PR for this test

Tanya-Solyanik (approved):
Looks good, but please create either a follow up issue or a new PR to add an ApplicationFont  test.

lonitra:
@LeafShi1 This will need to be documented so that users are aware of this new switch. With your most recent PR it looks like that this switch is meant to be by default off so there would be no impact to existing behavior unless users intentionally turn this switch on so this would be considered a new article request. Could you file a new docs issue to get this documented: https://github.com/dotnet/docs/issues

PR #10910, opened 2024-02-19
[release/8.0] WinForms TreeView null reference exception 'owningTreeView'

### LINKED ISSUE #10876: .NET Core 8.0 WinForms TreeView null reference exception 'owningTreeView'

Tanya-Solyanik:
Fixes #10876
Port of https://github.com/dotnet/winforms/pull/10880
(PR was created by cherry-picking commits one-by-one)

## Proposed changes
TreeNodeAccessibleObject can't be created when the parent TreeView is not found. This could happen when a node is being added or deleted from the Nodes collection. If AccessibleObject is not created on this attempt, it will be created on the next access. If TreeView is not available, we skip object creation. This fix was implemented based on the exception call stack from the customer.

## Customer Impact
Customer observes random crashes in their application when dragging TreeView items and clicking on TreeView subtrees with the mouse.
No workaround is available, 

## Regression? 
- Yes from NET7, introduced while enabling accessibility events to TreeNode operations.

## Risk
 - Low, we are exiting if accessible object can't be created. AccessibleObject will be created after the node is parented properly.

## Test 
- added unit tests
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/10910)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TreeNode.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TreeNode.cs
@@ -1169,7 +1169,7 @@ public partial class TreeNode : MarshalByRefObject, ICloneable, ISerializable
     }
 
     internal TreeNodeAccessibleObject AccessibilityObject
-        => _accessibleObject ??= new TreeNodeAccessibleObject(this, TreeView);
+        => _accessibleObject ??= TreeView is null ? null : new TreeNodeAccessibleObject(this, TreeView);
 
     /// <summary>
     ///  Adds a new child node at the appropriate sorted position
--- a/src/System.Windows.Forms/src/System/Windows/Forms/TreeView.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/TreeView.cs
@@ -2208,7 +2208,7 @@ public partial class TreeView : Control
         // if editing hasn't been canceled.
         if (IsAccessibilityObjectCreated && !e.CancelEdit)
         {
-            e.Node.AccessibilityObject.RaiseAutomationEvent(UiaCore.UIA.AutomationFocusChangedEventId);
+            e.Node.AccessibilityObject?.RaiseAutomationEvent(UiaCore.UIA.AutomationFocusChangedEventId);
         }
     }
 
@@ -2230,7 +2230,12 @@ public partial class TreeView : Control
         // Raise an event to announce a toggle state change.
         if (IsAccessibilityObjectCreated)
         {
-            AccessibleObject nodeAccessibleObject = e.Node.AccessibilityObject;
+            TreeNode.TreeNodeAccessibleObject nodeAccessibleObject = e.Node.AccessibilityObject;
+            if (nodeAccessibleObject is null)
+            {
+                return;
+            }
+
             UiaCore.ToggleState newState = nodeAccessibleObject.ToggleState;
             UiaCore.ToggleState oldState = newState == UiaCore.ToggleState.On
                 ? UiaCore.ToggleState.Off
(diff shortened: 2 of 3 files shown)


Tanya-Solyanik:
@Olina-Zhang  - could you please test this change for regression. We don't have a repro app from the customer.

Olina-Zhang:
> @Olina-Zhang - could you please test this change for regression. We don't have a repro app from the customer.

Tested this PR based on .Net 8.0 SDK for treeView with add/delete/edit treeNode, drag and drop treeNode, and Accessibility testing using Tools, no new issue found. In addition, ran migrated automation cases with no new issue found.

JeremyKuhne (approved):

RussKie:
FWIW, we've been observing the similar issue with at the LV groups level.

Issue #10902, opened 2024-02-16
Improve data collection for unit tests


Tanya-Solyanik:
@sharwell had added screenshots recording for failed integration tests - https://github.com/dotnet/winforms/pull/10552 Add Window messages data to the test failure information.
However for some tests we could use additional information to identify causes of Assertion failures. We could start by logging windows messages:
 
 - [ ] Register a custom logger that would log all messages sent to the main test window and all its children with timestamps. Use DataCollectionService.RegisterCustomLogger to register the new logger. [winforms/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/DataCollectionService.cs at f8e6d3b434f6e58bdc5ec1900586071d17650cc8 · dotnet/winforms (github.com)](https://github.com/dotnet/winforms/blob/f8e6d3b434f6e58bdc5ec1900586071d17650cc8/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/DataCollectionService.cs#L13)
The custom log will be saved to the same screenshots folder. Our tests are very quick, we can start by logging all windows messages, logs will not be too big. We can monitor messages using a hook - [SetWindowsHookExA function (winuser.h) - Win32 apps | Microsoft Learn](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setwindowshookexa)
Roslyn repo implements custom loggers, can be used as a reference.

- [ ] once we start recording messages, we could log controls state (visible, enabled, etc) for the visual tree.
- [ ] Sam's suggestion: record state of the keyboard (to validate a theory that some tests are not restoring keyboard state properly)
- [ ] Sam's suggestion: _It might be good to have a CI variable that basically saves the APNG data for every test, including passing ones. This variable is always false by default, but at any time someone could queue up a run to effectively gather baseline data that can be used to cross-reference failures against what is seen on success._

JeremyKuhne:
Where did this come from? Why would we log every message in the tests? "logs will not be too big" - what does "too big" mean?

We should be discussing, as a team, what we're trying to do here before we add complexity and overhead.

Tanya-Solyanik:
Screen shots that were added previously are not sufficient for our tests to understand the failure, screen shots only show that no unexpected windows are covering up our UI, but it's still not clear why test actions were not taken. A message log would provide more details on what happened in the test. Moving to untriaged to discuss in the meeting.

merriemcgaw:
[...]

PR #10916, opened 2024-02-20
Log windows messages in tests

sharwell:
Related to #10902


sharwell:
I can't get this to fail. I tried.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/ControlTestBase.cs
```diff
@@ -293,6 +293,7 @@ protected async Task RunSingleControlTestAsync<T>(Func<Form, T, Task> testDriver
 
     protected async Task RunFormAsync<T>(Func<(Form dialog, T control)> createDialog, Func<Form, T, Task> testDriverAsync)
     {
+        using var messageRecordService = new WindowsMessageRecordService();
         using var screenRecordService = new ScreenRecordService();
```
sharwell, line 297: I don't like having these build up. I plan to refactor it in a follow-up to have a single data collection service that delegates to items within it.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/WindowsMessageRecordService.cs
```diff
@@ -0,0 +1,223 @@
...
+                WINDOWS_HOOK_ID.WH_MSGFILTER,
+                &MessageProc,
+                HINSTANCE.Null,
+                PInvoke.GetCurrentThreadId());
+
+            s_getMsgProcHook = PInvoke.SetWindowsHookEx(
+                WINDOWS_HOOK_ID.WH_GETMESSAGE,
+                &GetMsgProc,
+                HINSTANCE.Null,
+                PInvoke.GetCurrentThreadId());
```
sharwell, line 155: I couldn't prove that any one of these was a superset of the others, so for now I collect all three.

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Infra/ControlTestBase.cs
Tanya-Solyanik: I don't mind having these very explicitly visible in the base class. 
Could you please add some way to enable message logging on specific tests only?

lonitra:
Closing as there has been no recent activity for some time now. Please feel free to reopen the PR.

PR #10917, opened 2024-02-20
Clean up code around the painting loop

JeremyKuhne:
Various cleanup with painting.

- Add a SuspendLayoutScope to simplify scoping and validation
- Don't update layout for the parent if you don't change size (SetBoundsCore)
- Fix remaining info messages in Control
- Don't paint transparent background twice in PaintBackground
- Cache checking image transparency in PaintBackground
- Suppress all relevant warnings in NotSupported.cs to avoid the 10000 warnings that pop when you accidentally open it

Stops caching high contrast setting. By caching and starting the SystemEvents window this was causing a nested paint on the main form during startup if you you had a background image. The call is not so expensive to need creating a window look for changes.

This also moves a few bitmap related APIs to Core to allow for Direct2D related scenarios.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/10917)


lonitra (approved):
👍

Issue #10871, opened 2024-02-13
Suppress CodeQL Warnings in Tests


lonitra:
We should suppress these warnings to reduce noise. These warnings do not need formal exceptions as these are tests.

CodeQL warnings: 
- [x] https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1666874/?view=edit
- [x] https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1775935/?view=edit
- [x] https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1775936/?view=edit
- [x] https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1779535/?view=edit
- [x] https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1779536/?view=edit
- [x] https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1823177/?view=edit
- [x] https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1823178/?view=edit
- [x] https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1823179/?view=edit
- [x] https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1823180/?view=edit
- [x] https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1845202/?view=edit
- [x] https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1845203/?view=edit
- [x] https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1845204/?view=edit
- [x] https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1845205/?view=edit
- [x] https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1845206/?view=edit
- [x] https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1845207/?view=edit
- [x] https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1845208/?view=edit
- [x] https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1845220/?view=edit
- [x] https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1845221/?view=edit
- [x] https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1951751/?view=edit
- [x] https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1951752/?view=edit
[...]

PR #10922, opened 2024-02-21
Suppress codeQL warnings in tests "BitmapTests" and "ImageListTests"

LeafShi1:
Related #10871

Suppress codeQL warnings in BitmapTests and ImageListTests 

## Related bug:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1666874/?view=edit
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1775936/?view=edit

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/10922)

### DIFF
--- a/src/System.Drawing.Common/tests/mono/System.Drawing/BitmapTests.cs
+++ b/src/System.Drawing.Common/tests/mono/System.Drawing/BitmapTests.cs
@@ -390,7 +390,7 @@ public class BitmapTests
         }
 
         // cs/weak-crypto
-        hash = MD5.Create().ComputeHash(pixels);// CodeQL [SM02196] Using MD5 for this specific scenario is acceptable.
+        hash = MD5.Create().ComputeHash(pixels); // CodeQL [SM02196] This hash is used in test to compare two bitmaps.
         return ByteArrayToString(hash);
     }
 
--- a/src/System.Drawing.Common/tests/mono/System.Drawing/BitmapTests.cs
+++ b/src/System.Drawing.Common/tests/mono/System.Drawing/BitmapTests.cs
@@ -389,7 +389,8 @@ public class BitmapTests
             }
         }
 
-        hash = MD5.Create().ComputeHash(pixels);
+        // cs/weak-crypto
+        hash = MD5.Create().ComputeHash(pixels);// CodeQL [SM02196] Using MD5 for this specific scenario is acceptable.
         return ByteArrayToString(hash);
     }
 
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ImageListTests.cs
```diff
@@ -444,7 +444,8 @@ private static T RoundtripSerialize<T>(T source)
         BinaryFormatter formatter = new();
         formatter.Serialize(stream, source);
         stream.Position = 0;
-        return (T)formatter.Deserialize(stream);
+        // cs/deserialization-unexpected-subtypes
+        return (T)formatter.Deserialize(stream);// CodeQL [SM02229] Deserialization is handled correctly here.
```
Tanya-Solyanik, line 448:
```suggestion
        return (T)formatter.Deserialize(stream); // CodeQL [SM02229] Testing legacy features: we are deserializing stream with controlled content.
```

src/System.Drawing.Common/tests/mono/System.Drawing/BitmapTests.cs
```diff
@@ -389,7 +389,8 @@ public string RotateBmp(Bitmap src, RotateFlipType rotate)
             }
         }
 
-        hash = MD5.Create().ComputeHash(pixels);
+        // cs/weak-crypto
+        hash = MD5.Create().ComputeHash(pixels);// CodeQL [SM02196] Using MD5 for this specific scenario is acceptable.
```
Tanya-Solyanik, line 393:
```suggestion
        hash = MD5.Create().ComputeHash(pixels);  // CodeQL [SM02196] This hash is used in test to compare two bitmaps.
```

JeremyKuhne: We should probably remove the MD5 usage in tests. We can switch these out to SHA256 to avoid having to deal with questions or accidents. If we're comparing pre-computed hashes, just update the values we have.

LeafShi1: > We should probably remove the MD5 usage in tests. We can switch these out to SHA256 to avoid having to deal with questions or accidents. If we're comparing pre-computed hashes, just update the values we have.

I will do this change in a separate PR and track this with another issue #10930

lonitra: Once we remove MD5 usage, I don't think this suppression will be necessary anymore. Please remove this suppression as part of that change

lonitra (approved):
🚀

src/System.Drawing.Common/tests/mono/System.Drawing/BitmapTests.cs
LeafShi1: Yes, once the use of MD5 is removed , I will also cancel this suppression

PR #10927, opened 2024-02-21
Make AxHost work without classic COM interop.

### LINKED ISSUE #10583: Support ComSourceGenerator-based ActiveX controls
### .NET version

.NET 8

Yes

Worked using the classic COM interop.

### Issue description

TLB support and tooling isn't ready yet in .NET 8 so its understandable that there has been limited attention to support the new COM source generators in WinForms, but there is still desire (https://github.com/dotnet/runtime/issues/66674#issuecomment-1856072113) to get it working eventually. I'm currently prototyping (https://github.com/Devolutions/MsRdpEx/issues/97) a hand written binding for the Microsoft RDP Client, which is apparently only provided in form of an ActiveX control, so should be a decent usecase and test scenario for ActiveX support in WinForms against the new COM source generators (besides the custom MFC based controls we previously used for testing compatibility, I might also extend those to provide test coverage, not sure how well a RDP client control is suited for unit testing)

A few things to note so far, I'll probably add more as the prototype comes along:
* returning a ComWrappers instance from `GetOcx` makes WinForms fail to resolve its interfaces
* passing a ComWrappers instance to `ConnectionPointCookie` makes WinForms fail to understand it
  * the current `ConnectionPointCookie` API is not well-suited to custom bindings since it wants an interface type for reflection
  * even if the event handler interface implements both ComWrapper and classic COM interfaces WinForms still fails

My primary goal of this issue will be tracking required changes to support getting the prototype running entirely with handwritten source generator bindings, I'll open a PR for suggested fixes once I figure out more about why WinForms fails interacting with ComWrapper based instances.

### Steps to reproduce

Write COM source generator bindings against an ActiveX control and try to use it with WinForms AxHost

weltkante:
Fixes #10583

## Proposed changes

- Adjust COM interop in WinForms to have AxHost usable together with classic COM interop disabled (`<BuiltInComInteropSupport>false</BuiltInComInteropSupport>`), which is a requirement for AOT.

## Customer Impact

- Allows using ActiveX controls in AOT builds.

## Regression? 

- No

## Risk

- I may have gotten some of the interop wrong, the PR requires review
- I have no unit test coverage. Test coverage can be added but requires #10754 to have a simple control. Since it has to be handwritten and cannot be generated yet it can't be one of the more complex Windows-provided ActiveX controls.

### Before

- Application fails with exceptions if an ActiveX control is used while having configured  `<BuiltInComInteropSupport>false</BuiltInComInteropSupport>` in the project file.

### After

- Application can support a simple ActiveX while having configured  `<BuiltInComInteropSupport>false</BuiltInComInteropSupport>` in the project file.

## Test methodology 

- so far only manual testing in the scratch project, not part of the PR since it depends on the ActiveX control of #10754 which is part of a separate PR

## Test environment(s) 

- 9.0.0-preview.2.24116.2
- `<BuiltInComInteropSupport>false</BuiltInComInteropSupport>` in the project file
- I've also used `[assembly: DisableRuntimeMarshalling]` in my test, in case its relevant

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/10927)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ActiveX/AxHost.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ActiveX/AxHost.cs
@@ -9,6 +9,7 @@ using System.Globalization;
 using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
+using System.Runtime.InteropServices.Marshalling;
 using Microsoft.VisualStudio.Shell;
 using Windows.Win32.System.Com;
 using Windows.Win32.System.Com.StructuredStorage;
@@ -3464,7 +3465,15 @@ public abstract unsafe partial class AxHost : Control, ISupportInitialize, ICust
         {
             if (_instance is not null)
             {
-                Marshal.ReleaseComObject(_instance);
+                if (_instance is ComObject)
+                {
+                    // TODO: how to release a ComWrappers ComObject?
+                }
+                else
+                {
+                    Marshal.ReleaseComObject(_instance);
+                }
+
                 _instance = null;
                 DisposeHelper.NullAndDispose(ref _iOleInPlaceActiveObjectExternal);
             }
--- a/src/System.Private.Windows.Core/src/Windows/Win32/System/Com/ComHelpers.cs
+++ b/src/System.Private.Windows.Core/src/Windows/Win32/System/Com/ComHelpers.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Runtime.InteropServices;
+using System.Runtime.InteropServices.Marshalling;
 using Windows.Win32.System.Com;
 
 namespace Windows.Win32;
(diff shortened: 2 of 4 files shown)


JeremyKuhne:
@weltkante We definitely plan to get this done for .NET 9.

src/System.Windows.Forms/src/System/Windows/Forms/ActiveX/AxHost.cs
```diff
@@ -3464,7 +3465,15 @@ private void ReleaseAxControl()
...
             if (_instance is not null)
             {
-                Marshal.ReleaseComObject(_instance);
+                if (_instance is ComObject)
+                {
+                    // TODO: how to release a ComWrappers ComObject?
+                }
+                else
+                {
+                    Marshal.ReleaseComObject(_instance);
```
weltkante, line 3474: This throws if `ComHelpers.BuiltInComSupported` is false but I believe it was a no-op for `ComObject` anyways, at least used to be. I'm not clear if it is possible to do manual disposal of a `ComObject` based RCW so maybe this just needs to be a no-op for this case.

Issue #4019, opened 2020-09-28
The "ListView" property is not set for a ListViewGroup when a ListViewItem with this ListViewGroup is added to the ListView


SergeySmirnov-Akvelon:
**Is your feature request related to a problem? Please describe.**
If user creates ListViewGroup, creates ListViewItem that linked to this ListViewGroup and adds ListViewItem to ListView then ListViewGroup does not appear in ListView. As a result ListView has ListViewItem linked with a ListViewGroup that is not associated with ListView. After execution of following code:
```
var listViewGroup = new ListViewGroup("Test group");
var listViewItem = new ListViewItem("Test item", listViewGroup);
listView.Items.Add(listViewItem);
```
ListViewGroup will not be displayed in ListView. Property "ListView" also will be empty for ListViewGroup.

**Describe the solution you'd like and alternatives you've considered**
I suggest to add ListViewGroup to ListView when user adds ListViewItem that associated with ListViewGroup.
As a result if user executes code above then:

- ListViewGroup should be added to ListView and should be displayed in ListView.Groups collection
- ListView property of ListViewGroup should have ListView to which the ListViewItem was added
- If ListViewGroup already has another ListView then we should throw IOE.

Also it will avoid the case where the user has ListViewItems that linked to a ListViewGroup but forgot to add the ListViewGroup to the ListView

**Will this feature affect UI controls?**
This issue is specific for ListView. Also it will help to avoid problem with ListViewGroupAccessibleObject, because we can not get AccessibleObject for ListViewGroup when it has not ListView.

Tanya-Solyanik:
@JeremyKuhne  - the sample code adds item to the default group right now. Developer should explicitly add the new group to the list to show the item under the new group. 

```
            var listViewGroup = new ListViewGroup("Test group");
            var listViewItem = new ListViewItem("Test item", listViewGroup);
            listView1.Groups.Add(listViewGroup);
            listView1.Items.Add(listViewItem);
```

I don't think Developer is blocked and I would wait for customer request to move forward with this change. From the accessibility point, if the group is not in the listview and is not visible to the user, there is no reason to announce it in the accessibility tools.

Thus closing

PR #10934, opened 2024-02-22
Intermittent exception when creating a ListViewGroupAccessibleObject

Tanya-Solyanik:
This fix is based on this call stack, I'm eliminating the top frame by creating this Accessibility object only when the parent control is available.
```
System.InvalidOperationException: ListView

   at System.Windows.Forms.ListViewGroup.ListViewGroupAccessibleObject..ctor(ListViewGroup owningGroup, Boolean owningGroupIsDefault)
   at System.Windows.Forms.ListViewGroup.get_AccessibilityObject()
```
This is a follow up on https://github.com/dotnet/winforms/pull/10910#issuecomment-1958261561
Related to https://github.com/dotnet/winforms/issues/4019

I couldn't reproduce the issue, but the call stack is clear enough to understand the problem. There are situations when AccessibilityObject is not available, for example when the corresponding control is not displayed or not parented, thus changing nullability on this property. We were already guarding against accessing objects for "invisible" groups in the parent ListViewAccessibleObject when enumerating children.
We should not attempt to create accessible object for unparented ListView groups, for example when a group is removed from the control.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/10934)

### DIFF
--- a/src/System.Windows.Forms/System/Windows/Forms/Controls/ListView/ListViewGroup.cs
+++ b/src/System.Windows.Forms/System/Windows/Forms/Controls/ListView/ListViewGroup.cs
@@ -83,7 +83,8 @@ public sealed partial class ListViewGroup : ISerializable
                 return _accessibilityObject;
             }
 
-            // Get ListView from the group item as a workaround for https://github.com/dotnet/winforms/issues/4019
+            // Using an item from the group to obtain the ListView because the group might not yet be associated
+            // with the ListView. See https://github.com/dotnet/winforms/issues/4019.
             var listView = ListView ?? (Items.Count > 0 ? Items[0].ListView : null);
             if (listView is not null)
             {
--- a/src/System.Windows.Forms/System/Windows/Forms/Controls/ListView/ListViewGroup.cs
+++ b/src/System.Windows.Forms/System/Windows/Forms/Controls/ListView/ListViewGroup.cs
@@ -74,8 +74,25 @@ public sealed partial class ListViewGroup : ISerializable
         _headerAlignment = headerAlignment;
     }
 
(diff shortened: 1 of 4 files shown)


src/System.Windows.Forms/System/Windows/Forms/Controls/ListView/ListViewGroup.cs
```diff
@@ -74,8 +74,25 @@ public ListViewGroup(string? header, HorizontalAlignment headerAlignment) : this
...
+    internal AccessibleObject? AccessibilityObject
+    {
+        get
+        {
+            if (_accessibilityObject is not null)
+            {
+                return _accessibilityObject;
+            }
+
+            // Get ListView from the group item as a workaround for https://github.com/dotnet/winforms/issues/4019
```
JeremyKuhne, line 86: Can we explain this in addition to having the link?

JeremyKuhne (changes requested):
Asked for a comment clarification.

src/System.Windows.Forms/System/Windows/Forms/Controls/ListView/ListViewGroup.cs
Tanya-Solyanik, line 86:
```suggestion
            // Using an item from the group to obtain the ListView because the group might not yet be associated
            // with the ListView. See https://github.com/dotnet/winforms/issues/4019.
```

JeremyKuhne (approved):

Tanya-Solyanik:
test failure is a duplicate of https://github.com/dotnet/winforms/issues/13291

PR #10971, opened 2024-02-29
Remove unsupported target framework from template

### LINKED ISSUE #10938: why are there unsupported target framework options?
https://github.com/dotnet/winforms/blob/b4c9c031e5451208ac12ad2b21ed3196a5c100e6/pkg/Microsoft.Dotnet.WinForms.ProjectTemplates/content/WinFormsApplication-CSharp/.template.config/template.json#L32-L41

LeafShi1:
Fixes #10938

Remove unsupported target framework .netcore3.1 and .net5.0 from template.json


Tanya-Solyanik:
@LeafShi1  - what does `dotnet -new winforms -h` show after this change? And can you create a NET6 application from CLI or VS after this change if NET6 is not installed?

LeafShi1:
> @LeafShi1 - what does `dotnet -new winforms -h` show after this change? And can you create a NET6 application from CLI or VS after this change if NET6 is not installed?

The .netcoreapp3.1 and .net5.0 are removed after the changes:
image

The Net6 application can be created after NET6 is not installed
image

image

Tanya-Solyanik:
>The Net6 application can be created after NET6 is not installed, 

 It can be created, but it will not run when you press f5? Also what happens when you try to create a NET6 application from the command line?

>but .net core 3.1 and .net 5.0 no longer show up in Framework option even though I installed these two components

I think, this is undesirable, we should be able to create new apps for the installed SDKs. @LeafShi1  - Did you install the SDK or only the runtime? What happens when using the command line? The same?

LeafShi1:
> > The Net6 application can be created after NET6 is not installed,
> 
> It can be created, but it will not run when you press f5? Also what happens when you try to create a NET6 application from the command line?
> 
> > but .net core 3.1 and .net 5.0 no longer show up in Framework option even though I installed these two components
> 
> I think, this is undesirable, we should be able to create new apps for the installed SDKs. @LeafShi1 - Did you install the SDK or only the runtime? What happens when using the command line? The same?

I installed the template following by https://github.com/dotnet/winforms/blob/main/pkg/Microsoft.Dotnet.WinForms.ProjectTemplates/readme.md and installed the SDK "" and "" by through VS installer

> > The Net6 application can be created after NET6 is not installed,
> 
> It can be created, but it will not run when you press f5? Also what happens when you try to create a NET6 application from the command line?
> 
> > but .net core 3.1 and .net 5.0 no longer show up in Framework option even though I installed these two components
> 
> I think, this is undesirable, we should be able to create new apps for the installed SDKs. @LeafShi1 - Did you install the SDK or only the runtime? What happens when using the command line? The same?

Yes, regardless of whether the SDK is installed or not, the result is the same.

- Install the new template following by https://github.com/dotnet/winforms/blob/main/pkg/Microsoft.Dotnet.WinForms.ProjectTemplates/readme.md
- Install .NET 5.0 runtime in VS installer
- Run the command "`dotnet new winforms -h`"
image
- Install the [.NET 5.0 SDK ](https://dotnet.microsoft.com/en-us/download/dotnet/5.0)
- Run the command "`dotnet new winforms -h`"
image

Tanya-Solyanik:
@LeafShi1 - we'll re-triage this issue

merriemcgaw:
@LeafShi1 can you experiment with making our template just target the single most recent SDK and see how that interacts with everything? Maybe check out ASP.NET templates, they are linked in the issue associated with this PR.

merriemcgaw:
We'll experiment on our end what changing the target framework to be a single target framework does.

RussKie:
It may worth considering adding a script that iterates over the following jsons and verifies that only supported/expected TFMs are defined in those:

* pkg/Microsoft.Dotnet.WinForms.ProjectTemplates/content/\*\*/.template.config/template.json
* pkg/Microsoft.Dotnet.WinForms.ProjectTemplates/content/\*\*/.template.config/localize/templatestrings.\*.json

LeafShi1:
> @LeafShi1 can you experiment with making our template just target the single most recent SDK and see how that interacts with everything? Maybe check out ASP.NET templates, they are linked in the issue associated with this PR.

If only .net9.0 option is kept, only ".NET 9.0" option will be displayed even if other SDK and runtime are installed
image

If only .net8.0（.net7.0 or.net6.0 ）option is kept and default platform set to .net8.0（.net7.0 or.net6.0 ）

- The project can be created successfully via the command line “`dotnet new winforms --framework=net8.0 -n testapp1`”
- The "Additional information" will not be show and VS close directly after showing "Loading template description" 

image

GangWang01:
@JanKrivanek not sure if you still remember why console project template has only the latest framework https://github.com/dotnet/sdk/blob/main/template_feed/Microsoft.DotNet.Common.ProjectTemplates.9.0/content/ConsoleApplication-CSharp/.template.config/template.json#L30-L44 while `dotnet new console -h` or VS could list other frameworks.

JanKrivanek:
> @JanKrivanek Jan Krivanek FTE not sure if you still remember why console project template has only the latest framework https://github.com/dotnet/sdk/blob/main/template_feed/Microsoft.DotNet.Common.ProjectTemplates.9.0/content/ConsoleApplication-CSharp/.template.config/template.json#L30-L44 while `dotnet new console -h` or VS could list other frameworks.

IIRC it has to do with the fact of 'combining' the templates of the same name (into a template group) - so if multiple TFMs shows up as availble for simple console template - it most likely means there are multiple templates - probably brought by multiple SDKs.

The winforms template seems to take stratgy of actually including the additional TFMs within the single template - so that they are available even if particular sdk is not installed

Tanya-Solyanik:
>The winforms template seems to take stratgy of actually including the additional TFMs within the single template - so that they are available even if particular sdk is not installed

@JanKrivanek - Based on Leaf's experiment above, it seems that winforms can't break with the strategy of including all supported TFMs into the template for NET9 because the downlevel TFMs are not picked up from the downlevel templates. Is this expected?
If it is expected, we can continue having a choice of multiple TFMs, but now we will restrict them to the supported ones.

Tanya-Solyanik (approved):

JanKrivanek:
Yes - it should work. Ideally you should increase the `precedence` number with each release, that needs to overlay and hid templates tagged with the same `groupIdentity`.

More reading: https://github.com/dotnet/templating/blob/main/docs/Using-Group-Identity.md

You seem to be already doing this for each major release - so you should be good to go. I'm mentioning this in case you'd want the changes to be properly picked up even between previews (as I'm not sure, there is any guarantee on which template is going to be selected from the group if there are multiple packages with same `groupIdentity` and `precedence`)

PR #10975, opened 2024-03-01
Add test "ToolStrip_Font_ApplyParentFontToMenus_GetReturnFont_SameAsA…

### LINKED ISSUE #10977: Add unit test for switch AppContext switch "System.Windows.Forms.ApplyParentFontToMenus"
The switch was added in PR #10906 , and lack of a unit test to verify the impact of Application.SetDefaultFont on ToolStrip font

**Expected result：**
The ToolStrip‘font same as the Application's default font when setting `System.Windows.Forms.ApplyParentFontToMenus: true`

LeafShi1:
Fixes #10977

## Proposed changes
Add test for switch "`ApplyParentFontToMenus`" in the interactive test - `WinformsControlsTest `app

- Set the default font of application `WinformsControlsTest ` to "`Microsoft Sans Serif 8.25F`"
- Set the switch `ApplyParentFontToMenus `to true

### Before
The toolstrip's font is `{Name = "Segoe UI" Size=9}`
image

### After
The toolstrip's font is `{Name = "Microsoft Sans Serif" Size=8.25}`
image

## Test methodology 

- interactive test

## Test environment(s) 

- .net 9.0.0-preview.3.24128.10

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/10975)


src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripTests.cs
```diff
@@ -1619,6 +1619,69 @@ public void ToolStrip_Font_SetWithItemsWithFontWithHandler_CallsFontChanged()
         Assert.Equal(3, callCount);
     }
 
+    [WinFormsFact]
+    public void Form_Font_GetReturnFont_SameAsApplication()
+    {
+        // Generally at this stage NativeWindow.AnyHandleCreated=true, we won't be able to set the font.
```
Tanya-Solyanik, line 1625: It's ok to test this scenario in the interactive test - `WinformsControlsApp`

LeafShi1: In the Unit test, the font of the Application cannot affect the font of the Form and the font of the control. The test for switch "ApplyParentFontToMenus" has been added to the WinformsControlsTest App

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -104,6 +104,11 @@ static bool GetSwitchDefaultValue(string switchName)
                 return false;
             }
 
+            if (switchName == ApplyParentFontToMenusSwitchName)
```
lonitra, line 107: Do we want this switch to be only available in .NET 9? If so I think we need to place this under the if statement below.

Tanya-Solyanik (commented):
> Do we want this switch to be only available in .NET 9? If so I think we need to place this under the if statement below. 

We don't have to. This is a good point though, this switch should be available after SetApplicatinFont API had beed introduced

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/ToolStripTests.cs
```diff
@@ -10,6 +11,7 @@ public partial class ToolStripTests : Form
     public ToolStripTests()
     {
         InitializeComponent();
+        LocalAppContextSwitches.SetLocalAppContextSwitchValue(LocalAppContextSwitches.ApplyParentFontToMenusSwitchName, true);
```
Tanya-Solyanik, line 14: Had you tried adding a runtimeconfig.json file? I would add the default values for other switches for easy experiments:

{
  "configProperties": {
    "System.Windows.Forms.AnchorLayoutV2": false,
    "System.Windows.Forms.ApplyParentFontToMenus": true,
    "System.Windows.Forms.DataGridViewUIAStartRowCountAtZero": false,
    "System.Windows.Forms.DoNotCatchUnhandledExceptions": true,
    "System.Windows.Forms.EnableMsoComponentManager": false
    "System.Windows.Forms.NoClientNotifications": false,
    "System.Windows.Forms.ScaleTopLevelFormMinMaxSizeForDpi": false,
    "System.Windows.Forms.ServicePointManagerCheckCrl": false,
    "System.Windows.Forms.TrackBarModernRendering": false,
  }
}

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/Program.cs
```diff
@@ -10,6 +10,7 @@
 ApplicationConfiguration.Initialize();
 
 Application.SetUnhandledExceptionMode(UnhandledExceptionMode.ThrowException);
+Application.SetDefaultFont(new Font(new FontFamily("Microsoft Sans Serif"), 8.25F, FontStyle.Regular, GraphicsUnit.Point, 0));
```
Tanya-Solyanik, line 13: the project file is already setting ApplicationFont.

Tanya-Solyanik (changes requested):

Tanya-Solyanik:
> > Do we want this switch to be only available in .NET 9? If so I think we need to place this under the if statement below.
> 
> We don't have to. This is a good point though, this switch should be available after SetApplicatinFont API had beed introduced

Never mind, this switch has 2 effects - inherit Font from the parent container if available or from the Application object. Parent container font is applicable to all versions. We don't have to special-case the switch availability, we just use the same default value everywhere, if the developer wants the switch ON, they have to opt-in.

src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/ToolStripTests.cs
Tanya-Solyanik: Also it would be nice to have another toolstrip with a custom font to show that custom font is preserved.

Tanya-Solyanik (dismissed):

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
Tanya-Solyanik, line 107: Do we still need this change?

LeafShi1: I don't think this is necessary, I will remove this.

SimonZhao888 (approved):

PR #11024, opened 2024-03-11
Clean up initialization messages

JeremyKuhne:
This change cleans up some of the instances of the following analyzer messages (using the code fixes primarily):

- Simplify type names (IDE0002)
- Redundant casts (IDE0004)
- Collection initialization syntax (IDE0028)
- Simplify object initialization (IDE0017)
- More specific types for performance (CA1859)

There are issues with applying initialization syntax to test theory data so I haven't touched much of the test code. Using collection initialization also doesn't work with some of our collections (such as `DesignerVerbCollection`). I'm pretty sure that is related to them not implementing a `IEnumerable<T>`.

The plan is to follow up and make sure our collections have at least `IEnumerable<T>` to enable using this syntax and that using said syntax doesn't go down the `Add(object)` path if said overload exists.

Additionally, I hope to turn some of these into at least warnings.

The hope is that with less for the analyzers to report it will improve the development experience in VS (and it appears to). I'd like to see our code files clean of all messages to also help us to write new code cleanly.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11024)


JeremyKuhne:
There are a number of other messages I intend to clean. Using `<para>` in the `<remarks>` for example.

Tanya-Solyanik:
Do you have any suggestions on how to review this change? This is  kind of thing that is verified by the CI build...

paul1956:
I have same issue with VB cleanup PR, code coverage goes to hell because I touch a lot of files that never had any coverage. For all the public functions I could add tests but a majority of the code is not public and I have no experience in writing tests for "Friend" functions.

JeremyKuhne:
> Do you have any suggestions on how to review this change? This is kind of thing that is verified by the CI build...

You could just walk through it to look for anything glaringly wrong. It was mostly done with the auto-fixers so you shouldn't have to think hard about it. Looking at the first 50 files again it didn't take too long for me.

Tanya-Solyanik (approved):
I scanned the changes, they look reasonable, and didn't wee any new warnings when building in the VS or on the command line.

PR #11025, opened 2024-03-11
Convert to 1ES Pipeline Template

lonitra:
Best to review without whitespace since the conversion tool removed any unnecessary indents
test run: https://dev.azure.com/dnceng/internal/_build/results?buildId=2400049&view=results
The error in the test run is being tracked by https://github.com/dotnet/arcade/issues/14560 but it does not block successful build

note- the PR validation does not use this new yml file, it uses `azure-pipelines-pr.yml` which is a dupe of the original azure-pipelines.yml definition
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11025)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -2,6 +2,11 @@
 # Please make sure to check all the converted content, it is your team's responsibility to make sure that the pipeline is still valid and functions as expected.
 # This pipeline will be extended to the OneESPT template
 # If you are not using the E+D shared hosted pool with windows-2022, replace the pool section with your hosted pool, os, and image name. If you are using a Linux image, you must specify an additional windows image for SDL: https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-docs/1es-pipeline-templates/features/sdlanalysis/overview#how-to-specify-a-windows-pool-for-the-sdl-source-analysis-stage
+
+# Parameters ARE available in template expressions, and parameters can have default values,
+# so they can be used to control yaml flow.
+
+# trigger ci builds for completed checkins into main and any release branches
 trigger:
   branches:
     include:
@@ -22,6 +27,8 @@ trigger:
     - README.md
     - SECURITY.md
     - THIRD-PARTY-NOTICES.TXT
+
+# trigger ci builds on pull requests into main and any release branches
 pr:
   autoCancel: true
   branches:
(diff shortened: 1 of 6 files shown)


azure-pipelines.yml
```diff
@@ -5,192 +5,204 @@
...
-      - THIRD-PARTY-NOTICES.TXT
+    - .github/*
+    - docs/*
+    - CODE-OF-CONDUCT.md
+    - CONTRIBUTING.md
+    - LICENSE.TXT
+    - PATENTS.TXT
+    - README.md
+    - SECURITY.md
+    - THIRD-PARTY-NOTICES.TXT
```
Shyam-Gupta, line 50: Now that we have separate yml for CI and PR builds, is it okay to remove `pr` section from this yml now?

azure-pipelines.yml
```diff
@@ -5,192 +5,204 @@
...
+  value: ${{ contains(variables['Build.SourceBranch'], 'main') }}
+- name: NativeToolsOnMachine
+  value: true
+- name: isPublic
+  value: ${{ eq(variables['System.TeamProject'], 'public') }}
+- name: isPR
+  value: ${{ eq(variables['Build.Reason'], 'PullRequest')}}
+
+# used for post-build phases, internal builds only
 - ${{ if and(ne(variables['isPublic'], 'true'), ne(variables['isPR'], 'true')) }}:
```
Shyam-Gupta, line 74: Since this yml file will only be used for CI builds, we can probably remove the checks for `isPR` throughout this file to remove conditionals as much as possible.

lonitra: Thanks for the catch! Ran new test run to make sure nothing got inadvertently changed with removing the conditionals: https://dev.azure.com/dnceng/internal/_build/results?buildId=2401893&view=results

RussKie:
FWIW, the way I approached the work was to separate it into two streams. First, I created, tested and updated the public pipelines (azure-pipelines-public.yml or azure-pipelines-PR.yml), and then concentrated on the internal pipelines. The former is much simpler and easier to do, and allowed me to concentrate on the complexities of the internal migrations.

The easiest way to test the public pipeline changes was to create the new yml files, but also _in your test branch only_ replace the azure-pipelines.yml with the content from the new yml. That way you can be sure that when you test-build your branch you build against your new yml. I learned it the hard way 😄

Shyam-Gupta:
Looks good. :shipit:

Shyam-Gupta (approved):

lonitra:
Thanks for sharing @RussKie ! That sounds much easier and will probably how I'll go about it with the servicing branches. 😄

PR #11045, opened 2024-03-13
More naming fixes

JeremyKuhne:
Bringing naming in-line with analyzers and adding appropriate suppressions where we can't satisfy the analyzer.

Also turn a couple more analyzers into errors.

This is the first step. Trying to keep these to less than 100 files.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11045)


.editorconfig
```diff
@@ -113,9 +116,16 @@ dotnet_diagnostic.DOC100.severity = error
...
+# IDE0002: Simplify Member Access
+dotnet_diagnostic.IDE0002.severity = error
+
 # IDE0003: Remove qualification
 dotnet_diagnostic.IDE0003.severity = error
 
+# IDE0079: Remove unnecessary suppression
+dotnet_diagnostic.IDE0079.severity = error
+
+
```
Tanya-Solyanik, line 128: Are the 2 blank lines intentional?

src/System.Private.Windows.Core/src/Windows/Win32/System/Com/GlobalInterfaceTable.cs
```diff
@@ -15,11 +15,12 @@ internal static unsafe partial class GlobalInterfaceTable
 
     static GlobalInterfaceTable()
     {
+        Guid guid = CLSID.StdGlobalInterfaceTable;
+
         fixed (IGlobalInterfaceTable** git = &s_globalInterfaceTable)
-        fixed (Guid* g = &CLSID.StdGlobalInterfaceTable)
```
Tanya-Solyanik, line 19: Why is `fixed` not needed anymore here?

src/System.Drawing.Common/src/System/Drawing/DrawingCom.ComWrappers.cs
```diff
@@ -56,7 +56,9 @@ internal static void ThrowExceptionForHR(int errorCode)
 
     internal interface IPicture : IDisposable
     {
-        static readonly Guid IID = new(0x7BF80980, 0xBF32, 0x101A, 0x8B, 0xBB, 0, 0xAA, 0x00, 0x30, 0x0C, 0xAB);
+#pragma warning disable IDE1006 // Naming Styles
+        internal static Guid IID = new(0x7BF80980, 0xBF32, 0x101A, 0x8B, 0xBB, 0, 0xAA, 0x00, 0x30, 0x0C, 0xAB);
```
Tanya-Solyanik, line 60: Why not make this one a get-only property, like you did above?

Tanya-Solyanik (approved):
Please see the comments.

.editorconfig
JeremyKuhne: Yes

src/System.Private.Windows.Core/src/Windows/Win32/System/Com/GlobalInterfaceTable.cs
JeremyKuhne: Because it is a local copy of the Guid, it is implicitly fixed.

src/System.Drawing.Common/src/System/Drawing/DrawingCom.ComWrappers.cs
JeremyKuhne: Because I'm going to remove this later this week. This is the last old interop in System.Drawing.

PR #11054, opened 2024-03-14
Re-enable CS0618

JeremyKuhne:
CS0618 was disabled in the project. Re-enable and add exclusions in source.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11054)


src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -10370,7 +10370,7 @@ protected virtual void ScaleCore(float dx, float dy)
             // enumerate
             for (int i = 0; i < controlsCollection.Count; i++)
             {
-                controlsCollection[i].Scale(dx, dy);
+                controlsCollection[i].Scale(new SizeF(dx, dy));
```
Tanya-Solyanik, line 10373: We can put this SizeF into a local variable and set outside the loop

Tanya-Solyanik (approved):

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
JeremyKuhne: Whoops, this was a mistake.

JeremyKuhne: Should not have changed as it changes the way layout works.

Tanya-Solyanik (approved):

PR #11057, opened 2024-03-14
Remove unused code

JeremyKuhne:
This enables a number of additional analyzers. The key one is turning unused code into warnings. Ends up we had quite a bit of unused parameters and fields.

I've tried to err on the side of caution when properties or methods might be used in the designer.

This also enforces using switch expressions when possible.

Other enablements (that didn't change much):

- Use keywords for types (int over System.Int) is now a warning
- Use null propagation is  now a warning
- Precompiled regex is now an error
- No protected members in sealed types
- Suppression format checks
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11057)


src/System.Drawing.Common/src/System/Drawing/Image.cs
```diff
@@ -160,7 +160,7 @@ public static Image FromStream(Stream stream, bool useEmbeddedColorManagement, b
     private static GpImage* LoadGdipImageFromStream(Stream stream, bool useEmbeddedColorManagement)
     {
         using var iStream = stream.ToIStream(makeSeekable: true);
-        return LoadGdipImageFromStream(iStream, useEmbeddedColorManagement: false);
```
lonitra, line 163: Did we have instances where we were calling this method with `true`?

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ToolStripItemDesigner.cs
```diff
@@ -334,7 +327,7 @@ internal virtual void CommitEdit(Type type, string text, bool commit, bool enter
 
         IDesignerHost designerHost = GetService<IDesignerHost>();
         ToolStripDesigner designer = (ToolStripDesigner)designerHost.GetDesigner(ToolStripItem.Owner);
-        designer?.EditManager?.ActivateEditor(null, false);
+        designer?.EditManager?.ActivateEditor(null);
```
lonitra, line 330: nit: param name

src/System.Windows.Forms.Design/tests/UnitTests/System/ComponentModel/Design/CollectionEditorTests.cs
```diff
@@ -1073,7 +1073,9 @@ private class ClassWithItem
 
     private class ClassWithPrivateItem
     {
+#pragma warning disable IDE0051 // Remove unused private members
         private int Item { get; set; }
+#pragma warning restore IDE0051
```
lonitra, line 1078: Do we need to keep these if they are unused and in tests?

lonitra (approved):
Minor comment/questions, LGTM 👍

src/System.Drawing.Common/src/System/Drawing/Image.cs
JeremyKuhne: Yeah, so this is a bug fix.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ToolStripItemDesigner.cs
JeremyKuhne: Catch in follow up

src/System.Windows.Forms.Design/tests/UnitTests/System/ComponentModel/Design/CollectionEditorTests.cs
JeremyKuhne: It is testing access to the members so we need to keep them in these cases. (Test would fail otherwise)

PR #11061, opened 2024-03-15
Remove finalization watch from System.Drawing

JeremyKuhne:
Removing code that does not compile as part of our normal builds.

In the future we can add debug code in the finalizer methods that will fail when the finalizer is hit (as we do with the scopes for HDC, etc.). Keeping an initialization stack isn't super useful.

Note that the try catch blocks are not needed as with direct interop we shouldn't see exceptions that we would handle.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11061)


src/System.Drawing.Common/src/System/Drawing/Imaging/ImageAttributes.cs
```diff
@@ -71,46 +67,21 @@ internal ImageAttributes(GpImageAttributes* newNativeImageAttributes)
...
+        _nativeImageAttributes = null;
+        Debug.Assert(status == Status.Ok, $"GDI+ returned an error status: {status}");
+        GC.SuppressFinalize(this);
     }
 
     /// <summary>
     ///  Cleans up Windows resources for this <see cref='ImageAttributes'/>.
     /// </summary>
-    ~ImageAttributes() => Dispose(disposing: false);
+    ~ImageAttributes() => Dispose();
```
Tanya-Solyanik, line 84: This would result in SuppressFinalize called from the finalizer. Is it safe?

Tanya-Solyanik (approved):

src/System.Drawing.Common/src/System/Drawing/Imaging/ImageAttributes.cs
JeremyKuhne: Yep, just sets a flag.

PR #11062, opened 2024-03-15
Remove TraceSwitch from CodeDomSerializer

JeremyKuhne:
Tracing predates debugger trace points, which accomplish what we typically need to do when debugging.

Not having the tracing makes it easier to read the code, as well as easier to make complexity-reducing automated changes (inverting blocks, returning early, etc.).

Some of the affected code here is currently indented over 10 levels deep.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11062)


src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/ResourceCodeDomSerializer.cs
```diff
@@ -351,43 +344,33 @@ public object SerializeInvariant(IDesignerSerializationManager manager, object v
     /// <summary>
     ///  Writes out the given metadata.
     /// </summary>
-    public void SerializeMetadata(IDesignerSerializationManager manager, string name, object? value, bool shouldSerializeValue)
+    public static void SerializeMetadata(IDesignerSerializationManager manager, string name, object? value, bool shouldSerializeValue)
```
Tanya-Solyanik, line 347: Nice!

Tanya-Solyanik (approved):

PR #11073, opened 2024-03-19
Enable more code analyzers

JeremyKuhne:
Enable a number of additional code analyzers. One of the more impactful ones here is that we don't want to throw "runtime" exceptions or Exception itself. Changed a number of `Exception` instances to `InvalidOperationException`.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11073)


src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerHost.Site.cs
```diff
@@ -240,10 +240,11 @@ internal bool Disposed
                         // allow renames that are just case changes of the current name.
                         if (namedComponent is not null && validateName)
                         {
-                            Exception ex = new(string.Format(SR.DesignerHostDuplicateName, value))
+                            InvalidOperationException ex = new(string.Format(SR.DesignerHostDuplicateName, value))
                             {
                                 HelpLink = SR.DesignerHostDuplicateName
                             };
+
                             throw ex;
```
lonitra, line 248: nit:
```suggestion
                            throw new InvalidOperationException(string.Format(SR.DesignerHostDuplicateName, value))
                            {
                                HelpLink = SR.DesignerHostDuplicateName
                            };
```

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/DesignerHost.cs
```diff
@@ -165,7 +165,7 @@ internal bool AddToContainerPreProcess(IComponent component, string? name, ICont
         // We should never add anything while we're unloading.
         if (_state[s_stateUnloading])
         {
-            Exception ex = new(SR.DesignerHostUnloading)
+            InvalidOperationException ex = new(SR.DesignerHostUnloading)
             {
                 HelpLink = SR.DesignerHostUnloading
             };
```
lonitra, line 171: Ditto here and some other places in this file as well

lonitra (approved):
Minor nit, but can be addressed in a follow up

PR #11118, opened 2024-03-24
Enable nullability in some ControlDesigner members

gpetrou:
## Proposed changes

- Enable nullability in some ControlDesigner members.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11118)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.ChildWindowTarget.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.ChildWindowTarget.cs
@@ -61,7 +61,7 @@ public partial class ControlDesigner
                 // If the designer disposed us, then we should follow suit.
                 if (_designer.DesignerTarget is null)
                 {
-                    designerTarget!.Dispose();
+                    designerTarget?.Dispose();
                 }
                 else
                 {
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.ChildWindowTarget.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.ChildWindowTarget.cs
@@ -45,7 +45,7 @@ public partial class ControlDesigner
             // We want these messages to go through the designer's WndProc method, and we want people to be able
             // to do default processing with the designer's DefWndProc.  So, we stuff the old window target into
             // the designer's target and then call their WndProc.
-            IDesignerTarget designerTarget = _designer.DesignerTarget;
+            IDesignerTarget? designerTarget = _designer.DesignerTarget;
             _designer.DesignerTarget = this;
 
             try
@@ -61,7 +61,7 @@ public partial class ControlDesigner
                 // If the designer disposed us, then we should follow suit.
                 if (_designer.DesignerTarget is null)
                 {
-                    designerTarget.Dispose();
+                    designerTarget!.Dispose();
                 }
                 else
                 {
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.cs
@@ -83,8 +83,7 @@ public partial class ControlDesigner : ComponentDesigner
         {
             Point loc = Control.Location;
 
-            ScrollableControl? parent = Control.Parent as ScrollableControl;
-            if (parent is not null)
+            if (Control.Parent is ScrollableControl parent)
             {
                 Point pt = parent.AutoScrollPosition;
                 loc.Offset(-pt.X, -pt.Y);
@@ -94,8 +93,7 @@ public partial class ControlDesigner : ComponentDesigner
         }
         set
         {
-            ScrollableControl? parent = Control.Parent as ScrollableControl;
-            if (parent is not null)
+            if (Control.Parent is ScrollableControl parent)
             {
                 Point pt = parent.AutoScrollPosition;
                 value.Offset(pt.X, pt.Y);
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.cs
@@ -1,8 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-#nullable disable
-
 using System.Collections;
 using System.ComponentModel;
 using System.ComponentModel.Design;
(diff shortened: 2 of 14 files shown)


src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.cs
```diff
@@ -216,7 +215,7 @@ public virtual SelectionRules SelectionRules
 
             if ((prop = props["Size"]) is not null && !prop.IsReadOnly)
             {
-                if (AutoResizeHandles && Component != _host.RootComponent)
+                if (AutoResizeHandles && Component != _host?.RootComponent)
```
lonitra, line 218: It looks like this is where our tests were originally expecting NRE. Should this be changed to throw `InvalidOperationException` instead of continuing on when `_host = null`? Thoughts @Tanya-Solyanik , @JeremyKuhne ?

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.cs
```diff
@@ -85,7 +83,7 @@ private Point Location
         {
             Point loc = Control.Location;
 
-            ScrollableControl p = Control.Parent as ScrollableControl;
+            ScrollableControl? p = Control.Parent as ScrollableControl;
```
lonitra, line 86: Could we update this to have more descriptive name
```suggestion
            ScrollableControl? parent = Control.Parent as ScrollableControl;
```

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.cs
```diff
@@ -96,7 +94,7 @@ private Point Location
         }
         set
         {
-            ScrollableControl p = Control.Parent as ScrollableControl;
+            ScrollableControl? p = Control.Parent as ScrollableControl;
```
lonitra, line 97: Here as well

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.ChildWindowTarget.cs
```diff
@@ -61,7 +61,7 @@ public void OnMessage(ref Message m)
                 // If the designer disposed us, then we should follow suit.
                 if (_designer.DesignerTarget is null)
                 {
-                    designerTarget.Dispose();
+                    designerTarget!.Dispose();
```
JeremyKuhne, line 64: I wouldn't let this fail.
```suggestion
                    designerTarget?.Dispose();
```

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.cs
JeremyKuhne: In this case I'd add a `Debug.Assert` and keep the `?.`. Presumption is that this probably should not happen, but there is no real bad in returning rules in this case. You'll need to add a no assert context to the test.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.cs
```diff
@@ -289,11 +288,15 @@ public virtual SelectionRules SelectionRules
     internal Point GetOffsetToClientArea()
     {
         Point nativeOffset = default;
-        PInvoke.MapWindowPoints(Control, Control.Parent, ref nativeOffset);
+        if (Control.Parent is not null)
```
JeremyKuhne, line 291: Please try to use pattern matching wherever you can.

```suggestion
        if (Control.Parent is {} parent)
```

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.cs
```diff
@@ -85,21 +83,21 @@ private Point Location
         {
             Point loc = Control.Location;
 
-            ScrollableControl p = Control.Parent as ScrollableControl;
-            if (p is not null)
+            ScrollableControl? parent = Control.Parent as ScrollableControl;
+            if (parent is not null)
```
lonitra, line 87: Let's change to use pattern matching here. Ditto with the setter as well.
```suggestion
            if (Control.Parent is ScrollableControl parent)
```

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.cs
```diff
@@ -159,13 +157,14 @@ private bool Locked
...
-    private string Name
+    private string? Name
     {
-        get => Component.Site.Name;
+        get => Component.Site?.Name;
         set
         {
             // Don't do anything here during loading, if a refactor changed it we don't want to do anything.
-            if (GetService(typeof(IDesignerHost)) is not IDesignerHost host || (host is not null && !host.Loading))
+            if ((GetService(typeof(IDesignerHost)) is not IDesignerHost host || (host is not null && !host.Loading))
```
lonitra, line 166: Could `TryGetService<IDesignerHost>()` be used here?

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.cs
```diff
@@ -304,23 +306,23 @@ internal Point GetOffsetToClientArea()
...
     {
         object component = Component;
         bool resizable = true;
         bool autoSize = false;
         bool growOnly = false;
 
-        AttributeCollection attributes = autoSizeProp?.Attributes;
+        AttributeCollection? attributes = autoSizeProp?.Attributes;
         if (attributes is not null && !(attributes.Contains(DesignerSerializationVisibilityAttribute.Hidden)
                 || attributes.Contains(BrowsableAttribute.No)))
```
lonitra, line 318:
```suggestion
        if (autoSizeProp?.Attributes is { }
        && !(attributes.Contains(DesignerSerializationVisibilityAttribute.Hidden)
             || attributes.Contains(BrowsableAttribute.No)))
```

lonitra (commented):
Few more minor comments, otherwise LGTM!

gpetrou:
Since we added the `Debug.Assert` some tests will fail now. Should I remove it?

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.cs
```diff
@@ -159,13 +156,14 @@ private bool Locked
...
-    private string Name
+    private string? Name
     {
-        get => Component.Site.Name;
+        get => Component.Site?.Name;
         set
         {
             // Don't do anything here during loading, if a refactor changed it we don't want to do anything.
-            if (GetService(typeof(IDesignerHost)) is not IDesignerHost host || (host is not null && !host.Loading))
+            if ((TryGetService(out IDesignerHost? host) || (host is not null && !host.Loading))
```
lonitra, line 165: Looks like we missed a `!` this might be why some tests are failing.
```suggestion
            if (!TryGetService(out IDesignerHost? host) || (host is not null && !host.Loading))
```

lonitra:
> Since we added the `Debug.Assert` some tests will fail now. Should I remove it?

Could you add `NoAssertContext` in the tests that are failing because of the Debug.Assert? This will eat Debug.Asserts

lonitra (approved):
👍

PR #11127, opened 2024-03-26

gpetrou:
## Proposed changes

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11127)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGridInternal/GridEntry.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGridInternal/GridEntry.cs
@@ -347,7 +347,7 @@ internal abstract partial class GridEntry : GridItem, ITypeDescriptorContext
 
             TypeConverter converter = TypeConverter;
             UITypeEditor? editor = UITypeEditor;
-            object value = Instance;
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Controls/PropertyGrid/PropertyGridInternal/GridEntry.cs
```diff
@@ -528,7 +528,7 @@ internal virtual PropertyGridView OwnerGridView
     /// <summary>
     ///  Retrieves the keyword that Visual Studio dynamic help window will use when this entry is selected.
     /// </summary>
-    public virtual string HelpKeyword => _parent?.HelpKeyword ?? string.Empty;
+    public virtual string? HelpKeyword => _parent?.HelpKeyword ?? string.Empty;
```
gpetrou, line 531: I had to make this property nullable because it is nullable in overridden properties.

lonitra: This change to make the virtual nullable seems fine to me.

lonitra (approved):

Issue #11083, opened 2024-03-20
NRE In Unmanaged Code When Pressing Keys In TreeView


stumper66:
### .NET version

.Net 8.0

### Did it work in .NET Framework?

Yes

### Did it work in any of the earlier releases of .NET Core or .NET 5+?

My project started in .Net Framework 4.x. I eventually migrated to .Net Core 3.1, then 5.0, 6.0, 7.0 and finally 8.0.
I can't remember exactly when this issue started but it might've been with .net core 3.1.

### Issue description

I have a TreeView control. I also have various keyboard shortcuts which are handled by a MenuStrip.
I've discovered that randomly when I'm using the arrow keys to move around inside the TreeView control, I will get an exception that doesn't contain any user code at all so I'm unable to handle it.
I've found the exact error spot in the source code and even made my own fork that handles the error graceful with one simple line change.
It appears to be attempting to send a ToolTip to the form despite me not having any ToolTips.

Error line in source: https://github.com/dotnet/winforms/blob/release/8.0/src/System.Windows.Forms/src/System/Windows/Forms/ToolTip.cs#L1576
In the source it is calling `tool.GetOwnerWindow()` however this returns a null value and the code seems to be expecting a non-null value.

### Steps to reproduce

I've been unable to reproduce this error in a brand new project. It only occurs on my mature project that originally started out in .Net Framework 4.x.
![Net Issue](https://github.com/dotnet/winforms/assets/18266662/20a5fef7-21d5-45aa-b1f8-640afecf3259)

stumper66:
Full callstack:
System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=System.Windows.Forms
  StackTrace:
   at System.Windows.Forms.Control.GetSafeHandle(IWin32Window window)
   at System.Windows.Forms.ToolTip.TryGetBubbleSize(IKeyboardToolTip tool, Size& bubbleSize)
   at System.Windows.Forms.ToolTip.ShowKeyboardToolTip(String text, IKeyboardToolTip tool, Int32 duration)
   at System.Windows.Forms.KeyboardToolTipStateMachine.ShowToolTip(IKeyboardToolTip tool, ToolTip toolTip)
   at System.Windows.Forms.KeyboardToolTipStateMachine.Transition(IKeyboardToolTip tool, ToolTip tooltip, SmEvent event)
[...]

PR #11134, opened 2024-03-27
Fix NPE Error

stumper66:
Fixes https://github.com/dotnet/winforms/issues/11083
Fixes https://github.com/dotnet/winforms/issues/7262

## Proposed changes

•	Added null-checks when accessing the owner DataGridView property from the DataGridView cell because this is a public property and could be set to null either by user code of when the owner control is re-created. Use default values when the owner DGV control is null.
•	Do not show the ToolTip when the owner window is null. This could happen when owner is available at the moment when the tool tip timer is started, but is destroyed by the end of the timer interval when the tooltip is ready to be shown. 

## Customer Impact
Fixes a rare application crash.

## Regression? 

•	Probably not a regression, the same code is present in the .NET Framework, where the keyboard tooltips were originally introduced. However, we don’t have customer reports about this issue in the framework. Most likely because .NET is faster.

## Risk
 Low – existing workarounds with  ThreadException handlers or UnhandledException handlers will not be affected, the application will end up showing an empty tooltip or will not show it at this time but will on the next `TAB`.

## Test methodology 
* I provided my project the updated System.Windows.Forms.dll and it resolved the issue.
*  Unit test is hard to add because this is an intermittent timing issue.
 

## Test environment(s) 

- 
.NET SDK:
 Version:           9.0.100-preview.2.24157.14
 Commit:            f7466905f9
 Workload version:  9.0.100-manifests.04914b26
 MSBuild version:   17.10.0-preview-24127-03+6f44380e4

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.22631
 OS Platform: Windows
 RID:         win-x64
 Base Path:   C:\Program Files\dotnet\sdk\9.0.100-preview.2.24157.14\

EN-US.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11092)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11134)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/ToolTip/ToolTip.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/ToolTip/ToolTip.cs
@@ -1481,12 +1481,15 @@ public partial class ToolTip : Component, IExtenderProvider, IHandle<HWND>
         int pointX = (toolRectangle.Left + toolRectangle.Right) / 2;
         int pointY = (toolRectangle.Top + toolRectangle.Bottom) / 2;
         var ownerWindow = tool.GetOwnerWindow();
-        Debug.Assert(ownerWindow is not null);
+        if (ownerWindow is null)
+        {
+            return;
+        }
 
         SetTool(ownerWindow, text, TipInfo.Type.Absolute, new Point(pointX, pointY));
 
         // Then look for a better ToolTip location.
-        if (TryGetBubbleSize(tool, out Size bubbleSize))
+        if (TryGetBubbleSize(tool, ownerWindow, out Size bubbleSize))
         {
             Point optimalPoint = GetOptimalToolTipPosition(tool, toolRectangle, bubbleSize.Width, bubbleSize.Height);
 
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/ToolTip/ToolTip.cs
```diff
@@ -1524,7 +1527,7 @@ internal void ShowKeyboardToolTip(string? text, IKeyboardToolTip tool, int durat
         }
     }
 
-    private bool TryGetBubbleSize(IKeyboardToolTip tool, out Size bubbleSize)
+    private bool TryGetBubbleSize(IKeyboardToolTip tool, IWin32Window ownerWindow, out Size bubbleSize)
```
lonitra, line 1530: Looks like the build is complaining that this `tool` parameter is no longer being used in this method. Let's go ahead and remove it.

Tanya-Solyanik (approved):

lonitra (approved):

PR #11164, opened 2024-04-05
Fix Few CodeQL Warnings

lonitra:
This PR fixes a few CodeQL warnings
- Add `NonSerializedAttribute` to explicitly indicate that pointers/handles are not being serialized ([ex](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1943327/?view=edit))
- Suppress [warning](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1845222/?view=edit) in `ResXSerializationBinder` as this is meant to redirect to .NET Framework type names, if we cannot do so we want to fall back to default binder
- Add more logic in BitmapBinderTest testing `BindToType` to ensure deserialized bitmap is equivalent to original
- Update `BitmapBinder` to avoid return null ([ex](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1931519/?view=edit))
- Change to use fluent assertions for `BitmapBinderTests`
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11164)

### DIFF
--- a/src/System.Drawing.Common/src/System/Drawing/Font.cs
+++ b/src/System.Drawing.Common/src/System/Drawing/Font.cs
@@ -18,6 +18,7 @@ namespace System.Drawing;
 [Runtime.CompilerServices.TypeForwardedFrom(AssemblyRef.SystemDrawing)]
 public sealed unsafe class Font : MarshalByRefObject, ICloneable, IDisposable, ISerializable
 {
+    [NonSerialized]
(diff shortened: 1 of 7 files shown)


JeremyKuhne (approved):
Nice!

Tanya-Solyanik (approved):

PR #11165, opened 2024-04-05
Mark ComNativeDescriptor-related code as RequiresUnreferencedCode

sbomer:
Addresses some warnings in a winforms app by propagating RUC to code related to ComNativeDescriptor.
The app still produces one related warning for dotnet/runtime's reference (by name) to ComNativeDescriptor, which I plan to address with a feature switch. The code path is not needed in the trimmed test app.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11165)


LakshanF (approved):
LGTM, looking forward to the feature switch work to enable apps that don't use these types to be warning free.

PR #11172, opened 2024-04-06
Replace as few Not Inheritable Classes with Module

paul1956:
Replace  NotInheritable Classes with Module which has been done in a few other place is WinForms

CODE CLEANUP and Improve Test Coverage

## Proposed changes
Change 3 NotInheritable Classes with Module which has been done in a few other place is WinForms

## Customer Impact

None this helps develops by simplifying code

## Regression? 

- No

## Risk

-Minimal 

## Test methodology 
Code will not function; this is all covered by existing tests

Visual Basic ONLY
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11172)


paul1956:
I don't understand why 2 files were deleted from CodeCove report is that because they no longer have any code, if so this is ready to review.

PR #11176, opened 2024-04-08
Fix TableLayoutSettings.StyleConverter trim warnings

sbomer:
`[DynamicallyAccessedMembers(PublicConstructors)]` on `TableLayoutStyle` will treat the return value of `this.GetType()` inside of methods on the type as satisfying `PublicConstructors`, which is required by the `GetConstructor` call.

The helper method is needed because trim analysis doesn't treat `style.GetType()` the same way. See https://github.com/dotnet/runtime/issues/100786.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11176)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Panels/TableLayoutPanel/TableLayoutStyle.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Panels/TableLayoutPanel/TableLayoutStyle.cs
@@ -75,6 +75,7 @@ public abstract class TableLayoutStyle
         _size = size;
     }
 
+    // Workaround for https://github.com/dotnet/runtime/issues/100786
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Panels/TableLayoutPanel/TableLayoutStyle.cs
```diff
@@ -7,6 +7,7 @@
 namespace System.Windows.Forms;
 
 [TypeConverter(typeof(TableLayoutSettings.StyleConverter))]
+[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicConstructors)]
```
LakshanF, line 10: There are only 2 classes that extend this type and should not cause too much of code bloat with this annotation.

Issue #10244, opened 2023-11-07
System.Windows.Automation.InvokePattern.Invoke blocks on a menu item.


lovettchris:
### .NET version

7.0.403

### Did it work in .NET Framework?

Yes

### Did it work in any of the earlier releases of .NET Core or .NET 5+?

Don't know, it works in .NET Framework 4.6.2, but not in latest version of .NET Framework 4.8

### Issue description

Run the unit test in the following solution. Notice that the unit test hangs on line 80 of UnitTest1.cs which is the InvokePattern.Invoke call on the FIle/Open... menu item.  This makes it impossible to test any menu item using System.Automation?  

### Steps to reproduce

1. on Windows 11 22H2 build 22621.2361
2. Load this solution into VS 2022
3. Run Test/Debug all tests
4. wait for windows open file dialog to appear
5. Notice it never disappears
6. hit break in the debugger and you will see it is blocked on line 80 of UnitTest1.cs.

[WinFormsApp1.zip](https://github.com/dotnet/winforms/files/13289996/WinFormsApp1.zip)

elachlan:
@Olina-Zhang can your team please test this to confirm?

Amy-Li03:
@elachlan we can repro this issue on .NET 7.0, 8.0 and 9.0.
![Repro](https://github.com/dotnet/winforms/assets/62929087/1fa8a294-978a-429a-8acc-af3758db63c8)

lovettchris:
[like]  Chris Lovett reacted to your message:
________________________________

Epica3055:
I and @Amy-Li03 discussed this issue, and we think this is not a a11y issue.

merriemcgaw:
@Epica3055 let's discuss in the next meeting.

Tanya-Solyanik:
@Amy-Li03  - Could you please verify if this issue reproduces on 4.7.2 and 4.8.1 and if it depends on any AccessibilityImprovements.Level?

lovettchris:
Attached is a .NET Framework 4.7.2 version of the above test app and the test passes (the invoke does **not** block)
[WindowsFormsApp1.zip](https://github.com/dotnet/winforms/files/13359676/WindowsFormsApp1.zip)

If you then change the project properties to target .NET Framework 4.8, the test fails as per my original bug report.

Amy-Li03:
[...]

PR #11190, opened 2024-04-10
Add unit test for Invoke pattern

LeafShi1:
Related #10244

## Proposed changes

- Add unit tests for Controls that support InvokePattern

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11190)

### DIFF
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjects/ToolStripMenuItem.ToolStripMenuItemAccessibleObjectTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjects/ToolStripMenuItem.ToolStripMenuItemAccessibleObjectTests.cs
@@ -35,9 +35,10 @@ public class ToolStripMenuItem_ToolStripMenuItemAccessibleObjectTests
         callCount.Should().Be(1);
     }
 
+    [ActiveIssue("https://github.com/dotnet/winforms/issues/10244")]
     [WinFormsFact]
-    [SkipOnArchitecture(TestArchitectures.Any,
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjects/ToolStripMenuItem.ToolStripMenuItemAccessibleObjectTests.cs
```diff
@@ -18,6 +18,25 @@ public void ToolStripMenuItemAccessibleObject_Ctor_Default()
...
     }
 
+    [WinFormsFact]
+    public void ToolStripMenuItemAccessibleObject_Supports_InvokePattern()
+    {
+        int callCount = 0;
+        EventHandler handler = (sender, e) =>
+        {
+            // This action is temporarily commented due to issue #10244
+            // MessageBox.Show("...");
```
Tanya-Solyanik, line 28: How would you close this messagebox? You need the same test form as you have in the button test code sample that you show in the issue description

Tanya-Solyanik:
The change looks good, but tests are failing :(

Tanya-Solyanik (commented):
This is strange. We actually don't run tests on ARM, so Any and x86+x64 should be equivalent.

Tanya-Solyanik (approved):

LeafShi1:
> This is strange. We actually don't run tests on ARM, so Any and x86+x64 should be equivalent.

The log of PR test failure does not have detailed reasons, so it is impossible to determine where the problem occurred. I just saw that only the X86 and X64 Flags were judged in the `SkipOnArchitectureDiscoverer` class, and TestArchitectures.Any has never been used, so replace Any with X86+X64

PR #11196, opened 2024-04-10
Fix some record issues

JeremyKuhne:
- BinaryArray was not putting itself in the record map
- Id negative and positive values are not equivalent
- Add "null" Id
- Add base ObjectRecord to surface ObjectId
- Handle BinaryLibrary records occuring before other record data
- Raise LibraryId to ClassRecord
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11196)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/Record.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/Record.cs
@@ -453,11 +453,29 @@ internal abstract class Record : IRecord
         BinaryReader reader,
         RecordMap recordMap,
         BinaryType type,
-        object? typeInfo) => type switch
+        object? typeInfo)
+    {
+        if (type == BinaryType.Primitive)
         {
-            BinaryType.Primitive => ReadPrimitiveType(reader, (PrimitiveType)typeInfo!),
-            BinaryType.String
-                or BinaryType.Object
+            return ReadPrimitiveType(reader, (PrimitiveType)typeInfo!);
+        }
+
+        if (type == BinaryType.String)
+        {
+            return ReadBinaryFormatRecord(reader, recordMap);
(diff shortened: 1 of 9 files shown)


JeremyKuhne:
@adamsitnik fyi

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/Record.cs
```diff
@@ -453,18 +453,37 @@ private protected static void WriteRecords(BinaryWriter writer, IReadOnlyList<ob
...
+        {
+            return ReadPrimitiveType(reader, (PrimitiveType)typeInfo!);
+        }
+
+        if (type == BinaryType.String)
+        {
+            return ReadBinaryFormatRecord(reader, recordMap);
+        }
+
+        // BinaryLibrary records can be dumped in front of any member reference.
```
Tanya-Solyanik, line 468: What is a binary library?

Tanya-Solyanik: Would we read a BinaryLibrary when type is set to `object`?

Tanya-Solyanik (approved):
Looks reasonable, but I'm not familiar with this format, would be good to get someone else's eyes on it.

src/System.Windows.Forms.Primitives/src/System/Windows/Forms/BinaryFormat/Record.cs
JeremyKuhne: Binary library records are information records that type information records point to. There is no type info record for object, members may be of type object, and there is a special member info type for it.

There are links to the spec in each of the record classes.

PR #11211, opened 2024-04-14
Enable nullability in DesignerToolStripControlHost

gpetrou:
## Proposed changes

- Enable nullability in DesignerToolStripControlHost.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11211)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DesignerToolStripControlHost.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DesignerToolStripControlHost.cs
@@ -1,8 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-#nullable disable
-
 using System.ComponentModel;
 using System.Drawing;
 using System.Windows.Forms.Design.Behavior;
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/DesignerToolStripControlHost.cs
```diff
@@ -32,7 +31,20 @@ protected override Size DefaultSize
...
-        _behaviorService ??= (BehaviorService)parent.Site.GetService(typeof(BehaviorService));
+        if (_behaviorService is null)
+        {
+            if (!parent.Site.TryGetService(out BehaviorService? behaviorService))
+            {
+                throw new InvalidOperationException();
+            }
+
+            _behaviorService = behaviorService;
+        }
```
lonitra, line 42: I believe we have `GetRequiredService` available here
```suggestion
        _behaviorService ??= parent.Site.GetRequiredService<BehaviorService>();
```

lonitra (commented):
Is it possible to add some test coverage here?

gpetrou:
> Is it possible to add some test coverage here?

Hmm, how is GetGlyphs being used? I don't see any usages.

Tanya-Solyanik:
Ultimately this method is used when a toolstrip on the design surface gets selected to draw the overlay glyphs on top of toolstrip items. Maybe we hadn't ported the relevant designers completely. ToolStrip designer is referenced by multiple other components and initially was ported only to support other designers. Even if this method is not used right now, I would keep it in case we enable the full scenario eventually.

lonitra:
If we hadn't ported relevant designers completely we can wait until then to add tests. @gpetrou please address feedback above about `GetRequiredService` and we can get this to completion.

lonitra (approved):
Thank you

PR #11213, opened 2024-04-14
fixed Font auto scaling when Handle is created

### LINKED ISSUE #11214: Font autoscaling doesn't function in .NET Core as it does in .NET Framework 4.8.
### Description

I've just migrated a Winform project from .NET Framework 4.8 to .NET Core 9, following the porting instructions provided in the Windows documentation. However, I've encountered an issue with font autoscaling. When I adjust the font for the entire application, the controls on the form don't scale appropriately.

### Reproduction Steps

1. Create small Multi-windows winform application (an app which can open multiple forms in the same app) based on .NET Framework 4.8 having a single form with basic ui control like labels, buttons, textboxes.
2. Create a form of changing font of entire application in the application.
3. Port above .NET 4.8 project to using .NET core with latest version (Also, keep copy of .Net 4.8 version)
4. Run both project, open a form, observe controls properties before and after changing font size.
5. For reference, .NET 4.8 uses default 'Microsoft san serif' with 8.25 Small font size and .Net core uses 'Segoe UI' with 9 small font size.

### Expected behavior

In a NET Core project, after changing the font, the properties of the controls like SizeX, SizeY, LocationX, and LocationY should be set to the same values as in the NET 4.8 project during events like onLoad, onShown, and FontChanged. 

I have attached my application expected screenshot.
Correct

### Actual behavior

In a NET Core project these properties are not getting set correctly, which ultimately causing winform UI to set weirdly.

I have attached my application actual screenshot.
weird

### Regression?

_No response_

### Known Workarounds

4 8 code

Above is NET 4.8 code from System.Windows.Forms->Systems->Windows->Forms->Layout->Containers->ContainerControls.cs

and below is the same code from NET Core

Core code

Look at difference, because of base.IsHandleCreated NET Core doesn't allow to take effect of FontChanged.
if we remove that ` base.IsHandleCreated` from code and test the application it works fine.

### Configuration

.NET version- 4.8 and .NET Core
Windows 10
x64 Architecture

### Other information

_No response_

swapso:
Fixes #11214 

Link for issue:  https://github.com/dotnet/winforms/issues/11214

## Proposed changes

- Updated if condition to bypass base.IsHandleCreated

## Regression? 

- No

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11213)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Layout/Containers/ContainerControl.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Layout/Containers/ContainerControl.cs
@@ -860,7 +860,7 @@ public class ContainerControl : ScrollableControl, IContainerControl
         // to scale in Dpi mode (during WM_DPICHANGED event).
         // This may require scaling/relayout of the form. AutoScaleFactor will take
         // AutoScaleMode into account while scaling the controls.
-        if (AutoScaleMode != AutoScaleMode.None && IsHandleCreated)
+        if ((AutoScaleMode != AutoScaleMode.None && IsHandleCreated) || AutoScaleMode is AutoScaleMode.Font or AutoScaleMode.Inherit)
         {
             _currentAutoScaleDimensions = SizeF.Empty;
 


elachlan:
@swapso can you update the PR description with the correct issue ID?

swapso:
> @swapso can you update the PR description with the correct issue ID?

Hi @elachlan I have updated correct Issue ID.

merriemcgaw:
@swapso thank you for raising this issue. A change in this part of the code is extremely risky and can have a lot of unintended consequences. @LeafShi1 - can your team do a deep dive into a workaround, perhaps enforcing `SuspendLayout` when handle is created?

LeafShi1:
Track this issue with PR https://github.com/dotnet/winforms/pull/11641

PR #11216, opened 2024-04-15
Enable nullability in ItemTypeToolStripMenuItem and StandardCommandToolStripMenuItem

gpetrou:
## Proposed changes

- Enable nullability in ItemTypeToolStripMenuItem and StandardCommandToolStripMenuItem.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11216)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ItemTypeToolStripMenuItem.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ItemTypeToolStripMenuItem.cs
@@ -11,7 +11,6 @@ namespace System.Windows.Forms.Design;
 /// </summary>
 internal class ItemTypeToolStripMenuItem : ToolStripMenuItem
 {
-    private static readonly string s_systemWindowsFormsNamespace = typeof(ToolStripItem).Namespace!;
     private static readonly ToolboxItem s_invalidToolboxItem = new();
     private readonly Type _itemType;
     private bool _convertTo;
@@ -19,8 +18,6 @@ internal class ItemTypeToolStripMenuItem : ToolStripMenuItem
     private Image? _image;
 
     public ItemTypeToolStripMenuItem(Type t) => _itemType = t;
-
-    public static string SystemWindowsFormsNamespace => s_systemWindowsFormsNamespace;
 
     public Type ItemType
     {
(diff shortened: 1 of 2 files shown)


LeafShi1 (commented):
LGTM! @lonitra Please review this PR further.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ItemTypeToolStripMenuItem.cs
```diff
@@ -13,12 +11,12 @@ namespace System.Windows.Forms.Design;
 /// </summary>
 internal class ItemTypeToolStripMenuItem : ToolStripMenuItem
 {
-    private static readonly string s_systemWindowsFormsNamespace = typeof(ToolStripItem).Namespace;
+    private static readonly string s_systemWindowsFormsNamespace = typeof(ToolStripItem).Namespace!;
```
lonitra, line 14: This doesn't seem to be used, could we remove this?

lonitra (commented):
Minor comment otherwise LGTM 👍

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ItemTypeToolStripMenuItem.cs
gpetrou: This is used in SystemWindowsFormsNamespace public property below.

lonitra: Believe we can remove both then since this class is internal and `SystemWindowsFormsNamespace` has no references afaict

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ItemTypeToolStripMenuItem.cs
```diff
@@ -48,21 +43,25 @@ public override Image Image
...
         {
         }
     }
 
-    public ToolboxItem TbxItem { get => _tbxItem; set => _tbxItem = value; }
+    public ToolboxItem TbxItem
+    {
+        get => _tbxItem;
+        set => _tbxItem = value;
+    }
```
lonitra, line 58: Missed this when reviewing the first time , but is this backing field needed? If not, let's make this into an auto prop.

lonitra (approved):
Thank you!

PR #11230, opened 2024-04-17
Disable binary formatter for trimming

LakshanF:
Added a feature switch check with the default set to true (enable using `BinaryFormatter` for compatibility). The feature switch flag is used so as to not generate a trim warning when the flag is false, which is the expectation in trimming. The trimmer will generate a warning if the feature is enabled to use the `BinaryFormatter`.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11230)


lonitra (commented):
couple minor comments but 👍

lonitra (approved):

PR #11250, opened 2024-04-22
Move the BinaryFormat code to the Private assembly

JeremyKuhne:
It has no dependencies on the current level of the stack and should be where it can be commonly used.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11250)


lonitra (approved):
:shipit:

PR #11253, opened 2024-04-23
fix issue 11243 TreeNodeCollection.AddRange has been broken in .NET 8 and no longer results in TreeNodes being drawn at the correct location

### LINKED ISSUE #11243: TreeNodeCollection.AddRange has been broken in .NET 8 and no longer results in TreeNodes being drawn at the correct location
### .NET version

8.0

Yes

.NET 7 and previous

### Issue description

### Steps to reproduce

image

image

Epica3055:
Fixes #11243 

### Before
image

### After
image

## Test methodology 

- 
-  manually 
- 

## Test environment(s) 
9.0.0-preview.4.24219.3

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNode.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNode.cs
@@ -1312,7 +1312,7 @@ public partial class TreeNode : MarshalByRefObject, ICloneable, ISerializable
 
             while (_childCount > 0)
             {
-                _children[^1].Remove(true);
+                _children[_childCount - 1].Remove(true);
             }
 
             _children = [];
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNode.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNode.cs
@@ -644,7 +644,7 @@ public partial class TreeNode : MarshalByRefObject, ICloneable, ISerializable
                 return null;
             }
 
-            return _children[^1];
+            return _children[_childCount - 1];
         }
     }
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNode.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNode.cs
@@ -89,6 +89,7 @@ public partial class TreeNode : MarshalByRefObject, ICloneable, ISerializable
 
     internal int _index;                  // our index into our parents child array
     internal int _childCount;
+    // this array should not be optimized as a list because we are inserting into the middle of it, not appending.
     internal TreeNode[] _children = [];
     internal TreeNode? _parent;
     internal TreeView? _treeView;
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNode.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNode.cs
@@ -1198,8 +1198,6 @@ public partial class TreeNode : MarshalByRefObject, ICloneable, ISerializable
             CompareInfo compare = Application.CurrentCulture.CompareInfo;
             for (int i = 0; i < _childCount; i++)
             {
-                // newOrder.Add(children[i]);
-
                 int min = -1;
                 for (int j = 0; j < _childCount; j++)
                 {
@@ -1234,8 +1232,6 @@ public partial class TreeNode : MarshalByRefObject, ICloneable, ISerializable
             IComparer sorter = parentTreeView.TreeViewNodeSorter;
             for (int i = 0; i < _childCount; i++)
             {
-                // newOrder.Add(children[i]);
-
                 int min = -1;
                 for (int j = 0; j < _childCount; j++)
                 {
@@ -1595,7 +1591,7 @@ public partial class TreeNode : MarshalByRefObject, ICloneable, ISerializable
     }
 
     /// <summary>
-    ///  Makes sure there is enough room to add n children
+    ///  Makes sure there is enough room to add <paramref name="num" /> children.
     /// </summary>
     internal void EnsureCapacity(int num)
     {
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNode.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNode.cs
@@ -88,7 +88,8 @@ public partial class TreeNode : MarshalByRefObject, ICloneable, ISerializable
     }
 
     internal int _index;                  // our index into our parents child array
-    internal List<TreeNode> _childNodes = [];
+    internal int _childCount;
+    internal TreeNode[] _children;
     internal TreeNode? _parent;
     internal TreeView? _treeView;
     private bool _expandOnRealization;
@@ -372,7 +373,7 @@ public partial class TreeNode : MarshalByRefObject, ICloneable, ISerializable
     ///  The first child node of this node.
     /// </summary>
     [Browsable(false)]
-    public TreeNode? FirstNode => _childNodes.Count == 0 ? null : _childNodes[0];
+    public TreeNode? FirstNode => _childCount == 0 ? null : _children[0];
 
     private TreeNode? FirstVisibleParent
     {
@@ -638,12 +639,12 @@ public partial class TreeNode : MarshalByRefObject, ICloneable, ISerializable
     {
         get
         {
-            if (_childNodes.Count == 0)
+            if (_childCount == 0)
             {
                 return null;
             }
 
-            return _childNodes[^1];
+            return _children[^1];
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNode.cs
```diff
@@ -88,7 +88,8 @@ internal TreeNodeImageIndexer StateImageIndexer
     }
 
     internal int _index;                  // our index into our parents child array
-    internal List<TreeNode> _childNodes = [];
+    internal int _childCount;
+    internal TreeNode[] _children;
```
elachlan, line 92: `error CS8618: Non-nullable field '_children' must contain a non-null value when exiting constructor. Consider declaring the field as nullable`

Tanya-Solyanik:
@Epica3055 , is this a straight revert of the https://github.com/dotnet/winforms/pull/9078 PR? It would be good to understand what specific change introduced the regression. Once we know it, it will be easier to decide on the direction of the fix.

src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNode.cs
```diff
@@ -1187,23 +1187,23 @@ internal int AddSorted(TreeView parentTreeView, TreeNode node)
...
-        List<TreeNode> newOrder = new(_childNodes.Count);
+        TreeNode[] newOrder = new TreeNode[_childCount];
         if (parentTreeView is null || parentTreeView.TreeViewNodeSorter is null)
         {
             CompareInfo compare = Application.CurrentCulture.CompareInfo;
-            for (int i = 0; i < _childNodes.Count; i++)
+            for (int i = 0; i < _childCount; i++)
             {
-                newOrder.Add(_childNodes[i]);
+                // newOrder.Add(children[i]);
```
Tanya-Solyanik, line 1201: Is this an experiment or part of the change?

src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNode.cs
```diff
@@ -1214,32 +1214,32 @@ private void SortChildren(TreeView? parentTreeView)
...
+            _children = newOrder;
         }
         else
         {
             IComparer sorter = parentTreeView.TreeViewNodeSorter;
-            for (int i = 0; i < _childNodes.Count; i++)
+            for (int i = 0; i < _childCount; i++)
             {
-                newOrder.Add(_childNodes[i]);
+                // newOrder.Add(children[i]);
```
Tanya-Solyanik, line 1237: Is this an experiment or part of the change?

src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNode.cs
```diff
@@ -1250,20 +1250,20 @@ private void SortChildren(TreeView? parentTreeView)
                         continue;
                     }
 
-                    if (sorter.Compare(_childNodes[j] /*previous*/, _childNodes[min] /*current*/) <= 0)
+                    if (sorter.Compare(_children[j] /*previous*/, _children[min] /*current*/) <= 0)
```
Tanya-Solyanik, line 1253:
```suggestion
                    if (sorter.Compare(previous: _children[j], current: _children[min]) <= 0)
```
(if these are indeed the parameter names)

src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNode.cs
```diff
@@ -1595,6 +1594,36 @@ public void EndEdit(bool cancel)
         PInvoke.SendMessage(tv, PInvoke.TVM_ENDEDITLABELNOW, (WPARAM)(BOOL)cancel);
     }
 
+    /// <summary>
+    ///  Makes sure there is enough room to add n children
```
Tanya-Solyanik, line 1598:
```suggestion
    ///  Makes sure there is enough room to add <paramref name="num" /> children.
```

ALso consider renaming num to `delta`, or something else that indicates an addition.

src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNode.cs
```diff
@@ -1157,10 +1157,10 @@ internal int AddSorted(TreeView parentTreeView, TreeNode node)
             {
                 IComparer sorter = parentTreeView.TreeViewNodeSorter;
                 // Insert at appropriate sorted spot
-                for (iMin = 0, iLim = nodeCount; iMin < iLim;)
+                for (iMin = 0, iLim = _childCount; iMin < iLim;)
                 {
                     iT = (iMin + iLim) / 2;
-                    if (sorter.Compare(_childNodes[iT] /*previous*/, node/*current*/) <= 0)
+                    if (sorter.Compare(_children[iT] /*previous*/, node/*current*/) <= 0)
```
Tanya-Solyanik, line 1163: use argument names instead of comments

src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNodeCollection.cs
```diff
@@ -219,9 +219,9 @@ public virtual void AddRange(params TreeNode[] nodes)
...
         }
 
-        _owner.Nodes.FixedIndex = _owner._childNodes.Count;
-        _owner._childNodes.EnsureCapacity(nodes.Length);
-        for (int i = 0; i < nodes.Length; i++)
+        _owner.Nodes.FixedIndex = _owner._childCount;
+        _owner.EnsureCapacity(nodes.Length);
+        for (int i = nodes.Length - 1; i >= 0; i--)
         {
             AddInternal(nodes[i], i);
```
Tanya-Solyanik, line 226:
```suggestion
            AddInternal(nodes[i], delta: i);
```

Tanya-Solyanik:
Please add a dedicated test for this scenario - call AddRange twice

gpetrou:
Please check https://github.com/dotnet/winforms/issues/11243#issuecomment-2087962400. Also, even if there is still a problem, can't we still use a List instead of an array?

Tanya-Solyanik:
> Please check [#11243 (comment)](https://github.com/dotnet/winforms/issues/11243#issuecomment-2087962400). Also, even if there is still a problem, can't we still use a List instead of an array?

Yes, this is a stable repro, when you call AddRange after the native tree view has been populated.

This code references a perf optimization for cases when we add large ranges to the treeview, and it seems that the workaround is to insert after the "fixed" node - the last node as previously populated treeview, as opposed to appending after the actual last node. To preserve this workaround as is, we would be inserting into the managed collection of child nodes at the "fixed index" as well, instead of appending, this seems complicated and an array is better suited for such use. To re-do this optimization, we need the specific scenario, Epica has experimented with large ranges and didn't see any difference with or without this perf optimization, I guess that scenario requires image lists, or other complications. But with no understanding of what this code is optimizing, it's risky to remove or modify it.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TreeViewTests.cs
```diff
@@ -7305,6 +7305,55 @@ public void Verify_NodeValue_AfterSortAndRemove()
...
+        TreeNode treeNode2 = new("b1");
+        TreeNode treeNode3 = new("c1");
+        TreeNode rootNode = new("Root", [new TreeNode("child")]);
+
+        treeView.Nodes.Add(rootNode);
+        form.Controls.Add(treeView);
+
+        form.Load += (s, e) =>
+        {
+            treeView.Nodes[0].Nodes.AddRange(treeNode1, treeNode2, treeNode3);
```
Tanya-Solyanik, line 7333: use `rootNode` local variable instead of the item accessor.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TreeViewTests.cs
```diff
@@ -7305,6 +7305,55 @@ public void Verify_NodeValue_AfterSortAndRemove()
...
+            treeView.Nodes[0].ExpandAll();
+        };
+        form.Show();
+
+        TreeNode childNode1 = treeView.Nodes[0].Nodes[0];
```
Tanya-Solyanik, line 7338: Verify that `treeView.Nodes[0]` is still the `rootNode` and then use `rootNode` local variable instead of the collection item accessor

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TreeViewTests.cs
Tanya-Solyanik, line 7333: Please add 4 nodes to force collection resize, as far as I remember initial size is 4, and we are adding exactly 4 nodes conting `child`

src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNode.cs
```diff
@@ -88,7 +88,8 @@ internal TreeNodeImageIndexer StateImageIndexer
     }
 
     internal int _index;                  // our index into our parents child array
-    internal List<TreeNode> _childNodes = [];
+    internal int _childCount;
+    internal TreeNode[] _children = [];
```
Tanya-Solyanik, line 92: Please add a comment that this array should not be optimized as a list because we are inserting into the middle of it, not appending.

Tanya-Solyanik (commented):
Added some minor comments, and please address the previous set of comments

src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNode.cs
Epica3055: actually they are not.

src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNode.cs
Epica3055: they are not argument names

Tanya-Solyanik (approved):
Thank you!

PR #11256, opened 2024-04-23
Update BinaryFormattedObject

JeremyKuhne:
This change updated BinaryFormattedObject to:

1. Support the ClassWithMembers and SystemClassWithMembers records correctly
2. Change to object? instead of object with ObjectNull records
3. Optimize reading all primitive array types
4. Always leave the incoming stream open
5. Support multidimensional and jagged arrays
6. Move some properties higher for simpler handling

Also made two small tweaks to tests. Added Skip= to one that hangs to keep it from running in Test Explorer. Also reduced the matrix for one exceptionally long running test.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11256)

### DIFF
--- a/src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/BinaryArray.cs
+++ b/src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/BinaryArray.cs
@@ -1,6 +1,8 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Runtime.Serialization;
+
 namespace System.Windows.Forms.BinaryFormat;
 
 /// <summary>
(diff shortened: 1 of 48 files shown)


src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/BinaryFormattedObject.ParseState.cs
```diff
@@ -0,0 +1,36 @@
...
+        public ParseState(BinaryReader reader, BinaryFormattedObject format)
+        {
+            Reader = reader;
+            _format = format;
+        }
+
+        public BinaryReader Reader { get; }
+        public RecordMap RecordMap => _format._recordMap;
+
+        [UnconditionalSuppressMessage(
```
JeremyKuhne, line 24: @sbomer @LakshanF I flowed the rest of the GetType related trimming attributes to this, more contained, spot to reduce the risk accidentally adding calls that could otherwise carry context or flow to the user in actionable ways.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/BinaryArray.cs
```diff
@@ -13,54 +15,141 @@ namespace System.Windows.Forms.BinaryFormat;
...
+        state.RecordMap[objectId] = array;
+        return array;
     }
 
-    public override void Write(BinaryWriter writer)
+    /// <summary>
+    ///  <see cref="BinaryArray"/> of objects.
+    /// </summary>
+    /// <inheritdoc cref="BinaryArray"/>
+    private sealed class ObjectBinaryArray : ArrayRecord<object?>, IRecord<ObjectBinaryArray>, IBinaryArray
```
lonitra, line 128: Should this be in its own file?

lonitra (approved):
👍 minor comments

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/BinaryArray.cs
JeremyKuhne: Yes, whoops. I'll catch it in the next PR.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/BinaryArray.cs
```diff
@@ -13,54 +15,141 @@ namespace System.Windows.Forms.BinaryFormat;
...
+        }
+
+        int rank = state.Reader.ReadInt32();
 
-        if (arrayType != BinaryArrayType.Single || rank != 1)
+        if (arrayType is BinaryArrayType.Single or BinaryArrayType.Jagged)
+        {
+            // Jagged array is an array of arrays, there should always be one rank
+            // for the "outer" array.
+            if (rank != 1)
```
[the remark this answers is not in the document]
JeremyKuhne: Want it not to have to check arrayType twice.

Issue #11232, opened 2024-04-18
Winforms VB application with Single Instance setting crashes for multiple users on terminal server


Olina-Zhang:
### .NET version

.Net 9.0 SDK: 9.0.100-preview.4.24217.27

### Did it work in .NET Framework?

Yes

### Did it work in any of the earlier releases of .NET Core or .NET 5+?

Yes, repro in .Net 8

### Issue description

Creating a Winforms .Net 9 or 8 VB application with setting "single instance application" to true, put it on a terminal server, more users with permission to work on this server using RDP. The first user starts the program just fine, the second user start the same app and it crashes on startup. 

Here is the log when it is reproduced:
```
Application: VBSingleInstance_NET9.exe
CoreCLR Version: 9.0.24.21505
.NET Version: 9.0.0-preview.4.24215.5
Description: The process was terminated due to an unhandled exception.
Exception Info: Microsoft.VisualBasic.ApplicationServices.CantStartSingleInstanceException: This single-instance application could not connect to the original instance.
   at Microsoft.VisualBasic.ApplicationServices.WindowsFormsApplicationBase.Run(String[] commandLine)
   at VBSingleInstance_NET9.My.MyApplication.Main(String[] Args) in :line 83
```

### Steps to reproduce

1. Create a Winforms VB .NET application
2. In Solution Explorer window, show all files, find Application.Designer.vb file, set IsSingleInstance to true
![image](https://github.com/dotnet/winforms/assets/26474449/33ab17a5-3abe-4e2a-b8ba-1980106fb482)
3. Build this application
4. Put this applicaiton to a prepared server machine with multiple users access
5. First user to log in this terminal server machine by RDP, launch this application
6. Second user to log in this terminal server machine by RDP, launch this application too

**MoreInfo:** this is the customer feedback: https://developercommunity.visualstudio.com/t/NET-8-single-instance-winforms-crashes-/10634133

elachlan:
Duplicate of #3715

elachlan:
@Olina-Zhang the exception is probably caused by the fact that the named pipe uses `PipeOptions.CurrentUserOnly`. So it can't connect to other users named pipes. Which is what we want. My suggested fix in #11258 should mitigate this issue.

PR #11258, opened 2024-04-23
Fix WindowsFormsApplicationBase.IsSingleInstance

### LINKED ISSUE #3715: WindowsFormsApplicationBase.IsSingleInstance applications are global on a machine, not per user
* .NET Core Version: 
v5.0.0-preview.7

No

**Problem description:**
The implementation of `WindowsFormsApplicationBase.IsSingleInstance` from #3200 uses a named pipe as its synchronization and communication tool, but named pipes are in a global namespace (at least on Windows) and thus any assembly can only be running once on any given machine.

**Expected behavior:**
Able to start this executable in multiple user accounts (under terminal services or fast user switching) at once.

**Actual behaviour:**
Starting the executable in a second user account produces an uncaught `CantStartSingleInstanceException` exception.

Strictly speaking this is a potential denial of service issue; by using a named pipe with a predictable name, users can prevent others on the same machine from ever starting the application. Adding the session identifier to the pipe name is enough to make the named pipe session-local, but not enough to prevent others from creating a pipe with the same name.

I don't know what other options there are; mutexes and event handles have a session namespace, but don't allow for data transfer. Local TCP runs into trouble with firewalls (as the old Remoting implementation did). Cursory reading shows that named memory mappings are Windows only, and require marshalling/synchronisation to a much greater degree, though perhaps a memory mapping containing only the randomly-generated named pipe name is the way to go (like [WCF](https://rodneyviana.com/named-pipes-in-wcf-are-named-but-not-by-you-and-how-to-find-the-actual-windows-object-name/)).

**Minimal repro:**
Borrowed from https://stackoverflow.com/a/19326/125549:

```C#
public class SingleInstanceApplication : System.Windows.Application
{
    protected override void OnStartup(System.Windows.StartupEventArgs e)
    {
        // Call the OnStartup event on our base class
        base.OnStartup(e);

        // Create our MainWindow and show it
        MainWindow window = new MainWindow();
        window.Show();
    }

    public void Activate()
    {
        // Reactivate the main window
        MainWindow.Activate();
    }
}

public class SingleInstanceManager : Microsoft.VisualBasic.ApplicationServices.WindowsFormsApplicationBase
{
    private SingleInstanceApplication _application;
    private System.Collections.ObjectModel.ReadOnlyCollection<string> _commandLine;

    public SingleInstanceManager()
    {
        IsSingleInstance = true;
    }

    protected override bool OnStartup(Microsoft.VisualBasic.ApplicationServices.StartupEventArgs eventArgs)
    {
        // First time _application is launched
        _commandLine = eventArgs.CommandLine;
        _application = new SingleInstanceApplication();
        _application.Run();
        return false;
    }

    protected override void OnStartupNextInstance(StartupNextInstanceEventArgs eventArgs)
    {
        // Subsequent launches
        base.OnStartupNextInstance(eventArgs);
        _commandLine = eventArgs.CommandLine;
        _application.Activate();
    }
}

public class EntryPoint
{
    [STAThread]
    public static void Main(string[] args)
    {
        SingleInstanceManager manager = new SingleInstanceManager();
        manager.Run(args);
    }
}
```

elachlan:
Add currentUserSID to `GetApplicationInstanceID` for use in `TryCreatePipeServer` to avoid system wide blocking. Used in conjunction with `PipeOptions.CurrentUserOnly`.

Fixes #3715

### DIFF
--- a/src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/ApplicationServices/WindowsFormsApplicationBase.vb
+++ b/src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/ApplicationServices/WindowsFormsApplicationBase.vb
@@ -1006,28 +1006,26 @@ Namespace Microsoft.VisualBasic.ApplicationServices
 
         ''' <summary>
         ''' Generates the name for the remote singleton that we use to channel multiple instances
-        ''' to the same application model thread.
+        ''' to the same application model thread for the current user.
         ''' </summary>
         ''' <returns>A string unique to the application that should be the same for versions of
         '''  the application that have the same Major and Minor Version Number
         ''' </returns>
         ''' <remarks>If GUID Attribute does not exist fall back to unique ModuleVersionId</remarks>
         Private Shared Function GetApplicationInstanceID(ByVal Entry As Assembly) As String
+            Dim guidAttribute As GuidAttribute = Entry.GetCustomAttribute(Of GuidAttribute)()
+            Dim currentUserSID As String = Principal.WindowsIdentity.GetCurrent().User.Value
 
-            Dim guidAttrib As GuidAttribute = Entry.GetCustomAttribute(Of GuidAttribute)()
-
-            If guidAttrib IsNot Nothing Then
-
+            If guidAttribute IsNot Nothing Then
                 Dim version As Version = Entry.GetName.Version
-
                 If version IsNot Nothing Then
-                    Return $"{guidAttrib.Value}{version.Major}.{version.Minor}"
+                    Return $"{guidAttribute.Value}{version.Major}.{version.Minor}-{currentUserSID}"
                 Else
-                    Return guidAttrib.Value
+                    Return $"{guidAttribute.Value}-{currentUserSID}"
                 End If
             End If
 
-            Return Entry.ManifestModule.ModuleVersionId.ToString()
+            Return $"{Entry.ManifestModule.ModuleVersionId}-{currentUserSID}"
         End Function
     End Class
 End Namespace


elachlan:
@weltkante had this suggestion and I ran with it. I think the concerns raised by @zanchey are technically valid, but block a pragmatic fix for most use cases. It might be mitigated by us using `PipeOptions.CurrentUserOnly` as well.

@Olina-Zhang could your team test this? I don't have a multi-user environment I can easily utilize for testing.

paul1956 (changes requested):
Can you remove ByVal and make private function parameters start with lower case character.  If touching code, it should conform to current VB Style otherwise it will conflict with other PR's trying to clean up VB Code Style.

Olina-Zhang:
@elachlan tested this PR using following 2 apps(one is Winforms VB app, another is Winforms C#) in .Net 9.0 SDK build on a server with multi-users. Every logged in user can launch the app, issue is fixed. Thanks!
[VB_SingleInstance_true.zip](https://github.com/dotnet/winforms/files/15087192/VB_SingleInstance_true.zip)
[SingleInstanceTrue_CSharp.zip](https://github.com/dotnet/winforms/files/15087194/SingleInstanceTrue_CSharp.zip)

src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/ApplicationServices/WindowsFormsApplicationBase.vb
```diff
@@ -1006,28 +1008,37 @@ Namespace Microsoft.VisualBasic.ApplicationServices
...
                 End If
             End If
 
-            Return Entry.ManifestModule.ModuleVersionId.ToString()
+            Return $"{pathGuid}-{entry.ManifestModule.ModuleVersionId}-{currentUserSID}"
+        End Function
+
+#Disable Warning CA5351 ' We want speed over collision resistance.
+        Private Shared Function GetFilePathAsGuid(filePath As String) As Guid
+            Using md5 As MD5 = MD5.Create()
```
elachlan, line 1036: Is `MD5` okay here? really we are just taking a path and making a guid. I don't expect it to be an issue. Because its in the startup path I believe its quite important to have a fast hashing algorithm here as well.

elachlan:
I have some concerns about using the filepath of the assembly because we are calling it using `Assembly.GetCallingAssembly()`.

This needs to be tested with self-contained and AOT scenarios. As far as I know for self contained applications I've had to use `Assembly.GetEntryAssembly()` to get the correct location of where the exe is located. If `Assembly.GetCallingAssembly()` is used it might get the location of the assembly that is extracted to the users appdata.

I am unsure if this is still the case.

@Olina-Zhang maybe your team could test this scenario?

Olina-Zhang:
> @Olina-Zhang maybe your team could test this scenario?

Use your new updated code to test these issues again, right?

elachlan:
The main issue to test is #11365 using the updated code.

Then additionally try #11365 but on a single file published app.

Olina-Zhang:
Hi @elachlan, tested the updated code, here is the result:
1. GH issue https://github.com/dotnet/winforms/issues/3715 and GH issue https://github.com/dotnet/winforms/issues/11232 -  Fixed
2. For GH issue: https://github.com/dotnet/winforms/issues/11365, the scenarios customer reported about launch exe from 2 different folders with one user - Fixed, we can see 2 app instances
3. For GH issue: https://github.com/dotnet/winforms/issues/11365, copy it on a server machine in 2 different folders, then launch exe from 2 different folders with different users - Launch exe successfully with different users
4. For GH issue: https://github.com/dotnet/winforms/issues/11365, published app with self-contained, copy it on a server machine in one folder, then launch exe with different users --- **first user can install and launch, second user can install but cannot launch,** get following exception in EventViewer:
```
Application: WinFormsApp1.exe
CoreCLR Version: 9.0.24.25601
.NET Version: 9.0.0-preview.5.24256.1
Description: The process was terminated due to an unhandled exception.
Exception Info: Microsoft.VisualBasic.ApplicationServices.CantStartSingleInstanceException: This single-instance application could not connect to the original instance.
   at Microsoft.VisualBasic.ApplicationServices.WindowsFormsApplicationBase.Run(String[] commandLine)
   at WinFormsApp1.Program.Main()


```

elachlan:
That last scenario with published self-contained app and two users is interesting and unexpected.

We will need to investigate further.

src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/ApplicationServices/WindowsFormsApplicationBase.vb
weltkante: that depends on whether you are treating the input (the file path) as a secret and/or want to prevent colissions with forged inputs - md5 is fine against accidental colissions - personally I don't think there's a security issue here, and if it were, then it needs to be adressed at the permission level since the named objects are visible and accessible to non-dotnet code as well

elachlan:
@Olina-Zhang can you retest the published self-contained failure scenario with the latest? The event viewer will have a bit more detail hopefully.

Olina-Zhang:
> @Olina-Zhang can you retest the published self-contained failure scenario with the latest? The event viewer will have a bit more detail hopefully.

I didn't see any difference with updated code, same as Yesterday's:
image

elachlan:
My change was to pass in the `GetApplicationInstanceID` value to the exception so we could see what it was. But it seems not to have worked.

elachlan:
@Tanya-Solyanik can someone please revisit this? I think most of the leg work is done.

src/Microsoft.VisualBasic.Forms/src/Microsoft/VisualBasic/ApplicationServices/WindowsFormsApplicationBase.vb
A9G-Data-Droid: MD5 is not okay, in general, because it is blocked by setting FIPS mode at the OS level. FIPS mode can't know that you are using MD5 in a non-security context, it just blocks any application using that algorithm. This is a pain for everyone in the government space.

elachlan: What would be appropriate then?

A9G-Data-Droid: 'SHA1()' is going to be the least expensive option. Looks like they will allow it until 2030, at which point SHA2 will be required.

elachlan: I suggest using one of these:
https://learn.microsoft.com/en-us/dotnet/api/system.io.hashing

PR #11266, opened 2024-04-24
fix issue 11243 TreeNodeCollection.AddRange has been broken in .NET 8 and no longer results in TreeNodes being drawn at the correct location

### LINKED ISSUE #11243: TreeNodeCollection.AddRange has been broken in .NET 8 and no longer results in TreeNodes being drawn at the correct location

Epica3055:
Fixes #11243 

### Before
image

### After
image

## Test methodology 

- 
-  manually 
- 

## Test environment(s) 
9.0.0-preview.4.24219.3

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNode.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNode.cs
@@ -810,14 +810,7 @@ public partial class TreeNode : MarshalByRefObject, ICloneable, ISerializable
                 return null;
             }
 
-            // fixedIndex is used for perf. optimization in case of adding big ranges of nodes
             int currentInd = _index;
-            int fixedInd = _parent.Nodes.FixedIndex;
-
-            if (fixedInd > 0)
-            {
-                currentInd = fixedInd;
-            }
 
             if (currentInd > 0 && currentInd <= _parent.Nodes.Count)
             {
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNodeCollection.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNodeCollection.cs
@@ -22,11 +22,6 @@ public class TreeNodeCollection : IList
         _owner = owner;
     }
 
-    // This index is used to optimize performance of AddRange
-    // items are added from last to first after this index
-    // (to work around TV_INSertItem comctl32 perf issue with consecutive adds in the end of the list)
-    internal int FixedIndex { get; set; } = -1;
-
     public virtual TreeNode this[int index]
     {
         get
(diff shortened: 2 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNode.cs
```diff
@@ -810,14 +810,7 @@ public TreeNodeCollection Nodes
                 return null;
             }
 
-            // fixedIndex is used for perf. optimization in case of adding big ranges of nodes
             int currentInd = _index;
-            int fixedInd = _parent.Nodes.FixedIndex;
```
Tanya-Solyanik, line 815: What this code added in NET8?

Epica3055: Actually no.

According to the comments I saw in the code, I think it was introduced to optimize `AddRange` performance.

But somehow after [PR 9078](https://github.com/dotnet/winforms/pull/9078), the logic about `FixedIndex` is not stable any more, which behaves like #11243.

So I tried to debug, and I found that the logic about `FixedIndex` didn't make scene to me about how to optimize performance and the logic is a little hard for me to understand. Then I tried to remove `FixedIndex` and I found that the issue is fixed. 

I also made some tests like, using `AddRange` to add an array whose length is 1000. And I think it is not slow.

Tanya-Solyanik: >found that the logic about FixedIndex didn't make scene to me

Interesting, maybe the original logic was broken by the refactoring changes? Could you please review this logic in the NET Framework sources or on the referencesource website?

I'm concerned about removing performance optimization, some application might be relying on it.

I remember that PR that introduced the standard comparer had regressed something else in the TreeView, could you please review the file history and bugs related to the treeview? I wonder if it makes sence to revert the refactoring change?

Epica3055: Yes, I checked NET Framework code, they are basically identical regarding `AddRange` logic and the code before [PR 9078](https://github.com/dotnet/winforms/pull/9078)

The difference between before and after [PR 9078](https://github.com/dotnet/winforms/pull/9078) is here:

this is main branch 
https://github.com/dotnet/winforms/blob/b75e07416960a2b2820689ee56af1058de7c876e/src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNodeCollection.cs#L207-L234

See line 224-227

https://github.com/dotnet/winforms/blob/9ec7d07e9c989b6dccd0cade990cb805e0fff619/src/System.Windows.Forms/src/System/Windows/Forms/TreeNodeCollection.cs#L263-L290

this is 7.0 branch 
See line 280-284

<hr/>

So now we use `List<TreeNode>` to store subTreeNodes instead of `TreeNode[]`, so it's impossible to do reverse iteration.

But if we don't do reverse iteration the logic here will break down, which I think is the cause of this issue.
https://github.com/dotnet/winforms/blob/b75e07416960a2b2820689ee56af1058de7c876e/src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNode.cs#L804-L831

> I wonder if it makes sence to revert the refactoring change?

Yes, that's the solution that I came up with in the first place see [PR 11253](https://github.com/dotnet/winforms/pull/11253)

Tanya-Solyanik: I see, we can't [insert ](https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/List.cs,770) an item into a generic list at an index past the actual list size, even when within the capacity, like we were inserting into an empty portion of the array. 
We can't revert the "FixedIndex" optimization because we don't know the specific scenario it solves. It might require a ListView with some specific properties and your test might not be sufficient.
I prefer reverting the change that replaced array of child nodes with the generic list. But please revert of that part, keep variable names cleanup.

PR #11272, opened 2024-04-24
Add a general deserializer for BinaryFormattedObject

JeremyKuhne:
This adds a constrained, simplified deserializer based on BinaryFormattedObject.

This deserializer does not support:

- IObjectReference
- Offset arrays

The deserialzer guarantees that value type members are initialized in SerializationInfo callbacks and deserialization events. It will throw if value types can not be applied in their final state.

Reference type members (including transitive ones) are only guaranteed to be in their final state if the object graph has no cycles (as is also the case for BinaryFormatter).

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11272)


JeremyKuhne:
Tests include most of the existing BinaryFormatter tests run against both deserializers. Only excluded types that would have required addtional package dependencies.

There are also additional tests and the expectation is to be adding more over the next several weeks in additional PRs.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/BinaryFormattedObject.cs
```diff
@@ -67,6 +71,29 @@ public BinaryFormattedObject(Stream stream, Options? options = null)
...
+    ///  Deserializes the <see cref="BinaryFormattedObject"/> back to an object.
+    /// </summary>
+    [RequiresUnreferencedCode("Ultimately calls Assembly.GetType for type names in the data.")]
+    public object Deserialize()
+    {
+        try
+        {
+            _typeResolver ??= new(_options, _recordMap);
+            Id rootId = ((SerializationHeader)_records[0]).RootId;
+            return Deserializer.Deserializer.Deserialize(rootId, _recordMap, _typeResolver, _options);
```
jacdavis, line 84: >Deserializer.Deserializer.Deserialize [](http://example.com/codeflow?start=19&length=37)

LOL, needs more deserializer identifiers :)

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/BinaryFormattedObject.cs
```diff
@@ -67,6 +71,29 @@ public BinaryFormattedObject(Stream stream, Options? options = null)
...
+    /// <summary>
+    ///  Deserializes the <see cref="BinaryFormattedObject"/> back to an object.
+    /// </summary>
+    [RequiresUnreferencedCode("Ultimately calls Assembly.GetType for type names in the data.")]
+    public object Deserialize()
+    {
+        try
+        {
+            _typeResolver ??= new(_options, _recordMap);
+            Id rootId = ((SerializationHeader)_records[0]).RootId;
```
jacdavis, line 83: >SerializationHeader [](http://example.com/codeflow?start=26&length=19)

Are you 100% confident this will be of this type here? Any way to confirm before the cast? Any magic number or watermark to confirm against?

JeremyKuhne: It should be, and I deliberately want it to blow up when it isn't (it will surface as a SerializationException).

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/BinaryFormattedObject.cs
JeremyKuhne: I figured adding a using for one line was a bit much. :)

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/BinaryFormattedObject.cs
```diff
@@ -67,6 +71,29 @@ public BinaryFormattedObject(Stream stream, Options? options = null)
...
+        }
+        catch (Exception ex) when (ex is ArgumentException or InvalidCastException or ArithmeticException or IOException)
+        {
+            // Make the exception easier to catch, but retain the original stack trace.
+            throw ex.ConvertToSerializationException();
+        }
+        catch (TargetInvocationException ex)
+        {
+            throw ExceptionDispatchInfo.Capture(ex.InnerException!).SourceException.ConvertToSerializationException();
+        }
```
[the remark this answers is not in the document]
JeremyKuhne: I believe it is the right choice. User callbacks can throw anything, of course. I'll discuss with @GrabYourPitchforks and others about what sort of configurability/changes we might want here.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/BinaryFormattedObject.cs
JeremyKuhne: Note that it is a deliberate design decision to hard cast and fail quickly throughout the deserializer.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Deserializer/ClassRecordDeserializer.cs
```diff
@@ -0,0 +1,91 @@
...
+        recordDeserializer.Continue();
+        return recordDeserializer;
+
+        static ObjectRecordDeserializer HandleObjectReference(Type type, ClassRecord classRecord, IDeserializer deserializer)
+        {
+            // Arbitrary IObjectReference handling is complicated and at risk of denial of service attacks.
+            // System.UnitySerializationHolder is the only .NET system type that implements this and is
+            // used to encapsulate DBNull, so we'll special case it.
+            if (type != deserializer.TypeResolver.GetType("System.UnitySerializationHolder", Id.Null)
+                || !0x0002.Equals(classRecord["UnityType"]))
```
jacdavis, line 70: >!0x0002.Equals(classRecord["UnityType"])) [](http://example.com/codeflow?start=19&length=41)

This seems like an odd pattern. Why object equals here?

JeremyKuhne: Because I wanted it to fall into the `SerializationException` with the message below if the data was something unexpected. It probably doesn't really matter and I can just hard cast here.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Deserializer/Deserializer.cs
```diff
@@ -0,0 +1,375 @@
...
+                }
+            }
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        [RequiresUnreferencedCode("Calls System.Windows.Forms.BinaryFormat.Deserializer.ObjectRecordParser.Create(Id, IRecord, IDeserializer)")]
+        object DeserializeNew(Id id)
+        {
+            // Strings, string arrays, and primitive arrays can be completed without creating a
+            // parser object. Single rimitives don't normally show up as records unless they are top
```
jacdavis, line 218: >rimitives [](http://example.com/codeflow?start=37&length=9)

Spelling

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Support/SerializationEvents.cs
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        // We should invoke the methods starting from base
+        attributedMethods?.Reverse();
+
+        return attributedMethods;
+    }
+
+    internal void InvokeOnDeserializing(object obj, StreamingContext context) =>
+        InvokeOnDelegate(obj, context, _onDeserializingMethods);
```
jacdavis, line 78: >InvokeOnDelegate [](http://example.com/codeflow?start=8&length=16)

Were these pre-existing code from the original implementation? Just curious if we care about exceptions getting translated into methodinvocationexception across the delegate invocation.

JeremyKuhne: SerializationEvents is an existing class that I simplified a bit (collapsed it with a separate cache class). In the outer entry point for creating the BinaryFormattedObject and deserializing I do unwrap the TargetInvocationException. Throws through delegates get hit in some of the tests.

jacdavis (approved):
:shipit:

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Deserializer/ClassRecordDeserializer.cs
```diff
@@ -0,0 +1,91 @@
...
+                throw new SerializationException($"Type '{type}' is not allowed to implement IObjectReference.");
+            }
+
+            deserializer.DeserializedObjects.Add(classRecord.ObjectId, DBNull.Value);
+            deserializer.CompleteObject(classRecord.ObjectId);
+            return new NoOpRecordDeserializer(classRecord, deserializer);
+        }
+    }
+
+    private class NoOpRecordDeserializer : ObjectRecordDeserializer
```
lonitra, line 81: Do we want this in its own file?

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Deserializer/ClassRecordSerializationInfoDeserializer.cs
```diff
@@ -0,0 +1,98 @@
...
+///  <para>
+///   This is used either because the class implements <see cref="ISerializable"/> or because a surrogate was used.
+///  </para>
+/// </remarks>
+internal sealed class ClassRecordSerializationInfoDeserializer : ClassRecordDeserializer
+{
+    private readonly ClassRecord _classRecord;
+    private readonly SerializationInfo _serializationInfo;
+    private readonly ISerializationSurrogate? _surrogate;
+    private int _currentField;
```
lonitra, line 23: Maybe we can rename this so it is more clear when reading the code that this is meant to be an `int`. Something like `_currentFieldPosition`?

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Support/TypeInfo.cs
```diff
@@ -44,6 +45,17 @@ internal static class TypeInfo
     internal static Assembly CorelibAssembly { get; } = typeof(string).Assembly;
     internal static string CorelibAssemblyString { get; } = CorelibAssembly.FullName!;
 
+    private static Type? s_surrogateForCyclicalReferenceType;
+
+    internal static bool IsSurrogateForCyclicalReference(Type type)
```
lonitra, line 50: I might've missed it, but I didn't catch any usages of this, is this needed?

lonitra (approved):
Overall 👍

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Support/TypeInfo.cs
JeremyKuhne: We might not. I currently always block surrogates that change the instance- BinaryFormatter only blocks if this is true.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Deserializer/ClassRecordDeserializer.cs
JeremyKuhne: For a single use only I felt it was better to have it here.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Deserializer/ClassRecordSerializationInfoDeserializer.cs
JeremyKuhne: I've renamed it to _currentFieldIndex.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Support/TypeInfo.cs
JeremyKuhne: Moved it to PendingSerializationInfo and #if defed it out.

lonitra (dismissed):
:shipit:

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/ClassRecord.cs
```diff
@@ -49,4 +51,77 @@ private protected ClassRecord(ClassInfo classInfo, MemberTypeInfo memberTypeInfo
...
+        return memberValues;
+    }
+
+    /// <summary>
+    ///  Writes <paramref name="memberValues"/> as specified by the <paramref name="memberTypeInfo"/>
+    /// </summary>
+    private protected static void WriteValuesFromMemberTypeInfo(
+        BinaryWriter writer,
+        MemberTypeInfo memberTypeInfo,
+        IReadOnlyList<object?> memberValues)
```
bartonjs, line 89: I don't know what all types end up in here, and I don't think I've brought it up already, but.. "is IReadOnlyList a good type to use here?"  If it's expecting user defined types, there are a lot of collections that don't implement the IReadOnlyList interface.

While the implementation currently uses `.Count` it's just in a loop, so it looks like IEnumerable would be workable...

bartonjs, line 89: The implementation expects only IRecord values (including `null`)... should this be `IReadOnlyList<IRecord?>`?

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Deserializer/ArrayRecordDeserialzer.cs
```diff
@@ -0,0 +1,117 @@
...
+    [RequiresUnreferencedCode("Calls System.Windows.Forms.BinaryFormat.BinaryFormattedObject.TypeResolver.GetType(String, Id)")]
+    internal ArrayRecordDeserialzer(ArrayRecord<object?> arrayRecord, IDeserializer deserializer)
+        : base(arrayRecord, deserializer)
+    {
+        _arrayRecord = arrayRecord;
+
+        if (arrayRecord is not IBinaryArray binaryArray)
+        {
+            // Other array types are handled directly (ArraySinglePrimitive and ArraySingleString).
+            Debug.Assert(arrayRecord is ArraySingleObject);
```
bartonjs, line 26: I'm assuming from context that "ArraySingleString" means "string[]", and not `RuntimeAssert(((string[])obj).Length == 1)`.  But the name keeps making me think "an array of a single string", versus "single-dimensional array of strings".

Don't know if that confusion comes from this library or something else (Adam's record reader thingy?)... but it's bothered me enough that I'm finally saying something.

bartonjs (commented):
20th comment and 5pm, posting what I have so far (then continuing)

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Deserializer/ArrayRecordDeserialzer.cs
```diff
@@ -0,0 +1,117 @@
...
+            {
+                throw new NotSupportedException("Only arrays with zero offsets are supported.");
+            }
+
+            _array = _arrayType is BinaryArrayType.Rectangular
+                ? Array.CreateInstance(_elementType, binaryArray.Lengths.ToArray())
+                : Array.CreateInstance(_elementType, arrayRecord.Length);
+        }
+
+        deserializer.DeserializedObjects.Add(arrayRecord.ObjectId, _array);
```
bartonjs, line 62: The layering here feels weird to me.  The consttructor for this object pushes tangentially-related state into one of the ctor parameters?

```C#
class ListItem
{
    public ListItem(object value, SomeList addMeToThis)
    {
        Value = value;
        addMeToThis.Add(this);
    }
}
```

seems an odd paradigm.  But, if that's the way it works...

src/System.Private.Windows.Core/src/System/IO/BinaryReaderExtensions.cs
```diff
@@ -56,10 +57,25 @@ public static long Remaining(this BinaryReader reader)
...
         where T : unmanaged
     {
         ArgumentOutOfRangeException.ThrowIfNegative(count);
+
+        if (typeof(T) == typeof(decimal) || typeof(T) == typeof(DateTime) || typeof(T) == typeof(TimeSpan))
+        {
+            // Decimal is persisted as a UTF-8 string. It has a 7-bit encoded length so it could be, in theory just
+            // a few bytes. Picking 2 bytes- once the minimum string length (and termination if applicable) are
+            // evaluated, this could be made more aggressive.
+            if (count > 0 && reader.Remaining() < count * (typeof(T) == typeof(decimal) ? 2 : sizeof(T)))
```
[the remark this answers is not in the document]
JeremyKuhne: I'll add the extra comment line. I'll also point out that this is one of two guards. The other is that we never reserve more than 10K of these in the List they go into.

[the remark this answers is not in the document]
JeremyKuhne: I'll double check other sizeof usages.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Deserializer/ArrayRecordDeserialzer.cs
```diff
@@ -0,0 +1,117 @@
...
+        IsParsingComplete = Continue().IsNull;
```
bartonjs, line 63: Why is the ctor calling Continue()?  Doesn't this destructively lose whatever data is in index 0?

It looks like Continue() already sets IsParsingComplete to true when it finishes, so why set it here?

src/System.Private.Windows.Core/src/System/IO/BinaryReaderExtensions.cs
```diff
@@ -128,5 +144,186 @@ public static unsafe T[] ReadPrimitiveArray<T>(this BinaryReader reader, int cou
...
+                {
+                    values.Add((T)(object)new TimeSpan(reader.ReadInt64()));
+                }
+                else
+                {
+                    throw new SerializationException($"Invalid primitive type '{typeof(T)}'");
+                }
+            }
+
+            return [.. values];
```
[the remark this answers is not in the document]
JeremyKuhne: Yeah, the analyzer we have in play forces it this way.

src/System.Private.Windows.Core/src/System/IO/BinaryReaderExtensions.cs
```diff
@@ -128,5 +144,186 @@ public static unsafe T[] ReadPrimitiveArray<T>(this BinaryReader reader, int cou
...
+    /// </summary>
+    /// <remarks>
+    ///  <para>
+    ///   Only supports <see langword="bool"/>, <see langword="byte"/>, <see langword="sbyte"/>, <see langword="char"/>,
+    ///   <see langword="short"/>, <see langword="ushort"/>, <see langword="int"/>, <see langword="uint"/>,
+    ///   <see langword="long"/>, <see langword="ulong"/>, <see langword="float"/>, <see langword="double"/>,
+    ///   <see langword="decimal"/>, <see cref="DateTime"/>, and <see cref="TimeSpan"/>.
+    ///  </para>
+    /// </remarks>
+    public static unsafe void WritePrimitives<T>(this BinaryWriter writer, IReadOnlyList<T> values)
```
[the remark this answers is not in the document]
JeremyKuhne: I'll be more consistent.

src/System.Private.Windows.Core/src/System/IO/BinaryReaderExtensions.cs
```diff
@@ -128,5 +144,186 @@ public static unsafe T[] ReadPrimitiveArray<T>(this BinaryReader reader, int cou
...
         return array;
+
+        static T[] ReadNonBlittableTypes(BinaryReader reader, int count)
+        {
+            // We've already made one check for remaining data. Decimal is a weird case as it is 16 bytes and is
+            // persisted as a UTF-8 string. Out of an abundance of caution, we'll max out what we pre-allocate to avoid
+            // untrusted data claiming a huge number of decimal strings. Worst case is that roughly 4x what the remaining
+            // data could contain at the smallest string size, but we'll still guard.
+
+            List<T> values = new(Math.Min(count, 1024 * 10));
```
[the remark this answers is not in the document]
JeremyKuhne: Good point. I'll look at creating an array abstraction that can prevent the copy when there isn't bad data and it is under the threshhold.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/ArrayRecord.cs
```diff
@@ -33,6 +34,47 @@ internal abstract class ArrayRecord : ObjectRecord, IEnumerable
...
+    ///  Reads a count of object member values of <paramref name="type"/> with optional clarifying <paramref name="typeInfo"/>.
+    /// </summary>
+    /// <exception cref="SerializationException"><paramref name="type"/> was unexpected.</exception>
+    private protected static IReadOnlyList<object?> ReadObjectArrayValues(
+        BinaryFormattedObject.IParseState state,
+        BinaryType type,
+        object? typeInfo,
+        int count)
+    {
+        List<object?> memberValues = new(Math.Min(count, BinaryFormattedObject.MaxNewCollectionSize));
```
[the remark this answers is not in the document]
JeremyKuhne: When we can validate how much data is needed to satisfy the count specification, we just allocate the space. When we can't, this is an attempt to mitigate OOM from corrupted data.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/BinaryFormattedObject.IParseState.cs
```diff
@@ -0,0 +1,18 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace System.Windows.Forms.BinaryFormat;
+
+internal sealed partial class BinaryFormattedObject
+{
+    /// <summary>
+    ///  Parsing state.
+    /// </summary>
+    internal interface IParseState
```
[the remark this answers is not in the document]
JeremyKuhne: I don't do interfaces for mocking generally, this is for composability. I'm using parts of the functionality to generate some of the test data.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/BinaryFormattedObject.ITypeResolver.cs
```diff
@@ -0,0 +1,21 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace System.Windows.Forms.BinaryFormat;
+
+internal sealed partial class BinaryFormattedObject
+{
+    /// <summary>
+    ///  Resolver for types.
+    /// </summary>
+    internal interface ITypeResolver
```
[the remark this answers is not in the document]
JeremyKuhne: I intend to be plugging other variants in here.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/ClassRecord.cs
```diff
@@ -49,4 +51,77 @@ private protected ClassRecord(ClassInfo classInfo, MemberTypeInfo memberTypeInfo
...
+        foreach ((BinaryType type, object? info) in memberTypeInfo)
+        {
+            object? value = ReadValue(state, type, info);
+            if (value is not ObjectNull nullValue)
+            {
+                memberValues.Add(value);
+                continue;
+            }
+
+            if (nullValue.NullCount != 1)
```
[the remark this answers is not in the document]
JeremyKuhne: The null annotations here weren't aggressive enough. It can never actually be `null` (either a record or a boxed primitive). Fixed.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/ClassRecord.cs
```diff
@@ -49,4 +51,77 @@ private protected ClassRecord(ClassInfo classInfo, MemberTypeInfo memberTypeInfo
...
+            {
+                throw new SerializationException("Member values can only have one null assigned.");
+            }
+
+            memberValues.Add(null);
+        }
```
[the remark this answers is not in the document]
JeremyKuhne: `null` isn't actually possible. Fixed the annotation.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/ClassRecord.cs
JeremyKuhne: This is just Records, nulls, and "primitives". I could loosen up to `IEnumerable` at some point if it is useful. I suspect that if a public API was ever made for an OM like this it would be a general `Serialize(object)` that kept all of this internal to avoid littering the ecosystem with wildly variant NRBF streams. That said, I haven't really dug deep on the idea.

JeremyKuhne: Boxed primitives can be in there too.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Deserializer/ArrayRecordDeserialzer.cs
JeremyKuhne: Naming here comes from the NRBF spec. :)

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Deserializer/ArrayRecordDeserialzer.cs
JeremyKuhne: There is a specific ordering that has to happen. I might be further abstracting this to `deserializer.AddObject()` for cycle checking reasons.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Deserializer/ArrayRecordDeserialzer.cs
JeremyKuhne: Because it might not need to iterate in the outer deserialize loop, all items might already be available in the deserialized objects. Just a minor optimization to avoid coming all the way back down the stack.
 
The set is vestigial. I'll remove it.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Deserializer/ArrayRecordDeserialzer.cs
JeremyKuhne: The specific ordering is that it has to come before members are evaluated as there can be cycles back to this object. I'll try to move this up.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/BinaryFormattedObject.ParseState.cs
```diff
@@ -20,17 +20,7 @@ public ParseState(BinaryReader reader, BinaryFormattedObject format)
...
-            Justification = """
-                Incoming type names are coming off of the formatted stream. There is no way for user code to pass compile
-                time context for preserialized data. If a type can't be found on deserialization it won't matter any more
-                than any other case where the type can't be found (e.g. a missing assembly). The deserializer will fail
-                with information on the missing type that can be used to attribute to keep said type.
-                """)]
-        [return: DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.All)]
-        public Type GetType(string typeName, Id libraryId) => _format.GetType(typeName, libraryId);
+        public Options Options => _format._options;
+        public ITypeResolver TypeResolver => _format.TypeResolver;
```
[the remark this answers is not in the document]
JeremyKuhne: It is a bit tricky as I want the `TypeResolver` to be lazy. Normally you shouldn't need it to parse, and in fact, I want to block it by default (as figuring out the stream contents by reflecting on the current state of types is not the safest thing in the world to do).

src/System.Private.Windows.Core/src/System/IO/BinaryReaderExtensions.cs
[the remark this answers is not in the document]
JeremyKuhne: @bartonjs I put up a collection that encapsulates what is going on here. `CappedArray`

src/System.Private.Windows.Core/src/System/Collections/Generic/CappedArray.cs
```diff
@@ -0,0 +1,124 @@
...
+    {
+        if (_items.Length == _maxCount)
+        {
+            throw new ArgumentOutOfRangeException(nameof(item), "Collection is at max capacity.");
+        }
+
+        Debug.Assert(Count == _items.Length);
+        int size = Count;
+
+        int newCapacity = _items.Length == 0 ? DefaultCapacity : 2 * _items.Length;
```
bartonjs, line 80:
```suggestion
        int newCapacity = _items.Length == 0 ? DefaultCapacity : Math.Min(_maxCount, 2 * _items.Length);
```

Then ToArray() can just throw if Count != _maxCount.

src/System.Private.Windows.Core/src/System/Collections/Generic/CappedArray.cs
```diff
@@ -0,0 +1,124 @@
...
+    public bool Contains(T item) => Array.IndexOf(_items, item, 0, Count) != -1;
+    public void CopyTo(T[] array, int arrayIndex) => Array.Copy(_items, 0, array, arrayIndex, Count);
+    bool ICollection<T>.Remove(T item) => throw new NotSupportedException();
+
+#pragma warning disable IDE0305 // Simplify collection initialization
+    // We deliberately don't want to use a collection expression here because it would allocate a new array.
+    public static implicit operator T[](CappedArray<T> list) => list.ToArray();
+#pragma warning restore IDE0305
+
+    public static explicit operator ArraySegment<T>(CappedArray<T> list) => new(list._items, 0, list.Count);
```
bartonjs, line 123: I'd remove all of these, and the interface declarations, and rename the class to `ArrayBuilder<T>`; as it looks like it's only ever used to add things into and it's never itself passed around.

JeremyKuhne: It is passed around as IReadOnlyList<T>

bartonjs (commented):
Today's progress

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Deserializer/ClassRecordFieldInfoDeserializer.cs
```diff
@@ -0,0 +1,99 @@
...
+            }
+
+            _currentFieldIndex++;
+        }
+
+        if (!_hasFixups || !_isValueType)
+        {
+            // We can be used for completion even with fixups if we're not a value type as our fixups won't need to be
+            // copied to propogate them. Note that surrogates cannot replace our created instance for reference types.
+            Deserializer.CompleteObject(_classRecord.ObjectId);
```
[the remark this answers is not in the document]
JeremyKuhne: When it's `_incompleteDependencies` are cleared in the `Deserializer` via the dependencies being `CompleteObject`ed.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Deserializer/ClassRecordFieldInfoDeserializer.cs
```diff
@@ -0,0 +1,99 @@
...
+
+            if (!reference.IsNull && field.FieldType.IsValueType && Deserializer.IncompleteObjects.Contains(reference))
+            {
+                // Need to track a fixup for this field.
+                _hasFixups = true;
+                Deserializer.PendValueUpdater(new FieldValueUpdater(_classRecord.ObjectId, reference, field));
+            }
+
+            _currentFieldIndex++;
+        }
```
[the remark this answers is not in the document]
JeremyKuhne: Yeah, it isn't a requirement. I'll comment.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Deserializer/ClassRecordSerializationInfoDeserializer.cs
```diff
@@ -0,0 +1,98 @@
...
+            _serializationInfo.AddValue(memberName, memberValue);
+            _currentMemberIndex++;
+        }
+
+        // If we were confident that there were no cycles in the graph to this point we could apply directly
+        // if there were no pending value types (which should also not happen if there are no cycles).
+
+        PendingSerializationInfo pending = new(_classRecord.ObjectId, _serializationInfo, _surrogate);
+
+        if (_hasAnyReferences && Deserializer.IncompleteObjects.Count > 1)
```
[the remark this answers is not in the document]
JeremyKuhne: The trick here is that with `SerializationInfo` we have no idea what the surrogate or serialization constructor is going to do with it. If references are dereferenced, what will the state be? Doesn't matter for direct field sets as no user code is involved.

When cycles are involved full dependency analysis gets costly. By pending the work to the end we avoid having to keep track of this.

I believe it is relatively easy to track when we encounter our first cycle (by checking the parse stack of `int[]` to see if member ids are on it). If we haven't hit that condition yet we can finish all of these as they occur. After the first hit I don't think it is easy to know whether the current object might be part of a cycle without doing a separate, full walk.

I've removed the extra condition because it doesn't really buy enough to warrant the extra cognitive overhead.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Deserializer/Deserializer.cs
```diff
@@ -0,0 +1,378 @@
...
+            if (_recordMap[completedId] is ClassRecord classRecord)
+            {
+                // Hook any finished events for this object. Doing at the end of deserialization for simplicity.
+                // (If we knew there were no cycles in the graph we could fire these as we go.)
+
+                Type type = _typeResolver.GetType(classRecord.Name, classRecord.LibraryId);
+                object @object = _deserializedObjects[completedId];
+
+                OnDeserialized += SerializationEvents.GetSerializationEventsForType(type)
+                    .AddOnDeserialized(@object, OnDeserialized);
```
bartonjs, line 323: This does +=, and the helper it's calling also does +=.  This feels like it's MASSIVELY over-enqueuing (N^2)

[sharplab.io sample](https://sharplab.io/#v2:EYLgtghglgdgNAFxAJwK4wD4AEBMAGAWAChcBGY42BAU2RggBsACXF0gdmIG9im+mADsigA3CDTYA2JlSYBnAPoIA9gkYBuXvyGjx1KU2ojqMBGxwAeZcABW1AMYIAfEwCix05qL8mWvjrEJLFJpLAAWJgBZaBgACgBKPyYebx9+ADNlZCZY2QgmAF4mPHUmfIsmHFKIAGoaxNS05KSmvncTMxqi6IBramCcOViIODcPBHivVqYAXwpGtPbTWJhUBgZJlv5ggE5YxRU1Dan+OaIk4NDSS2s7Rxde/uuhvNGBq1sHZ0MEewafFLTTLZXKmJjAQrFUoQioAZmhdX+rUB0341F+TC6UQgfQGw1GwFG6L+JyaZy2fCw7B+9lJs3mPku5g+d2+jzxrxkYMJzNuXxcxKRfBRTXefPuTBg1AAHmYispCi4uGw9gASABEXAgM2SwBm6sm8iUqkYdVKMzpPmJmKKUtllu21OJdLOMyAA=)

This sample simulates 2 classes each adding 3 callbacks (should be 6 total). Then it fires them.  70 invocations occur.

Unless I'm missing some nuance, this should not be passing the existing OnDeserialized in to AddOnDeserialized; and AddOnDeserialized should thus do something like

```C#
Action<object>? ret = null;

foreach (MemberInfo memberInfo in relevantMemberInfos)
{
    ret += next;
}

return ret;
```

(Doesn't need to special-case ret being null, because apparently delegate += handles that... otherwise this code would already fail :smile:)

And once Add isn't doing += on an input, it should probably be renamed to GetOnDeserialized.

If the adds are chaining somewhere else, then maybe this line just needs to pass in null as the existing basis.

bartonjs: Actually, N^2 is too low.  This is 2^classes.

Initial: []

As long as nothing gets added, it the delegate stays null.

Now have a class with two relevant members:

State: [0, 1].

Call it again on a class that doesn't care.  [0, 1] += [0, 1] => [0, 1, 0, 1]  (2 became 4, while adding nothing).

So, if the types walk in an order 0, 0, 1, 0, 2, 0, 1, 0:

* 0
* 0
* 1
* 2
* 6
* 12
* 25
* 50

instead of 4.

After fixing my test to remove a layer of += that I don't think was correctly reflecting the code here, I observe for 3, 4, 5:

* 3
* 10
* 25

when the correct answer is 12.

bartonjs: [improved sample](https://sharplab.io/#v2:EYLgtghglgdgNAFxAJwK4wD4AEBMBGAWAChjYEBTZGCAGwAJcG8B2Ygb2Lq7oAdkoAbhApMAbHTJ0AzgH0EAewS0A3J259Bw8mLrkB5GAiY4APPOAArcgGMEAPjoBRfYdVFudNVw1CRWPOJYACx0ALLQMAAUAJRedBzuHtxkANoAunTW8ugIUnQAvHQpAAxwdKV0eGUVFVXlaW5JyYYSBeWNTZ6JTQBm8sjkENYAFnSRklk5EjCZ2YZSsd1JCZ1JzgZGANSF4QDW5P44UuNlk4Zl64bRHatcUJubN0kAvsRxHpcIkTCoNDTX724/gAnDEnh4QZFZAolP8nq8SEt/IE8KZzFZbA49gdUcdJBAypIaFAwFAEGVDmZLDZ7LoENZFh4Vp0+sgxpJgG1iso6JyTHRiaSEDzgA9GZ1mbc6dY6NswhB9odIgTeQClh4EYCuFhmNL4W8kQFjFSMbTsUr8YSWsBxVxJUlKeiaQ4YOQAB5GQryAoONh0ADC8hgUnkNHIADoAOr8CiRAAkACI2BBngBaNjAZ5wBPXaRyRS0B4857goG610e/VEZ5AA=)

Fun fact: if `counts` is only composed of zeros and ones, it's a binary decoder.

JeremyKuhne: Yeah, Loni just caught this as well and we're adding the relevant test to catch it. :)

JeremyKuhne: I think the idea of the original code here was to avoid creating multiple intermediate delegates.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Deserializer/Deserializer.cs
```diff
@@ -0,0 +1,378 @@
...
+                    // Push our current parser.
+                    _parseStack.Push(currentId);
+                    _parserStack.Push(currentParser);
+
+                    // Push the required parser so we can complete it.
+                    _parseStack.Push(requiredId);
+                    _parserStack.Push(requiredParser);
+
+                    break;
+                }
```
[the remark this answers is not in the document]
JeremyKuhne: I think we're covered here. There isn't a lot we can track about Continue() getting nowhere as objects can have the same required object multiple members in a row. The _parseStack check will catch any infinite loops.

I did add a debug assert to make sure that DeserializeNew always returns something that is not IRecord.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Deserializer/Deserializer.cs
JeremyKuhne: It is too messy. Simplifying to avoid making this mistake again.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Support/SerializationEvents.cs
```diff
@@ -0,0 +1,109 @@
...
+        // Traverse the hierarchy to find all methods with the particular attribute
+        Type? baseType = type;
+        while (baseType is not null && baseType != typeof(object))
+        {
+            // Get all methods which are declared on this type, instance and public or nonpublic
+            MethodInfo[] methods = baseType.GetMethods(BindingFlags.DeclaredOnly | BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Public);
+            foreach (MethodInfo method in methods)
+            {
+                // For each method find if attribute is present, the return type is void and the method is not virtual
+                if (method.IsDefined(attribute, inherit: false))
```
bartonjs, line 61: The comment says it checks 3 things... the code only checks the first one.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Support/SerializationExtensions.cs
```diff
@@ -42,4 +42,51 @@ public static Span<T> GetArrayData<T>(this Array array)
...
+
+        static void SetArrayValueByFlattenedIndex(Array array, int flattenedIndex, object? value)
+        {
+            int ranks = array.Rank;
+
+            if (ranks == 2)
+            {
+                int length = array.GetLength(1);
+                int index1 = flattenedIndex / length;
+                int index2 = flattenedIndex - (index1 * length);
```
bartonjs, line 70: Isn't this just

```suggestion
                (int index1, int index2) = int.DivRem(flattenedIndex, length);
```

I mean, this is fine... but (IIUC) Div and DivRem are the same instruction, so might as well use the one instruction to get both parts and save on an extra multiplication and subtraction.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Support/SerializationExtensions.cs
```diff
@@ -42,4 +42,51 @@ public static Span<T> GetArrayData<T>(this Array array)
...
+            }
+
+            Span<int> indices = stackalloc int[ranks];
+            for (int i = 0; i < ranks; i++)
+            {
+                indices[i] = flattenedIndex / rankProducts[i];
+                flattenedIndex -= indices[i] * rankProducts[i];
+            }
+
+            array.SetValue(value, indices.ToArray());
```
bartonjs, line 89: @tannergooding Can you double-check this, since I know you're currently mired in code doing the same thing?

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Support/SerializationInfoExtensions.cs
```diff
@@ -0,0 +1,18 @@
...
+
+internal static class SerializationInfoExtensions
+{
+    private static readonly Action<SerializationInfo, string, object, Type> s_updateValue =
+        typeof(SerializationInfo)
+        .GetMethod("UpdateValue", BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance)!
+        .CreateDelegate<Action<SerializationInfo, string, object, Type>>();
+
+    internal static void UpdateValue(this SerializationInfo si, string name, object value, Type type) =>
+        s_updateValue(si, name, value, type);
```
bartonjs, line 17: On the one hand, the whole thing depends on private reflection.  On the other... "private reflection, really?"

I don't see any calls to UpdateValue in current corelib, so hopefully this type somewhere is telling the trimmer to not delete it.

bartonjs (approved):
Assuming that the code vs comment behavior in finding serialization callback methods is reconciled, I have no further feedback.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Support/SerializationEvents.cs
JeremyKuhne: Copy paste, same is true for the source. :) Rechecking the comments.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Support/SerializationInfoExtensions.cs
JeremyKuhne: [UpdateValue is used here in the runtime.](https://github.com/dotnet/runtime/blob/71f8fb65a5a28018901823c19de57fe9451ab3b1/src/libraries/System.Runtime.Serialization.Formatters/src/System/Runtime/Serialization/ObjectManager.cs#L1628)

JeremyKuhne: Added the `[DynamicDependency]` to this.

lonitra (approved):

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Support/SerializationExtensions.cs
tannergooding: Looks correct to me.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Support/SerializationExtensions.cs
```diff
@@ -42,4 +42,50 @@ public static Span<T> GetArrayData<T>(this Array array)
 
         return data;
     }
+
+    /// <summary>
+    ///  Sets a value in an array by its flattened index.
+    /// </summary>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static void SetArrayValueByFlattenedIndex(this Array array, object? value, int flattenedIndex)
```
adamsitnik, line 50: Calling this method in a loop can be very expensive, as it always allocates an array of indices for arrays with ranks > 1.

This could be avoided by hoisting the array as I did here: https://github.com/adamsitnik/SafePayloadReader/blob/199c0a5b35456472193b7e597ffc64efc7465f18/System.Runtime.Serialization.BinaryFormat/Records/RectangularOrCustomOffsetArrayRecord.cs#L50-L75

Since you don't need to target .NET Standard, you can use `MemoryMarshal.GetArrayDataReference(Array)` if WinForms accepts the usage of `Unsafe`: https://github.com/adamsitnik/SafePayloadReader/blob/199c0a5b35456472193b7e597ffc64efc7465f18/System.Runtime.Serialization.BinaryFormat/Records/RectangularOrCustomOffsetArrayRecord.cs#L77-L123

adamsitnik (commented):
I've reviewed the changes related to reading the payload, without the later stages of deserialization. Overall it LGTM, but PTAL at my comments. Thanks!

PR #11283, opened 2024-04-28
Enable nullability in TabPageCollectionEditor

gpetrou:
## Proposed changes

- Enable nullability in TabPageCollectionEditor.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11283)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/CollectionEditor.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/CollectionEditor.cs
@@ -108,9 +108,16 @@ public partial class CollectionEditor : UITypeEditor
             }
         }
 
-        return itemType.UnderlyingSystemType == typeof(string)
-            ? string.Empty
-            : TypeDescriptor.CreateInstance(host, itemType, argTypes: null, args: null)!;
+        if (itemType.UnderlyingSystemType == typeof(string))
+        {
+            return string.Empty;
+        }
+
+        object? o = TypeDescriptor.CreateInstance(host, itemType, argTypes: null, args: null);
+
+        return o is null
+            ? throw new InvalidOperationException($"Neither {nameof(IDesignerHost)} nor {nameof(TypeDescriptor)} can create an instance of {itemType.FullName}.")
+            : o;
     }
 
     /// <summary>
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/CollectionEditor.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/CollectionEditor.cs
@@ -116,7 +116,12 @@ public partial class CollectionEditor : UITypeEditor
         object? o = TypeDescriptor.CreateInstance(host, itemType, argTypes: null, args: null);
 
         return o is null
-            ? throw new InvalidOperationException($"Neither {nameof(IDesignerHost)} nor {nameof(TypeDescriptor)} can create an instance of {itemType.FullName}.")
+            ? throw new InvalidOperationException(
+                string.Format(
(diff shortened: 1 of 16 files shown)


Epica3055 (commented):
looks good 😁

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/TabPageCollectionEditor.cs
```diff
@@ -13,29 +11,31 @@ namespace System.Windows.Forms.Design;
...
+        object? retValue = base.SetItems(editValue, value);
 
         tabControl?.ResumeLayout();
+
         return retValue;
     }
 
     protected override object CreateInstance(Type itemType)
     {
         object instance = base.CreateInstance(itemType);
```
lonitra, line 36: Changes look good! But while taking a look at the base method it looks like in rare cases we may be allowing null to be returned with `TypeDescriptor.CreateInstance`. Could you update the base method to check for if `TypDescriptor.CreateInstance` returns null and if so, throw an `InvalidOperationException` with description "Neither {nameof(IDesignHost)} nor {nameof(TypeDescriptor)} can create an instance of {itemType.FullName}." in either this PR or a separate PR?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/CollectionEditor.cs
```diff
@@ -108,9 +108,16 @@ protected virtual object CreateInstance(Type itemType)
...
-            : TypeDescriptor.CreateInstance(host, itemType, argTypes: null, args: null)!;
+        if (itemType.UnderlyingSystemType == typeof(string))
+        {
+            return string.Empty;
+        }
+
+        object? o = TypeDescriptor.CreateInstance(host, itemType, argTypes: null, args: null);
+
+        return o is null
+            ? throw new InvalidOperationException($"Neither {nameof(IDesignerHost)} nor {nameof(TypeDescriptor)} can create an instance of {itemType.FullName}.")
```
lonitra, line 119: We'd want to localize this since this would surface to users, could you place this string inside of [SR.resx](https://github.com/dotnet/winforms/blob/main/src/System.Windows.Forms.Design/src/Resources/SR.resx)? Afterwards this can be changed to the following
```suggestion
            ? throw new InvalidOperationException(string.Format(SR.CollectionEditorCreateInstanceError, nameof(IDesignerHost), nameof(TypeDescriptor), itemType.FullName))
```

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/CollectionEditor.cs
```diff
@@ -108,9 +108,21 @@ protected virtual object CreateInstance(Type itemType)
...
+        object? o = TypeDescriptor.CreateInstance(host, itemType, argTypes: null, args: null);
+
+        return o is null
+            ? throw new InvalidOperationException(
+                string.Format(
+                    SR.CollectionEditorCreateInstanceError,
+                    nameof(IDesignerHost),
+                    nameof(TypeDescriptor),
+                    itemType.FullName))
+            : o;
```
LeafShi1, line 125: Using `is {} obj` can more simpler

```
      if (TypeDescriptor.CreateInstance(host, itemType, argTypes: null, args: null) is { } obj)
      {
          return obj;
      }

      throw new InvalidOperationException(
          string.Format(
          SR.CollectionEditorCreateInstanceError,
          nameof(IDesignerHost),
          nameof(TypeDescriptor),
          itemType.FullName));
```

lonitra (approved):
👍

PR #11298, opened 2024-05-01
Enable nullability in TransparentBehavior

gpetrou:
## Proposed changes

- Enable nullability in TransparentBehavior.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11298)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.TransparentBehavior.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.TransparentBehavior.cs
@@ -1,8 +1,6 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
-#nullable disable
-
 using System.Drawing;
 using System.Windows.Forms.Design.Behavior;
 
@@ -23,7 +21,7 @@ public partial class ControlDesigner
         /// <summary>
         ///  Constructor that accepts the related ControlDesigner.
         /// </summary>
-        internal TransparentBehavior(ControlDesigner designer) => _designer = designer;
+        internal TransparentBehavior(ControlDesigner designer) => _designer = designer.OrThrowIfNull();
 
         /// <summary>
         ///  This property performs a hit test on the ControlDesigner to determine if the BodyGlyph should return
@@ -34,7 +32,7 @@ public partial class ControlDesigner
         /// <summary>
         ///  Forwards DragDrop notification from the BehaviorService to the related ControlDesigner.
         /// </summary>
-        public override void OnDragDrop(Glyph g, DragEventArgs e)
+        public override void OnDragDrop(Glyph? g, DragEventArgs e)
         {
             _controlRect = Rectangle.Empty;
             _designer.OnDragDrop(e);
@@ -43,11 +41,11 @@ public partial class ControlDesigner
         /// <summary>
         ///  Forwards DragDrop notification from the BehaviorService to the related ControlDesigner.
         /// </summary>
-        public override void OnDragEnter(Glyph g, DragEventArgs e)
+        public override void OnDragEnter(Glyph? g, DragEventArgs e)
         {
-            if (_designer is not null && _designer.Control is not null)
+            if (_designer.Control is { } control)
             {
-                _controlRect = _designer.Control.RectangleToScreen(_designer.Control.ClientRectangle);
+                _controlRect = control.RectangleToScreen(control.ClientRectangle);
             }
 
             _designer.OnDragEnter(e);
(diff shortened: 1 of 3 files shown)


src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.TransparentBehavior.cs
```diff
@@ -17,55 +15,55 @@ public partial class ControlDesigner
...
         /// <summary>
         ///  Forwards DragDrop notification from the BehaviorService to the related ControlDesigner.
         /// </summary>
-        public override void OnDragEnter(Glyph g, DragEventArgs e)
+        public override void OnDragEnter(Glyph? g, DragEventArgs e)
         {
-            if (_designer is not null && _designer.Control is not null)
+            if (_designer?.Control is not null)
             {
                 _controlRect = _designer.Control.RectangleToScreen(_designer.Control.ClientRectangle);
```
lonitra, line 48:
```suggestion
            if (_designer?.Control is { } control)
            {
                _controlRect = control.RectangleToScreen(_designer.Control.ClientRectangle);
```

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.TransparentBehavior.cs
```diff
@@ -23,7 +21,7 @@ internal class TransparentBehavior : Behavior.Behavior
         /// <summary>
         ///  Constructor that accepts the related ControlDesigner.
         /// </summary>
-        internal TransparentBehavior(ControlDesigner designer) => _designer = designer;
+        internal TransparentBehavior(ControlDesigner designer) => _designer = designer.OrThrowIfNull();
```
LeafShi1, line 24: The designer here can be changed to be nullable `ControlDesigner? designer` because `designer.OrThrowIfNull()` is added, so that you can avoid changing the Public method in `ControlBodyGlyph.cs`

gpetrou: This was suggested in https://github.com/dotnet/winforms/pull/11298#discussion_r1588480259. @lonitra should we revert the changes?

lonitra: Hmm.. I see that `ControlBodyGlyph` constructor has nullable `ControlDesigner` as a param. To me it does not make much sense for the designer to be nullable. In all instances that we call `ControlBodyGlyph` constructor we ensure that it is called with nonnull designer value, so I am leaning towards that we need to change the constructor for `ControlBodyGlyph`'s `designer` value to be nonnullable. @Tanya-Solyanik , @KlausLoeffelmann are there any cases that I may be missing where a null `ControlDesigner` is valid when constructing a `ControlBodyGlyph`?

Tanya-Solyanik: I don't think you are missing anything, this looks like an oversight in the previous changes, or a fallout from adding nullability chunk by chunk.

LeafShi1 (approved):

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.TransparentBehavior.cs
LeafShi1: I see, when developers call `ControlBodyGlyph ` they should get a warning or suggestion that Designer cannot be null, and if they still force a null value, the exception will be caught here. This does make more sense.

lonitra: Agree, fallout from adding nullability chunk by chunk seems to be likely what happened here. @gpetrou could we go ahead and update `ControlBodyGlyph`'s constructor to have nonnullable `ControlDesigner`?

lonitra (changes requested):

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.TransparentBehavior.cs
gpetrou: Isn't that already done? Or you mean something else?

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.TransparentBehavior.cs
```diff
@@ -56,16 +54,16 @@ public override void OnDragEnter(Glyph g, DragEventArgs e)
         /// <summary>
         ///  Forwards DragDrop notification from the BehaviorService to the related ControlDesigner.
         /// </summary>
-        public override void OnDragLeave(Glyph g, EventArgs e)
+        public override void OnDragLeave(Glyph? g, EventArgs e)
         {
             _controlRect = Rectangle.Empty;
-            _designer.OnDragLeave(e);
+            _designer?.OnDragLeave(e);
```
lonitra, line 60: `_designer` is now never null here so I don't think we need the ? anymore

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.TransparentBehavior.cs
lonitra: Oh, I thought it was not added yet, Apologies seems I had missed the change.

lonitra (approved):

PR #11300, opened 2024-05-01
8768 :: Resolve Addrange issue after sorting.

### LINKED ISSUE #8768: TreeNodeCollection.AddRange works incorrectly for non-empty collection of the sorted TreeView

RutulPatel8:
Fixes #8768 

## Proposed changes
if tree is already sorted then we don't want to set fixed Index value. due to these reason It is misbehave. 

## Customer Impact

- No

## Regression? 

- Yes

## Risk

- No

### Before

When the user adds a value to the treenode list and sorts it using treeview.sort(), it works fine. Subsequently, when the user adds individual trees through the add event, it functions as expected and sorts the entire tree automatically. However, if the user adds multiple values through AddRange, it does not work as expected.

### After
I've fixed the sorting issue we discussed earlier, and now everything is working smoothly. When users add values individually or use treeview.sort(), the sorting works fine. Even when they add multiple values at once using AddRange, the sorting behaves as expected.

## Test methodology 

- Testcase added

## Test environment(s)

- tested with .net core 8.0

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11300)

### DIFF
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
@@ -305,6 +305,109 @@ public class ComboBoxTests
         Assert.False(property.ShouldSerializeValue(control));
     }
 
+    [WinFormsFact]
+    public void VerifyAutoCompleteEntries()
+    {
+        void AssertAutoCompleteCustomSource(string[] items, bool isHandleCreated)
+        {
+            using ComboBox control = new();
+
+            if (items is not null)
+            {
+                AutoCompleteStringCollection autoCompleteCustomSource = new();
+                autoCompleteCustomSource.AddRange(items);
+                control.AutoCompleteCustomSource = autoCompleteCustomSource;
+                control.AutoCompleteCustomSource.Should().BeEquivalentTo(autoCompleteCustomSource);
+            }
+            else
+            {
+                control.AutoCompleteCustomSource = null;
+                control.AutoCompleteCustomSource.Should().NotBeNull();
+                control.AutoCompleteCustomSource.Count.Should().Be(0);
+            }
+
+            control.IsHandleCreated.Should().Be(isHandleCreated);
+        }
+
+        AssertAutoCompleteCustomSource(new[] { "item1", "item2" }, false);
+        AssertAutoCompleteCustomSource(null, false);
+        AssertAutoCompleteCustomSource(new[] { "item3", "item4" }, false);
+    }
+
+    [WinFormsFact]
+    public void ComboBox_BeginEndUpdate()
+    {
+        using ComboBox control1 = new();
+        control1.BeginUpdate();
+        control1.EndUpdate();
+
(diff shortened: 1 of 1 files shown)


Tanya-Solyanik:
@RutulPatel8  - please investigate test failures and take a look at another PR in TreeNodesCollection area that might be addressing this regression as well
https://github.com/dotnet/winforms/pull/11253.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TreeNodeCollectionTests.cs
```diff
@@ -214,6 +214,39 @@ public void TreeNodeCollection_Find_InvokeNoSuchKey_ReturnsEmpty(string key)
...
+        };
+        TreeNode child3 = new()
+        {
+            Name = "7"
+        };
+
+        treeView.Nodes.Add(child1);
+        treeView.Nodes.Add(child2);
+        treeView.Nodes.Add(child3);
+
```
Tanya-Solyanik, line 238: After the nodes are added you can verify that they are not sorted. And after you sort them, you can verify that they are sorted

Tanya-Solyanik: As far as I understand, this issue repros only when AddRange is called from the Load event handler, i.e. when AddRange is called after the native TreeView had been initialized, please include that in this test.

Tanya-Solyanik (commented):
The approach to this fix seems correct, but this fix causes failures in TreeView tests, please investigate them. Also this kind on a change requires an opt-in switch, like the one @elachlan added here - https://github.com/dotnet/winforms/pull/10493/files

src/System.Windows.Forms/src/System/Windows/Forms/Controls/TreeView/TreeNodeCollection.cs
```diff
@@ -219,7 +219,11 @@ public virtual void AddRange(params TreeNode[] nodes)
             tv.BeginUpdate();
         }
 
-        _owner.Nodes.FixedIndex = _owner._childNodes.Count;
+        if (tv is not null && !tv.Sorted)
```
Tanya-Solyanik, line 222:
```suggestion
        if (tv is null || !tv.Sorted)
```

This is the reason of the test failures

RutulPatel8: Let me correct it.

Issue #10453, opened 2023-12-08
Ensure that all WinForms controls have unit tests coverage for public APIs


Tanya-Solyanik:
- [x]  Investigate if we can re-enable code coverage on internal builds
        We enabled it in #514, however Coverage option is disabled in the build pipeline -  https://github.com/dotnet/winforms/blob/95e83ed6e86c93c10020a7c92a6c9d459f65f6a5/eng/pipelines/build.yml#L33.
- [x] Generate list of controls with missing test coverage for public methods
- [ ] Work with @Olina-Zhang's team on adding the missing scenarios

Following controls are missing test coverage for public methods:
**System.Windows.Forms**
- [x] Button https://github.com/dotnet/winforms/pull/10693
- [x] CheckBox https://github.com/dotnet/winforms/pull/10693, https://github.com/dotnet/winforms/pull/11249
- [x] FlatButton https://github.com/dotnet/winforms/pull/10693, https://github.com/dotnet/winforms/pull/12882
- [x] RadioButton https://github.com/dotnet/winforms/pull/10693, https://github.com/dotnet/winforms/pull/11687
- [x] ComboBox https://github.com/dotnet/winforms/pull/11381, https://github.com/dotnet/winforms/pull/11533, https://github.com/dotnet/winforms/pull/11565, https://github.com/dotnet/winforms/pull/11699, https://github.com/dotnet/winforms/pull/12038
- [x] DataGridView https://github.com/dotnet/winforms/pull/11541, https://github.com/dotnet/winforms/pull/11556, https://github.com/dotnet/winforms/pull/11618, https://github.com/dotnet/winforms/pull/11691, https://github.com/dotnet/winforms/pull/12622, https://github.com/dotnet/winforms/pull/12814, https://github.com/dotnet/winforms/pull/12768, https://github.com/dotnet/winforms/pull/12766, https://github.com/dotnet/winforms/pull/12673
- [x] DateTimePicker https://github.com/dotnet/winforms/pull/11033, https://github.com/dotnet/winforms/pull/11106, https://github.com/dotnet/winforms/pull/11318, https://github.com/dotnet/winforms/pull/11385, https://github.com/dotnet/winforms/pull/11534, https://github.com/dotnet/winforms/pull/11463
- [x] GroupBox https://github.com/dotnet/winforms/pull/11688
[...]

PR #11317, opened 2024-05-07
Add unit test for CheckedListBox.CheckedIndexCollection

Olina-Zhang:
Related https://github.com/dotnet/winforms/issues/10453

## Proposed changes

- Add unit tests for CheckedListBox.CheckedIndexCollection to test its properties: Count, SyncRoot, IsSynchronized, IsFixedSize, IsReadOnly
- Add unit tests for CheckedListBox.CheckedIndexCollection to test its method: Add, Clear, Insert, Remove, RemoveAt, Contains, IndexOf, CopyTo, GetEnumerator

## Customer Impact

- None

## Regression? 

- No

## Risk

- Minimal

## Test methodology 

- Unit tests

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11317)

PR #11331, opened 2024-05-08
Add unit tests for DomainUpDown

Syareel-Sukeri:
related https://github.com/dotnet/winforms/issues/10453

## Proposed changes

- Add unit tests for DomainUpDown to test DownButton and UpButton methods: Verify the behavior of the DownButton and UpButton methods in the DomainUpDown class and how they change the SelectedIndex under different conditions
-  Add unit test for DomainUpDown to test ToString method: Test ToString method's output, ensuring it correctly reflects the state of the Items collection and the SelectedIndex.

## Customer Impact

- None

## Regression? 

- No

## Risk

- Minimal

## Test methodology 

- Unit tests

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11331)

### DIFF
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DomainUpDownTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DomainUpDownTests.cs
@@ -1395,6 +1395,127 @@ public class DomainUpDownTests
         Assert.False(control.IsHandleCreated);
     }
 
+    [WinFormsTheory]
+    [InlineData(0, 1, false)]
+    [InlineData(1, 2, false)]
+    [InlineData(2, 2, false)] 
+    [InlineData(2, 0, true)] 
+    public void DomainUpDown_DownButton_InvokeWithItems_ChangesSelectedIndex(int initialIndex, int expectedIndex, bool wrap)
+    {
+        using DomainUpDown control = new();
+        control.Items.Add("Item1");
+        control.Items.Add("Item2");
+        control.Items.Add("Item3");
+        control.SelectedIndex = initialIndex;
+        control.Wrap = wrap;
+
+        control.DownButton();
+
+        control.SelectedIndex.Should().Be(expectedIndex);
+    }
+
+    [WinFormsTheory]
+    [InlineData(0, 2, false)]
+    [InlineData(1, 2, false)]
+    [InlineData(2, 2, false)] 
+    [InlineData(2, 1, true)] 
+    public void DomainUpDown_DownButton_InvokeMultipleTimes_ChangesSelectedIndex(int initialIndex, int expectedIndex, bool wrap)
+    {
+        using DomainUpDown control = new();
+        control.Items.Add("Item1");
+        control.Items.Add("Item2");
+        control.Items.Add("Item3");
+        control.SelectedIndex = initialIndex;
+        control.Wrap = wrap;
+
+        control.DownButton();
+        control.DownButton(); 
+
+        control.SelectedIndex.Should().Be(expectedIndex);
+    }
+
+    [WinFormsTheory]
+    [InlineData(new string[] { "Item1", "Item2", "Item3" }, 1)]
+    [InlineData(new string[] { "ItemA", "ItemB", "ItemC", "ItemD" }, 3)]
+    public void DomainUpDown_ToString_Invoke_ReturnsExpected(string[] items, int selectedIndex)
+    {
+        using DomainUpDown control = new();
+        for (int i = 0; i < items.Length; i++)
+        {
+            string item = items[i];
+            control.Items.Add(item);
+        }
+
+        control.SelectedIndex = selectedIndex;
+
+        string s = control.ToString();
+
+        s.Should().Contain($"Items.Count: {items.Length}");
+        s.Should().Contain($"SelectedIndex: {selectedIndex}");
+    }
+
+    [WinFormsTheory]
+    [InlineData("Item1", 0)]
+    [InlineData("Item2", 1)]
+    public void DomainUpDown_UpButton_MatchFound(string text, int expectedIndex)
+    {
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DomainUpDownTests.cs
```diff
@@ -1395,6 +1395,127 @@ public void DomainUpDown_UpdateEditText_InvokeNotEmptyWithSelection_Success(bool
         Assert.False(control.IsHandleCreated);
     }
 
+    [WinFormsTheory]
```
ricardobossan, line 1398: How about implementing the [constructor and dispose pattern](https://xunit.net/docs/shared-context#constructor)? This would allow to share the context setup and cleanup code without sharing object instances between tests, reducing unnecessary repetition. e.g.:

Add a constructor with body:

```cs
using DomainUpDown control = new();
control.Items.AddRange(new string[] { "Item1", "Item2", "Item3" });
```

and a Dispose method clearing the items and disposing of the control.

As done in [this commit](https://github.com/dotnet/winforms/pull/11317/commits/5ea3dd602ab428d3dd20418e9076082173f50903).

LeafShi1:
LGTM! @lonitra Please review this PR.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DomainUpDownTests.cs
```diff
@@ -1395,6 +1395,121 @@ public void DomainUpDown_UpdateEditText_InvokeNotEmptyWithSelection_Success(bool
         Assert.False(control.IsHandleCreated);
     }
 
+     public class DomainUpDownTests1 : IDisposable
```
lonitra, line 1398: Could `_control` and the disposable pattern here be applied to the other tests in this file as well so that we can remove this inner class? If not, please give this a more descriptive name. Perhaps `DomainUpDownDisposableWithItems` and put this into a separate file (this should make the outer test class into a partial class)

Tanya-Solyanik: @lonitra  - do you know if we can use the disposable pattern in tests? Can the constructor be called on a non-STA thread and are we guaranteed that Dispose methos is called in the case of an exception?

Tanya-Solyanik: Test call only the constructor, not the CreateControl method, we should be good.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DomainUpDownTests.cs
```diff
@@ -7,142 +7,157 @@
...
-        Assert.Same(control.BindingContext, control.BindingContext);
-        Assert.Equal(BorderStyle.Fixed3D, control.BorderStyle);
-        Assert.Equal(control.PreferredHeight, control.Bottom);
-        Assert.Equal(new Rectangle(0, 0, 120, control.PreferredHeight), control.Bounds);
-        Assert.False(control.CanEnableIme);
-        Assert.False(control.CanFocus);
-        Assert.True(control.CanRaiseEvents);
-        Assert.True(control.CausesValidation);
-        Assert.False(control.ChangingText);
+        Assert.Null(_sub.ActiveControl);
```
Tanya-Solyanik, line 32: Please use fluent assertions.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DomainUpDownTests.cs
```diff
@@ -7,273 +7,278 @@
...
 
-public class DomainUpDownTests
+public class DomainUpDownTests : IDisposable
 {
+    private readonly DomainUpDown _control;
+    private readonly SubDomainUpDown _sub;
+
+    public DomainUpDownTests()
+    {
+        _control = new DomainUpDown();
```
Tanya-Solyanik, line 17:
```suggestion
        _control = new();
```

We prefer "target-typed new" syntax

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DomainUpDownTests.cs
```diff
@@ -1175,48 +1137,46 @@ public static IEnumerable<object[]> OnChanged_TestData()
...
+    callCount.Should().Be(1);
 
-        // Remove handler.
-        control.SelectedItemChanged -= handler;
-        control.OnSelectedItemChanged(source, eventArgs);
-        Assert.Equal(1, callCount);
+    // Remove handler.
+    _sub.SelectedItemChanged -= handler;
+    _sub.OnSelectedItemChanged(source, eventArgs);
+    callCount.Should().Be(1);
```
lonitra, line 1156: Please indent this

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DomainUpDownTests.cs
```diff
@@ -7,273 +7,278 @@
...
-public class DomainUpDownTests
+public class DomainUpDownTests : IDisposable
 {
+    private readonly DomainUpDown _control;
+    private readonly SubDomainUpDown _sub;
+
+    public DomainUpDownTests()
+    {
+        _control = new();
+        _sub= new();
```
lonitra, line 18:
```suggestion
        _sub = new();
```

lonitra (commented):
Minor comments, otherwise LGTM!

lonitra (approved):
Thank you

PR #11341, opened 2024-05-09
Extend PayloadReader APIs to meet the WinForms requirements

adamsitnik:
@JeremyKuhne I was working on the MSRCs today and managed to make only a small progress with transition to SafePayloadReader APIs. What I have here compiles, but you can revert the last revert to have a temporary state where some APIs have been transitioned and there are compiler errors.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11341)


Winforms.sln
```diff
@@ -177,6 +177,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "TrimTestBinaryDeserializati
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "BinaryFormatTests", "src\System.Private.Windows.Core\tests\BinaryFormatTests\BinaryFormatTests.csproj", "{57EC5288-9513-46CF-8FB7-626199690D90}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "System.Runtime.Serialization.BinaryFormat", "src\System.Runtime.Serialization.BinaryFormat\System.Runtime.Serialization.BinaryFormat.csproj", "{DC3DD357-7FB4-4384-92DC-5D34864D010B}"
```
adamsitnik, line 180: I've moved all the code to a separate project to ensure that we don't accidentally use any of the `internal` APIs

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/BinaryFormattedObjectExtensions.cs
```diff
@@ -484,23 +359,6 @@ static bool Get(BinaryFormattedObject format, [NotNullWhen(true)] out object? ex
             || format.TryGetPointF(out value)
             || format.TryGetNotSupportedException(out value);
 
-    /// <summary>
-    ///  Dereferences <see cref="MemberReference"/> records.
-    /// </summary>
-    public static IRecord Dereference(this IReadOnlyRecordMap recordMap, IRecord record) => record switch
```
adamsitnik, line 490: The PayloadReader API keeps references an implementation detail, all the public APIs simply dereference OOTB:

https://github.com/adamsitnik/SafePayloadReader/blob/f9beee7176ce0bf29fdea794d4f79de5c024103c/System.Runtime.Serialization.BinaryFormat/Records/ClassRecord.cs#L134

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Deserializer/ArrayRecordDeserializer.cs
```diff
@@ -110,4 +98,63 @@ internal override Id Continue()
...
+            return null;
+        }
+
+        Type expectedArrayType = arrayRecord.ArrayType switch
+        {
+            ArrayType.Rectangular => arrayRecordElementType.MakeArrayType(arrayRecord.Rank),
+            _ => arrayRecordElementType.MakeArrayType()
+        };
+
+        return arrayRecord.ToArray(expectedArrayType, maxLength: Array.MaxLength);
```
adamsitnik, line 149: PayloadReader  can instantiate arrays of selected primitive types, we just need to provide the expected type

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Deserializer/Deserializer.cs
```diff
@@ -230,27 +231,22 @@ object DeserializeNew(Id id)
...
-                    IPrimitiveTypeRecord primitiveArray => primitiveArray.GetPrimitiveArray(),
-                    _ => null
-                };
-
-                if (values is not null)
-                {
-                    _deserializedObjects.Add(arrayRecord.ObjectId, values);
-                    return values;
-                }
+                RecordType.BinaryObjectString => ((PrimitiveTypeRecord<string>)record).Value,
```
adamsitnik, line 238: Payload Reader API does not expose `BinaryObjectString ` or `MemberPrimitiveTypedRecord` but we do provide a public abstraction: `PrimitiveTypeRecord<T>` (so users need to learn fewer concepts)

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Deserializer/Deserializer.cs
```diff
@@ -230,27 +231,22 @@ object DeserializeNew(Id id)
...
-            IRecord record = _recordMap[id];
-            if (record is BinaryObjectString binaryString)
-            {
-                _deserializedObjects.Add(id, binaryString.Value);
-                return binaryString.Value;
-            }
+            SerializationRecord record = _recordMap[id];
 
-            if (record is ArrayRecord arrayRecord)
+            object? value = record.RecordType switch
```
adamsitnik, line 236: I could cast the record to all various types, but comparing the enum value is simpler and allows us to avoid plenty of allocations

src/System.Private.Windows.Core/tests/BinaryFormatTests/FormatTests/Common/PrimitiveTests.cs
```diff
@@ -1,18 +1,8 @@
...
-
 namespace FormatTests.Common;
 
 public abstract class PrimitiveTests<T> : SerializationTest<T> where T : ISerializer
 {
-    [Fact]
-    public void TypesWhenNeeded_Integer()
-    {
-        int value = 42;
-        Stream stream = Serialize(value, typeStyle: FormatterTypeStyle.TypesWhenNeeded);
```
adamsitnik, line 14: TypesWhenNeeded are not supported by design (and I hope it's acceptable)

src/System.Private.Windows.Core/tests/BinaryFormatTests/FormatTests/FormattedObject/ArrayTests.cs
```diff
@@ -17,49 +15,13 @@ public override void Roundtrip_ArrayContainingArrayAtNonZeroLowerBound()
         action.Should().Throw<NotSupportedException>();
     }
 
-    [Theory]
-    [MemberData(nameof(ArrayInfo_ParseSuccessData))]
-    public void ArrayInfo_Parse_Success(Stream stream, int expectedId, int expectedLength)
-    {
-        using BinaryReader reader = new(stream);
```
adamsitnik, line 24: I've removed these and similar tests as they belong to Payload Reader test suite

src/System.Private.Windows.Core/tests/BinaryFormatTests/FormatTests/FormattedObject/HashTableTests.cs
```diff
@@ -66,13 +66,14 @@ public void HashTable_CustomComparer()
...
-        BinaryFormattedObject format = new(Serialize(hashtable));
-        SystemClassWithMembersAndTypes systemClass = format.RootRecord.Should().BeOfType<SystemClassWithMembersAndTypes>().Subject;
-        systemClass.Name.Should().Be("System.Collections.Hashtable");
-        format[(MemberReference)systemClass["Comparer"]!].Should().BeOfType<SystemClassWithMembersAndTypes>().Which.Name.Should().Be("System.OrdinalComparer");
-        format[(MemberReference)systemClass["HashCodeProvider"]!].Should().BeOfType<ClassWithMembersAndTypes>().Which.Name.Should().Be("FormatTests.FormattedObject.HashtableTests+CustomHashCodeProvider");
-        format[(MemberReference)systemClass["Keys"]!].Should().BeOfType<ArraySingleObject>();
-        format[(MemberReference)systemClass["Values"]!].Should().BeOfType<ArraySingleObject>();
+        System.Windows.Forms.BinaryFormat.BinaryFormattedObject format = new(Serialize(hashtable));
+        ClassRecord systemClass = (ClassRecord)format.RootRecord;
+        systemClass.RecordType.Should().Be(RecordType.SystemClassWithMembersAndTypes);
```
adamsitnik, line 71: SystemClassWithMembersAndTypes is not public, `ClassRecord` is and by checking the `RecordType` we can check the actual type (I did not want to expose types that derive from `ClassRecord` to confuse the users and force them to learn all different types)

src/System.Private.Windows.Core/tests/BinaryFormatTests/FormatTests/Common/BasicObjectTests.cs
```diff
@@ -75,7 +75,7 @@ public void DeserializeStoredObjects(object value, TypeSerializableValue[] seria
             // Explicitly not supporting offset arrays
             from value in BinaryFormatterTests.RawSerializableObjects()
             from FormatterAssemblyStyle assemblyFormat in new[] { FormatterAssemblyStyle.Full, FormatterAssemblyStyle.Simple }
-            from FormatterTypeStyle typeFormat in new[] { FormatterTypeStyle.TypesAlways, FormatterTypeStyle.TypesWhenNeeded, FormatterTypeStyle.XsdString }
+            from FormatterTypeStyle typeFormat in new[] { FormatterTypeStyle.TypesAlways/*, FormatterTypeStyle.TypesWhenNeeded, FormatterTypeStyle.XsdString*/ }
```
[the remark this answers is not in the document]
JeremyKuhne: I don't want to support it, just made sure it was possible to do so.

src/System.Runtime.Serialization.BinaryFormat/Records/ArrayRecord.cs
```diff
@@ -28,7 +30,11 @@ private protected ArrayRecord(ArrayInfo arrayInfo)
     /// </summary>
     public ArrayType ArrayType => ArrayInfo.ArrayType;
 
-    internal override int ObjectId => ArrayInfo.ObjectId;
+    public abstract TypeName ElementTypeName { get; }
+
+    public abstract AssemblyNameInfo ElementTypeLibraryName { get; }
```
adamsitnik, line 35: these are the new public API that were needed: ElementTypeName and ElementTypeLibraryName

src/System.Runtime.Serialization.BinaryFormat/Records/BinaryArrayRecord.cs
```diff
@@ -200,7 +209,7 @@ private static Type MapElementType(Type arrayType)
             arrayNestingDepth++;
         }
 
-        if (PrimitiveTypes.Contains(elementType))
+        if (PrimitiveTypes.Contains(elementType) || (Nullable.GetUnderlyingType(elementType) is Type nullable && PrimitiveTypes.Contains(nullable)))
```
adamsitnik, line 212: Your test suite allowed me to discover a bug with arrays of nullables

src/System.Runtime.Serialization.BinaryFormat/Records/RectangularOrCustomOffsetArrayRecord.cs
```diff
@@ -144,14 +152,17 @@ internal override (AllowedRecordTypes allowed, PrimitiveType primitiveType) GetA
...
-        return memberTypeInfo.Infos[0].BinaryType switch
+        Type elementType = memberTypeInfo.Infos[0].BinaryType switch
         {
-            BinaryType.Primitive => new(MapPrimitive((PrimitiveType)memberTypeInfo.Infos[0].AdditionalInfo!), arrayInfo, memberTypeInfo, lengths, offsets, recordMap),
-            BinaryType.String => new(typeof(string), arrayInfo, memberTypeInfo, lengths, offsets, recordMap),
-            BinaryType.Object => new(typeof(object), arrayInfo, memberTypeInfo, lengths, offsets, recordMap),
-            BinaryType.SystemClass or BinaryType.Class => new(typeof(ClassRecord), arrayInfo, memberTypeInfo, lengths, offsets, recordMap),
-            _ => throw ThrowHelper.InvalidBinaryType(memberTypeInfo.Infos[0].BinaryType),
+            BinaryType.Primitive => MapPrimitive((PrimitiveType)memberTypeInfo.Infos[0].AdditionalInfo!),
+            BinaryType.PrimitiveArray => MapPrimitive((PrimitiveType)memberTypeInfo.Infos[0].AdditionalInfo!).MakeArrayType(),
```
adamsitnik, line 158: another test gap discovered during porting (jagged arrays with custom offsets ;) )

src/System.Runtime.Serialization.BinaryFormat/Records/SerializationRecord.cs
```diff
@@ -23,7 +23,7 @@ public abstract class SerializationRecord
 
     public abstract RecordType RecordType { get; }
 
-    internal virtual int ObjectId => NoId;
+    public virtual int ObjectId => NoId;
```
adamsitnik, line 26: this is another new public API that I had to expose

src/System.Runtime.Serialization.BinaryFormat/Records/SystemClassWithMembersAndTypesRecord.cs
```diff
@@ -44,30 +44,42 @@ internal SerializationRecord TryToMapToUserFriendly()
...
+                    double value when TypeName.FullName == typeof(double).FullName => Create(value),
+                    _ => this
+                };
+            }
+            else if (HasMember("value"))
+            {
+                return MemberValues[0] switch
+                {
+                    // there can be a value match, but no TypeName match
+                    long value when TypeName.FullName == typeof(IntPtr).FullName => Create(new IntPtr(value)),
```
adamsitnik, line 74: another test gap: root records that are simply `IntPtr` or `UIntPtr`

src/System.Windows.Forms/src/Properties/AssemblyInfo.cs
```diff
@@ -12,6 +12,7 @@
 [assembly: InternalsVisibleTo("System.Windows.Forms.UI.IntegrationTests, PublicKey=00000000000000000400000000000000")]
 [assembly: InternalsVisibleTo("ScratchProjectWithInternals, PublicKey=00000000000000000400000000000000")]
 [assembly: InternalsVisibleTo("ComDisabled.Tests, PublicKey=00000000000000000400000000000000")]
+[assembly: InternalsVisibleTo("BinaryFormatTests, PublicKey=00000000000000000400000000000000")]
```
adamsitnik, line 15: I've moved all your code related to records to `System.Windows.Forms` to avoid name conflicts in the `Core` project

src/System.Windows.Forms/src/System/Windows/Forms/BinaryFormat/ArrayInfo.cs
```diff
@@ -24,13 +24,6 @@ public ArrayInfo(Id objectId, Count length)
         ObjectId = objectId;
     }
 
-    public static Id Parse(BinaryReader reader, out Count length)
```
adamsitnik, line 27: I've removed all the code related to parsing and kept only things for writing

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/BinaryFormattedObject.TypeResolver.cs
```diff
@@ -15,71 +16,46 @@ internal sealed class DefaultTypeResolver : ITypeResolver
...
+        Type ITypeResolver.GetType(TypeName typeName, AssemblyNameInfo libraryName)
         {
-            if (libraryId == _lastLibraryId && string.Equals(typeName, _lastTypeName))
+            if (_types.TryGetValue((typeName.FullName, libraryName.FullName), out Type? cachedType))
             {
-                Debug.Assert(_lastType is not null);
-                return _lastType;
-            }
-
-            _lastLibraryId = libraryId;
```
JeremyKuhne, line 52: Last retrieved is an important perf optimization, we should try and keep that.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Deserializer/ArrayRecordDeserializer.cs
```diff
@@ -1,72 +1,58 @@
...
-            {
-                throw new NotSupportedException("Only arrays with zero offsets are supported.");
-            }
+            elementType = elementType.GetElementType()!;
+        }
 
-            _array = _arrayType is BinaryArrayType.Rectangular
-                ? Array.CreateInstance(_elementType, binaryArray.Lengths.ToArray())
-                : Array.CreateInstance(_elementType, arrayRecord.Length);
+        int[] lengths = new int[arrayRecord.Rank];
```
JeremyKuhne, line 42: This should be stackalloc'ed, should have made that more explicit in my own code.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Deserializer/ClassRecordSerializationInfoDeserializer.cs
```diff
@@ -32,33 +33,40 @@ internal sealed class ClassRecordSerializationInfoDeserializer : ClassRecordDese
...
+                    Deserializer.PendValueUpdater(new SerializationInfoValueUpdater(
+                        _classRecord.ObjectId,
+                        reference,
+                        _serializationInfo,
+                        memberName));
+                }
+
+                _serializationInfo.AddValue(memberName, memberValue);
             }
+            while (_memberNamesIterator.MoveNext());
```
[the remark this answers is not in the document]
JeremyKuhne: It does have to be maintained for this to be compatible.

I'm not sure guarding against a lot of member names and member data is that useful. Are they really going to be able to DOS with this? Wouldn't an upper cap on number of members be more practical? I can't imagine any real-world class having more than 64 members that get serialized, picking some large number beyond that might be more useful.

src/System.Private.Windows.Core/tests/BinaryFormatTests/FormatTests/Common/BasicObjectTests.cs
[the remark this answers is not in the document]
JeremyKuhne: The thing we would want to add back here is `FormatterTypeStyle.TypesAlways | FormatterTypeStyle.XsdString`. That always forces strings to go as records.

src/System.Private.Windows.Core/tests/BinaryFormatTests/FormatTests/Common/NullRecordTests.cs
```diff
@@ -1,256 +0,0 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
-// The .NET Foundation licenses this file to you under the MIT license.
-
-using System.Runtime.Serialization;
-using System.Windows.Forms;
-using System.Windows.Forms.BinaryFormat;
-
-namespace FormatTests.Common;
-
-public abstract class NullRecordTests<T> : SerializationTest<T> where T : ISerializer
```
JeremyKuhne, line 10: These bad input tests need to be retained. As a quick hack you can make them just have the raw data streams.

src/System.Private.Windows.Core/tests/BinaryFormatTests/FormatTests/Common/PrimitiveTests.cs
JeremyKuhne: I only kept the implementation to know that I could re-add if it was deemed necessary.

src/System.Private.Windows.Core/tests/BinaryFormatTests/FormatTests/FormattedObject/ArrayTests.cs
JeremyKuhne: They need to be fully covered, so should at least have a tracking issue to make sure we don't lose any of these scenarios.

src/System.Private.Windows.Core/tests/BinaryFormatTests/FormatTests/FormattedObject/HashTableTests.cs
JeremyKuhne: I didn't plan to make these public either.

src/System.Runtime.Serialization.BinaryFormat/Utils/RecordMap.cs
```diff
@@ -1,20 +1,38 @@
...
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Collections;
 using System.IO.Hashing;
 using System.Runtime.InteropServices;
 
 namespace System.Runtime.Serialization.BinaryFormat;
 
-internal sealed class RecordMap
+internal sealed class RecordMap : IReadOnlyDictionary<int, SerializationRecord>
```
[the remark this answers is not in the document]
JeremyKuhne: Yeah, I ended up making this IReadOnlyRecordMap

src/System.Private.Windows.Core/tests/BinaryFormatTests/FormatTests/Common/BasicObjectTests.cs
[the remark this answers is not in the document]
JeremyKuhne: Note that with that option tests fail again. There is what I believe to be a bug in the serializer where it will put out duplicate negative ids. My `RecordMap` allows for that as they aren't supposed to be referenced. I didn't try to constrain further, this may only happen for the root record, or there might be some other constraint.

src/System.Runtime.Serialization.BinaryFormat/Records/ArraySinglePrimitiveRecord.cs
JeremyKuhne, line 1: `ReadPrimitiveTypes` should be replaced with the implementation I have in `BinaryReaderExtensions.ReadPrimitiveArray`. `ToArray` needs to just pass back the existing array. Critical for the allocations. We need to be able to do the same for `BinaryArray` when it is primitive as well (think `byte[,]`), but that can come as a work item as long as the object model allows for it. I seem to recall some single dimensional primitive arrays ending up in `BinaryArray` as well, but I'm not positive.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Deserializer/ClassRecordSerializationInfoDeserializer.cs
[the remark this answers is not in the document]
adamsitnik: > It does have to be maintained for this to be compatible.

I tried to write a failing test for that and I've failed, it seems that the order is preserved as long as we don't mutate the dictionary afterwards: https://stackoverflow.com/a/64596145/5852046 But is it safe to rely on that assumption? I am not convinced.

We could use `OrderedDictionary` which offers `O(1)` lookups, but the inserts can be `O(n)` at worst, but the same is true for `Dictionary`. ([source](https://stackoverflow.com/a/2565514/5852046))

> Wouldn't an upper cap on number of members be more practical? 

I was thinking about introducing such limit: https://github.com/adamsitnik/SafePayloadReader/blob/f9beee7176ce0bf29fdea794d4f79de5c024103c/System.Runtime.Serialization.BinaryFormat/PayloadOptions.cs#L14

@GrabYourPitchforks Thoughts?

src/System.Private.Windows.Core/tests/BinaryFormatTests/FormatTests/Common/BasicObjectTests.cs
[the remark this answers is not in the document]
adamsitnik: > I don't want to support it, just made sure it was possible to do so.

Awesome! I've changed the code to throw `NotSupportedException` and removed the related code.

> `FormatterTypeStyle.TypesAlways | FormatterTypeStyle.XsdString`

Thanks for sharing that! I had no idea that `FormatterTypeStyle` an be used as `[Flags]` (it's not marked as such...)

> There is what I believe to be a bug in the serializer where it will put out duplicate negative ids

Thanks a lot for the hint, it has saved me a lot of debugging

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Deserializer/ArrayRecordDeserializer.cs
adamsitnik: > This should be stackalloc'ed

But there is no overload for `Array.CreateInstance` that accepts a span (only arrays?)?

From [apisof.net](https://apisof.net/catalog/6e066e154cccdfce98e6cd1215e926bd):

image

src/System.Runtime.Serialization.BinaryFormat/Records/ArraySinglePrimitiveRecord.cs
adamsitnik: > ToArray needs to just pass back the existing array

I am going to offer that in the API proposal. Something like [MoveToImmutable](https://learn.microsoft.com/en-us/dotnet/api/system.collections.immutable.immutablearray-1.builder.movetoimmutable?view=net-8.0) or a boolean flag for `ToArray`. Do you have any preferences?

> ReadPrimitiveTypes should be replaced with the implementation I have in BinaryReaderExtensions.ReadPrimitiveArray

That implementation may be faster, but it has some bugs:

1. It assumes that every `Stream` can return `Length` and `Position` while non-seekable streams will simply [throw](https://github.com/dotnet/runtime/blob/ddff593e32e3f5d469f41a06ff4287948077862f/src/libraries/System.Private.CoreLib/src/System/IO/FileStream.cs#L477-L480):
https://github.com/dotnet/winforms/blob/31176a4d203711e766b8b587414d04c27d8cc173/src/System.Private.Windows.Core/src/System/IO/BinaryReaderExtensions.cs#L51-L55


2. For certain types it works only for big endian:

https://github.com/dotnet/winforms/blob/31176a4d203711e766b8b587414d04c27d8cc173/src/System.Private.Windows.Core/src/System/IO/BinaryReaderExtensions.cs#L129

My goal is to get the API design accepted and merged. Once we get there we can talk about perf (if there are not other more important things to work on). What is important is that all the mentioned optimizations can be achieved with current public API shape.

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/BinaryFormattedObject.TypeResolver.cs
adamsitnik: `libraryId == _lastLibraryId` was much faster when the API was accepting `Id libraryId` (integer). Now the public API provides `AssemblyNameInfo` and we need to compare strings to check for equality (in this particular case).
 I doubt it would move the needle now, as what we have (a dictionary lookup) is very fast (`O(1)`)

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Deserializer/ArrayRecordDeserializer.cs
```diff
@@ -1,72 +1,58 @@
...
-            Debug.Assert(arrayRecord is ArraySingleObject);
-            _elementType = typeof(object);
-            _arrayType = BinaryArrayType.Single;
-            _array = Array.CreateInstance(_elementType, arrayRecord.Length);
-        }
-        else
+            ArrayType.Rectangular => _elementType.MakeArrayType(arrayRecord.Rank),
+            _ => _elementType.MakeArrayType()
+        };
+        // Tricky part: for arrays of classes/structs the following record allocates and array of class records
```
JeremyKuhne, line 32: @adamsitnik Is this allocating an additional array or is it just returning the internal data structure?

src/System.Private.Windows.Core/src/System/Windows/Forms/BinaryFormat/Deserializer/ArrayRecordDeserializer.cs
JeremyKuhne: A feature request is in order, perhaps. You can allocate arrays on the stack with https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-12#inline-arrays.

adamsitnik:
Closing in favor of https://github.com/dotnet/winforms/pull/11745

PR #11356, opened 2024-05-11
fix issue 11243 TreeNodeCollection.AddRange has been broken in .NET 8…

### LINKED ISSUE #11243: TreeNodeCollection.AddRange has been broken in .NET 8 and no longer results in TreeNodes being drawn at the correct location

Epica3055:
Fixes #11243 

### Before
image

### After
image

## Test methodology 

- 
-  manually and unit test
-


Olina-Zhang:
Tested this PR based on .NET 8.0, GH issue: https://github.com/dotnet/winforms/issues/11243 is fixed. And did Accessibility testing in Inspect and Narrator, no change compared with .NET 7.0, no regression issue found.

lonitra:
Looks like branding occured a few days ago. @Tanya-Solyanik double checking here -- is this good to merge now?

Issue #11334, opened 2024-05-08
Massive GDI (region) leak. Help needed.


kirsan31:
### .NET version

.Net7 .Net8 .Net9

### Did it work in .NET Framework?

Yes

### Did it work in any of the earlier releases of .NET Core or .NET 5+?

We didn't see these problems before .Net7.

### Issue description

For several month we are trying to investigate huge GDI (regions) leak in our app. This leak is critical because can reach GDI limit (10k) in one day.

- The leaking GDI objects are regions.
- This is not managed leak - we have no leaking managed objects in dumps.
- It's start leaking suddenly and most of the time continue to the the limit or app restart, but some time stopped.
- The leak is related to elements redraw because of regions of course and most of the time it's happens around rdp connect/disconnect.
- We can't repro this leak :(
- We once managed to catch this behavior using [performance HUD](https://www.microsoft.com/en-us/download/details.aspx?id=100813). This is quite problematic, because HUD slows down the working PC. And what we saw was very strange, it felt like something was starting to leak that had not leak before. Unfortunately, the call stack saved in HUD (`*.hudinsight`) does not show the names of the methods when viewed (**may be some one knew how to overcome this?**) :( And even those copied manually as text also turned out to be cropped due to the large size. Therefore, I will present here what I managed to get (sorry for this).

1. 479 leaked regions due to redrawing (almost all are system calls):  
![1](https://github.com/dotnet/winforms/assets/17767561/48908205-b923-4569-9075-b56a6765803b)
[1.csv](https://github.com/dotnet/winforms/files/15249049/1.csv)

2. 84 leaked regions due to close all opened child mdi forms. Closing all windows is done through the menu, when opened, the child menu is filled with open forms (15 in our case). Big part of it is leaking in `MenuStrip`  -> `Control.SetBoundsCore` -> `SetWindowPos`.  Call stacks `ToolStripDropDownItem.OnDropDownOpened` -> `EtwWriteTransfer` and `ToolStripDropDownItem.OnDropDownClosed` -> `EtwWriteTransfer` are full.
![01](https://github.com/dotnet/winforms/assets/17767561/3b89e3a0-3aea-4b69-8796-499eab7f2b56)
![02](https://github.com/dotnet/winforms/assets/17767561/6948d43b-a8fb-48e3-877d-1507e44e9919)
![03](https://github.com/dotnet/winforms/assets/17767561/8277ff01-1ed3-4d86-9824-214518a38023)
[...]

PR #11358, opened 2024-05-11
Fix ToolStrip memory leak due to MouseHoverTimer and #4808

### LINKED ISSUE #4808: ToolStripMenuItem can lead to memory leaks
* .NET Core Version: all up to latest (5.0.5)

**Problem description:**
`ToolStripMenuItem` can lead to memory leaks in some special conditions.

- Menu with submenu.
``` 
 windowsToolStripMenuItem 
  -> listToolStripMenuItem
```
- Populate and clear submenu in runtime. In fact, only cleaning is important.
```cs
private void WindowsToolStripMenuItem_DropDownOpened(object sender, EventArgs e)
{
    for (int i = 0; i < 4; i++)
        listToolStripMenuItem.DropDownItems.Add("MenuItem" + i);
}

private void windowsToolStripMenuItem_DropDownClosed(object sender, EventArgs e)
{
    while (listToolStripMenuItem.DropDownItems.Count > 0)
        listToolStripMenuItem.DropDownItems[listToolStripMenuItem.DropDownItems.Count - 1].Dispose();
}
```
In certain conditions (I think that is - open only main menu and not opening submenu `listToolStripMenuItem`) after closing main menu some of `listToolStripMenuItem` subitems will remain in memory. See video below.
This happens due to [`_displayedItems`](https://github.com/dotnet/winforms/blob/c9287e9ca780cd1828633fd3c6e43a6601896147/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs#L40) collection still have live references to removed menu items. I think this collection ( and may be `_overflowItems` too ) must be checked on item removal.

https://user-images.githubusercontent.com/17767561/115139643-acf23080-a03b-11eb-9be9-c0e4c2527f38.mp4

*Workaround:*
You can call `listToolStripMenuItem.DropDown.PerformLayout();` after items removal. This lead to call of [`SetDisplayedItems()`](https://github.com/dotnet/winforms/blob/c9287e9ca780cd1828633fd3c6e43a6601896147/src/System.Windows.Forms/src/System/Windows/Forms/ToolStrip.cs#L4561) which will clear `_displayedItems` of removed elements.

**Expected behavior:**
All removed and disposed menu elements must be eligible to GC (have no live references).

**Minimal repro:**
[WinFormsCoreTest_Grid_Menu.zip](https://github.com/dotnet/winforms/files/6330797/WinFormsCoreTest_Grid_Menu.zip)

kirsan31:
Fix #4808 and problem that was mentioned in #11334.

## Proposed changes
Replace `ToolStripItem? _currentItem` with `WeakReference<ToolStripItem?> _currentItem` in `MouseHoverTimer`
Implement [this suggestion.](https://github.com/dotnet/winforms/issues/4808#issuecomment-1849319594)

## Customer Impact
No more memory leak if `ToolStripItem` will be disposed after MouseHoverTimer start. 
No more memory leak of chilled elements due to `DisplayedItems` and `OverflowItems` collections.. 

## Regression? 

- No

## Risk
Minimal.

## Screenshots 
`ToolStripItem` below was disposed, but remain in memory:
timer1
timer2

## Test methodology 
- Existing unit tests.
- New unit tests.
- Manual with test app from #4808.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11358)


kirsan31:
Added commit that fix #4808 (thanks to @elachlan).

elachlan:
@Olina-Zhang Could your team please test this pr to confirm the issues are resolved?

Olina-Zhang:
> @Olina-Zhang Could your team please test this pr to confirm the issues are resolved?

@elachlan @kirsan31 Tested this PR, just can confirm that GH issue https://github.com/dotnet/winforms/issues/4808 is fixed, I am not clear how to verify https://github.com/dotnet/winforms/issues/11334, can you give some more information about how to repro it?

- GH issue https://github.com/dotnet/winforms/issues/4808 validation
image

kirsan31:
@Olina-Zhang 

> can you give some more information about how to repro it?

Same repro app, the key is to remove item immediately after `MouseHoverTimer` has started:

https://github.com/dotnet/winforms/assets/17767561/aa6c7f35-347e-4661-9ead-bc17c88b02ab


image

Olina-Zhang:
@kirsan31 thanks for your info.
Adding the GH issue https://github.com/dotnet/winforms/issues/11334 validation: fixed
image

src/System.Windows.Forms/src/System/Windows/Forms/Controls/ToolStrips/ToolStrip.cs
```diff
@@ -4779,9 +4779,16 @@ internal void OnItemAddedInternal(ToolStripItem item)
...
 
-    internal void OnItemRemovedInternal(ToolStripItem item)
+    internal void OnItemRemovedInternal(ToolStripItem item, ToolStripItemCollection itemCollection)
     {
         KeyboardToolTipStateMachine.Instance.Unhook(item, ToolTip);
+        if (itemCollection == _toolStripItemCollection)
+        {
+            // To prevent memory leaks when item removed from main collection,
+            // we need to remove it from _displayedItems and _overflowItems too.
+            _displayedItems?.Remove(item);
```
Tanya-Solyanik, line 4789: Is it possible to make _displayedItems and _overflowItems collections of weak references instead?

I wonder if we have cases when item is removed but is not disposed and will be reused.

kirsan31: > Is it possible to make _displayedItems and _overflowItems collections of weak references instead?

That was the first thing I thought of. But there about 100 references, it will be too overcomplicated...

> I wonder if we have cases when item is removed but is not disposed and will be reused.

Both collections insertions appears only in `SetDisplayedItems` and both are cleared at start of the method:
https://github.com/dotnet/winforms/blob/0aa3a4dfea1dad493ecbfe15cd513604468b3916/src/System.Windows.Forms/src/System/Windows/Forms/Controls/ToolStrips/ToolStrip.cs#L4333-L4337

So, the logic must preserve.

Also. `SetDisplayedItems` are called from `OnLayout` and both `ToolStripItemCollection.Insert` and `ToolStripItemCollection.Add` are calling `OnItemAdded` -> `DoLayoutIfHandleCreated` -> ... `SetDisplayedItems`.


P.s. I noticed a strange difference in the `ToolStripItemCollection.Insert` and `ToolStripItemCollection.Add` methods:
https://github.com/dotnet/winforms/blob/0aa3a4dfea1dad493ecbfe15cd513604468b3916/src/System.Windows.Forms/src/System/Windows/Forms/Controls/ToolStrips/ToolStripItemCollection.cs#L106-L110
https://github.com/dotnet/winforms/blob/0aa3a4dfea1dad493ecbfe15cd513604468b3916/src/System.Windows.Forms/src/System/Windows/Forms/Controls/ToolStrips/ToolStripItemCollection.cs#L310-L324
this code is not present in `ToolStripItemCollection.Add`:
```cs
if (_owner.IsHandleCreated) 
{ 
   LayoutTransaction.DoLayout(_owner, value, PropertyNames.Parent); 
} 
else 
{ 
   // next time we fetch the preferred size, recalc it. 
   CommonProperties.xClearPreferredSizeCache(_owner); 
} 
```
🤔

kirsan31: > Do we own all items in these collections?

No of course. More of it I doubt that we own any of them.

elachlan: Keep in mind that we are removing the `ToolStripItem` from our collections here and not disposing of them. So we free up the references, but aren't responsible for the lifetime.

JeremyKuhne (dismissed):

JeremyKuhne (approved):

PR #11363, opened 2024-05-13
fix winforms-designer/issues/2707  The comboBox incorrect display on form designer when setting its DropDownStyle as Simple after building the project

Epica3055:
Fixes [designer/issues/2707](https://github.com/microsoft/winforms-designer/issues/2707)

## Root Cause

We use `_requestedHeight` to set the height of a ComboBox when `DropDownStyle` is `ComboBoxStyle.Simple`

https://github.com/dotnet/winforms/blob/70a6a7b4a9af452160e636838f946b3f4327383c/src/System.Windows.Forms/src/System/Windows/Forms/Controls/ComboBox/ComboBox.cs#L61-L63

Let's go through the repro steps that was described here https://github.com/microsoft/winforms-designer/issues/2707#issuecomment-799222994

The following steps also can repro this issue:

1. Create a Winforms .Net Core project with a ComboBox
2. Close and reopen form designer
3. Set the DropDownStyle of comboBox as Simple

Step I : Create a Winforms .Net Core project with a ComboBox

I think this is the generated code
``` c#
        private void InitializeComponent()
        {
            comboBox1 = new ComboBox();
            SuspendLayout();
            // 
            // comboBox1
            // 
            comboBox1.FormattingEnabled = true;
            comboBox1.Location = new Point(44, 71);
            comboBox1.Name = "comboBox1";
            comboBox1.Size = new Size(121, 23);
            comboBox1.TabIndex = 0;
```
Step II: Close and reopen form designer
First ComboBox constrictor will be invoked 

https://github.com/dotnet/winforms/blob/70a6a7b4a9af452160e636838f946b3f4327383c/src/System.Windows.Forms/src/System/Windows/Forms/Controls/ComboBox/ComboBox.cs#L116-L126

`_requestedHeight` will be assigned to `DefaultSimpleStyleHeight`, which is 150.

https://github.com/dotnet/winforms/blob/70a6a7b4a9af452160e636838f946b3f4327383c/src/System.Windows.Forms/src/System/Windows/Forms/Controls/ComboBox/ComboBox.cs#L47

then `comboBox1.Size = new Size(121, 23);` will be executed, which will involve `SetBoundsCore` method

https://github.com/dotnet/winforms/blob/70a6a7b4a9af452160e636838f946b3f4327383c/src/System.Windows.Forms/src/System/Windows/Forms/Controls/ComboBox/ComboBox.cs#L3371-L3382

then height(23) will be assigned to `_requestedHeight`.

Step III: Set the DropDownStyle of comboBox as Simple

please see line 1170-1173

https://github.com/dotnet/winforms/blob/70a6a7b4a9af452160e636838f946b3f4327383c/src/System.Windows.Forms/src/System/Windows/Forms/Controls/ComboBox/ComboBox.cs#L1135-L1177

So step 3 will trigger HandleRecreate.
please see line 2435-2438, so eventually `SetBoundsCore` method will be triggered. And `_requestedHeight`(23)  will be assigned to `Height`

https://github.com/dotnet/winforms/blob/70a6a7b4a9af452160e636838f946b3f4327383c/src/System.Windows.Forms/src/System/Windows/Forms/Controls/ComboBox/ComboBox.cs#L2380-L2469

So the parameter  height here is 23, which is supposed to be 150.

https://github.com/dotnet/winforms/blob/70a6a7b4a9af452160e636838f946b3f4327383c/src/System.Windows.Forms/src/System/Windows/Forms/Controls/ComboBox/ComboBox.cs#L3371-L3382

## Proposed changes

- 
- Add a condition when `_requestedHeight` is assigned.
- 

## Before

Core

## After

https://github.com/dotnet/winforms/assets/135201996/aaef3b5d-2ca0-4722-bbe1-f4b643eb7e1a

## Test methodology 

- 
- manually (build the binaries and copy System.windows.forms.dll to the program files\dotnet\shared)
-

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Controls/ComboBox/ComboBox.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Controls/ComboBox/ComboBox.cs
@@ -119,7 +119,7 @@ public partial class ComboBox : ListControl
                  ControlStyles.UseTextForAccessibility |
                  ControlStyles.StandardClick, false);
 
-        _requestedHeight = DefaultSimpleStyleHeight;
+        _requestedHeight = ScaleHelper.ScaleToInitialSystemDpi(DefaultSimpleStyleHeight);
 
         // this class overrides GetPreferredSizeCore, let Control automatically cache the result
         SetExtendedState(ExtendedStates.UserPreferredSizeCache, true);
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Controls/ComboBox/ComboBox.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Controls/ComboBox/ComboBox.cs
@@ -3333,8 +3333,7 @@ public partial class ComboBox : ListControl
     {
         // If we are changing height, store the requested height.
         // Requested height is used if the style is changed to simple.
-        // (
-        if ((specified & BoundsSpecified.Height) != BoundsSpecified.None)
+        if ((specified & BoundsSpecified.Height) != BoundsSpecified.None && DropDownStyle == ComboBoxStyle.Simple)
         {
             _requestedHeight = height;
         }
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
@@ -2645,6 +2645,7 @@ public class ComboBoxTests
 
         comboBox.DropDownStyle = ComboBoxStyle.Simple;
 
+        comboBox.DropDownStyle.Should().Be(ComboBoxStyle.Simple);
         handleCreatedInvoked.Should().Be(2);
     }
 
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
@@ -2631,7 +2631,7 @@ public class ComboBoxTests
     [WinFormsFact]
     public void ComboBox_CorrectHeightAfterSetDropDownStyleSimple()
     {
-        using ComboBox comboBox = new ComboBox();
+        using ComboBox comboBox = new();
 
         int handleCreatedInvoked = 0;
         comboBox.HandleCreated += (s, e) =>
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
@@ -2627,6 +2627,27 @@ public class ComboBoxTests
         control.SelectionLength.Should().Be(control.Text.Length);
     }
 
+    // Unit test for https://github.com/microsoft/winforms-designer/issues/2707
+    [WinFormsFact]
+    public void ComboBox_CorrectHeightAfterSetDropDownStyleSimple()
+    {
+        using ComboBox comboBox = new ComboBox();
+
+        int handleCreatedInvoked = 0;
+        comboBox.HandleCreated += (s, e) =>
+        {
+            handleCreatedInvoked++;
+        };
+
+        comboBox.CreateControl();
+
+        comboBox.DropDownStyle.Should().Be(ComboBoxStyle.DropDown);
+
+        comboBox.DropDownStyle = ComboBoxStyle.Simple;
+
+        handleCreatedInvoked.Should().Be(2);
+    }
+
     private void InitializeItems(ComboBox comboBox, int numItems)
     {
         for (int i = 0; i < numItems; i++)
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
@@ -2648,6 +2648,7 @@ public class ComboBoxTests
 
         comboBox.DropDownStyle = ComboBoxStyle.Simple;
 
+        // DefaultSimpleStyleHeight is 150 in ComboBox class
         comboBox.Height.Should().Be(150);
         comboBox.DropDownStyle.Should().Be(ComboBoxStyle.Simple);
         handleCreatedInvoked.Should().Be(2);
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
@@ -2639,12 +2639,16 @@ public class ComboBoxTests
             handleCreatedInvoked++;
         };
 
+        comboBox.Height.Should().Be(23);
+
         comboBox.CreateControl();
 
+        comboBox.Height.Should().Be(23);
         comboBox.DropDownStyle.Should().Be(ComboBoxStyle.DropDown);
 
         comboBox.DropDownStyle = ComboBoxStyle.Simple;
 
+        comboBox.Height.Should().Be(150);
         comboBox.DropDownStyle.Should().Be(ComboBoxStyle.Simple);
         handleCreatedInvoked.Should().Be(2);
     }


Tanya-Solyanik:
>I haven't tested it yet. because this change is made in runtime repo while Designer is in another repo.

build the binaries and copy System.windows.forms.dll to the program files\dotnet\shared

But more importantly, how do you prove that this is not regressing any runtime scenarios?

src/System.Windows.Forms/src/System/Windows/Forms/Controls/ComboBox/ComboBox.cs
```diff
@@ -119,7 +119,7 @@ public ComboBox()
                  ControlStyles.UseTextForAccessibility |
                  ControlStyles.StandardClick, false);
 
-        _requestedHeight = DefaultSimpleStyleHeight;
+        _requestedHeight = ScaleHelper.ScaleToInitialSystemDpi(DefaultSimpleStyleHeight);
```
Tanya-Solyanik, line 122: I agree that this is the correct change, not sure if it is related to this fix though.

src/System.Windows.Forms/src/System/Windows/Forms/Controls/ComboBox/ComboBox.cs
```diff
@@ -3372,8 +3372,7 @@ protected override void SetBoundsCore(int x, int y, int width, int height, Bound
     {
         // If we are changing height, store the requested height.
         // Requested height is used if the style is changed to simple.
-        // (
-        if ((specified & BoundsSpecified.Height) != BoundsSpecified.None)
+        if ((specified & BoundsSpecified.Height) != BoundsSpecified.None && DropDownStyle == ComboBoxStyle.Simple)
```
Tanya-Solyanik, line 3375: In this particular repro case, the height that is set in SetBoundsCore is not intended for the combo box with simple style. But simple combos are the only ones that are consuming custom heights because the rest are rendered by the OS based on the font height.  Why then is BoundsSpecified.Height flag on? I guess because we had done some DPI-related rescaling. I think the right fix would be to add flags to BoundsSpecified that would distinguish between changes caused by scaling and sizes requested by the developer.

Consider this sequence:
1. combo box with style other than simple is created
2. developer causes resizing
3. style changed to simple

I would guess that developer expects that the new size is applied, but you would have skipped it.

I think this issue should be revisited in NET10 when we get to the scaling work. But please create a PR with the other change.

Epica3055: winforms-designer_issues_2707_3

Tanya-Solyanik: Can the Height property be changes in code behind when style is set to simple?

Epica3055: I suppose you were saying that:

Can the Height property be changes in code behind when style is **not** set to simple?


this is before the change

https://github.com/dotnet/winforms/assets/135201996/a2855e8a-da0d-4c93-a431-c780d8b59dac



this is after the change

https://github.com/dotnet/winforms/assets/135201996/5d153642-7a47-40fc-b348-773132afde46

Tanya-Solyanik: I meant a runtime scenario, not in the designer, application can change the style in response to a button click for example.

1 show a not simple style combo
2 set the height
3 click a button and change the style to simple

Epica3055: I tried something like this:

``` c#
using System;
using System.Collections.Generic;
using System.ComponentModel;
using System.Data;
using System.Drawing;
using System.Linq;
using System.Text;
using System.Threading.Tasks;
using System.Windows.Forms;

namespace WinFormsApp5
{
    public partial class Form3 : Form
    {

        private System.ComponentModel.IContainer components = null;
        private Button button1;
        private ComboBox comboBox1;

        protected override void Dispose(bool disposing)
        {
            if (disposing && (components != null))
            {
                components.Dispose();
            }
            base.Dispose(disposing);
        }

        private void InitializeComponent()
        {
            button1 = new Button();
            comboBox1 = new ComboBox();
            SuspendLayout();
            // 
            // button1
            // 
            button1.Location = new Point(262, 68);
            button1.Name = "button1";
            button1.Size = new Size(94, 29);
            button1.TabIndex = 0;
            button1.Text = "button1";
            button1.UseVisualStyleBackColor = true;
            button1.Click += button1_Click;
            // 
            // comboBox1
            // 
            comboBox1.FormattingEnabled = true;
            comboBox1.Location = new Point(18, 49);
            comboBox1.Name = "comboBox1";
            comboBox1.Size = new Size(151, 28);
            comboBox1.TabIndex = 1;
            // 
            // Form3
            // 
            AutoScaleDimensions = new SizeF(8F, 20F);
            AutoScaleMode = AutoScaleMode.Font;
            ClientSize = new Size(394, 505);
            Controls.Add(comboBox1);
            Controls.Add(button1);
            Name = "Form3";
            Text = "Form3";
            ResumeLayout(false);
        }

        public Form3()
        {
            InitializeComponent();
            this.Shown += Form3_Shown;
        }

        private void Form3_Shown(object? sender, EventArgs e)
        {
            this.comboBox1.Height = 100;
        }

        private void button1_Click(object sender, EventArgs e)
        {
            this.comboBox1.DropDownStyle = ComboBoxStyle.Simple;
        }
    }
}

```

### before the change

https://github.com/dotnet/winforms/assets/135201996/a1e6cc18-deea-425f-8d0a-ff34552f5e18


### after the change

https://github.com/dotnet/winforms/assets/135201996/9d8ab08a-bcdf-4f63-9532-c812b735ca32


I think the height of `after the change` is not correct.
but I think we can address this problem by this
``` c#


        private void button1_Click(object sender, EventArgs e)
        {
            this.comboBox1.DropDownStyle = ComboBoxStyle.Simple;
            this.comboBox1.Height = 100;
        }
```

Epica3055: @Tanya-Solyanik hi, tanya. could you please take a look at this when you have time?

LeafShi1: After the modification, the logic of Designer and Runtime is consistent. Only after the DropdownStyle is set to Simple, it can record and change the height of the control. This is more make sense.

Tanya-Solyanik:
@Epica3055  - pleas add a unit test.

JeremyKuhne (dismissed):
This looks good.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
```diff
@@ -2627,6 +2627,27 @@ public void ComboBox_SelectAll_InvokeWithoutItems_Success()
         control.SelectionLength.Should().Be(control.Text.Length);
     }
 
+    // Unit test for https://github.com/microsoft/winforms-designer/issues/2707
+    [WinFormsFact]
+    public void ComboBox_CorrectHeightAfterSetDropDownStyleSimple()
+    {
+        using ComboBox comboBox = new ComboBox();
```
Tanya-Solyanik, line 2634:
```suggestion
        using ComboBox comboBox = new();
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
```diff
@@ -2627,6 +2627,27 @@ public void ComboBox_SelectAll_InvokeWithoutItems_Success()
         control.SelectionLength.Should().Be(control.Text.Length);
     }
 
+    // Unit test for https://github.com/microsoft/winforms-designer/issues/2707
+    [WinFormsFact]
+    public void ComboBox_CorrectHeightAfterSetDropDownStyleSimple()
```
Tanya-Solyanik, line 2632: You could set Height before changing style and verify that height is still the same after.

Epica3055: After the handle is created, height won't be the same

Tanya-Solyanik: Right, I forgot details of this issue :). Then test that they are not the same. Or better yet calculate what he new one should be and compare to the actual.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ComboBoxTests.cs
```diff
@@ -2627,6 +2627,32 @@ public void ComboBox_SelectAll_InvokeWithoutItems_Success()
...
+        comboBox.Height.Should().Be(23);
+
+        comboBox.CreateControl();
+
+        comboBox.Height.Should().Be(23);
+        comboBox.DropDownStyle.Should().Be(ComboBoxStyle.DropDown);
+
+        comboBox.DropDownStyle = ComboBoxStyle.Simple;
+
+        comboBox.Height.Should().Be(150);
```
Tanya-Solyanik, line 2651: Could you please add a comment that explains where 150 is coming from?

Tanya-Solyanik (approved):

lonitra (approved):

PR #11376, opened 2024-05-15
Add test coverage for MaskedTextBox

MelonWang1:
Related https://github.com/dotnet/winforms/issues/10453

## Proposed changes

- Add unit tests for MaskedTextBox to test its properties: MaskedTextBox_RejectInputOnFirstFailure, ResetOnPrompt, ValidatingType, PromptChar, PasswordChar.
- Add unit tests for MaskedTextBox to test its events: TypeValidationCompletedEvent.
- Add unit tests for MaskedTextBox to test its methods:
GetCharFromPosition, GetCharIndexFromPosition, GetPositionFromCharIndex, ValidateText.

### DIFF
--- a/src/System.Windows.Forms/tests/UnitTests/MaskedTextBoxTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/MaskedTextBoxTests.cs
@@ -2,12 +2,25 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.ComponentModel;
+using System.Drawing;
 using System.Globalization;
 
 namespace System.Windows.Forms.Tests;
 
-public class MaskedTextBoxTests
+public class MaskedTextBoxTests : IDisposable
 {
+    private readonly MaskedTextBox _maskedTextBox;
+
+    public MaskedTextBoxTests()
+    {
+        _maskedTextBox = new();
+    }
+
+    public void Dispose()
+    {
+        _maskedTextBox.Dispose();
+    }
+
     [WinFormsFact]
     public void MaskedTextBox_Constructor()
     {
@@ -28,6 +41,16 @@ public class MaskedTextBoxTests
         mtb.ValidatingType.Should().BeNull();
         mtb.TextAlign.Should().Be(HorizontalAlignment.Left);
         mtb.FormatProvider.Should().Be(null);
+        mtb.ResetOnPrompt.Should().BeTrue();
+        mtb.ValidatingType.Should().BeNull();
+        mtb.PromptChar.Should().Be('_');
+        mtb.GetLineFromCharIndex(2).Should().Be(0);
+        mtb.GetFirstCharIndexFromLine(100).Should().Be(0);
+        mtb.GetFirstCharIndexOfCurrentLine().Should().Be(0);
+        mtb.PasswordChar.Should().Be('\0');
+        mtb.RejectInputOnFirstFailure.Should().BeFalse();
+        mtb.Text.Should().Be(string.Empty);
+        mtb.UseSystemPasswordChar.Should().BeFalse();
 
         mtb.IsHandleCreated.Should().BeFalse();
     }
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/tests/UnitTests/MaskedTextBoxTests.cs
```diff
@@ -670,4 +693,184 @@ public void MaskedTextBox_TextAlignChangedEvent_AddRemove_Success()
...
+    [BoolData]
+    public void MaskedTextBox_RejectInputOnFirstFailure_GetSet_ReturnsExpected(bool value)
+    {
+        _maskedTextBox.RejectInputOnFirstFailure = value;
+        _maskedTextBox.RejectInputOnFirstFailure.Should().Be(value);
+        _maskedTextBox.IsHandleCreated.Should().BeFalse();
+
+        _maskedTextBox.RejectInputOnFirstFailure = value;
+        _maskedTextBox.RejectInputOnFirstFailure.Should().Be(value);
+        _maskedTextBox.IsHandleCreated.Should().BeFalse();
```
SimonZhao888, line 707: I don't think this step is necessary.

src/System.Windows.Forms/tests/UnitTests/MaskedTextBoxTests.cs
```diff
@@ -670,4 +693,184 @@ public void MaskedTextBox_TextAlignChangedEvent_AddRemove_Success()
...
+
+        _maskedTextBox.RejectInputOnFirstFailure = !value;
+        _maskedTextBox.RejectInputOnFirstFailure.Should().Be(!value);
+        _maskedTextBox.IsHandleCreated.Should().BeFalse();
+    }
+
+    [WinFormsTheory]
+    [BoolData]
+    public void MaskedTextBox_ResetOnPrompt_GetSet_ReturnsExpected(bool value)
```
SimonZhao888, line 716: We can merge the 'GetSet' tests for the attributes into one.

src/System.Windows.Forms/tests/UnitTests/MaskedTextBoxTests.cs
```diff
@@ -670,4 +693,184 @@ public void MaskedTextBox_TextAlignChangedEvent_AddRemove_Success()
...
+        _maskedTextBox.CreateControl();
+
+        Point position = _maskedTextBox.GetPositionFromCharIndex(4);
+        int result = _maskedTextBox.GetCharIndexFromPosition(position);
+
+        result.Should().Be(4);
+    }
+
+    [WinFormsFact]
+    public void MaskedTextBox_GetPositionFromCharIndex_ReturnsExpected()
```
SimonZhao888, line 812: We can merge `MaskedTextBox_GetPositionFromCharIndex_ReturnsExpected`,
`MaskedTextBox_GetCharIndexFromPosition_ReturnsExpected`,
`MaskedTextBox_GetCharFromPosition_ReturnsExpected` these 3 cases into one.

src/System.Windows.Forms/tests/UnitTests/MaskedTextBoxTests.cs
```diff
@@ -670,4 +693,184 @@ public void MaskedTextBox_TextAlignChangedEvent_AddRemove_Success()
...
+        _maskedTextBox.Text = "Hello, World!";
+        _maskedTextBox.CreateControl();
+
+        Point result = _maskedTextBox.GetPositionFromCharIndex(4);
+
+        result.Should().NotBe(Point.Empty);
+    }
+
+    [WinFormsFact]
+    public void MaskedTextBox_ValidateText_ReturnsExpected()
```
SimonZhao888, line 823: This test case can be merged into `MaskedTextBox_TypeValidationCompletedEvent_AddRemove_Success`.

src/System.Windows.Forms/tests/UnitTests/MaskedTextBoxTests.cs
```diff
@@ -670,4 +693,184 @@ public void MaskedTextBox_TextAlignChangedEvent_AddRemove_Success()
...
+    }
+
+    [WinFormsFact]
+    public void MaskedTextBox_PasswordChar_SetInvalid_ThrowsArgumentException()
+    {
+        Action act = () => _maskedTextBox.PasswordChar = '\t';
+        act.Should().Throw<ArgumentException>();
+    }
+
+    [WinFormsFact]
```
SimonZhao888, line 855: This test case duplicates the code in `MaskedTextBox_PromptChar_Set_GetReturnsExpected`.
Could you try to merge PasswordChar related test cases into one?

src/System.Windows.Forms/tests/UnitTests/MaskedTextBoxTests.cs
```diff
@@ -670,4 +693,184 @@ public void MaskedTextBox_TextAlignChangedEvent_AddRemove_Success()
...
+        _maskedTextBox.ResetOnPrompt = !value;
+        _maskedTextBox.ResetOnPrompt.Should().Be(!value);
+        _maskedTextBox.IsHandleCreated.Should().BeFalse();
+    }
+
+    [WinFormsTheory]
+    [InlineData(typeof(int))]
+    [InlineData(typeof(string))]
+    [InlineData(null)]
+    public void MaskedTextBox_ValidatingType_Set_GetReturnsExpected(Type value)
```
SimonZhao888, line 735: We can merge this case into the following one.

src/System.Windows.Forms/tests/UnitTests/MaskedTextBoxTests.cs
```diff
@@ -670,4 +693,184 @@ public void MaskedTextBox_TextAlignChangedEvent_AddRemove_Success()
...
+        int result = _maskedTextBox.GetCharIndexFromPosition(position);
+
+        result.Should().Be(4);
+    }
+
+    [WinFormsFact]
+    public void MaskedTextBox_GetPositionFromCharIndex_ReturnsExpected()
+    {
+        _maskedTextBox.Text = "Hello, World!";
+        _maskedTextBox.CreateControl();
```
Epica3055, line 815: Is this necessary?

src/System.Windows.Forms/tests/UnitTests/MaskedTextBoxTests.cs
```diff
@@ -670,4 +693,184 @@ public void MaskedTextBox_TextAlignChangedEvent_AddRemove_Success()
...
+    [InlineData('%', '&')]
+    public void MaskedTextBox_PromptChar_Set_GetReturnsExpected(char originalValue, char value)
+    {
+        _maskedTextBox.PromptChar = originalValue;
+        _maskedTextBox.PromptChar.Should().Be(originalValue);
+
+        _maskedTextBox.PromptChar = value;
+        _maskedTextBox.PromptChar.Should().Be(value);
+
+        _maskedTextBox.Handle.Should().NotBe(IntPtr.Zero);
```
Epica3055, line 775: please avoid using `.Handle` because it will force control to create Handle if handle is not created.
you can use `.IsHandleCreated`

src/System.Windows.Forms/tests/UnitTests/MaskedTextBoxTests.cs
MelonWang1: Yes, in this case, the GetPositionFromCharIndex, GetCharFromPosition, and GetCharIndexFromPosition rely on the underlying Windows control handles for their operation.

src/System.Windows.Forms/tests/UnitTests/MaskedTextBoxTests.cs
```diff
@@ -670,4 +693,184 @@ public void MaskedTextBox_TextAlignChangedEvent_AddRemove_Success()
...
+    [WinFormsTheory]
+    [InlineData(typeof(int))]
+    [InlineData(typeof(string))]
+    [InlineData(null)]
+    public void MaskedTextBox_ValidatingType_Set_GetReturnsExpected(Type value)
+    {
+        _maskedTextBox.ValidatingType = value;
+        _maskedTextBox.ValidatingType.Should().Be(value);
+        _maskedTextBox.IsHandleCreated.Should().BeFalse();
+    }
```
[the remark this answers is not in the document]
MelonWang1: This code doesn't need to be run twice because it's setting up the state for the test. Running it a second time wouldn't change the state of the MaskedTextBox or provide any additional verification. If the property involves some state changes that require repeated validation, we can write it twice. ValidatingType can be set once to validate most of the behavior.

src/System.Windows.Forms/tests/UnitTests/MaskedTextBoxTests.cs
```diff
@@ -2,12 +2,26 @@
...
 
 using System.ComponentModel;
+using System.Drawing;
 using System.Globalization;
+using System.Reflection;
 
 namespace System.Windows.Forms.Tests;
 
-public class MaskedTextBoxTests
+public class MaskedTextBoxTests : IDisposable
```
Tanya-Solyanik, line 11: I would prefer to create a new control in every test to be sure that we start in a clean state every time.

src/System.Windows.Forms/tests/UnitTests/MaskedTextBoxTests.cs
```diff
@@ -28,6 +42,16 @@ public void MaskedTextBox_Constructor()
         mtb.ValidatingType.Should().BeNull();
         mtb.TextAlign.Should().Be(HorizontalAlignment.Left);
         mtb.FormatProvider.Should().Be(null);
+        mtb.ResetOnPrompt.Should().BeTrue();
```
Tanya-Solyanik, line 45: I know that this is not part of your change, but since you are editing this file anyway, could you please rename this local variable to a full word - maskedTextBox?

src/System.Windows.Forms/tests/UnitTests/MaskedTextBoxTests.cs
Tanya-Solyanik: I see that test class constructor is executed before every test runs, but are we guaranteed that constructor runs on the STA thread and that Dispose is called even in the case of exception?

Tanya-Solyanik: I see that Dispose pattern is recommended by xUnit, and we are not creating control in the test constructor, so the handle is not created, that will be safe to use.

src/System.Windows.Forms/tests/UnitTests/MaskedTextBoxTests.cs
```diff
@@ -2,50 +2,74 @@
...
+    public void Dispose()
+    {
+        _maskedTextBox.Dispose();
+    }
+
     [WinFormsFact]
     public void MaskedTextBox_Constructor()
     {
-        using MaskedTextBox mtb = new();
+        using MaskedTextBox maskedTextBox = new();
```
lonitra, line 28: Since constructor is running before every test, let's use `_maskedTextBox` here instead of creating a new one.

lonitra (commented):
Small comment, otherwise LGTM

src/System.Windows.Forms/tests/UnitTests/MaskedTextBoxTests.cs
```diff
@@ -2,50 +2,72 @@
...
     }
 
     [WinFormsFact]
     public void MaskedTextBox_ConstructorMaskedTextProvider()
     {
-        using MaskedTextBox mtb = new(new MaskedTextProvider("Hello World!"));
+        using MaskedTextBox maskedTextBox = new(new MaskedTextProvider("Hello World!"));
 
-        Assert.NotNull(mtb);
+        Assert.NotNull(maskedTextBox);
```
Epica3055, line 70: and here

src/System.Windows.Forms/tests/UnitTests/MaskedTextBoxTests.cs
```diff
@@ -2,50 +2,72 @@
...
     }
 
     [WinFormsFact]
     public void MaskedTextBox_ConstructorString()
     {
-        using MaskedTextBox mtb = new("Hello World!");
+        using MaskedTextBox maskedTextBox = new("Hello World!");
 
-        Assert.NotNull(mtb);
+        Assert.NotNull(maskedTextBox);
```
Epica3055, line 62: and here

src/System.Windows.Forms/tests/UnitTests/MaskedTextBoxTests.cs
```diff
@@ -2,50 +2,72 @@
...
+        _maskedTextBox.Dispose();
+    }
+
     [WinFormsFact]
     public void MaskedTextBox_Constructor()
     {
-        using MaskedTextBox mtb = new();
-
-        Assert.NotNull(mtb);
+        Assert.NotNull(_maskedTextBox);
```
Epica3055, line 28: why not use fluent api here?

lonitra (approved):

PR #11384, opened 2024-05-16
Add test coverage for CheckedListBox.CheckedItemCollection and CheckedIndexCollection

Olina-Zhang:
Related https://github.com/dotnet/winforms/issues/10453

## Proposed changes

-  Add unit tests for CheckedListBox.CheckedItemCollection  to test its properties: SyncRoot, IsSynchronized, IsFixedSize, IsReadOnly
- Add unit tests for CheckedListBox.CheckedItemCollection to test its methods: Add, Clear, Insert, Remove, RemoveAt, Contains, IndexOf, CopyTo, GetEnumerator
- Update unit tests for CheckedListBox.CheckedIndexCollection to test its methods: Contains, IndexOf

### DIFF
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckedListBox.CheckedIndexCollectionTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckedListBox.CheckedIndexCollectionTests.cs
@@ -52,32 +52,49 @@ public class CheckedListBox_CheckedIndexCollectionTests: IDisposable
     }
 
     [WinFormsTheory]
-    [InlineData([new string[] { "item1", "item2" }, new bool[] { true, false }, 0, true])]
-    [InlineData([new string[] { "item1", "item2" }, new bool[] { true, false }, 1, false])]
-    [InlineData([new string[] { "item1", "item2" }, new bool[] { false, false }, 0, false])]
-    public void CheckedIndexCollection_Contains_ReturnsExpected(string[] items, bool[] checkedStatus, int index, bool expected)
+    [InlineData(new string[] { "item1", "item2", "item3" }, new bool[] { true, false, true }, 0, true)]
+    [InlineData(new string[] { "item1", "item2", "item3" }, new bool[] { true, false, true }, 1, false)]
+    [InlineData(new string[] { "item1", "item2", "item3" }, new bool[] { false, true, false }, 1, true)]
+    [InlineData(new string[] { "item1", "item2", "item3" }, new bool[] { false, false, false }, 0, false)]
+    [InlineData(new string[] { "item1", "item2", "item3" }, new bool[] { true, false, true }, "invalid", false)]
+    public void CheckedIndexCollection_ContainsIntAndObject_ReturnsExpected(string[] items, bool[] checkedStates, object index, bool expected)
     {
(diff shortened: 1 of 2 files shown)


ricardobossan (commented):
Other than small comment, LGTM.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckedListBox.CheckedItemCollectionTests.cs
```diff
@@ -0,0 +1,119 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Collections;
+
+namespace System.Windows.Forms.Tests;
+
+public class CheckedListBox_CheckedItemCollectionTests : IDisposable
```
Tanya-Solyanik, line 8: I would prefer to create controls for each test to ensure that we start with a clean control every time

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckedListBox.CheckedIndexCollectionTests.cs
```diff
@@ -52,32 +52,49 @@ public void CheckedIndexCollection_Methods_ThrowNotSupportedException()
...
         for (int i = 0; i < items.Length; i++)
         {
-            _checkedListBox.Items.Add(items[i], checkedStatus[i]);
+            _checkedListBox.Items.Add(items[i], checkedStates[i]);
         }
 
-        // Test Contains method
-        _collection.Contains(index).Should().Be(expected);
+        // Test Contains(int index) method
+        if (index is int indexAsInt)
```
Tanya-Solyanik, line 68: BTW, you could use `@` to escape the keyword name - `@int`

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/CheckedListBox.CheckedItemCollectionTests.cs
Tanya-Solyanik: I see that test class is created before every test is run, but are we guaranteed that the constructor is executed on the STA thread? and that DIspose method is always called even if the test fals?

Olina-Zhang: We are following this reference: [constructor and dispose pattern](https://xunit.net/docs/shared-context#constructor).
A TestInitialize like method does not exist in xunit, we have added an empty construct for its uniform operation.
As far as we know, when test cases are run in pipeline, they are randomly assigned to the machine each time and then executed by the vstester, one test case at a time. So, there will be no case of multiple test cases running at the same time. 
Are we going to continue to write like this or are we going to use the original creating controls for each test?

Tanya-Solyanik: Thank you for the pointer to docs. My other concern was that winforms controls should be used from the same thread they are created on. But in your cases you are not calling CreateControl, only the constructor. So Disposable pattern looks fine.

lonitra (approved):

PR #11385, opened 2024-05-16
Add unit test for DateTimePicker 12 properties

John-Qiao:
related https://github.com/dotnet/winforms/issues/10453

## Proposed changes

- Add unit tests for below 12 properties

1. CalendarFont
2. Checked
3. CustomFormat
4. DropDownAlign
5. MaxDate
6. MaximumDateTime
7. MinDate
8. MinimumDateTime
9. RightToLeftLayout
10. ShowUpDown
11. Text
12. Value

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11385)

### DIFF
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DateTimePickerTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DateTimePickerTests.cs
@@ -3,13 +3,26 @@
 
 using System.ComponentModel;
 using System.Drawing;
+using System.Globalization;
 using Point = System.Drawing.Point;
 using Size = System.Drawing.Size;
 
 namespace System.Windows.Forms.Tests;
 
-public class DateTimePickerTests
+public class DateTimePickerTests: IDisposable
 {
+    private readonly DateTimePicker _dateTimePicker;
+
+    public DateTimePickerTests()
+    {
+        _dateTimePicker = new();
+    }
+
+    public void Dispose()
+    {
+        _dateTimePicker.Dispose();
+    }
+
     [WinFormsFact]
     public void DateTimePicker_Ctor_Default()
     {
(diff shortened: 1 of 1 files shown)


src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DateTimePickerTests.cs
```diff
@@ -3,13 +3,26 @@
...
 using System.ComponentModel;
 using System.Drawing;
+using System.Globalization;
 using Point = System.Drawing.Point;
 using Size = System.Drawing.Size;
 
 namespace System.Windows.Forms.Tests;
 
-public class DateTimePickerTests
+public class DateTimePickerTests: IDisposable
```
Tanya-Solyanik, line 12: I would prefer to create a new control in every test to be sure that we start in a clean state every time.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DateTimePickerTests.cs
```diff
@@ -271,6 +284,170 @@ public void DateTimePicker_CalendarMonthBackground_SetInvalid_ThrowsArgumentExce
...
+    [WinFormsFact]
+    public void DateTimePicker_MinimumDateTime_ReturnsExpected()
+    {
+        DateTime minSupportedDateTime = CultureInfo.CurrentCulture.Calendar.MinSupportedDateTime;
+        DateTime expectedDate = minSupportedDateTime.Year < 1753 ? new(1753, 1, 1) : minSupportedDateTime;
+
+        DateTime result = DateTimePicker.MinimumDateTime;
+
+        result.Should().Be(expectedDate);
+    }
```
Epica3055, line 388: this is official site
> https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.datetimepicker.maximumdatetime?view=windowsdesktop-9.0
> 
> ## Remarks
> The maximum date is set to December 31, 9998.

I think we don't have to test it by using the same logic.

Maybe can simple write something like this: 
DateTimePicker.MinimumDateTime.Should().Be(new DateTime(December 31, 9998));

what do you think @Tanya-Solyanik

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DateTimePickerTests.cs
```diff
@@ -271,6 +284,170 @@ public void DateTimePicker_CalendarMonthBackground_SetInvalid_ThrowsArgumentExce
...
+
+    [WinFormsFact]
+    public void DateTimePicker_MinimumDateTime_ReturnsExpected()
+    {
+        DateTime minSupportedDateTime = CultureInfo.CurrentCulture.Calendar.MinSupportedDateTime;
+        DateTime expectedDate = minSupportedDateTime.Year < 1753 ? new(1753, 1, 1) : minSupportedDateTime;
+
+        DateTime result = DateTimePicker.MinimumDateTime;
+
+        result.Should().Be(expectedDate);
```
LeafShi1, line 387: It is enough to verify that `MinimumDateTime` is in the correct range

`Assert.True(DateTimePicker.MinimumDateTime >= new DateTime(1753, 1, 1, 0, 0, 0));`

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DateTimePickerTests.cs
```diff
@@ -271,6 +284,170 @@ public void DateTimePicker_CalendarMonthBackground_SetInvalid_ThrowsArgumentExce
...
+        _dateTimePicker.ShowUpDown = false;
+        _dateTimePicker.ShowUpDown.Should().Be(false);
+    }
+
+    [WinFormsFact]
+    public void DateTimePicker_Text_GetSet_ReturnsExpected()
+    {
+        string validDateString = "2022-01-01";
+        _dateTimePicker.Text = validDateString;
+        _dateTimePicker.Value.Should().Be(DateTime.Parse(validDateString, CultureInfo.CurrentCulture));
```
LeafShi1, line 419: Why not verify the value of the `_dateTimePicker.Text`?

LeafShi1 (changes requested):

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DateTimePickerTests.cs
Tanya-Solyanik: XUnit creates the new test class object before running the test, but I don't know if we are guaranteed that Dispose is called in the case of an exception, or if constructor is called on the same thread the test executes on.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DateTimePickerTests.cs
Tanya-Solyanik: It's possible that the min date is different on different cultures, I think it's ok to keep the more defensive logic.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DateTimePickerTests.cs
Tanya-Solyanik: I see that Dispose pattern is recommended by xUnit, and we are not creating control in the test constructor, so the handle is not created, that will be safe to use.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DateTimePickerTests.cs
John-Qiao: According to Tanya's suggestion, still using that original codes.

LeafShi1:
LGTM! @lonitra Please review this PR

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DateTimePickerTests.cs
```diff
@@ -271,6 +284,178 @@ public void DateTimePicker_CalendarMonthBackground_SetInvalid_ThrowsArgumentExce
...
+        _dateTimePicker.DropDownAlign = LeftRightAlignment.Left;
+        _dateTimePicker.DropDownAlign.Should().Be(LeftRightAlignment.Left);
+    }
+
+    [WinFormsFact]
+    public void DateTimePicker_MaxDate_GetSet_ReturnsExpected()
+    {
+        _dateTimePicker.MaxDate.ToString().Should().Be("12/31/9998 12:00:00 AM");
+
+        var expectedDate = new DateTime(2022, 12, 31);
```
lonitra, line 347: Please use target type new
```suggestion
        DateTime expectedDate = new(2022, 12, 31);
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DateTimePickerTests.cs
```diff
@@ -271,6 +284,178 @@ public void DateTimePicker_CalendarMonthBackground_SetInvalid_ThrowsArgumentExce
...
+
+        result.Should().Be(expectedDate);
+    }
+
+    [WinFormsFact]
+    public void DateTimePicker_MinDate_GetSet_ReturnsExpected()
+    {
+        _dateTimePicker.MinDate.ToString().Should().Be("1/1/1753 12:00:00 AM");
+
+        var expectedDate = new DateTime(2022, 1, 1);
```
lonitra, line 377:
```suggestion
        DateTime expectedDate = new(2022, 1, 1);
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DateTimePickerTests.cs
```diff
@@ -271,6 +284,178 @@ public void DateTimePicker_CalendarMonthBackground_SetInvalid_ThrowsArgumentExce
...
+        _dateTimePicker.Text.Should().Be(DateTime.Parse(DateTime.Now.Date.ToString(), CultureInfo.CurrentCulture).ToString("dddd, MMMM d, yyyy"));
+
+        _dateTimePicker.Text = string.Empty;
+        _dateTimePicker.Text.Should().Be(DateTime.Parse(DateTime.Now.Date.ToString(), CultureInfo.CurrentCulture).ToString("dddd, MMMM d, yyyy"));
+    }
+
+    [WinFormsFact]
+    public void DateTimePicker_Value_GetSet_ReturnsExpected()
+    {
+        var expectedDate = new DateTime(2022, 1, 1);
```
lonitra, line 438: Ditto here as well

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DateTimePickerTests.cs
```diff
@@ -3,13 +3,26 @@
...
 
-public class DateTimePickerTests
+public class DateTimePickerTests: IDisposable
 {
+    private readonly DateTimePicker _dateTimePicker;
+
+    public DateTimePickerTests()
+    {
+        _dateTimePicker = new();
+    }
```
lonitra, line 19:
```suggestion
    public DateTimePickerTests() => _dateTimePicker = new();
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DateTimePickerTests.cs
```diff
@@ -3,13 +3,26 @@
...
+
+    public void Dispose()
+    {
+        _dateTimePicker.Dispose();
+    }
```
lonitra, line 24:
```suggestion
    public void Dispose() => _dateTimePicker.Dispose();
```

lonitra (commented):
Few small comments, otherwise LGTM!

LeafShi1 (approved):

lonitra (approved):
Thank you!

PR #11401, opened 2024-05-19
Improve `ListViewSubItem.Text` performance

### LINKED ISSUE #10963: Updating ListView.ListViewItem.ListViewSubItem Text or Name values updates them all, why? is this normal?
### .NET version

Tested and Checked on .NET Runtime 7.0 and 8.0

Not tested/verified

n/a.

### Issue description

Talking about WinForms, ListView Control, ListViewItem and ListViewSubItems

when you have a ListViewItem , lets say, with 10 ListViewSubItems, on a Details view, and you change the text, it triggers a update text on ALL of the subitems.

If you check on the Text Setter, it reads as follows:

ListViewSubItem class
```
        public string Text {
            get => text ?? string.Empty;
            set {
                text = value;
                _owner?.UpdateSubItems(-1);
            }
        }

```

_owner being the ListViewItem, so if we go there, we find this:

```
internal void UpdateSubItems(int index) => UpdateSubItems(index, SubItemCount);
internal void UpdateSubItems(int index, int oldCount) {
    // trim down, not modified, just formated.
    if (_listView is not null && _listView.IsHandleCreated) {
        int subItemCount = SubItemCount;
        int itemIndex = Index;

        if (index != -1) {
            _listView.SetItemText(itemIndex, index, _subItems[index].Text);
        } else {
            for (int i = 0; i < subItemCount; i++) {
                _listView.SetItemText(itemIndex, i, _subItems[i].Text);
            }
        }

        for (int i = subItemCount; i < oldCount; i++) {
            _listView.SetItemText(itemIndex, i, string.Empty);
        }
    }
}

```

the ListView itself is then the one in charge of triggering the PInvoke.SendMessage with each of the subitems actual text.

So my question is, why does this happen? and even worst, why does changing the NAME of the item also triggers an update on the text?

Seems a little bit weird and overwhelming, but then again I am just checking and making sure this is not a problem..

I have found some laggy and performance issues and found myself that the ListView was killing the UI Thread with message loops because I am updating a text on a subitem on column 20 , of course, it then updates all 30 of them which are all the columns I have.

### Steps to reproduce

Create a WinForms app with a ListView on it, add a couple of ListViewItems with several subitems, and try to change the text of Nth subitem only and check how many of them get triggered to update.

elachlan:
Fixes #10963

Change `ListViewSubItem.Text` to pass index to `ListViewItem.UpdateSubItems` to avoid loop.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11401)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Controls/ListView/ListViewItem.ListViewSubItem.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Controls/ListView/ListViewItem.ListViewSubItem.cs
@@ -206,7 +206,7 @@ public partial class ListViewItem
             set
             {
                 text = value;
-                _owner?.UpdateSubItems(-1);
+                _owner?.UpdateSubItems(_owner.SubItems.IndexOf(this));
             }
         }
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Controls/ListView/ListViewItem.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Controls/ListView/ListViewItem.cs
@@ -1319,26 +1319,31 @@ public partial class ListViewItem : ICloneable, ISerializable
 
     internal void UpdateSubItems(int index, int oldCount)
     {
-        if (_listView is not null && _listView.IsHandleCreated)
+        if (_listView is null || !_listView.IsHandleCreated)
         {
-            int subItemCount = SubItemCount;
-            int itemIndex = Index;
-            if (index != -1)
-            {
-                _listView.SetItemText(itemIndex, index, _subItems[index].Text);
-            }
-            else
-            {
-                for (int i = 0; i < subItemCount; i++)
-                {
-                    _listView.SetItemText(itemIndex, i, _subItems[i].Text);
-                }
-            }
+            return;
+        }
 
-            for (int i = subItemCount; i < oldCount; i++)
+        int subItemCount = SubItemCount;
+        int itemIndex = Index;
+        if (index != -1)
+        {
+            // Update the specified subitem text.
+            _listView.SetItemText(itemIndex, index, _subItems[index].Text);
+        }
+        else
+        {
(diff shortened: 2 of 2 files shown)


elachlan:
@Olina-Zhang can your team please test this?

Olina-Zhang:
Hi @elachlan, tested before and after this PR change, I didn't see any behavior difference when updating one ListViewSubItem's Text property, noticed entire row is blinking once(The recorded video doesn't seem to show it), is it the issue? Not sure if I understand the client's question correctly.

https://github.com/dotnet/winforms/assets/26474449/72f65fe2-7e5f-4a5e-97bf-d40396ffa6e6

elachlan:
Hi @Olina-Zhang, That is fantastic. There shouldn't be any behavior changes. 

The issue will arise if there is a large number of sub items (1000) it will be very slow to update. After the change it should be very quick.

Olina-Zhang:
That's because my test was a bit small, I'll try a bigger one.

Olina-Zhang:
If ListView has a large number of sub items (>1000), updating the text of ListViewSubItem is also quite fast with before and after your PR change, looks like no difference.

https://github.com/dotnet/winforms/assets/26474449/e21f1014-a2e0-4b6f-be9d-c13b2d72a044

Tanya-Solyanik (approved):
The change looks reasonable. Probably updating text for all subitems was an oversight in the original implementation.

John-Qiao:
Verified this issue on 9.0.100-preview.5.24273.4 with dlls built from winforms repo of main branch, the result is same as above which Olina did.

https://github.com/dotnet/winforms/assets/45864985/1586d2ba-6809-4396-9d24-542d1097ac79

John-Qiao:
Verified this issue with 9.0.100-preview.5.24304.3 test pass build, it was fixed. Test result is same as above.

Issue #11180, opened 2024-04-09
[Tracking] BindingContext getter isn't trim compatible


LakshanF:
The trimming pattern in WinForms that we are striving for, is to use a feature switch to allow folks to trim feature areas if they do not plan to take advantage of certain capabilities that are not trim compatible. The feature switch will ensure that these capabilities are trimmed in the application and attempting to use these capabilities will throw at runtime. But we would like to generate a warning at build time if they do take advantage of these capabilities and do trim.

Binding is an example where we try to follow this pattern. `BindingContext` allows binding and will generate trim warnings, given current binding isn't trim compatible. We check if `BindingContext` is not null ([here ](https://github.com/dotnet/winforms/blob/0e3964e75d345cdb4d849e676d97042cf3e5f723/src/System.Windows.Forms/src/System/Windows/Forms/Controls/ListControl/ListControl.cs#L681)and [here](https://github.com/dotnet/winforms/pull/11175) on `ContainerControl`) if the feature switch is on. To address case to warn developers who want to use traditional binding and trim, we have annotated the `BindingContext` setter with [RUC](https://github.com/dotnet/winforms/blob/0e3964e75d345cdb4d849e676d97042cf3e5f723/src/System.Windows.Forms/src/System/Windows/Forms/DataBinding/IBindableComponent.cs#L15). This works in the current WinForms code base on the 2 instances mentioned above. But it would be possible to workaround this pattern in WinForms via setting the `BindingContext` outside of the setter (in the getter for example).

LakshanF:
This has been fixed with #11407

PR #11407, opened 2024-05-20
Get TrimTest to 0 trim warnings

LakshanF:
Use the new `ComponentModel` type registration [APIs](https://github.com/dotnet/runtime/pull/102094) to make the `TrimTest` project trim compatible.
- Use a feature switch flag to use the new `ComponentModel` APIs (non-trimmed applications will continue to use the legacy `ComponentModel` APIs)
- Some features of WinForms are not supported in trimming and are disabled via feature switches (a complete set can be seen in [TrimTest](url) project)
- Primitive types (string, int, bool, double) are not registered with `ComponentModel` since these are assumed to be supported by default by `ComponentModel`
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11407)


src/System.Windows.Forms/src/System/Drawing/Design/UITypeEditor.cs
```diff
@@ -14,6 +14,11 @@ namespace System.Drawing.Design;
 /// </summary>
 public class UITypeEditor
 {
+    [FeatureSwitchDefinition("System.Drawing.Design.UITypeEditor.IsSupported")]
+#pragma warning disable IDE0075 // Simplify conditional expression - the simpler expression is hard to read
+    private static bool IsSupported => AppContext.TryGetSwitch("System.Drawing.Design.UITypeEditor.IsSupported", out bool isSupported) ? isSupported : true;
```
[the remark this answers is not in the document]
Tanya-Solyanik: Would `IsIncluded` be a more descriptive name? `Supported` makes me think about the OS dependencies, like COM is not available, while trimming is about the application author design decision

src/System.Windows.Forms/src/System/Windows/Forms/Controls/ToolStrips/OpacityConverter.cs
```diff
@@ -45,7 +45,15 @@ public override bool CanConvertFrom(ITypeDescriptorContext? context, Type source
...
             {
-                percent = (double)TypeDescriptor.GetConverter(typeof(double)).ConvertFrom(context, culture, text)!;
+                if (!Control.UseComponentModelRegisterTypes)
+                {
+                    percent = (double)TypeDescriptor.GetConverter(typeof(double)).ConvertFrom(context, culture, text)!;
+                }
+                else
+                {
+                    // Call the trim safe API
+                    percent = (double)TypeDescriptor.GetConverterFromRegisteredType(typeof(double)).ConvertFrom(context, culture, text)!;
```
Tanya-Solyanik, line 55: Should we register `double` here?

src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.ComposedDataObject.NativeDataObjectToWinFormsAdapter.cs
```diff
@@ -18,7 +18,7 @@ internal unsafe partial class ComposedDataObject
     {
         [FeatureSwitchDefinition("System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization")]
 #pragma warning disable IDE0075 // Simplify conditional expression - the simpler expression is hard to read
-        private static bool EnableUnsafeBinaryFormatterInNativeObjectSerialization { get; } = AppContext.TryGetSwitch("System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization", out bool isEnabled) ? isEnabled : true;
+        internal static bool EnableUnsafeBinaryFormatterInNativeObjectSerialization { get; } = AppContext.TryGetSwitch("System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization", out bool isEnabled) ? isEnabled : true;
```
Tanya-Solyanik, line 21: Maybe we  need a new helper class that would store "System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization" switch value.

src/System.Windows.Forms/src/System/Windows/Forms/Panels/TableLayoutPanel/TableLayoutSettings.cs
```diff
@@ -485,6 +496,11 @@ internal List<ControlInformation> GetControlsInformation()
         {
             List<ControlInformation> controlsInfo = new(Owner!.Children.Count);
 
+            if (Control.UseComponentModelRegisterTypes)
+            {
+                TypeDescriptor.RegisterType<Control>();
```
Tanya-Solyanik, line 501: This pulls in `object` type, does not it? How is that handled? If object-types property is set by the developer to something custom, how would the developer know that they should Register that custom type?

src/System.Windows.Forms/src/System/Windows/Forms/Panels/TableLayoutPanel/TableLayoutSettings.cs
```diff
@@ -493,7 +509,17 @@ internal List<ControlInformation> GetControlsInformation()
...
                     // since it is shadowed.
-                    PropertyDescriptor? prop = TypeDescriptor.GetProperties(c)["Name"];
+                    PropertyDescriptor? prop;
+                    if (!Control.UseComponentModelRegisterTypes)
+                    {
+                        prop = TypeDescriptor.GetProperties(c)["Name"];
+                    }
+                    else
+                    {
+                        // Call the trim safe API
```
Tanya-Solyanik, line 519: And here again, this is the developer responsibility to register their custom controls, right? How does that error bubble up?

src/System.Windows.Forms/src/System/Windows/Forms/ActiveX/Control.ActiveXImpl.cs
```diff
@@ -1134,6 +1144,11 @@ bool SetValue(PropertyDescriptor currentProperty, object data)
 #pragma warning disable SYSLIB0011 // Type or member is obsolete
                     if (!success)
                     {
+                        if (!DataObject.ComposedDataObject.EnableUnsafeBinaryFormatterInNativeObjectSerialization)
+                        {
+                            throw new NotSupportedException(SR.BinaryFormatterNotSupported);
+                        }
```
MichalStrehovsky, line 1150: It feels weird to have a feature switch to control use of binary formatter which is also under a feature switch.

We annotated BinaryFormatter as RequiresUnreferencedCode, but at the same time we introduced a feature switch to disable it and defaulted it to "disabled" when trimming. Should we actually drop the RUC on BinaryFormatter? The rationale is that there is no behavioral difference between trimmed and untrimmed app (both throw) and we'll probably generate trimming warnings from BinaryFormatter internals should someone dare to enable the BinaryFormatter feature switch.

It doesn't feel right that people need to create feature switch for a feature switch.

(In a normal PublishTrimmed app, `new BinaryFormatter().Deserialize(stream)` below is going to throw a NotSupportedException anyway.)

Cc @eerhardt

eerhardt: > It feels weird to have a feature switch to control use of binary formatter which is also under a feature switch.

`DataObject.ComposedDataObject.EnableUnsafeBinaryFormatterInNativeObjectSerialization` is just a wrapper over the existing `System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization` feature switch. It isn't a new feature switch itself.

> Should we actually drop the RUC on BinaryFormatter?

Not until BinaryFormatter's logic is actually removed - https://github.com/dotnet/runtime/issues/98245. Once it always throws everywhere - then yes I think we can drop the annotations. No use in giving trim/AOT warnings when it will always throw.

src/System.Windows.Forms/tests/IntegrationTests/TrimTest/TrimTest.csproj
```diff
@@ -22,4 +22,50 @@
...
+    <RuntimeHostConfigurationOption Include="System.Windows.Forms.Binding.IsSupported"
+                                    Value="false"
+                                    Trim="true" />
+    <RuntimeHostConfigurationOption Include="System.ComponentModel.DefaultValueAttribute.IsSupported"
+                                    Value="false"
+                                    Trim="true" />
+    <RuntimeHostConfigurationOption Include="System.ComponentModel.TypeDescriptor.IsComObjectDescriptorSupported"
+                                    Value="false"
+                                    Trim="true" />
+    <RuntimeHostConfigurationOption Include="System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization"
```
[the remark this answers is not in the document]
LakshanF: Created issues in [runtime ](https://github.com/dotnet/runtime/issues/102619)and [SDK ](https://github.com/dotnet/sdk/issues/41122)to track this work.

src/System.Windows.Forms/tests/IntegrationTests/TrimTest/TrimTest.csproj
```diff
@@ -22,4 +22,50 @@
...
+                                    Value="false"
+                                    Trim="true" />
+    <RuntimeHostConfigurationOption Include="System.Drawing.Design.UITypeEditor.IsSupported"
+                                    Value="false"
+                                    Trim="true" />
+    <RuntimeHostConfigurationOption Include="System.Windows.Forms.ActiveXImpl.IsSupported"
```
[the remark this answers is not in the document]
LakshanF: It will be documented in the feature switch documentation and tracked in [WinForms ](https://github.com/dotnet/winforms/issues/10813)and [runtime](https://github.com/dotnet/runtime/issues/102619).

src/System.Windows.Forms/src/System/Drawing/Design/UITypeEditor.cs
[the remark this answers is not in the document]
LakshanF: Its good to use `IsSupported` if possible, to follow the pattern with feature switches (for example, [EventSource](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Tracing/EventSource.cs#L228))

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -1029,7 +1042,6 @@ public void ResetBindings()
     public virtual BindingContext? BindingContext
     {
         get => BindingContextInternal;
-        [RequiresUnreferencedCode(IBindableComponent.ComponentModelTrimIncompatibilityMessage)]
```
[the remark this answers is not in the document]
LakshanF: The setter wasn't the best place for the RUC and added the RUC to the `BindingContext` constructor. See #11180 for more details.

src/System.Windows.Forms/src/System/Windows/Forms/Controls/ToolStrips/OpacityConverter.cs
LakshanF: `ComponentModel` registers primitive types by default. @steveharter

src/System.Windows.Forms/src/System/Windows/Forms/Controls/ListControl/ListControl.cs
```diff
@@ -527,7 +537,15 @@ private protected int FindStringInternal(string? str, IList? items, int startInd
...
         // Try Formatter.FormatObject
-        s_stringTypeConverter ??= TypeDescriptor.GetConverter(typeof(string));
+        if (!UseComponentModelRegisterTypes)
+        {
+            s_stringTypeConverter ??= TypeDescriptor.GetConverter(typeof(string));
+        }
+        else
+        {
+            // Call the trim safe API
+            s_stringTypeConverter ??= TypeDescriptor.GetConverterFromRegisteredType(typeof(string));
```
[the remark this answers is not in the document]
LakshanF: Yes. @steveharter

src/System.Windows.Forms/src/System/Windows/Forms/Panels/TableLayoutPanel/TableLayoutSettings.cs
LakshanF: Only the relevant members of the Control would be kept since only Control is registered (and not any derived types). `ComponentModel` would throw in a predictable manner if a WinForms application that have custom control types are trimmed since those controls are not supported in trimming.

src/System.Windows.Forms/src/Resources/SR.resx
```diff
@@ -6970,4 +6970,19 @@ Stack trace where the illegal operation occurred was:
   <data name="BinaryFormatterNotSupported" xml:space="preserve">
     <value>Using the BinaryFormatter is not supported.</value>
   </data>
+  <data name="UITypeEditorNotSupported" xml:space="preserve">
+    <value>Using the UITypeEditor is not supported.</value>
```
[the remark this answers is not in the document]
Tanya-Solyanik: Do we have a work item to track changes for .targets file?

src/System.Windows.Forms/src/System/Windows/Forms/Panels/TableLayoutPanel/TableLayoutSettings.cs
LakshanF: Yes, developers will need to register custom types if they want to trim (this is an advanced scenario not planned to be supported in the initial trim support). Those applications will get a predictable runtime error in trimmed scenarios from the new `ComponentModel` API.

src/System.Windows.Forms.Primitives/src/System/ComponentModel/TypeConverterHelper.cs
```diff
@@ -7,6 +7,15 @@ namespace System.ComponentModel;
 
 internal static class TypeConverterHelper
 {
+    // Feature switch, when set to true, used for trimming to access ComponentModel in a trim safe manner
+    [FeatureSwitchDefinition("System.ComponentModel.TypeConverterHelper.UseComponentModelRegisteredTypes")]
+#pragma warning disable IDE0075 // Simplify conditional expression - the simpler expression is hard to read
+    private static bool UseComponentModelRegisteredTypes { get; } =
+        AppContext.TryGetSwitch("System.ComponentModel.TypeConverterHelper.UseComponentModelRegisteredTypes", out bool isEnabled)
```
eerhardt, line 14: I don't think this switch's name should be in the `System.ComponentModel` namespace. We should be using the namespace/type name in `System.Windows.Forms`. If someone saw this name with no context, they would think it would affect ALL System.ComponentModel things, but it is only for WinForms.

src/System.Windows.Forms/src/System/Windows/Forms/ActiveX/Control.ActiveXImpl.cs
LakshanF: The feature switch, `EnableUnsafeBinaryFormatterInNativeObjectSerialization`, using the binary formatter feature switch, is used to control trim warnings related to using `BinaryFormatter` in WinForms trimming. The trimmer will generate a warning when set to true and will not generate a warning when set to false to work around the issue of the RUC on the `BinaryFormatter` as mentioned above.

Tanya-Solyanik (dismissed):

src/System.Windows.Forms/src/System/Windows/Forms/Panels/TableLayoutPanel/TableLayoutSettings.cs
```diff
@@ -466,7 +466,18 @@ internal TableLayoutPanelCellPosition GetPositionFromControl(IArrangedElement? e
...
-        TypeConverter converter = TypeDescriptor.GetConverter(this);
+        TypeConverter converter;
+        if (!Control.UseComponentModelRegisteredTypes)
+        {
+            converter = TypeDescriptor.GetConverter(this);
+        }
+        else
+        {
+            // Call the trim safe API
+            TypeDescriptor.RegisterType<TableLayoutSettings>();
```
steveharter, line 477: Can RegisterType() be done at a higher level to avoid calling it more than once? Such as a constructor or static initializer. Applies elsewhere too.

If not, or to avoid registering when it is not necessary, perhaps we can add a `TypeDescriptor.GetConverterAndRegisterType<T>()` to avoid an extra hashtable lookup if max perf is necessary here. In either case, a minor point.

steveharter (approved):

steveharter (approved):

Tanya-Solyanik (approved):

PR #11419, opened 2024-05-23
Implement missing `RemoveToolTip` on `UpDownBase` to cleanup `ToolTip`

### LINKED ISSUE #11417: Can't remove tooltip once added to NumericUpDown
### .NET version

8.0.4

No

I don't think so.

### Issue description

    var f = new Form();
    var n = new NumericUpDown(); // can't realy clear tooltip once set
    // var n = new TextBox();    // we can clear tooltip
    var t = new ToolTip();
    var b = new Button { Left = 100, Top = 50, Text = "Clear"};
    b.Click += (_, _) => t.SetToolTip(n, null);
    
    f.Controls.Add(n);
    f.Controls.Add(b);

    t.SetToolTip(n, "haha ha");

    f.ShowDialog();

Once tooltip set on NumericUpDown, it can't be cleared. In this sample, After click the clear button, tooltip removed on textbox but not on numericupdown.

### Steps to reproduce

Create a winforms app, set tooltip to numericupdown, try to clear the tooltip, and you will see it doesn't work.

It was shocking to find out this issue exists  in .net framework, too.

elachlan:
Makes sure that the `_upDownEdit` and `_upDownButtons` have their tooltip removed.

Fixes #11417

I am unsure how to add tests for this. I manually tested the fix.

### DIFF
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/UpDownBaseTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/UpDownBaseTests.cs
@@ -3032,6 +3032,13 @@ public class UpDownBaseTests
 
         Assert.Equal(text, actualEditToolTipText);
         Assert.Equal(text, actualButtonsToolTipText);
+
+        toolTip.SetToolTip(upDownBase, null); // Invokes UpDownBase's SetToolTip inside
+        actualEditToolTipText = toolTip.GetToolTip(upDownBase._upDownEdit);
+        actualButtonsToolTipText = toolTip.GetToolTip(upDownBase._upDownButtons);
+
+        Assert.Equal("", actualEditToolTipText);
+        Assert.Equal("", actualButtonsToolTipText);
     }
 
     private class CustomValidateUpDownBase : UpDownBase
--- a/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/UpDownBaseTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/UpDownBaseTests.cs
@@ -3021,24 +3021,24 @@ public class UpDownBaseTests
         string actualEditToolTipText = toolTip.GetToolTip(upDownBase._upDownEdit);
         string actualButtonsToolTipText = toolTip.GetToolTip(upDownBase._upDownButtons);
 
-        Assert.Empty(actualEditToolTipText);
-        Assert.Empty(actualButtonsToolTipText);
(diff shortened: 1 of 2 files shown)


Olina-Zhang:
> @Olina-Zhang can your team please test this fix?

Tested this PR based on the latest .NET 9.0 SDK build for GH Issue: https://github.com/dotnet/winforms/issues/11417, it is fixed with no new issue found.

Fixed

elachlan:
Thank you @Olina-Zhang!!

lonitra:
Looks like we have a test that checks what we set the tooltip for UpDownBase if what is expected. 
https://github.com/dotnet/winforms/blob/f53f153808bbbaddda3f0c1e994e8a3b5cc085a9/src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/UpDownBaseTests.cs#L3015-L3035

We can use this as a base to write another test checking if setting to empty works as expected. Could you test if setting the tooltip multiple times works as expected as well?

elachlan:
Thanks I'll write one up and test asap.

elachlan:
@lonitra I added the code to the existing test, since we would have to have the whole test replicated again anyway to set the initial tooltip text.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/UpDownBaseTests.cs
```diff
@@ -3032,6 +3032,13 @@ public void UpDownBase_Invokes_SetToolTip_IfExternalToolTipIsSet()
 
         Assert.Equal(text, actualEditToolTipText);
         Assert.Equal(text, actualButtonsToolTipText);
+
+        toolTip.SetToolTip(upDownBase, null); // Invokes UpDownBase's SetToolTip inside
+        actualEditToolTipText = toolTip.GetToolTip(upDownBase._upDownEdit);
+        actualButtonsToolTipText = toolTip.GetToolTip(upDownBase._upDownButtons);
+
+        Assert.Equal("", actualEditToolTipText);
```
lonitra, line 3040:
```suggestion
        Assert.Equal(string.Empty, actualEditToolTipText);
```

Ditto below as well. Could we go ahead and change this test to use fluent assertions since we are already here? (no need to do the entire file, unless you would like to of course :) )

elachlan: I've pushed the change to fluent assertions for the test.

lonitra (approved):
🚀

Tanya-Solyanik (approved):

PR #11423, opened 2024-05-24
 8768 :: Resolve Addrange issue after sorting.

### LINKED ISSUE #8768: TreeNodeCollection.AddRange works incorrectly for non-empty collection of the sorted TreeView

Epica3055:
Fixes #8768 

## Proposed changes
if tree is already sorted then we don't want to set fixed Index value. due to these reason It is misbehave. 

## Customer Impact

- No

## Regression? 

- Yes

## Risk

- No

### Before

When the user adds a value to the treenode list and sorts it using treeview.sort(), it works fine. Subsequently, when the user adds individual trees through the add event, it functions as expected and sorts the entire tree automatically. However, if the user adds multiple values through AddRange, it does not work as expected.

### After
I've fixed the sorting issue we discussed earlier, and now everything is working smoothly. When users add values individually or use treeview.sort(), the sorting works fine. Even when they add multiple values at once using AddRange, the sorting behaves as expected.

## Test methodology 

- Testcase added

## Test environment(s)

- tested with .net core 8.0

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11423)

### DIFF
--- a/src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
@@ -24,6 +24,7 @@ internal static partial class LocalAppContextSwitches
     internal const string DataGridViewUIAStartRowCountAtZeroSwitchName = "System.Windows.Forms.DataGridViewUIAStartRowCountAtZero";
     internal const string NoClientNotificationsSwitchName = "Switch.System.Windows.Forms.AccessibleObject.NoClientNotifications";
     internal const string EnableMsoComponentManagerSwitchName = "Switch.System.Windows.Forms.EnableMsoComponentManager";
+    internal const string TreeNodePrevNodeDoNotUseFixedIndexWithSortedTreeViewSwitchName = "System.Windows.Forms.TreeNodeDoNotUseFixedIndexWithSortedTreeView";
 
     private static int s_scaleTopLevelFormMinMaxSizeForDpi;
     private static int s_anchorLayoutV2;
@@ -34,6 +35,7 @@ internal static partial class LocalAppContextSwitches
     private static int s_dataGridViewUIAStartRowCountAtZero;
     private static int s_noClientNotifications;
     private static int s_enableMsoComponentManager;
+    private static int s_treeNodePrevNodeDoNotUseFixedIndexWithSortedTreeView;
 
     private static FrameworkName? s_targetFrameworkName;
 
@@ -200,6 +202,16 @@ internal static partial class LocalAppContextSwitches
     {
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         get => GetCachedSwitchValue(EnableMsoComponentManagerSwitchName, ref s_enableMsoComponentManager);
+    }
+
+    /// <summary>
+    ///  Indicates whether TreeNode.PrevNode uses a fixed index to return its value.
+    ///  Thus mirroring the behavior of the TreeNodeCollection.
+    /// </summary>
+    public static bool TreeNodePrevNodeDoNotUseFixedIndexWithSortedTreeView
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        get => GetCachedSwitchValue(TreeNodePrevNodeDoNotUseFixedIndexWithSortedTreeViewSwitchName, ref s_treeNodePrevNodeDoNotUseFixedIndexWithSortedTreeView);
     }
 
     internal static void SetLocalAppContextSwitchValue(string switchName, bool value)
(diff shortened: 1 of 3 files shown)


Tanya-Solyanik (commented):
Hi @RutulPatel8 - I apologize that we had accidentally destroyed your branch when reviewing and trying to rebase it on the latest. This PR restores your changes with minor modifications, see the commit history, we applied the fix to the failing test, used fluent assertions in the new test and used [WinFormsTheory] instead of the [WinFormFact] on the test. You can continue working in this branch by pulling it down using GH Client `gh pr checkout 11423`, GitHub Desktop, or VS-Code. Or you can cherry-pick your commit https://github.com/dotnet/winforms/pull/11423/commits/c6dd01833519da19ea7a7ca41daed45322d76966 and create a new PR. 

Because this change could be breaking for someone who is not sorting nodes intentionally, it requires an opt-out switch. If you are not planning to pick it up, a team member can add this switch and finish this work. Please let us know if you have the bandwidth to wrap it up in 2 weeks. This will give us time to get this change into the next preview.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TreeNodeCollectionTests.cs
```diff
@@ -214,6 +214,39 @@ public void TreeNodeCollection_Find_InvokeNoSuchKey_ReturnsEmpty(string key)
         Assert.Empty(collection.Find(key, searchAllChildren: false));
     }
 
+    [WinFormsTheory]
+    [InlineData("7")]
```
ricardobossan, line 218: InlineData with just one key seems redundant. Convert to WinformsFact or maybe add add more test cases.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -24,6 +24,7 @@ internal static partial class LocalAppContextSwitches
     internal const string DataGridViewUIAStartRowCountAtZeroSwitchName = "System.Windows.Forms.DataGridViewUIAStartRowCountAtZero";
     internal const string NoClientNotificationsSwitchName = "Switch.System.Windows.Forms.AccessibleObject.NoClientNotifications";
     internal const string EnableMsoComponentManagerSwitchName = "Switch.System.Windows.Forms.EnableMsoComponentManager";
+    internal const string TreeNodePrevNodeDoNotUseFixedIndexWithSortedTreeViewSwitchName = "System.Windows.Forms.TreeNodeDoNotUseFixedIndexWithSortedTreeView";
```
Tanya-Solyanik, line 27: Please use `TreeNodeCollectionAddRangeRespectsSortOrder ` as a switch name, and make it an opt-out switch, i.e. default value should be true.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/TreeNodeCollectionTests.cs
```diff
@@ -214,6 +214,52 @@ public void TreeNodeCollection_Find_InvokeNoSuchKey_ReturnsEmpty(string key)
         Assert.Empty(collection.Find(key, searchAllChildren: false));
     }
 
+    [WinFormsFact]
+    public void TreeNodeCollection_Sort_ShouldAfterAddingItems()
+    {
+        using TreeView treeView = new();
```
Tanya-Solyanik, line 220: We also need tests that exercise the new switch, we should run this code with the switch on and off.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -202,6 +209,16 @@ public static bool EnableMsoComponentManager
         get => GetCachedSwitchValue(EnableMsoComponentManagerSwitchName, ref s_enableMsoComponentManager);
     }
 
+    /// <summary>
+    ///  Indicates whether TreeNode.PrevNode uses a fixed index to return its value.
```
Tanya-Solyanik, line 213: Suggestion:

When set to <see langword="true" /> (default), <see cref="Treeview.AddRange" /> API will insert nodes in the sorted order. To get behavior compatible with the previous versions of .NET and .NET Frameworek, set this switch to <see langword="false" />.

Tanya-Solyanik (commented):
Added a minor comment, otherwise looks good

Tanya-Solyanik:
@RutulPatel8  - thank you! We are done with the new appcontext switch and tests. This fix  will ship in the next preview, assuming everything goes well, please test it on your side. But we have many other issues we are [looking for help with](https://github.com/dotnet/winforms/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22).

Tanya-Solyanik (approved):

PR #11513, opened 2024-06-08
Enable nullability in remaining ControlDesigner members

gpetrou:
## Proposed changes

- Enable nullability in remaining ControlDesigner members.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11513)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.ChildSubClass.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.ChildSubClass.cs
@@ -58,7 +58,7 @@ public partial class ControlDesigner
             }
             catch (Exception ex)
             {
-                _designer.SetUnhandledException(Control.FromChildHandle(m.HWnd), ex);
+                _designer.SetUnhandledException(Control.FromChildHandle(m.HWnd)!, ex);
             }
             finally
             {
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.ChildSubClass.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.ChildSubClass.cs
@@ -58,7 +58,7 @@ public partial class ControlDesigner
             }
             catch (Exception ex)
             {
-                _designer.SetUnhandledException(Control.FromChildHandle(m.HWnd)!, ex);
+                _designer.SetUnhandledException(Control.FromChildHandle(m.HWnd), ex);
             }
             finally
             {
--- a/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.cs
+++ b/src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.cs
@@ -2397,6 +2397,8 @@ public partial class ControlDesigner : ComponentDesigner
     {
         Point pt = PARAM.ToPoint(lParam);
         pt = Control.PointToScreen(pt);
+
+        // We have already checked if Parent is null before calling the method.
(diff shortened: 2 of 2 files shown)


src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.cs
```diff
@@ -2312,7 +2312,7 @@ private void PaintException(PaintEventArgs e, Exception ex)
         stringFormat.Dispose();
     }
 
-    private IOverlayService OverlayService => _overlayService ??= GetService<IOverlayService>();
+    private IOverlayService? OverlayService => _overlayService ??= GetService<IOverlayService>();
```
ricardobossan, line 2315: Use TryGet pattern, to ensure non-null return values, like this:

```cs
   private IOverlayService? OverlayService
   {
       get
       {
           if (_overlayService is null)
           {
               if (TryGetService<IOverlayService>(out var service))
               {
                   _overlayService = service;
               }
           }

           return _overlayService;
       }
   }
```

gpetrou: Not sure I understand the suggestion. This is still nullable.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.cs
```diff
@@ -2490,8 +2490,8 @@ internal void SetUnhandledException(Control owner, Exception exception)
         _thrownException = exception;
         owner ??= Control;
 
-        string[] exceptionLines = exception.StackTrace.Split('\r', '\n');
-        string typeName = owner.GetType().FullName;
+        string[] exceptionLines = exception.StackTrace!.Split('\r', '\n');
+        string typeName = owner.GetType().FullName!;
```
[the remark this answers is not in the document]
gpetrou: I doubt that these will ever be null, but I added a couple of null checks anyway.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.cs
ricardobossan: We recommend the TryGet pattern, like TryGetService, to ensure non-null return values and maintain consistency. While the value can still be nullable, this approach standardizes our code. It’s encouraged for consistency but not mandatory.

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.ChildSubClass.cs
```diff
@@ -58,7 +58,7 @@ protected override void WndProc(ref Message m)
             }
             catch (Exception ex)
             {
-                _designer.SetUnhandledException(Control.FromChildHandle(m.HWnd), ex);
+                _designer.SetUnhandledException(Control.FromChildHandle(m.HWnd)!, ex);
```
lonitra, line 61: Looking at `SetUnhandledException` seems like it is ok for first param to be null. We should adjust the signature in that method and avoid the `!` here

src/System.Windows.Forms.Design/src/System/Windows/Forms/Design/ControlDesigner.cs
lonitra: We typically use `TryGetService<T>` if we are going to do different logic based off of whether we can get a non-null service or not. In this case, what we return is still nullable so leaving this as `GetService<T>` is acceptable.

lonitra (approved):
Thank you!

PR #11526, opened 2024-06-12
Added an opt-in switch to use BinaryFormater.Deserialize in Clipboard get methods.

Tanya-Solyanik:
The switch is not fully used yet. It requires that app opts in into binary formatter use globally.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11526)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/OLE/ClipboardHelper.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/OLE/ClipboardHelper.cs
@@ -1,0 +1,32 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Reflection.Metadata;
+using Switches = System.Windows.Forms.Primitives.LocalAppContextSwitches;
+
+namespace System.Windows.Forms;
+
+internal static class ClipboardHelper
+{
+    /// <devdoc>
+    ///  This resolver is used in deserialization of Clipboard and when wrapped into a serialization binder, suppresses all dangerous type.
+    /// </devdoc>
+    /// <exception cref="NotSupportedException"></exception>
+    internal static Func<TypeName, Type> SafeResolver { get; } = (typeName) =>
+        throw new NotSupportedException("Using BinaryFormatter is not supported in WinForms Clipboard data deserialization.");
+
+    /// <devdoc>
+    ///  This resolver is used in deserialization of Clipboard content and and when wrapped into a serialization binder, passes all types through.
+    ///  It is used in compatibility scenarios when user explicitly opts into its use.
+    /// </devdoc>
+    /// <exception cref="NotSupportedException"></exception>
+    internal static Func<TypeName, Type>? UnsafeResolver { get; } = (typeName) =>
+    {
+        if (!Switches.ClipboardEnableUnsafeBinaryFormatterDeserialization)
+        {
+            throw new NotSupportedException("Using BinaryFormatter is not supported in WinForms Clipboard data deserialization.");
+        }
+
+        return null!;
+    };
+}
--- a/src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/runtimeconfig.template.json
+++ b/src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/runtimeconfig.template.json
@@ -5,10 +5,12 @@
     "System.Windows.Forms.ApplyParentFontToMenus": true,
     "System.Windows.Forms.DataGridViewUIAStartRowCountAtZero": false,
     "System.Windows.Forms.DoNotCatchUnhandledExceptions": true,
-    "System.Windows.Forms.EnableMsoComponentManager": false,
-    "System.Windows.Forms.NoClientNotifications": false,
+    "Switch.System.Windows.Forms.EnableMsoComponentManager": false,
+    "Switch.System.Windows.Forms.NoClientNotifications": false,
     "System.Windows.Forms.ScaleTopLevelFormMinMaxSizeForDpi": false,
     "System.Windows.Forms.ServicePointManagerCheckCrl": false,
-    "System.Windows.Forms.TrackBarModernRendering": false
+    "System.Windows.Forms.TrackBarModernRendering": false,
+    "System.Windows.Forms.TreeNodeCollectionAddRangeRespectsSortOrder": true,
+    "System.Windows.Forms.Clipboard.EnableUnsafeBinaryFormatterDeserialization": false
   }
 }
--- a/src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
+++ b/src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
@@ -9,8 +9,7 @@ namespace System.Windows.Forms.Primitives;
 // Borrowed from https://github.com/dotnet/runtime/blob/main/src/libraries/Common/src/System/LocalAppContextSwitches.Common.cs
 internal static partial class LocalAppContextSwitches
 {
-    // Enabling switches in Core is different from Framework. See https://learn.microsoft.com/dotnet/core/runtime-config/
-    // for details on how to set switches.
+    // See https://learn.microsoft.com/dotnet/core/runtime-config/  for details on how to set switches.
 
     // Switch names declared internal below are used in unit/integration tests. Refer to
     // https://github.com/dotnet/winforms/blob/tree/main/docs/design/anchor-layout-changes-in-net80.md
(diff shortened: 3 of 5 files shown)


src/System.Windows.Forms/tests/IntegrationTests/WinformsControlsTest/runtimeconfig.template.json
```diff
@@ -5,10 +5,12 @@
     "System.Windows.Forms.ApplyParentFontToMenus": true,
     "System.Windows.Forms.DataGridViewUIAStartRowCountAtZero": false,
     "System.Windows.Forms.DoNotCatchUnhandledExceptions": true,
-    "System.Windows.Forms.EnableMsoComponentManager": false,
-    "System.Windows.Forms.NoClientNotifications": false,
+    "Switch.System.Windows.Forms.EnableMsoComponentManager": false,
+    "Switch.System.Windows.Forms.NoClientNotifications": false,
```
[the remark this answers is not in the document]
Tanya-Solyanik: No, they are already in use, at least the second one is in use in the designer repo.

src/System.Windows.Forms/src/System/Windows/Forms/OLE/SafeRestrictiveBinder.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Runtime.Serialization;
+
+namespace System.Windows.Forms;
+
+internal class SafeRestrictiveBinder : SerializationBinder
```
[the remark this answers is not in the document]
Tanya-Solyanik: Switch allows use of BF.Deserialize with the unsafe binder. It can be used when porting an app with the "old" APIs to NET9 and making it BF-safe with a single switch with no recompile. 
Clipboard.GetData APIs are always using the unsafe binder behind the scenes and can be controlled by this appcompat switch.
Clipboard.TryGetData APIs that don't take a custom binder parameter, always use the safe binder. They are used for a new app that wants to be safe. Theoretically this should be equivalent to skipping BF.Deserialize, but I don't know how to prove that. This binder also makes the implementation simpler as we always have a binder parameter. Also, because the switch does not control this binder, the same application can use both the old and the new APIs. This sounds inconsistent, but I'm thinking of 3rd party dlls that are not under the app author control and should be backward compatible. 
Clipboard.TryGet API that takes a custom binder is controlled only by the "global" app compat switch. This is for a new app, or for a ported app that is being recompiled and wants to deserialize a hashtable with a custom comparer (I had not double-checked that yet).
So far comments in the code are for tracking what I want to do next, i'll edit them when we have a better picture.
What you are saying is a clearer model, but we might need more flexibility.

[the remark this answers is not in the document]
lonitra: I see, thanks laying out some of the details here! I can see that always having a binder does make the implementation simpler and I think with either approach we are ultimately achieving what we want here, but I agree it'll be clearer when we get a better picture and other changes are out.

lonitra (approved):

src/System.Windows.Forms/src/System/Windows/Forms/OLE/SafeRestrictiveBinder.cs
JeremyKuhne, line 8: Seal the class

JeremyKuhne (changes requested):
We'll want to update this to a static helper class that has static `Func<TypeName, Type>` methods.

We'll need an internal `SerializationBinder` that takes said `Func<TypeName, Type>` and creates the `TypeName` based on the string arguments.

Tanya-Solyanik:
>We'll want to update this to a static helper class that has static Func<TypeName, Type> methods.

Yeah, I updated this branch and then push was incomplete and I didn't notice it. This is why I pinged the thread for a review.

src/System.Windows.Forms/src/System/Windows/Forms/OLE/UnsafeCompatibleBinder.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Runtime.Serialization;
+using System.Windows.Forms.Primitives;
+
+namespace System.Windows.Forms;
+
+internal class UnsafeCompatibleBinder : SerializationBinder
```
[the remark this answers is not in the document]
Tanya-Solyanik: removed this class

lonitra (approved):
👍

src/System.Windows.Forms/src/System/Windows/Forms/OLE/SafeRestrictiveBinder.cs
Tanya-Solyanik: not applicable

src/System.Windows.Forms/src/System/Windows/Forms/OLE/UnsafeCompatibleBinder.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    // TanyaSo: We might unseal it to allow for composition of binders.
+    private UnsafeCompatibleBinder()
+    {
+    }
+
+    internal static readonly UnsafeCompatibleBinder s_instance = new ();
```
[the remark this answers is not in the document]
Tanya-Solyanik: removed this code

src/System.Windows.Forms/src/System/Windows/Forms/OLE/ClipboardHelper.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace System.Windows.Forms;
+
+internal static class ClipboardHelper
+{
+    /// <devdoc>
+    ///  This resolver is used in deserialization of Clipboard and when wrapped into a serialization binder, suppresses all dangerous type.
+    /// </devdoc>
+    /// <exception cref="NotSupportedException"></exception>
+    internal static Func<TypeName, Type> SafeResolver { get; } = (typeName) =>
```
JeremyKuhne, line 15: I believe the best way to do this is simply have static methods:

``` C#
internal static Type NotSupportedResolver(TypeName typeName) => throw new ...
```

Don't say that it is safe, just say what it does.

src/System.Windows.Forms/src/System/Windows/Forms/OLE/ClipboardHelper.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Reflection.Metadata;
+using Switches = System.Windows.Forms.Primitives.LocalAppContextSwitches;
+
+namespace System.Windows.Forms;
+
+internal static class ClipboardHelper
```
JeremyKuhne, line 9: There is no need for this to get its own class, just put it in `Clipboard`.

src/System.Windows.Forms/src/System/Windows/Forms/OLE/ClipboardHelper.cs
```diff
@@ -0,0 +1,32 @@
...
+        throw new NotSupportedException("Using BinaryFormatter is not supported in WinForms Clipboard data deserialization.");
+
+    /// <devdoc>
+    ///  This resolver is used in deserialization of Clipboard content and and when wrapped into a serialization binder, passes all types through.
+    ///  It is used in compatibility scenarios when user explicitly opts into its use.
+    /// </devdoc>
+    /// <exception cref="NotSupportedException"></exception>
+    internal static Func<TypeName, Type>? UnsafeResolver { get; } = (typeName) =>
+    {
+        if (!Switches.ClipboardEnableUnsafeBinaryFormatterDeserialization)
```
JeremyKuhne, line 25: It is unclear that this would be the best place for the switch check.

src/System.Windows.Forms/src/System/Windows/Forms/OLE/ClipboardHelper.cs
```diff
@@ -0,0 +1,32 @@
...
+using Switches = System.Windows.Forms.Primitives.LocalAppContextSwitches;
+
+namespace System.Windows.Forms;
+
+internal static class ClipboardHelper
+{
+    /// <devdoc>
+    ///  This resolver is used in deserialization of Clipboard and when wrapped into a serialization binder, suppresses all dangerous type.
+    /// </devdoc>
+    /// <exception cref="NotSupportedException"></exception>
```
JeremyKuhne, line 14: If you don't think a comment is needed for this XML block, then close it `/>`

JeremyKuhne (changes requested):
Added some feedback. It is a little odd to approve this without the consuming code. I presume you'll add additional commits with the rest of the implementation.

Issue #11368, opened 2024-05-14
Investigate JSON Serialization/Deserialization in OLE


lonitra:
A known way to serialize/deserialize data is by using System.Text.Json. We should investigate how we might be able to use JSON to serialize/deserialize clipboard/drag data. JSON serialization would occur in [WinFormsDataObjectToNativeAdapter.SaveObjectToHGLOBAL ](https://github.com/dotnet/winforms/blob/main/src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.ComposedDataObject.WinFormsDataObjectToNativeAdapter.cs#L322) and ensure enough information is provided to be able to deserialize on the other side in [NativeDataObjectToWinFormsAdapter.ReadObjectFromHGLOBAL](https://github.com/dotnet/winforms/blob/main/src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.ComposedDataObject.NativeDataObjectToWinFormsAdapter.cs#L176). As part of this investigation we should be mindful about how this would work in harmony with the other work that is being done in the area and ensuring serialization/deserialization across different process still works as it does today

Useful links:
https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/how-to
https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/deserialization

PR #11527, opened 2024-06-13
Add JSON Serialization Support for OLE

lonitra:
Related: https://github.com/dotnet/winforms/issues/11368

This is a start to support JSON serialization in OLE scenarios.
`JsonData`: This is our wrapper class that holds the JSON serialized data along with the original type information. Note that the data that is being JSON serialized uses [default JsonSerializer behavior](https://learn.microsoft.com/dotnet/standard/serialization/system-text-json/how-to#serialization-behavior).  Users can follow the remarks of the public APIs if custom JSON serialization behavior is needed.

Public APIs:
- `DataObject.SetDataAsJson()` - Allows users to add data to `DataObject`  that will be JSON serialized. This is how we support JSON serialization for drag/drop scenario.
- `Clipboard.SetDataAsJson()` - Allows users to conveniently JSON serialize a type onto the clipboard, but note that this does not accept `DataObject` being passed in. If users want to place a `DataObject` on the clipboard, users should ensure data they want to be JSON serialized is already in the `DataObject` by utilizing `DataObject.SetDataAsJson` and then use `Clipboard.SetDataObject()`.
- `Control.DoDragDropAsJson()` - Allows users to conveniently JSON serialize a type for drag operation, but note that this does not accept `DataObject` being passed in. If users wants to start a drag operation with a `DataObject`, user should ensure data they want to be JSON serialized is already in the `DataObject` by utilizing `DataObject.SetDataAsJson` and then use `Control.DoDragDrop()`

Tests:
- Unit tests added for new APIs, some of the same tests also added in ComDisabledTests to ensure same behavior if built in com support is turned off.
- Manually tested cross process scenario -- drag/drop and copy/paste from 2 WinForms apps and confirmed able to retrieve the original data without exposing `JsonData<T>`. If using legacy GetData APIs, user should change to use TryGetData APIs, otherwise follow recommendations to manually deserialize the data.
- Manually tested copying data onto clipboard from app running on main branch and pasting into .NET 8 app for informational purposes. This scenario does not work as one may expect likely due to `JsonData` not existing in .NET 8. In order to properly deserialize, users will need to follow recommendations to manually deserialize the stream.
- Test behavior in trimmed app: `JsonSerializer` uses reflection to serialize by default, which will not be available when trimming is turned on. In the future, we will need to either use [source generator](https://learn.microsoft.com/dotnet/standard/serialization/system-text-json/reflection-vs-source-generation) APIs to support trimming or configure [JsonSerializerOptions.TypeInfoResolver
](https://learn.microsoft.com/dotnet/api/system.text.json.jsonserializeroptions.typeinforesolver) to support trim scenarios.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11527)

### DIFF
--- a/src/System.Windows.Forms/src/Resources/SR.resx
+++ b/src/System.Windows.Forms/src/Resources/SR.resx
@@ -7035,7 +7035,5 @@ Stack trace where the illegal operation occurred was:
   </data>
   <data name="ClipboardOrDragDrop_InvalidFormatTypeCombination" xml:space="preserve">
     <value>The specified type '{0}' is not compatible with the specified format '{1}'.</value>
-  <data name="InvalidTypeForSetDataAsJson" xml:space="preserve">
-    <value>DataObject cannot be JSON serialized meaningfully. Set the data by using {0} instead.</value>
   </data>
 </root>
--- a/src/System.Windows.Forms/src/Resources/SR.resx
+++ b/src/System.Windows.Forms/src/Resources/SR.resx
@@ -7035,5 +7035,7 @@ Stack trace where the illegal operation occurred was:
   </data>
   <data name="ClipboardOrDragDrop_InvalidFormatTypeCombination" xml:space="preserve">
     <value>The specified type '{0}' is not compatible with the specified format '{1}'.</value>
+  <data name="InvalidTypeForSetDataAsJson" xml:space="preserve">
+    <value>DataObject cannot be JSON serialized meaningfully. Set the data by using {0} instead.</value>
   </data>
 </root>
--- a/src/System.Windows.Forms/src/System/Windows/Forms/BinaryFormat/WinFormsBinaryFormatWriter.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/BinaryFormat/WinFormsBinaryFormatWriter.cs
@@ -57,10 +57,11 @@ internal static class WinFormsBinaryFormatWriter
         using BinaryFormatWriterScope writer = new(stream);
         new BinaryLibrary(libraryId: 2, IJsonData.CustomAssemblyName).Write(writer);
         new ClassWithMembersAndTypes(
-            new ClassInfo(1, jsonData.TypeFullName, [$"<{nameof(jsonData.JsonBytes)}>k__BackingField"]),
+            new ClassInfo(1, typeof(JsonData).FullName!, [$"<{nameof(jsonData.JsonBytes)}>k__BackingField", $"<{nameof(jsonData.OriginalAssemblyQualifiedTypeName)}>k__BackingField"]),
             libraryId: 2,
-            new MemberTypeInfo[] { new(BinaryType.PrimitiveArray, PrimitiveType.Byte) },
-            new MemberReference(idRef: 3)).Write(writer);
+            new MemberTypeInfo[] { new(BinaryType.PrimitiveArray, PrimitiveType.Byte), new(BinaryType.String, null) },
+            new MemberReference(idRef: 3),
+            new BinaryObjectString(objectId: 4, jsonData.OriginalAssemblyQualifiedTypeName)).Write(writer);
 
         new ArraySinglePrimitive<byte>(objectId: 3, jsonData.JsonBytes).Write(writer);
     }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/BinaryFormat/WinFormsBinaryFormatWriter.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/BinaryFormat/WinFormsBinaryFormatWriter.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Drawing;
+using System.Private.Windows;
 using System.Private.Windows.Core.BinaryFormat;
 using System.Private.Windows.Core.BinaryFormat.Serializer;
 
@@ -51,17 +52,17 @@ internal static class WinFormsBinaryFormatWriter
         new ArraySinglePrimitive<byte>(3, data).Write(writer);
     }
 
-    public static void WriteJsonData(Stream stream, JsonData jsonData)
+    public static void WriteJsonData(Stream stream, IJsonData jsonData)
     {
         using BinaryFormatWriterScope writer = new(stream);
-        new BinaryLibrary(2, s_currentWinFormsFullName).Write(writer);
+        new BinaryLibrary(libraryId: 2, "System.Private.Windows.VirtualJson").Write(writer);
         new ClassWithMembersAndTypes(
             new ClassInfo(1, jsonData.TypeFullName, [$"<{nameof(jsonData.JsonBytes)}>k__BackingField"]),
             libraryId: 2,
             new MemberTypeInfo[] { new(BinaryType.PrimitiveArray, PrimitiveType.Byte) },
-            new MemberReference(3)).Write(writer);
+            new MemberReference(idRef: 3)).Write(writer);
 
-        new ArraySinglePrimitive<byte>(3, jsonData.JsonBytes).Write(writer);
+        new ArraySinglePrimitive<byte>(objectId: 3, jsonData.JsonBytes).Write(writer);
     }
 
     /// <summary>
@@ -85,7 +86,7 @@ internal static class WinFormsBinaryFormatWriter
                 WriteBitmap(stream, bitmap);
                 return true;
             }
-            else if (value is JsonData jsonData)
+            else if (value is IJsonData jsonData)
             {
                 WriteJsonData(stream, jsonData);
                 return true;
--- a/src/System.Windows.Forms/src/System/Windows/Forms/BinaryFormat/WinFormsBinaryFormatWriter.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/BinaryFormat/WinFormsBinaryFormatWriter.cs
@@ -55,7 +55,7 @@ internal static class WinFormsBinaryFormatWriter
     public static void WriteJsonData(Stream stream, IJsonData jsonData)
     {
         using BinaryFormatWriterScope writer = new(stream);
-        new BinaryLibrary(libraryId: 2, "System.Private.Windows.VirtualJson").Write(writer);
+        new BinaryLibrary(libraryId: 2, IJsonData.CustomAssemblyName).Write(writer);
         new ClassWithMembersAndTypes(
             new ClassInfo(1, jsonData.TypeFullName, [$"<{nameof(jsonData.JsonBytes)}>k__BackingField"]),
             libraryId: 2,
--- a/src/System.Windows.Forms/src/System/Windows/Forms/BinaryFormat/WinFormsBinaryFormatWriter.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/BinaryFormat/WinFormsBinaryFormatWriter.cs
@@ -57,10 +57,11 @@ internal static class WinFormsBinaryFormatWriter
         using BinaryFormatWriterScope writer = new(stream);
         new BinaryLibrary(libraryId: 2, IJsonData.CustomAssemblyName).Write(writer);
         new ClassWithMembersAndTypes(
-            new ClassInfo(1, jsonData.TypeFullName, [$"<{nameof(jsonData.JsonBytes)}>k__BackingField"]),
+            new ClassInfo(1, $"{typeof(IJsonData).Namespace}.JsonData", [$"<{nameof(jsonData.JsonBytes)}>k__BackingField", $"<{nameof(jsonData.InnerTypeAssemblyQualifiedName)}>k__BackingField"]),
             libraryId: 2,
-            new MemberTypeInfo[] { new(BinaryType.PrimitiveArray, PrimitiveType.Byte) },
-            new MemberReference(idRef: 3)).Write(writer);
+            new MemberTypeInfo[] { new(BinaryType.PrimitiveArray, PrimitiveType.Byte), new(BinaryType.String, null) },
+            new MemberReference(idRef: 3),
+            new BinaryObjectString(objectId: 4, jsonData.InnerTypeAssemblyQualifiedName)).Write(writer);
 
         new ArraySinglePrimitive<byte>(objectId: 3, jsonData.JsonBytes).Write(writer);
     }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/BinaryFormat/WinFormsBinaryFormatWriter.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/BinaryFormat/WinFormsBinaryFormatWriter.cs
@@ -57,11 +57,10 @@ internal static class WinFormsBinaryFormatWriter
         using BinaryFormatWriterScope writer = new(stream);
         new BinaryLibrary(libraryId: 2, IJsonData.CustomAssemblyName).Write(writer);
         new ClassWithMembersAndTypes(
-            new ClassInfo(1, typeof(JsonData).FullName!, [$"<{nameof(jsonData.JsonBytes)}>k__BackingField", $"<{nameof(jsonData.OriginalAssemblyQualifiedTypeName)}>k__BackingField"]),
+            new ClassInfo(1, jsonData.TypeFullName, [$"<{nameof(jsonData.JsonBytes)}>k__BackingField"]),
             libraryId: 2,
-            new MemberTypeInfo[] { new(BinaryType.PrimitiveArray, PrimitiveType.Byte), new(BinaryType.String, null) },
-            new MemberReference(idRef: 3),
-            new BinaryObjectString(objectId: 4, jsonData.OriginalAssemblyQualifiedTypeName)).Write(writer);
+            new MemberTypeInfo[] { new(BinaryType.PrimitiveArray, PrimitiveType.Byte) },
+            new MemberReference(idRef: 3)).Write(writer);
 
         new ArraySinglePrimitive<byte>(objectId: 3, jsonData.JsonBytes).Write(writer);
     }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/BinaryFormat/WinFormsBinaryFormatWriter.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/BinaryFormat/WinFormsBinaryFormatWriter.cs
@@ -16,7 +16,7 @@ internal static class WinFormsBinaryFormatWriter
 
     private static readonly string s_currentWinFormsFullName = typeof(WinFormsBinaryFormatWriter).Assembly.FullName!;
 
-    public static unsafe void WriteBitmap(Stream stream, Bitmap bitmap)
+    public static void WriteBitmap(Stream stream, Bitmap bitmap)
     {
         using MemoryStream memoryStream = new();
         bitmap.Save(memoryStream);
@@ -51,6 +51,19 @@ internal static class WinFormsBinaryFormatWriter
         new ArraySinglePrimitive<byte>(3, data).Write(writer);
     }
 
+    public static void WriteJsonData(Stream stream, JsonData jsonData)
+    {
+        using BinaryFormatWriterScope writer = new(stream);
+        new BinaryLibrary(2, s_currentWinFormsFullName).Write(writer);
+        new ClassWithMembersAndTypes(
+            new ClassInfo(1, jsonData.TypeFullName, [$"<{nameof(jsonData.JsonBytes)}>k__BackingField"]),
+            libraryId: 2,
+            new MemberTypeInfo[] { new(BinaryType.PrimitiveArray, PrimitiveType.Byte) },
+            new MemberReference(3)).Write(writer);
+
+        new ArraySinglePrimitive<byte>(3, jsonData.JsonBytes).Write(writer);
+    }
+
     /// <summary>
     ///  Writes the given <paramref name="value"/> if supported.
     /// </summary>
@@ -70,6 +83,11 @@ internal static class WinFormsBinaryFormatWriter
             else if (value is Bitmap bitmap)
             {
                 WriteBitmap(stream, bitmap);
+                return true;
+            }
+            else if (value is JsonData jsonData)
+            {
+                WriteJsonData(stream, jsonData);
                 return true;
             }
 
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Nrbf/WinFormsSerializationRecordExtensions.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Nrbf/WinFormsSerializationRecordExtensions.cs
@@ -63,24 +63,26 @@ internal static class WinFormsSerializationRecordExtensions
     {
         @object = null;
 
-        if (record.TypeName.AssemblyName!.FullName != IJsonData.CustomAssemblyName)
+        if (record.TypeName.AssemblyName?.FullName != IJsonData.CustomAssemblyName)
         {
             // The data was not serialized as JSON.
             return false;
         }
 
         if (record is not ClassRecord types
-            || !types.HasMember("<JsonBytes>k__BackingField")
             || types.GetRawValue("<JsonBytes>k__BackingField") is not SZArrayRecord<byte> byteData
-            || !TypeName.TryParse(types.TypeName.FullName, out TypeName? result)
-            || Type.GetType(result.GetGenericArguments()[0].AssemblyQualifiedName) is not Type genericType)
+            || types.GetRawValue("<OriginalAssemblyQualifiedTypeName>k__BackingField") is not string typeData
+            || !TypeName.TryParse(typeData, out TypeName? result)
+            || Type.GetType(result.AssemblyQualifiedName) is not Type originalType)
         {
             // This is supposed to be JsonData, but somehow the binary formatted data is corrupt.
             throw new InvalidOperationException();
         }
 
+        // TODO: If the full name of the type user is asking for doesn't match the type that is saved, return false.
+
         // TODO: We should get the type from the Func<TypeName, Type> that will be passed down instead of using Type.GetType()
-        @object = JsonSerializer.Deserialize(byteData.GetArray(), genericType);
+        @object = JsonSerializer.Deserialize(byteData.GetArray(), originalType);
 
         return true;
     }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Nrbf/WinFormsSerializationRecordExtensions.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Nrbf/WinFormsSerializationRecordExtensions.cs
@@ -3,6 +3,9 @@
 
 using System.Drawing;
 using System.Formats.Nrbf;
+using System.Private.Windows;
+using System.Reflection.Metadata;
+using System.Text.Json;
 
 namespace System.Windows.Forms.Nrbf;
 
@@ -54,6 +57,35 @@ internal static class WinFormsSerializationRecordExtensions
     }
 
     /// <summary>
+    ///  Tries to deserialize this object if it was serialized as JSON.
+    /// </summary>
+    public static bool TryGetObjectFromJson(this SerializationRecord record, out object? @object)
+    {
+        @object = null;
+
+        if (record.TypeName.AssemblyName!.FullName != IJsonData.CustomAssemblyName)
+        {
+            // The data was not serialized as JSON.
+            return false;
+        }
+
+        if (record is not ClassRecord types
+            || !types.HasMember("<JsonBytes>k__BackingField")
+            || types.GetRawValue("<JsonBytes>k__BackingField") is not SZArrayRecord<byte> byteData
+            || !TypeName.TryParse(types.TypeName.FullName, out TypeName? result)
+            || Type.GetType(result.GetGenericArguments()[0].AssemblyQualifiedName) is not Type genericType)
+        {
+            // This is supposed to be JsonData, but somehow the binary formatted data is corrupt.
+            throw new InvalidOperationException();
+        }
+
+        // TODO: We should get the type from the Func<TypeName, Type> that will be passed down instead of using Type.GetType()
+        @object = JsonSerializer.Deserialize(byteData.GetArray(), genericType);
+
+        return true;
+    }
+
+    /// <summary>
     ///  Try to get a supported object. This supports common types used in WinForms that do not have type converters.
     /// </summary>
     public static bool TryGetResXObject(this SerializationRecord record, [NotNullWhen(true)] out object? value) =>
@@ -63,5 +95,6 @@ internal static class WinFormsSerializationRecordExtensions
 
     public static bool TryGetCommonObject(this SerializationRecord record, [NotNullWhen(true)] out object? value) =>
         record.TryGetResXObject(out value)
-        || record.TryGetDrawingPrimitivesObject(out value);
+        || record.TryGetDrawingPrimitivesObject(out value)
+        || record.TryGetObjectFromJson(out value);
 }
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Nrbf/WinFormsSerializationRecordExtensions.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Nrbf/WinFormsSerializationRecordExtensions.cs
@@ -4,7 +4,9 @@
 using System.Drawing;
 using System.Formats.Nrbf;
 using System.Private.Windows;
+using System.Private.Windows.Core.BinaryFormat;
 using System.Reflection.Metadata;
+using System.Runtime.Serialization;
 using System.Text.Json;
 
 namespace System.Windows.Forms.Nrbf;
@@ -59,7 +61,12 @@ internal static class WinFormsSerializationRecordExtensions
     /// <summary>
     ///  Tries to deserialize this object if it was serialized as JSON.
     /// </summary>
-    public static bool TryGetObjectFromJson(this SerializationRecord record, out object? @object)
+    /// <returns>
+    ///  <see langword="true"/> if the data was serialized as JSON and was successfully deserialized. Otherwise, <see langword="false"/>.
+    /// </returns>
+    /// <exception cref="SerializationException">If the data was supposed to be our <see cref="JsonData{T}"/>, but was serialized incorrectly./></exception>
+    /// <exception cref="NotSupportedException">If an exception occurred while JSON deserializing.</exception>
+    public static bool TryGetObjectFromJson<T>(this SerializationRecord record, ITypeResolver resolver, out object? @object)
     {
         @object = null;
 
@@ -71,21 +78,27 @@ internal static class WinFormsSerializationRecordExtensions
 
         if (record is not ClassRecord types
             || types.GetRawValue("<JsonBytes>k__BackingField") is not SZArrayRecord<byte> byteData
-            || !TypeName.TryParse(types.TypeName.FullName, out TypeName? result)
-            || Type.GetType(result.GetGenericArguments().Single().AssemblyQualifiedName) is not Type genericType)
+            || types.GetRawValue("<InnerTypeAssemblyQualifiedName>k__BackingField") is not string innerTypeFullName
+            || !TypeName.TryParse(innerTypeFullName, out TypeName? genericTypeName))
         {
             // This is supposed to be JsonData, but somehow the binary formatted data is corrupt.
-            throw new InvalidOperationException();
+            throw new SerializationException();
         }
 
-        // TODO: We should get the type from the Func<TypeName, Type> that will be passed down instead of using Type.GetType()
+        Type genericType = resolver.GetType(genericTypeName);
+        if (!genericType.IsAssignableTo(typeof(T)))
+        {
+            // Not the type the caller asked for.
+            return false;
+        }
+
         try
         {
-            @object = JsonSerializer.Deserialize(byteData.GetArray(), genericType);
+            @object = JsonSerializer.Deserialize<T>(byteData.GetArray());
         }
         catch (Exception ex)
         {
-            @object = new NotSupportedException(ex.Message);
+            throw new NotSupportedException(ex.Message);
         }
 
         return true;
@@ -101,6 +114,5 @@ internal static class WinFormsSerializationRecordExtensions
 
     public static bool TryGetCommonObject(this SerializationRecord record, [NotNullWhen(true)] out object? value) =>
         record.TryGetResXObject(out value)
-        || record.TryGetDrawingPrimitivesObject(out value)
-        || record.TryGetObjectFromJson(out value);
+        || record.TryGetDrawingPrimitivesObject(out value);
 }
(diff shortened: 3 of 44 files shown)


src/System.Windows.Forms/src/Resources/SR.resx
```diff
@@ -6976,4 +6976,7 @@ Stack trace where the illegal operation occurred was:
   <data name="DesignTimeFeaturesNotSupported" xml:space="preserve">
     <value>Design time features are not supported.</value>
   </data>
+  <data name="InvalidTypeForSetDataAsJson" xml:space="preserve">
+    <value>DataObject cannot be JSON serialized meaningfully. Set the data by using {0} instead.</value>
```
Tanya-Solyanik, line 6980: It will be difficult to merge the xlf files.... I'd rather add all resources as the last step

src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.cs
```diff
@@ -91,9 +93,49 @@ internal IDataObject TryUnwrapInnerIDataObject()
...
+    ///   If custom JSON serialization behavior is needed, manually JSON serialize the data and then use <see cref="SetData(object?)"/>,
+    ///   or create a custom <see cref="Text.Json.Serialization.JsonConverter"/>, attach the
+    ///   <see cref="Text.Json.Serialization.JsonConverterAttribute"/>, and then recall this method.
+    ///   See <see href="https://learn.microsoft.com/dotnet/standard/serialization/system-text-json/converters-how-to"/> for more details
+    ///   on custom converters for JSON serialization.
+    ///  </para>
+    /// </remarks>
+    public void SetDataAsJson<T>(string format, T data)
+    {
+        if (data is DataObject)
```
Tanya-Solyanik, line 123: Should we exclude any IDataObject?

src/System.Windows.Forms/src/System/Windows/Forms/BinaryFormat/WinFormsBinaryFormatWriter.cs
```diff
@@ -49,6 +49,19 @@ public static void WriteImageListStreamer(Stream stream, ImageListStreamer strea
         new ArraySinglePrimitive<byte>(3, data).Write(writer);
     }
 
+    public static void WriteJsonData(Stream stream, JsonData jsonData)
+    {
+        using BinaryFormatWriterScope writer = new(stream);
+        new BinaryLibrary(2, s_currentWinFormsFullName).Write(writer);
```
Tanya-Solyanik, line 55: What is "2", resord number? Is there a way to get the "next available record number"?

src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.cs
```diff
@@ -91,9 +93,49 @@ internal IDataObject TryUnwrapInnerIDataObject()
...
+    }
+
     #region IDataObject
-    public virtual object? GetData(string format, bool autoConvert) =>
-        ((IDataObject)_innerData).GetData(format, autoConvert);
+    public virtual object? GetData(string format, bool autoConvert)
+    {
+        object? data = ((IDataObject)_innerData).GetData(format, autoConvert);
+#pragma warning disable SYSLIB0050 // Type or member is obsolete
+        return data is JsonData jsonData && jsonData is IObjectReference reference ? reference.GetRealObject(default) : data;
```
Tanya-Solyanik, line 136: Developers might need JsonData public in order to override this virtual

Tanya-Solyanik, line 136: Did you plan to put the conversion from JsonData to the real object into the com to winforms adapter?

src/System.Windows.Forms/src/System/Windows/Forms/OLE/JsonData.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace System.Windows.Forms;
+
+/// <summary>
+///  Wrapper which contains JSON serialized data along with the JSON data's original type information
+///  to be deserialized later.
+/// </summary>
+[Serializable]
+#pragma warning disable SYSLIB0050 // Type or member is obsolete
+internal struct JsonData<T> : IObjectReference, JsonData
```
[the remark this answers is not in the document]
lonitra: Yes the Deserializer will check if the object it deserializes is an `IObjectReference`

src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.cs
lonitra: 😮 Thanks! I hadn't thought about this.. Let me see if I can find a way to avoid making JsonData public. I'll test out your suggestion below.

src/System.Windows.Forms/src/System/Windows/Forms/BinaryFormat/WinFormsBinaryFormatWriter.cs
lonitra: 2 is the library ID -- I'll update to add the param names for these. This is the valid binary format for JsonData that I got. I'm not sure if having a way to get "next available" would be useful as this should not change.

src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.cs
lonitra: This is a scenario I am not sure about. If we get an `IDataObject`, we are getting a user type and I am not quite sure what the intent the user has here.. They might be able to meaningfully JSON serialize it by implementing their own JSON converter? I'm not sure if there is any valid scenario, but I had decided not to block it in case there was. I am open to excluding if we think that it should be blocked though.

Tanya-Solyanik: Makes sence

src/System.Windows.Forms/src/System/Windows/Forms/OLE/JsonData.cs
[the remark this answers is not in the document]
Tanya-Solyanik: It will check that the object is an IObjectReference and will unwrap it and return the real object, right? So we would receive JsonData from deserialization only if it went through the BinaryFormatter. But it should never come back to us from the BinaryFormatter.Deserialize, because BinaryFormattedObject takes precedence, and shouldn't fail for JsonData. Am I missing something?
If BFO had failed to Deserialize, we can read the root record, see that it is JsonData and if deserialize failed thow instead of going to the binaryfarmatter for the second attemps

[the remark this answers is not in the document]
lonitra: That's right, I don't think you are missing anything. If the binary serialized object really is a JsonData, BFO shouldn't fail to deserialize it. I can only imagine it would fail if the binary formatted data somehow got corrupted, though I'm not sure how that would be possible. I will add a throw for this case.

src/System.Windows.Forms/src/System/Windows/Forms/BinaryFormat/WinFormsBinaryFormatWriter.cs
```diff
@@ -49,6 +50,19 @@ public static void WriteImageListStreamer(Stream stream, ImageListStreamer strea
         new ArraySinglePrimitive<byte>(3, data).Write(writer);
     }
 
+    public static void WriteJsonData(Stream stream, IJsonData jsonData)
+    {
+        using BinaryFormatWriterScope writer = new(stream);
+        new BinaryLibrary(libraryId: 2, "System.Private.Windows.VirtualJson").Write(writer);
```
JeremyKuhne, line 56: I'd put this string in the JsonData type as a constant with a comment.

src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.cs
```diff
@@ -91,9 +94,50 @@ internal IDataObject TryUnwrapInnerIDataObject()
     /// <inheritdoc cref="ComposedDataObject.OriginalIDataObject"/>
     internal IDataObject? OriginalIDataObject => _innerData.OriginalIDataObject;
 
+    /// <summary>
+    ///  Stores the specified data and its associated format in this instance as JSON.
+    /// </summary>
+    /// <exception cref="InvalidOperationException">
+    ///  If <see cref="DataObject"/> is passed in as the data. <see cref="DataObject"/> cannot be JSON serialized meaningfully.
```
JeremyKuhne, line 101: Nit: the normal verbiage is past tense "something was", or in this case "<paramref> was <see cref="DataObject">. Details..."

src/System.Windows.Forms/src/System/Windows/Forms/OLE/JsonData.cs
```diff
@@ -0,0 +1,35 @@
...
+[Serializable]
+#pragma warning disable SYSLIB0050 // Type or member is obsolete
+internal struct JsonData<T> : IObjectReference, IJsonData
+#pragma warning restore SYSLIB0050
+{
+    public byte[] JsonBytes { get; set; }
+
+    public readonly string TypeFullName => $"{typeof(JsonData<T>).FullName}";
+
+    public readonly object GetRealObject(StreamingContext context) =>
```
JeremyKuhne, line 22: While you don't technically need `IObjectReference` for our usages, the workaround for `Clipboard` versions that don't know of this type would be to copy this with `IObjectReference`. You can comment to that effect. Writing a test to capture this behavior would be good too.

The test should use the writer code to manually write a `JsonData` object to a stream, then you can read it with the deserializer and a binder that redirects to this actual type.

JeremyKuhne (commented):
Generally looks good, some minor comments.

src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.cs
lonitra: Looking at this again, users should really call the base method even if they override as there is not a way for them to get the data that is stored otherwise. We do not expose anything in DataObject that would allow users to be able to meaningfully override `GetData()` without calling the base method.

lonitra: Hm, I don't think this is needed. 🤔 users would have to go through this method to retrieve the data.

elachlan:
@willibrandon you might be interested in the clipboard work here.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataObjectTests.cs
```diff
@@ -2502,4 +2529,127 @@ public unsafe void DataObject_Native_GetFormats_ReturnsExpected()
...
+        string offsetFormat = "MM/dd/yyyy";
+        deserialized.Date.ToString(offsetFormat).Should().Be(forecast.Date.ToString(offsetFormat));
+        deserialized.TemperatureCelsius.Should().Be(forecast.TemperatureCelsius);
+        deserialized.Summary.Should().Be($"{forecast.Summary} custom!");
+    }
+
+    [JsonConverter(typeof(WeatherForecastJsonConverter))]
+    private class WeatherForecast
+    {
+        public DateTimeOffset Date { get; set; }
```
Tanya-Solyanik, line 2601: Theoretically this class does not need custom Json serialization. Maybe if these were fields instead of properties, it would be more useful?

src/System.Windows.Forms/src/System/Windows/Forms/Nrbf/WinFormsSerializationRecordExtensions.cs
```diff
@@ -53,6 +56,35 @@ public static bool TryGetBitmap(this SerializationRecord record, out object? bit
...
+        if (record.TypeName.AssemblyName?.FullName != IJsonData.CustomAssemblyName)
+        {
+            // The data was not serialized as JSON.
+            return false;
+        }
+
+        if (record is not ClassRecord types
+            || types.GetRawValue("<JsonBytes>k__BackingField") is not SZArrayRecord<byte> byteData
+            || !TypeName.TryParse(types.TypeName.FullName, out TypeName? result)
+            || result.GetGenericArguments().FirstOrDefault() is not { } genericTypeName
```
Tanya-Solyanik, line 75:
```suggestion
            || result.GetGenericArguments().Single() is not { } genericTypeName
```
THere should be exactly 1 generic parameter

lonitra: Oh thanks! I didn't know about this method 😄

willibrandon:
Thanks for working on these new APIs 🤤 I plan on testing your branch this evening, and will post any feedback that I have here.

willibrandon:
Looks good! I am able to drag and drop JSON cats now 😸 

```csharp
var data = new DataObject();
var cat = new NyanCat()
{
    Ascii = _nyanCatAscii
};

data.SetDataAsJson(nameof(_nyanCatAscii), cat);
toolStripItem.DoDragDrop(
    data: data,
    allowedEffects: DragDropEffects.All,
    dragImage: _nyanCatAscii301Bmp,
    cursorOffset: new Point(0, 111),
    useDefaultDragImage: false);
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ClipboardTests.cs
```diff
@@ -618,4 +621,129 @@ public unsafe void Clipboard_RawClipboard_SetClipboardData_ReturnsExpected()
...
+            stream = new MemoryStream(bytes, index, bytes.Length - index);
+        }
+        finally
+        {
+            PInvokeCore.GlobalUnlock(hglobal);
+        }
+
+        stream.Should().NotBeNull();
+        // Use NrbfDecoder to decode the stream and rehydrate the type.
+        SerializationRecord record = NrbfDecoder.Decode(stream, leaveOpen: true);
```
Tanya-Solyanik, line 726: Why do we want to keep this stream open, we not disposing it on our own?

lonitra: Good call. I had copied this from what we are doing internally in Clipboard and we leave the stream open there. Here, it can be closed and I need to dispose stream. I'll update.

src/System.Windows.Forms/src/System/Windows/Forms/Nrbf/WinFormsSerializationRecordExtensions.cs
```diff
@@ -53,6 +56,41 @@ public static bool TryGetBitmap(this SerializationRecord record, out object? bit
         return true;
     }
 
+    /// <summary>
+    ///  Tries to deserialize this object if it was serialized as JSON.
+    /// </summary>
+    public static bool TryGetObjectFromJson(this SerializationRecord record, out object? @object)
```
Tanya-Solyanik, line 62: To plug this into the stream reader code, this method should provide 3 pieces of information:
1 is this a JsonData struct?
2 does T match or IsAssignableFrom the record type?
3.if the above answers are "yes" then get the object out.
if something went wrong, we throw.

src/System.Windows.Forms/src/System/Windows/Forms/Nrbf/WinFormsSerializationRecordExtensions.cs
```diff
@@ -53,6 +56,41 @@ public static bool TryGetBitmap(this SerializationRecord record, out object? bit
...
+            return false;
+        }
+
+        if (record is not ClassRecord types
+            || types.GetRawValue("<JsonBytes>k__BackingField") is not SZArrayRecord<byte> byteData
+            || !TypeName.TryParse(types.TypeName.FullName, out TypeName? result)
+            || Type.GetType(result.GetGenericArguments().Single().AssemblyQualifiedName) is not Type genericType)
+        {
+            // This is supposed to be JsonData, but somehow the binary formatted data is corrupt.
+            throw new InvalidOperationException();
```
Tanya-Solyanik, line 78: Should this be a SerializationException? Payload had not been serialized correctly.

src/System.Windows.Forms/src/System/Windows/Forms/Nrbf/WinFormsSerializationRecordExtensions.cs
```diff
@@ -53,6 +56,41 @@ public static bool TryGetBitmap(this SerializationRecord record, out object? bit
...
+        }
+
+        // TODO: We should get the type from the Func<TypeName, Type> that will be passed down instead of using Type.GetType()
+        try
+        {
+            @object = JsonSerializer.Deserialize(byteData.GetArray(), genericType);
+        }
+        catch (Exception ex)
+        {
+            @object = new NotSupportedException(ex.Message);
```
Tanya-Solyanik, line 88: This is not necessarily assignable to T, the new APIs might not be able to return it to the user. Should we throw instead?

src/System.Windows.Forms/src/System/Windows/Forms/OLE/Clipboard.cs
```diff
@@ -292,6 +294,59 @@ public static void SetData(string format, object data)
...
+    ///  </para>
+    ///  <para>
+    ///   If custom JSON serialization behavior is needed, manually JSON serialize the data and then use <see cref="SetData"/>
+    ///   to save the data onto the clipboard, or create a custom <see cref="Text.Json.Serialization.JsonConverter"/>, attach the
+    ///   <see cref="Text.Json.Serialization.JsonConverterAttribute"/>, and then recall this method.
+    ///   See <see href="https://learn.microsoft.com/dotnet/standard/serialization/system-text-json/converters-how-to"/> for more details
+    ///   on custom converters for JSON serialization.
+    ///  </para>
+    /// </remarks>
+    [RequiresUnreferencedCode("Uses default System.Text.Json behavior which is not trim-compatible.")]
```
Tanya-Solyanik, line 327: Do we have the follow up work item to re-evaluate and provide guidance on how to make drag-drop cases trimmable? We can extend this one - https://github.com/dotnet/winforms/issues/12611 if not.

src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.cs
```diff
@@ -91,9 +93,81 @@ internal IDataObject TryUnwrapInnerIDataObject()
...
+        {
+            // TODO: Localize string.
+            throw new InvalidOperationException($"DataObject will serialize as empty. JSON serialize the data within  {nameof(data)}, then use {nameof(SetData)}  instead.");
+        }
+
+        SetData(typeof(T), new JsonData<T>() { JsonBytes = JsonSerializer.SerializeToUtf8Bytes(data) });
+    }
+
+    /// <summary>
+    ///  Stores the specified data and its associated format in this instance as JSON.
```
Tanya-Solyanik, line 124: format is not stored as JSON, we have no control over the format. Also Unicode will not be stored as JSON. How about:

Stores the specified data in this data object in the specified format. If the data is a managed object and format allows for serialization of managed objects, the object will be serialized as JSON text.

lonitra: Yes, bad wording here on my part. I meant that the data is serialized as JSON and is associated with the specified format. I had decided to only use the first portion of your suggestion since the data is always serialized as JSON no matter the format. Though I think it would be good to call out that if the data they are trying to pass in is a type that can be handled without binary serialization SetData should be called to reduce overhead. 🤔

src/System.Windows.Forms/src/System/Windows/Forms/OLE/JsonData.cs
```diff
@@ -0,0 +1,140 @@
...
+/// </remarks>
+/// <example>
+/// <![CDATA[
+///  // Recommended: deserialize using NrbfDecoder.
+///  void DeserializeJsonData(DataObject dataObject)
+///  {
+///     // Manually retrieve serialized stream.
+///     System.Runtime.InteropServices.ComTypes.FORMATETC formatetc = new()
+///     {
+///         cfFormat = (short) DataFormats.GetFormat("yourDataFormat").Id,
```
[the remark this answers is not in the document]
lonitra: I have a test that has this code for nrbf usage scenario: https://github.com/dotnet/winforms/pull/11527/files#diff-0102a9243799084fe22a0af3ec77d7c7d421b954b0a1d556561987dd354b0099R979
For BinaryFormatter scenario I couldn't add a real test because BinaryFormatter should only be used for this scenario if user does not want to go with NRBF route and JsonData doesn't exist, which it does here.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataObjectTests.cs
```diff
@@ -2800,4 +2830,173 @@ public unsafe void DataObject_Native_GetData_SerializationFailure()
...
+        deserializedString.Should().Be("test");
+    }
+
+    [WinFormsFact]
+    public void DataObject_SetDataAsJson_GetData_ReturnsNonDeserialized()
+    {
+        Point point = new() { X = 1, Y = 1 };
+        DataObject data = new();
+        data.SetDataAsJson("point", point);
+        data.GetData("point").Should().BeOfType<JsonData<Point>>();
```
lonitra, line 2893: Legacy GetData will expose our internal JsonData<T> in inproc scenarios, should we be more aggressive about not exposing this? On one hand this is an indicator that user should be using TryGetData API versus if we are aggressive and simply return null to avoid exposure, user can interpret other reasons why null is returned.

src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.cs
Tanya-Solyanik: First portion is a copy/paste from the docs :).

> is always serialized as JSON no matter the format.

Right.... I see that now. But is it deserialized as JSON only if it is not the OLE format.

src/System.Windows.Forms/src/System/Windows/Forms/OLE/Clipboard.cs
```diff
@@ -516,6 +518,60 @@ public static void SetData(string format, object data)
...
+        {
+            // TODO: Localize string
+            throw new InvalidOperationException($"'DataObject' will serialize as empty. JSON serialize the data within {nameof(data)}, then use {nameof(SetDataObject)} API instead.");
+        }
+
+        JsonData<T> jsonData = new()
+        {
+            JsonBytes = JsonSerializer.SerializeToUtf8Bytes(data),
+            InnerTypeAssemblyQualifiedName = typeof(T).ToTypeName().AssemblyQualifiedName
+    };
```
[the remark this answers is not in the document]
Tanya-Solyanik: This doesn't have to happen in the clipboard class, AssemblyQualifiedName can be generated by the constructor  JsonData<T> because this is an implementation detail, JsonData.InnerTypeAssemblyQualifiedName can be a read only property

src/System.Windows.Forms/src/System/Windows/Forms/Nrbf/WinFormsSerializationRecordExtensions.cs
```diff
@@ -53,6 +58,52 @@ public static bool TryGetBitmap(this SerializationRecord record, out object? bit
...
+            return false;
+        }
+
+        try
+        {
+            @object = JsonSerializer.Deserialize<T>(byteData.GetArray());
+        }
+        catch (Exception ex)
+        {
+            throw new NotSupportedException(ex.Message);
```
willibrandon, line 101: Consider passing through the original inner exception when JSON deserialization fails in order to preserve the debugging information. This way, the original stack trace and exception type will be maintained, making debugging easier.
```csharp
catch (Exception ex)
{
    throw new NotSupportedException("Failed to deserialize JSON data.", ex);
}
```

src/System.Windows.Forms/src/System/Windows/Forms/Nrbf/WinFormsSerializationRecordExtensions.cs
```diff
@@ -53,6 +58,52 @@ public static bool TryGetBitmap(this SerializationRecord record, out object? bit
...
+        if (record.TypeName.AssemblyName?.FullName != IJsonData.CustomAssemblyName)
+        {
+            // The data was not serialized as JSON.
+            return false;
+        }
+
+        if (record is not ClassRecord types
+            || types.GetRawValue("<JsonBytes>k__BackingField") is not SZArrayRecord<byte> byteData
+            || types.GetRawValue("<InnerTypeAssemblyQualifiedName>k__BackingField") is not string innerTypeFullName
+            || !TypeName.TryParse(innerTypeFullName, out TypeName? genericTypeName))
```
Tanya-Solyanik, line 82: `genericTypeName`  is not a descriptive name.

src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.cs
lonitra: Updated to avoid JSON serializing restricted formats to avoid overhead. Only time there can be overhead at this point is if custom format or `DataFormats.Serializable` is used with a type that is intrinsically handled by us. I placed in remarks it is recommended to use SetData API in this case and also updated summary to essentially take the suggestion now that this is done.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataObjectTests.cs
Tanya-Solyanik: GetData will generate a compile time warning - you don't see it in this class because the whole class disables it. So we already have an indication that this is not the best API to use..
Exactly how will the user see that type - it is a fake one...

src/System.Windows.Forms/tests/TestUtilities/DataObjectTestHelpers.cs
```diff
@@ -0,0 +1,52 @@
...
+         DataFormats.Rtf,
+         DataFormats.Html,
+         DataFormats.OemText,
+         DataFormats.FileDrop,
+         "FileName",
+         "FileNameW"
+    ];
+
+    public static TheoryData<string> UnboundedFormat() =>
+[
```
Tanya-Solyanik, line 26: Please fix indentation

src/System.Windows.Forms/tests/TestUtilities/DataObjectTestHelpers.cs
```diff
@@ -0,0 +1,52 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace System.Windows.Forms.TestUtilities;
+
+/// <summary>
+///  Test utilities relating to <see cref="DataObject"/>.
+/// </summary>
+public static class DataObjectTestHelpers
```
Tanya-Solyanik, line 9: This is cool, we could share theory data elsewhere later!

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataObjectTests.cs
```diff
@@ -2700,29 +2704,56 @@ public unsafe void DataObject_MockRoundTrip_OutData_IsSame(object data)
...
-            inData.Should().NotBeSameAs(data);
+            inData.Should().BeSameAs(data);
         }
 
         using var inDataPtr = ComHelpers.GetComScope<Com.IDataObject>(inData);
         IDataObject outData = dropTargetAccessor.CreateDelegate<CreateWinFormsDataObjectForOutgoingDropData>()(inDataPtr);
         outData.Should().BeSameAs(data);
     }
 
+    public static IEnumerable<object[]> DataObjectWithJsonMockRoundTripData()
```
Tanya-Solyanik, line 2722: Consider replacing with TheoryData time permitting

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/DataObjectTests.cs
```diff
@@ -2801,4 +2832,203 @@ public unsafe void DataObject_Native_GetData_SerializationFailure()
...
+    [WinFormsFact]
+    public void DataObject_SetDataAsJson_PredefinedTypes_NotJsonSerialized()
+    {
+        string format = "test";
+        using Bitmap bitmap = new(10, 10);
+        DataObject dataObject = new();
+
+        dataObject.SetDataAsJson(format, bitmap);
+        object result = dataObject.GetData(format);
+        result.Should().NotBeOfType<IJsonData>();
```
Tanya-Solyanik, line 3016: IJsonData is an internal type... we shouldn't return it

Tanya-Solyanik: Never mind, we are not returning it here, but on line 3032.

src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.cs
```diff
@@ -100,13 +102,148 @@ internal IDataObject TryUnwrapInnerIDataObject()
...
         Obsoletions.DataObjectGetDataMessage,
         error: false,
         DiagnosticId = Obsoletions.ClipboardGetDataDiagnosticId,
         UrlFormat = Obsoletions.SharedUrlFormat)]
-    public virtual object? GetData(string format, bool autoConvert) => _innerData.GetData(format, autoConvert);
+    public virtual object? GetData(string format, bool autoConvert)
+    {
+        object? result = _innerData.GetData(format, autoConvert);
+        // Avoid exposing our internal JsonData<T>
+        return result is IJsonData ? null : result;
```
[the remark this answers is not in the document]
Tanya-Solyanik: You are right, but still user applications that have no access to JsonData type will throw a first chance  SerializationException from BInaryFormatter. If this scenario is blocked before the binary formatter, then there is no exception.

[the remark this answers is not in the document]
lonitra: Wouldn't the exception be a good indicator for user if they want to deserialize this with BinaryFormatter, they need to follow recommended steps? Otherwise, user will just get null and that gives very little information to user what went wrong.

src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.cs
```diff
@@ -100,13 +102,148 @@ internal IDataObject TryUnwrapInnerIDataObject()
...
+
+    /// <summary>
+    ///  JSON serialize the data only if the format is not a restricted deserialization format and the data is not an intrinsic type.
+    /// </summary>
+    /// <returns>
+    ///  The passed in <paramref name="data"/> as is if the format is restricted. Otherwise the JSON serialized <paramref name="data"/>.
+    /// </returns>
+    private static object TryJsonSerialize<T>(string format, T data)
+    {
+        format.OrThrowIfNull(nameof(format));
```
Tanya-Solyanik, line 186: We have a specific message for this error.
```
        if (string.IsNullOrWhiteSpace(format.OrThrowIfNull()))
        {
            throw new ArgumentException(SR.DataObjectWhitespaceEmptyFormatNotAllowed, nameof(format));
        }
```

src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.cs
```diff
@@ -100,13 +102,148 @@ internal IDataObject TryUnwrapInnerIDataObject()
...
+        return IsRestrictedFormat(format) || Composition.Binder.IsKnownType<T>()
+            ? data
+            : new JsonData<T>() { JsonBytes = JsonSerializer.SerializeToUtf8Bytes(data) };
+    }
+
+    /// <summary>
+    ///  Check if the <paramref name="format"/> is one of the restricted formats, which formats that
+    ///  correspond to primitives or are pre-defined in the OS such as strings, bitmaps, and OLE types.
+    /// </summary>
+    internal static bool IsRestrictedFormat(string format) => RestrictDeserializationToSafeTypes(format)
```
Tanya-Solyanik, line 197:
```suggestion
    private static bool IsRestrictedFormat(string format) => RestrictDeserializationToSafeTypes(format)
```
This is used in the DataObject class and its inner classes only

src/System.Windows.Forms/tests/TestUtilities/DataObjectTestHelpers.cs
```diff
@@ -0,0 +1,62 @@
...
+         DataFormats.OemText,
+         DataFormats.FileDrop,
+         "FileName",
+         "FileNameW"
+    ];
+
+    public static TheoryData<string> UnboundedFormat() =>
+    [
+        DataFormats.Serializable,
+            "something custom"
```
Tanya-Solyanik, line 30:
```suggestion
        "something custom"
```

src/System.Windows.Forms/tests/TestUtilities/DataObjectTestHelpers.cs
```diff
@@ -0,0 +1,62 @@
...
+        DataFormats.Palette
+    ];
+
+    public static TheoryData<string> BitmapFormat() =>
+    [
+        DataFormats.Bitmap,
+        "System.Drawing.Bitmap"
+    ];
+
+    [Serializable]
```
Tanya-Solyanik, line 55: If we are using this class only with JSON, then it does not have to be serializable

src/System.Windows.Forms/src/System/Windows/Forms/OLE/JsonData.cs
```diff
@@ -0,0 +1,152 @@
...
+}
+
+/// <summary>
+///  Represents an object that contains JSON serialized data. This interface is used to
+///  identify a <see cref="JsonData{T}"/> without needing to have the generic type information.
+/// </summary>
+internal interface IJsonData
+{
+    // We use a custom assembly name to allow versions where JsonData<T> doesn't exist to still be able rehydrate JSON serialized data.
+    public const string CustomAssemblyName = "System.Private.Windows.VirtualJson";
```
Tanya-Solyanik, line 137:
```suggestion
    const string CustomAssemblyName = "System.Private.Windows.VirtualJson";
```

src/System.Windows.Forms/src/System/Windows/Forms/OLE/JsonData.cs
```diff
@@ -0,0 +1,152 @@
...
+
+    /// <summary>
+    ///  The assembly qualified name of the T in <see cref="JsonData{T}"/>. This name should
+    ///  have any <see cref="TypeForwardedFromAttribute"/> names taken into account.
+    /// </summary>
+    string InnerTypeAssemblyQualifiedName { get; }
+
+    /// <summary>
+    ///  Deserializes the data stored in the JsonData. This is a convenience method
+    ///  to deserialize teh data when we are not dealing with a binary formatted record.
```
Tanya-Solyanik, line 149:
```suggestion
    ///  to deserialize the data when we are not dealing with a binary formatted record.
```

src/System.Windows.Forms/tests/TestUtilities/DataObjectTestHelpers.cs
lonitra: This is also used in tests which uses BinaryFormatter to serialize.

src/System.Windows.Forms/src/System/Windows/Forms/Nrbf/WinFormsSerializationRecordExtensions.cs
```diff
@@ -53,6 +58,53 @@ public static bool TryGetBitmap(this SerializationRecord record, out object? bit
...
+        {
+            // This is supposed to be JsonData, but somehow the binary formatted data is corrupt.
+            throw new SerializationException();
+        }
+
+        Type serializedType = resolver.GetType(serializedTypeName);
+        if (!serializedType.IsAssignableTo(typeof(T)))
+        {
+            // Not the type the caller asked for.
+            return false;
```
Tanya-Solyanik, line 92: "not the type the user asked for" should return a true and null object because the caller should exit and return a null to the user

src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.cs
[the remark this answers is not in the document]
Tanya-Solyanik: I agree with argument about exception stirring the user in the right direction. But are you sure that `result is IJsonData` can happen outside an internalsvisible to assembly?

src/System.Windows.Forms/src/System/Windows/Forms/Nrbf/WinFormsSerializationRecordExtensions.cs
lonitra: Ah true. We shouldn't continue trying to deserialize at this point. Thanks for the catch!

src/System.Windows.Forms/src/System/Windows/Forms/OLE/DataObject.cs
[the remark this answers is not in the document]
lonitra: Yes, I had tested this in our scratch project which does not have InternalsVisibleTo and this is the following results:
```c#
        TestData testData = new() { x = 1 };
        DataObject dataObject = new DataObject();
        dataObject.SetDataAsJson("test", testData);

        // This is null.
        var test1 = dataObject.GetData("test");


        Clipboard.SetDataObject(dataObject, copy: false);
        // This is null.
        var test2 = Clipboard.GetData("test");

        Clipboard.SetDataObject(dataObject, copy: true);
        // Empty memory stream. If first chance exception on, throws SerializationException when path to use BinaryFormatter enabled
        var test3 = Clipboard.GetData("test");
```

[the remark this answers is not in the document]
Tanya-Solyanik: Bu are you hitting a breakpoint on line 243 with result not null?

src/System.Windows.Forms/src/System/Windows/Forms/OLE/JsonData.cs
```diff
@@ -0,0 +1,152 @@
...
+///         || !System.Reflection.Metadata.TypeName.TryParse(innerTypeName.ToCharArray(), out System.Reflection.Metadata.TypeName? result))
+///     {
+///         // This is supposed to be JsonData, but somehow the data is corrupt.
+///         throw new InvalidOperationException();
+///     }
+///
+///     // TODO: Additional checking on result TypeName to ensure it is expected type.
+///
+///     // This should return the original data that was JSON serialized.
+///     System.Text.Json.JsonSerializer.Deserialize(byteData.GetArray(), genericType);
```
Tanya-Solyanik, line 75:
```suggestion
///     var result = System.Text.Json.JsonSerializer.Deserialize(byteData.GetArray(), genericType);
///     // TODO: Process the payload as needed.      
```

Tanya-Solyanik (approved):

elachlan:
Congratulations on getting this merged!

Issue #11227, opened 2024-04-17
Unstable unit tests that need to be disabled and fixed


LeafShi1:
**Disable failing tests and open tracking bugs**

1. When test fails in the CI or PR build, open a bug for each test fail to investigate the test further and add a `disabled-test` label to the bug.
2. Disable test by adding a [skip] attribute, If only a specific theory data causes the failure, comment that data out. (In order to prevent some problems from being caused by changes in the PR itself or machine problems, if it fails only once in a month, we will not disable it. If it fails a second time, we will disable it.)
4. Add a link to the tracking bug to the SkipAttribute of the comment line.
5. Collect all data relevant to the failure into the bug.
6. Try to root-cause the flakiness, and fix these tests:
      - Replacing SendInput or cursor-positioning steps with direct method invocations, or by sending WM_COMMAND
     -  If UI element is easily identifiable, for example by and accessible name, or we have access to the AccessibleObject, we can use UIA 
        APIs to invoke methods
      - Add logging if possible.
      - If the test must use `SendInput`, then make sure it is moved to the UI Integration tests assembly

**The following unit tests exist flaky issues**:

- [X] System.Windows.Forms.Tests.ImageCollectionTests.ImageCollection_Item_Get32bppColorDepth_Success(pixelFormat: Format1bppIndexed, pixel00Color: Color [Empty], givenPixel10Color: Color [Empty], expectedPixel10Color: Color [A=255, R=0, G=0, B=0])
- [X] System.Windows.Forms.Tests.ToolTipTests.ToolTip_SetToolTip_TabControl_DoesNotAddToolForTabControlItself
- [X] System.Windows.Forms.Tests.ToolTipTests.ToolTip_WmShow_Invokes_AnnounceText_WithExpectedText_ForTabControlTabs
- [x] System.Windows.Forms.Tests.DataGridViewTests.DataGridView_OnColumnHeadersHeightChanged_InvokeWithHandle_CallsColumnHeadersHeightChanged(columnHeadersWidthSizeMode: EnableResizing, columnHeadersVisible: True, eventArgs: null)
- [x] System.Windows.Forms.UITests.ButtonTests.Button_CancelButton_EscapeClicksCancelButtonAsync
- [x] System.Windows.Forms.UITests.ButtonTests.Button_DialogResult_SpaceToClickFocusedButtonAsync
- [x] System.Windows.Forms.UITests.ButtonTests.Button_Hotkey_Fires_OnClickAsync
[...]

PR #11215, opened 2024-04-14
Code cleanup only and Add VB Tests

paul1956:
Fixes part of #10090
Standalone PR to do some cleanup of VB Code
Removes VB Option from source files and moves to Project File
Correct Spelling and Capitalization Errors
Add a few tests
Update editorConfig to support converting 3 NotInheritable classes to Modules.
Add a few tests
Change format of XLM Comment to indent text by 1 space
Change a few Private functions to Friend for testing
Not all issues are addressed in files I expect to replace.
Not done is reorganizing code in files to be in standard order, this will be a follow-on PR

## Proposed changes
Fix above with very minimal code changes 

## Customer Impact

- None only developers will notice
- 

## Regression? 

 No

## Risk

- Minimal due to almost no code changes

## Test methodology 

Added tests where required otherwise covered by existing tests

Visual Basic only

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11215)

Issue #1999, opened 2019-10-01
Tests deadlock


RussKie:
* .NET Core Version: 5.0

**Problem description:**

Running tests from the master branch locally I got tests hung. I have attached a debugger and there was a dozen of hung threads.

9 threads were stuck unable to acquire the lock in `NativeWindow.CreateHandle()` method (originating stacks were slightly different, but they all ended in the same critical section):
![image](https://user-images.githubusercontent.com/4403806/65936205-37297280-e45f-11e9-88f1-bceed2dee9a7.png)

1 thread was stuck restoring the default trace listener:
![image](https://user-images.githubusercontent.com/4403806/65936554-570d6600-e460-11e9-9c15-77b331d5fe41.png)

2 thread were stuck while writing a trace:
![image](https://user-images.githubusercontent.com/4403806/65936617-7e643300-e460-11e9-85b9-c80a33254db5.png)
![image](https://user-images.githubusercontent.com/4403806/65936653-9cca2e80-e460-11e9-9661-8eb754b89cff.png)

What appears to have happened is that one test was in a middle of creating a handle and attempted to add a trace, which was locked by another thread, that was restoring the default trace listener:
![image](https://user-images.githubusercontent.com/4403806/65937120-25959a00-e462-11e9-8900-9984cb95047a.png)

**Expected behavior:**

No deadlock

**Minimal repro:**

This is very intermittent...

RussKie:
Few questions here:
- do we need to lock in `NoAssertContext`? Can we make `s_defaultListener` volatile or use `Interlocked` object?
- should all out tests dealing with UI (e.g. instantiating controls) be decorated with `StaFactAttribute` or [`WinFormsFactAttribute`](https://github.com/AArnott/Xunit.StaFact/blob/master/src/Xunit.StaFact/WinFormsFactAttribute.cs)

/cc: @hughbe

weltkante:
Creating a control handle on a non-STA thread is a bad idea, it turns the thread pool thread into a UI thread by allocating a message queue etc. but without telling COM that the thread is STA, so controls using COM internally (e.g. RichTextBox) will be pretty confused, if working at all.

JeremyKuhne:
> Should all out tests dealing with UI (e.g. instantiating controls) be decorated with StaFactAttribute or WinFormsFactAttribute

I think `WinFormsFactAttribute` is the right response here

weltkante:
[...]

PR #11530, opened 2024-06-13
Add disable parallelization for collection sequential

LeafShi1:
related #11227 

## Proposed changes

- Add `[CollectionDefinition("Sequential", DisableParallelization = true)]` for collection `Sequential`

### DIFF
--- a/src/Microsoft.VisualBasic/tests/UnitTests/Microsoft/VisualBasic/MyServices/ClipboardProxyTests.cs
+++ b/src/Microsoft.VisualBasic/tests/UnitTests/Microsoft/VisualBasic/MyServices/ClipboardProxyTests.cs
@@ -9,6 +9,7 @@ using TextDataFormat = System.Windows.Forms.TextDataFormat;
 namespace Microsoft.VisualBasic.MyServices.Tests;
 
 [Collection("Sequential")]
+[CollectionDefinition("Sequential", DisableParallelization = true)]
 public class ClipboardProxyTests
 {
     [WinFormsFact]
--- a/src/System.Windows.Forms/tests/UnitTests/TextBoxBaseTests.cs
+++ b/src/System.Windows.Forms/tests/UnitTests/TextBoxBaseTests.cs
@@ -4131,7 +4131,10 @@ public partial class TextBoxBaseTests
         Assert.Equal(0, createdCallCount);
     }
 
+    [ActiveIssue("https://github.com/dotnet/winforms/issues/11497")]
     [WinFormsFact]
+    [SkipOnArchitecture(TestArchitectures.X86,
+        "Flaky tests, see: https://github.com/dotnet/winforms/issues/11497")]
     public void TextBoxBase_ClearUndo_CanUndo_Success()
     {
         using SubTextBox control = new()
@@ -4216,7 +4219,10 @@ public partial class TextBoxBaseTests
         Assert.Equal(0, createdCallCount);
     }
 
+    [ActiveIssue("https://github.com/dotnet/winforms/issues/11498")]
     [WinFormsFact]
+    [SkipOnArchitecture(TestArchitectures.X86,
+        "Flaky tests, see: https://github.com/dotnet/winforms/issues/11498")]
     public void TextBoxBase_Copy_PasteNotEmptyWithHandle_Success()
     {
         using SubTextBox control = new()


LeafShi1:
Adopt Paul's comment in commit [4d471144 ](https://github.com/paul1956/winforms/commit/42459afe8df15476d4c116c500b0546eb159c215#commitcomment-142992766)

LeafShi1:
@paul1956 Please help review this PR

src/Microsoft.VisualBasic/tests/UnitTests/Microsoft/VisualBasic/MyServices/ClipboardProxyTests.cs
```diff
@@ -9,6 +9,7 @@
 namespace Microsoft.VisualBasic.MyServices.Tests;
 
 [Collection("Sequential")]
+[CollectionDefinition("Sequential", DisableParallelization = true)]
```
Tanya-Solyanik, line 12: @lonitra  - I seem to remember that this had not worked in the past, to you recall any details? 
@LeafShi1 - had you tested this change?

Tanya-Solyanik: This attribute should be placed on the "collection definition class" - https://xunit.net/docs/running-tests-in-parallel so it would apply only to this class... Ideally we would want all clipboard tests across the assembly in a single collection.

lonitra: We do not have a `CollectionDefinition` for "Sequential". I believe we had only been using `Collection("Sequential")` to ensure all tests within the collection do not run at the same time, but hadn't restricted to also not run in parallel with tests in other collections.

Tanya-Solyanik: I see, xunit runs all tests within a collection, test class is always a collection and tests within a class run sequentially. We use the [Collection()] attribute to extend the set of tests that run sequentially. 
So `[CollectionDefinition("Sequential", DisableParallelization = true)]` is not needed.

paul1956:
@LeafShi1 @Tanya-Solyanik I think there may be two issues here. In my experience without
```
[CollectionDefinition("Sequential", DisableParallelization = true)]
```
my VB tests #11215 failed randomly with same error others reported. With above Attribute all my new tests in VisualBasic Runitme (Written in both C# and VB) in separate projects pass when I run just the tests I changed against the VB runtime. But when I run all tests, it sometimes fails with the Clipboard being corrupted between setting and reading.

I believe the issue is some existing tests (not part of Visual Basic Runtime) use the Clipboard (at least the ones being disabled as Flakey) plus possibly more need to be added to the same collection. I don't know if doing that removes the requirement to add the CollectionDefination Attribute, all the searching seems to say you need collection name and collection definition.

My other point was is the Clipboard the only shared resource in the repo? If it is than using "Sequential" as the name is fine, but if there are other similar resources they should have their own collection name so that they can run in parallel with tests using Clipboard.

Tanya-Solyanik:
>[CollectionDefinition("Sequential", DisableParallelization = true)]

is an incorrect use of this attribute according to xunit docs. Your other 2 points are valid but shoudl be addressed in a different PR
1. Tests that exercise clipboard directly (ClipboardTests class) as well as all other tests that use clipboard indirectly(for example copy/paste in some datagrid tests) should be in a single collection
2. Other tests that use global resources don't have to be in the same collection as the clipboard tests are.

paul1956:
@Tanya-Solyanik 
> 1. Tests that exercise clipboard directly (ClipboardTests class) as well as all other tests that use clipboard indirectly(for example copy/paste in some datagrid tests) should be in a single collection

There are also and will be additional Clipboard Proxy tests in VB runtime that depend on Clipboard.

Whatever is decided for [CollectionDefinition("Sequential", DisableParallelization = true)] I will just follow.

Its interesting that CursorTestsCollection uses both
[Collection(nameof(CursorTestsCollection))]
[CollectionDefinition(nameof(CursorTestsCollection), DisableParallelization = true)]

For Clipboard the nameof will not work because multiple different classes use Clipboard

paul1956:
There are many different test classes in different projects and namespaces that use clipboard that is why including the second attribute is needed unless you want all control test mixed into one class. Sent from my iPhoneI apologize for any typos Siri might have made.(503) 803-6077On Jun 13, 2024, at 9:06 AM, Tanya Solyanik ***@***.***> wrote:﻿
@Tanya-Solyanik commented on this pull request.



In src/Microsoft.VisualBasic/tests/UnitTests/Microsoft/VisualBasic/MyServices/ClipboardProxyTests.cs:
> @@ -9,6 +9,7 @@
 namespace Microsoft.VisualBasic.MyServices.Tests;
 
 [Collection("Sequential")]
+[CollectionDefinition("Sequential", DisableParallelization = true)]

This attribute should be placed on the "collection definition class" - https://xunit.net/docs/running-tests-in-parallel so it would apply only to this class... Ideally we would want all clipboard tests across the assembly in a single collection.

paul1956:
The issue is tests are in different classes. Attribute [CollectionDefinition(DisableParallelization = true)] doesn't prevent parallel execution between class · Issue #1999 · xunit/xunitgithub.comSent from my iPhoneI apologize for any typos Siri might have made.(503) 803-6077On Jun 14, 2024, at 11:22 AM, Tanya Solyanik ***@***.***> wrote:﻿
@Tanya-Solyanik commented on this pull request.



In src/Microsoft.VisualBasic/tests/UnitTests/Microsoft/VisualBasic/MyServices/ClipboardProxyTests.cs:
> @@ -9,6 +9,7 @@
 namespace Microsoft.VisualBasic.MyServices.Tests;
 
 [Collection("Sequential")]
+[CollectionDefinition("Sequential", DisableParallelization = true)]

I see, xunit runs all tests within a collection, test class is always a collection and tests within a class run sequentially. We use the [Collection()] attribute to extend the set of tests that run sequentially.
So [CollectionDefinition("Sequential", DisableParallelization = true)] is not needed.

src/Microsoft.VisualBasic/tests/UnitTests/Microsoft/VisualBasic/MyServices/ClipboardProxyTests.cs
LeafShi1: > I see, xunit runs all tests within a collection, test class is always a collection and tests within a class run sequentially. We use the [Collection()] attribute to extend the set of tests that run sequentially. So `[CollectionDefinition("Sequential", DisableParallelization = true)]` is not needed.

So the attribute should be placed on the test suite definition class and not here, right?

paul1956:
@LeafShi1 @Tanya-Solyanik the problem with moving “Collection” attribute to a class is the Clipboard is shared by many test classes in multiple projects. Logically tests are not groups by those that use Clipboard and those that don’t, they are grouped by control or class the objects they are testing are in. Does it make sense to group the VB Clipboard Proxy tests with TextBox tests I don’t think so.

LeafShi1:
And according to the document, add attribute `[CollectionDefinition(DisableParallelization = true)]`  only affect current assembly
image

Issue #11536, opened 2024-06-15
Define new scopes for AppContext switches for use in tests


Tanya-Solyanik:
1. replace use cases of [SetLocalAppContextSwitchValue](https://github.com/search?q=repo%3Adotnet%2Fwinforms%20SetLocalAppContextSwitchValue&type=code) in tests with custom scope for each switch see [BinaryFormatterScope](https://github.com/search?q=repo%3Adotnet%2Fwinforms%20BinaryFormatterScope&type=code) example.
2. Replace use cases of TestAccessor like this one - https://github.com/dotnet/winforms/blob/29528171490190df28fd4afc89dee271ebfef68b/src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/Dpi/FormDpiTests.cs#L111 with AppContext switch scopes as well

elachlan:
@Tanya-Solyanik can we resolve this now? I am not sure if the PR covered everything. I thought I got all the usages, but could be wrong.

PR #11539, opened 2024-06-16
Replacing usages of `LocalAppContextSwitches` with `AppContextSwitchScope`

elachlan:
Working on #11536

Only thing left is `LocalAppContextSwitchesTest`, but I am unsure how to proceed on that. Since its directly testing `LocalAppContextSwitches`.
https://github.com/dotnet/winforms/blob/79472b41928fe7147527f93651f828088f617c1e/src/System.Windows.Forms.Primitives/tests/UnitTests/System/Windows/Forms/LocalAppContextSwitchesTest.cs#L9-L44
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11539)


src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/AnchorLayoutTests.cs
```diff
@@ -44,23 +43,14 @@ public void Control_ResizeAnchoredControls_ParentHandleCreated_NewAnchorsApplied
...
-        try
-        {
-            LaunchFormAndVerify(anchors, expectedX, expectedY, expectedWidth, expectedHeight);
-        }
-        finally
-        {
-            // Reset switch.
-            SetAnchorLayoutV2Switch(previousLayoutSwitch);
-        }
+        using AnchorLayoutV2Scope scope = new(true);
```
Tanya-Solyanik, line 46:
```suggestion
        using AnchorLayoutV2Scope scope = new(enabled: true);
```

src/Common/tests/TestUtilities/ScaleTopLevelFormMinMaxSizeForDpiScope.cs
```diff
@@ -0,0 +1,32 @@
...
+namespace System;
+
+/// <summary>
+/// Scope for enabling / disabling the ScaleTopLevelFormMinMaxSizeForDpi Switch. Use in a <see langword="using"/> statement.
+/// </summary>
+public readonly ref struct ScaleTopLevelFormMinMaxSizeForDpiScope
+{
+    private readonly AppContextSwitchScope _switchScope;
+
+    public ScaleTopLevelFormMinMaxSizeForDpiScope(bool enable)
```
Tanya-Solyanik, line 13: Add a constructor that takes target framework and uses s_targetFrameworkName to set it on the switch. It will be used to test default values in LocalAppContextSwitchesTest

src/Common/tests/TestUtilities/AppContextSwitchNames.cs
```diff
@@ -7,6 +7,25 @@ namespace System;
 
 public static class AppContextSwitchNames
 {
+    /// <summary>
+    /// The switch that controls whether AnchorLayoutV2 feature is enabled.
+    /// </summary>
+    public static string AnchorLayoutV2 { get; }
+        = "System.Windows.Forms.AnchorLayoutV2";
```
Tanya-Solyanik, line 14: Had you tried making Primitives' internals visible to tis test?

Tanya-Solyanik: Never mind, if we make Common.TestUtilities dependent on Primitives, Analyzer tests don't compile

src/Common/tests/TestUtilities/AppContextSwitchNames.cs
```diff
@@ -7,6 +7,25 @@ namespace System;
 
 public static class AppContextSwitchNames
 {
+    /// <summary>
+    /// The switch that controls whether AnchorLayoutV2 feature is enabled.
+    /// </summary>
+    public static string AnchorLayoutV2 { get; }
```
Tanya-Solyanik, line 13: These could be const fields rather than properties.

src/Common/tests/TestUtilities/AppContextSwitchScope.cs
```diff
@@ -17,15 +19,15 @@ namespace System;
     private readonly string _switchName;
     private readonly bool _originalState;
 
-    public AppContextSwitchScope(string switchName, bool enable)
+    public AppContextSwitchScope(string? targetFrameworkName, string switchName, bool enable)
```
Tanya-Solyanik, line 22: There is no reason to set both target framework and enabled state because these parameters will be used in different tests. You will set `enabled` in tests that verify functionality, but the framework name is used in tests that verify default switch value because default enabled state depends on the framework.

src/Common/tests/TestUtilities/AppContextSwitchScope.cs
```diff
@@ -17,15 +19,15 @@ namespace System;
...
     {
         if (!AppContext.TryGetSwitch(AppContextSwitchNames.LocalAppContext_DisableCaching, out bool isEnabled)
             || !isEnabled)
         {
             // It doesn't make sense to try messing with AppContext switches if they are going to be cached.
             throw new InvalidOperationException("LocalAppContext switch caching is not disabled.");
         }
-
+
```
Tanya-Solyanik, line 30: remove the white spaces

src/Common/tests/TestUtilities/AppContextSwitchScope.cs
Tanya-Solyanik: Target framework defined the default switch value, and 'enable' always overrides the default, so there is no reason to set both in tests.

elachlan: I ended up creating `TargetFrameworkNameScope`. It isolates the logic. Otherewise I had to move the files to `System.Windows.Forms.Primitives.TestUtilities`. Which I wasn't a fan of, since only some of the Scope classes needed to support targetframework.

Tanya-Solyanik: This is a better approach.

src/Common/tests/TestUtilities/AnchorLayoutV2Scope.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace System;
+
+/// <summary>
+/// Scope for enabling / disabling the AnchorLayoutV2 Switch. Use in a <see langword="using"/> statement.
```
Tanya-Solyanik, line 7:
```suggestion
///  Scope for enabling / disabling the AnchorLayoutV2 Switch. Use in a <see langword="using"/> statement.
```

src/Common/tests/TestUtilities/AnchorLayoutV2Scope.cs
```diff
@@ -0,0 +1,32 @@
...
+public readonly ref struct AnchorLayoutV2Scope
+{
+    private readonly AppContextSwitchScope _switchScope;
+
+    public AnchorLayoutV2Scope(bool enable)
+    {
+        // Prevent multiple AnchorLayoutV2Scope instances from running simultaneously.
+        // Using Monitor to allow recursion on the same thread.
+        Monitor.Enter(typeof(AnchorLayoutV2Scope));
+        _switchScope = new AppContextSwitchScope(AppContextSwitchNames.AnchorLayoutV2, enable);
```
Tanya-Solyanik, line 18:
```suggestion
        _switchScope = new (AppContextSwitchNames.AnchorLayoutV2, enable);
```

src/Common/tests/TestUtilities/AnchorLayoutV2Scope.cs
Tanya-Solyanik: This applies to several files in this PR

src/System.Windows.Forms.Primitives/tests/TestUtilities/TargetFrameworkNameScope.cs
```diff
@@ -0,0 +1,41 @@
...
+
+using System.Runtime.Versioning;
+using System.Windows.Forms.Primitives;
+
+namespace System;
+
+#nullable enable
+
+/// <summary>
+/// Scope for setting the TargetFrameworkName temporarily. Use in a <see langword="using"/> statement.
```
Tanya-Solyanik, line 12:
```suggestion
/// Scope for setting the see cref="LocalAppContextSwitches.TargetFrameworkName" /> temporarily. Use in a <see langword="using"/> statement.
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjects/DataGridViewCellAccessibleObjectTests.cs
```diff
@@ -1460,17 +1459,23 @@ private DataGridView CreateDataGridView(int columnCount, bool createControl = tr
...
-        using DataGridView dataGridView = new();
-        dataGridView.Columns.Add(new DataGridViewTextBoxColumn());
-        dataGridView.Rows.Add(new DataGridViewRow());
-
-        Assert.Equal($"{string.Format(SR.DataGridView_AccRowName, 0)}, Not sorted.", dataGridView.Rows[0].Cells[0].AccessibilityObject.Name);
-
-        LocalAppContextSwitches.SetLocalAppContextSwitchValue(LocalAppContextSwitches.DataGridViewUIAStartRowCountAtZeroSwitchName, false);
+        try
+        {
+            using var scope = new DataGridViewUIAStartRowCountAtZeroScope(enable: true);
```
Tanya-Solyanik, line 1466: Please use the target-typed new in this pattern, this comment applies to several files in this PR.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjects/DataGridViewRowAccessibleObjectTests.cs
```diff
@@ -2390,17 +2389,23 @@ public void DataGridViewRowAccessibleObject_GetPropertyValue_ValueValuePropertyI
...
-        using DataGridView dataGridView = new();
-        dataGridView.Columns.Add(new DataGridViewTextBoxColumn());
-        dataGridView.Rows.Add(new DataGridViewRow());
-
-        Assert.Equal(string.Format(SR.DataGridView_AccRowName, 0), dataGridView.Rows[0].AccessibilityObject.Name);
-
-        LocalAppContextSwitches.SetLocalAppContextSwitchValue(LocalAppContextSwitches.DataGridViewUIAStartRowCountAtZeroSwitchName, false);
-
-        Assert.Equal(string.Format(SR.DataGridView_AccRowName, 1), dataGridView.Rows[0].AccessibilityObject.Name);
+        try
```
Tanya-Solyanik, line 2394: Why did you have to replace a using with try/finally?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PictureBoxTests.cs
```diff
@@ -736,13 +735,10 @@ public void PictureBox_ImageLocation_SetValidWithWaitOnLoadTrueLocal_GetReturnsE
     [BoolData]
     public void PictureBox_ImageLocation_SetValidWithWaitOnLoadTrueUri_ConfigSwitch_CheckCRL_GetReturnsExpected(bool switchValue)
     {
-        dynamic testAccessor = typeof(LocalAppContextSwitches).TestAccessor().Dynamic;
+        using ServicePointManagerCheckCrlScope scope = new(enable: switchValue);
```
Tanya-Solyanik, line 738: we usually use argument names before literals, In this case the test parameter name is descriptive enough.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripTests.cs
```diff
@@ -1622,34 +1621,43 @@ public void ToolStrip_Font_SetWithItemsWithFontWithHandler_CallsFontChanged()
     [WinFormsFact]
     public void ToolStrip_Font_ApplyParentFontToMenus_GetReturnFont_SameAsForm()
     {
-        LocalAppContextSwitches.SetLocalAppContextSwitchValue(LocalAppContextSwitches.ApplyParentFontToMenusSwitchName, true);
-
-        using Font font = new("Microsoft Sans Serif", 8.25f);
-        using Form form = new();
-        using ToolStrip toolStrip1 = new();
-        using SubToolStripItem item1 = new();
-        using SubToolStripItem item2 = new();
```
Tanya-Solyanik, line 1631: Why did you change a using expression to explicit Dispose?

src/System.Windows.Forms/tests/IntegrationTests/UIIntegrationTests/AnchorLayoutTests.cs
```diff
@@ -78,15 +68,14 @@ public void Control_NotParented_AnchorsNotComputed()
         finally
         {
             // Reset switch.
```
LeafShi1, line 70: The comments here and in the following tests should be modified or removed

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjects/DataGridViewRowAccessibleObjectTests.cs
elachlan: My assumption is that the scope struct needs to be created prior the the datagridview and then disposed prior to the second `Assert.Equal`. If I add in the braces to do this, it puts dataGridView out of scope for the second assert.

We can avoid it if the scope doesn't need to be created prior to the dataGridView. But we can't assume all consumers of the switch work the same or that they aren't used in the control setup.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripTests.cs
elachlan: The using scope ends up hiding the variables from the final Assert calls. The inverse would be to create our struct to a variable and manually dispose of it in the finally.

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjects/DataGridViewRowAccessibleObjectTests.cs
RussKie: Is there a reason we can't dispose the scope explicitly?

```cs
        DataGridViewUIAStartRowCountAtZeroScope scope = new(enable: true);
        using DataGridView dataGridView = new();
        dataGridView.Columns.Add(new DataGridViewTextBoxColumn());
        dataGridView.Rows.Add(new DataGridViewRow());

        Assert.Equal(string.Format(SR.DataGridView_AccRowName, 0), dataGridView.Rows[0].AccessibilityObject.Name);

        scope.Dispose();

        Assert.Equal(string.Format(SR.DataGridView_AccRowName, 1), dataGridView.Rows[0].AccessibilityObject.Name);
```

elachlan: We could do that, do we need a try finally in case? or will it be cleaned up when it leaves scope?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/ToolStripTests.cs
Tanya-Solyanik: Is it possible to refactor this test?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjects/DataGridViewRowAccessibleObjectTests.cs
Tanya-Solyanik: @elachlan - if you don't have time for the refactor, let me know, it can be done in a follow up PR.

elachlan: I have already refactored to remove the try catch, is this new approach okay?

Tanya-Solyanik: scope.Dispose() will not be called in case of an exception, and that scope might interfere with scopes in other tests.

elachlan: Done.

Tanya-Solyanik (approved):
Thank you!

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/AccessibleObjects/DataGridViewCellAccessibleObjectTests.cs
```diff
@@ -1457,20 +1456,16 @@ private DataGridView CreateDataGridView(int columnCount, bool createControl = tr
...
-    public void DataGridView_SwitchConfigured_AdjustsCellRowStartIndices()
+    [WinFormsTheory]
+    [InlineData([false, 1])]
+    [InlineData([true, 0])]
+    public void DataGridView_SwitchConfigured_AdjustsCellRowStartIndices(bool switchValue, int expectedIndex)
     {
-        LocalAppContextSwitches.SetLocalAppContextSwitchValue(LocalAppContextSwitches.DataGridViewUIAStartRowCountAtZeroSwitchName, true);
-
-        using DataGridView dataGridView = new();
+        using DataGridViewUIAStartRowCountAtZeroScope scope = new(enable: switchValue);
```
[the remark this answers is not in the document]
elachlan: Done!

Tanya-Solyanik (approved):

PR #2124, opened 2019-10-18
use httpclient instead of webrequest

flerka:
Fixes https://github.com/dotnet/winforms/issues/1756
Refactor `PictureBox` to use `HttpClient` instead of using `WebRequest`
Before implementing read comments from this pr https://github.com/dotnet/winforms/pull/1952
Introduced HttpClient field as suggested by  @RussKie  in https://github.com/dotnet/winforms/pull/1952#discussion_r329871280

## Proposed changes
- introduce optional `HttpClient` parameter in `PictureBox` constructor
- use `HttpClient` instead of using `WebRequest`
- use `CancellationToken` to cancel request
- update error handling due to  `HttpClient` behavior 

## Customer Impact

-  may be a breaking change:  introduce optional `HttpClient` parameter in `PictureBox` constructor

## Regression? 

- Yes

## Risk

- Low/Medium

## Test methodology 

- Run all tests in the solution
- Manual testing and debugging with test project according to debugging guide for various cases (success, unsuccessful such as remote server fail, non-existed image etc.)

 

## Test environment(s) 
```

.NET Core SDK (reflecting any global.json):
 Version:   3.0.100
 Commit:    04339c3a26

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.18362
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\3.0.100\

Host (useful for support):
  Version: 3.0.0
  Commit:  7d57652f33

.NET Core SDKs installed:
  2.1.801 [C:\Program Files\dotnet\sdk]
  2.2.401 [C:\Program Files\dotnet\sdk]
  3.0.100 [C:\Program Files\dotnet\sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.All 2.1.12 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.13 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.2.6 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.2.7 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.1.12 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
[...]

PR #11542, opened 2024-06-17
use HttpClient to replace WebClient

kasperk81:
this will unblock https://github.com/dotnet/sdk/pull/41616
see https://github.com/dotnet/runtime/pull/103456

fix https://github.com/dotnet/winforms/issues/11544

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11542)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Controls/PictureBox/PictureBox.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Controls/PictureBox/PictureBox.cs
@@ -26,6 +26,9 @@ public partial class PictureBox : Control, ISupportInitialize
         !AppContext.TryGetSwitch("System.Windows.Forms.PictureBox.UseWebRequest", out bool useWebRequest)
         || useWebRequest;
 
+    private static readonly HttpClient s_httpClient = !LocalAppContextSwitches.ServicePointManagerCheckCrl ? new() :
+        new(new HttpClientHandler { CheckCertificateRevocationList = true });
+
     /// <summary>
     ///  The type of border this control will have.
     /// </summary>
@@ -464,23 +467,26 @@ public partial class PictureBox : Control, ISupportInitialize
         try
         {
             DisposeImageStream();
-            if (UseWebRequest())
+            Uri uri = CalculateUri(_imageLocation);
+            if (uri.IsFile)
             {
-                LoadImageViaWebClient();
+                _localImageStreamReader = new StreamReader(uri.LocalPath);
+                Image img = Image.FromStream(_localImageStreamReader.BaseStream);
+                InstallNewImage(img, ImageInstallationType.FromUrl);
+            }
+            else if (UseWebRequest())
+            {
+                // Run async operation synchronously to avoid blocking UI thread and potential deadlocks.
+                Task.Run(async () =>
+                {
+                    _uriImageStream = await s_httpClient.GetStreamAsync(uri).ConfigureAwait(false);
+                    Image img = Image.FromStream(_uriImageStream);
+                    InstallNewImage(img, ImageInstallationType.FromUrl);
+                }).GetAwaiter().GetResult();
             }
             else
             {
-                Uri uri = CalculateUri(_imageLocation);
-                if (uri.IsFile)
-                {
-                    _localImageStreamReader = new StreamReader(uri.LocalPath);
-                    Image img = Image.FromStream(_localImageStreamReader.BaseStream);
-                    InstallNewImage(img, ImageInstallationType.FromUrl);
-                }
-                else
-                {
-                    throw new NotSupportedException(SR.PictureBoxRemoteLocationNotSupported);
-                }
+                throw new NotSupportedException(SR.PictureBoxRemoteLocationNotSupported);
             }
         }
         catch
@@ -495,34 +501,6 @@ public partial class PictureBox : Control, ISupportInitialize
                 InstallNewImage(ErrorImage, ImageInstallationType.ErrorOrInitial);
             }
         }
-    }
-
-    private void LoadImageViaWebClient()
-    {
-        Image img;
-        Uri uri = CalculateUri(_imageLocation!);
-        if (uri.IsFile)
-        {
-            _localImageStreamReader = new StreamReader(uri.LocalPath);
-            img = Image.FromStream(_localImageStreamReader.BaseStream);
-        }
-        else
-        {
-            if (LocalAppContextSwitches.ServicePointManagerCheckCrl)
-            {
-#pragma warning disable SYSLIB0014 // Type or member is obsolete
-                ServicePointManager.CheckCertificateRevocationList = true;
-#pragma warning restore SYSLIB0014
-            }
-
-#pragma warning disable SYSLIB0014 // Type or member is obsolete
-            using WebClient webClient = new(); // lgtm[cs/webrequest-checkcertrevlist-disabled] - Having ServicePointManager.CheckCertificateRevocationList set to true has a slim chance of resulting in failure. We have an opt-out for this rare event.
-#pragma warning restore SYSLIB0014 // Type or member is obsolete
-            _uriImageStream = webClient.OpenRead(uri.ToString());
-            img = Image.FromStream(_uriImageStream);
-        }
-
-        InstallNewImage(img, ImageInstallationType.FromUrl);
     }
 
     [SRCategory(nameof(SR.CatAsynchronous))]


akoeplinger:
this seems like a more involved change, I think I'd prefer suppressing the warning (the code is already doing that in some places) and file a follow-up issue.

akoeplinger:
There was an earlier attempt in https://github.com/dotnet/winforms/pull/2124. There's also https://github.com/dotnet/winforms/issues/1756 which I think was inadvertently closed when https://github.com/dotnet/winforms/pull/6684 was merged.

akoeplinger:
@kasperk81 the winforms maintainers need to ultimately decide but I think we'd get the flow unblocked more easily if we suppress the warning for now.

paul1956:
> I think was working on similar things for Winforms VB.Net.

@elachlan I already did all the work for download in PR #9867 and PR #11215 but they are not being reviewed. Upload is a trivial change but writing a test server for upload is well beyond my ability. None of the VB Code Upload/Download or FileIO has tests on Main branch, I wrote then to make sure error codes/exceptions are exactly the same. One minor issue is getting the error codes exactly the same requires SR entries and translation that I don't understand.

lonitra:
@kasperk81 Is this still blocking for https://github.com/dotnet/sdk/pull/41616? This was suppressed in a recent depenedency update.

src/System.Windows.Forms/src/System/Windows/Forms/Controls/PictureBox/PictureBox.cs
```diff
@@ -464,23 +464,29 @@ public void Load()
...
-                if (uri.IsFile)
+                Task.Run(async () =>
                 {
-                    _localImageStreamReader = new StreamReader(uri.LocalPath);
-                    Image img = Image.FromStream(_localImageStreamReader.BaseStream);
+                    HttpResponseMessage response = await _httpClient.GetAsync(uri).ConfigureAwait(false);
+                    response.EnsureSuccessStatusCode();
+
+                    using Stream stream = await response.Content.ReadAsStreamAsync().ConfigureAwait(false);
+                    _uriImageStream = stream;
```
lonitra, line 482: This seems equivalent to `HttpClient.GetStreamAsync`
```suggestion
                    _uriImageStream = await _httpClient.GetStreamAsync(uri);
```

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PictureBoxTests.cs
```diff
@@ -2002,17 +1994,13 @@ public void PictureBox_Load_UrlValidWithWaitOnLoadTrueLocal_GetReturnsExpected()
         Assert.Equal(new Size(110, 100), pictureBox.Image.Size);
     }
 
-    [WinFormsTheory]
-    [BoolData]
-    public void PictureBox_Load_UrlValidWithWaitOnLoadTrueUri_ConfigSwitch_CheckCRL_GetReturnsExpected(bool switchValue)
+    [WinFormsFact]
+    public void PictureBox_Load_UrlValidWithWaitOnLoadTrueUri_ConfigSwitch_CheckCRL_GetReturnsExpected()
```
lonitra, line 1998:
```suggestion
    public void PictureBox_Load_UrlValidWithWaitOnLoadTrueUri_GetReturnsExpected()
```

src/System.Windows.Forms/src/System/Windows/Forms/Controls/PictureBox/PictureBox.cs
```diff
@@ -62,6 +61,7 @@ public partial class PictureBox : Control, ISupportInitialize
     private SendOrPostCallback? _loadProgressDelegate;
     private bool _handleValid;
     private readonly object _internalSyncObject = new();
+    private readonly HttpClient _httpClient = new();
```
elachlan, line 64: Why is this not static?

src/System.Windows.Forms/tests/UnitTests/System/Windows/Forms/PictureBoxTests.cs
```diff
@@ -2002,17 +1994,13 @@ public void PictureBox_Load_UrlValidWithWaitOnLoadTrueLocal_GetReturnsExpected()
         Assert.Equal(new Size(110, 100), pictureBox.Image.Size);
     }
 
-    [WinFormsTheory]
-    [BoolData]
-    public void PictureBox_Load_UrlValidWithWaitOnLoadTrueUri_ConfigSwitch_CheckCRL_GetReturnsExpected(bool switchValue)
+    [WinFormsFact]
+    public void PictureBox_Load_UrlValidWithWaitOnLoadTrueUri_GetReturnsExpected()
     {
         dynamic testAccessor = typeof(LocalAppContextSwitches).TestAccessor().Dynamic;
```
RussKie, line 2000: This is no longer needed. Nor try/catch for that matter - if we fail to load the image then the test must fail.

Ditto above.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -18,7 +18,6 @@ internal static partial class LocalAppContextSwitches
     private const string ScaleTopLevelFormMinMaxSizeForDpiSwitchName = "System.Windows.Forms.ScaleTopLevelFormMinMaxSizeForDpi";
```
RussKie, line 18: The link above is broken, it should read:
```
https://github.com/dotnet/winforms/blob/main/docs/design/anchor-layout-changes-in-net80.md
```

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
```diff
@@ -18,7 +18,6 @@ internal static partial class LocalAppContextSwitches
...
     internal const string AnchorLayoutV2SwitchName = "System.Windows.Forms.AnchorLayoutV2";
     internal const string ApplyParentFontToMenusSwitchName = "System.Windows.Forms.ApplyParentFontToMenus";
-    internal const string ServicePointManagerCheckCrlSwitchName = "System.Windows.Forms.ServicePointManagerCheckCrl";
```
RussKie, line 21: ❗ This switch is documented in https://learn.microsoft.com/dotnet/api/system.windows.forms.picturebox.load, so it's removal must be reflected in the docs.

src/System.Windows.Forms/src/System/Windows/Forms/Controls/PictureBox/PictureBox.cs
```diff
@@ -464,23 +465,25 @@ public void Load()
...
-                    Image img = Image.FromStream(_localImageStreamReader.BaseStream);
+                    _uriImageStream = await s_httpClient.GetStreamAsync(uri).ConfigureAwait(false);
+                    Image img = Image.FromStream(_uriImageStream);
                     InstallNewImage(img, ImageInstallationType.FromUrl);
-                }
-                else
-                {
-                    throw new NotSupportedException(SR.PictureBoxRemoteLocationNotSupported);
-                }
+                }).GetAwaiter().GetResult();
```
RussKie, line 482: Why do we need to wrap the whole thing in task just to synchronously block on it to get the result?
Why not get the result straight from `GetStreamAsync`? E.g., something like this:
```cs
                    _uriImageStream = await s_httpClient.GetStreamAsync(uri).Result;
                    Image img = Image.FromStream(_uriImageStream);
                    InstallNewImage(img, ImageInstallationType.FromUrl);
```

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
RussKie: Thank you 🤝 

I'm mainly highlighting this for the @dotnet/dotnet-winforms team.

src/System.Windows.Forms/src/System/Windows/Forms/Controls/PictureBox/PictureBox.cs
RussKie: I'm not sure :) I asked why the code you proposed was better as it seemed excessive yet (at least on the surface) appeared to do the same thing.

I did some reading since (e.g., [this SO thread seems good](https://stackoverflow.com/a/38530225/2338036)), and appears you original code was better in a sense that it'd still be blocking but at least it wouldn't be wrapping exceptions, if one were to occur on a background thread. 

I retract my original proposal and suggest going back to you original implementation. Though it's worth adding a comment explaining _why_ we're doing it like this.

paul1956: Are the error codes the same with WebClient and httpClient, for example when the remote address does not exist or you don’t have permission to read the file? I had to do a lot to return identical errors that I don’t see here.

paul1956: Yes I did all that and added the missing SR entries in my VB PR so tests pass before and after change. I wish I had know it was acceptable to change exceptions codes. The problem with not doing this is the user (or VB runtime in my case) had no tests for network errors so would just crashed the application in the unlikely case of an error.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
MihaZupan: FWIW CRL checking is **not** on by default with HttpClient if that was the intention (controlled by `handler.CheckCertificateRevocationList`).

src/System.Windows.Forms/src/System/Windows/Forms/Controls/PictureBox/PictureBox.cs
paul1956: @kasperk81 the 2 expected WebException are SR.net_webstatus_Unauthorized  and  SR.net_webstatus_Timeout are these being lost by throwing  InvalidOperationException? The first is permanent and the second transient maybe this doesn't matter in PictureBox.

src/System.Windows.Forms.Primitives/src/System/LocalAppContextSwitches/LocalAppContextSwitches.cs
lonitra: It would be good to update HttpClient to check CRL

lonitra (approved):
Changes LGTM, thank you! 🥳 We'll still need to do some slight documentation adjustments to reflect WebClient is no longer being used

RussKie:
@lonitra I think this should be filed under "breaking change" with relevant docs, release docs and breaking change docs updated/generated.

paul1956:
I think this is worse than a breaking change. Current code can distinguish between a wrong password and a offline server or bad URL. After this change everything is an IO Error. This may not apply to loading an image but does apply to replacing WebClient.Sent from my iPhoneI apologize for any typos Siri might have made.(503) 803-6077On Jun 19, 2024, at 2:51 PM, Igor Velikorossov ***@***.***> wrote:﻿
@lonitra I think this should be filed under "breaking change" with relevant docs, release docs and breaking change docs updated/generated.

paul1956:
@kasperk81 my comments refer to the VB network code which currently uses WebClient also and does expose WebClient Exceptions to code calling the network functions. I have not looked at other uses. The comments here seem to say this breaking change is OK but I was hoping for a more holistic solution that covered all WebClient usage.

lonitra:
> Current code can distinguish between a wrong password and a offline server or bad URL. After this change everything is an IO Error.

I'm not sure I follow this. My understanding here is WebClient throws WebException for most of its methods and if we want to find out more about the error, we would need to refer to [WebExceptionStatus](https://learn.microsoft.com/en-us/dotnet/api/system.net.webexceptionstatus?view=net-8.0). HttpRequestException similarly has [HttpStatusCode](https://learn.microsoft.com/en-us/dotnet/api/system.net.httpstatuscode?view=net-8.0) which can be checked to determine more information about the error that occurred. I'm not familiar with VB network code usage of WebClient, but if we had previously handled status codes of the WebExceptions then we should make adjustments to continue to handle them in the same way with HttpRequestExceptions, but in PictureBox case we had not handled it. We have filed breaking change so it is up to caller to handle this for their specific scenario

paul1956:
@lonitra 

All of the VB upload and download functions look like this and they rethrow the WebClient exception. This is from Main
```
        Public Sub DownloadFile(address As Uri, destinationFileName As String)
            Debug.Assert(m_WebClient IsNot Nothing, "No WebClient")
            Debug.Assert(address IsNot Nothing, "No address")
            Debug.Assert((Not String.IsNullOrWhiteSpace(destinationFileName)) AndAlso Directory.Exists(Path.GetDirectoryName(Path.GetFullPath(destinationFileName))), "Invalid path")

            ' If we have a dialog we need to set up an async download
            If m_ProgressDialog IsNot Nothing Then
                m_WebClient.DownloadFileAsync(address, destinationFileName)
                m_ProgressDialog.ShowProgressDialog() 'returns when the download sequence is over, whether due to success, error, or being canceled
            Else
                m_WebClient.DownloadFile(address, destinationFileName)
            End If

            'Now that we are back on the main thread, throw the exception we encountered if the user didn't cancel.
            If _exceptionEncounteredDuringFileTransfer IsNot Nothing Then
                If m_ProgressDialog Is Nothing OrElse Not m_ProgressDialog.UserCanceledTheDialog Then
                    Throw _exceptionEncounteredDuringFileTransfer
                End If
            End If

        End Sub
```
Then code like the following works, this is a test I wrote to cover this, but this also exists in applications including one I am maintaining. On bad password it prompts user to reenter and on Timeout it prompts for new download address.
```
        <WinFormsFact>
        Public Sub DownloadFile_UrlWithTimeOut_Throws()
            Dim testDirectory As String = CreateTempDirectory()
            Dim destinationFileName As String = GetDestinationFileName(testDirectory)
            Dim webListener As New WebListener(DownloadLargeFileSize)
            Dim listener As HttpListener = webListener.ProcessRequests()

            Try
                Dim ex As Exception = Assert.Throws(Of WebException)(  ' <======= this is an issue
                        Sub()
                            My.Computer.Network _
                                .DownloadFile(webListener.Address,
                                              destinationFileName,
                                              userName:="",
                                              password:="",
                                              showUI:=False,
                                              connectionTimeout:=1,
                                              overwrite:=True)

                        End Sub)

                Assert.Equal(SR.net_webstatus_Timeout, ex.Message) ' <======= this is an issue
                Assert.False(File.Exists(destinationFileName))
            Finally
                CleanUp(listener, testDirectory)
            End Try
        End Sub

```
Separate issue SR.net_webstatus_Timeout and SR.net_webstatus_Unauthorized are not accessible to the VB code in WinForms Repo.

elachlan:
@paul1956 I think because its internal here its not so much an issue. VB.net will be different.

lonitra:
@paul1956 Let's chat about VB scenario in your PR.

PR #11555, opened 2024-06-19
Fix missing types in System.Drawing.Common netframework facade

ericstj:
Fixes https://github.com/dotnet/runtime/issues/103702

## Proposed changes

- Add missing type forwards to System.Drawing.Common .NETFramework assembly.
 
## Customer Impact

- Compiling against netstandard2.0 and running on .NETFramework results in TypeLoadException for a few types in System.Drawing.Common.

## Regression? 

- *No* - I don't think this ever worked correctly.   Originally the types mentioned here were not exposed to .NETSTandard - only .NETFramework and .NETCore.  They were later pushed down into System.Drawing.Common but not exposed in netstandard2.0 because that wasn't being built.  Later on when the netstandard2.0 assembly was resurrected for building these types were added, but the forwards were never introduced.  Probably it went so long unnoticed because folks don't typically directly reference these types - instead they get them through TypeDescriptor.GetConverter() which would not be impacted by this bug.

## Risk

-  Very low, just adding type forwards to existing assembly.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11555)


ericstj:
cc @lonitra @ViktorHofer

lonitra (approved):

lonitra:
@ericstj does this need to be backported at all?

ViktorHofer:
Did you find out why APICompat doesn't flag this?

ViktorHofer (approved):

ericstj:
> Did you find out why APICompat doesn't flag this? EDIT: Oh lol, it did.

Yeah, I'm looking at more instances of CP0001 that are suppressed in runtime to see if anywhere else we're ignoring problems like this.

> @ericstj does this need to be backported at all?

It could be.  It's pretty low risk.  The bug exists in both 8.0 and 6.0.

Issue #8954, opened 2023-04-07
[API Proposal] System.Drawing.Drawing2D.GraphicsPathIterator.CopyData spanify


wstaelens:
### Background and motivation

When converting graphicsPaths I'm iterating over a lot of subpaths.

```
for (int i = ...
{
    // ...
    var subPathPoints = new GDIP.PointF[cnt];
    var subPathTypes = new byte[cnt];
    pathIterator.CopyData(ref subPathPoints, ref subPathTypes, startIndexSubPath, endIndexSubPath);
 } 
``` 

I'd like rent the subPathPoints and subPathTypes from an ArrayPool. Because the arraypool would give me larger arrays, I should need to use Spans and slice them to the actual size.

_original question:_ Can a `CopyData` overload using Spans be added?

### API Proposal

``` C#
namespace System.Drawing.Drawing2D;

public class GraphicsPathIterator
{
    // Existing API (note that the `ref` is completely unnecessary)
    public int Enumerate(ref PointF[] points, ref byte[] types);
    public int CopyData(ref PointF[] points, ref byte[] types, int startIndex, int endIndex);

    // Proposed
    public int Enumerate(Span<PointF> points, Span<byte> types);
    public int CopyData(Span<PointF> points, Span<byte> types, int startIndex, int endIndex);
}
```

### API Usage

```csharp
pathIterator.CopyData(pointsSpan, pathTypesSpan, startIndexSubPath, endIndexSubPath);
```

### Alternative Designs

None

### Risks

None

### Will this feature affect UI controls

No

Only impact is performanceboost and less allocations! 😄

elachlan:
@wstaelens You should format this as an API proposal. The team will explain a bit more about the process. I don't see any issues with adding span overloads.

https://github.com/dotnet/winforms/blob/main/.github/ISSUE_TEMPLATE/feature_request.yml

wstaelens:
Updated the template. Thanks for pointing out.

wstaelens:
@JeremyKuhne sounds fine for me!

elachlan:
@wstaelens I think GraphicsPath was spanified for .net9. If you could review it and let us know of any suggestions that would be great.

KrzysztofCwalina:
Should the source span of the copy method be read-only?

JeremyKuhne:
> Should the source span of the copy method be read-only?

Both of the spans are targets from the `GraphicsPath`. The existing API is really misleading with the `ref`.

[...]

PR #11594, opened 2024-06-27
Add span overloads to GraphicsPathIterator

JeremyKuhne:
Related to #8954.

This also adds more aggressive input validation to avoid buffer overruns.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11594)

### DIFF
--- a/src/System.Drawing.Common/src/System/Drawing/Drawing2D/GraphicsPathIterator.cs
+++ b/src/System.Drawing.Common/src/System/Drawing/Drawing2D/GraphicsPathIterator.cs
@@ -147,13 +147,28 @@ public sealed unsafe class GraphicsPathIterator : MarshalByRefObject, IDisposabl
         GC.KeepAlive(this);
     }
 
+    /// <inheritdoc cref="CopyData(ref PointF[], ref byte[], int, int)"/>
     public unsafe int Enumerate(ref PointF[] points, ref byte[] types)
+        => Enumerate(points.OrThrowIfNull().AsSpan(), types.OrThrowIfNull().AsSpan());
+
+    /// <inheritdoc cref="CopyData(ref PointF[], ref byte[], int, int)"/>
+#if NET9_0_OR_GREATER
+    public
+#else
+    private
+#endif
+    unsafe int Enumerate(Span<PointF> points, Span<byte> types)
     {
-        if (points.Length != types.Length)
+        if (points.Length != types.Length
+            || points.Length < Count)
+        {
             throw Status.InvalidParameter.GetException();
+        }
 
         if (points.Length == 0)
+        {
             return 0;
+        }
 
         fixed (PointF* p = points)
         fixed (byte* t = types)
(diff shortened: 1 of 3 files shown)


src/System.Drawing.Common/src/System/Drawing/Drawing2D/GraphicsPathIterator.cs
```diff
@@ -171,10 +186,36 @@ public unsafe int Enumerate(ref PointF[] points, ref byte[] types)
...
+    unsafe int CopyData(Span<PointF> points, Span<byte> types, int startIndex, int endIndex)
     {
-        if ((points.Length != types.Length) || (endIndex - startIndex + 1 > points.Length))
+        int count = endIndex - startIndex + 1;
+
+        if ((points.Length != types.Length)
+            || endIndex < 0
+            || startIndex < 0
+            || count > points.Length
+            || endIndex >= Count)
```
lonitra, line 215: Double checking, should this be `count` or `Count`?

src/System.Drawing.Common/tests/System/Drawing/Drawing2D/GraphicsPathIteratorTests.cs
```diff
@@ -320,14 +333,14 @@ public static IEnumerable<object[]> CopyData_StartEndIndexesOutOfRange_TestData(
...
-    public void CopyData_StartEndIndexesOutOfRange_ReturnsExpected(PointF[] points, byte[] types, int startIndex, int endIndex)
+    public void CopyData_WithData_StartEndIndexesOutOfRange_ThrowsArgumentException(PointF[] points, byte[] types, int startIndex, int endIndex)
     {
         PointF[] resultPoints = new PointF[points.Length];
         byte[] resultTypes = new byte[points.Length];
 
         using GraphicsPath gp = new(points, types);
         using GraphicsPathIterator gpi = new(gp);
-        Assert.Equal(0, gpi.CopyData(ref resultPoints, ref resultTypes, startIndex, endIndex));
+        AssertExtensions.Throws<ArgumentException>(null, () => gpi.CopyData(ref resultPoints, ref resultTypes, startIndex, endIndex));
```
lonitra, line 343: Do we need to update documentation since this is now throwing for these cases?
Also.. how are we getting exception for the last case where startIndex=2 -- is it from the PInvoke?

src/System.Drawing.Common/src/System/Drawing/Drawing2D/GraphicsPathIterator.cs
JeremyKuhne: Count of points in the actual path.

src/System.Drawing.Common/tests/System/Drawing/Drawing2D/GraphicsPathIteratorTests.cs
JeremyKuhne: It is from our own check. That said, I'm going to be more explicit in the checks so it throws if start is greater than end.

I don't think we need to update the docs as we already throw for some invalid inputs, we just weren't consistent.

lonitra (approved):

PR #11603, opened 2024-06-27
Namespace cleanup for System.Private.Windows.Core

JeremyKuhne:
We don't want to have "Windows.Forms" in types in this assembly so the code is cleaner for sharing (say with WPF).

This must not be squashed to retain history.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11603)


Tanya-Solyanik (approved):

lonitra (approved):

Zheng-Li01:
The System.Windows.Forms.BinaryFormat has been changed to System.Private.Windows.Core.BinaryFormat in the latest winforms main repro for the System.Private.Windows.Core namespace.
image

Issue #9416, opened 2023-07-03
Unable to resize `PictureBox` control at runtime in designer - DemoConsole


elachlan:
### .NET version

Main branch.

### Did it work in .NET Framework?

Not tested/verified

### Did it work in any of the earlier releases of .NET Core or .NET 5+?

Yes. We had a similar issue previously and fixed it in #8522

### Issue description

From #9411:

I cannot change the size of the picturebox control in the designer workspace, but can resize it through the properties window.

https://github.com/dotnet/winforms/assets/104609169/0a5c9320-1638-42d4-a390-54fe6794085f

_Originally posted by @Nora-Zhou01 in https://github.com/dotnet/winforms/issues/9411#issuecomment-1617719621_
            

### Steps to reproduce

Run DemoConsole and resize picture box.

elachlan:
@Nora-Zhou01, I've created the ticket here for the resize issue. It was working in #8522, so we can bisect from there.

CC: @Tanya-Solyanik

elachlan:
Additionally the delete functionality doesn't work properly. The button isn't removed.

elachlan:
I have bisected it to changes in #8353

elachlan:
![image](https://github.com/dotnet/winforms/assets/2433737/340e923c-6f87-4a4f-9f32-48c33cacf204)
I think the issue is in `CodeDomComponentSerializationService` because of its dependency on `BinaryFormatter`.

CC: @JeremyKuhne

elachlan:
If I add `<EnableUnsafeBinaryFormatterSerialization>true</EnableUnsafeBinaryFormatterSerialization>` to the project file it fixes it. But I don't think this is a long term solution.

JeremyKuhne:
There is some hope that we can tackle some part of this with the `BinaryFormat` stuff I wrote.

#6267

elachlan:
We need a follow up issue around removing the binary formatter and adding in the type converter to allow moving of picture box.

John-Qiao:
Verified this on latest .NET 8.0.100-preview.7.23362.29, issue was fixed in [DemoConsole](https://github.com/dotnet/winforms/tree/main/src/System.Windows.Forms/tests/IntegrationTests/DesignSurface/DemoConsole) project, we can change the size of the picturebox control in the designer workspace now.

![9416-testresult](https://github.com/dotnet/winforms/assets/45864985/23b1300f-e9df-4adc-96dc-ce697c618d4c)

[...]

Issue #9428, opened 2023-07-05
Look at limiting BinaryFormatter usage with CodeDomSerializationStore


JeremyKuhne:
The resource stream is a `HashTable` of `string` to `object`. While we support `HashTable` in `BinaryFormattedObject` related code, it's currently only for primitives and strings. We would need to look into what `object`s get added typically to fully understand this and characterize how useful the current `HashTable` support is and whether or not we could add coverage for other types to make it useful. (Say, an embedded `byte[]`.)

It is entirely possible that we may be stuck with `BinaryFormatter` for runtime design scenarios.

JeremyKuhne:
#9416 is one repro scenario.

elachlan:
This would be fantastic.

elachlan:
Why does serialization have to happen when resizing a control via the runtime designer, but not via the property window? In the `PictureBox` example, is it the image being resized causing the serialization?

PR #11614, opened 2024-06-30
Attempt to use `BinaryFormattedObject`/`BinaryFormatWriter` before `BinaryFormatter`

### LINKED ISSUE #11607: BinaryFormatter error when clicking "Add Tab" command for TabControl in DemoConsole application
### .NET version

.NET 9.0: Main branch of WinForms repo (.NET 9.0.0-preview.7.24326.3)

Yes

Yes, this is a regression issue. Not repro on .NET 9.0.0-preview.6.24305.2

### Issue description

In [DemoConsole](https://github.com/dotnet/winforms/tree/main/src/System.Windows.Forms/tests/IntegrationTests/DesignSurface/DemoConsole) test app of Winforms repo, BinaryFormatter error when clicking "Add Tab" command for TabControl.

image

https://github.com/dotnet/winforms/assets/86937911/f86da6ba-fee0-452b-bb45-ca4cdb663f3f

### Steps to reproduce

### More Info:
This issue also reproduces on ImageList when clicking the "Add" command.
image

elachlan:
Fixes #11607 
Related: #9428

Work was based on other usages in `System.Windows.Forms`.

#9416 is not fixed, because I was unable to use `WinFormsBinaryFormatWriter` as its internal to `System.Windows.Forms`. Which allows bitmap serializing.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11614)

### DIFF
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.CodeDomSerializationStore.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.CodeDomSerializationStore.cs
@@ -226,7 +226,7 @@ public sealed partial class CodeDomComponentSerializationService
                 delegator.Manager.Container = container;
             }
 
-            if (provider?.GetService(typeof(IDesignerSerializationManager)) is DesignerSerializationManager hostManager)
+            if (provider?.GetService<IDesignerSerializationManager>() is DesignerSerializationManager hostManager)
             {
                 foreach (IDesignerSerializationProvider serProvider in hostManager.SerializationProviders)
                 {
@@ -244,7 +244,7 @@ public sealed partial class CodeDomComponentSerializationService
             // Recreate resources
             if (_resourceStream is not null)
             {
-                long startPosition = _resourceStream.Position;
+                _resourceStream.Seek(0, SeekOrigin.Begin);
                 try
                 {
                     if (new BinaryFormattedObject(_resourceStream).TryGetPrimitiveHashtable(out Hashtable? value))
@@ -257,21 +257,22 @@ public sealed partial class CodeDomComponentSerializationService
                     // Couldn't parse for some reason, let the BinaryFormatter try to handle it.
                 }
 
-                // This check is to help in trimming scenarios with a trim warning on a call to BinaryFormatter.Deserialize(), which has a RequiresUnreferencedCode annotation.
-                // If the flag is false, the trimmer will not generate a warning, since BinaryFormatter.Deserialize() will not be called,
-                // If the flag is true, the trimmer will generate a warning for calling a method that has a RequiresUnreferencedCode annotation.
-                if (!EnableUnsafeBinaryFormatterInNativeObjectSerialization)
+                if (_resources is null)
                 {
-                    throw new NotSupportedException(SR.BinaryFormatterNotSupported);
-                }
+                    // This check is to help in trimming scenarios with a trim warning on a call to BinaryFormatter.Deserialize(), which has a RequiresUnreferencedCode annotation.
+                    // If the flag is false, the trimmer will not generate a warning, since BinaryFormatter.Deserialize() will not be called,
+                    // If the flag is true, the trimmer will generate a warning for calling a method that has a RequiresUnreferencedCode annotation.
+                    if (!EnableUnsafeBinaryFormatterInNativeObjectSerialization)
+                    {
+                        throw new NotSupportedException(SR.BinaryFormatterNotSupported);
+                    }
 
-                _resourceStream.Position = startPosition;
-
-                _resourceStream.Seek(0, SeekOrigin.Begin);
+                    _resourceStream.Seek(0, SeekOrigin.Begin);
 #pragma warning disable SYSLIB0011 // Type or member is obsolete
-                Hashtable? resources = new BinaryFormatter().Deserialize(_resourceStream) as Hashtable;
+                    Hashtable? resources = new BinaryFormatter().Deserialize(_resourceStream) as Hashtable;
 #pragma warning restore SYSLIB0011 // Type or member is obsolete
-                _resources = new LocalResourceManager(resources);
+                    _resources = new LocalResourceManager(resources);
+                }
             }
 
             using (delegator.Manager.CreateSession())
--- a/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.CodeDomSerializationStore.cs
+++ b/src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.CodeDomSerializationStore.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Collections;
+using System.Private.Windows.Core.BinaryFormat;
 using System.Reflection;
 using System.Runtime.Serialization;
 using System.Runtime.Serialization.Formatters.Binary;
@@ -48,6 +49,13 @@ public sealed partial class CodeDomComponentSerializationService
 
         // These fields are available after serialization or deserialization
         private ICollection? _errors;
+
+#pragma warning disable IDE0075 // Simplify conditional expression - the simpler expression is hard to read
+        internal static bool EnableUnsafeBinaryFormatterInNativeObjectSerialization { get; } =
+            AppContext.TryGetSwitch("System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization", out bool isEnabled)
+                ? isEnabled
+                : true;
+#pragma warning restore IDE0075
 
         /// <summary>
         ///  Creates a new store.
@@ -162,10 +170,23 @@ public sealed partial class CodeDomComponentSerializationService
                 if (_resourceStream is null)
                 {
                     _resourceStream = new MemoryStream();
+                    bool success = false;
 
+                    try
+                    {
+                        success = BinaryFormatWriter.TryWriteFrameworkObject(_resourceStream, _resources.Data);
+                    }
+                    catch (Exception ex) when (!ex.IsCriticalException())
+                    {
+                        Debug.Fail($"Failed to write with BinaryFormatWriter: {ex.Message}");
+                    }
+
+                    if (!success)
+                    {
 #pragma warning disable SYSLIB0011 // Type or member is obsolete
-                    new BinaryFormatter().Serialize(_resourceStream, _resources.Data);
+                        new BinaryFormatter().Serialize(_resourceStream, _resources.Data);
 #pragma warning restore SYSLIB0011 // Type or member is obsolete
+                    }
                 }
             }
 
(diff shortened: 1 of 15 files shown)


elachlan:
@JeremyKuhne I just did a quick look into this and it seems to have fixed #11607

It does not fix #9416, because I was unable to use `WinFormsBinaryFormatWriter` as its internal to `System.Windows.Forms`.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.CodeDomSerializationStore.cs
```diff
@@ -49,6 +50,13 @@ private sealed partial class CodeDomSerializationStore : SerializationStore, ISe
         // These fields are available after serialization or deserialization
         private ICollection? _errors;
 
+#pragma warning disable IDE0075 // Simplify conditional expression - the simpler expression is hard to read
+        internal static bool EnableUnsafeBinaryFormatterInNativeObjectSerialization { get; } =
```
Tanya-Solyanik, line 54: When the switch is cached in a static, it can't be turned on/off when tests are entering the scope.

elachlan:
@Tanya-Solyanik would using `InternalsVisibleTo` be inappropriate to resolve the issue of `WinFormsBinaryFormatWriter` not being accessible from `System.Windows.Forms.Design`? It feels a bit wrong to me.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.CodeDomSerializationStore.cs
elachlan: Its a static property, so I think the getter will check each time? The same pattern is used all over winforms.

elachlan:
I've played around with it. `InternalsVisibleTo` causes a bunch of warnings/errors when used to allow `Design` to access `Forms`.

Because of the internal APIs used by `TryGetImageListStreamer`/`WriteImageListStreamer`, we can't share the code between the two projects like we do with `Common.

I can't think of an easy fix for this that isn't complicated. Something like:
```csharp
namespace System.Windows.Forms.Primitives
{
    internal interface IBinaryFormattedObjectHandler
    {
        bool TryWriteObject(Stream stream, object value);
        bool TryGetObject(BinaryFormattedObject format, out object? value);
    }

    internal static class WinFormsBinaryFormatterRegistry
    {
        private static readonly List<IBinaryFormattedObjectHandler> _handlers = new();

        public static void RegisterHandler(IBinaryFormattedObjectHandler handler)
        {
            _handlers.Add(handler);
        }

        public static bool TryWriteObject(Stream stream, object value)
        {
            if (BinaryFormatWriter.TryWriteFrameworkObject(stream, value))
            {
                return true;
            }

            foreach (var handler in _handlers)
            {
                if (handler.TryWriteObject(stream, value))
                {
                    return true;
                }
            }
            return false;
        }

        public static bool TryGetObject(BinaryFormattedObject format, out object? value)
        {
			if (format.TryGetFrameworkObject(stream, value))
            {
                return true;
            }
			
            foreach (var handler in _handlers)
            {
                if (handler.TryGetObject(format, out value))
                {
                    return true;
                }
            }
            value = null;
            return false;
        }
    }
}
```

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.CodeDomSerializationStore.cs
```diff
@@ -162,10 +170,23 @@ public override void Close()
                 if (_resourceStream is null)
                 {
                     _resourceStream = new MemoryStream();
+                    bool success = false;
+
+                    try
+                    {
+                        success = BinaryFormatWriter.TryWriteFrameworkObject(_resourceStream, _resources.Data);
```
lonitra, line 177: Looks like we are always writing a Hashtable so perhaps change this to utilize `BinaryFormatWriter.TryWriteHashtable`

lonitra (commented):
Changes RLGTM. @JeremyKuhne is out this week, but I think we might want to consider pulling some of the writer and maybe even the reader code to System.Private.Windows.Core if we are starting to need these across different assemblies. We'll need to discuss this.

elachlan:
@lonitra I was able to refactor the `WinFormsBinaryFormatWriter` to move the Bitmap part into `PrimitivesBinaryFormatWriter`.

Which will allow each project to have access to `WriteBitmap`.

I had to add `FXAssembly`/`AssemblyRef` to the Primitives project, which has caused some compilation problems. So I might just define `AssemblyRef.SystemDrawing` in `PrimitivesBinaryFormatWriter` as a constant. The other option is to move all the `AssemblyRef`/`FXAssembly` to Primitives and remove them from `System.Windows.Forms`? But I am unsure if it is required to stay inside the project?

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.CodeDomSerializationStore.cs
elachlan: Done. #9416 fails here because `Bitmap` is not a primitive. 

CoreImageExtensions could add:
```csharp
internal static byte[] Save(this IImage image)
{
    using MemoryStream stream = new();
    image.Save(stream);
    return stream.ToArray();
}
```

Then we can save it to a `byte[]` which we can maybe handle somehow?

lonitra:
Oops, I realized we had already moved BinaryFormatWriter to System.Private.Window.Core ignore my previous comment about moving it.🤦‍♀️ There is a possibility we ship System.Private.Windows.Core in the future for code sharing so we do not want it to contain code related to items that are specific to WinForms. We could consider moving the Bitmap part to System.Private.Windows.Core since it is fairly common and not necessarily WinForms specific, but as you mentioned there would still be issues with `ImageListStreamer`. Your approach with `WinFormsBinaryFormatterRegistry` seems like the way to go, but ultimately I would want Jeremy to weigh in.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.CodeDomSerializationStore.cs
```diff
@@ -223,6 +244,29 @@ private void Deserialize(IServiceProvider? provider, IContainer? container, bool
             // Recreate resources
             if (_resourceStream is not null)
             {
+                long startPosition = _resourceStream.Position;
```
JeremyKuhne, line 247: We should seek to the beginning here, there is no need to save the position.

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.CodeDomSerializationStore.cs
```diff
@@ -223,6 +244,29 @@ private void Deserialize(IServiceProvider? provider, IContainer? container, bool
...
+                try
+                {
+                    if (new BinaryFormattedObject(_resourceStream).TryGetPrimitiveHashtable(out Hashtable? value))
+                    {
+                        _resources = new LocalResourceManager(value);
```
JeremyKuhne, line 252: You should return here if you were successful.

JeremyKuhne (changes requested):
Couple of comments

JeremyKuhne:
> I've played around with it. `InternalsVisibleTo` causes a bunch of warnings/errors when used to allow `Design` to access `Forms`.
> 
> Because of the internal APIs used by `TryGetImageListStreamer`/`WriteImageListStreamer`, we can't share the code between the two projects like we do with `Common.
> 
> I can't think of an easy fix for this that isn't complicated.

We should never give `WinForms.Design` internal visibility to `WinForms`.

The better inversion here is to do something like this in Core (which has the right visibility):

``` C#
internal delegate bool DeserializationDelegate(Stream stream, out object value);

internal static class ObjectSerializers
{
    private static readonly ConcurrentDictionary<Type, DeserializationDelegate> _deserializers = new();

    // This should be called in the assembly initializer for all known types
    internal static void AddDeserializer<T>(DeserializationDelegate deserializer)
    {
        if (!_deserializers.TryAdd(typeof(T), deserializer))
        {
        }
    }

    internal static bool TryDeserialize(Stream stream, out object value)
    {
        foreach ((_, var deserializer) in _deserializers)
        {
            if (deserializer(stream, out value))
            {
                return true;
            }
        }

        value = default!;
        return false;
    }

    internal static bool TrySerialize(Stream stream, object value)
    {
        if (value is IStreamSerializable serializable)
        {
             return serializable.TrySerialize(stream);
        }

        // Try Framework serializers here

        return false;
    }
}

internal interface IStreamSerializable
{
    bool TrySerialize(Stream stream);
}

// Upstream somewhere
public class Foo : IStreamSerializable
{
    bool IStreamSerializable.TrySerialize(Stream stream)
    {
        throw new NotImplementedException();
    }
}
``

src/System.Windows.Forms.Design/src/System/ComponentModel/Design/Serialization/CodeDomComponentSerializationService.CodeDomSerializationStore.cs
elachlan: What about the section below around `using (delegator.Manager.CreateSession())`?

elachlan:
@JeremyKuhne I don't think the `ObjectSerializers` structure you outlined helps us with the `CodeDomSerializationStore` Serialization, because we are serializing a `Hashtable` which contains objects. So the `BinaryFormatWriter.TryWriteHashtable` would have to be adjusted to handle `IStreamSerializable` values. 

Does that sound right?

JeremyKuhne:
> cause we are serializing a `Hashtable` which contains objects. S

It is really limited, so it doesn't buy much. It may actually be too difficult for users to figure out what works and doesn't so we might not want to do this.

elachlan:
@JeremyKuhne Have you got any suggestions on being able to handle `Bitmap` and `ImageListStreamer` in the hashtable? or doe we leave that to another PR?

JeremyKuhne:
> @JeremyKuhne Have you got any suggestions on being able to handle `Bitmap` and `ImageListStreamer` in the hashtable? or do we leave that to another PR?

It is a tricky question. While we could expand the manual handling, it gets a bit unwieldly. With a general deserializer we can handle that a little easier, but we don't have a good answer for serialization.

What other WinForms and Framework types can we see in this hashtable?

elachlan:
Technically any type that has the `[Serializable]` attribute can be added to it as far as I can tell. which is pretty broad. Its not super easy to follow in terms of seeing what resources are added to the hashtable.

elachlan:
For Winforms types `Tag` is the most problematic property I think, because its `object` and user input.

elachlan:
@JeremyKuhne should we implement the `ObjectSerializers` in a follow up PR? Or did you want me to add it here?

JeremyKuhne:
@elachlan my point is that I'm not sure we should do this as we can only handle a very limited set of things. It would be more confusing than helpful I think. If you're using the CodeDomSerializer you really should be taking the dependency on BinaryFormatter.

elachlan:
I think limited coverage for our winforms controls would be positive. Anything we can cover in a safer manner than the `BinaryFormatter` is a good thing.

Happy to not implement it and we can leave this PR as is.

PR #11651, opened 2024-07-10
Simplify some logic in Control.cs

JeremyKuhne:
Simplifies some logic in Control.cs.

- Pull in some methods with only a single private caller,
- Invert ifs to reduce nesting
- Use ternary expressions
- Use better variable names
- Use pattern matching

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/dotnet/winforms/pull/11651)

### DIFF
--- a/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
+++ b/src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
@@ -570,11 +570,11 @@ public unsafe partial class Control :
 
             if (Site is not null)
             {
-                ambientProperties = Site.GetService(typeof(AmbientProperties)) as AmbientProperties;
+                ambientProperties = Site.GetService<AmbientProperties>();
             }
             else
             {
-                ambientProperties = (AmbientProperties?)GetService(typeof(AmbientProperties));
+                ambientProperties = GetService<AmbientProperties>();
             }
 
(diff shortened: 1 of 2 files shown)


src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
```diff
@@ -605,32 +559,36 @@ public virtual bool AllowDrop
...
-                {
-                    Properties.SetObject(s_ambientPropertiesServiceProperty, props);
-                }
+            if (Site is not null)
+            {
+                ambientProperties = Site.GetService(typeof(AmbientProperties)) as AmbientProperties;
+            }
+            else
+            {
+                ambientProperties = (AmbientProperties?)GetService(typeof(AmbientProperties));
```
lonitra, line 577: I believe we have a `GetService<T>` overload which would be good to use here

lonitra (approved):
Just small comment but 👍

src/System.Windows.Forms/src/System/Windows/Forms/Control.cs
JeremyKuhne: Updated all of the GetService calls in Control

lonitra (approved):

